PR #12482, opened 2022-10-07
Adding ISO 8601 timestamp utility functions

lumberyard-employee-dm:
The ISO 8601 timestamp utility functions support the extended format which includes characters such as as dashes(`-`) and colons (`:`). So the timestamp would like the following "2022-10-10T12:00:00Z"

The ISO8601 basic format support has been added as well. A timestamp in that format would look as follows "20221010T120000Z"

Finally a ISO 8601 like format has been added to support a mix of the extended format for human readability in the Date portion and the basic format for filesystem compatibility with Windows in the Time portion in order to support timestamps that can be used as part of filenames. The Windows Kernel doesn't support paths with colons in them.

That format current looks like the following: "2022-10-10T120000Z" That format is up for debate and feedback to change it is welcome.

Only support for the Date and Time portion of the ISO8601 format is being added.
The ability to get
[Duration](https://en.wikipedia.org/wiki/ISO_8601#Durations) or [Week based](https://en.wikipedia.org/wiki/ISO_8601#Week_dates) timestamp support is not added at this time.

Switching to C++20 would alliviate the need for this functions since the chrono date library support stream operators for outputing ISO 8601 timestamps.

## How was this PR tested?

Added UnitTest to test all the ISO8601 timestamp functions


Code/Framework/AzCore/AzCore/Date/DateFormat.h
```diff
@@ -0,0 +1,456 @@
...
+
+            return true;
+    }
+
+    //! Time formats
+    // formats an extended format ISO 8601 timestamp time to seconds
+    // This format is not suitable for filenames as it contains the colon character, which isn't allowed
+    // for windows filesystems(NTFS)
+    // Returns true if utcTime is valid
+    // Use the basic format for filesystem safe timestamp
```
hultonha, line 118: question: Do we explicitly not what these comments to be picked up by doxygen? Otherwise we should likely update them all to be `//!` style (repeated across the other functions too)

Code/Framework/AzCore/AzCore/Date/DateFormat.h
```diff
@@ -0,0 +1,456 @@
...
+            "T%.2u:%.2u:%.2u.%.6uZ",
+            static_cast<unsigned>(currentTimeOfDay.hours().count()),
+            static_cast<unsigned>(currentTimeOfDay.minutes().count()),
+            static_cast<unsigned>(currentTimeOfDay.seconds().count()),
+            static_cast<unsigned>(AZStd::chrono::duration_cast<AZStd::chrono::microseconds>(currentTimeOfDay.subseconds()).count()));
+
+        return true;
+    }
+
+    // formats an basic format ISO 8601 timestamp time
```
hultonha, line 198: uber nit (repeated): 'formats an' -> 'formats a'

Code/Framework/AzCore/AzCore/Date/DateFormat.h
```diff
@@ -0,0 +1,456 @@
...
+    }
+
+    // The GetIso8601*Now versions queries the current utc time and formats it into a fixed string
+    bool GetIso8601ExtendedFormatNow(Iso8601TimestampString& utcTimestamp);
+    bool GetIso8601ExtendedFormatNowWithMilliseconds(Iso8601TimestampString& utcTimestamp);
+    bool GetIso8601ExtendedFormatNowWithMicroseconds(Iso8601TimestampString& utcTimestamp);
+
+    bool GetIso8601BasicFormatNow(Iso8601TimestampString& utcTimestamp);
+    bool GetIso8601BasicFormatNowWithMilliseconds(Iso8601TimestampString& utcTimestamp);
+    bool GetIso8601BasicFormatNowWithMicroseconds(Iso8601TimestampString& utcTimestamp);
```
hultonha, line 449: suggestion: Might it be useful to include a sample string for each of these functions similar to what you've done in the PR description to make it clear to the reader what output they'd expect to get? Could be helpful to know at a glance what the output would be.

Code/Framework/AzCore/Tests/Date/DateFormatTests.cpp
```diff
@@ -0,0 +1,365 @@
...
+        : public ScopedAllocatorSetupFixture
+    {
+    };
+
+    namespace Iso8601Timestamp
+    {
+        TEST_F(DateFormatTest, DateFormat_ExtendedFormatDate_Succeeds)
+        {
+            using namespace AZStd::chrono;
+            using namespace AZStd::chrono_literals;
```
hultonha, line 28: suggestion: Instead of `using namespace` here you could instead use `using AZStd::chrono::year_month_day;` and `using AZStd::chrono::operator""_y;` etc.. (a little more typing but avoids any weird name lookup rules with `using namespace`

Code/Framework/AzCore/AzCore/Date/DateFormat.h
hultonha: Or maybe better, a path to the tests, as these are guaranteed to be correct and could be useful for people to check to see the format in the tests themselves?

hultonha (approved):
LGTM! I just have a few minor suggestions/comments

Code/Framework/AzCore/Tests/Date/DateFormatTests.cpp
lumberyard-employee-dm: Well this `using namespace` is localized to the test function which is only using chrono function and not mixing with other libraries, so it should cause any issues.

Code/Framework/AzCore/AzCore/Date/DateFormat.h
lumberyard-employee-dm: Yeah I will add the `!`

Code/Framework/AzCore/Tests/Date/DateFormatTests.cpp
```diff
@@ -0,0 +1,365 @@
...
+
+            constexpr AZStd::chrono::year_month_day testDate{ 2025_y / March / 18_d };
+            constexpr AZStd::chrono::hh_mm_ss testTime{ AZStd::chrono::hours(11)
+            + AZStd::chrono::minutes(5)
+            + AZStd::chrono::seconds(37)
+            + AZStd::chrono::microseconds(71041) };
```
amzn-phist, line 34: nit: indent continuation lines (repeated)

Code/Framework/AzCore/AzCore/Date/DateFormat.h
```diff
@@ -0,0 +1,456 @@
...
+    //! Date Formats
+    // formats an extended format ISO 8601 timestamp date
+    // Returns true if utcTime is valid
+    template<class Duration>
+    bool GetIso8601ExtendedFormatDate(Iso8601TimestampString& utcTimestamp, const AZStd::chrono::utc_time<Duration>& utcTime)
+    {
+        using DurationType = typename AZStd::remove_cvref_t<decltype(utcTime)>::duration;
+
+        AZStd::chrono::year_month_day currentYmd;
+        [[maybe_unused]] AZStd::chrono::hh_mm_ss<DurationType> currentTimeOfDay;
```
amzn-phist, line 69: `currentTimeOfDay` appears to have similar "usage" pattern as `utcTime`, I don't know why one would need `[[maybe_unused]]` and not the other.  We can see that `currentTimeOfDay` is an "out" parameter and `utcTime` is a read-only "in" parameter, so it does seem correct that `currentTimeOfDay` is `[[maybe_unused]]` (maybe not read from), but does the compiler investigate that `currentTimeOfDay` is only written to in the internals of `GetYearMonthDayWithHoursMinutesSecondsFromUTCTime`?  I mean, if it doesn't look into that function, how else would it know whether `hoursMinutesSeconds` is not read and written?

amzn-phist (approved):

Code/Framework/AzCore/AzCore/Date/DateFormat.h
lumberyard-employee-dm: We could link to the ISO 8601 wikipedia page which shows the expected timestamp formats: https://en.wikipedia.org/wiki/ISO_8601#Calendar_dates

Code/Framework/AzCore/Tests/Date/DateFormatTests.cpp
hultonha: Unfortunately that isn't the case, `using namespace` will inject these names into the global scope and can lead to really weird problems to do with name look-up rules (see this article for more details - https://abseil.io/tips/153)

Code/Framework/AzCore/AzCore/Date/DateFormat.h
hultonha: Yeah that's a good compromise 👍

Code/Framework/AzCore/AzCore/Date/DateFormat.h
lumberyard-employee-dm: The clang compiler is really smart and detects that the memory address where currentTimeOfDay resides on the stack is only being written to, but not read.
It can track this because the `GetYearMonthDayWithHoursMinutesSecondsFromUTCTime` is inlined due to be a template function.

Code/Framework/AzCore/Tests/Date/DateFormatTests.cpp
lumberyard-employee-dm: These names are not getting injected into the global cope. Because the nearest namespace is `UnitTest::Iso8601Timestamp`

Code/Framework/AzCore/AzCore/Date/DateFormat.h
amzn-phist: Okay that makes sense if that function is inlined.  Thanks!

amzn-phist (approved):

PR #12483, opened 2022-10-07
Adding registration of a Core metrics event logger

lumberyard-employee-dm:
The ComponentApplication registers a Core metrics event logger that outputs to relative path inside the project user path `<project-root>/user` directory with the format of `Metrics/core_metrics.<target>.<timestamp>.json`

Only the simulation frame duration value is being recorded currently.


Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp
```diff
@@ -587,6 +535,133 @@ namespace AZ
...
+    {
+        // Register Core Event logger with Component Application
+
+        // Get the name of the running build target if available
+        AZ::IO::FixedMaxPath uniqueFilenameSuffix = AZ::Metrics::CoreMetricsFilenameStem;
+        if (AZ::IO::FixedMaxPathString buildTargetName;
+            m_settingsRegistry->Get(buildTargetName, AZ::SettingsRegistryMergeUtils::BuildTargetNameKey))
+        {
+            // append the build target name as injected from CMake if known
+            uniqueFilenameSuffix.Native() = AZ::IO::FixedMaxPathString::format(".%s", buildTargetName.c_str());
```
amzn-phist, line 634: This appears to set the extension rather than append (as the comment suggests).

Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp
```diff
@@ -587,6 +535,133 @@ namespace AZ
...
+        }
+        // Retrieve a filename safe timestamp with milliseconds precision
+
+        if (AZ::Date::Iso8601TimestampString timestampMilli;
+            AZ::Date::GetFilenameCompatibleFormatNowWithMilliseconds(timestampMilli))
+        {
+            uniqueFilenameSuffix.Native() = AZ::IO::FixedMaxPathString::format(".%s", timestampMilli.c_str());
```
amzn-phist, line 641: Same here, it appears to set the extension, overwriting anything that happened in the logic above it.  Is that intended?  Or do we want to append?

Code/Framework/AzCore/AzCore/Component/ComponentApplication.h
```diff
@@ -43,6 +43,10 @@ namespace AZ
 namespace AZ::Metrics
 {
     class IEventLoggerFactory;
+
+    enum class EventLoggerId : AZ::u32;
+
+    extern const EventLoggerId CoreEventLoggerId;;
```
amzn-phist, line 49: Extraneous `;`

Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp
lumberyard-employee-dm: Well the extension is always joining to be .`json`.
The other parts of the name between the the `<dot>` character isn't an extension.
For a file with the name of "core_metrics.editor.json", the term `editor` isn't part of the extension.

lumberyard-employee-dm: Yeah this should be `+=`

Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp
lumberyard-employee-dm: Yup an append is what is needed.

amzn-phist: Er, yeah "suffix" is what I meant.

amzn-phist (approved):

Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp
```diff
@@ -587,6 +537,172 @@ namespace AZ
...
+        m_projectNameChangedHandler = m_settingsRegistry->RegisterNotifier(ProjectNameChangedEventHandler{
+            *m_settingsRegistry });
+        m_commandLineUpdatedHandler = m_settingsRegistry->RegisterNotifier(UpdateCommandLineEventHandler{
+            *m_settingsRegistry, m_commandLine });
+
+        // Merge Command Line arguments
+        constexpr bool executeRegDumpCommands = false;
+
+#if defined(AZ_DEBUG_BUILD) || defined(AZ_PROFILE_BUILD)
+        // Skip over merging the User Registry in non-debug and profile configurations
```
spham-amzn, line 581: This comment doesn't seem to match `#if`

lumberyard-employee-dm: It does. The comment is indicating that the `~/.o3de/Registry` User Registry is skipped over in non-debug and non-profile configurations.

The "non" applies to both the "debug" and "profile" terms

Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp
```diff
@@ -587,6 +537,172 @@ namespace AZ
...
+        else
+        {
+            return;
+        }
+
+        // Append the build configuration(debug, release, profile) to the metrics filename
+        if (AZStd::string_view buildConfig(AZ_BUILD_CONFIGURATION_TYPE); !buildConfig.empty())
+        {
+            uniqueFilenameSuffix.Native() += AZ::IO::FixedMaxPathString::format(".%.*s", AZ_STRING_ARG(buildConfig));
+        }
```
spham-amzn, line 648: Is this even possible (AZ_BUILD_CONFIGURATION_TYPE being empty). 
If so, maybe just use a '#if' 
```suggestion
#if AZ_BUILD_CONFIGURATION_TYPE
    AZStd::string_view buildConfig(AZ_BUILD_CONFIGURATION_TYPE); 
    uniqueFilenameSuffix.Native() += AZ::IO::FixedMaxPathString::format(".%.*s", AZ_STRING_ARG(buildConfig));
#endif // AZ_BUILD_CONFIGURATION_TYPE
```

lumberyard-employee-dm: Well `AZ_BUILD_CONFNIGURATION_TYPE` cannot be empty, but even if it is empty, it defaults to a `AZStd::string_view()` which is an empty string_view

Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp
```diff
@@ -587,6 +537,172 @@ namespace AZ
...
+            (AZStd::chrono::steady_clock::time_point monotonicTime) mutable -> bool
+        {
+            // Retrieve the record rate setting from the Setting Registry
+            using namespace AZStd::chrono_literals;
+            AZStd::chrono::microseconds recordTickMicroseconds = 10s;
+            if (AZ::s64 recordRateMicrosecondValue;
+                registry.Get(recordRateMicrosecondValue, AZ::Metrics::CoreMetricsRecordRateMicrosecondsKey))
+            {
+                recordTickMicroseconds = AZStd::chrono::microseconds(recordRateMicrosecondValue);
+            }
```
spham-amzn, line 693: minor:
Is it possible that registry value changes during runtime? If not, maybe `recordTickMicroseconds` can be moved out of this lambda and just captured in.

Code/Framework/AzCore/AzCore/Metrics/IEventLogger.h
```diff
@@ -23,8 +24,47 @@ namespace AZ::IO
...
+    //! for an Event Logger are queried.
+    //! Currently  supports the following settings
+    //!
+    //! * "/O3DE/Metrics/<EventLoggerName>/Active" - If set to false, the event logger will not record new events
+    //!   If not set or true, the event logger will record events
+    struct SettingsKey_t
+    {
+        constexpr AZStd::fixed_string<128> operator()(AZStd::string_view name) const
+        {
+            AZStd::fixed_string<128> settingsKey(MetricsSettingsPrefix);
```
spham-amzn, line 41: an Assert check on the size of name could be helpful here so it doesn't overflow the fixed string

Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp
lumberyard-employee-dm: It is possible to update the record rate during run time via the SettingsRegistryInterface::Set function in C++ or through scripting.

Code/Framework/AzCore/AzCore/Metrics/IEventLogger.h
lumberyard-employee-dm: I can add that

lawsonamzn (approved):
Would be interested to know why we need to ignore whitespace anyway, for the tests considering we control the encoder, the writer, the serializer, the format, and the test data

spham-amzn (approved):

lumberyard-employee-dm:
> Would be interested to know why we need to ignore whitespace anyway, for the tests considering we control the encoder, the writer, the serializer, the format, and the test data

So that the test is reobust against to any whitespace that may occur in the encoder.

PR #12491, opened 2022-10-07
Fixed error deploying assets to android after the first deploy

moraaar:

## What does this PR do?

Second attempt to deploy assets to an android device was failing because `mkdir` command fails since the folder already exists. Added function to create a path on device that checks if it exists first.

## How was this PR tested?

Deployed assets to android device from scratch, made a change to one asset and deployed again. Both deploys work now.

### DIFF
--- a/cmake/Tools/Platform/Android/android_deployment.py
+++ b/cmake/Tools/Platform/Android/android_deployment.py
@@ -437,6 +437,15 @@ class AndroidDeployment(object):
         self.adb_call(arg_list=['install', '-t', '-r', str(self.apk_path.resolve())],
                       device_id=target_device)
 
+    def check_path_exists_on_device(self, path, target_device):
+        try:
+            result, output = self.adb_ls(path=path,
+                                        args=None,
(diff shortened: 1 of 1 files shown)


cmake/Tools/Platform/Android/android_deployment.py
```diff
@@ -437,6 +437,19 @@ def install_apk_to_device(self, target_device):
         self.adb_call(arg_list=['install', '-t', '-r', str(self.apk_path.resolve())],
                       device_id=target_device)
 
+    def path_exists_on_device(self, path, device_id):
+        try:
+            result, output = self.adb_ls(path=path,
```
amzn-phist, line 442: nit: If you don't use the `output` variable, use the `_` placeholder.
`result, _ = self.adb_ls(...`

moraaar: Good call! Done

lumberyard-employee-dm (approved):

PR #12499, opened 2022-10-07
Implement custom patching for FilterAdapter

alexmontAmazon:
## What does this PR do?
- Adds custom (at least twice as efficient) patching to FilterAdapter
- Adds filter to DPE-based AssetEditor
- Small fix for row handling in DocumentPropertyEditor
- Small fix for patch debugging in DomUtils

## How was this PR tested?
Locally, both with the DPE standalone, and with Editor


Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/FilterAdapter.h
```diff
@@ -21,10 +21,13 @@ namespace AZ::DocumentPropertyEditor
         ~RowFilterAdapter();
 
         void SetSourceAdapter(DocumentAdapterPtr sourceAdapter);
+        void SetIncludeAllMatchDescendants(bool includeAll);
 
     protected:
         struct MatchInfoNode
         {
+            MatchInfoNode(const MatchInfoNode& nodeToCopy) = delete;
```
lumberyard-employee-dm, line 29: This seems unnecessary given that MatchInfoNode is already a protected member of the RowFilterAdapter, so it is not as if outside code would attempt to copy it.

Also if copy constructor is to be deleted, then the copy assignment operator should also be as well

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/FilterAdapter.cpp
```diff
@@ -368,15 +360,144 @@ namespace AZ::DocumentPropertyEditor
...
+                    else
+                    {
+                        valueIter = rowValue.ArrayErase(valueIter);
+                        continue; // we've already moved valueIter forward, skip the ++valueIter below
+                    }
+                }
+                ++valueIter;
+            }
+        }
+    }
```
lumberyard-employee-dm, line 404: This function doesn't touch any member variables.
It can be marked as static.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/FilterAdapter.cpp
```diff
@@ -368,15 +360,144 @@ namespace AZ::DocumentPropertyEditor
...
+
+    void RowFilterAdapter::GeneratePatch(PatchType patchType)
+    {
+        Dom::Patch patch;
+
+        AZStd::function<void(MatchInfoNode*, Dom::Path, const Dom::Value&)> generatePatchOperations =
+            [&](MatchInfoNode* matchNode, Dom::Path mappedPath, const Dom::Value& sourceValue)
```
lumberyard-employee-dm, line 411: No need to store this in an AZStd::function prevents the `operator()` call on the lambda from being inlined.
`auto` works in this case because the lambda `operator()` can be invoked directly.
```suggestion
        auto generatePatchOperations =
            [&](MatchInfoNode* matchNode, Dom::Path mappedPath, const Dom::Value& sourceValue)
```

Code/Framework/AzToolsFramework/AzToolsFramework/AssetEditor/AssetEditorWidget.cpp
```diff
@@ -210,8 +211,9 @@ namespace AzToolsFramework
             else
             {
                 m_adapter = AZStd::make_shared<AZ::DocumentPropertyEditor::ReflectionAdapter>();
-                m_dpe = new DocumentPropertyEditor(this);
-                propertyEditor = m_dpe;
+                m_filteredWidget = new FilteredDPE(this);
+                m_dpe = m_filteredWidget->GetDPE();
+                propertyEditor = m_filteredWidget;
```
amzn-tmryan, line 216: Is there a reason why the FilteredDPE can't just be a DPE?

amzn-tmryan (approved):
Looks good to me. Approved under the assumption that Derric's feedback will be addressed.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/FilterAdapter.cpp
alexmontAmazon: I tried that, but since it can call itself recursively on line 439, auto isn't working for me. It gives me the error:
`error C3536: 'generatePatchOperations': cannot be used before it is initialized`
Is there another way around that rather than storing it in a named function?

Code/Framework/AzToolsFramework/AzToolsFramework/AssetEditor/AssetEditorWidget.cpp
alexmontAmazon: do you mean, why don't I derive FilteredDPE from DocumentPropertyEditor? DocumentPropertyEditor is a QScrollArea (so it can scroll its contents if necessary, while FilteredDPE has to hold a DPE but keep its search box outside of the scroll area, so it doesn't scroll. That's why it's a "hasA" rather than an "isA"

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/FilterAdapter.cpp
alexmontAmazon: good call, fixed!

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/FilterAdapter.cpp
lumberyard-employee-dm: Well you could make a struct with an `operator()` which would allow to avoid the auto?

alexmontAmazon: This is quite an interesting topic, so I did a deep dive on recursive lambdas and found this:
https://artificial-mind.net/blog/2020/09/12/recursive-lambdas

I followed this which worked as long as I gave the lambda a trailing return type, which I did. See if you like it now...

lumberyard-employee-dm: That works as well

lumberyard-employee-dm (approved):

dtamkin1 (approved):

cgalvan (approved):

PR #12500, opened 2022-10-07
reenables EditorTestTesting module

evanchia-ly-sdets:
Signed-off-by: evanchia-ly-sdets <evanchia@amazon.com>

## What does this PR do?
This re-enables the EditorTestTesting module and updates one of the pytest legacy methods.

## How was this PR tested?
Tested by locally running 1000 attempts without reproduction. Further investigated original bug logs and found errors not related to the test that caused the failures.


Kadino (approved):
Looks good! Definitely something to be aware of when PyTest and/or Python gets a version change.

cgalvan (approved):

AMZN-scspaldi (approved):

PR #12501, opened 2022-10-07
Prism/create gem implement tags field

tkothadev:
## What does this PR do?
This PR implements an updated Tag field for the Create a Gem workflow. This mimics existing filter tag fields in the O3DE editor and allows the user to click on tags to remove them from the list. It's written in such a way that the tag field can extend beyond gems in the future.

This is what the field looked like before the changes:
CreateGemBeforeUserTagChanges

And this is what it looks like after the changes:
CreateGemUserTagChanges

I chose the two row approach for the tag field because it lined up with the existing features we have in the O3DE editor. For example, in the `FilterSearchWidget` for the editor:
image (10)

I also found that in general, implementation went way smoother by doing this split up. Trying to inline the tags led to numerous layout glitches inside the Qt framework that were hard to discern. There's also an edge case @AMZN-daimini pointed out where a single line cannot handle an overflow of tags. Dedicating a separate row to that allows us to gracefully deal with that issue. This also allows the tags to be adjustable via qss without much fuss.

Primary code changes are the introduction of a new class `FormLineEditTagsWidget` which inherits directly from `FormLineEditWidget`. It only contains the extra functionality required to handle the tag rendering and auto-completion, including the drop down button to show the available common tags.

In the line edit field, tags are space separated, so pressing either space or enter will create a new tag.

## How was this PR tested?

Testing was performed manually. The gif showing the changes demonstrates the testing I have done regarding the UI interactions. I also verified that upon gem creation it correctly exports the desired gem tags in a gem json file.


tkothadev:
Some additional details regarding the tag on the second row:

It was found during testing that having the tag line on the same row as the line edit caused fighting for real estate, making it uncomfortable visually when attempting text entry with a certain number of tags (around 8-10). Thus adding the second row allows us to freely type on the line edit field without interference.

Currently there's only one row for tags. Overflow is dealt with by the `QHBoxLayout` scrunching all tags together if they get too crammed into the line.
image (12)

Dealing with overflow via multiple rows of tags is out of scope for the current PR, and will be backlogged for now. The main justification is that in practice (seeing the existing gems in the gem catalog), most gems have around 4-5 tags, each with text that is not too long.

Code/Tools/ProjectManager/Source/FormLineEditTagsWidget.h
```diff
@@ -0,0 +1,72 @@
...
+#include <QWidget>
+#include <FormLineEditTagsWidget.h>
+#endif
+
+QT_FORWARD_DECLARE_CLASS(QCompleter)
+QT_FORWARD_DECLARE_CLASS(QPushButton)
+QT_FORWARD_DECLARE_CLASS(QLineEdit)
+QT_FORWARD_DECLARE_CLASS(QLabel)
+QT_FORWARD_DECLARE_CLASS(QFrame)
+QT_FORWARD_DECLARE_CLASS(QKeyEvent)
```
pollend, line 21: I think its more consistent to just forward it using `class X`

Code/Tools/ProjectManager/Source/FormLineEditTagsWidget.cpp
```diff
@@ -0,0 +1,201 @@
...
+#include <QVBoxLayout>
+#include <QHBoxLayout>
+#include <QCheckBox>
+#include <QLineEdit>
+#include <QLabel>
+#include <QMovie>
+#include <QFrame>
+#include <QValidator>
+#include <QStyle>
+#include <QKeyEvent>
```
pollend, line 27: can we split qt and AZ components and also sort these alphabetically.

Code/Tools/ProjectManager/Source/FormLineEditTagsWidget.cpp
```diff
@@ -0,0 +1,201 @@
...
+        if(!found)
+        {
+            return;
+        }
+        //remove the offending string
+        m_tags.removeOne(string);
+
+        //now reconstruct the tag list
+        refreshTagFrame();
+    }
```
AMZN-daimini, line 191: Since the individual QCheckbox that was clicked is sending the signal, you should be able to just use `sender()` to get a pointer to it and avoid the loop to search?

Code/Tools/ProjectManager/Resources/ProjectManager.qss
```diff
@@ -161,6 +165,31 @@ QTabBar::tab:focus {
...
+    spacing:0px;
+    color:#94D2FF;
+}
+
+#formTagField QCheckBox::indicator{
+    background-color: transparent;
+    image:url(:/X.svg);
+    border:0px;
+    width: 10%;
+    padding-left:px;
```
AMZN-daimini, line 190: nit: remove empty property

AMZN-daimini (approved):

Code/Tools/ProjectManager/Source/FormLineEditTagsWidget.cpp
```diff
@@ -0,0 +1,201 @@
...
+#include <QValidator>
+#include <QStyle>
+#include <QKeyEvent>
+
+namespace O3DE::ProjectManager
+{
+
+    void FormLineEditTagsWidget::setupCompletionTags()
+    {
+        m_completionTags << "Audio" << "Animation" << "Physics" << "GameDevelopment" << "IK" << "Blender" << "Rendering" << "Terrain";
```
AMZN-Phil, line 34: What was the reasoning behind these choices? IK seems very specific when tags that are used a lot (Framework, Scripting,  Tools, Utility) are not in this list.

Code/Tools/ProjectManager/Source/FormLineEditTagsWidget.cpp
```diff
@@ -0,0 +1,201 @@
...
+        m_completer->setObjectName("formCompleter");
+        m_completer->setCaseSensitivity(Qt::CaseInsensitive);
+        m_completer->setCompletionMode(QCompleter::UnfilteredPopupCompletion);
+        m_completer->popup()->setVerticalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
+        m_completer->popup()->setFixedWidth(m_lineEdit->rect().width()*3);
+        m_completer->popup()->setMouseTracking(true);
+        m_completer->popup()->setObjectName("formCompleterPopup");
+
+
+        /* NOTE(tkothadev): Manually setting the stylesheet of this popup widget is not desired.
```
AMZN-Phil, line 63: We usually avoid usernames in comments.

Code/Tools/ProjectManager/Source/FormLineEditTagsWidget.cpp
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    void FormLineEditTagsWidget::textChanged(const QString& text)
+    {
+        if(text.contains(" "))
+        {
+            //do tag adding
+            QString interim = m_lineEdit->text();
+            m_lineEdit->clear();
+            for(auto str : interim.split(" "))
```
AMZN-Phil, line 159: Maybe tagStr or similar to make it more obvious what the string is holding.

Code/Tools/ProjectManager/Source/FormLineEditTagsWidget.cpp
```diff
@@ -0,0 +1,201 @@
...
+            refreshTagFrame();
+        }
+    } 
+
+    void FormLineEditTagsWidget::processTagDelete([[maybe_unused]] int unused)
+    {
+        auto checkBoxes = m_tagFrame->findChildren<QCheckBox* >();
+        QString string;
+        bool found = false;
+        for(auto c : checkBoxes)
```
AMZN-Phil, line 172: checkbox instead of c

Code/Tools/ProjectManager/Source/FormLineEditTagsWidget.cpp
lumberyard-employee-dm: I find it strange that is hardcoded and not data driven from a file.

Code/Tools/ProjectManager/Source/FormLineEditTagsWidget.cpp
```diff
@@ -0,0 +1,201 @@
...
+            //do tag adding
+            QString interim = m_lineEdit->text();
+            m_lineEdit->clear();
+            for(auto str : interim.split(" "))
+            {
+                addToTagList(str);
+            }
+            refreshTagFrame();
+        }
+    }
```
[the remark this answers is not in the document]
tkothadev: So when the interim string is split by space, there are 3 distinct cases that the text line will detect ( _ as space):

1.  _word

2.  word_

3.  wo_rd

In the first case, the validator regex actually prevents a space from ever preceding before the word. The line edit must always start with a non-whitespace character.

In the second cases the "word" is picked up as an entire string tag, while the latter part is basically an empty string, which gets filtered out in `addToTagList`. 

In the third case "wo" and "rd" are two separate strings, and rendered as tags separately. This was verified in testing the tag field.

All cases get the intended text pieces as tags.

Code/Tools/ProjectManager/Source/FormLineEditTagsWidget.cpp
tkothadev: Making it data driven was intended to be the next step after implementing the tags field. I went with a preliminary set of tags primarily to test the auto-completion features. Most of the effort for this PR went into fixing the interactions and styling of the completer.

As for tag choice, these were the tags I was most aware of at the time, but I can add more tags in. 

Regarding the data file, should I add a text file in the existing resources folder (i.e. where we keep the qss stylesheets)? I could call it something like `completionTags.txt`

Code/Tools/ProjectManager/Source/CreateAGemScreen.cpp
```diff
@@ -269,8 +269,8 @@ namespace O3DE::ProjectManager
         m_licenseURL = new FormLineEditWidget(tr("License URL"), "", tr("Link to the license web site i.e. https://opensource.org/licenses/Apache-2.0"), "");
         gemDetailsLayout->addWidget(m_licenseURL);
 
-        m_userDefinedGemTags = new FormLineEditWidget(tr("User-defined Gem Tags (Comma separated list)"), "");
-        m_userDefinedGemTags->lineEdit()->setValidator(new QRegularExpressionValidator(QRegularExpression("(\\w+)(,\\s?\\w*)*"), this));
+        m_userDefinedGemTags = new FormLineEditTagsWidget(tr("User-defined Gem Tags (Press Space or Return with non-empty text to create a new tag)"), "");
```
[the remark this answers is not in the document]
tkothadev: Done (as tooltip).
image

Code/Tools/ProjectManager/Source/FormLineEditTagsWidget.cpp
tkothadev: Next commit will have this part updated to a data driven format. I opted to use tags commonly recognized in the existing gem catalog.

Code/Tools/ProjectManager/Source/FormLineEditTagsWidget.cpp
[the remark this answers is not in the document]
AMZN-daimini: I was watching the gif in this PR, and it shows "AtomRendering" being written, then a space being added in the middle, and only "Atom" becoming a tag. Rewatching, I realized "Rendering" was already a tag, so I guess that's why it wasn't added again - and what caused my confusion. 

I wonder if we could clarify the situation when users try to add an already existing tag... but that's more of a follow-up question than anything.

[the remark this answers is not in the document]
tkothadev: Agreed, that is a follow up item, because primary correctness is not affected in this PR. Most likely I would need to add an additional 'information frame' that would allow me to write messages such as :
"Note: Rendering tag already exists"

And then it fades after 3 seconds. I would most likely add it to the bottom of the tags.

Code/Tools/ProjectManager/Source/FormLineEditTagsWidget.cpp
```diff
@@ -0,0 +1,196 @@
...
+        m_completer->popup()->setVerticalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
+        m_completer->popup()->setFixedWidth(m_lineEdit->rect().width()*3);
+        m_completer->popup()->setMouseTracking(true);
+        m_completer->popup()->setObjectName("formCompleterPopup");
+
+        /* Manually setting the stylesheet of this popup widget is not desired.
+         * However, the current styling rules of the Project Manager make the background color blend in
+         * with surroundings, making it difficult to see. Currently, attempts at rectifying this in the main
+         * stylesheet proved very difficult, so a stop-gap measure of hard-coding the stylesheet was used.
+         * I have discussed this with AMZN-alexpete.
```
AMZN-Phil, line 76: Missed another username here

Code/Tools/ProjectManager/Resources/ProjectManagerCompletionTags.txt
```diff
@@ -0,0 +1,46 @@
+AWS
```
AMZN-Phil, line 1: I'm not sure if we should be suggesting tags specific to any one company or product out of the box, same goes for some of the other tags (Barrier, Twitch, Synergy).

Code/Tools/ProjectManager/Resources/ProjectManagerCompletionTags.txt
```diff
@@ -0,0 +1,46 @@
...
+Profiler
+RemoteTools
+Rendering
+SDK
+Sample
+ScriptAutomation
+Scripting
+Simulation
+Slices
+Stars
```
AMZN-Phil, line 38: Should check with UX on this list whether we want a large suggestion list or not, a couple of entries are quite specific that are not going to be used that often.

Code/Tools/ProjectManager/Resources/ProjectManagerCompletionTags.txt
tkothadev: I'll remove company and product specific tags. Waiting for response from UX before I address shrinking the tag list any further.

Code/Tools/ProjectManager/Source/FormLineEditTagsWidget.cpp
```diff
@@ -0,0 +1,195 @@
...
+#include <QStyle>
+#include <QValidator>
+#include <QVBoxLayout>
+
+namespace O3DE::ProjectManager
+{
+
+    void FormLineEditTagsWidget::setupCompletionTags()
+    {
+        QFile completionTagFile(":/ProjectManager/text/ProjectManagerCompletionTags.txt");
```
AMZN-alexpete, line 36: We need to make sure this file ends up in the installer build (SDK output) or is optional - have you tested this where the file is absent?  I'm OK with this change going in a separate PR, but just wanted to make sure it won't break for the installer version.

Code/Tools/ProjectManager/Source/FormLineEditTagsWidget.h
```diff
@@ -0,0 +1,72 @@
...
+{
+    class StyledLineEdit;
+}
+
+namespace O3DE::ProjectManager
+{
+    class FormLineEditTagsWidget
+        : public FormLineEditWidget 
+    {
+
```
AMZN-alexpete, line 33: I think you need the Q_OBJECT macro on this class
> The Q_OBJECT macro must appear in the private section of a class definition that declares its own signals and slots or that uses other services provided by Qt's meta-object system.
```suggestion
     Q_OBJECT
```

AMZN-alexpete (changes requested):
Just a couple of items I think we should address - the Q_OBJECT macro and testing with that tags file missing.  May want to look into what needs to be done to make sure it ends up in the SDK too.

While it looks like the file should be included in the binary, I recommend having some kind of error checking.

Code/Tools/ProjectManager/Source/FormLineEditTagsWidget.cpp
lumberyard-employee-dm: If you want to add this to the SDK layout, you can us the [ly_install_files](https://github.com/o3de/o3de/blob/development/cmake/Install.cmake#L116-L125) or [ly_install_directory](https://github.com/o3de/o3de/blob/development/cmake/Install.cmake#L37-L52) command to copy this file to the SDK layout.

You can then build the `install` target(called INSTALL) in visual studio to create an SDK layout and validate that the file is there

AMZN-alexpete: I missed that this file is part of the .qrc so it should be included automatically with the other Qt resource files.

Code/Tools/ProjectManager/Source/FormLineEditTagsWidget.h
tkothadev: Per AMZN-alexpete's recent comments, I think the tags file portion is resolved, so I'll focus on the Q_OOBJECT macro here along with doxygen comment.

AMZN-alexpete (approved):
Nice!

tkothadev:
Per my sync with @amzn-leenguy for UX, I implemented the following changes:
* Add explanation text (italicized) next to the field title stating: Press enter to create Gem Tag
* Add hint text in the line edit stating: Tags without spaces: i.e. SampleGame
* Disable ability to type spaces in line edit (only “enter” creates the new tag)

These are what the changes look like:
image (13)


Follow up items (after PR is merged):
* When clicking on an auto-completion suggestion, it should create a tag
* Create a ticket for a future task to see how we can contribute back to the QT library; How All the bespoke custom UI Components in Prism (such as the Oversized input field, dropdown and so on) can be added to the BJDS QT Control gallery (located at C:\O3DE\22.10.0\bin\Windows\profile\Default\O3DEQtControlGallery.exe)

AMZN-alexpete (approved):

lumberyard-employee-dm (approved):

amzn-leenguy (approved):
Approved from UX POV

PR #12505, opened 2022-10-07
[Terrain] Create paint brush blend modes

mbalfour-amzn:
## What does this PR do?

This introduces a set of blend modes to the Image Gradient Paintbrush. In the process, I also followed a suggestion from @invertednormal to use an intermediate tile cache while painting. This fixes the painting logic for each brush stroke to match how paint programs work, where each pixel only gets modified once in a brush stroke, instead of accumulating changes each time the brush touches a pixel on the same brush stroke. It also significantly optimizes the undo/redo buffer itself, both in creation time and in memory usage.

Before:
Notice how each single brush stroke with low opacity has inconsistent paint values based on how fast the paint brush was moving.
https://user-images.githubusercontent.com/82224783/194671483-b7113be2-879c-4fe2-8899-097b8f9849b4.mp4

After:
Each brush stroke now has consistent paint values, and there are also a set of blend modes available now too.
https://user-images.githubusercontent.com/82224783/194673280-128d4c83-4f20-42e6-a52f-9ea5572df3c9.mp4

## How was this PR tested?

Manually tested the paintbrush with different image gradients, blend modes, and sizes.


Code/Framework/AzToolsFramework/AzToolsFramework/PaintBrushSettings/PaintBrushSettings.cpp
```diff
@@ -46,7 +47,19 @@ namespace AzToolsFramework
...
+                    ->EnumAttribute(PaintBrushBlendMode::Normal, "Normal")
+                    ->EnumAttribute(PaintBrushBlendMode::Multiply, "Multiply")
+                    ->EnumAttribute(PaintBrushBlendMode::Screen, "Screen")
+                    ->EnumAttribute(PaintBrushBlendMode::Add, "Linear Dodge (Add)")
+                    ->EnumAttribute(PaintBrushBlendMode::Subtract, "Subtract")
+                    ->EnumAttribute(PaintBrushBlendMode::Darken, "Darken (Min)")
+                    ->EnumAttribute(PaintBrushBlendMode::Lighten, "Lighten (Max)")
+                    ->EnumAttribute(PaintBrushBlendMode::Average, "Average")
+                    ->EnumAttribute(PaintBrushBlendMode::Overlay, "Overlay")
+                    ;
```
gadams3, line 62: If you switch to AZ_ENUM_CLASS_WITH_UNDERLYING_TYPE then the EnumAttribute calls can be reduced to:
->Attribute(AZ::Edit::Attributes::EnumValues, AZ::Edit::GetEnumConstantsFromTraits<PaintBrushBlendMode>())

It would use the exact enum name so you would have to rename them to whatever you want the user to see in the drop down.

mbalfour-amzn: The problem is that the names that we want the user to see contain parentheses, so I don't believe GetEnumConstantsFromTraits() is a workable approach here.

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
```diff
@@ -209,21 +227,91 @@ namespace AzToolsFramework
...
+                        opacities.emplace_back(opacity);
                     }
                 }
             });
 
+            // Set the blending operation based on the current paintbrush blend mode setting.
+            PaintBrushNotifications::BlendFn blendFn;
+            switch (blendMode)
+            {
+                case PaintBrushBlendMode::Normal:
```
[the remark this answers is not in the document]
mbalfour-amzn: This is one of the few times I'm deliberately _not_ making a more extensible approach. :) The design behind the paint brush tool is that it contains all of the paint brush logic possible. Other systems shouldn't need to extend with custom blend types. There's a finite set of blend modes defined in paint programs.

That said, once we have a second tangible use case for the paint brush, it's possible that we'll find a reason to add custom types, and it could extracted at that point. It just feels like overkill to add that support right now, when the intent is _not_ to have a need for it.

As for the blend functions being the same as mixed gradient, I'm definitely open to suggestions here. I stared at it a bunch but didn't see a great way to share them. The blend math itself is the same, but the lerping/clamping is a bit different between the two use cases. I just wasn't seeing enough logic to pull out into shared code, especially if we also keep the enums separate between the two.

Code/Framework/AzToolsFramework/AzToolsFramework/PaintBrushSettings/PaintBrushSettings.cpp
gadams3: I understand. My suggestion implied removing the more descriptive naming or renaming the enums to include that wording. It could possibly replace parentheses with "_", if that's even necessary. The details for the more artist friendly names could also be moved to the property description tooltip or documentation. In any case, all of the comments were about sharing and simplifying code. Nothing blocking.

gadams3 (approved):

cgalvan (approved):

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponentMode.cpp
```diff
@@ -27,9 +27,198 @@
...
+        //! Size of each modified image tile that we'll cache off.
+        //! This size is chosen somewhat arbitrarily to keep the number of tiles balanced at a reasonable size.
+        static inline constexpr uint32_t ImageTileSize = 32;
+
+        //! Keeps track of all the unmodified and modified gradient values for an NxN tile.
+        //! We store it a pair of arrays instead of an array of pairs for better compatibility with the ImageGradient APIs,
+        //! where we can just pass in a full array of values to update a full tile of values at once.
+        using ImageTile = AZStd::pair<
+                AZStd::array<float, ImageTileSize * ImageTileSize>,         // Unmodified data
+                AZStd::array<float, ImageTileSize * ImageTileSize>>;        // Modified data
```
invertednormal, line 200: nit: I think it'd be a little cleaner to just have a simple struct for this. Having to reference `first` / `second` isn't as straightforward as members named something like `m_originalData` / `m_modifiedData`;

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponentMode.cpp
```diff
@@ -216,99 +373,134 @@ namespace GradientSignal
...
+        // Hand over ownership of the paint stroke buffer to the undo/redo buffer.
+        m_paintBrushUndoBuffer->SetUndoBufferAndDirtyArea(AZStd::move(m_paintStrokeBuffer), m_paintStrokeDirtyRegion);
+
         EndUndoBatch();
+
+        // Make sure we've cleared out our paint stroke data until the next paint stroke begins.
+        m_paintStrokeBuffer.reset();
+        m_metersPerPixelX = 0.0f;
+        m_metersPerPixelY = 0.0f;
+        m_paintStrokeDirtyRegion = AZ::Aabb::CreateNull();
```
invertednormal, line 424: nit : might make sense to put these in their own separate struct that you could easily reset with `= {};`

invertednormal (approved):
Looks good, just a few nits, but definitely nothing blocking.

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponentMode.cpp
```diff
@@ -27,9 +27,198 @@
...
+                        m_paintedImageTiles[tileIndex]->first);
+                }
+                else
+                {
+                    ImageGradientModificationBus::Event(
+                        m_imageGradientEntityId,
+                        &ImageGradientModificationBus::Events::SetPixelValuesByPixelIndex,
+                        pixelIndices,
+                        m_paintedImageTiles[tileIndex]->second);
+                }
```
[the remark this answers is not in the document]
mbalfour-amzn: i'll condense this in my next PR to:
```
                ImageGradientModificationBus::Event(
                    m_imageGradientEntityId,
                    &ImageGradientModificationBus::Events::SetPixelValuesByPixelIndex,
                    pixelIndices,
                    undo ? m_paintedImageTiles[tileIndex]->m_unmodifiedData : m_paintedImageTiles[tileIndex]->m_modifiedData);
```

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponentMode.cpp
mbalfour-amzn: i'll replace with a struct in my next PR.

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponentMode.cpp
mbalfour-amzn: i'll structify it in my next PR.

Issue #12013, opened 2022-09-15
Creating a prefab requires the user to save the level twice


forkercat:
**Describe the bug**

Creating a prefab requires the user to save the level twice

**Steps to reproduce**
1. Open any level
2. Create an entity and select it
3. Create a prefab for the entity
4. Hit Ctrl+S or file->save. Notice that the level is still marked dirty
5. Hit Ctrl+S again. Notice that it is saved now.

**Expected behavior**
Level is not marked dirty after hitting ctrl+s for the first time.

**Actual behavior**
Level is still marked dirty after hitting ctrl+s for the first time.

**Found in branch**
stabilization/2210 and development

ParasiteSt:
A second case of this is when you already have a prefab created and alter the prefab. Once altered if you save the level it will save the prefab's changes, but not the level, a second save is needed for the level here as well.

forkercat:
> A second case of this is when you already have a prefab created and alter the prefab. Once altered if you save the level it will save the prefab's changes, but not the level, a second save is needed for the level here as well.

From what I understand in the workflow, the right state after create-prefab operation is only the level is marked as dirty, not the prefab itself. The prefab does not need to be saved because it was already saved to disk in this operation. In-memory and on-disk representations should be in sync already.

ParasiteSt:
Ok, I understand what your saying now. I misunderstood the issue, yes these two are unrelated, and you are correct that the level is all that should need to be saved after creating a prefab as that prefab is not dirty upon creation (or at least it shouldn't be.)

PR #12506, opened 2022-10-07
Prefab | Improve create prefab code in public handler

forkercat:

## What does this PR do?

This PR is to refactor code in the create-prefab workflow, specifically in the **CreatePrefabInMemory** function. When we create a prefab in editor, it first calls **CreatePrefabInDisk**, in which it calls **CreatePrefabInMemory** and then save the template to disk.

**CreatePrefabInMemory** function has the following steps:
1. Prep step: Identify common parent root and related nested entities and instances
2. Detach the entities and nested instances
3. Create the new prefab instance object
4. Set up links between templates
5. Update the new prefab template

### Changes:
1. Replaced UpdatePrefabInstance undo function with RemoveEntities and UpdateEntity undo functions
2. Renamed some variables to avoid confusion. E.g. `containerEntity` -> `newContainerEntity` or `nestedContainerEntity`
3. Encapsulated and simplified the code of updating the new prefab template without transform info
    1. Previously, we have separate code to update prefab template and to remove transform info in the template.
    2. ~In this PR, we combine the code in a helper function UpdateContainerEntityAndResetTransformInTemplate.~
    3. Update: No reset is needed now in the workflow.
4. Make template DOM updates more selective for the parent template and the newly created template
5. Added comments to make each step more distinguishable and about whether it supports undo/redo.

Also fixes: https://github.com/o3de/o3de/issues/12013

## How was this PR tested?

- [x] Manually tested
- [x] Passed auto tests


Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -206,34 +229,25 @@ namespace AzToolsFramework
...
                 }
                 EditorEntityContextNotificationBus::Broadcast(&EditorEntityContextNotification::SetForceAddEntitiesToBackFlag, false);
 
-                // Update the template of the instance since the entities are modified since the template creation.
-                Prefab::PrefabDom serializedInstance;
-                if (m_instanceToTemplateInterface->GenerateDomForInstance(serializedInstance, instanceToCreate->get()))
-                {
-                    m_prefabSystemComponentInterface->UpdatePrefabTemplate(instanceToCreate->get().GetTemplateId(), serializedInstance);
-                }
-
```
srikappa-amzn, line 236: This needs to be added back because this was capturing the parenting information of top level entities and updating the template with that information. This still needs to happen and it needs to happen before adding in nested prefabs too.

I'm surprised that this is not wrapped in undo-redo logic. You can consider adding that too.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -279,33 +293,25 @@ namespace AzToolsFramework
                     }
                 });
 
-                // Create a link between the templates of the newly created instance and the instance it's being parented under.
-                CreateLink(
-                    instanceToCreate->get(), commonRootEntityOwningInstance->get().GetTemplateId(), undoBatch.GetUndoBatch(),
-                    AZStd::move(patch));
+                // Update the prefab template for the new instance.
+                // This update includes new child information and excludes transform data seen from its parent.
+                UpdateTemplateForInstance(newInstanceTemplateId, instanceToCreate->get());
```
srikappa-amzn, line 298: Discussed about this offline. This can just be limited to updating the newContainerEntityId with latest information rather than updating the entire prefab instance.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -375,6 +381,31 @@ namespace AzToolsFramework
...
+            const AZ::EntityId containerEntityId = instanceToUpdate.GetContainerEntityId();
+
+            AZ::EntityId parentEntityIdBeforeReset;
+            AZ::TransformBus::EventResult(parentEntityIdBeforeReset, containerEntityId, &AZ::TransformBus::Events::GetParentId);
+
+            AZ::Transform localTransformBeforeReset = AZ::Transform::CreateIdentity();
+            AZ::TransformBus::EventResult(localTransformBeforeReset, containerEntityId, &AZ::TransformBus::Events::GetLocalTM);
+
+            AZ::TransformBus::Event(containerEntityId, &AZ::TransformBus::Events::SetParent, AZ::EntityId());
+            AZ::TransformBus::Event(containerEntityId, &AZ::TransformBus::Events::SetLocalTM, AZ::Transform::CreateIdentity());
```
srikappa-amzn, line 399: All of this is not needed because the parent entity and transform info is not set on the new entity at this point.

srikappa-amzn (changes requested):

michabr (commented):
I'm curious whether this will work properly when creating a prefab out of entities that are part of an instance that's not in "Prefab Edit Mode". It seems like it will always update the template.

forkercat:
> creating a prefab out of entities that are part of an instance that's not in "Prefab Edit Mode"

```
Level
|-- SuperCar.prefab      (being edited)
    |-- Driver
    |-- Car.prefab
        |-- Engine
        |-- Wheel.prefab
|-- Taxi
```
In this example, SuperCar is in "Prefab Edit Mode", Level and Wheel prefabs are not.
1. If we create a prefab out of Driver and Car.prefab, it works. Engine and Wheel.prefab are still unchanged and owned by Car.
2. We cannot create a prefab out of Driver and Taxi, because they are not sharing the same owning template. There is a check at the beginning of the function. [#L96](https://github.com/o3de/o3de/blob/development/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp#L96)
3. We cannot create a prefab out of Taxi, because it is grayed out and cannot be selected.

So I think we always update only one template in this operation, which is the one being edited (i.e. SuperCar).

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
forkercat: good callout. The current approach seems like an overkill and it can be reduced to only update what it needs. I'll try out to see if it works.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
forkercat:
```
// Example
SuperCar.prefab      (being edited)
|-- Car.prefab       (been created)
    |-- Engine
        |-- Piston
    |-- Wheel.prefab
```

> This needs to be added back because this was capturing the parenting information of top level entities and updating the template with that information. This still needs to happen and it needs to happen before adding in nested prefabs too.

Yes. I took a close look. This update currently does two things in the template:
1. Update Car's container entity. After `prefabSystemComponent->CreatePrefab()`, the container entity DOM has 0 component. So after this step, it has all components in place and the child sort component has top level entities in the array.
2. Update transform component of top level entities (e.g. Engine) with new parent info

So, I suggest we keep this original code to do both work at once rather than selectively doing update for each.

forkercat: > I'm surprised that this is not wrapped in undo-redo logic. You can consider adding that too.

After the discussion, I thought about it and it seems like no undo-redo logic for the newly created template is expected and right behavior.

Imagine if we add undo-redo support for it and we hit undo after prefab creation. The Car template will be in a state with no entities and instances, or it can be worse if the template is in a weird state with some stuff but it is not complete (e.g. missing parent/child info). What happens if at this time we try to instantiate a Car prefab? I assume the Car template needs not be reloaded from disk because we already have the in-memory representation. We didn't clear the Car template object during undo.

***If there is no redo/undo support for Car, how redo works to put back the Car prefab instance?***

For the whole Car instance DOM, it can be redo because the link creation between SuperCar and Car supports redo/undo.

***If there is no redo/undo support for Car, how undo/redo works for top level entities (e.g. Engine)?***

For top level entities, their parent data in transform component is stored in SuperCar template "Entities". It supports the undo operation. For redo,  Car template is valid and complete and it contains the parent info in top level entity DOM. I think the redo of Car instance DOM is captured in the link creation between SuperCar and Car.

michabr:
I was referring to the mode where "EnableOverridesUx" is true. Are we not worrying about that yet?

forkercat:
> I was referring to the mode where "EnableOverridesUx" is true. Are we not worrying about that yet?

We're not worrying about that. I tried the above steps that I mentioned when flag=true. The check is still valid.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.h
```diff
@@ -130,6 +130,12 @@ namespace AzToolsFramework
             PrefabDom ApplyContainerTransformAndGeneratePatch(AZ::EntityId containerEntityId, AZ::EntityId parentEntityId,
                 const AZ::Vector3& translation, const AZ::Quaternion& rotation);
 
+            //! Updates prefab template with a given prefab instance object.
+            //! This will deserialize an instance DOM that excludes transform information and update the template without undo support.
+            //! \param templateId The template to be updated.
+            //! \param instanceToUpdate The instance object used to update the template.
+            void UpdateTemplateForInstance(const TemplateId templateId, const Instance& instanceToUpdate);
```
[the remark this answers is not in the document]
forkercat: I changed it to `UpdateContainerEntityAndResetTransformInTemplate` to call out the transform reset purpose. It makes sense to me to have two responsibilities in one function:
1. Update contain entity DOM
2. Make sure the DOM does have reset transform

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
forkercat: In the new revision, I put it back. So it worked like before. It currently does two things as mentioned above: 
1. Update container entity DOM with all the components
2. Update top level entity DOMs and non-top level entity DOMs

It looks good for now. Later when we address the sort issue, we can change it to be more selective.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
forkercat: New container entity object's transform is updated by ApplyContainerTransformAndGeneratePatch. The call was placed before setting top level entities' parent. It looks like it was placed there by purpose. Still couldn't figure out why.

I tried to delay calling ApplyContainerTransformAndGeneratePatch to one line before CreateLink, but some prefab tests fail due to transform position mismatch (not for the new container entity, but its child).

> assert current_entity_pos.IsClose(expected_pos), \\\n  AssertionError: Entity 'Entity_0' position '(x=200.0000000,y=200.0000000,z=200.0000000)' is not located at expected position '(x=100.0000000,y=100.0000000,z=100.0000000)'\nTest result:  FAILURE"}

forkercat:
Noticed a child sort issue in Create Prefab even before this PR. I created a separate issue: https://github.com/o3de/o3de/issues/12603

Discussed offline to address that issue in a separate PR.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
srikappa-amzn: But you can avoid the resetting if you do the entity dom update this way.
- store the before state of entity in a dom by fetching it from the template just after the previous template update
- serialize the entity again in this function to get the after state
- generate a patch by comparing the doms
- apply that patch to the prefab template (with undo/redo support---because we need this info again if user does undo followed by redo)

I think this way of updating the template would be better because it captures all the diff of the container entity since the last template update and you don't have to do additional resetting logic here.

forkercat: Discussed offline: (1) The reset can be avoided if comparing two entity DOMs that both have the same transform info (even if it contains parent info). (2) The undo/redo is not needed.

forkercat:
The latest commit addressed some feedbacks. Here is summary of what changes:
1. Rename CreatePrefabInDisk and add comment about undo/redo
2. Remove reset logic and the helper function
3. Make template updates more selective (e.g. remove the instance DOM update). It should make it clear on how template DOMs get updated in each step.
4. Add comments regarding what does support undo/redo and what does not

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicInterface.h
```diff
@@ -44,24 +44,24 @@ namespace AzToolsFramework
...
              * Automatically detects descendants of entities, and discerns between entities and child instances.
+             * Note: It calls CreatePrefabInMemory internally and then saves the new template in disk.
              * @param entityIds The entities that should form the new prefab (along with their descendants).
              * @param filePath The absolute path for the new prefab file.
              * @return An outcome object with an entityId of the new prefab's container entity;
              *  on failure, it comes with an error message detailing the cause of the error.
              */
-            virtual CreatePrefabResult CreatePrefabInDisk(
-                const EntityIdList& entityIds, AZ::IO::PathView filePath) = 0;
+            virtual CreatePrefabResult CreatePrefabAndSaveTemplateInDisk(const EntityIdList& entityIds, AZ::IO::PathView filePath) = 0;
```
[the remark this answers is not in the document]
forkercat: good suggest. Thanks for the link!

michabr (approved):
This function does a lot of stuff. It's a bit hard to follow even with the comments. Might be worth splitting it up into smaller chunks at some point.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -183,21 +199,28 @@ namespace AzToolsFramework
...
 
-                    instancePtrs.emplace_back(AZStd::move(outInstance));
+                    detachedNestedInstancePtrs.emplace_back(AZStd::move(detachedInstance));
                 }
 
+                // Update common parent entity DOM with latest child information.
+                PrefabDom commonRootEntityDomAfterDetaching;
+                m_instanceToTemplateInterface->GenerateDomForEntity(commonRootEntityDomAfterDetaching, *commonRootEntity);
+
+                PrefabUndoHelpers::UpdateEntity(commonRootEntityDomBeforeDetaching, commonRootEntityDomAfterDetaching,
```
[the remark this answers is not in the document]
forkercat: Good point! Some time ago I had the thought of removing it and forgot about it when focusing on understanding how DOM changes in template. I'll give it a try.

PR #12508, opened 2022-10-08
Fix View Directory Label Link with proper directory link after gem creation

tkothadev:
## What does this PR do?
There was a weird issue discovered where shortly after a gem is created via Create a Gem workflow, trying to "View Directory" of said gem in the catalog produced a malformed directory link, leaving us unable to access the folder unless you close the program and restart.

This is the issue before the proposed fix:
CreateGemViewDirectoryWeirdPath

After the fix:
CreateGemViewDirectoryWorkingPath

The main issue was shown in the first gif. The rendered URL in the View Directory dialog uses %5C as the separator. This is an URL encoding for backward slash, which means the links were being rendered as windows paths instead of unix paths (which are compatible with URL links). The main fix was to update the CreateGem function in the PythonBindings to make sure to replace '\\' with '/' in the directory link field of the output `GemInfo` before returning the result.

### DIFF
--- a/Code/Tools/ProjectManager/Source/PythonBindings.cpp
+++ b/Code/Tools/ProjectManager/Source/PythonBindings.cpp
@@ -878,8 +878,8 @@ namespace O3DE::ProjectManager
         }
         else
         {
-            //For Windows paths, need to normalize directory link to forward slashes
-            gemInfoResult.m_directoryLink.replace("\\", "/");
+            //Make sure directory link is a normalized path that can be rendered in "View Directory" dialog
+            gemInfoResult.m_directoryLink = QDir::cleanPath(gemInfoResult.m_directoryLink);
             return AZ::Success(AZStd::move(gemInfoResult));
         }
     }
--- a/Code/Tools/ProjectManager/Source/PythonBindings.cpp
+++ b/Code/Tools/ProjectManager/Source/PythonBindings.cpp
@@ -878,6 +878,8 @@ namespace O3DE::ProjectManager
         }
         else
         {
+            //For Windows paths, need to normalize directory link to forward slashes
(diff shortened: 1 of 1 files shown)


Code/Tools/ProjectManager/Source/PythonBindings.cpp
```diff
@@ -878,6 +878,8 @@ namespace O3DE::ProjectManager
         }
         else
         {
+            //For Windows paths, need to normalize directory link to forward slashes
+            gemInfoResult.m_directoryLink.replace("\\", "/");
```
cgalvan, line 882: Could run this through `AZ::IO::Path` normalize (`LexicallyNormal`)

lumberyard-employee-dm: This can use the `AZ::IO::PathView` `AsPosix` function

lumberyard-employee-dm (approved):

moraaar (approved):

Code/Tools/ProjectManager/Source/PythonBindings.cpp
AMZN-alexpete: Eventually we should be storing all of our path type objects in AZ::IO::Path or similar types, instead of QStrings.  This has been on the backlog because it will touch a lot of code and will require careful testing.  I'm on the fence of doing a string replacement here with code that is specific to Windows - PALifying just this is a bit overkill, I'd rather run it through AZ::IO::Path.

tkothadev: Addressed @AMZN-alexpete's suggestion in the latest commit via QDir.

AMZN-alexpete (approved):
Thanks for updating!

Issue #11926, opened 2022-09-13
The Editor fails to launch if it was previously closed while still processing assets for the project with all gems enabled


LB-TomaszSerwanski:
**Describe the bug**
The Editor crashes on the splash screen if the User has previously closed it while still processing the assets. Please note that the issue only occurs on projects with all gems enabled.

Please note that the issue does not always occur is most likely to reproduce by closing the Editor while there is still over 800 jobs to be processed in the Asset Processor.

Additionally - we have noticed that closing the Editor on project with all gems enabled when still processing assets also closes the Asset Processor. This is not the case when doing so on the default template project, on which the Asset Processor remains launched and keeps processing the assets.

The workaround for the issue is to delete the cache folder of the project and reprocess all the assets.

Please refer to the attached video and log files for more details.

**Call Stack Trace**
```
 9) 00007FFD3ED70057 (KERNELBASE) : UnhandledExceptionFilter
 8) 00007FFD3E1ED6D8 (InProcessClient64) : (function-name not available)
 7) 00007FFD3E1ED321 (InProcessClient64) : (function-name not available)
 6) 00007FFD40F753B0 (ntdll) : memset
 5) 00007FFD40F5C766 (ntdll) : _C_specific_handler
 4) 00007FFD40F7229F (ntdll) : _chkstk
 3) 00007FFD40F21454 (ntdll) : RtlRaiseException
 2) 00007FFD40F70DCE (ntdll) : KiUserExceptionDispatcher
 1) D:\Git\o3de\Gems\Atom\RPI\Code\Source\RPI.Public\Model\Model.cpp (62) : AZ::RPI::Model::GetLods
```

**Steps to reproduce**
1. Create a new project with all gems enabled and build it.
2. Launch the Editor for the project.
3. Close the Editor while the Asset Processor is still working on the assets (ideally when there is ~850 jobs left).
4. Relaunch the Editor.

**Expected behavior**
The Editor launches without issues.

**Actual behavior**
The Editor crashes on the splash screen.

**Screenshots/Video**

https://user-images.githubusercontent.com/86953108/189912846-a02bb6b0-2562-4ff0-bd29-9bfdfe15b6c6.mp4

**Found in Branch**
Development (511a88c)
Stabilization/2210 (021d32a)

**Desktop/Device:**
 - Device: PC
 - OS: Windows
 - Version 10
 - CPU AMD Ryzen 5 3600
 - GPU Nvidia RTX 2060 SUPER
 - Memory 16GB

[...]

PR #12509, opened 2022-10-08
Asset Processor - Broadcast assets on connection - Dev Fix

amzn-mike:
## What does this PR do?

When starting editor from a cache that does not contain an assetcatalog.xml, editor relies on network messages from the AP to build its in-memory catalog. If the asset processor/editor are restarted before the first time starutp has completed, AP will resume where it was in the processing without resending the previously completed assets to the editor.

This fixes AP to send out all the existing assets in the catalog when a new connection is established. Also fixed issue in a few usages of CompileAssetSync where the Unknown return case was not being handled, which can occur when an asset has already been compiled.

Fixes https://github.com/o3de/o3de/issues/11926

## How was this PR tested?

Manually following repro steps in above linked issue
Added and ran automated test


santorac (commented):
The Atom changes look fine to me.

AMZN-stankowi (approved):

Code/Framework/AzFramework/AzFramework/Asset/AssetCatalog.cpp
```diff
@@ -808,96 +808,118 @@ namespace AzFramework
...
-            if (!isNewAsset)
-            {
-                // the following deliveries must happen on the main thread of the application:
-                AZ::SystemTickBus::QueueFunction([assetId]() 
+                if (!isNewAsset)
                 {
-                    AzFramework::AssetCatalogEventBus::Broadcast(&AzFramework::AssetCatalogEventBus::Events::OnCatalogAssetChanged, assetId);
-                });
+                    // the following deliveries must happen on the main thread of the application:
+                    AZ::SystemTickBus::QueueFunction(
```
junbo75, line 870: Just curious: Does it mean that all these queued notifications will be handled on the main thread (similar to TickEvents at https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Component/TickBus.h#L63) and is that expected?

amzn-mike: Yes, that is intended for this message, this isn't new code though and this branch won't run for my changes because of the check on line 834, this message isn't intended to trigger OnAssetReloaded.

Code/Tools/AssetProcessor/native/AssetManager/AssetCatalog.cpp
```diff
@@ -589,6 +589,54 @@ namespace AssetProcessor
...
+                const auto& legacyIds =
+                    m_registries[platform].GetLegacyMappingSubsetFromRealIds(AZStd::vector<AZ::Data::AssetId>{ assetInfo.second.m_assetId });
+
+                for (auto& legacyId : legacyIds)
+                {
+                    message.m_legacyAssetIds.emplace_back(legacyId.first);
+                }
+
+                bulkMessage.m_messages.push_back(AZStd::move(message));
+            }
```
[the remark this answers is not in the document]
amzn-mike: From looking at the lower level code, this is using QTcpSockets, and I don't believe a large message would be any different than many small ones, the sockets are treated like file streams, reading until all the expected data is read (there is a fixed size header at the front containing the length).  I'm not an expert in networking though so I'm not sure if there is some other concern here.

junbo75 (approved):

Code/Framework/AzFramework/AzFramework/Asset/AssetCatalog.cpp
```diff
@@ -808,96 +808,118 @@ namespace AzFramework
     // note that this function is called from within a network job thread in order to update
     // the catalog as soon as new data is available, before other systems try to reach in and query
     // the results.  For this reason, it is a direct call and protects its structures with a lock.
-    void AssetCatalog::AssetChanged(AzFramework::AssetSystem::AssetNotificationMessage message)
+    void AssetCatalog::AssetChanged(const AZStd::vector<AzFramework::AssetSystem::AssetNotificationMessage>& messages, bool isCatalogInitialize)
```
lmbr-pip, line 811: Is this a public API?

Code/Framework/AzFramework/AzFramework/Asset/AssetCatalog.cpp
```diff
@@ -808,96 +808,118 @@ namespace AzFramework
...
+                        // Nothing to do here - catalog initialize messages are intended to sync with the AP catalog.
+                        // Since this asset is already known, just stop processing to avoid triggering reloads.
+                        return;
+                    }
 
-                const AZ::Data::AssetType& assetType = isNewAsset ? message.m_assetType : assetInfoPair->second.m_assetType;
+#if defined(AZ_ENABLE_TRACING)
+                    if (message.m_assetType == AZ::Data::s_invalidAssetType)
+                    {
+                        AZ_TracePrintf(
```
lmbr-pip, line 844: nit: should we add how you set the type? Maybe "type is not set in ..."

Code/Framework/AzFramework/AzFramework/Asset/AssetCatalog.cpp
```diff
@@ -808,96 +808,118 @@ namespace AzFramework
...
+                    AZ::SystemTickBus::QueueFunction(
+                        [assetId]()
+                        {
+                            AZ::Data::AssetManager::Instance().ReloadAsset(assetId, AZ::Data::AssetLoadBehavior::Default, true);
+                        });
+                }
+            }
+            else
+            {
+                AZ_TracePrintf("AssetCatalog", "AssetChanged: invalid asset id: %s\n", assetId.ToString<AZStd::string>().c_str());
```
lmbr-pip, line 921: nit: Does not seem actionable.

Code/Framework/AzFramework/AzFramework/Asset/AssetCatalog.h
```diff
@@ -98,8 +98,8 @@ namespace AzFramework
 
         //////////////////////////////////////////////////////////////////////////
         // NetworkAssetUpdateInterface
-        void AssetChanged(AzFramework::AssetSystem::AssetNotificationMessage message) override;
-        void AssetRemoved(AzFramework::AssetSystem::AssetNotificationMessage message) override;
+        void AssetChanged(const AZStd::vector<AzFramework::AssetSystem::AssetNotificationMessage>& messages, bool isCatalogInitialize = false) override;
```
lmbr-pip, line 101: This are public APIs, did we consider deprecation as a strategy here? This will break legacy code integrated with these notifications. 

Or are these notifications still internal?

Code/Framework/AzFramework/AzFramework/Asset/AssetProcessorMessages.h
```diff
@@ -628,6 +628,24 @@ namespace AzFramework
             AZStd::vector<AZ::Data::ProductDependency> m_dependencies;
         };
 
+        // Bulk message for sending updates for multiple assets all at once.
```
lmbr-pip, line 631: nit: Use doxygen compatible comments for self-documenting code.

Code/Framework/AzFramework/AzFramework/Asset/AssetSystemComponent.cpp
```diff
@@ -167,6 +167,58 @@ namespace AzFramework
...
 
+            m_bulkMessageHandle = m_socketConn->AddMessageHandler(
+                BulkAssetNotificationMessage::MessageType,
+                [context](unsigned int /*typeId*/, unsigned int /*serial*/, const void* data, unsigned int dataLength)
+                {
+                    if (dataLength)
+                    {
+                        BulkAssetNotificationMessage bulkMessage;
+
+                        // note that we forbid asset loading and we set STRICT mode.  These messages are all the kind of message that is
```
lmbr-pip, line 178: Not quite sure what "Note: We forbid asset loading and we set strict mode." has to do with bulk asset notification processing.

jackalbe (approved):

Code/Tools/AssetProcessor/native/AssetManager/AssetCatalog.cpp
```diff
@@ -589,6 +589,54 @@ namespace AssetProcessor
...
+    void AssetCatalog::OnConnect(unsigned int connectionId, QStringList platforms)
+    {
+        // Send out a message for each asset to make sure the connected tools are aware of the existence of all previously built assets
+        // since the assetcatalog might not have been written out to disk previously.
+        for (QString platform : platforms)
+        {
+            QMutexLocker locker(&m_registriesMutex);
+            auto itr = m_registries.find(platform);
+
+            if(itr == m_registries.end())
```
lmbr-pip, line 601: nit: formatting "if (itr =="

Code/Tools/AssetProcessor/native/AssetManager/AssetCatalog.cpp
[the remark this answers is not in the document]
lmbr-pip: We can have packet fragementation where we exceed MTU/MSS sizes which will slightly degrade performance and potentially lead to a lot of retransmission costs if routing issues occur. As we are over a local socket its not as important beyond the cost to split, send and recombine. We are not thought sending a massive amount of overall data consistently. It would be more of a concern for remote AP IMHO

lmbr-pip (commented):
LG

Amount of notifications is a concern, so with this redesign do we have any performance data that can talk to the worst case overhead (ie nothing in asset cache, everything requires a notification)?

Code/Framework/AzFramework/AzFramework/Asset/AssetCatalog.cpp
amzn-mike: This API is specifically marked internal

Code/Framework/AzFramework/AzFramework/Asset/AssetCatalog.h
amzn-mike: Its the same internal API:

```
//! This internal interface exists only to communicate to an Asset Catalog or manager of Asset Catalogs
//! such as the PlatformAddressedAssetCatlogManager
//! If you want to know when new assets are ready to load or have changed, listen to the AssetCatalogEventBus
//! which will notify you when assets have been updated on the main thread only, and only after the catalog itself has
//! updated with the new information.
class NetworkAssetUpdateInterface
```

Code/Framework/AzFramework/AzFramework/Asset/AssetSystemComponent.cpp
amzn-mike: Hm, I copied it from the regular message.  The bulk one is just a wrapper around the regular message.  I think perhaps without explicitly disabling the asset loading, the objectstream might see the AssetIds and try to load them as assets, so that might be what that is about.  And strict mode I think is related to error tolerance when parsing, which is set since this is not a message coming from a file a user can mess up.  I think the comment does still apply here.

Code/Framework/AzFramework/AzFramework/Asset/AssetCatalog.cpp
amzn-mike: Updated a bit but these are kind of odd messages, they might as well be asserts because this is data coming from a network messages from AP, which is all coming from code.  I imagine if this error occurs its either some weird data corruption happened or most likely a coding error, so there's no user action to recommend.  Perhaps they should be upgraded to asserts?

Issue #12290, opened 2022-09-29
Feature Request: Rename sv_isTransient and update description


lmbr-pip:
**Is your feature request related to a problem? Please describe.**
Split from https://github.com/o3de/o3de/issues/12198

The netowrking CVAR `sv_isTransient` does not accurately describe what this CVAR controls. Suggest adding more semantic context to CVAR to highlight this will cause the server to close once all connected players leave. 

The CVAR is important when running on cloud compute to ensure servers correct clean up ie on Amazon GameLift and other hosting services, to prevent zombie servers that incur unexpected hosting costs.

Consider renaming ie `sv_AutoShutdown`, `sv_terminateOnPlayerExit`

**Describe the solution you'd like**
* Propose new CVAR
* Deprecate old CVAR
* Update CVAR description in code and docs to be clear to avoid confusion (https://www.o3de.org/docs/user-guide/networking/settings/)

**Describe alternatives you've considered**
Just updating docs https://www.o3de.org/docs/user-guide/networking/settings/ 

**Additional context**

neelmewada:
Hello @lmbr-pip @AMZN-Olex! I have created a pull request for this issue here: https://github.com/o3de/o3de/pull/12515

lmbr-pip:
Code is merged

PR #12515, opened 2022-10-10
added sv_terminateOnPlayerExit and deprecated sv_isTransient

neelmewada:
## What does this PR do?

This PR is related to issue: https://github.com/o3de/o3de/issues/12290. According to the issue, I have deprecated `sv_isTransient` CVAR and added `sv_terminateOnPlayerExit` CVAR with it's description.

## How was this PR tested?

It isn't fully tested as it's a really small fix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ _savebackup/
 *.swatches
 /imgui.ini
 
+
--- a/.gitignore
+++ b/.gitignore
@@ -21,4 +21,3 @@ _savebackup/
 *.swatches
 /imgui.ini
 
-
--- a/Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
+++ b/Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
@@ -972,7 +972,7 @@ namespace Multiplayer
         }
 
         // Signal to session management when there are no remaining players in a dedicated server for potential cleanup
-        // We avoid this for client server as the host itself is a user and non-transient dedicated servers
+        // We avoid this for client server as the host itself is a user and dedicated servers that do not terminate when all players have exited
         // Replaced sv_isTransient with sv_terminateOnPlayerExit
         if (sv_terminateOnPlayerExit && m_agentType == MultiplayerAgentType::DedicatedServer && connection->GetConnectionRole() == ConnectionRole::Acceptor)
         {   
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -21,3 +21,4 @@ _savebackup/
 *.swatches
 /imgui.ini
 
+
```
AMZN-puvvadar, line 24: Nitpick: It looks like a whitespace change snuck in. This should be removed.

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
```diff
@@ -972,7 +973,8 @@ namespace Multiplayer
 
         // Signal to session management when there are no remaining players in a dedicated server for potential cleanup
         // We avoid this for client server as the host itself is a user and non-transient dedicated servers
```
AMZN-puvvadar, line 975:
```suggestion
        // We avoid this for client server as the host itself is a user and dedicated servers that do not terminate when all players have exited
```

If we're moving away from transient as a term, it'd be good to update this comment as well.

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
```diff
@@ -972,7 +973,8 @@ namespace Multiplayer
...
-        if (sv_isTransient && m_agentType == MultiplayerAgentType::DedicatedServer && connection->GetConnectionRole() == ConnectionRole::Acceptor)
+        // Replaced sv_isTransient with sv_terminateOnPlayerExit
```
AMZN-puvvadar, line 976: I think this comment can be removed given sv_isTransient is already marked DEPRECATED. That way the comments remain descriptive of the logic and not of naming history.

.gitignore
neelmewada: Hi @AMZN-puvvadar, yes I'll remove it for sure :) Should I make a new commit (along with other changes you mentioned) with a whitespace removed?

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
neelmewada: @AMZN-puvvadar Just fixed this and made a commit

.gitignore
neelmewada: @AMZN-puvvadar I have removed this whitespace in the latest commit. Do let me know if you need any other changes 👍

neelmewada:
Hi @AMZN-puvvadar, I have fixed/implemented all of your suggestions and pushed everything. Let me know if there's anything else needed. Thanks!

AMZN-puvvadar (approved):

lmbr-pip (approved):

lmbr-pip:
First automated test run, failed with unrelated error : `/data/workspace/o3de/AutomatedTesting/Gem/PythonTests/editor_test_testing/EditorTest_That_Fails.py` failed to fail to fail on Linux.

lmbr-pip:
Once this in, we need to submit PR for https://www.o3de.org/docs/user-guide/networking/settings/

neelmewada:
@lmbr-pip do I have to do anything on my side about the error? As you said it seems unrelated

lmbr-pip:
@neelmewada Am discussing with SIG/Testing folks. We can make an exception here most likely as branch is still failing on AR but on a test that has nothing todo with your change.

lmbr-pip:
Am requesting AR bypass for change due to known issue block AR approval, as per https://github.com/o3de/community/blob/main/contributors/docs/override_pr_status_check.md

"Some integ tests inside editor_test_testing intentionally fail, but assert on that failure. The Jenkins TestResults page still shows the test failure, even though it is intentional. This is causing confusion as users think that a real test failure has occurred."

Issue #11411, opened 2022-08-19
Deleting an entity does not close its context menu


LB-ArturZieba:
**Describe the bug**
If a context menu is opened on an entity, and that entity is deleted using the Delete key, the context menu remains opened. This can lead to unintended behavior, and potentially crashing (for example, as shown on the video, attempting to detach a prefab on a deleted prefab causes the Editor to freeze, and crash).

Please note that the crash shown on the video only occurs on Windows.

**Steps to reproduce**
1. Create an entity.
2. Right-click the created entity.
3. Press the Delete key.

**Expected behavior**
The context menu closes.

**Actual behavior**
Context menu remains open, and can be interacted with.

**Video**

https://user-images.githubusercontent.com/86952082/185622202-fef4e72e-8804-4e3c-afdb-a5d2fcd811cc.mp4

**Found in Branch**
Development (e876c36)

**Desktop**
 - Device: PC
 - OS: Windows
 - Version 11
 - CPU AMD Ryzen 5 3600
 - GPU NVIDIA GeForce RTX 2060 SUPER
 - Memory 16GB

rainbj:
Reviewed issue. We (sig-ui-ux team) agree with needing to fix this

LB-TomaszSerwanski:
The issue verified as Fixed on the latest Development (a912a04) branch.

PR #12518, opened 2022-10-10
Fix for delete key shortcut operating when outliner context menu is o…

sphrose:
…pen.

## What does this PR do?

This pr fixes https://github.com/o3de/o3de/issues/11411. The problem is caused by the menu being synchronous, but the delete hotkey still being available during its lifetime. This could result in the selected entity being deleted whilst the menu is still active.

## How was this PR tested?

Tested in the editor as described, ensured menu otherwise operates correctly and ensures other context menus were not affected.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/UI/Outliner/EntityOutlinerWidget.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/UI/Outliner/EntityOutlinerWidget.cpp
@@ -342,6 +342,13 @@ namespace AzToolsFramework
     //  Currently, the first behavior is implemented.
     void EntityOutlinerWidget::OnSelectionChanged(const QItemSelection& selected, const QItemSelection& deselected)
     {
+        // If we have an open context menu when the selection changes, it is no longer relevant and needs closing.
+        QMenu* menu = qobject_cast<QMenu*>(QApplication::activePopupWidget());
+        if (menu && menu->parentWidget() == this)
+        {
+            menu->close();
+        }
+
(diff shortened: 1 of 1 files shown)


AMZN-daimini (commented):
I think the same menu can also be triggered from the viewport?

gadams3 (approved):

sphrose:
> I think the same menu can also be triggered from the viewport?

It can, but the shortcut keys do not operate in the viewport while the menu is showing.

AMZN-daimini:
Ok, this sounds like the sort of issue caused by the shortcut dispatcher then :/

AMZN-daimini (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Outliner/EntityOutlinerWidget.cpp
```diff
@@ -342,6 +342,13 @@ namespace AzToolsFramework
     //  Currently, the first behavior is implemented.
     void EntityOutlinerWidget::OnSelectionChanged(const QItemSelection& selected, const QItemSelection& deselected)
     {
+        // If we have an open context menu when the selection changes, it is no longer relevant and needs closing.
+        QMenu* menu = qobject_cast<QMenu*>(QApplication::activePopupWidget());
+        if (menu && menu->parentWidget() == this)
```
pollend, line 347: might be able to simplify this logic a bit
```suggestion
        if (auto* menu = qobject_cast<QMenu*>(QApplication::activePopupWidget()) && menu->parentWidget() == this)
```

pollend (approved):

PR #12520, opened 2022-10-10
fix issue with animation editor debug draw disappearing when unfocused

greerdv:

## What does this PR do?
Moves animation editor debug drawing from happening OnTick to OnBeginPrepareRender, which means that it continues to be drawn when the animation editor is not focused. 

## How was this PR tested?
Manual testing in the animation editor.


greerdv:
Behaviour before

https://user-images.githubusercontent.com/82226198/194920022-5c6e912d-6bcb-4d72-ae19-f445067eb30d.mp4

Behaviour after

https://user-images.githubusercontent.com/82226198/194920036-a9c27a9b-8d09-4c70-97c9-53fbb7b749e6.mp4

Gems/AtomLyIntegration/EMotionFXAtom/Code/Tools/EMStudio/AnimViewportRenderer.cpp
```diff
@@ -392,4 +392,14 @@ namespace EMStudio
             iblFeatureProcessor, m_skyboxFeatureProcessor, exposureControlSettingInterface, m_directionalLightFeatureProcessor,
             cameraConfig, m_lightHandles, false);
     }
+
+    AZ::RPI::SceneId AnimViewportRenderer::GetRenderSceneId()
```
hultonha, line 396: update: Looks like this could be `const`

Gems/AtomLyIntegration/EMotionFXAtom/Code/Tools/EMStudio/AnimViewportRenderer.cpp
```diff
@@ -392,4 +392,14 @@ namespace EMStudio
...
+    {
+        return m_scene->GetId();
+    }
+
+    const AZStd::vector<AZ::Entity*>& AnimViewportRenderer::GetActorEntities()
```
hultonha, line 401: update: Also looks like this could be marked `const`

Gems/AtomLyIntegration/EMotionFXAtom/Code/Tools/EMStudio/AnimViewportWidget.cpp
```diff
@@ -410,4 +421,19 @@ namespace EMStudio
         m_renderOverlay.setGeometry(geometry());
         m_viewportUiManager.Update();
     }
+
+    void AnimViewportWidget::OnBeginPrepareRender()
+    {
+        if (m_debugDisplay)
+        {
```
hultonha, line 428: nit: We could early return here if `m_debugDisplay` is nullptr to avoid the indentation (or if this _should_ never happen we could add an Assert too?)

hultonha (approved):
Nice! 😄 Change looks great to me, really nice fix! 🙂

moraaar (approved):
Nice change.

Gems/AtomLyIntegration/EMotionFXAtom/Code/Tools/EMStudio/AnimViewportRenderer.cpp
greerdv: Done.

Gems/AtomLyIntegration/EMotionFXAtom/Code/Tools/EMStudio/AnimViewportWidget.cpp
greerdv: I don't think it makes much difference here because it's such a short function, but will bear in mind for the future.

Issue #12147, opened 2022-09-22
Setting cylinder collider radius or height to zero terminates the Editor


piotr-zyskowski-rai:
**Describe the bug**
When the cylinder collider radius or length is set to zero, the Editor terminates

**Assets required**
n/a

**Steps to reproduce**
1. Create a cylinder collider using `createComponent` API and set the length or radius to 0
```
  PhysX::EditorColliderComponentRequestBus::Event(
      AZ::EntityComponentIdPair(entityId, component->GetId()),
      &PhysX::EditorColliderComponentRequests::SetCylinderHeight,
      0.0;
  PhysX::EditorColliderComponentRequestBus::Event(
      AZ::EntityComponentIdPair(entityId, component->GetId()),
      &PhysX::EditorColliderComponentRequests::SetCylinderRadius,
      0.0);

```
2. When executed the `'std::bad_optional_access'` exception will be thrown

Another way to reproduce:
1. Open any project
2. Create an entity and add a collider component
3. Set collider component to be "Cylinder"
4. Set radius or height to 0
5. Editor finishes with `Process finished with exit code 134 (interrupted by signal 6: SIGABRT)`

**Expected behavior**
No Editor termination; requested cylinder added - not a great cylinder but technically correct

**Actual behavior**
Editor terminates

**Screenshots/Video**
n/a

**Found in Branch**
n/a

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Ubuntu
 - Version 20.04
 - CPU Intel i7-10750H 
 - GPU Nvidia GeForce 3060
 - Memory 32GB

**Additional context**

PR #12522, opened 2022-10-10
#12147 Fixed crash when cylinder sizes are set to zero

SergeyAMZN:

## What does this PR do?

Fixes a crash when either radius or height of the cylinder collider is set to zero.
Issue #12147

## How was this PR tested?

Unit tests, manual testing in the Editor

### DIFF
--- a/Gems/PhysX/Code/Source/EditorColliderComponent.cpp
+++ b/Gems/PhysX/Code/Source/EditorColliderComponent.cpp
@@ -1538,19 +1538,22 @@ namespace PhysX
         const float height = m_shapeConfiguration.m_cylinder.m_height;
         const float radius = m_shapeConfiguration.m_cylinder.m_radius;
 
-        if (height > 0.0f && radius > 0.0f)
+        if (height <= 0.0f)
         {
-            Utils::Geometry::PointList samplePoints = Utils::CreatePointsAtFrustumExtents(height, radius, radius, subdivisionCount).value();
+            AZ_Error("PhysX", false, "Cylinder height must be greater than zero. Entity: %s", GetEntity()->GetName().c_str());
+            return;
+        }
 
-            const AZ::Vector3 scale = m_shapeConfiguration.m_cylinder.m_configuration.m_scale;
-            m_shapeConfiguration.m_cylinder.m_configuration = Utils::CreatePxCookedMeshConfiguration(samplePoints, scale).value();
-        }
-        else
+        if (radius <= 0.0f)
         {
-            AZ_Error("PhysX", false, "Cylinder radius and height must be greater than zero. Entity: %s",
-                GetEntity()->GetName().c_str());
-            m_shapeConfiguration.m_cylinder.m_configuration = {};
+            AZ_Error("PhysX", false, "Cylinder radius must be greater than zero. Entity: %s", GetEntity()->GetName().c_str());
+            return;
         }
+
+        Utils::Geometry::PointList samplePoints = Utils::CreatePointsAtFrustumExtents(height, radius, radius, subdivisionCount).value();
+
+        const AZ::Vector3 scale = m_shapeConfiguration.m_cylinder.m_configuration.m_scale;
+        m_shapeConfiguration.m_cylinder.m_configuration = Utils::CreatePxCookedMeshConfiguration(samplePoints, scale).value();
     }
 
     void EditorColliderComponentDescriptor::Reflect(AZ::ReflectContext* reflection) const
--- a/Gems/PhysX/Code/Source/EditorColliderComponent.cpp
+++ b/Gems/PhysX/Code/Source/EditorColliderComponent.cpp
@@ -1464,9 +1464,16 @@ namespace PhysX
 
     void EditorColliderComponent::SetCylinderRadius(float radius)
     {
-        m_shapeConfiguration.m_cylinder.m_radius = radius;
-        UpdateCylinderCookedMesh();
-        CreateStaticEditorCollider();
+        if (radius > 0.0f)
+        {
+            m_shapeConfiguration.m_cylinder.m_radius = radius;
+            UpdateCylinderCookedMesh();
+            CreateStaticEditorCollider();
+        }
+        else
+        {
+            AZ_Error("PhysX", false, "SetCylinderRadius: radius must be greater than zero.");
+        }
     }
 
     float EditorColliderComponent::GetCylinderRadius() const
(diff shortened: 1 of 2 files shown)


Gems/PhysX/Code/Source/EditorColliderComponent.cpp
```diff
@@ -1464,9 +1464,16 @@ namespace PhysX
 
     void EditorColliderComponent::SetCylinderRadius(float radius)
     {
-        m_shapeConfiguration.m_cylinder.m_radius = radius;
-        UpdateCylinderCookedMesh();
-        CreateStaticEditorCollider();
+        if (radius > 0.0f)
```
pollend, line 1467: I think I would flip this if statement
```
if(radius <= 0.0f) {
    AZ_Error("PhysX", false, "SetCylinderRadius: radius must be greater than zero.");
    return;
} 


```

Gems/PhysX/Code/Tests/RigidBodyComponentTests.cpp
```diff
@@ -140,4 +140,52 @@ namespace PhysXEditorTests
...
+            if (i == 0)
+            {
+                PhysX::EditorColliderComponentRequestBus::Event(
+                    idPair, &PhysX::EditorColliderComponentRequests::SetCylinderRadius, 0.0f);
+            }
+            else
+            {
+                PhysX::EditorColliderComponentRequestBus::Event(
+                    idPair, &PhysX::EditorColliderComponentRequests::SetCylinderHeight, 0.0f);
+            }
```
pollend, line 178: this looks really opaque I would probably just split this into two test cases. more code but the setup and assertions would be more cleaer.

Gems/PhysX/Code/Source/EditorColliderComponent.cpp
```diff
@@ -1476,9 +1483,16 @@ namespace PhysX
 
     void EditorColliderComponent::SetCylinderHeight(float height)
     {
-        m_shapeConfiguration.m_cylinder.m_height = height;
-        UpdateCylinderCookedMesh();
-        CreateStaticEditorCollider();
+        if (height > 0.0f)
```
pollend, line 1486: similar suggestion as above

Gems/PhysX/Code/Source/EditorColliderComponent.cpp
```diff
@@ -1525,11 +1539,20 @@ namespace PhysX
         const AZ::u8 subdivisionCount = m_shapeConfiguration.m_cylinder.m_subdivisionCount;
         const float height = m_shapeConfiguration.m_cylinder.m_height;
         const float radius = m_shapeConfiguration.m_cylinder.m_radius;
-        Utils::Geometry::PointList samplePoints = Utils::CreatePointsAtFrustumExtents(height, radius, radius, subdivisionCount).value();
 
-        const AZ::Vector3 scale = m_shapeConfiguration.m_cylinder.m_configuration.m_scale;
-        m_shapeConfiguration.m_cylinder.m_configuration
-            = Utils::CreatePxCookedMeshConfiguration(samplePoints, scale).value();
+        if (height > 0.0f && radius > 0.0f)
```
pollend, line 1543: do the same as above but split this into two separate error messages.

Gems/PhysX/Code/Source/EditorColliderComponent.cpp
```diff
@@ -1525,11 +1539,20 @@ namespace PhysX
...
+        {
+            Utils::Geometry::PointList samplePoints = Utils::CreatePointsAtFrustumExtents(height, radius, radius, subdivisionCount).value();
+
+            const AZ::Vector3 scale = m_shapeConfiguration.m_cylinder.m_configuration.m_scale;
+            m_shapeConfiguration.m_cylinder.m_configuration = Utils::CreatePxCookedMeshConfiguration(samplePoints, scale).value();
+        }
+        else
+        {
+            AZ_Error("PhysX", false, "Cylinder radius and height must be greater than zero. Entity: %s",
```
pollend, line 1552: its not clear if the height or the radius is invalid.

Gems/PhysX/Code/Source/EditorColliderComponent.cpp
SergeyAMZN: yep good idea, thanks!

Gems/PhysX/Code/Tests/RigidBodyComponentTests.cpp
SergeyAMZN: Yeah I did consider having two test cases initially but thought its testing pretty much the same functionality and it wasn't enough to justify extra code. 
I agree having separate test cases is more "correct" here, so changed now, thanks!

Gems/PhysX/Code/Source/EditorColliderComponent.cpp
SergeyAMZN: Yep, updated to two error messages, thanks!

pollend (approved):

gadams3 (approved):

moraaar (approved):

Gems/PhysX/Code/Tests/RigidBodyComponentTests.cpp
```diff
@@ -140,4 +140,65 @@ namespace PhysXEditorTests
...
+
+        // Set collider to be a cylinder
+        const Physics::ShapeType shapeType = Physics::ShapeType::Cylinder;
+        PhysX::EditorColliderComponentRequestBus::Event(idPair, &PhysX::EditorColliderComponentRequests::SetShapeType, shapeType);
+
+        // Set collider cylinder radius to zero
+        PhysX::EditorColliderComponentRequestBus::Event(idPair, &PhysX::EditorColliderComponentRequests::SetCylinderRadius, 0.0f);
+
+        // Notify listeners that collider has changed
+        Physics::ColliderComponentEventBus::Event(editorEntity->GetId(), &Physics::ColliderComponentEvents::OnColliderChanged);
```
greerdv, line 166: Is it not a bit weird having to do this manually? Or is it just because it's in a test and something that would normally handle this is not stood up?

greerdv (approved):

hultonha (approved):

Gems/PhysX/Code/Tests/RigidBodyComponentTests.cpp
SergeyAMZN: We get these automatically when working with colliders via UI in the Editor. 
Here `OnColliderChanged` is an indication that we finished making changes to the collider and RigidBodyComponent can re-create the actual rigid body. 
Doing it on every single change (i.e. inside SetCylinderRadius) wouldn't be great either as we can have incomplete setups used for constructing the rigid body which could spawn unnecessary warnings/errors..

greerdv: Cool, thanks for the explanation.

PR #12523, opened 2022-10-10
Fix hierarchy for child transform nodes of a parent mesh node

### LINKED ISSUE #12021: ProcPrefab issues with complex Hierarchy with empty transforms and child transforms with local transforms.
**Describe the bug**
A clear and concise description of what the bug is. Try to isolate the issue to help the community to reproduce it easily and increase chances for a fast fix.

Hierarchy works great if all the mesh transforms pivots are 0,0,0.
image

When Exporting a model with a more complex Hierarchy, where I might have Parents with 0,0,0 Transforms, but have child meshes with Pivots that are local, customized positions for a custom pivot point.
image

This is where issues are shown.
image

If I where to just export the meshes with no empty transform parents:
image

If I open the failed export in Maya with the complex transforms it will import correctly:
image

Here is the complex test file used: (I created this file)
[RobotCart_Complex.zip](https://github.com/o3de/o3de/files/9578857/RobotCart_Complex.zip)

I did a simple test model which work correctly:
image
image

Detaching, everything looks correct.
image
Here is the file: 
[Simple_Complex.zip](https://github.com/o3de/o3de/files/9578936/Simple_Complex.zip)

Would be interesting to see what caused the car to flip while the simple model is successful. 

**Steps to reproduce**
1. Build a model with a complex Hierarchy
2. Import into o3de with a procedural prefab. 

**Expected behavior**
Model transforms, even if they are not 0,0,0 are respected.

**Actual behavior**
Model will flip, however when testing in another tool such as Maya, the Blender file is correct.

**Found in Branch**
Nightly 9-15-2022

 - Device: PC
 - OS: Windows
 - Version 10
 - CPUIntel I9-9900k
 - GPU NVidia RTX 2080ti
 - Memory [e.g. 16GB]

**Additional context**

michabr:

## What does this PR do?

The prefab builder was ignoring transform nodes that were children of a mesh node. This PR fixes that by creating unique entities for transform nodes that are not associated with a mesh but are direct children of a mesh node.

More context on scene graph:
The first child transform of a mesh node in a graph contains the transform data associated with that mesh. All other child transform nodes of the mesh are considered actual children when preserving the parent/child hierarchy.

Fixes #12021 

## How was this PR tested?

Manual testing with the asset provided in the GHI #12021. Updated unit test to include this scenario and checked that it passed.

### DIFF
--- a/Gems/Prefab/PrefabBuilder/PrefabGroup/tests/PrefabBehaviorTests.cpp
+++ b/Gems/Prefab/PrefabBuilder/PrefabGroup/tests/PrefabBehaviorTests.cpp
@@ -388,7 +388,7 @@ namespace UnitTest
         ASSERT_TRUE(AzToolsFramework::Prefab::PrefabDomUtils::LoadInstanceFromPrefabDom(instance, *prefabDomRef));
         EXPECT_TRUE(IsChildOfParent(instance, "3", "2")); // Mesh entity is child of a transform entity
         EXPECT_TRUE(IsChildOfParent(instance, "6", "2")); // Transform entity is child of another transform entity
-        EXPECT_FALSE(IsChildOfParent(instance, "5", "4")); // First transform entity is not a child of the mesh entity
+        EXPECT_FALSE(IsChildOfParent(instance, "4", "3")); // First transform entity is not a child of the mesh entity
         EXPECT_TRUE(IsChildOfParent(instance, "5", "3")); // Second transform entity is child of the mesh entity
     }
 
--- a/Gems/Prefab/PrefabBuilder/PrefabGroup/tests/PrefabBehaviorTests.cpp
+++ b/Gems/Prefab/PrefabBuilder/PrefabGroup/tests/PrefabBehaviorTests.cpp
@@ -331,15 +331,21 @@ namespace UnitTest
         auto scene = CreateEmptyMockSceneWithRoot("Manifest", "//o3de/watch.folder/manifest_src_file.xml", "//o3de/watch.folder");
 #endif
         /*---------------------------------------\
+        Notes on the graph hierarchy:
+        - 3m is a node with mesh data. Proc prefab builder will create an entity for this node
+        - 4t contains the transform data for the mesh. It's a child of 3m because a node can only contain one data item.
+          Proc prefab builder will apply the data of the first child transform to the mesh entity's transform component
+        - 5t is a standalone child transform node of the mesh node. Proc prefab builder will create an entity for this node
+
                     Root
                      |
                      1
                      |
                      2t
                    /   \
-                  3m    5t
-                 /
-                4t
+                  3m    6t
+                 /  \
+                4t  5t
         \---------------------------------------*/
 
         AZ::Matrix3x4 nonIdentityMatrix = AZ::Matrix3x4::CreateScale(AZ::Vector3(10.0f));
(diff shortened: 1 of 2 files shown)


AMZN-stankowi (approved):

AMZN-stankowi:
The description on the graph hierarchy in the automated test helps a ton, thanks for adding it.

amzn-mike (approved):

Gems/Prefab/PrefabBuilder/PrefabGroup/tests/PrefabBehaviorTests.cpp
```diff
@@ -379,7 +387,9 @@ namespace UnitTest
         AzToolsFramework::Prefab::Instance instance;
         ASSERT_TRUE(AzToolsFramework::Prefab::PrefabDomUtils::LoadInstanceFromPrefabDom(instance, *prefabDomRef));
         EXPECT_TRUE(IsChildOfParent(instance, "3", "2")); // Mesh entity is child of a transform entity
-        EXPECT_TRUE(IsChildOfParent(instance, "5", "2")); // Transform entity is child of another transform entity
+        EXPECT_TRUE(IsChildOfParent(instance, "6", "2")); // Transform entity is child of another transform entity
+        EXPECT_FALSE(IsChildOfParent(instance, "5", "4")); // First transform entity is not a child of the mesh entity
```
srikappa-amzn, line 391: Should this have been IsChildOfParent(instance, "5", "3")?

Gems/Prefab/PrefabBuilder/PrefabGroup/tests/PrefabBehaviorTests.cpp
```diff
@@ -331,15 +331,21 @@ namespace UnitTest
         auto scene = CreateEmptyMockSceneWithRoot("Manifest", "//o3de/watch.folder/manifest_src_file.xml", "//o3de/watch.folder");
 #endif
         /*---------------------------------------\
+        Notes on the graph hierarchy:
+        - 3m is a node with mesh data. Proc prefab builder will create an entity for this node
+        - 4t contains the transform data for the mesh. It's a child of 3m because a node can only contain one data item.
```
srikappa-amzn, line 336: > '.... because a node can only contain one data item'

Is this limitation coming from the fbx file format or a design choice made while building the nodes? I ask because this approach of making the first child contain transform data seems counterintuitive to me. Ideally if the mesh node itself can also hold the transform data, it'll be an easier mental model to work through and also potentially won't cause node-sync issues if the tree gets modified later.

jackalbe (approved):

Gems/Prefab/PrefabBuilder/PrefabGroup/tests/PrefabBehaviorTests.cpp
michabr: It's how the SceneGraph class stores content. I don't know the decision process on this but perhaps @AMZN-stankowi or @jackalbe can give more info.

Gems/Prefab/PrefabBuilder/PrefabGroup/tests/PrefabBehaviorTests.cpp
michabr: Yes, thanks! Changed and test still passes :)

PR #12549, opened 2022-10-12
Fix profiler when loading a capture from a file

### LINKED ISSUE #12548: [PC][Profiler] Issues when loading a CPU capture file
**Describe the bug**
When loading a CPU capture from a file that was recorded in another device, the CPU profile doesn't show the profiled data correctly.
The graphic visualizer doesn't show the captured the bars for the captures regions
The names of the regions are not complete (a "%s" is showed instead of the name)
The times for the captured regions is incorrect (it shows seconds instead of milliseconds)

**Assets required**
Sync to the Atom Sample Viewer project

**Steps to reproduce**
1. Launch ASV 
2. Select "Profile" from the Menu
3. Select "CPU Profiler"
4. Click the "Capture" button
5. Observe the file being saved
6. Press "Load File"
7. Select the generated file from the list
8. Click "Swap to visualizer"
9. See error

**Expected behavior**
The regions of the capture are not displayed correctly. The names of the regions are not properly named.

**Actual behavior**
The regions of the capture in the visualizer are displayed correctly. The names of the regions include the full name and no "%s" is shown.

**Screenshots/Video**
If applicable, add screenshots and/or a video to help explain your problem.

**Found in Branch**
Development

akioCL:
Add ticks per second property when recording a CPU capture
Fix region's names when including "print" arguments
Fix displaying of regions when loading a single frame capture
Fixes #12548

## What does this PR do?

Fixes multiple issues with the CPU profiler when loading a single frame capture

## How was this PR tested?

Created a capture on an Android device. Loaded the capture in the CPU profiler on PC.


Gems/Profiler/Code/Source/CpuProfiler.cpp
```diff
@@ -98,9 +98,10 @@ namespace Profiler
             {
                 // Lazy initialization, creates an instance of the Thread local data if it's not created, and registers it
                 RegisterThreadStorage();
-
+                va_list args;
+                va_start(args, eventNameArgCount);
                 // Push it to the stack
-                CachedTimeRegion timeRegion({budget->Name(), eventName});
+                CachedTimeRegion timeRegion({ budget->Name(), AZStd::string::format_arg(eventName, args).c_str() });
```
lumberyard-employee-dm, line 104: This would be better to use a `AZStd::fixed_string<512>::format_arg` to avoid any heap allocations

Gems/Profiler/Code/Source/CpuProfiler.cpp
```diff
@@ -350,7 +351,7 @@ namespace Profiler
             // Add the cached regions to the map
             for (auto& cachedTimeRegion : m_cachedTimeRegions)
             {
-                const AZStd::string regionName = cachedTimeRegion.m_groupRegionName.m_regionName;
+                const AZStd::string regionName = cachedTimeRegion.m_groupRegionName.m_regionName.GetCStr();
```
lumberyard-employee-dm, line 354: This should use GetStringView instead of GetCStr to avoid an unnecessary strlen call on a c-string.
```suggestion
                const AZStd::string regionName = cachedTimeRegion.m_groupRegionName.m_regionName.GetStringView();
```

Gems/Profiler/Code/Source/CpuProfiler.cpp
```diff
@@ -317,7 +318,7 @@ namespace Profiler
     // Gets called when region ends and all data is set
     void CpuTimingLocalStorage::AddCachedRegion(const CachedTimeRegion& timeRegionCached)
     {
-        if (auto iter = m_hitSizeLimitMap.find(timeRegionCached.m_groupRegionName.m_regionName);
+        if (auto iter = m_hitSizeLimitMap.find(timeRegionCached.m_groupRegionName.m_regionName.GetCStr());
```
lumberyard-employee-dm, line 321: Because this is a map of `AZStd::string` the `m_regionName.GetStringView` function should be used instead to avoid retrieving a null-terminated c-string and needing needing the string constructor iterate over the string to determine it's string length again. The `AZStd::string` has a constructor which accepts a string_view that just reads the size directly from it.
```suggestion
        if (auto iter = m_hitSizeLimitMap.find(timeRegionCached.m_groupRegionName.m_regionName.GetStringView());
```

Gems/Profiler/Code/Source/CpuProfiler.h
```diff
@@ -227,5 +227,6 @@ namespace Profiler
         CpuProfilingStatisticsSerializer(const AZStd::ring_buffer<TimeRegionMap>& continuousData);
 
         AZStd::vector<CpuProfilingStatisticsSerializerEntry> m_cpuProfilingStatisticsSerializerEntries;
+        AZStd::sys_time_t m_timeTicksPerSecond = 0;
```
lumberyard-employee-dm, line 230: It is simpler to just store an [AZStd::chrono::microseconds](https://en.cppreference.com/w/cpp/chrono/duration)

lumberyard-employee-dm:
```suggestion
        AZStd::chrono::microseconds m_timeMicroseconds = 0;
```

Gems/Profiler/Code/Source/ImGuiCpuProfiler.cpp
```diff
@@ -126,6 +111,19 @@ namespace Profiler
...
+        const AZStd::sys_time_t ticksPerSecond = m_ticksPerSecondFromFile > 0 ? m_ticksPerSecondFromFile : AZStd ::GetTimeTicksPerSecond();
+        // Note: converting to microseconds integer before converting to milliseconds float
+        AZ_Assert(ticksPerSecond >= 1000, "Error in converting ticks to ms, expected ticksPerSecond >= 1000");
+        return static_cast<float>((ticks * 1000) / (ticksPerSecond / 1000)) / 1000.0f;
+    }
+
+    float ImGuiCpuProfiler::TicksToMs(AZStd::sys_time_t ticks)
+    {
+        return TicksToMs(static_cast<double>(ticks));
+    }
```
lumberyard-employee-dm, line 125: Ms is normally meant to mean "milliseconds", while Us is meant to mean "microseconds"

But instead of needing a ticks to microsecond, you could just have a function that performs a returns `AZStd::chrono::duration<float, AZStd::chrono::micro>` 
```suggestion
    float ImGuiCpuProfiler::TicksToMs(double ticks)
    {
        const AZStd::sys_time_t ticksPerSecond = m_ticksPerSecondFromFile > 0 ? m_ticksPerSecondFromFile : AZStd ::GetTimeTicksPerSecond();
        // Note: converting to microseconds integer before converting to milliseconds float
        AZ_Assert(ticksPerSecond >= 1000, "Error in converting ticks to ms, expected ticksPerSecond >= 1000");
        return static_cast<float>((ticks * 1000) / (ticksPerSecond / 1000)) / 1000.0f;
    }

    float ImGuiCpuProfiler::ToFloatMicroseconds(AZStd::sys_time_t ticks)
    {
        return AZStd::chrono::duration<float, AZStd::chrono::micro>(m_timeMicroseconds).count();
    }
```

Gems/Profiler/Code/Source/CpuProfiler.cpp
```diff
@@ -406,6 +407,7 @@ namespace Profiler
                 }
             }
         }
+        m_timeTicksPerSecond = AZStd ::GetTimeTicksPerSecond();
```
lumberyard-employee-dm, line 410: This should use [AZStd::chrono::steady_clock::now()](https://en.cppreference.com/w/cpp/chrono/steady_clock) to retrieve a monotonic clock value that isn't affected by the system clock.

moraaar (approved):
LGTM but I'm not familiarized with this code. Agree with @lumberyard-employee-dm comments.

Gems/Profiler/Code/Source/CpuProfiler.cpp
lumberyard-employee-dm: Because TimeTicksPeSecond is hard-coded value based on a certain frequency of time.
On Linux and Android this is just returns [1'000'000'000](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/Platform/Common/UnixLike/AzCore/std/time_UnixLike.cpp#L17-L19).
On Windows it is value of [QueryPerformanceFrequency](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/Platform/Windows/AzCore/std/time_Windows.cpp#L19)

There is no reason to store the value or even serialize the value out

Gems/Profiler/Code/Source/ImGuiCpuProfiler.cpp
lumberyard-employee-dm: This is meant to convert from integer microseconds to float milliseconds

Gems/Profiler/Code/Source/CpuProfiler.cpp
lumberyard-employee-dm: Well if the profiler works in ticks, then the existing data is already not portable.

But that does seem like a larger refactor of the entire, so I'll drop this request.

moudgils, line 410:
```suggestion
        m_timeTicksPerSecond = AZStd::GetTimeTicksPerSecond();
```

moudgils (approved):

lumberyard-employee-dm:
> Is there anything else to be done in this PR? Can it be be submitted? @lumberyard-employee-dm do you have more feedback?

I don't have anymore feedback for this PR.

PR #12552, opened 2022-10-12
Fix screenshot resizing issue

jiaweig-amzn:

ASV change:
https://github.com/o3de/o3de-atom-sampleviewer/pull/535

## What does this PR do?
The issue is: the screenshot tests (both dx12 and Vulkan on Windows) in ASV failed to match the size of the golden images, because Windows forced the window size to be within the screen rectangle, thus when the screen resolution is too small or scaling is too big, the screenshot will be clipped to a smaller size.

The solution is actually as simple as adding a flag SWP_NOSENDCHANGING when setting the new window pos. Without this flag, windows will send out WM_WINDOWPOSCHANGING message and then WM_GETMINMAXINFO message, so that the window size will be clipped to what the screen max size.
https://learn.microsoft.com/en-us/windows/win32/winmsg/wm-windowposchanging
https://learn.microsoft.com/en-us/windows/win32/winmsg/wm-getminmaxinfo

This fix can fix both dx12 and vulkan end. There was an attempt to fix it by forcing to send out window size change event and triggers swap chain recreation. However, this only works for dx12. Vulkan has the surface layer that is read only, and if we force the resizing to exceed the surface capability, it will trigger crash. 

## How was this PR tested?
Ran the full test suit in both dx12 and vulkan. Using a screen resolution at 1920 x 1080, scaled at 125%. The height limit will be clipped to 845 pixels before the change, this will fail at some test cases like material tests which sets the client size to 999 x 999. Now it can have the correct size.
Note that there are a couple of failures, they are all reported before this change and none of them is size mismatch
Vulkan
dx12


Code/Framework/AzFramework/AzFramework/Windowing/NativeWindow.h
```diff
@@ -175,7 +175,7 @@ namespace AzFramework
 
             virtual void SetWindowTitle(const AZStd::string& title);
             virtual WindowSize GetClientAreaSize() const;
-            virtual void ResizeClientArea(WindowSize clientAreaSize);
+            virtual void ResizeClientArea(WindowSize clientAreaSize, bool ignoreScreenSizeLimit);
```
lumberyard-employee-dm, line 178: Instead of updating this function to accept a bool, how about adding a structure for accepting window position parameters, so that the API of this function wouldn't have to change, if for some reason a another window position flag is needed.
```suggestion
            struct WindowPosOptions
            {
                bool m_ignoreScreenSizeLimit{};
            };
            virtual void ResizeClientArea(WindowSize clientAreaSize, const WindowPosOptions&);
```

That way this can still be called in an EBus as
```c++
AzFramework::WindowRequestBus::Event(windowHandle, &AzFramework::WindowRequestBus::Events::ResizeClientArea, clientAreaSize, WindowPosOptions{ ignoreScreenSizeLimit });
```

If for some reason, we need to add other options for Windows, Mac or Linux, only the handlers, that need to use the new options would need to be updated

lumberyard-employee-dm (approved):

rgba16f (approved):
I assume this leaves Linux and Mac as still broken? or were they unaffected by the issue?

jiaweig-amzn:
> I assume this leaves Linux and Mac as still broken? or were they unaffected by the issue?

The ticket only includes Windows platforms. I haven't looked into Linux and Mac. We can have a follow up ticket for that.
This PR will at least unblock the ASV nightly run.

Issue #12270, opened 2022-09-28
Editor Crash After Filtering Python Scripts


santorac:
**Describe the bug**
If you close the Python Scripts tool after filtering for a file, it will crash inside QT code.

**Assets required**
I tested with the AtomSampleViewer project. I suspect it will happen with other projects though.

**Steps to reproduce**
1. Run the Editor
2. Don't open any levels. Just close the "Welcome to O3DE" dialog.
3. Tools > Other > Python Scripts
4. Type "material" in the filter text
5. Expand the filtered tree to reveal some file
6. Close the dialog

**Expected behavior**
Clean exit from the dialog.

**Actual behavior**
The Editor crashes with the following callstack...
```
 	Qt5Guid.dll!00007ff946a08ba5()	Unknown
 	Qt5Guid.dll!00007ff946a07aef()	Unknown
 	Qt5Cored.dll!00007ff94b0b5a6e()	Unknown
 	Qt5Cored.dll!00007ff94b0f431c()	Unknown
 	Qt5Cored.dll!00007ff94b0f06a6()	Unknown
 	Qt5Cored.dll!00007ff94b0f1b6b()	Unknown
 	Qt5Cored.dll!00007ff94b0b5a6e()	Unknown
 	Qt5Cored.dll!00007ff94b0a518c()	Unknown
 	Qt5Cored.dll!00007ff94b0a0ee6()	Unknown
 	Qt5Cored.dll!00007ff94b0f7990()	Unknown
 	Qt5Cored.dll!00007ff94b0f7915()	Unknown
 	Qt5Cored.dll!00007ff94b0f4b9d()	Unknown
 	Qt5Cored.dll!00007ff94b0ee3d4()	Unknown
 	Qt5Cored.dll!00007ff94b1bb8ef()	Unknown
 	Qt5Cored.dll!00007ff94b1a7e77()	Unknown
 	Qt5Cored.dll!00007ff94b09f756()	Unknown
 	Qt5Guid.dll!00007ff946a0b632()	Unknown
 	Qt5Guid.dll!00007ff946a0512f()	Unknown
>	EditorLib.dll!CFolderTreeCtrl::CTreeItem::~CTreeItem() Line 479	C++
 	[External Code]	
 	EditorLib.dll!CFolderTreeCtrl::~CFolderTreeCtrl() Line 132	C++
 	[External Code]	
 	EditorLib.dll!CPythonScriptsDialog::~CPythonScriptsDialog() Line 115	C++
[...]

PR #12556, opened 2022-10-12
Clear search filter when closing dialog

sphrose:

## What does this PR do?
This PR fixes https://github.com/o3de/o3de/issues/12270. The crash only happens with a valid filter term, i.e. one that matches one or more items. Clearing the filter when the dialog is closing solves the problem.

## How was this PR tested?

Multiple runs with the filter in various states.


gadams3 (approved):

pollend:
Could you look at how the tree view is getting cleaned up. There is probably a better fix.

pollend (changes requested):
@sphrose something is really weird about this CTreeItem its attached itself to the invisiteRootItem for m_model but its also managed with a unique_ptr. There is some really strange ownership happening here. I think that is probably why it crashing its been freed by m_model but its also getting deleted by the unique_ptr. 

https://doc.qt.io/qt-6/qstandarditem.html#dtor.QStandardItem

sphrose:
> @sphrose something is really funny about CTreeItem its attached itself to the invisiteRootItem for m_model but its also managed with a unique_ptr. There is some really strange ownership happening here. I think that is probably why it crashing its been freed by m_model but its also getting deleted by the unique_ptr.
> 
> https://doc.qt.io/qt-6/qstandarditem.html#dtor.QStandardItem

@pollend That looks like exactly the problem, but it only happens when items are selected by the filter, which is all hidden away inside a QSortFilterProxyModel. A filter with no hits doesn't break, and no filter at all is fine. This leads me to believe that Qt is keeping a reference somewhere internally, although why, and why it should be deleting it is beyond me. I;ll add a comment explaining it.

pollend:
> > @sphrose something is really funny about CTreeItem its attached itself to the invisiteRootItem for m_model but its also managed with a unique_ptr. There is some really strange ownership happening here. I think that is probably why it crashing its been freed by m_model but its also getting deleted by the unique_ptr.
> > https://doc.qt.io/qt-6/qstandarditem.html#dtor.QStandardItem
> 
> @pollend That looks like exactly the problem, but it only happens when items are selected by the filter, which is all hidden away inside a QSortFilterProxyModel. A filter with no hits doesn't break, and no filter at all is fine. This leads me to believe that Qt is keeping a reference somewhere internally, although why, and why it should be deleting it is beyond me. I;ll add a comment explaining it.

I think it makes more sense to fix CTreeItem. might be sufficient to remove the unique_ptr since the root item is already owned under the invisible item in the model.

pollend (approved):

sphrose:
> > > @sphrose something is really funny about CTreeItem its attached itself to the invisiteRootItem for m_model but its also managed with a unique_ptr. There is some really strange ownership happening here. I think that is probably why it crashing its been freed by m_model but its also getting deleted by the unique_ptr.
> > > https://doc.qt.io/qt-6/qstandarditem.html#dtor.QStandardItem
> > 
> > 
> > @pollend That looks like exactly the problem, but it only happens when items are selected by the filter, which is all hidden away inside a QSortFilterProxyModel. A filter with no hits doesn't break, and no filter at all is fine. This leads me to believe that Qt is keeping a reference somewhere internally, although why, and why it should be deleting it is beyond me. I;ll add a comment explaining it.
> 
> I think it makes more sense to fix CTreeItem. might be sufficient to remove the unique_ptr since the root item is already owned under the invisible item in the model. clearing the filter might just mask the problem. Is this the only place this control is used?

Yes, it does seem to be the only place it's used. Your suggestion makes sense, I'll have a look.

pollend:
> > > > @sphrose something is really funny about CTreeItem its attached itself to the invisiteRootItem for m_model but its also managed with a unique_ptr. There is some really strange ownership happening here. I think that is probably why it crashing its been freed by m_model but its also getting deleted by the unique_ptr.
> > > > https://doc.qt.io/qt-6/qstandarditem.html#dtor.QStandardItem
> > > 
> > > 
> > > @pollend That looks like exactly the problem, but it only happens when items are selected by the filter, which is all hidden away inside a QSortFilterProxyModel. A filter with no hits doesn't break, and no filter at all is fine. This leads me to believe that Qt is keeping a reference somewhere internally, although why, and why it should be deleting it is beyond me. I;ll add a comment explaining it.
> > 
> > 
> > I think it makes more sense to fix CTreeItem. might be sufficient to remove the unique_ptr since the root item is already owned under the invisible item in the model. clearing the filter might just mask the problem. Is this the only place this control is used?
> 
> Yes, it does seem to be the only place it's used. Your suggestion makes sense, I'll have a look.

do you plan to work on a more permanent fix? I think this is good for now, will probably find another situation where this crashes.

PR #12559, opened 2022-10-12
Fix crash when running on Quest 2 with GPU detailed profiling mode enabled

moraaar:

## What does this PR do?

ASV RHI and RPI samples crashed when running on Quest2 with GPU detailed profiling mode enabled (`'adb shell ovrgpuprofiler -e`), which allows the use of `GPU systrace` and `ovrgpuprofiler` tools.

There were 2 crashes when GPU detailed profiling mode is enabled:
1. XR pipeline calling `LoadProcAddresses` passing the physical device leads to vulkan extensions enabled when they are not supported in an XR environment. To fix this issue the RHI::Device just uses the glad context returned from XR, that way only XR's device is responsible to do `LoadProcAddresses`.
2. With an XR pipeline it crashes when creating the default swapchain. Since the default swapchain isn't used anyway in non-host platforms when there is an xr system, now it's not created, saving memory and avoiding the crash (which was a segfault happing in the vulkan internals when creating the swapchain).

This PR goes in conjunction with this PR in o3de-extras: https://github.com/o3de/o3de-extras/pull/43

## How was this PR tested?

Run ASV RHI and RPI samples on Quest 2 with GPU detailed profiling mode enabled and disabled.
Able to capture render stage metrics with both `GPU systrace` and `ovrgpuprofiler` tools.

### DIFF
--- a/Gems/Atom/RHI/Vulkan/Code/Source/RHI/Device.cpp
+++ b/Gems/Atom/RHI/Vulkan/Code/Source/RHI/Device.cpp
@@ -271,6 +271,7 @@ namespace AZ
                 AZ::RHI::ResultCode result = xrSystem->CreateDevice(&xrDevicDescriptor);
                 AZ_Assert(result == RHI::ResultCode::Success, "Xr Vk device creation was not successful");
                 m_nativeDevice = xrDevicDescriptor.m_outputData.m_xrVkDevice;
+                m_context = xrDevicDescriptor.m_outputData.m_context;
                 RETURN_RESULT_IF_UNSUCCESSFUL(result);
                 m_isXrNativeDevice = true;
             }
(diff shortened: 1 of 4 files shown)


Gems/Atom/RHI/Vulkan/Code/Source/RHI/Device.cpp
```diff
@@ -280,20 +281,20 @@ namespace AZ
                     instance.GetContext().CreateDevice(physicalDevice.GetNativePhysicalDevice(), &deviceInfo, nullptr, &m_nativeDevice);
                 AssertSuccess(vkResult);
                 RETURN_RESULT_IF_UNSUCCESSFUL(ConvertResult(vkResult));
+
+                if (!instance.GetFunctionLoader().LoadProcAddresses(
+                        &m_context, instance.GetNativeInstance(), physicalDevice.GetNativePhysicalDevice(), m_nativeDevice))
+                {
+                    AZ_Warning("Vulkan", false, "Could not initialized function loader.");
```
amzn-phist, line 288: nit: initialized -> initialize

amzn-phist (approved):

Gems/Atom/RHI/Vulkan/Code/Source/RHI/SwapChain.cpp
```diff
@@ -117,8 +130,12 @@ namespace AZ
                 auto& presentationQueue = device.GetCommandQueueContext().GetOrCreatePresentationCommandQueue(*this);
                 m_presentationQueue = &presentationQueue;
 
-                result = CreateSwapchain();
-                RETURN_RESULT_IF_UNSUCCESSFUL(result);
+                if (IsDefaultSwapChainNeeded())
```
moudgils, line 133: I thought we added the check during pipeline creation time so why do we need this check again here. It should never come here if the default pipeline is never created for Android?

moraaar: It's not creating the default render pipeline (so it'll eventually won't render to the default swapchain), but the swapchains are always created by the WindowsContext, it's a different initialization path. After this change now it's doing both, not creating the default swapchain and not creating the default render pipeline.

moudgils (approved):

PR #12564, opened 2022-10-12
[Terrain] Add paint brush hardness, flow, distance controls

mbalfour-amzn:
## What does this PR do?

Added controls for hardness, flow, and distance to the paint brush tool. To get these to work correctly, I needed to overhaul a lot of the painting logic itself, and I also did a UX pass on all the options. The paint controls and behaviors should now be identical (or nearly identical) to Photoshop and Gimp.

This also includes a small bugfix to the Slider control so that it respects the "Suffix" attribute.

Here's an example showing all of the controls in action:
https://user-images.githubusercontent.com/82224783/195397096-a3cd3a8a-c0d2-43b3-8270-a91339a53c38.mp4

## How was this PR tested?

Manually validated all the control combinations in-Editor.


Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponentMode.cpp
```diff
@@ -61,26 +60,31 @@ namespace GradientSignal
...
         {
             uint32_t tileIndex = GetTileIndex(pixelIndex);
             uint32_t pixelTileIndex = GetPixelTileIndex(pixelIndex);
 
             AZ_Assert(m_paintedImageTiles[tileIndex], "Cached image tile hasn't been created yet!");
 
-            m_paintedImageTiles[tileIndex]->second[pixelTileIndex] = modifiedValue;
+            m_paintedImageTiles[tileIndex]->m_modifiedData[pixelTileIndex] = modifiedValue;
+            m_paintedImageTiles[tileIndex]->m_modifiedDataOpacity[pixelTileIndex] = opacity;
         }
```
invertednormal, line 88: nit: I'd do tile creation on set, not get. get could just return defaults if the tile didn't exist.

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
```diff
@@ -159,84 +190,150 @@ namespace AzToolsFramework
...
+                            // hardness falloff gives per-pixel opacity within the stamp.
+                            // For the falloff function, we use a nonlinear falloff that's approximately the same as a squared cosine:
+                            // 2x^3 - 3x^2 + 1
+                            float shortestDistanceNormalized = sqrt(shortestDistanceSquared) / radius;
+                            float hardnessDistance = AZStd::max(shortestDistanceNormalized - hardness, 0.0f) * inverseHardnessReciprocal;
+                            float curHardness = (2.0f * hardnessDistance * hardnessDistance * hardnessDistance) -
+                                (3.0f * hardnessDistance * hardnessDistance) + 1.0f;
+
+                            // For the opacity at this point, combine the opacity from previous stamps with the current hardness and flow.
+                            opacity = AZStd::min(opacity + (curHardness * flow), 1.0f);
```
invertednormal: If I understand this correctly, I don't think this is right. This also applies to 'OnPaint()' below (in EditorImageGradientComponentMode.cpp).

This appears to just be adding opacities and clamping at 1.0. So, if you set flow to 0.5 and the hardness was 1.0, any stamp that overlapped the previous would be 0.5 + 0.5 = 1.0, completely opaque. But it should be only 0.75 opaque. The math should be something like
```
float stampOpacity = curHardness * flow;
opacity = opacity + (1.0f - opacity) * stampOpacity;
```

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponentMode.cpp
```diff
@@ -473,21 +472,26 @@ namespace GradientSignal
...
         {
-            // Blend the pixel and store it back into our paint stroke buffer.
-            float gradientValue = m_paintStrokeBuffer->GetOriginalPixelValue(pixelIndices[index]);
-            float blendedValue = blendFn(gradientValue, intensities[index], opacities[index]);
-            m_paintStrokeBuffer->SetModifiedPixelValue(pixelIndices[index], blendedValue);
+            auto [gradientValue, opacityValue] = m_paintStrokeData.m_strokeBuffer->GetOriginalPixelValueAndOpacity(pixelIndices[index]);
 
-            // Store the blended value into a second buffer that we'll use to immediately modify the image gradient.
+            // Add the new per-pixel opacity to the existing opacity in our stroke layer.
+            opacityValue = AZStd::clamp(perPixelOpacities[index] + opacityValue, 0.0f, 1.0f);
```
invertednormal, line 484: See above comment about opacity.

invertednormal (approved):
Looks good - the only real issue I see is how you're blending opacity. That should be fixed (assuming I'm not misunderstanding something) but I'm also fine with that change happening in a different PR if it's inconvenient to do it in this one.

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
```diff
@@ -137,12 +151,29 @@ namespace AzToolsFramework
...
+        else if (mouseInteraction.m_mouseEvent == AzToolsFramework::ViewportInteraction::MouseEvent::DoubleClick)
+        {
+            if (mouseInteraction.m_mouseInteraction.m_mouseButtons.Left())
+            {
+                // Swallow up double-click events so that the default component mode behavior doesn't detect it and try
+                // to leave component mode. It's too easy to double-click by accident when painting, so double-clicking
+                // to exit just doesn't work well.
+                return true;
+            }
+        }
```
hultonha, line 175: question: If this is the case maybe we should do away with this behavior entirely now otherwise being inconsistent across Component Modes could be confusing. Maybe something to ask @o3de/sig-ui-ux about?

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
```diff
@@ -159,84 +190,162 @@ namespace AzToolsFramework
...
             {
-                m_previousCenter = m_center;
+                brushStamps.emplace_back(currentCenter2D);
+                m_lastBrushCenter = currentCenter2D;
+                m_distanceSinceLastDraw = 0.0f;
             }
 
-            // Get our current paint brush settings.
+            // Get the direction that we've moved the mouse since the last mouse movement we handled.
+            AZ::Vector2 direction = (currentCenter2D - m_lastBrushCenter).GetNormalized();
```
hultonha, line 237: question: There may be situations potentially where this are the same and normalizing the direction fails? Might we want to use the safe version or early out if `currentCenter2d-m_lastBrusCenter` is zero?

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
```diff
@@ -159,84 +190,162 @@ namespace AzToolsFramework
...
+                        // Since each stamp is a circle, we can just compare distance to the center of the circle vs radius.
+                        if (float shortestDistanceSquared = brushCenter.GetDistanceSq(point2D);
+                            shortestDistanceSquared <= manipulatorRadiusSq)
+                        {
+                            // It's a valid point, so calculate the opacity. The per-point opacity for a paint stamp is a combination
+                            // of the hardness falloff and the flow. The flow value gives the overall opacity for each stamp, and the
+                            // hardness falloff gives per-pixel opacity within the stamp.
+
+                            // Normalize the distance into the 0-1 range, where 0 is the center of the stamp, and 1 is the edge.
+                            float shortestDistanceNormalized = sqrt(shortestDistanceSquared) / radius;
```
hultonha, line 317: nit: `AZStd::sqrt`

Code/Framework/AzToolsFramework/AzToolsFramework/PaintBrushSettings/PaintBrushSettingsWindow.cpp
```diff
@@ -103,7 +88,7 @@ namespace PaintBrush
         // and this is only visible while in a painting component mode, so we don't ever need to disable the controls.
         viewOptions.isDisabledInComponentMode = false;
         // Default size of the window
-        viewOptions.paneRect = QRect(50, 50, 350, 145);
+        viewOptions.paneRect = QRect(50, 50, 350, 210);
```
hultonha, line 91: question: Maybe we could move this to an AZ_CVAR for easier tweaking if we expect it to change much more? Could use an AZ::Vector4 for it to store the dimensions potentially?

hultonha (approved):
LGTM! I just have a few minor comments, really cool to see the manipulator and component mode usage 🙂

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
```diff
@@ -159,84 +190,150 @@ namespace AzToolsFramework
...
+            // Get the direction that we've moved the mouse since the last mouse movement we handled.
+            AZ::Vector2 direction = (currentCenter2D - m_lastBrushCenter).GetNormalized();
+
+            // Get the total distance that we've moved since the last time we drew a brush stamp (which might
+            // have been many small mouse movements ago).
+            float totalDistance = m_lastBrushCenter.GetDistance(currentCenter2D) + m_distanceSinceLastDraw;
+
+            // Find the location for each brush stamp that we can draw based on the total distance the mouse moved since
+            // the last time we drew a stamp.
+            for (; totalDistance >= distanceBetweenBrushStamps; totalDistance -= distanceBetweenBrushStamps)
```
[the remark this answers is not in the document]
mbalfour-amzn: Expanded the comments to be more descriptive.

Code/Framework/AzToolsFramework/AzToolsFramework/PaintBrushSettings/PaintBrushSettings.h
```diff
@@ -14,6 +14,41 @@
...
+       closer to pointillism, where the stroke consists of many discrete brush "stamps", where a "stamp" is a single discrete imprint
+       of the brush shape.
+
+       With that description in mind, here's how the settings play together:
+
+       Stroke settings:
+       - Intensity: The color of the paint on the paintbrush. We only support monochromatic painting, so this is just a greyscale
+         intensity from black to white. This color is fully applied to each brush stroke.
+       - Opacity: The opacity of an entire brush stroke. Because it's per-stroke, a stroke that crosses itself won't blend with
+         itself. The blending only occurs after the stroke is complete.
```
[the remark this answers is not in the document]
mbalfour-amzn: Clarified the comment.

Code/Framework/AzToolsFramework/AzToolsFramework/PaintBrushSettings/PaintBrushSettings.h
```diff
@@ -40,43 +75,69 @@ namespace AzToolsFramework
...
+        void SetIntensityPercent(float intensityPercent);
+        void SetOpacityPercent(float opacityPercent);
         void SetBlendMode(PaintBrushBlendMode blendMode);
 
+        // Stamp settings
+
+        float GetSize() const
+        {
+            return m_size;
+        }
```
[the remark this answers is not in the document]
mbalfour-amzn: I'll skip this suggestion for now. I'm deliberately starting to push all mention of "radius" down into the implementation instead of the API at this point in the hopes that we'll get around to adding non-circular stamps at some point.

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponentMode.cpp
mbalfour-amzn: I should probably do lazy tile creates on _both_ for safety, but it needs to at least be on Get() for the code to work. The "unmodifiedData" in the tile is the cached original gradient values. If Get() just returned default values instead of the gradient values, then painting would erase the gradient instead of blend with it.

Code/Framework/AzToolsFramework/AzToolsFramework/PaintBrushSettings/PaintBrushSettingsWindow.cpp
mbalfour-amzn: I think a CVAR / AZ::Vector4 would be overkill for this. It's not something that needs to be arbitrarily tweaked, since it's just the initial pane size. Once a user adjusts it, the resized version is serialized out and these numbers never affect anything again without deleting a windows registry setting (I wish I was kidding). 

The only reason you're seeing these numbers change a lot in my PRs is because I keep adding new fields to the window as I add paintbrush functionality, so the default window size needs to keep growing. But there shouldn't be a need to adjust them when new fields _aren't_ being added.

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
mbalfour-amzn: Awesome catch, thx! I really thought these were additive in the paint programs, not blended. You're 100% correct. Changed!

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
mbalfour-amzn: I'll be discussing all of the recent paintbrush changes with UX later this afternoon, I'll make sure to ask about this as well.

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
mbalfour-amzn: There _shouldn't_ be a situation that can cause that, but I admit I was being too subtle and clever for my own good here. Basically, if the distance is 0, then the totalDistance will be < distanceBetweenBrushStamps, and the direction will never get used. So it _ought_ to be safe, but there aren't any guard rails here.

I'll change it to use GetNormalizedSafe() just to be 100% safe.

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
mbalfour-amzn: I asked about this in the sig-ui-ux office hours just now, and it sounds like there's _already_ been conversations about removing the default double-click-to-exit behavior. So as long as you're good with it, I'll put up a separate PR to remove that default behavior, along with my special-case override here, since it will no longer be necessary.

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
hultonha: Cool thanks!

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
hultonha: Sounds good to me thanks!

Issue #9309, opened 2022-04-29
AP has issues processing models with history and nGons. Test case assets attached.


shawstar:
**Describe the bug**
AP has issues processing models with history and nGons. History in FBX is from Maya DCCs tool. Will add a few test cases to this ticket. 

**Steps to reproduce**

**Test case One: SimpleSkinedMesh_with_history_and_ngons.fbx**
This is a .fbx from Maya, this is a cylinder that has bones skinned to it with animation. There is history in the mesh from taking the top of the cylender face of and replacing it will a fill, making the top an nGon.
When the o3de AP sees the mesh, it will import it without an issue:
![image](https://user-images.githubusercontent.com/87207603/165971227-edafa184-c621-4510-ad4e-666dcd2fad74.png)
The file is not usable in o3de, it is a dead file.
![image](https://user-images.githubusercontent.com/87207603/165971338-2c5527d5-1d5f-48bf-968b-66cab58fa580.png)

**Test case Two: SimpleSkinedMesh_with_history.fbx**
This is a .fbx from Maya, this is a cylinder that has bones skinned to it with animation. There is history in the mesh from taking the top of the cylender face of and replacing it will a fill, making the top an nGon, but now we select the nGon and Triangle the nGon.
**Test case Two: SimpleSkinedMesh_with_history.fbx**
![image](https://user-images.githubusercontent.com/87207603/165971773-834b9ae3-ffbc-4232-903b-d5794d7f7e0c.png)
The file is not usable in o3de, it is a dead file.
![image](https://user-images.githubusercontent.com/87207603/165972365-0689227e-b567-4913-9e9f-a57064a7f236.png)

**Test case Three: SimpleSkinedMesh_with_non_deform_history_removed.fbx**
This is a .fbx from Maya, this is a cylinder that has bones skinned to it with animation. There is history in the mesh from taking the top of the cylender face of and replacing it will a fill, making the top an nGon, but now we select the nGon and Triangle the nGon, then the non-deform history was removed. This resulted in the AP throwing a warning.
![image](https://user-images.githubusercontent.com/87207603/165973375-27540053-9b9b-420e-a828-4ce4b548fe38.png)
![image](https://user-images.githubusercontent.com/87207603/165973749-351262bb-b1d7-4c67-b7fb-dad0f736352f.png)
However this assets is actually usable in o3de
![image](https://user-images.githubusercontent.com/87207603/165974069-1a9b2b8f-926c-44e2-b00f-a3812ab453c4.png)
[...]

PR #12587, opened 2022-10-13
EMFX: Downgrade the AZ_Error to warnings when multiple mesh group is detected. 

amzn-rhhong:
Many asset are failing now due to the EMotionFX: Ambigious mesh for actor asset. More than one mesh group found. Defaulting to the first one.
EMotionFX had this as AZ_Error because atom should only ever produce one true mesh group. Now that Procedural Prefab system also creates mesh group, it is only making sense to downgrade to a AZ_Warning.

P.S this PR related to https://github.com/o3de/o3de/issues/9309 
P.S.2 Is there a pattern of when Procedural Prefab create additional mesh group and when not? This seems to happen on certain assets, but not all of them.

### DIFF
--- a/Gems/EMotionFX/Code/EMotionFX/Pipeline/RCExt/Actor/ActorGroupExporter.cpp
+++ b/Gems/EMotionFX/Code/EMotionFX/Pipeline/RCExt/Actor/ActorGroupExporter.cpp
@@ -173,7 +173,7 @@ namespace EMotionFX
             }
 
             // Default to the first mesh group until we get a way to choose it via the scene settings (ATOM-13590).
-            AZ_Warning("EMotionFX", atomModelAssets.size() <= 1, "Ambigious mesh for actor asset. More than one mesh group found. Defaulting to the first one.");
+            AZ_Warning("EMotionFX", atomModelAssets.size() <= 1, "Ambiguous mesh for actor asset. More than one mesh group found. Defaulting to the first one.");
             if (!atomModelAssets.empty())
             {
                 const AZ::SceneAPI::Events::ExportProduct& meshProduct = atomModelAssets[0];
--- a/Gems/EMotionFX/Code/EMotionFX/Pipeline/RCExt/Actor/ActorGroupExporter.cpp
+++ b/Gems/EMotionFX/Code/EMotionFX/Pipeline/RCExt/Actor/ActorGroupExporter.cpp
@@ -173,7 +173,7 @@ namespace EMotionFX
             }
 
             // Default to the first mesh group until we get a way to choose it via the scene settings (ATOM-13590).
-            AZ_Error("EMotionFX", atomModelAssets.size() <= 1, "Ambigious mesh for actor asset. More than one mesh group found. Defaulting to the first one.");
+            AZ_Warning("EMotionFX", atomModelAssets.size() <= 1, "Ambigious mesh for actor asset. More than one mesh group found. Defaulting to the first one.");
             if (!atomModelAssets.empty())
             {
                 const AZ::SceneAPI::Events::ExportProduct& meshProduct = atomModelAssets[0];


moraaar (commented):
[suggestion] I'd be ok with the change, but before approving I'd like to make a suggestion. 

My only concern with this change is that a warning most likely will be missed by the user and then ultimately they won't get the assets they expect and wonder why. This is a core issue of how actor groups work with mesh groups in Scene Settings and I'm not expecting to be fixed by this PR, but I wonder if it'd be better to keep is as an AZ_Error (so it's crystal clear when users do what they are not suppose to) and instead look into the procedural process that makes this situation happen and change it so it avoids making an actor with multiple meshes groups.

amzn-rhhong:
> [suggestion] I'd be ok with the change, but before approving I'd like to make a suggestion and get people's feedback.
> 
> My only concern with this change is that a warning most likely will be missed by the user and then ultimately they won't get the assets they expect and wonder why. This is a core issue of how actor groups work with mesh groups in Scene Settings and I'm not expecting to be fixed by this PR, but I wonder if it'd be better to keep it as an AZ_Error (so it's crystal clear when users do what they are not suppose to) and instead look into the procedural process that makes this situation happen and change it so it avoids making an actor with multiple mesh groups.

Proc prefab also create a mesh group for mesh node by default. There is no data error in this case to the user, I don't even think this should be a warning. I do wonder if we should export multiple actors (one for atom, one for proc prefab).

moraaar:
> > [suggestion] I'd be ok with the change, but before approving I'd like to make a suggestion and get people's feedback.
> > My only concern with this change is that a warning most likely will be missed by the user and then ultimately they won't get the assets they expect and wonder why. This is a core issue of how actor groups work with mesh groups in Scene Settings and I'm not expecting to be fixed by this PR, but I wonder if it'd be better to keep it as an AZ_Error (so it's crystal clear when users do what they are not suppose to) and instead look into the procedural process that makes this situation happen and change it so it avoids making an actor with multiple mesh groups.
> 
> Proc prefab also create a mesh group for mesh node by default. There is no data error in this case to the user, I don't even think this should be a warning because nothing is wrong with user's data. I do wonder if we should export multiple actors (one for atom, one for proc prefab).

Having multiple mesh groups with no actor groups is not a problem, it will just export out 1 static mesh per mesh group. The issue comes when there is one actor group and multiple mesh groups, which mesh group should the actor group use to get the mesh data from? Since there is no way at the moment to specify that in Scene Settings it was decided to always use the first mesh group and error if there were more than one. The user has to know that if they want to export actors there can be only 1 mesh group.

Scene Settings tool at the moment allows multiple mesh groups and multiple actor groups, but there is not way to say "actor group 1 wants to use mesh group 3 for its data. And actor group 2 wants to use mesh group 1 for its data". So my guess is that to avoid letting the user make a mistakes an AZ_Error was added to be clear.

lumberyard-employee-dm:
Honestly my thoughts on using `AZ_Error` in general is that instead of showing polluting the trace system with an error message that the user might ignore or a message that might not be relevant, it would be better to return an `AZ::Outcome` instead with the error type being the failure message with a type of `AZStd::fixed_string<N>` so that dynamic memory allocations don't have to be for.

Gems/EMotionFX/Code/EMotionFX/Pipeline/RCExt/Actor/ActorGroupExporter.cpp
```diff
@@ -173,7 +173,7 @@ namespace EMotionFX
             }
 
             // Default to the first mesh group until we get a way to choose it via the scene settings (ATOM-13590).
-            AZ_Error("EMotionFX", atomModelAssets.size() <= 1, "Ambigious mesh for actor asset. More than one mesh group found. Defaulting to the first one.");
+            AZ_Warning("EMotionFX", atomModelAssets.size() <= 1, "Ambigious mesh for actor asset. More than one mesh group found. Defaulting to the first one.");
```
lmbr-pip, line 176: Consider logging number of mesh groups?

Also ATOM-13590 reference in comment is stale, need to link to actual GHI or public roadmap item.

lmbr-pip (approved):

Gems/EMotionFX/Code/EMotionFX/Pipeline/RCExt/Actor/ActorGroupExporter.cpp
lmbr-pip, line 176: nit: consider logging number of mesh groups? Also ATOM-13590 is a stale issue reference, consider updating to GitHub issue
```suggestion
            AZ_Warning("EMotionFX", atomModelAssets.size() <= 1, "Ambiguous mesh for actor asset. More than one mesh group found. Defaulting to the first one.");
```

amzn-tommy, line 176: Part of the problem is that the code is not actually checking for how many mesh groups there are any more. It's checking for how many .azmodels there are. It used to be there was a 1:1 correlation between mesh groups and azmodels, so the error about "more than one mesh group found" was accurate. Now, with procedural prefabs, counting the # of product models is no longer a valid way to count the # of mesh groups.

Is there a way at this point in the pipeline to still check the scene settings for the asset? That way you can get the scene settings, check the number of mesh groups, and have an error when there is actually more than one mesh group (which is ambiguous), but have no error at all with 1 mesh group and multiple product azmodels (which is valid, and not (as) ambiguous since the model you want is the only non-procedural prefab model that exists.

It's also worth verifying that the logic here of choosing the 'first' product is actually going to give you the atom model associated with the mesh group in the scene settings, and not one of the atom models that come from procedural prefabs.

AMZN-stankowi:
I think it makes sense to downgrade this from an error to a warning. A warning makes sense when there's a fallback that will be OK and generate useful data often. An error should be used when data will be lost, or there isn't a good fallback.

> My only concern with this change is that a warning most likely will be missed by the user

We've been doing a lot of work to improve how warning / error messages are surfaced and displayed in the Asset Processor, as well as the Scene Settings UI. We're not done, though, there's a lot more work to be done in that space.

It's unfortunately very common for teams to ignore warnings in their project. I think the use in that case is, if something starts going wrong, then they'll check the logs, and a warning is a useful hint at resolving the issue.

I also want to find some time to do a deeper revamp of the scene settings UI: Make it more obvious what's happened and why, what the defaults are, how settings relate to each other.

> With this said I think we can do much better to show this behavior to the user than just throwing out an AZ_Error or a AZ_Warning. I'd be in favor of removing the error entirely and instead display a message in the Actor tab in Scene Settings that say something in the lines of "the mesh data for actor groups will be taken from the first mesh group". That way the user will just know what's happening.

I think that's a great idea in addition to the error/warning in asset processor, but I think we'd still want a message in the asset processor. Not all content creators will load up the scene settings UI.

> Scene Settings tool at the moment allows multiple mesh groups and multiple actor groups, but there is not way to say "actor group 1 wants to use mesh group 3 for its data. And actor group 2 wants to use mesh group 1 for its data". So my guess is that to avoid letting the user make a mistakes an AZ_Error was added to be clear.

Is there a ticket in a backlog for this?

> Part of the problem is that the code is not actually checking for how many mesh groups there are any more. It's checking for how many .azmodels there are. It used to be there was a 1:1 correlation between mesh groups and azmodels, so the error about "more than one mesh group found" was accurate. Now, with procedural prefabs, counting the # of product models is no longer a valid way to count the # of mesh groups.

Aren't procedural prefabs using mesh groups to generate the additional stand alone azmodels?

If there are multiple mesh group, in other situations we're checking for a mesh group that has a filename that matches the source scene file and using that as a default.

AMZN-stankowi (approved):
I know the typo was already there, but it was bothering me seeing this error message in a bunch of places with the typo that lmbr-pip pointed out, so I'd really like to see that fixed.

Gems/EMotionFX/Code/EMotionFX/Pipeline/RCExt/Actor/ActorGroupExporter.cpp
amzn-rhhong: Is there a way to check if the atom model is created by the proc prefab?

moraaar:
>>Scene Settings tool at the moment allows multiple mesh groups and multiple actor groups, but there is not way to say "actor group 1 wants to use mesh group 3 for its data. And actor group 2 wants to use mesh group 1 for its data". So my guess is that to avoid letting the user make a mistakes an AZ_Error was added to be clear.
>
>Is there a ticket in a backlog for this?

Not that I'm aware of. @amzn-rhhong, @hultonha, it might be a good idea to add a ticket for this (if there isn't one already).

hultonha:
Good call @moraaar, I'll follow-up with @amzn-rhhong about this tomorrow :+1:

PR #12594, opened 2022-10-13
GraphCanvas node palette in EMotionFX node palette dock

### LINKED ISSUE #11691: Port Node Palette in EMotionFX to `GraphCanvas::NodePaletteWidget`
In [10963](https://github.com/o3de/o3de/issues/10963) the node palette dock widget was hidden by default. However, It wasn't migrated to the common `GraphCanvas::NodePaletteWidget` like the context menu. It stayed as a custom tree view without a search box.

Image

To achieve similar UX across all tools that use nodes palette, it should be replaced with a solution that uses `GraphCanvas::NodePaletteWidget`

Likely, the classes developed in #10963 (`BlendGraphMimeEvent`, `BlendGraphNodePaletteTreeItem`) can be re-used for this view.

The class is located in `Gems\EMotionFX\Code\EMotionFX\Tools\EMotionStudio\Plugins\StandardPlugins\Source\AnimGraph\NodePaletteWidget.{h,cpp}`

MiKom:
## What does this PR do?

This PR migrates node palette dock (`EMStudio::NodePaletteWidget`) from custom solution to `GraphCanvas::NodePaletteWidget`.

Also, common code for building the tree model for the node palette in EMFX context menu and dock was extracted to a common helper class.

A bugfix for `GraphCanvas::NodePaletteAutoCompleteModel` also was discovered and fixed.

I encourage reviewing commit by commit.

Here's how it looks:
https://user-images.githubusercontent.com/104767/195676131-92acd1ab-137f-4fc2-bd71-50156a27e4f5.mp4
Note that my screen recorder didn't grab the nice pixmap that is following the cursor during the drag.

## Failing tests (Fixed)

~~I'm putting this PR as a draft because I don't know how to fix memory allocation errors related to the new code:~~
The tests are fixed now
```
[ RUN      ] UIFixture.CanAddAnimGraphNode
[  MEMORY  ] Allocation Name: "BlendGraphMimeEvent" Addr: 000000191461DE4D0 Size: 80 Alignment: 8
[  MEMORY  ]  C:\kdab\dev\amazon\code\o3de\Gems\GraphCanvas\Code\StaticLib\GraphCanvas\Widgets\NodePalette\NodePaletteWidget.cpp (748) : GraphCanvas::NodePaletteWidget::HandleSelectedItem
[  MEMORY  ]  C:\kdab\dev\amazon\code\o3de\Gems\GraphCanvas\Code\StaticLib\GraphCanvas\Widgets\NodePalette\NodePaletteWidget.cpp (598) : GraphCanvas::NodePaletteWidget::OnSelectionChanged
[  MEMORY  ]  00007FF9BDDE09AA (Qt5Core) : QObject::qt_static_metacall
[  MEMORY  ]  00007FF9BDD8EA81 (Qt5Core) : QItemSelectionModel::emitSelectionChanged
[  MEMORY  ]  00007FF9BDD93B36 (Qt5Core) : QItemSelectionModel::select
[  MEMORY  ] Allocation Name: "BlendGraphMimeEvent" Addr: 000000191461DE520 Size: 80 Alignment: 8
[  MEMORY  ]  C:\kdab\dev\amazon\code\o3de\Gems\GraphCanvas\Code\StaticLib\GraphCanvas\Widgets\NodePalette\NodePaletteWidget.cpp (748) : GraphCanvas::NodePaletteWidget::HandleSelectedItem
[  MEMORY  ]  C:\kdab\dev\amazon\code\o3de\Gems\GraphCanvas\Code\StaticLib\GraphCanvas\Widgets\NodePalette\NodePaletteWidget.cpp (598) : GraphCanvas::NodePaletteWidget::OnSelectionChanged
[  MEMORY  ]  00007FF9BDDE09AA (Qt5Core) : QObject::qt_static_metacall
[  MEMORY  ]  00007FF9BDD8EC20 (Qt5Core) : QItemSelectionModel::emitSelectionChanged
[  MEMORY  ]  00007FF9BDD93B36 (Qt5Core) : QItemSelectionModel::select
[  MEMORY  ] Allocation Name: "AZStd::allocator" Addr: 00000019147911540 Size: 48 Alignment: 1
[  MEMORY  ]  C:\kdab\dev\amazon\code\o3de\Code\Framework\AzCore\AzCore\std\string\string.h (396) : AZStd::basic_string<char,AZStd::char_traits<char>,AZStd::allocator>::assign
[  MEMORY  ]  C:\kdab\dev\amazon\code\o3de\Gems\EMotionFX\Code\EMotionFX\Tools\EMotionStudio\Plugins\StandardPlugins\Source\AnimGraph\BlendGraphWidget.cpp (130) : EMStudio::BlendGraphNodePaletteTreeItem::CreateMimeEvent
[  MEMORY  ]  C:\kdab\dev\amazon\code\o3de\Gems\GraphCanvas\Code\StaticLib\GraphCanvas\Widgets\NodePalette\NodePaletteWidget.cpp (748) : GraphCanvas::NodePaletteWidget::HandleSelectedItem
[  MEMORY  ]  C:\kdab\dev\amazon\code\o3de\Gems\GraphCanvas\Code\StaticLib\GraphCanvas\Widgets\NodePalette\NodePaletteWidget.cpp (598) : GraphCanvas::NodePaletteWidget::OnSelectionChanged
[  MEMORY  ]  00007FF9BDDE09AA (Qt5Core) : QObject::qt_static_metacall
[  MEMORY  ] Allocation Name: "AZStd::allocator" Addr: 00000019147911600 Size: 48 Alignment: 1
[  MEMORY  ]  C:\kdab\dev\amazon\code\o3de\Code\Framework\AzCore\AzCore\std\string\string.h (396) : AZStd::basic_string<char,AZStd::char_traits<char>,AZStd::allocator>::assign
[  MEMORY  ]  C:\kdab\dev\amazon\code\o3de\Gems\EMotionFX\Code\EMotionFX\Tools\EMotionStudio\Plugins\StandardPlugins\Source\AnimGraph\BlendGraphWidget.cpp (130) : EMStudio::BlendGraphNodePaletteTreeItem::CreateMimeEvent
[  MEMORY  ]  C:\kdab\dev\amazon\code\o3de\Gems\GraphCanvas\Code\StaticLib\GraphCanvas\Widgets\NodePalette\NodePaletteWidget.cpp (748) : GraphCanvas::NodePaletteWidget::HandleSelectedItem
[  MEMORY  ]  C:\kdab\dev\amazon\code\o3de\Gems\GraphCanvas\Code\StaticLib\GraphCanvas\Widgets\NodePalette\NodePaletteWidget.cpp (598) : GraphCanvas::NodePaletteWidget::OnSelectionChanged
[  MEMORY  ]  00007FF9BDDE09AA (Qt5Core) : QObject::qt_static_metacall
C:\kdab\dev\amazon\code\o3de\Code\Framework\AzCore\AzCore/UnitTest/UnitTest.h(175): error: We still have 4 allocations on record! They must be freed prior to destroy!

[  FAILED  ] UIFixture.CanAddAnimGraphNode (1032 ms)
````

As you can see, these tests complain about allocation of `BlendGraphMimeEvent` which happens only in `BlendGraphNodePaletteTreeItem::CreateMimeEvent`. I do it similar to other classes in the engine (e.g. `ScriptEventsHandlerEventPaletteTreeItem::CreateMimeEvent`). This method is called by the internals of `GraphCanvasTreeModel` and the result is serialized. I don't have control of it from the tree item side:

```cpp
    BlendGraphMimeEvent* BlendGraphNodePaletteTreeItem::CreateMimeEvent() const
    {
        return IsEnabled() ? new BlendGraphMimeEvent(FromQtString(m_typeString), FromQtString(GetName())) : nullptr;
    }
```
I couldn't find a working combination of `aznew` and `new` that would make the tests happy. Help appreciated.

## How was this PR tested?

I checked the functionality of the node palette and context menu manually, also, apart from failing memory management requirements, the unit tests run functionally correct.

Fixes #11691

### DIFF
--- a/Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/BlendGraphWidget.cpp
+++ b/Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/BlendGraphWidget.cpp
@@ -35,6 +35,7 @@
 #include <MysticQt/Source/KeyboardShortcutManager.h>
 #include <MCore/Source/LogManager.h>
 #include <GraphCanvas/Widgets/GraphCanvasTreeItem.h>
+#include <GraphCanvas/Widgets/GraphCanvasMimeContainer.h>
 #include <GraphCanvas/Widgets/NodePalette/TreeItems/NodePaletteTreeItem.h>
 // qt includes
 #include <QDropEvent>
@@ -46,8 +47,20 @@
 
 namespace EMStudio
 {
-    BlendGraphMimeEvent::BlendGraphMimeEvent(QString typeString, QString namePrefix)
-        : GraphCanvas::GraphCanvasMimeEvent(), m_typeString(typeString), m_namePrefix(namePrefix)
+    void BlendGraphMimeEvent::Reflect(AZ::ReflectContext* context)
+    {
+        if (auto serializeContext = azrtti_cast<AZ::SerializeContext*>(context))
+        {
+            serializeContext->Class<BlendGraphMimeEvent, GraphCanvas::GraphCanvasMimeEvent>()
+                ->Version(0)
+                ->Field("typeString", &BlendGraphMimeEvent::m_typeString)
+                ->Field("namePrefix", &BlendGraphMimeEvent::m_namePrefix);
+        }
+    }
+    BlendGraphMimeEvent::BlendGraphMimeEvent(AZStd::string_view typeString, AZStd::string_view namePrefix)
+        : GraphCanvas::GraphCanvasMimeEvent()
+        , m_typeString(typeString)
+        , m_namePrefix(namePrefix)
     {
     }
 
@@ -62,19 +75,19 @@ namespace EMStudio
         return false;
     }
 
-    QString BlendGraphMimeEvent::GetTypeString() const
+    AZStd::string BlendGraphMimeEvent::GetTypeString() const
     {
         return m_typeString;
     }
 
-    QString BlendGraphMimeEvent::GetNamePrefix() const
+    AZStd::string BlendGraphMimeEvent::GetNamePrefix() const
     {
         return m_namePrefix;
     }
 
     BlendGraphNodePaletteTreeItem::BlendGraphNodePaletteTreeItem(
-        AZStd::string_view name, const QString& typeString, GraphCanvas::EditorId editorId, const AZ::Color& color)
-        : GraphCanvas::IconDecoratedNodePaletteTreeItem(name, editorId)
+        const AZStd::string_view name, const QString& typeString, GraphCanvas::EditorId editorId, const AZ::Color& color)
+        : GraphCanvas::DraggableNodePaletteTreeItem(name, editorId)
         , m_typeString(typeString)
     {
         // Draw a pixmap with the provided color to use as an icon adjacent to the name text
@@ -109,12 +122,12 @@ namespace EMStudio
         {
             return QVariant(m_colorPixmap);
         }
-        return NodePaletteTreeItem::OnData(index, role);
+        return GraphCanvas::DraggableNodePaletteTreeItem::OnData(index, role);
     }
 
     BlendGraphMimeEvent* BlendGraphNodePaletteTreeItem::CreateMimeEvent() const
     {
-        return IsEnabled() ? new BlendGraphMimeEvent(m_typeString, GetName()) : nullptr;
+        return IsEnabled() ? new BlendGraphMimeEvent(FromQtString(m_typeString), FromQtString(GetName())) : nullptr;
     }
 
     // constructor
@@ -155,6 +168,30 @@ namespace EMStudio
             return;
         }
 
+        // calculate the drop position
+        QPoint localPos = LocalToGlobal(event->pos());
+
+        // Dragged from node palette
+        if (event->mimeData()->hasFormat(BlendGraphMimeEvent::BlendGraphMimeEventType))
+        {
+            GraphCanvas::GraphCanvasMimeContainer mimeContainer;
+            const QByteArray mimeDataBuffer = event->mimeData()->data(BlendGraphMimeEvent::BlendGraphMimeEventType);
+            mimeContainer.FromBuffer(mimeDataBuffer.constData(), mimeDataBuffer.size());
+
+            // When adding multiple nodes together, position them a bit differently
+            // so they don't end up on top of each other looking like only one was added
+            int offset = 0;
+            for (GraphCanvas::GraphCanvasMimeEvent* graphCanvasEvent : mimeContainer.m_mimeEvents)
+            {
+                if (auto* blendGraphEvent = azrtti_cast<BlendGraphMimeEvent*>(graphCanvasEvent))
+                {
+                    const QPoint nudgedPosition = { localPos.x(), localPos.y() + offset };
+                    offset += 10;
+                    CreateNodeFromMimeEvent(blendGraphEvent, nudgedPosition);
+                }
+            }
+        }
+
         // only accept copy actions
         if (event->dropAction() != Qt::CopyAction || event->mimeData()->hasText() == false)
         {
@@ -166,8 +203,6 @@ namespace EMStudio
         AZStd::string dropText = FromQtString(event->mimeData()->text());
         MCore::CommandLine commandLine(dropText.c_str());
 
-        // calculate the drop position
-        QPoint offset = LocalToGlobal(event->pos());
         QModelIndex targetModelIndex;
         if (GetActiveGraph())
         {
@@ -207,10 +242,10 @@ namespace EMStudio
                         if (targetModelIndex.isValid())
                         {
                             EMotionFX::AnimGraphNode* currentNode = targetModelIndex.data(AnimGraphModel::ROLE_NODE_POINTER).value<EMotionFX::AnimGraphNode*>();
-                            CommandSystem::CreateAnimGraphNode(&commandGroup, currentNode->GetAnimGraph(), AZ::Uuid::CreateName("BlendTreeMotionNode"), "Motion", currentNode, offset.x(), offset.y(), serializedMotionNode.GetValue());
+                            CommandSystem::CreateAnimGraphNode(&commandGroup, currentNode->GetAnimGraph(), AZ::Uuid::CreateName("BlendTreeMotionNode"), "Motion", currentNode, localPos.x(), localPos.y(), serializedMotionNode.GetValue());
 
                             // setup the offset for the next motion
-                            offset.setY(offset.y() + 60);
+                            localPos.setY(localPos.y() + 60);
                         }
                     }
                 }
@@ -278,10 +313,10 @@ namespace EMStudio
                         if (serializedContent.IsSuccess())
                         {
                             EMotionFX::AnimGraphNode* currentNode = targetModelIndex.data(AnimGraphModel::ROLE_NODE_POINTER).value<EMotionFX::AnimGraphNode*>();
-                            CommandSystem::CreateAnimGraphNode(&commandGroup, currentNode->GetAnimGraph(), azrtti_typeid<EMotionFX::AnimGraphMotionNode>(), "Motion", currentNode, offset.x(), offset.y(), serializedContent.GetValue());
+                            CommandSystem::CreateAnimGraphNode(&commandGroup, currentNode->GetAnimGraph(), azrtti_typeid<EMotionFX::AnimGraphMotionNode>(), "Motion", currentNode, localPos.x(), localPos.y(), serializedContent.GetValue());
 
                             // setup the offset for the next motion
-                            offset.setY(offset.y() + 60);
+                            localPos.setY(localPos.y() + 60);
                         }
                     }
                     else
@@ -316,43 +351,18 @@ namespace EMStudio
                         AZ::Outcome<AZStd::string> serializedMotionNode = MCore::ReflectionSerializer::Serialize(&tempMotionNode);
                         if (serializedMotionNode.IsSuccess())
                         {
-                            EMotionFX::AnimGraphNode* currentNode = targetModelIndex.data(AnimGraphModel::ROLE_NODE_POINTER).value<EMotionFX::AnimGraphNode*>();
-                            CommandSystem::CreateAnimGraphNode(&commandGroup, currentNode->GetAnimGraph(), azrtti_typeid<EMotionFX::AnimGraphMotionNode>(), "Motion", currentNode, offset.x(), offset.y(), serializedMotionNode.GetValue());
+                            EMotionFX::AnimGraphNode* currentNode = targetModelIndex.data(AnimGraphModel::ROLE_NODE_POINTER)
+                                .value<EMotionFX::AnimGraphNode*>();
+                            CommandSystem::CreateAnimGraphNode(&commandGroup, currentNode->GetAnimGraph(),
+                                azrtti_typeid<EMotionFX::AnimGraphMotionNode>(), "Motion", currentNode, localPos.x(), localPos.y(),
+                                serializedMotionNode.GetValue());
 
                             // setup the offset for the next motion
-                            offset.setY(offset.y() + 60);
+                            localPos.setY(localPos.y() + 60);
                         }
                     }
                 }
             }
-        }
-        // default handling, we're drag & dropping from the palette window
-        else
-        {
-            AZStd::vector<AZStd::string> parts;
-            AzFramework::StringFunc::Tokenize(dropText.c_str(), parts, ";", false, true);
-            if (parts.size() != 3)
-            {
-                MCore::LogError("BlendGraphWidget::dropEvent() - Incorrect syntax using drop data '%s'", FromQtString(event->mimeData()->text()).c_str());
-                event->ignore();
-                return;
-            }
-
-            AZStd::string commandString;
-            AZStd::string resultString;
-            EMotionFX::AnimGraphNode* currentNode = targetModelIndex.data(AnimGraphModel::ROLE_NODE_POINTER).value<EMotionFX::AnimGraphNode*>();
-            if (!currentNode)
-            {
-                event->ignore();
-                return;
-            }
-
-            // build the name prefix
-            AZStd::string& namePrefix = parts[2];
-            AzFramework::StringFunc::Strip(namePrefix, MCore::CharacterConstants::space, true /* case sensitive */);
-
-            const AZ::TypeId typeId = AZ::TypeId::CreateString(parts[1].c_str(), parts[1].size());
-            CommandSystem::CreateAnimGraphNode(&commandGroup, currentNode->GetAnimGraph(), typeId, namePrefix, currentNode, offset.x(), offset.y());
         }
 
         if (!commandGroup.IsEmpty())
@@ -368,8 +378,15 @@ namespace EMStudio
     }
 
 
-    bool BlendGraphWidget::OnEnterDropEvent(QDragEnterEvent* event, EMotionFX::AnimGraphNode* currentNode, NodeGraph* activeGraph)
+    bool BlendGraphWidget::OnEnterDropEvent(QDragEnterEvent* event, EMotionFX::AnimGraphNode* currentNode)
     {
+        if (event->mimeData()->hasFormat(BlendGraphMimeEvent::BlendGraphMimeEventType))
+        {
+            // If it's coming from node palette it should be legal to add because node palette
+            // filters for only the permitted ones
+            return true;
+        }
+
         if (event->mimeData()->hasText() == false)
         {
             return false;
@@ -395,69 +412,7 @@ namespace EMStudio
             {
                 return true;
             }
-
-            return false;
         }
-
-        AZStd::vector<AZStd::string> parts;
-        AzFramework::StringFunc::Tokenize(dropText.c_str(), parts, MCore::CharacterConstants::semiColon, true /* keep empty strings */, true /* keep space strings */);
-
-        if (parts.size() != 3)
-        {
-            //MCore::LogError("BlendGraphWidget::dropEvent() - Incorrect syntax using drop data '%s'", FromQtString(event->mimeData()->text()).AsChar());
-            return false;
-        }
-
-        if (parts[0].find("EMotionFX::AnimGraphNode") != 0)
-        {
-            return false;
-        }
-
-        // check if the dropped node is a state machine node
-        bool canActAsState = false;
-        const AZ::TypeId typeId = AZ::TypeId::CreateString(parts[1].c_str(), parts[1].size());
-        if (!typeId.IsNull())
-        {
-            EMotionFX::AnimGraphObject* registeredObject = EMotionFX::AnimGraphObjectFactory::Create(typeId);
-            MCORE_ASSERT(azrtti_istypeof<EMotionFX::AnimGraphNode>(registeredObject));
-
-            EMotionFX::AnimGraphNode* registeredNode = static_cast<EMotionFX::AnimGraphNode*>(registeredObject);
-            if (registeredNode->GetCanActAsState())
-            {
-                canActAsState = true;
-            }
-
-            delete registeredObject;
-        }
-
-        // in case the current node is nullptr and the active graph is a valid graph it means we are showing the root graph
-        if (currentNode == nullptr)
-        {
-            if (activeGraph)
-            {
-                // only state machine nodes are allowed in the root graph
-                if (canActAsState)
-                {
-                    return true;
-                }
-            }
-        }
-        else
-        {
-            // check if we need to prevent dropping of non-state machine nodes
-            if (azrtti_typeid(currentNode) == azrtti_typeid<EMotionFX::AnimGraphStateMachine>())
-            {
-                if (canActAsState)
-                {
-                    return true;
-                }
-            }
-            else
-            {
-                return true;
-            }
-        }
-
         return false;
     }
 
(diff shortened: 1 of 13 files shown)


Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/BlendGraphWidget.cpp
```diff
@@ -155,6 +168,30 @@ namespace EMStudio
...
         }
 
+        // calculate the drop position
+        QPoint localPos = LocalToGlobal(event->pos());
+
+        // Dragged from node palette
+        if (event->mimeData()->hasFormat(BlendGraphMimeEvent::BlendGraphMimeEventType))
+        {
+            GraphCanvas::GraphCanvasMimeContainer mimeContainer;
+            const QByteArray ba = event->mimeData()->data(BlendGraphMimeEvent::BlendGraphMimeEventType);
```
hultonha, line 178: uber nit: We try to avoid single or double letter variable names so a name more inline with what the variable is rather than the type itself would be great, so perhap `blendGraphMimeData` or something similar?

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/BlendGraphWidget.cpp
```diff
@@ -317,43 +352,15 @@ namespace EMStudio
                         if (serializedMotionNode.IsSuccess())
                         {
                             EMotionFX::AnimGraphNode* currentNode = targetModelIndex.data(AnimGraphModel::ROLE_NODE_POINTER).value<EMotionFX::AnimGraphNode*>();
-                            CommandSystem::CreateAnimGraphNode(&commandGroup, currentNode->GetAnimGraph(), azrtti_typeid<EMotionFX::AnimGraphMotionNode>(), "Motion", currentNode, offset.x(), offset.y(), serializedMotionNode.GetValue());
+                            CommandSystem::CreateAnimGraphNode(&commandGroup, currentNode->GetAnimGraph(), azrtti_typeid<EMotionFX::AnimGraphMotionNode>(), "Motion", currentNode, localPos.x(), localPos.y(), serializedMotionNode.GetValue());
```
hultonha, line 355: nit: Pretty minor but while we're here can we shorten the length of this line (if you select it in VS and run `Ctrl-K, F` it'll format just the selected text (line limit is supposed to be 140 characters now). (repeated in a couple of places). Thanks!

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/BlendGraphWidget.h
```diff
@@ -37,20 +37,29 @@ namespace EMStudio
     class BlendGraphMimeEvent : public GraphCanvas::GraphCanvasMimeEvent
     {
     public:
-        explicit BlendGraphMimeEvent(QString typeString, QString namePrefix);
+        AZ_RTTI(BlendGraphMimeEvent, "{AA7C8960-C7BA-4F26-B52B-97CF4AC3CB39}", GraphCanvas::GraphCanvasMimeEvent);
+        AZ_CLASS_ALLOCATOR(BlendGraphMimeEvent, AZ::SystemAllocator, 0);
+        static void Reflect(AZ::ReflectContext* context);
+
+        BlendGraphMimeEvent() = default;
+        explicit BlendGraphMimeEvent(const AZStd::string_view typeString, const AZStd::string_view namePrefix);
```
hultonha, line 45: nit: We can drop the explicit here as we have two arguments being passed to the constructor making implicit conversions a lot less likely

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/BlendGraphWidget.h
```diff
@@ -37,20 +37,29 @@ namespace EMStudio
...
+        AZStd::string m_typeString;
+        AZStd::string m_namePrefix;
     };
 
-    class BlendGraphNodePaletteTreeItem : public GraphCanvas::IconDecoratedNodePaletteTreeItem
+    class BlendGraphNodePaletteTreeItem : public GraphCanvas::DraggableNodePaletteTreeItem
     {
     public:
-        BlendGraphNodePaletteTreeItem(AZStd::string_view name, const QString& typeString, GraphCanvas::EditorId editorId, const AZ::Color& color);
+        BlendGraphNodePaletteTreeItem(const AZStd::string_view name, const QString& typeString, GraphCanvas::EditorId editorId, const AZ::Color& color);
```
hultonha, line 62: suggestion: We don't usually add `const` to the header/interface when the type is pass by value as it doesn't add anything and is redundant. It's fine in the .cpp/implementation to indicate to the reader (and compiler) the value won't change but here it can be removed

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/ContextMenu.cpp
```diff
@@ -185,35 +153,22 @@ namespace EMStudio
...
 
                     auto* action = new QWidgetAction(menu);
-                    auto* rootItem = new GraphCanvas::NodePaletteTreeItem("root", AnimGraphEditorId);
+
+                    NodePaletteModelHelper modelHelper{ plugin };
+                    modelHelper.InitForNode(currentNode);
+
                     GraphCanvas::NodePaletteConfig config;
-                    config.m_rootTreeItem = rootItem;
+                    config.m_rootTreeItem = modelHelper.GetRootItem();
```
hultonha, line 165: question: Is there a potential lifetime issue here when `modelHelper` goes out of scope as we're taking a reference to `modelHelper.GetRootItem()` below and it is a local variable? It might be fine but I just wanted to double check

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/NodePaletteModelHelper.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include "NodePaletteModelHelper.h"
+#include <EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/AnimGraphPlugin.h>
+#include <EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/BlendGraphWidget.h>
+#include <EMotionFX/Source/AnimGraphObjectFactory.h>
+#include <GraphCanvas/Widgets/NodePalette/TreeItems/NodePaletteTreeItem.h>
+
+namespace EMStudio
+{
+    NodePaletteModelHelper::NodePaletteModelHelper(AnimGraphPlugin* plugin)
+        : m_rootItem(aznew GraphCanvas::NodePaletteTreeItem("root", AnimGraphEditorId))
```
hultonha, line 18: question: If this is an AZ type then `aznew` is correct but if this does derive from a QWidget or QObject I believe we should just plain `new`

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/NodePaletteModelHelper.h
```diff
@@ -0,0 +1,52 @@
...
+     * node is focused.
+     *
+     * Nodes will be sorted into categories.
+     *
+     * Root node returned by GetRootItem() is typically passed in a config
+     * for GraphCanvas::NodePaletteWidget setup. Object of this class should live
+     * no longer than the widget because the widget takes the ownership of the
+     * root node and the nodes in this class will become dangling pointers if the
+     * wigdet is destroyed.
+     */
```
hultonha, line 35: nit: Please now prefer `//!` style comments instead of `/**/` block comments for all API comments. They're much easier to maintain and less fiddly to format (repeated below)

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/NodePaletteWidget.h
```diff
@@ -56,22 +60,18 @@ namespace EMStudio
...
         void OnFocusChanged(const QModelIndex& newFocusIndex, const QModelIndex& newFocusParent, const QModelIndex& oldFocusIndex, const QModelIndex& oldFocusParent);
 
     private:
-        void SaveExpandStates();
-        void RestoreExpandStates();
 
         AnimGraphPlugin*            m_plugin;
-        NodePaletteModel*           m_model;
-        QTreeView*                  m_treeView;
+        AZStd::unique_ptr <NodePaletteModelHelper> m_modelHelper;
```
hultonha, line 69: update: As the type this is being used in is a QWidget and we're not always sure about the lifetime, I'd have this use a raw pointer and `new` rather than a `AZStd::unique_ptr` to see if that makes any difference with the memory issues you were running into.

hultonha (commented):
Overall code looks good and I don't see anything immediately obvious that might be causing the memory failures in the tests 🤔 I think @burelc-amzn might be a good person to ask if my suggestions prove fruitless. The other suggestion is to do the classic bisect to earlier commits to try and determine the change that started causing the tests to fail to help isolate the change. If I think of anything else I'll let you know and feel free to poke me if you have more questions.

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/NodePaletteModelHelper.cpp
MiKom: Thanks for guidance.

`NodePaletteTreeItem` however doesn't inherit `QObject`. It derives from `GraphCanvas::GraphCanvasTreeItem` which is a plain class:
```cpp
class GraphCanvasTreeItem
{
```

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/ContextMenu.cpp
MiKom: I don't thing there's a lifetime issue since the palette widget takes ownership of the root pointer. It's passed in the `config` to the widget in `m_palette->SetupNodePalette(config);` couple of lines below.

Inside `SetupNodePalette` the root item is passed to the `GraphCanvasTreeModel` which in turns takes ownership of the item by storing it in a unique pointer:

```cpp
    class GraphCanvasTreeModel
        : public QAbstractItemModel
    {
/*snip*/
        AZStd::unique_ptr<GraphCanvasTreeItem> m_treeRoot;
    };
```

The only thing for the user of `NodePaletteModelHelper` is to not use it after the palette widget is destroyed because the item pointer will be invalid.

Since on one hand, the palette widget takes ownership of the root node and on the other, we still need a reference to it when updating the tree contents I don't see an easy way out. I could devise some class that would tie the palette widget and the helper class but that might be an overkill.

This `NodePaletteModelHelper` is more of a builder one might say. It can however rebuild the tree when it gets reinitialized for different focused node.  Maybe `NodePaletteModelUpdater` would make more sense?

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/NodePaletteWidget.h
MiKom: Now I realized that this doesn't have to be a pointer and can be a regular field. And this makes the question of memory management moot :)

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/BlendGraphWidget.cpp
MiKom: Done

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/BlendGraphWidget.cpp
MiKom: Done

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/BlendGraphWidget.h
MiKom: Done

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/BlendGraphWidget.h
MiKom: Done

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/ContextMenu.cpp
MiKom: I went ahead and renamed the class to `NodePaletteModelUpdater`. It's a bit more descriptive than `*Helper`

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/NodePaletteModelHelper.h
MiKom: Done

MiKom:
I managed to eliminate one error in tests by using a static c-style array instead of a static vector.

hultonha:
> I managed to eliminate one error in tests by using a static c-style array instead of a static vector.

Oh that'll do it! Static vectors aren't good as the allocators won't have been created at that point at static initialization time. Do you have any other AZ types that try and allocate and are being created statically like that?

pollend:
> > I managed to eliminate one error in tests by using a static c-style array instead of a static vector.
> 
> Oh that'll do it! Static vectors aren't good as the allocators won't have been created at that point at static initialization time. Do you have any other AZ types that try and allocate and are being created statically like that?

there is an C++ array type in AZStd

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/ContextMenu.cpp
hultonha: Okay I see, thanks for the explanation @MiKom. The updated name definitely helps 👍

hultonha:
@MiKom Are you still hitting the allocator test failures with this change? Are there any other types you might be creating statically that need allocation? I think something was improved in this area recently that @burelc-amzn worked on to make it possible to allocate things before `main` using the AZ allocators IIRC but I don't recall all the exact details.

MiKom:
Hey @hultonha 

warnings from static initialization are gone.

The only one left is from non-static allocations in this method:
```cpp
    BlendGraphMimeEvent* BlendGraphNodePaletteTreeItem::CreateMimeEvent() const
    {
        return IsEnabled() ? new BlendGraphMimeEvent(FromQtString(m_typeString), FromQtString(GetName())) : nullptr;
    }
```

I tried changing it to `aznew` but it didn't help.

hultonha:
> Hey @hultonha
> 
> warnings from static initialization are gone.
> 
> The only one left is from non-static allocations in this method:
> 
> ```c++
>     BlendGraphMimeEvent* BlendGraphNodePaletteTreeItem::CreateMimeEvent() const
>     {
>         return IsEnabled() ? new BlendGraphMimeEvent(FromQtString(m_typeString), FromQtString(GetName())) : nullptr;
>     }
> ```
> 
> I tried changing it to `aznew` but it didn't help.
> 
> I updated the test run log to show only this allocation failure.

Hmm interesting, so if those types are members of a fixture but aren't 'newed'/allocated in the fixture `SetUp`, then the `AZStd::strings` will potentially be created before the `AZ` allocators have been created. I think that might be the problem more than the `new` itself in that call maybe? 🤔

MiKom:
> I think that might be the problem more than the new itself in that call maybe? 🤔

I think so as well, the log complains not only about the allocation of `BlendGraphMimeEvent` but also about internal allocations of `AZstd::string` constructor in `FromQtString`.

Do you know if there are some ways to force the initialization of the allocators? I could put it in the text fixture setup function.

hultonha:
@MiKom  All fixtures usually derive from `AllocatorsTestFixture` and you have to remember to call the base `SetUp/TearDown` functions, so maybe one or other of those are missing in this particular test/fixture?

MiKom:
> @MiKom All fixtures usually derive from `AllocatorsTestFixture` and you have to remember to call the base `SetUp/TearDown` functions, so maybe one or other of those are missing in this particular test/fixture?

Hmm, then everything should work as expected. The problematic test is set up by `EMotionFX::UIFixture` which has the following inheritance hierarchy:
`EMotionFX::UIFixture <- EMotionFX::ComponentFixture (as UIFixtureBase) <- UnitTest::ScopedAllocatorSetupFixture <- UnitTest::ScopedAllocatorFixture <- UnitTest::AllocatorsBase`

hultonha:
> > @MiKom All fixtures usually derive from `AllocatorsTestFixture` and you have to remember to call the base `SetUp/TearDown` functions, so maybe one or other of those are missing in this particular test/fixture?
> 
> Hmm, then everything should work as expected. The problematic test is set up by `EMotionFX::UIFixture` which has the following inheritance hierarchy: `EMotionFX::UIFixture <- EMotionFX::ComponentFixture (as UIFixtureBase) <- UnitTest::ScopedAllocatorSetupFixture <- UnitTest::ScopedAllocatorFixture <- UnitTest::AllocatorsBase`

Hmm I see 🤔 We could maybe hop on a quick call this morning to go over this and have a look together?

MiKom:
For PR readers: the memory leaks were a result of the combination of a peculiar way that EMotionFX test suite simulated mouse clicks and slightly incorrect way of memory management in `GraphCanvas::NodePaletteWidget`. Look at individual commits for details.

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/BlendGraphWidget.cpp
```diff
@@ -155,6 +168,30 @@ namespace EMStudio
...
+
+            // When adding multiple nodes together, position them a bit differently
+            // so they don't end up on top of each other looking like only one was added
+            int offset = 0;
+            for (GraphCanvas::GraphCanvasMimeEvent* graphCanvasEvent : mimeContainer.m_mimeEvents)
+            {
+                if (auto* blendGraphEvent = azrtti_cast<BlendGraphMimeEvent*>(graphCanvasEvent))
+                {
+                    const QPoint nudgedPosition = { localPos.x(), localPos.y() + offset };
+                    offset += 10;
```
hultonha, line 189: suggestion: Is it worth moving the `10` here to a constant of named variable to make it more clearer what it is and why it's needed?

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/NodePaletteModelUpdater.h
```diff
@@ -0,0 +1,54 @@
...
+    //! root node and the nodes in this class will become dangling pointers if the
+    //! wigdet is destroyed.
+    class NodePaletteModelUpdater
+    {
+    public:
+        explicit NodePaletteModelUpdater(AnimGraphPlugin* plugin);
+        GraphCanvas::NodePaletteTreeItem* GetRootItem();
+        /**
+         * Rebuild the list of available/enabled nodes when a given node is focused
+         */
```
hultonha, line 45: nit: Prefer `//!` style (thanks for updating the above)

hultonha (approved):
LGTM! Nice work and very neat and clear commits! 😀 I only have super minor comments that aren't crazy important, more a note for future reference unless you want to make the changes now

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/NodePaletteModelUpdater.h
MiKom: Ah right, one leftover I missed when converting. Will change.

bhanuja-s:
> Tagging @bhanuja-s for visibility, please take a look at the video in the description when you have a moment and let us know if you have any feedback. Thanks!

This looks great, thanks for tagging me! I have a question, can a user create a new animgraph file when they drag out a node to the empty graph? This would follow Script Canvas' paradigm, and it would be great to be consistent throughout the engine on this behaviour.

MiKom:
@bhanuja-s I left the current behavior intact.

It would be possible to change this behavior to be in line with Script Canvas but unfortunately, it wouldn't be trivial. I suggest creating a new task for it.

MiKom:
@hultonha 

The CI failure we have (https://jenkins.build.o3de.org/job/O3DE/job/PR-12594/1/display/redirect) doesn't seem to be related to this change. It seems like some exception was thrown in Asset Processor tests.

To be honest, I can't even pinpoint the failure. I see some exceptions thrown in the log but the "Tests" tab shows that all tests are passing.

cgalvan (approved):

PR #12595, opened 2022-10-13
Remove double-click-to-exit for component mode

mbalfour-amzn:
## What does this PR do?

This removes the default behavior for component mode where double-clicking off of an entity would exit component mode. 

This was removed for the painting tool because double-clicking was too easy to do by mistake. After discussion on https://github.com/o3de/o3de/pull/12564 and with the UX team, it sounds like we should remove the default behavior as a whole, because other component modes (such as prefab editing) also don't want this behavior.

## How was this PR tested?

Went into component mode and verified that double-clicking didn't exit component mode.


hultonha (approved):
LGTM! Might be worth double checking if there's a test for this, not 100% there is but something might fail for a legit reason if there is (might also want to make a follow-up note for the docs team to update any info about this in the Component Mode docs). Tagging @FiniteStateGit for visibility.

greerdv (approved):
LGTM

AMZN-daimini (approved):

PR #12598, opened 2022-10-13
Added filtering to DPE-based ComponentEditor

alexmontAmazon:
## What does this PR do?
- Removes Qt dependency in FilterAdapter by switching from QString to AZStd::string
- Integrates filtering into DPE-based ComponentEditor

## How was this PR tested?

locally with Editor


Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/FilterAdapter.cpp
```diff
@@ -598,15 +598,15 @@ namespace AZ::DocumentPropertyEditor
     {
     }
 
-    void ValueStringFilter::SetFilterString(QString filterString)
+    void ValueStringFilter::SetFilterString(const AZStd::string& filterString)
```
AMZN-daimini, line 601: This string is being moved on line 605, is it intentional? Or should it be passed by value?

AMZN-daimini (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/FilterAdapter.cpp
alexmontAmazon: good catch! My brain just reverted to const& by default. Fixed now, thanks :)

amzn-tmryan (approved):

dtamkin1 (approved):

Issue #9666, opened 2022-05-19
Multiplayer: As a Dev I want a Warning When Loading Network Levels At Bad Times Depending on Multiplayer Status


AMZN-Gene:
**Is your feature request related to a problem? Please describe.**
The following workflow won't work, but is very easy to mistake as possible. As a dev, I want a warning whenever attempting to do the following...
_Server > loadlevel > host_

The reason this order of operation doesn't work is that when loading a networked level, the Multiplayer System still is uninitialized (ie the app doesn't know if it's a client, dedicatedserver, or clientserver).

**Describe the solution you'd like**
Check whenever a prefab (level) is loaded, and if that prefab is associated with a network.spawnable then check if the IMultiplayer system is initialized to be a client or server. If it's uninitialized, then display a prominent warning.

**Describe alternatives you've considered**
It might be good even to block loading the prefab entirely. Why would you load a network spawnable if there's no network?

AMZN-Gene:
This came up again in WF6. There's general confusion around when it's okay to loadlevel while hosting a multiplayer session

lmbr-pip:
Have log warnings now, need to add onscreen debug text to make this more visible

PR #12630, opened 2022-10-14
Removing Cry Dependency for Getting the Current Level Name

AMZN-Gene:
- Adding GetCurrentLevelName() method to AzFramework LevelSystem in order to remove Cry dependency. 
- Multiplayer will need to use this value in an upcoming PR
- Rename level blocking with its own EBus

Signed-off-by: Gene Walters <genewalt@amazon.com>

A piece of #9666 

## How was this PR tested?
Making sure the updated code that was using the old GetLevelName is still working.

### DIFF
--- a/Code/Framework/AzFramework/AzFramework/API/ApplicationAPI.h
+++ b/Code/Framework/AzFramework/AzFramework/API/ApplicationAPI.h
@@ -203,19 +203,22 @@ namespace AzFramework
     };
     using LevelLoadBlockerBus = AZ::EBus<LevelLoadBlockerRequests>;
 
-    class LevelSystemLifecycleRequests
-        : public AZ::EBusTraits
+    class ILevelSystemLifecycle
     {
     public:
-        static const AZ::EBusHandlerPolicy HandlerPolicy = AZ::EBusHandlerPolicy::Single;
-        static const AZ::EBusAddressPolicy AddressPolicy = AZ::EBusAddressPolicy::Single;
+        AZ_TYPE_INFO(ILevelSystemLifecycle, "{BDF3616A-4725-4B5D-AB71-34DFCDF9C5B0}");
+        virtual ~ILevelSystemLifecycle() = default;
 
         //! Returns the name of the currently loaded level.
         //! Note: for spawnable level system, this is the cache folder path to the level asset. Example: levels/mylevel/mylevel.spawnable
         //! @return Level name or empty string if no level loaded.
-        virtual AZStd::string GetCurrentLevelName() = 0;
+        virtual const char* GetCurrentLevelName() const = 0;
+
+        //! Checks if a level is loaded
+        //! @return true if a level is loaded; otherwise false.
+        virtual bool IsLevelLoaded() const = 0;
     };
-    using LevelSystemLifecycleRequestBus = AZ::EBus<LevelSystemLifecycleRequests>;
+    using LevelSystemLifecycleInterface = AZ::Interface<ILevelSystemLifecycle>;
 
     class LevelSystemLifecycleNotifications
         : public AZ::EBusTraits
--- a/Code/Framework/AzFramework/AzFramework/API/ApplicationAPI.h
+++ b/Code/Framework/AzFramework/AzFramework/API/ApplicationAPI.h
@@ -189,17 +189,31 @@ namespace AzFramework
         virtual void OnApplicationAboutToStop() {}
     };
 
+    class LevelLoadBlockerRequests
+        : public AZ::EBusTraits
+    {
+    public:
+        static const AZ::EBusHandlerPolicy HandlerPolicy = AZ::EBusHandlerPolicy::Multiple;
+        static const AZ::EBusAddressPolicy AddressPolicy = AZ::EBusAddressPolicy::Single;
+
+        //! Gives handlers the opportunity to block the loadlevel command.
+        //! There can be multiple handlers for this bus; if any one of them return true the loadlevel command will be stopped.
+        //! @return Return true to stop loading.
+        virtual bool ShouldBlockLevelLoading([[maybe_unused]]const char* levelName) { return false; }
+    };
+    using LevelLoadBlockerBus = AZ::EBus<LevelLoadBlockerRequests>;
+
     class LevelSystemLifecycleRequests
         : public AZ::EBusTraits
     {
     public:
-        // Bus Configuration
-        static const AZ::EBusHandlerPolicy HandlerPolicy = AZ::EBusHandlerPolicy::Multiple;
+        static const AZ::EBusHandlerPolicy HandlerPolicy = AZ::EBusHandlerPolicy::Single;
         static const AZ::EBusAddressPolicy AddressPolicy = AZ::EBusAddressPolicy::Single;
 
-        // Gives handlers the opportunity to block the loadlevel command.
-        // Return true to stop loading.
-        virtual bool ShouldBlockLevelLoading([[maybe_unused]]const char* levelName) { return false; }
+        //! Returns the name of the currently loaded level.
+        //! Note: for spawnable level system, this is the cache folder path to the level asset. Example: levels/mylevel/mylevel.spawnable
+        //! @return Level name or empty string if no level loaded.
+        virtual AZStd::string GetCurrentLevelName() = 0;
     };
     using LevelSystemLifecycleRequestBus = AZ::EBus<LevelSystemLifecycleRequests>;
 
--- a/Code/Legacy/CryCommon/ILevelSystem.h
+++ b/Code/Legacy/CryCommon/ILevelSystem.h
@@ -68,10 +68,12 @@ struct ILevelSystem
 
     //! Deprecated.
     //! @deprecated ILevelSystem is part of the legacy CryCommon module, please use AzFramework::LevelSystemLifecycleInterface::Get()->IsLevelLoaded instead.
+    //! O3DE_DEPRECATION_NOTICE(GHI-12715)
     virtual bool IsLevelLoaded() const = 0;
 
     //! Deprecated.
-    //! @deprecated ILevelSystem is part of the legacy CryCommon module, please use AzFramework::LevelSystemLifecycleInterface::Get()->GetCurrentLevelName instead.
+    //! @deprecated ILevelSystem is part of the legacy CryCommon module, please use AzFramework::LevelSystemLifecycleInterface::Get()->GetCurrentLevelName instead. O3DE_DEPRECATION_NOTICE(GHI-12715)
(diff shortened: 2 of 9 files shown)


Code/Legacy/CrySystem/LevelSystem/LevelSystem.h
```diff
@@ -92,6 +94,16 @@ class CLevelSystem
...
     void UnloadLevel() override;
     bool IsLevelLoaded() override { return m_bLevelLoaded; }
+
+    AZStd::string GetCurrentLevelName() override
+    {
+        if (m_pCurrentLevel && m_pCurrentLevel->GetLevelInfo())
+        {
+            return m_pCurrentLevel->GetLevelInfo()->GetName();
+        }
+        return "";
```
lumberyard-employee-dm, line 104: returning "" is a pessimization.
A default constructed AZStd::string should be returned instead
```suggestion
        return AZStd::string{};
```

Code/Framework/AzFramework/AzFramework/API/ApplicationAPI.h
```diff
@@ -189,18 +189,32 @@ namespace AzFramework
...
+        : public AZ::EBusTraits
+    {
+    public:
+        static const AZ::EBusHandlerPolicy HandlerPolicy = AZ::EBusHandlerPolicy::Single;
+        static const AZ::EBusAddressPolicy AddressPolicy = AZ::EBusAddressPolicy::Single;
+
+        //! Returns the name of the currently loaded level.
+        //! Note: for spawnable level system, this is the cache folder path to the level asset. Example: levels/mylevel/mylevel.spawnable
+        //! @return Level name or empty string if no level loaded.
+        virtual AZStd::string GetCurrentLevelName() = 0;
```
lumberyard-employee-dm, line 216: This function should be const.
But honestly, this shouldn't even be an EBus.

It should be an interface that gets used with AZ::Interface.

Also this function can return a `const char*`
No need for a `string` at this moment.

```
class ILevelSystemLifecycle
{
    virtual const char* GetCurrentLevelName() const = 0;
};

using LevelSystemLifecycleInterface = AZ::Interface<ILevelSystemLifecycle>;
```

Code/Legacy/CrySystem/LevelSystem/SpawnableLevelSystem.cpp
```diff
@@ -82,6 +82,7 @@ namespace LegacyLevelSystem
         }
 
         AzFramework::RootSpawnableNotificationBus::Handler::BusConnect();
+        AzFramework::LevelSystemLifecycleRequestBus::Handler::BusConnect();
```
lumberyard-employee-dm, line 85: This should instead register itself with an AZ::Interface.
```suggestion
        auto registerOutcome = AzFramework::LevelSystemLifecycleInterface::Register(this);
        AZ_Error("SpawnableLevelSystem", registerOutcome, "Failed to register the SpawnableLevelSystem with the LevelSystemLifecycleInterface: %s", registerOutcome.GetError().c_str());
```

Code/Framework/AzFramework/AzFramework/API/ApplicationAPI.h
AMZN-Gene: Much better :D
Side question: is there still rule against using const EBus functions? I recall that being a thing even for simple stuff like Getters().

Code/Legacy/CrySystem/LevelSystem/LevelSystem.h
AMZN-Gene: Using const char* thanks :)

Code/Legacy/CrySystem/LevelSystem/SpawnableLevelSystem.cpp
AMZN-Gene: Fixed using Interface now! Deriving from Interface<>::Registrar seems fine too, or should I explicitly keep the error checking?

lumberyard-employee-dm: I think either work fine, though I would prefer the error message path.

lumberyard-employee-dm (approved):

Code/Framework/AzFramework/AzFramework/API/ApplicationAPI.h
lumberyard-employee-dm: No, that paradigm of avoiding using const EBus functions was never built on a solid reason.

The thought process by those recommending it was what "if the handler function need to modify the class, but the interface imposed a const function it".

It is not a strong reason why functions shouldn't be able to marked const

AMZN-Gene: Thank you. I'm a const fan.

Code/Legacy/CrySystem/LevelSystem/SpawnableLevelSystem.cpp
AMZN-Gene: Good idea. Keep the RAII, and the error.

Code/Legacy/CryCommon/ILevelSystem.h
```diff
@@ -65,7 +65,11 @@ struct ILevelSystem
 
     virtual bool LoadLevel(const char* levelName) = 0;
     virtual void UnloadLevel() = 0;
-    virtual bool IsLevelLoaded() = 0;
+
+    // Deprecated. Use AzFramework::LevelSystemLifecycleInterface::Get()->IsLevelLoaded instead.
+    virtual bool IsLevelLoaded() const = 0;
+
+    // Deprecated. Use AzFramework::LevelSystemLifecycleInterface::Get()->GetCurrentLevelName instead.
```
pollend, line 72: can you follow this: https://github.com/o3de/sig-core/blob/main/governance/API-Ref-Guidelines-Update.md#deprecations

lumberyard-employee-dm (approved):

pollend:
@hultonha I know you worked on the deprecation guide is this the latest up to date one I remember there had to be an issue or something to track what was deprecated?

hultonha:
> @hultonha I know you worked on the deprecation guide is this the latest up to date one I remember there had to be an issue or something to track what was deprecated?

Well spotted @pollend. @AMZN-Gene could you please refer to this document https://github.com/o3de/community/blob/main/guides/o3de-deprecation-guidelines.md and create a follow-up GHI (that is referenced at the callsite of the deprecation) to track its removal. This will help us actually track the deprecation and know when we can remove it. Thanks!

Code/Legacy/CryCommon/ILevelSystem.h
AMZN-Gene: fixed! thank you

AMZN-Gene:
> > @hultonha I know you worked on the deprecation guide is this the latest up to date one I remember there had to be an issue or something to track what was deprecated?
> 
> Well spotted @pollend. @AMZN-Gene could you please refer to this document https://github.com/o3de/community/blob/main/guides/o3de-deprecation-guidelines.md and create a follow-up GHI (that is referenced at the callsite of the deprecation) to track its removal. This will help us actually track the deprecation and know when we can remove it. Thanks!

Nice! I added the GHI and updated this PR to include the GHI #12715

Gems/ImGui/Code/Source/LYCommonMenu/ImGuiLYCommonMenu.cpp
```diff
@@ -305,11 +305,14 @@ namespace ImGui
                 auto lvlSystem = (gEnv && gEnv->pSystem) ? gEnv->pSystem->GetILevelSystem() : nullptr;
                 if (lvlSystem && ImGui::BeginMenu("Levels"))
                 {
-                    if (lvlSystem->IsLevelLoaded())
+                    if (AzFramework::LevelSystemLifecycleInterface::Get()->IsLevelLoaded())
```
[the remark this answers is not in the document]
AMZN-Gene: Fixed!

AMZN-puvvadar (approved):

Gems/ImGui/Code/Source/LYCommonMenu/ImGuiLYCommonMenu.cpp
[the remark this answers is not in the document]
lumberyard-employee-dm: You could store this in a pointer for reuse later instead of making two calls to LevelSystemLifecycleInterface::`Get()`

lumberyard-employee-dm (approved):

pollend (approved):

hultonha:
> > > @hultonha I know you worked on the deprecation guide is this the latest up to date one I remember there had to be an issue or something to track what was deprecated?
> > 
> > 
> > Well spotted @pollend. @AMZN-Gene could you please refer to this document https://github.com/o3de/community/blob/main/guides/o3de-deprecation-guidelines.md and create a follow-up GHI (that is referenced at the callsite of the deprecation) to track its removal. This will help us actually track the deprecation and know when we can remove it. Thanks!
> 
> Nice! I added the GHI and updated this PR to include the GHI #12715

Awesome thanks @gene-walt! 🥳

Code/Legacy/CrySystem/LevelSystem/SpawnableLevelSystem.h
```diff
@@ -20,34 +19,40 @@ namespace LegacyLevelSystem
...
 
-        // ILevelSystem
+        //! ILevelSystem overrides.
+        //! @{
         void Release() override;
 
         void AddListener(ILevelSystemListener* pListener) override;
         void RemoveListener(ILevelSystemListener* pListener) override;
 
         bool LoadLevel(const char* levelName) override;
```
kberg-amzn, line 35: We really just need this level system itself pulled out of legacy, or at least the interface, ILevelSystem, and have it registered as an AZ::Interface. This would be better than having a singleton level interface in the legacy codebase, and another ebus interface in azframework that we combine together in this manner. What we have here, we're going to increasingly need to move methods over to the LevelSystemLifecycle interface to expose it to the rest of the engine, and we'll likely end up needing to rename/refactor the LevelSystemLifecycle interface itself because the added functionality won't be lifecycle specific necessarily, resulting in more interface changes than strictly required.

AMZN-Gene: Agreed! My next step was to pull this out entirely and potentially even pull CryCommon out of the other gems that need to use the level system. I’ll create a GHI for that task.

AMZN-Gene: https://github.com/o3de/o3de/issues/12772

PR #12636, opened 2022-10-14
Adding implementation of C++23 std::expected

lumberyard-employee-dm:
Updated the `AZ::Outcome` class to use std::expected.

Fixed the SkeletonOutlinerPlugin::GetSelectedRowIndices function to properly returns an `Outcome<QModelIndexList>`.
It was incorrectly returning a `Outcome<const QModelIndexList&>`, but the [QSelectionModel::selectedRows](https://doc.qt.io/qt-5/qitemselectionmodel.html#selectedRows) function that populates that Outcome returns a value not a reference

## How was this PR tested?

Validated that all the Outcome test passes.

Added unit-test to validate `AZStd::expected` functionality


amzn-rhhong:
Thanks for fixing the emfx code!

Code/Editor/Lib/Tests/test_AzAssetBrowserRequestHandler.cpp
```diff
@@ -226,17 +226,17 @@ namespace UnitTest
                         {
                             if (comptype != AZ::Uuid(MockEditorComponent::s_MockEditorComponentTypeId))
                             {
-                                return ThisOutcome(AZStd::string("Failed - wrong component"));
```
hultonha, line 229: question: Would it be possible to move the `AZStd::unexpect` inside the `ThisOutcome` constructor so it's better hidden and fewer callsites need to be updated?

Code/Editor/Settings.cpp
```diff
@@ -1027,7 +1029,8 @@ AzToolsFramework::EditorSettingsAPIRequests::SettingOutcome SEditorSettings::Set
     }
     else
     {
-        return { AZStd::string("Invalid Value Type - supported types: string, bool, int, float") };
+        return AzToolsFramework::EditorSettingsAPIRequests::SettingOutcome{ AZStd::unexpect,
+            AZStd::string("Invalid Value Type - supported types: string, bool, int, float") };
```
hultonha, line 1033: question: Same here, is it possible to keep `AZStd::unexpect` internal (for now) until all usages move over to `AZStd::expected` without needing to change all callsites?

Code/Framework/AzCore/Platform/Common/VisualStudio/AzCore/Natvis/azcore.natvis
```diff
@@ -643,25 +643,29 @@
     </Expand>
   </Type>
 
-  <Type Name="AZ::Outcome&lt;*&gt;">
-    <DisplayString Condition="m_isSuccess" >{{Value = {*(SuccessType*)&amp;m_success}}}</DisplayString>
-    <DisplayString Condition="!m_isSuccess">{{Error = {*(FailureType*)&amp;m_failure}}}</DisplayString>
+  <Type Name="AZStd::expected&lt;*&gt;">
```
hultonha, line 646: Great work updating the natvis file too! People will thank you for that! 😅

Code/Framework/AzCore/Tests/AZStd/ExpectedTests.cpp
```diff
@@ -0,0 +1,317 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#include <AzCore/std/string/fixed_string.h>
+#include <AzCore/std/utility/expected.h>
+#include <AzCore/UnitTest/TestTypes.h>
+
+namespace UnitTest
```
hultonha, line 13: Great to see the thorough test coverage, awesome work!

hultonha (commented):
This does look like a great update, excellent we're standardizing things, but this does also look like an immediate breaking API change for `AZ::Outcome` that could impact downstream consumers who are using it. Is it possible to insulate `AZ::Outcome` from API changes for this addition and follow the deprecation process to get people to update to `AZStd::expected` for the next or following release (https://github.com/o3de/community/blob/main/guides/o3de-deprecation-guidelines.md).

Code/Framework/AzCore/AzCore/std/utility/expected.h
```diff
@@ -0,0 +1,326 @@
...
+
+    private:
+        E m_unexpected;
+    };
+
+    template<class E>
+    unexpected(E) -> unexpected<E>;
+
+    //! Implementation of C++23 [std::expected](https://en.cppreference.com/w/cpp/utility/expected)
+    //! It inherits from AZ::Outcome and is pretty much similar class with the following differences
```
AMZN-puvvadar, line 53:
```suggestion
    //! It inherits from AZ::Outcome and is a pretty much similar class with the following differences
```
NP: Typo

Code/Framework/AzCore/AzCore/std/utility/expected_internal.h
```diff
@@ -0,0 +1,524 @@
...
+    constexpr bool is_std_unexpected_specialization_v = false;
+
+    template<class E>
+    constexpr bool is_std_unexpected_specialization_v<unexpected<E>> = true;
+
+    // Due to using not having C++20 support, the requires block cannot be used
+    // to specify constraints on a class special member functions(default constructor, copy/move constructor, destructor)
+    // without using function templates on the AZStd::expected class template.
+    // The problem with using a function template though, is that it isn't default'able.
+    // Therefore it means that the class cannot have trivial version of these special member functions
```
AMZN-puvvadar, line 56:
```suggestion
    // Therefore it means that the class cannot have trivial versions of these special member functions
```

NP: Typo

AMZN-puvvadar (approved):
AzNetworking changes specifically seem reasonable. AzNetworking units exercise Serializers so if those are passing then I have no concerns specifically from the sig-network perspective.

I reviewed the change in general as well and it seems reasonable to me though I don't have strog opinions on std::expected.

Code/Editor/Lib/Tests/test_AzAssetBrowserRequestHandler.cpp
hultonha: I see, we still likely need to issue a deprecation notice around this though as this could break people. If the current code could be at risk of already being broken we might be able to argue making this change but otherwise for a fairly widely used type we're breaking the API without any notice period for people to upgrade without being broken

lumberyard-employee-dm: I will add a deprecation notice for the API.
But based on a search of where AZ::Outcome is used across known git repositories using sourcegraph, it is only referenced in @AMZN-Olex book(https://sourcegraph.com/search?q=context:global+AZ::Outcome+NOT+repo:o3de&patternType=standard)

hultonha: Understood, thanks for adding issue #12696 @lumberyard-employee-dm 👍 Once this is merged is likely makes sense to send an impactful change notification to Discord too

Code/Framework/AzCore/AzCore/std/utility/expected.h
```diff
@@ -0,0 +1,326 @@
...
+ *
+ */
+#pragma once
+
+#include <AzCore/std/typetraits/is_swappable.h>
+#include <AzCore/std/utility/expected_internal.h>
+
+namespace AZStd
+{
+    //! Used to pass wrap the error type supplied to the construct an expected object
```
hultonha, line 15: nit/update: `//! Used to pass wrap the error type supplied to the construct an expected object` -> `//! Wraps the error type supplied to construct an expected object`

hultonha: Also should this say 'unexpected' object?

Code/Framework/AzCore/AzCore/std/utility/expected_internal.h
```diff
@@ -0,0 +1,524 @@
...
+        || !is_constructible_v<unexpected<E>, const expected<U, G>&>
+        || !is_constructible_v<unexpected<E>, const expected<U, G>>;
+
+    template<class T>
+    constexpr bool is_std_unexpected_specialization_v = false;
+
+    template<class E>
+    constexpr bool is_std_unexpected_specialization_v<unexpected<E>> = true;
+
+    // Due to using not having C++20 support, the requires block cannot be used
```
hultonha, line 52: nit: remove 'using' here

Code/Framework/AzCore/AzCore/std/utility/expected_internal.h
```diff
@@ -0,0 +1,524 @@
...
+    // to specify constraints on a class special member functions(default constructor, copy/move constructor, destructor)
+    // without using function templates on the AZStd::expected class template.
+    // The problem with using a function template though, is that it isn't default'able.
```
hultonha, line 55: nit: default'able -> defaultable?

hultonha: Or maybe this is the accepted way of writing it as I see it appearing elsewhere

hultonha (approved):
Overall looks good to me but would be good to get some more sig-core eyes on it.

Also I wanted to check is it now preferred/recommended to use AZStd::expected directly in place of AZ::Outcome? Or should AZ::Outcome still be used? Is this something we want to mention in the coding standards in future? Thanks!

Code/Editor/Lib/Tests/test_AzAssetBrowserRequestHandler.cpp
lumberyard-employee-dm: Will do.
I'll post it on Discord

Code/Framework/AzCore/AzCore/std/utility/expected_internal.h
lumberyard-employee-dm: Well I don't think `defaultable` is a word, so I put the apostrophe to separate it

Code/Framework/AzCore/AzCore/std/utility/expected.h
lumberyard-employee-dm: What makes the terminology confusing between "unexpected" and "error" is that the type alias is [error_type](https://en.cppreference.com/w/cpp/utility/expected) which has the description of "the type of the unexpected value"

PR #12637, opened 2022-10-14
Asset Processor - Use absolute paths internally

amzn-mike:
## What does this PR do?

Adds a SourceAssetReference class which refers to a single asset on disk and provides absolute path, relative path, scanfolder path, and scanfolder id information for said file.

Updated a large portion of AssetProcessor code to use SourceAssetReference instead of random strings which may contain relative or absolute paths.

The net result of this change should be that AP behaves exactly as before with no difference (although a few existing bugs were fixed along the way).  This is groundwork for enabling future support for AP to work with multiple files having the same relative path.

## How was this PR tested?

Added several new unit tests
Updated and ran all unit tests
Manually tested running AP and reprocessing a bunch of assets


Code/Tools/AssetProcessor/native/utilities/assetUtils.cpp
```diff
@@ -1426,22 +1427,37 @@ namespace AssetUtilities
...
 
         if (!topLevelSource)
         {
-            AzToolsFramework::AssetDatabase::SourceDatabaseEntry source;
-            db->GetSourceBySourceName(relativeSourcePath.FixedMaxPathStringAsPosix().c_str(), source);
+            AzToolsFramework::AssetDatabase::SourceDatabaseEntryContainer source;
+            db->GetSourcesBySourceNameScanFolderId(sourceAsset.RelativePath().c_str(), sourceAsset.ScanfolderId(), source);
 
-            topLevelSource = source;
+            AZ_Assert(
```
jackalbe, line 1442: This is data coming in from the database and it can AZ_Assert
If this happens, will the user just need to nuke the database because the code will just assert each time now.

Code/Tools/AssetProcessor/native/utilities/IPathConversion.h
```diff
@@ -0,0 +1,33 @@
...
+        AZ_RTTI(IPathConversion, "{8838D113-8BC0-4270-BF4D-4DFEAB628102}");
+
+        virtual ~IPathConversion() = default;
+
+        virtual bool ConvertToRelativePath(QString fullFileName, QString& databaseSourceName, QString& scanFolderName) const = 0;
+
+        //! given a full file name (assumed already fed through the normalization funciton), return the first matching scan folder
+        virtual const AssetProcessor::ScanFolderInfo* GetScanFolderForFile(const QString& fullFileName) const = 0;
+
+        virtual const AssetProcessor::ScanFolderInfo* GetScanFolderById(AZ::s64 id) const = 0;
```
jackalbe, line 31: Instead of a naked pointer, why not use an AZStd::optional wrapper?

Code/Tools/AssetProcessor/native/ui/MainWindow.cpp
```diff
@@ -1841,21 +1841,7 @@ void MainWindow::ShowJobLogContextMenu(const QPoint& pos)
...
-    AssetSystemRequestBus::BroadcastResult(result, &AssetSystemRequestBus::Events::GetSourceInfoBySourcePath, assetNameUtf8.constData(), info, watchFolder);
-    if (!result)
-    {
-        AZ_Error("AssetProvider", false, "Failed to locate asset info for '%s'.", assetNameUtf8.constData());
-    }
-
-    return result
-        ? QDir(watchFolder.c_str()).absoluteFilePath(info.m_relativePath.c_str())
-        : QString();
+    return cachedJobInfo->m_elementId.GetSourceAssetReference().AbsolutePath().c_str();
```
jackalbe, line 1844: Should the "cachedJobInfo" naked pointer be checked or null?

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.h
```diff
@@ -499,8 +489,8 @@ namespace AssetProcessor
         JobRunKeyToJobInfoMap m_jobRunKeyToJobInfoMap;
         AZStd::multimap<AZStd::string, AZ::u64> m_jobKeyToJobRunKeyMap;
 
-        using SourceUUIDToSourceInfoMap = AZStd::unordered_map<AZ::Uuid, SourceInfo>;
-        SourceUUIDToSourceInfoMap m_sourceUUIDToSourceInfoMap; // contains UUID -> SourceInfo, which includes database name and relative to watch folder:
+        using SourceUUIDToSourceInfoMap = AZStd::unordered_map<AZ::Uuid, SourceAssetReference>;
+        SourceUUIDToSourceInfoMap m_sourceUUIDToSourceInfoMap; // contains UUID -> SourceAssetReference
```
jackalbe, line 493: Is this UUID the Source Asset UUID?

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
```diff
@@ -208,29 +207,29 @@ namespace AssetProcessor
     {
         //this function just adds an removes to a maps to speed up job status, we don't actually write
         //to the database until it either succeeds or fails
-        AZ::Uuid sourceUUID = AssetUtilities::CreateSafeSourceUUIDFromName(jobEntry.m_databaseSourceName.toUtf8().data());
-        AZ::Uuid legacySourceUUID = AssetUtilities::CreateSafeSourceUUIDFromName(jobEntry.m_databaseSourceName.toUtf8().data(), false); // legacy source uuid
+        AZ::Uuid sourceUUID = AssetUtilities::CreateSafeSourceUUIDFromName(jobEntry.m_sourceAssetReference.RelativePath().c_str());
+        AZ::Uuid legacySourceUUID = AssetUtilities::CreateSafeSourceUUIDFromName(jobEntry.m_sourceAssetReference.RelativePath().c_str(), false); // legacy source uuid
```
jackalbe, line 211: is legacySourceUUID the "former source UUID"?

AMZN-stankowi (commented):
I've made it halfway through today, I'll go through the second half tomorrow.

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
```diff
@@ -3655,6 +3660,21 @@ namespace AssetProcessor
...
+                QString absolutePath = m_platformConfig->FindFirstMatchingFile(databaseSourceName);
+
+                if (!absolutePath.isEmpty())
+                {
+                    databaseSourceName = absolutePath;
+                }
+                else
+                {
+                    // If we can't resolve the dependency, it usually means it doesn't exist
+                    job.m_jobDependencyList.erase(jobDependencyInternal);
```
jackalbe, line 3674: so the tech does not support "future dependent" files that might be created from another job, correct?

Code/Tools/AssetProcessor/native/AssetManager/SourceAssetReference.cpp
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    bool SourceAssetReference::operator<(const SourceAssetReference& other) const
+    {
+        return m_absolutePath < other.m_absolutePath;
+    }
+
+    SourceAssetReference::operator bool() const
+    {
+        return !m_absolutePath.empty();
```
junbo75, line 72: Can the source asset can be uniquely identified via its scan folder and relative path as well? If that's the case, maybe this function should return true if !m_relativePath.empty() && !m_scanfolderPath.empty()

Code/Tools/AssetProcessor/native/AssetManager/SourceAssetReference.h
```diff
@@ -0,0 +1,86 @@
...
+
+#include <QString>
+#include <AzCore/Interface/Interface.h>
+#include <AzCore/IO/Path/Path.h>
+#include <utilities/IPathConversion.h>
+#include <native/AssetManager/assetScanFolderInfo.h>
+
+namespace AssetProcessor
+{
+    class SourceAssetReference
```
junbo75, line 19: Nit: It may worth having some comments for explaining the responsibility of this class. I'm also wondering whether we want to move this to AzFramework or AzToolFramework since other content tools may also reference source assets via this structure going forward.

Code/Tools/AssetProcessor/native/tests/SourceAssetReferenceTests.cpp
```diff
@@ -0,0 +1,182 @@
...
+#include <AzTest/AzTest.h>
+#include <AzCore/UnitTest/TestTypes.h>
+#include <AssetManager/SourceAssetReference.h>
+#include "UnitTestUtilities.h"
+
+namespace UnitTests
+{
+    using SourceAssetReferenceTests = ::UnitTest::ScopedAllocatorSetupFixture;
+
+    TEST_F(SourceAssetReferenceTests, ConstructBasic)
```
junbo75, line 18: Nit: It might be better to name the tests following the Osherove naming method: https://wiki.agscollab.com/display/lmbr/Test+Writing+Standards#TestWritingStandards-TheOsheroveNamingMethod

Code/Tools/AssetProcessor/native/tests/UnitTestUtilities.h
```diff
@@ -12,6 +12,7 @@
 #include <utilities/assetUtils.h>
 #include <AzCore/Component/ComponentApplicationBus.h>
 #include <AzCore/Interface/Interface.h>
+#include <AzCore/Math/Uuid.h>
```
junbo75, line 15: Nit: Any specific reason to include this header? Seems like the new mock structure doesn't need UUID.

Code/Tools/AssetProcessor/native/unittests/AssetProcessorManagerUnitTests.cpp
```diff
@@ -463,9 +467,8 @@ namespace AssetProcessor
         {
             UNIT_TEST_EXPECT_TRUE(processResults[checkIdx].m_jobEntry.m_computedFingerprint != 0);
             UNIT_TEST_EXPECT_TRUE(processResults[checkIdx].m_jobEntry.m_jobRunKey != 0);
-            UNIT_TEST_EXPECT_TRUE(processResults[checkIdx].m_jobEntry.m_watchFolderPath == AssetUtilities::NormalizeFilePath(watchFolderPath));
-            UNIT_TEST_EXPECT_TRUE(processResults[checkIdx].m_jobEntry.m_pathRelativeToWatchFolder == "uniquefile.txt");
-            UNIT_TEST_EXPECT_TRUE(processResults[checkIdx].m_jobEntry.m_databaseSourceName == "uniquefile.txt");
+            UNIT_TEST_EXPECT_TRUE(QString(processResults[checkIdx].m_jobEntry.m_sourceAssetReference.ScanfolderPath().c_str()) == AssetUtilities::NormalizeFilePath(watchFolderPath));
```
junbo75, line 470: Just a heads up that I have another ongoing PR (https://github.com/o3de/o3de/pull/12607) for migrating this test to the modern test framework as well. Not sure which PR will go in first and it may require either of us to resolve the conflict.

Code/Tools/AssetProcessor/native/utilities/assetUtils.cpp
amzn-mike: This API is honestly poorly designed.  It makes no sense for it to return a container of sources when it should be impossible to have more than 1 source with the same name and scanfolder ID.  That would mean 2 files exist at the same absolute path, which could only happen if the database is corrupted or some code screwed up.

I could remove the assert I suppose, since I doubt it will ever fire, I think I just felt it explained the assumption made below where I'm just blindly grabbing index 0.

Code/Tools/AssetProcessor/native/utilities/IPathConversion.h
amzn-mike: GetScanFolderForFile is an existing API that's already used everywhere so I can't really change that.  I could for the other API, but I feel like a pointer is a perfectly valid use-case here: there may not be a scanfolder, so a reference isn't a good choice, and a pointer already provides the "this might not exist" functionality and avoids unnecessary copying.

Code/Tools/AssetProcessor/native/AssetManager/SourceAssetReference.cpp
```diff
@@ -0,0 +1,108 @@
...
+        m_relativePath = relativePath.toUtf8().constData();
+        m_scanfolderPath = scanfolderPath.toUtf8().constData();
+
+        Normalize();
+    }
+
+    SourceAssetReference::SourceAssetReference(AZ::IO::PathView scanFolderPath, AZ::IO::PathView pathRelativeToScanFolder)
+    {
+        AZ_Assert(!scanFolderPath.Native().empty(), "scanfolderPath is empty");
+        AZ_Assert(!pathRelativeToScanFolder.Native().empty(), "pathRelativeToScanFolder is empty");
```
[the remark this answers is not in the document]
amzn-mike: I'm on the fence about this.  If it were an error or anything else instead, it would likely happen silently and we'd never know about it or fix it, because unfortunately that's what happens in a lot of cases in AP since most things get routed to logs that no one looks at.  I like it because it helps catch bugs quickly (it already has pointed out a few) but if someone is strongly against it I can take them out.

Code/Tools/AssetProcessor/native/ui/MainWindow.cpp
amzn-mike: It wasn't being checked before but I can add that

Code/Tools/AssetProcessor/native/resourcecompiler/RCQueueSortModel.cpp
```diff
@@ -61,7 +61,14 @@ namespace AssetProcessor
                     if (jobDepedencyInternal.m_jobDependency.m_type == AssetBuilderSDK::JobDependencyType::Order || jobDepedencyInternal.m_jobDependency.m_type == AssetBuilderSDK::JobDependencyType::OrderOnce)
                     {
                         const AssetBuilderSDK::JobDependency& jobDependency = jobDepedencyInternal.m_jobDependency;
-                        QueueElementID elementId(jobDependency.m_sourceFile.m_sourceFileDependencyPath.c_str(), jobDependency.m_platformIdentifier.c_str(), jobDependency.m_jobKey.c_str());
+                        AZ_Assert(
```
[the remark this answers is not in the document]
amzn-mike: It comes from the builders and is run through some processing code which is responsible for converting whatever type of input is provided into the expected format here.

Code/Tools/AssetProcessor/native/resourcecompiler/JobsModel.cpp
```diff
@@ -416,11 +423,13 @@ namespace AssetProcessor
     QModelIndex JobsModel::GetJobFromProduct(const AzToolsFramework::AssetDatabase::ProductDatabaseEntry& productEntry, AzToolsFramework::AssetDatabase::AssetDatabaseConnection& assetDatabaseConnection)
     {
         AZStd::string sourceForProduct;
+        AZ::s64 scanFolderId;
         assetDatabaseConnection.QuerySourceByProductID(
             productEntry.m_productID,
             [&](AzToolsFramework::AssetDatabase::SourceDatabaseEntry& sourceEntry)
         {
             sourceForProduct = sourceEntry.m_sourceName;
+                scanFolderId = sourceEntry.m_scanFolderPK;
```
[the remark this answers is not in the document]
amzn-mike: Sure is, fixed

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.h
amzn-mike: Appears to be, I can update the comment to be more explicit about that.

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
amzn-mike: It's the legacy UUID format, which did not lowercase the path before hashing.  Updated the comment a bit

Code/Tools/AssetProcessor/native/AssetManager/SourceAssetReference.cpp
amzn-mike: It can, but SourceAssetReference will always resolve the absolute path, since that's really what its meant to use for referencing files

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
amzn-mike: This is for job dependencies specifically, so that is correct.  However this is also just part of the in-memory dependency resolution, the dependency should still be stored in the db beforehand.

Code/Tools/AssetProcessor/native/AssetManager/SourceAssetReference.cpp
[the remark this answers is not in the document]
amzn-mike: I've removed these since Joe also brought it up

Code/Tools/AssetProcessor/native/tests/SourceAssetReferenceTests.cpp
amzn-mike: Updated

Code/Tools/AssetProcessor/native/AssetManager/AssetCatalog.cpp
```diff
@@ -637,20 +637,16 @@ namespace AssetProcessor
...
     {
         AZStd::lock_guard<AZStd::mutex> lock(m_sourceUUIDToSourceNameMapMutex);
 
-        SourceInfo sourceInfo = { rootPath, relativeFilePath };
-        m_sourceUUIDToSourceNameMap.insert({ sourceUuid, sourceInfo });
+        m_sourceUUIDToSourceNameMap.insert({ sourceUuid, sourceAsset });
 
         //adding legacy source uuid as well
-        m_sourceUUIDToSourceNameMap.insert({ legacyUuid, sourceInfo });
+        m_sourceUUIDToSourceNameMap.insert({ legacyUuid, sourceAsset });
```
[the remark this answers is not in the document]
amzn-mike: Updated

Code/Tools/AssetProcessor/native/AssetManager/SourceAssetReference.cpp
```diff
@@ -0,0 +1,108 @@
...
+        m_absolutePath = absolutePath;
+        m_relativePath = relativePath.toUtf8().constData();
+        m_scanfolderPath = scanfolderPath.toUtf8().constData();
+
+        Normalize();
+    }
+
+    SourceAssetReference::SourceAssetReference(AZ::IO::PathView scanFolderPath, AZ::IO::PathView pathRelativeToScanFolder)
+    {
+        AZ_Assert(!scanFolderPath.Native().empty(), "scanfolderPath is empty");
```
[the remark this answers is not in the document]
amzn-mike: I've removed the asserts and added an IsValid (there is an operator bool already but I figured it doesn't hurt to have both)

AMZN-stankowi (approved):

junbo75 (approved):

Code/Tools/AssetProcessor/native/resourcecompiler/RCQueueSortModel.cpp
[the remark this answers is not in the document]
jackalbe: So my question still stands. This seems like some input that will happen again and again for the customer. The AZ_Assert() will crash this application due to bad input; and it seems that this asset builder will continue getting the same input. 

How can this detection inform the customer about how to avoid/fix this?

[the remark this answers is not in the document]
amzn-mike: Sorry I wasn't clear.  The customer can't fix this because it's a coding error.  This code assumes the previous processing code did it's job correctly (converting the various inputs into an absolute path), so the assert is there to catch that in case something changes (or a unit test hasn't set things up correctly...).  So this wouldn't happen because of a data error, only because someone changes the code.

siliconvoodoo:
good change, feels cleaner.

Code/Tools/AssetProcessor/native/utilities/assetUtils.cpp
jackalbe: So here. GetSourcesBySourceNameScanFolderId() comes back with a SourceDatabaseEntryContainer named source. This is incoming data from the data base. If the "source.size() > 1" then the assert will fire and kill the AP due to this assert. 

This seems to be triggering an assert based on input data which means what for the user? How will coding fix this?

amzn-mike: I've removed the assert since it is an impossible situation and having more than 1 result won't affect anything

PR #12641, opened 2022-10-15
Improved culling performance by 2.5ms (roughly 50%)

antonmic:
Changes:

**Culling.cpp**: changed WorkListCapacity from 5 to 9. This increases performance quite a bit as it increases the size of the array of visibility nodes we fill before we spawn a job to process that list (so less job spawning, longer jobs, better performance)
The rest of Culling.cpp is just cosmetic changes

**RenderPipeline.cpp**: Removed unnecessary call to reset views. Resetting the view resets the DrawListContext, which kills all the array and so the array capacity during the next draw list context gather will be 0, resulting in array growing and memory allocations. This changes lets the arrays keep their capacity for better draw item insertion.

**View.cpp**: Don't reset the DrawListContext if the mask is the same (for the same reason as just stated)

Update:
Added a commit with the option to use entry lists instead of visibility node lists for job work. This creates more balanced jobs and results in better performance when tested. There is a CVAR to switch between this and the old node list method as well as a CVARs to control the amount of entities accumulated into the node list / entity list respectively before the lists are handed off to jobs.

Signed-off-by: antonmic <56370189+antonmic@users.noreply.github.com>

### DIFF
--- a/Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
+++ b/Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
@@ -166,7 +166,7 @@ namespace AZ
             {
                 if (m_debugCtx->m_debugDraw && (m_view->GetName() == m_debugCtx->m_currentViewSelectionName))
                 {
-                    AuxGeomFeatureProcessorInterface::GetDrawQueueForScene(worklistData->m_scene);
+                    AuxGeomFeatureProcessorInterface::GetDrawQueueForScene(m_scene);
                 }
                 return nullptr;
             }
@@ -290,7 +290,7 @@ namespace AZ
                             if (worklistData->m_debugCtx->m_drawBoundingBoxes)
                             {
                                 auxGeomPtr->DrawObb(c->m_cullData.m_boundingObb, Matrix3x4::Identity(),
-                                    nodeIsContainedInFrustum ? Colors::Lime : Colors::Yellow, AuxGeomDraw::DrawStyle::Line);
+                                    parentNodeContainedInFrustum ? Colors::Lime : Colors::Yellow, AuxGeomDraw::DrawStyle::Line);
                             }
 
                             if (worklistData->m_debugCtx->m_drawBoundingSpheres)
--- a/Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
+++ b/Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
@@ -168,7 +168,10 @@ namespace AZ
             return worklistData;
         }
 
-        constexpr size_t WorkListCapacity = 5;
+        // Any value larger than 9 triggers a crash the job creation following the message:
+        // "Allocation size (560) is too big (max: 512) for pools!"
+        // (each additional count here increases the allocation size by 48, value of 9 makes the size exactly 512)
+        constexpr size_t WorkListCapacity = 9;
         using WorkListType = AZStd::fixed_vector<AzFramework::IVisibilityScene::NodeData, WorkListCapacity>;
 
 #if AZ_TRAIT_MASKED_OCCLUSION_CULLING_SUPPORTED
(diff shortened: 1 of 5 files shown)


Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
```diff
@@ -271,13 +273,14 @@ namespace AZ
                                     // There are ways to write this without [[maybe_unused]], but they are brittle.
                                     // For example, using #else could cause a bug where the function's parameter
                                     // is changed in #ifdef but not in #else.
-                                    [[maybe_unused]] const uint32_t drawPacketCount=AddLodDataToView(c->m_cullData.m_boundingSphere.GetCenter(), c->m_lodData, *worklistData->m_view);
+                                    [[maybe_unused]]
+                                    const uint32_t drawPacketCount = AddLodDataToView(c->m_cullData.m_boundingSphere.GetCenter(), c->m_lodData, *worklistData->m_view);
```
lumberyard-employee-dm, line 277: Why is the `[[maybe_unused]]` on the previous line?
It works, but it is unusual to put it on a line by itself.

amzn-tommy (approved):

Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
```diff
@@ -168,7 +168,10 @@ namespace AZ
             return worklistData;
         }
 
-        constexpr size_t WorkListCapacity = 5;
+        // Any value larger than 9 triggers a crash the job creation following the message:
+        // "Allocation size (560) is too big (max: 512) for pools!"
+        // (each additional count here increases the allocation size by 48, value of 9 makes the size exactly 512)
+        constexpr size_t WorkListCapacity = 9;
```
amzn-tommy, line 174: I'm curious what impact it would have to go in the other direction, but with the taskgraph. A WorkListCapacity of 1 +TG. The TaskGraph is supposed to have lower overhead, and the purpose of accumulating nodes into a worklist is to avoid job system overhead. I would expect being more distributed to help, especially on a ThreadRipper.

In my experiments with batching up draw call creation, I was able to achieve the best utilization with smaller batches using the TaskGraph, otherwise several threads/cores were just idling.

Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
antonmic: It was a long line and [[maybe_unused]] was just making it longer/harder to read. No real reason other than that

gadams3 (approved):

amzn-tommy (approved):

Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
```diff
@@ -168,30 +171,81 @@ namespace AZ
             return worklistData;
         }
 
-        constexpr size_t WorkListCapacity = 5;
-        using WorkListType = AZStd::fixed_vector<AzFramework::IVisibilityScene::NodeData, WorkListCapacity>;
+        // Used to accumulate NodeData into lists to be handed off to jobs for processing
+        struct WorkListType
+        {
+            AZStd::vector<AzFramework::IVisibilityScene::NodeData> m_nodes;
```
rgba16f, line 177: Would a fixed_vector or an array be a better choice here?

Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
```diff
@@ -168,30 +171,81 @@ namespace AZ
...
+        };
+
```
rgba16f, line 179: What's the reason to wrap this in a single member struct? Do we plan to add more members?

Gems/Atom/RPI/Code/Source/RPI.Public/RenderPipeline.cpp
```diff
@@ -470,8 +470,6 @@ namespace AZ
                 }
                 else if (pipelineViews.m_type == PipelineViewType::Persistent)
                 {
-                    // Reset persistent view: clean draw list mask and draw lists
-                    pipelineViews.m_views[0]->Reset();
```
rgba16f, line 474: why remove this? it's just clearing the draw list tag and the various draw list data pieces, I don't see any deallocation happening.

antonmic: View reset will reset the draw list context which will clear the array of arrays which means the capacity for all the arrays gets zeroed and adding new draw items will result in the usual re-allocations as the arrays grow.

Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
antonmic: Added a new variable to the struct

Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
antonmic: Size is controlled via a CVAR
Also is flexible depending on the job allocation strategy that is also controlled via CVAR

Gems/Atom/RPI/Code/Source/RPI.Public/RenderPipeline.cpp
rgba16f: Since clear doesn't reset the vector's m_end parameter the capacity is unchanged, m_last is changed so all the entries are removed.

amzn-tommy: The issue is that DrawListsByTag is an array of vectors. So if you clear the array, the vectors get destroyed, not cleared.

rgba16f: Array doesn't have a clear function, it's actually calling clear on all the contained vectors.

Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
```diff
@@ -205,164 +246,163 @@ namespace AZ
...
+            s32 startIdx = 0, size = s32(nodeData.m_entries.size());
+            const AZStd::vector<AzFramework::VisibilityEntry*>& entries = nodeData.m_entries;
+
+            if (worklistData->m_taskGraphEvent)
+            {
+                static const AZ::TaskDescriptor descriptor{ "AZ::RPI::ProcessWorklist", "Graphics" };
+
+                AZ::TaskGraph taskGraph{ "ProcessCullableEntries" };
+                AZ::TaskGraphEvent taskGraphEvent{ "ProcessCullableEntries Wait" };
```
rgba16f, line 336: The TaskGraphEvent looks to be unused

rgba16f (approved):

Gems/Atom/RPI/Code/Source/RPI.Public/RenderPipeline.cpp
amzn-tommy: My bad. It's actually a vector of unique_ptrs to arrays of vectors. So when the vector of unique_ptrs is cleared, the arrays are released.

Issue #12602, opened 2022-10-13
Script Canvas: Script Canvas component's label formatting gets scattered after assigning and reassigning scripts with exposed transform variables


michleza:
**Describe the bug**
If you assign and reassign script canvas files to a script canvas component the labels will begin to get off center or not removed from the component inspector.

**Steps to reproduce**
1. Open O3DE and Script canvas editor
2. Make a new script canvas file
3. Add several variables to the script canvas file. Make sure they are different types such as color and include a transform.
4. Expose the variables to component
5. Make a new entity in O3DE with a script canvas component 
6. Add the new script canvas file to the component
7. Remove the script canvas file from the component

**Expected behavior**
The component returns to its clean state. No leftover fields or labels

**Actual behavior**
Labels for transform appear to stay on the component. Continuing to add and remove the script canvas file ot the component will further add labels and make the formatting very difficult to understand

**Screenshots/Video**
![image](https://user-images.githubusercontent.com/98843754/195725869-f65823cb-1677-442c-b56e-69bb891b8746.png)
![image](https://user-images.githubusercontent.com/98843754/195725898-430e9f55-9a5f-456a-86d1-dd3b915ec873.png)

**Found in Branch**
development

**Desktop/Device (please complete the following information):**
 - Device:. PCWindows 10, 
 - CPU Ryzen 5950x, ]
 - GPU NVidia RTX 3070]
 - Memory [e.g. 32GB]

**Additional context**

lsemp3d:
This no longer happens

PR #12643, opened 2022-10-16
bugfix(ToolsFramework): ensure TransformRowWidget widget deletes additional components 

pollend:
ref: https://github.com/o3de/o3de/issues/12602

## What does this PR do?

this just makes sure that the modifications that are introduced with the TransformwWidget is cleaned up with the component is deleted.  this is because PropertyRowWidget is recycled and reused so any change to the component will need to be undone when the widget is return to the pool

## How was this PR tested?

follow step in the referenced issue

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyRowWidget.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyRowWidget.cpp
@@ -170,6 +170,7 @@ namespace AzToolsFramework
         m_handler = nullptr;
         m_isMultiSizeContainer = false;
         m_isFixedSizeOrSmartPtrContainer = false;
+        m_custom = false;
         if (m_selectionEnabled) 
         {
             SetSelected(false);
--- a/Code/Tools/SceneAPI/SceneUI/RowWidgets/TransformRowWidget.cpp
+++ b/Code/Tools/SceneAPI/SceneUI/RowWidgets/TransformRowWidget.cpp
@@ -103,12 +103,12 @@ namespace AZ
             TransformRowWidget::TransformRowWidget(QWidget* parent)
                 : QWidget(parent)
             {
-                QWidget* hider = new QWidget();
+                m_containerWidget = new QWidget();
                 QGridLayout* layout = new QGridLayout();
                 layout->setMargin(0);
                 QGridLayout* layout2 = new QGridLayout();
                 setLayout(layout);
-                AzToolsFramework::PropertyRowWidget* parentWidget = reinterpret_cast<AzToolsFramework::PropertyRowWidget*>(parent);
+                AzToolsFramework::PropertyRowWidget* parentWidget = static_cast<AzToolsFramework::PropertyRowWidget*>(parent);
                 QToolButton* toolButton = parentWidget->GetIndicatorButton();
                 QVBoxLayout* layoutOriginal = parentWidget->GetLeftHandSideLayoutParent();
                 parentWidget->SetAsCustom(true);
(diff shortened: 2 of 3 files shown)


pollend:
found there is some extra spacing: 
image

can't think of a why this is happening.

hultonha:
@pollend would it be possible to include a before/after screen shot to make the change clearer. Also tagging @cgalvan and @AMZN-daimini for visibility as they will likely be interested in this change and have more background on why things are the way they are.

hultonha:
Also tagging @lsemp3d for visibility

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyRowWidget.cpp
```diff
@@ -170,6 +170,7 @@ namespace AzToolsFramework
         m_handler = nullptr;
         m_isMultiSizeContainer = false;
         m_isFixedSizeOrSmartPtrContainer = false;
+        m_custom = false;
```
cgalvan, line 173: This looks unused

Code/Tools/SceneAPI/SceneUI/RowWidgets/TransformRowWidget.cpp
```diff
@@ -103,12 +103,12 @@ namespace AZ
             TransformRowWidget::TransformRowWidget(QWidget* parent)
                 : QWidget(parent)
             {
-                QWidget* hider = new QWidget();
```
cgalvan, line 106: If you parent `hider` to `this` I think you can omit the rest of the changes, since it will get deleted when `TransformRowWidget` gets deleted. It doesn't look like the `hider` widget is getting deleted/re-created whenever the custom row widget is returned to the pool, it should just be a one-time initialization.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyRowWidget.cpp
pollend: the flag is set by TransformRowWidget.cpp and it disables a part of the PropertyRowWidget. I have to reset the flag to restore PropertyRowWidget when it goes back to the pool.

Code/Tools/SceneAPI/SceneUI/RowWidgets/TransformRowWidget.cpp
pollend: I though the layout would change the owning parent.

pollend: I just tried this today, @cgalvan the parent is changed when you add it to the layout.

pollend:
just checked out from upstream development and opened the same script looks like the layout is consistent

image

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyRowWidget.cpp
cgalvan: OK sounds good!

Code/Tools/SceneAPI/SceneUI/RowWidgets/TransformRowWidget.cpp
cgalvan: I was originally going to say that this should just get cleaned up normally by the layout taking ownership, but I just realized the `TransformRowWidget` is actually modifying the layout of its parent, which is pretty yucky. But I think re-working that is outside the scope of this, so I think this is OK for now. We should look at this deeper in the future though @alexmontAmazon @dtamkin1

cgalvan (approved):

pollend:
@cgalvan I think I would like a way to create additional rows with the handler not sure how that would work but the way this works seems a bit invasive. the API probably needs to be more restrictive to guarantee that things are reset if we still would like to recycle the widget.

mbalfour-amzn (approved):

PR #12654, opened 2022-10-17
Improves AssetProcessorBatch startup time

lawsonamzn:
Also fixes a unit test that was always failing on mac (nobody was running the unit tests on mac).

## What does this PR do?

Improves the startup time of AP in general (but especially noticible when running the batch version) when all assets are already compiled.

This change comes from profiling the AP startup time when the cache is already mostly populated (Zero Analyis Mode).  It optimizes the functions that were consuming most of the time as well as does the file scanner regexing in a thread pool.

On this mac, the resulting difference was ~17 seconds before to 6 seconds after.   The profiler shows the remaining time now consists of starting the builder satellite executable (which itself is mostly loading and initializing all the gem modules).
Besides this, the most time consuming function is QString::toLower because of the file cache needing toLower its inputs in order to use them to look files up in the hash.

## How was this PR tested?

This was tested by running the AP Batch version with --zeroAnalysisMode parameter, as well as --project-path parameter to point it at AutomatedTesting.

The unit tests were also run, via 
```
AzTestRunner libAssetProcessor.Tests.dylib AzRunUnitTests
```
During this, a problem with one of the tests was discovered.  Since I don't like to check any change in that doesn't pass on the unit tests, part of this PR has the fix (canonical path interactions with the Intermediate Assets folder) as well.


Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
```diff
@@ -2634,35 +2645,26 @@ namespace AssetProcessor
             // examination occurs here.
             // first, is it a source or is it a product in the cache folder?
             QString normalizedPath = examineFile.m_fileName;
-
-            AZ_TracePrintf(AssetProcessor::DebugChannel, "ProcessFilesToExamineQueue: %s delete: %s.\n", examineFile.m_fileName.toUtf8().constData(), examineFile.m_isDelete ? "true" : "false");
-
-            // debug-only check to make sure our assumption about normalization is correct.
-            Q_ASSERT(normalizedPath == AssetUtilities::NormalizeFilePath(normalizedPath));
```
lawsonamzn, line 2641: Authors note: Github can't detect it but these lines were just moved downwards, after the fixup.

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
```diff
@@ -3067,8 +3074,6 @@ namespace AssetProcessor
         m_AssetProcessorIsBusy = true;
         Q_EMIT AssetProcessorManagerIdleState(false);
 
-        AZ_TracePrintf(AssetProcessor::DebugChannel, "AssesFileInternal: %s %s\n", normalizedFullFile.toUtf8().constData(), isDelete ? "true" : "false");
```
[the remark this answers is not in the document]
gadams3: Was most of the time lost to the logging or the conversion toUtf8?

gadams3 (approved):
Can't wait to see the impact of this launching all of the tools that connect to the AP.

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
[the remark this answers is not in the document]
lawsonamzn: > Was most of the time lost to the logging or the conversion toUtf8?

It was actually logging, according to the profiler - that is, the stack (including file IO and stuff) that happens inside AZ::Trace.  Interestingly though, on mac, this was not the case - there is something much more expensive about logging on windows than mac.

Code/Tools/AssetProcessor/native/AssetManager/assetScannerWorker.cpp
```diff
@@ -118,14 +124,28 @@ void AssetScannerWorker::ScanForSourceFiles(const ScanFolderInfo& scanFolderInfo
...
-            m_fileList.insert(AZStd::move(assetFileInfo));
+            // Entry is a file - add it to the check list which will be processed multi-threaded later.
+            m_checkForExclusionList.insert(AZStd::move(assetFileInfo));
         }
     }
 }
 
 void AssetScannerWorker::EmitFiles()
 {
+    // multi-threaded processing of regular expresisons.  Note that each call to IsFileExcluded
```
AMZN-stankowi, line 135: expresisons -> expressions

AMZN-stankowi (approved):
This is going to be a huge improvement for automated testing especially, I can't wait for this to get in.

Code/Tools/AssetProcessor/native/AssetManager/assetScannerWorker.cpp
```diff
@@ -101,15 +108,14 @@ void AssetScannerWorker::ScanForSourceFiles(const ScanFolderInfo& scanFolderInfo
...
-            m_excludedList.insert(AZStd::move(assetFileInfo));
-            continue;
-        }
-
         if (isDirectory)
         {
+            // Filtering out excluded directories immediately (not in a thread pool) since that prevents us from recursing.
+            if (m_platformConfiguration->IsFileExcluded(absPath))
+            {
+                m_excludedList.insert(AZStd::move(assetFileInfo));
```
jackalbe, line 116: So now m_excludedList is just an excluded list of folders, right? 

Or should should the m_excludedList be filled out with the file entries during the blockingFilteredReduced() ?

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
```diff
@@ -2634,35 +2645,26 @@ namespace AssetProcessor
...
                 {
                     // found in canonical location, update its normalized path
                     QString withoutCachePath = normalizedPath.mid(canonicalRootDir.length() + 1);
                     // the extra +1 is to consume the slash that is after the root dir.
                     normalizedPath = AssetUtilities::NormalizeFilePath(m_cacheRootDir.absoluteFilePath(withoutCachePath));
                 }
             }
-#endif // AZ_TRAIT_OS_PLATFORM_APPLE
+
+            AZ_TracePrintf(AssetProcessor::DebugChannel, "ProcessFilesToExamineQueue: %s delete: %s.\n", examineFile.m_fileName.toUtf8().constData(), examineFile.m_isDelete ? "true" : "false");
```
lmbr-pip, line 2660: Nit: To me its unclear what the printf is trying to say, I believe its more that "ProcessFilesToExamineQueue: File %s is marked for delete %s" 

* Not sure why spamming about all files here is useful esp as we are not currently deleting, or removing files.

Code/Tools/AssetProcessor/native/AssetManager/assetScannerWorker.cpp
```diff
@@ -118,14 +124,28 @@ void AssetScannerWorker::ScanForSourceFiles(const ScanFolderInfo& scanFolderInfo
...
+            // Entry is a file - add it to the check list which will be processed multi-threaded later.
+            m_checkForExclusionList.insert(AZStd::move(assetFileInfo));
         }
     }
 }
 
 void AssetScannerWorker::EmitFiles()
 {
+    // multi-threaded processing of regular expresisons.  Note that each call to IsFileExcluded
+    // checks the file path against every exclusion rule, which turns out to be a decent chunk of work
```
lmbr-pip, line 136: Nit: comment appears to be for the PR contributor, would suggest writing it for future maintainers. Not clear what "a decent chunk of work" means without context nor if this is a warning, general note etc. I would suggest removing or adding more context. Other comments of this nature have clear guidance to contributors.

Code/Tools/AssetProcessor/native/AssetManager/assetScannerWorker.cpp
junbo75: Seems like m_excludedList is a list of file entries (AssetFileInfo) instead of folders here.

junbo75 (commented):
Could we add any automation tests for this change or do we plan to backfill the tests in future PRs?

Code/Tools/AssetProcessor/native/AssetManager/assetScannerWorker.cpp
lawsonamzn: thats a really good question.  I didn't add m_excludedList, it was there before, but _whats it for???_

Code/Tools/AssetProcessor/native/AssetManager/assetScannerWorker.cpp
lawsonamzn: good idea, I'll reword or simplify

lawsonamzn:
> Could we add any automation tests for this change or do we plan to backfill the tests in future PRs?

The multithreaded regexing would already be covered by existing tests, as the actual input/outputs and API is unchanged, just the internal computations.  

The same goes for the other functions touched - they're such core functions (especially FindFirstMatchingFile) that they are invoked in many of the automated tests already and would already produce errors if broken by something I did here. 

 This is often the case when optimizing the body of a function instead of modifying its expected inputs/outputs or API.

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
lawsonamzn: Note that I didn't add this, it was present earlier.  GITHUB diff seems to think I deleted it and added it instead of just moved it around.

I think this is part of the scrub done a couple years ago so that the log file contains the absolute minimal amount of information to know what branches were taken.   For example you'll see no other output anywhere else that indicates that this file made it to the final analysis queue except if something goes wrong.

I'm hesitant to remove this... really, if we do, it should be part of a much bigger scrub that moves messages like this into some "debug trace" severity and turns them off by default unless you change a checkbox or config somewhere...

Code/Tools/AssetProcessor/native/AssetManager/assetScannerWorker.cpp
lawsonamzn: It looks like the excluded list is only used to add information about excluded files into the file system cache (the one you ask about modtimes and stuff).  But why?

lawsonamzn: I don't think we should be sending the excluded files into the file system cache at all.  I'll investigate some more.

Code/Tools/AssetProcessor/native/utilities/PlatformConfiguration.cpp
```diff
@@ -1569,29 +1572,49 @@ namespace AssetProcessor
 
         auto* fileStateInterface = AZ::Interface<AssetProcessor::IFileStateRequests>::Get();
 
-        QDir cacheRoot;
-        AssetUtilities::ComputeProjectCacheRoot(cacheRoot);
-
+        QString intermediateAssetsAsQString; // lazily computed on first use
```
amzn-mike, line 1575: This seems like it could just be inside the loop, it's only run once I believe since skip is set to false after the first iteration

amzn-mike (approved):

Code/Tools/AssetProcessor/native/utilities/PlatformConfiguration.cpp
```diff
@@ -1750,10 +1773,9 @@ namespace AssetProcessor
     //! Given a scan folder path, get its complete info
     const AssetProcessor::ScanFolderInfo* PlatformConfiguration::GetScanFolderByPath(const QString& scanFolderPath) const
     {
-        AZ::IO::Path scanFolderPathView(scanFolderPath.toUtf8().constData());
         for (int pathIdx = 0; pathIdx < m_scanFolders.size(); ++pathIdx)
         {
-            if (AZ::IO::PathView(m_scanFolders[pathIdx].ScanPath().toUtf8().constData()) == scanFolderPathView)
+            if (m_scanFolders[pathIdx].ScanPath() == scanFolderPath)
```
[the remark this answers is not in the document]
lumberyard-employee-dm: PathView comparison also deals with case insensitivity comparison and slash differences on Windows.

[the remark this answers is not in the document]
lawsonamzn: I'll double check this, but where we build all the scan folders, I saw it normalizing and absoluting it before storing it in the scanfolders.

Code/Tools/AssetProcessor/native/utilities/PlatformConfiguration.cpp
lawsonamzn: I'll move it

Code/Tools/AssetProcessor/native/AssetManager/assetScannerWorker.cpp
jackalbe: So it seems that the FileStateCache uses the file info for API such as FileStateCache::Exists() and FileStateCache::GetHash() so maybe the excluded files info is valid?

Code/Tools/AssetProcessor/native/utilities/PlatformConfiguration.cpp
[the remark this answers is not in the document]
lawsonamzn: I found one case where it wasn't normalized and fixed that.  I've also added an assert if someone calls AddScanFolder w/o normalizing first.

lawsonamzn:
turns out multithreading regex is slower on windows even if its a tad faster on Mac.   I'll strip that out.  the other thing I can do is convert regexes to globs which are significantly faster, and prune any non-relevant ones.

Code/Tools/AssetProcessor/native/utilities/PlatformConfiguration.cpp
lawsonamzn: so it can't be in the loop, but it can be computed up front, once.  The result (intermediateAssetsAsQstring) has to survive looping, because its not necessarily the case that the first scan folder is the intermediate assets folder, so it could use it over and over until it finds it.

lawsonamzn:
I added a test to cover the new API that pre-warms the excluded folder cache, and de-recursed the file scanner, allowing the caching of more things.
I also added a cheaper version of the exclusion test and updated the datafiles to use globs instead of regexes where they could.

Note the same number of exclusions happen before and after this change, no difference in # of files processed, telling us that the exclusions are working.

lawsonamzn:
Latest perf:
Asset Processor runs Windows X64

Before
AssetProcessor: Total Assets Processing Time: 30.469002s
AssetProcessor: Total Assets Processing Time: 30.724001s
AssetProcessor: Total Assets Processing Time: 30.502001s

AFTER:

AssetProcessor: Total Assets Processing Time: 14.927001s
AssetProcessor: Total Assets Processing Time: 15.169001s
AssetProcessor: Total Assets Processing Time: 15.019001s

Code/Tools/AssetProcessor/native/AssetManager/assetScannerWorker.cpp
lawsonamzn: I've kept the exclusion list working as it did before, essentially.  We can examine getting rid of it later, but I did other optimizations such that it doesn't actually matter anymore in terms of time taken.

Code/Tools/AssetProcessor/native/AssetManager/assetScannerWorker.cpp
lawsonamzn: Fixed.  More like, removed.  We don't try to multihtread, its actually slower to do so.  The cuhnks of work are actually too small :(

lawsonamzn:
> Could we add any automation tests for this change or do we plan to backfill the tests in future PRs?

I've added a test of the cache - the rest of the functions have been optimized but behave as before, and are covered by existing tests.

AMZN-stankowi (approved):

Code/Tools/AssetProcessor/native/AssetManager/assetScannerWorker.cpp
```diff
@@ -74,64 +79,118 @@ void AssetScannerWorker::ScanForSourceFiles(const ScanFolderInfo& scanFolderInfo
...
+            QString absPath = entry.absoluteFilePath();
+            const bool isDirectory = entry.isDir();
+            QDateTime modTime = entry.lastModified();
+            AZ::u64 fileSize = isDirectory ? 0 : entry.size();
+            AssetFileInfo assetFileInfo(absPath, modTime, fileSize, &rootScanFolder, isDirectory);
+            QString relPath = absPath.mid(rootScanFolder.ScanPath().length() + 1);
+
+            if (isDirectory)
+            {
+                // in debug, assert that the paths coming from qt directory info iteration is already normalized
```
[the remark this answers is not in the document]
lawsonamzn: The problem is that  these checks actually cost a lot of time and its the hot path of the application.
AssetUtilities::NormalizeDirectoryPath(absPath) is not free and is in fact what we're trying to avoid having to do.

Code/Tools/AssetProcessor/native/utilities/PlatformConfiguration.cpp
```diff
@@ -1241,10 +1241,13 @@ namespace AssetProcessor
...
+
+                QString watchFolderPath = QString::fromUtf8(scanFolderEntry.m_watchPath.c_str());
+                watchFolderPath = AssetUtilities::NormalizeDirectoryPath(watchFolderPath);
                 AddScanFolder(ScanFolderInfo(
-                    QString::fromUtf8(scanFolderEntry.m_watchPath.c_str(), aznumeric_cast<int>(scanFolderEntry.m_watchPath.Native().size())),
-                    QString::fromUtf8(scanFolderEntry.m_scanFolderDisplayName.c_str(), aznumeric_cast<int>(scanFolderEntry.m_scanFolderDisplayName.size())),
-                    QString::fromUtf8(scanFolderEntry.m_scanFolderIdentifier.c_str(), aznumeric_cast<int>(scanFolderEntry.m_scanFolderIdentifier.size())),
+                    watchFolderPath,
+                    QString::fromUtf8(scanFolderEntry.m_scanFolderDisplayName.c_str()),
+                    QString::fromUtf8(scanFolderEntry.m_scanFolderIdentifier.c_str()),
```
lumberyard-employee-dm, line 1250: Removing the size parameter would actually make this code slightly slower, since the string length would have to be recalculated again.

lawsonamzn: ah, nicely spotted.  i'll fix

lawsonamzn:
I had to resolve conflicts via rebase/forcepush.   I'll kick off another build once it completes.

Code/Tools/AssetProcessor/native/utilities/PlatformConfiguration.cpp
```diff
@@ -1462,9 +1465,20 @@ namespace AssetProcessor
...
+            if (source.ScanFolderID() == 0)
+            {
+                m_scanFolders.back().SetScanFolderID(m_scanFolders.size() - 1);
+            }
             return;
         }
 
+        AZ_Assert(
+            source.ScanPath() == AssetUtilities::NormalizeDirectoryPath(source.ScanPath()),
+            "AddScanFolder expects normalized name input but %s was given instead.",
```
jackalbe, line 1479: So this is condition comes from bad input? If so, why not just skip it and show an error? Is this really a coding error that requires a programmer to fix?

lawsonamzn: I'm not quite sure what you would like to change here.   It is indeed a programming error (somewhere in AP) if we ever end up with a scan folder being added that isn't normalized, since all the places that call this function already normalize it beforehand, and thus this assert never triggers but future proofs it for if someone adds scan folders (in c++ code) by some other new means.

However, maybe I can change this so that the scan folder input isn't const, or I could copy it and normalize it here, instead of trying to assert, so it simply can't be wrong.  (This is not a bottleneck / hot path in the app)

lawsonamzn: I'll just erase the assert.  It turns out all data members of ScanFolderInfo class are read-only (ie, there are no setters) and the constructor normalizes the path from whatever was passed in, so its impossible for this assert to trigger.

lumberyard-employee-dm (approved):

lawsonamzn:
this failed due to an intermittent test problem.  Once dev is green again I'll try again.

lawsonamzn:
Turns out the build system needed me to rebase onto main to get files that the jenkins nodes expect, that aren't there from older branches.  will rebase, then retest.

lawsonamzn:
Rebased without any conflicts or similar issues.

lawsonamzn:
This failed due to linux node running out of file handles.   This is not caused by this change.
But it needs to be fixed

lawsonamzn:
https://github.com/o3de/o3de/pull/13651 will fix the file handles thing, this is blocked till that one goes in

lawsonamzn:
https://github.com/o3de/o3de/pull/13651 is now fixed, I can return to this.

Looks like another issue is that due to the Intermediate Assets feature, the asset processor batch no longer uses the file hash cache (ie, it rehashes every file every time).  I need to now fix that too, as part of this, or else only the GUI version will improve.

lawsonamzn:
I updated it so that the batch version of AP uses the file cache.  It doesn't miss intermediates because I created a signal for intermediate emission and directly connected it to the file cache invalidation.  (The file cache has a mutex).  This means that for intermediates this is actually even more reliable than the file monitor.
 
I also removed a test which was useles.  It was creating a file on disk and then asserting that the file cache instantly knew about the file.  This is not possible without something telling the file cache (or if the file cache is a dummy passthru like it was previously).   Even an OS file monitor runs async and takes some non zero amount of time to notify.  And that file monitor has its own tests.

Code/Tools/AssetProcessor/native/utilities/ApplicationManagerBase.cpp
```diff
@@ -1445,8 +1447,39 @@ bool ApplicationManagerBase::Activate()
 
     InitFileStateCache();
     InitFileProcessor();
-
     InitUuidManager();
+
+    // now that apm, statecache, processor, and uuid manager are all alive, hook them up to the signal that AP
+    // gives when it modifies an intermediate asset.  For the file cache, we hook it up directly so that there is
+    // no delay between the notificatin and the invalidation/creation of its cache entry.
```
AMZN-stankowi, line 1454: nitpick typo: notificatin -> notification

AMZN-stankowi (approved):

lawsonamzn:
rebased on dev.  Trying again.

lawsonamzn:
Its still failing for seemingly unrelated reasons.

PR #12655, opened 2022-10-18
Automatically set the matching Diffuse or Specular IBL image in the Editor

dmcdiarmid-ly:
This change will automatically set the matching Diffuse or Specular IBL image in the EditorImageBasedLightComponent.

Tested Editor DX12/Vulkan

Signed-off-by: dmcdiarmid-ly <63674186+dmcdiarmid-ly@users.noreply.github.com>

### DIFF
--- a/Gems/AtomLyIntegration/CommonFeatures/Code/Source/ImageBasedLights/EditorImageBasedLightComponent.cpp
+++ b/Gems/AtomLyIntegration/CommonFeatures/Code/Source/ImageBasedLights/EditorImageBasedLightComponent.cpp
@@ -28,7 +28,11 @@ namespace AZ
             if (AZ::SerializeContext* serializeContext = azrtti_cast<AZ::SerializeContext*>(context))
             {
                 serializeContext->Class<EditorImageBasedLightComponent, BaseClass>()
-                    ->Version(1, ConvertToEditorRenderComponentAdapter<1>);
+                    ->Version(2, ConvertToEditorRenderComponentAdapter<1>)
+                    ->Field("diffuseImageAsset", &EditorImageBasedLightComponent::m_diffuseImageAsset)
+                    ->Field("specularImageAsset", &EditorImageBasedLightComponent::m_specularImageAsset)
+                    ->Field("exposure", &EditorImageBasedLightComponent::m_exposure)
+                    ;
 
                 if (AZ::EditContext* editContext = serializeContext->GetEditContext())
                 {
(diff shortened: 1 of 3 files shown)


antonmic (approved):

Gems/AtomLyIntegration/CommonFeatures/Code/Source/ImageBasedLights/EditorImageBasedLightComponent.cpp
```diff
@@ -75,5 +83,74 @@ namespace AZ
...
+
+                assetId = AZ::RPI::AssetUtils::GetAssetIdForProductPath(assetPath.c_str(), AZ::RPI::AssetUtils::TraceLevel::None);
+                if (!assetId.IsValid())
+                {
+                    // unable to locate the matching image asset
+                    return false;
+                }
+            }
+
+            if (asset2.Get())
```
gadams3, line 137: Check against the asset ID instead of the asset pointer in case the asset was never queued or loaded.

```suggestion
            if (asset2.GetId().IsValid())
```

Gems/AtomLyIntegration/CommonFeatures/Code/Source/ImageBasedLights/EditorImageBasedLightComponent.cpp
```diff
@@ -75,5 +83,74 @@ namespace AZ
...
+            const char* asset2Suffix,
+            const char* asset2Name)
+        {
+            AZStd::string assetPath = asset1.GetHint();
+            AZ::Data::AssetId assetId;
+
+            if (!assetPath.empty())
+            {
+                // try to load the matching image asset by replacing the asset1 suffix with the asset2 suffix
+                AZ::StringFunc::Replace(assetPath, asset1Suffix, asset2Suffix);
```
gadams3, line 127: I recommend checking the asset ID first and getting the path from the asset ID instead of using the hint because the hint will not always be set.

Gems/AtomLyIntegration/CommonFeatures/Code/Source/ImageBasedLights/EditorImageBasedLightComponent.cpp
dmcdiarmid-ly: Done.

Gems/AtomLyIntegration/CommonFeatures/Code/Source/ImageBasedLights/EditorImageBasedLightComponent.cpp
dmcdiarmid-ly: Done.

gadams3 (approved):

PR #12657, opened 2022-10-18
refactor: fix crash for CFolderTreeCtrl and cleanup logic

pollend:
ref: https://github.com/o3de/o3de/issues/12270

## What does this PR do?

this resolves the ownership problems with CFolderTreeCtrl. This should resolve the crash problem observed in the referenced ticket. I also migrated the logic from the standard library and removed a lot of additional methods. I think the logic should be more clear with these changes. 

## How was this PR tested?

follow the steps in the referenced ticket.

### DIFF
--- a/Code/Editor/Controls/FolderTreeCtrl.cpp
+++ b/Code/Editor/Controls/FolderTreeCtrl.cpp
@@ -16,7 +16,8 @@
 #include <QStandardItemModel>
 
 #include <AzCore/std/algorithm.h>
-// AzQtComponents
+#include <AzCore/IO/SystemFile.h>
+
 #include <AzQtComponents/Utilities/DesktopUtilities.h> // for AzQtComponents::ShowFileOnDesktop
 
 //////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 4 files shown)


cgalvan (approved):

Code/Editor/Controls/FolderTreeCtrl.cpp
```diff
@@ -16,87 +15,76 @@
...
+    m_folders.clear();
+    AZStd::ranges::copy_if(
+        folders,
+        AZStd::back_inserter(m_folders),
+        [](const auto& path)
+        {
+            if (CFileUtil::PathExists(path) || Path::IsFolder(path.toLocal8Bit().constData()))
+            {
+                return true;
+            }
```
mbalfour-amzn, line 52: Suggestion: Consider using AZ::IO::SystemFile::Exists(path) and IsDirectory(path) instead, it would be nice if we could eventually deprecate FileUtil / PathUtil since they overlap with AzCore/AzFramework methods.

Code/Editor/Controls/FolderTreeCtrl.cpp
```diff
@@ -16,87 +15,76 @@
...
+    for (auto& item: m_folders)
+    {
+        if (!item.isEmpty())
+        {
+            LoadTreeRec(item);
+        }
+    }
 
-    InitTree();
+    if (!m_bDisableMonitor)
```
mbalfour-amzn, line 84: Suggestion: rename the variable to "m_bEnableMonitor" and flip the logic. Every time it's used it's causing a double-negative ("if not disabled"), which is a bit harder to understand than a single-positive ("if enabled"). :)

mbalfour-amzn (approved):

pollend:
looks like it crashed, the error seems unrelated to anything that was changed here?

cgalvan:
> looks like it crashed, the error seems unrelated to anything that was changed here?

Yeah I can't tell if it ran out of memory or what, but like you said it looks un-related to your changes, so I've kicked off a clean build that should (hopefully) do the trick.

PR #12680, opened 2022-10-19
Attribute Patch handling for add/remove/morphing property editors

dtamkin1:
## What does this PR do?

Appropriately check and update attributes when a property editor is added, removed, or morphed.
Implemented a map in the DPERowWidget that keeps track of a dom index and the attributes associated with that index.
Reworked attributes in the DPELayout and DPERowWidget to utilize that map.

## How was this PR tested?

Locally with the inspector, asset editor, and DPEDebugViewStandalone


Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
```diff
@@ -545,6 +575,120 @@ namespace AzToolsFramework
...
+        auto foundEntry = m_domOrderToAttributeInfo.find(domIndex);
+        if (foundEntry != m_domOrderToAttributeInfo.end())
+        {
+            return foundEntry->second;
+        }
+        else
+        {
+            return new AttributeInfo;
+        }
+    }
```
dtamkin1, line 670: I think ideally I would want to return the attributes only if they exist, since this creates a default AttributeInfo every time, so maybe have a separate method that returns true if an attribute exists at this domIndex?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
```diff
@@ -325,54 +336,72 @@ namespace AzToolsFramework
...
+                    }
+                    // If the widget is the second member of the group, remove the first widget in the group
+                    else if (currentGroupIndex == 1)
+                    {
+                        currentGroup.erase(currentGroup.begin());
+                        return;
+                    }
+                    else
+                    {
+                        //! to-do: handle case where we remove widget from the middle of a share prior group
```
dtamkin1, line 387: This is the case that needs to be handled for shared groups larger than 4 elements, because if the second element or third element is removed from this group, then we need to update behavior according to what the user would expect to happen.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
```diff
@@ -696,6 +841,7 @@ namespace AzToolsFramework
                         // CreateWidgetForHandler will add a new entry to m_widgetToPropertyHandlerInfo, kill the old entry
                         GetDPE()->ReleaseHandler(AZStd::move(foundEntry->second.hanlderInterface));
                         m_widgetToPropertyHandlerInfo.erase(foundEntry);
+                        m_columnLayout->RemoveSharePriorColumn(childIndex);
```
alexmontAmazon, line 844: this handles the morph case, and I see that AddColumnWidget handles the add case, but what about if the attributes just change? Like, what if the handler type stays the same, but suddenly it wants to sharePriorColumn? Is that handled somewhere I'm missing? It seems like there's code missing below in the area of "// handler is the same, set the existing handler with the new value"

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
alexmontAmazon: I probably would just return null here, so the caller knows it doesn't exist yet, and then let them make a new one themselves, if that's reasonable. It's not like this else case is doing anything special...

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
dtamkin1: Yeah I wasn't sure whether we needed to add attribute checks there or not, since I wasn't sure of instances where attributes can change if the handler type stays the same. I opted not to because it was using that code path a lot and we were making those checks every single time but never doing anything. 
The SetPropertyEditorAttributes method can handle checking attributes for a domValue and a domIndex, so it just be added there :)

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
dtamkin1: Good call, will do that

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
dtamkin1: I'd like to discuss it, but I have an idea in theory of what we could do here
This is really important to do correctly for shared groups to scale, but for now I left it since I haven't seen anything in the inspector/asset browser have more than 3 elements in a shared group

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
```diff
@@ -814,27 +1025,14 @@ namespace AzToolsFramework
...
-            auto shouldDisable = AZ::Dpe::Nodes::PropertyEditor::Disabled.ExtractFromDomNode(domValue).value_or(false);
-
-            // if this row doesn't already have a tooltip, use the first valid
-            // tooltip from a child PropertyEditor (like the RPE)
-            if (!descriptionString.empty() && toolTip().isEmpty())
-            {
-                setToolTip(QString::fromUtf8(descriptionString.data(), aznumeric_cast<int>(descriptionString.size())));
-            }
+            //auto descriptionString = AZ::Dpe::Nodes::PropertyEditor::Description.ExtractFromDomNode(domValue).value_or("");
+            //auto shouldDisable = AZ::Dpe::Nodes::PropertyEditor::Disabled.ExtractFromDomNode(domValue).value_or(false);
```
amzn-tmryan, line 1029: Are these supposed to be removed or uncommented?

dtamkin1: Good catch, they're supposed to be removed. Fixed

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
```diff
@@ -325,54 +342,95 @@ namespace AzToolsFramework
...
-    bool DPELayout::ShouldSharePrior()
+    void DPELayout::RemoveSharePriorColumn(size_t widgetIndex)
     {
-        return m_shouldSharePrior;
+        for (auto groupIt = m_sharePriorColumn.begin(); groupIt != m_sharePriorColumn.end(); ++groupIt)
+        {
+            AZStd::vector<size_t> currentGroup = *groupIt;
+            for (int currentGroupIndex = 0; currentGroupIndex < currentGroup.size(); currentGroupIndex++)
+            {
+                if (widgetIndex == currentGroup[currentGroupIndex])
```
cgalvan, line 368: Nit: Since all the below logic is underneath this conditional, could change it to be an early out for a continue to save some tab spacing:

```suggestion
                if (widgetIndex != currentGroup[currentGroupIndex])
                {
                    continue;
                }
                
                ...
```

cgalvan (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
dtamkin1: That early out will be helpful! Just added it, thannks

alexmontAmazon (approved):
looks good, and it sounds like @amzn-tmryan tested it with his read-only attribute work, so I'm glad to hear that attribute patching without widget morphing is working. Good work!

amzn-tmryan (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
```diff
@@ -545,6 +608,193 @@ namespace AzToolsFramework
...
+                        childWidget->setToolTip(QString::fromUtf8(descriptionString.data(), aznumeric_cast<int>(descriptionString.size())));
+                    }
+                    foundEntry->second->m_descriptionString = descriptionString;
+                }
+
+                // Check Disabled
+                if (shouldDisable != foundEntry->second->m_shouldDisable)
+                {
+                    foundEntry->second->m_shouldDisable = shouldDisable;
+                }
```
amzn-tmryan, line 760: I believe this needs an else case to cover when the Disabled attribute is removed.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
```diff
@@ -545,6 +608,193 @@ namespace AzToolsFramework
...
+                newAttribute->m_descriptionString = descriptionString;
+                validAttribute = true;
+            }
+
+            // Check Disabled
+            if (shouldDisable)
+            {
+                childWidget->setEnabled(false);
+                validAttribute = true;
+            }
```
amzn-tmryan, line 680: Here too, or maybe just here? One or both of these needs an else case for re-enabling a widget when the Disabled attribute is cleared.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
```diff
@@ -545,6 +608,193 @@ namespace AzToolsFramework
...
     }
 
+    void DPERowWidget::SetPropertyEditorAttributes(size_t domIndex, const AZ::Dom::Value& domArray, QWidget* childWidget)
+    {
+        // Extract all attributes from dom value
+        auto alignment = AZ::Dpe::Nodes::PropertyEditor::Alignment.ExtractFromDomNode(domArray).value_or(AZ::Dpe::Nodes::PropertyEditor::Align::UseDefaultAlignment);
+        auto sharePrior = AZ::Dpe::Nodes::PropertyEditor::SharePriorColumn.ExtractFromDomNode(domArray).value_or(false);
+        auto minimumWidth = AZ::Dpe::Nodes::PropertyEditor::UseMinimumWidth.ExtractFromDomNode(domArray).value_or(false);
+        auto descriptionString = AZ::Dpe::Nodes::PropertyEditor::Description.ExtractFromDomNode(domArray).value_or("");
+        auto shouldDisable = AZ::Dpe::Nodes::PropertyEditor::Disabled.ExtractFromDomNode(domArray).value_or(false);
```
amzn-tmryan, line 618: While we're at it, there actually is another attribute we need to watch for here. It totally slipped my mind when you asked me yesterday; apologies.

This needs to also check for `PropertyEditor::AncestorDisabled`. Should be able to || it together with the `Disabled` attribute here.

Issue #12665, opened 2022-10-18
Motion Event Inspector window does not update itself with changes until the modified motion event is deselected


LB-MateuszSzweda:
**Describe the bug**
Any changes provided to the selected **motion event** by using its **Inspector** window will not update the window and the event itself until the modified motion event or motion is deselected and selected again in **Animation Editor** .

Please note that in order to update the **Inspector** window with new _Event Data_ brackets only the **motion event** has to be deselected and selected again but to update event _Start_ and _End time_ modifier the whole motion has to be deselected and selected again in the **Motion Set** window.

For further information please refer to the attached video file.

**Prerequisites**

**EMotion FX Animation** Gem is enabled and built for the active project.

**Steps to reproduce**

1. Launch the Editor and create a new level.
2. Open the **Animation Editor** tool.
3. Change its layout to _Animgraph_ and navigate to the _Motion Sets_ tab.
4. Add any motion to the **Motion Set** window (for instance _rin_Idle.fbx_).
5. Select the added motion and add a new event track in its **Time View** window.
6. Right-click the track and select _Add motion event_ option.
7. Click the added motion event and navigate to the **Inspector** window.
8. Add **TwoStringEventData** or **EventDataFootIK** to _Event Data_ bracket.
9. Modify _Start_ and _End time_ modifiers.
10. Deselect the Motion Event or the Motion itself and select it again.

**Expected behavior**
Motion Event Inspector window updates itself and the Motion event when changes are introduced to it.

**Actual behavior**
Motion Event Inspector window and the Motion itself are updated only after the Inspector window or motion is deselected and selected again.

**Screenshots/Video**

https://user-images.githubusercontent.com/86953456/196435390-15423fcd-052e-48de-9f88-ab22dcb3f576.mp4

**Found in Branch**
Development (b47b11d)

**Desktop/Device:**
- Device: PC
- OS: Windows
- Version: 10 Pro
- CPU: AMD Ryzen 9 3900X
- GPU: NVIDIA GeForce RTX 2060 SUPER
- Memory: 16GB

hultonha:
This should now be fixed @LB-MateuszSzweda (see commit https://github.com/o3de/o3de/commit/ccba544b5f9b4e581012c6c025fede384b849faf)

LB-MateuszSzweda:
Unfortunately the issue persisted for me even after I've merged Development (92af95a) with Development (ccba544) via Git fetch/Git merge actions and rebuilding the engine afterwards.
[...]

PR #12692, opened 2022-10-19
bugfix(EmotionFX): update Motion Event Inspector when event data has changed

pollend:
ref: https://github.com/o3de/o3de/issues/12665

## What does this PR do?

I added some logic to update the event data when its changed. not sure if this is the best way to approach this, but I think this should handle the case where an event is update remotely/deleted since this logic will just re-query the data from TimerViewerPlugin.cpp.

There is quite a lot here, I think can be refactored. This can be pretty hard to follow. 

## How was this PR tested?

follows steps in linked PR.

### DIFF
--- a/Gems/EMotionFX/Code/EMotionFX/Source/Event.cpp
+++ b/Gems/EMotionFX/Code/EMotionFX/Source/Event.cpp
@@ -16,6 +16,17 @@ namespace EMotionFX
 {
     AZ_CLASS_ALLOCATOR_IMPL(Event, MotionEventAllocator, 0)
 
+    Event::Event(const Event& data)
+        : m_eventDatas(data.m_eventDatas)
+    {
+    }
+
+    Event::Event(Event&& event)
+        : m_eventDatas(AZStd::move(event.m_eventDatas)),
+          m_eventDatasChangeEvent(AZStd::move(event.m_eventDatasChangeEvent))
+    {
+    }
+
     Event::Event(EventDataPtr&& data)
         : m_eventDatas{ AZStd::move(data) }
     {
@@ -24,6 +35,18 @@ namespace EMotionFX
     Event::Event(EventDataSet&& datas)
         : m_eventDatas(AZStd::move(datas))
     {
+    }
+    
+    Event& Event::operator=(const Event& other) 
+    {
+        m_eventDatas = other.m_eventDatas;
+        return *this;
+    }
+
+    Event& Event::operator=(Event&& other) 
+    {
+        m_eventDatas = AZStd::move(other.m_eventDatas);
+        return *this;
     }
 
     void Event::Reflect(AZ::ReflectContext* context)
@@ -65,30 +88,30 @@ namespace EMotionFX
     void Event::AppendEventData(EventDataPtr&& newData)
     {
         m_eventDatas.emplace_back(AZStd::move(newData));
-        m_eventContainer.m_eventDatasChangeEvent.Signal();
+        m_eventDatasChangeEvent.Signal();
     }
 
     void Event::RemoveEventData(size_t index)
     {
         m_eventDatas.erase(AZStd::next(m_eventDatas.begin(), index));
-        m_eventContainer.m_eventDatasChangeEvent.Signal();
+        m_eventDatasChangeEvent.Signal();
     }
 
     void Event::SetEventData(size_t index, EventDataPtr&& newData)
     {
         m_eventDatas[index] = AZStd::move(newData);
-        m_eventContainer.m_eventDatasChangeEvent.Signal();
+        m_eventDatasChangeEvent.Signal();
     }
 
     void Event::InsertEventData(size_t index, EventDataPtr&& newData)
     {
         m_eventDatas.emplace(AZStd::next(m_eventDatas.begin(), index), AZStd::move(newData));
-        m_eventContainer.m_eventDatasChangeEvent.Signal();
+        m_eventDatasChangeEvent.Signal();
     }
 
     void Event::SetEventDataChangeEvent(Event::EventDataChangeEvent::Handler& handler)
     {
-        handler.Connect(m_eventContainer.m_eventDatasChangeEvent);
+        handler.Connect(m_eventDatasChangeEvent);
     }
 
 } // end namespace EMotionFX
--- a/Gems/EMotionFX/Code/EMotionFX/Source/Event.cpp
+++ b/Gems/EMotionFX/Code/EMotionFX/Source/Event.cpp
@@ -65,21 +65,30 @@ namespace EMotionFX
     void Event::AppendEventData(EventDataPtr&& newData)
     {
         m_eventDatas.emplace_back(AZStd::move(newData));
+        m_eventContainer.m_eventDatasChangeEvent.Signal();
     }
 
     void Event::RemoveEventData(size_t index)
     {
         m_eventDatas.erase(AZStd::next(m_eventDatas.begin(), index));
+        m_eventContainer.m_eventDatasChangeEvent.Signal();
     }
 
     void Event::SetEventData(size_t index, EventDataPtr&& newData)
     {
         m_eventDatas[index] = AZStd::move(newData);
+        m_eventContainer.m_eventDatasChangeEvent.Signal();
     }
 
     void Event::InsertEventData(size_t index, EventDataPtr&& newData)
     {
         m_eventDatas.emplace(AZStd::next(m_eventDatas.begin(), index), AZStd::move(newData));
+        m_eventContainer.m_eventDatasChangeEvent.Signal();
+    }
+
+    void Event::SetEventDataChangeEvent(Event::EventDataChangeEvent::Handler& handler)
+    {
+        handler.Connect(m_eventContainer.m_eventDatasChangeEvent);
     }
 
 } // end namespace EMotionFX
--- a/Gems/EMotionFX/Code/EMotionFX/Source/Event.cpp
+++ b/Gems/EMotionFX/Code/EMotionFX/Source/Event.cpp
@@ -45,6 +45,7 @@ namespace EMotionFX
 
     Event& Event::operator=(Event&& other) 
     {
+        m_eventDatasChangeEvent = AZStd::move(other.m_eventDatasChangeEvent);
         m_eventDatas = AZStd::move(other.m_eventDatas);
         return *this;
     }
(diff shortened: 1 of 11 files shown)


Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/MotionEvents/MotionEventWidget.cpp
```diff
@@ -36,9 +40,13 @@ namespace EMStudio
...
 
-
     void MotionEventWidget::ReInit(EMotionFX::Motion* motion, EMotionFX::MotionEvent* motionEvent)
     {
+        m_changeEventHandler.Disconnect();
+        if (motionEvent)
+        {
+            motionEvent->SetEventDataChangeEvent(m_changeEventHandler);
+        }
         m_editor->SetMotionEvent(motion, motionEvent);
```
hultonha, line 50: question: Should this be moved inside the null check too?

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/TimeView/TimeViewPlugin.h
```diff
@@ -182,6 +182,7 @@ namespace EMStudio
         void MotionEventTrackChanged(size_t eventNr, float startTime, float endTime, const char* oldTrackName, const char* newTrackName)            { UnselectAllElements(); CommandSystem::CommandHelperMotionEventTrackChanged(eventNr, startTime, endTime, oldTrackName, newTrackName); }
         void OnManualTimeChange(float timeValue);
         void toggleMotionEventPresetsPane();
+        void onRefreshSelection();
```
hultonha, line 185: update: I think as we're not overriding here this should be of the form `OnRefreshSelection` to be consistent with the coding standards

Gems/EMotionFX/Code/EMotionFX/Source/Event.h
```diff
@@ -42,7 +45,21 @@ namespace EMotionFX
...
+        void SetEventDataChangeEvent(Event::EventDataChangeEvent::Handler& handler);
     protected:
+        struct EventContainer {
+            EventContainer() = default;
+             
+            // event container is not copyable 
+            EventContainer(const EventContainer&) {}
+            EventContainer& operator=(const EventContainer&) {
+                return *this;
+            }
```
hultonha, line 57: update: This doesn't look correct to me, these operations are not correct, should we `=delete` them if they can't be used. Currently they'll run but do the wrong thing (not actually copy anything)

hultonha (commented):
Thanks very much for taking a look at this @pollend, there are a couple of things I've got questions about it would be good to sanity check. Would you also please be able to include a video of how the operation behaves now to show it compared to that described in the bug. Thanks!

amzn-rhhong:
I will review this later this week. Awesome work @pollend

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/MotionEvents/MotionEventWidget.cpp
pollend: This is the existing logic. It will clear of the data  is null. The api is not clear so we will have to revisit this I think

Gems/EMotionFX/Code/EMotionFX/Source/Event.h
pollend: I'm cheating a bit event is not a copyable type is in stubbing the copy and assignment.

Gems/EMotionFX/Code/EMotionFX/Source/Event.h
```diff
@@ -42,7 +45,21 @@ namespace EMotionFX
         void SetEventData(size_t index, EventDataPtr&& newData);
         void InsertEventData(size_t index, EventDataPtr&& newData);
 
+        void SetEventDataChangeEvent(Event::EventDataChangeEvent::Handler& handler);
     protected:
+        struct EventContainer {
```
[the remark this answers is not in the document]
amzn-rhhong: Why are you copying an event? I think AZ::event aren't supposed to be copied by design. EventHanderA is suppose to only handle EventA. If there is an EventB that copied from EventA, it will trigger undefined behavior.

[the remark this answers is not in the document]
pollend: I'm not. The events are discarded when the emfx event is copied. It's done in a couple places.

Gems/EMotionFX/Code/EMotionFX/Source/Event.h
hultonha: Yeah I don't think we should do this, there definitely has to be another option

hultonha: I think in this case we should provide a proper implementation of the Event copy/move constructor and move/assignment operator to correctly clone the event and its callback, rather than this cheat. This is much like the same thing as `std::unique_ptr` which isn't copyable but you can of course do a deep copy of the underlying value if supported which I think should be possible here right? Or we need to make `Event` itself non-copyable

pollend: I end up having the implement the copy constructor for both Event and MotionEvent I think just cheating a bit here but yes this is probably the correct thing I should of done originally.

hultonha: Yeah in cases like this the rule of 5 is the way to go (implement all special members) - you can use the `swap` idiom to implement some in terms of the others but it's up to you what you decide - this talk on the special members is really great too if you haven't seen it - https://www.youtube.com/watch?v=vLinb2fgkHk.

pollend: yep, I ended up implementing all 5 operators.

hultonha: Okay excellent thanks @pollend, I'll take another look shortly

Gems/EMotionFX/Code/EMotionFX/Source/Event.cpp
```diff
@@ -25,6 +36,18 @@ namespace EMotionFX
...
+    {
+        m_eventDatas = other.m_eventDatas;
+        return *this;
+    }
+
+    Event& Event::operator=(Event&& other) 
+    {
+        m_eventDatas = AZStd::move(other.m_eventDatas);
+        return *this;
+    }
```
hultonha, line 50: It looks like `m_eventDatasChangeEvent` is missing here?

To be honest we really should write some tests for this logic to ensure all the right things are getting copied/moved as expected otherwise we won't be sure this is working correctly

Gems/EMotionFX/Code/EMotionFX/Source/Event.h
```diff
@@ -42,7 +48,15 @@ namespace EMotionFX
         void SetEventData(size_t index, EventDataPtr&& newData);
         void InsertEventData(size_t index, EventDataPtr&& newData);
 
+        void SetEventDataChangeEvent(Event::EventDataChangeEvent::Handler& handler);
+
+        Event& operator=(const Event& other);
+
+        Event& operator=(Event&& other);
```
hultonha, line 55: update: Please can you group these with the other special members

Gems/EMotionFX/Code/EMotionFX/Source/MotionEvent.cpp
```diff
@@ -30,6 +30,23 @@ namespace EMotionFX
...
+    MotionEvent::MotionEvent(const MotionEvent& event):
+        Event(event),
+        m_startTime(event.m_startTime),
+        m_endTime(event.m_endTime),
+        m_isSyncEvent(event.m_isSyncEvent)
+    {
+
+    }
+    MotionEvent::MotionEvent(MotionEvent&& event):
+        Event(event),
```
hultonha, line 42: update: You need to add `AZStd::move` to this parameter (`event`) otherwise it won't be moved

Gems/EMotionFX/Code/EMotionFX/Source/MotionEvent.cpp
```diff
@@ -62,6 +79,24 @@ namespace EMotionFX
...
+        Event::operator=(other);
+        m_startTime = other.m_startTime;
+        m_endTime = other.m_endTime;
+        m_isSyncEvent = other.m_isSyncEvent;
+        return *this;
+    }
+
+    MotionEvent& MotionEvent::operator=(MotionEvent&& other)
+    {
+        Event::operator=(other);
```
hultonha, line 93: update: Need to `AZStd::move` `other` here too otherwise it will call the normal assignment operator - as this stuff is complex we really should add tests to cover it

Gems/EMotionFX/Code/EMotionFX/Source/MotionEvent.h
```diff
@@ -96,6 +98,10 @@ namespace EMotionFX
          */
         MotionEvent(float startTimeValue, float endTimeValue, EventDataSet&& datas);
 
+        MotionEvent& operator=(const MotionEvent& other);
+        
+        MotionEvent& operator=(MotionEvent&& other);
```
hultonha, line 103: update: Please group these with the other special member calls

Gems/EMotionFX/Code/EMotionFX/Source/Event.cpp
```diff
@@ -16,6 +16,17 @@ namespace EMotionFX
 {
     AZ_CLASS_ALLOCATOR_IMPL(Event, MotionEventAllocator, 0)
 
+    Event::Event(const Event& data)
+        : m_eventDatas(data.m_eventDatas)
+    {
+    }
```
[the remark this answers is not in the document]
hultonha: That would be really great, thank you!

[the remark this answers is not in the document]
pollend: you can't clone `m_eventDatasChangeEvent`

hultonha:
After taking a closer look at this issue @pollend I'm closing in favour of https://github.com/o3de/o3de/pull/13383 which keeps all changes at a slightly higher level and does not require changes to `Motion` or `MotionEvent`. I think this is a simpler approach and achieves the same ends. Let me know what you think. Thanks!

Issue #12687, opened 2022-10-19
Depracation Notice: AZ::RPI::RenderPipeline related notification APIs


VickyAtAZ:
## Deprecated APIs
The following APIs are deprecated
-  `void AZ::RPI::SceneNotification::OnRenderPipelineAdded(RenderPipelinePtr)`
-  `void AZ::RPI::SceneNotification::OnRenderPipelinePassesChanged(RenderPipeline*)`
-  `void AZ::RPI::SceneNotification::OnRenderPipelineRemoved(RenderPipeline*)`
-  `void AZ::RPI::FeatureProcessor::ApplyRenderPipelineChange(RenderPipeline* )`

                        
## Alternatives APIs

- Use `void AZ::RPI::SceneNotification::OnRenderPipelineChanged(RenderPipeline*, RenderPipelineChangeType)` to replace `OnRenderPipelineAdded`, `OnRenderPipelinePassesChanged `and `OnRenderPipelineRemoved`
- Use `void AZ::RPI::FeatureProcessor::AddRenderPasses(RenderPipeline* ) `to replace `ApplyRenderPipelineChange`

## Last Release
Last release version: `stabilization/2210`

## Additional Context
This is to simplify the render pipeline change notification to one function because the operation inside each are often the same.

PR #12700, opened 2022-10-19
GHI-12687 AZ::RPI::RenderPipeline related notification APIs

VickyAtAZ:
## What does this PR do?
This is to simplify the render pipeline change notification to one function because the operation inside each are often the same.
It also rename AZ::RPI::FeatureProcessor::ApplyRenderPipelineChange() to AZ::RPI::FeatureProcessor::AddRenderPasses() to clarify the intention of this function.
(https://github.com/o3de/o3de/issues/12687)

## How was this PR tested?
O3DE editor
ASV Screenshot tests.

### DIFF
--- a/Gems/Atom/Feature/Common/Code/Source/AuxGeom/AuxGeomFeatureProcessor.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/AuxGeom/AuxGeomFeatureProcessor.cpp
@@ -157,7 +157,8 @@ namespace AZ
             m_fixedShapeProcessor->SetUpdatePipelineStates();
         }
 
-        void AuxGeomFeatureProcessor::OnRenderPipelineChanged(RPI::RenderPipeline*, RPI::SceneNotification::RenderPipelineChangeType)
+        void AuxGeomFeatureProcessor::OnRenderPipelineChanged([[maybe_unused]] RPI::RenderPipeline* pipeline,
+            [[maybe_unused]] RPI::SceneNotification::RenderPipelineChangeType changeType)
         {
             OnSceneRenderPipelinesChanged();
         }
--- a/Gems/Atom/Feature/Common/Code/Source/AuxGeom/AuxGeomFeatureProcessor.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/AuxGeom/AuxGeomFeatureProcessor.cpp
@@ -157,15 +157,9 @@ namespace AZ
             m_fixedShapeProcessor->SetUpdatePipelineStates();
         }
 
-        void AuxGeomFeatureProcessor::OnRenderPipelineAdded(RPI::RenderPipelinePtr pipeline)
+        void AuxGeomFeatureProcessor::OnRenderPipelineChanged(RPI::RenderPipeline*, RPI::SceneNotification::RenderPipelineChangeType)
         {
             OnSceneRenderPipelinesChanged();
         }
-
-        void AuxGeomFeatureProcessor::OnRenderPipelineRemoved([[maybe_unused]] RPI::RenderPipeline* pipeline)
-        {
-            OnSceneRenderPipelinesChanged();
-        }
-
     } // namespace Render
 } // namespace AZ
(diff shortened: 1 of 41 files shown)


Gems/Atom/RPI/Code/Include/Atom/RPI.Public/FeatureProcessor.h
```diff
@@ -88,10 +88,17 @@ namespace AZ
...
+            
+            //! O3DE_DEPRECATION_NOTICE(GHI-12687)
+            //! @deprecated use AddRenderPasses(RenderPipeline*)
             //! Apply changes and add additional render passes to the render pipeline from the feature processors
             virtual void ApplyRenderPipelineChange(RenderPipeline* ) {}
 
+            //! Add additional render passes to the render pipeline before it's finalized
+            //! The render pipeline must have m_allowModification set to true (see Scene::TryApplyRenderPipelineChanges() function)
+            //! This function is called when the render pipeline is added or rebuilt
+            virtual void AddRenderPasses(RenderPipeline* ) {}
```
dmcdiarmid-ly, line 100: Same as above, [[maybe_unused]] might be preferred here.

dmcdiarmid-ly (approved):

Gems/Atom/Feature/Common/Code/Source/AuxGeom/AuxGeomFeatureProcessor.cpp
```diff
@@ -157,15 +157,9 @@ namespace AZ
             m_fixedShapeProcessor->SetUpdatePipelineStates();
         }
 
-        void AuxGeomFeatureProcessor::OnRenderPipelineAdded(RPI::RenderPipelinePtr pipeline)
+        void AuxGeomFeatureProcessor::OnRenderPipelineChanged(RPI::RenderPipeline*, RPI::SceneNotification::RenderPipelineChangeType)
```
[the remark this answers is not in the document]
VickyAtAZ: Yes, I changed that intentionally. The function OnSceneRenderPipelinesChanged() should also be called when pass change event is triggered since it may change the output layout which need pipeline states update.

Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
```diff
@@ -518,12 +518,8 @@ namespace AZ
             m_forceRebuildDrawPackets = true;
         }
 
-        void MeshFeatureProcessor::OnRenderPipelineAdded(RPI::RenderPipelinePtr pipeline)
-        {
-            m_forceRebuildDrawPackets = true;;
-        }
-
-        void MeshFeatureProcessor::OnRenderPipelineRemoved([[maybe_unused]] RPI::RenderPipeline* pipeline)
+        void MeshFeatureProcessor::OnRenderPipelineChanged([[maybe_unused]] RPI::RenderPipeline* pipeline,
```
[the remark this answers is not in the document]
VickyAtAZ: Same as above which I think the draw packets should be rebuilt when any pass changed.

gadams3 (approved):

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/SceneBus.h
```diff
@@ -39,14 +39,37 @@ namespace AZ
...
 
+            //! O3DE_DEPRECATION_NOTICE(GHI-12687)
+            //! @deprecated use OnRenderPipelineChanged(RenderPipeline*, RenderPipelineChangeType::Removed)
             //! Notifies when a render pipeline is removed from this scene.
             //! @param pipeline The render pipeline which was removed
             virtual void OnRenderPipelineRemoved([[maybe_unused]] RenderPipeline* pipeline) {};
+
+            //! Notifies when a render pipeline was added, removed or changed
+            //! @param pipeline The render pipeline which was added
+            virtual void OnRenderPipelineChanged(RenderPipeline* , RenderPipelineChangeType) {};
```
jiaweig-amzn, line 72: There is RenderPipelinePtr only used for adding behavior in the original API. But why enventually RenderPipeline* is used over RenderPipelinePtr?

jiaweig-amzn (approved):

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/SceneBus.h
VickyAtAZ: We don't want user to keep a smart pointer of the render pipeline especially when it's about to be removed.

PR #12709, opened 2022-10-19
{lyn98594} flag for the cached assets

jackalbe:
## What does this PR do?

This adds a flag to product asset database entries to track the ones that were cached using the Shared Cache system.

This flag will be used to indicate which product assets are cached in the UX of the Asset Processor.

## How was this PR tested?

Added a unit test + manual testing of the asset cache server to see the flag being set correctly.


cgalvan (approved):

Code/Tools/AssetProcessor/native/resourcecompiler/rcjob.cpp
```diff
@@ -542,6 +546,13 @@ namespace AssetProcessor
                                         builderParams.m_rcJob->GetJobEntry().m_pathRelativeToWatchFolder.toUtf8().data(), builderParams.m_rcJob->GetJobKey().toUtf8().data(),
                                         builderParams.m_rcJob->GetPlatformInfo().m_identifier.c_str(), builderParams.m_rcJob->GetOriginalFingerprint());
                                 }
+                                else
+                                {
+                                    AZStd::for_each(result.m_outputProducts.begin(), result.m_outputProducts.end(), [](auto& product)
+                                    {
+                                        product.m_outputFlags |= AssetBuilderSDK::ProductOutputFlags::CachedAsset;
+                                    });
```
amzn-mike, line 554: Is there a benefit to using for_each instead of a for range loop?

amzn-mike (approved):

Code/Tools/AssetProcessor/native/resourcecompiler/rcjob.cpp
jackalbe: not really, I can switch it to a for loop

amzn-mike (approved):

Issue #12550, opened 2022-10-12
Integrate Radeon GPU Analyzer for Register Statistics


santorac:
**Is your feature request related to a problem? Please describe.**
When analyzing GPU performance, one of the more important factors to consider is register (VGPR) pressure. We need a convenient way to determine the register usage for compiled shaders, to better understand the impact of specific shader options and shader variants. Render capture tools such as Nsight and PIX typically report this information, but it can be time consuming for developers to run the tools, run the engine, get into an appropriate runtime state, take a capture, and then browse to find information about VGPR usage. 

**Describe the solution you'd like**
[Radeon GPU Analyzer (RGA)](https://gpuopen.com/rga/) is a MIT licensed tool, capable of performing various offline analysis of HLSL code, including register usage. We should integrate this tool with the shader build pipeline to automatically report register usage (and other statistics) when compiling shader assets.

Since running this analysis could take considerable time on large data sets, and most users don't need this information, it should by off by default in most cases. When enabled, it would ideally be downloaded and/or installed automatically by the build system. Graphics engineers likely will only need to analyze a small number of shaders, so even when the tool is enabled, there should be away to enable only for specific shaders. This could be in the .shader file, in the .shadervariantlist file, or a .setreg file (it's up to the implementer to determine the best course).

Since the command line configuration for RGA is complex, and experience shows that usage of the tool is rather fragile, we should try to incorporate it into the Automated Review process, so we'll know right away if the integration breaks. This should be pretty simple, by enabling the statistics only one or two core shaders (like StandardPbr_ForwardPass.shader for example) in the AutomatedTesting project. That if the integration ever breaks, the shader fail to process, and this will fail the AR asset processing jobs on Jenkins. Since only one or two shaders would get this treatment, it won't have much impact on AR times.

**Describe alternatives you've considered**

**Additional context**
Some experimentation was done under this other task to determine how RGA can work together with O3DE's intermediate HLSL files in the cache. https://github.com/o3de/o3de/issues/11792

galibzon:
@santorac RGA is integrated in CMake, but is this task complete?

santorac:
I think we should still do this part before closing the ticket:
[...]

PR #12710, opened 2022-10-20
Add Radeon GPU analyzer tool and integrate to Asset Processor

etherealchain:

## What does this PR do?

Addressing #12550 

Add cmake file to fetch rga executsble from Github
Add parameter `EnableAnalysis` and `Asic` in .shadervariantlist
Users can easily enable/disable the analysis for any variant, using any gpu target rga provided
The output files will be moved to the Cache folder, the same place as the result  .azshadervariant file

## How was this PR tested?

Test on my pc, the executable is downloaded before building, analysis file will be output when .shadervariantlist, .shader or .azsl file is updated, if the switch is enabled.


Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantAssetBuilder.cpp
```diff
@@ -1140,10 +1141,137 @@ namespace AZ
...
+            if (shaderVariantInfo.m_enableRegisterAnalysis && strcmp(creationContext.m_shaderPlatformInterface.GetAPIName().GetCStr(), "vulkan") == 0)
+            {
+                AZ::IO::FixedMaxPath projectBuildPath = AZ::Utils::GetExecutableDirectory();
+                projectBuildPath = projectBuildPath.RemoveFilename(); // profile
+                projectBuildPath = projectBuildPath.RemoveFilename(); // bin
+
+                AZStd::string spirvPath;
+                AZStd::string spirvFile = AZStd::string::format(
+                    "%s_vulkan_%u.spirv.bin", creationContext.m_shaderStemNamePrefix.c_str(), shaderVariantInfo.m_stableId);
+                AZ::StringFunc::Path::Join(creationContext.m_tempDirPath.c_str(), spirvFile.c_str(), spirvPath, true, true);
```
lumberyard-employee-dm, line 1153: This can use the operator/= or Append function of AZ::IO::Path
```suggestion
                AZ::IO::FixedMaxPath spirvPath(AZStd::string_view(creationContext.m_tempDirPath));
                spirvPath /= AZ::IO::FixexMaxPathString::format(
                    "%s_vulkan_%u.spirv.bin", creationContext.m_shaderStemNamePrefix.c_str(), shaderVariantInfo.m_stableId);
```

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantAssetBuilder.cpp
```diff
@@ -1140,10 +1141,137 @@ namespace AZ
...
+
+                AZStd::string rgaCommand = AZStd::string::format(
+                    "-s vk-spv-offline --isa ./disassem_%u.txt --livereg ./livereg_%u.txt --asic %s",
+                    shaderVariantInfo.m_stableId,
+                    shaderVariantInfo.m_stableId,
+                    shaderVariantInfo.m_asic.c_str());
+
+                AZStd::string command = AZStd::string::format("%s%s %s %s", projectBuildPath.c_str(), RgaExePath, rgaCommand.c_str(), spirvPath.c_str());
+                AZ_TracePrintf(ShaderVariantAssetBuilderName, "Rga command %s", command.c_str());
```
lumberyard-employee-dm, line 1162: FYI `AZ_TracePrintf doesn't automatically add a new line to the end of the message so the next log is slammed directly next to the previous one.
```suggestion
                AZ_TracePrintf(ShaderVariantAssetBuilderName, "Rga command %s\n", command.c_str());
```

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantAssetBuilder.cpp
```diff
@@ -1140,10 +1141,137 @@ namespace AZ
...
+
             Data::Asset<RPI::ShaderVariantAsset> shaderVariantAsset;
             variantCreator.End(shaderVariantAsset);
             return AZ::Success(AZStd::move(shaderVariantAsset));
         }
 
+        bool ShaderVariantAssetBuilder::LaunchRadeonGPUAnalyzer(AZStd::string& command, const AZStd::string& workingDirectory, AZStd::string& failMessage)
+        {
+            AzFramework::ProcessLauncher::ProcessLaunchInfo processLaunchInfo;
+            processLaunchInfo.m_commandlineParameters.emplace<AZStd::string>(command);
```
lumberyard-employee-dm, line 1184: I suggest using the AZStd::vector<AZStd::string> version for the command line parameters as that can properly deal with paths that have spaces in them.
For example if the user had a project at `C:\Users\My User Name\My Documents\TestProject`, then this command will have issues.

So to avoid that, this function should accept an `AZStd::vector<AZStd::string> command` and then move it into the processLaunchInfo.

```suggestion
        bool ShaderVariantAssetBuilder::LaunchRadeonGPUAnalyzer(AZStd::vector<AZStd::string> command, const AZStd::string& workingDirectory, AZStd::string& failMessage)
        {
            AzFramework::ProcessLauncher::ProcessLaunchInfo processLaunchInfo;
            processLaunchInfo.m_commandlineParameters.emplace<AZStd::vector<AZStd::string>>(AZStd::move(command));
```

cmake/3rdParty/FetchRGA.cmake
```diff
@@ -0,0 +1,19 @@
...
+if (NOT FETCHED_RGA)
+    include(FetchContent)
+    FetchContent_Declare(
+        RGA
+        URL https://github.com/GPUOpen-Tools/radeon_gpu_analyzer/releases/download/2.6.2/rga-windows-x64-2.6.2.zip
+        DOWNLOAD_EXTRACT_TIMESTAMP true
+    )
+
+    FetchContent_MakeAvailable(RGA)
+    message("rga.exe will be store in ${rga_SOURCE_DIR}")
```
santorac, line 16:
```suggestion
    message("rga.exe will be stored in ${rga_SOURCE_DIR}")
```

Gems/Atom/RPI/Code/Include/Atom/RPI.Edit/Shader/ShaderVariantListSourceData.h
```diff
@@ -40,6 +40,16 @@ namespace AZ
...
                 ShaderOptionValuesSourceData m_options;
+
+                
+                // Output register analysis data
+                bool m_enableRegisterAnalysis = false;
+
+                // The GPU target to use on register analysis
+                // The value depend on the version of RGA we use
+                // Current RGA is 2.6.2
+                // Support value: gfx900 gfx902 gfx906 gfx90c gfx1010 gfx1011 gfx1012 gfx1030 gfx1031 gfx1032 gfx1034 gfx1035
```
santorac, line 51:
```suggestion
                // The GPU target to use on register analysis
                // The value depends on the version of RGA we use
                // Current RGA is 2.6.2
                // Supported values: gfx900 gfx902 gfx906 gfx90c gfx1010 gfx1011 gfx1012 gfx1030 gfx1031 gfx1032 gfx1034 gfx1035 
```

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantAssetBuilder.cpp
```diff
@@ -1140,10 +1141,137 @@ namespace AZ
                 }
             }
 
+            if (shaderVariantInfo.m_enableRegisterAnalysis && strcmp(creationContext.m_shaderPlatformInterface.GetAPIName().GetCStr(), "vulkan") == 0)
```
santorac, line 1144: I think you can compare like this:
```suggestion
            if (shaderVariantInfo.m_enableRegisterAnalysis && creationContext.m_shaderPlatformInterface.GetAPIName() == "vulkan")
```
or this:
```suggestion
            if (shaderVariantInfo.m_enableRegisterAnalysis && creationContext.m_shaderPlatformInterface.GetAPIName().GetStringView() == "vulkan")
```

Gems/Atom/RPI/Code/Include/Atom/RPI.Edit/Shader/ShaderVariantListSourceData.h
```diff
@@ -40,6 +40,16 @@ namespace AZ
                 AZ::u32 m_stableId = 0;
 
                 ShaderOptionValuesSourceData m_options;
+
+                
+                // Output register analysis data
+                bool m_enableRegisterAnalysis = false;
```
santorac, line 46: Maybe add a comment about how this only works on vulkan.

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantAssetBuilder.cpp
santorac: Maybe report a warning when this is not vulkan, so users aren't confused why the data doesn't show up for other platforms.

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantAssetBuilder.cpp
```diff
@@ -1140,10 +1141,137 @@ namespace AZ
...
+
+                AZStd::string failMessage;
+                if (!LaunchRadeonGPUAnalyzer(command, creationContext.m_tempDirPath, failMessage))
+                {
+                    return AZ::Failure(failMessage);
+                }
+
+                // add rga output to the by product list
+                outputByproducts->m_intermediatePaths.insert(AZStd::string::format("./%s_disassem_%u_frag.txt",shaderVariantInfo.m_asic.c_str(), shaderVariantInfo.m_stableId));
+                outputByproducts->m_intermediatePaths.insert(AZStd::string::format("./%s_livereg_%u_frag.txt", shaderVariantInfo.m_asic.c_str(),shaderVariantInfo.m_stableId));
```
santorac, line 1172:
```suggestion
                outputByproducts->m_intermediatePaths.insert(AZStd::string::format("./%s_disassem_%u_frag.txt", shaderVariantInfo.m_asic.c_str(), shaderVariantInfo.m_stableId));
                outputByproducts->m_intermediatePaths.insert(AZStd::string::format("./%s_livereg_%u_frag.txt", shaderVariantInfo.m_asic.c_str(), shaderVariantInfo.m_stableId));
```

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantAssetBuilder.cpp
etherealchain: Thanks for the info!

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantAssetBuilder.cpp
etherealchain: I think it is not necessary, I'm not sure about Linux, but on Windows Asset Processor always produces 3 hlsl (dx12, null, vulkan), no matter which RHI is used. So data will always be created when it is enabled on Windows.

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantAssetBuilder.cpp
```diff
@@ -62,6 +62,7 @@ namespace AZ
     namespace ShaderBuilder
     {
         static constexpr char ShaderVariantAssetBuilderName[] = "ShaderVariantAssetBuilder";
+        static constexpr char RgaExePath[] = "/_deps/rga-src/rga.exe";
```
siliconvoodoo, line 67: always windows for now I suppose?

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantAssetBuilder.cpp
```diff
@@ -1140,10 +1141,138 @@ namespace AZ
...
+            }
+
+            AZ_Assert(!watcherPtr->IsProcessRunning(), "Rga execution failed to terminate");
+
+            // Pump one last time to make sure the streams have been flushed
+            pumpOuputStreams();
+
+            if (timedOut)
+            {
+                failMessage = AZStd::string("Rga takes too long to finish");
```
siliconvoodoo, line 1254: Maybe in the past tense: "Rga execution timed out"

siliconvoodoo (changes requested):
tiny nits

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantAssetBuilder.cpp
etherealchain: I added Linux support, but I have no machine to test

cmake/3rdParty/FetchRGA.cmake
```diff
@@ -0,0 +1,26 @@
...
+# file of the target platform.
+
+message(STATUS "started FetchRGA")
+
+if (NOT FETCHED_RGA)
+    include(FetchContent)
+    if (WIN32)
+        FetchContent_Declare(
+            RGA
+            URL https://github.com/GPUOpen-Tools/radeon_gpu_analyzer/releases/download/2.6.2/rga-windows-x64-2.6.2.zip
```
AMZN-Olex, line 12: Does this work with an installer? By that I mean if you produce an installer for O3DE by building INSTALL target, can you use this feature still? I am guessing this is a binary payload, does it exist in an installer SDK layout?

etherealchain: I try to build CMakePredefinedTargets/INSTALL
But FetchGRA.cmake is not invoked

I include FetchGRA.cmake in 3rdParty.cmake, which is included in o3de/CMakeList.txt
And it is only included `if(NOT INSTALLED_ENGINE)`, following the logic of other 3rdParty cmake files
Should it be working with the installer? What is a better way to do this?

etherealchain: FetchRGA.cmake will only be invoked when create/recreate the project.
It is included in 3rdParty.cmake, which is included in o3de/CMakeLists.txt.
I added in `if(NOT INSTALLED_ENGINE)`, to follow the manner of other 3rdParty cmake files

I try to build CMakePredefinedTargets/INSTALL
The downloaded binary payload will not be copied to the install folder, if that's what you ask

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantAssetBuilder.cpp
santorac: There is a setting where you could turn off RHI backends. Some people do this to save asset compilation time, they'll turn off everything except dx12 for example.

etherealchain: Warning added!

cmake/3rdParty/FetchRGA.cmake
AMZN-Olex: If a binary wont be going to the installer folder then it means only users who are building from source code directly can use the tool. For example, if this goes in then the next O3DE installer wont have this functioning out of the box, right?

I believe there are are ways to add these binaries to install payloads as well. I would ask in sig-core. Depending on the sig that owns this PR, perhaps create a separate GHI for that work. 

Thanks!

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantAssetBuilder.cpp
```diff
@@ -1140,10 +1142,162 @@ namespace AZ
...
+                    AZStd::string command = AZStd::string::format(
+                        "%s%s %s %s", projectBuildPath.c_str(), RgaPath.c_str(), rgaCommand.c_str(), spirvPath.c_str());
+                    AZ_TracePrintf(ShaderVariantAssetBuilderName, "Rga command %s\n", command.c_str());
+
+                    AZStd::vector<AZStd::string> fullCommand;
+                    fullCommand.push_back(command);
+                    AZStd::string failMessage;
+                    if (!LaunchRadeonGPUAnalyzer(fullCommand, creationContext.m_tempDirPath, failMessage))
+                    {
+                        return AZ::Failure(failMessage);
```
santorac, line 1182: I think we should report a warning here, but don't return Failure. If I understand correctly, the .exe will not be downloaded when the o3de installer is used. In that case, we don't want to fail the job because the main product here is the shader variant asset, which is not impacted by the failed rga process, that's only debug information.

cmake/3rdParty/FetchRGA.cmake
AMZN-Olex: I think you can add this instruction to fix the installer. (Or something like that.)
```
ly_install_directory(DIRECTORIES .)
```

santorac: Since this is mainly for graphics engineers to optimize shaders, it's probably okay for now to have it only available when building from source code. But I agree we should create a GHI to follow up by making this available to installed builds. Also, see my comment below about not failing the job when rga is missing.

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantAssetBuilder.cpp
antonmic: Agree, no need to fail here, just need to warn the user

Gems/Atom/RPI/Code/Source/RPI.Edit/Shader/ShaderVariantListSourceData.cpp
```diff
@@ -21,9 +21,11 @@ namespace AZ
             if (auto serializeContext = azrtti_cast<SerializeContext*>(context))
             {
                 serializeContext->Class<VariantInfo>()
-                    ->Version(1)
+                    ->Version(2)
```
antonmic, line 24: Can you add a // comment on this line so it triggers a merge conflict if anyone else simultaneously bumps the version to the same number? i.e.
->Version(2)    // Added Radeon GPU Analyzer

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantAssetBuilder.cpp
```diff
@@ -1177,16 +1177,18 @@ namespace AZ
...
-                        return AZ::Failure(failMessage);
+                        // add rga output to the by product list
+                        outputByproducts->m_intermediatePaths.insert(AZStd::string::format(
+                            "./%s_disassem_%u_frag.txt", shaderVariantInfo.m_asic.c_str(), shaderVariantInfo.m_stableId));
+                        outputByproducts->m_intermediatePaths.insert(AZStd::string::format(
+                            "./%s_livereg_%u_frag.txt", shaderVariantInfo.m_asic.c_str(), shaderVariantInfo.m_stableId));
+                    }
+                    else
+                    {
+                        AZ_Warning(ShaderVariantAssetBuilderName, false, failMessage.c_str());
```
santorac, line 1190: When reporting strings, we should also pass through "%s" just in case the string has "%" symbols that could become a stability or security issue.
```suggestion
                        AZ_Warning(ShaderVariantAssetBuilderName, false, "%s", failMessage.c_str());
```

santorac (approved):

santorac:
Automated Review started: https://jenkins.build.o3de.org/job/O3DE/view/change-requests/job/PR-12710/5/

antonmic (approved):

PR #12717, opened 2022-10-20
Material Pipeline System Phase 1

santorac:
## What does this PR do?

Added the first phase of implementation of the material pipieline system. 
See https://github.com/o3de/sig-graphics-audio/blob/main/rfcs/MaterialPipelineAbstraction.md for the full design.

At this stage you can author a material type with custom vertex manipulation, custom surface parameterization, and make it work with any render pipeline. It supports the MainPipeline and LowEndPipeline, as well as a proof-of-concept deferred pipeline that I have been working on here: https://github.com/aws-lumberyard-dev/o3de-atom-sampleviewer/tree/Atom/santorac/DeferredPipelinePOC

### Details
- Added new .materialpipeline file
- The .materialtype file can now follow an "abstract" format where a shader list is not provided, and instead the material type provides custom material-specific shader code that will be combined with code from the material pipeline.
- Updated MaterialTypeBuilder to process these abstract .materialtype files by stitching them together with each material pipeline. It uses the new intermediate asset feature to generate intermediate .shader files and an intermediate .materialtype file that uses the original "direct" format.
- Updated the asset database version since the MaterialTypeBuilder has new job keys.
- There are shader templates for each pass.
- Multiple material pipelines can share common shader templates, or provide their own unique ones.
- There are a couple example material types to demonstrate the functionality and test the system.
- The original "direct" .materialtype format is still supported.
- The list of available material pipelines is configured through a .setreg file. Be default, only MainPipeline is enabled. In the AutomatedTesting project we override the list to also include LowEndPipeline for regression testing purposes, so any asset failures will block Jenkins.
- Improved comments AssetBuilderSDK.h

### Reviewer Guide

The most important thing to understand is how the shaders get stitched together. Consider the case of running a basic material type on the MainPipeline. The final generated AZSL file looks like this: 

**materialpipelinetest_basic_mainpipeline_forwardpass.azsl**
```
// This code was generated by MaterialTypeBuilder. Do not modify.
#include <D:\o3de\Gems\Atom\Feature\Common\Assets\Materials\Pipelines\MainPipeline\ForwardPass.azsli>
#include <D:\o3de\Gems\Atom\TestData\TestData\Materials\Types\MaterialPipelineTest_Basic.azsli>
 
#if !MaterialFunction_AdjustVertexData_DEFINED                                                    
    void MaterialFunction_AdjustVertexData(in float3 positionLS, inout VertexData vertexData) {}  
    #define MaterialFunction_AdjustVertexData_DEFINED 1                                           
#endif                                                                                            
                                                                                                  
#if !MaterialFunction_AdjustSurface_DEFINED && MATERIALPIPELINE_SHADER_HAS_PIXEL_STAGE            
    void MaterialFunction_AdjustSurface(inout Surface outSurface) {}                              
    #define MaterialFunction_AdjustSurface_DEFINED 1                                              
#endif                                                                                            
```

`MaterialPipelineTest_Basic.materialtype` provides some material shader code in [MaterialPipelineTest_Basic.azsli](https://github.com/o3de/o3de/pull/12717/files#diff-c1dc1e170303a091d675875d5b6ea3932a6567f32e378d2a50366d1ae9cfd5df). `MainPipeline.materialpipeline` provides an AZSLI file specific to the forward pass in [MainPipeline\ForwardPass.azsli](https://github.com/o3de/o3de/pull/12717/files#diff-121c51b0570e9658655c90959d44ff4ef478f75827b568a4476d29792917c693). The MaterialTypeBuilder stitches these two files together and also appends a bit of utility code. 

1. With this in mind, go review [MaterialPipelineCallbacks.azsli](https://github.com/o3de/o3de/pull/12717/files#diff-1b4867fd2d06b48d8fa9a9acd5ee4c34847f86075b8a72242cd35b4d19e8081f), [MainPipeline\ForwardPass.azsli](https://github.com/o3de/o3de/pull/12717/files#diff-121c51b0570e9658655c90959d44ff4ef478f75827b568a4476d29792917c693), and [MaterialPipelineTest_Basic.azsli](https://github.com/o3de/o3de/pull/12717/files#diff-c1dc1e170303a091d675875d5b6ea3932a6567f32e378d2a50366d1ae9cfd5df).
1. Next you'll likely be wondering how vertex modifications work, for supporting things like vertex animation, so go look at [MaterialPIpelineTest_Animated.azsli](https://github.com/o3de/o3de/pull/12717/files#diff-416175da592a239d73344edf385f437a5dc23e6d38e8c763f4ce8341dfda5e67).
1. [MaterialTypeSourceData.h](https://github.com/o3de/o3de/pull/12717/files#diff-1ccaa2448dc890218058844104716370ce934f863e4302407575fda7cd96edac) has a good explanation of the "abstract" and "direct" forms that .materialtype can have.
1. [MaterialTypeBuilder.h](https://github.com/o3de/o3de/pull/12717/files#diff-2a7193e991e4dbc8cb661b94207f5374112651c7f752717a7eba1925a615cc53) shows how processing is broken into two phases.
1. [MaterialTypeBuilder.cpp](https://github.com/o3de/o3de/pull/12717/files#diff-ccbd2a4b5acec4cea74f13936d702da932985a9cbebbd4bb2f8b09b247e68126) is where all the magic happens.
1. Go through the other files at your leisure.

## How was this PR tested?
- Created and opened materials in the Material Editor. 
- Rotated a model to make sure transforms were applied.
- Tested switching between pipelines in AtomSampleViewer's Mesh sample, including my prototype deferred pipeline.
- Run AtomSampleViewer full test suite. No new failures.
- Lots of testing of asset dependencies and reprocessing...
    - Changed the material_pipelines.setreg list, restarted the AP, and saw the changes get applied in the AP job list.
    - Removed and added a shader template in a .materialpipeline and saw the changes get applied in the AP job list.
    - Changed the .materialtype and saw the appropriate assets get reprocessed (intermediate .materialtype and .material that uses it).
    - Changed the material type's AZSLI file and saw the appropriate assets get reprocessed (intermediate .materialtype, .material that uses it, and each of the shaders).
    - Changed a pipeline-specific shader template file and saw the necessary assets get reprocessed (source material types, the corresponding intermediate .shader files). It also reprocessed the intermediate material types because those depend on the generated .shader files which changed, and the relevant .material files; this is expected.
    - Changed a pipeline-specific AZSLI template file and saw the intermediate .shader files get reprocessed. It also reprocessed the intermediate material types because those depend on the generated .shader files which changed, and the relevant .material files; this is expected. The source .materialfile was NOT reprocessed, which is good, it shouldn't need to be reprocessed, only intermediate assets are impacted here.
    - Changed a common shader template file and saw the necessary assets get reprocessed (source material types, the corresponding intermediate .shader files). It also reprocessed the intermediate material types similar to above.
    - Changed a common AZSLI template file and saw the intermediate .shader files get reprocessed. It also reprocessed the intermediate material types similar to above.
    - Changed ForwardPassVertexAndPixel.azsli which is only used in the main pipeline and saw the expected files get reprocessed. Only the main pipeline shaders were reprocessed, not the low end pipeline shaders. It also reprocessed the intermediate material types similar to above.


Gems/Atom/Feature/Common/Assets/Materials/Pipelines/LowEndPipeline/ForwardPass.azsli
```diff
@@ -0,0 +1,123 @@
...
+
+#include <viewsrg.srgi>
+#include <Atom/RPI/ShaderResourceGroups/DefaultDrawSrg.azsli>
+#include <Atom/Features/PBR/DefaultObjectSrg.azsli>
+#include <Atom/Features/Pipeline/Forward/ForwardPassSrg.azsli>
+#include <Atom/Features/PBR/AlphaUtils.azsli>
+#include <Atom/Features/SrgSemantics.azsli>
+#include <Atom/Features/ColorManagement/TransformColor.azsli>
+
+// TODO: It a future iteration, the material pipeline will support multiple lighting models instead of just one.
```
HogJonny-AMZN, line 29: nit: typo, 'In a ...'

HogJonny-AMZN (approved):

Gems/Atom/RPI/Code/Source/RPI.Edit/Material/MaterialPipelineSourceData.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+                    editContext->Class<MaterialPipelineSourceData>("MaterialPipelineSourceData", "")
+                        ->ClassElement(AZ::Edit::ClassElements::EditorData, "")
+                        ->Attribute(AZ::Edit::Attributes::AutoExpand, true)
+                        ->DataElement(AZ::Edit::UIHandlers::Default, &MaterialPipelineSourceData::m_shaderTemplates, "Shader Templates", "List of templates used to generate material-specific shaders.")
+                        ;
+                }
+            }
+
+            if (auto behaviorContext = azrtti_cast<AZ::BehaviorContext*>(context))
```
gadams3, line 50: Thank you for making sure all of the contexts are filled out

Code/Framework/AzToolsFramework/AzToolsFramework/AssetDatabase/AssetDatabaseConnection.h
```diff
@@ -72,6 +72,7 @@ namespace AzToolsFramework
             AddedStatsTable,
             ChangedSourceDependencySourceColumn,
             SplitMaterialBuilderAndMaterialAssetBuilder,
+            NewMaterialTypeBuildPipeline,
```
jeremyong-az, line 75: Placeholder name I presume?

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/Common/DepthPass.shader.template
```diff
@@ -0,0 +1,20 @@
+{
+    "Source" : "INSERT_AZSL_HERE",
+
+    "DepthStencilState" : { 
+        "Depth" : { "Enable" : true, "CompareFunc" : "GreaterEqual" }
```
jeremyong-az, line 5: Assuming reverse-z is probably a safe assumption but we can keep this in the back of our minds (in some cases, shader implementation will be coupled to PSO state like this)

jeremyong-az: Also do we want >= and not > here?

Gems/Atom/RPI/Code/Source/RPI.Builders/Material/MaterialTypeBuilder.cpp
```diff
@@ -199,6 +268,269 @@ namespace AZ
...
+                    "#if !MaterialFunction_AdjustVertexData_DEFINED                                                    \n"
+                    "    void MaterialFunction_AdjustVertexData(in float3 positionLS, inout VertexData vertexData) {}  \n"
+                    "    #define MaterialFunction_AdjustVertexData_DEFINED 1                                           \n"
+                    "#endif                                                                                            \n"
+                    "                                                                                                  \n"
+                    "#if !MaterialFunction_AdjustSurface_DEFINED && MATERIALPIPELINE_SHADER_HAS_PIXEL_STAGE            \n"
+                    "    void MaterialFunction_AdjustSurface(inout Surface outSurface) {}                              \n"
+                    "    #define MaterialFunction_AdjustSurface_DEFINED 1                                              \n"
+                    "#endif                                                                                            \n"
+                    ;
```
gadams3, line 454: Other than the include file, will the generated code ever changed dynamically? Could this be moved to a template file and copied over instead? Then it can be updated if necessary without recompiling the builder.

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/LowEndPipeline/ForwardPass.azsli
```diff
@@ -0,0 +1,123 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+// These setting used to be in ShaderQualityOptions.azsli as a basic/limited solution for multiple pipelines.
+// TODO, instead of UNIFIED_FORWARD_OUTPUT, fork ForwardPassOutput.azsli
```
jeremyong-az, line 10: Suggest using a searchable string like `TODO(MaterialPipeline)` to make things easy to find later

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/LowEndPipeline/ForwardPass.shader.template
```diff
@@ -0,0 +1,43 @@
...
+    "Source" : "INSERT_AZSL_HERE",
+
+    "DepthStencilState" :
+    {
+        "Depth" :
+        {
+            "Enable" : true,
+            "CompareFunc" : "GreaterEqual"
+        },
+        "Stencil" :
```
jeremyong-az, line 11: We will probably need some way to control this eventually, fine as-is

Gems/Atom/RPI/Code/Source/RPI.Builders/Material/MaterialTypeBuilder.cpp
```diff
@@ -199,6 +268,269 @@ namespace AZ
...
+
+                AZ::IO::Path shaderName = shaderTemplate.m_shader;
+                shaderName = shaderName.Filename(); // Removes the folder path
+                shaderName = shaderName.ReplaceExtension(""); // This will remove the ".template" extension
+                shaderName = shaderName.ReplaceExtension(""); // This will remove the ".shader" extension
+
```
gadams3, line 460: As we discussed before, I got rid of the template file extension for material canvas templates and it might be worth doing the same here. The purpose of the template file extension was to make the template files easily identifiable while also hiding them from the AP. But hiding them from the AP made it easy for other changes, like include files being moved around, to break the templates. Now, templates use the standard file extension, which allows them to be automatically recognized, processed, and validated by the AP. The template is just defined by references in the output node configuration but files for the current template are also contained in the same folder. Most of the template files are treated as text files that will be scanned for markers that can be filled in or substituted. The material type and shader files can be deserialized directly to make other modifications to the data structures.

Gems/Atom/RPI/Assets/ShaderLib/Atom/RPI/MaterialPipelineCallbacks.azsli
```diff
@@ -0,0 +1,93 @@
...
+// The material type may optionally implement any of these functions for custom behavior.
+// Any functions that the material type does not implement will be automatically stubbed out with default behavior.
+
+
+
+// This struct includes fields for all vertex data that can possibly be supported by any material pipeline.
+// The actual vertex stage inputs and outputs will vary between shaders. If any of these fields isn't needed
+// by a given shader, then the material pipeline will initialize them to 0 and will not use them.
+// This way, the material type implementation of the callback function can read and set any of these values
+// without caring whether they are actually used. If the value is not used then any extra code will
```
jeremyong-az, line 21: I'm not sure if DCE will necessarily apply here, since if you write 0 to a field which ultimately isn't consumed by the subsequent stage, won't the compiler still need to emit a `mov v0, 0` instruction at the very least and keep the ISBE slot for that data?

Gems/Atom/RPI/Code/Source/RPI.Builders/Material/MaterialTypeBuilder.cpp
```diff
@@ -199,6 +268,269 @@ namespace AZ
...
+                // Intermediate shader file
+
+                AZStd::string azslFileReference = AZ::IO::Path{outputAzslFilePath.Filename()}.c_str();
+                AZStd::to_lower(azslFileReference.begin(), azslFileReference.end());
+                AzFramework::StringFunc::Replace(shaderFile.GetValue(), "INSERT_AZSL_HERE", azslFileReference.c_str());
+
+                AZ::IO::Path outputShaderFilePath = request.m_tempDirPath;
+                outputShaderFilePath /= AZStd::string::format("%s_%s_%s.shader", materialTypeName.c_str(), materialPipelineName.c_str(), shaderName.c_str());
+
+                if (AZ::Utils::WriteFile(shaderFile.GetValue(), outputShaderFilePath.c_str()).IsSuccess())
```
gadams3, line 488: Instead of reading in the text for the shader file, I recommend loading the shader source data structure from the file, adding the file reference, and saving the output shader source data file. Then you don't need INSERT_AZSL_HERE.

Gems/Atom/RPI/Code/Source/RPI.Builders/Material/MaterialTypeBuilder.cpp
```diff
@@ -199,6 +268,269 @@ namespace AZ
...
+                        resolveTemplateFilePathReference(materialPipelineFilePath, normalizedShaderTemplate.m_shader);
+                        resolveTemplateFilePathReference(materialPipelineFilePath, normalizedShaderTemplate.m_azsli);
+
+                        shaderTemplateReferences[normalizedShaderTemplate].push_back(materialPipelineName.c_str());
+                    }
+                }
+
+                if (missingSomeFiles)
+                {
+                    return;
```
jeremyong-az, line 371: Should we emit an error here?

Gems/Atom/RPI/Code/Source/RPI.Builders/Material/MaterialTypeBuilder.cpp
```diff
@@ -199,6 +268,269 @@ namespace AZ
...
+                    materialPipelineName = *materialPipelineList.begin();
+                }
+                else if(materialPipelineList.size() > 1)
+                {
+                    // Multiple material pipelines reference the same shader, so it should have a generic common name.
+                    materialPipelineName = PipelineNameForCommonShaders;
+                }
+                else
+                {
+                    AZ_Assert(false, "There should be at least one material pipeline referencing the shader");
```
jeremyong-az, line 410: Question, are asset processor builders expected to handle errors differently through some error reporting API as opposed to asserting?

Gems/Atom/RPI/Code/Source/RPI.Builders/Material/MaterialTypeBuilder.cpp
```diff
@@ -199,6 +268,269 @@ namespace AZ
...
+                materialType.m_shaderCollection.push_back({});
+                materialType.m_shaderCollection.back().m_shaderFilePath = AZ::IO::Path{outputShaderFilePath.Filename()}.c_str();
+            }
+
+            AZ::IO::Path outputMaterialTypeFilePath = request.m_tempDirPath;
+            // The "_generated" postfix is necessary because AP does not allow intermediate file to have the same relative path as a source file.
+            outputMaterialTypeFilePath /= AZStd::string::format("%s_generated.materialtype", materialTypeName.c_str());
+
+            AZ_Assert(!materialType.IsAbstractFormat(),
+                "The output material type must not use the abstract format, this will likely causing the '%s' job to run in an infinite loop.", PipelineStageJobKey);
```
jeremyong-az, line 513: s/causing/cause

jeremyong-az (commented):
Shaping up really well I think!

Gems/Atom/Asset/Shader/Code/Tests/ShaderBuilderUtilityTests.cpp
```diff
@@ -34,11 +34,13 @@ namespace UnitTest
...
             "# include <a\\dire-ctory\\valid-file6.azsli>\n"
             "#includ \"a\\dire-ctory\\invalid-file7.azsli\"\n"
+            " #include <..\\Relative\\Path\\To\\File.azsi>\n"
+            " #include <C:\\Absolute\\Path\\To\\File.azsi>\n"
         );
 
         AZ::ShaderBuilder::ShaderBuilderUtility::IncludedFilesParser includedFilesParser;
         auto fileList = includedFilesParser.ParseStringAndGetIncludedFiles(haystack);
-        EXPECT_EQ(fileList.size(), 5);
+        //EXPECT_EQ(fileList.size(), 5);
```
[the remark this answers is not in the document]
santorac: Oops, this is was an oversight, thanks.

Gems/Atom/RPI/Code/Source/RPI.Builders/Material/MaterialTypeBuilder.cpp
santorac: Yes I have been thinking about moving this out. I didn't bother for now since I'll be iterating on this code a lot in the next couple weeks, I'll have lots of other opportunities to move this out. If I have to touch this snippet at all, I'll probably move it to a file at that time.

Gems/Atom/RPI/Code/Source/RPI.Builders/Material/MaterialTypeBuilder.cpp
santorac: >But hiding them from the AP made it easy for other changes, like include files being moved around, to break the templates.
This isn't a concern for the material pipeline system because there is an asset builder that uses these files, and will readily fail if any breaking changes are made.

Gems/Atom/RPI/Code/Source/RPI.Builders/Material/MaterialTypeBuilder.cpp
santorac: I kind of like having "INSERT_AZSL_HERE" since the file is a template and not a complete valid shader asset, this makes it clear to readers what aspect of the file is a template and will be filled in later.

Code/Framework/AzToolsFramework/AzToolsFramework/AssetDatabase/AssetDatabaseConnection.h
santorac: This is bumping the asset database version number to force all assets to reprocess. In many cases, per-asset-builder version numbers can be incremented to reprocess assets from a specific builder. However in this case I changed the builder's job key which invalidates a bunch of data in the asset database and requires a full rebuild. The enum name is just a rough description of why we needed to rebuild the database.

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/Common/DepthPass.shader.template
santorac: This is just a duplicate of the main depth shader (Gems\Atom\Feature\Common\Assets\Shaders\Depth\DepthPass.shader), so I won't change it here. That said, I don't remember the history behind deciding to use reverse-z or why we have >=

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/LowEndPipeline/ForwardPass.azsli
santorac: Sure, I can do that.

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/LowEndPipeline/ForwardPass.shader.template
santorac: Yeah, stencil is one of those things I've had in the back of my mind that needs to be decoupled. I think material pipeline will give us some ability to configure this on a material-type basis at least, through the material pipeline properties and lua script.

Gems/Atom/RPI/Assets/ShaderLib/Atom/RPI/MaterialPipelineCallbacks.azsli
santorac: Hm, you might be right, I'm not sure. I'll remove that part of the comment.

Gems/Atom/RPI/Code/Source/RPI.Builders/Material/MaterialTypeBuilder.cpp
santorac: AZ_Error is for problems in the data that a user can fix. AZ_Assert is for problems in the code that a programmer must fix. AZ_Assert is appropriate in this case, because it shouldn't be possible to get here since the loop that filled shaderTemplateReferences should always put at least one pipeline in the list. All add a code comment to clarify.

Gems/Atom/RPI/Code/Source/RPI.Builders/Material/MaterialTypeBuilder.cpp
gadams3: I see. The same idea could be conveyed by moving the generated code block with the placeholder material functions to its own file and setting that as the default path instead using INSERT_AZSL_HERE. It might even simplify the builder. In addition to that, since the generated file will now be a “real file”, the user could just include it at the end of their custom AZSL to provide the defaults for whatever functions they didn't implement.

Also, when I initially read this, INSERT_AZSL_HERE seemed like it was expecting code to be inserted into .shader file. Sticking with this token, I recommend making it something more explicit to convey that it's a path.

santorac: The thing the actual entry point .azsl file does not exist, that's generated by the asset builder. Both the template .azsli file and the material type's .azsli file are just snippets. The builder generates the .azsl that stitches them together, and that's what the .shader needs to reference.

I can change the macro name in my next PR, so I don't have to re-run Jenkins AR.

gadams3 (approved):

PR #12719, opened 2022-10-20
Added quit option for performance collection.

galibzon:
## What does this PR do?
Added new CVAR named `r_metricsQuitUponCompletion`, defaults to false. When true the application will quit when all frame batches have been measured for performance.

Added OS name and RHI name to the Log Category name: `Graphics-<OS>-<RHI>`. It will make it easy to keep vulkan and dx12 results side by side and easy to identify.

## How was this PR tested?

This is a cosmetic change but I made sure the GameLauncher  quits when all frames have been measured for performance.
Also made sure the output json file has the new name format: "Performance_Graphics-OS-RHI_DATE.json".

Signed-off-by: galibzon <66021303+galibzon@users.noreply.github.com>

### DIFF
--- a/Gems/Atom/RPI/Code/Source/RPI.Private/RPISystemComponent.cpp
+++ b/Gems/Atom/RPI/Code/Source/RPI.Private/RPISystemComponent.cpp
@@ -20,9 +20,11 @@
 #include <AzCore/NativeUI/NativeUIRequests.h>
 #include <AzCore/Serialization/SerializeContext.h>
 #include <AzCore/Settings/SettingsRegistry.h>
+#include <AzCore/PlatformId/PlatformId.h>
 
 #include <AzFramework/API/ApplicationAPI.h>
 #include <AzFramework/CommandLine/CommandLine.h>
+#include <AzFramework/Components/ConsoleBus.h>
 
 #ifdef RPI_EDITOR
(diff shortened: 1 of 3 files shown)


Gems/Atom/RPI/Code/Source/RPI.Private/RPISystemComponent.cpp
```diff
@@ -185,21 +187,36 @@ namespace AZ
...
+                        &AzFramework::ConsoleRequests::ExecuteConsoleCommand, "quit");
+                }
             };
 
             auto performanceMetrics = AZStd::to_array<AZStd::string_view>({
                 PerformanceSpecGraphicsSimulationTime,
                 PerformanceSpecGraphicsRenderTime,
                 PerformanceSpecEngineCpuTime,
                 });
+            AZStd::string logCategory = GetLogCategory();
```
rgba16f, line 217: nit: no need to store in a local variable, function can be called as part of the make_unique call.

rgba16f (approved):

Gems/Atom/RPI/Code/Source/RPI.Private/RPISystemComponent.cpp
galibzon: Yes, but I do that to facilitate debugging.

antonmic (approved):

PR #12728, opened 2022-10-20
Bugfix for switcher defaulting back to Transform.

mbalfour-amzn:
## What does this PR do?

The Image Gradient Component Mode wasn't defaulting back to Transform on exit. This was happening because in its exit code, it ran multiple event ticks while showing a progress bar during image asset saving, and the component switcher was using a timer to defer some logic to switch to Transform after the Component Mode ended. Since the mode hadn't fully ended at the point the timer triggered, the switch never occurred.

This change adds more information to make it clear that the component mode has begun ending, so the switcher no longer needs to defer its logic and it can handle things immediately.

## How was this PR tested?

Manually switched between Tube, Spline, Image Gradient, and Transform multiple times successfully.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeCollection.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeCollection.cpp
@@ -213,7 +213,7 @@ namespace AzToolsFramework
         void ComponentModeCollection::BeginComponentMode()
         {
             m_selectedComponentModeIndex = 0;
-            m_componentMode = true;
+            m_componentModeState = ComponentModeState::Active;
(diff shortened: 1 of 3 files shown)


cgalvan (approved):

amzn-ahmadkrm:
Looks good to me! I believe you can get rid of the QTimer include now too as it isn't being used anywhere

amzn-ahmadkrm (approved):

mbalfour-amzn:
> Looks good to me! I believe you can get rid of the QTimer include now too as it isn't being used anywhere

Since I've already got a passing AR, I'll leave the include cleanup for another day.

hultonha (commented):
Thanks for the updates here @mbalfour-amzn and for adding the enum state 👍

PR #12732, opened 2022-10-21
Prism/edit gem workflow

tkothadev:
## What does this PR do?

A new feature is implemented to edit existing gem information from the Project Manager.

EditGemWorkflow

Primary changes are focused around the existing `CreateAGemScreen` class. A new edit workflow is introduced, and entry points are modified to change the existing workflow UI and actions depending on the type of workflow initiated (whether edit or creation based).

## How was this PR tested?

Tests were done manually for the following:
* Initiating the edit workflow from multiple gems (including backtracking, and picking a new gem). In each case, the intended gem information selected by the user is correctly loaded into the workflow for editing and publishing.
* Verifying existing UI interactions such as tabs, next and back buttons, and form fields.
* Successive operations of editing gems and creating new gems (in various combinations). In each case the application behaved as expected.
* Adding and removing tags. This also factors in the gem name change which is implicitly included with gem tags in the JSON files.

After discussing with @AMZN-Phil, I plan to create a follow up Github issue for implementing unit tests for the newly created `PythonBindings::EditGem` function. This will most likely be lumped in with a larger effort to improve unit testing for `PythonBindings` overall.

### DIFF
--- a/Code/Tools/ProjectManager/Source/CreateAGemScreen.cpp
+++ b/Code/Tools/ProjectManager/Source/CreateAGemScreen.cpp
@@ -483,4 +483,26 @@ namespace O3DE::ProjectManager
         }
     }
 
+    void CreateGem::ResetWorkflow(const GemInfo& oldGemInfo)
+    {
+        //details page
+        m_gemDisplayName->lineEdit()->setText(oldGemInfo.m_displayName);
+        m_gemName->lineEdit()->setText(oldGemInfo.m_name);
+        m_gemSummary->lineEdit()->setText(oldGemInfo.m_summary);
+        m_requirements->lineEdit()->setText(oldGemInfo.m_requirement);
+        m_license->lineEdit()->setText(oldGemInfo.m_licenseText);
+        m_licenseURL->lineEdit()->setText(oldGemInfo.m_licenseLink);
+        m_documentationURL->lineEdit()->setText(oldGemInfo.m_documentationLink);
+        m_gemLocation->lineEdit()->setText(oldGemInfo.m_path);
+        m_gemIconPath->lineEdit()->setText(oldGemInfo.m_iconPath);
+        m_userDefinedGemTags->setTags(oldGemInfo.m_features);
+
+        //creator details page
+        m_origin->lineEdit()->setText(oldGemInfo.m_origin);
+        m_originURL->lineEdit()->setText(oldGemInfo.m_originURL);
+        m_repositoryURL->lineEdit()->setText(oldGemInfo.m_repoUri);
+
+        //this will be ported over to a new EditGem class. There are a lot of subtle details, like removing the template selection, pre-filling gem values like type, and disabling folder editing, that are better handled on a separate class.
+    }
+
 } // namespace O3DE::ProjectManager
--- a/Code/Tools/ProjectManager/Source/CreateAGemScreen.cpp
+++ b/Code/Tools/ProjectManager/Source/CreateAGemScreen.cpp
@@ -45,12 +45,6 @@ namespace O3DE::ProjectManager
             this,
             [&]()
             {
-                //if previously editing a gem, cancel the operation and revert to old creation workflow
-                //this way we prevent accidental stale data for an already existing gem.
-                if(m_isEditGem)
-                {
-                    ClearWorkflow();
-                }
                 emit GoToPreviousScreenRequest();
             });
         screenLayout->addWidget(m_header);
@@ -88,6 +82,9 @@ namespace O3DE::ProjectManager
 
         setObjectName("createAGemBody");
         setLayout(screenLayout);
+
+        m_gemActionString = tr("Create");
+        m_indexBackLimit = GemTemplateSelectionScreen;
     }
 
     QFrame* CreateGem::CreateTabButtonsFrame()
@@ -135,21 +132,13 @@ namespace O3DE::ProjectManager
     {
         m_stackWidget->setCurrentIndex(GemDetailsScreen);
         m_nextButton->setText(tr("Next"));
-        m_backButton->setVisible(!m_isEditGem);
+        m_backButton->setVisible(m_stackWidget->currentIndex() != m_indexBackLimit);
     }
 
     void CreateGem::HandleGemCreatorDetailsTab()
     {
         m_stackWidget->setCurrentIndex(GemCreatorDetailsScreen);
-        if(m_isEditGem)
-        {
-            m_nextButton->setText(tr("Edit"));
-        }
-        else
-        {
-            m_nextButton->setText(tr("Create"));
-        }
-        
+        m_nextButton->setText(m_gemActionString);
         m_backButton->setVisible(true);
     }
 
@@ -377,17 +366,7 @@ namespace O3DE::ProjectManager
 
         QDir chosenGemLocation = QDir(m_gemLocation->lineEdit()->text());
         
-        bool locationValid = false;
-        
-        if(m_isEditGem)
-        {
-            //If in edit gem workflow, users cannot change this value, so this check should suffice
-            locationValid = chosenGemLocation.exists();
-        }
-        else
-        {
-            locationValid = !chosenGemLocation.exists() || chosenGemLocation.isEmpty();
-        }
+        bool locationValid = ValidateGemLocation(chosenGemLocation);
 
         m_gemLocation->setErrorLabelVisible(!locationValid);
         return locationValid;
@@ -415,15 +394,7 @@ namespace O3DE::ProjectManager
 
     void CreateGem::HandleBackButton()
     {
-        int indexBackLimit = 0;
-
-        if(m_isEditGem)
-        {
-            //this will prevent us from reaching the templates page in the edit gem workflow
-            indexBackLimit = 1;
-        }
-
-        if (m_stackWidget->currentIndex() > indexBackLimit)
+        if (m_stackWidget->currentIndex() > m_indexBackLimit)
         {
             const int newIndex = m_stackWidget->currentIndex() - 1;
             m_stackWidget->setCurrentIndex(newIndex);
@@ -438,7 +409,7 @@ namespace O3DE::ProjectManager
             }
         }
 
-        if (m_stackWidget->currentIndex() == indexBackLimit)
+        if (m_stackWidget->currentIndex() == m_indexBackLimit)
         {
             m_backButton->setVisible(false);
         }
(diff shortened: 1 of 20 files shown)


Code/Tools/ProjectManager/Source/PythonBindings.cpp
```diff
@@ -884,6 +885,74 @@ namespace O3DE::ProjectManager
...
+                    "new_type"_a = pybind11::none(), //For the Project manager, it does not make sense to change the type of a gem once it's created
+                    "new_summary"_a = QString_To_Py_String(newGemInfo.m_summary),
+                    "new_icon"_a = QString_To_Py_String(newGemInfo.m_iconPath),
+                    "new_requirements"_a = QString_To_Py_String(newGemInfo.m_requirement),
+                    "new_documentation_url"_a = QString_To_Py_String(newGemInfo.m_documentationLink),
+                    "new_license"_a = QString_To_Py_String(newGemInfo.m_licenseText),
+                    "new_license_url"_a = QString_To_Py_String(newGemInfo.m_licenseLink),
+                    "new_tags"_a = QString_To_Py_String(newList.join(" ")),  //the python code seems to interpret these lists as space separated
+                    "remove_tags"_a = QString_To_Py_String(removeList.join(" ")),
+                    "replace_tags"_a = pybind11::none())
```
AMZN-alexpete, line 935: you can omit unused/unnecessary arguments
```suggestion
                auto editGemResult = m_gemProperties.attr("edit_gem_props")(
                    "gem_path"_a = gemPath,
                    "gem_name"_a = QString_To_Py_String(oldGemInfo.m_name),
                    "new_name"_a = QString_To_Py_String(newGemInfo.m_name),
                    "new_display"_a = QString_To_Py_String(newGemInfo.m_displayName),
                    "new_origin"_a = QString_To_Py_String(newGemInfo.m_origin),
                    "new_summary"_a = QString_To_Py_String(newGemInfo.m_summary),
                    "new_icon"_a = QString_To_Py_String(newGemInfo.m_iconPath),
                    "new_requirements"_a = QString_To_Py_String(newGemInfo.m_requirement),
                    "new_documentation_url"_a = QString_To_Py_String(newGemInfo.m_documentationLink),
                    "new_license"_a = QString_To_Py_String(newGemInfo.m_licenseText),
                    "new_license_url"_a = QString_To_Py_String(newGemInfo.m_licenseLink),
                    "new_tags"_a = QString_To_Py_String(newList.join(" ")),  //the python code interprets these lists as space separated
                    "remove_tags"_a = QString_To_Py_String(removeList.join(" ")),
```

Also, you could convert the tags to a list like how it is done in UpdateProject() - might warrant a helper or macro to convert `QStringList` to `pybind11::list`

Code/Tools/ProjectManager/Source/PythonBindings.cpp
```diff
@@ -884,6 +885,74 @@ namespace O3DE::ProjectManager
...
+        QStringList removeList;
+
+        for (auto feature : oldGemInfo.m_features)
+        {
+            if(!newGemInfo.m_features.contains(feature))
+            {
+                //this must have been removed by the user
+                removeList << feature;
+            }
+        }
```
AMZN-alexpete, line 912: Rather than requiring old gem info and new gem info to populate new_tags and remove_tags, I believe you can just use replace_tags parameter like we do in  `UpdateProject`.  then you can remove `oldGemInfo` parameter.

Code/Tools/ProjectManager/Source/GemCatalog/GemModel.h
```diff
@@ -65,6 +65,7 @@ namespace O3DE::ProjectManager
         QVector<Tag> GetDependingGemTags(const QModelIndex& modelIndex);
         bool HasDependentGems(const QModelIndex& modelIndex) const;
 
+        static const GemInfo GetGemInfo(const QModelIndex& modelIndex);
```
AMZN-alexpete, line 68: can you mark this function as const also?
```suggestion
        static const GemInfo GetGemInfo(const QModelIndex& modelIndex) const;
```

Code/Tools/ProjectManager/Source/GemCatalog/GemInspector.h
```diff
@@ -44,10 +44,16 @@ namespace O3DE::ProjectManager
...
+        QModelIndex GetCurrentModelIndex()
+        {
+            return m_curModelIndex;
+        }
+
     signals:
         void TagClicked(const Tag& tag);
         void UpdateGem(const QModelIndex& modelIndex);
         void UninstallGem(const QModelIndex& modelIndex);
+        void EditGem();
```
AMZN-alexpete, line 56: If you pass in the modelIndex in `EditGem` as a parameter it looks like you won't need the `GetCurrentModelIndex` function at all

AMZN-alexpete (commented):
See above comments for suggestions.  

Have you considered removing the edit functionality from `CreateAGemScreen` and putting it in separate class that inherits from `CreateAGemScreen`? Most of the other create/edit screens and controllers separate the two.  You will likely have to move some functionality to protected methods that can be overridden in the edit class, e.g. the back/next button visibility.

amzn-leenguy (approved):
Great work. Approved from UX POV

Code/Tools/ProjectManager/Source/GemCatalog/GemModel.h
tkothadev: Looks like I'm getting a compiler error stating that modifier is not allowed on static methods:

```
'GetGemInfo': modifiers not allowed on static member functions	
```

tkothadev:
Regarding @AMZN-alexpete 's suggestion, I will create a common base class that the create/edit gem workflows share from. That way functionality/logic unique to editing is distinct from creation.

AMZN-Phil:
It looks like some fields aren't covered by the Python function (looks like at least Origin URL and Repo URL fall into this) but are visible in the edit gem workflow. These should be added to edit_gem_props so they are updated too, having editable fields that do not update the corresponding keys in the gem feels like unexpected behavior.

tkothadev:
Per my discussion with @AMZN-Phil, I plan on addressing the issue he raised regarding the python function in a future PR where I address platform support in the Python CLI.

AMZN-Phil (approved):

Code/Tools/ProjectManager/Source/EditAGemScreen.h
```diff
@@ -0,0 +1,67 @@
...
+
+#endif
+
+QT_FORWARD_DECLARE_CLASS(QButtonGroup)
+QT_FORWARD_DECLARE_CLASS(QDialogButtonBox)
+QT_FORWARD_DECLARE_CLASS(QRadioButton)
+QT_FORWARD_DECLARE_CLASS(QScrollArea)
+QT_FORWARD_DECLARE_CLASS(QVBoxLayout)
+QT_FORWARD_DECLARE_CLASS(QStackedWidget)
+QT_FORWARD_DECLARE_CLASS(QDir)
```
AMZN-alexpete, line 31: you can remove most of the unused includes here because they are no longer needed - most of the Qt types are in the parent class header.

Code/Tools/ProjectManager/Source/CreateAGemScreen.h
```diff
@@ -34,36 +38,54 @@ namespace O3DE::ProjectManager
...
         bool ValidateFormNotEmpty(FormLineEditWidget* form);
         bool ValidateRepositoryURL();
+        
+        void ProceedToGemDetailsPage();
+        void ProceedToGemCreatorDetailsPage();
+        void ProceedToGemAction();
+
+        virtual void GemAction();
+
+        virtual bool ValidateGemLocation(QDir chosenGemLocation)
```
AMZN-alexpete, line 75: It'd be better to not pass QDir by value, but instead by const reference or pointer.  By doing this and moving the implementation to the .cpp you could move the QDir header include to the .cpp file also and just forward declare it here.  It's good to get in the habit of minimizing header includes which will help keep compile times down.

Also, this function can be const.

Code/Tools/ProjectManager/Source/CreateAGemScreen.h
```diff
@@ -34,36 +38,54 @@ namespace O3DE::ProjectManager
         explicit CreateGem(QWidget* parent = nullptr);
         ~CreateGem() = default;
 
+        void ClearFields();
+
+        ProjectManagerScreen GetScreenEnum() override
```
AMZN-alexpete, line 43: Not sure why we haven't made this function const everywhere - but that'll touch a lot of files so we can do that in a separate review

Code/Tools/ProjectManager/Source/CreateAGemScreen.h
```diff
@@ -34,36 +38,54 @@ namespace O3DE::ProjectManager
         explicit CreateGem(QWidget* parent = nullptr);
         ~CreateGem() = default;
 
+        void ClearFields();
```
AMZN-alexpete, line 41: Lets move this function into `protected` scope if it's only meant to be called by this class and classes that inherit from it.

Code/Tools/ProjectManager/Source/CreateAGemScreen.h
```diff
@@ -24,6 +27,7 @@ QT_FORWARD_DECLARE_CLASS(QRadioButton)
 QT_FORWARD_DECLARE_CLASS(QScrollArea)
 QT_FORWARD_DECLARE_CLASS(QVBoxLayout)
 QT_FORWARD_DECLARE_CLASS(QStackedWidget)
+QT_FORWARD_DECLARE_CLASS(QDir)
```
AMZN-alexpete, line 30: Note that when you include the actual header <Qdir> you should not forward declare the class - but it is preferable to not have the include in the header if possible and to use a forward declare so you can keep compile times optimized

Code/Tools/ProjectManager/Source/CreateAGemScreen.h
```diff
@@ -16,6 +16,9 @@
 #include <FormComboBoxWidget.h>
 #include <GemCatalog/GemInfo.h>
 #include <PythonBindings.h>
+#include <ScreenHeaderWidget.h>
+#include <ScreenDefs.h>
```
AMZN-alexpete, line 20: You should not need this include because `ScreenWidget.h` includes it.
```suggestion
```

Code/Tools/ProjectManager/Source/CreateAGemScreen.h
```diff
@@ -92,12 +114,24 @@ namespace O3DE::ProjectManager
         QRadioButton* m_gemDetailsTab = nullptr;
         QRadioButton* m_gemCreatorDetailsTab = nullptr;
 
-        GemInfo m_createGemInfo;
+        GemInfo m_focalGemInfo;
```
AMZN-alexpete, line 117: focal? as in the focused gem? maybe it'd be simple enough to just call it `m_gemInfo` and not characterize what kind of gem info it is in this class?

Code/Tools/ProjectManager/Source/EditAGemScreen.cpp
```diff
@@ -0,0 +1,146 @@
...
+namespace O3DE::ProjectManager
+{
+
+    EditGem::EditGem(QWidget* parent)
+        : CreateGem(parent)
+    {
+        //undo the connections of the parent class
+        m_header->backButton()->disconnect();
+        m_backButton->disconnect();
+        m_nextButton->disconnect();
```
AMZN-alexpete, line 20: Another option - instead of disconnecting the parent events, would be to move the parent's calls to `connect` into a function called `Init` or something that is virtual and `EditGem` could override this method and have its own connect methods there.

Code/Tools/ProjectManager/Source/CreateAGemScreen.h
tkothadev: I call it `m_focalGemInfo` because the `EditGem` class actually retains the `m_oldGemInfo`  temporarily when pre-filling values and supplying necessary arguments to the python edit gems function.

Code/Tools/ProjectManager/Source/GemCatalog/GemCatalogScreen.cpp
```diff
@@ -640,6 +650,22 @@ namespace O3DE::ProjectManager
...
     }
 
+    void GemCatalogScreen::HandleEditGem(const QModelIndex& currentModelIndex)
+    {
+        if (m_screensControl)
+        {
+            ScreenWidget* editGemScreen = m_screensControl->FindScreen(ProjectManagerScreen::EditGem);
+            if (editGemScreen)
+            {
+                auto editGem = qobject_cast<EditGem*>(editGemScreen);
```
AMZN-alexpete, line 660:
```suggestion
            auto* editGemScreen = qobject_cast<EditGem*>(m_screensControl->FindScreen(ProjectManagerScreen::EditGem));
            if (editGemScreen)
            {
```

Code/Tools/ProjectManager/Source/EditAGemScreen.h
```diff
@@ -0,0 +1,49 @@
...
+    private: 
+        void GemAction() override;
+
+        bool ValidateGemLocation(QDir chosenGemLocation) override
+        {
+            return chosenGemLocation.exists();
+        }
+
+        //Edit Gem workflow
+        bool m_isEditGem = false;
```
AMZN-alexpete, line 43: m_isEditGem is unused
```suggestion

```

Code/Tools/ProjectManager/Source/CreateAGemScreen.h
AMZN-alexpete: It looks like you just need to save the old gem name - perhaps you could use `m_gemInfo` to store the final info and `m_originalName` to store the name needed to update the gem info?

Code/Tools/ProjectManager/Source/EditAGemScreen.cpp
```diff
@@ -0,0 +1,128 @@
...
+        m_userDefinedGemTags->setTags(tagsToEdit);
+
+        //creator details page
+        m_origin->lineEdit()->setText(oldGemInfo.m_origin);
+        m_origin->setErrorLabelVisible(false);
+
+        m_originURL->lineEdit()->setText(oldGemInfo.m_originURL);
+        m_repositoryURL->lineEdit()->setText(oldGemInfo.m_repoUri);
+
+        m_oldGemInfo = oldGemInfo;
```
AMZN-alexpete, line 98: It looks like you could just store the original name - perhaps `m_originalName`? and avoid copying the whole structure

Code/Tools/ProjectManager/Source/CreateAGemScreen.cpp
```diff
@@ -406,81 +409,156 @@ namespace O3DE::ProjectManager
...
-            }
+            ProceedToGemAction();
         }
     }
 
+    //this ignores the templates page
+    void CreateGem::ClearFields()
+    {
+        //details page
+        m_gemDisplayName->lineEdit()->clear();
```
AMZN-alexpete, line 525: hmm wouldn't it be nice if you could just issue a single line "reset" signal to any fields in a form listening for it and they would clear/reset error labels?
e.g. our field widgets would connect to the parent (or maybe some passed in QObject?) listening for the signal
```
connect(parent, SIGNAL(reset()), this, SLOT(reset()));
```
and forms/screens could call it with
```
emit reset();
```
A possible future handy bit of functionality

Code/Tools/ProjectManager/Source/CreateAGemScreen.h
```diff
@@ -34,36 +38,54 @@ namespace O3DE::ProjectManager
...
         bool ValidateGemTemplateLocation();
         bool ValidateGemDisplayName();
         bool ValidateGemName();
         bool ValidateGemPath();
         bool ValidateFormNotEmpty(FormLineEditWidget* form);
         bool ValidateRepositoryURL();
+        
+        void ProceedToGemDetailsPage();
+        void ProceedToGemCreatorDetailsPage();
+        void ProceedToGemAction();
```
AMZN-alexpete, line 71: These look like there are only used by `CreateGem` - in which case they can be moved to the `private:` section

Code/Tools/ProjectManager/Source/CreateAGemScreen.h
```diff
@@ -34,36 +38,54 @@ namespace O3DE::ProjectManager
...
-        QScrollArea* CreateGemCreatorScrollArea();
-        QFrame* CreateTabButtonsFrame();
-        QFrame* CreateTabPaneFrame();
+    protected:
         bool ValidateGemTemplateLocation();
         bool ValidateGemDisplayName();
         bool ValidateGemName();
         bool ValidateGemPath();
         bool ValidateFormNotEmpty(FormLineEditWidget* form);
         bool ValidateRepositoryURL();
```
AMZN-alexpete, line 67: It looks like all these non-virtual functions that are not called by `EditGem` could be moved to the `private:` section

Code/Tools/ProjectManager/Source/CreateAGemScreen.cpp
```diff
@@ -406,81 +409,156 @@ namespace O3DE::ProjectManager
...
+    void CreateGem::GemAction()
+    {
+        QString templateLocation;
+        if (m_formFolderRadioButton->isChecked())
+        {
+            templateLocation = m_gemTemplateLocation->lineEdit()->text();
+        }
+        else
+        {
+            int templateID = m_radioButtonGroup->checkedId();
```
AMZN-alexpete, line 469:
```suggestion
            const int templateID = m_radioButtonGroup->checkedId();
```

Code/Tools/ProjectManager/Source/CreateAGemScreen.h
tkothadev: I'll opt for `m_oldGemName` for further clarity

Code/Tools/ProjectManager/Source/EditAGemScreen.cpp
tkothadev: Instead of calling it `Init` we could just call it `HookConnections` so that the purpose is clear. It seems like we can call `connect` at any point in initialization, so long as it's done before the user can start interacting with the application.

Code/Tools/ProjectManager/Source/CreateAGemScreen.cpp
tkothadev: I'll mark this as a github issue so that it can be addressed in the future.

Code/Tools/ProjectManager/Source/GemCatalog/GemModel.h
AMZN-alexpete: Ah yeah good point, I didn't notice it was static.

Code/Tools/ProjectManager/Source/ScreenWidget.h
```diff
@@ -54,7 +54,9 @@ namespace O3DE::ProjectManager
         virtual void GoToScreen([[maybe_unused]] ProjectManagerScreen screen)
         {
         }
-
+        virtual void HookConnections()
```
AMZN-alexpete, line 57: nit: I would prefer to call this something more generic like `Init()` as it could be used for more than just setting up connections, but am fine with calling it this for now.

AMZN-alexpete (approved):
One minor suggestion but over-all I'm good with this.

PR #12737, opened 2022-10-21
O3DE launcher running on Quest 2 natively

moraaar:
## What does this PR do?

Added support for running the launcher on Quest 2 natively. Also support enable/disabling the default render pipeline on host while there is a xr system present (launcher only).

**Approach**
There are 3 key elements at play:
- Render Pipelines. Created by BootstrapSystemComponent.
- Render Swapchains. Created by WindowsContext.
- Views. Created by cameras and other elements.

Each one of these elements have one for default rendering and 2 for XR rendering (one for each eye). 

This PR changes BootstrapSystemComponent to create only render pipelines that's needed. So for example, if there is no XR system then the XR pipelines are not created, or if we're running on an xr system on mobile then there is no default pipeline.

The same happens with render swapchains. This was already present before, this PR just improves the query of what to create, now centralized in XR gem. This change goes along with this PR: https://github.com/o3de/o3de-extras/pull/49

ViewGroup is the new class that holds the 3 views (default + XRs). Before, a view was pushed/pop to viewport contexts, now the interface push/pop view groups. Since ViewGroups are used outside Atom scope, to keep its use simple and not spread the logic of "when to created default vs XR views", users will always call `ViewGroup::CreateMainView` and `ViewGroup::CreateStereoscopicViews` and internally the ViewGroup will create XR views only when there is an Xr system present.

ViewportContext is where views and render pipelines come together. **Views queried always have to be valid**, then ViewportContext will obtain the pipeline of the view type and it will handle if the pipeline is null or not. ViewportContext also handles enabling/disable default pipeline rendering when there is an xr system. Code now handles correctly if a render pipeline is null.

**Limitation**
At the moment all cameras create xr views when there is an xr system, it cannot be chosen that for a certain camera do not create the xr views (for example editor camera). This limitation comes from pipelines always need to have a valid view and it's not trivial to modify the internals of RenderPipeline to handle null views. Because of this limitation when running VR with editor opened, the xr pipeline is still rendering to the device even outside game mode (and since editor camera does not update xr view matrices then the headset can show odd images while not in game mode). This is something that can be improved in the future.

## How was this PR tested?

Using a project with XR and OpenXRVk gems enabled:
- Run launcher using a default level on Quest 2 natively.
- Run editor on PC with a Quest 2 connected via link cable. Tested that entering in game mode it renders on Quest 2.
- Run launcher using a default level on PC with a Quest 2 connected via link cable. Tested the level renders correctly on pc and quest 2. Using remote console also tested the cvar `r_EnableHostRenderPipelineOnXR` to enable/disable rendering on PC.

Also used `XR_Office` level from `OpenXRTest` project.


amzn-phist (approved):
Looks good to me, although I'm not an Atom expert.

Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
```diff
@@ -306,62 +306,77 @@ namespace AZ
...
+                        }
+                    }
                 }
 
+                // Apply MSAA state to all the render pipelines.
+                // It's important to do this after all the pipelines have
+                // been created so the same values are applied to all.
+                // As it cannot be applied MSAA values per pipeline,
+                // it's setting the MSAA state from the last pipeline loaded.
+                AZ::RPI::RPISystemInterface::Get()->SetApplicationMultisampleState(multisampleState);
```
moudgils, line 378: This is problematic. There is/was a bug where you couldnt set MSAA state after the pipeline is loaded and added to the scene. I had to revert this specific change due to that bug (https://github.com/o3de/o3de/pull/12359). Maybe it is fixed due to recent changes made by Vicky here https://github.com/o3de/o3de/pull/12700 but this needs to be confirmed by @VickyAtAZ and @dmcdiarmid-ly. If that bug is fixed I am ok with this change as that is how I intended it to work but if not we need to revisit this code.

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/ViewportContext.h
```diff
@@ -61,11 +61,21 @@ namespace AZ
...
+            ConstViewGroupPtr GetViewGroup() const;
+
             //! Gets the default view associated with this ViewportContext.
-            //! Alternatively, use  AZ::RPI::ViewportContextRequests::Get()->GetCurrentView().
+            //! Alternatively, use AZ::RPI::ViewportContextRequests::Get()->GetCurrentViewGroup()->GetView().
             ViewPtr GetDefaultView();
             ConstViewPtr GetDefaultView() const;
 
+            //! Gets the stereoscopic view associated with this ViewportContext.
+            //! Alternatively, use AZ::RPI::ViewportContextRequests::Get()->GetCurrentViewGroup()->GetView(AZ::RPI::ViewType::XrLeft|XrRight).
```
moudgils, line 75: The comment seems to imply you can pass AZ::RPI::ViewType::XrLeft|XrRight into this function which I dont think is correct. Maybe just remove XrLeft|XrRight part of the comment.

moudgils (commented):
Most of the changes look good except the one issue related to setting MSAA state. Also make sure to do thorough testing across all Editor flows by checking that stuff like material editor, ui editor, animation editor works as before.

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/ViewportContext.h
moraaar: done

Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
moraaar: It's funny that I changed it to be the same code you had it in first place in #12359  :) 

Unfortunately without this change it lead to errors on Quest 2. Previously it was applying MSAA for default renderer only and not for XR pipelines. That causes errors when running on Quest2 because there is no default renderer and that means the MSAA state never gets applied and that leads to errors with the light culling pass mismatching the depth image descriptor having 1 sample set and the depth texture in the shader using TextureMS sampler (which expects image descriptors with samples >1 set).

Then I modified the code to set the MSAA at least once, which means on PC it'll be when loading the default renderer and on VR that'll be on the first XR pipeline (left eye pipeline, and not the right eye pipeline). That get rid of the error that it was rendering differently left eye and right eye, the shadows didn't appear in one of them.

I think I tested to always apply it and that didn't work either, but I can't remember what was the result.

The only way I got it to work for both PC and XR was by applying MSAA at the end after all pipelines were created.

The function `SetApplicationMultisampleState` internally does this

````
            // reinitialize pipelines for all scenes
            for (auto& scene : m_scenes)
            {
                for (auto& renderPipeline : scene->GetRenderPipelines())
                {
                    renderPipeline->GetRenderSettings().m_multisampleState = multisampleState;
                    renderPipeline->MarkPipelinePassChanges(PipelinePassChanges::MultisampleStateChanged);
                }
            }
````

Since it goes through all the render pipelines to set the MSAA state I thought it seems to make sense to apply it after the render pipelines have been created and added to the scenes.

@moudgils , @VickyAtAZ @dmcdiarmid-ly please advice on how to proceed. I'm happy to test other variants of when to set the MSAA state if you have other suggestions.

Thanks

moudgils: Lets first confirm if this bug was fixed. Can you check that DiffuseProbeGrid component is working properly within the Editor with these changes?. Try adding DiffuseProbeGrid volume on top of the shaderball in the default scene, the bottom of the shaderball should receive light from the ground. Ping @dmcdiarmid-ly if you need further clarification for testing this as he is the author of that component.

moraaar: Tested DiffuseProbeGrid component in a default level, I sent a screenshot to @dmcdiarmid-ly for verification and he confirmed: "The default irradiance color is white, so that looks like it's working.". I also tested changing the Sun's directional light color and it did affect the probes.

@moudgils please proceed to approve the PR if there is nothing else. Thanks.

moudgils (approved):

Issue #12628, opened 2022-10-14
Component Mode Switcher doesn't handle component dependency changes


mbalfour-amzn:
**Describe the bug**
If a component with Component Mode gets activated or deactivated by having a component it depends on added or removed, the Component Mode Switcher doesn't refresh.

**Steps to reproduce**
1. Run the Editor and create or open a level.
2. Create a new entity.
3. Add a tube shape component.
4. Add a spline component.

**Expected behavior**
Both the tube shape and spline components show up in the Component Mode Switcher.

**Actual behavior**
Only the spline shows up until you click off of the entity and back onto it.

**Screenshots/Video**
https://user-images.githubusercontent.com/82224783/195917605-fbd399ac-8831-429f-94d8-d107ce86c53e.mp4

**Found in Branch**
development

hultonha:
Fixed by #12746

PR #12746, opened 2022-10-21
Bug fix for handling component dependency changes

amzn-ahmadkrm:

## What does this PR do?

Addresses https://github.com/o3de/o3de/issues/12628 where the switcher would not update if a component that has dependents gets added/removed/disabled.

For example, the tube shape component requires a spline component, when an entity has a tube shape but not a spline the user cannot enter tube shape component mode. If the user clicks "Add Required Components" to add the spline the switcher should add both the tube component and spline component but did not. This PR changes the way components are added to the switcher when a new component is added or enabled/disabled to allow for the dependency changes.

## How was this PR tested?

Manually tested with Tube shape dependency and Joints.

Editor_kdNWI8TsRn


Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeSwitcher.cpp
```diff
@@ -384,19 +384,23 @@ namespace AzToolsFramework::ComponentModeFramework
...
-            else
+            else if (m_addOrRemove == AddOrRemoveComponent::Add && m_callCount < 1)
             {
-                RemoveComponentButton(m_componentModePair);
+                m_callCount += 1;
+                return;
             }
+
+            UpdateSwitcher(entityIdList, {});
+            m_callCount = 0;
```
mbalfour-amzn, line 403: Is there an alternate way to handle this logic that doesn't make as many assumptions about its calling pattern? This seems pretty fragile because any change to the order or frequency of calls happening somewhere else in the code can cause this to break.

Ideally, it seems like all composition changes could just perform logic similar to UpdateSwitcherOnEntitySelectionChange() where it clears everything out and rebuilds the button list based on the set of common components that now exist, and not need to track the number of calls or whether or not an addOrRemove happened.

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeSwitcher.h
```diff
@@ -130,6 +130,9 @@ namespace AzToolsFramework
             AddOrRemoveComponent m_addOrRemove = AddOrRemoveComponent::Add; //!< Setting to either add or remove component.
             //! Protects the switcher from being opened by OnImGuiDropDownShown if it has been hidden elsewhere.
             bool m_hiddenByImGui = false;
+            //! The notification OnEntityCompositionChanged is called twice when an Entities composition changes, this is used to
```
mbalfour-amzn, line 133: Nit: "Entities" -> "Entity's"

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeSwitcher.cpp
amzn-ahmadkrm: I was thinking to do that but I don't want the switcher to have to clear and repopulate the switcher everytime something happens, but do you think that has a relatively low overhead?

mbalfour-amzn: In a general sense, I'm trying to push for correctness and maintainability over speed of execution here. The current design is maintaining and merging partial state knowledge ("are we adding or removing"? "how many times are we called"?) across separate API entry points, which seems like it's opening the door to errors and isn't _necessarily_ making things perceptibly faster, though I haven't measured the system to say concretely one way or the other.

Instead of trying to maintain partial state, I'm wondering if it's possible to just react to whatever state has currently changed (ideally in an efficient way), and not worry as much about reacting multiple times to multiple rapid state changes.

At a high level, it seems like at any point in time, the Switcher has a set of active entities and components that it is displaying, and it receives notifications that something has changed about one or more of those entities and components. So instead of trying to defer processing on the notifications, it seems like we just want an efficient way to process changes. I think these just boil down to something like this:
* Component added: Check to see if this component is already displayed. If not, check to see if it *should* be displayed by seeing if it exists on all selected entities and has a component mode. If so, then add it.
* Component removed: Check to see if this component is already displayed. If yes, then remove it.
* Entity changed: For each component on the entity, go through component added. For each component on the switcher, if it's not on the entity, remove it.

And then the Switcher would only send out change notifications if components have *actually* been changed in the Switcher list. With these changes, I think both m_callCount and m_addOrRemove could get removed entirely.

It's certainly possible that this could cause performance problems by performing some redundant logic, and it would be interesting to profile selection and modification of 1000 duplicate entities with several editable components on them, to see if there _are_ in fact any performance problems here. My _intuition_ is that if any performance problems exist, the solution to them would either be faster ways to merge the entity and component lists, or triggering less updates at a higher level, rather than trying to work around extra updates inside the Switcher.

But right now, I'm just asking this all more as a question. If you've tried alternatives, and measured this already, and this is the best way to optimize it, that's totally OK to say so. My request in that case would just be to add some super explicit comments in the code to this effect. :)

amzn-ahmadkrm: I agree, the checking of the amount of calls is not necessarily sustainable especially if it changes to get called once for example in above example, it previously just accepted that it was being called twice and dealt with it so I can just revert that. With the switcher I guess there will only be a max of 10s of components for the foreseeable so performance wouldn't be negatively impacted. 

With regards to m_addOrRemove, when I receive a notification OnEntityComponentAdded originally I wanted to process the component then but the component doesn't "exist" on the entity until I receive a notification from OnEntityCompositionChanged which is why they are paired and share state knowledge. I don't like it but have tried searching for alternatives for getting the absolute state of if a component has been added or removed but couldn't find any.

I appreciate your feedback thank you :)

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeSwitcher.h
```diff
@@ -104,7 +104,7 @@ namespace AzToolsFramework
             // EntityCompositionNotificationBus overrides ...
             void OnEntityComponentAdded(const AZ::EntityId& entityId, const AZ::ComponentId& componentId) override;
             void OnEntityComponentRemoved(const AZ::EntityId& entityId, const AZ::ComponentId& componentId) override;
-            void OnEntityComponentEnabled(const AZ::EntityId& entityId, const AZ::ComponentId& componentId) override;
+            void OnEntityComponentEnabled(const AZ::EntityId& entityId, [[maybe_unused]] const AZ::ComponentId& componentId) override;
             void OnEntityComponentDisabled(const AZ::EntityId& entityId, [[maybe_unused]] const AZ::ComponentId& componentId) override;
```
hultonha, line 108: nit: [[maybe_unused]] can be removed here in both instances

hultonha (commented):
@amzn-ahmadkrm I'll take a look at this once the changes have been made after @mbalfour-amzn's feedback

Code/Framework/AzToolsFramework/Tests/ComponentModeTestFixture.h
```diff
@@ -10,14 +10,32 @@
...
 
 namespace UnitTest
 {
     class ComponentModeTestFixture
         : public ToolsApplicationFixture
+        , public AzToolsFramework::EditorDisabledCompositionRequestBus::Handler
     {
+    public:
+        // EditorDisabledCompositionRequestBus overrides ...
+        void GetDisabledComponents(AZStd::vector<AZ::Component*>& components) override;
```
hultonha, line 25: update: Is this missing an `override` keyword if it's overriding a function?

hultonha (approved):
These updates look good to me. It might be nice to look at seeing where we can simplify in future but for now if this does the job and we have tests to sanity check the behavior that's a good place to be 👍

Code/Framework/AzToolsFramework/Tests/ComponentModeTestFixture.h
```diff
@@ -10,14 +10,32 @@
...
+        void AddDisabledComponent(AZ::Component* componentToAdd) override;
+        void RemoveDisabledComponent(AZ::Component* componentToRemove) override;
+
+        void Connect(AZ::EntityId entityId);
+        void Disconnect();
+        // Adds disabled component to
```
jonawals, line 31: Is this comment needed? It looks incomplete and most likely the function name `AddDisabledComponentToBus` is descriptive enough :p

Code/Framework/AzToolsFramework/Tests/ComponentModeTestFixture.h
```diff
@@ -10,14 +10,32 @@
...
 #include <AzToolsFramework/UnitTest/AzToolsFrameworkTestHelpers.h>
+#include <AzToolsFramework/ToolsComponents/EditorDisabledCompositionBus.h>
+
 #include <AzCore/UnitTest/TestTypes.h>
 
 namespace UnitTest
 {
     class ComponentModeTestFixture
         : public ToolsApplicationFixture
+        , public AzToolsFramework::EditorDisabledCompositionRequestBus::Handler
```
jonawals, line 21: I *think* this can be `private` instead of `public`, no? 🤔

jonawals (approved):
Some minor suggestions but otherwise LGTM

mbalfour-amzn (approved):

PR #12747, opened 2022-10-21
Image suffix support

mbalfour-amzn:
## What does this PR do?

The Atom Image Builder uses a set of user-definable file name suffixes to select the correct image settings for processing the image. These suffixes take the form “_albedo”, “_gsi”, etc. See https://www.o3de.org/docs/user-guide/assets/texture-settings/texture-presets/ for more details. The change that’s being made is that an additional condition is being added to the suffix check so they will only compare the filename between the last underscore and the first period. For example, the name “image_gsi.0000.tif” used to try comparing “_gsi.0000” to the image suffixes. Now it will compare “_gsi” and ignore the “.0000”.

There's also a small mostly-unrelated null crash fix in this PR that I found while debugging the builder code in the Asset Processor. I wasn't making it to the image builder because some other malformed assets on my drive were crashing the EditorDisabledComponent. This crash fix prevents malformed assets from crashing. :) 

## How was this PR tested?

Verified that assets following the pattern "image_gsi.0000.tif" were getting processed as GSI files.
Verified that assets following regular patterns ("chair_normal.tif" etc) were getting processed as normals, etc.

### DIFF
--- a/Gems/Atom/Asset/ImageProcessingAtom/Code/Source/BuilderSettings/BuilderSettingManager.cpp
+++ b/Gems/Atom/Asset/ImageProcessingAtom/Code/Source/BuilderSettings/BuilderSettingManager.cpp
@@ -557,6 +557,14 @@ namespace ImageProcessingAtom
                     AZ_Warning(LogWindow, false, "File mask '%s' is invalid. It must contain only a single '%c' character.", filemask.c_str(), FileMaskDelimiter);
                     continue;
                 }
+                else if (filemask.find(AZ_FILESYSTEM_EXTENSION_SEPARATOR) != AZStd::string::npos)
+                {
+                    AZ_Warning(LogWindow, false,
+                        "File mask '%s' is invalid. It must not contain a file extension separator ('%c').",
+                        filemask.c_str(),
+                        AZ_FILESYSTEM_EXTENSION_SEPARATOR);
+                    continue;
+                }
 
                 extraString += (filemask + preset.m_name.GetCStr());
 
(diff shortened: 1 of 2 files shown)


cgalvan (approved):

Gems/Atom/Asset/ImageProcessingAtom/Code/Source/BuilderSettings/BuilderSettingManager.cpp
```diff
@@ -599,7 +607,14 @@ namespace ImageProcessingAtom
         size_t lastUnderScore = fileName.find_last_of(FileMaskDelimiter);
         if (lastUnderScore != AZStd::string::npos)
         {
-            return fileName.substr(lastUnderScore);
+            // If the complete file name contains multiple extension separators ('.'), the GetFileName() call above will
```
VickyAtAZ, line 610: We may use QFileInfo which we can just use baseName() to get the file name without any '.' parts at line 604.

VickyAtAZ (approved):

Gems/Atom/Asset/ImageProcessingAtom/Code/Source/BuilderSettings/BuilderSettingManager.cpp
mbalfour-amzn: Great suggestion, I'll change this and also simplify the explanatory comment. :)

VickyAtAZ (approved):

PR #12758, opened 2022-10-21
add further test for transforming AABBs

greerdv:

## What does this PR do?
Adds a test for Aabb::GetTransformedAabb, which was previously not directly tested.

## How was this PR tested?
Running unit tests.


greerdv:
I added this test because while working on another feature I wasn't completely sure GetTransformedAabb was returning the correct result. I checked the tests and discovered that nothing was directly testing that function, only comparing it with the result from calling ApplyTransform (which is what GetTransformedAabb also does under the hood, so it's possible they could both have been wrong). The test I added (with expected values I calculated by hand) is passing, so it looks like the behaviour is correct, but now we have a test to give evidence for that.

greerdv:
I also figured out the reason for the results I was getting which caused me to doubt GetTransformedAabb in the first place, and it had nothing to do with GetTransformedAabb. Still think it's useful to have a test which directly targets that function though.

Code/Framework/AzCore/Tests/Math/AabbTests.cpp
```diff
@@ -410,6 +410,20 @@ namespace UnitTest
         EXPECT_TRUE(aabb.GetMax().IsClose(transAabb.GetMax()));
     }
 
+    TEST(MATH_AabbTransform, GetTransformedAabbCorrectResult)
```
pollend, line 413: probably can parameterize this to add more variations.

pollend (approved):

moraaar (approved):

hultonha (approved):
Nice addition @greerdv, great to see! 🙂

Code/Framework/AzCore/Tests/Math/AabbTests.cpp
greerdv: Apologies for the delay in replying, was working on something else last week. 

I'd say there is value in going from 0 tests to 1, but I think going from 1 to many tests would have diminishing returns here, especially since there aren't multiple code paths in the implementation. I'm also a bit reluctant to add a  parameterized test because it would make it a bit harder to read/debug the test.

PR #12761, opened 2022-10-22
Improve prefab instance API and add Doxygen comments

forkercat:
## What does this PR do?

This PR improves prefab Instance class API as discussed and adds missing Doxygen comments for the class.

#### Major Changes
1. Resolve confusion about non-recursive and recursive operations
    1. **Non-recursive name pattern:** GetEntities, GetEntityIds, GetNestedInstances
    2. **Recursive name pattern:** GetAllEntitiesInHierarchy, GetAllEntityIdsInHierarchy
2. Add missing Doxygen comments in public and private functions

#### Minor Changes
1. Rename internal function DetachEntity to EraseEntityFromMap to avoid name confusion with the public function
2. Reorder some functions in Instance.cpp to match the new order in header file
3. Try to add [Doxygen grouping](https://www.doxygen.nl/manual/grouping.html) `@{}` as I saw the usage in other O3DE files

## How was this PR tested?

It does not alter any implementation logic.


Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/Instance.h
```diff
@@ -80,182 +83,275 @@ namespace AzToolsFramework
...
+            //! @param templateId Template id to be set.
             void SetTemplateId(TemplateId templateId);
 
+            //! Gets template source path.
+            //! @return Template source path of the instance.
             const AZ::IO::Path& GetTemplateSourcePath() const;
+
+            //! Sets template id.
+            //! @param sourcePath Template source path to be set.
             void SetTemplateSourcePath(AZ::IO::Path sourcePath);
```
forkercat, line 115: notice a typo. Will fix

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/Instance.cpp
```diff
@@ -223,10 +223,10 @@ namespace AzToolsFramework
                 }
             }
 
-            return DetachEntity(entityAliasToRemove);
+            return EraseEntityFromMap(entityAliasToRemove);
         }
 
-        AZStd::unique_ptr<AZ::Entity> Instance::DetachEntity(const EntityAlias& entityAlias)
+        AZStd::unique_ptr<AZ::Entity> Instance::EraseEntityFromMap(const EntityAlias& entityAlias)
```
srikappa-amzn, line 229: I don't think 'EraseEntityFromMap' is a better name than DetachEntity(). Detach implies that it removes and returns entity for further action. Erase doesn't imply that. I don't think having the same name for public and private functions is an issue because it's only the public one that's accessible. If you still prefer a different name, you can go with DetachEntityHelper()

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/Instance.h
```diff
@@ -71,6 +72,8 @@ namespace AzToolsFramework
             using AliasToInstanceMap = AZStd::unordered_map<InstanceAlias, AZStd::unique_ptr<Instance>>;
             using AliasToEntityMap = AZStd::unordered_map<EntityAlias, AZStd::unique_ptr<AZ::Entity>>;
 
+            //! Construtors.
```
srikappa-amzn, line 75: Doxygen grouping sounds like a cool idea. Just make sure that it appears like you would expect it to on the website documentation.

srikappa-amzn (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/Instance.h
```diff
@@ -80,182 +83,275 @@ namespace AzToolsFramework
...
+            //! Replaces the entity stored under the provided alias with a new one.
+            //! @param entity The new entity.
+            //! @param alias The alias of the entity that will be replaced.
+            //! @return The original entity or a nullptr if not found.
+            AZStd::unique_ptr<AZ::Entity> ReplaceEntity(AZStd::unique_ptr<AZ::Entity>&& entity, EntityAliasView alias);
 
-            void Reset();
+            //! Detaches entities in the instance hierarchy based on the provided filter.
+            //! @param filter A user provided filter used to decide whether an entity should be detached. Returns true to detach.
+            void RemoveEntitiesInHierarchy(const AZStd::function<bool(const AZStd::unique_ptr<AZ::Entity>&)>& filter);
```
michabr, line 155: Should this be called RemoveAllEntitiesInHierarchy for consistency?

michabr (approved):
Thanks! Just had one question/suggestion.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/Instance.h
forkercat: I did not change this one because it does not remove all entities but remove entities based on a filter. I think this would better express what it is doing.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/Instance.h
forkercat: Yeah. I like this idea too. I think it is epecially good for overloaded members.

I checked on documentation website for one existing example. They did end up being grouped together. I put it here for reference.

```cpp
//! Normalizes the vector in-place and returns the previous length.
//! This takes a few more instructions than calling just Normalize().
//! @{
float NormalizeWithLength();
float NormalizeWithLengthEstimate();
//! @}
```

https://www.o3de.org/docs/api/frameworks/azcore/class_a_z_1_1_vector2.html

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/Instance.h
michabr: Oh yeah, I thought it was just a callback. Never mind. :)

PR #12763, opened 2022-10-22
Shape offset part one

greerdv:
## What does this PR do?

## How was this PR tested?

### DIFF
--- a/Code/Framework/AzFramework/AzFramework/UnitTest/TestDebugDisplayRequests.cpp
+++ b/Code/Framework/AzFramework/AzFramework/UnitTest/TestDebugDisplayRequests.cpp
@@ -114,4 +114,21 @@ namespace UnitTest
             m_transforms.pop();
         }
     }
+
+    AZ::Matrix3x4 TestDebugDisplayRequests::PopPremultipliedMatrix()
+    {
+        if (m_transforms.size() == 1)
+        {
+            AZ_Error("TestDebugDisplayRequest", false, "Invalid call to PopPremultipliedMatrix when no matrices were pushed.");
+            return AZ::Matrix3x4::CreateIdentity();
+        }
+        AZ::Matrix3x4 matrix = AZ::Matrix3x4::CreateFromTransform(m_transforms.top());
+        m_transforms.pop();
+        return matrix;
+    }
+
+    void TestDebugDisplayRequests::PushPremultipliedMatrix(const AZ::Matrix3x4& matrix)
+    {
+        m_transforms.push(AZ::Transform::CreateFromMatrix3x4(matrix));
+    }
 } // namespace UnitTest
--- a/Code/Framework/AzFramework/AzFramework/UnitTest/TestDebugDisplayRequests.h
+++ b/Code/Framework/AzFramework/AzFramework/UnitTest/TestDebugDisplayRequests.h
@@ -47,6 +47,8 @@ namespace UnitTest
         void DrawLines(const AZStd::vector<AZ::Vector3>& lines, const AZ::Color& color) override;
         void PushMatrix(const AZ::Transform& tm) override;
         void PopMatrix() override;
+        void PushPremultipliedMatrix(const AZ::Matrix3x4& matrix) override;
+        AZ::Matrix3x4 PopPremultipliedMatrix() override;
     private:
         void DrawPoints(const AZStd::vector<AZ::Vector3>& points);
         AZStd::vector<AZ::Vector3> m_points;
--- a/Gems/LmbrCentral/Code/Source/Shape/BoxShape.cpp
+++ b/Gems/LmbrCentral/Code/Source/Shape/BoxShape.cpp
@@ -267,7 +267,11 @@ namespace LmbrCentral
         }
 
         // transform to world space
-        return m_currentTransform.TransformPoint(AZ::Vector3(x, y, z));
+        AZ::Transform worldTransformWithoutScale = m_currentTransform;
+        const float entityScale = worldTransformWithoutScale.ExtractUniformScale();
+        const AZ::Vector3 scaledTranslationOffset = entityScale * m_currentNonUniformScale * m_boxShapeConfig.m_translationOffset;
+
+        return worldTransformWithoutScale.TransformPoint(AZ::Vector3(x, y, z) + scaledTranslationOffset);
     }
 
     AZ::Vector3 BoxShape::GetTranslationOffset()
--- a/Gems/LmbrCentral/Code/Source/Shape/BoxShape.cpp
+++ b/Gems/LmbrCentral/Code/Source/Shape/BoxShape.cpp
@@ -297,7 +297,7 @@ namespace LmbrCentral
         {
             ShapeComponentNotificationsBus::Event(
                 m_entityId, &ShapeComponentNotificationsBus::Events::OnShapeChanged,
-                ShapeComponentNotifications::ShapeChangeReasons::TransformChanged);
+                ShapeComponentNotifications::ShapeChangeReasons::ShapeChanged);
         }
     }
 
--- a/Gems/LmbrCentral/Code/Source/Shape/BoxShape.cpp
+++ b/Gems/LmbrCentral/Code/Source/Shape/BoxShape.cpp
@@ -288,7 +288,7 @@ namespace LmbrCentral
             if (!m_boxShapeConfig.m_translationOffset.IsClose(translationOffset))
             {
                 m_boxShapeConfig.m_translationOffset = translationOffset;
-                m_intersectionDataCache.InvalidateCache(InvalidateShapeCacheReason::TransformChange);
+                m_intersectionDataCache.InvalidateCache(InvalidateShapeCacheReason::ShapeChange);
                 shapeChanged = true;
             }
         }
@@ -342,8 +342,8 @@ namespace LmbrCentral
         const ShapeDrawParams& shapeDrawParams, const BoxShapeConfig& boxShapeConfig,
         AzFramework::DebugDisplayRequests& debugDisplay, const AZ::Vector3& nonUniformScale)
     {
-        const AZ::Vector3 boxMin = boxShapeConfig.m_dimensions * nonUniformScale * -0.5f;
-        const AZ::Vector3 boxMax = boxShapeConfig.m_dimensions * nonUniformScale * 0.5f;
+        const AZ::Vector3 boxMin = nonUniformScale * (boxShapeConfig.m_dimensions * -0.5f + boxShapeConfig.m_translationOffset);
+        const AZ::Vector3 boxMax = nonUniformScale * (boxShapeConfig.m_dimensions * 0.5f + boxShapeConfig.m_translationOffset);
 
         if (shapeDrawParams.m_filled)
         {
--- a/Gems/LmbrCentral/Code/Source/Shape/BoxShape.cpp
+++ b/Gems/LmbrCentral/Code/Source/Shape/BoxShape.cpp
@@ -142,7 +142,8 @@ namespace LmbrCentral
     void BoxShape::GetTransformAndLocalBounds(AZ::Transform& transform, AZ::Aabb& bounds)
     {
         const AZ::Vector3 extent(m_boxShapeConfig.m_dimensions * m_currentNonUniformScale * 0.5f);
-        bounds = AZ::Aabb::CreateFromMinMax(-extent, extent);
+        const AZ::Vector3 scaledOffset(m_boxShapeConfig.m_translationOffset * m_currentNonUniformScale);
+        bounds = AZ::Aabb::CreateFromMinMax(scaledOffset - extent, scaledOffset + extent);
         transform = m_currentTransform;
     }
 
@@ -212,9 +213,8 @@ namespace LmbrCentral
         AZ::Vector3 boxMin = m_intersectionDataCache.m_scaledDimensions * -0.5f * insideMargin;
         AZ::Vector3 boxMax = m_intersectionDataCache.m_scaledDimensions * 0.5f * insideMargin;
 
-        // As std:normal_distribution requires a std:random_engine to be passed in,
-        // As std:normal_distribution requires a std:random_engine to be passed in, create one using a random seed that is guaranteed to be properly
-        // random each time it is called
+        // As std:normal_distribution requires a std:random_engine to be passed in, create one using a random seed
+        // that is guaranteed to be properly random each time it is called
         time_t seedVal;
         seedVal = AZ::Sfmt::GetInstance().Rand64();
         std::default_random_engine generator;
@@ -270,22 +270,50 @@ namespace LmbrCentral
         return m_currentTransform.TransformPoint(AZ::Vector3(x, y, z));
     }
 
+    AZ::Vector3 BoxShape::GetTranslationOffset()
+    {
+        return m_boxShapeConfig.m_translationOffset;
+    }
+
+    void BoxShape::SetTranslationOffset(const AZ::Vector3& translationOffset)
+    {
+        bool shapeChanged = false;
+
+        {
+            AZStd::unique_lock lock(m_mutex);
+            if (!m_boxShapeConfig.m_translationOffset.IsClose(translationOffset))
+            {
+                m_boxShapeConfig.m_translationOffset = translationOffset;
+                m_intersectionDataCache.InvalidateCache(InvalidateShapeCacheReason::TransformChange);
+                shapeChanged = true;
+            }
+        }
+
+        if (shapeChanged)
+        {
+            ShapeComponentNotificationsBus::Event(
+                m_entityId, &ShapeComponentNotificationsBus::Events::OnShapeChanged,
+                ShapeComponentNotifications::ShapeChangeReasons::TransformChanged);
+        }
+    }
+
     void BoxShape::BoxIntersectionDataCache::UpdateIntersectionParamsImpl(
         const AZ::Transform& currentTransform, const BoxShapeConfig& configuration, const AZ::Vector3& currentNonUniformScale)
     {
         AZ::Transform worldFromLocalNormalized = currentTransform;
         const float entityScale = worldFromLocalNormalized.ExtractUniformScale();
+        const AZ::Vector3 scaledTranslationOffset = entityScale * currentNonUniformScale * configuration.m_translationOffset;
 
-        m_currentPosition = worldFromLocalNormalized.GetTranslation();
+        m_currentPosition = worldFromLocalNormalized.TransformPoint(scaledTranslationOffset);
         m_scaledDimensions = configuration.m_dimensions * currentNonUniformScale * entityScale;
 
         AZ::Quaternion worldFromLocalQuaternion = worldFromLocalNormalized.GetRotation();
         if (worldFromLocalQuaternion.IsClose(AZ::Quaternion::CreateIdentity()))
         {
-            AZ::Vector3 boxMin = m_scaledDimensions * -0.5f;
+            AZ::Vector3 boxMin = m_scaledDimensions * -0.5f + scaledTranslationOffset;
             boxMin = worldFromLocalNormalized.TransformPoint(boxMin);
 
-            AZ::Vector3 boxMax = m_scaledDimensions * 0.5f;
+            AZ::Vector3 boxMax = m_scaledDimensions * 0.5f + scaledTranslationOffset;
             boxMax = worldFromLocalNormalized.TransformPoint(boxMax);
 
             m_aabb = AZ::Aabb::CreateFromMinMax(boxMin, boxMax);
--- a/Gems/LmbrCentral/Code/Source/Shape/BoxShape.cpp
+++ b/Gems/LmbrCentral/Code/Source/Shape/BoxShape.cpp
@@ -281,8 +281,12 @@ namespace LmbrCentral
 
     void BoxShape::SetTranslationOffset(const AZ::Vector3& translationOffset)
     {
-        bool shapeChanged = false;
+        if (!IsShapeComponentTranslationEnabled())
+        {
+            return;
+        }
 
+        bool shapeChanged = false;
         {
             AZStd::unique_lock lock(m_mutex);
             if (!m_boxShapeConfig.m_translationOffset.IsClose(translationOffset))
--- a/Gems/LmbrCentral/Code/Source/Shape/EditorAxisAlignedBoxShapeComponent.cpp
+++ b/Gems/LmbrCentral/Code/Source/Shape/EditorAxisAlignedBoxShapeComponent.cpp
@@ -171,5 +171,13 @@ namespace LmbrCentral
     {
         return AZ::Vector3(m_aaboxShape.GetCurrentTransform().GetUniformScale());
     }
+
+    AZ::Aabb EditorAxisAlignedBoxShapeComponent::GetLocalBounds()
+    {
+        AZ::Transform transform = AZ::Transform::CreateIdentity();
+        AZ::Aabb aabb = AZ::Aabb::CreateNull();
+        m_aaboxShape.GetTransformAndLocalBounds(transform, aabb);
+        return aabb.GetTransformedAabb(AZ::Transform::CreateFromQuaternion(GetWorldTM().GetRotation().GetInverseFast()));
+    }
 } // namespace LmbrCentral
 
--- a/Gems/LmbrCentral/Code/Source/Shape/EditorAxisAlignedBoxShapeComponent.cpp
+++ b/Gems/LmbrCentral/Code/Source/Shape/EditorAxisAlignedBoxShapeComponent.cpp
@@ -172,3 +172,4 @@ namespace LmbrCentral
         return AZ::Vector3(m_aaboxShape.GetCurrentTransform().GetUniformScale());
     }
 } // namespace LmbrCentral
+
--- a/Gems/LmbrCentral/Code/Tests/AxisAlignedBoxShapeTest.cpp
+++ b/Gems/LmbrCentral/Code/Tests/AxisAlignedBoxShapeTest.cpp
@@ -6,8 +6,6 @@
  *
  */
 
-#include <AzTest/AzTest.h>
-
 #include <AZTestShared/Math/MathTestHelpers.h>
 #include <AzCore/Component/ComponentApplication.h>
 #include <AzCore/Math/Matrix3x3.h>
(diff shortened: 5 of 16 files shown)


Code/Framework/AzFramework/AzFramework/UnitTest/TestDebugDisplayRequests.cpp
```diff
@@ -98,6 +98,27 @@ namespace UnitTest
...
+            const float theta = (static_cast<float>(sliceIndex) + 0.5f) / static_cast<float>(numSlices) * AZ::Constants::Pi;
+            float sinTheta;
+            float cosTheta;
+            AZ::SinCos(theta, sinTheta, cosTheta);
+            for (int angularIndex = 0; angularIndex < numAngularDivisions; angularIndex++)
+            {
+                const float phi = static_cast<float>(angularIndex) / static_cast<float>(numAngularDivisions) * AZ::Constants::TwoPi;
+                points.push_back(pos + radius * AZ::Vector3(sinTheta * AZ::Sin(phi), sinTheta * AZ::Cos(phi), cosTheta));
+            }
+        }
```
mbalfour-amzn, line 118: Suggestion: You could potentially also use AZ::Geometry3DUtils::GenerateIcoSphere() and scale/translate it. Totally reasonable to leave this code as-is, just mentioning it since it's a relatively new method and I'm trying to raise awareness of it. :)

Code/Framework/AzFramework/AzFramework/UnitTest/TestDebugDisplayRequests.h
```diff
@@ -45,8 +45,12 @@ namespace UnitTest
         void DrawLine(const AZ::Vector3& p1, const AZ::Vector3& p2) override;
         void DrawLine(const AZ::Vector3& p1, const AZ::Vector3& p2, const AZ::Vector4& col1, const AZ::Vector4& col2) override;
         void DrawLines(const AZStd::vector<AZ::Vector3>& lines, const AZ::Color& color) override;
+        void DrawWireSphere(const AZ::Vector3& pos, float radius) override;
+        void DrawWireSphere([[maybe_unused]] const AZ::Vector3& pos, [[maybe_unused]] const AZ::Vector3 radius) override {};
```
mbalfour-amzn, line 49: Nit: No semicolon needed at the end.

Gems/LmbrCentral/Code/Source/Shape/BoxShape.cpp
```diff
@@ -267,25 +267,61 @@ namespace LmbrCentral
...
+                m_intersectionDataCache.InvalidateCache(InvalidateShapeCacheReason::ShapeChange);
+                shapeChanged = true;
+            }
+        }
+
+        if (shapeChanged)
+        {
+            ShapeComponentNotificationsBus::Event(
+                m_entityId, &ShapeComponentNotificationsBus::Events::OnShapeChanged,
+                ShapeComponentNotifications::ShapeChangeReasons::ShapeChanged);
```
mbalfour-amzn, line 304: Crazy semantics question here - should this send "ShapeChanged" or "TransformChanged", since technically it's only the position of the shape that changed?

Gems/LmbrCentral/Code/Source/Shape/BoxShape.cpp
```diff
@@ -267,25 +267,61 @@ namespace LmbrCentral
...
+        }
     }
 
     void BoxShape::BoxIntersectionDataCache::UpdateIntersectionParamsImpl(
         const AZ::Transform& currentTransform, const BoxShapeConfig& configuration, const AZ::Vector3& currentNonUniformScale)
     {
         AZ::Transform worldFromLocalNormalized = currentTransform;
```
mbalfour-amzn, line 311: Why does this transform get copied instead of using currentTransform directly? Unless I'm missing something it doesn't seem to be changing any of the values.

mbalfour-amzn (approved):

Gems/LmbrCentral/Code/Source/Shape/BoxShape.cpp
greerdv: Yeah I debated that for a while. But I was thinking of the offset as part of the shape definition.

Gems/LmbrCentral/Code/Source/Shape/BoxShape.cpp
greerdv: I think ExtractUniformScale on the next line will modify it.

Code/Framework/AzFramework/AzFramework/UnitTest/TestDebugDisplayRequests.cpp
greerdv: Yeah, that's a good call and will remove a few lines of unnecessary code here, will update.

Gems/LmbrCentral/Code/Source/Shape/BoxShape.cpp
mbalfour-amzn: i can totally buy that. Just wanted to ask the question. :)

Gems/LmbrCentral/Code/Source/Shape/BoxShape.cpp
mbalfour-amzn: Eeek, yep, you're right, I did NOT read "Extract" as a destructive action. That's a method name that could use a bit more verbosity. :(

Code/Framework/AzFramework/AzFramework/UnitTest/TestDebugDisplayRequests.cpp
```diff
@@ -114,4 +135,21 @@ namespace UnitTest
             m_transforms.pop();
         }
     }
+
+    AZ::Matrix3x4 TestDebugDisplayRequests::PopPremultipliedMatrix()
```
hultonha, line 139: question: Is it worth updating `PopMatrix` to also return a `Matrix3x4`? Also it would be great to have a comment (there may be already and I haven't got to it yet) in the API for 'PreMultipliedPush/Pop' to make it clear it's not a separate stack and exactly what it does

Gems/LmbrCentral/Code/Source/Shape/BoxShape.cpp
```diff
@@ -267,25 +267,61 @@ namespace LmbrCentral
...
         // transform to world space
-        return m_currentTransform.TransformPoint(AZ::Vector3(x, y, z));
+        AZ::Transform worldTransformWithoutScale = m_currentTransform;
+        const float entityScale = worldTransformWithoutScale.ExtractUniformScale();
+        const AZ::Vector3 scaledTranslationOffset = entityScale * m_currentNonUniformScale * m_boxShapeConfig.m_translationOffset;
+
+        return worldTransformWithoutScale.TransformPoint(AZ::Vector3(x, y, z) + scaledTranslationOffset);
+    }
+
+    AZ::Vector3 BoxShape::GetTranslationOffset()
```
hultonha, line 277: update: This should be `const` ideally

Gems/LmbrCentral/Code/Source/Shape/BoxShape.cpp
hultonha: I'd personally say this is shape data that's changing even if it's in a new place which is why we should prefer `OnShapeChanged` (the entity transform hasn't changed)

Gems/LmbrCentral/Code/Source/Shape/BoxShape.cpp
hultonha: Because of the call to `worldFromLocalNormalized.ExtractUniformScale()` below which mutates `worldFromLocalNormalized`

Gems/LmbrCentral/Code/Source/Shape/CapsuleShape.cpp
```diff
@@ -174,7 +175,7 @@ namespace LmbrCentral
         // If its not in either sphere check the cylinder
         return AZ::Intersect::PointCylinder(
             m_intersectionDataCache.m_basePlaneCenterPoint, m_intersectionDataCache.m_axisVector,
-            powf(m_intersectionDataCache.m_internalHeight, 2.0f), radiusSquared, point);
+            m_intersectionDataCache.m_internalHeight * m_intersectionDataCache.m_internalHeight, radiusSquared, point);
```
hultonha, line 178: Good spot and nice change (this is likely something stupid I did in the past)

Gems/LmbrCentral/Code/Source/Shape/EditorAxisAlignedBoxShapeComponent.cpp
```diff
@@ -171,4 +171,13 @@ namespace LmbrCentral
     {
         return AZ::Vector3(m_aaboxShape.GetCurrentTransform().GetUniformScale());
     }
+
+    AZ::Aabb EditorAxisAlignedBoxShapeComponent::GetLocalBounds()
```
hultonha, line 175: question: Should be implement `GetWorldBounds` too? (or is there a default implementation implemented in terms of this?)

Gems/LmbrCentral/Code/Source/Shape/EditorCapsuleShapeComponent.cpp
```diff
@@ -84,10 +84,18 @@ namespace LmbrCentral
...
+            [this]()
+            {
+                return CanDraw();
+            },
             [this](AzFramework::DebugDisplayRequests& debugDisplay)
             {
-                DrawShape(debugDisplay, { m_shapeColor, m_shapeWireColor, m_displayFilled }, m_capsuleShapeMesh);
+                DrawShape(
+                    debugDisplay,
+                    { m_shapeColor, m_shapeWireColor, m_displayFilled },
```
hultonha, line 96: nit: Could we add the type before the `{...}` object created here as I've found searching for instances of it is made much easier when the type is included (this again is a fault of mine from the past and I've since changed my ways 😅)

Gems/LmbrCentral/Code/Tests/AxisAlignedBoxShapeTest.cpp
```diff
@@ -6,21 +6,23 @@
...
 #include <Shape/AxisAlignedBoxShapeComponent.h>
+#include <ShapeTestUtils.h>
 #include <ShapeThreadsafeTest.h>
 
 namespace UnitTest
 {
-    class AxisAlignedBoxShapeTest : public AllocatorsFixture
+    class AxisAlignedBoxShapeTest
+        : public AllocatorsFixture
+        , public ShapeOffsetFixture
```
hultonha, line 25: question: Is inheriting from two fixtures a good idea? 🤔 Or is one not really a fixture, just checking there's no off multiple inheritance going on here 😝

Gems/LmbrCentral/Code/Tests/AxisAlignedBoxShapeTest.cpp
```diff
@@ -194,4 +198,150 @@ namespace UnitTest
...
+        const AZ::Vector3 translationOffset(4.0f, 3.0f, -1.0f);
+        LmbrCentral::ShapeComponentRequestsBus::Event(
+            entity.GetId(), &LmbrCentral::ShapeComponentRequestsBus::Events::SetTranslationOffset, translationOffset);
+
+        AZ::Aabb aabb = AZ::Aabb::CreateNull();
+        LmbrCentral::ShapeComponentRequestsBus::EventResult(
+            aabb, entity.GetId(), &LmbrCentral::ShapeComponentRequests::GetEncompassingAabb);
+
+        EXPECT_THAT(aabb.GetMin(), IsClose(AZ::Vector3(1.75f, 7.0f, -8.0f)));
+        EXPECT_THAT(aabb.GetMax(), IsClose(AZ::Vector3(6.25f, 10.0f, 1.0f)));
```
hultonha, line 240: suggestion: There's actually an IsClose for Aabbs too which you could use instead/in addition if you wanted (repeated in a few places)

Gems/LmbrCentral/Code/Tests/AxisAlignedBoxShapeTest.cpp
```diff
@@ -194,4 +198,150 @@ namespace UnitTest
...
+        // test some pairs of nearby points which should be just either side of the surface of the box
+        EXPECT_TRUE(IsPointInside(entity, AZ::Vector3(-4.9f, -15.0f, -20.0f)));
+        EXPECT_FALSE(IsPointInside(entity, AZ::Vector3(-5.1f, -15.0f, -20.0f)));
+        EXPECT_TRUE(IsPointInside(entity, AZ::Vector3(0.0f, -11.1f, -25.0f)));
+        EXPECT_FALSE(IsPointInside(entity, AZ::Vector3(0.0f, -10.9f, -25.0f)));
+        EXPECT_TRUE(IsPointInside(entity, AZ::Vector3(1.9f, -17.9f, -35.7f)));
+        EXPECT_FALSE(IsPointInside(entity, AZ::Vector3(2.1f, -18.1f, -35.8f)));
+    }
+
+    TEST_F(AxisAlignedBoxShapeTest, DistanceFromPointWithTranslationRotationScaleAndTranslationOffset)
```
hultonha, line 287: These tests are excellent @greerdv, really awesome work 🙂

Gems/LmbrCentral/Code/Tests/BoxShapeTest.cpp
```diff
@@ -6,23 +6,24 @@
...
+#include <AzTest/AzTest.h>
+#include <Shape/BoxShapeComponent.h>
+#include <ShapeTestUtils.h>
 #include <ShapeThreadsafeTest.h>
 
 namespace UnitTest
 {
     class BoxShapeTest
         : public AllocatorsFixture
+        , public ShapeOffsetFixture
```
hultonha, line 26: question: Same question about Fixture (maybe it should just be renamed?)

Gems/LmbrCentral/Code/Tests/BoxShapeTest.cpp
```diff
@@ -682,4 +725,263 @@ namespace UnitTest
...
+        // test some pairs of nearby points which should be just either side of the surface of the box
+        EXPECT_TRUE(IsPointInside(entity, AZ::Vector3(-0.15f, 10.4f, 4.66f)));
+        EXPECT_FALSE(IsPointInside(entity, AZ::Vector3(-0.15f, 10.42f, 4.66f)));
+        EXPECT_TRUE(IsPointInside(entity, AZ::Vector3(-0.17f, 8.13f, 4.49f)));
+        EXPECT_FALSE(IsPointInside(entity, AZ::Vector3(-0.17f, 8.13f, 4.47f)));
+        EXPECT_TRUE(IsPointInside(entity, AZ::Vector3(-6.34f, 5.58f, 5.47f)));
+        EXPECT_FALSE(IsPointInside(entity, AZ::Vector3(-6.36f, 5.58f, 5.47f)));
+    }
+
+    TEST_F(BoxShapeTest, IsPointInsideUnrotatedWithTranslationOFfset)
```
hultonha, line 880: uber nit: OFfset -> Offset

Gems/LmbrCentral/Code/Tests/ShapeTestUtils.h
```diff
@@ -0,0 +1,34 @@
...
+
+namespace AZ
+{
+    class Entity;
+    class Vector3;
+} // namespace AZ
+
+namespace UnitTest
+{
+    class ShapeOffsetFixture
```
hultonha, line 22: suggestion: Might be better to not call this Fixture and instead have it as `ShapeOffsetTestsBase` or something like that, just to avoid any confusion with the other main Fixture that gets inherited from (feel free to think of a better name 😅)

Gems/LmbrCentral/Code/include/LmbrCentral/Shape/ShapeComponentBus.h
```diff
@@ -31,6 +31,23 @@ namespace LmbrCentral
...
+    /// Helper function for checking whether feature flag for in progress shape component translation offsets is enabled.
+    /// See https://github.com/o3de/sig-simulation/issues/26 for more details.
+    inline bool IsShapeComponentTranslationEnabled()
+    {
+        bool isShapeComponentTranslationEnabled = false;
+
+        if (auto* registry = AZ::SettingsRegistry::Get())
+        {
+            registry->Get(isShapeComponentTranslationEnabled, ShapeComponentTranslationOffsetEnabled);
+        }
```
hultonha, line 46: suggestion: I think you can just return here and not use an out param (see ViewportSettings.h/cpp for some examples)

Gems/LmbrCentral/Code/include/LmbrCentral/Shape/ShapeComponentBus.h
```diff
@@ -174,6 +191,19 @@ namespace LmbrCentral
             return false;
         }
 
+        /// Get the translation offset for the shape relative to its entity.
+        virtual AZ::Vector3 GetTranslationOffset()
```
hultonha, line 195: nit: Would be great to add `const` to this (I realize it's a Bus interface but still helps in this case I think and is worth adding - in an emergency people can always use `mutable` 😝)

hultonha (approved):
Change is looking fantastic @greerdv. The thoroughness of the tests is first rate, super impressive to see. Really awesome work! 😄

Gems/LmbrCentral/Code/Source/Shape/EditorAxisAlignedBoxShapeComponent.cpp
greerdv: EditorBaseShapeComponent (which this component inherits from) already implements both GetLocalBounds and GetWorldBounds. Because of the unique way the AA box works, the base implementation of GetLocalBounds does not work for it, but the base implementation of GetWorldBounds does.

hultonha: Aaah I see perfect, makes sense thanks 👍

Code/Framework/AzFramework/AzFramework/UnitTest/TestDebugDisplayRequests.cpp
```diff
@@ -98,6 +98,27 @@ namespace UnitTest
...
 
+    void TestDebugDisplayRequests::DrawWireSphere(const AZ::Vector3& pos, float radius)
+    {
+        AZStd::vector<AZ::Vector3> points;
+        const int numSlices = 16;
+        const int numAngularDivisions = 16;
+        points.reserve(numSlices * numAngularDivisions);
+        for (int sliceIndex = 0; sliceIndex < numSlices; sliceIndex++)
+        {
+            const float theta = (static_cast<float>(sliceIndex) + 0.5f) / static_cast<float>(numSlices) * AZ::Constants::Pi;
```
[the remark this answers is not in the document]
greerdv: This code is gone now due to mbalfour-amzn's suggestion.

Code/Framework/AzFramework/AzFramework/UnitTest/TestDebugDisplayRequests.cpp
greerdv: This is just implementing the interface in AzFramework::DebugDisplayRequests. I'd say changes to that interface are out of scope here.

Gems/LmbrCentral/Code/Source/Shape/BoxShape.cpp
greerdv: Updated.

Gems/LmbrCentral/Code/Tests/BoxShapeTest.cpp
greerdv: Renamed to ShapeOffsetTestsBase as suggested in your other comment.

Gems/LmbrCentral/Code/Tests/AxisAlignedBoxShapeTest.cpp
greerdv: Renamed as suggested in your other comment.

Gems/LmbrCentral/Code/Tests/AxisAlignedBoxShapeTest.cpp
greerdv: Tbh I think this is easier to read.

Gems/LmbrCentral/Code/Tests/BoxShapeTest.cpp
greerdv: Nice spot, fixed!

Gems/LmbrCentral/Code/Tests/ShapeTestUtils.h
greerdv: Renamed.

Gems/LmbrCentral/Code/include/LmbrCentral/Shape/ShapeComponentBus.h
greerdv: I think there's a function in ViewportSettings.h to support that, which I'd need to copy or move to the registry settings to make that work. I also don't really see the problem with the current version.

Gems/LmbrCentral/Code/include/LmbrCentral/Shape/ShapeComponentBus.h
greerdv: Done.

Gems/LmbrCentral/Code/Source/Shape/CapsuleShape.cpp
```diff
@@ -213,31 +215,63 @@ namespace LmbrCentral
         return intersection;
     }
 
+    AZ::Vector3 CapsuleShape::GetTranslationOffset() const
+    {
+        return m_capsuleShapeConfig.m_translationOffset;
+    }
```
moraaar, line 221: The same way that `SetTranslationOffset` is protected to do nothing when shape component translation is not enabled, `GetTranslationOffset` should do the same and return offset 0 if it's disabled.

As it is now since `m_capsuleShapeConfig.m_translationOffset` is serialized then it still returns the offset value it was saved last while translation was enabled.

But I guess that's still dangerous having `m_capsuleShapeConfig.m_translationOffset` with a value and `GetTranslationOffset` returning 0. I wonder if there is a better way during reflection to ignore `m_capsuleShapeConfig.m_translationOffset` and set it to 0 if `IsShapeComponentTranslationEnabled` is false. At the moment it's only setting visibility to false, but the variable will still be there with the last saved value.

For example, `UpdateIntersectionParamsImpl` uses `m_translationOffset` directly, doesn't check if translation is enabled or not, so even if translation is disabled it can still apply an offset if it read a value from serialization.

Do we need a runtime check for `IsShapeComponentTranslationEnabled` at all? Why do we need to disable it? It's making this more complicated to manage.

Gems/LmbrCentral/Code/Tests/ShapeTestUtils.cpp
```diff
@@ -0,0 +1,46 @@
...
+        {
+            AZ::SettingsRegistry::Register(m_oldSettingsRegistry);
+            m_oldSettingsRegistry = nullptr;
+        }
+        m_settingsRegistry.reset();
+    }
+
+    bool IsPointInside(const AZ::Entity& entity, const AZ::Vector3& point)
+    {
+        bool inside;
```
moraaar, line 41: In case the bus doesn't set any value, please initialize `inside` to `false`.

Gems/LmbrCentral/Code/include/LmbrCentral/Shape/ShapeComponentBus.h
```diff
@@ -174,6 +191,19 @@ namespace LmbrCentral
...
+        {
+            AZ_Warning("ShapeComponentRequests", false, "GetTranslationOffset not implemented");
+            return AZ::Vector3::CreateZero();
+        }
+
+        /// Set the translation offset for the shape relative to its entity.
+        virtual void SetTranslationOffset([[maybe_unused]] const AZ::Vector3& translationOffset)
+        {
+            AZ_Warning("ShapeComponentRequests", false, "SetTranslationOffset not implemented");
+        }
```
moraaar, line 205: `GetTranslationOffset` and `SetTranslationOffset` shouldn't through a warning if translation is disabled.

moraaar (commented):
Great work @greerdv! The change looks very good and thanks for adding all those unit tests.

Before I give my approval I have a question related to `m_translationOffset` still having values while `IsShapeComponentTranslationEnabled` is false, which it seems it can lead to issues. More details in a comment below. I'll wait for your comment before approving.

Thanks

Code/Framework/AzFramework/AzFramework/UnitTest/TestDebugDisplayRequests.cpp
```diff
@@ -114,4 +135,21 @@ namespace UnitTest
             m_transforms.pop();
         }
     }
+
+    AZ::Matrix3x4 TestDebugDisplayRequests::PopPremultipliedMatrix()
+    {
+        if (m_transforms.size() == 1)
```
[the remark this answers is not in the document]
greerdv: I think that would be fine, it would just cause the test to fail (which you would want because it would indicate a problem in your debug draw code).

Code/Framework/AzFramework/AzFramework/UnitTest/TestDebugDisplayRequests.cpp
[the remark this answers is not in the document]
hultonha: Awesome thanks!

Code/Framework/AzFramework/AzFramework/UnitTest/TestDebugDisplayRequests.cpp
hultonha: Ah I see understood, makes sense thanks

Gems/LmbrCentral/Code/Tests/BoxShapeTest.cpp
hultonha: Perfect thank you!

Gems/LmbrCentral/Code/include/LmbrCentral/Shape/ShapeComponentBus.h
hultonha: Sure no worries just a random suggestion to streamline the code but this is fine as-is

Gems/LmbrCentral/Code/Source/Shape/CapsuleShape.cpp
hultonha: Thinking about it too it might be useful to do the check...

```
if (offsetSupported) {
    return translationOffset;
}
return zero;
```

But I'm not sure how much overhead there is to that (both from runtime and code complexity) - ultimately your call @greerdv but a good callout from @moraaar as something to consider

moraaar: >  the place where the setting is defined makes it clear that it is work in progress

That's fair enough. If that check is ultimately going to be removed once the work is completely then it's not big worry.

moraaar (approved):

Gems/LmbrCentral/Code/Source/Shape/CapsuleShape.cpp
greerdv: Yeah, this should only be for a few weeks while this work is going on.

Gems/LmbrCentral/Code/Tests/ShapeTestUtils.cpp
greerdv: I've got a passing AR now, so I'm going to submit this PR and make a follow up PR for this suggestion and the one below if that's ok.

greerdv: https://github.com/o3de/o3de/pull/12983

PR #11381, opened 2022-08-18
Fix a Vulkan validation error that was spamming every frame

moudgils:
Signed-off-by: moudgils <47460854+moudgils@users.noreply.github.com>

## What does this PR do?

It fixes the following vulkan validation that is spamming each frame 
[Error] (vkDebugMessage) - [ERROR][Validation] Validation Error: [ VUID-vkCmdPipelineBarrier-srcAccessMask-02815 ] Object 0: handle = 0x20fa61bc598, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x24d88c2b | vkCmdPipelineBarrier(): .pImageMemoryBarriers[0].srcAccessMask bit VK_ACCESS_TRANSFER_WRITE_BIT is not supported by stage mask (VK_PIPELINE_STAGE_VERTEX_SHADER_BIT|VK_PIPELINE_STAGE_TESSELLATION_CONTROL_SHADER_BIT|VK_PIPELINE_STAGE_TESSELLATION_EVALUATION_SHADER_BIT|VK_PIPELINE_STAGE_GEOMETRY_SHADER_BIT|VK_PIPELINE_STAGE_FRAGMENT_SHADER_BIT|VK_PIPELINE_STAGE_COMPUTE_SHADER_BIT). The Vulkan spec states: The srcAccessMask member of each element of pMemoryBarriers must only include access flags that are supported by one or more of the pipeline stages in srcStageMask, as specified in the table of supported access types (https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VUID-vkCmdPipelineBarrier-srcAccessMask-02815)

The main issue was that we werent filtering out the source mask based on the supported shader stage.

## How was this PR tested?

Ran editor with validation enabled and ran full test suite on dx12 and Vk on pc

Issue #11225, opened 2022-08-11
Lighting is broken on Linux with and AMD gpu


rgba16f:
**Describe the bug**
A large block of hte screen doesn't receive direct lighting

**Assets required**
AutomatedTesting/Levels/AtomFeatureIntegrationBenchmark

**Steps to reproduce**
1. On Linux open the editor
2. Load AtomFeatureIntegrationBenchmark level
3. See error

**Expected behavior**
Level lighting matches Windows DX12

**Actual behavior**
The lower portion of the screen has incorrect lighting, the portion changes with camera orientation and window size.

**Screenshots/Video**
![Screenshot from 2022-08-11 15-58-38](https://user-images.githubusercontent.com/82187279/184217822-389205b8-0858-4cad-bb20-83d01d8895f6.png)

**Found in Branch**
o3de development@2aac761cc65ad86a9e77ca54435ad5d7f1555a2f

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Windows 10
 - Version 20H2 19042.1826
 - CPU Intel(R) Core(TM) i7-6850K
 - GPU AMD Radeon RX Vega 64
 - Memory 32GB

pollend:
@rgba16f can you try this test with this branch: https://github.com/o3de/o3de/pull/10876

rgba16f:
> for the world if you attach the SSAO shader and disable it does the black boxes disappear?
> 
> seems similar to this issue: #4379
> 
> @rgba16f can you try this test with this PR: #10876

Huzzah @pollend, I can confirm that your PR does fix the lighting issue I'm seeing.

pollend:
@rgba16f I'm seeing the same issue again can you re-verify this with development?

https://github.com/o3de/o3de/pull/11381

rgba16f:
@pollend I'm seeing this issue on AMD as well

pollend:
We need to tweak the barriers on that call. Pipeline barrier is getting ignored when we try to set it

pollend:
@rgba16f did you verify this with the recent change that got in: https://github.com/o3de/o3de/pull/12764

rgba16f:
> @rgba16f did you verify this with the recent change that got in: #12764

I didn't

gralco:
I built 2210 on Fedora 37 and have this issue (I did not experience this on Fedora 36). All textures are shown as black. This does not seem to be a lighting issue.

I will try building and testing the head of development (92af95a) to see if the problem persists

[...]

PR #10876, opened 2022-07-26
bugfix(Vulkan): set access bit on image barrier

pollend:
Signed-off-by: Michael Pollind <mpollind@gmail.com>

documentation: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkAccessFlagBits.html
```
VK_ACCESS_TRANSFER_WRITE_BIT specifies write 
access to an image or buffer in a [clear](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#clears) or [copy](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#copies) operation. Such access occurs in the VK_PIPELINE_STAGE_2_ALL_TRANSFER_BIT pipeline stage.
```

VK_ACCESS_TRANSFER_WRITE_BIT specifies a write access on an image this is placed on the barrier proceeding the  vkcmdclearimage. This appears to be reletaed to AMD where the clear and read seem to be happening at the same time. this produces the square artifacts on the image. (https://github.com/o3de/o3de/issues/4379)

## How was this PR tested?

ref: https://gitlab.freedesktop.org/mesa/mesa/-/issues/6702#note_1481580
ref: https://github.com/o3de/o3de/issues/4379
ref: https://github.com/o3de/o3de/pull/5034

A
B
C
D
E
Verified with AtomSampleViewer full test suit. Some tests are broken at the moment due to unrelated code changes. Verified with both with and without the changes.

PR #12764, opened 2022-10-22
bugfix(RHI): set VK_PIPELINE_STAGE_TRANSFER_BIT when pipeline has an explicit clear

pollend:
ref: https://github.com/o3de/o3de/issues/11225

## What does this PR do?

The last attempt at this introduced a breaking change where this would fail with the vulkan validation layer: https://github.com/o3de/o3de/pull/11381

after reviewing the specification it seems just enough to add the stage to mark a transfer. should be able to then set the access for a write transfer. 

## How was this PR tested?

need to repeat the steps here: https://github.com/o3de/o3de/pull/10876

### DIFF
--- a/Gems/Atom/RHI/Vulkan/Code/Source/RHI/FrameGraphCompiler.h
+++ b/Gems/Atom/RHI/Vulkan/Code/Source/RHI/FrameGraphCompiler.h
@@ -103,8 +103,8 @@ namespace AZ
             // Add VK_ACCESS_TRANSFER_WRITE_BIT in case we want to do a clear operation.
             if (HasExplicitClear(scopeAttachment, scopeAttachment.GetDescriptor()))
             {
-                srcAccessFlags |= VK_ACCESS_TRANSFER_WRITE_BIT;
-                srcAccessFlags = RHI::FilterBits(srcAccessFlags, GetSupportedAccessFlags(srcPipelineStageFlags));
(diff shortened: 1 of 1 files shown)


Gems/Atom/RHI/Vulkan/Code/Source/RHI/FrameGraphCompiler.h
```diff
@@ -103,8 +103,8 @@ namespace AZ
             // Add VK_ACCESS_TRANSFER_WRITE_BIT in case we want to do a clear operation.
             if (HasExplicitClear(scopeAttachment, scopeAttachment.GetDescriptor()))
             {
-                srcAccessFlags |= VK_ACCESS_TRANSFER_WRITE_BIT;
-                srcAccessFlags = RHI::FilterBits(srcAccessFlags, GetSupportedAccessFlags(srcPipelineStageFlags));
+                srcPipelineStageFlags |= VK_PIPELINE_STAGE_TRANSFER_BIT;
+                srcAccessFlags = RHI::FilterBits(srcAccessFlags | VK_ACCESS_TRANSFER_WRITE_BIT, GetSupportedAccessFlags(srcPipelineStageFlags));
```
rgba16f, line 107: Should it be the same bit added on both lines?

moudgils (approved):

Issue #12365, opened 2022-10-03
UI Editor: Elements zoom incorrectly when Editor's window is out of focus


LB-KrzysztofPinkowski:
**Describe the bug**
When the UI Editor's window is out of focus, scrolling it does not zoom UI elements. Focusing the Editor afterwards, updates the size and position of the elements.

**Steps to reproduce**
1. Open UI Editor.
2. Add element with a component to the canvas (e.g. Image or text).
3. Click on the desktop/other window.
4. With the Editor out of focus, scroll the UI Editor's window.

**Expected behavior**
UI elements zoom correctly.

**Actual behavior**
UI elements size and position is not updated.

**Screenshots/Video**
https://user-images.githubusercontent.com/110379578/193585143-17469a6f-5ed7-4668-8aed-f6bcc0df9f76.mp4

**Found in Branch**
Stabilization/2210 (38005c3)
Development (780d8fc)

**Desktop/Device:**
 - Device: PC
 - OS: Windows
 - Version 10
 - CPU AMD Ryzen 5 5600H
 - GPU NVIDIA GeForce RTX 3070 Laptop GPU
 - Memory 16GB

pollend:
when the window loses focus the OnTick event does not occur so canvas scene graph does not re-evaluate. CryEdit.cpp:2353.

in file CryEdit.cpp
```
GetIEditor()->GetGameEngine()->Update()
```

for now I guess it should be simple to block the scroll event if the window loses focus.

pollend:
One other thing to note is there is this this slight difference between the canvas zooming the and checkered background. the canvas seems to update a bit later but the data is used directly from the ViewportInteraction?

ParasiteSt:
https://github.com/o3de/o3de/pull/12765
This PR has just been merged, I would check again on the newest dev branch to make sure its still an issue.

PR #12765, opened 2022-10-23
bugfix(UI Editor): prevent zoom when canvas window loses focus

pollend:

## What does this PR do?

OnTick is paused when the window loses focus so the viewport does not correctly handle the change. 

this is the call that would need to happen: `UpdateCanvasInEditorViewport`

## How was this PR tested?

follow steps listed from issue: https://github.com/o3de/o3de/issues/12365

### DIFF
--- a/Gems/LyShine/Code/Editor/ViewportInteraction.cpp
+++ b/Gems/LyShine/Code/Editor/ViewportInteraction.cpp
@@ -985,6 +985,13 @@ void ViewportInteraction::SetCanvasZoomPercent(float percent)
 
 void ViewportInteraction::SetCanvasToViewportScale(float newScale, Vec2i* optionalPivotPoint)
 {
+    // only allow setting the viewport scale if current window is active
+    // OnTick for ViewportWidget is need to re-evaluate the layout but does not update if window loses focus
+    if (!m_editorWindow->isActiveWindow())
+    {
+        return;
+    }
+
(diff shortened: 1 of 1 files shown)


Gems/LyShine/Code/Editor/ViewportInteraction.cpp
```diff
@@ -985,6 +985,13 @@ void ViewportInteraction::SetCanvasZoomPercent(float percent)
 
 void ViewportInteraction::SetCanvasToViewportScale(float newScale, Vec2i* optionalPivotPoint)
 {
+    // only allow setting the viewport scale if current window is active
+    // OnTick for ViewportWidget is need to re-evaluate the layout but does not update if window loses focus
```
hultonha, line 989: nit:  `OnTick for ViewportWidget is needed to reevaluate the layout...`

hultonha (approved):
Change looks solid but I'm not super familiar with this code, if there's a video of the 'after' fix that would definitely help. Thanks!

pollend:
> Change looks solid but I'm not super familiar with this code, if there's a video of the 'after' fix that would definitely help. Thanks!

not much of a video, just unfocusing the window and scrolling. nothing should happen since the scroll would be discarded.

gadams3 (approved):

cgalvan (approved):

amzn-tmryan (approved):

PR #12773, opened 2022-10-24
Asset Processor - Added asset count column for folders in Asset tab view

amzn-mike:
## What does this PR do?

Added a column to the Assets tab which shows a recursive count of all contained assets.
Removed the SourceAssetTreeFilterModel class as it was added as a special case to handle sorting a numeric column.  I replaced this with QCollator which can handle sorting strings with a numeric sort, so we don't need the special case filter anymore.

## How was this PR tested?

Manually tested Assets tab

image

Sorting:
image


AMZN-stankowi (approved):
Looks good to me, but this will probably need UX eyes on it, too

Code/Tools/AssetProcessor/native/ui/MainWindow.cpp
```diff
@@ -386,7 +386,7 @@ void MainWindow::Activate()
         this, writeJobFilterSettings);
 
     // Asset view
-    m_sourceAssetTreeFilterModel = new AssetProcessor::SourceAssetTreeFilterModel(this);
+    m_sourceAssetTreeFilterModel = new AssetProcessor::AssetTreeFilterModel(this);
```
jackalbe, line 389: if the class name is being renamed (SourceAssetTreeFilterModel -> AssetTreeFilterModel ) why not the data member "m_sourceAssetTreeFilterModel" ?

jackalbe (approved):

Code/Tools/AssetProcessor/native/ui/MainWindow.cpp
amzn-mike: There's actually 3 of these, the other 2 are below: 
`m_intermediateAssetTreeFilterModel = new AssetProcessor::AssetTreeFilterModel(this);` and `m_productAssetTreeFilterModel = new AssetProcessor::AssetTreeFilterModel(this);`

SourceAssetTreeFilterModel is actually the most accurate name for this given the other 2 variables. The class I removed was likely named to match this variable rather than the other way around.

rainbj:
Question. Why does it look like some columns have no numbers? I would assume if the value is 0 it would say 0 not blank.

amzn-mike:
> Question. Why does it look like some columns have no numbers? I would assume if the value is 0 it would say 0 not blank.

@rainbj  Those are for rows that refer to files, since they can't contain anything they're left blank.

bhanuja-s:
> > Question. Why does it look like some columns have no numbers? I would assume if the value is 0 it would say 0 not blank.
> 
> @rainbj Those are for rows that refer to files, since they can't contain anything they're left blank.

It might be a better experience, if there was a hyphen, or an em-dash?

bhanuja-s:
That looks great! Thank you for the changes ☺️

PR #12774, opened 2022-10-24
Fix spam caused by preview and thumbnail render request before the asset catalog was loaded

gadams3:
## What does this PR do?

Replace this hard coded product path references with asset IDs copied from asset processor.
Only attempts to queue the assets for load after the asset Catalog has been loaded just stop warnings about assets that can't be resolved.

## How was this PR tested?

Cleared asset cache and launched main editor and material editor before and after changes to see that messages no longer show

### DIFF
--- a/Gems/AtomLyIntegration/CommonFeatures/Code/Source/Material/EditorMaterialSystemComponent.cpp
+++ b/Gems/AtomLyIntegration/CommonFeatures/Code/Source/Material/EditorMaterialSystemComponent.cpp
@@ -106,7 +106,13 @@ namespace AZ
             AzToolsFramework::EditorMenuNotificationBus::Handler::BusConnect();
             AzToolsFramework::EditorEvents::Bus::Handler::BusConnect();
             AzToolsFramework::ToolsApplicationNotificationBus::Handler::BusConnect();
-            AZ::SystemTickBus::Handler::BusConnect();
+            AzFramework::AssetCatalogEventBus::Handler::BusConnect();
+
+            // All material previews use the same model and lighting preset assets
+            // models/sphere.azmodel
+            m_materialPreviewModelAsset.Create(AZ::Data::AssetId("{6DE0E9A8-A1C7-5D0F-9407-4E627C1F223C}", 284780167));
+            // lightingpresets/thumbnail.lightingpreset.azasset
+            m_materialPreviewLightingPresetAsset.Create(AZ::Data::AssetId("{4F3761EF-E279-5FDD-98C3-EF90F924FBAC}", 0));
 
             m_materialBrowserInteractions.reset(aznew MaterialBrowserInteractions);
         }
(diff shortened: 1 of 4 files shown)


Gems/AtomLyIntegration/CommonFeatures/Code/Source/Material/EditorMaterialSystemComponent.cpp
```diff
@@ -106,7 +106,13 @@ namespace AZ
...
             AzToolsFramework::EditorEvents::Bus::Handler::BusConnect();
             AzToolsFramework::ToolsApplicationNotificationBus::Handler::BusConnect();
-            AZ::SystemTickBus::Handler::BusConnect();
+            AzFramework::AssetCatalogEventBus::Handler::BusConnect();
+
+            // All material previews use the same model and lighting preset assets
+            // models/sphere.azmodel
+            m_materialPreviewModelAsset.Create(AZ::Data::AssetId("{6DE0E9A8-A1C7-5D0F-9407-4E627C1F223C}", 284780167));
+            // lightingpresets/thumbnail.lightingpreset.azasset
+            m_materialPreviewLightingPresetAsset.Create(AZ::Data::AssetId("{4F3761EF-E279-5FDD-98C3-EF90F924FBAC}", 0));
```
mbalfour-amzn, line 115: Suggestion: Could the asset IDs get looked up using code like this once the OnCatalogLoaded is fired? It seems like it would be a bit more maintainable to have the asset names hardcoded instead of the asset IDs...

```
                    AZStd::string watchFolder;
                    AZ::Data::AssetInfo assetInfo;
                    AssetSystemRequestBus::BroadcastResult(
                        foundIt, &AssetSystemRequestBus::Events::GetSourceInfoBySourcePath, "models/sphere.azmodel", assetInfo,
                        watchFolder);

```

gadams3: I agree. I prefer readability and the hint to find the original file. Using the path should also continue to work if the asset ID format changes. However, there seems to be a push to remove hard coded asset paths. I thought I would get ahead of it.

It is worth noting that I had to go to the asset processor, look up the asset IDs, and paste them into code. That's not a very convenient workflow. In addition to that, I was originally using the function AZ::Data::AssetId::CreateString and pasted the value directly from the AP into the function. I discovered that it is parsing the asset ID sub ID as an unsigned long and assigning it to an unsigned int, which breaks for values out of the supported range. It was also explained that different systems display and serialize the sub ID in either base 10 or base 16. So the asset team is looking into that.

jiaweig-amzn (approved):

Gems/AtomLyIntegration/CommonFeatures/Code/Source/SharedPreview/SharedThumbnailRenderer.cpp
```diff
@@ -24,23 +24,48 @@ namespace AZ
...
         {
-            m_defaultModelAsset.Create(SharedPreviewUtils::GetAssetIdForProductPath(DefaultModelPath), true);
-            m_defaultMaterialAsset.Create(SharedPreviewUtils::GetAssetIdForProductPath(DefaultMaterialPath), true);
-            m_defaultLightingPresetAsset.Create(SharedPreviewUtils::GetAssetIdForProductPath(DefaultLightingPresetPath), true);
+            //models/sphere.azmodel
+            m_defaultModelAsset.Create(AZ::Data::AssetId("{6DE0E9A8-A1C7-5D0F-9407-4E627C1F223C}", 284780167));
+            //lightingpresets/thumbnail.lightingpreset.azasset
+            m_defaultLightingPresetAsset.Create(AZ::Data::AssetId("{4F3761EF-E279-5FDD-98C3-EF90F924FBAC}", 0));
+            //materials/reflectionprobe/reflectionprobevisualization.azmaterial
+            m_reflectionMaterialAsset.Create(AZ::Data::AssetId("{4322FBCB-8916-5572-9CDA-18582E22D238}", 0));
```
santorac, line 49: Just heads up, I noticed that the sphere and lighting preset are in the seedList.seed file, but reflectionprobevisualization is not.

PR #12783, opened 2022-10-24
feat: add SettingRegistry for LYShine ViewportInteraction

pollend:

## What does this PR do?

This is an early attempt at moving away from QSettings and moving the configuration to `AZ::SettingsRegistry`. I ended up copying things across since I didn't want to reference large parts of ViewportSettings.h. 

The settings are not saved unless you go through Global Preferences and press saved? Just a little confused how this is intended to work actually?

## How was this PR tested?

open UI Editor and set coordinate and interaction mode.

### DIFF
--- a/Code/Framework/AzCore/AzCore/Settings/SettingRegisteryUtility.h
+++ b/Code/Framework/AzCore/AzCore/Settings/SettingRegisteryUtility.h
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#pragma once
+
+#include <AzCore/Settings/SettingsRegistry.h>
+
+namespace AZ::SettingsRegistryUtils
+{
+    template<typename T>
+    inline void SetRegistry(SettingsRegistryInterface* registry, const AZStd::string_view setting, T&& value)
+    {
+        if (registry)
+        {
+            registry->Set(setting, AZStd::forward<T>(value));
+        }
+    }
+
+    template<typename T>
+    inline AZStd::remove_cvref_t<T> GetRegistry(SettingsRegistryInterface* registry, const AZStd::string_view setting, T&& defaultValue)
+    {
+        AZStd::remove_cvref_t<T> value = AZStd::forward<T>(defaultValue);
+        if (registry)
+        {
+            T potentialValue;
+            if (registry->Get(potentialValue, setting))
+            {
+                value = AZStd::move(potentialValue);
+            }
+        }
+
+        return value;
+    }
+} // namespace AZ::SettingsRegistryUtils
--- a/Gems/LyShine/Code/Editor/LyShineEditorSettings.cpp
+++ b/Gems/LyShine/Code/Editor/LyShineEditorSettings.cpp
@@ -8,57 +8,29 @@
 #include "LyShineEditorSettings.h"
 
 #include <AzCore/Settings/SettingsRegistry.h>
+#include <AzCore/Settings/SettingRegisteryUtility.h>
 
 namespace LyShine
 {
-
-    namespace Internal
-    {
-        template<typename T>
-        void SetRegistry(const AZStd::string_view setting, T&& value)
-        {
-            if (auto* registry = AZ::SettingsRegistry::Get())
-            {
-                registry->Set(setting, AZStd::forward<T>(value));
-            }
-        }
-
-        template<typename T>
-        AZStd::remove_cvref_t<T> GetRegistry(const AZStd::string_view setting, T&& defaultValue)
-        {
-            AZStd::remove_cvref_t<T> value = AZStd::forward<T>(defaultValue);
-            if (const auto* registry = AZ::SettingsRegistry::Get())
-            {
-                T potentialValue;
-                if (registry->Get(potentialValue, setting))
-                {
-                    value = AZStd::move(potentialValue);
-                }
-            }
-
-            return value;
-        }
-    }
-
     ViewportInteraction::InteractionMode GetInteractionMode()
     {
         return static_cast<ViewportInteraction::InteractionMode>(
-            Internal::GetRegistry(InteractionModeSetting, aznumeric_cast<AZ::u64>(ViewportInteraction::InteractionMode::SELECTION)));
+            AZ::SettingsRegistryUtils::GetRegistry(AZ::SettingsRegistry::Get(), InteractionModeSetting, aznumeric_cast<AZ::u64>(ViewportInteraction::InteractionMode::SELECTION)));
     }
 
     void SetInteractionMode(ViewportInteraction::InteractionMode mode)
     {
-        Internal::SetRegistry(InteractionModeSetting, aznumeric_cast<AZ::u64>(mode));
+        AZ::SettingsRegistryUtils::SetRegistry(AZ::SettingsRegistry::Get(),InteractionModeSetting, aznumeric_cast<AZ::u64>(mode));
     }
 
     ViewportInteraction::CoordinateSystem GetCoordinateSystem()
     {
         return static_cast<ViewportInteraction::CoordinateSystem>(
-            Internal::GetRegistry(CoordinateSystemSetting, aznumeric_cast<AZ::u64>(ViewportInteraction::CoordinateSystem::LOCAL)));
+            AZ::SettingsRegistryUtils::GetRegistry(AZ::SettingsRegistry::Get(), CoordinateSystemSetting, aznumeric_cast<AZ::u64>(ViewportInteraction::CoordinateSystem::LOCAL)));
     }
 
     void SetCoordinateSystem(ViewportInteraction::CoordinateSystem system)
     {
-        Internal::SetRegistry(CoordinateSystemSetting, aznumeric_cast<AZ::u64>(system));
+        AZ::SettingsRegistryUtils::SetRegistry(AZ::SettingsRegistry::Get(),CoordinateSystemSetting, aznumeric_cast<AZ::u64>(system));
     }
 }
--- a/Gems/LyShine/Code/Editor/LyShineEditorSettings.cpp
+++ b/Gems/LyShine/Code/Editor/LyShineEditorSettings.cpp
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+#include "LyShineEditorSettings.h"
+
+#include <AzCore/Settings/SettingsRegistry.h>
+
+namespace LyShine
+{
+
+    namespace Internal
+    {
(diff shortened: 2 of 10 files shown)


Gems/LyShine/Code/Editor/LyShineEditorSettings.cpp
```diff
@@ -0,0 +1,64 @@
...
+    ViewportInteraction::CoordinateSystem GetCoordinateSystem()
+    {
+        return static_cast<ViewportInteraction::CoordinateSystem>(
+            Internal::GetRegistry(CoordinateSystemSetting, aznumeric_cast<AZ::u64>(ViewportInteraction::CoordinateSystem::LOCAL)));
+    }
+
+    void SetCoordinateSystem(ViewportInteraction::CoordinateSystem system)
+    {
+        Internal::SetRegistry(CoordinateSystemSetting, aznumeric_cast<AZ::u64>(system));
+    }
```
lumberyard-employee-dm, line 63: In theory you could write out the `/O3DE/Preferences/Editor/LyShine` section of the Settings Registry whenever SetInteractionMode or SetCoordinateMode is invoked automatically.

Or an UI option could be added to the EditorWindow to save the settings out.
The settings can be saved to any filename, as long as it is output to the `<project-path>/user/Registry` folder

Here is the documentation on dumping the SettingsRegistry to an `AZ::IO::GenericStream` derived class such as `AZ::IO::ByteContainerStream` or `AZ::IO::SystemFileStream`
https://www.o3de.org/docs/user-guide/settings/output-settings-registry/#storing-editor-preferences-using-a-key-and-an-anchor

```
AZ::SettingsRegistryMergeUtils::DumperSettings dumperSettings;
dumperSettings.m_prettifyOutput = true;
dumperSettings.m_jsonPointerPrefix = "/O3DE/Preferences/Editor/LyShine";

AZStd::string stringBuffer;
AZ::IO::ByteContainerStream stringStream(&stringBuffer);
if (auto registry = AZ::SettingsRegistry::Get();
    registry == nullptr || !AZ::SettingsRegistryMergeUtils::DumpSettingsRegistryToStream(*registry, "/O3DE/Preferences/Editor/LyShine", stringStream, dumperSettings))
{
    AZ_Warning("LyShine Editor Settings", false, R"(Unable to dump the LyShine Preferences settings from the Settings Registry)");
    return;
}

// stringBuffer now contains the dumped lyshine settings in JSON
// Now write it to disk

//! Get the path to the current projects `<project-path>/user` directory and append to the Registrt/lyshine_preferences.setreg file to it
auto lyShinePreferencesFilepath = AZ::IO::FixedMaxPath(AZ::Utils::GetProjectUserPath()) / "Registry/lyshine_preferences.setreg";

bool saved{};
constexpr auto openMode = AZ::IO::OpenMode::ModeWrite;
if (AZ::IO::SystemFileStream outputFileStream(lyShinePreferencesFilepath.c_str(), openMode);
    outputFileStream.IsOpen())
{
    saved = outputFile.Write(stringBuffer.size(), stringBuffer.data()) == stringBuffer.size();
}

AZ_Warning("LyShine Editor Settings", saved, R"(Unable to save LyShine Editor Preferences registry file to path "%s"\n)",
    lyShinePreferencesFilepath.c_str());
```

The Settings Registry will automatically load any .setreg/.setregpatch files from the current project's `<project-path>/user/Registry` directory that doesn't have Settings Registry specialization in it. 
NOTE: This only occurs automatically on application launch though.

Here is the [reference](https://www.o3de.org/docs/user-guide/settings/developer-documentation/#search-locations) to the search locations that applications look for .setreg/.setregpatch files and the order that they merge such files

Gems/LyShine/Code/Editor/LyShineEditorSettings.cpp
```diff
@@ -0,0 +1,64 @@
...
+                T potentialValue;
+                if (registry->Get(potentialValue, setting))
+                {
+                    value = AZStd::move(potentialValue);
+                }
+            }
+
+            return value;
+        }
+    }
```
hultonha, line 41: update: Please don't duplicate these, just bring in the include from `ViewportSettings.h/cpp` (or move the functions to a separate file we can include in both places)

Gems/LyShine/Code/Editor/LyShineEditorSettings.h
```diff
@@ -0,0 +1,24 @@
...
+
+#pragma once
+
+#include "AzCore/base.h"
+#include "ViewportInteraction.h"
+
+namespace LyShine
+{
+    static constexpr AZStd::string_view InteractionModeSetting = "/O3DE/Preferences/Editor/LyShine/InteractionMode";
+    static constexpr AZStd::string_view CoordinateSystemSetting = "/O3DE/Preferences/Editor/LyShine/CoordinateSystem";
```
hultonha, line 17: update: I'd suggest moving this to the .cpp file so they don't leak out and people use the functions instead (also it's best not to use `static` when defining variables in the .h, instead use `inline` or move the definition to the .cpp).

Gems/LyShine/Code/Editor/LyShineEditorSettings.h
```diff
@@ -0,0 +1,24 @@
...
+    
+    ViewportInteraction::InteractionMode GetInteractionMode();
+    void SetInteractionMode(ViewportInteraction::InteractionMode mode);
+
+    ViewportInteraction::CoordinateSystem GetCoordinateSystem();
+    void SetCoordinateSystem(ViewportInteraction::CoordinateSystem length);
+}
```
hultonha, line 24: uber nit: Closing namespace

hultonha (commented):
Looks good so far but would be good to get some more eyes on it (tagging @michabr, @AMZN-daimini and @cgalvan for visibility)

Gems/LyShine/Code/Editor/LyShineEditorSettings.cpp
pollend: umm I guess It should make sense to move this utility to SettingRegistry

Code/Framework/AzCore/AzCore/Settings/SettingRegisteryUtility.h
```diff
@@ -0,0 +1,39 @@
...
+            T potentialValue;
+            if (registry->Get(potentialValue, setting))
+            {
+                value = AZStd::move(potentialValue);
+            }
+        }
+
+        return value;
+    }
+} // namespace AZ::SettingsRegistryUtils
```
lumberyard-employee-dm, line 39: There are the same functions in [ViewportSettings.h](https://github.com/o3de/o3de/blob/development/Code/Framework/AzToolsFramework/AzToolsFramework/Viewport/ViewportSettings.h#L15-L38), instead of adding this code, that file can just be included.

Also both of these functions have flaws, that the member Get/Set functions on the SettingsRegistry does not experience.

First this `SetRegistry` function doesn't return a boolean or `AZ::Outcome<void, AZStd::string>` indicating if the value was actually set
```
    using ErrorString = AZStd::fixed_string<128>;
    using RegistryOutcome = AZ::Outcome<void, ErrorString>;

    template<typename T>
    inline RegistryOutcome SetRegistry(SettingsRegistryInterface* registry, const AZStd::string_view setting, T&& value)
    {
        if (registry == nullptr)
        {
             return AZ::Failure(ErrorString::format(R"(Failed to set registry entry for setting "%.*s". The Settings Registry is nullptr.)", AZ_STRING_ARG(setting));
        }
        if (!registry->Set(setting, AZStd::forward<T>(value))
        {
            return AZ::Failure(ErrorString::format(R"(Failed to set registry entry for setting "%.*s". The SettingsRegistry::Set has failed.)", AZ_STRING_ARG(setting));
        }

        return AZ::Success();
    }
```
Next the `GetRegistry1 function should defer assigning the defaultValue until after the query of the value from the SettingsRegistry fails to avoid an unneccessary assignment.

i.e
```
    template<typename T>
    inline AZStd::remove_cvref_t<T> GetRegistry(SettingsRegistryInterface* registry, const AZStd::string_view setting, T&& defaultValue)
    {
        if (registry == nullptr)
        {
            return AZStd::forward<T>(defaultValue);
        }

        if (AZStd::remove_cvref_t<T> value; !registry->Get(value, setting))
        {
            return AZStd::forward<T>(defaultValue);
        }

        return value;
    }
```

Also final note, the filename of "SettingsRegist**e**ryUtility.h" has an extra e in it.
The PR description also has an extra e in the term "SettingsRegist**e**ry"

pollend: if were going to change the functionality of the utility method can just deprecate those methods under [ViewportSettings.h](https://github.com/o3de/o3de/blob/development/Code/Framework/AzToolsFramework/AzToolsFramework/Viewport/ViewportSettings.h#L15-L38)

hultonha: The utility methods were just little helpers not intended to be used widely (that's why they weren't added to the Settings Registry core itself). They are all known values that we know will work which is why there's less error checking to make things less verbose and noisy. If we can keep a nice clean interface and update them to what @lumberyard-employee-dm suggested I'm fine with that but please don't make all the existing calls super noisy with lots of extra checks, that was the whole point of having these helper functions for the viewport settings in the first place.

Code/Framework/AzCore/AzCore/Settings/SettingRegisteryUtility.h
```diff
@@ -0,0 +1,39 @@
...
+    inline void SetRegistry(SettingsRegistryInterface* registry, const AZStd::string_view setting, T&& value)
+    {
+        if (registry)
+        {
+            registry->Set(setting, AZStd::forward<T>(value));
+        }
+    }
+
+    template<typename T>
+    inline AZStd::remove_cvref_t<T> GetRegistry(SettingsRegistryInterface* registry, const AZStd::string_view setting, T&& defaultValue)
```
hultonha, line 25: question: Is there a reason why we're passing the `SettingsRegistryInterface` in? Can we not just call it internally (as it's an interface it's possible to swap out in tests if necessary too) - will just make the call sites a bit cleaner

lumberyard-employee-dm: Yes actually.

The Settings Registry is not a singleton. There is a global instance that is available that can be queried from using `SettingsRegistry::Get()`, but local instances of the Settings Registry do exist and are being used.
For example the SettingsRegistryBuilder uses a local Settings Registry to isolate Settings that are applicable for the GameLauncher.

Another example is the [SettingsRegistryMergeUtils::VisitManifestJson](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Settings/SettingsRegistryMergeUtils.cpp#L1451-L1468) function uses a local Settings Registry to load in the manifest.json files like `o3de_manifest.json`, `engine.json`, `project.json` and `gem.json` to read the manifest object names (ex. `project_name` from the project.json, `engine_name` from the engine.json, `gem_name` from the gem.json).

hultonha:
@pollend It looks like this needs to be updated slightly after the PR feedback, let me know when we should take another look. Thanks!

hultonha:
Is this something you'll be able to pick-up in future @pollend or would it be best to close this for now? Thanks!

pollend:
> Is this something you'll be able to pick-up in future @pollend or would it be best to close this for now? Thanks!

probably not anytime in the near future it seems.

PR #12786, opened 2022-10-25
Update RayTracing data when the material instance is changed

dmcdiarmid-ly:
The MeshFeatureProcessor was unaware of changes to the material instance, which required reloading the entity to update the mesh RayTracing data.

Added a MaterialAssignmentNotificationBus and handled the notification to update the mesh RayTracing data.

Tested Editot DX12/Vulkan

Signed-off-by: dmcdiarmid-ly <63674186+dmcdiarmid-ly@users.noreply.github.com>

### DIFF
--- a/Gems/Atom/Feature/Common/Code/Include/Atom/Feature/Material/MaterialAssignmentBus.h
+++ b/Gems/Atom/Feature/Common/Code/Include/Atom/Feature/Material/MaterialAssignmentBus.h
@@ -1,0 +1,29 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+#pragma once
+
+#include <AzCore/EBus/EBus.h>
+
+namespace AZ
+{
+    namespace Render
+    {
+        //! Components should inherit from the MaterialAssignmentNotificationBus to receive notifications
+        class MaterialAssignmentNotifications
+            : public AZ::EBusTraits
+        {
+        public :
+            static const AZ::EBusHandlerPolicy HandlerPolicy = AZ::EBusHandlerPolicy::Multiple;
+            static const AZ::EBusAddressPolicy AddressPolicy = EBusAddressPolicy::ById;
+            using BusIdType = AZ::Data::AssetId;
+
+            virtual void OnRebuildMaterialInstance() = 0;
+        };
+        using MaterialAssignmentNotificationBus = AZ::EBus<MaterialAssignmentNotifications>;
+    } // namespace Render
+} // namespace AZ
--- a/Gems/Atom/Feature/Common/Code/Include/Atom/Feature/Mesh/MeshFeatureProcessor.h
+++ b/Gems/Atom/Feature/Common/Code/Include/Atom/Feature/Mesh/MeshFeatureProcessor.h
@@ -13,6 +13,7 @@
 #include <Atom/RPI.Public/MeshDrawPacket.h>
 #include <Atom/RPI.Public/Shader/ShaderSystemInterface.h>
 #include <Atom/Feature/Material/MaterialAssignment.h>
+#include <Atom/Feature/Material/MaterialAssignmentBus.h>
 #include <Atom/Feature/TransformService/TransformServiceFeatureProcessor.h>
 #include <Atom/Feature/Mesh/ModelReloaderSystemInterface.h>
 #include <RayTracing/RayTracingFeatureProcessor.h>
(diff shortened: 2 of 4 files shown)


rgba16f (approved):

santorac (approved):

gadams3 (changes requested):
Before you submit this, what is this bus trying to accomplish? The mesh component and mesh feature processor should already be receiving updated material instances from the material component notification bus.

dmcdiarmid-ly:
> Before you submit this, what is this bus trying to accomplish? The mesh component and mesh feature processor should already be receiving updated material instances from the material component notification bus.
> 
> Here is the notification bus. https://github.com/o3de/o3de/blob/development/Gems/AtomLyIntegration/CommonFeatures/Code/Include/AtomLyIntegration/CommonFeatures/Material/MaterialComponentBus.h#L169-L191

There are a few problems with using that bus:
1. When the material assignment is changed to a different material: The notification sets the material assignment map before the material instance is rebuilt.  This means that the ray tracing data is not updated with the new material information, but instead uses the previous (now outdated) data.  It's already removing and re-adding the RT data, but it has no effect when changing the material assignment since it hasn't rebuilt yet.
2. When material instance properties are edited with the dialog box: The notification is not called for the Main scene, only the PreviewRenderer.
3. That bus will not handle updates to the material instance that are done outside of LyIntegration.

> Another observation is that with the bus that you've created, sending notifications every time the material assignment rebuilds its instances, you could potentially be receiving that notification for every overridden material on every entity whether they are sharing material instances or creating unique ones. There can be more than one instance referencing the same asset ID.

This is by design, the RayTracing data is mesh specific so receiving an update for every entity using the mesh is required.

gadams3:
After discussing offline, removing my block. However, the new bus is providing redundant functionality that is already provided by the material component notification bus. Material assignment related data types were created exclusively for use with the material component. They will soon be moved into LY integration with the material component. 

The mesh feature processor should use a map of material instances instead of a map of material assignments so that it does not carry around all of the extra data for property overrides and assets for each mesh handle.

i am concerned that the new notification, as well as the instance created notification, is being sent excessively for each instance instead of being sent once per tick for all of the updated instances. Looking at the code, it also does not seem like the notification will be sent consistently for property changes.

dmcdiarmid-ly:
> After discussing offline, removing my block. However, the new bus is providing redundant functionality that is already provided by the material component notification bus.
The currently existing bus does not provide a notification on material instance property editing.

> Material assignment related data types were created exclusively for use with the material component. They will soon be moved into LY integration with the material component.  The mesh feature processor should use a map of material instances instead of a map of material assignments so that it does not carry around all of the extra data for property overrides and assets for each mesh handle.
The refactor will need to make sure that the material data is updated with the latest data when the material is changed, either inside or outside of the Editor, i.e., without the presence of AtomLyIntegration.  Note that it will need to do this when the Material is fully rebuilt, not in response to Editor events.  There appears to be a race condition where the Editor can inform the MeshFeatureProcessor via SetMaterialAssignmentMap prior to the RPI::Material data being fully up-to-date.

>  Looking at the code, it also does not seem like the notification will be sent consistently for property changes.
In my testing it is sent consistently, and for every property change.

gadams3:
How was it tested? Editor, game launcher, atom sample viewer? Other factors like editor only code or the prefab system might be generating false positives. Atom sample viewer does not have access to the material component, which is the only thing that will invoke this code path. Testing the game launcher will eliminate any of the editor only cases while keeping the material component active. Should be tested and verified using lua or script canvas to change properties in the game launcher without interference from other systems. If there is a strong desire to keep this separate from the material component and LY integration then a better option would be to set up and listen for property change events/notifications directly from the RPI material instance. @santorac

amzn-tommy:
>     Another observation is that with the bus that you've created, sending notifications every time the material assignment rebuilds its instances, you could potentially be receiving that notification for every overridden material on every entity whether they are sharing material instances or creating unique ones. There can be more than one instance referencing the same asset ID.

> This is by design, the RayTracing data is mesh specific so receiving an update for every entity using the material is required.

I think the issue would be if you have a model with N sub-meshes/sub-materials, and all of those materials are rebuilt, you're going to then handle the event N times in a single frame, when really SetRayTracingData only needs to be called once. But that's easy enough to fix after the fact.

>  set up and listen for property change events/notifications directly from the RPI material instance. 
I think this is the right approach. Eventually, we want to get more specific about the kinds of updates that happen, and how they are handled. E.g., if a property on a material changes, right now we will re-build the draw packet 'just in case'. When really, only certain changes actually require rebuilding a draw packet.

The same idea applies here - presumably we only need to update the material properties in the ray tracing data, and could skip the part of SetRayTracingData that gets the vertex streams for the meshes.

The material instance should have more context to signal more granular events, knowing what specifically changed.

PR #12789, opened 2022-10-25
Atom Tools: Deferring open document actions to the system tick bus and fixing other timer usage

gadams3:
## What does this PR do?

Problems were occurring when opening documents from shortcut and menu actions. Modal message boxes can be displayed for documents that log warning and error messages. Opening a document triggers notifications that could subsequently cause the actions that triggered them to be deleted as part of menus refreshing. By deferring the open document commands to use a timer or tick bus, they will be invoked as part of the regular system tick instead of nested inside of the action callback with the potential warning or error message boxes.

Replaced all QTimers that could not be connected to a parent object for lifetime management with system tick bus.

The callbacks bound to most timers and the tick bus were updated to not capture the this pointer.

Some direct function calls were converted to bus calls so the this pointer did not have to be captured.

Created and updated test scripts that do exhaustive document opening and closing to make sure things are being cleaned up correctly.

## How was this PR tested?

Manually opening, editing, closing different material canvas graphs repeatedly.
Created test scripts to open and close all material canvas graphs and a project several times in random order.

### DIFF
--- a/Gems/Atom/Tools/AtomToolsFramework/Code/Source/Document/AtomToolsDocumentSystem.cpp
+++ b/Gems/Atom/Tools/AtomToolsFramework/Code/Source/Document/AtomToolsDocumentSystem.cpp
@@ -82,6 +82,8 @@ namespace AtomToolsFramework
                 ->Event("SaveDocumentAsChild", &AtomToolsDocumentSystemRequestBus::Events::SaveDocumentAsChild)
                 ->Event("SaveAllDocuments", &AtomToolsDocumentSystemRequestBus::Events::SaveAllDocuments)
                 ->Event("SaveAllModifiedDocuments", &AtomToolsDocumentSystemRequestBus::Events::SaveAllModifiedDocuments)
+                ->Event("QueueReopenModifiedDocuments", &AtomToolsDocumentSystemRequestBus::Events::QueueReopenModifiedDocuments)
+                ->Event("ReopenModifiedDocuments", &AtomToolsDocumentSystemRequestBus::Events::ReopenModifiedDocuments)
                 ->Event("GetDocumentCount", &AtomToolsDocumentSystemRequestBus::Events::GetDocumentCount)
                 ->Event("IsDocumentOpen", &AtomToolsDocumentSystemRequestBus::Events::IsDocumentOpen)
                 ->Event("AddRecentFilePath", &AtomToolsDocumentSystemRequestBus::Events::AddRecentFilePath)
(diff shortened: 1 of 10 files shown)


Gems/Atom/Tools/AtomToolsFramework/Code/Source/Document/AtomToolsDocumentSystem.cpp
```diff
@@ -517,110 +516,38 @@ namespace AtomToolsFramework
         return result;
     }
 
-    AZ::u32 AtomToolsDocumentSystem::GetDocumentCount() const
```
gadams3, line 520: most of this big block is just reordering a couple of functions

Gems/Atom/Tools/MaterialCanvas/Scripts/TestOpenClose.py
```diff
@@ -0,0 +1,30 @@
+"""
```
cgalvan, line 1: Nice test!

cgalvan (approved):

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Window/AtomToolsMainWindow.cpp
```diff
@@ -195,22 +195,22 @@ namespace AtomToolsFramework
         if (!m_updateMenus)
         {
             m_updateMenus = true;
-            QTimer::singleShot(0, [this]() {
+            QTimer::singleShot(0, this, [this]() {
```
rgba16f, line 198: Why leave this one as a `QTimer::singleShot` rather than put it on the SystemTickBus?

rgba16f (approved):

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Window/AtomToolsMainWindow.cpp
gadams3: This one still requires access to members like the menu bar. By adding this as the second parameter, it's supposed to guarantee that the timer will be cancelled if this is destroyed before it executes. I will continue to look at rearranging it so that I can get rid of the timer completely. The main reason to switch away from timers for some things is the guarantee synchronous updates In sync with the system tick. But that also requires that the lambda or callback function is not holding on to pointers or any other data that might be invalidated before the tick or timer triggers. That's why I changed some things to use buses that won't have any effect if the handler is destroyed or disconnected.

Issue #11388, opened 2022-08-18
[SMC] Improve the Shader Variant List File Save Workflow


santorac:
**Is your feature request related to a problem? Please describe.**
The .shadervariantlist file can only be stored in two specific locations, in order to be found at runtime. It can either be in the same folder as the .shader file, or it can be in a specific path under the user's project folder. However Shader Management Console's (SMC's) current File menu items allow the user to save the file to any location. There is no good way for the average user to know where the file ought to be saved.

**Describe the solution you'd like**
Users can save the .shadervariantlist to either the original .shader location or to their game project folder, in the appropriate required path. If the shader is {SomeGem}/Assets/Ocean/Shaders/ComputeWaves.shader then the shader variant list must be in {MyProject}/ShaderVariants/Ocean/Shaders/ComputeWaves.shadervariantlist.

These are the changes we should make to the File menu:
1. Change Save As to only give the user the two valid options. This can just be a dialog with radio buttons for "Next to the .shader file" and "In the project folder".
1. Remove Save As Child, it doesn't make sense for variant lists.
2. If the file has not been saved before, the Save menu item should show the same dialog as Save As
3. (optional) Provide a third "Custom Location" radio button option that opens a typical Save As dialog where the user can pick their location. This would mostly be used by developers for testing.

**Describe alternatives you've considered**
n/a

**Additional context**
The user should be able to create some data first, and then save it later (or discard it if they want). Normally, an application would prompt the user for a save location when the file is saved the first time, but since the .shadervariantlist is a special circumstance, we expect the user to know where to save the file and have it actually work.

etherealchain:
Update:
The basic workflow is working now

1. Right click .shader file > GenerateShaderVariantListForMaterials.py
2. File > Save as Child
3. A window popup to select saving location ( in project or in engine)
4. File saved successfully
![image](https://user-images.githubusercontent.com/4178768/188085794-ef4be1e4-b8bd-428f-8b82-12b4ee3c13b5.png)

To do:

1. Make sure it worked in all condition (both generated and opened shader variant list)
2. The Save and Save as button is grey out when document is created but not opened.

santorac:
There are a couple issues with the current implementation:
1. Once you click Save or Save As and the dialog is open, there is no way to cancel. If you click the "X" it does "Save to engine". There should be a cancel button, and clicking "X" should cancel as well.
2. Save all produces an error message: 
[...]

PR #12796, opened 2022-10-25
Address issue for #11388: Add cancel button to save shader variant list process

etherealchain:

## What does this PR do?

Resolve the issue in #11388 that users couldn't cancel the saving process.
Add a cancel button in the save document window.
When users click cancel or "X", the saving process will be canceled. 

## How was this PR tested?

Test locally on my machine


Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleWindow.cpp
```diff
@@ -94,6 +95,10 @@ namespace ShaderManagementConsole
             shaderFullPath.ReplaceExtension("shadervariantlist");
             result = shaderFullPath;
         }
+        else if (msgBox.clickedButton() == cancelBtn || ret == QMessageBox::Close)
+        {
+            result = "";
+        }
```
pollend, line 101: you don't need this.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleWindow.cpp
```diff
@@ -63,7 +63,8 @@ namespace ShaderManagementConsole
         msgBox.setText("Where do you want to save the list?");
         QPushButton* projectBtn = msgBox.addButton(QObject::tr("Save to project"), QMessageBox::YesRole);
         QPushButton* engineBtn = msgBox.addButton(QObject::tr("Save to engine"), QMessageBox::NoRole);
-        msgBox.exec();
+        QPushButton* cancelBtn = msgBox.addButton(QObject::tr("Cancel"), QMessageBox::RejectRole);
+        int ret = msgBox.exec();
 
         AZ::IO::Path result;
```
pollend, line 69: declare this as an empty path but I think this should already be empty with the default constructor.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleWindow.cpp
```diff
@@ -94,6 +95,10 @@ namespace ShaderManagementConsole
             shaderFullPath.ReplaceExtension("shadervariantlist");
             result = shaderFullPath;
         }
+        else if (msgBox.clickedButton() == cancelBtn || ret == QMessageBox::Close)
+        {
+            result = "";
+        }
 
         return result.Native();
```
pollend, line 103: isn't it bad that this returns result.Native(). the result goes out of scope?

santorac: It returns by value, so this will copy the string. It would only be a problem if the return type was string&

santorac (approved):

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleWindow.cpp
etherealchain: Thanks for the tip!

pollend (approved):

PR #12797, opened 2022-10-25
Memory allocator availability

burelc-amzn:
## What does this PR do?

Make allocator instances always available, before `main()` starts

This changes the lifetime of the allocator instances to outlive any
variable that uses them. Allocators are now created on demand, and
destroyed at program exit.

Allocators that use other allocators internally should call
`AZ::AllocatorInstance<SubAllocator>::Get()` in their constructor, to
ensure the correct destruction order of the suballocator.

The OSAllocator now tracks allocations, to provide better visibility into
potential memory leaks.

The tests have been changed to check for leaks directly, instead of
checking for leaks in the allocator's destructor. This has been done to
accommodate the allocator's new lifetime.

Some of the existing API has been kept, like the `Destroy` method, even
though it is now a no-op. This was done to minimize the diff for the
initial change.

## How was this PR tested?

Existing unit tests are leveraged to ensure that the current engine startup and shutdown is unaffected


hultonha:
There are a lot of files in this PR @burelc-amzn 😅 Any advice on how best to approach this? Areas to focus on? etc... It just seems like a lot to attempt to review in one go

Also are there any breaking API changes here that customers should be notified about? If so we may want to create a deprecation issue to go along with this PR as well.

Code/Framework/AzCore/AzCore/Memory/AllocationRecords.cpp
```diff
@@ -397,6 +353,97 @@ namespace AZ::Debug
...
+                            Debug::SymbolStorage::DecodeFrames(&ai->m_stackFrames[iFrame], numToDecode, lines);
+                            numStackLevels -= numToDecode;
+                            iFrame += numToDecode;
+                        }
+                    }
+                }
+            }
+        }
+
+        AllocatorManager::Instance().DebugBreak(address, *ai);
```
lumberyard-employee-dm, line 436: I learned that [AllocatorManager::DebugBreak](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Memory/AllocatorManager.cpp#L413-L449) is completely different from [AZ::Debug::Trace::DebugBreak](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/Platform/Common/WinAPI/AzCore/Debug/Trace_WinAPI.cpp#L88-L91)

Code/Framework/AzCore/AzCore/Memory/AllocatorDebug.h
```diff
@@ -0,0 +1,66 @@
...
+    public:
+        AZ_ALLOCATOR_DEFAULT_TRAITS
+
+        DebugAllocator() = default;
+        virtual ~DebugAllocator() = default;
+
+        pointer allocate(size_type byteSize, align_type alignment = 1);
+        void deallocate(pointer ptr, size_type byteSize = 0, align_type alignment = 0);
+        pointer reallocate(pointer ptr, size_type newSize, align_type alignment = 1);
+    };
```
lumberyard-employee-dm, line 55: What is the difference between this and the OS Allocator?

Code/Framework/AzCore/AzCore/Memory/AllocatorInstance.h
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#pragma once
+
+#include <AzCore/Module/Environment.h>
+#include <AzCore/Memory/IAllocator.h>
+#include <iostream>
```
lumberyard-employee-dm, line 13: The iostream include is unneccessary

lumberyard-employee-dm: Also the `IAllocator.h` include isn't needed either.

Only an IAllocator reference is between returned so that include can be omitted as well

Code/Framework/AzCore/AzCore/Memory/AllocatorManager.cpp
```diff
@@ -12,96 +12,54 @@
...
-        s_allocManager = Environment::CreateVariable<AllocatorManager>(AZ_CRC_CE("AZ::AllocatorManager::s_allocManager"));
+        allocatorManager = Environment::CreateVariable<AllocatorManager>(AZ_CRC_CE("AZ::AllocatorManager::s_allocManager"));
 
-        s_allocManagerDebug = &(*s_allocManager);
+        s_allocManagerDebug = &(*allocatorManager);
     }
 
-    return *s_allocManager;
+    return *allocatorManager;
 }
```
lumberyard-employee-dm, line 54: It would be nice if this was placed in an AZ::Interface in a future pass on this code.

For the [NameDictionary](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Name/NameDictionary.cpp#L31-L78) we were able to update it to use an AZ::Interface while maintaining backwards compatibility

Code/Framework/AzCore/AzCore/Memory/HphaAllocator.h
```diff
@@ -0,0 +1,84 @@
...
+    class HphaSchema
+        : public HphaSchemaBase<Internal::HphaDebugAllocator>
+    {
+    public:
+        using HphaSchemaBase<Internal::HphaDebugAllocator>::HphaSchemaBase;
+    };
+} // namespace AZ
+
+#endif // AZ_HPHA_ALLOCATION_SCHEME_ALLOCATOR_H
+#pragma once
```
lumberyard-employee-dm, line 84: Can you remove the file guards and move this `#pragma once` to the top of the file?

Code/Framework/AzCore/AzCore/Memory/HphaAllocator.h
```diff
@@ -0,0 +1,84 @@
...
+        // A static assert inside of HphaSchema.cpp validates that this is the case
+        // as of commit https://github.com/o3de/o3de/commit/92cd457c256e1ec91eeabe04b56d1d4c61f8b1af
+        // When MULTITHREADED and USE_MUTEX_PER_BUCKET is defined
+        // the largest sizeof for HpAllocator is 16640 on MacOS
+        // On Windows the sizeof HpAllocator is 8384
+        // Up this value to 18 KiB to be safe
+        static constexpr size_t hpAllocatorStructureSize = 18 * 1024;
+
+        HpAllocator*        m_allocator;
+        AZStd::aligned_storage_t<hpAllocatorStructureSize, 16> m_hpAllocatorBuffer;    ///< Memory buffer for HpAllocator
```
lumberyard-employee-dm, line 62: FYI aligned_storage is [deprecated](https://stackoverflow.com/questions/71828288/why-is-stdaligned-storage-to-be-deprecated-in-c23-and-what-to-use-instead) in the future

Code/Framework/AzCore/AzCore/Memory/IAllocator.h
```diff
@@ -95,70 +57,178 @@ namespace AZ
...
+            return 0;
+        }
+
+        /**
+         * Returns memory allocated by the allocator and available to the user for allocations.
+         * IMPORTANT: this is not the overhead memory this is just the memory that is allocated, but not used. Example: the pool allocators
+         * allocate in chunks. So we might be using one elements in that chunk and the rest is free/unallocated. This is the memory
+         * that will be reported.
+         */
+        //AZ_DEPRECATED_MESSAGE("Use GetFragmentedSize instead, this method was problematic because not all overhead memory is avaialble for allocations. "
```
lumberyard-employee-dm, line 189: typo: avaialble -> available
```suggestion
        //AZ_DEPRECATED_MESSAGE("Use GetFragmentedSize instead, this method was problematic because not all overhead memory is available for allocations. "
```

Code/Framework/AzCore/AzCore/Memory/Memory.h
```diff
@@ -838,17 +472,17 @@ namespace AZ
             : m_allocator(rhs.m_allocator)
             , m_name(name) { (void)rhs; }
         AZ_FORCE_INLINE AZStdIAllocator& operator=(const AZStdIAllocator& rhs) { m_allocator = rhs.m_allocator; m_name = rhs.m_name; return *this; }
-        AZ_FORCE_INLINE pointer_type allocate(size_t byteSize, size_t alignment, int flags = 0)
+        AZ_FORCE_INLINE pointer allocate(size_t byteSize, size_t alignment)
```
lumberyard-employee-dm, line 475: Isn't AZStdIAllocator the same as the AllocatorPointer wrapper in AllocatorWrapper.h?

Code/Framework/AzCore/AzCore/Memory/SimpleSchemaAllocator.h
```diff
@@ -19,19 +19,21 @@ namespace AZ
...
     */
-    template <class Schema, class DescriptorType=typename Schema::Descriptor, bool ProfileAllocations=true, bool ReportOutOfMemory=true>
+    template <class Schema, bool ProfileAllocations=true, bool ReportOutOfMemory=true>
     class SimpleSchemaAllocator
         : public AllocatorBase
     {
     public:
-        using Descriptor = DescriptorType;
-        using pointer_type = typename Schema::pointer_type;
+        AZ_RTTI(((SimpleSchemaAllocator<Schema>), "{32019C72-6E33-4EF9-8ABA-748055D94EB2}", Schema), AllocatorBase)
```
lumberyard-employee-dm, line 27: There inner set of parentheses isn't needed here because `SimpleSchemaAllocator<Schema>` doesn't contain a comma.
This also messes up the type name when queried using `AZ::TypeInfo<T>::Name()`.

This would output the name as "(SimpleSchemaAllocator\<Schema>)<_ActualSchemaTemplateParameter>_".

The AZ_TYPE_INFO macros automatically appends the template argument to the "_ClassName" paramtter and therefore only "SimpleSchemaAllocator" should be specified.
```suggestion
        AZ_RTTI((SimpleSchemaAllocator, "{32019C72-6E33-4EF9-8ABA-748055D94EB2}", Schema), AllocatorBase)
```

Code/Framework/AzCore/AzCore/Memory/SystemAllocator.cpp
```diff
@@ -15,140 +15,54 @@
 #include <AzCore/std/functional.h>
 
 #include <AzCore/Debug/Profiler.h>
+#include <iostream>
```
lumberyard-employee-dm, line 18: This include is unused and should be removed.

Code/Framework/AzCore/AzCore/Serialization/SerializeContext.cpp
```diff
@@ -3178,17 +3178,17 @@ namespace AZ
...
+        auto genericClassInfoContainer = AZStd::move(m_moduleLocalGenericClassInfos);
+        auto serializeContextSet = AZStd::move(m_serializeContextSet);
         // Un-reflect GenericClassInfo from each serialize context registered with the module
         // The SerailizeContext uses the SystemAllocator so it is required to be ready in order to remove the data
         if (AZ::AllocatorInstance<AZ::SystemAllocator>::IsReady())
         {
             for (AZ::SerializeContext* serializeContext : serializeContextSet)
             {
-                for (const AZStd::pair<AZ::Uuid, AZ::GenericClassInfo*>& moduleGenericClassInfoPair : genericClassInfoContainer)
+                for (const auto& [specializedTypeId, genericClassInfo] : genericClassInfoContainer)
```
lumberyard-employee-dm, line 3189: It is inconsistent to change this iterator over the genericClassInfoContainer, but not the one right below this if block.

The for loop below should use a structured binding to be consistent with this change.

Code/Framework/AzCore/AzCore/UnitTest/UnitTest.h
```diff
@@ -226,54 +221,54 @@ namespace UnitTest
...
+                    const int numAllocators = allocatorManager.GetNumAllocators();
+                    allocators.resize(numAllocators);
+                    // Iterate in reverse order since some allocators could depend on others to do garbage collection.
+                    // If allocatorB depends on allocatorA, allocatorA will be registered before into the allocator manager.
+                    for (int i = numAllocators - 1; i >= 0; --i)
                     {
-                        invalidAllocatorCount++;
+                        AZ::IAllocator* allocator = allocatorManager.GetAllocator(i);
+                        allocator->GarbageCollect();
+                        allocators[i] = allocator; // keep the same order so allocators that depend on others are reporter latter
```
lumberyard-employee-dm, line 237: Why not `reserve` and `push_back` instead of resize and assign.
At the very least the resize doesn't need to construct, so `resize_no_construct` can also be used

Code/Framework/AzCore/AzCore/base.h
```diff
@@ -10,6 +10,7 @@
 #include <AzCore/variadic.h>
 #include <AzCore/PlatformDef.h> ///< Platform/compiler specific defines
 #include <AzCore/base_Platform.h>
+#include <tuple>
```
lumberyard-employee-dm, line 13: Why is this include needed in base.h?

Code/Framework/AzCore/AzCore/std/allocator_static.h
```diff
@@ -126,21 +107,13 @@ namespace AZStd
     template<AZStd::size_t Size, AZStd::size_t Alignment>
     AZ_FORCE_INLINE bool operator==(const static_buffer_allocator<Size, Alignment>& a, const static_buffer_allocator<Size, Alignment>& b)
     {
-        if (&a == &b)
-        {
-            return true;
-        }
-        return false;
+        return static_cast<bool>(&a == &b);
```
lumberyard-employee-dm, line 110: Why the static cast.
This just comparing two pointers.

Code/Framework/AzCore/AzCore/std/allocator_static.h
```diff
@@ -126,21 +107,13 @@ namespace AZStd
...
 
     template<AZStd::size_t Size, AZStd::size_t Alignment>
     AZ_FORCE_INLINE bool operator!=(const static_buffer_allocator<Size, Alignment>& a, const static_buffer_allocator<Size, Alignment>& b)
     {
-        if (&a != &b)
-        {
-            return true;
-        }
-        return false;
+        return static_cast<bool>(&a != &b);
```
lumberyard-employee-dm, line 116: repeat: Why the static_cast to bool?

Code/Framework/AzCore/AzCore/std/allocator_static.h
```diff
@@ -182,14 +153,10 @@ namespace AZStd
...
-        AZ_FORCE_INLINE static_pool_allocator(const this_type& rhs)
-            : m_name(rhs.m_name)    {  reset(); }
-        AZ_FORCE_INLINE static_pool_allocator(const this_type& rhs, const char* name)
-            : m_name(name) { (void)rhs; reset(); }
-        AZ_FORCE_INLINE this_type& operator=(const this_type& rhs)      { m_name = rhs.m_name; return *this; }
-
-        AZ_FORCE_INLINE const char*  get_name() const           { return m_name; }
-        AZ_FORCE_INLINE void         set_name(const char* name) { m_name = name; }
+        AZ_FORCE_INLINE static_pool_allocator(const this_type&)
+            {  reset(); }
```
lumberyard-employee-dm, line 157: nitpick: extra space before reset().
```suggestion
            { reset(); }
```

Code/Framework/AzCore/AzCore/std/allocator_traits.h
```diff
@@ -31,7 +31,7 @@ namespace AZStd
 
         //! pointer
         //! Attempts to use the C++ standard allocator_type::pointer typedef if avialable,
-        //! otherwise attempt to fallback to the AZStd pointer_type typedef.
+        //! otherwise attempt to fallback to the AZStd pointer typedef.
```
lumberyard-employee-dm, line 34: Can you do me a favor and fix the typo in the term avialable?

Code/Framework/AzCore/AzCore/std/allocator_traits.h
```diff
@@ -107,6 +107,18 @@ namespace AZStd
             using type = typename Allocator::difference_type;
         };
 
+        //! align_type
+        template<typename Allocator, typename AlignType, typename = void >
+        struct get_align_type
+        {
+            using type = typename std::make_unsigned<AlignType>::type;
```
lumberyard-employee-dm, line 114: `AZStd::make_ununsigned_t` should be used here.

Code/Framework/AzCore/Tests/AZStd/String.cpp
```diff
@@ -2470,7 +2470,7 @@ namespace UnitTest
         // Is the size of the pointer (used for storing the memory address of the string)
         // + the size of the string "size" member used to store the size of the string
         // + the size of the string "capacity" member used to store the capacity of the string
-        size_t constexpr ExpectedBasicStringSize = sizeof(void*) + 2 * sizeof(size_t) + sizeof(AZStd::allocator);
+        size_t constexpr ExpectedBasicStringSize = sizeof(void*) + 2 * sizeof(size_t);
```
lumberyard-employee-dm, line 2473: You have reduced the `AZStd::allocator` down to an empty class, therefore it counts for zero bytes when the empty base optimization is being used?
Good job.

lumberyard-employee-dm: You can update the static_assert comment to mention that when the basic_string class is used with a stateless_allocator it should result in 24 byte class.

Code/Framework/AzCore/Tests/AZStd/VectorAndArray.cpp
```diff
@@ -424,10 +424,9 @@ namespace UnitTest
 
         int_vector10.set_allocator(allocator2);
         AZ_TEST_VALIDATE_VECTOR(int_vector10, 100);
-        // now we move the allocated size from menager1 to manager2 (without freeing menager1)
-        AZ_TEST_ASSERT(myMemoryManager1.get_allocated_size() == myMemoryManager2.get_allocated_size());
-
-        myMemoryManager1.reset(); // flush manager 1 again (int_vector10 is stored in manager 2)
+        // now we move the allocated size from menager1 to manager2
+        AZ_TEST_ASSERT(myMemoryManager2.get_allocated_size() == 100 * sizeof(int));
+        AZ_TEST_ASSERT(myMemoryManager1.get_allocated_size() == 0); // setting the allocator moved the allocation from manage1 to manager2, freeing manager1
```
lumberyard-employee-dm, line 429: Might as well change these to `EXPECT_EQ`

Code/Framework/AzCore/Tests/Console/ConsoleTests.cpp
```diff
@@ -66,7 +66,6 @@ namespace AZ
         {
             AZ::Interface<AZ::IConsole>::Unregister(m_console.get());
             m_console = nullptr;
-            TeardownAllocator();
```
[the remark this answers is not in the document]
lumberyard-employee-dm: Or a simpler approach, change the base class to ScopedAllocatorSetupFixture

Code/Framework/AzCore/Tests/EBus/ScheduledEventTests.cpp
```diff
@@ -44,7 +44,6 @@ namespace UnitTest
             m_loggerComponent.reset();
 
             AZ::NameDictionary::Destroy();
-            TeardownAllocator();
```
lumberyard-employee-dm, line 47: repeat: Shouldn't SetupAllocator class also be removed?

Code/Framework/AzCore/Tests/Memory.cpp
```diff
@@ -2557,88 +2034,63 @@ namespace UnitTest
...
-                HeapSchema::Descriptor dlMallocDesc;
-                dlMallocDesc.m_numMemoryBlocks = 1;
-                dlMallocDesc.m_memoryBlocksByteSize[0] = hphaDesc.m_fixedMemoryBlockByteSize;
-                dlMallocDesc.m_memoryBlocks[0] = AZ_OS_MALLOC(dlMallocDesc.m_memoryBlocksByteSize[0], hphaDesc.m_fixedMemoryBlockAlignment);
-                PoolSchema::Descriptor poolDesc;
-                poolDesc.m_pageAllocator = &da;
-                poolDesc.m_numStaticPages = 100;
-                printf("\t\t\tPrealocated memory %.2f MB!\n", (float)hphaDesc.m_fixedMemoryBlockByteSize / (1024.f*1024.f));
+            {
+                // TODO Switch to using instance of HphaAllocator with a sub allocator of a fixed size
```
lumberyard-employee-dm, line 2038: TODO: Add a GHI issue ticket for making this change, so that the work can be added to the backlog

Code/Framework/AzCore/Tests/Memory/LeakDetection.cpp
```diff
@@ -6,13 +6,47 @@
...
+        bool anyAllocatorHasAllocations = false;
+        allocatorManager.GarbageCollect();
+        for (int i = numAllocators - 1; i >= 0; --i)
+        {
+            auto* allocator = allocatorManager.GetAllocator(i);
+            const AZ::Debug::AllocationRecords* records = allocator->GetRecords();
+            if (records)
+            {
+                const bool hasAllocations = records->RequestedBytes() != 0;
+                anyAllocatorHasAllocations |= hasAllocations;
```
lumberyard-employee-dm, line 32: This should use logical or.
```suggestion
                anyAllocatorHasAllocations = anyAllocatorHasAllocations || hasAllocations;
```

Code/Framework/AzFramework/Tests/FrameworkApplicationFixture.h
```diff
@@ -25,7 +25,7 @@ namespace UnitTest
      * Test fixture that starts up an AzFramework::Application.
      */
     class FrameworkApplicationFixture
-        : public ::testing::Test
+        : public UnitTest::AllocatorsTestFixture
```
lumberyard-employee-dm, line 28: Why not switch this to `ScopedAllocatorSetupFixture`?

Code/Framework/AzManipulatorTestFramework/Include/AzManipulatorTestFramework/AzManipulatorTestFrameworkTestHelpers.h
```diff
@@ -52,8 +52,9 @@ namespace UnitTest
 
     //! Fixture to provide the indirect call viewport interaction that inherits from ToolsApplicationFixture for the
     //! dependent on AzToolsFramework::ToolsApplication.
+    template<bool CheckForLeaksOnDestruction = true>
```
lumberyard-employee-dm, line 55: Does `IndirectCallManipulatorViewportInteractionFixtureMixin` fixture ever get used with a value of false?
If not, then the template parameter isn't needed.

Gems/Terrain/Code/Tests/ClipmapBoundsTests.cpp
```diff
@@ -18,7 +18,7 @@
 namespace UnitTest
 {
     class ClipmapBoundsTests
-        : public UnitTest::AllocatorsTestFixture
+        : public testing::Test
```
lumberyard-employee-dm, line 21: Did you mean to do this?
Should this be `ScopedAllocatorSetupFixture`?

Gems/PhysX/Code/Tests/PhysXColliderComponentModeTests.cpp
```diff
@@ -24,7 +24,7 @@
 namespace UnitTest
 {
     class PhysXColliderComponentModeTest
-        : public ToolsApplicationFixture
+        : public ToolsApplicationFixture<false>
```
lumberyard-employee-dm, line 27: So we don't check for leaks in the PhysXColliderCompoentModeTest fixture?

Gems/PhysX/Code/Tests/EditorTestUtilities.h
```diff
@@ -41,7 +41,7 @@ namespace PhysXEditorTests
     //! The application is created for the whole test case, rather than individually for each test, due to a known
     //! problem with buses when repeatedly loading and unloading gems. A new default world is created for each test.
     class PhysXEditorFixture
-        : public UnitTest::AllocatorsTestFixture
+        : public testing::Test
```
lumberyard-employee-dm, line 44: Was switching the base class to the Googletest `::testing::Test` intended?

Code/Framework/AzCore/AzCore/Memory/AllocatorDebug.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+    void DebugAllocator::deallocate(pointer ptr, [[maybe_unused]] size_type byteSize, [[maybe_unused]] align_type alignment)
+    {
+        AZ_OS_FREE(ptr);
+    }
+
+    DebugAllocator::pointer DebugAllocator::reallocate(pointer ptr, size_type newSize, align_type alignment)
+    {
+        return AZ_OS_REALLOC(ptr, newSize, alignment);
+    }
```
lumberyard-employee-dm, line 27: Isn't this the same as the stateless_allocator?

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Entries/RootAssetBrowserEntry.cpp
```diff
@@ -5,6 +5,7 @@
  * SPDX-License-Identifier: Apache-2.0 OR MIT
  *
  */
+#include "AzCore/IO/Path/Path_fwd.h"
```
lumberyard-employee-dm, line 8: This should angle brackets for the include
```suggestion
#include <AzCore/IO/Path/Path_fwd.h>
```

Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorLayerComponent.cpp
```diff
@@ -67,6 +67,14 @@ namespace AzToolsFramework
             }
         }
 
+        EditorLayer::~EditorLayer()
+        {
+            for (AZ::Entity* entity : m_layerEntities)
```
lumberyard-employee-dm, line 72: We can change the `m_layerEntities` type to `AZStd;:vector<AZStd::unique_ptr<AZ::Entity>>`

Code/Tools/AssetProcessor/native/tests/ApplicationManagerTests.h
```diff
@@ -9,7 +9,9 @@
 #pragma once
 
 #include <utilities/BatchApplicationManager.h>
+#if !defined(Q_MOC_RUN)
```
lumberyard-employee-dm, line 12: Why are these UnitTest files being run through QMoc?

Code/Tools/AWSNativeSDKInit/include/AWSNativeSDKInit/AWSMemoryInterface.h
```diff
@@ -15,31 +15,12 @@
...
-            return "AWSNativeSDKAllocator";
-        }
-
-        const char* GetDescription() const override
-        {
-            return "Allocator used by the AWSNativeSDK";
-        }
-        ///////////////////////////////////////////////////////////////////////////////////////////
-    };
+    AZ_ALLOCATOR_DEFAULT_GLOBAL_WRAPPER(AWSNativeSDKAllocator, AZ::SystemAllocator, "{8B4DA42F-2507-4A5B-B13C-4B2A72BC161E}")
```
lumberyard-employee-dm, line 23: We can actually delete the entire Code/Tools/AWSNativeSDKInit folder from the repo.

Gems/AtomLyIntegration/CommonFeatures/Code/Tests/MeshComponentControllerTests.cpp
```diff
@@ -53,7 +53,7 @@ namespace UnitTest
         m_lastEntityIdChanged = entityId;
     }
 
-    class MeshComponentControllerFixture : public ToolsApplicationFixture
+    class MeshComponentControllerFixture : public ToolsApplicationFixture<false>
```
lumberyard-employee-dm, line 56: The MeshComponentFixtureFixture doesn't track memory leaks either?

Gems/Blast/Code/Source/Components/BlastSystemComponent.h
```diff
@@ -99,10 +99,12 @@ namespace Blast
...
 
             void* allocate(
-                size_t size, const char* typeName, [[maybe_unused]] const char* filename,
+                size_t size,
+                [[maybe_unused]] const char* typeName,
+                [[maybe_unused]] const char* filename,
                 [[maybe_unused]] int line) override
             {
-                return azmalloc_4(size, alignment, AZ::SystemAllocator, typeName);
+                return azmalloc_3(size, alignment, AZ::SystemAllocator);
```
lumberyard-employee-dm, line 107: Why not call `azmalloc` here?

Gems/LmbrCentral/Code/Tests/SpawnerComponentTest.cpp
```diff
@@ -106,13 +106,13 @@ class SpawnerApplication
 };
 
 class SpawnerComponentTest
-    : public testing::Test
+    : public UnitTest::AllocatorsTestFixture
```
lumberyard-employee-dm, line 109: If this was changed to inherit from UnitTest::ScopedAllocatorSetupFixture, then the SystemAllocator Create call can be removed.

Gems/LyShine/Code/Tests/TextInputComponentTest.cpp
```diff
@@ -96,14 +96,14 @@ namespace UnitTest
     };
 
     class UiTextInputComponentTest
-        : public ::testing::Test
+        : public UnitTest::AllocatorsTestFixture
```
lumberyard-employee-dm, line 99: repeat: This can be changed to inherit from ScopedAllocatorSetupFixture and the `AllocatorInstance<SystemAllocator>` Create and Destroy calls can be removed.

Gems/LyShine/Code/Tests/UiDynamicScrollBoxComponentTest.cpp
```diff
@@ -64,16 +64,14 @@ namespace UnitTest
     };
 
     class UiDynamicScrollBoxComponentTest
-        : public testing::Test
+        : public UnitTest::AllocatorsTestFixture
```
lumberyard-employee-dm, line 67: repeat: This can inherit from ScopedAllocatorSetupFixture

Gems/LyShine/Code/Tests/UiScrollBarComponentTest.cpp
```diff
@@ -61,14 +61,14 @@ namespace UnitTest
     };
 
     class UiScrollBarComponentTest
-        : public testing::Test
+        : public UnitTest::AllocatorsTestFixture
```
lumberyard-employee-dm, line 64: repeat: This can inherit from ScopedAllocatorSetupFixture

Gems/LyShine/Code/Tests/UiTooltipComponentTest.cpp
```diff
@@ -84,7 +84,7 @@ namespace UnitTest
     };
 
     class UiTooltipComponentTest
-        : public testing::Test
+        : public UnitTest::AllocatorsTestFixture
```
lumberyard-employee-dm, line 87: repeat: This can inherit from ScopedAllocatorSetupFixture

lumberyard-employee-dm (commented):
Overall the changes look good, there is a lot of feedback to address.

Nothing that would be blocking though.

Gems/AudioSystem/Code/Platform/Common/Default/AudioSystemGemSystemComponent_default.cpp
```diff
@@ -14,41 +14,14 @@ namespace Audio::Platform
...
-            allocDesc.m_heap.m_numFixedMemoryBlocks = 1;
-            allocDesc.m_heap.m_fixedMemoryBlocksByteSize[0] = heapSize;
-
-            allocDesc.m_heap.m_fixedMemoryBlocks[0] = AZ::AllocatorInstance<AZ::OSAllocator>::Get().Allocate(
-                allocDesc.m_heap.m_fixedMemoryBlocksByteSize[0],
-                allocDesc.m_heap.m_memoryBlockAlignment
-            );
-
-            AZ::AllocatorInstance<AudioSystemAllocator>::Create(allocDesc);
+            AZ::AllocatorInstance<AudioSystemAllocator>::Create();
```
amzn-phist, line 19: Is there no longer a way to make a fixed-size block for an allocator to use?  Sometimes platforms have strict memory budgets, how do we express that (i.e. keep these cvars meaningful)?

Code/Framework/AzCore/AzCore/std/allocator_static.h
burelc-amzn: I just applied clang-tidy's [simplify-boolean-expr](https://clang.llvm.org/extra/clang-tidy/checks/readability/simplify-boolean-expr.html) fix-it, I'm not sure why it determined it should get a `static_cast`. It looks like that check is only supposed to add the `static_cast` if the comparison involves types that have an `explicit operator bool()`, which these do not. I'll take the cast out.

Code/Framework/AzCore/Tests/AZStd/VectorAndArray.cpp
burelc-amzn: There's enough code that uses `AZ_TEST_ASSERT` in these files already, I think keeping the existing style for these tests is ok. Let's update the assertion macros these tests are using as a separate task.

Code/Framework/AzCore/Tests/Console/ConsoleTests.cpp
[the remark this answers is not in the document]
burelc-amzn: Yes, there's definitely still some cleanup to be done here. It will happen separately.

Code/Framework/AzCore/Tests/Serialization/Json/MapSerializerTests.cpp
```diff
@@ -173,9 +173,29 @@ namespace JsonSerializationTests
...
+                    *entry.first = {242, 242.0};
+                    *entry.second = {};
+                }
+                else if (i == 2)
+                {
+                    *entry.first = {342, 342.0};
+                    *entry.second = {388, 388.0};
+                }
+                ++i;
+            }
```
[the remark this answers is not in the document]
burelc-amzn: AZStd::map is an ordered container. When the key type for a map is a pointer, then the address that pointer points to is used for the sorting algorithm for where that item should appear in the map. When a map is serialized to json, it is merely iterated over in (sorted) order.

The json tests serialize a map out to a string, then compare it against a known string, so the order of the elements is important. That means that care must be taken when inserting elements into the map, that things are inserted in the correct order. For a map of pointers, that means that the memory address of the allocated objects is what determines the map element's order.

These tests were previously relying on the allocator to always return a higher address for later allocations, which is not a guarantee of the allocator. This fixes that error in the test by allocating the objects first, inserting them in the map so they are sorted by address, then traversing the map to set the values appropriately.

Gems/Terrain/Code/Tests/ClipmapBoundsTests.cpp
burelc-amzn: That part is a bit of a mess right now. There's a few changes that need to be done next, but since these changes are so big already, I wanted to merge what is done before continuing.
* `ScopedAllocatorSetupFixture` needs to be removed now. The global allocator instances that are managed by `AZ::AllocatorInstance` can no longer be controlled by a scope. If you want to make one inside a scope, you do that by instantiating one of them directly, on the stack.
* `AllocatorsTestFixture` now serves the purpose of checking for memory leaks after each individual test, and nothing else. It needs to be renamed to `LeakDetectionFixture` (and will be soon). Most of the time, you want that, as it makes it easier to isolate where the leak is coming from. The problem is that for this fixture to work, it has to call `AZ::GetCurrentSerializeContextModule().Cleanup();`, so that the allocations made there don't show up as a leak. This conflicts with tests that use the `GemTestEnvironment`, which creates a ComponentApplication in environment setup and then reuse it across multiple tests.

Code/Tools/AssetProcessor/native/tests/ApplicationManagerTests.h
burelc-amzn: `Code/Tools/AssetProcessor/native/tests/ApplicationManagerTests.cpp` includes `MockAssetProcessorManager.h`, which contains a `Q_OBJECT`, so `moc` must run on it. `ApplicationManagerTests.cpp` also includes `ApplicationManagerTests.h`, which includes a bunch of stuff that eventually contains a raw string literal, which [moc does not support](https://bugreports.qt.io/browse/QTBUG-55962).

Gems/PhysX/Code/Tests/PhysXColliderComponentModeTests.cpp
lumberyard-employee-dm: Well we could move the Per-Module Generic Class Info to the stateless allocator.
It is not in a critical path for runtime use and it is used to store instances to template classes reflected with the current module memory space.

Alternatively could we move the Per-Module Generic Class info to be an AZ::EnvironmentVariable.
Do those get checked in the memory leak checks?

Code/Framework/AzCore/Tests/Serialization/Json/MapSerializerTests.cpp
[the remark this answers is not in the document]
lumberyard-employee-dm: Ok thanks for the clarification.

This means the test is invalid testing the order of elements, and it is something we should look into refactoring so that each JSON object stringified entry is compared instead of the entire stringified JSON object, but that is something that can be done in the future.

Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp
```diff
@@ -325,28 +303,15 @@ namespace AZ
...
                     ->DataElement(Edit::UIHandlers::CheckBox, &Descriptor::m_allocationRecordsSaveNames, "Record allocations with name saving", "Saves names/filenames information on each allocation made, useful for tracking down leaks in dynamic modules (ignored in Release builds)")
                     ->DataElement(Edit::UIHandlers::CheckBox, &Descriptor::m_allocationRecordsAttemptDecodeImmediately, "Record allocations and attempt immediate decode", "Decode callstacks for each allocation when they occur, used for tracking allocations that fail decoding. Very expensive. (ignored in Release builds)")
                     ->DataElement(Edit::UIHandlers::ComboBox, &Descriptor::m_recordingMode, "Stack recording mode", "Stack record mode. (Ignored in final builds)")
-                    ->DataElement(Edit::UIHandlers::SpinBox, &Descriptor::m_stackRecordLevels, "Stack entries to record", "Number of stack levels to record for each allocation (ignored in Release builds)")
                         ->Attribute(Edit::Attributes::Step, 1)
                         ->Attribute(Edit::Attributes::Max, 1024)
```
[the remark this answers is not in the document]
burelc-amzn: Eek, good catch.

Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp
```diff
@@ -325,28 +303,13 @@ namespace AZ
...
-                    ->DataElement(Edit::UIHandlers::SpinBox, &Descriptor::m_pageSize, "Page size", "Memory page size in bytes (must be OS page size aligned)")
-                        ->Attribute(Edit::Attributes::Step, 1024)
-                    ->DataElement(Edit::UIHandlers::SpinBox, &Descriptor::m_poolPageSize, "Pool page size", "Memory pool page size in bytes (must be a multiple of page size)")
-                        ->Attribute(Edit::Attributes::Max, &Descriptor::m_pageSize)
-                        ->Attribute(Edit::Attributes::Step, 1024)
-                    ->DataElement(Edit::UIHandlers::SpinBox, &Descriptor::m_memoryBlockAlignment, "Block alignment", "Memory block alignment in bytes (must be multiple of the page size)")
-                        ->Attribute(Edit::Attributes::Step, &Descriptor::m_pageSize)
                     ->DataElement(Edit::UIHandlers::SpinBox, &Descriptor::m_memoryBlocksByteSize, "Block size", "Memory block size in bytes (must be multiple of the page size)")
-                        ->Attribute(Edit::Attributes::Step, &Descriptor::m_pageSize)
```
amzn-phist, line 347: I noticed this `Attribute` was also deleted from the `DataElement` `m_memoryBlocksByteSize`.  It made the step size match what's mentioned in the description about multiples of page size.  But `m_pageSize` is removed.  Do we want to introduce a step size or update the description?

Code/Framework/AzCore/AzCore/Memory/AllocationRecords.cpp
burelc-amzn: Yes it is 😄

Code/Framework/AzCore/AzCore/Memory/AllocatorInstance.h
burelc-amzn: Fixed.

Code/Framework/AzCore/AzCore/Memory/HphaAllocator.h
burelc-amzn: Fixed.

Code/Framework/AzCore/AzCore/Memory/HphaAllocator.h
burelc-amzn: Indeed. However, this code was just moved from `HphaSchema.h`, not really changed.

Code/Framework/AzCore/AzCore/Memory/SimpleSchemaAllocator.h
burelc-amzn: Fixed.

Code/Framework/AzCore/AzCore/Serialization/SerializeContext.cpp
burelc-amzn: Fixed.

Code/Framework/AzCore/AzCore/base.h
burelc-amzn: Removed.

Code/Framework/AzCore/AzCore/std/allocator_static.h
burelc-amzn: Updated to remove the static cast.

Code/Framework/AzCore/AzCore/std/allocator_static.h
burelc-amzn: Updated to remove the static cast.

Code/Framework/AzCore/AzCore/std/allocator_static.h
burelc-amzn: Fixed.

Code/Framework/AzCore/AzCore/std/allocator_traits.h
burelc-amzn: Fixed.

Code/Framework/AzCore/AzCore/std/allocator_traits.h
burelc-amzn: Fixed.

Code/Framework/AzCore/Tests/Memory/LeakDetection.cpp
burelc-amzn: Fixed.

Code/Framework/AzManipulatorTestFramework/Include/AzManipulatorTestFramework/AzManipulatorTestFrameworkTestHelpers.h
burelc-amzn: Fixed.

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Entries/RootAssetBrowserEntry.cpp
burelc-amzn: Fixed. It actually doesn't need `Path_fwd.h` at all, it includes `Path.h` already.

Code/Framework/AzCore/AzCore/Memory/AllocatorDebug.cpp
burelc-amzn: Yes, duplicate allocators like this one will be removed separately.

Code/Tools/AWSNativeSDKInit/include/AWSNativeSDKInit/AWSMemoryInterface.h
burelc-amzn: Is there already a separate task for that work?

Gems/Blast/Code/Source/Components/BlastSystemComponent.h
burelc-amzn: Fixed.

Code/Framework/AzCore/AzCore/Memory/AllocatorManager.h
```diff
@@ -42,9 +41,8 @@ namespace AZ
 
         AllocatorManager();
 
-        typedef AZStd::function<void (IAllocator* allocator, size_t /*byteSize*/, size_t /*alignment*/, int/* flags*/, const char* /*name*/, const char* /*fileName*/, int lineNum /*=0*/)>    OutOfMemoryCBType;
+        typedef AZStd::function<void (IAllocator* allocator, size_t /*byteSize*/, size_t /*alignment*/)>    OutOfMemoryCBType;
```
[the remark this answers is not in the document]
burelc-amzn: I'd agree, I think it'd be nice to make a whole pass to change all of our `typedef` declarations to `using` declarations, but that can be done separately.

Code/Framework/AzCore/AzCore/Memory/Memory.h
```diff
@@ -55,19 +51,19 @@ void* operator new[](std::size_t, const AZ::Internal::AllocatorDummy*);
...
+    #define azmalloc_1(_1)                                          AZ::AllocatorInstance< AZ::SystemAllocator >::Get().allocate(_1)
 /// azmalloc(size,alignment)
-    #define azmalloc_2(_1, _2)                                      AZ::AllocatorInstance< AZ::SystemAllocator >::Get().Allocate(_1, _2, 0)
+    #define azmalloc_2(_1, _2)                                      AZ::AllocatorInstance< AZ::SystemAllocator >::Get().allocate(_1, _2)
 /// azmalloc(size,alignment,Allocator)
-    #define azmalloc_3(_1, _2, _3)                                  AZ::AllocatorInstance< _3 >::Get().Allocate(_1, _2, 0)
+    #define azmalloc_3(_1, _2, _3)                                  AZ::AllocatorInstance< _3 >::Get().allocate(_1, _2)
 /// azmalloc(size,alignment,Allocator,allocationName)
-    #define azmalloc_4(_1, _2, _3, _4)                              AZ::AllocatorInstance< _3 >::Get().Allocate(_1, _2, 0, _4)
+    #define azmalloc_4(_1, _2, _3, _4)                              AZ::AllocatorInstance< _3 >::Get().allocate(_1, _2)
```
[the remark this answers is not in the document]
burelc-amzn: I'll make a separate task to remove the ones that are no longer useful.

Code/Tools/AWSNativeSDKInit/include/AWSNativeSDKInit/AWSMemoryInterface.h
lumberyard-employee-dm: No there is not.

Code/Framework/AzCore/AzCore/Memory/SystemAllocator.cpp
burelc-amzn: Fixed.

Code/Framework/AzCore/AzCore/Memory/IAllocator.h
burelc-amzn: Fixed.

Code/Editor/Include/Command.h
```diff
@@ -94,7 +93,7 @@ class CCommand
             return m_args[i];
         }
     private:
-        AZStd::vector<AZStd::string,AZ::StdLegacyAllocator> m_args;
```
byrcolin, line 97: nit: space after ,

amzn-phist (approved):

Code/Framework/AzCore/AzCore/Memory/SystemAllocator.cpp
```diff
@@ -234,11 +115,11 @@ namespace AZ
         }
 
         AZ_Assert(
-            address != nullptr, "SystemAllocator: Failed to allocate %d bytes aligned on %d (flags: 0x%08x) %s : %s (%d)!", byteSize,
-            alignment, flags, name ? name : "(no name)", fileName ? fileName : "(no file name)", lineNum);
+            address != nullptr, "SystemAllocator: Failed to allocate %d bytes aligned on %d!", byteSize,
+            alignment);
```
lumberyard-employee-dm, line 119: This should be %zu for both byteSize and alignment as they are `size_type`

lumberyard-employee-dm (approved):

PR #12799, opened 2022-10-25
Int Conversion Error

dtamkin1:
## What does this PR do?

Fix for warnings that are being treated as errors by converting unsigned ints to int

## How was this PR tested?
Locally


lumberyard-employee-dm (approved):

lumberyard-employee-dm:
> nit: you could use `aznumeric_cast`

I think `static_cast` is fine here.
numeric_cast requires an additional include, the numeric_cast checks are only active in the C++ UnitTest and the edges it is meant to catch is not going to with use of this container. If the value is undetermined, then there is memory corruption somewhere before the numeric_cast and it will not help with catching it.
If there was somehow legitimately added 2 billion columns added, then the Editor would have been running for centuries without crashing and without needing any more RAM which is impossible.

AMZN-daimini:
Started AR.

PR #12808, opened 2022-10-26
Fix sky using wrong SRG and use AddRenderPasses

AMZN-alexpete:

## What does this PR do?

Fixes issue where the SRG changed and the sky parent pass did not update the new SRG.
Moves the addition of new passes to the `AddRenderPasses` function instead of in RenderPipelineChanged.

## How was this PR tested?

Verified the bug existed before the change using the steps:
1. Create a level with a skyatmosphere component.
1. save the level and close the editor
2. open the editor and open the level
The sky loads but disappears - verified SRG values are zeroed in RenderDoc capture

After fix, the issue no longer repros in the Editor following the steps above

### DIFF
--- a/Gems/Atom/Feature/Common/Code/Source/SkyAtmosphere/SkyAtmospherePass.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/SkyAtmosphere/SkyAtmospherePass.cpp
@@ -201,7 +201,7 @@ namespace AZ::Render
         }
     }
 
-    bool SkyAtmospherePass::NeedsShaderDataRebuild()
+    bool SkyAtmospherePass::NeedsShaderDataRebuild() const
     {
         if (m_children.empty())
         {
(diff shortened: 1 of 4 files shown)


invertednormal (approved):

jeremyong-az (approved):

Gems/Atom/Feature/Common/Code/Source/SkyAtmosphere/SkyAtmospherePass.cpp
```diff
@@ -200,8 +201,40 @@ namespace AZ::Render
...
+        }
+
+        return false;
+    }
+
     void SkyAtmospherePass::FrameBeginInternal(AZ::RPI::Pass::FramePrepareParams params)
     {
+        if (NeedsShaderDataRebuild())
+        {
+            BuildShaderData();
```
VickyAtAZ, line 235: I think you need to set m_updateConstants to true after the shader data was rebuilt

AMZN-alexpete: oh, I moved that line inside of `BuildShaderData` since I always need to update the constants after building the shader data.

PR #12252, opened 2022-09-28
Performance Metrics API: Refactored the AZ EventLogger interface

lumberyard-employee-dm:
## What does this PR do?
The EventLogger interface supports logging events in the google trace event format
(https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/)

The following events are supported: Duration Events, Complete Events, Instant Events, Counter Events and Async Events.

The Flow Events, Object Events, Metadata Events, Memory Dump Events, Mark Events and Clock Sync Event and Context Events are not supported.

Recording of C++ stack frames within events are not supported.

Added an implementation of a JSON Trace Event Logger that logs to the Google Trace Event Format.

Added UnitTest and Benchmarks test that verifies the implementation of the JSON Trace Event Logger.

Add implementation of a stack allocator for rapidjson to use to allocate memory from a pre-defined buffer.
This is used to have a `rapidjson::Writer` produce the JSON output for an event instead of hand-rolling a JSON solution.

Signed-off-by: lumberyard-employee-dm <56135373+lumberyard-employee-dm@users.noreply.github.com>

## How was this PR tested?

Added UnitTest which outputs sample test data

```json
[
  {"name":"Duration Event","id":"0","cat":"Test","ph":"B","ts":1664329933375019,"pid":31760,"tid":36036,"args":{"string":"Hello world","int64_t":-2,"uint64_t":18446462603027808255,"bool":true,"double":64.0,"array":["Hello world",-2,18446462603027808255,true,64.0],"object":{"Field1":"Hello world","Field2":-2,"Field3":18446462603027808255,"Field4":true,"Field5":64.0}}},
  {"name":"Duration Event","id":"0","cat":"Test","ph":"E","ts":1664329933375119,"pid":31760,"tid":36036,"args":{"string":"Hello world","int64_t":-2,"uint64_t":18446462603027808255,"bool":true,"double":64.0,"array":["Hello world",-2,18446462603027808255,true,64.0],"object":{"Field1":"Hello world","Field2":-2,"Field3":18446462603027808255,"Field4":true,"Field5":64.0}}},
[...]

PR #12824, opened 2022-10-26
Add BufferedSystemFileStream files to azcore_files.cmake

allisaurus:
## What does this PR do?

Adds the `BufferedSystemFileStream` files added in https://github.com/o3de/o3de/pull/12252 to the cmake project for AzCore

## How was this PR tested?

After making this change and building the `ZERO_CHECK` CMake target, project reloaded with these files visible
image


allisaurus:
Per offline discussion, these file were not indented to be added, so closing this.

Issue #12813, opened 2022-10-26
Clicking White Box Manipulator Mode switch icon multiple times results in grey square artifacts appearing in Viewport layout


LB-MateuszSzweda:
**Describe the bug**
Entering the Edit Mode of an entity based on **White Box** component and clicking the _Switch to Manipulator mode_ icon few times in a row results in grey square artifacts appearing in the Viewport, beneath the other Edit mode icons.

These grey squares persists in Viewport and Game mode until the Editor is restarted or the Viewport layout is reloaded.

Please note that White Box Manipulator mode is an experimental edit mode that is not turned on by default and requires the user to enable it manually by modifying the project's _editorpreferences.setreg_ file.

For further information please refer to the attached video file.

**Prerequisites**:
 * The selected project was built and Whitebox ManipulatorMode was enabled for it by including
```
    "O3DE": {
        "Preferences": {
            "WhiteBox": {
                "TransformFeature": true
```
lines in project's _editorpreferences.setreg_ file.

**Steps to Reproduce**:
 1. Launch the Editor and create a new level.
 2. Create an entity with the **White Box** component.
 3. Enter the **White Box** component _Edit mode_.
 4. Locate the _Switch to Manipulator mode_ icon in the Viewport and click it few times in a row. 

**Expected behavior**
White Box Manipulator mode is being turned on and off few times without leaving any artifacts in the Viewport window.

**Actual behavior**
Clicking _Switch to Manipulator mode_ icon few times in a row creates grey square artifacts in the Viewport window.

**Screenshots/Video**

https://user-images.githubusercontent.com/86953456/197989599-36c7ba9c-123f-4e6c-bed8-9a64626d232a.mp4

**Found in Branch**
Development (52059cb)

**Desktop/Device:**
- Device: PC
- OS: Windows
- Version: 10 Pro
- CPU: AMD Ryzen 9 3900X
- GPU: NVIDIA GeForce RTX 2060 SUPER
- Memory: 16GB

PR #12827, opened 2022-10-26
bugfix(Whitebox, AZToolsFramework): correctly generate unique ids for new Cluster, Switcher, TextField

pollend:
ref: https://github.com/o3de/o3de/issues/12813

## What does this PR do?

Looks like the ID's when you register a cluster are not unique so the overlapping id's will delete the wrong cluster when the transform mode icon is selected multiple times

Create Cluster --> 1 [1]
Create Cluster --> 2 [1,2]
Delete Cluster --> 1 [2]
Create Cluster --> 2 [2,2]

## How was this PR tested?

repeat steps in the linked ticket: https://github.com/o3de/o3de/issues/12813

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/ViewportUi/ViewportUiManager.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/ViewportUi/ViewportUiManager.cpp
@@ -331,7 +331,7 @@ namespace AzToolsFramework::ViewportUi
 
     ClusterId ViewportUiManager::RegisterNewCluster(AZStd::shared_ptr<Internal::ButtonGroup>& buttonGroup)
     {
-        ClusterId newId = ClusterId(++m_currentId);
+        ClusterId newId = ClusterId(++m_nextViewportUiElementId);
         m_clusterButtonGroups.insert({ newId, buttonGroup });
 
         return newId;
@@ -339,7 +339,7 @@ namespace AzToolsFramework::ViewportUi
 
     SwitcherId ViewportUiManager::RegisterNewSwitcher(AZStd::shared_ptr<Internal::ButtonGroup>& buttonGroup)
     {
-        SwitcherId newId = SwitcherId(++m_currentId);
+        SwitcherId newId = SwitcherId(++m_nextViewportUiElementId);
         m_switcherButtonGroups.insert({ newId, buttonGroup });
 
         return newId;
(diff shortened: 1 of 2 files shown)


Code/Framework/AzToolsFramework/AzToolsFramework/ViewportUi/ViewportUiManager.cpp
```diff
@@ -331,23 +332,23 @@ namespace AzToolsFramework::ViewportUi
 
     ClusterId ViewportUiManager::RegisterNewCluster(AZStd::shared_ptr<Internal::ButtonGroup>& buttonGroup)
     {
-        ClusterId newId = ClusterId(m_clusterButtonGroups.size() + 1);
+        ClusterId newId = ClusterId(AZ::Sfmt::GetInstance().Rand64());
```
hultonha, line 335: Hmm I'm not sure just using a random number here is the way to go 🤔 I know it's very unlikely but what happens if/when there is a collision? Is it because we should be using a different size value (should it be buttonGroups.size() + number of buttons per group.size() instead?) Or we should use a separate monotonically increasing counter to give a unique id and not rely on the size value at all, that might be the simplest option.

pollend: I think a counter should be find. exhausting a u64 would be a challenge.

hultonha: Nice, yeah let's go with that! We'd have to wait for the universe to end multiple times with a u64 😅 Thanks!

Code/Framework/AzToolsFramework/AzToolsFramework/ViewportUi/ViewportUiManager.h
```diff
@@ -91,5 +92,7 @@ namespace AzToolsFramework::ViewportUi
         void UpdateSwitcherButtonGroupUi(Internal::ButtonGroup* buttonGroup);
         //! Update the corresponding ui element for the given text field.
         void UpdateTextFieldUi(Internal::TextField* textField);
+
+        AZ::u64 m_currentId = 0;
```
hultonha, line 96: suggestion: This looks good but can we make the variable name a little clearer, I  guess this is `m_nextViewportUiElementId` technically?

hultonha (approved):
Looks good I just have a suggestion about the name, thanks @pollend!

gadams3 (approved):

hultonha:
Looks great @pollend, I've just kicked off a build now (https://jenkins.build.o3de.org/job/O3DE/view/change-requests/job/PR-12827/1/)

cgalvan (approved):

PR #12841, opened 2022-10-27
Improve keyboard UX of FilteredSearchWidget

MiKom:
Couple of quality-of-life improvements:
 * When the filtering menu is opened, focus on the search field so that user can start typing right away
 * When down arrow is pressed when in the search box, the list gets focus and the first visible item in the list is selected. User can then quickly check/uncheck it with spacebar
 * When focus is in the tree view and the first item is selected, pressing the up arrow will cause the search box to get focus

Overall, these changes allow for a very natural keyboard interaction with the filter menu.

Drive-By: I sorted the Qt includes

## What does this PR do?

Implements some quality of life improvements for interaction with filtered search widget

https://user-images.githubusercontent.com/104767/198277156-c68549e9-874f-4cdd-ad41-733c8ce47ef9.mp4

## How was this PR tested?

Manual testing


Code/Framework/AzQtComponents/AzQtComponents/Components/FilteredSearchWidget.cpp
```diff
@@ -232,6 +234,43 @@ namespace AzQtComponents
 
         QMenu::showEvent(e);
     }
+    bool SearchTypeSelector::eventFilter(QObject* obj, QEvent* event)
```
hultonha, line 237: uber nit: Can we add a line break between function calls please

Code/Framework/AzQtComponents/AzQtComponents/Components/FilteredSearchWidget.cpp
```diff
@@ -232,6 +234,43 @@ namespace AzQtComponents
...
+    {
+        // Focus the first visible item if pressing down from the search field
+        if (obj == m_searchField && event->type() == QEvent::KeyPress)
+        {
+            auto* keyEvent = static_cast<QKeyEvent*>(event);
+            if (keyEvent->key() == Qt::Key_Down && keyEvent->modifiers() == Qt::NoModifier)
+            {
+                const QModelIndex firstVisibleIdx = m_tree->indexAt({ 0, 0 });
+                if (firstVisibleIdx.isValid())
```
hultonha, line 246: nit/suggestion: Could use an if initializer here to reduce the scope

```
if (const QModelIndex firstVisibleIdx = m_tree->indexAt({ 0, 0 });
    firstVisibleIdx.isValid())
```

hultonha (commented):
LGTM! Nice change and thanks for including the video :+1:

Code/Framework/AzQtComponents/AzQtComponents/Components/FilteredSearchWidget.cpp
MiKom: Done

Code/Framework/AzQtComponents/AzQtComponents/Components/FilteredSearchWidget.cpp
hultonha: Thanks! 😁

Code/Framework/AzQtComponents/AzQtComponents/Components/FilteredSearchWidget.cpp
```diff
@@ -233,6 +235,44 @@ namespace AzQtComponents
...
+                {
+                    m_tree->setFocus();
+                    m_tree->setCurrentIndex(firstVisibleIdx);
+                    return true;
+                }
+            }
+        }
+
+        // Focus the search field when pressing up if the first item is selected
+        if (obj == m_tree && event->type() == QEvent::KeyPress)
```
gadams3, line 257: nit: Things could be reduced if the common checks for the event type, modifier state, and key event cast are moved to an outer if or switch statement. The check for the up and down keys could also be converted to a switch statement since both cases are mutually exclusive.

gadams3 (approved):

amzn-leenguy (approved):
Reviewing this as a UX community member in the SIG UI UX.  This is fantastic. and the video was great. Thank you.

Approved.

Code/Framework/AzQtComponents/AzQtComponents/Components/FilteredSearchWidget.cpp
MiKom: Hey,
I took a look again and I moved the cast outside:
```cpp
    bool SearchTypeSelector::eventFilter(QObject* obj, QEvent* event)
    {
        if (event->type() == QEvent::KeyPress)
        {
            const auto* keyEvent = static_cast<QKeyEvent*>(event);

            // Focus the first visible item if pressing down from the search field
            if (obj == m_searchField &&
                keyEvent->key() == Qt::Key_Down && keyEvent->modifiers() == Qt::NoModifier)
            {
                if (const QModelIndex firstVisibleIdx = m_tree->indexAt({ 0, 0 }); firstVisibleIdx.isValid())
                {
                    m_tree->setFocus();
                    m_tree->setCurrentIndex(firstVisibleIdx);
                    return true;
                }
            }

            // Focus the search field when pressing up if the first item is selected
            if (obj == m_tree &&
                keyEvent->key() == Qt::Key_Up && keyEvent->modifiers() == Qt::NoModifier)
            {
                const QModelIndex firstItem = m_tree->model()->index(0, 0);
                const bool isFirstSelected = m_tree->selectionModel()->selectedIndexes().contains(firstItem);
                if (isFirstSelected)
                {
                    m_searchField->setFocus();
                    m_tree->clearSelection();
                    return true;
                }
            }
        }

        return QMenu::eventFilter(obj, event);
    }
```

I could maybe also take the modifier check outside but I don't think it would be for much benefit. It would be another if level. Also, in the above version both characteristics of the keystroke (key and modifier) are checked together in one line making it easy to see on what type of keystroke are we reacting.

hultonha (approved):

Code/Framework/AzQtComponents/AzQtComponents/Components/FilteredSearchWidget.cpp
MiKom: @gadams3 Please let me know if you think that's better. I'll do the change if yes.

rainbj:
Thanks for these changes. Sig-UI-UX approved.

cgalvan (approved):

PR #12843, opened 2022-10-27
Change UpdateManifest's default behaviour

umesh-huawei:
Signed-off-by: Umesh Rajesh Ramchandani <umesh.rajesh.ramchandani@huawei.com>

## What does this PR do?

Change the behaviour of UpdateManifest in PrefabGroupBehaviour when it does not find manifest updates from Events::ProcessingResult::Failure to Events::ProcessingResult::Ignored as it breaks custom prefabs that do not contain content of type IMeshData in them. This allows prefabs to have content of custom data type and still process the asset and not throw error.

## How was this PR tested?

Manual Testing


Gems/Prefab/PrefabBuilder/PrefabGroup/PrefabGroupBehavior.cpp
```diff
@@ -203,7 +203,7 @@ namespace AZ::SceneAPI::Behaviors
 
         if (!manifestUpdates)
         {
-            return Events::ProcessingResult::Failure;
+            return Events::ProcessingResult::Ignored;
```
jackalbe, line 206: This means that the pipeline has been configured to generate default proc prefabs and the scene has a mesh, but the system does not have enough information (manifest updates) in order to generate a default procedural prefab. So ignoring this situation might be confusing. 

Maybe add an AZ_Warning and return ignored would be better?

umesh-huawei: Should the warning contain a sort of suggestion on what more needs to be added i.e. Maybe an IMeshData or ITransform node? Or just a warning that things might not work as expected?

jackalbe: The Scene Graph is sensitive to having a IMeshData in the scene, so having a warning that suggests adding at least one mesh data node makes sense to me.

umesh-huawei: I've added the warning in my latest commit

jackalbe (approved):

amzn-mike (approved):

Gems/Prefab/PrefabBuilder/PrefabGroup/PrefabGroupBehavior.cpp
```diff
@@ -203,7 +203,8 @@ namespace AZ::SceneAPI::Behaviors
 
         if (!manifestUpdates)
         {
-            return Events::ProcessingResult::Failure;
+            AZ_Warning("prefab", false, "Scene doesn't contain IMeshData, add atleast 1 IMeshData to generate Manifest Updates");
```
amzn-mike, line 206:
```suggestion
            AZ_Warning("prefab", false, "Scene doesn't contain IMeshData, add at least 1 IMeshData to generate Manifest Updates");
```

Issue #12812, opened 2022-10-26
Material Canvas: Asset Processor Fails When Using Normal as Output Color


AMZN-Gene:
**Describe the bug**
A simple material canvas shader which uses the vertex color as the base color will fail to compile asset processor

**Steps to reproduce**
1. Open Material Canvas
2. File > New > "New Material Canvas Document"
3. Create this canvas and save
![image](https://user-images.githubusercontent.com/32776221/197939769-42167370-5e8d-4f98-9c31-11070defa632.png)

**Expected behavior**
A material is produced which uses the vertex normal as the base color

**Actual behavior**
Asset Processor fails to compile the material asset. Maybe casting from float3 -> float4 is causing the issue. Not sure how to create a float4 from values. There's a float4 constant, but it doesn't accept input.
![image](https://user-images.githubusercontent.com/32776221/197939961-71eac040-af6f-465a-96cc-0b9e88c3e9ed.png)

**Screenshots/Video**
If applicable, add screenshots and/or a video to help explain your problem.

**Found in Branch**
development October 14th 2022

PR #12885, opened 2022-10-28
Material Canvas: Basic matrix node configs and vector casting improvements (mostly JSON changes)

gadams3:
## What does this PR do?

Created basic node configurations for floating point matrix types. Support for connecting to these data types will be added to other nodes after another change that adds more intuitive settings for specifying supported scalar, vector, matrix data types and sizes.

Extended all of the color and vector variable and input nodes with XYZW/RGBA output slots as a convenience to give immediate access to all of the vector and color components.

Updated experimental PBR template and a couple of notes to use explicit C style casting so that scalars and vectors can be converted to the correct type according to HLSL specifications. I was previously attempting to do this with implicit type conversions or accessing vector elements. The C style casting handles the cases uniformly and stops implicit type conversion warnings.

Resolves https://github.com/o3de/o3de/issues/12812

## How was this PR tested?

image
image
image

### DIFF
--- a/Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/TestData/Nodes/Functions/clamp.materialcanvasnode
+++ b/Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/TestData/Nodes/Functions/clamp.materialcanvasnode
@@ -1,4 +1,4 @@
-{
+\{
     "Type": "JsonSerialization",
     "Version": 1,
     "ClassName": "DynamicNodeConfig",
@@ -8,9 +8,9 @@
         "title": "clamp",
         "inputSlots": [
             {
-                "name": "inValue1",
-                "displayName": "Value1",
-                "description": "Value1",
+                "name": "inValue",
+                "displayName": "Value",
+                "description": "Value",
                 "supportedDataTypes": [
                     "float",
                     "float2",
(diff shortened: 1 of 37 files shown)


lsemp3d (approved):

Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/TestData/Nodes/Functions/clamp.materialcanvasnode
```diff
@@ -1,4 +1,4 @@
-{
```
onecent1101, line 1: extra back slash?

onecent1101 (approved):
Should we move the node configuration files into proper directory? they are under `TestData`, the name is so confusing

gadams3:
> Should we move the node configuration files into proper directory? they are under `TestData`, the name is so confusing

Technically, this is still an experimental release, moving into preview. I make periodic changes to the settings and formats. Will be adding better descriptions, display names, and so on. At least for the next few weeks, I want to continue calling it test data. All of the node configurations used to be referenced by paths but I changed them to be registered and looked up by UUID so the files can be moved, renamed, reorganized at any point.

Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/TestData/Nodes/Functions/clamp.materialcanvasnode
gadams3: Oh no. I missed that. Submitting a fix

gadams3: https://github.com/o3de/o3de/pull/12889

PR #12891, opened 2022-10-29
Material Pipeline System Phase 2

santorac:
## What does this PR do?

Made material pipelines use a lua script to select a shader with the lighting model indicated in the material type.
See https://github.com/o3de/sig-graphics-audio/blob/main/rfcs/MaterialPipelineAbstraction.md for the full design.

Added support for a lua "pipeline script" to the .materialpipeline file, which can process settings from the .materialtype and configure the shader build process. The .materialtype file now indicates a lighting model setting, which tells the material pipeline which shaders to select when building assets (more settings can be added in the future). This script is facilitated by a new MaterialPipelineScriptRunner class, which provides a narrow API for configuring the builder.

- Did some cleanup in MaterialTypeSourceData code to make the difference between "abstract" format and "direct" format more clear. 
- Changed the existing forward pass template into a "Base" lighting template, and added new shader templates for "Standard" and "Enhanced" lighting. The new shader code is nearly identical to the original "Base" code, with only minor changes, mostly to initialize additional Surface data. 
- Added scripts for both material pipelines that enable the appropriate forward pass shader depending on which lighting model is requested. 
- Added new test material types for Standard and Enhanced lighting models. These are very similar to MaterialPipelineTest_Basic.materialtype, but with additional properties for clear coat and subsurface scattering. 
- Changed INSERT_AZSL_HERE to INSERT_AZSL_PATH_HERE to be a bit more clear. 
- Fixed a minor oversight in StandardPBR_ClearCoatEnableFeature.lua where it failed to report a shader option dependency, which caused a warning at runtime. 
- Factored out a ApplyClearCoatToSpecularF0() utility function from several shaders. 
- Moved GetSubsurfaceScatteringFactorAndQuality() to ForwardPassVertexAndPixel.azsli because it's all about packing render targets, not actually about the Surface struct. This removed a dependencies on LightingOptions.azsli from SurfaceData files. 
- Added some #include LightingOptions.azsli in files where the lighting options were references, as I was getting some compile errors. 
- Added missing reflection of MaterialFunctorSourceData which was necessary for saving the intermediate .materialtype file. 
- Moved some common lua bindings into a new LuaScriptUtilities class, for sharing between LuaMaterialFunctor and MaterialPipelineScriptRunner. 
- Fixed some dependency problems in MaterialTypeBuilder where a file would be missing, and replacing the file did not wake up the builder.

## Reviewer Guide

1. Look at [MaterialPipelineTest_Basic.materialtype](https://github.com/o3de/o3de/pull/12891/files#diff-9e5347902ac9ad44779e43efb38bc396c1e4a347373671eb397be09a59ee4718) and [MaterialPipelineTest_Animated.materialtype](https://github.com/o3de/o3de/pull/12891/files#diff-727b783a5cf4e4fbaeeb1888ff6a9cc6e8f59fcad7eba68bd2560cdd4a5229ae) to see that .materialtype now includes a "lightingModel" setting.
1. Look at [LowEndPipeline.materialpipeline](https://github.com/o3de/o3de/pull/12891/files#diff-59278a8a03064b30d32441c0d2e2fe4665e4fc81697a71c0dd572f55ef6b9ca7) and [MainPipeline.materialpipeline](https://github.com/o3de/o3de/pull/12891/files#diff-9de60abe1cf1aa8ddd7f1845acbb7946ff0c8c631e4ddacb5a23ae984bc1b4c3) to see that we now include three different forward pass shaders. Notice that there is also a new "pipelineScript" field that references a lua script.
1. Look at [LowEndPipelineScript.lua](https://github.com/o3de/o3de/pull/12891/files#diff-4edaf33f3c2e2b6df6f4def696524caa2e2f6cd65aba11edb8cac6f0eef1f83f) and [MainPipelineScript.lua](https://github.com/o3de/o3de/pull/12891/files#diff-d4f45fcf47cf642db5c4d2690fd7170cebd80661670dfd029e2ce10bcc180a7d) to see how these scripts choose which forward pass shader to use, based on the "lightingModel" setting from the .materialtype.
1. Look at [MaterialPipelineScriptRunner.h](https://github.com/o3de/o3de/pull/12891/files#diff-ac3a0134eaffe334e4cb610fa2c7a391b465affb02b2bd8689e1a0bc3c95956d) and [MaterialPipelineScriptRunner.cpp](https://github.com/o3de/o3de/pull/12891/files#diff-6712c9c51d9ca995df7b55061eb5f3c615a7c38eefda76340db3f8d3c04c609a) to see how the lua script and corresponding API is set up. (This is similar to LuaMaterialFunctorSourceData and LuaMaterialFunctor).
1. [MaterialTypeBuilder.cpp](https://github.com/o3de/o3de/pull/12891/files#diff-ccbd2a4b5acec4cea74f13936d702da932985a9cbebbd4bb2f8b09b247e68126) has new dependencies and hooks up the MaterialPipelineScriptRunner.
2. You're on your own...

## How was this PR tested?

- Created materials using the new material types in the Material Editor and compared to StandardPbr and EnhancedPbr.
- AtomSampleViewer full test suite showed no changes.
- Added new ASV test cases for the new material types.


jeremyong-az (approved):

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/MainPipeline/ForwardPass_StandardLighting.shader.template
```diff
@@ -0,0 +1,43 @@
...
+    "DepthStencilState" :
+    {
+        "Depth" :
+        {
+            "Enable" : true,
+            "CompareFunc" : "GreaterEqual"
+        },
+        "Stencil" :
+        {
+            "Enable" : true,
```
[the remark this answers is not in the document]
santorac: Yes they do, and that's actually fine since this shader template is defined inside the material pipeline. Some other pipeline would have a different shader template and could therefore have different stencil settings.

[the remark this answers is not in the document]
santorac: Something that would be interesting to consider is whether the material type could have some influence over the shader's depth/stencil settings. I think in that case we would use the material pipeline script concept, possibly expanding its API if needed. I think we can address that when the use case arises, though.

Gems/Atom/RPI/Code/Source/RPI.Builders/Material/MaterialTypeBuilder.cpp
```diff
@@ -403,15 +441,8 @@ namespace AZ
...
-            if (!materialType.m_shaderCollection.empty())
-            {
-                // It technically is possible for the shader list to include shaders and have these combined with the auto-generated shaders below.
-                // However, there isn't a clear use case for this, so it is not allowed just in case it would introduce new edge cases. If a valid
-                // use case is presented, we could consider keeping the m_shaderCollection.
-                AZ_Warning(MaterialTypeBuilderName, false, "The shader list in the .materialtype file is not empty. This is unexpected and will be ignored.");
-                materialType.m_shaderCollection.clear();
-            }
+            materialType.m_lightingModel.clear();
+            materialType.m_shaderCollection.clear(); // This should already be clear, but just in case
```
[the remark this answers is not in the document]
santorac: I think the fact that it's already clear is more of a coincidence of the fact that we already check for this in CreateJobs, not something we need to assert to be true. I could see us reverting this change to put the warning back in place, but at the time I felt it was excessive.

Gems/Atom/RPI/Code/Source/RPI.Reflect/Material/LuaScriptUtilities.cpp
```diff
@@ -0,0 +1,39 @@
...
+    namespace RPI
+    {
+        void LuaScriptUtilities::Reflect(AZ::BehaviorContext* behaviorContext)
+        {
+            behaviorContext->Method("Error", &Error);
+            behaviorContext->Method("Warning", &Warning);
+            behaviorContext->Method("Print", &Print);
+        }
+
+        void LuaScriptUtilities::Error([[maybe_unused]] const AZStd::string& message)
```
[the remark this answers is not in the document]
santorac: Yes, because the trace macros will be stripped out in Release builds.

Gems/Atom/RPI/Code/Source/RPI.Edit/Material/MaterialPipelineSourceData.cpp
```diff
@@ -43,6 +44,7 @@ namespace AZ
                         ->ClassElement(AZ::Edit::ClassElements::EditorData, "")
                         ->Attribute(AZ::Edit::Attributes::AutoExpand, true)
                         ->DataElement(AZ::Edit::UIHandlers::Default, &MaterialPipelineSourceData::m_shaderTemplates, "Shader Templates", "List of templates used to generate material-specific shaders.")
+                        ->DataElement(AZ::Edit::UIHandlers::Default, &MaterialPipelineSourceData::m_pipelineScript, "Pipeline Script Path", "The pipeline can use a lua script to configure shader compilation.")
```
gadams3, line 47: Later, we should update these with file picker controls. There is one that @cgalvan implemented but it requires AZ::IO::Path and one that I implemented but I have to move it from AtomTF

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/Common/DepthPass.shader.template
```diff
@@ -1,5 +1,5 @@
 {
-    "Source" : "INSERT_AZSL_HERE",
+    "Source" : "INSERT_AZSL_PATH_HERE",
```
gadams3, line 2: This is a much better name. Thank you. I would still recommend having it use a default file path instead and removing the template extension. Besides allowing the AP to pick up and validate the hand edited file as is, material canvas allows you to open and edit shader files. 

Hopefully we can keep all of the new formats and changes to existing format simple enough that we can easily register material related data types as documents in material canvas, material editor, SMC, etc. 

I need to revisit some structures to make them a little more user friendly and add descriptions for all of the fields. 

We could also reregister the file type with the template extension. Please excuse the low res screenshots.

image
image

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/LowEndPipeline/LowEndPipelineScript.lua
```diff
@@ -0,0 +1,39 @@
...
+
+        context:IncludeShader("DepthPass")
+        context:IncludeShader("ShadowmapPass")
+        context:IncludeShader("ForwardPass_StandardLighting")
+        return true
+    end
+    
+    Error('Unsupported lighting model "' .. lightingModel .. '".')
+    return false
+end
```
gadams3, line 38: What other use cases do you see for requiring lua? This is cool and potentially powerful but my concern is that it means somebody has to write lua code. I have not looked through all of the lua files in the PR yet. So far, it kind of looks like inclusion and exclusion lists, basic rules that could be set up using a table in a reflected data structure instead of script. Doing it that way means it can be configured from one of our tools or Python scripts.

gadams3 (approved):

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/Common/DepthPass.shader.template
santorac: It might work out the way you suggest at some point, I'll keep it in mind.

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/LowEndPipeline/LowEndPipelineScript.lua
santorac: The Material Pipeline section of the RFC goes into those details: https://github.com/o3de/sig-graphics-audio/blob/main/rfcs/MaterialPipelineAbstraction.md#material-pipeline

There could be other settings besides lighting model that dictate which shaders should be compiled. You could also use the script to set preprocessor flags (though I don't think that was covered in the RFC). Not every material pipeline will necessarily use these features, but I think supporting a script is the most flexible way to enable use cases that we haven't thought of yet. The lua API can be easily extended to support new things, like configuring depth/stencil states.

PR #12898, opened 2022-10-31
Address serveral issues in SMC

etherealchain:

## What does this PR do?

Address some issues mentioned in #11388 

1. Fix Save All path issue
2. Fix GenerateShaderVariantListForMaterials.py will generate an empty list
3. Fix another path issue, which occurs when users choose to save a new document on closing.

## How was this PR tested?

Test on my local machine.


gadams3 (commented):
What problems was this trying to resolve? There was an issue if the user tried to close the application or all of the windows while having unsafe changes. If they chose to cancel it would close the application anyway and lose their changes. I fixed that so if they cancelled it would not close the main window or application, leaving the window and unsaved documents open.

Going through the document system should do all of the checks and send all the notifications that will keep the window and application from closing. You should not have to loop over the tabs manually.

Either way, let's try to keep the changes isolated to the document system itself. The changes also need to be tested in material editor and material canvas.

etherealchain:
As I said on slack, the main problem this PR trying to solve is the empty path issue.
It can be divided into saving and closing actions, but they are actually the same problem.

As Chris mentioned in #11388, Save All will not work if there are newly generated documents.
The way we are dealing with empty paths now is using `GetSaveDocumentParams()`, which is defined in `AtomToolsDocumentMainWindow` and unaccessible from `AtomToolsDocumentSystem`, so we have to loop through tabs to check if the path exists for each of them.

The same goes to the closing action, if you close a newly generated document, a popup window will ask you if you want to save the changes, but the saving will fail if you chose yes, because the path is empty.
The logic for this is living in `AtomToolsDocumentSystem::CloseDocument()` now and it had no way to handle empty path, because it doesn't have access to `GetSaveDocumentParams()`.
So I have to move code around and add some closing checking functions in `AtomToolsDocumentMainWindow`, and again we have to loop through tabs because `AtomToolsDocumentSystem` is not handling this.

When you said 

>  Either way, let's try to keep the changes isolated to the document system itself. The changes also need to be tested in material editor and material canvas.

Do you mean we shouldn't modify `AtomToolsDocumentMainWindow`?
What is the better approach here? If my explanation here makes sense.

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Document/AtomToolsDocumentMainWindow.cpp
```diff
@@ -211,6 +196,115 @@ namespace AtomToolsFramework
...
+            }
+            if (selection == QMessageBox::Yes)
+            {
+                if (!SaveDocument(documentId))
+                {
+                    const QString title = QObject::tr("Document could not be closed");
+                    const QString text = QObject::tr("Close document failed because document was not saved: \n%1").arg(documentPath.c_str());
+                    AZ_Error("AtomToolsDocumentMainWindow", false, "%s: %s", title.toUtf8().constData(), text.toUtf8().constData());
+                    QMessageBox::critical(
+                        GetToolMainWindow(), title, QObject::tr("%1\nThese messages can be disabled from settings.").arg(text));
```
gadams3, line 263: This message cannot be disabled because it is not checking the registry setting that was being used in the other file. So I would just remove this part of the message.

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Document/AtomToolsDocumentMainWindow.cpp
```diff
@@ -150,29 +125,39 @@ namespace AtomToolsFramework
...
-                result, m_toolId, &AtomToolsDocumentSystemRequestBus::Events::SaveAllDocuments);
-            if (!result)
+
+            for (int index = 0; index < m_tabWidget->count(); index++)
             {
-                SetStatusError(tr("Document save all failed"));
+                if (!SaveDocument(GetDocumentTabId(index)))
+                {
+                    // Stop if there is any save failed or cancel
+                    break;
```
gadams3, line 134: continue;

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Document/AtomToolsDocumentMainWindow.cpp
```diff
@@ -150,29 +125,39 @@ namespace AtomToolsFramework
...
+                AtomToolsDocumentSystemRequestBus::Event(m_toolId, &AtomToolsDocumentSystemRequestBus::Events::CloseDocument, documentId);
+            }
         }, QKeySequence::Close);
 
         m_actionCloseAll = CreateActionAtPosition(m_menuFile, insertPostion, "Close All", [this]() {
-            AtomToolsDocumentSystemRequestBus::Event(m_toolId, &AtomToolsDocumentSystemRequestBus::Events::CloseAllDocuments);
+            if (CloseAllDocumentCheck())
+            {
+                AtomToolsDocumentSystemRequestBus::Event(m_toolId, &AtomToolsDocumentSystemRequestBus::Events::CloseAllDocuments);
+            }
```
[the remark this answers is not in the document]
etherealchain: I tried this before and there is a problem here.
Because `AtomToolsDocumentSystemRequestBus::Events::CloseDocument` actually closes the document, `m_tabWidget->count()` will change, and this will mess up the loop.
That's why I change it to check first, then close the tabs with respective functions.

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Document/AtomToolsDocumentMainWindow.cpp
etherealchain: If we continue here it would be a different behavior than what Chris mentioned in #11388 

> And I think the Cancel button should abort the entire Save All operation, not just the current file.

gadams3: The purpose of the tools framework is to keep things as consistent as possible, ideally requiring as little custom code as possible, between tools. That being said, using continue wouldn't do anything in this case. I must have been reading the wrong block. 

What you mentioned elsewhere makes more sense. Selecting no when asked about saving the file should continue the loop. Selecting cancel should break the loop. That should be consistent across all of the save all, close all, close all except actions. We will have to consolidate code and make sure things are behaving consistently between the document system class and the main window class.

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Document/AtomToolsDocumentMainWindow.cpp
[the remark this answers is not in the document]
gadams3: I see. That would just require a minor update to loop over the tabs and build a set of document IDs. Then loop over the document IDs and close them. You could even pass in the set of documents to close instead of passing in the document to ignore. Then the function would be used consistently for all close actions.


```
const AZStd::vector<AZ::Uuid> GetOpenDocumentIds() const
{
    AZStd::vector<AZ::Uuid> documentIds;
    documentIds.reserve(m_tabWidget->count());
    for (int index = 0; index < m_tabWidget->count(); index++)
    {
        documentIds.push_back(GetDocumentTabId(index));
    }
    return documentIds;
}

bool CloseDocuments(const AZStd::vector<AZ::Uuid>& documentIds)
{
    for (const auto& documentId : documentIds)
    {
        if (!CloseDocumentCheck(documentId))
        {
            return false;
        }

        bool result = false;
        AtomToolsDocumentSystemRequestBus::EventResult(
            result, m_toolId, &AtomToolsDocumentSystemRequestBus::Events::CloseDocument, documentId);
        if (!result)
        {
            return false;
        }
    }
    return true;
}

CloseDocuments(GetOpenDocumentIds());
```

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Document/AtomToolsDocumentMainWindow.cpp
gadams3:
```
for (const auto& documentId : GetOpenDocumentIds())
{
    if (!SaveDocument(documentId))
    {
        break;
    }
}
```

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Document/AtomToolsDocumentMainWindow.cpp
```diff
@@ -150,29 +125,39 @@ namespace AtomToolsFramework
...
         });
         m_menuFile->insertSeparator(insertPostion);
 
         m_actionClose = CreateActionAtPosition(m_menuFile, insertPostion, "&Close", [this]() {
             const AZ::Uuid documentId = GetCurrentDocumentId();
-            AtomToolsDocumentSystemRequestBus::Event(m_toolId, &AtomToolsDocumentSystemRequestBus::Events::CloseDocument, documentId);
+            if (CloseDocumentCheck(documentId))
+            {
+                AtomToolsDocumentSystemRequestBus::Event(m_toolId, &AtomToolsDocumentSystemRequestBus::Events::CloseDocument, documentId);
+            }
```
gadams3, line 145: CloseDocuments({GetCurrentDocumentId()});

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Document/AtomToolsDocumentMainWindow.cpp
```diff
@@ -150,29 +125,39 @@ namespace AtomToolsFramework
...
+            {
+                AtomToolsDocumentSystemRequestBus::Event(m_toolId, &AtomToolsDocumentSystemRequestBus::Events::CloseAllDocuments);
+            }
         });
 
         m_actionCloseOthers = CreateActionAtPosition(m_menuFile, insertPostion, "Close Others", [this]() {
             const AZ::Uuid documentId = GetCurrentDocumentId();
-            AtomToolsDocumentSystemRequestBus::Event(
-                m_toolId, &AtomToolsDocumentSystemRequestBus::Events::CloseAllDocumentsExcept, documentId);
+            if (CloseAllDocumentExceptCheck(documentId))
```
gadams3, line 157: auto documentIds = GetOpenDocumentIds();
AZStd::erase(documentIds. GetCurrentDocumemtId()); // whichever algorithm is needed for the container
CloseDocuments(documentIds);

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Document/AtomToolsDocumentMainWindow.cpp
```diff
@@ -389,7 +483,10 @@ namespace AtomToolsFramework
 
         connect(m_tabWidget, &QTabWidget::tabCloseRequested, this, [this]() {
             const AZ::Uuid documentId = GetCurrentDocumentId();
-            AtomToolsDocumentSystemRequestBus::Event(m_toolId, &AtomToolsDocumentSystemRequestBus::Events::CloseDocument, documentId);
+            if (CloseDocumentCheck(documentId))
+            {
+                AtomToolsDocumentSystemRequestBus::Event(m_toolId, &AtomToolsDocumentSystemRequestBus::Events::CloseDocument, documentId);
+            }
```
gadams3, line 489: CloseDocuments({GetCurrentDocumentId()});

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Document/AtomToolsDocumentMainWindow.cpp
```diff
@@ -581,18 +678,24 @@ namespace AtomToolsFramework
             AtomToolsDocumentNotificationBus::Event(m_toolId, &AtomToolsDocumentNotificationBus::Events::OnDocumentOpened, documentId);
         });
         menu.addAction("Close", [this, documentId]() {
-            AtomToolsDocumentSystemRequestBus::Event(m_toolId, &AtomToolsDocumentSystemRequestBus::Events::CloseDocument, documentId);
+            if (CloseDocumentCheck(documentId))
+            {
+                AtomToolsDocumentSystemRequestBus::Event(m_toolId, &AtomToolsDocumentSystemRequestBus::Events::CloseDocument, documentId);
+            }
```
gadams3, line 684: CloseDocuments({documentId});

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Document/AtomToolsDocumentMainWindow.cpp
```diff
@@ -581,18 +678,24 @@ namespace AtomToolsFramework
...
         });
         menu.addAction("Close Others", [this, documentId]() {
-            AtomToolsDocumentSystemRequestBus::Event(m_toolId, &AtomToolsDocumentSystemRequestBus::Events::CloseAllDocumentsExcept, documentId);
+            if (CloseAllDocumentExceptCheck(documentId))
+            {
+                AtomToolsDocumentSystemRequestBus::Event(m_toolId, &AtomToolsDocumentSystemRequestBus::Events::CloseAllDocumentsExcept, documentId);
+            }
```
gadams3, line 690: auto documentIds = GetOpenDocumentIds();
AZStd::erase(documentIds. documemtId); // whichever algorithm is needed for the container
CloseDocuments(documentIds);

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Document/AtomToolsDocumentMainWindow.cpp
```diff
@@ -664,8 +767,16 @@ namespace AtomToolsFramework
...
+        if (CloseAllDocumentCheck())
+        {
+            AtomToolsDocumentSystemRequestBus::EventResult(canClose, m_toolId, &AtomToolsDocumentSystemRequestBus::Events::CloseAllDocuments);
+        }
+        else
+        {
+            canClose = false;
+        }
+        
+        if(!canClose)
```
gadams3, line 779: if (!CloseDocuments(GetOpenDocumentIds()))
 {
            closeEvent->ignore();
            return;
}

gadams3:
I forgot to mention that I looked at some other multi document tools like Visual Studio. 

If a new document is created but not modified then Visual Studio does not prompt to close it. It's actually creating and immediately saving every new document to the temp folder.

It will only prompt the user if any files or modified, bringing up a confirmation dialog listing all of the modified files, asking if the user wants to save any of the files or ignore them all the files before proceeding. However, it prompts the user and opens a save dialog for each individual unsaved file as they are being iterated over.

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Document/AtomToolsDocumentMainWindow.cpp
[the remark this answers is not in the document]
etherealchain: Thanks for the idea!

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Document/AtomToolsDocumentMainWindow.cpp
```diff
@@ -211,6 +187,110 @@ namespace AtomToolsFramework
...
+            }
+        }
+        return true;
+    }
+
+    const AZStd::vector<AZ::Uuid> AtomToolsDocumentMainWindow::GetOpenDocumentIds() const
+    {
+        AZStd::vector<AZ::Uuid> documentIds;
+        documentIds.reserve(m_tabWidget->count());
+        for (int index = 0; index < m_tabWidget->count(); index++)
```
gadams3, line 287: Super minor compared to the rest of the function but ++index is more optimal and preferred in this case, also more consistent with the rest of the file

galibzon (approved):

PR #12899, opened 2022-10-31
Replace the python image comparison method with the one in c++

jiaweig-amzn:

## What does this PR do?
1. Changed the C++ image comparison API to return the result as a struct, so that it can be reflected through behavior context.
2. Replace the python image comparison method with the one in C++. Change the test behavior a bit so that screenshot comparison is also done by the engine instead of external python code. In this way, screenshot test can happen in any platform and a unified algorithm for comparison is used.

## How was this PR tested?

Run through the updated tests. Vulkan area lights don't show up in the first 3 screenshots. Other screenshots all passed.
Those tests remain as skipped in this PR.


AMZN-alexpete:
If we no longer need the Python modules for comparing image captures and we could remove them, I think that would be a significant install size savings.   Perhaps they're used for other things but I'm thinking of numpy (60mb), scipy(120mb)

jiaweig-amzn:
> 

Thanks for the notice. It looks to be safe to remove [scipy](https://github.com/o3de/o3de/search?q=scipy) since it is only used in image comparison. The other [numpy](https://github.com/o3de/o3de/search?q=numpy) seems to be used more widely. I'll double check the possibility to remove them in a separate PR.
And yes, the python comparison should be deprecated. The engine comparison can allow testing on platforms such as Android to be possible.

jiaweig-amzn:
> If we no longer need the Python modules for comparing image captures and we could remove them, I think that would be a significant install size savings. Perhaps they're used for other things but I'm thinking of [numpy ](https://github.com/o3de/o3de/search?q=numpy)(60mb), [scipy](https://github.com/o3de/o3de/search?q=scipy)(120mb)
> 
> image

Thanks for the notice. It looks to be safe to remove [scipy](https://github.com/o3de/o3de/search?q=scipy) since it is only used in image comparison. The other [numpy](https://github.com/o3de/o3de/search?q=numpy) seems to be used more widely. I'll double check the possibility to remove them in a separate PR.
And yes, the python comparison should be deprecated. The engine comparison can allow testing on platforms such as Android to be possible.

Gems/Atom/Utils/Code/Source/ImageComparison.cpp
```diff
@@ -9,11 +9,55 @@
...
 {
     namespace Utils
     {
+        AZ_ENUM_DEFINE_REFLECT_UTILITIES(ImageDiffResultCode);
+
+        void ImageDiffResult::Reflect(ReflectContext* context)
+        {
+            if (auto* serializeContext = azrtti_cast<SerializeContext*>(context))
+            {
+                ImageDiffResultCodeReflect(*serializeContext);
```
gadams3, line 24: There is an equivalent function to call for the behavior context so you don't have to reflect the enum manually.

Gems/Atom/Utils/Code/Source/ImageComparison.cpp
```diff
@@ -9,11 +9,55 @@
...
+                    ->Attribute(AZ::Script::Attributes::Module, "utils");
+                behaviorContext->EnumProperty<static_cast<int>(ImageDiffResultCode::SizeMismatch)>("ImageDiffResultCode_SizeMismatch")
+                    ->Attribute(AZ::Script::Attributes::Scope, AZ::Script::Attributes::ScopeFlags::Automation)
+                    ->Attribute(AZ::Script::Attributes::Module, "utils");
+                behaviorContext->EnumProperty<static_cast<int>(ImageDiffResultCode::FormatMismatch)>("ImageDiffResultCode_FormatMismatch")
+                    ->Attribute(AZ::Script::Attributes::Scope, AZ::Script::Attributes::ScopeFlags::Automation)
+                    ->Attribute(AZ::Script::Attributes::Module, "utils");
+                behaviorContext->EnumProperty<static_cast<int>(ImageDiffResultCode::UnsupportedFormat)>("ImageDiffResultCode_UnsupportedFormat")
+                    ->Attribute(AZ::Script::Attributes::Scope, AZ::Script::Attributes::ScopeFlags::Automation)
+                    ->Attribute(AZ::Script::Attributes::Module, "utils");
```
gadams3, line 46: AZ_ENUM macros produce a function that you can call and set this up automatically

jiaweig-amzn: Per our offline discussion, this is still needed for exposing the enum to pytest, before we have more support from AZ_ENUM_DEFINE_REFLECT_UTILITIES.

gadams3 (approved):

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/screenshot_utils.py
```diff
@@ -91,19 +92,23 @@ def wait_until_screenshot(self):
...
+    ScreenshotHelper(general.idle_wait_frames).capture_screenshot_blocking(f"{screenshot_name}.png")
     general.idle_wait(1.0)
     general.exit_game_mode()
     helper.wait_for_condition(lambda: not general.is_in_game_mode(), 2.0)
     general.log(f"{entity_name}_test: Exit game mode: {not general.is_in_game_mode()}")
+
+def compare_screenshots(imageA, imageB):
+    result = azlmbr.atom.FrameCaptureTestRequestBus(
+        azlmbr.bus.Broadcast, "CompareScreenshots", imageA, imageB, 0.01)
+    return result
```
smurly, line 114: before you return result you should assert result_code == ImageDiffResultCode::Success however you would represent that in the code here. that depends on what result is in this case but it seems to be an ebus outcome which might have IsSuccess() so you could assert result.IsSuccess()

AutomatedTesting/Gem/PythonTests/Atom/tests/hydra_AtomGPU_AreaLightScreenshotTest.py
```diff
@@ -180,13 +201,20 @@ def AtomGPU_LightComponent_AreaLightScreenshotsMatchGoldenImages():
...
 
         # 15. Delete the Area Light entity.
         area_light_entity.delete()
         Report.result(Tests.area_light_entity_deleted, not area_light_entity.exists())
 
-        # 16. Look for errors.
+        # 16. Compare screenshots to golden images.
+        for screenshot_name, screenshot_threshold in zip(screenshot_names, screenshot_thresholds):
+            compResult = compare_screenshot_to_golden_image(FOLDER_PATH, screenshot_name, screenshot_name)
+            Report.result(("Screenshot comparison succeeded.", "Screenshot comparison failed."), compResult.result_code == azlmbr.utils.ImageDiffResultCode_Success)
```
smurly, line 213: include screenshot_name in your tuple strings here for failure so we can tell which one failed.
("Screenshot succeeded", f"Screenshot {screenshot_name} failed comparison")
also this is actually not comparison failed so much as comparison was not possible because result_code != azlmbr.utils.ImageDiffResultCode_Success
looks like other than success here is AZ_Error condition for mostly unsupported image conditions.

AutomatedTesting/Gem/PythonTests/Atom/tests/hydra_AtomGPU_AreaLightScreenshotTest.py
```diff
@@ -180,13 +201,20 @@ def AtomGPU_LightComponent_AreaLightScreenshotsMatchGoldenImages():
...
+            if compResult.result_code == azlmbr.utils.ImageDiffResultCode_Success:
+                Report.result((f"{screenshot_name} diff score {compResult.diff_score} under threshold {screenshot_threshold}.", f"{screenshot_name} diff score {compResult.diff_score} over threshold {screenshot_threshold}."), compResult.diff_score < screenshot_threshold)
```
smurly, line 215: line length here is making it hard to see the threshold in action
declare your tuple outside of the Report.result call:
```suggestion
                screen_compare_test = (
                    f"{screenshot_name} diff score {compResult.diff_score} under threshold {screenshot_threshold}.",
                    f"{screenshot_name} diff score {compResult.diff_score} over threshold {screenshot_threshold}.")
                Report.result(screen_compare_test, compResult.diff_score < screenshot_threshold)
```

AutomatedTesting/Gem/PythonTests/Atom/tests/hydra_AtomGPU_SpotLightScreenshotTest.py
```diff
@@ -233,9 +256,16 @@ def AtomGPU_LightComponent_SpotLightScreenshotsMatchGoldenImages():
...
         # 17. Enter game mode and take a screenshot then exit game mode.
-        enter_exit_game_mode_take_screenshot("SpotLight_6.ppm", Tests.enter_game_mode, Tests.exit_game_mode)
+        enter_exit_game_mode_take_screenshot(screenshot_names[5], Tests.enter_game_mode, Tests.exit_game_mode)
+
+        # 18. Compare the screenshots to golden images.
+        for screenshot_name, screenshot_threshold in zip(screenshot_names, screenshot_thresholds):
+            compResult = compare_screenshot_to_golden_image(FOLDER_PATH, screenshot_name, screenshot_name)
+            Report.result(("Screenshot comparison succeeded.", "Screenshot comparison failed."), compResult.result_code == azlmbr.utils.ImageDiffResultCode_Success)
+            if compResult.result_code == azlmbr.utils.ImageDiffResultCode_Success:
+                Report.result((f"{screenshot_name} diff score {compResult.diff_score} under threshold {screenshot_threshold}.", f"{screenshot_name} diff score {compResult.diff_score} over threshold {screenshot_threshold}."), compResult.diff_score < screenshot_threshold)
```
smurly, line 266: line length. consider breaking this up for readability

AutomatedTesting/Gem/PythonTests/Atom/tests/hydra_AtomGPU_BasicLevelSetup.py
```diff
@@ -299,9 +312,16 @@ def AtomGPU_BasicLevelSetup_SetsUpLevel():
...
         # 21. Enter/Exit game mode taking screenshot.
-        enter_exit_game_mode_take_screenshot(f"{SCREENSHOT_NAME}.ppm", Tests.enter_game_mode, Tests.exit_game_mode)
+        enter_exit_game_mode_take_screenshot(screenshot_names[0], Tests.enter_game_mode, Tests.exit_game_mode)
+
+        # 22. Compare the screenshots to golden images.
+        for screenshot_name, screenshot_threshold in zip(screenshot_names, screenshot_thresholds):
+            compResult = compare_screenshot_to_golden_image(FOLDER_PATH, screenshot_name, screenshot_name)
+            Report.result(("Screenshot comparison succeeded.", "Screenshot comparison failed."), compResult.result_code == azlmbr.utils.ImageDiffResultCode_Success)
+            if compResult.result_code == azlmbr.utils.ImageDiffResultCode_Success:
+                Report.result((f"{screenshot_name} diff score {compResult.diff_score} under threshold {screenshot_threshold}.", f"{screenshot_name} diff score {compResult.diff_score} over threshold {screenshot_threshold}."), compResult.diff_score < screenshot_threshold)
```
smurly, line 322: line length makes this hard to follow

smurly (commented):
seems like the ebus call should return an outcome with the IsSuccess() method exposed in python outcome.GetValue() can unpack the resulting diff score or a tuple of diffScore and filteredDiffScore. that would put the ebus call in line with other bus calls for use and function.

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/screenshot_utils.py
```diff
@@ -91,19 +92,23 @@ def wait_until_screenshot(self):
...
     helper.wait_for_condition(lambda: general.is_in_game_mode(), 2.0)
     general.log(f"{entity_name}_test: Entered game mode: {general.is_in_game_mode()}")
-    ScreenshotHelper(general.idle_wait_frames).capture_screenshot_blocking(f"{screenshot_name}.ppm")
+    ScreenshotHelper(general.idle_wait_frames).capture_screenshot_blocking(f"{screenshot_name}.png")
     general.idle_wait(1.0)
     general.exit_game_mode()
     helper.wait_for_condition(lambda: not general.is_in_game_mode(), 2.0)
     general.log(f"{entity_name}_test: Exit game mode: {not general.is_in_game_mode()}")
+
+def compare_screenshots(imageA, imageB):
```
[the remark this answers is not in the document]
jiaweig-amzn: It is documented.

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/screenshot_utils.py
jiaweig-amzn: This is actually a good suggestion. But I think if we want to change the comparison part, we have to change the whole `FrameCaptureSystemComponent` to match the same style. It contains `CaptureScreenshot` that returns a `FrameCaptureId`, and `BuildScreenshotFilePath` that returns a `string`. Those can be all put into outcome based on their usage.
It will expand the scope of this PR. I'll open a ticket to refactor this.

Kadino (approved):
While it reduces the contexts it can be used in, I think it is overall a good idea to have this image capture tool be Atom-specific.

Thanks for removing the old code!

smurly (approved):

PR #12906, opened 2022-10-31
Atom/antonmic/math pow

antonmic:
Added more PowX entries in Math.azsli


amzn-tommy (approved):

rgba16f (approved):

jeremyong-az:
Note that just using `pow` I believe should just be a single `v_exp_f32` instruction on AMD asics so it'd be interesting to understand where and if these helpers should be used

jeremyong-az (approved):

PR #12909, opened 2022-10-31
Asset Processor - Add critical and escalation values to InProgress status

amzn-mike:
## What does this PR do?

Indicates if in-progress jobs are critical or escalated.  Note that the escalation/critical status is not saved so this only applies to in-progress jobs.

## How was this PR tested?

Manually ran AP and reprocessed assets to view critical status.
Ran editor with clean cache and observed AP status.

image


AMZN-stankowi (approved):
I like the concept a lot, I think some UX polish will go a long way here. Probably a different in-progress icon for critical?

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.h
```diff
@@ -290,7 +290,7 @@ namespace AssetProcessor
         virtual void AssessDeletedFile(QString filePath);
         void OnAssetScannerStatusChange(AssetProcessor::AssetScanningStatus status);
         void FinishAssetScan();
-        void OnJobStatusChanged(JobEntry jobEntry, JobStatus status);
+        void OnJobStatusChanged(JobEntry jobEntry, JobStatus status, int escalation, bool critical);
```
lmbr-pip, line 293: Nit: Consider comments on public methods, unclear what "escalation means" here.

As a thought experiment, are there other statuses we may care about in the future ie should the bool by an enum?

amzn-mike: Added a comment on the signal (this is the slot).  I think bool is fine for the critical because this comes directly from `JobDetails` where critical is stored as a separate bool.

rainbj:
I just took a look at this new status.

The concept make sense. However, it feels odd to mix this current In progress with escalated or critical. Feels like there should be different columns. It could read to a user that this is a critical Asset OR a critically failed asset not yet done processing

My question would be that what use case is this supposed to solve? As a user, I marked an asset to be "critical" when running through AP, so now I want to see those assets first?

sptramer:
From a docs perspective: LGTM. However I agree with @rainbj's concern about mixing "In progress" with "escalation" and "critical" - it feels like maybe what we want isn't in the "Status" column, but part of the description? For example:

`` In Progress: (Priority %n, `[escalated ? "escalation" : ""]`) `[Current message]` ``

Code/Tools/AssetProcessor/native/resourcecompiler/JobsModel.cpp
```diff
@@ -335,7 +335,13 @@ namespace AssetProcessor
                 return message;
             }
             case JobStatus::InProgress:
-                return tr("InProgress");
+                QString message = tr("InProgress");
+                QString extra;
+
+                Append(message, critical ? "Critical" : "", " ");
+                Append(message, escalation != 0 ? QStringLiteral("(%1)").arg(escalation) : "", " ");
```
sptramer, line 342: It feels like we don't want the `Critical` or `Escalation` as part of the status - we want it in the message itself. That way users can filter on Status for all In-Progress jobs without assigning them a specific status code based on multiple priorities, and then they can further filter (or only filter) on messages containing a `(Critical)` or `(High-priority)` marker.

sptramer (commented):
Called out in the code where status messages are set, regarding my previous comment, just to make it official and mark me as a reviewer.

Code/Tools/AssetProcessor/native/resourcecompiler/JobsModel.cpp
amzn-mike: I'm not sure I follow.  There's no description column and the message field here is just the name of the variable which holds the data that's displayed in the Status column.  The filter system only works on the status column, so if we move this info to another column you wouldn't be able to filter on it.  I'm not sure how useful it is to filter beyond "InProgress" though, its difficult to work with the job list while AP is processing things.  If you filter to InProgress and try to find some file or set of files, the list will be constantly moving and rearranging, with items appearing and disappearing rapidly.

I can move this status info into its own column easy enough, just realize the info is not stored anywhere so once a job exits "InProgress" status, the column will go blank, which could be confusing to users.

sptramer: Understood - what I'm asking for may be a future improvement then, rather than something that can be achieved right now. This was more of an "ideal state", meaning we may want to add a form of "priority" column or report on current asset processing status via live updates in a message area. @yuyihsu @amzn-leenguy  is this already a known area of improvement for UX as part of our asset processing message initiative?

PR #12914, opened 2022-10-31
Action Manager | Fixes to Open Recent menu displaying invalid levels

AMZN-daimini:
## What does this PR do?

Introduces fixes to the `Open Recent...` menu with the new Action Manager enabled.

- In the previous implementation, the full level path would be displayed instead of the display version;
- Fixed the list generation to skip levels that do not belong to the current project;
- Fixed index number to show the correct count (bug that is still in the legacy implementation).

Verified that the menu items open the correct files (had to refactor the code since now there is no longer 1:1 correspondence between the indices).

Legacy menu (new action manager flag off)
image

New menu (new action manager flag on)
image

(Note, the ordering of the levels is different due to testing, but the behavior is correct).

## How was this PR tested?
Manual testing.

### DIFF
--- a/Code/Editor/Core/EditorActionsHandler.cpp
+++ b/Code/Editor/Core/EditorActionsHandler.cpp
@@ -33,8 +33,10 @@
 #include <QtViewPaneManager.h>
 #include <ToolBox.h>
 #include <ToolsConfigPage.h>
+#include <Util/PathUtil.h>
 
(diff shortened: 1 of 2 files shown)


AMZN-daimini:
Fixes #12914 12526

cgalvan (approved):

amzn-tmryan (approved):

PR #12921, opened 2022-11-01
[Terrain] Add autosave options to the Image Gradient

mbalfour-amzn:
## What does this PR do?

This adds the concept of "save modes" to the Image Gradient. 
* "Auto save" (the default) will automatically save the source asset after every paint session is exited, though for the first save of that asset after level load, it will prompt the user for a save location. This is done to ensure that the user is deliberately choosing the destructive action of overwriting a source asset instead of just letting it happen in a way that can't be undone.
* "Auto save with incrementing names" will automatically append an incrementing number at the end of the file name on each save. It won't prompt the user for a save location unless the generated name will overwrite an existing asset.
* "Save as..." will prompt the user for a save location on every save.

There are a few other small cleanups / bugfixes in here as well:
* Removed the Dependency Monitor from ImageGradientComponent because it wasn't really serving any useful purpose. The component already subscribes to all of the buses that affect it.
* Removed the "default" label on the dropdown enums from UX feedback that it wasn't something users truly needed to know.
* Fixed painting bug in the case where the paintbrush tries to paint in a location where the image doesn't exist.
* Split the Enable/Disable Component Mode logic into separate methods in preparation for an eventual bugfix to the component switcher where the component mode shouldn't always be active and visible.

## How was this PR tested?

Tried editing images with all three save modes and verified that they behaved as expected.


Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponent.cpp
```diff
@@ -563,11 +706,47 @@ namespace GradientSignal
...
+                    watchFolder.c_str(), assetInfo.m_relativePath.c_str(), sourcePath, true);
+
+                if (success)
+                {
+                    return sourcePath;
+                }
+            }
+        }
+
+        // Invalid image asset or failed path creation, try creating a new name.
```
cgalvan, line 737: Should we log a warning if we get to this point?

cgalvan (approved):

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponent.cpp
mbalfour-amzn: We can end up here through the "Create Image" flow with an intentionally-invalid image asset, so we shouldn't log a warning. I _think_ we already log an error in other cases where we could get here, but I'll examine the code further and see if there's anywhere else that needs warnings.

PR #12926, opened 2022-11-01
[Terrain] Add specialization of GetImageDataPixelValue<> for AZ::Color.

mbalfour-amzn:
## What does this PR do?

We currently have a GetImageDataPixelValue<> API that's specialized for floats and ints that return a single channel of data from the pixel. This adds a third specialization for AZ::Color that returns all of the RGBA channels at once.

The main changes below are the addition of RetrieveColorValue() and changing GetBlockColor() in BC1Block to return an AZ::Color instead of a single float. The ScaleSNorm* methods are just code that was extracted from the existing Retrieve*Value() methods so that it could be reused for the Color specialization. The BC1Block already existed but for some reason is showing up as all new in the diff.

The rest of the changes below are just automatic clang formatting cleanups.

## How was this PR tested?

This was tested locally with some experimental code for painting on terrain macro color materials in the Editor. This specialization makes painting nearly 4x faster than trying to do the same type of pixel color extraction through 4 calls per-pixel to the previously-existing APIs.


Gems/Atom/RPI/Code/Source/RPI.Public/RPIUtils.cpp
```diff
@@ -110,6 +110,26 @@ namespace AZ
...
             }
 
+            float ScaleSNorm8Value(float value)
+            {
+                // Scale the value from AZ::s8 min/max to -1 to 1
+                // We need to treat -128 and -127 the same, so that we get a symmetric
+                // range of -127 to 127 with complementary scaled values of -1 to 1
+                constexpr float signedMax = static_cast<float>(std::numeric_limits<AZ::s8>::max());
+                constexpr float signedMin = -signedMax;
+                return ScaleValue(AZStd::max(value, signedMin), signedMin, signedMax, -1.0f, 1.0f);
```
VickyAtAZ, line 120: I wonder why you only make sure the value not smaller than signedMin but not 'not larger than signedMax' since the input value type is float.

Maybe it's better to use int8_t as input rather than float.

Gems/Atom/RPI/Code/Source/RPI.Public/RPIUtils.cpp
```diff
@@ -110,6 +110,26 @@ namespace AZ
                 return ((value - origMin) / (origMax - origMin)) * (scaledMax - scaledMin) + scaledMin;
```
VickyAtAZ, line 110: Maybe add assert to make sure value >= origMin and value >= origMax. Otherwise you would need to trim it.

Gems/Atom/RPI/Code/Source/RPI.Public/RPIUtils.cpp
```diff
@@ -110,6 +110,26 @@ namespace AZ
...
+            }
+
+            float ScaleSNorm16Value(float value)
+            {
+                // Scale the value from AZ::s16 min/max to -1 to 1
+                // We need to treat -32768 and -32767 the same, so that we get a symmetric
+                // range of -32767 to 32767 with complementary scaled values of -1 to 1
+                constexpr float signedMax = static_cast<float>(std::numeric_limits<AZ::s16>::max());
+                constexpr float signedMin = -signedMax;
+                return ScaleValue(AZStd::max(value, signedMin), signedMin, signedMax, -1.0f, 1.0f);
```
VickyAtAZ, line 130: Same as above

Gems/Atom/RPI/Code/Source/RPI.Public/RPIUtils.cpp
```diff
@@ -139,86 +248,277 @@ namespace AZ
...
+                            1.0f);
+                    }
+                case AZ::RHI::Format::R16G16B16A16_UNORM:
+                    {
+                        auto actualMem = reinterpret_cast<const AZ::u16*>(mem);
+                        return AZ::Color(
+                            actualMem[indices.first + 0] / static_cast<float>(std::numeric_limits<AZ::u16>::max()),
+                            actualMem[indices.first + 1] / static_cast<float>(std::numeric_limits<AZ::u16>::max()),
+                            actualMem[indices.first + 2] / static_cast<float>(std::numeric_limits<AZ::u16>::max()),
+                            actualMem[indices.first + 3] / static_cast<float>(std::numeric_limits<AZ::u16>::max()));
```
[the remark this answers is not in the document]
mbalfour-amzn: Whoops, these were cut-n-pasted from other existing std::numeric_limits lines in the file. I'll replace them all.

Gems/Atom/RPI/Code/Source/RPI.Public/RPIUtils.cpp
mbalfour-amzn: Fair point, I didn't change the existing logic here, but I _did_ put it in a function that takes in a float. I'll change the input types to int8_t and int16_t.

invertednormal (approved):
Good improvement!

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/BlockCompression.h
```diff
@@ -48,56 +48,59 @@ namespace AZ
...
+                    // Index 3 uses 1/3 of color 0 and 2/3 of color 1
+                    return AZ::Color(
+                        ((((m_color0 >> 11) & 0x1F) / aznumeric_cast<float>(0x1F)) * (1.0f / 3.0f)) +
+                            ((((m_color1 >> 11) & 0x1F) / aznumeric_cast<float>(0x1F)) * (2.0f / 3.0f)),
+                        ((((m_color0 >> 5) & 0x3F) / aznumeric_cast<float>(0x3F)) * (1.0f / 3.0f)) +
+                            ((((m_color1 >> 5) & 0x3F) / aznumeric_cast<float>(0x3F)) * (2.0f / 3.0f)),
+                        ((((m_color0 >> 0) & 0x1F) / aznumeric_cast<float>(0x1F)) * (1.0f / 3.0f)) +
+                            ((((m_color1 >> 0) & 0x1F) / aznumeric_cast<float>(0x1F)) * (2.0f / 3.0f)),
+                        1.0f);
+                    break;
```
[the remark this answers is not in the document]
mbalfour-amzn: Love it, thanks for the suggestion!

Gems/Atom/RPI/Code/Source/RPI.Public/RPIUtils.cpp
VickyAtAZ: sounds good.

VickyAtAZ (approved):

PR #12933, opened 2022-11-01
fix mistake in double precision quaternion conjugate

greerdv:

## What does this PR do?
Fixes an error where the implementation of GetConjugate for double precision quaternions was returning the negative of the conjugate. The negative of a quaternion is in some senses an equivalent rotation, so I don't think this would have had a big impact, and doesn't seem to have affected the automatic joint fitting (which is the only feature using the double precision quaternion), but it was technically incorrect.

## How was this PR tested?
Ran unit tests. Tested automatic joint limit fitting in the animation editor.


moraaar (approved):

hultonha:
Nice find! 🙂

PR #12936, opened 2022-11-01
Material Canvas: Replace dynamic node supported type list with regex, matrix transform nodes, other improvements

gadams3:
## What does this PR do?

102 of the files are data files that can be glanced over because the same change is repeated throughout.

This PR changes the dynamic node slot configuration class to specify compatible data types using a regular expression instead of an explicit list. This is much more compact and flexible for selecting multiple data types and dimensions. It will also allow new data types to be automatically supported if they are registered with names that match the regular expression. For example, currently there are only a subset of floating-point matrix types registered. If new types come online that match the ranges for allowed dimensions, then they will immediately be supported.

Improvements have been made to the dynamic node slot configuration layout and property editor. There is no longer a need to manually clear or select the default value. Updating the regular expression or default data type will automatically update affected fields and change the default value if it does not match the current supported types. This validation is also done when loading the node configuration.

This includes some clean up to a few graph model classes. Part of that was the removal of cashed connections from the slot glass. Instead, they will be looked up directly from the graph object. This may be less efficient but it is cleaner, easier to read, and less error prone.

Most of the remaining changes are repeated data changes for the switch to regular expressions.

## How was this PR tested?

Verified that all of the existing test graphs continue to open and generate materials without error.

Verified that the regular expression allows bool and integer types to be connected to floating point scalars and vectors.

Verified that matrix and vector nodes can be connected to transform nodes and changes are visualized.

image


cgalvan (approved):

Gems/Atom/Tools/AtomToolsFramework/Code/Include/AtomToolsFramework/DynamicNode/DynamicNodeSlotConfig.h
```diff
@@ -34,11 +35,19 @@ namespace AtomToolsFramework
             const AZStd::string& displayName,
             const AZStd::string& description,
             const AZStd::any& defaultValue,
-            const AZStd::vector<AZStd::string>& supportedDataTypes,
+            const AZStd::string& supportedDataTypeRegex,
             const DynamicNodeSettingsMap& settings);
         DynamicNodeSlotConfig() = default;
         ~DynamicNodeSlotConfig() = default;
 
+        AZ::Crc32 ValidateDataTypes();
```
santorac, line 43: I'd like a comment about what this returns.

Gems/Atom/Tools/AtomToolsFramework/Code/Include/AtomToolsFramework/DynamicNode/DynamicNodeSlotConfig.h
```diff
@@ -34,11 +35,19 @@ namespace AtomToolsFramework
...
+        AZStd::any GetDefaultValue() const;
+        AZStd::string GetDefaultDataTypeName() const;
+        GraphModel::DataTypePtr GetDefaultDataType() const;
+        AZStd::vector<AZStd::string> GetSupportedDataTypeNames() const;
```
santorac, line 47: I'd like a comment about how this relates to supportedDataTypeRegex that's passed to the constructor.

Gems/Atom/Tools/AtomToolsFramework/Code/Source/DynamicNode/DynamicNodeManager.cpp
```diff
@@ -161,25 +162,24 @@ namespace AtomToolsFramework
...
-                    "DynamicNodeConfig \"%s\" could not be validated because DynamicNodeSlotConfig \"%s\" references unregistered data type \"%s\".",
-                    configId.ToFixedString().c_str(),
-                    slotConfig.m_displayName.c_str(),
-                    dataTypeName.c_str());
-                return false;
-            }
+            AZ_Error(
+                "DynamicNodeManager",
+                false,
+                "DynamicNodeConfig \"%s\" could not be validated because DynamicNodeSlotConfig \"%s\" does not match any rich to data "
```
santorac, line 178: Looks like a mistake in "any rich to data types", I don't understand.

Gems/Atom/Tools/AtomToolsFramework/Code/Source/DynamicNode/DynamicNodeSlotConfig.cpp
```diff
@@ -97,75 +95,124 @@ namespace AtomToolsFramework
...
-        DynamicNodeManagerRequestBus::BroadcastResult(registeredDataTypes, &DynamicNodeManagerRequestBus::Events::GetRegisteredDataTypes);
+        const auto& supportedDataTypes = GetSupportedDataTypes();
+        for (const auto& dataType : supportedDataTypes)
+        {
+            if (dataType->GetDisplayName() == m_defaultDataType)
+            {
+                return m_defaultDataType;
+            }
+        }
+        return !supportedDataTypes.empty() ? supportedDataTypes.front()->GetDisplayName() : AZStd::string();
```
santorac, line 175: Can this just call GetDefaultDataType()->GetDisplayName()?

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Inspector/PropertyWidgets/PropertyStringBrowseEditCtrl.cpp
```diff
@@ -64,7 +64,7 @@ namespace AtomToolsFramework
     {
         if (attrib == AZ_CRC_CE("LineEditReadOnly"))
         {
-            if (bool value = true; attrValue->Read<bool>(value))
+            if (bool value = true; attrValue->Read<decltype(value)>(value))
```
santorac, line 67: Do you even need the template specification? I think it can infer the type.
```suggestion
            if (bool value = true; attrValue->Read(value))
```

santorac (approved):

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Inspector/PropertyWidgets/PropertyStringBrowseEditCtrl.cpp
gadams3: It may not be required. I just changed the model to be consistent. I've been bitten by that before after submitting changes to AR for different platforms. Better safe than sorry.

Gems/Atom/Tools/AtomToolsFramework/Code/Source/DynamicNode/DynamicNodeSlotConfig.cpp
gadams3: Yes! It's always simplified compared to what I had before but that's yet another simplification.

PR #12945, opened 2022-11-01
Fix for ReflectionProbes on transparent meshes

dmcdiarmid-ly:
Moved the ReflectionProbeFeatureProcessor update of the MeshFeatureProcessor to OnRenderEnd to avoid a race in Simulate.

Tested Editor DX12/Vulkan


rgba16f (approved):

gadams3 (approved):

invertednormal (approved):
Frame of latency is probably ok with this kind of feature, but it highights that it'd be nice to have the concept of feature processor dependencies to enforce some ordering between updates when required.

Also, it seems like this could get quite expensive on large scenes - telling 10,000 meshes to update their object srgs isn't cheap, especially when those srgs are recalculating all of their indices and updating the entire srg. Might be better to move the affected data out of the object srg in the future into a buffer referenced by ID.

dmcdiarmid-ly:
> Frame of latency is probably ok with this kind of feature, but it highights that it'd be nice to have the concept of feature processor dependencies to enforce some ordering between updates when required.

Yes in this case it's not a major issue, any changes to Reflection Probes are usually from editor actions or something where a frame delay won't be noticed.  For example, a time of day cubemap change.

We've discussed the dependency issue in the past, and it's still an open topic.  Right now there isn't a great way to solve a Simulate() race between FPs.

> Also, it seems like this could get quite expensive on large scenes - telling 10,000 meshes to update their object srgs isn't cheap, especially when those srgs are recalculating all of their indices and updating the entire srg. Might be better long term to move the affected data out of the object srg in the future into a buffer referenced by ID.

It's actually not quite as expensive as it appears - it only affects meshes that are using forward-pass Reflection Probes.  Most meshes use the Reflections pass, so typically this will only affect a small subset of meshes.

PR #12946, opened 2022-11-01
fix issue with not all vertices in icosphere getting normalized

greerdv:

## What does this PR do?
Fixes an issue where some vertices from GenerateIcoSphere were not normalized to unit radius.

## How was this PR tested?
Added a unit test to check that all vertices in the icosphere have unit radius.


mbalfour-amzn (approved):
Thanks for finding and fixing this!

hultonha (approved):
Great fix @greerdv and awesome work on the test too! 🙂

PR #12950, opened 2022-11-01
[Terrain] Change underlying paint brush settings code to support color.

mbalfour-amzn:
## What does this PR do?

This replaces intensity/opacity in the paintbrush APIs with a single AZ::Color value, and adds a paintbrush color mode so that each component mode can specify whether they want linear color, SRGB, or greyscale. 

The paintbrush settings themselves will show either a color picker or an intensity slider depending on whether color or greyscale is selected. The opacity slider remains as well for better UX usability, even though the opacity value just gets merged into the AZ::Color from an API perspective.

The Image Gradient painting code only uses the greyscale mode, nothing currently uses the color mode.

## How was this PR tested?

The color mode was tested in a local branch in which the Terrain Macro Material supports color painting. I was able to use this to verify that the color picking works for component modes that support color, and intensity works for component modes that support greyscale.


invertednormal (approved):
Looks good! 😄

cgalvan (approved):

PR #12953, opened 2022-11-01
Averaging metrics per save period

AMZN-Olex:
Signed-off-by: Olex Lozitskiy <5432499+AMZN-Olex@users.noreply.github.com>

## What does this PR do?

Changes network metrics to save average values over the save period, which is configured with `bg_networkingMetricCollectionPeriod`. Previously, the value would be average over a set number of items instead, which didn't always correspond with the collection period.

## How was this PR tested?

Currently testing in o3de-multiplayersample


Gems/Multiplayer/Code/Source/MultiplayerStatSystemComponent.h
```diff
@@ -58,8 +58,11 @@ namespace Multiplayer
 
         struct CumulativeAverage
         {
+            using Type = AZ::IO::AverageWindow<double, double, AZ::IO::s_statisticsWindowSize>;
```
AMZN-puvvadar, line 61: I'm not sure how I feel about Type as a name.

AMZN-Olex: Fair, updated.

Gems/Multiplayer/Code/Source/MultiplayerStatSystemComponent.cpp
```diff
@@ -221,9 +222,21 @@ namespace Multiplayer
                     AZStd::for_each(
```
kberg-amzn, line 222: Why not just `for (auto& stat : group.m_stats.m_items)`. Ever since ranged for came into the standard, I'm a bit confused why people still use for_each.

Gems/Multiplayer/Code/Source/MultiplayerStatSystemComponent.h
```diff
@@ -58,9 +58,11 @@ namespace Multiplayer
 
         struct CumulativeAverage
         {
+            using AverageWindowType = AZ::IO::AverageWindow<double, double, AZ::IO::s_statisticsWindowSize>;
+
             AZStd::string m_name;
-            AZ::IO::AverageWindow<double, double, AZ::IO::s_statisticsWindowSize> m_average;
-            AZ::u64 m_sampleCount = 0;
+            AverageWindowType m_average;
+            double m_lastValue = 0.f;
```
kberg-amzn, line 65: Drop the .f for a double.. '0.0' or '0.'. Adding the '.f' means you explicitly want a single precision constant.

kberg-amzn (approved):

AMZN-Gene (approved):

Gems/Multiplayer/Code/Source/MultiplayerStatSystemComponent.h
AMZN-Olex: Done

PR #12957, opened 2022-11-02
Adding a `ParseConfigFile` function to AzCore

lumberyard-employee-dm:
The `ParseConfigFile` function can parse a Windows style INI file without any dynamic memory allocations and is suitable to be used in early application startup before O3DE allocators are available.

The SettingsRegistryMergeUtils MergeSettingsToRegistry_ConfigFile function has been updated to use the ParseConfigFile function.

`ParseConfigFile` can can be used for example to parse .ini files containing VFS settings for connecting to the AssetProcessor on Android or allocator configuration settings

Added a fix to the Comment Prefix Filter functions. If a line contained multiple comment start characters such as `";#"` and a commented line of `# This is a comment; More text`, then the Comment Prefix Filter function would only remove up to the first comment start character that was found (in this case it remove up to the semicolon and not the number sign)

## How was this PR tested?

Added UnitTest to validate the ParseConfigFile logic to ConfigParserTests.cpp
Also validated that the SettingsRegistryMergeUtilsTest that read configuration files also suceeded

### DIFF
--- a/Code/Framework/AzCore/AzCore/Settings/ConfigParser.cpp
+++ b/Code/Framework/AzCore/AzCore/Settings/ConfigParser.cpp
@@ -1,0 +1,211 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#include <AzCore/IO/GenericStreams.h>
+#include <AzCore/Settings/ConfigParser.h>
+#include <AzCore/std/string/conversions.h>
+#include <AzCore/std/string/fixed_string.h>
+#include <AzCore/std/containers/fixed_vector.h>
+#include <AzCore/StringFunc/StringFunc.h>
+
+namespace AZ::Settings
+{
+    AZStd::string_view ConfigParserSettings::DefaultCommentPrefixFilter(AZStd::string_view line)
+    {
+        constexpr AZStd::string_view commentPrefixes = ";#";
+        for (char commentPrefix : commentPrefixes)
+        {
+            if (size_t commentOffset = line.find(commentPrefix); commentOffset != AZStd::string_view::npos)
+            {
+                line = line.substr(0, commentOffset);
+            }
+        }
+
+        return line;
+    }
+
+    AZStd::string_view ConfigParserSettings::DefaultSectionHeaderFilter(AZStd::string_view line)
+    {
+        AZStd::string_view sectionName;
+        constexpr char sectionHeaderStart = '[';
+        constexpr char sectionHeaderEnd = ']';
+        if (line.starts_with(sectionHeaderStart) && line.ends_with(sectionHeaderEnd))
+        {
+            // View substring of line between the section '[' and ']' characters
+            sectionName = line.substr(1);
+            sectionName = sectionName.substr(0, sectionName.size() - 1);
+            // strip any leading and trailing whitespace from the section name
+            size_t startIndex = 0;
+            for (; startIndex < sectionName.size() && isspace(sectionName[startIndex]); ++startIndex);
+            sectionName = sectionName.substr(startIndex);
+            size_t endIndex = sectionName.size();
+            for (; endIndex > 0 && isspace(sectionName[endIndex - 1]); --endIndex);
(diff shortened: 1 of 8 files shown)


Code/Framework/AzCore/AzCore/Settings/ConfigParser.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+        ParseOutcome parseOutcome;
+        if (!configParserSettings.m_parseConfigEntryFunc)
+        {
+            return AZStd::unexpected(ParseErrorString::format("The Parse Config Entry function is not valid for parsing a Windows Style INI line"));
+        }
+
+        // The ConfigFile is parsed using into a fixed_vector of ConfigBufferMaxSize below
+        // which avoids performing any dynamic memory allocations during parsing
+        // Only the user supplied callback functions as part of the ConfigParserSettings can potentailly allocate memory
```
cgalvan, line 92:
```suggestion
        // The ConfigFile is parsed into a fixed_vector of ConfigBufferMaxSize below
        // which avoids performing any dynamic memory allocations during parsing
        // Only the user supplied callback functions as part of the ConfigParserSettings can potentially allocate memory
```

cgalvan (approved):

Code/Framework/AzCore/AzCore/Settings/ConfigParser.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+        return line;
+    }
+
+    AZStd::string_view ConfigParserSettings::DefaultSectionHeaderFilter(AZStd::string_view line)
+    {
+        AZStd::string_view sectionName;
+        constexpr char sectionHeaderStart = '[';
+        constexpr char sectionHeaderEnd = ']';
+        if (line.starts_with(sectionHeaderStart) && line.ends_with(sectionHeaderEnd))
```
moraaar, line 37: When I reached this code I wondered what happens if there are trailing spaces, then I saw in the parsing loop that for each line the trailing spaces are removed before calling this function. Maybe it'd be good to add a small clarification note here indicating that the line received by argument won't have trailing spaces.

Code/Framework/AzCore/AzCore/Settings/ConfigParser.cpp
```diff
@@ -0,0 +1,211 @@
...
+        // Only the user supplied callback functions as part of the ConfigParserSettings can potentially allocate memory
+        size_t remainingFileLength = configStream.GetLength();
+        if (remainingFileLength == 0)
+        {
+            return AZStd::unexpected(ParseErrorString::format(R"(Config stream "%s" is empty, nothing to merge)" "\n", configStream.GetFilename()));
+        }
+
+        constexpr size_t ConfigBufferMaxSize = 4096;
+        AZStd::fixed_vector<char, ConfigBufferMaxSize> configBuffer;
+        size_t readSize = (AZStd::min)(configBuffer.max_size(), remainingFileLength);
```
moraaar, line 101: maybe throw a warning if `configBuffer` is larger than the max size to let the user known that some parts of the stream won't be read.

moraaar (approved):
Sorry for the wait. This looks great @lumberyard-employee-dm. Thanks for adding this. I've got a few minor comments, but nothing blocking. Approved.

Code/Framework/AzCore/AzCore/Settings/ConfigParser.cpp
lumberyard-employee-dm: Well that is the normal part of line parsing that leading and trailing whitespace are not signficant or kept.
I'll add a comment noting as such

Code/Framework/AzCore/AzCore/Settings/ConfigParser.cpp
lumberyard-employee-dm: We can't actually throw a warning here.
This code is meant to be used in early startup before the allocators and trace system are available for reading startup files.

That is why this function returns an Outcome.

The preconditions of this function is that a single line cannot be larger than 4096 and that is documented in the function [declaration](https://github.com/o3de/o3de/pull/12957/files/66f05b8be6c8614771eed9531901d53f0b270a59#diff-ccedb6a20bd3fcb1df6f384991b4c3046b1b41d1ddc4fac09edbeb24fffa74f3R129-R130).

It is fine if the config file is larger than 4096 because it acts like a sliding 4096 byte window

PR #12964, opened 2022-11-02
Prefab/overrides/visualization v1

srikappa-amzn:
## What does this PR do?

This PR merges previously approved below PRs into development:
https://github.com/aws-lumberyard-dev/o3de/pull/458
https://github.com/aws-lumberyard-dev/o3de/pull/461
https://github.com/aws-lumberyard-dev/o3de/pull/466

RFC: https://github.com/o3de/sig-content/issues/75

## How was this PR tested?

- New unit tests added
- Existing unit tests and automated tests passed in AR
- Additional manual testing done to make sure nothing else is broken around overrides.


Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabUiHandler.cpp
```diff
@@ -234,6 +242,29 @@ namespace AzToolsFramework
...
         }
 
+        if (descendantIndex.column() == EntityOutlinerListModel::ColumnName)
+        {
+            AZ::EntityId descendantEntityId = GetEntityIdFromIndex(descendantIndex);
+
+            // Container entities will always have overrides because they need to maintain unique positions in the scene.
+            // We are skipping checking for overrides on container entities for this reason.
+            if (!m_prefabPublicInterface->IsInstanceContainerEntity(descendantEntityId) &&
+                m_prefabOverridePublicInterface->AreOverridesPresent(descendantEntityId))
```
AMZN-daimini, line 252: nit: Which of these calls is cheaper? We could call it first for a minor performance gain.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabUiHandler.cpp
```diff
@@ -234,6 +242,29 @@ namespace AzToolsFramework
...
+            {
+                // Build the rect that will be used to paint the icon
+                QRect overrideRectangle =
+                    QRect(option.rect.topLeft() + s_overrideIconOffset, QSize(s_overrideIconSize * 2, s_overrideIconSize * 2));
+
+                painter->save();
+                painter->setRenderHint(QPainter::Antialiasing, true);
+                painter->setPen(Qt::NoPen);
+                painter->setBrush(s_overrideIconBackgroundColor);
+                painter->drawEllipse(overrideRectangle.center(), s_overrideIconSize, s_overrideIconSize);
```
AMZN-daimini, line 262: Is it a rectangle or an ellipse? :)

If it's not painted and just used for sizing, maybe we could call it `overrideIconBounds` or something clearer.

AMZN-daimini (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceToTemplateInterface.h
```diff
@@ -46,6 +46,11 @@ namespace AzToolsFramework
             //! @return The string matching the path to the entity alias
             virtual AZStd::string GenerateEntityAliasPath(AZ::EntityId entityId) = 0;
 
+            //! Generates a path to the entity matching the id from the focused prefab.
+            //! @param entityId The entity id to fetch the path for.
+            //! @return The path to the entity with the given id.
+            virtual AZ::Dom::Path GenerateEntityPathFromFocusedPrefab(AZ::EntityId entityId) = 0;
```
forkercat, line 52: question: Want to know more about your thoughts of using AZ::Dom::Path here instead of AZStd::string. One point I may already notice is with this Path type we can easily skip an instance from top in the path. Are there other advantages over string?

forkercat (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceToTemplateInterface.h
srikappa-amzn: A couple of advantages actually:
1. A string can be any combination of literals whereas with a Dom::Path, we know that the format is going to be a valid path that can be used for looking up in DOMs. It also provides easy-to-use functions like push() and pop() whereas with strings, you would need to find the first '/', do a lot of substring math.
2. AZ::Dom::Path internally uses AZ::Name to store the path tokens and AZ::Name internally uses a dictionary that hashes and ref counts used strings. So it generally results in faster copying and comparing operations.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabUiHandler.cpp
srikappa-amzn: Good call. The IsInstanceContainerEntity will be the cheaper one because its just 2 function calls to lookup whereas the AreOverridesPresent() call needs to check 'm' components within an entity and 'n' properties within an entity to see if any overrrides exist. So the order here would be the optimized one.

forkercat (approved):

PR #12967, opened 2022-11-02
Updated font schema & missing dependency test to work on Linux and Windows

AMZN-stankowi:
Updated font schema & missing dependency test to work on Linux and Windows

## What does this PR do?

Updated font schema & missing dependency test to work on Linux and Windows.

This test was failing on Linux for two reasons: The way it was setting up the test environment resulted in file write permission issues, so the Asset Processor was failing to process the font files. The search pattern was failing in the asset database.

This update works around both issues, and makes the test run faster on Windows and Linux.

## How was this PR tested?

Ran this on Windows and Linux, the test passes with this change, fails without on Linux and passes without on Windows.

### DIFF
--- a/AutomatedTesting/Gem/PythonTests/assetpipeline/asset_processor_tests/asset_processor_batch_dependency_tests2.py
+++ b/AutomatedTesting/Gem/PythonTests/assetpipeline/asset_processor_tests/asset_processor_batch_dependency_tests2.py
@@ -17,7 +17,6 @@ from ..ap_fixtures.asset_processor_fixture import asset_processor as asset_proce
 from ..ap_fixtures.ap_setup_fixture import ap_setup_fixture as ap_setup_fixture
 
 
-
 # Import LyShared
 from automatedtesting_shared import file_utils as file_utils
 from ly_test_tools.o3de.ap_log_parser import APLogParser, APOutputParser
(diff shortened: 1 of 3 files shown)


AutomatedTesting/Gem/PythonTests/assetpipeline/asset_processor_tests/asset_processor_batch_dependency_tests2.py
```diff
@@ -44,53 +43,71 @@ def local_resources(request, workspace, ap_setup_fixture):
...
+class TestsAssetProcessorBatch_DependencyTests(object):
     """
     AssetProcessorBatch Dependency tests
     """
     @pytest.mark.test_case_id("C18108049")
     @pytest.mark.BAT
     @pytest.mark.assetpipeline
-    def test_WindowsMacPlatforms_SchemaGem_BatchLoadsSchemaFromGem(
-            self, asset_processor, workspace):
+    def test_WindowsMacPlatforms_SchemaGem_BatchLoadsSchema(
```
jackalbe, line 53: the naming seems to indicate "test_WindowsMacPlatforms" that it is only for Windows and Mac... but this is supposed to fix the Linux issues. Maybe this test should be renamed to "test_SchemaGem_BatchLoadsSchema" ?

jackalbe (approved):

AutomatedTesting/Gem/PythonTests/assetpipeline/asset_processor_tests/asset_processor_batch_dependency_tests2.py
AMZN-stankowi: Good point, I've removed windows & mac from the test name

amzn-mike (approved):

junbo75 (approved):

Issue #9669, opened 2022-05-19
As a Dev I want to the Know which Version of the Engine my Game is Running


AMZN-Gene:
**Is your feature request related to a problem? Please describe.**
Connect a client to a host was failing due to both client and server using different versions of the engine. 

**Describe the solution you'd like**
I want a simple way to compare the engine versions when running a gamelauncher or serverlauncher. I also want to see a warning when versions are different. O3DE needs a generic versioning solution.

svitlani:
This should be a parent issue for https://github.com/o3de/o3de/issues/9041 and will be addressed in one effort

AMZN-Gene:
blocked by https://github.com/o3de/sig-core/issues/44

lmbr-pip:
Plan is for networking will propose own RFC + APIs for identifying the network boundary version. 

The challenge for networking is ensuring that the binary serial protocols are the same between client  and server (ie different autocomponent versions may be incompatible).

PR #12975, opened 2022-11-02
Multiplayer Component Version Mismatch Detection

AMZN-Gene:
- Server will detect when a connecting client has different multiplayer components and will send back a list of all its components to compare what has been modified, added, or removed.
- A connection with a mismatch will be disconnected once all the components are synced up
- After disconnecting the client can be updated to match the server and reconnect without the server shutting down
- Unit Tests

image

Part 1 implementation for #9669 and https://github.com/o3de/sig-network/issues/81 
Upcoming PR
- Events when version mismatch occurs so that game-specific logic can be ran
- On-screen debug information
 
## How was this PR tested?
1. Modify a multiplayer component, add a multiplayer component on client, add a multiplayer component on server and ensure error logs are seen containing information about which components are mismatched.


Gems/Multiplayer/Code/Source/Components/MultiplayerComponentRegistry.cpp
```diff
@@ -70,8 +75,31 @@ namespace Multiplayer
...
+
+        return false;
+    }
+
+    const AZStd::unordered_map<AZ::Name, AZ::HashValue64>& MultiplayerComponentRegistry::GetMultiplayerComponentVersionHashes() const
+    {
+        return m_componentVersionHashes;
+    }
+
     void MultiplayerComponentRegistry::Reset()
```
AMZN-puvvadar, line 100: Should this clear `m_componentVersionHash `as well?

Gems/Multiplayer/Code/Include/Multiplayer/Components/MultiplayerComponentRegistry.h
```diff
@@ -67,11 +67,26 @@ namespace Multiplayer
...
+        //! @return True if we found the multiplayer component and filled out the hash value; otherwise false.
+        bool FindComponentVersionHashByName(const AZ::Name& multiplayerComponentName, AZ::HashValue64& hash) const;
+
         //! This releases all owned memory, should only be called during multiplayer shutdown.
         void Reset();
 
     private:
         NetComponentId m_nextNetComponentId = NetComponentId{ 0 };
         AZStd::unordered_map<NetComponentId, ComponentData> m_componentData;
+        AZ::HashValue64 m_componentVersionHash = AZ::HashValue64{ 0 };
```
AMZN-puvvadar, line 89: `m_componentVersionHash` and `m_componentVersionHashes` are very similar names despite representing different things. I'd consider renaming the former to something more descriptive.

AMZN-puvvadar:
The approach looks good generally. One question I have is do we want to send the full dataset of hashes for the mismatch? That can certainly be useful for debugging but I wonder if that's too descriptive in a production environment.

For example, should a client get that much information about server hashes or should it just get that it is behind the server version and needs to upda[te?

Gems/Multiplayer/Code/Include/Multiplayer/Components/MultiplayerComponentRegistry.h
AMZN-Gene: Renamed! Good call, the RFC referred to it as a "holistic" hash, and the Getter method refers to it that way as well... fixed!

Gems/Multiplayer/Code/Source/Components/MultiplayerComponentRegistry.cpp
AMZN-Gene: thank you!

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
```diff
@@ -842,6 +873,91 @@ namespace Multiplayer
...
+                    AZLOG_ERROR(
+                        "Multiplayer component mismatch! %s has a different version hash. Please make sure both client and server have "
+                        "matching multiplayer components.",
+                        theirComponent.m_componentName.GetCStr());
+                }
+            }
+            else
+            {
+                // Connected machine is using a multiplayer component we don't have
+                AZLOG_ERROR(
```
AMZN-Olex, line 922: I am wondering if this should be an ebus to report on? This way it could be easy to add it to the Editor for example, if there is some odd reason when this fails.

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
```diff
@@ -842,6 +873,91 @@ namespace Multiplayer
...
+            {
+                AZLOG_ERROR(
+                    "Multiplayer component mismatch! We have a component named %s which the connected application is missing!",
+                    ourComponentName.GetCStr());
+            }
+        }
+
+        // Disconnect and clear out the list of components in case the user fixes up their end and tries to reconnect
+        theirMultiplayerComponents.clear();
+        connection->Disconnect(DisconnectReason::VersionMismatch, TerminationEndpoint::Local);
```
AMZN-Olex, line 957: Hmm, is there an ebus/event here that a user can sign up and detect this version mismatch? Perhaps, we should pass some extra data here as well for debug-ability?

AMZN-Olex: For example, I am curious what would be needed to expose to the Editor? Looking into the logs can often difficult.

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
AMZN-Gene: Yes! Events coming in the next PR along with on-screen debug information, and unit tests.

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
AMZN-Gene: Yup! Events coming in the next PR
I wasn't planning anything special just for the editor, but if the mp viewport debug message system is enabled, I'll display a popup saying there's a mismatch. Also the event could be used for game-specific logic (like having the user download the latest version).

Gems/Multiplayer/Code/Include/Multiplayer/MultiplayerTypes.h
```diff
@@ -113,6 +113,20 @@ namespace Multiplayer
         Enabled
     };
 
+        //! The maximum number of component version hashes we want to stuff into a single packet
```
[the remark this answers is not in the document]
AMZN-Gene: fixed!

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
```diff
@@ -600,6 +600,37 @@ namespace Multiplayer
                 return true;
             }
         }
+
+        // Make sure the client that's trying to connect has the same multiplayer components
```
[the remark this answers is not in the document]
AMZN-Gene: Thought about that as well; i think the only time i'd disable it is if the system itself was broken and wasn't reliably finding mismatches. Excluding not working as expected, is there a reason you'd want to run a client and server with different components?

AMZN-Gene:
> The approach looks good generally. One question I have is do we want to send the full dataset of hashes for the mismatch? That can certainly be useful for debugging but I wonder if that's too descriptive in a production environment.
> 
> For example, should a client get that much information about server hashes or should it just get that it is behind the server version and needs to update?

I was thinking if it occurred it would be super useful for debugging if players uploaded their logs to the developer. You're thinking the component names could be maliciously used for hacking? I could make a cvar to bypass sending individual hashes and just call disconnect(versionMismatch). Although there's also talks of a cvar to not disconnect mismatch, in which case, I could send an different less-information-mismatch packet.

AMZN-puvvadar:
> > The approach looks good generally. One question I have is do we want to send the full dataset of hashes for the mismatch? That can certainly be useful for debugging but I wonder if that's too descriptive in a production environment.
> > For example, should a client get that much information about server hashes or should it just get that it is behind the server version and needs to update?
> 
> I was thinking if it occurred it would be super useful for debugging if players uploaded their logs to the developer and they could see what component went wrong. You're thinking the component names could be maliciously used for hacking? I could make a cvar to bypass sending individual hashes and just call disconnect(versionMismatch). Although there's also talks of a cvar to not disconnect on mismatch, in which case, I could create a new less-information-mismatch packet.



> > The approach looks good generally. One question I have is do we want to send the full dataset of hashes for the mismatch? That can certainly be useful for debugging but I wonder if that's too descriptive in a production environment.
> > For example, should a client get that much information about server hashes or should it just get that it is behind the server version and needs to update?
> 
> I was thinking if it occurred it would be super useful for debugging if players uploaded their logs to the developer and they could see what component went wrong. You're thinking the component names could be maliciously used for hacking? I could make a cvar to bypass sending individual hashes and just call disconnect(versionMismatch). Although there's also talks of a cvar to not disconnect on mismatch, in which case, I could create a new less-information-mismatch packet.

That's my concern yeah. I think rather than relying on users sending logs it'd be better to send a packet back to the server and then let the developer decide what they want to do with that data.

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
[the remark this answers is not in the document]
AMZN-Gene: Fixed! Added sv_versionMismatch_autoDisconnect which when false will still allow players to join even if there is a mismatch.

AMZN-Gene:
> That's my concern yeah. I think rather than relying on users sending logs it'd be better to send a packet back to the server and then let the developer decide what they want to do with that data.

Fixed! Added sv_versionMismatch_sendAllComponentHashesToClient. Set to false and the client will only hear about the mismatch, but not get any details on the specific components.

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
```diff
@@ -842,6 +871,125 @@ namespace Multiplayer
...
+            {
+                // DANGER: We're accepting the player connection even though there's a component mismatch
+                AZLOG_WARN(
+                    "Multiplayer component mismatch was found but we're allowing the player to connect anyways. Please set sv_versionMismatch_autoDisconnect=true if this is undesired behavior!");
+                AttemptPlayerConnect(connection, m_originalConnectPackets[connection->GetConnectionId()]);
+                m_originalConnectPackets.erase(connection->GetConnectionId());
+            }
+            else
+            {
+                AZLOG_ERROR(
```
AMZN-puvvadar, line 984: NP: Not sure it's worth saying "please file a bug" here.

Also could this be construed as not trying to accept a specific mismatched connection with the cvar off? That use case seems rare though.

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
```diff
@@ -1485,6 +1638,32 @@ namespace Multiplayer
...
+        componentMismatchPacket.SetTotalComponentCount(totalMultiplayerComponentCount);
+
+        for (const auto& multiplayerComponentData : GetMultiplayerComponentRegistry()->GetMultiplayerComponentVersionHashes())
+        {
+            componentMismatchPacket.ModifyComponentVersions().emplace_back(
+                multiplayerComponentData.first, aznumeric_cast<uint64_t>(multiplayerComponentData.second));
+
+            if (componentMismatchPacket.GetComponentVersions().full())
+            {
+                // we've hit the packet limit, send the current packet, and begin filling up another
```
AMZN-puvvadar, line 1654: This alone makes the "send no manifest" cvar worth it. On a bulky mismatch, it'll keep things quick.

One thing you can consider here is doing one large packet of dynamic size where you store the size of the buffer followed by the buffer. That way you'll leverage AzNetworking's large packet fragmentation. Not strictly necessary but something to think about.

AMZN-Gene: sv_versionMismatch_sendAllComponentHashesToClient = false stops sending all information to the client. Think there should be another cvar that eliminates sending all the components to both client and server?

AMZN-Gene: Ah, I wasn't aware I could send non-fixed size vectors. I didn't want to assume a large fixed size since there really isn't any cap on the number of components a game can have. Dynamic packet size would let me send just one packet, point me at any existing packet example if you get a chance.

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
AMZN-Gene: Maybe an assert is better here? Basically if that code ever gets hit... the code is wrong and a developer needs to fix it.

AMZN-Gene: _"Also could this be construed"_
Ah I see, the logic flow was weird to read with the if/else-if/else. Updated so it's just an if/else and then another if/else inside the else, which is easier on the eyes. Lemme know!

Gems/Multiplayer/Code/Include/Multiplayer/AutoGen/AutoComponent_Header.jinja
```diff
@@ -533,6 +533,7 @@ namespace {{ Component.attrib['Namespace'] }}
         bool HasController() const override;
         Multiplayer::MultiplayerController* GetController() override;
         const Multiplayer::MultiplayerController* GetController() const override;
+        static AZ::HashValue64 GetVersionHash() { return s_versionHash; }
```
kberg-amzn, line 536: Really try to keep implementations out of headers.. even trivial guys like this, as it has a tendency to get out of control quickly in a large codebase like what we're dealing with..

Gems/Multiplayer/Code/Include/Multiplayer/Components/MultiplayerComponentRegistry.h
```diff
@@ -67,11 +67,26 @@ namespace Multiplayer
         //! @return reference to the requested component data, an empty container will be returned if the NetComponentId does not exist
         const ComponentData& GetMultiplayerComponentData(NetComponentId netComponentId) const;
 
+        //! Returns the combined hashes of all the multiplayer components creating a single holistic version hash that can be compared between the server and client app.
+        //! @return a 64-bit hash value representing of all of the multiplayer components
+        AZ::HashValue64 GetMultiplayerComponentVersionHolisticHash() const;
```
kberg-amzn, line 72: Holistic seems a bit weird too 😆 CombinedHash or AggregateHash might read a bit better? Perhaps even ComponentRegistryHash, although that is definitely a bit more esoteric..

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
AMZN-puvvadar: I don't think it's been attempted but look at AzContainerSerializers.h. Non-fixed size containers have an implementation along with non-fixed size strings.

The question on sv_versionMismatch_sendAllComponentHashesToClient is an implementation detail for you to consider.

Gems/Multiplayer/Code/Include/Multiplayer/Components/MultiplayerComponentRegistry.h
AMZN-puvvadar: Maybe just GetMultiplayerSystemVersionHash? I think the focus here should be on describing what it is not how it's formed.

AMZN-Gene: I like it! Name changed to SystemHash.

Gems/Multiplayer/Code/Include/Multiplayer/AutoGen/AutoComponent_Header.jinja
AMZN-Gene: Thank you, fixed

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
AMZN-Gene: Replaced with assert.

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
AMZN-Gene: Love it! I'm using a dynamically sized map and the code is a lot cleaner. Check out AzContainerSerializers.h, I updated it to include pairs since that's needed for the map iterator, but otherwise it worked without a problem.

Gems/Multiplayer/Code/Include/Multiplayer/Components/MultiplayerComponentRegistry.h
```diff
@@ -67,11 +67,26 @@ namespace Multiplayer
         //! @return reference to the requested component data, an empty container will be returned if the NetComponentId does not exist
         const ComponentData& GetMultiplayerComponentData(NetComponentId netComponentId) const;
 
+        //! Returns the combined hashes of all the multiplayer components creating a single system version hash that can be quickly compared between the server and client app.
+        //! @return a 64-bit system-wide hash value representing of all of the multiplayer components
+        AZ::HashValue64 GetSystemVersionHash() const;
```
lmbr-pip, line 72: nit: GetMultiplayerSystemVersionHash or GetCombinedMultiplayerVersionHash? 

Feel that GetSystemVersionHash is a little too brief and not as self-describing as we could make it

Gems/Multiplayer/Code/Source/AutoGen/Multiplayer.AutoPackets.xml
```diff
@@ -11,17 +11,22 @@
...
         <Member Type="uint64_t" Name="temporaryUserId" Init="0" />
         <Member Type="Multiplayer::LongNetworkString" Name="ticket" />
+        <Member Type="AZ::HashValue64" Name="multiplayerComponentVersionHash" />
     </Packet>
 
     <Packet Name="Accept" HandshakePacket="true" Desc="Server accept packet">
         <Member Type="Multiplayer::LongNetworkString" Name="map" />
     </Packet>
 
+    <Packet Name="SyncComponentMismatch" HandshakePacket="true" Desc="Sends a vector of multiplayer components with their name and version hashes in order to compare which particular components are mismatched. Note: Marked as a handshake packet so it can be sent back to the client without having received the server acceptance packet.">
```
lmbr-pip, line 21: Consider changing the name and description so that its clear this is synching component hashes.

Unclear from desc if this is all the hashes or just the ones which are different.

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
```diff
@@ -679,6 +662,52 @@ namespace Multiplayer
...
+            if (!sessionRequests->ValidatePlayerJoinSession(config))
+            {
+                auto visitor = [](IConnection& connection) { connection.Disconnect(DisconnectReason::TerminatedByUser, TerminationEndpoint::Local); };
+                m_networkInterface->GetConnectionSet().VisitConnections(visitor);
+                return true;
+            }
+        }
+
+        // Make sure the client that's trying to connect has the same multiplayer components
+        if (GetMultiplayerComponentRegistry()->GetSystemVersionHash() != packet.GetMultiplayerComponentVersionHash())
```
lmbr-pip, line 689: Is this the right check? These should compare "system hash" (ours) vs "system hash" theirs right? Naming makes that unclear IMHO

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
```diff
@@ -855,6 +884,104 @@ namespace Multiplayer
...
+                        "Multiplayer component mismatch! %s has a different version hash. Please make sure both client and server have "
+                        "matching multiplayer components.",
+                        theirComponentName.GetCStr());
+                }
+            }
+            else
+            {
+                // Connected machine is using a multiplayer component we don't have
+                AZLOG_ERROR(
+                    "Multiplayer component mismatch! We're missing a component with version hash 0x%llx. "
```
lmbr-pip, line 913: nit: avoid personal pronouns in error messages. Unclear to reader who "we" is.

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
```diff
@@ -855,6 +884,104 @@ namespace Multiplayer
...
+                    {
+                        theyHaveComponent = true;
+                        break;
+                    }
+                }
+
+                if (!theyHaveComponent)
+                {
+                    AZLOG_ERROR(
+                        "Multiplayer component mismatch! We have a component named %s which the connected application is missing!",
```
lmbr-pip, line 942: nit: personal pronouns in error message

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
```diff
@@ -855,6 +884,104 @@ namespace Multiplayer
...
+            {
+                // Disconnect from the connector; we have all the mismatch information we need now for debugging.
+                connection->Disconnect(DisconnectReason::VersionMismatch, TerminationEndpoint::Local);
+            }
+            else
+            {
+                if (m_originalConnectPackets.contains(connection->GetConnectionId()))
+                {
+                    // DANGER: We're accepting the player connection even though there's a component mismatch
+                    AZLOG_WARN("Multiplayer component mismatch was found but we're allowing the player to connect anyways. Please set "
```
lmbr-pip, line 969:
```suggestion
                    AZLOG_WARN("Multiplayer component mismatch was found. Server configured to allow the player to connect anyways, which may result in undefined behavior. Set "
```

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
AMZN-puvvadar: Second @lmbr-pip 's sentiment.

Gems/Multiplayer/Code/Include/Multiplayer/Components/MultiplayerComponentRegistry.h
AMZN-puvvadar: I would say GetMultiplayerSystemVersionHash between the two. Personally I find the namespace and class qualifies GetSystemVersionHash reasonably well enough.

AMZN-puvvadar (approved):
With pending resolution of issues around naming/comments in mind, this looks good to me.

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
```diff
@@ -97,7 +97,9 @@ namespace Multiplayer
         "slow down quicker and may be better suited to connections with highly variable latency");
     AZ_CVAR(bool, bg_multiplayerDebugDraw, false, nullptr, AZ::ConsoleFunctorFlags::Null, "Enables debug draw for the multiplayer gem");
     AZ_CVAR(bool, cl_connect_onstartup, false, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Whether to call connect as soon as the Multiplayer SystemComponent is activated.");
-    
+    AZ_CVAR(bool, sv_versionMismatch_autoDisconnect, true, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Should the server automatically disconnect a client that is attempting connect who is running a build containing different/modified multiplayer components.");
```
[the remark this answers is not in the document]
AMZN-Gene: I'm conflicted on this; would users care how the mismatch is detected (using a hash), or simply that's there was a mismatch?

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
AMZN-Gene: <3

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
AMZN-Gene: Fixed. Replaced "We" with "This application"

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
AMZN-Gene: Fixed!

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
AMZN-Gene: good called! changed to packet.GetSystemVersionHash

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
```diff
@@ -842,6 +871,127 @@ namespace Multiplayer
...
+                        "Multiplayer component mismatch! %s has a different version hash. Please make sure both client and server have "
+                        "matching multiplayer components.",
+                        theirComponent.m_componentName.GetCStr());
+                }
+            }
+            else
+            {
+                // Connected machine is using a multiplayer component we don't have
+                AZLOG_ERROR(
+                    "Multiplayer component mismatch! We're missing a component with version hash 0x%" PRIx64 ". "
```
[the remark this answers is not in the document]
AMZN-Gene: good call. fixed to "this application"

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
```diff
@@ -97,7 +97,9 @@ namespace Multiplayer
         "slow down quicker and may be better suited to connections with highly variable latency");
     AZ_CVAR(bool, bg_multiplayerDebugDraw, false, nullptr, AZ::ConsoleFunctorFlags::Null, "Enables debug draw for the multiplayer gem");
     AZ_CVAR(bool, cl_connect_onstartup, false, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Whether to call connect as soon as the Multiplayer SystemComponent is activated.");
-    
+    AZ_CVAR(bool, sv_versionMismatch_autoDisconnect, true, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Should the server automatically disconnect a client that is attempting connect who is running a build containing different/modified multiplayer components.");
+    AZ_CVAR(bool, sv_versionMismatch_sendAllComponentHashesToClient, true, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Should the server send all its individual multiplayer component version hashes to the client when there's a mismatch? Upon receiving the information, the client will print the mismatch information to the game log. This is good for debugging during development, but you may want to mark false for release builds.");
```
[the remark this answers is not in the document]
AMZN-Gene: thanks, fixed!

Gems/Multiplayer/Code/Source/AutoGen/Multiplayer.AutoPackets.xml
AMZN-Gene: Updated the description to mention that all component information is sent in order to compare. Both the server and client receive this packet from one another. Wondering about the name change though, since I'm more concerned about the event (mismatch) than the under-lying hashes. Maybe "component" is making it confusing; I'm thinking "VersionMismatch" instead of "SyncComponentMismatch" would be a better name. Like a "version mismatch" event just occurred.

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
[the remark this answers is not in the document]
AMZN-Gene: Seeing the related review comment below about the packet name "SyncComponentMismatch". What if I changed the packet name to "VersionMismatch"? Trying to lower the user's burden of having to know how the mismatch is detected (via hashes) vs just knowing that there is a mismatch.

Gems/Multiplayer/Code/Include/Multiplayer/Components/MultiplayerComponentRegistry.h
AMZN-Gene: Thanks for the feedback. Totally, both of the same thoughts crossed my mind, but I went with the less redundant "GetSystemVersion". The namespace and class covers the base nicely. Plus the calls for an outsider look like...
_GetMultiplayerComponentRegistry()->GetSystemVersionHash()_

AMZN-Olex (approved):
Looks good!

PR #12978, opened 2022-11-02
Remove the hardcoded sleep for setting the file creation timestamp

junbo75:
Signed-off-by: Junbo Liang <68558268+junbo75@users.noreply.github.com>

## What does this PR do?

- Replaced the hardcoded sleep in the asset processor manager unit test which is used to set a different timestamp for each file.
- Replace the anonymous namespace in the test code

## How was this PR tested?

asset processor manager unit tests passed locally

### DIFF
--- a/Code/Tools/AssetProcessor/native/unittests/AssetProcessorManagerUnitTests.cpp
+++ b/Code/Tools/AssetProcessor/native/unittests/AssetProcessorManagerUnitTests.cpp
@@ -171,7 +171,7 @@ namespace AssetProcessor
         EXPECT_EQ(jobDetails.m_intermediatePath, intermediateAssetsFolder);
     };
 
-    namespace Internal
+    namespace AssetProcessorManagerUnitTestUtils
     {
         class MockAssetBuilderInfoHandler
             : public AssetProcessor::AssetBuilderInfoBus::Handler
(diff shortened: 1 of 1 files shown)


gadams3 (approved):

cgalvan (approved):

jackalbe (approved):

amzn-mike (approved):

Code/Tools/AssetProcessor/native/unittests/AssetProcessorManagerUnitTests.cpp
```diff
@@ -190,22 +171,44 @@ namespace AssetProcessor
         EXPECT_EQ(jobDetails.m_intermediatePath, intermediateAssetsFolder);
     };
 
-    namespace
+    namespace Internal
```
lmbr-pip, line 174: Why move from the anonymous namespace? 

If we are adding this to provide semantic meaning, then maybe give it a better name APUnitTestUtils or similar?

AMZN-stankowi (approved):

Code/Tools/AssetProcessor/native/unittests/AssetProcessorManagerUnitTests.cpp
junbo75: In my previous PR, Allen mentioned that O3DE coding practices do not use anonymous namespaces. I can replace it with a more meaningful namespace.

PR #12991, opened 2022-11-03
Fixes for enabling prefabs in PhysX editor tests

michabr:

## What does this PR do?

This PR includes fixes to issues that were exposed when prefabs were enabled in the PhysX editor unit tests
- Previously when an entity was being deleted, it wasn't removed from its owning instance and thus leaving a dangling pointer. Now the entity being deleted is properly detached from its owning instance whether it's a container entity or a regular entity
- Fix in EditorColliderComponent::Deactivate where it was using m_sceneInterface after AzPhysics::SceneInterface was already deleted
- Change GetWarningCount() to GetExpectedWarningCount() in a couple of tests so that warnings unrelated to the test don't affect the outcome
- Enable prefabs now that issues have been fixed

Update:
- Add an EntityOwnershipService::HandleEntityBeingDestroyed method that gets called when an entity is being destroyed so that it can detach it from its owning instance if it has one. This replaces the call to EntityOwnershipService::DestroyEntityById which now actually deletes the entity (as the API comments suggest)
- Add Instance::DestroyEntity to allow a specific entity to be detached and deleted from its owning instance

Note: In the future, we might want to consider moving some stuff out of PrefabEditorEntityOwnershipService and into PrefabEntityOwnershipService.

## How was this PR tested?

Ran all PhysX tests and made sure they passed. Also ran Editor and tested loading/saving a level.


Code/Framework/AzToolsFramework/AzToolsFramework/Entity/PrefabEditorEntityOwnershipService.cpp
```diff
@@ -175,6 +180,11 @@ namespace AzToolsFramework
         AZ_Assert(IsInitialized(), "Tried to destroy an entity without initializing the Entity Ownership Service");
         AZ_Assert(m_entitiesRemovedCallback, "Callback function for DestroyEntityById has not been set.");
         OnEntityRemoved(entityId);
+        auto owningInstance = m_instanceEntityMapperInterface->FindOwningInstance(entityId);
+        if (owningInstance.has_value())
```
hultonha, line 184: nit: We could update this to...

```c++

if (auto owningInstance = m_instanceEntityMapperInterface->FindOwningInstance(entityId);
    owningInstance.has_value())
```

Just to limit the scope of `owningInstance`, also a guide comment here could be useful to indicate why this is required.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/Instance.cpp
```diff
@@ -198,32 +198,44 @@ namespace AzToolsFramework
 
         AZStd::unique_ptr<AZ::Entity> Instance::DetachEntity(const AZ::EntityId& entityId)
```
hultonha, line 199: question: Is there an existing test or a test we could add to verify this function to ensure it is doing what's expected? It might be cool to have a test that verified the previous error state and then run it again with this change to show it passes? It might be tricky to do but wanted to check as it could be really helpful in future

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/Instance.cpp
```diff
@@ -198,32 +198,44 @@ namespace AzToolsFramework
...
-                entityAliasToRemove = instanceToTemplateEntityIdIterator->second;
-                if (m_entityIdInstanceRelationship == EntityIdInstanceRelationship::OneToOne)
+                return DetachContainerEntity();
+            }
+            else
+            {
+                EntityAlias entityAliasToRemove;
+                auto instanceToTemplateEntityIdIterator = m_instanceToTemplateEntityIdMap.find(entityId);
+
+                if (instanceToTemplateEntityIdIterator != m_instanceToTemplateEntityIdMap.end())
```
hultonha, line 210: nit: We could potentially update to

```c++

if (auto instanceToTemplateEntityIdIterator = m_instanceToTemplateEntityIdMap.find(entityId);
    instanceToTemplateEntityIdIterator != m_instanceToTemplateEntityIdMap.end())
```

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/Instance.cpp
```diff
@@ -198,32 +198,44 @@ namespace AzToolsFramework
...
+                                "Prefab",
+                                false,
+                                "An owning instance couldn't be found corresponding to the entity requested to be detached");
+                            return nullptr;
+                        }
+                    }
+                    else
+                    {
+                        [[maybe_unused]] bool isEntityRemoved =
+                            m_templateToInstanceEntityIdMap.erase(entityAliasToRemove) && m_instanceToTemplateEntityIdMap.erase(entityId);
```
hultonha, line 227: question: Should these `erase` calls be explicitly checking `m_templateToInstanceEntityIdMap.erase(entityAliasToRemove) != m_templateToInstanceEntityIdMap.end()` and `m_instanceToTemplateEntityIdMap.erase(entityId) != m_instanceToTemplateEntityIdMap.end()`? What is it they return if the `erase` doesn't succeed?

hultonha (commented):
Looks great I just have a couple of small suggestions and one question. Once they're been reviewed I'd be happy to approve, thanks so much for looking into this and fixing it! 😄

Code/Framework/AzToolsFramework/AzToolsFramework/Entity/PrefabEditorEntityOwnershipService.cpp
```diff
@@ -175,6 +180,11 @@ namespace AzToolsFramework
...
+        {
+            owningInstance->get().DetachEntity(entityId).release();
```
srikappa-amzn, line 186: If a system tries to call the EOS to  DestroyEntityById(), this function now has the likelihood of creating a memory leak if the caller is not aware that they need to destroy the entity themselves.

Rather than doing it this way,
~Entity()->OnEntityDestruction()->DetachEntity(entityId).release() might be a better way to handle this.

This would require 
1. Adding an OnEntityDestroyed() call to EntityOwnershipService.h
2. Making the SliceEOS override it to call DestroyEntityById() (no change here)
3. Making the PrefabEditorEOS override it to detach and release.

Ideally, I don't think other systems should be managing the lifecycle of entities because it should be the prefab system that does it. That's why prefab instances have AZStd::unique_ptr<AZ::Entity> in them. But access to raw pointer to entities has been given prior to prefabs and exists for legacy reasons.

If anyone from physx is reading this, I recommend that they change their unit tests implementation to not manage entity life cycles themselves.

srikappa-amzn (changes requested):

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/Instance.cpp
michabr: I believe this was just a formatting change for this PR, but good question. I checked this and the erase returns the number of elements that were actually erased. I'll change the code to check against 0 to make it more clear.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/Instance.cpp
michabr: I'll look into adding a test.

Code/Framework/AzToolsFramework/AzToolsFramework/Entity/PrefabEditorEntityOwnershipService.cpp
michabr: Even if entity deletion was managed by the prefab system, DestroyEntity still wouldn't have detached the entity from the Instance. Which call would a user make if they wanted to delete an entity?

srikappa-amzn: @michabr, can you create a github issue asking physx team to refactor their tests to not manage entity lifecycles themselves?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/Instance.cpp
srikappa-amzn: Do we need this change at all? So far, DetachEntity() only detaches entities within a prefab except the container entity. And if it is a container entity, users can call DetachContainerEntity() themselves. Any reason for changing this behavior now.
It is similar to other calls in the api. GetEntities() only returns entities in the prefab instance except the container entity.

Code/Framework/AzToolsFramework/AzToolsFramework/Entity/PrefabEditorEntityOwnershipService.cpp
michabr: It looks like DestroyEntityById and DestroyEntity have different purposes which is a bit confusing. DestroyEntityById is called via ~Entity() and shouldn't delete the entity since it's already being deleted. DestroyEntity is an API called by the user and so the function should delete the entity after detaching it. I'll double check that this is happening.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/Instance.cpp
srikappa-amzn: I think that's fine for now because that's how the API calls are defined so far: entities and container entity have different functions. We can change that behavior in the future if needed and let that be consistent across all functions Detach, Get, etc. but that'll be a breaking change since existing functionality might have made different calls for them separately and we need to take that into account when making the change. It'll also need to be accompanied with a 'impactful change' announcement

Code/Framework/AzToolsFramework/AzToolsFramework/Entity/PrefabEditorEntityOwnershipService.cpp
srikappa-amzn: > Even if entity deletion was managed by the prefab system, DestroyEntity still wouldn't have detached the entity from the Instance. Which call would a user make if they wanted to delete an entity?

They can call ToolsApplication::DeleteEntityById() or EditorEntityManager::DeleteEntityById or  PrefabPublicHandler::DeleteEntitiesAndAllDescendantsInInstance() which would correctly detach and also update corresponding prefab templates. 

> DestroyEntityById is called via ~Entity() and shouldn't delete the entity since it's already being deleted. 

yes but I don't think that's the function we should be calling. DesroyEntityById() is called from EntityContext::OnEntityDestruction(). Instead of doing that, I am recommending the introducton of a new EOS function to avoid confusion that we won't be actually destroying the entity but rather let the EOS do cleanup for the destroyed entity. Feel free to reach out to me if you have questions on this approach.

michabr: I wasn't familiar with the EditorEntityManager but it looks like that is a wrapper around prefab public handler? Would it be correct to say that EditorEntityManager/PrefabPublicHandler is for calls that also handle undo/redo nodes for you, and direct API calls to EOS should not do any undo/redo processing?

Since the EOS also has an API to destroy an entity, I think we should still handle that correctly and actually detach and delete the entity when called.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/Instance.cpp
michabr: OK, I left the DetachEntity function as is and moved the new logic outside of it.

Code/Framework/AzToolsFramework/AzToolsFramework/Entity/PrefabEditorEntityOwnershipService.cpp
michabr: Made DestroyEntity and DestroyEntityById behave the same (detaches and deletes the entity), and added a new method to replace the DestroyEntityById in EntityContext::OnEntityDestruction() that only detaches the entity from its instance.

Code/Framework/AzFramework/AzFramework/Entity/SliceEntityOwnershipService.cpp
```diff
@@ -146,11 +143,9 @@ namespace AzFramework
     {
         AZ_Assert(m_rootAsset && m_rootAsset->GetComponent(), "Root slice has not been created.");
         AZ_Assert(m_entitiesRemovedCallback, "Callback function for DestroyEntityById has not been set.");
+        SliceEntityRequestBus::MultiHandler::BusDisconnect(entityId);
         m_entitiesRemovedCallback({ entityId });
-
-        // Entities removed through the application (as in via manual 'delete'),
-        // should be removed from the root slice, but not again deleted.
-        return m_rootAsset->GetComponent()->RemoveEntity(entityId, false);
+        return m_rootAsset->GetComponent()->RemoveEntity(entityId);
```
srikappa-amzn, line 148: This changes the behavior because previously this function didn't delete the entity but now it will because RemoveEntity() takes 'true' as the default argument. With the above changes, I think it is fine to delete the entity here. Just keep an eye on this because if any bugs raise because of this we can always flip it back to 'false' again.

Code/Framework/AzToolsFramework/AzToolsFramework/Entity/PrefabEditorEntityOwnershipService.cpp
srikappa-amzn: > Would it be correct to say that EditorEntityManager/PrefabPublicHandler is for calls that also handle undo/redo nodes for you, and direct API calls to EOS should not do any undo/redo processing

That's a fair callout but one thing to remember is that slices uses EOS for both editor context and game context. So, being able to do something without undo/redo support makes sense. But prefabs only have an Editor EOS and prefabs are only an editor concept. So I don't think providing entity operations without undo/redo support for prefabs is something that we should be supporting. I can't think of an actual real life editor use case for being able to modify prefabs without undo/redo support. Let's say for example, you remove an entity from a prefab instance, what use is that prefab instance if it is not in sync with the underlying prefab template?

Also, PrefabEditorEOS::AddEntity() and PrefabEditorEOS::AddEntities() do support undo/redo. So, right now there doesn't appear to be a streamlined approach defined to say what the PrefabEditorEOS should be doing. In the long run, we should try to unify PrefabEditorEOS and PrefabPublicHandler into one api because they both serve the same purpose- provide a way to modify editor entities while keeping the underlying in-memory prefabs in the correct states.

Overall, I think it is okay to add support for DestroyEntityById() because it satisfies the EOS requirements but we should really streamline the EOS and PublicHandler APIs and decide how they are going to co-exist going forward. @michabr , can you create a github issue to track this work?

srikappa-amzn (approved):
Looks good. left a minor comment and couple of follow up tasks.

Code/Framework/AzToolsFramework/AzToolsFramework/Entity/PrefabEditorEntityOwnershipService.cpp
forkercat: Curious: Is release() the right call to destroy the entity object?

Ref: https://stackoverflow.com/questions/25609457/does-unique-ptrrelease-call-the-destructor

lumberyard-employee-dm: No release explicitly means that the unique pointer gives up owernship of it's address.
For destruction, `reset()` would need to be called

michabr: Here, we don't want to destroy the entity object.

Code/Framework/AzToolsFramework/AzToolsFramework/Entity/PrefabEditorEntityOwnershipService.cpp
```diff
@@ -167,14 +172,25 @@ namespace AzToolsFramework
 
     bool PrefabEditorEntityOwnershipService::DestroyEntity(AZ::Entity* entity)
     {
+        AZ_Assert(entity, "Tried to destroy a null entity");
```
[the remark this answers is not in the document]
michabr: Will do. I thought in our coding guidelines we shouldn't be preventing crashes for programmer error/asserts, but that could have changed.

Code/Framework/AzFramework/AzFramework/Entity/SliceEntityOwnershipService.cpp
michabr: Yeah, I will keep an eye on it. It makes sense to delete the entity here especially because in SliceEntityOwnershipService.h it reads:
`        //! Deletes the entity id from the root-slice and destroys it.
        //! @param entityId
        bool DestroyEntityById(AZ::EntityId entityId) override;`

Code/Framework/AzToolsFramework/AzToolsFramework/Entity/PrefabEditorEntityOwnershipService.cpp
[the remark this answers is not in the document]
moraaar: Agree with @michabr, asserts are not needed to be protected, if the function is never meant to be called with entity null then an assert is enough and failing asserts are meant to leave the app in an unknown state because it's never meant to happen. If it's possible to happen then it's not an az_assert, but an az_error that needs to be handled.

Code/Framework/AzFramework/AzFramework/Entity/EntityOwnershipService.h
```diff
@@ -92,6 +92,13 @@ namespace AzFramework
         virtual void SetEntitiesRemovedCallback(OnEntitiesRemovedCallback onEntitiesRemovedCallback) = 0;
         virtual void SetValidateEntitiesCallback(ValidateEntitiesCallback validateEntitiesCallback) = 0;
 
+        /**
+         * Called when an entity is in the process of being destroyed
+         *
+         * \param entityId The entity Id of the entity being destroyed.
+         */
+        virtual void HandleEntityBeingDestroyed(const AZ::EntityId& entityId) = 0;
```
hultonha, line 100: nit: This is super minor and we don't do this every where already, but as `EntityId` is just a type safe wrapper around a `u64`, we can pass by value and not `const` ref

Code/Framework/AzToolsFramework/AzToolsFramework/Entity/PrefabEditorEntityOwnershipService.cpp
```diff
@@ -443,6 +459,7 @@ namespace AzToolsFramework
 
     void PrefabEditorEntityOwnershipService::OnEntityRemoved(AZ::EntityId entityId)
     {
+        AZ_Assert(m_entitiesRemovedCallback, "Callback function for entity removal has not been set.");
```
moraaar, line 462: Is the callback always meant to be set? If yes then an assert is correct, if it's a choice to set it or not, then it seems better to remove the assert and do this instead:

````
    AzFramework::SliceEntityRequestBus::MultiHandler::BusDisconnect(entityId);
    if (m_entitiesRemovedCallback)
    {
        m_entitiesRemovedCallback({ entityId });
    }
````

File: Code/Framework/AzFramework/AzFramework/Entity/SliceEntityOwnershipService.cpp [hunk 10]
```diff
@@ -661,6 +656,17 @@ namespace AzFramework
...
     }
 
+    void SliceEntityOwnershipService::HandleEntityBeingDestroyed(const AZ::EntityId& entityId)
+    {
+        AZ_Assert(m_rootAsset && m_rootAsset->GetComponent(), "Root slice has not been created.");
+        AZ_Assert(m_entitiesRemovedCallback, "Callback function for entity removal has not been set.");
+        m_entitiesRemovedCallback({ entityId });
+
+        // Entities removed through the application (as in via manual 'delete'),
+        // should be removed from the root slice, but not again deleted.
```
hultonha, line 666: +1 to the helpful guide comment 😄

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/Instance.h
```diff
@@ -226,6 +227,7 @@ namespace AzToolsFramework
             void SetContainerEntity(AZ::Entity& entity);
 
             AZStd::unique_ptr<AZ::Entity> DetachContainerEntity();
+            bool DestroyContainerEntity();
```
hultonha, line 230: suggestion: Brief API comment about the return value (true deleted, false not) might be helpful

hultonha (approved):
LGTM! Thank you again for looking into this and fixing these issues @michabr, great work!

moraaar (approved):
LGTM. Please test closing editor and launcher won't crash with this new code:
- Open editor, open default level, close editor pressing X button. It shouldn't crash.
- Run launcher with a default level, close launcher by entering `quit` console command. Launcher should gracefully close without crashing.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/Instance.cpp
hultonha: Aah I see, thanks for updating, much clearer now 👍

Code/Framework/AzToolsFramework/AzToolsFramework/Entity/PrefabEditorEntityOwnershipService.cpp
```diff
@@ -167,14 +172,25 @@ namespace AzToolsFramework
...
+            if (owningInstance->get().GetContainerEntityId() == entityId)
+            {
+                owningInstance->get().DestroyContainerEntity();
+            }
+            else
+            {
+                owningInstance->get().DestroyEntity(entityId);
+            }
+        }
         return true;
```
forkercat, line 194: looks like it always returns true (like what the function did before). Is it the right way of continuing do this? 🤔

forkercat (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Entity/PrefabEditorEntityOwnershipService.cpp
michabr: This should be fixed to return the correct outcome. I'll add that.

Code/Framework/AzToolsFramework/AzToolsFramework/Entity/PrefabEditorEntityOwnershipService.cpp
michabr: Yeah, the callback is always meant to be set.

Tested opening/closing level in editor and launcher.

moraaar: Thanks @michabr

hultonha:
@michabr it looks like the most recent build has this test fail - `SliceEditorEntityOwnershipTests.RestoreSliceEntity_SliceEntityDeleted_SliceEntityRestored` which looks like it could be related to these changes. Might be worth doing a quick investigation before starting another build. Thanks!

michabr:
> @michabr it looks like the most recent build has this test fail - `SliceEditorEntityOwnershipTests.RestoreSliceEntity_SliceEntityDeleted_SliceEntityRestored` which looks like it could be related to these changes. Might be worth doing a quick investigation before starting another build. Thanks!

Hey @hultonha yes this test was failing due to the entity actually being deleted now when calling DestroyEntityById. I've fixed the test and it now succeeds for me locally but still fails in AR. I'm rebuilding one more time with a clean workspace and if that doesn't pass, I'll investigate further.

PR #12993, opened 2022-11-03
Fix PrefabUndoAddEntityTests crash

chiyenteng:
Signed-off-by: chiyenteng <82238204+chiyenteng@users.noreply.github.com>

## What does this PR do?

Fix a crash from unit test PrefabUndoAddEntityTests:: PrefabUndoAddEntity_AddEntityUnderUnfocusedInstance by using instance after propagation for validation instead of before.

## How was this PR tested?

Ran all unit tests in AzToolsFramework.Tests and passed.


srikappa-amzn (approved):
Makes sense because the instance (or at least parts of it ) would have been rebuilt due to propagation.

forkercat (approved):

PR #12999, opened 2022-11-04
Adding PAL functions for query CPUThreadTime

lumberyard-employee-dm:

## How was this PR tested?

Added a UnitTest to validate that the CpuThread time is non-zero after waiting 50 milliseconds after the first call


gadams3 (approved):

invertednormal (approved):
The solution on windows feels a bit hacky but guessing there aren't any other options. It's odd to see a comment that basically says "the docs say don't do this, but we're doing it anyway" with no other explanation why.

lumberyard-employee-dm:
> The solution on windows feels a bit hacky but guessing there aren't any other options. It's odd to see a comment that basically says "the docs say don't do this, but we're doing it anyway" with no other explanation why.

Well that is because there is no other way to accurately grab the CPU time on Windows.
I'll update the comment though.

spham-amzn (approved):

PR #13004, opened 2022-11-04
Add cmake option for RGA

etherealchain:

## What does this PR do?

Following #12710 
Add a cmake Cache Entry LY_RADEON_GPU_ANALYZER_ENABLED to control whether RGA should be download.

## How was this PR tested?

Test on my PC.


cmake/3rdParty.cmake
```diff
@@ -6,6 +6,8 @@
 #
 #
 
+set(LY_RADEON_GPU_ANALYZER_ENABLED FALSE CACHE BOOL "Whether to download Radeon GPU Analyzer from Github.")
```
lumberyard-employee-dm, line 9: Can use `O3DE_RADEON_GPU_ANALYZER_ENABLED` for the name of the Cache Variable?
The `LY_*` prefix is a holdover from Lumberyard and going forward we are wanting to use O3DE specific names.

lumberyard-employee-dm (approved):

jeremyong-az (approved):

lumberyard-employee-dm (approved):

lawsonamzn:
(Not gonna block this, but...) Long term it really feels like this should be a ui driven option somewhere in the asset processor or editor, when you try to turn on analysis or something, and you don' ahve it, it should offer to get it for you.  I don't see why this would ultimately be driven by the build system at all since its not really involved in build or deploy.

cgalvan (approved):

PR #13013, opened 2022-11-04
Add texture tags and quality

CyaNinix:
Hi,

This commit adds support for texture tags (allowing applications to register tags and then tag textures with them) along with a way to setup a quality level per tag, allowing to change the maximum mipchain that will be uploaded to the GPU.

Quality level defaults on 0 (= current behavior, all mipchains are uploaded), each value means a mipchain will not be uploaded to the GPU (quality level 1 means that the first mipchain won't be uploaded, etc.).  
Textures with multiple tags take their highest quality (= lowest value = biggest mipchain).

Texture quality can be changed at runtime which will trigger an asset reload (since recreating the texture is currently the only way we can prevent unused mipmaps VRAM consumption, although I know this can be improved with the coming sparse texture support).

This comes with an updated texture settings UI to add/remove tags to textures.

### DIFF
--- a/Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Image/ImageQuality.h
+++ b/Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Image/ImageQuality.h
@@ -1,0 +1,22 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#pragma once
+
+#include <AzCore/base.h>
+
+namespace AZ
+{
+    namespace RPI
+    {
+        using ImageQuality = u32;
+
+        constexpr ImageQuality ImageQualityHighest = 0;
+        constexpr ImageQuality ImageQualityLowest = 0xFFFFFFFF;
+    }
+}
--- a/Gems/Atom/RPI/Code/Include/Atom/RPI.Reflect/Image/StreamingImageAsset.h
+++ b/Gems/Atom/RPI/Code/Include/Atom/RPI.Reflect/Image/StreamingImageAsset.h
@@ -13,6 +13,7 @@
 #include <Atom/RPI.Reflect/Image/ImageAsset.h>
 #include <Atom/RPI.Reflect/Image/StreamingImagePoolAsset.h>
 #include <Atom/RPI.Reflect/Image/ImageMipChainAsset.h>
+#include <AzCore/std/containers/unordered_set.h>
 
 namespace AZ
 {
@@ -104,6 +105,9 @@ namespace AZ
             //! Whether the image has all referenced ImageMipChainAssets loaded
             bool HasFullMipChainAssets() const;
 
+            //! Returns the image tags
+            const AZStd::unordered_set<AZ::Name>& GetTags() const;
+
         private:
             struct MipChain
             {
@@ -143,6 +147,8 @@ namespace AZ
             //! mip chain index, not the level. And will fail for any level
             //! that resides in the tail mip chain.
             const ImageMipChainAsset* GetImageMipChainAsset(AZ::u32 mipLevel) const;
+
+            AZStd::unordered_set<AZ::Name> m_tags;
         };
     }
 }
--- a/Gems/Atom/RPI/Code/Source/RPI.Public/Image/ImageTagSystemComponent.cpp
+++ b/Gems/Atom/RPI/Code/Source/RPI.Public/Image/ImageTagSystemComponent.cpp
@@ -1,0 +1,133 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+/**
+* @file RPISystemComponent.cpp
+* @brief Contains the definition of the RPISystemComponent methods that aren't defined as inline
+*/
+
+#include <Atom/RPI.Public/Image/ImageTagSystemComponent.h>
+#include <AzCore/Asset/AssetManager.h>
+#include <AzCore/Component/TickBus.h>
+#include <AzCore/std/sort.h>
+#include <AzCore/Serialization/SerializeContext.h>
+#include <AzCore/Settings/SettingsRegistry.h>
+
+namespace AZ
+{
+    namespace RPI
+    {
+        ImageTagSystemComponent::~ImageTagSystemComponent() = default;
+
+        void ImageTagSystemComponent::Reflect(ReflectContext* context)
+        {
+            if (auto* serializeContext = azrtti_cast<SerializeContext*>(context))
+            {
+                serializeContext->Class<ImageTagSystemComponent::TagData>()
+                    ->Version(0)
+                    ->Field("Quality", &ImageTagSystemComponent::TagData::quality)
+                    ->Field("RegisteredImages", &ImageTagSystemComponent::TagData::registeredImages)
+                ;
+
+                serializeContext->Class<ImageTagSystemComponent, Component>()
+                    ->Version(0)
+                    ->Field("ImageTags", &ImageTagSystemComponent::m_imageTags)
+                ;
+            }
+        }
+
+        void ImageTagSystemComponent::GetRequiredServices(ComponentDescriptor::DependencyArrayType& /*required*/)
+        {
+        }
+
+        void ImageTagSystemComponent::GetProvidedServices(ComponentDescriptor::DependencyArrayType& provided)
+        {
+            provided.push_back(AZ_CRC_CE("TextureTagSystemComponent"));
+        }
+
+        void ImageTagSystemComponent::GetDependentServices(AZ::ComponentDescriptor::DependencyArrayType& /*dependent*/)
+        {
+        }
+
+        void ImageTagSystemComponent::Activate()
+        {
+            ImageTagBus::Handler::BusConnect();
+        }
+
+        void ImageTagSystemComponent::Deactivate()
+        {
+            ImageTagBus::Handler::BusDisconnect();
+        }
+
+        ImageQuality ImageTagSystemComponent::GetQuality(const AZ::Name& imageTag) const
+        {
+            auto it = m_imageTags.find(imageTag);
+            if (it == m_imageTags.end())
+            {
+                AZ_Warning("ImageTagSystemComponent", false, "Image tag %s has not been registered", imageTag.GetCStr());
+                return ImageQualityHighest;
+            }
+
+            return it->second.quality;
+        }
+
+        AZStd::vector<AZ::Name> ImageTagSystemComponent::GetTags() const
+        {
+            AZStd::vector<AZ::Name> tags;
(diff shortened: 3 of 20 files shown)


Gems/Atom/RPI/Code/Include/Atom/RPI.Reflect/Image/StreamingImageAsset.h
```diff
@@ -143,6 +147,8 @@ namespace AZ
             //! mip chain index, not the level. And will fail for any level
             //! that resides in the tail mip chain.
             const ImageMipChainAsset* GetImageMipChainAsset(AZ::u32 mipLevel) const;
+
+            AZStd::unordered_set<AZ::Name> m_tags;
```
invertednormal, line 151: Suggestion - Looking at the places this is used I'm not sure it needs to be an unordered set. Vector is faster to iterate through and likely to be faster for lookups with small numbers of tags (which I assume is the case here). Only thing I think you'd need to be careful about is duplicates being added in `AddTag()` but that's a simple check.

Gems/Atom/RPI/Code/Source/RPI.Public/Image/StreamingImage.cpp
```diff
@@ -170,7 +206,10 @@ namespace AZ
 
                     // We want to store off the id, not the AssetData instance. This simplifies the fetch / evict logic which
                     // can do more strict assertions.
-                    m_mipChains.push_back(Data::Asset<ImageMipChainAsset>(assetId, azrtti_typeid<ImageMipChainAsset>()));
+                    if (mipChainIndex >= highestMipChain)
+                        m_mipChains.push_back(Data::Asset<ImageMipChainAsset>(assetId, azrtti_typeid<ImageMipChainAsset>()));
+                    else
+                        m_mipChains.push_back(Data::Asset<ImageMipChainAsset>{});
```
invertednormal, line 212: o3de style: o3de style guidelines require `{` `}`
```suggestion
                    if (mipChainIndex >= highestMipChain)
                    {
                        m_mipChains.push_back(Data::Asset<ImageMipChainAsset>(assetId, azrtti_typeid<ImageMipChainAsset>()));
                    }
                    else
                    {
                        m_mipChains.push_back(Data::Asset<ImageMipChainAsset>{});
                    }
```

invertednormal (approved):
Looks like a good change overall to me @VickyAtAZ should definitely review though since this is her space. I just have a few minor suggestions and style considerations.

One other thing worth considering is storing the quality directly with the tag in image assets to avoid an extra lookup on the bus - it's a tiny amount of data and would be far cheaper than having to look up over ebus for every tag.

Gems/Atom/RPI/Code/Include/Atom/RPI.Reflect/Image/StreamingImageAsset.h
CyaNinix: Yes I went for a set just to quickly ensure there won't be any duplicate. Vector is faster to iterate but I doubt it'll change anything since tags are iterated only once on image creation

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Image/ImageQuality.h
```diff
@@ -0,0 +1,22 @@
...
+
+#pragma once
+
+#include <AzCore/base.h>
+
+namespace AZ
+{
+    namespace RPI
+    {
+        using ImageQuality = u32;
```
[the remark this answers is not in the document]
CyaNinix: 8bits would be more than enough, as this value won't really go over like 4-5. I put u32 as it's what's used to number the mipchains but it can be changed without any issue.

Gems/Atom/RPI/Code/Source/RPI.Public/Image/ImageTagSystemComponent.cpp
```diff
@@ -0,0 +1,133 @@
...
+        {
+            auto it = m_imageTags.find(imageTag);
+            if (it == m_imageTags.end())
+            {
+                AZ_Warning("ImageTagSystemComponent", false, "Image tag %s has not been registered", imageTag.GetCStr());
+                return;
+            }
+
+            TagData& tagData = it->second;
+            tagData.quality = quality;
```
VickyAtAZ, line 122: You may check if the quality is change and only reload assets if the quality is changed.

VickyAtAZ:
@CyaMinix The image tag system makes sense and I assume we can use vector<tag> for future function expansion?

Although I think it would be better the image tag quality setting is applied to StreamingImageAsset directly instead of StreamingImage. Here are the reasons:
1. The quality setting is for StreamingImageAsset and the implementation would be clean if we handle the quality when loading StreamingImageAsset. And there is no need to add all the condition checks in StreamingImage.
2. There are other places other than StreamingImage which use StreamingImageAsset. Such as Terrain height map, Decal system. The tag would still be functional without any specific implementation for these places. 
3. We have individual control for StreamingImage's target mip level at runtime via StreamingImage::SetTargetMip(). This would work well with the quality set by tag system.

To change to such an implementation, you may need
1. Revert the changes with StreamingImage
2. Register the StreamingImageAsset id to ImageTagSystem and apply image quality in StreamingImageAssetHandler::LoadAssetData
- Update the ImageDescriptor
- Update StreamingImageAsset::m_mipChains and StreamingImageAsset::m_mipLevelToChainIndex

Please let me know if you want to discuss more about the suggestion I proposed. (You may ping me on discord)

CyaNinix:
> @CyaMinix The image tag system makes sense and I assume we can use vector for future function expansion?

I'm not sure to understand, you mean having a vector instead of a map and using linear search instead of lookup for tag storage?

> Although I think it would be better the image tag quality setting is applied to StreamingImageAsset directly instead of StreamingImage. Here are the reasons:
> 
> 1. The quality setting is for StreamingImageAsset and the implementation would be clean if we handle the quality when loading StreamingImageAsset. And there is no need to add all the condition checks in StreamingImage.
> 2. There are other places other than StreamingImage which use StreamingImageAsset. Such as Terrain height map, Decal system. The tag would still be functional without any specific implementation for these places.
> 3. We have individual control for StreamingImage's target mip level at runtime via StreamingImage::SetTargetMip(). This would work well with the quality set by tag system.

I agree for point 1 and 2, I didn't see that function which seems a better fit to handle this.  
As for point 3, the main reason we implemented this is mainly to lower video memory usage, does the dynamic mipmap system handles this on all RHI? I'm not totally familiar with sparse textures but are they applicable for all textures? From what I know it's pretty heavy on the GPU and may no be suitable for all textures.


> To change to such an implementation, you may need
> 
> 1. Revert the changes with StreamingImage
> 2. Register the StreamingImageAsset id to ImageTagSystem and apply image quality in StreamingImageAssetHandler::LoadAssetData
> 
> * Update the ImageDescriptor
> * Update StreamingImageAsset::m_mipChains and StreamingImageAsset::m_mipLevelToChainIndex
> 
> Please let me know if you want to discuss more about the suggestion I proposed. (You may ping me on discord)

This seems like a good idea and I'm looking forward to update this PR as soon as I get some time.

VickyAtAZ:
> I'm not sure to understand, you mean having a vector instead of a map and using linear search instead of lookup for tag storage?
Sorry, I didn't mean you need to change anything. The set<tag> (or unorder_set) you used in texture settings and StreamingImageAsset can contain more than one tag which is more than a quality tag you need. 

> As for point 3, the main reason we implemented this is mainly to lower video memory usage, does the dynamic mipmap system handles this on all RHI? I'm not totally familiar with sparse textures but are they applicable for all textures? From what I know it's pretty heavy on the GPU and may no be suitable for all textures.

Yes, the quality tag system you have is useful and doesn't need backend support which is very nice. It doesn't have conflict with the image streaming system.
The image streaming system is supposed to reduce vram usage too when evict mipmaps. Currently we only have have dx12 backend actually support that. 
The dx12's streaming system implementation uses reserved resource (tiled resource, sparse image) . It supports most of streaming images except images doesn't have mipmap, image array with very small mip. It doesn't impact gpu performance on Nvidia card (RTX3070) I tested. I also did limited test on an AMD GPU which even has better gpu performance than committed resource (which is a bit of strange consider AMD suggested otherwise). I do have plan to look into the pix captures to see why it's faster. 

An alternative solution is to re-create the device resource when the streaming image mipmap count changes.  We will implement that if we find enough proof that reserved image has big performance impact on certain GPUs.

CyaNinix:
> The image streaming system is supposed to reduce vram usage too when evict mipmaps. Currently we only have have dx12 backend actually support that.
The dx12's streaming system implementation uses reserved resource (tiled resource, sparse image) . It supports most of streaming images except images doesn't have mipmap, image array with very small mip. It doesn't impact gpu performance on Nvidia card (RTX3070) I tested. I also did limited test on an AMD GPU which even has better gpu performance than committed resource (which is a bit of strange consider AMD suggested otherwise). I do have plan to look into the pix captures to see why it's faster.

Yes, those are called "sparse textures" on Vulkan, which is the name I know them, and as far as I know it's quite heavy (for example a single `vkQueueBindSparse` call, used to update sparse texture bindings, takes about ~600 microseconds on the CPU and about 3.5 milliseconds for full completion - tested with a fence). Sparse textures (or tiled resources) are not designed for all textures as far as I know (but as I said I'm not a sparse texture expert), did you test on a full scene representative of a full game, with a lot of textures being updated?

> An alternative solution is to re-create the device resource when the streaming image mipmap count changes. We will implement that if we find enough proof that reserved image has big performance impact on certain GPUs.

This is exactly what this PR is implementing, when a texture tag quality changes all textures using this tag will be recreated (which is costly but should only happen in an option menu).

CyaNinix:
> @CyaMinix The image tag system makes sense and I assume we can use vector for future function expansion?
> 
> Although I think it would be better the image tag quality setting is applied to StreamingImageAsset directly instead of StreamingImage. Here are the reasons:
> 
> 1. The quality setting is for StreamingImageAsset and the implementation would be clean if we handle the quality when loading StreamingImageAsset. And there is no need to add all the condition checks in StreamingImage.
> 2. There are other places other than StreamingImage which use StreamingImageAsset. Such as Terrain height map, Decal system. The tag would still be functional without any specific implementation for these places.
> 3. We have individual control for StreamingImage's target mip level at runtime via StreamingImage::SetTargetMip(). This would work well with the quality set by tag system.
> 
> To change to such an implementation, you may need
> 
> 1. Revert the changes with StreamingImage
> 2. Register the StreamingImageAsset id to ImageTagSystem and apply image quality in StreamingImageAssetHandler::LoadAssetData
> 
> * Update the ImageDescriptor
> * Update StreamingImageAsset::m_mipChains and StreamingImageAsset::m_mipLevelToChainIndex
> 
> Please let me know if you want to discuss more about the suggestion I proposed. (You may ping me on discord)

I began to work on this, if I understand correctly I have to change the `StreamingImageAsset` **before** the call to `Utils::LoadObjectFromStreamInPlace<StreamingImageAsset>(*stream, *assetData, m_serializeContext)` is made?

VickyAtAZ:
> I began to work on this, if I understand correctly I have to change the `StreamingImageAsset` **before** the call to `Utils::LoadObjectFromStreamInPlace<StreamingImageAsset>(*stream, *assetData, m_serializeContext)` is made?

It should be after this call. This call loads data from the asset file to a StreamingImageAsset. You need the loaded data to get the tags which were saved in the asset file.

CyaNinix:
Superseded by https://github.com/o3de/o3de/pull/14129

Issue #12682, opened 2022-10-19
Bug Report: Null Draw Packet Console Spam


ParasiteSt:
**Describe the bug**
The Console spams a Null Draw Packet Error when having an entity selected. 

**Steps to reproduce**
1. Start up O3DE and open a level.
2. Place an entity with a mesh component and mesh attached.
3. Select the entity in the viewport.

**Expected behavior**
There is no Null Draw Packet console spamming.

**Actual behavior**
There is one message that keeps getting executed on the console every frame.

**Screenshots/Video**
![image](https://user-images.githubusercontent.com/100494538/196579303-a7f9f134-3271-4129-9014-e14dd376a735.png)

**Found in Branch**
Development [ce8bfb7]

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Windows
 - Version: 11

LB-ArturZieba:
This also occurred on Development (18bd190), however we have noticed that restarting the Editor after first launch fixes this issue.

ParasiteSt:
Hmm, I have had this issue for a couple development updates/rebuilds, and restarting the editor does not seem to fix the issue on my end. (Around 2 weeks now)

ParasiteSt:
I just grabbed the newest Dev branch [a5e2145] and this issue still occurs. The upside is it does not seem to affect development in terms of a crash, but it does overtake the console, and makes it hard to debug aspects of your project.

PR #13015, opened 2022-11-04
Fix editor mode feedback, the pass was only added to the pipeline once.

moraaar:

## What does this PR do?

Fixes https://github.com/o3de/o3de/issues/12682

The pass of editor mode feedback was only added to the render pipeline once because it was doing an early return if the pass template was created. But when the pipelines are reinitialized (for example when setting the MSAA state) then it doesn't re-add the pass because the pass template was already created.

Updated the early return to check if the pass is in the render pipeline fixed the issue.

This PR includes other small improvements done while investigating the issue.
Now the editor mode feedback won't add null draw packets to the system and instead it will print an error if the draw packet generated was incorrect, this way it'll be easier in the future to identify the error since the message will come directly from the visual mode system.

## How was this PR tested?

- Run the editor and selected the shader ball model to check if editor visual feedback works.
- Entered into prefab mode and check the highlighting of objects worked and also the monochromatic effect worked as well.
- Tested adding a Diffuse Probe Grid component, enabled the probes, modified the sun's light and verified the probes were updated accordingly.

### DIFF
--- a/Gems/AtomLyIntegration/EditorModeFeedback/Code/Source/Pass/EditorStatePassSystem.cpp
+++ b/Gems/AtomLyIntegration/EditorModeFeedback/Code/Source/Pass/EditorStatePassSystem.cpp
@@ -49,107 +49,118 @@ namespace AZ::Render
     void EditorStatePassSystem::AddPassesToRenderPipeline(RPI::RenderPipeline* renderPipeline)
     {
         const auto templateName = Name(MainPassParentTemplateName);
-        if (RPI::PassSystemInterface::Get()->GetPassTemplate(templateName))
+
+        // Early return if pass is already found in render pipeline.
+        auto passFilter = AZ::RPI::PassFilter::CreateWithTemplateName(templateName, renderPipeline);
+        auto foundPass = AZ::RPI::PassSystemInterface::Get()->FindFirstPass(passFilter);
+        if (foundPass)
         {
-            // Template was created by another pipeline, do not to create again
             return;
         }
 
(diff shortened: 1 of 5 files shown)


Gems/AtomLyIntegration/EditorModeFeedback/Code/Source/Pass/EditorStatePassSystem.cpp
```diff
@@ -49,107 +49,118 @@ namespace AZ::Render
...
-        m_masks = CreateMaskPassTemplatesFromEditorStates(m_editorStates);
-        for (const auto& drawList : m_masks)
+        auto mainParentPassTemplate = RPI::PassSystemInterface::Get()->GetPassTemplate(templateName);
+        if (!mainParentPassTemplate)
         {
-            RPI::PassRequest pass;
-            pass.m_passName = GetMaskPassNameForDrawList(drawList);
-            pass.m_templateName = GetMaskPassTemplateNameForDrawList(drawList);
+            // Create the pass template and add it to the pass system.
+            auto newPassTemplate = AZStd::make_shared<RPI::PassTemplate>();
```
moraaar, line 65: All the code inside this block is exactly the same but using `newPassTemplate` variable and indented. Slack didn't pick up the changes very well.

VickyAtAZ (approved):

jonawals (approved):
@moraaar walked through the changes with me in a call and the EMVF look good to me 👍

hultonha:
Great work @moraaar! Excellent improvements, thank you very much for your help with this 🙂

PR #13018, opened 2022-11-04
Turned off a test on Linux that was failing because file permissions …

AMZN-stankowi:
…in our Python libraries don't work on Linux.

Updated the lock/unlock file functions to emit a log and not a warning when they do what they are called to do.

## What does this PR do?

Disables a test on Linux that does not function on Linux.

## How was this PR tested?

Tested on Windows and Linux, test now passes on Linux, continues to pass on Windows.


cgalvan (approved):

AutomatedTesting/Gem/PythonTests/assetpipeline/asset_processor_tests/asset_relocator_tests.py
```diff
@@ -344,6 +344,10 @@ def test_WindowsMacPlatforms_DeleteWithParameters_DeleteSuccess(
         4. Assert file existence or nonexistence based on the test case
         5. Validate the relocation report based on expected and unexpected messages
         """
+        if str(type(workspace)) == "<class 'ly_test_tools._internal.managers.platforms.linux.LinuxWorkspaceManager'>":
+            logger.info("Skipping test on Linux, Python based file locking does not function on Linux.")
+            return
```
smurly, line 349: won't this test execute and log as a passing test on linux? that accomplishes your goal of not failing on linux but I'm not sure that's the best way to do this.

AMZN-stankowi: Yeah, I'd rather have the test be marked skipped on Linux, but I couldn't find a clear way to conditionally skip tests that way.

Kadino: You can annotate the test:
```python
@pytest.mark.skipif(ly_test_tools.LINUX, reason="please mention why!")
```
or raise a skip-exception from inside the test:
```python
if ly_test_tools.LINUX:
    pytest.skip(reason="please mention why!")
```

smurly: I've seen CMakeLists.txt used if you look in .\o3de\AutomatedTesting\Gem\PythonTests\assetpipeline\asset_processor_tests\CMakeLists.txt line 150 there's a section you can place a suite of tests that are not to be run on linux. after that check I believe the 3 suites below there are windows mac only. so your test could go into a suite there or you could add a suite for windows mac only that resides below that platform check

AMZN-stankowi: Thanks all, that works. I've updated the test, verified it's skipping and not causing all test execution to stop on Linux, and verified it still runs on Windows.

smurly (approved):

PR #13021, opened 2022-11-04
Added an option to batch rigid body transform syncs for the entire simulation pass

SergeyAMZN:

## What does this PR do?

Added an option to batch rigid body transform syncs for the entire simulation pass
Disabled by default as needs documentation update and more real world testing

Gives up to 40% cut on physics simulation pass cost: 

batchedUpdate

## How was this PR tested?
Unit tests, manual testing various conditions: boxes, ragdolls, with & without interpolation

### DIFF
--- a/Gems/PhysX/Code/Source/ForceRegion.cpp
+++ b/Gems/PhysX/Code/Source/ForceRegion.cpp
@@ -238,7 +238,11 @@ namespace PhysX
         entityParams.m_id = entityId;
 
         AzPhysics::RigidBody* rigidBody = nullptr;
-        Physics::RigidBodyRequestBus::EventResultReverse(rigidBody, entityId, &Physics::RigidBodyRequestBus::Events::GetRigidBody);
+
+        if (auto* handler = Physics::RigidBodyRequestBus::FindFirstHandler(entityId))
+        {
+            rigidBody = handler->GetRigidBody();
+        }
 
         if (!rigidBody)
         {
--- a/Gems/PhysX/Code/Source/ForceRegion.cpp
+++ b/Gems/PhysX/Code/Source/ForceRegion.cpp
@@ -236,16 +236,21 @@ namespace PhysX
     {
         EntityParams entityParams;
         entityParams.m_id = entityId;
-        AZ::TransformBus::EventResult(entityParams.m_position
-            , entityId
-            , &AZ::TransformBus::Events::GetWorldTranslation);
-        Physics::RigidBodyRequestBus::EventResultReverse(entityParams.m_velocity
-            , entityId
-            , &Physics::RigidBodyRequestBus::Events::GetLinearVelocity);
-        Physics::RigidBodyRequestBus::EventResultReverse(entityParams.m_mass
-            , entityId
-            , &Physics::RigidBodyRequestBus::Events::GetMass);
+
+        AzPhysics::RigidBody* rigidBody = nullptr;
+        Physics::RigidBodyRequestBus::EventResultReverse(rigidBody, entityId, &Physics::RigidBodyRequestBus::Events::GetRigidBody);
+
+        if (!rigidBody)
+        {
+            AZ_Error("PhysX", false, "ForceRegionUtil::CreateEntityParams: No rigid body for entity [%llu]", entityId);
+            return entityParams;
+        }
+
+        entityParams.m_position = rigidBody->GetPosition();
+        entityParams.m_velocity = rigidBody->GetLinearVelocity();
+        entityParams.m_mass = rigidBody->GetMass();
         entityParams.m_aabb = GetForceRegionAabb(entityId);
+
         return entityParams;
     }
 
--- a/Gems/PhysX/Code/Source/Scene/PhysXScene.cpp
+++ b/Gems/PhysX/Code/Source/Scene/PhysXScene.cpp
@@ -23,7 +23,9 @@
 #include <PhysX/MathConversion.h>
 #include <Joint/PhysXJoint.h>
 
+#include <AzCore/Console/IConsole.h>
 #include <AzCore/Debug/ProfilerBus.h>
+#include <AzCore/std/algorithm.h>
 #include <AzCore/std/containers/variant.h>
 #include <AzCore/std/containers/vector.h>
 #include <AzCore/std/smart_ptr/make_shared.h>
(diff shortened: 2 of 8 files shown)


Gems/PhysX/Code/Source/Scene/PhysXScene.cpp
```diff
@@ -1287,4 +1307,46 @@ namespace PhysX
...
+        if (m_set.insert(bodyIndex).second)
+        {
+            m_array.emplace_back(bodyIndex);
+        }
+    }
+
+    void PhysXScene::QueuedActiveBodyIndices::Reserve(size_t reserveSize)
+    {
+        m_array.reserve(m_array.size() + reserveSize);
+    }
```
moraaar, line 1338: This function is doing something different than a `reserve`, it's always increasing memory of `m_array`. Why `m_array.reserve(reserveSize);` wouldn't work? 

If always increasing memory is what's needed, then maybe use another name that reflects better what's doing, for example `IncreaseCapacity` or something on those lines which is clear it's always incrementing.

Gems/PhysX/Code/Source/Scene/PhysXScene.cpp
```diff
@@ -1287,4 +1307,46 @@ namespace PhysX
...
+        AZ_PROFILE_SCOPE(Physics, "PhysX::FlushTransformSync");
+
+        m_queuedActiveBodyIndices.Apply(
+            [this](AzPhysics::SimulatedBodyIndex bodyIndex)
+            {
+                if (bodyIndex < m_simulatedBodies.size() && m_simulatedBodies[bodyIndex].second)
+                {
+                    m_simulatedBodies[bodyIndex].second->SyncTransform(m_accumulatedDeltaTime);
+                }
+            });
```
moraaar, line 1321: This loop would be much faster if what's stored in the queue is `m_simulatedBodies[bodyIndex].second` directly instead of `AzPhysics::SimulatedBodyIndex`, that way when calling `Apply` it could just do:

`simulatedBody->SyncTransform(m_accumulatedDeltaTime);`

It removes the branching called for each element in the queue. Any reason not to store the pointer to the rigid body directly?

moraaar (approved):
Awesome work @SergeyAMZN! I left a couple of improvement suggestions, but this is great.

Gems/PhysX/Code/Source/Scene/PhysXScene.cpp
SergeyAMZN: That's the point of all these SimulatedBodyHandles (which is just scene name CRC + body Index) in the first place - we don't end up with dangling pointers which could happen if we stored `SimulatedBody*` directly.

Hence we do `if (bodyIndex < m_simulatedBodies.size() && m_simulatedBodies[bodyIndex].second)` check for each bodyIndex first

Gems/PhysX/Code/Source/ForceRegion.cpp
```diff
@@ -236,16 +236,21 @@ namespace PhysX
...
-            , &AZ::TransformBus::Events::GetWorldTranslation);
-        Physics::RigidBodyRequestBus::EventResultReverse(entityParams.m_velocity
-            , entityId
-            , &Physics::RigidBodyRequestBus::Events::GetLinearVelocity);
-        Physics::RigidBodyRequestBus::EventResultReverse(entityParams.m_mass
-            , entityId
-            , &Physics::RigidBodyRequestBus::Events::GetMass);
+
+        AzPhysics::RigidBody* rigidBody = nullptr;
+        Physics::RigidBodyRequestBus::EventResultReverse(rigidBody, entityId, &Physics::RigidBodyRequestBus::Events::GetRigidBody);
```
hultonha, line 241: question: Just curious why this is `EventResultReverse` and not `EventResult` 🤔 (I know it was like this before but it just seemed unusual to me and I wondered what the reason was).

Gems/PhysX/Code/Source/Scene/PhysXScene.h
```diff
@@ -79,7 +79,27 @@ namespace PhysX
...
+        {
+        public:
+            void Insert(AzPhysics::SimulatedBodyIndex bodyIndex);
+            void Reserve(size_t reserveSize);
+            void Clear();
+            void Apply(const AZStd::function<void(AzPhysics::SimulatedBodyIndex)>& applyFunction);
+
+        private:
+            AZStd::unordered_set<AzPhysics::SimulatedBodyIndex> m_set;
+            AZStd::vector<AzPhysics::SimulatedBodyIndex> m_array;
```
hultonha, line 100: suggestion: It might be nicer to name these members after what they are and not the types, so this could be `m_uniqueIndices` for the set and `m_packedIndices` for the vector? Not a huge deal but generally tends to make the code more resilient if the types change in future.

Gems/PhysX/Code/Source/Scene/PhysXScene.h
```diff
@@ -93,11 +113,24 @@ namespace PhysX
...
+
+        // Accumulated delta time over multiple simulation sub-steps.
+        // When we run the batched transform sync, the accumulated simulation time is provided
+        // to tell how much time was simulated in this full pass.
+        float m_accumulatedDeltaTime = 0.0f;
+
         AzPhysics::SceneConfiguration m_config;
         AzPhysics::SceneHandle m_sceneHandle;
+
+        // Delta time for the current simulation sub-step
```
hultonha, line 130: +1 to the helpful comments here in the class 👍

hultonha (approved):
LGTM! Just curious with the tests/benchmarks, how do you set the CVAR? Do you have the tests running with/without? Might be worth adding a comment to the benchmark fixtures to mention how to do this? Otherwise great update! We could also talk to some LB folks to try out some more challenging scenarios to see if they find any issues.

Gems/PhysX/Code/Source/Scene/PhysXScene.h
SergeyAMZN: Nice names, I'll use them, thanks!

hultonha: Awesome thank you!

hultonha:
@SergeyAMZN Looks like a legit failure in AR

```
[2022-11-08T16:42:14.012Z] /data/workspace/o3de/Gems/PhysX/Code/Tests/Benchmarks/PhysXBenchmarksCommon.cpp:53:28: error: no type named 'PhysXScene' in namespace 'PhysX'

[2022-11-08T16:42:14.012Z]         static_cast<PhysX::PhysXScene*>(m_defaultScene)->FlushTransformSync();

[2022-11-08T16:42:14.012Z]                     ~~~~~~~^

[2022-11-08T16:42:14.012Z] /data/workspace/o3de/Gems/PhysX/Code/Tests/Benchmarks/PhysXBenchmarksCommon.cpp:53:58: error: no member named 'FlushTransformSync' in 'AzPhysics::Scene'

[2022-11-08T16:42:14.012Z]         static_cast<PhysX::PhysXScene*>(m_defaultScene)->FlushTransformSync();

[2022-11-08T16:42:14.012Z]                                         ~~~~~~~~~~~~~~   ^

```

PR #13022, opened 2022-11-04
[Terrain] Smoothing feature improvements

mbalfour-amzn:
## What does this PR do?

Thanks to a suggestion from @invertednormal , I've added "Smoothing Radius" and "Smoothing Spacing" controls to the smoothing mode. 

"Smoothing Radius" controls the number of pixels in each direction from the base pixel to use in the smoothing algorithm. It's currently capped at a range of 1-10 for performance reasons, which gives a 3x3 to 21x21 kernel for use in smoothing.

"Smoothing Spacing" controls how many pixels to skip between sampled pixels when gathering values for the smoothing algorithm. For example, a spacing of 0 used with a 3x3 smoothing will use blocks of consecutive 3x3 pixels. A spacing of 1 will sample every other pixel in a block of 6x6 pixels. A spacing of 2 will sample every third pixel in a block of 9x9 pixels. etc. It's currently capped at a range of 0-50. This provides a cheaper alternative to allowing a larger Smoothing Radius by limiting the number of total pixels sampled, but allowing the range to expand, so that more distant pixel values can have an effect in the overall smoothing. 

By using high radius and/or spacing values, it's possible to erase large features like mountains, and then use small radius/spacing values for cleaning up smaller bumps.

This PR also adds the '{' and '}' hotkeys for adjusting brush hardness.

## How was this PR tested?

Tried smoothing many features.

https://user-images.githubusercontent.com/82224783/200069323-0f61d138-d8d0-4d10-9901-e800c7781109.mp4


HogJonny-AMZN (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/PaintBrushSettings/PaintBrushSettings.h
```diff
@@ -204,6 +220,19 @@ namespace AzToolsFramework
...
         float m_distancePercent = 25.0f;
 
+        static inline constexpr size_t MinSmoothingRadius = 1;  // We need to use at least 1 pixel in each direction for smoothing
+        static inline constexpr size_t MaxSmoothingRadius = 10; // Limit the max to 10 pixels in each direction due to performance
+
+        //! Brush smoothing radius - the number of pixels in each direction to use for smoothing calculations.
+        size_t m_smoothingRadius = MinSmoothingRadius;
+
+        static inline constexpr size_t MinSmoothingSpacing = 0;     // A spacing of 0 means use adjacent pixels
+        static inline constexpr size_t MaxSmoothingSpacing = 50;    // Reasonable limit that doesn't spread the pixels too far out
```
cgalvan, line 230: Are these max limits specific to terrain performance? Would it be beneficial to expose these as overridable by different paint brush types?

cgalvan (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/PaintBrushSettings/PaintBrushSettings.h
mbalfour-amzn: They're not exactly specific to terrain, it's a bit hard for me to say if there's any benefit in exposing them per brush type. I'm inclined to leave them hard-coded for now, until we have a new brush type that clearly wants/needs different limits for these.

The smoothing radius limits here set our smoothing data queries to use anywhere from 9 - 441 data points per point we're smoothing. The smoothing spacing limits let us use a range from every point to every 50th point for querying for the smoothing, which also seems like a pretty big range.

It's possible that different underlying data types will want different ranges on these, but we can always expose these outwards later once we know for sure that different ranges are truly needed.

invertednormal:
I'm not sold on the way this is presented to the user, here's a few thoughts.

"Smoothing Radius" makes me think it's the radius over which the smoothing will happen, but it's actually the kernel size. The size of the kernel is really about _quality_. You could have a 3x3 kernel (9 samples) spread over a 50x50 area or a 10x10 kernel (100 samples) spread over the same area. They'll both do roughly the same thing, but the 10x10 kernel will give you a higher quality result. I don't know if it's valuable to give the user 10 levels of quality. 

Forcing the user to figure out a good combination for kernel size and spacing isn't very intuitive. Say they want the smoothing to average pixels over a radius of 50 pixels at medium quality. Do they go with a 10x10 kernel with a spacing of 5? or a 7x7 kernel with a spacing of 7 (except that's 49).

I'd recommend giving people a radius value that's actually the radius over which the smoothing will happen - the distance of the furthest pixel that's factored as part of the average. Then separately have a "quality" or "kernel size" setting that controls how many samples are used inside that radius.

mbalfour-amzn:
> "Smoothing Radius" makes me think it's the radius over which the smoothing will happen, but it's actually the kernel size. The size of the kernel is really about _quality_. You could have a 3x3 kernel (9 samples) spread over a 50x50 area or a 10x10 kernel (100 samples) spread over the same area. They'll both do roughly the same thing, but the 10x10 kernel will give you a higher quality result. 

I thought about calling it "kernel size", but that seems like a pretty engineering-centric description, and I'd also have to change it (I guess to a dropdown?) to make it actually select the kernel size (3x3 to 21x21). Radius feels a bit easier to work with as a slider than a dropdown, but I get your point that it's a bit confusing between that and spacing.

I disagree that kernel size means "quality" though. I'm going to speculate that in this context, you mean "quality" as "ability to reduce noise without adding artifacts", and not "subjectively better blurring results". The latter starts to go into what level of blurring is _appropriate_ for the specific image being blurred, so we'll skip that topic. :) A larger kernel will do a better job at reducing and redistributing noise across a larger area, but it can also cause more artifacts and attenuation on edges and other high-frequency details, so I don't think it's appropriate to present kernel size as "quality".

I'm definitely open to other suggestions. The UX challenge that we face here is that the other brush settings operate in world space, but "Smoothing Radius" and "Smoothing Spacing" operate in the more abstract "value space". In the context of image gradients, this is pixel space (number of pixels to use, how many pixels to skip, etc), and it still is similar enough to world space that it's easy to mistake the two. However, once paintbrushes start getting used for things like painting weights onto cloth meshes, the value space can be pretty warped in world space, so it would be nice to find a way to describe this that correctly describes the more abstract concept without being super confusing.

> I don't know if it's valuable to give the user 10 levels of quality.

I don't know that it's _not_. :P From a technical side of things, they each produce different results, and whether that's valuable or not seems to depend on the specific image I'm editing. Some images benefit more from the higher end of the range, some images benefit more from the lower end, so I'd prefer to keep this range higher instead of lower.

> Forcing the user to figure out a good combination for kernel size and spacing isn't very intuitive. Say they want the smoothing to average pixels over a radius of 50 pixels at medium quality. Do they go with a 10x10 kernel with a spacing of 5? or a 7x7 kernel with a spacing of 7 (except that's 49).

After playing with these sliders a bit, at the end of the day, I'm not sure if it really _matters_ if users understand exactly what they're doing. They'll be more focused on what result it's providing, and they'll slide the sliders back and forth to try and get the on-screen results to match what's in their heads as efficiently as possible. I don't think anyone will have a specific mental model here.

To put it another way, when I'm blurring things in Photoshop, I _never_ count the number of pixels that I'm trying to blur to make sure that my brush size is set correctly. I also don't perform the hardness calculations in my head to make sure that I've correctly selected a hardness of 51% instead of 57%.

> I'd recommend giving people a radius value that's actually the radius over which the smoothing will happen - the distance of the furthest pixel that's factored as part of the average. Then separately have a "quality" or "kernel size" setting that controls how many samples are used inside that radius.

Ideally, we'd avoid calling _either_ of these parameters a "radius", because that sounds too much like it belongs in world space.

What about the following names / tooltips?
Smoothing Values: The number of values in each direction to use for smoothing calculations (1 value in each direction = 3x3 smoothing kernel).
Smoothing Value Spacing: The spacing of values used for smoothing calculations (0 uses adjacent values, 1 uses every other value, etc).

mbalfour-amzn:
A bit more follow-up on this. 
* Even though Photoshop _labels_ the field as "radius" and "pixels", the results that it produces seems to match up with the value being used as the standard deviation.
* Gimp has a second blur filter called "Selective Gaussian Filter" that uses "Radius" to describe the smoothing kernel size.

Neither Photoshop nor Gimp exposes the smoothing kernel size for the blur _brush_, only for the blur _filter_ that works across the entire image. The blur brush in both cases uses a fixed 3x3 kernel whose size can't be changed anywhere. Paint.net doesn't have a blur brush at all.

Both Photoshop and Gimp _do_ have a "Rate" parameter on the Smoothing brush that controls the amount of blurring in a way that's similar to opacity but a little bit different. It might be controlling the amplitude of the Gaussian curve? Really not sure.

Based on all of this, I'm going to do the following:
* Continue exposing "Smoothing Radius" to control the kernel size. I'll expose it as pixel count, not standard deviation, because standard deviation is about the most unfriendly way to expose this that I can imagine. I feel OK deviating from Photoshop and Gimp here, since they don't have an equivalent control on the brush anyways.
* I've left in the underlying code for Smoothing Spacing at least for now, but removed it from the UX so that users can't set it. We'll hold off on exposing this until a day in which we explore the usefulness and UX of it further.

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
```diff
@@ -509,6 +528,58 @@ namespace AzToolsFramework
...
+
+        // We track the sum of all the weights so that we can normalize them at the end.
+        float sum = 0.0f;
+
+        // Keep a floating-point version of this so that we don't have to do int/float conversions in our loops below.
+        const float radius = aznumeric_cast<float>(smoothingRadius);
+
+        // In practice, only pixels within a distance of 3 * sigma affect the result, so we'll pick a sigma that's 1/3 of the distance
+        // in each direction.
+        const float sigma = AZStd::max(smoothingRadius / 3.0f, 1.0f);
```
invertednormal, line 549: nit: better to use radius here to avoid the cast. Compiler probably optimizes, but still seems weird to not use the value you just saved off to avoid unnecessary casts.

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
```diff
@@ -509,6 +528,58 @@ namespace AzToolsFramework
...
+        gaussianWeights.reserve(kernelSize * kernelSize);
+
+        // Compute the Gaussian weights for each point spreading outward from the origin.
+        // The order in which we compute this is important, because we'll expect the pixel values that we're smoothing to have been
+        // gathered in the same order.
+        for (float y = -radius; y <= radius; y++)
+        {
+            for (float x = -radius; x <= radius; x++)
+            {
+                float kernelValue = peakHeight * AZStd::exp(expConstant * ((x * x) + (y * y)));
```
invertednormal, line 566: super-duper-nit: `y * y` could be calculated outside the x loop.

invertednormal (approved):
Looks good, just some nits.

For future reference - As we talked about in person, it'd be nice to eventually move this to a two-pass solution. Since gaussian blurs are separable the entire blur could be done horizontally first, then vertically to turn the problem from N * N samples per pixel to just 2N. For fancy brushes this may mean throwing a lot of pixel values away (since the blur would need to be done for the entire region of the brush) but the improvement in cost would still be worth it, particularly for large blurs. For instance, with a kernel size of 100, you can afford to throw away a lot of pixels and still be much cheaper (N * N = 10,000 vs 2N = 200).

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
```diff
@@ -29,16 +28,30 @@ namespace AzToolsFramework
...
     {
         static constexpr AZ::Crc32 PaintbrushIncreaseSize = AZ_CRC_CE("org.o3de.action.paintbrush.increase_size");
         static constexpr AZ::Crc32 PaintbrushDecreaseSize = AZ_CRC_CE("org.o3de.action.paintbrush.decrease_size");
+        static constexpr AZ::Crc32 PaintbrushIncreaseHardness = AZ_CRC_CE("org.o3de.action.paintbrush.increase_hardness_percent");
+        static constexpr AZ::Crc32 PaintbrushDecreaseHardness = AZ_CRC_CE("org.o3de.action.paintbrush.decrease_hardness_percent");
 
         static constexpr const char* PaintbrushIncreaseSizeTitle = "Increase Size";
         static constexpr const char* PaintbrushDecreaseSizeTitle = "Decrease Size";
+        static constexpr const char* PaintbrushIncreaseHardnessTitle = "Increase Hardness Percent";
+        static constexpr const char* PaintbrushDecreaseHardnessTitle = "Decrease Hardness Percent";
```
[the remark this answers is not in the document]
mbalfour-amzn: Nearly all of the other component mode actions use this style of capitalization, so I'll continue to stay consistent with them for now.

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
```diff
@@ -390,6 +403,82 @@ namespace AzToolsFramework
...
 
+    PaintBrushNotifications::BlendFn PaintBrushManipulator::GetBlendFunction(const PaintBrushSettings& brushSettings)
+    {
+        switch (brushSettings.GetBlendMode())
+        {
+        case PaintBrushBlendMode::Normal:
+            return [](float baseValue, float intensity, float opacity)
+            {
+                const float operationResult = intensity;
+                return AZStd::clamp(AZStd::lerp(baseValue, operationResult, opacity), 0.0f, 1.0f);
```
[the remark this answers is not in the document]
mbalfour-amzn: I've marginally improved this. :)

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
```diff
@@ -509,6 +528,58 @@ namespace AzToolsFramework
...
+
+        // Keep a floating-point version of this so that we don't have to do int/float conversions in our loops below.
+        const float radius = aznumeric_cast<float>(smoothingRadius);
+
+        // In practice, only pixels within a distance of 3 * sigma affect the result, so we'll pick a sigma that's 1/3 of the distance
+        // in each direction.
+        const float sigma = AZStd::max(smoothingRadius / 3.0f, 1.0f);
+
+        // We can precompute the peakHeight and expConstant terms outside of the loop.
+        const float peakHeight = (1.0f / (AZ::Constants::TwoPi * sigma * sigma));
```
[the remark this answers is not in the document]
mbalfour-amzn: Yeah fair point, I think I prefer it the way I've written it though, since it's nice and easy to see how it matches up with the Gaussian function. These are only calculated once per paint operation, not per-pixel, so I'm not worried about the cost savings from condensing it.

PR #13023, opened 2022-11-04
Force Component Mode to close when exiting the Editor.

mbalfour-amzn:
## What does this PR do?

It was possible to select "Exit" while in Component Mode, which could cause the work currently being done in Component Mode to get lost. This gives Component Mode a chance to save the work before exiting.

In particular, this improves two behaviors for the Image Gradient Component Mode:
1. Ensures that the user has a chance to save the image when Exiting the Editor.
2. Ensures that the Paint Brush Settings window closes cleanly and doesn't reappear on the next run of the Editor.

## How was this PR tested?

Exited the Editor while in Component Mode and verified that it ended Component Mode before exiting.

### DIFF
--- a/Code/Editor/CryEditDoc.cpp
+++ b/Code/Editor/CryEditDoc.cpp
@@ -28,6 +28,7 @@
 #include <AzFramework/API/ApplicationAPI.h>
 
 // AzToolsFramework
+#include <AzToolsFramework/ComponentMode/EditorComponentModeBus.h>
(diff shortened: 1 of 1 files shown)


AMZN-daimini (approved):

cgalvan (approved):

hultonha (approved):
LGTM! Thanks for adding this @mbalfour-amzn

PR #13024, opened 2022-11-04
Improve DPE performance by recycling label and row widgets

alexmontAmazon:
## What does this PR do?
- creates a new InstancePool class that allows for classes to be recycled instead of constructed and destructed
- consumes the InstancePool in the DPE for recycling labels and row wdigets

## How was this PR tested?

locally with the Editor and DPE Standalone apps


Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
```diff
@@ -435,53 +472,64 @@ namespace AzToolsFramework
...
+            if (childWidget)
             {
-                QWidget* propertyWidget = propertyWidgetIter->first;
-                auto toRemove = AZStd::remove(m_domOrderedChildren.begin(), m_domOrderedChildren.end(), propertyWidget);
-                m_domOrderedChildren.erase(toRemove, m_domOrderedChildren.end());
-                propertyWidget->hide();
-                m_columnLayout->removeWidget(propertyWidget);
-                propertyWidget->setParent(nullptr);
-                dpe->ReleaseHandler(AZStd::move(propertyWidgetIter->second.hanlderInterface));
+                if (auto foundPropertIter = m_widgetToPropertyHandlerInfo.find(childWidget);
```
amzn-tmryan, line 501: nitpick: typo here in "foundPropertIter" -> "foundPropertyIter"

amzn-tmryan (approved):
Looks good to me!

Code/Framework/AzCore/AzCore/Instance/InstancePool.h
```diff
@@ -0,0 +1,157 @@
...
+#include <AzCore/std/containers/vector.h>
+#include <AzCore/std/containers/unordered_map.h>
+#include <AzCore/std/smart_ptr/unique_ptr.h>
+#include <AzCore/std/smart_ptr/weak_ptr.h>
+#include <AzCore/std/smart_ptr/shared_ptr.h>
+#include <AzCore/std/smart_ptr/make_shared.h>
+#include <AzCore/Outcome/Outcome.h>
+#include <AzCore/std/string/string.h>
+#include <AzCore/Name/Name.h>
+#include <AzCore/std/utils.h>
```
lsemp3d, line 20: Are all these absolutely required? It's a lot of headers in a header, maybe some can be fwd declared

lsemp3d (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
```diff
@@ -32,14 +32,50 @@ AZ_CVAR(
...
+        m_cachedMinLayoutSize = QSize();
+        for (auto index = count() - 1; index >= 0; --index)
+        {
+            auto theItem = takeAt(index);
+            auto theWidget = theItem->widget();
+            if (theWidget)
+            {
+                theWidget->deleteLater();
+            }
+        }
```
lumberyard-employee-dm, line 76: FYI:"This only works when `count()` returns a signed type.

When an unsigned type is used, this would result in an infinite loop.
Luckily Qt `count()` function returns an `int`, but the code can be authored in a way that works for both signed and unsigned cases by deferring decrementing the index until it is used
```suggestion
        for (int index = count(); index > 0; --index)
        {
            auto theItem = takeAt(index - 1);
            auto theWidget = theItem->widget();
            if (theWidget)
            {
                theWidget->deleteLater();
            }
        }
```

alexmontAmazon: good catch! I'll take that change directly :)

Code/Framework/AzCore/AzCore/Instance/InstancePool.h
alexmontAmazon: I got rid of a couple! I can't get rid of any more because it's all templated :-/

Issue #12745, opened 2022-10-21
Intermediate Assets Can Cause Transient Asset Processor Failures


santorac:
**Describe the bug**
I encountered this issue while utilizing intermediate asset processing for the new material pipeline system. See https://github.com/o3de/o3de/pull/12717

I have asset A producing intermediate assets B and C. It looks like the AP sees B before it sees C, so it processes B and fails (because when a job dependency is missing, the AP ignores it the dependency). Then later it discovers C, processes C, and then processes B again and passes. (This occurred in 1 of 3 Jenkins runs)

Similarly (but possibly a separate underlying issue), there is another source file D that has a source dependency on the intermediate file B. The AP processes D before processing A, so the intermediate file is missing, and D fails.

In both cases, everything builds correctly in the end. C is finally discovered, the AP reprocesses B and passes, then reprocesses D and passes. **But** Jenkins finds failure messages in the log and so fails the AR process. This blocks us from being able to check in assets that use the material pipeline. (This occurred in 3 of 3 Jenkins runs)

There are couple instances of this situation in the log, one for a "Basic" material type and one for an "Animated" material type. Here is a concrete breakdown of the files involved for one of these instances:
A=MaterialPipelineTest_Animated.materialtype (Source) It produces B and C.
B=materialpipelinetest_animated_generated.materialtype (Intermediate) It has a job dependency on C.
C=MaterialPipelineTest_Animated_LowEndPipeline_ForwardPass.shader (Intermediate)
D=MaterialPipelineTest_Animated.material (Source) It has a source dependency on B.

The original failure log is here: https://jenkins.build.o3de.org/job/O3DE/view/change-requests/job/PR-12717/2/execution/node/501/log/?consoleFull
I have attached a copy here: [APLog.txt](https://github.com/o3de/o3de/files/9841382/APLog.txt)

For now I am going to work around this issue by disabling the material pipeline system in the AutomatedTesting project.

**Assets required**
This will be a bit tricky, since I have changes in flight on a feature branch, and I'm going to work around the issue by disabling the material pipeline system in the AutomatedTesting project. If the code has not been merged yet, you'll need to check out my branch at https://github.com/aws-lumberyard-dev/o3de/tree/Atom/santorac/MaterialPipeline1. Also checkout AtomSampleViewer, again using this branch if the changes have not been merged yet: https://github.com/aws-lumberyard-dev/o3de-atom-sampleviewer/tree/Atom/santorac/MaterialPipeline1

[...]

PR #13027, opened 2022-11-04
Asset Processor - Disable file state cache for asset processor batch

amzn-mike:
## What does this PR do?

Disables the file state cache for APB.

The file state cache requires the file monitor to update its state but the file monitor is disabled in batch mode. Since intermediate assets are created while running, this meant the cache would be out of date.

This was causing a particular problem for atom systems which attempt to look up dependency files using the GetSourceInfo API before declaring a dependency.  The API was always returning that intermediate asset files did not exist, resulting in the atom builder reporting the dependency as a source dependency instead of a job dependency (this is common behavior for atom builders because they work with relative paths that can refer to multiple locations and don't want to spam job dependency errors by reporting dependencies on files that will never exist).

Running batch on a clean cache with the file state cache ENABLED took 737s (approx 12.2 minutes)
With the file state cache DISABLED the same test took 843s (approx 14 minutes)

This change only affects asset processor BATCH and does not affect the GUI mode.

Partial fix for https://github.com/o3de/o3de/issues/12745

## How was this PR tested?

Manually ran APB and verified the passthrough cache is created instead of the actual state cache.

### DIFF
--- a/Code/Tools/AssetProcessor/native/tests/ApplicationManagerTests.cpp
+++ b/Code/Tools/AssetProcessor/native/tests/ApplicationManagerTests.cpp
@@ -16,13 +16,14 @@
 #include <native/resourcecompiler/rcjob.h>
 #include <AzCore/Utils/Utils.h>
 #include <AzCore/Settings/SettingsRegistryImpl.h>
+#include <unittests/UnitTestUtils.h>
 
 namespace UnitTests
 {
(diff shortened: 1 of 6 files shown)


junbo75 (commented):
Looks good to me. Any automation tests we can add for this change?

AMZN-stankowi (approved):

jackalbe (approved):

PR #13029, opened 2022-11-05
Removed strict dependency of O3DE on the MovieSystem (which is a gem)

lsemp3d:
Signed-off-by: Luis Sempé <58790905+lsemp3d@users.noreply.github.com>

## What does this PR do?

The MovieSystem is implemented by the Maestro gem, if this gem is not part of the project or built into the engine the engine would fail to run. Given that gems are optional components, I removed the strict dependency and safeguarded calls to retrieve the IMovieSystem. 

No longer show TrackView as an option if no MovieSystem is present.

## How was this PR tested?

Ran the editor, verified no crashes, then verified TrackView no longer shows in the tools menu


cgalvan (approved):
Nice!

Issue #12986, opened 2022-11-03
Material Canvas nodes have inconsistent letter casing


LB-ArturZieba:
**Describe the bug**
Nodes in the Material Canvas have inconsistent letter casing. Nodes in the *Math Functions* and *Scene* categories are starting with a lower case letter, while all others are starting with upper case letter.

**Steps to reproduce**
1. Go to Tools > Material Canvas.
2. Expand all of the categories in the Node Palette.

**Expected behavior**
All node names have consistent letter casing.

**Actual behavior**
Node names in different categories have different letter casing.

**Material Canvas Node Palette Screenshots**
![MaterialCanvasNodes0](https://user-images.githubusercontent.com/86952082/199730567-38ddb0b7-c73f-4ace-b77e-567a86d97776.png)
![MaterialCanvasNodes1](https://user-images.githubusercontent.com/86952082/199730575-8b9874df-c3e5-4180-8ad8-bd9565854b77.png)
![MaterialCanvasNodes2](https://user-images.githubusercontent.com/86952082/199730578-dcc56b1c-5f5c-41f9-92af-86db242b12eb.png)

**Found in Branch**
Development (1e41006)

**Desktop**
 - Device: PC
 - OS: Linux
 - Version: Ubuntu 20.04.3 LTS
 - CPU AMD Ryzen 5 3600
 - GPU NVIDIA GeForce RTX 2060 SUPER
 - Memory 16GB

lsemp3d:
@gadams3 I recommend using the GraphCanvas Translation system for this, it makes the updating of all these elements data driven, even in gems (for example, if someone were to create a new gem with custom nodes). This way you can take advantage of the `.names` files which allow users to provide names, tooltips, and more, both on the node palette and on the nodes themselves.

LB-ArturZieba:
On the Development (7f642a1) build there is still one node with lower case name - Math Functions > Noise:

![MaterialCanvasLowerCase](https://user-images.githubusercontent.com/86952082/200563820-4f807e1a-0565-47de-8978-259c2d641257.png)

It should be noted that all other letter cases were fixed.

Issue #12492, opened 2022-10-07
Material Canvas nodes do not have distinct coloration


LB-ArturZieba:
**Describe the bug**
All Material Canvas nodes have indistinct colors (or in the case of the Utility category none at all), while for example Script Canvas has different colors for different types of nodes/categories.

This issue occurs on both Windows and Linux OS.

This issue also occurs on the Development (3f742f7) build.

**Steps to reproduce**
1. Go to Tools > Material Canvas (Experimental).
2. Look at the Node Palette.

**Expected behavior**
All categories/types of nodes have distinct coloration.

**Actual behavior**
All categories/types of nodes are colored black or in the case of Utilities no color is assigned.

**Screenshot**
![MaterialCanvasNodesColors](https://user-images.githubusercontent.com/86952082/194562011-a1db0e47-8715-4651-9a8c-5027b5abf37e.png)

**Found in Branch**
Stabilization/2210 (7c4e3ca)

**Desktop**
 - Device: PC
 - OS: Linux
 - Version: Ubuntu 20.04.3 LTS
 - CPU AMD Ryzen 5 3600
 - GPU NVIDIA GeForce RTX 2060 SUPER
 - Memory 16GB

monroegm:
SIG-Graphics-Audio own Material Canvas

LB-ArturZieba:
On the Development (7f642a1) build there are still some nodes with no color:
* Math Functions > Acos
* Math Functions > Noise
* Utilities > Comment
* Utilities > Node Group

![MaterialCanvasNoColor0](https://user-images.githubusercontent.com/86952082/200564353-66a268cf-8d87-4fae-b419-8ebe3e21281d.png)
![MaterialCanvasNoColor1](https://user-images.githubusercontent.com/86952082/200564373-7329c8f8-1821-4469-ac42-95356e065c4b.png)
![MaterialCanvasNoColor2](https://user-images.githubusercontent.com/86952082/200564389-4c624124-1541-4271-a155-e45d4dddbfad.png)

It should be noted that all other nodes have a color assigned.

Issue #12152, opened 2022-09-22
Non-zero return code after test success when running MaterialCanvas automated tests (successes should return exit code 0).


jromnoa:
**Describe the bug**
Exit code 0xc0000005 is being returned from the MaterialCanvas automated tests referencing a crash in C:/git/o3de/Gems/AtomTressFX/Code/Passes/HairGeometryRasterPass.cpp. The tests are not yet merged to our repo but are currently in a PR: https://github.com/o3de/o3de/pull/11915 
I may mark the tests skip if the bug is hard to fix or requires substantial time. I can remove the skip once this bug is fixed.

**Assets required**
You just need the MaterialCanvas program built and have the ability to launch python automated tests locally. Also make sure to process the assets before running the test so that no timeout failures occur during the automated test run.

**Steps to reproduce**
1. Run the test from https://github.com/o3de/o3de/pull/11915 locally (its the first automated MaterialCanvas test): `C:\git\o3de\python\runtime\python-3.10.5-rev1-windows\python\python.exe -m pytest -vv -s --build-directory="C:\git\o3de\build\bin\profile" C:\git\o3de\AutomatedTesting\Gem\PythonTests\Atom\TestSuite_Main_Null_Render_MaterialCanvas_01.py`
2. Observe the test output (side note: the test doesn't really do anything other than run MaterialCanvas, verify it launched, then closes it so I 'm not sure why the HairGeometryRasterPass.cpp code would cause a non-zero return code on exit since there are no visible errors during the run).

**Expected behavior**
 An exit code of 0 success is generated with no log errors.

**Actual behavior**
An exit code of 0xc0000005 is returned with a crash referencing C:/git/o3de/Gems/AtomTressFX/Code/Passes/HairGeometryRasterPass.cpp:
```
E   Starting test MaterialCanvas_Launched_SuccessfullyLaunched...
E   Test MaterialCanvas_Launched_SuccessfullyLaunched finished.
E   Report:
E   [SUCCESS] Success: Inspector pane is visible, MaterialCanvas launch succeeded
E   Test result:  SUCCESS
E   JSON_START({"name": "MaterialCanvas_Atom_LaunchMaterialCanvas", "success": true, "exception": null, "output": "Test MaterialCanvas_Launched_SuccessfullyLaunched finished.\nReport:\n[SUCCESS] Success: Inspector pane is visible, MaterialCanvas launch succeeded\nTest result:  SUCCESS"})JSON_END
[...]

PR #13039, opened 2022-11-07
Material Canvas: Stability, Simplification, Styling, Symbols, Substitutions, Slot Conversion, Slot Value Promotion, Shut Down Crash Fixes (Except for one QT crash TBD)

gadams3:
## What does this PR do?

- Material canvas node configurations and code generation no longer require slot ID and node ID prefixes. Undecorated slot and simple names can be used throughout instruction and material input settings blocks. New macros have been added for accessing node and slot information from other slots.
- A lot of stabilization and clean up related to code generation and type conversions in material canvas document.
- New node configuration settings for setting up groups of slots that should all share the same data type.
- Separated settings for slot editability and visibility on node UI.
- Fixed multiple shutdown crashes caused by the graph model system component and graph controller manager not cleaning up before the system gonna deactivates.
- Changed the graph model clipboard serialization data to actually serialize nodes into byte buffers instead of holding onto share pointers that would not be released until application shut down. This was causing problems with nodes with slots containing image assets as values. The assets needed to be released before their Asset handlers.
- Updated the graph model integer display data interface to be template sized so that it could be used for other integer types. Registered unsigned int, long, unsigned long.
- Fixed a bug where node title palette styling was not being applied to previously serialized nodes.
- More graph model node and slot convenience functions.

Resolves https://github.com/o3de/o3de/issues/12986
Resolves https://github.com/o3de/o3de/issues/12492
Progress on https://github.com/o3de/o3de/issues/12152 

## How was this PR tested?

creating several test graphs with different combinations of nodes and confirming they all generate the expected results
a lot of undo redo shutting down restarting
QT crash still shows up occasionally but will discuss with sig content 

https://user-images.githubusercontent.com/82461473/200394100-01201540-5c25-4fcf-baf2-ec11813cda63.mp4

Still investigating this
image


Gems/Atom/Tools/AtomToolsFramework/Code/Source/DynamicNode/DynamicNodeConfig.cpp
```diff
@@ -146,20 +156,25 @@ namespace AtomToolsFramework
...
+    AZStd::vector<AZStd::string> DynamicNodeConfig::GetSlotNames() const
+    {
+        AZStd::vector<AZStd::string> slotNames;
+        VisitDynamicNodeSlotConfigs(
+            *this,
+            [&](const DynamicNodeSlotConfig& slotConfig)
+            {
+                slotNames.push_back(slotConfig.m_name);
+            });
+        return slotNames;
```
rgba16f, line 176: nit: only need to capture slotNames rather than everything. Not a blocker

Gems/Atom/Tools/AtomToolsFramework/Code/Include/AtomToolsFramework/Application/AtomToolsApplication.h
```diff
@@ -32,8 +32,8 @@ namespace AtomToolsFramework
 {
     //! Base class for Atom tools to inherit from
     class AtomToolsApplication
-        : public AzFramework::Application
-        , public AzQtComponents::AzQtApplication
+        : public AzQtComponents::AzQtApplication
+        , public AzFramework::Application
```
cgalvan, line 36: What does this ordering change do?

Gems/Atom/Tools/AtomToolsFramework/Code/Source/DynamicNode/DynamicNodeManager.cpp
```diff
@@ -120,8 +120,11 @@ namespace AtomToolsFramework
             auto categoryItr = categoryMap.find(config.m_category);
             if (categoryItr == categoryMap.end())
             {
-                categoryItr = categoryMap.emplace(config.m_category,
-                    rootItem->CreateChildNode<GraphCanvas::IconDecoratedNodePaletteTreeItem>(config.m_category, m_toolId)).first;
+                // Creating a new node pilot tree item category and setting the title palette based on the 1st dynamic node config added
```
cgalvan, line 123:
```suggestion
                // Creating a new node palette tree item category and setting the title palette based on the 1st dynamic node config added
```

Gems/Atom/Tools/AtomToolsFramework/Code/Source/DynamicNode/DynamicNodeSlotConfig.cpp
```diff
@@ -31,7 +31,9 @@ namespace AtomToolsFramework
                 ->Field("supportedDataTypeRegex", &DynamicNodeSlotConfig::m_supportedDataTypeRegex)
                 ->Field("defaultDataType", &DynamicNodeSlotConfig::m_defaultDataType)
                 ->Field("defaultValue", &DynamicNodeSlotConfig::m_defaultValue)
-                ->Field("supportsEditingOnNode", &DynamicNodeSlotConfig::m_supportsEditingOnNode)
```
cgalvan, line 34: Are we losing data with this change? I didn't see a json converter. Or is it OK since Material Canvas is still experimental?

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Util/Util.cpp
```diff
@@ -88,23 +88,36 @@ namespace AtomToolsFramework
...
+        // Tokenize the string where separated by whitespace
         QStringList displayNameParts = displayName.split(QRegExp("\\s"), Qt::SkipEmptyParts);
         for (QString& part : displayNameParts)
         {
+            // Capitalize the first character of every token
             part.replace(0, 1, part[0].toUpper());
         }
+        // Recombine all of the strings separated by a single space
         return displayNameParts.join(" ").toUtf8().constData();
     }
```
rgba16f, line 122: Seems like these could have a shared impl that takes a lambda for the non-shared code.

cgalvan (approved):

Gems/Atom/Tools/MaterialCanvas/Code/Source/Document/MaterialCanvasDocument.cpp
```diff
@@ -793,67 +971,108 @@ namespace MaterialCanvas
...
         if (auto slot = node->GetSlot(slotConfig.m_name))
         {
             AtomToolsFramework::CollectDynamicNodeSettings(slotConfig.m_settings, "materialInputs", materialInputsForSlot);
 
-            ReplaceStringsInContainer("SLOTSTANDARDSRGMEMBER", ConvertSlotValueToSrgMember(node, slotConfig), materialInputsForSlot);
-            ReplaceStringsInContainer("SLOTNAME", slot->GetName().c_str(), materialInputsForSlot);
-            ReplaceStringsInContainer("SLOTTYPE", ConvertSlotTypeToAZSL(slot->GetDataType()->GetDisplayName()), materialInputsForSlot);
-            ReplaceStringsInContainer("SLOTVALUE", ConvertSlotValueToAZSL(slot->GetValue()), materialInputsForSlot);
+            ReplaceSymbolsInContainer(substitutionSymbols, materialInputsForSlot);
+            ReplaceSymbolsInContainer("SLOTSTANDARDSRGMEMBER", GetAzslSrgMemberFromSlot(node, slotConfig), materialInputsForSlot);
```
rgba16f, line 1072: This line is missing from the earlier function that also does the name substitution, does it need to be added there?

rgba16f (approved):

Gems/Atom/Tools/AtomToolsFramework/Code/Include/AtomToolsFramework/Application/AtomToolsApplication.h
gadams3: I was experimenting with destruction order for the crash

Gems/Atom/Tools/AtomToolsFramework/Code/Source/DynamicNode/DynamicNodeSlotConfig.cpp
gadams3: No data loss. All of the notes in the test folder were updated to use the new flags. It was only used in a handful of places.

PR #13043, opened 2022-11-07
[Math] Fixed the fundamentally broken ClipRayWithAabb.

mbalfour-amzn:
## What does this PR do?

This method calculated rayEnd wrong, and calculated tClipStart / End wrong when one end was inside the AABB. The problems were discovered while debugging some broken terrain raycasts, which is the only code in the engine that currently uses ClipRayWithAabb.

## How was this PR tested?

Created unit tests that validated all the behaviors that were broken, then fixed the function, then verified the unit tests passed.


cgalvan (approved):

lumberyard-employee-dm (approved):

Code/Framework/AzCore/Tests/Math/IntersectionTests.cpp
```diff
@@ -1297,4 +1297,96 @@ namespace UnitTest
...
+
+        float tClipStart = AZStd::numeric_limits<float>::max();
+        float tClipEnd = AZStd::numeric_limits<float>::max();
+
+        bool clipped = Intersect::ClipRayWithAabb(aabb, rayStart, rayEnd, tClipStart, tClipEnd);
+
+        // We expect the ray to get clipped.
+        EXPECT_TRUE(clipped);
+        EXPECT_EQ(rayStart, AZ::Vector3(5.0f, 5.0f, 10.0f));
+        EXPECT_EQ(rayEnd, AZ::Vector3(5.0f, 5.0f, -10.0f));
```
hultonha, line 1342: update: It would likely be good to update this to `EXPECT_THAT(rayStart, IsClose(AZ::Vector3(...));` as that will do an epsilon check and also give more informative output if it fails

Code/Framework/AzCore/Tests/Math/IntersectionTests.cpp
```diff
@@ -1297,4 +1297,96 @@ namespace UnitTest
...
+
+        float tClipStart = AZStd::numeric_limits<float>::max();
+        float tClipEnd = AZStd::numeric_limits<float>::max();
+
+        bool clipped = Intersect::ClipRayWithAabb(aabb, rayStart, rayEnd, tClipStart, tClipEnd);
+
+        // We expect the ray to get clipped.
+        EXPECT_TRUE(clipped);
+        EXPECT_EQ(rayStart, AZ::Vector3(5.0f, 5.0f, 0.0f));
+        EXPECT_EQ(rayEnd, AZ::Vector3(5.0f, 5.0f, -10.0f));
```
hultonha, line 1365: update: Same comment about `IsClose` matcher

Code/Framework/AzCore/Tests/Math/IntersectionTests.cpp
```diff
@@ -1297,4 +1297,96 @@ namespace UnitTest
...
+
+        float tClipStart = AZStd::numeric_limits<float>::max();
+        float tClipEnd = AZStd::numeric_limits<float>::max();
+
+        bool clipped = Intersect::ClipRayWithAabb(aabb, rayStart, rayEnd, tClipStart, tClipEnd);
+
+        // We expect the ray to get clipped.
+        EXPECT_TRUE(clipped);
+        EXPECT_EQ(rayStart, AZ::Vector3(5.0f, 5.0f, 10.0f));
+        EXPECT_EQ(rayEnd, AZ::Vector3(5.0f, 5.0f, 0.0f));
```
hultonha, line 1388: update: `IsClose` and `EXPECT_THAT` as mentioned before

hultonha (approved):
Change looks great and thank you for adding the tests. I do have a suggestion about using the `IsClose` matcher for the `Vector3` tests but otherwise awesome stuff!

Code/Framework/AzCore/Tests/Math/IntersectionTests.cpp
hultonha: 5.0 and 10.0 can be represented exactly by floating point but it's possible error may be introduced in `ClipRayWithAabb` and so this test has a higher chance of failing with other values in future potentially

mbalfour-amzn:
Going to open a different more expansive PR that adds fixes to the Terrain Raycast code, since they both need to be submitted at the same time to avoid breaking the paintbrush.

PR #13051, opened 2022-11-07
Asset Processor - Fix APB early termination

amzn-mike:
## What does this PR do?

Fixed an issue where asset processor batch would exit early before all processing had finished on intermediate assets.

APB was previously set up to operate under the assumption that once APM went idle followed by RC going idle, that there was no more work to do and shut down could start.  With intermediate assets, RC can finish its work and end up creating more work for APM to do.  This fix adds an additional check for the m_remainingJobsForEachSourceFile to hit zero before shutdown can occur.  The count is emitted in a new signal, FinishedAnalysis, which APB will listen to.

This also fixes 2 cases where FinishAnalysis was not being called because rccontroller was discarding duplicate jobs without notifying APM.

Additionally fixes a minor issue where Jobs loaded from the database were not setting the m_watchFolder property, causing the hash of the Job to be different when compared to in-memory generated jobs.  This resulted in job logs sometimes being unable to be loaded when an error occurred.

Partial fix for: https://github.com/o3de/o3de/issues/12745

## How was this PR tested?

Manually running APB locally.  Also added and ran automated tests.


Code/Tools/AssetProcessor/native/tests/assetmanager/AssetManagerTestingBase.h
```diff
@@ -43,9 +43,9 @@ namespace UnitTests
         }
 
         void WaitForFinish()
-        {
-            m_signal.acquire();
-        }
+            {
```
AMZN-stankowi, line 46: why did this get tabbed in?

AMZN-stankowi (approved):

Code/Tools/AssetProcessor/native/tests/assetmanager/AssetManagerTestingBase.h
amzn-mike: Probably something with VS auto formatting, mine's been acting weird lately.  Fixed

cgalvan (approved):

PR #13061, opened 2022-11-08
Added performance metrics to PhysX gem

SergeyAMZN:
## What does this PR do?

PhysX gem integration with Performance Metrics system. 
Currently only simulation sub-step time is measured. 
More metrics to come in the future PRs.

## How was this PR tested?

Manual testing with and without CVars.


moraaar (approved):
The code change looks good, but maybe I'd move all the code to another cpp file to avoid keep making PhysXSystem bigger, for example to a `PhysXDebug.cpp`. That way all the cvars would be in that debug cpp.

In addition all the code could be wrapped around a debug class and PhysXSystem only instantiate it if it's not release configuration.

I'll approve it since the code is correct, but it'd be great to separate it a bit and avoid creating it in release configurations.

lumberyard-employee-dm:
> The code change looks good, but maybe I'd move all the code to another cpp file to avoid keep making PhysXSystem bigger, for example to a `PhysXDebug.cpp`. That way all the cvars would be in that debug cpp.
> 
> In addition all the code could be wrapped around a debug class and PhysXSystem only instantiate it if it's not release configuration.
> 
> I'll approve it since the code is correct, but it'd be great to separate it a bit and avoid creating it in release configurations.

The Json TraceEvent Logger by default doesn't record in [release](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Metrics/JsonTraceEventLogger.cpp#L160-L169) configuration.

Gems/PhysX/Code/Source/System/PhysXSystem.cpp
```diff
@@ -41,6 +45,51 @@ namespace PhysX
...
+        AZ::ConsoleFunctorFlags::DontReplicate, "Collects and reports PhysX performance in this number of batches.");
+
+    AZ_CVAR(AZ::CVarFixedString, physx_metricsDataLogType,
+        "statistical", // (s)(S)tatistical Summary (average, min, max, stdev) / (a)(A)ll Samples. Default=s.
+        [](const AZ::CVarFixedString& newValue)
+        {
+            PhysX::GetPhysXSystem()->GetPerformanceCollector()->UpdateDataLogType(GetDataLogTypeFromCVar(newValue));
+        },
+        AZ::ConsoleFunctorFlags::DontReplicate, "Defines the kind of data collection and logging. "
+        "If starts with 's' it will log statistical summaries, if starts with 'a' will log each sample of data (high verbosity).");
```
hultonha, line 83: suggestion: Is it worth mentioning the `s`/`S` bit from the code comment in this block for users?

Gems/PhysX/Code/Source/System/PhysXSystem.cpp
```diff
@@ -41,6 +45,51 @@ namespace PhysX
...
+
+    AZ_CVAR(AZ::u32, physx_metricsWaitTimePerCaptureBatch, 0,
+        [](const AZ::u32& newValue)
```
hultonha, line 86: uber nit: Could remove the `&` here as this is just a small builtin type

Gems/PhysX/Code/Source/System/PhysXSystem.cpp
```diff
@@ -51,6 +100,30 @@ namespace PhysX
...
         InitializePhysXSdk(cookingParams);
+
+        InitializePerformanceCollector();
+    }
+
+    void PhysXSystem::InitializePerformanceCollector()
+    {
+        auto performanceMetrics = AZStd::to_array<AZStd::string_view>({
+            PerformanceSpecPhysXSimulationTime,
+        });
```
hultonha, line 111: uber nit/standards: Brace placement (not a big deal and you can ignore just couldn't help mentioning it 😅)

hultonha (approved):
LGTM! Thanks for adding these @SergeyAMZN, I look forward to seeing the output! 🙂

Gems/PhysX/Code/Source/System/PhysXSystem.cpp
```diff
@@ -46,6 +50,51 @@ namespace PhysX
...
+    {
+        if (newCaptureType[0] == 'a' || newCaptureType[0] == 'A')
+        {
+            return AZ::Debug::PerformanceCollector::DataLogType::LogAllSamples;
+        }
+        else
+        {
+            return AZ::Debug::PerformanceCollector::DataLogType::LogStatistics;
+        }
+    }
```
lumberyard-employee-dm, line 64: `fixed_string<N>::starts_with` should be used here as this will cause an out of bound access if the `newCaptureType` is an empty string

lumberyard-employee-dm (changes requested):

Gems/PhysX/Code/Source/System/PhysXSystem.cpp
SergeyAMZN: awesome, thanks! 
Updated the Atom code with the fix too

Gems/PhysX/Code/Source/System/PhysXSystem.cpp
SergeyAMZN: That's actually how VS auto-formats it for me (using clang-format?). 
Maybe the global settings need updating to comply with coding standard?

Gems/PhysX/Code/Source/System/PhysXSystem.cpp
SergeyAMZN: Yeah, merged the code comment with the description. No need to duplicate the same info. 
Thanks!

hultonha: Nice one thanks!

Gems/PhysX/Code/Source/System/PhysXSystem.cpp
hultonha: Ah that's fine, not sure if we can maybe update it but this isn't a huge problem, thanks for checking

lumberyard-employee-dm (approved):

PR #13071, opened 2022-11-08
AZ Networking ISerialize Support for AZ::s64 and AZ::u64

### LINKED ISSUE #13019: AzNetworking: Add ISerializer Support for AZ::u64 and AZ::s64
The AZNetworking ISerializer supports serializing int64_t and uint64_t but not AZ::u64 and AZ::s64 because on some platforms this will be long and on others this will be long long, leading to a compiler error.

**Describe possible solutions you'd like**
There's a few approaches...
1. Stick to AZ types, basically enforce anyone using network serializer to stay in the AZ ecosystem and stick with AZ::u64 and AZ::s64. 

2. Add something to hide the standard int64_t
#define int64_t int64_typeconsistent
#define int64_typeconsistent signed long long
And then just use int64_t (and uint64_t) everywhere in code normally

3. Add a virtual method to ISerializer specifically for AZ::s64 and AZ::u64 #if AZ_TRAIT_COMPILER_INT64_T_IS_LONG

AMZN-Gene:
## What does this PR do?
Allows AzNetworking to serialize AZ::s64 and AZ::u64.
AzNetworking ISerialize current supports int64_t and uint64_t, which (depending on the system) is sometimes long and sometimes long long. AZ types such as AZ::u64 are set up to always be defined as long long. This change makes it so that if int64_t is long (not long long), then we'll define a new method to support long long.

Fixes #13019

### DIFF
--- a/Code/Framework/AzNetworking/AzNetworking/Serialization/ISerializer.h
+++ b/Code/Framework/AzNetworking/AzNetworking/Serialization/ISerializer.h
@@ -9,6 +9,7 @@
 #pragma once
 
 #include <stdint.h>
+#include <AzCore/base.h>
 #include <AzCore/std/limits.h>
 
 namespace AzNetworking
@@ -100,6 +101,12 @@ namespace AzNetworking
         //! @return boolean true for success, false for failure
         virtual bool Serialize(int64_t& value, const char* name, int64_t minValue = AZStd::numeric_limits<int64_t>::min(), int64_t maxValue = AZStd::numeric_limits<int64_t>::max()) = 0;
 
+        //! Serialize a signed 64-bit integer.
+        //! @param value    signed 64-bit integer input value to serialize
+        //! @param name     string name of the value being serialized
+        //! @param minValue the minimum value expected during serialization
+        //! @param maxValue the maximum value expected during serialization
+        //! @return boolean true for success, false for failure
         #if AZ_TRAIT_COMPILER_INT64_T_IS_LONG
         virtual bool Serialize(AZ::s64& value, const char* name, AZ::s64 minValue = AZStd::numeric_limits<AZ::s64>::min(), AZ::s64 maxValue = AZStd::numeric_limits<AZ::s64>::max()) = 0;
         #endif
--- a/Code/Framework/AzNetworking/AzNetworking/Serialization/ISerializer.h
+++ b/Code/Framework/AzNetworking/AzNetworking/Serialization/ISerializer.h
@@ -100,6 +100,10 @@ namespace AzNetworking
         //! @return boolean true for success, false for failure
         virtual bool Serialize(int64_t& value, const char* name, int64_t minValue = AZStd::numeric_limits<int64_t>::min(), int64_t maxValue = AZStd::numeric_limits<int64_t>::max()) = 0;
 
+        #if AZ_TRAIT_COMPILER_INT64_T_IS_LONG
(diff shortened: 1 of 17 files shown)


Code/Framework/AzNetworking/AzNetworking/Serialization/ISerializer.h
```diff
@@ -9,6 +9,7 @@
 #pragma once
 
 #include <stdint.h>
+#include <AzCore/base.h>
```
AMZN-puvvadar, line 12: NP: You could macro wrap this include as well if you wanted to

AMZN-puvvadar (approved):

Code/Framework/AzNetworking/AzNetworking/Serialization/ISerializer.h
AMZN-Gene: Nice catch! Thank you :)

lumberyard-employee-dm: Actually conditionally including this macro is not correct.

The `AZ_TRAIT_COMPILER_INT64_T_IS_LONG` comes from the `AzCore_Traits_Platform.h` that include that base.h brings in.

For example on Linux this macro is defined by the platform specific [AzCore_Traits_Linux.h](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/Platform/Linux/AzCore/AzCore_Traits_Linux.h#L59).

Without the include, the macro is defined.

lumberyard-employee-dm (changes requested):

Code/Framework/AzNetworking/AzNetworking/Serialization/ISerializer.h
AMZN-Gene: Fixed! Thanks :)

AMZN-puvvadar (approved):

lumberyard-employee-dm (approved):

Code/Framework/AzNetworking/AzNetworking/Serialization/ISerializer.h
```diff
@@ -100,6 +101,16 @@ namespace AzNetworking
...
         virtual bool Serialize(int64_t& value, const char* name, int64_t minValue = AZStd::numeric_limits<int64_t>::min(), int64_t maxValue = AZStd::numeric_limits<int64_t>::max()) = 0;
 
+        //! Serialize a signed 64-bit integer.
+        //! @param value    signed 64-bit integer input value to serialize
+        //! @param name     string name of the value being serialized
+        //! @param minValue the minimum value expected during serialization
+        //! @param maxValue the maximum value expected during serialization
+        //! @return boolean true for success, false for failure
+        #if AZ_TRAIT_COMPILER_INT64_T_IS_LONG
+        virtual bool Serialize(AZ::s64& value, const char* name, AZ::s64 minValue = AZStd::numeric_limits<AZ::s64>::min(), AZ::s64 maxValue = AZStd::numeric_limits<AZ::s64>::max()) = 0;
```
kberg-amzn, line 111: Couldn't this be non virtual, and in the inline simply cast to int64_t/uint64_t so that you only touch the header and no other interfaces are affected?

```
bool Serialize(AZ::s64& value, const char* name, AZ::s64 minValue = AZStd::numeric_limits<AZ::s64>::min(), AZ::s64 maxValue = AZStd::numeric_limits<AZ::s64>::max())
{
    return Serialize(static_cast<int64_t>(value), name);
}
```

AMZN-Gene:
Moved simpler implementation to https://github.com/o3de/o3de/pull/13116

PR #13073, opened 2022-11-08
[Terrain] Fixes to Terrain raycast and the underlying ClipRayWithAabb in AzCore

mbalfour-amzn:
## What does this PR do?

This fixes Terrain raycasting at multiple levels.

Math - ClipRayWithAabb: This method calculated rayEnd wrong, and calculated tClipStart / End wrong when one end was inside the AABB. The problems were discovered while debugging some broken terrain raycasts, which is the only code in the engine that currently uses ClipRayWithAabb.

Terrain - FindNearestIntersection: This method had some fundamentally broken math that was apparently relying on the broken results from ClipRayWithAabb. The unit tests that I wrote for this method failed 100% of the time once ClipRayWithAabb was fixed, and failed ~10% of the time even with the broken ClipRayWithAabb. The newly-rewritten method passes 100% of the time with the fixed ClipRayWithAabb.

## How was this PR tested?

Created unit tests for both ClipRayWithAabb and FindNearestIntersection that demonstrated the failures of the previous code, and demonstrates the success of the current code.

Also manually tested by using the paintbrush on terrain. I was previously seeing intermittent issues with the paintbrush failing to move (due to the collision failing), but now it moves consistently.


cgalvan (approved):

invertednormal (approved):
Looks good, just one suggestion for the future for how to make this faster. I worry a little about floating point precision as the squares are iterated through but it's probably ok.

Gems/Terrain/Code/Tests/TerrainSystemTest.cpp
```diff
@@ -1421,6 +1423,64 @@ namespace UnitTest
...
+                EXPECT_NEAR(result.m_worldPosition.GetZ(), 0.0f, 0.001f);
+                EXPECT_THAT(result.m_worldNormal, IsClose(AZ::Vector3::CreateAxisZ()));
+
+                // Track any intersection failures
+                numFailures += (result ? 0 : 1);
+            }
+        }
+
+        // This is here just to give us a final tally of how many rays failed this test.
+        ASSERT_EQ(numFailures, 0);
```
invertednormal, line 1481: Should this be `EXPECT_EQ` instead of `ASSERT_EQ`?

Gems/Terrain/Code/Source/TerrainRaycast/TerrainRaycastContext.cpp
```diff
@@ -77,158 +77,167 @@ namespace
...
+        |    |    |    |    |    |    |    | /X |
+        |____|____|____|____|____|____|____|/___|
+        |    |    |    |    |    |    | X  / X  |
+        |____|____|____|____|____|____|___/|____|
+        |    |    |    |    |    |    | T/ |    |
+        |____|____|____|____|____|____|____|____|
+        |    |    |    |    |    |    |    |    |
+        |____|____|____|____|____|____|____|____|
+    */
+    inline void FindNearestIntersectionIterative(
```
hultonha, line 115: nit: could remove `inline` here as we're in the .cpp (if this is only used in the .cpp then `static` would be a better option

Gems/Terrain/Code/Source/TerrainRaycast/TerrainRaycastContext.cpp
```diff
@@ -77,158 +77,167 @@ namespace
...
+        const TerrainSystem& terrainSystem,
+        const AZ::Vector2& terrainResolution,
+        const AZ::Aabb& terrainWorldBounds,
+        const AZ::Vector3& rayStart,
+        const AZ::Vector3& rayEnd,
+        AzFramework::RenderGeometry::RayResult& result)
```
hultonha, line 121: suggestion: Maybe outside the scope of this PR but having this function actually return the result instead of using an out param might be a nice change 🙂

hultonha: or as this is in an anonymous namespace you can just omit entirely

hultonha (approved):
LGTM and great comments + tests 👍

Gems/Terrain/Code/Source/TerrainRaycast/TerrainRaycastContext.cpp
mbalfour-amzn: Good point, there's really no reason this even exists as a separate method, I'll merge the code into RayIntersect.

Gems/Terrain/Code/Source/TerrainRaycast/TerrainRaycastContext.cpp
```diff
@@ -77,158 +77,167 @@ namespace
...
+        // We'll check each square to see if the ray intersections actually intersect the terrain triangles in the square.
+        for (int terrainSquare = 0; terrainSquare < numTerrainSquares; terrainSquare++)
         {
+            // Create a bounding volume for this terrain square.
+            AZ::Aabb currentVoxel = AZ::Aabb::CreateFromMinMax(
+                AZ::Vector3(curGridCorner, terrainWorldBounds.GetMin().GetZ()),
+                AZ::Vector3(curGridCorner + terrainResolution, terrainWorldBounds.GetMax().GetZ()));
+
+            // Check for a hit against the terrain triangles in this square.
             TriangulateAndFindNearestIntersection(terrainSystem, currentVoxel, hitTester, result);
```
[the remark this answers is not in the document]
mbalfour-amzn: 1 - I like the idea in concept, but I'm not sure it would be any (or much) cheaper in practice. Assuming the segment hits the terrain *somewhere*, then the min/max heights of the entire segment *should* surround most voxel boxes. So then we'd have to calculate just the min/max heights of the portion of the segment that intersects with the box, which then effectively trades "ray vs triangle" with "ray vs AABB". If you look at the code in IntersectRayAABB vs SegmentTriangleHitTester::Intersect, it's a bit hard to say which one would be faster.

2 - Really good point, this would definitely make the whole thing 50% faster. I'll save this one for a future improvement, but I'll at least add a comment in the code describing this as a future optimization so that the idea is captured.

lumberyard-employee-dm (approved):

Gems/Terrain/Code/Source/TerrainRaycast/TerrainRaycastContext.cpp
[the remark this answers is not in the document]
invertednormal: We chatted about this in person, but I wanted to give anyone else that comes across this PR some context. For the first suggestion, what I meant is to get the min/max height values for the ray inside a single terrain quad and compare them with the min/max height values for the 4 samples on the corners of that quad. In the vast majority of cases the ray will be above or below that quad completely, so a triangle test is unnecessary.

Issue #12897, opened 2022-10-30
Feature Request: Casting Shadows should be a flag on materials


ParasiteSt:
**Is your feature request related to a problem? Please describe.**
Right now the only way to make an object not cast a shadow is a hack by setting the objects opacity to 0.99, as shadows do not cast from transparent objects.

**Describe the solution you'd like**
Having a flag that could be toggled on or off to allow a material to either cast shadows or not cast shadows is the best approach.

**Describe alternatives you've considered**
The only alternative right now is setting the objects opacity to 0.99 to stop shadows from casting, but this is not a good way to do this.

**Additional context**
The small conversation from the discord server ;
![image](https://user-images.githubusercontent.com/100494538/198902704-d4c68778-cabc-400f-b0d4-14c6e7f28e27.png)
![image](https://user-images.githubusercontent.com/100494538/198902708-a1ab2faf-ed15-4be5-85a9-825421f58801.png)

santorac:
This would be really easy to implement on a per-material basis.

Issue #10678, opened 2022-07-14
Material Editor Shadow Catcher Always In Shadow


santorac:
**Describe the bug**
In the Material Editor the preview has a a shadow-catcher feature that is supposed to render a shadow on an otherwise invisible plane. Something is broken such that the entire plane is rendered as being in shadow.

**Steps to reproduce**
1. Open the Material Editor
2. Ensure the Shadow Catcher is enabled (it is usually on by default) ![image](https://user-images.githubusercontent.com/55155825/179101003-c72c68d1-d39c-42fa-9a81-d7d2bb642908.png)
3. Notice the background

**Expected behavior**
A shadow is cast under the preview object
![image](https://user-images.githubusercontent.com/55155825/179101097-32d971b7-bd07-4269-973d-1f614358ea24.png)

**Actual behavior**
The entire bottom half of the scene appears to be cast in shadow.
![image](https://user-images.githubusercontent.com/55155825/179101109-7b8cf779-320f-43d0-bd2f-adfd9fb7c082.png)

**Found in Branch**
development

**Additional context**
I think this issue was introduce by https://github.com/o3de/o3de/pull/8312

PR #13082, opened 2022-11-09
Added a flag to disable shadow casting on the core material types.

santorac:
## What does this PR do?

Added a flag to disable shadow casting on the core material types.
This addresses https://github.com/o3de/o3de/issues/12897.

I also added the capability for material properties to connect directly to the Enable state of a shader, similar to a direct shader constant or shader option connection. I ended up only being able to use this in one material type, but I think it will become more useful in the future.

Fixed ShadowCatcher shader (https://github.com/o3de/o3de/issues/10678). The opacity mode was set to Opaque which caused it to attempt using the fullscreen shadow buffer, but fullscreen shadows aren't supported for transparent materials. I also noticed some functions had an unused screen position parameter, which I removed.

## How was this PR tested?

AtomSampleViewer full test suite on vulkan saw no changes.
Tested each material type in Material Editor, flipping both the Receive Shadows and Cast Shadows flags. These affect self-shadowing and the shader catcher plane as expected.


antonmic (approved):

Gems/Atom/Feature/Common/Assets/Materials/Special/ShadowCatcher.azsl
```diff
@@ -15,7 +15,7 @@ option enum class OpacityMode
     Blended, 
     TintedTransparent
     
-} o_opacity_mode = OpacityMode::Opaque;
+} o_opacity_mode = OpacityMode::Blended;
```
rgba16f, line 18: is this a debugging change? Doesn't seem related to the shadow enabled change.

rgba16f (approved):

gadams3 (approved):

ParasiteSt:
Is there a required PR in front of this one before it can be merged?

Gems/Atom/Feature/Common/Assets/Materials/Special/ShadowCatcher.azsl
santorac: This is "Fixed ShadowCatcher shader" mentioned in the description.

santorac:
> Is there a required PR in front of this one before it can be merged?

No, it's ready to go, we just need to get passed Automated Review.

PR #13090, opened 2022-11-09
RHI/Vulkan: Fix error message in ImageView.cpp

CyaNinix:
Signed-off-by: JLeclercq <99494374+CyaNinix@users.noreply.github.com>

When this assert triggers (which is does incorrectly with cubemaps and default image view because of this line:
```cpp
            const uint16_t arrayLayers = AZStd::min(static_cast<uint16_t>(imgViewDesc.m_arraySliceMax - imgViewDesc.m_arraySliceMin), static_cast<uint16_t>(imgDesc.m_arraySize - 1)) + 1;
```), a crash follows because the format string is incorrect (single % have to be escaped as %%)


rgba16f (approved):

rgba16f:
AR run started: https://jenkins.build.o3de.org/job/O3DE/view/change-requests/job/PR-13090/1

PR #13104, opened 2022-11-10
Add support for EnumValue attribute to DPE

amzn-tmryan:
**Description**
Before this PR, EnumValue attributes were not being converted to DPE node attributes correctly in the LegacyReflectionBridge.

That was partly due to the way that attributes are being identified by a CRC of their string name; there was an EnumValue attribute in the SerializeContext and another in the EditContext.

This PR does a few things in order to handle EnumValue attributes in the DPE:
- Changes the CRC of the SerializeContext's EnumValue attribute key
- Templatizes EnumValueAttributeDefinition so it can work with the enum's original type
- Registers AttributeDefinitions for those enum types supported by the combo box property controls
- Caches related EnumValue attributes in the LegacyReflectionBridge and puts them into an EnumValues DOM node for the DPE to use
- Adds a char Visit function to the reader writer interfaces in Visitor.h as well as registering a char primitive handler in the LegacyReflectionBridge; this allows the Value attribute of a PropertyEditor node to be parsed correctly when the reflected enum values are of type char

**Testing**
- Verified that combo boxes are populated with their full list of reflected EnumValue attributes
- Verified the flow of control in the debugger and saw that the reflected attributes are being turned into DOM nodes in the DPE
- Verified that values specified as member defaults by any class reflecting an enum using EnumValue attributes is the value selected in DPE ComboBox handlers

### DIFF
--- a/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/DocumentSchema.cpp
+++ b/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/DocumentSchema.cpp
@@ -104,48 +104,6 @@ namespace AZ::DocumentPropertyEditor
         return Dom::Value(result.GetStringView(), false);
     }
 
-    Dom::Value EnumValueAttributeDefinition::ValueToDom(const EnumConstantValue& attribute) const
-    {
-        Dom::Value result(Dom::Type::Object);
-        result[EntryNameKey] = Dom::Value(attribute.m_name, true);
-        result[EntryValueKey] = Dom::Value(static_cast<uint64_t>(attribute.m_value));
-        return result;
-    }
-
-    AZStd::optional<EnumConstantValue> EnumValueAttributeDefinition::DomToValue(const Dom::Value& value) const
-    {
-        if (!value.IsObject() || !value.HasMember(EntryNameKey) || !value.HasMember(EntryValueKey))
-        {
-            return {};
-        }
-
-        return EnumConstantValue{ value[EntryNameKey].GetString(), static_cast<AZ::u64>(value[EntryValueKey].GetUint64()) };
-    }
-
-    AZ::Dom::Value EnumValueAttributeDefinition::LegacyAttributeToDomValue(void* instance, AZ::Attribute* attribute) const
-    {
-        if (attribute == nullptr)
-        {
-            return AZ::Dom::Value();
-        }
-
-        using EnumConstantBasePtr = AZStd::unique_ptr<SerializeContextEnumInternal::EnumConstantBase>;
-        if (auto internalEnumConsantBase = azdynamic_cast<AttributeData<EnumConstantBasePtr>*>(attribute); internalEnumConsantBase)
-        {
-            EnumConstantValue* value = reinterpret_cast<EnumConstantValue*>(internalEnumConsantBase->Get(instance).get());
-            return ValueToDom(*value);
-        }
-        else if (auto azEditEnumConstant = azdynamic_cast<AttributeData<AZ::Edit::EnumConstant<AZ::u64>>*>(attribute); azEditEnumConstant)
-        {
-            AZ::Edit::EnumConstant<AZ::u64> value = static_cast<AZ::Edit::EnumConstant<AZ::u64>>(azEditEnumConstant->Get(instance));
-            Dom::Value domValue(Dom::Type::Object);
-            domValue[EntryDescriptionKey] = Dom::Value(value.m_description, true);
-            domValue[EntryValueKey] = Dom::Value(static_cast<uint64_t>(value.m_value));
-            return domValue;
-        }
-        return Dom::Value();
-    }
-
     Dom::Value EnumValuesAttributeDefinition::ValueToDom(const EnumValuesContainer& attribute) const
     {
         Dom::Value result(Dom::Type::Array);
--- a/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/DocumentSchema.cpp
+++ b/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/DocumentSchema.cpp
@@ -107,19 +107,43 @@ namespace AZ::DocumentPropertyEditor
     Dom::Value EnumValueAttributeDefinition::ValueToDom(const EnumConstantValue& attribute) const
     {
         Dom::Value result(Dom::Type::Object);
-        result[EntryDescriptionKey] = Dom::Value(attribute.m_description, true);
+        result[EntryNameKey] = Dom::Value(attribute.m_name, true);
         result[EntryValueKey] = Dom::Value(static_cast<uint64_t>(attribute.m_value));
         return result;
     }
 
     AZStd::optional<EnumConstantValue> EnumValueAttributeDefinition::DomToValue(const Dom::Value& value) const
     {
-        if (!value.IsObject() || !value.HasMember(EntryDescriptionKey) || !value.HasMember(EntryValueKey))
+        if (!value.IsObject() || !value.HasMember(EntryNameKey) || !value.HasMember(EntryValueKey))
         {
             return {};
         }
 
-        return EnumConstantValue{ static_cast<AZ::u64>(value[EntryValueKey].GetUint64()), value[EntryDescriptionKey].GetString() };
+        return EnumConstantValue{ value[EntryNameKey].GetString(), static_cast<AZ::u64>(value[EntryValueKey].GetUint64()) };
+    }
+
+    AZ::Dom::Value EnumValueAttributeDefinition::LegacyAttributeToDomValue(void* instance, AZ::Attribute* attribute) const
+    {
+        if (attribute == nullptr)
+        {
+            return AZ::Dom::Value();
+        }
+
+        using EnumConstantBasePtr = AZStd::unique_ptr<SerializeContextEnumInternal::EnumConstantBase>;
+        if (auto internalEnumConsantBase = azdynamic_cast<AttributeData<EnumConstantBasePtr>*>(attribute); internalEnumConsantBase)
+        {
+            EnumConstantValue* value = reinterpret_cast<EnumConstantValue*>(internalEnumConsantBase->Get(instance).get());
+            return ValueToDom(*value);
+        }
+        else if (auto azEditEnumConstant = azdynamic_cast<AttributeData<AZ::Edit::EnumConstant<AZ::u64>>*>(attribute); azEditEnumConstant)
+        {
+            AZ::Edit::EnumConstant<AZ::u64> value = static_cast<AZ::Edit::EnumConstant<AZ::u64>>(azEditEnumConstant->Get(instance));
+            Dom::Value domValue(Dom::Type::Object);
+            domValue[EntryDescriptionKey] = Dom::Value(value.m_description, true);
+            domValue[EntryValueKey] = Dom::Value(static_cast<uint64_t>(value.m_value));
+            return domValue;
+        }
+        return Dom::Value();
     }
 
     Dom::Value EnumValuesAttributeDefinition::ValueToDom(const EnumValuesContainer& attribute) const
--- a/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/DocumentSchema.cpp
+++ b/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/DocumentSchema.cpp
@@ -104,6 +104,24 @@ namespace AZ::DocumentPropertyEditor
         return Dom::Value(result.GetStringView(), false);
     }
 
+    Dom::Value EnumValueAttributeDefinition::ValueToDom(const EnumConstantValue& attribute) const
+    {
+        Dom::Value result(Dom::Type::Object);
+        result[EntryDescriptionKey] = Dom::Value(attribute.m_description, true);
+        result[EntryValueKey] = Dom::Value(static_cast<uint64_t>(attribute.m_value));
+        return result;
+    }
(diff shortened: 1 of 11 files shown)


Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/PropertyEditorNodes.h
```diff
@@ -138,7 +138,7 @@ namespace AZ::DocumentPropertyEditor::Nodes
 
         static constexpr auto EnumType = TypeIdAttributeDefinition("EnumType");
         static constexpr auto EnumUnderlyingType = TypeIdAttributeDefinition("EnumUnderlyingType");
-        static constexpr auto EnumValue = AttributeDefinition<Dom::Value>("EnumValue");
+        static constexpr auto EnumValue = EnumValueAttributeDefinition("EnumValue");
```
NicholasVanSickle, line 141: Sorry, refresh my memory on what `EnumValue` is used for? On a superficial search it looks like it only takes one enum value, so I just want to make sure I understand what the RPE-side handlers are trying to do with it before I dig in further on the DPE side.

amzn-tmryan: `EnumValue` is an internal attribute that the serialize context uses. I don't see it being used to reflect a given value anywhere, with a global search. I do see its case in `PropertyEnumBoxComboCtrl::ConsumeAttribute` (ln 107-ish) being hit often enough when loading DPEs for some assets/components. Some of those have broken combo boxes, hence the trying to fix `EnumValue` parsing.

I guess it's possible we don't actually need to handle this attribute at all in the reflection bridge since it's supposed to be internal to the serialize context? The legacy reflection bridge does currently attempt to parse them; it just fails to retrieve something that the `AttributeReader` can read.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/DocumentSchema.cpp
```diff
@@ -104,6 +104,40 @@ namespace AZ::DocumentPropertyEditor
...
+        }
+
+        using EnumConstantBasePtr = AZStd::unique_ptr<SerializeContextEnumInternal::EnumConstantBase>;
+        if (auto attributeData = azdynamic_cast<AttributeData<EnumConstantBasePtr>*>(attribute); attributeData != nullptr)
+        {
+            EnumConstantValue* value = reinterpret_cast<EnumConstantValue*>(attributeData->Get(instance).get());
+            return ValueToDom(*value);
+        }
+        return Dom::Value();
+    }
```
amzn-tmryan, line 139: To preface the words below, this code properly parses the `EnumValue` attributes, but has to use a `reinterpret_cast` which isn't awesome. I haven't verified the DPE loads properly yet because something is preventing the combo boxes from showing up now (might be serialize vs edit related but I'm unsure) but I'm definitely able to see the values being pulled out in `LegacyAttributeToDomValue` using the debugger.

Overrode this function for `EnumValue`s in order to single them out. Made it easier to debug at first but, once Alex and I realized what was going on, it also made it clear to me that skipping the `AttributeReader` might help show the problem better.

So, the `EnumValue` attribute is stored as a `unique_ptr<SerializeContextEnumInternal::EnumConstantBase>` in its `AttributeData`. I'm assuming this is fine for whatever is done with it internal to the reflection system, but it's causing problems for us trying to pull the data out and into a `Dom::Value`. Problems I was running into:
- Can't copy the `unique_ptr` (as expected for unique_ptr)
- `AttributeData::Get` returns a `const` qualified value; in this case it yields a `const unique_ptr<SerializeContextEnumInternal::EnumConstantBase>` (this might have been causing me problems when trying to `AZStd::move` the `unique_ptr` into a `shared_ptr` so I could try `AZStd::rtti_pointer_cast`
- The dereferenced type is an abstract base class (`EnumConstantBase`) so we can't do much with it other than cast or pass it
- It's a `unique_ptr`, and detecting that an attribute's data is a unique_ptr inside of the `AttributeReader` seems like overkill for this case (I'm assuming not many attributes would be stored this way)

Those problems taken together are making it very hard (seemingly impossible) to use the `AttributeReader` or do anything other than `reinterpret_cast` the base type ptr into something we can use.

I'm sure there is another way to handle this. In fact, I'm still a little curious what all would be involved in turning the context's `EnumValue` attributes into `EnumValues` in the `LegacyReflectionBridge` similar to how we turn all read-only related attributes into `Disabled` in the DPE.

amzn-tmryan:
Added a commit that at least successfully pulls the data out of its AttributeData. It's a hack though, so mostly for informational purposes at this point.

Left a comment with more details.

amzn-tmryan:
Updated the last commit to support EnumValue attributes in some components as well. They were using the `AZ::Edit::EnumConstant<AZ::u64>` type that we were looking for originally. So there are at least two ways that `EnumValue` can be stored.

I was also able to verify that although this does create a Dom::Value, it's still not populating the combo box. Investigating that.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/PropertyEditorNodes.h
amzn-tmryan: More comments below.

amzn-tmryan:
There is still some work left to do since only a single enum value is showing in each ComboBox and the default value for those controls is currently null. I have some ideas on how to resolve those issues as commented in the PR description, but let me know what you think.

lumberyard-employee-dm (approved):
This change looks good to me now.
The Enum type is preserved with the EnumValueAttributeDefintion and the base AttributeDefinitionBase class can use ValueToDom and DomToValue to convert enum values to and from DOM values.

lumberyard-employee-dm:
> There is still some work left to do since only a single enum value is showing in each ComboBox and the default value for those controls is currently null. I have some ideas on how to resolve those issues as commented in the PR description, but let me know what you think. Most of this latest stuff was done with @lumberyard-employee-dm's guidance/advice so adding them to the PR.

One idea is to to a vector of all attributes that has the key "EnumValue" for the AZ::Edit::ElementData instance being iteted and then have the PropertyHandlerBase* instance invoke [ConsumeAttribute](https://github.com/o3de/o3de/blob/development/Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyEnumComboBoxCtrl.hxx#L107-L116)  on those attributes, which would add the enum string as a drop down option for each call.

amzn-tmryan:
@NicholasVanSickle I think I needed to add a primitive handler for type char in order to fully support enums that are reflected as EnumValue attributes. Please let me know if there was a reason Visit(char, ...) was left out of Visitor::IReadWrite etc.

alexmontAmazon (approved):
looks good!

dtamkin1 (approved):

NicholasVanSickle (approved):
Looks reasonable to me, though I suspect you don't actually need a `char` visitor because it should alias to `AZ::u8` on all platforms I'm aware of.

amzn-tmryan:
> Looks reasonable to me, though I suspect you don't actually need a `char` visitor because it should alias to `AZ::u8` on all platforms I'm aware of.

Thanks for taking a look!

The behavior I saw before adding the char visitor was that the PropertyEditor::Value attribute had a Dom::Value with a pointer to the char instead of the char itself. If there is another way to do this then I'd be happy to fix it up in one of the upcoming related PRs.

lumberyard-employee-dm:
> Looks reasonable to me, though I suspect you don't actually need a `char` visitor because it should alias to `AZ::u8` on all platforms I'm aware of.

`AZ::u8` alias to `unsigned char`, which is different type than `char`

amzn-tmryan:
Linux platforms are having trouble determining the right Dom::Value constructor to use on line 268 in `Dom::Value ValueToDom(const EnumConstantValue& attribute)`. I tried to sort it out using AZStd::to_underlying, but we don't have access to the actual enum by this point; it's wrapped in an AZ::Edit::EnumConstant. Since AZ::Edit::EnumConstant is storing its value as an AZ::u64, we can get by without casting here. We still cast back to the original enum value type when extracting data from the Dom node.

Hopefully this will solve the issue on Linux platforms (not 100% confident), but let me know if there is a better way to do it.

lumberyard-employee-dm:
> Linux platforms are having trouble determining the right Dom::Value constructor to use on line 268 in `Dom::Value ValueToDom(const EnumConstantValue& attribute)`. I tried to sort it out using AZStd::to_underlying, but we don't have access to the actual enum by this point; it's wrapped in an AZ::Edit::EnumConstant. Since AZ::Edit::EnumConstant is storing its value as an AZ::u64, we can get by without casting here. We still cast back to the original enum value type when extracting data from the Dom node.
> 
> Hopefully this will solve the issue on Linux platforms (not 100% confident), but let me know if there is a better way to do it.

What do you mean, we don't have access to the actual enum? We have the actual C++ template type.
We can use AZStd::underlying_type_t to get the underlying integral type and use that as part of the static_cast.

See [base.h](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/base.h#L340)

amzn-tmryan:
> > Linux platforms are having trouble determining the right Dom::Value constructor to use on line 268 in `Dom::Value ValueToDom(const EnumConstantValue& attribute)`. I tried to sort it out using AZStd::to_underlying, but we don't have access to the actual enum by this point; it's wrapped in an AZ::Edit::EnumConstant. Since AZ::Edit::EnumConstant is storing its value as an AZ::u64, we can get by without casting here. We still cast back to the original enum value type when extracting data from the Dom node.
> > Hopefully this will solve the issue on Linux platforms (not 100% confident), but let me know if there is a better way to do it.
> 
> What do you mean, we don't have access to the actual enum? We have the actual C++ template type. We can use AZStd::underlying_type_t to get the underlying integral type and use that as part of the static_cast.
> 
> See [base.h](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/base.h#L340)

Attempting to do something like the following is giving me `Failed to specialize alias template` errors:
```
using UnderlyingType = AZStd::underlying_type_t<EnumType>;
...
result[EntryValueKey] = Dom::Value(static_cast<UnderlyingType>(attribute.m_value));
...
```

lumberyard-employee-dm:
> > > Linux platforms are having trouble determining the right Dom::Value constructor to use on line 268 in `Dom::Value ValueToDom(const EnumConstantValue& attribute)`. I tried to sort it out using AZStd::to_underlying, but we don't have access to the actual enum by this point; it's wrapped in an AZ::Edit::EnumConstant. Since AZ::Edit::EnumConstant is storing its value as an AZ::u64, we can get by without casting here. We still cast back to the original enum value type when extracting data from the Dom node.
> > > Hopefully this will solve the issue on Linux platforms (not 100% confident), but let me know if there is a better way to do it.
> > 
> > 
> > What do you mean, we don't have access to the actual enum? We have the actual C++ template type. We can use AZStd::underlying_type_t to get the underlying integral type and use that as part of the static_cast.
> > See [base.h](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/base.h#L340)
> 
> Attempting to do something like the following is giving me `Failed to specialize alias template` errors:
> 
> ```
> using UnderlyingType = AZStd::underlying_type_t<EnumType>;
> ...
> result[EntryValueKey] = Dom::Value(static_cast<UnderlyingType>(attribute.m_value));
> ...
> ```

You need to cast the attribute value to an `UnderlyingType*`
```
result[EntryValueKey] = Dom::Value(*static_cast<UnderlyingType*>(attribute.m_value));
```

amzn-tmryan:
>

amzn-tmryan:
Is this true on our Linux machines that run AR?
``` AZ_TRAIT_COMPILER_INT64_T_IS_LONG ```

Wondering if the problem could be related to this:
```
#   if AZ_TRAIT_COMPILER_INT64_T_IS_LONG // int64_t is long
    typedef signed long long        s64;
    typedef unsigned long long      u64;
#   else
    typedef int64_t   s64;
    typedef uint64_t  u64;
#   endif //
```

`uint64_t` isn't being defined as `u64` when `AZ_TRAIT_COMPILER_INT64_T_IS_LONG` is true. Looks like `unsigned long long` is u64 in that case, and there is no Dom::Value ctor for `unsigned long long`. There is a `explicit Value(uint64_t value)` though.

Not sure why this sort of error wouldn't have popped up before though. I'm still open to the other explanation, but the error makes it sound like the ambiguity results from trying to determine which ctor to use. Looks like the types are all covered and they're explicit, so maybe we aren't providing a ctor for a type that we'd encounter on systems with a Linux kernel?

File: Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/DocumentSchema.h [hunk 3]
```diff
@@ -265,7 +265,8 @@ namespace AZ::DocumentPropertyEditor
         {
             Dom::Value result(Dom::Type::Object);
             result[EntryDescriptionKey] = Dom::Value(attribute.m_description, true);
-            result[EntryValueKey] = Dom::Value(attribute.m_value);
+            result[EntryValueKey] = Dom::Value();
+            result[EntryValueKey].SetUint64(attribute.m_value);
```
amzn-tmryan, line 269: I made this change so the Dom::Value ctor is no longer called. That "fixed" the ambiguity issue in Linux builds. We still need to figure out what caused it and find a resolution.

Merging as-is for now with a note to revisit this with lumberyard-employee-dm after the break.

PR #13105, opened 2022-11-10
Use cached values to determine if entities are selectable from the viewport

AMZN-daimini:

## What does this PR do?

Changes the Viewport code to use the cached values to determine whether an entity is selectable over re-computing them for every query.

## How was this PR tested?

Manual testing


Code/Framework/AzToolsFramework/AzToolsFramework/ViewportSelection/EditorHelpers.cpp
```diff
@@ -382,11 +382,28 @@ namespace AzToolsFramework
...
+        auto entityCacheId = m_entityDataCache->GetVisibleEntityIndexFromId(entityId);
+        if (entityCacheId.has_value())
+        {
+            return IsSelectableInViewport(entityCacheId.value());
+        }
+
+        return false;
+    }
+
+    bool EditorHelpers::IsSelectableInViewport(size_t entityCacheId) const
```
hultonha, line 394: nit: Could you please update the parameter name to be `entityCacheIndex`

Code/Framework/AzToolsFramework/AzToolsFramework/ViewportSelection/EditorHelpers.cpp
```diff
@@ -382,11 +382,28 @@ namespace AzToolsFramework
...
+    {
+        return m_entityDataCache->IsVisibleEntityIndividuallySelectableInViewport(entityCacheId);
     }
 
     bool EditorHelpers::IsSelectableAccordingToFocusMode(const AZ::EntityId entityId) const
     {
+        auto entityCacheId = m_entityDataCache->GetVisibleEntityIndexFromId(entityId);
```
hultonha, line 401: nit: I think it's better to refer to this as `entityCacheIndex` as it's slightly closer to what it really is and can't be confused with the entityId

Code/Framework/AzToolsFramework/AzToolsFramework/ViewportSelection/EditorHelpers.h
```diff
@@ -92,6 +92,10 @@ namespace AzToolsFramework
         //! to the current Editor Focus Mode and Container Entity setup.
         bool IsSelectableInViewport(AZ::EntityId entityId) const;
 
+        //! Returns whether the entityCacheId can be selected in the viewport according
+        //! to the current Editor Focus Mode and Container Entity setup.
+        bool IsSelectableInViewport(size_t entityCacheId) const;
```
hultonha, line 97: update: I wouldn't make this a public function and instead just make a wrapper in the .cpp, and then call it from line 333 with `entityCacheIndex` directly). If it does make it easier to have this be a member function then private would be better.

hultonha (commented):
Great change, I just have a few minor suggestions, thanks very much for taking a look at this! 😄

Code/Framework/AzToolsFramework/AzToolsFramework/ViewportSelection/EditorHelpers.cpp
```diff
@@ -382,11 +382,28 @@ namespace AzToolsFramework
 
     bool EditorHelpers::IsSelectableInViewport(const AZ::EntityId entityId) const
     {
-        return IsSelectableAccordingToFocusMode(entityId) && IsSelectableAccordingToContainerEntities(entityId);
+        auto entityCacheId = m_entityDataCache->GetVisibleEntityIndexFromId(entityId);
+        if (entityCacheId.has_value())
```
hultonha, line 386: nit: Also I can't help mentioning it but these could all be if initializers 🙂 

```
if (auto entityCacheId = m_entityDataCache->GetVisibleEntityIndexFromId(entityId);
    entityCacheId.has_value())
```

AMZN-daimini: Done!

Code/Framework/AzToolsFramework/AzToolsFramework/ViewportSelection/EditorHelpers.cpp
AMZN-daimini: Fixed everywhere!

Code/Framework/AzToolsFramework/AzToolsFramework/ViewportSelection/EditorHelpers.h
AMZN-daimini: I just removed the function completely since it was only called in the entityId version either way,

cgalvan (approved):

hultonha (commented):
Sorry I think we had some cross wires, can you please keep the function with the EntityIndex, and then change this.... (line 333)

```
if (const AZ::EntityId entityId = m_entityDataCache->GetVisibleEntityId(entityCacheIndex);
    m_entityDataCache->IsVisibleEntityVisible(entityCacheIndex) && IsSelectableInViewport(entityId))
```

to...

```
if (m_entityDataCache->IsVisibleEntityVisible(entityCacheIndex) && IsSelectableInViewport(entityCacheIndex))
```

So we don't need to do the extra conversion from entityId to index

Code/Framework/AzToolsFramework/AzToolsFramework/ViewportSelection/EditorHelpers.cpp
hultonha: Thank you! 😄

Code/Framework/AzToolsFramework/AzToolsFramework/ViewportSelection/EditorHelpers.h
hultonha: See my later comment which is why I think it is worth keeping (albeit private) as it'll be faster to call in the loop as we already have the `entityCacheIndex`

AMZN-daimini:
> Sorry I think we had some cross wires, can you please keep the function with the EntityIndex, and then change this.... (line 333)
> 
> ```
> if (const AZ::EntityId entityId = m_entityDataCache->GetVisibleEntityId(entityCacheIndex);
>     m_entityDataCache->IsVisibleEntityVisible(entityCacheIndex) && IsSelectableInViewport(entityId))
> ```
> 
> to...
> 
> ```
> if (m_entityDataCache->IsVisibleEntityVisible(entityCacheIndex) && IsSelectableInViewport(entityCacheIndex))
> ```
> 
> So we don't need to do the extra conversion from entityId to index

The contents of that if block will still need the entityId. We can still avoid the back and forth though.

hultonha (approved):
Looks great! Thanks @AMZN-daimini 👍

PR #13106, opened 2022-11-10
Added supported platform text rendering for gem catalog cards.

tkothadev:
Signed-off-by: T.J. Kotha <112996779+tkothadev@users.noreply.github.com>

## What does this PR do?

This PR provides text rendering for gems to indicate their supported platforms. Please see the testing section for examples.

_Note on implementation:_ The original plan was to implement the text rendering using a QLabel and adjusting layout (i.e. implementing from scratch). However, rendering the text via QPainter was chosen, because the `GemItemDelegate` class uses QPainter to render everything. To go with the original plan would have meant overhauling the class, which is too drastic for this PR.

## How was this PR tested?

The rendering logic was visually tested for various scenarios. Note that at the time of writing this PR,  there is currently no logic for fetching the supported platforms from `GemModel`. That will be addressed in a future PR. These scenarios were tested via manually inputting values for the `platforms` field @ `GemItemDelegate.cpp:L380`. Rendering logic was verified via this approach.

When platforms are unspecified:
image(1)

When at least four are specified:
image (14)

When there is wrap around:
image (15)

When there is extreme wrap around:
image (16)
GemCatalogSupportedPlatformWordWrap

Showcasing both cases where unspecified and specified are present:
image (17)

### DIFF
--- a/Code/Tools/ProjectManager/Source/GemCatalog/GemItemDelegate.cpp
+++ b/Code/Tools/ProjectManager/Source/GemCatalog/GemItemDelegate.cpp
@@ -375,18 +375,23 @@ namespace O3DE::ProjectManager
     }
 
 
-    void GemItemDelegate::DrawPlatformText(QPainter* painter, const QRect& contentRect, const QFont& standardFont, [[maybe_unused]] const QModelIndex& modelIndex) const
+    void GemItemDelegate::DrawPlatformText(QPainter* painter, const QRect& contentRect, const QFont& standardFont, const QModelIndex& modelIndex) const
     {
         const GemInfo::Platforms platforms = GemModel::GetPlatforms(modelIndex);
         
         auto xbounds = CalcColumnXBounds(HeaderOrder::Name);
-        int startX = s_platformTextleftMarginCorrection + xbounds.first;
+        const int startX = s_platformTextleftMarginCorrection + xbounds.first;
         QFont platformFont(standardFont);
         platformFont.setPixelSize(s_featureTagFontSize);
         platformFont.setBold(false);
         painter->setFont(platformFont);
         QString platformText = "";
 
+        //If no platforms are specified, there is nothing to draw
+        if(platforms == 0)
+        {
+            return;
+        }
         
         //UX prefers that we show platforms in reverse alphabetical order
         for(int i = GemInfo::NumPlatforms-1; i >= 0; i--)
@@ -394,23 +399,21 @@ namespace O3DE::ProjectManager
             const GemInfo::Platform platform = static_cast<GemInfo::Platform>(1 << i);
             if (platforms & platform)
             {
-                QString singlePlatformText = GemInfo::GetPlatformString(platform);
+                const QString& singlePlatformText = GemInfo::GetPlatformString(platform);
                 if(i != GemInfo::NumPlatforms-1)
                 {
                     platformText.append(", ");
                 }
                 platformText.append(singlePlatformText);
-                
             }
         }
-        
 
         //figure out the ideal rect size for the platform text space constraints
-        QRect platformRect = QRect(contentRect.left() + startX, contentRect.bottom() - s_platformTextHeightAdjustment,
+        const QRect platformRect = QRect(contentRect.left() + startX, contentRect.bottom() - s_platformTextHeightAdjustment,
                                    xbounds.second -xbounds.first - s_platformTextWrapAroundMargin,
                                    (s_featureTagFontSize + s_platformTextLineBottomMargin) * s_platformTextWrapAroundLineMaxCount);
 
-        DrawText(platformText, painter, platformRect, platformFont, !platforms);
+        DrawText(platformText, painter, platformRect, platformFont);     
     }
 
     void GemItemDelegate::DrawFeatureTags(
@@ -470,7 +473,7 @@ namespace O3DE::ProjectManager
         return doc;
     }
 
-    void GemItemDelegate::DrawText(const QString& text, QPainter* painter, const QRect& rect, const QFont& standardFont, const bool& greyedOut) const 
+    void GemItemDelegate::DrawText(const QString& text, QPainter* painter, const QRect& rect, const QFont& standardFont) const 
     {
         painter->save();
 
@@ -489,14 +492,7 @@ namespace O3DE::ProjectManager
         else
         {
             painter->setFont(standardFont);
-            if(greyedOut)
-            {
-                painter->setPen(m_greyedOutTextColor);
-            }
-            else
-            {
-                painter->setPen(m_textColor);
-            }
+            painter->setPen(m_textColor);
             painter->drawText(rect, Qt::AlignLeft | Qt::TextWordWrap, text);
         }
 
--- a/Code/Tools/ProjectManager/Source/GemCatalog/GemItemDelegate.cpp
+++ b/Code/Tools/ProjectManager/Source/GemCatalog/GemItemDelegate.cpp
@@ -168,7 +168,7 @@ namespace O3DE::ProjectManager
         {
             DrawButton(painter, buttonRect, modelIndex);
         }
-        DrawPlatformIcons(painter, contentRect, modelIndex);
+        DrawPlatformText(painter, contentRect, standardFont, modelIndex);
         DrawFeatureTags(painter, contentRect, featureTags, standardFont, summaryRect);
 
         painter->restore();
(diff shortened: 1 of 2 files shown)


tkothadev:
Note: Due to the caveat mentioned for testing, currently if the PR should be accepted, all gems will default to stating platform support is `Unspecified`. This behavior will be updated to fetch actual supported platforms from the `GemModel` in my next task.

tkothadev:
Updated the Unspecified keyword to now be blank whenever a gem lacks data for supported platforms.

It will look like this:
image (18)

Code/Tools/ProjectManager/Source/GemCatalog/GemItemDelegate.cpp
```diff
@@ -374,6 +374,45 @@ namespace O3DE::ProjectManager
         }
     }
 
+
+    void GemItemDelegate::DrawPlatformText(QPainter* painter, const QRect& contentRect, const QFont& standardFont, [[maybe_unused]] const QModelIndex& modelIndex) const
```
AMZN-alexpete, line 378: you are using this param so you don't need `maybe_unused`
```suggestion
    void GemItemDelegate::DrawPlatformText(QPainter* painter, const QRect& contentRect, const QFont& standardFont, const QModelIndex& modelIndex) const
```

Code/Tools/ProjectManager/Source/GemCatalog/GemItemDelegate.cpp
```diff
@@ -374,6 +374,45 @@ namespace O3DE::ProjectManager
...
+    {
+        const GemInfo::Platforms platforms = GemModel::GetPlatforms(modelIndex);
+        
+        auto xbounds = CalcColumnXBounds(HeaderOrder::Name);
+        int startX = s_platformTextleftMarginCorrection + xbounds.first;
```
AMZN-alexpete, line 383:
```suggestion
        const int startX = s_platformTextleftMarginCorrection + xbounds.first;
```

Code/Tools/ProjectManager/Source/GemCatalog/GemItemDelegate.cpp
```diff
@@ -374,6 +374,45 @@ namespace O3DE::ProjectManager
         }
     }
 
+
+    void GemItemDelegate::DrawPlatformText(QPainter* painter, const QRect& contentRect, const QFont& standardFont, [[maybe_unused]] const QModelIndex& modelIndex) const
+    {
+        const GemInfo::Platforms platforms = GemModel::GetPlatforms(modelIndex);
```
AMZN-alexpete, line 380: can we just early out here to avoid the rest of the logic?
```suggestion
        const GemInfo::Platforms platforms = GemModel::GetPlatforms(modelIndex);
        if(!platforms)
        {
            return;
        }
```

Code/Tools/ProjectManager/Source/GemCatalog/GemItemDelegate.cpp
```diff
@@ -374,6 +374,45 @@ namespace O3DE::ProjectManager
...
+        QString platformText = "";
+
+        
+        //UX prefers that we show platforms in reverse alphabetical order
+        for(int i = GemInfo::NumPlatforms-1; i >= 0; i--)
+        {
+            const GemInfo::Platform platform = static_cast<GemInfo::Platform>(1 << i);
+            if (platforms & platform)
+            {
+                QString singlePlatformText = GemInfo::GetPlatformString(platform);
```
AMZN-alexpete, line 397: compiler is probably smart enough to avoid the copy here, but you can probably get away with something like this
```suggestion
                const QString& singlePlatformText = GemInfo::GetPlatformString(platform);
```

Code/Tools/ProjectManager/Source/GemCatalog/GemItemDelegate.cpp
```diff
@@ -374,6 +374,45 @@ namespace O3DE::ProjectManager
...
+                if(i != GemInfo::NumPlatforms-1)
+                {
+                    platformText.append(", ");
+                }
+                platformText.append(singlePlatformText);
+
```
AMZN-alexpete, line 403: nit: blank line
```suggestion
```

Code/Tools/ProjectManager/Source/GemCatalog/GemItemDelegate.cpp
```diff
@@ -374,6 +374,45 @@ namespace O3DE::ProjectManager
...
+                QString singlePlatformText = GemInfo::GetPlatformString(platform);
+                if(i != GemInfo::NumPlatforms-1)
+                {
+                    platformText.append(", ");
+                }
+                platformText.append(singlePlatformText);
+                
+            }
+        }
+
```
AMZN-alexpete, line 406: nit: blank line
```suggestion
```

Code/Tools/ProjectManager/Source/GemCatalog/GemItemDelegate.cpp
```diff
@@ -374,6 +374,45 @@ namespace O3DE::ProjectManager
...
+        const GemInfo::Platforms platforms = GemModel::GetPlatforms(modelIndex);
+        
+        auto xbounds = CalcColumnXBounds(HeaderOrder::Name);
+        int startX = s_platformTextleftMarginCorrection + xbounds.first;
+        QFont platformFont(standardFont);
+        platformFont.setPixelSize(s_featureTagFontSize);
+        platformFont.setBold(false);
+        painter->setFont(platformFont);
+        QString platformText = "";
+
```
AMZN-alexpete, line 389: nit: extra blank line
```suggestion
```

Code/Tools/ProjectManager/Source/GemCatalog/GemItemDelegate.cpp
```diff
@@ -374,6 +374,45 @@ namespace O3DE::ProjectManager
...
+                    platformText.append(", ");
+                }
+                platformText.append(singlePlatformText);
+                
+            }
+        }
+        
+
+        //figure out the ideal rect size for the platform text space constraints
+        QRect platformRect = QRect(contentRect.left() + startX, contentRect.bottom() - s_platformTextHeightAdjustment,
```
AMZN-alexpete, line 409:
```suggestion
        const QRect platformRect = QRect(contentRect.left() + startX, contentRect.bottom() - s_platformTextHeightAdjustment,
```

Code/Tools/ProjectManager/Source/GemCatalog/GemItemDelegate.cpp
```diff
@@ -431,7 +470,7 @@ namespace O3DE::ProjectManager
         return doc;
     }
 
-    void GemItemDelegate::DrawText(const QString& text, QPainter* painter, const QRect& rect, const QFont& standardFont) const 
+    void GemItemDelegate::DrawText(const QString& text, QPainter* painter, const QRect& rect, const QFont& standardFont, const bool& greyedOut) const
```
AMZN-alexpete, line 473: I'm starting to wonder why we don't just add a `color` parameter to this function and pass in the color at the call sites instead...

Code/Tools/ProjectManager/Source/GemCatalog/GemItemDelegate.cpp
```diff
@@ -450,7 +489,14 @@ namespace O3DE::ProjectManager
...
             painter->setFont(standardFont);
-            painter->setPen(m_textColor);
+            if(greyedOut)
+            {
+                painter->setPen(m_greyedOutTextColor);
+            }
+            else
+            {
+                painter->setPen(m_textColor);
+            }
```
AMZN-alexpete, line 499: This is a bit tricky, once we go down the road of multiple colors we probably don't want to have that logic in this helper function.  Right now it's just the 2 colors, but worth thinking ahead - if we have a warning color or an error color or hyperlink color etc we probably don't want to have a large `if` block in here

AMZN-alexpete (commented):
looking good just have some suggestions

Code/Tools/ProjectManager/Source/GemCatalog/GemItemDelegate.cpp
tkothadev: this might tie into your suggestion earlier with using color as an explicit parameter. Doing that may require adding that as a parameter at every other call site (which is technically outside scope of PR), but that may help with the long run.

tkothadev: actually... I removed the only portion of text that would require a greyed out color. So I can just remove the greyed out logic altogether.

Code/Tools/ProjectManager/Source/GemCatalog/GemItemDelegate.cpp
tkothadev: per my update in the second comment regarding this, I realized I can remove the colored text drawing for this PR, since it's not being used at all. As such this topic may need to be revisited at a later time.

amzn-leenguy:
Thanks for the update TJ on removing the Unspecified wording. From a UX POV, i was concerned it was going to cause too much visual content clutter. If anyone has any data points or reasoning else wise, please comment below. 

UX approved.

amzn-leenguy (approved):

Code/Tools/ProjectManager/Source/GemCatalog/GemItemDelegate.cpp
```diff
@@ -374,6 +374,48 @@ namespace O3DE::ProjectManager
...
+        }
+        
+        //UX prefers that we show platforms in reverse alphabetical order
+        for(int i = GemInfo::NumPlatforms-1; i >= 0; i--)
+        {
+            const GemInfo::Platform platform = static_cast<GemInfo::Platform>(1 << i);
+            if (platforms & platform)
+            {
+                const QString& singlePlatformText = GemInfo::GetPlatformString(platform);
+                if(i != GemInfo::NumPlatforms-1)
```
AMZN-Phil, line 403: Shouldn't we just be checking that platformText isn't empty? It looks like if a gem doesn't support Windows but supports other platforms then we'll be adding a comma to the start of the string.

tkothadev: I just confirmed the bug occurs as you listed here. I think it would be better to switch to building a QStringList and then using join on `,`

tkothadev: this will be fixed in my upcoming commit for this PR

AMZN-Phil (approved):

AMZN-alexpete (approved):

PR #13108, opened 2022-11-10
Scroll to component card when component mode activates

amzn-ahmadkrm:

## What does this PR do?

A QOL fix that was not an issue previously as the user had to scroll to the component card to click on the "Edit" button, but now the user can click on the switcher to do the same thing. The Entity Inspector doesn't scroll to the component entering component mode so the user cannot automatically see the editing details unless they scroll manually. This PR auto scrolls to the card.

Before:

https://user-images.githubusercontent.com/105638312/201067901-f515fdd9-2965-40be-a994-50054cf8d1d8.mp4

After:

https://user-images.githubusercontent.com/105638312/201068050-5fc6e3b5-2df2-4614-ad69-08366d709cfd.mp4

## How was this PR tested?

Manual testing.


Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/EntityPropertyEditor.cpp
```diff
@@ -5899,7 +5900,12 @@ namespace AzToolsFramework
 
             for (auto componentEditor : m_componentEditors)
             {
-                componentEditor->EnteredComponentMode(componentModeTypes);
+                auto enteredComponentMode = componentEditor->EnteredComponentMode(componentModeTypes);
```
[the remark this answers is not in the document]
amzn-ahmadkrm: The function name is a little confusing, It checks whether the componentEditor in the for loop has entered component mode or not and if not disables them, it should be named something like ProcessComponentEditorOnEnteredComponentMode but thats horrible. Currently the false return does actually mean that component mode was not entered for that component so hasHighlightedItem doesn't make too much sense either, any suggestions?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/ComponentEditor.hxx
```diff
@@ -101,7 +101,7 @@ namespace AzToolsFramework
 
         void SetComponentOverridden(const bool overridden);
 
-        void EnteredComponentMode(const AZStd::vector<AZ::Uuid>& componentModeTypes);
+        bool EnteredComponentMode(const AZStd::vector<AZ::Uuid>& componentModeTypes);
```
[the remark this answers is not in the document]
amzn-ahmadkrm: Yeah I agree! I was writing a reply to your above comment but then saw this one. It isn't aptly named, I'll make a change to something a bit more appropriate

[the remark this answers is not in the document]
hultonha: +1 a comment here would be helpful - be careful changing the name as that's technically an API break so would be good to check this isn't used in many other places and/or is public/private.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/EntityPropertyEditor.cpp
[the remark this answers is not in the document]
hultonha: +1 good spot - it is also a bit confusing having the return value here be a bool as it does really look like it might be 'component mode was enabled' as @mbalfour-amzn describes - I know this came up before but maybe an enum here might actually be useful to make the code more readable? (depending on your perspective 😝)

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/EntityPropertyEditor.cpp
```diff
@@ -5922,6 +5928,9 @@ namespace AzToolsFramework
                 componentEditor->LeftComponentMode(componentModeTypes);
             }
 
+            // return to scroll position before EditorMode was activated
+            m_gui->m_componentList->verticalScrollBar()->setValue(m_verticalScrollPos);
+
```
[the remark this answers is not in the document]
hultonha: To go one step further you could even make the value `optional` and only give it a valid state while Component Mode is running to make it even harder to misuse 🙂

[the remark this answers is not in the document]
hultonha: update: It might also be worth calling it `componentModeVerticalScrollOffset` or something more specific to Component Mode too

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/EntityPropertyEditor.hxx
```diff
@@ -254,6 +254,9 @@ namespace AzToolsFramework
         void OnEditorModeDeactivated(
             const AzToolsFramework::ViewportEditorModesInterface& editorModeState, AzToolsFramework::ViewportEditorMode mode) override;
 
+        // Save the position of the scroll bar which is returned to OnEditorModeDeactivated
```
hultonha, line 257: update: Would be good if the comment mentioned Component Mode here too

hultonha (commented):
Great feature to add but I agree with @mbalfour-amzn's comments and think there's a few nice little improvements to be made before merging. Good spot though and great bias for action 🙂

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/EntityPropertyEditor.cpp
```diff
@@ -5899,10 +5900,15 @@ namespace AzToolsFramework
...
+                // if this component editor entered component mode
+                if (componentEditor->EnteredComponentMode(componentModeTypes) == ComponentEditor::ComponentEditorState::DidEnterComponentMode)
+                {
+                    // scroll to the relevant component card
+                    m_gui->m_componentList->verticalScrollBar()->setValue(componentEditor->pos().y());
+                }
             }
 
-            // record the selected state after entering component mode
+            // record the selected state after entering component modse
```
hultonha, line 5911: update: Need to fix the typo here

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/EntityPropertyEditor.hxx
```diff
@@ -254,6 +254,9 @@ namespace AzToolsFramework
         void OnEditorModeDeactivated(
             const AzToolsFramework::ViewportEditorModesInterface& editorModeState, AzToolsFramework::ViewportEditorMode mode) override;
 
+        /// Save the offset of the scroll bar before entering Component Mode to return to after leaving Component Mode
+        AZStd::optional<int> m_componentModeVerticalScrollOffset = AZStd::nullopt;
```
hultonha, line 258: update: This will be default constructed, you can omit `= AZStd::nullopt`

Code/Framework/AzToolsFramework/Tests/UI/EntityPropertyEditorTests.cpp
```diff
@@ -279,4 +297,44 @@ namespace UnitTest
         EXPECT_EQ(selectedEntityIds[0], m_levelEntity);
     }
 
+    TEST_F(LevelEntityPropertyEditorRequestTest, GetSelectedEntitiesForLevelInspectorWhenLevelI2sLoaded)
+    {
+        m_levelOpen = true;
+        AZ::Entity* entity = nullptr;
+        AZ::EntityId entityId = CreateDefaultEditorEntity("ComponentModeEntity", &entity);
+
+        // connect to EditorDisabledCompositionRequestBus with entityI
```
hultonha, line 306: update: Looks like a 'd' is missing here

Code/Framework/AzToolsFramework/Tests/UI/EntityPropertyEditorTests.cpp
```diff
@@ -279,4 +297,44 @@ namespace UnitTest
...
+
+        AzToolsFramework::ComponentModeFramework::ComponentModeSystemRequestBus::Broadcast(
+            &AzToolsFramework::ComponentModeFramework::ComponentModeSystemRequests::AddSelectedComponentModesOfType,
+            DependentPlaceholderEditorComponent::RTTI_Type());
+
+        auto a = m_levelEditor->GetVerticalScroll();
+
+        // Make sure the correct number of entities are returned
+        EXPECT_EQ(a, 0);
+        //EXPECT_EQ(selectedEntityIds[0], m_levelEntity);
```
hultonha, line 337: update: Please don't check in commented out code and please avoid using single letter variable names (other than in mathsy equation heavy code where it can be appropriate at times)

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/ComponentEditor.cpp
```diff
@@ -922,23 +922,26 @@ namespace AzToolsFramework
...
         {
             if (!componentModeTypes.empty())
             {
                 // only set the first item to be selected/highlighted
                 SetSelected(componentModeTypes.front() == m_componentType);
+                return ComponentEditor::ComponentEditorState::DidEnterComponentMode;
             }
         }
+        return ComponentEditor::ComponentEditorState::DidNotEnterComponentMode;
     }
```
hultonha, line 945: update: I don't think the enum naming here is right, as in both cases we do actually enter component mode afaik, it's just change how we visualize the Entity Inspector. I think here it's more like... `ComponentCardDisabled` and `ComponentCardSelected` - that's what's actually happening to the UI (and we only select the 'first' component card when we're entering two dependent Component Modes at once).

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/ComponentEditor.hxx
```diff
@@ -101,7 +101,16 @@ namespace AzToolsFramework
 
         void SetComponentOverridden(const bool overridden);
 
-        void EnteredComponentMode(const AZStd::vector<AZ::Uuid>& componentModeTypes);
+        enum ComponentEditorState
+        {
+            DidEnterComponentMode,
+            DidNotEnterComponentMode
+        };
```
hultonha, line 108: update: Thanks for adding the enum @amzn-ahmadkrm but if you see my earlier comment I think we can have more accurate names - also we should use `enum class` here not just `enum`

hultonha (commented):
Few more bits of feedback but definitely heading in the right direction 👍

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/ComponentEditor.hxx
```diff
@@ -101,7 +101,16 @@ namespace AzToolsFramework
...
+
+        /// Disables the ComponentEditor if the componentModeType has not entered component mode.
+        /// Returns true and highlights the ComponentEditor widget if the componentModeType has enterered component mode.
+        ComponentEditorState EnteredComponentMode(const AZStd::vector<AZ::Uuid>& componentModeTypes);
```
hultonha, line 112: update: This comment isn't accurate I'm afraid - if we update the names we can say 'Returns the state of this specific component card when Component Mode is entered (inactive - that component isn't in component mode, active - that component is in component mode, selected - that component is in component mode and is highlighted/selected (orange outline).

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/ComponentEditor.cpp
```diff
@@ -930,18 +930,18 @@ namespace AzToolsFramework
...
             if (!componentModeTypes.empty())
             {
                 // only set the first item to be selected/highlighted
                 SetSelected(componentModeTypes.front() == m_componentType);
-                return ComponentEditor::ComponentEditorState::DidEnterComponentMode;
+                return ComponentEditor::ComponentEditorState::ComponentCardSelected;
             }
         }
-        return ComponentEditor::ComponentEditorState::DidNotEnterComponentMode;
+        return ComponentEditor::ComponentEditorState::ComponentCardDisabled;
```
hultonha, line 944: update: Sorry I think we actually need three enum states here as if we get to this point, the component card hasn't been updated, it's been left 'as-is' (not disabled and not selected). Try running the code path when you have a Spline and Tube Shape on an entity at the same time to see what I mean. Can you please add this case too and return it here 'ComponentCardUnmodified' maybe?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/ComponentEditor.hxx
```diff
@@ -101,7 +101,15 @@ namespace AzToolsFramework
 
         void SetComponentOverridden(const bool overridden);
 
-        void EnteredComponentMode(const AZStd::vector<AZ::Uuid>& componentModeTypes);
+        enum class ComponentEditorState
+        {
+            ComponentCardSelected,
+            ComponentCardDisabled
+        };
```
mbalfour-amzn, line 108: Naming suggestion, take it or leave it:
"ComponentEditorState" -> "ComponentModeActiveState". 

(Because this enum is specifically used to say whether or not the component card is active and editable during Component Mode, and doesn't reflect the state of the ComponentEditor at any other time)

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/EntityPropertyEditor.cpp
```diff
@@ -5922,6 +5929,13 @@ namespace AzToolsFramework
                 componentEditor->LeftComponentMode(componentModeTypes);
             }
 
+            // return to the scroll offset before Component Mode was activated
+            if (m_componentModeVerticalScrollOffset.has_value())
+            {
+                m_gui->m_componentList->verticalScrollBar()->setValue(m_componentModeVerticalScrollOffset.value());
+                m_componentModeVerticalScrollOffset = AZStd::nullopt;
+            }
+
```
mbalfour-amzn, line 5938: Hm. After looking at this a bit more, should we actually be doing this? It definitely makes sense to scroll _to_ the selected component when entering Component Mode, but it seems like it would be a bit jarring to scroll _away_ from it when exiting Component Mode, especially if the user has scrolled that pane since then.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/EntityPropertyEditor.cpp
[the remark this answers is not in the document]
hultonha: Resolving as I believe this has now been addressed

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/ComponentEditor.hxx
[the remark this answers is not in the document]
hultonha: Resolving as I believe this has now been addressed

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/ComponentEditor.cpp
hultonha: I am pretty certain the final state can be reached, but it only happens in the case of something like the Tube Shape Component and Spline Component both being on the Entity at the same time and Tube Shape Component Mode starts. We keep both Spline and Tube active, but one is the 'selected' one (for the Component Mode you're in). That's what I'm saying the enum should reflect

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/EntityPropertyEditor.cpp
amzn-ahmadkrm: Good point! I've made a change that it only scrolls back to the original position if the user hasn't scrolled themselves during component mode, if they do scroll, it'll stay in the position that they scroll to instead.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/ComponentEditor.cpp
```diff
@@ -922,22 +922,21 @@ namespace AzToolsFramework
...
-            if (!componentModeTypes.empty())
-            {
-                // only set the first item to be selected/highlighted
-                SetSelected(componentModeTypes.front() == m_componentType);
-            }
+            // only set the first item to be selected/highlighted
+            SetSelected(componentModeTypes.front() == m_componentType);
+            return ComponentEditor::ComponentModeActiveState::ComponentCardSelected;
         }
     }
```
hultonha, line 941: update: I don't think this change is correct  - please can you put the logic back to how it was before with the three returns as I'm pretty certain they will all be hit. We need an extra enum state added to cover this (I'm going to test locally to be 100% sure but I'm pretty confident this is the case)

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/ComponentEditor.hxx
amzn-ahmadkrm: I took it :)

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/ComponentEditor.cpp
amzn-ahmadkrm: Tom and I looked into this further and it could not be reached but we want it to be reached to accurately reflect what's happening with the cards. We were returning that both cards were selected when the Tube's component mode was active but only the Tube card was being selected while the Spline card was being left alone (but returning that it was selected) The change I just pushed should make it more clear

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/EntityPropertyEditor.cpp
```diff
@@ -627,23 +627,23 @@ namespace AzToolsFramework
...
+        EditorEntityContextNotificationBus::Handler::BusDisconnect();
         ReadOnlyEntityPublicNotificationBus::Handler::BusDisconnect();
         EditorWindowUIRequestBus::Handler::BusDisconnect();
         EntityPropertyEditorRequestBus::Handler::BusDisconnect();
-        ToolsApplicationEvents::Bus::Handler::BusDisconnect();
         AZ::EntitySystemBus::Handler::BusDisconnect();
-        EditorEntityContextNotificationBus::Handler::BusDisconnect();
-        AzToolsFramework::ComponentModeFramework::EditorComponentModeNotificationBus::Handler::BusDisconnect();
-        ViewportEditorModeNotificationsBus::Handler::BusDisconnect();
+        ToolsApplicationEvents::Bus::Handler::BusDisconnect();
```
amzn-ahmadkrm, line 647: Changed the order of the buses to be FILO

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/ComponentEditor.cpp
```diff
@@ -922,21 +922,31 @@ namespace AzToolsFramework
...
         else
         {
-            if (!componentModeTypes.empty())
+            if (componentModeTypes.front() == m_componentType)
             {
                 // only set the first item to be selected/highlighted
-                SetSelected(componentModeTypes.front() == m_componentType);
+                SetSelected(true);
+                return ComponentCardState::Selected;
+            }
```
mbalfour-amzn, line 946: I apologize for the conflicting guidance here, but to reiterate from my previous comment, there's no guarantee that SetSelected() is actually selecting the item:
```
    void ComponentEditor::SetSelected(bool selected)
    {
        // do not allow cards to be selected when they are disabled
        // (show the yellow outline)
        if (!isEnabled())
        {
            return;
        }

        AzQtComponents::Card::setSelected(selected);
    }
```
I realize that to get to this point where the component is on the list of componentModeTypes, it should very _likely_ always be selectable based on the current editor logic, I'm just suggesting that this is introducing a point of fragility should any logic elsewhere in the Editor ever change. From that point of view, is there any reason to use this enum at all, vs just calling IsSelected() in the places that are adjusting the scrolling?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/EntityPropertyEditor.cpp
mbalfour-amzn: Oh, I was trying to suggest to _never_ change the scroll offset back to what it was, but I guess that works too.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/ComponentEditor.cpp
hultonha: This is a fair point and I discussed this with @amzn-ahmadkrm but I think for this change this is outside the scope of the PR as `SetSelected` already doesn't return if it succeeded or not and we were already relying on this logic before so I think leaving as-is for now is okay and we can always revisit this in future if other problems arise.

hultonha (approved):
Changes look good, thanks for making the updates. I only have very minor nits now. Thanks!

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/ComponentEditor.cpp
```diff
@@ -924,19 +924,22 @@ namespace AzToolsFramework
...
             SetWidgetInteractEnabled(this, false);
         }
         else
         {
-            if (!componentModeTypes.empty())
+            if (componentModeTypes.front() == m_componentType)
             {
                 // only set the first item to be selected/highlighted
-                SetSelected(componentModeTypes.front() == m_componentType);
+                SetSelected(true);
```
mbalfour-amzn, line 942: TLDR: I think all of this can get reverted?

Longer version:

This part is redundant, because the find() call will _also_ set things to false if componentModeTypes is empty, since the find() would always return false.
```
       if (componentModeTypes.empty())
        {
            SetWidgetInteractEnabled(this, false);
        }
```

This part I think breaks the logic in here, because now there's nothing calling SetSelected(false) in the case where componentModeTypes is NOT empty and the widget is in componentModeTypes and the widget isn't the first one:
```
            if (componentModeTypes.front() == m_componentType)
            {
                // only set the first item to be selected/highlighted
                SetSelected(true);
```

My suggestion would actually be to shorten the implementation down a little bit to this:
```
        if (AZStd::find(
            componentModeTypes.begin(),
            componentModeTypes.end(), m_componentType) == componentModeTypes.end())
        {
            // disable all component cards that aren't in the active component mode.
            SetWidgetInteractEnabled(this, false);
        }
        else
        {
            // For components that *are* in the active component mode, try to set the first one
            // to selected, and the rest to unselected.
            SetSelected(componentModeTypes.front() == m_componentType);
        }
```

mbalfour-amzn (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/EntityPropertyEditor.hxx
```diff
@@ -245,24 +247,29 @@ namespace AzToolsFramework
...
             const AzToolsFramework::ViewportEditorModesInterface& editorModeState, AzToolsFramework::ViewportEditorMode mode) override;
 
-        // EntityPropertEditorRequestBus
+        // EditorComponentModeNotificationBus overrides ...
+        void ActiveComponentModeChanged(const AZ::Uuid& componentType) override;
+
+        // Save the offset of the scroll bar before entering Component Mode to return to after leaving Component Mode
+        AZStd::optional<int> m_verticalScrollOffset;
+        // Save the offset of scroll bar when Component Mode has been entered
+        AZStd::optional<int> m_componentModeVerticalScrollOffset;
```
hultonha, line 262: update: Should these now be delete now? I don't see them referenced anywhere else

hultonha (approved):
LGTM, can you please just remove the unused variables now, thanks!

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/EntityPropertyEditor.cpp
```diff
@@ -5900,6 +5901,13 @@ namespace AzToolsFramework
             for (auto componentEditor : m_componentEditors)
             {
                 componentEditor->EnteredComponentMode(componentModeTypes);
+
+                // if this component editor is active and editable during component mode
+                if (componentEditor->IsSelected())
+                {
+                    // scroll to the relevant component card
+                    m_gui->m_componentList->verticalScrollBar()->setValue(componentEditor->pos().y());
```
mbalfour-amzn, line 5909: Just double-checking, it seemed like this might have some issues with scrolling to the right location for some combinations of components and overall Editor window size. Are those all working correctly now?

If not, maybe this should be using m_componentList->ensureWidgetVisible() or something instead...

hultonha: Agreed it feels like we should be factoring the overall visible scroll area size, the component card size, the total scrollable distance (including wrapped area that's offscreen) and the current position - hopefully you can access those values from Qt.

hultonha (approved):
New change looks good! Thanks for getting to the bottom of this @amzn-ahmadkrm :+1:

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/EntityPropertyEditor.hxx
amzn-ahmadkrm: Thank you for spotting this!

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/EntityPropertyEditor.cpp
amzn-ahmadkrm: I changed it to use ensureWidgetVisible and it works consistently now :)

hultonha (approved):

mbalfour-amzn (approved):

PR #13116, opened 2022-11-10
AZ Networking ISerialize Support for AZ::s64 and AZ::u64

AMZN-Gene:
Allows AzNetworking to serialize AZ::s64 and AZ::u64.
AzNetworking ISerialize current supports int64_t and uint64_t, which (depending on the system) is sometimes long and sometimes long long. AZ types such as AZ::u64 are set up to always be defined as long long. This change makes it so that if int64_t is long (not long long), then we'll define a new method to support long long.

Fixes https://github.com/o3de/o3de/issues/13019

Signed-off-by: Gene Walters <genewalt@amazon.com>

### DIFF
--- a/Code/Framework/AzNetworking/AzNetworking/Serialization/ISerializer.h
+++ b/Code/Framework/AzNetworking/AzNetworking/Serialization/ISerializer.h
@@ -108,10 +108,7 @@ namespace AzNetworking
         //! @param maxValue the maximum value expected during serialization
         //! @return boolean true for success, false for failure
         #if AZ_TRAIT_COMPILER_INT64_T_IS_LONG
-        bool Serialize(AZ::s64& value, const char* name, AZ::s64 minValue = AZStd::numeric_limits<AZ::s64>::min(), AZ::s64 maxValue = AZStd::numeric_limits<AZ::s64>::max())
-        {
-            return Serialize(static_cast<int64_t>(value), name, minValue, maxValue);
-        }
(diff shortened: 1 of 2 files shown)


AMZN-puvvadar (approved):

Code/Framework/AzNetworking/AzNetworking/Serialization/ISerializer.h
```diff
@@ -100,6 +101,19 @@ namespace AzNetworking
...
 
+        //! Serialize a signed 64-bit integer.
+        //! @param value    signed 64-bit integer input value to serialize
+        //! @param name     string name of the value being serialized
+        //! @param minValue the minimum value expected during serialization
+        //! @param maxValue the maximum value expected during serialization
+        //! @return boolean true for success, false for failure
+        #if AZ_TRAIT_COMPILER_INT64_T_IS_LONG
+        bool Serialize(AZ::s64& value, const char* name, AZ::s64 minValue = AZStd::numeric_limits<AZ::s64>::min(), AZ::s64 maxValue = AZStd::numeric_limits<AZ::s64>::max())
+        {
```
kberg-amzn, line 112: Definitely put these definitions in the inline file however (ISerializer.inl).. we should strive to keep our interfaces interfacey so they stay easy for customers to read.

AMZN-Gene: Fixed! Thank you.

kberg-amzn (approved):

Issue #13064, opened 2022-11-08
Baking image using default file name in the Gradient Baker causes an Asset Processor fail to pop up


LB-ArturZieba:
**Describe the bug**
If the image baked using the *Gradient Baker* component is saved under its default name (baked_output_gsi.png), an Asset Processor fail will pop up, even though the image saves correctly and when viewed in the Asset Processor it shows no issues.

It should be noted that this does not occur with other names.

**Steps to reproduce**
1. Create an entity and add *Perlin Noise Gradient*, *Gradient Transform Modifier* and *Box Shape* components to it.
2. Create another entity and add *Gradient Baker* component to it.
3. In *Gradient Baker* set the *Gradient Entity Id* and *Input Bounds* to the entity created in step 1.
4. In *Gradient Baker* set the *Output Path* to any path (for example the project's *Assets* folder), do not change the image's name (should be baked_output_gsi.png by default).
5. Click the *Bake image* button on the *Gradient Baker* component.

**Expected behavior**
No fail pops up.

**Actual behavior**
An Asset Processor fail pops up.

**Video and Asset Processor screenshot**

https://user-images.githubusercontent.com/86952082/200578295-d5b71366-b2db-433e-9de4-f360512e29ab.mp4

Images generated on the video in the Asset Processor:
![GradientBakerImagesAssetProcessor](https://user-images.githubusercontent.com/86952082/200578313-6c81debf-19ff-426c-ac38-93aac3d8b0fd.png)

**Found in Branch**
Development (7f642a1)

**Desktop**
 - Device: PC
 - OS: Windows
 - Version 11
 - CPU AMD Ryzen 5 3600
 - GPU NVIDIA GeForce RTX 2060 SUPER
 - Memory 16GB

AMZN-stankowi:
This isn't just for this asset type, this occurs with any non-instant write operation to asset files. If you copy a large file, or perform a longer file save operation, the asset processor will start attempting to process the file before the operation finishes, fail, and then need to re-process the asset when the write operation actually finishes.

LB-ArturZieba:
Issue verified as fixed on the Development (ba92844) build.

PR #13117, opened 2022-11-10
[Terrain] Prevent AP from processing image until after it's done saving

mbalfour-amzn:
## What does this PR do?

When painting or baking an image gradient, it was possible for the Asset Processor to start processing the asset before the image was fully written, which would result in an incorrect processed asset. Furthermore, the AP didn't know to reprocess the asset without a manual trigger.

This change uses special temp file naming to make the AP ignore the image while it's being written, then renames it _after_ it's done so that the AP processes the fully-written image.

Also fixed a tiny null dereference problem in the image builder that could happen occasionally if the source image couldn't be read in for some reason.

## How was this PR tested?

Tried painting on images that were 64x64 in size and saving them. Oddly enough, tiny images like these were triggering the problem 100% of the time for me. They now trigger it 0% of the time.

### DIFF
--- a/Gems/GradientSignal/Code/Source/Editor/EditorGradientImageCreatorUtils.cpp
+++ b/Gems/GradientSignal/Code/Source/Editor/EditorGradientImageCreatorUtils.cpp
@@ -211,22 +211,22 @@ namespace GradientSignal::ImageCreatorUtils
             return false;
         };
 
-        bool result = outputImage->write_image(
+        bool writeResult = outputImage->write_image(
             pixelFormat, pixelBuffer.data(), OIIO::AutoStride, OIIO::AutoStride, OIIO::AutoStride, WriteProgressCallback, saveDialog);
-        if (!result)
+        if (!writeResult)
         {
-            AZ_Error("EditorImageGradientComponent", result, "Failed to write out gradient image to path: %s", tempSavePath.c_str());
+            AZ_Error("EditorImageGradientComponent", writeResult, "Failed to write out gradient image to path: %s", tempSavePath.c_str());
         }
 
         outputImage->close();
 
         // Now that we're done saving the temporary image, rename it to the correct file name.
-        [[maybe_unused]] auto moveResult = AZ::IO::SmartMove(tempSavePath.c_str(), absolutePath.c_str());
+        auto moveResult = AZ::IO::SmartMove(tempSavePath.c_str(), absolutePath.c_str());
         AZ_Error("EditorImageGradientComponent", moveResult,
             "Failed to rename temporary image asset %s to %s", tempSavePath.c_str(), absolutePath.c_str());
 
         delete saveDialog;
-        return result;
+        return writeResult && moveResult;
     }
 
 }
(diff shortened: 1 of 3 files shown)


lawsonamzn (approved):

Gems/GradientSignal/Code/Source/Editor/EditorGradientImageCreatorUtils.cpp
```diff
@@ -211,11 +215,16 @@ namespace GradientSignal::ImageCreatorUtils
...
-            AZ_Error("EditorImageGradientComponent", result, "Failed to write out gradient image to path: %s", absolutePath.c_str());
+            AZ_Error("EditorImageGradientComponent", result, "Failed to write out gradient image to path: %s", tempSavePath.c_str());
         }
 
         outputImage->close();
 
+        // Now that we're done saving the temporary image, rename it to the correct file name.
+        [[maybe_unused]] auto moveResult = AZ::IO::SmartMove(tempSavePath.c_str(), absolutePath.c_str());
+        AZ_Error("EditorImageGradientComponent", moveResult,
+            "Failed to rename temporary image asset %s", tempSavePath.c_str(), absolutePath.c_str());
```
invertednormal, line 226: You're passing both path strings into this message, but it only takes one.

invertednormal (changes requested):
Just one small thing needs to be fixed.

Gems/GradientSignal/Code/Source/Editor/EditorGradientImageCreatorUtils.cpp
mbalfour-amzn: Whoops ty! It surprisingly didn't complain!

lawsonamzn: nicely spotted!

cgalvan (approved):

cgalvan:
@mbalfour-amzn  Does this fix https://github.com/o3de/o3de/issues/13064 ?

invertednormal (approved):

Gems/GradientSignal/Code/Source/Editor/EditorGradientImageCreatorUtils.cpp
```diff
@@ -211,11 +215,16 @@ namespace GradientSignal::ImageCreatorUtils
...
         {
-            AZ_Error("EditorImageGradientComponent", result, "Failed to write out gradient image to path: %s", absolutePath.c_str());
+            AZ_Error("EditorImageGradientComponent", result, "Failed to write out gradient image to path: %s", tempSavePath.c_str());
         }
 
         outputImage->close();
 
+        // Now that we're done saving the temporary image, rename it to the correct file name.
+        [[maybe_unused]] auto moveResult = AZ::IO::SmartMove(tempSavePath.c_str(), absolutePath.c_str());
+        AZ_Error("EditorImageGradientComponent", moveResult,
```
jackalbe, line 225: Should "moveResult" be used in the return value?

mbalfour-amzn: Yep, good catch!

PR #13118, opened 2022-11-10
Fixed Android running on VFS. 

moraaar:

## What does this PR do?

The following fix https://github.com/o3de/o3de/pull/12509 caused Android to start crashing when running with VFS in AtomSampleViewer project.

This happens because as soon as the connection with Asset Processor is stablished it's sending a `BulkAssetNotificationMessage` with `AssetChanged` events for all the assets, but with VFS that's happening before the catalog is initialized (with VFS the initialization happens after connecting to AP and reads the catalog from the remote system). Skipping `AssetChanged` event for this scenario when the catalog is supposed to be initialized fixes the issue.

Also added `CriticalAssetsCompiled` printfs as it's useful to know when they happen in the log. The editor already has one, added the one for the launcher and atom tools.

## How was this PR tested?

- Run ASV on Android using VFS now doesn't crash.
- Run editor with ASV and XR samples.

### DIFF
--- a/Code/Framework/AzFramework/AzFramework/Asset/AssetCatalog.cpp
+++ b/Code/Framework/AzFramework/AzFramework/Asset/AssetCatalog.cpp
@@ -810,6 +810,16 @@ namespace AzFramework
     // the results.  For this reason, it is a direct call and protects its structures with a lock.
     void AssetCatalog::AssetChanged(const AZStd::vector<AzFramework::AssetSystem::AssetNotificationMessage>& messages, bool isCatalogInitialize)
     {
+        if (isCatalogInitialize && !m_initialized)
+        {
+            // This can happen when running with VFS, where the AP connection is done first
+            // and it's too early to send asset changed messages since catalog is not initialized yet.
+            AZ_TracePrintf(
+                "AssetCatalog",
+                "Skipping AssetChanged to update the catalog since it's too early, catalog has not been initialized yet.");
+            return;
+        }
+
         for (const auto& message : messages)
         {
             AZStd::string relativePath = message.m_data;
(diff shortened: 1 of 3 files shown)


amzn-mike (approved):

Code/Framework/AzFramework/AzFramework/Asset/AssetCatalog.cpp
```diff
@@ -810,6 +810,16 @@ namespace AzFramework
...
     {
+        if (isCatalogInitialize && !m_initialized)
+        {
+            // This can happen when running with VFS, where the AP connection is done first
+            // and it's too early to send asset changed messages since catalog is not initialized yet.
+            AZ_TracePrintf(
+                "AssetCatalog",
+                "Skipping AssetChanged to update the catalog since it's too early, catalog has not been initialized yet.");
+            return;
+        }
```
lumberyard-employee-dm, line 821: @AMZN-stankowi @amzn-mike Will this cause assets not in the AssetCatalog.xml on the host machine within the `<project-path>/Cache/android` directory to not send over the asset changed events if the AP happened to have process new android assets during it's run (because it had `android` platform enabled), before the remote Android system connected to it?

moraaar: Not sure that's an scenario expected to work for android at this time. Depending on the packing mode used:
- Assets in APK: the android assets must be fully generated before building android.
- Assets in sdcard: the android assets must be fully generated before deploying to device.
- Assets by VFS: the minimum assets (.setreg files) must be generated in the cache before deploying to device (otherwise it'll crash), and any the other assets are picked up at runtime when the app is launched, read by the host. I'd say that protecting against the host not processed the android asset when android app is launched is a bit of a stretched scenario to worry about at this point.

lawsonamzn (approved):
seems legit since if the asset catalog is not initialized, it will initialize later and get caught up on these assets in the interim.  IIRC when the asset catalog is initialized it also makes sure to flush the AP's write of the asset catalog, which it reads over VFS.

lumberyard-employee-dm (approved):

PR #13121, opened 2022-11-10
Made an example to execute a lua file from command line

jiaweig-amzn:

## What does this PR do?

Added a console command to execute lua script via console.
Made an example to take a screenshot in the default level.

This is a rough sample of the current framework, which needs more a bigger revision and add more abilities, such as
1. adding more notifications instead of doing a time based wait (when loading level).
2. adding richer abilities to control level setting up.
3. add a reporting module like the one in ASV, or something with an equivalent function.
4. add the ability to hide Imgui after launching a script.

## How was this PR tested?

Tested running the script in both launcher and editor.


AutomatedTesting/Assets/AutomatedTestScripts/screenshot.sa.lua
```diff
@@ -9,12 +9,12 @@
 --
 ----------------------------------------------------------------------------------------------------
 
-g_screenshotOutputFolder = GetScreenshotOutputPath(true)
-Print('Saving screenshots to ' .. NormalizePath(g_screenshotOutputFolder))
+g_screenshotOutputFolder = '@user@/ScriptAutomation/Screenshots/'
+Print('Saving screenshots to ' .. ResolvePath(g_screenshotOutputFolder))
```
rgba16f, line 13: Do these lines do anything? CaptureScreenshot builds it's own capture path now.

gadams3 (approved):

rgba16f (approved):

AutomatedTesting/Assets/AutomatedTestScripts/screenshot.sa.lua
jiaweig-amzn: It just prints out the message to the console, which doesn't happen when a capture is taken.
I can make this a script call so that things can become neater.

jiaweig-amzn: It just prints out the message to the console, which doesn't happen when a capture is taken.
I can make this a script call so that things can become neater.

Issue #13114, opened 2022-11-10
AssetProcessor - Removing a gem with assets causes a crash on next AP start


lawsonamzn:
**Describe the bug**
If you deactivate a gem that has assets and restart, AP crashes.

**Assets required**
Nothing in particular. 

**Steps to reproduce**
1. Use the Project Manager to activate an gem not currently activated which has assets.
2. Get editor/ap started so that the assets appear and have been built (or even errored is ok)
3. quit editor and ap
4. Use the Project Manager to deactivate that gem
5. Build and try to start editor/ap again

**Expected behavior**
AP should start, the files that are no longer relevant should be removed from the DB and possibly from disk

**Actual behavior**
AP Crashes.  VArious parts of it try to dereference the now deleted scanfolder which no longer exists.

**Found in Branch**
dev

**Desktop/Device (please complete the following information):**
its gonna happen on everything.

PR #13125, opened 2022-11-10
Asset Processor - Fix missing scanfolder crash

amzn-mike:
## What does this PR do?

Fixed issue where AP crashed when running after having processed with a scanfolder which no longer exists.  A few places in the code were assuming a 'find scanfolder' call would always have a return value.

Fixes https://github.com/o3de/o3de/issues/13114

## How was this PR tested?

Followed repro steps in https://github.com/o3de/o3de/issues/13114

### DIFF
--- a/Code/Tools/AssetProcessor/native/AssetManager/SourceAssetReference.cpp
+++ b/Code/Tools/AssetProcessor/native/AssetManager/SourceAssetReference.cpp
@@ -104,6 +104,22 @@ namespace AssetProcessor
         Normalize();
     }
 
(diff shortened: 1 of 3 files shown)


cgalvan (approved):

lawsonamzn (approved):
It looks okay, but would this actually clean the files up as it should?

amzn-mike:
> It looks okay, but would this actually clean the files up as it should?

@lawsonamzn  I've updated it (and tested to confirm) so that the files from missing scanfolders actually get cleared out of the cache.

lawsonamzn (approved):

PR #13127, opened 2022-11-10
Prevent entities being marked dirty outside of an undo batch

michabr:

## What does this PR do?

This PR checks if there is a current undo batch in progress before adding an entity to the dirty entity list. This prevents entities being dirtied unintentionally such as during level load. An example of this is the TransformComponent calling SetDirty() from its Activate().

This manifested as a performance issue for large levels in the prefab system because the dirty entity list became bloated after level load causing the next user undo operation, such as entity selection, to take a long time since it was creating undo commands for all those dirty entities.

## How was this PR tested?

Tested common undo/redo operations in editor. Opened/closed/switched levels.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Application/ToolsApplication.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Application/ToolsApplication.cpp
@@ -1383,6 +1383,14 @@ namespace AzToolsFramework
             return;
         }
 
+        // Only accept entities as dirty between a begin/end undo batch. This filters out entities
+        // that have been marked dirty by a non-user operation such as entity activation. This can
+        // occur if SetDirty is called from a common function used in both user and non-user actions
+        if (!m_currentBatchUndo)
+        {
+            return;
+        }
+
         m_dirtyEntities.insert(entityId);
 
         // Check if this dirty entity is in a layer by walking up its parenting hierarchy.
--- a/Code/Framework/AzToolsFramework/Tests/ToolsComponents/EditorLayerComponentTests.cpp
+++ b/Code/Framework/AzToolsFramework/Tests/ToolsComponents/EditorLayerComponentTests.cpp
@@ -580,6 +580,8 @@ namespace AzToolsFramework
         m_layerEntity.m_layer->ClearUnsavedChanges();
 
         AZ::Entity* childEntity = CreateEditorReadyEntity("ChildEntity");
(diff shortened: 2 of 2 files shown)


srikappa-amzn (approved):
Looks good to me! It'll be good to send an impactful change broadcast along with this change. In the broadcast message, you can provide best practices and guidelines for other systems and individual component owners for marking an entity dirty

forkercat (approved):

cgalvan (approved):

lawsonamzn (approved):
(For future) maybe we can figure out a way to do some sort of check when we compute the actual undo to warn if various entities are dirty but result in no patch.

AMZN-daimini:
@michabr and I had a short chat about this and while I don't think this change will break anything, I believe it kind of conflicts with our current stance with backwards compatibility in the Prefab system.

For example, the View Bookmark system will check whether the level that is being loaded has a View Bookmark component on its prefab container entity, and if not it will add it. This is a backwards-compatibility check to ensure old levels can still be loaded and the Editor works as expected in new versions.

When this happens, the expectation is that the prefab template is updated, and the level is shown as dirty in the Outliner so that the user knows a change was applied to it, and they can press Ctrl+S to actually save those changes to disk.
Will this change actually prevent that kind of change from being registered? If so, it conflicts with what I thought was the current setup for this kind of event.

Note that I'm not saying this has to be the way it works - we can change how things should work and find another solution, as long as it's standardized and we don't have different systems or components behaving in different ways and complicating the UX.

lumberyard-employee-dm (approved):

hultonha:
@AMZN-daimini To fix/improve the view bookmark case maybe we could start an undo batch ourselves if we know we're going to modify the level? I know it only resolves this one case but I don't think would be hard to do. The change does seem like a good idea, especially for the performance win it'll generate, but I agree it's always good to be cautious about these things and ensure there aren't any unintended consequences 😅

AMZN-daimini:
> @AMZN-daimini To fix/improve the view bookmark case maybe we could start an undo batch ourselves if we know we're going to modify the level? I know it only resolves this one case but I don't think would be hard to do. The change does seem like a good idea, especially for the performance win it'll generate, but I agree it's always good to be cautious about these things and ensure there aren't any unintended consequences 😅

Yes that's what we considered when chatting about it. I do feel like we need a clear standard here though before we end up with completely different solutions in each system.

lawsonamzn (approved):

lumberyard-employee-dm (approved):

Issue #9882, opened 2022-06-02
Bug: Shadow Artifacts On Terrain


galibzon:
**Describe the bug**
In the editor, as the camera move above the terrain, there are some visible shadow artifacts where some triangles get considerable darker than what they should be.

**Steps to reproduce**
Build the Editor, with `AutomatedTesting`, and load the attached level.

**Expected behavior**
No shadow artifacts in the terrain

**Actual behavior**
Some triangles are rendered darker as the camera moves in the Editior.

**Assets required**
[goal_9ks_1kd.zip](https://github.com/o3de/o3de/files/8824883/goal_9ks_1kd.zip)

**Screenshots/Video**
![image](https://user-images.githubusercontent.com/66021303/171654898-6980beea-bc0f-4558-8074-f23dd0baf867.png)

**Found in Branch**
`development` branch:
commit f648fe91c71e824798ad06a2eea3a17e686e121e (HEAD -> development, upstream/development)
Author: Allen Jackson <23512001+jackalbe@users.noreply.github.com>
Date:   Thu Jun 2 08:31:26 2022 -0500

**Desktop/Device (please complete the following information):**
 - Device: [PC] 
 - OS: [Windows]
 - Version [10]
 - CPU [AMD ThreadRipper 32/64]
 - GPU [NVidia RTX 3070]
 - Memory [64GB]

**Additional context**

galibzon:
FYI @mbalfour-amzn , @invertednormal , @jiaweig-amzn

mbalfour-amzn:
Note - this artifact is specifically related to the Shadow Filter method. It happens with "None" and "PCF", but not with "ESM".

invertednormal:
@antonmic I noticed this as well, I don't _think_ It's anything in particular that terrain is doing wrong, especially considering that esm filtering fixes the problem Do you have any more insight here?

invertednormal:
Another example - 
![image](https://user-images.githubusercontent.com/1105143/176021658-e2af8e80-bbef-4bcf-bc0c-53663c86a323.png)

galibzon:
@antonmic Today I noticed the terrain was not rendering the shadow artifacts, but the code I have is 10 days old. Then I noticed I had MSAA set to 1 (aka NOMSAA). Just a lead.

ParasiteSt:
> Note - this artifact is specifically related to the Shadow Filter method. It happens with "None" and "PCF", but not with "ESM".

[...]

PR #13131, opened 2022-11-10
Changed MSAA and SpecularAA

antonmic:
Changed MSAA to be 2x by default (better performance than 4x) with custom sample points (fix some artifacts, including this bug: https://github.com/o3de/o3de/issues/9882) and turned on specular AA by default for better image quality and less aliasing across the board

Signed-off-by: antonmic <56370189+antonmic@users.noreply.github.com>

### DIFF
--- a/Gems/Atom/Feature/Common/Assets/Materials/Types/MaterialInputs/GeneralCommonPropertyGroup.json
+++ b/Gems/Atom/Feature/Common/Assets/Materials/Types/MaterialInputs/GeneralCommonPropertyGroup.json
@@ -14,7 +14,7 @@
             "displayName": "Apply Specular AA",
             "description": "Whether to apply specular anti-aliasing in the shader.",
             "type": "Bool",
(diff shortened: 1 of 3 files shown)


rgba16f (approved):

invertednormal (approved):

Gems/Atom/Feature/Common/Assets/Materials/Types/MaterialInputs/GeneralCommonPropertyGroup.json
```diff
@@ -14,7 +14,7 @@
             "displayName": "Apply Specular AA",
             "description": "Whether to apply specular anti-aliasing in the shader.",
             "type": "Bool",
-            "defaultValue": false,
+            "defaultValue": true,
```
santorac, line 17: Not every material type uses this file, some have inlined the same content. Search for "Apply Specular AA" to see those other locations.

PR #13136, opened 2022-11-11
Prism/platform support create edit gem ux

tkothadev:
## What does this PR do?

This PR for the Create Gem and Edit Gem workflows of the Project Manager enables users to specify Platform Support for their gems.

A new UI Widget for Platform Options is introduced in this PR as well:
![PlatformSupportCheckBoxField](https://user-images.githubusercontent.com/112996779/201237012-79a97707-ec22-443a-a2a9-d81341d6684b.gif)


Platform Support in Create Gem Workflow:
![PlatformSupportCreateGemWorkflow](https://user-images.githubusercontent.com/112996779/201237084-87ed433e-7c72-45c0-8d9c-02e70fd19096.gif)

Platform Support in Edit Gem Workflow:
![PlatformSupportEditGemWorkflow](https://user-images.githubusercontent.com/112996779/201237230-0c0fbbe0-fc70-4aaf-999c-b46e33d14d93.gif)


Please note that the use of icons for platform support indication is temporary. There is a parallel PR which is aimed at text rendering for platforms: https://github.com/o3de/o3de/pull/13106

_Implementation Notes_: Some Python scripts were updated in this PR. For instance, I discovered an edge case where the scripts do not correctly set the values of newly updated platform values _if_ the user provides an empty list. The scripts use `if variable` as a stand-in check for sentinel values. The problem though is that both `None` and `[]` can be interpreted as `false` by the Python Interpreter. I've updated the code to make the checks more explicit for sentinel values (they strictly match against `None`).

In order for Create Gem to correctly add platform values to `gem.json`, the template files had to be updated to check for `${Platform}`.
## How was this PR tested?

Testing the UI was handled manually by going through various trials, such as the gifs recorded above.

Any edits made to the Python scripts also included updates to their unit tests. All such files are passing their unit tests:
```
C:\workspace\o3de(Prism/PlatformSupportCreateEditGemUX)                                                                                        
λ python\python.cmd -m pytest scripts\o3de\tests\test_engine_template.py                                                                       
Pytest custom argument "--build-directory" was not provided, tests using LyTestTools will fail                                                 
Pytest custom argument "--output-path" was not provided, defaulting Test Results output to: C:\workspace\o3de\pytest_results\2022-11-10T19-49-0
7-148016                                                                                                                                       
============================================================ test session starts ============================================================= 
platform win32 -- Python 3.10.5, pytest-6.2.5, py-1.11.0, pluggy-0.13.1                                                                        
rootdir: C:\workspace\o3de, configfile: pytest.ini                                                                                             
plugins: cov-4.0.0, mock-3.8.2, timeout-2.1.0, ly-test-tools-1.0.0                                                                             
timeout: 1200.0s                                                                                                                               
timeout method: thread                                                                                                                         
timeout func_only: False                                                                                                                       
collected 8 items                                                                                                                              
                                                                                                                                               
scripts\o3de\tests\test_engine_template.py ........                                                                                     [100%] 
                                                                                                                                               
============================================================= 8 passed in 0.24s ============================================================== 

C:\workspace\o3de(Prism/PlatformSupportCreateEditGemUX)
λ python\python.cmd -m pytest scripts\o3de\tests\test_gem_properties.py
Pytest custom argument "--build-directory" was not provided, tests using LyTestTools will fail
Pytest custom argument "--output-path" was not provided, defaulting Test Results output to: C:\workspace\o3de\pytest_results\2022-11-10T19-49-10-502611
============================================================ test session starts ============================================================= platform win32 -- Python 3.10.5, pytest-6.2.5, py-1.11.0, pluggy-0.13.1
rootdir: C:\workspace\o3de, configfile: pytest.ini
plugins: cov-4.0.0, mock-3.8.2, timeout-2.1.0, ly-test-tools-1.0.0
timeout: 1200.0s
timeout method: thread
timeout func_only: False
collected 3 items

scripts\o3de\tests\test_gem_properties.py ...                                                                                           [100%]

============================================================= 3 passed in 0.05s ==============================================================

```


amzn-leenguy:
If no text is entered for a Gem, can we allow for default placeholder text to be placed in the field? I'm thinking in the lines of generic text like: "_None specified_" (there could be better word crafting there)

amzn-leenguy (approved):
UX approved for the toggle work. Great work.

Code/Tools/ProjectManager/Source/FormOptionsWidget.h
```diff
@@ -0,0 +1,58 @@
...
+        void disable(const QString& option);
+
+        void disable(const QStringList& options);
+
+        void enableAll();
+
+        QStringList getOptions() const;
+
+    private:
+        int getCheckedCount() const;
```
AMZN-alexpete, line 51: These member function names do not match our style guide for member function names - they should all start with upper case letters.

Code/Tools/ProjectManager/Source/FormOptionsWidget.h
```diff
@@ -0,0 +1,58 @@
...
+{
+    class FormOptionsWidget : public QWidget
+    {
+        Q_OBJECT
+
+    public:
+        FormOptionsWidget(const QString& labelText,
+                          const QStringList& options,
+                          const QString& allOptionsText,
+                          const int optionItemSpacing = 24,
```
AMZN-alexpete, line 33: normally we do not mark pass-by-value parameters as const - are you intending to signal some intent about how it is used?
```suggestion
                          int optionItemSpacing = 24,
```

Code/Tools/ProjectManager/Source/GemCatalog/GemInfo.cpp
```diff
@@ -100,4 +100,56 @@ namespace O3DE::ProjectManager
...
+        for (int i = 0; i < GemInfo::NumPlatforms; ++i)
+        {
+            const GemInfo::Platform platform = static_cast<GemInfo::Platform>(1 << i);
+            if(m_platforms & platform)
+            {
+                platformStrings << GetPlatformString(platform);
+            }
+        }
+        return platformStrings;
+    }
```
AMZN-alexpete, line 154: I'm seeing this repeated code enough in the last couple of reviews I'm starting to think we might benefit from a utility `GetPlatformsAsStringList` function on the `GemInfo` class that takes in a bit mask and returns a QStringList instead of having this code copied in several places

AMZN-alexpete: Although, now that I say that, most of those other places want the list in reverse order, which complicates things

scripts/o3de/o3de/engine_template.py
```diff
@@ -2103,15 +2103,29 @@ def create_gem(gem_path: pathlib.Path,
...
         replacements.append(("${Origin}", origin))
+    else:
+        replacements.append(("${Origin}", ""))
+    
     if origin_url:
         replacements.append(("${OriginURL}", origin_url))
-    
+    else:
+        replacements.append(("${OriginURL}", ""))
+
```
AMZN-alexpete, line 2128: If we do this won't it remove all the helper text in each gem templates' gem.json or is that the goal?

Code/Tools/ProjectManager/Source/GemCatalog/GemInfo.cpp
tkothadev: Yea, there's one instance actually where we want the list in reverse order, namely in `GemItemDelegate.cpp` in this PR: https://github.com/o3de/o3de/pull/13106

I think refactoring this method may require a survey of all instances where we do this kind of pattern. Without that data, it's hard to generalize this.

Code/Tools/ProjectManager/Source/FormOptionsWidget.h
tkothadev: Main intent is that the spacing value does not change over the course of the function. If that is not necessary I can remove this.

scripts/o3de/o3de/engine_template.py
tkothadev: It shouldn't affect the helper text, since only the `${VAR_NAME}` pattern is affected in these string replacements. Any pattern that matches `${VAR_NAME}` will be affected in the destination gem.json file (the file that the create/edit gem workflow creates) however. 

If you do not add these empty string replacements, the destination file ends up looking like this (actual output of a test gem I made from the workflow):
```
{
    "gem_name": "kevingem6",
    "display_name": "6th gem of kevin",
    "license": "mit",
    "license_url": "${LicenseURL}",
    "origin": "tkothadev",
    "origin_url": "${OriginURL}",
    "type": "Code",
    "summary": "${Summary}",
    "canonical_tags": [
        "Gem"
    ],
    "user_tags": [
        "kevingem6"
    ],
    "icon_path": ".",
    "requirements": "${Requirements}",
    "documentation_url": "${DocumentationURL}",
    "dependencies": [],
    "repo_uri": "${RepoURI}",
    "restricted": "kevingem6"
}
```

In practice, dealing with this in the edit workflow was extremely annoying. Especially the last page when I go to 'Edit' the gem, it would keep autofilling the field with `${RepoURI}`, which should not exist as a value in the workflow. This would have to be cleared out everytime before I proceed. The `RepoURI` is a non-trivial example, since I am simulating creating a local gem in which I have no plans to host in a repository.

Code/Tools/ProjectManager/Source/CreateAGemScreen.h
```diff
@@ -13,6 +13,7 @@
 #include <FormFolderBrowseEditWidget.h>
 #include <FormLineEditWidget.h>
 #include <FormLineEditTagsWidget.h>
+#include <FormOptionsWidget.h>
 #include <FormComboBoxWidget.h>
```
AMZN-alexpete, line 17: because most of these are used as pointer member variables in this header, we should be forward declaring them here, and move these includes to the .cpp files at some point, doesn't have to be this review.

scripts/o3de/o3de/engine_template.py
AMZN-alexpete: OK, thanks for the clear explanation!

Code/Tools/ProjectManager/Source/CreateAGemScreen.cpp
```diff
@@ -269,6 +269,23 @@ namespace O3DE::ProjectManager
         m_requirements = new FormLineEditWidget(tr("Requirements"), "", tr("Notice of any requirements your Gem. i.e. This requires X other gem"), "");
         gemDetailsLayout->addWidget(m_requirements);
 
+
```
AMZN-Phil, line 272: Extra newlines

Code/Tools/ProjectManager/Source/CreateAGemScreen.h
```diff
@@ -30,6 +24,18 @@ QT_FORWARD_DECLARE_CLASS(QDir)
 
 namespace O3DE::ProjectManager
 {
+    class FormFolderBrowseEditWidget;
+
```
AMZN-Phil, line 28: Usually we wouldn't have empty lines between these forward declarations.

AMZN-Phil (approved):

Code/Tools/ProjectManager/Source/EditAGemScreen.cpp
```diff
@@ -107,7 +107,19 @@ namespace O3DE::ProjectManager
...
+
+        //load the supported platforms
+        for(int i = GemInfo::NumPlatforms-1; i >= 0; i--)
+        {
+            const GemInfo::Platform platform = static_cast<GemInfo::Platform>(1 << i);
+            if(platform & oldGemInfo.m_platforms)
+            {
+                m_platformOptions->Enable(GemInfo::GetPlatformString(platform));
+            }
+        }
```
lumberyard-employee-dm, line 122: spacing: There should be a space around the subtraction operator.
However a more foolproof way of iterating backwards is to defer decrementing the index until the array it needs to be accessed.

This makes it so even the index type is an unsigned integer type, then it will still work regardless.

The method currently being used only work for signed types, as unsigned types would result in an infinite loop

```suggestion
        for(int i = GemInfo::NumPlatforms; i > 0; i--)
        {
            const GemInfo::Platform platform = static_cast<GemInfo::Platform>(1 << (i - 1));
            if(platform & oldGemInfo.m_platforms)
            {
                m_platformOptions->Enable(GemInfo::GetPlatformString(platform));
            }
        }
```

scripts/o3de/o3de/engine_template.py
```diff
@@ -2103,15 +2103,29 @@ def create_gem(gem_path: pathlib.Path,
 
     if summary:
         replacements.append(("${Summary}", summary))
+    else:
+        replacements.append(("${Summary}", ""))
```
lumberyard-employee-dm, line 2107: Alternatively you could use the [comprehension-if](https://docs.python.org/3/reference/expressions.html#conditional-expressions) syntax of python do do this in one line

```suggestion
    replacements.append(("${Summary}", summary if summary else ""))
```

scripts/o3de/o3de/gem_properties.py
```diff
@@ -31,13 +33,13 @@ def update_values_in_key_list(existing_values: list, new_values: list or str, re
 
     returns updated existing value list
     """
-    if new_values:
+    if new_values is not None and new_values != []:
         new_values = new_values.split() if isinstance(new_values, str) else new_values
         existing_values.extend(new_values)
-    if remove_values:
+    if remove_values is not None and remove_values != []:
```
lumberyard-employee-dm, line 39: This is overly verbose and not pythonic and doesn't work for empty string.

This should revert to the if new_values check

scripts/o3de/o3de/gem_properties.py
```diff
@@ -85,24 +88,26 @@ def edit_gem_props(gem_path: pathlib.Path = None,
...
-    if new_documentation_url:
+    if new_documentation_url is not None:
         update_key_dict['documentation_url'] = new_documentation_url
-    if new_license:
+    if new_license is not None:
         update_key_dict['license'] = new_license
-    if new_license_url:
+    if new_license_url is not None:
         update_key_dict['license_url'] = new_license_url
+    if new_repo_uri is not None:
```
lumberyard-employee-dm, line 109: This could use python comprehension if "expression" as well

```python
update_key_dict['display_name'] = new_display if isinstance(new_display, str) else ""
```

At the very least a better check is to make sure the instance of the variable is `str` instead of checking against it not being `None`

Code/Tools/ProjectManager/Source/GemCatalog/GemInfo.cpp
```diff
@@ -100,4 +100,56 @@ namespace O3DE::ProjectManager
...
+            return GemInfo::Platform::Android;
+        }
+        else if(platformText == "iOS")
+        {
+            return GemInfo::Platform::iOS;
+        }
+        else if(platformText == "macOS")
+        {
+            return GemInfo::Platform::macOS;
+        }
```
lumberyard-employee-dm, line 125: This shouldn't be hardcoded.

Instead the existing set of Platform "defines" defined in [PlatformDefaults.h](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/PlatformId/PlatformDefaults.h) should be used.

The [PlatformHelper::GetPlatformName](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/PlatformId/PlatformDefaults.h#L109-L110) function can be used to translated an enum to a string

scripts/o3de/o3de/gem_properties.py
tkothadev: Just want to double check that this only applies to `new_values` and `remove_values` correct? `replace_values` needs the more verbose check in the event the user wishes to clear a list by supplying an empty list `[]`.

lumberyard-employee-dm: Yes, this only applies to `new_values` and `remove_values` variables

Code/Tools/ProjectManager/Source/GemCatalog/GemInfo.cpp
tkothadev: I definitely agree in principle, but I'm noticing there's a mismatch between the `Windows` value in the ProjectManager codebase as compared to the `PlatformDefaults` values of `AZ::PC` and `PlatformId::PLATFORM_WINDOWS_64 --> Win64`.

As such I'm not sure I can take action on this suggestion in the scope of this PR. I'd need to come back to this one and revise, because there are more places in the Project Manager where these OS values are hardcoded in this manner.

tkothadev: however I won't be able to use the codename for Mac `PlatformCodeNameMac`, because the string value is `Mac`, whereas in Project Manager it's `macOS`.

AMZN-alexpete: We should be consistent in our naming.  Let's find out if there's any reason why we cannot use the AzCore names or at least the code names.

lumberyard-employee-dm: Well "macOS" kind of break this situation of using the Codename.
However the term "macOS" should be capitalized to "MacOS".

tkothadev: Sounds good. So would it be fine if I note this in a github issue and address it after this PR?

AMZN-alexpete: OK I'm fine with deferring this for a later PR

scripts/o3de/o3de/gem_properties.py
```diff
@@ -88,26 +88,17 @@ def edit_gem_props(gem_path: pathlib.Path = None,
...
+    update_key_dict['display_name'] = new_display if isinstance(new_display, str) else ""
+    update_key_dict['origin'] = new_origin if isinstance(new_origin, str) else ""
+    update_key_dict['type'] = new_type if isinstance(new_type, str) else ""
+    update_key_dict['summary'] = new_summary if isinstance(new_summary, str) else ""
+    update_key_dict['icon_path'] = new_icon if isinstance(new_icon, str) else ""
+    update_key_dict['requirements'] = new_requirements if isinstance(new_requirements, str) else ""
+    update_key_dict['documentation_url'] = new_documentation_url if isinstance(new_documentation_url, str) else ""
+    update_key_dict['license'] = new_license if isinstance(new_license, str) else ""
+    update_key_dict['license_url'] = new_license_url if isinstance(new_license_url, str) else ""
+    update_key_dict['repo_uri'] = new_repo_uri if isinstance(new_repo_uri, str) else ""
```
AMZN-alexpete, line 101: would you get the same behavior and simpler code without checking if it's a string? i.e.
```
update_key_dict['display_name'] = new_display if new_display else ""
```
if `new_display` is `None` then you'll get "", and if `new_display` is "" you'll get "" and if it's a string you'll get that string

scripts/o3de/o3de/gem_properties.py
```diff
@@ -85,24 +88,17 @@ def edit_gem_props(gem_path: pathlib.Path = None,
...
+    update_key_dict['display_name'] = new_display if new_display else ""
+    update_key_dict['origin'] = new_origin if new_origin else ""
+    update_key_dict['type'] = new_type if new_type else ""
+    update_key_dict['summary'] = new_summary if new_summary else ""
+    update_key_dict['icon_path'] = new_icon if new_icon else ""
+    update_key_dict['requirements'] = new_requirements if new_requirements else ""
+    update_key_dict['documentation_url'] = new_documentation_url if new_documentation_url else ""
+    update_key_dict['license'] = new_license if new_license else ""
+    update_key_dict['license_url'] = new_license_url if new_license_url else ""
+    update_key_dict['repo_uri'] = new_repo_uri if new_repo_uri else ""
```
AMZN-alexpete, line 101: As discussed - this logic would update every value that wasn't passed into this function with an empty string so we just want the `if` block not the `else`
e.g.
```
if new_display not None:
        update_key_dict['display_name'] = new_display
```
That or an `isinstance` check.
That would let a user set the value to an empty string.

scripts/o3de/o3de/gem_properties.py
tkothadev: After speaking with AMZN-alexpete, it looks like using the comprehension suggested here creates a possible edge case from the CLI side. 

If someone tries to call the edit_gem_props function like this:
`edit_gem_props(gem_name="blah", new_display="new!")`
with only the intention to edit `gem_name` and `new_display`, then they end up unintentionally clearing every other field.

tkothadev: most likely I plan on reverting back this change

Code/Tools/ProjectManager/Source/GemCatalog/GemInfo.cpp
```diff
@@ -100,4 +100,56 @@ namespace O3DE::ProjectManager
...
+        else
+        {
+            return GemInfo::Platforms();
```
AMZN-alexpete, line 128: In a later PR we should make sure we have some fallback for unknown platforms and not use the default which could be any value.

AMZN-alexpete (approved):

Code/Tools/ProjectManager/Source/GemCatalog/GemInfo.cpp
tkothadev: will file github issue for this.

lumberyard-employee-dm (approved):

scripts/o3de/o3de/gem_properties.py
lumberyard-employee-dm: That works for me.

PR #13139, opened 2022-11-11
DPE: Support invocable ReadOnly attributes

amzn-tmryan:
**Description**
Before this PR, it was possible to encounter an assertion caused by the incorrect assumption that `ReadOnly` attributes were always `bool` values. In fact, `ReadOnly` attributes can be invocable values that reference member functions/data that will yield a `bool` value.

This PR adds support for those invocable `ReadOnly` values to the `LegacyReflectionBridge`.

**Testing**
- Verified opening the EntityInspector no longer throws an assertion for attempting to access a non-bool Value
- Added a test case to the standalone debug DPE app and verified that the property editor is enabled/disabled at runtime as appropriate


amzn-tmryan:
Added case to DPE debug standalone app that uses a ReadOnly attribute with an invocable value.

dtamkin1 (approved):

cgalvan (approved):

alexmontAmazon (approved):
looks good!

PR #13140, opened 2022-11-11
The GameLauncher now starts at the screen resolution

galibzon:
## What does this PR do?
The GameLauncher now starts at the screen resolution
and fullscreen mode as specified in the command line via the CVARs: r_width, r_height, r_fullscreen.

REMARK: It was found that Atom's BootstrapSystemComponent::Activate() is called BEFORE Legacy/System.cpp::ExecuteCommandLine(), the solution for now is to our own command line parsing for the required CVARs.

## How was this PR tested?
Validated that AutomatedTesting.GameLauncher starts at the request screen resolution specified by --r_width and --r_height when given in the command line arguments.
Also made sure it enters fullscreen mode if --r_fullscreen=true

Signed-off-by: galibzon <66021303+galibzon@users.noreply.github.com>

### DIFF
--- a/Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
+++ b/Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
@@ -16,6 +16,7 @@
 #include <AzCore/Serialization/SerializeContext.h>
 #include <AzCore/std/smart_ptr/make_shared.h>
 #include <AzCore/Utils/Utils.h>
+#include <AzCore/StringFunc/StringFunc.h>
 
 #include <AzFramework/API/ApplicationAPI.h>
 #include <AzFramework/Components/TransformComponent.h>
@@ -45,6 +46,9 @@
 #include <BootstrapSystemComponent_Traits_Platform.h>
 
 AZ_CVAR(AZ::CVarFixedString, r_default_pipeline_name, AZ_TRAIT_BOOTSTRAPSYSTEMCOMPONENT_PIPELINE_NAME, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Default Render pipeline name");
+AZ_CVAR(uint32_t, r_width, 1920, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Starting window width in pixels.");
+AZ_CVAR(uint32_t, r_height, 1080, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Starting window height in pixels.");
+AZ_CVAR(uint32_t, r_fullscreen, false, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Starting fullscreen state.");
 
 namespace AZ
 {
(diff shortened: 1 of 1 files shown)


Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
```diff
@@ -107,6 +111,72 @@ namespace AZ
...
+                    auto numValues = pCmdLine->GetNumSwitchValues(fullscreenCvarName);
+                    if (numValues > 0)
+                    {
+                        auto valueStr = pCmdLine->GetSwitchValue(fullscreenCvarName);
+                        if (AZ::StringFunc::LooksLikeBool(valueStr.c_str()))
+                        {
+                            r_fullscreen = AZ::StringFunc::ToBool(valueStr.c_str());
+                        }
+                    }
+                }
```
rgba16f, line 140: Not a blocker but it would be really nice if these command line had a functions like this to avoid all this boiler plate code to query the value.
```
     bool GetBoolValue(AZStd::string_view switchName, bool default value = false);
     uint32_t GetUInt32Value(AZStd::string_view switchName, uint32_t default value = 1);
```

rgba16f (approved):

Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
galibzon: Was thinking about that too. Will leave it for later.

galibzon:
> Does this change also work with values passed in a cfg file or a setreg value?

It's only for changes coming from the command line. CFG parsing seems to occur also afterwards. In other words, the clean solution is to fix the order of instantiation so System.cpp gets to read command line arguments + CFG file data, then later the BootstrapSystemcomponent only has to read the CVAR values.

Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
```diff
@@ -45,6 +46,9 @@
 #include <BootstrapSystemComponent_Traits_Platform.h>
 
 AZ_CVAR(AZ::CVarFixedString, r_default_pipeline_name, AZ_TRAIT_BOOTSTRAPSYSTEMCOMPONENT_PIPELINE_NAME, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Default Render pipeline name");
+AZ_CVAR(uint32_t, r_width, 1920, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Starting window width in pixels.");
+AZ_CVAR(uint32_t, r_height, 1080, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Starting window height in pixels.");
+AZ_CVAR(uint32_t, r_fullscreen, false, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Starting fullscreen state.");
```
moraaar, line 51: It seems more appropriate to use setting registry for these so they can be put in .setreg files. And more specifically platform dependent setting registry (read with `AZ::SettingsRegistryMergeUtils::PlatformGet`), so different values per platform could be specified.

These CVars will appear on the console and could be modified at runtime, but they won't make any difference.

Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
```diff
@@ -119,8 +189,14 @@ namespace AZ
...
 
-                    m_nativeWindow = AZStd::make_unique<AzFramework::NativeWindow>(projectTitle.c_str(), AzFramework::WindowGeometry(0, 0, 1920, 1080));
+                    // It is important to call this before using r_width, r_height or r_fullscreen
+                    // because at the moment this system component initializes before Legacy System.cpp gets to parse
+                    // command line arguments into cvars.
+                    UpdateCVarsFromCommandLine();
+
+                    m_nativeWindow = AZStd::make_unique<AzFramework::NativeWindow>(projectTitle.c_str(), AzFramework::WindowGeometry(0, 0, r_width, r_height));
                     AZ_Assert(m_nativeWindow, "Failed to create the game window\n");
+                    m_nativeWindow->SetFullScreenState(r_fullscreen);
```
moraaar, line 199: how will affect calling `SetFullScreenState` on mobile platforms? Is this suppose to be called on host platforms only?

moraaar:
Is there some way to do resolution scaling at the moment? If yes, how does code change works with it?

Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
lumberyard-employee-dm: Actually `PlatformGet` should not be used.
That was added just for porting over the bootstrap.cfg settings.

The recommended way to specify platform dependent settings registry settings is by adding a `Platform/<platform-name>` folder underneath the `Registry` directory that is being used (https://www.o3de.org/docs/user-guide/settings/developer-documentation/#merge-settings-files)

moraaar: Thanks for the clarification @lumberyard-employee-dm. I wasn't away of that recommendation.

galibzon:
> Is there some way to do resolution scaling at the moment? If yes, how does this code work in conjunction with it?

This is a quick workaround (aka Manager's code) to unblock performance collection on Windows & Linux. I opened a ticket to address fullscreen issues: https://github.com/o3de/o3de/issues/13147

antonmic (approved):

Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
galibzon: I understand your concern, and the problem we have is that there's no unified way at the moment. There are things in CVars, others in Registry settings, others in CFG files, etc. I'll leave it to the engineers to eventually clean that up. For now, this PR enables screen resolution setup while minimizing verbosity.

Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
galibzon: See my comment below.

Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
```diff
@@ -119,8 +189,14 @@ namespace AZ
 
                     auto projectTitle = AZ::Utils::GetProjectDisplayName();
 
-                    m_nativeWindow = AZStd::make_unique<AzFramework::NativeWindow>(projectTitle.c_str(), AzFramework::WindowGeometry(0, 0, 1920, 1080));
+                    // It is important to call this before using r_width, r_height or r_fullscreen
+                    // because at the moment this system component initializes before Legacy System.cpp gets to parse
+                    // command line arguments into cvars.
+                    UpdateCVarsFromCommandLine();
```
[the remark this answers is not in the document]
galibzon: Let us know when it gets fixed and we'll make changes to this code. Yes, I tried that API and it didn't work.

PR #13142, opened 2022-11-11
DPE: Add default transient in-memory expansion state saving

amzn-tmryan:
**Description**
Before this PR, only persistent expansion state storage existed in the DPE. This meant that editors needed to create and set a GUID state key if they wanted the DPE to remember expansion states and there was no way to activate it only for the time a document was open (transient in-memory storage only).

This PR makes the DPE's default expansion state saving behavior the non-persistent in-memory only (wiped after the document is closed) behavior. Setting a save state key will continue to work as before where settings are saved to disk when a document is closed.

**Testing**
- Tested in the AssetEditor by disabling the current save to disk behavior
- Tested in the standalone debug DPE app by collapsing and expanding rows in various configs as well as switching between the available adapters

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditorSettings.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditorSettings.cpp
@@ -24,6 +24,7 @@ namespace AzToolsFramework
             .ReplaceExtension(SettingsRegistrar::SettingsRegistryFileExt);
 
         m_wereSettingsLoaded = LoadExpanderStates();
+        m_shouldSettingsPersist = true;
     }
 
     DocumentPropertyEditorSettings::~DocumentPropertyEditorSettings()
@@ -85,7 +86,7 @@ namespace AzToolsFramework
             }
         }
 
-        if (!m_expandedElementStates.empty())
+        if (m_shouldSettingsPersist && !m_expandedElementStates.empty())
         {
             SaveExpanderStates();
         }
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
@@ -1240,8 +1240,10 @@ namespace AzToolsFramework
             });
         m_adapter->ConnectMessageHandler(m_domMessageHandler);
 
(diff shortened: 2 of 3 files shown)


Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditorSettings.cpp
```diff
@@ -12,6 +12,11 @@
 
 namespace AzToolsFramework
 {
+    DocumentPropertyEditorSettings::DocumentPropertyEditorSettings()
+    {
+        m_shouldSaveSettings = false;
```
alexmontAmazon, line 17: this is set to false here, and to true inline in the header declaration. By convention, we use the header declaration

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditorSettings.cpp
```diff
@@ -85,7 +90,7 @@ namespace AzToolsFramework
             }
         }
 
-        if (!m_expandedElementStates.empty())
+        if (m_shouldSaveSettings && !m_expandedElementStates.empty())
```
alexmontAmazon, line 93: what can currently set this to true?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditorSettings.cpp
amzn-tmryan: That makes sense. I'll default it to false and set to true in the other ctor, then we can still have our `= default` ctor for this class.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditorSettings.cpp
amzn-tmryan: Right now this will only be true if the object was initialized with the data needed to store things in a unique registry.

It was true by default, but I'm going to swap it around so that the value is set to true in the appropriate ctor.

amzn-tmryan:
Set the "should we store this stuff?" flag to false by default and added some comments for the two ctors.

dtamkin1 (approved):

cgalvan (approved):

alexmontAmazon (approved):
Good for now, but this will also change with our mapToSource saving, etc

phuchau1989 (approved):

phuchau1989 (approved):

PR #13145, opened 2022-11-11
Fixed Physics Material RTTI name

moraaar:

## What does this PR do?

Fixed Physics Material RTTI name. It's not expected to cause any issues since the type uuid of the class used is the same.

## How was this PR tested?

Run editor and test a physics collider with different materials.


greerdv (approved):

hultonha (approved):
Thanks for fixing this @moraaar and again good spot @rgba16f! 😄

SergeyAMZN (approved):

PR #13148, opened 2022-11-11
connect collider and rigid body components to visibility bounds

### LINKED ISSUE #2466: PhysX Force Region disappears depending on camera orientation
PhysX Force Region disappears depending on camera orientation

**To Reproduce**
ForceFields_lq

**Expected behavior**
Force field is always visible and just gets skipped to render when the debug visualization is fully out of the view frustum

greerdv:

## What does this PR do?
Connects PhysX Collider and PhysX Rigid Body components to visibility bounds bus. Fixes #2466

## How was this PR tested?
Tested in editor, with variety of different collider geometries, position and rotation offsets, transform rotation, translation and scale, with and without non-uniform scale, single and multiple colliders, with and without rigid body component.

### DIFF
--- a/Gems/PhysX/Code/Source/EditorColliderComponent.cpp
+++ b/Gems/PhysX/Code/Source/EditorColliderComponent.cpp
@@ -425,6 +425,7 @@ namespace PhysX
         AZ::Render::MeshComponentNotificationBus::Handler::BusConnect(GetEntityId());
         EditorColliderComponentRequestBus::Handler::BusConnect(AZ::EntityComponentIdPair(GetEntityId(), GetId()));
         EditorColliderValidationRequestBus::Handler::BusConnect(GetEntityId());
+        AzFramework::BoundsRequestBus::Handler::BusConnect(GetEntityId());
         m_nonUniformScaleChangedHandler = AZ::NonUniformScaleChangedEvent::Handler(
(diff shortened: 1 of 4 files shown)


Gems/PhysX/Code/Source/EditorColliderComponent.cpp
```diff
@@ -1570,6 +1572,22 @@ namespace PhysX
         m_shapeConfiguration.m_cylinder.m_configuration = Utils::CreatePxCookedMeshConfiguration(samplePoints, scale).value();
     }
 
+    AZ::Aabb EditorColliderComponent::GetWorldBounds()
+    {
+        return GetAabb();
```
SergeyAMZN, line 1577: Do we need to have `IsValid()` check for the world bounds like we do in `GetLocalBounds` or the visibility system is OK about dealing with invalid AABBs?

SergeyAMZN (approved):

hultonha (approved):
Looks great! Thanks for adding this @greerdv 👍

moraaar (approved):

Gems/PhysX/Code/Source/EditorColliderComponent.cpp
greerdv: The visibility system is fine with invalid AABBs. The reason I put the IsValid check in the other place was just in case anything weird could happen if you inverse transform a null AABB.

PR #13110, opened 2022-11-10
Support for shape component translation offsets in PhysX gem

greerdv:
## What does this PR do?
Adds support for shape translation offsets for box, capsule and sphere shape components when used with the PhysX shape collider component.

## How was this PR tested?
This PR adds a lot of unit tests. Ran all the PhysX and PhysX editor unit tests and performed manual testing in the editor.

PR #13149, opened 2022-11-11
Physx collider refactor

greerdv:
## What does this PR do?
Some refactoring following up on #13110

## How was this PR tested?
Ran unit tests.


Gems/PhysX/Code/Source/Utils.cpp
```diff
@@ -1177,17 +1177,11 @@ namespace PhysX
             }
         }
 
-        AZ::Vector3 GetTransformScale(AZ::EntityId entityId)
+        float GetTransformScale(AZ::EntityId entityId)
```
moraaar, line 1180: Shouldn't this function be `GetUniformTransformScale` now to be clear?

moraaar (approved):
LGTM. Just 1 question. Approved.

Gems/PhysX/Code/Source/Utils.cpp
greerdv: Transform scale can't be anything but uniform.

Gems/PhysX/Code/Editor/DebugDraw.cpp
```diff
@@ -499,26 +499,17 @@ namespace PhysX
...
-
-            if (forceUniformScaling)
-            {
-                scaledBoxParameters *= resultantScale.GetMaxElement();
-            }
-            else
-            {
-                scaledBoxParameters *= resultantScale;
-            }
+            AZ::Vector3 scaledBoxParameters = boxShapeConfig.m_dimensions * 0.5f * resultantScale;
```
hultonha, line 512: uber nit: This could be `const`

Gems/PhysX/Code/Source/EditorShapeColliderComponent.cpp
```diff
@@ -440,25 +440,24 @@ namespace PhysX
 
         SetShapeConfig(ShapeType::Capsule, capsuleShapeConfig);
 
-        m_shapeConfigs.back()->m_scale = scale;
-        const float scalarScale = scale.GetMaxElement();
-        m_geometryCache.m_radius = scalarScale * capsuleShapeConfig.m_radius;
-        m_geometryCache.m_height = scalarScale * capsuleShapeConfig.m_height;
+        m_shapeConfigs.back()->m_scale = AZ::Vector3(scale);
```
hultonha, line 443: question: Should we check `shapeConfigs` isn't empty just in case?

Gems/PhysX/Code/Source/EditorShapeColliderComponent.cpp
```diff
@@ -440,25 +440,24 @@ namespace PhysX
...
     {
         float radius = 0.0f;
         LmbrCentral::SphereShapeComponentRequestsBus::EventResult(radius, GetEntityId(),
             &LmbrCentral::SphereShapeComponentRequests::GetRadius);
 
         SetShapeConfig(ShapeType::Sphere, Physics::SphereShapeConfiguration(radius));
         
-        m_shapeConfigs.back()->m_scale = scale;
-        m_geometryCache.m_radius = scale.GetMaxElement() * radius;
+        m_shapeConfigs.back()->m_scale = AZ::Vector3(scale);
```
hultonha, line 456: question: Same here, is the `.back()` call always safe?

hultonha (approved):
LGTM! Just a couple of minor questions

Gems/PhysX/Code/Editor/DebugDraw.cpp
greerdv: Done.

Gems/PhysX/Code/Source/EditorShapeColliderComponent.cpp
greerdv: This was always happening immediately after a call to SetShapeConfig, so I've folded the scale change into that function. Inside there, it's guaranteed not to be empty.

Gems/PhysX/Code/Source/EditorShapeColliderComponent.cpp
greerdv: See other comment.

Gems/PhysX/Code/Source/EditorShapeColliderComponent.cpp
hultonha: Ah nice, sounds good thanks!

SergeyAMZN (approved):

hultonha (approved):

Code/Framework/AzFramework/AzFramework/Physics/ShapeConfiguration.cpp
```diff
@@ -72,8 +77,9 @@ namespace Physics
         }
     }
 
-    SphereShapeConfiguration::SphereShapeConfiguration(float radius)
+    SphereShapeConfiguration::SphereShapeConfiguration(float radius, const AZ::Vector3& scale)
         : m_radius(radius)
+        , ShapeConfiguration(scale)
```
moraaar, line 82: [nit] normally the parent class input goes first because that's the order it'll happen. Having it at the end looks like the order it'll happen is different, so it's preferred to put it first.

Same in the other classes.

moraaar (approved):
Change looks good to me. One minor comment.

PR #13151, opened 2022-11-11
Remove rapidjson default type customizations

lawsonamzn:
## What does this PR do?
The previous attempt to wrangle the memory usage of rapidjson worked but was incompatible with all 3rd party plugins that also used rapidjson internally or shipped with their own custom version of rapidjson, becuase it modified the actual default types for rapidjson::Document, and rapidjson::Value.  While this works, the vast majority of all existing software that interfaces with rapidjson does not do that, causing all sorts of weird errors to crop up when a piece of software (for example Cesium) uses 'vanilla' rapidjson internally links to or includes headers from AzCore when it redefines basic rapidjson types such as rapidjson::Document.

Note that rapidjson::Pointer code (from the actual rapidjson library) has several bugs in it that makes it incompatible with non-default rapidjson::Document and rapidjson::Value, also, so even if we were to, for example, define our own type instead of modify the default, this would still not function.   A comment is included that explains this for future maintainers.

This change reverts the previous change to rapidjson::Document and removes the special rapidjson_ly namespace.  Instead, it works at the allocator level, so that the types remain the default types, but memory is shunted into a special rapidjson allocator so that its still tracked by the memory system.

## How was this PR tested?
Turning on the Cesium addon and ensuring it compiles without modification to any of the rapidjson stuff.  
Using Visual Studio's "Heap Capture" memory profiler to check for memory leaks or memory accumulation before and after opening levels, after performing large numbers of undo operations, after undoing them and removing the stack, etc.

With this change memory consumption is even better than before, better than even using the default non-tracked rapidjson allocators.

Note that as part of this change, several cases where rapidjson was accumulating memory were replaced with swaps and clears from default, and some places where the wrong allocator was being used were also repaired.

## other notes

The structure of rapidjson.h was modified to be forward-declaring and a companion rapidjson.cpp was added which is the only place where the system allocator header is pulled in.  This allows the inclusion of rapidjson.h very lightweight (just the rapidjson stuff itself and azcore/base.h), reducing include chain complexity significantly.

### DIFF
--- a/Code/Framework/AzCore/AzCore/JSON/rapidjson.h
+++ b/Code/Framework/AzCore/AzCore/JSON/rapidjson.h
@@ -8,52 +8,97 @@
 
 #pragma once
 
-#include <AzCore/Memory/SystemAllocator.h>
+#include <AzCore/base.h>
 
-// Skip the error as we are including the recommended way.
-#define RAPIDJSON_SKIP_AZCORE_ERROR
-
-#if defined(AZ_ENABLE_TRACING)
-#   define RAPIDJSON_ASSERT(x) AZ_Assert(x, "Assert[rapidjson]: " #x)
-#else
-#   define RAPIDJSON_ASSERT(x) (void)(0)
+#if defined(RAPIDJSON_RAPIDJSON_H_)
+// if this happens, someone has alrady included the default rapidjson.h already, which is a bad idea as it won't
+// include any customizations such as memory management.
+#error vanilla rapidjson was included before including <AzCore/JSON/rapidjson.h>
 #endif
-#define RAPIDJSON_STATIC_ASSERT(x) static_assert(x, "Assert[rapidjson]: " #x)
-#define RAPIDJSON_HAS_CXX11_TYPETRAITS 1
-#define RAPIDJSON_HAS_CXX11_RVALUE_REFS 1
 
-namespace rapidjson_ly_internal
-{
-    template<typename T>
-    void Delete(T* instance)
+// Only override RAPIDJSON_ customization macros which are not already defined.
+// This allows the user to pre-define these before including O3DE if they want to use default values
+// or their own values.
+
+#if !defined(RAPIDJSON_ASSERT)
+#   if defined(AZ_ENABLE_TRACING)
+#       define RAPIDJSON_ASSERT(x) AZ_Assert(x, "Assert[rapidjson]: " #x)
+#   else
+#       define RAPIDJSON_ASSERT(x) (void)(0)
+#   endif
+#endif
+
+#if !defined(RAPIDJSON_STATIC_ASSERT)
+   #define RAPIDJSON_STATIC_ASSERT(x) static_assert(x, "Assert[rapidjson]: " #x)
+#endif
+
+#if !defined(RAPIDJSON_HAS_CXX11_TYPETRAITS)
+#   define RAPIDJSON_HAS_CXX11_TYPETRAITS 1
+#endif
+
+#if !defined(RAPIDJSON_HAS_CXX11_RVALUE_REFS)
+#   define RAPIDJSON_HAS_CXX11_RVALUE_REFS 1
+#endif
+
+// note for future maintainers:
+// RapidJSON allows you to customize the default allocator type for documents by overriding with RAPIDJSON_DEFAULT_ALLOCATOR.
+// However, customizing the default type of rapidjson::Document and rapidjson::Value in that way will cause incompatibilities
+// with external software libraries that use unmodified RapidJSON.   It is not thus recommended
+// to actually set the value of RAPIDJSON_DEFAULT_ALLOCATOR or redefine what a rapidjson::Document itself is.
+// The other way to get a 'custom' allocator installed in RapidJSON is to make your own Document/Value/Pointer type, since it's templated
+// on allocator type - for example, implement your own allocator and do
+// using MyDocument = rapidjson::GenericDocument<rapidjson::UTF8<char>, MyAllocator, MyAllocator> and then use MyDocument in your code
+// instead of RapidJSON::Document.  Note GenericDocument<> customized must use customized GenericValue<> and GenericPointer<>
+// or else you haven't acutally plugged into all the places where memory is allocated and will lose some.
+// 
+// Unfortunately, rapidjson::Pointer can't deal with these customized values or documents due to a bug in the code where functions
+// like Set/Get/Create use the default template parameter instead of deducing it:
+// ie, they are set up like somefunction<T>(GenericValue<char> val) which will only accept default GenericValue<char, CrtAllocator>.
+// correct form: somefunction<T>(GenericValue<char, T::AllocatorType> val, which would allow MyValue<char, MyAllocator> to be passed in.
+// thus, it is not recommended to use customized documents, values, or pointers either, (until that bug is fixed).
+
+// this leaves just one avenue for customization of memory of Rapidjson - overriding the macros RAPIDJSON_xxxx (new,delete, malloc, realloc..)
+
+// forward all allocations to our own allocators.  But preferrably without acutally having to drag our allocators into this header:
+
+// note that we cannot partially define allocation forwarding - for example, it is not okay to just override new, but not delete.
+// So either we define them all, or we define none.  If a user has defined any of them, we don't define any of them:
+// Users can manually switch the entire system off by setting USING_AZCORE_RAPIDJSON_ALLOCATION_OVERRIDES 0
+
+#if !defined(USING_AZCORE_RAPIDJSON_ALLOCATION_OVERRIDES)
+#   if  !defined(RAPIDJSON_MALLOC) && !defined(RAPIDJSON_REALLOC) && !defined(RAPIDJSON_FREE) && !defined(RAPIDJSON_NEW) && !defined(RAPIDJSON_DELETE)
+#       define USING_AZCORE_RAPIDJSON_ALLOCATION_OVERRIDES 1
+#   else
+#       define USING_AZCORE_RAPIDJSON_ALLOCATION_OVERRIDES 0
+#   endif
+#endif
+
+#if USING_AZCORE_RAPIDJSON_ALLOCATION_OVERRIDES
+
+    namespace AZ::JSON
     {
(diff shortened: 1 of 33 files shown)


Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndo.cpp
```diff
@@ -121,7 +121,7 @@ namespace AzToolsFramework
                 {
                     // Create a copy of the DOM of the end state so that it shares the lifecycle of the cached DOM.
                     PrefabDom endStateCopy;
-                    endStateCopy.CopyFrom(endState, cachedDom->get().GetAllocator());
+                    endStateCopy.CopyFrom(endState, endStateCopy.GetAllocator());
```
lawsonamzn, line 124: (note from author) I'm actually curious as to why this was that way in the first place, the comment is a bit confusing.
In general, the allocator you want to use is the allocator of the DOM itself, not someone else...

Code/Framework/AzCore/AzCore/JSON/rapidjson.h
```diff
@@ -22,38 +19,48 @@
...
+// instead of RapidJSON::Document.  Note GenericDocument<> customized must use customized GenericValue<> and GenericPointer<>
+// or else you haven't acutally plugged into all the places where memory is allocated and will lose some.
+// 
+// Unfortunately, rapidjson::Pointer can't deal with these customized values or documents due to a bug in the code where functions
+// like Set/Get/Create use the default template parameter instead of deducing it:
+// ie, they are set up like somefunction<T>(GenericValue<char> val) which will only accept default GenericValue<char, CrtAllocator>.
+// correct form: somefunction<T>(GenericValue<char, T::AllocatorType> val, which would allow MyValue<char, MyAllocator> to be passed in.
+// thus, it is not recommended to use customized documents, values, or pointers either, (until that bug is fixed).
+
+// this leaves just one avenue for customization of memory of Rapidjson - overriding the macros RAPIDJSON_xxxx (new,delete, malloc, realloc..)
```
lumberyard-employee-dm, line 39: Should we also add in ifdef checks to make sure the macros aren't already defined before defining them?

So uses can technically avoid defined there own allocators through compiler definitions instead?

lumberyard-employee-dm (approved):

Code/Framework/AzCore/AzCore/JSON/rapidjson.h
lawsonamzn: I can do that, yes.  Its a good idea, since that's what rapidjson already does also.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/Instance.cpp
```diff
@@ -921,6 +921,7 @@ namespace AzToolsFramework
 
         void Instance::SetCachedInstanceDom(PrefabDomValueConstReference instanceDom)
         {
+            m_cachedInstanceDom = PrefabDom(); // force a flush of memory by clearing first.
             m_cachedInstanceDom.CopyFrom(instanceDom->get(), m_cachedInstanceDom.GetAllocator());
```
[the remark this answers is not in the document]
lawsonamzn: No, since ~GenericValue() destroys the baseclass, but not the derived class (GenericDocument) which is what owns the allocator.

Code/Framework/AzCore/AzCore/JSON/rapidjson.h
```diff
@@ -22,38 +19,48 @@
 #define RAPIDJSON_HAS_CXX11_TYPETRAITS 1
 #define RAPIDJSON_HAS_CXX11_RVALUE_REFS 1
 
-namespace rapidjson_ly_internal
+// note for future maintainers:
+// RapidJSON allows you to customize the default allocator type for documents by overriding with RAPIDJSON_DEFAULT_ALLOCATOR.
+// However, customizing the default type of rapidjson::Document and rapidjson::Value in that way will cause incompatibilities
+// with external software libraries that use unmodified RapidJSON.   It is not thus recommended
+// to actually the value of RAPIDJSON_DEFAULT_ALLOCATOR or redefine what a rapidjson::Document itself is.
```
AMZN-puvvadar, line 26: NP: Missing a verb between actually and the
```suggestion
// to actually set the value of RAPIDJSON_DEFAULT_ALLOCATOR or redefine what a rapidjson::Document itself is.
```

AMZN-puvvadar (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndo.cpp
lawsonamzn: I'm not sure what your code above refers to for cachedOwningInstanceDom, or entityDom, neither of those are present in the Capture function as of yet?

lawsonamzn: Do you perhaps mean this instead?
```
                    PrefabDomValue endStateCopy(endState, cachedDom->get().GetAllocator());
                    // Update the cached instance DOM corresponding to the entity so that the same modified entity isn't reloaded again.
                    PrefabDomPath entityPathInDom(entityAliasPath.c_str());
                    entityPathInDom.Set(cachedDom->get(), endStateCopy.Move());
```

forkercat: oh sorry I was referring to some code elsewhere. Yeah, this is what I mean.

lawsonamzn: I think thats worse, perhaps I don't understand. Looking at the code, here's my analysis
```
// the following line makes a deep copy of endState into endStateCopy and allocates memory to do so from the allocator
// from cachedDom.  cachedDom's allocator thus would grow by the size of endState's values.
PrefabDomValue endStateCopy(endState, cachedDom->get().GetAllocator());

PrefabDomPath entityPathInDom(entityAliasPath.c_str());

// important note - rapidjson Pointer.Set(x, y) does not have a move form that takes &&y, only &y, and const &y
// so this invokes GenericValue.Set(non-const GenericDocument&, non-const Value&), not value&& which
// removes the rvalue-reference from value.
// which itself invokes Pointer::Create(document) = value.
// The call to Create ensures that all parent values exist, so it uses cachedDom's allocator to create them if necessary.
// cachedDom's memory will grow by that amount.
// the assignment operator is then called on the new child, without a rvalue reference.
// this = operator then just deep copies the value
 entityPathInDom.Set(cachedDom->get(), endStateCopy.Move());

 // the end result is that cachedDom's allocator grows by the size of endstate,
 // plus the size of the elements required to create the tree of parents to match the pointer
// plus the size ofendState again.
```

Code/Framework/AzCore/AzCore/JSON/rapidjson.h
lawsonamzn: Should we use our own macros or adopt rapidjsons ones?  Hmm.

lumberyard-employee-dm: I vote for using the rapidjson macros.

The benefits are that there is only one set of macros that users need to have knowledge about and they could lookup that information in the rapidjson documentation.
Furthermore it helps avoid any discovery issues with determining that we supply macros for overriding the allocator.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndo.cpp
lawsonamzn: Ok, so it turns out GenericValue::operator=(GenericValue& other) adopts the pointers of other (without destroying other).
This means the only option here that mkes sense is to copy into the final dom.
This can be achieved by the one liner
```
 entityPathInDom.Set(cachedDom->get(), endState);
 ```
because endState is already a const& and thus will already trigger a copy into the dom using the dom's allocator

forkercat: discussed offline. We can avoid doing the copy by ourselves and let Set() do it. But in this case we need to be careful about the type of endState. It needs to be const& (not non-const &) to trigger the copy version of Set().

forkercat (approved):

lawsonamzn:
Progress update.  Need to resolve conflicts and also find a different approach as garbage collecting directly during propogation causes tests to invalidate.  Note that this is because the tests retain pointers into the actual document (as in, rapidjson document) DOM values, and expect those pointers to still be valid after the document has been modified by the test - which itself is iffy, (especially if you modify a vector and it resizes) but in these specific cases, they are just modifying values, which is not expected to invalidate pointers to those values.

I'll move the garbage collection to be after the fact and explicit.

lawsonamzn:
I have updated it to work with latest development, via rebase (too many changes otherwise).

Not much is different, the major difference is the garbage collection of templates is now outside the instance update loop and is its own thing.  I found several other places, too, where we "accumulate" memory in rapidjson and clear it before we set it.  The net result is according to MSVC heap profiler, we have no more bytes unaccounted for in the prefab system after load, we have exactly as much memory as you'd expect for the prefab templates, and their instance dom cache, etc.   As you start doing operations the undo stack starts accumulating, but unloading levels, erasing the undo stack, etc etc, causes memory to drop back to expected (at least in terms of the prefab system).

The final result here is that a 10k+ entity level that originally cost 2.5g (before any changes) and then 0.8g (after the initial set of rapidjson changes) now costs 0.5g and of that 0.5g only 0.1g is attributed to the prefab system, the rest is render stuff, tracking stuff, physics, octrees growing, entity instance memory, etc.

lawsonamzn:
This failed due to an unrelated problem already in PR.
Will have to merge it in and try again.

lawsonamzn:
The thing blocking this has been checked in, I can go back to this now, but it looks like I got a bunch of merges to do :)

lumberyard-employee-dm (approved):

PR #13152, opened 2022-11-11
[Terrain] Separated out the paintbrush logic.

### LINKED ISSUE #13089: Gradient Paint Brush: Allow configuring color of Viewport brush gizmos
The Viewport brush gizmo is red by default. When working with terrain that had a red macro material applied, the gizmo was very hard to see. 

We should allow configuring the color of the gizmo for working with different color terrain. Perhaps just a few different choices to swap between with a CVar.

mbalfour-amzn:
## What does this PR do?

This PR splits the core paintbrush logic apart from the paintbrush manipulator logic. This will make it significantly easier to unit test the underlying paint logic without requiring as much scaffolding around it to set up the test conditions.

A couple of other changes:
* Added CVars to control the colors of the inner and outer circle on the manipulator. This way, if you're painting on a red terrain, you don't have to be stuck with an "invisible" red paintbrush manipulator. ;) 
* Fixed the logic with paint strokes that go off the edge of a surface and back onto it. It used to draw a line connecting the exit and entry points together. Now it correctly handles the discontinuities.

By splitting the PaintBrush code out like this, the eventual goal is to move the core PaintBrush code down from AzToolsFramework to AzFramework so that it can be used at runtime as well as tool time. I'm saving this for a separate PR though because there are some tricky bits to sort out around how to manage the color picker configuration in its Edit Context, since the color picker configuration structures are defined up in AzToolsFramework.

This will close #13089 .

## How was this PR tested?

Manually verified that painting still worked the same as before.


Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.h
```diff
@@ -10,35 +10,25 @@
...
-    //! 3. The paintbrush responds with the specific painted values for each of those positions based on the brush shape and settings.
-    //! This back-and-forth is needed so that we can keep a clean separation between the paintbrush and the listener. The paintbrush
-    //! doesn't have knowledge of which points in the world (or at which resolution) the listener needs, and the listener doesn't have
-    //! knowledge of the exact shape, size, and hardness of the paintbrush. The separation allows the paintbrush manipulator to add
-    //! features over time without having to change the listeners, and it enables more systems to use the paintbrush without requiring
-    //! specialized logic inside the paintbrush itself.
+    //! PaintBrushManipulator contains the core logic for managing the paintbrush in the Editor.
+    //! It uses the global paintbrush settings for painting and provides logic for converting mouse events into paintbrush actions.
+    //! It also draws the paintbrush circles in the viewport and provides the UX for switching paint modes.
+    //! The PaintBrushManipulator relies a PaintBrush instance to do the actual painting logic and on a component-specific
```
cgalvan, line 28:
```suggestion
    //! The PaintBrushManipulator relies on a PaintBrush instance to do the actual painting logic and on a component-specific
```

cgalvan (approved):

lsemp3d (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/PaintBrush/PaintBrush.h
```diff
@@ -0,0 +1,136 @@
...
+    {
+
+    public:
+        AZ_RTTI(PaintBrush, "{D003175F-12BF-4E7E-BD51-2F4B010C9B5E}");
+        AZ_CLASS_ALLOCATOR(PaintBrush, AZ::SystemAllocator, 0);
+
+        PaintBrush() = delete;
+        PaintBrush(const PaintBrush&) = delete;
+        PaintBrush& operator=(const PaintBrush&) = delete;
+        explicit PaintBrush(const AZ::EntityComponentIdPair& entityComponentIdPair);
```
hultonha, line 42: question: Likely outside the scope of this PR but I wonder if it'd be possible to eliminate `entityComponentIdPair` from the core PaintBrush too so we could use it more easily in other viewports (e.g. The animation editor in future for painting vertex weights for example). We actually get around this with core manipulators by having an optional API to add an EntityComponentIdPair when needed which isn't required to be called. Not critical but could be something useful in future

hultonha (approved):
Nice change! 😄 Great to see 🙂

Code/Framework/AzToolsFramework/AzToolsFramework/PaintBrush/PaintBrush.h
mbalfour-amzn: The entity component pair shouldn't prevent it from being used in other viewports, should it? It's just a mechanism for having a unique key for talking back and forth between the paintbrush and the component it's attached to. There are definitely other ways of generating a key and passing it around, but right now, since components are the only things consuming the paintbrush, the entity / component ID pair has been a pretty transparent way to connect the two. I'd prefer not to move to a more opaque key unless it's necessary.

PR #12173, opened 2022-09-23
Move TBN construction to pixel shader

jiaweig-amzn:
Signed-off-by: jiaweig <51759646+jiaweig-amzn@users.noreply.github.com>

## What does this PR do?

Move TBN construction to pixel shader. This will save LDS usage and interpolant overheads at big scenes where vertices are rendered into a few pixels.

## How was this PR tested?
Tested on ASV. 3 failures are not related to this PR, will fix them separately.
Capture

PR #13154, opened 2022-11-11
Temporarily bring bitangent back in vertex shader to fix the screenshot test

jiaweig-amzn:

## What does this PR do?

The calculated bitangent seems to be inconsistent from the vertex stream. Bring the bitangent back before further investigation to determine the correctness of the bitangent.

## How was this PR tested?

The test should pass with little difference.
eye

And for reference:
Before and after this change:
https://github.com/o3de/o3de/pull/12173
Calculated TBN in vertex shader:
BeforeTBNChange
Calculated TBN in pixel shader:
AfterTBNChange

### DIFF
--- a/Gems/Atom/Feature/Common/Assets/Shaders/Materials/Eye/Eye_PixelGeometryEval.azsli
+++ b/Gems/Atom/Feature/Common/Assets/Shaders/Materials/Eye/Eye_PixelGeometryEval.azsli
@@ -81,6 +81,7 @@ PixelGeometryData EvaluatePixelGeometry_Eye(VsOutput IN, bool isFrontFace)
 
     float3 vertexNormal, vertexTangent, vertexBitangent;
     ConstructTBN(IN.normal, IN.tangent, objectToWorld, objectToWorldIT, vertexNormal, vertexTangent, vertexBitangent);
+    vertexBitangent = normalize(mul(objectToWorld, float4(IN.bitangent, 0.0)).xyz);
 
     return EvaluatePixelGeometry_Eye(
         IN.worldPosition,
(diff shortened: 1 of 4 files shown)


antonmic (approved):

Gems/Atom/Feature/Common/Assets/Shaders/Materials/Eye/Eye_PixelGeometryEval.azsli
```diff
@@ -81,6 +81,7 @@ PixelGeometryData EvaluatePixelGeometry_Eye(VsOutput IN, bool isFrontFace)
 
     float3 vertexNormal, vertexTangent, vertexBitangent;
     ConstructTBN(IN.normal, IN.tangent, objectToWorld, objectToWorldIT, vertexNormal, vertexTangent, vertexBitangent);
+    vertexBitangent = normalize(mul(objectToWorld, float4(IN.bitangent, 0.0)).xyz);
```
santorac, line 84: I think we should have a comment here to explain why we're doing this.

santorac: Also we should have an Issue to investigate further and remove the bitangent again if possible, or provide an explanation about why the bitangent is necessary.

jiaweig-amzn: Right. That's what I plan to do. I'll add a comment to explain.

jiaweig-amzn: https://github.com/o3de/o3de/issues/13159

santorac (approved):

santipaprika (approved):

PR #13158, opened 2022-11-11
Add AttachSubTree and DetachSubTree functions for AZ::Dom::DomPrefixTree

srikappa-amzn:
Signed-off-by: srikappa-amzn <82230713+srikappa-amzn@users.noreply.github.com>

## What does this PR do?
This PR adds 2 new public functions to attach and detach sub trees within an AZ::Dom::PrefixTree. It also adds a few private helper functions and a constructor to achieve this.

These functions are required in order to support several override operations for Prefabs. Once these changes are in, they will be integrated immediately with the 'prefab revert overrides' feature. The draft PR for that can be found here along with a working video of the feature: https://github.com/aws-lumberyard-dev/o3de/pull/468

## How was this PR tested?

Tested manually to accomplish a prefab feature. Once this direction is agreed upon, I'll also add unit tests to this same PR.

### DIFF
--- a/Code/Framework/AzCore/AzCore/DOM/DomPrefixTree.inl
+++ b/Code/Framework/AzCore/AzCore/DOM/DomPrefixTree.inl
@@ -48,9 +48,66 @@ namespace AZ::Dom
     }
 
     template<class T>
+    DomPrefixTree<T>::DomPrefixTree(Node&& node)
+    {
+        m_rootNode = AZStd::move(node);
+    }
+
+    template<class T>
+    bool DomPrefixTree<T>::DomPrefixTree::Node::IsEmpty() const
+    {
+        return (m_values.empty() && m_data);
+    }
+
+    template<class T>
     auto DomPrefixTree<T>::GetNodeForPath(const Path& path) -> Node*
     {
         return const_cast<Node*>(AZStd::as_const(*this).GetNodeForPath(path));
+    }
+
+    template<class T>
+    auto DomPrefixTree<T>::DetachNodeAtPath(const Path& path) -> Node
+    {
+        Node* node = &m_rootNode;
+        const size_t entriesToIterate = path.Size() - 1;
+        for (size_t i = 0; i < entriesToIterate; ++i)
+        {
+            const PathEntry& entry = path[i];
+            auto nodeIt = node->m_values.find(entry);
+            if (nodeIt == node->m_values.end())
+            {
+                return Node();
+            }
+            node = &nodeIt->second;
+        }
+        auto nodeIt = node->m_values.find(path[path.Size() - 1]);
+        if (nodeIt != node->m_values.end())
+        {
+            Node detachedNode = AZStd::move(nodeIt->second);
+            node->m_values.erase(nodeIt);
+            return detachedNode;
+        }
+        return Node();
+    }
+
+    template<class T>
+    bool DomPrefixTree<T>::AttachNodeAtPath(const Path& path, const Node& nodeToAttach)
+    {
+        Node* node = &m_rootNode;
+        const size_t entriesToIterate = path.Size() - 1;
+        for (size_t i = 0; i < entriesToIterate; ++i)
+        {
+            const PathEntry& entry = path[i];
+            auto nodeIt = node->m_values.find(entry);
+            if (nodeIt == node->m_values.end())
+            {
+                return false;
+            }
+            node = &nodeIt->second;
+        }
+
+        node->m_values[path[path.Size() - 1]] = nodeToAttach;
+        return true;
     }
 
     template<class T>
@@ -326,6 +383,24 @@ namespace AZ::Dom
                 nodeIt->second.m_data = {};
             }
         }
+    }
+
+    template<class T>
+    DomPrefixTree<T> DomPrefixTree<T>::DetachSubTree(const Path& path)
+    {
+        Node node = DetachNodeAtPath(path);
+        if (!node.IsEmpty())
+        {
+            return DomPrefixTree<T>(AZStd::move(node));
+        }
+        return DomPrefixTree<T>();
+    }
+
+    template<class T>
+    bool DomPrefixTree<T>::AttachSubTree(const Path& path, const DomPrefixTree& subTree)
+    {
+        const Node* node = subTree.GetNodeForPath(AZ::Dom::Path());
+        return AttachNodeAtPath(path, *node);
     }
 
     template<class T>
--- a/Code/Framework/AzCore/AzCore/DOM/DomPrefixTree.inl
+++ b/Code/Framework/AzCore/AzCore/DOM/DomPrefixTree.inl
@@ -91,7 +91,7 @@ namespace AZ::Dom
     }
 
     template<class T>
-    bool DomPrefixTree<T>::AttachNodeAtPath(const Path& path, const Node& nodeToAttach)
+    bool DomPrefixTree<T>::AttachNodeAtPath(const Path& path, Node&& nodeToAttach)
     {
         Node* node = &m_rootNode;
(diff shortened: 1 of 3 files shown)


srikappa-amzn:
@NicholasVanSickle , This might be a good one for you to look at. I've made sure that the internals of DomPrefixTree (Node) aren't exposed to any public functions or constructors.

srikappa-amzn:
@lumberyard-employee-dm , the only other major system that uses AZ::Dom::PrefixTree is the SettingsRegistryOriginTracker. If that class wants to use these 2 functions in the future, the SettingsRegistryOrigin struct would need to add copy and move constructors. Just wanted to call it out here to see if that is okay.

lumberyard-employee-dm:
> @lumberyard-employee-dm , the only other major system that uses AZ::Dom::PrefixTree is the SettingsRegistryOriginTracker. If that class wants to use these 2 functions in the future, the SettingsRegistryOrigin struct would need to add copy and move constructors. Just wanted to call it out here to see if that is okay.

I think the `AttachSubTree` function should only accept an rvalue `Node` and therefore the only requirement would be that the class would need to be move constructible.

srikappa-amzn:
> > @lumberyard-employee-dm , the only other major system that uses AZ::Dom::PrefixTree is the SettingsRegistryOriginTracker. If that class wants to use these 2 functions in the future, the SettingsRegistryOrigin struct would need to add copy and move constructors. Just wanted to call it out here to see if that is okay.
> 
> I think the `AttachSubTree` function should only accept an rvalue `Node` and therefore the only requirement would be that the class would need to be move constructible.
> 
> Alternatively a second overload of `AttachSubTree` can be added to support move-only types.

The reason I added a copy constructor is because I want to be able to store a copy of DomPrefixTree in the undo object so that when a user hits a redo, I can simply copy the DomPrefixTree and attach it to the main tree. I wanted the tree to be stored in the undo object because undo/redo objects should be capable enough to switch between 2 states without any additional data. 

So, the current workflow is
1. user hits detach
2. We create an undo object and move the detached tree into it
3. If user hits undo, we copy the tree in the undo object and use it to attach.
4. If user hits redo, we simply detach and discard the detached tree because we already have a copy.

An alternate workflow that would can avoid the copy constructor is 
1. user hits detach
2. We create an undo object and move the detached tree into it
3. If user hits undo, we move the tree in the undo object and use it to attach. At this point the undo object will have an empty tree. Since users can't undo the same action twice, this tree being empty won't be an issue
4. If user hits redo, we simply detach and store the detached tree again in the undo object.

The reason I didn't go with the alternate workflow is because it feels like a bad design to be constantly keep updating the data stored in undo objects with undo/redo calls instead of just capturing once and using the captured data for subsequent undo/redo calls. @lawsonamzn , do you have any thoughts on whether we can go with the alternate approach?

lumberyard-employee-dm:
> > > @lumberyard-employee-dm , the only other major system that uses AZ::Dom::PrefixTree is the SettingsRegistryOriginTracker. If that class wants to use these 2 functions in the future, the SettingsRegistryOrigin struct would need to add copy and move constructors. Just wanted to call it out here to see if that is okay.
> > 
> > 
> > I think the `AttachSubTree` function should only accept an rvalue `Node` and therefore the only requirement would be that the class would need to be move constructible.
> > Alternatively a second overload of `AttachSubTree` can be added to support move-only types.
> 
> The reason I added a copy constructor is because I want to be able to store a copy of DomPrefixTree in the undo object so that when a user hits a redo, I can simply copy the DomPrefixTree and attach it to the main tree. I wanted the tree to be stored in the undo object because undo/redo objects should be capable enough to switch between 2 states without any additional data.
> 
> So, the current workflow is
> 
>     1. user hits detach
> 
>     2. We create an undo object and move the detached tree into it
> 
>     3. If user hits undo, we copy the tree in the undo object and use it to attach.
> 
>     4. If user hits redo, we simply detach and discard the detached tree because we already have a copy.
> 
> 
> An alternate workflow that would can avoid the copy constructor is
> 
>     1. user hits detach
> 
>     2. We create an undo object and move the detached tree into it
> 
>     3. If user hits undo, we move the tree in the undo object and use it to attach. At this point the undo object will have an empty tree. Since users can't undo the same action twice, this tree being empty won't be an issue
> 
>     4. If user hits redo, we simply detach and store the detached tree again in the undo object.
> 
> 
> The reason I didn't go with the alternate workflow is because it feels like a bad design to be constantly keep updating the data stored in undo objects with undo/redo calls instead of just capturing once and using the captured data for subsequent undo/redo calls. Probably someone with more knowledge about the undo/redo system can pitch in here. @lawsonamzn , do you have any thoughts on whether we can go with the alternate approach?

How about the second overload for AttachSubTree that does accept an rvalue then?

lawsonamzn:
It _would_ be weird to actually modify the data inside already captured undo objects when you do the undo or redo.  

That being said it is not "illegal" or something though.  
Every undo node merely always has to be able to (on undo) move the 'world' into the prior state, and on redo, move it into the following state, and there's not a lot of rules about how that is accomplished.  As long as each undo node in isolation is able to do that, it doesn't really matter if acutally undoing mutates them.  The danger comes from if something later in the undo chain depends on something earlier in the undo chain (and vice versa).

srikappa-amzn:
> It _would_ be weird to actually modify the data inside already captured undo objects when you do the undo or redo.
> 
> That being said it is not "illegal" or something though. Every undo node merely always has to be able to (on undo) move the 'world' into the prior state, and on redo, move it into the following state, and there's not a lot of rules about how that is accomplished. As long as each undo node in isolation is able to do that, it doesn't really matter if acutally undoing mutates them. The danger comes from if something later in the undo chain depends on something earlier in the undo chain (and vice versa). Modifying the data in an earlier undo node when there are nodes later on that follow it is dangerous for that reason. If you don't do it perfectly, then its possible for a following node later on in the chain (when you then redo a few times) to get messed up because history isn't as it expects.

After weighing the pros and cons, I've decided to go with the alternate approach of avoiding the copy constructor. The reasons to do so are:
1. It avoids systems using the DomPrefixTree to implement both copy and move constructors if they want to use the new detach and attach calls
2. It avoids confusion on whether you should copy or move the tree while attaching and results in a cleaner API
3. Integration with prefabs would avoid copying of rapidjson::Document data, which based on recent investigations proved to leak memory and take a lot of time too as the documents grows in size
4. The case of undo nodes depending on other undo nodes is something I don't see being an issue with prefabs and since all of those nodes are custom undo nodes that the prefab system owns and maintains, working around that issue will not be a problem.

lumberyard-employee-dm (approved):

Code/Framework/AzCore/AzCore/DOM/DomPrefixTree.inl
```diff
@@ -328,9 +385,33 @@ namespace AZ::Dom
...
 
+    template<class T>
+    DomPrefixTree<T> DomPrefixTree<T>::DetachSubTree(const Path& path)
+    {
+        Node node = DetachNodeAtPath(path);
+        if (!node.IsEmpty())
+        {
+            return DomPrefixTree<T>(AZStd::move(node));
+        }
+        return DomPrefixTree<T>();
```
NicholasVanSickle, line 396: nit: seems like this can be a one-liner `return DomPrefixTree<T>(AZStd::move(node));` since `DetachNodeAtPath` always returns a valid node.

NicholasVanSickle (approved):
Looks like a nice API addition to me!

lawsonamzn (approved):

Code/Framework/AzCore/AzCore/DOM/DomPrefixTree.inl
srikappa-amzn: I'll update this in the next PR since this one is ready for a merge

Issue #13111, opened 2022-11-10
It is not possible to Delete assets and folders via Asset Browser if engine is built using source configuration


LB-JakubSkorupka:
**Describe the bug**
In **Asset Browser**, it is not possible to delete files or folders by clicking "Delete" in selected asset context menu or pressing Delete key with asset or folder being selected in **Asset Browser**.

This issue seems to occur only on source configuration engine, as it was reproduced on 2 Windows and 1 Linux machines. This issue did not occur on one machine with Pre-built engine configuration.

**Steps to reproduce**
1. Create a new level.
2. In **Asset Browser** find unused asset.
3. Right click on the found asset.
4. Click the **Delete** option in context menu.
5. Make sure the asset is still selected.
6. Press **Delete** key.

**Expected behavior**
It is possible to delete assets or folders via **Asset Browser**.

**Actual behavior**
It is **not** possible to delete assets or folders via **Asset Browser**.

**Video**

https://user-images.githubusercontent.com/86951967/201119437-ff802bdf-660f-4f32-8d47-2bce6de50897.mp4

**Found in Branch**
Development (522b935)

**Desktop:**
 - Device: PC
 - OS: Windows
 - Version: 10
 - CPU: Intel Core i7-8700K
 - GPU: NVIDIA GeForce RTX 2080 Ti
 - Memory: 16GB

jjjoness:
@LB-JakubSkorupka @lsemp3d  This was caused by the context menu not checking whether the asset was a source asset or a product asset. It now checks and the menu will only appear if the asset is source. Also fixes Rename and Move-to.

LB-JakubSkorupka:
Reopening the issue as this issue still occurs on build Development (5234076).
Please note that this issue also occurred on one machine with Pre-built SDK engine.

I also noticed that "Move To" option in context menu does not work. Renaming empty folder or using "Move To" option from context menu on empty folder, results in deleting that folder (this issue also occurs with Pre-built engine).

jjjoness:
@LB-KatarzynaDylska Working on this along with 15320

LB-JakubSkorupka:
Reopening the issue as it reproduced on 3 machines, two of those machines was Development (9f4341f) builds with Source configuration engines.
We used Stabilization/2305 installer on 3rd machine.
The issue slightly differs from the originaly reported one, currently:
 1. It is not possible to delete folders with any assets or folders in it.
 2. It is not possible to delete assets.
 3. It is possible to delete empty folders.

[...]

PR #13168, opened 2022-11-14
Fixed by checking if source or product before showing the menu

jjjoness:
Signed-off-by: John Jones-Steele <82226755+jjjoness@users.noreply.github.com>

## What does this PR do?

Fixes bug #13111 

This bug was caused by not checking the asset for being source not product before adding to the context menu.

## How was this PR tested?

Checked many items in the Asset Browser to see that the menu works correctly.


LB-JakubSkorupka:
We noticed similar issue with "Move To" option in Asset Browser. Folders and assets cannot be moved by using "Move to" option. Will this PR resolve also this issue or should we create separate issue?

jjjoness:
@LB-JakubSkorupka Also fixes Rename and Move-to.

lsemp3d (approved):

PR #13169, opened 2022-11-14
Added parallel transform sync for rigid bodies when batched sync is enabled

SergeyAMZN:
Added parallel transform sync for rigid bodies when batched sync is enabled

## What does this PR do?

Enables parallelized transform sync for rigid bodies using `AZ::TaskGraph`. 

## How was this PR tested?

Manual testing with 4320 rigid bodies, unit tests

Parallel_sync

### DIFF
--- a/Gems/PhysX/Code/Source/Scene/PhysXScene.cpp
+++ b/Gems/PhysX/Code/Source/Scene/PhysXScene.cpp
@@ -30,17 +30,22 @@
 #include <AzCore/std/containers/vector.h>
 #include <AzCore/std/smart_ptr/make_shared.h>
 #include <AzCore/Debug/Profiler.h>
+#include <AzCore/Task/TaskGraph.h>
 #include <AzFramework/Physics/Character.h>
 #include <AzFramework/Physics/Collision/CollisionEvents.h>
 #include <AzFramework/Physics/Configuration/RigidBodyConfiguration.h>
 #include <AzFramework/Physics/Configuration/StaticRigidBodyConfiguration.h>
 #include <AzFramework/Physics/Material/PhysicsMaterialManager.h>
 
-
-AZ_CVAR_EXTERNED(bool, physx_batchTransformSync);
-
 namespace PhysX
 {
+    AZ_CVAR_EXTERNED(bool, physx_batchTransformSync);
+
+    AZ_CVAR(bool, physx_parallelTransformSync, true, nullptr, AZ::ConsoleFunctorFlags::Null, "Multithreaded transform update for rigid bodies. "
+        "Only relevant if batched transform update is enabled.");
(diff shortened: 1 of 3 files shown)


Gems/PhysX/Code/Source/Scene/PhysXScene.cpp
```diff
@@ -1348,5 +1361,42 @@ namespace PhysX
...
+                taskGraph.AddTask(
+                    taskDescriptor,
+                    [start = i, end = AZStd::min(i + batchSize, fullSize), &applyFunction, pxScene, this]()
+                    {
+                        AZ_PROFILE_SCOPE(Physics, "Sync Task");
+
+                        // Note: It is important to keep the scene locked for read for the entire task execution.
+                        // Otherwise the functions reading data from the rigid body will have to lock it locally.
+                        // This causes a huge amount of context switches making the execution of each task ~20x slower. 
+                        PHYSX_SCENE_READ_LOCK(pxScene);
```
hultonha, line 1386: question: I'm just curious about how this works... If we lock the `pxScene` for each task, will that mean other tasks will have to wait and run serially? Or if they're all readers, is it fine for them to all hold this lock at the same time so they can all run in parallel? Thanks!

Gems/PhysX/Code/Source/Scene/PhysXScene.cpp
```diff
@@ -1348,5 +1361,42 @@ namespace PhysX
...
+            AZ_PROFILE_SCOPE(Physics, "Sync Setup");
+
+            size_t batchSize = physx_parallelTransformSyncBatchSize;
+            size_t fullSize = m_packedIndices.size();
+            for (size_t i = 0; i < fullSize; i += batchSize)
+            {
+                AZ::TaskDescriptor taskDescriptor{"SyncTask", "Physics"};
+                taskGraph.AddTask(
+                    taskDescriptor,
+                    [start = i, end = AZStd::min(i + batchSize, fullSize), &applyFunction, pxScene, this]()
```
hultonha, line 1379: Nice way to wrap the last elements that don't make up a full batch at the end 🙂

hultonha (approved):
This looks great to me! Awesome to see the speed improvement too 😁

moraaar (approved):

Gems/PhysX/Code/Source/Scene/PhysXScene.cpp
SergeyAMZN: All readers can read simultaneously and nested read-locks cost nothing. 
It's the first read lock per thread is expensive, so we do this one higher level read-lock to ensure all further read-locks nested.

hultonha: I see, that's great thanks for clarifying :+1:

Issue #13150, opened 2022-11-11
PrefabTests_Periodic.py is failing in test_AddEntity_UnderUnfocusedInstanceAsOverride


jckand-amzn:
**Describe the bug**
Newly added test test_AddEntity_UnderUnfocusedInstanceAsOverride is failing with a timeout. It appears the test may be hitting some kind of deadlock.

**Failed Jenkins Job Information:**
https://jenkins.build.o3de.org/blue/organizations/jenkins/O3DE_periodic-incremental-daily/detail/development/160/pipeline/1137

**Attachments**
[log(1).txt](https://github.com/o3de/o3de/files/9991496/log.1.txt)

PR #13170, opened 2022-11-14
Organizational updates and test fix for Prefab TestSuite_Periodic

jckand-amzn:
Signed-off-by: jckand-amzn <82226555+jckand-amzn@users.noreply.github.com>

## What does this PR do?

- Adds necessary setreg flag for running test_AddEntity_UnderUnfocusedInstanceAsOverride - Fixes https://github.com/o3de/o3de/issues/13150
- Moves all tests to use EditorBatchedTest to avoid parallel Editor issues
- Combined separate test classes, and now utilize both -BatchMode and -autotest_mode for all tests
- Updates a test level to remove deprecated SelectionComponent warning

## How was this PR tested?

- Ran updated periodic suite several times with 0 failures

### DIFF
--- a/AutomatedTesting/Gem/PythonTests/Prefab/TestSuite_Periodic.py
+++ b/AutomatedTesting/Gem/PythonTests/Prefab/TestSuite_Periodic.py
@@ -7,7 +7,7 @@ SPDX-License-Identifier: Apache-2.0 OR MIT
 
 import pytest
 
-from ly_test_tools.o3de.editor_test import EditorSingleTest, EditorSharedTest, EditorParallelTest, EditorTestSuite
+from ly_test_tools.o3de.editor_test import EditorSingleTest, EditorBatchedTest, EditorParallelTest, EditorTestSuite
 
 
 @pytest.mark.SUITE_periodic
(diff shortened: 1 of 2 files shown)


AutomatedTesting/Gem/PythonTests/Prefab/TestSuite_Periodic.py
```diff
@@ -7,73 +7,71 @@
...
 
 
 @pytest.mark.SUITE_periodic
 @pytest.mark.parametrize("launcher_platform", ['windows_editor'])
 @pytest.mark.parametrize("project", ["AutomatedTesting"])
 class TestAutomationNoAutoTestMode(EditorTestSuite):
 
-    # Enable only -BatchMode for these tests. Some tests cannot run in -autotest_mode due to UI interactions
-    global_extra_cmdline_args = ["-BatchMode"]
+    global_extra_cmdline_args = ["--regset=O3DE/Preferences/Prefabs/EnableOverridesUx=true"]
```
forkercat, line 18: Does it mean after this PR we now run all prefab automated tests with the flag on by default?

forkercat (approved):
looks good!

AutomatedTesting/Gem/PythonTests/Prefab/TestSuite_Periodic.py
jckand-amzn: I think it's the way to go, however have only tested with the periodic suite so far. I'll have another PR ready hopefully by EoD to move a bunch of tests over to EditorBatchedTest, and will likely make this change the Prefab TestSuite_Main as well.

cgalvan (approved):

AMZN-daimini (approved):

AutomatedTesting/Gem/PythonTests/Prefab/TestSuite_Periodic.py
srikappa-amzn: @jckand-amzn , eventually overrides will be the default editing mechanism but switching all the tests right now to use override mode by default seems like the wrong thing to do now. Every editor user uses the features with the flag turned off. So, by turning this option on, we will be opening ourselves to being disconnected from how the majority of users use and see the editor. 

I recommend that we enable this flag only for the override tests for now and make it a global setting when we are ready to release the default override mode.

PR #13171, opened 2022-11-14
[Prefab Overrides] Add support for delete entity and prefab as override

forkercat:

## What does this PR do?

This PR adds support for deleting entities and prefabs as overrides. It should work for both the old Prefab Edit Mode workflow and the new Override workflow. New workflow can be seen by turning on the overrides UX flag.

Users can do 4 kinds of deletions for one item (entity or prefab instance) selected:
1. Prefab Edit Mode
    1. Delete an entity owned by focused instance
    1. Delete an instance (container entity) owned by focused instance
1. Prefab Override Mode
    1. Delete an entity that is not owned by focused instance
    2. Delete an instance (container entity) that is not owned by focused instance

https://user-images.githubusercontent.com/11157226/199858383-93b8a0f4-bf99-48af-a7ba-91733942788d.mp4

Note: At the time I recorded it didn't include visualization changes. Now there should be an override icon in parent entity.

#### Major Changes
1. Add delete undo nodes that support previous Prefab Edit Mode workflow and the new override workflow.
1. Improvement: Removes copying initial parent entity DOM values in public handler. The entity before states come from focused template or root template if it is the focused container entity.

#### How Override Patching Works?
1. Generate removal patches and update patches
3. Apply all patches to the currently focused/edited template DOM
8. Update the link between focused template and the template of the top instance

## How was this PR tested?

The changes are behind a feature flag: `EnableOverridesUx`

- Manually tested with flag on and off. Passed existing unit tests and automated tests that are only for old workflow.
- Tested with new unit tests. New tests will be added in a follow-up PR.
- Tested with new automated tests. New tests will be added in a follow-up PR


Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -1340,87 +1347,102 @@ namespace AzToolsFramework
...
+                    if (entity->GetId().IsValid())
+                    {
+                        entitiesThatWillBeRemoved.insert(entity->GetId());
+                    }
+                });
 
-            // Map for capturing the initial DOM state before parent entities get updated.
-            AZStd::unordered_map<AZ::EntityId, PrefabDom> parentEntityDomBeforeRemovalMap;
+            // Set of parent entities that need to be updated. It should not include entities that will be moved.
+            AZStd::unordered_set<const AZ::Entity*> parentEntitysThatNeedUpdate;
```
srikappa-amzn, line 1366: spelling: parentEntities...

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -1340,87 +1347,102 @@ namespace AzToolsFramework
...
+                    parentEntitysThatNeedUpdate.insert(GetEntityById(parentEntityId));
                 }
 
-                const AZ::Entity* parentEntity = GetEntityById(parentEntityId);
-                if (parentEntity)
-                {
-                    PrefabDom parentEntityDomAfter;
-                    m_instanceToTemplateInterface->GenerateDomForEntity(parentEntityDomAfter, *parentEntity);
+                commonOwningInstance->get().DetachEntity(entityId).release();
+                AZ::ComponentApplicationBus::Broadcast(&AZ::ComponentApplicationRequests::DeleteEntity, entityId);
```
srikappa-amzn, line 1422: These are not new changes but it is interesting that to delete entities, we are calling this ebus but to delete entities in nested prefabs, we are simply calling detachedInstance.reset();. This is not consistent behavior for entities and prefabs. We should simply be calling reset() in both cases here. Can you give that a try and see if that works? If it breaks anything, please add a ghi to track this.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -1340,87 +1347,102 @@ namespace AzToolsFramework
...
 
-                    entitiesThatGotRemoved.insert(entityId);
+                if (isOverrideEditing)
+                {
+                    detachedInstanceAliasPaths.push_back(AZStd::move(PrefabDomUtils::PathStartingWithInstances + instanceAlias));
+                }
+                else
+                {
+                    // Removes the link if it is source template editing.
+                    RemoveLink(detachedInstance, commonOwningInstance->get().GetTemplateId(), undoBatch.GetUndoBatch());
```
michabr, line 1400: Would it make sense to move this to the DeleteEntities? That way we would get a nice separation from entity/instance object manipulation vs. prefab DOM manipulation.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoDelete.cpp
```diff
@@ -0,0 +1,96 @@
...
+                        continue;
+                    }
+
+                    PrefabDom parentEntityDomAfterRemovingChildren;
+                    m_instanceToTemplateInterface->GenerateDomForEntity(parentEntityDomAfterRemovingChildren, *parentEntity);
+
+                    PrefabUndoUtils::AppendUpdateEntityPatch(
+                        m_redoPatch, *parentEntityDomInFocusedTemplate, parentEntityDomAfterRemovingChildren, parentEntityAliasPath);
+                    PrefabUndoUtils::AppendUpdateEntityPatch(
+                        m_undoPatch, parentEntityDomAfterRemovingChildren, *parentEntityDomInFocusedTemplate, parentEntityAliasPath);
```
michabr, line 110: I'm wondering if we can optimize this by generating the remove patches manually rather than doing a "GenerateDomForEntity".

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoUtils.h
```diff
@@ -27,18 +26,26 @@ namespace AzToolsFramework
                 const PrefabDomValue& newEntityDom,
                 const AZStd::string& newEntityAliasPath);
 
-            //! Create remove-entity patch from entity's alias path, and append it to a given instance 
-            //! patch which is an array of entity patches.
+            //! Create remove patch from an alias path, and append it to a given instance patch.
             //! @param patch An array object of DOM values which stores entity patches.
```
michabr, line 30: Should we remove the word "entity" from this comment?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabUndoHelpers.h
```diff
@@ -58,6 +59,32 @@ namespace AzToolsFramework
                 const PrefabDomValue& entityDomAfterUpdatingEntity,
                 AZ::EntityId entityId,
                 UndoSystem::URSequencePoint* undoBatch);
+
+            //! Helper function for deleting entities to source template with undo-redo support.
+            //! @param entityAliasPathList The alias path list for entities that will be removed.
+            //! @param parentEntityList The parent entity list that will be updated.
+            //! @param focusedInstance The current focused prefab instance.
+            //! @param undoBatch The undo batch node to register the removal undo nodes to.
+            void DeleteEntities(
```
michabr, line 68: Can you describe why we need a DeleteEntities function and also a RemoveEntities function? It's not obvious by the function names what the differences are. I suggest renaming to make the differences more apparent.

michabr (commented):
Nice! Added some minor comments/questions.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoUtils.h
forkercat: yeah! I am reviewing all the comments in this util class to make sure they are correct and accurate.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoDelete.cpp
forkercat: AFAIK, AppendUpdateEntityPatch is doing two jobs:

1. Update component properties for entity
2. Update parent entity's child sort order

For the first job, I think it might be yes. But we need to do some profiling first to see if it brings noticeable improvement, because imo generating entity DOM is acceptable compared to generating an instance DOM. Another pain point would be how we can locate a particular component rather than doing a linear search.

For the second job, I think it might be hard. This type of patches can be >1 depending on which entity you're updating in the N children list. If it is at the first position, it would be N (or N-1) patches. It is like inserting an element to vector like what `m_numbers.insert(0, newNumber)` does to move elements behind forward by 1 position.

So due to multiple factors, we are still generating entity DOM from serialization and doing diff to find what needs to updated. But I think we already started being smart about this for updates like adding/removing an entity DOM.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
forkercat: Do you mean DeleteEntities helper function or delete undo node? Two of reasons would be reusing RemoveLink in handler and managing Instance lifecycle in handler instead of passing down instance pointers to helper.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabUndoHelpers.h
forkercat: I was planning to remove it but recent changes in CreatePrefabInMemory just started depending on it.

I put a note:
```
//! Note: It does not include updating relevant parent entities.
void RemoveEntities(
```
But yeah I agree I should put more info about the differences or maybe rename it to "RemoveEntityDoms".

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoDelete.cpp
```diff
@@ -0,0 +1,96 @@
...
+                    PrefabUndoUtils::UpdateCachedOwningInstanceDomWithRemoval(cachedOwningInstanceDom, entityAliasPath);
+                }
+            }
+
+            // Updates parent entity DOMs.
+            for (const AZ::Entity* parentEntity : parentEntityList)
+            {
+                // This is actually a from-focused path because focused instance is the owning instance in the current workflow.
+                const AZStd::string parentEntityAliasPath = m_instanceToTemplateInterface->GenerateEntityAliasPath(parentEntity->GetId());
+                PrefabDomPath parentEntityAliasDomPathFromFocused(parentEntityAliasPath.c_str());
```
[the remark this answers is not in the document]
forkercat: I didn't follow. Do you mean "is there an API that can be used to pass in owningInstance"?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
forkercat: I tried it and it seemed to work with no issue.

I think the difference between the two ways is ComponentApplication::DeleteEntity also signals "entity removed event". Code Ref:

https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp#L1171

But I did not notice any "entity removed event" handlers being registered for this. I also did a code search and are not seeing DetachEntity(entitiyId).reset() usage in other places.

srikappa-amzn: It is a good thing to be worrying about backwards compatibility. However, there is also the consistency issue where we are only calling this api for direct entities only
. And since there doesn't appear to be other systems that are using this additional event, I think it's a safe thing to start calling reset() now
.

If you want to separate it from your PR, you can try it in a follow up PR so that it can be easily reverted if needed

forkercat: Discussed offline. The inconsistency is between deleting direct entities and deleting entities in nested instances (during detachedInstance.reset()). The first one involves calling the DeleteEntity event API.

I will put up a follow-up PR for this change.

srikappa-amzn (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoDelete.cpp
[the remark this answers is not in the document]
michabr: Yeah, I'm wondering if we should replace m_instanceToTemplateInterface->GenerateEntityAliasPath with a call that we can pass focusedInstance to as an argument.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
michabr: The benefit I see is the code would be easier to follow. PrefabUndoHelpers::DeleteEntitiesAndPrefabsAsOverride would be more consistent with PrefabUndoHelpers::DeleteEntities.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoDelete.cpp
michabr: That's fair. Thanks for the details.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoDelete.cpp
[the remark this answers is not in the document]
forkercat: I think it is not needed for now. All similar APIs in this class are taking in an entity id. fyi We recently added a new API: GenerateEntityPathFromFocusedPrefab(entityId).

I'm wondering if later it would be good to have single generic interface that handles:
1. AliasPath from owning instance
2. AliasPath from focused instance (if provided)

@srikappa-amzn any thoughts?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
forkercat: We can definitely extend from the current classes. I can create a backlog task for this. For now I want to keep it simple and reuse what we already have to make sure delete-prefab override works correctly.

Once deletion in multiple owning instances and instantiate-prefab as override gets in, at that time I think we can do the refactoring for link updates.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoDelete.cpp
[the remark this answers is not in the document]
michabr: `        void PrefabUndoDeleteEntity::Capture(
            const AZStd::vector<AZStd::string>& entityAliasPathList,
            const AZStd::vector<const AZ::Entity*> parentEntityList,
            Instance& focusedInstance)`

Since this function takes an instance reference as an argument, it seems confusing to later call other API methods that retrieve the "focused instance" on their own. My suggestion is to either not pass in focusedInstance at all and use API calls that know what the focused instance is, or pass in the focused instance but use API calls that also take a focused instance as an argument.

I may be missing something though, so let me know if I am. :)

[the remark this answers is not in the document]
srikappa-amzn: I think it's better to use GenerateEntityPathFromFocusedPrefab() in this use case rather than adding a comment to explain the intention. It partly addresses @michabr point about not being obvious where the path is generated from. It doesn't use the focused instance you already have in this function though. If you want to avoid that too, you can provide a new private helper function in this class that implements something like GenerateEntityPathFromFocusedPrefab itself. But I don't think it's worth the effort to do that here. Getting the focused instance twice shouldn't be an issue performance wise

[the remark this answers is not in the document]
michabr: I'm more concerned about having the API be super intuitive rather than this being a performance issue, but I'll go ahead and approve the PR since this doesn't seem like a major concern.

michabr (approved):

PR #13172, opened 2022-11-14
[Terrain] First batch of unit tests for PaintBrush

mbalfour-amzn:
## What does this PR do?

This adds a set of unit tests for PaintBrush that test most of the base functionality:
- Paint Mode begin/end
- Brush Stroke begin/end
- PaintToLocation, SmoothToLocation, ResetPaintBrushTracking all trivially work and handle locations correctly

The one functional change introduced here is that I realized the PaintBrush code was inconsistently converting things into 2D space for painting calculations. I've made this more explicit and consistent and (not really) conditional based on the constant "PaintIn2D". It's structured this way because we might eventually want to open this up as a runtime parameter so that we can have full 3D painting as well, but since we don't have any use cases for that yet I'd like to avoid adding extra parameters and unit tests before we have the need.

## How was this PR tested?

Ran the unit tests and verified they all passed.

### DIFF
--- a/Code/Framework/AzToolsFramework/Tests/PaintBrush/PaintBrushBaseTests.cpp
+++ b/Code/Framework/AzToolsFramework/Tests/PaintBrush/PaintBrushBaseTests.cpp
@@ -1,0 +1,249 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#include <AzCore/Serialization/SerializeContext.h>
+#include <AzCore/UnitTest/TestTypes.h>
+#include <AzTest/AzTest.h>
+#include <AzToolsFramework/PaintBrush/PaintBrush.h>
+#include <AZTestShared/Math/MathTestHelpers.h>
+#include <Tests/PaintBrush/MockPaintBrushNotificationHandler.h>
+
+/*
+This set of unit tests validates that the basic PaintBrush functionality and notifications work:
+- Begin/End Paint Mode
+- Begin/End Brush Stroke
+- Use Eyedropper
+*/
+
+namespace UnitTest
+{
+    class PaintBrushTestFixture : public ScopedAllocatorSetupFixture
+    {
+    public:
+
+        // Some common default values that we'll use in our tests.
+        const AZ::EntityComponentIdPair EntityComponentIdPair{ AZ::EntityId(123), AZ::ComponentId(456) };
+        const AZ::Vector3 TestBrushCenter{ 10.0f, 20.0f, 30.0f };
+        const AZ::Vector3 TestBrushCenter2d{ 10.0f, 20.0f, 0.0f };  // This should be the same as TestBrushCenter, but with Z=0
+        const AZ::Color TestColor{ 0.25f, 0.50f, 0.75f, 1.0f };
+
+        // Useful helpers for our tests
+        AzToolsFramework::PaintBrushSettings m_settings;
+    };
+
+    TEST_F(PaintBrushTestFixture, TrivialConstruction)
+    {
+        // We can construct / destroy a PaintBrush without any errors.
+
+        AzToolsFramework::PaintBrush paintBrush(EntityComponentIdPair);
+    }
+
+    TEST_F(PaintBrushTestFixture, BeginAndEndPaintModeMustBePairedCorrectly)
+    {
+        // BeginPaintMode() / EndPaintMode() can only be called in the correct order in pairs.
+        // They will error if they're called in the wrong order or more than once in a row.
+
+        AzToolsFramework::PaintBrush paintBrush(EntityComponentIdPair);
+
+        // End before Begin should assert.
+        AZ_TEST_START_TRACE_SUPPRESSION;
+        paintBrush.EndPaintMode();
+        AZ_TEST_STOP_TRACE_SUPPRESSION(1);
+
+        paintBrush.BeginPaintMode();
+
+        // Calling Begin twice should assert.
+        AZ_TEST_START_TRACE_SUPPRESSION;
+        paintBrush.BeginPaintMode();
+        AZ_TEST_STOP_TRACE_SUPPRESSION(1);
+
+        paintBrush.EndPaintMode();
+
+        // Calling End twice should assert.
+        AZ_TEST_START_TRACE_SUPPRESSION;
+        paintBrush.EndPaintMode();
+        AZ_TEST_STOP_TRACE_SUPPRESSION(1);
+    }
+
+    TEST_F(PaintBrushTestFixture, IsInPaintModeFunctionsCorrectly)
+    {
+        // IsInPaintMode() correctly identifies when we're between a BeginPaintMode() and EndPaintMode() call.
+
+        AzToolsFramework::PaintBrush paintBrush(EntityComponentIdPair);
+
+        EXPECT_FALSE(paintBrush.IsInPaintMode());
+        paintBrush.BeginPaintMode();
+        EXPECT_TRUE(paintBrush.IsInPaintMode());
+        paintBrush.EndPaintMode();
+        EXPECT_FALSE(paintBrush.IsInPaintMode());
+    }
+
+    TEST_F(PaintBrushTestFixture, BeginAndEndBrushStrokeMustBePairedCorrectlyInsidePaintMode)
+    {
+        // BeginBrushStroke() / EndBrushStroke() can only be called in the correct order in pairs.
+        // They will error if they're called in the wrong order, more than once in a row, or outside of BeginPaintMode().
+
+        AzToolsFramework::PaintBrush paintBrush(EntityComponentIdPair);
+
+        // BeginBrushStroke without BeginPaintMode should assert.
+        AZ_TEST_START_TRACE_SUPPRESSION;
+        paintBrush.BeginBrushStroke(m_settings);
+        AZ_TEST_STOP_TRACE_SUPPRESSION(1);
+
+        paintBrush.BeginPaintMode();
+
+        // End before Begin should assert.
+        AZ_TEST_START_TRACE_SUPPRESSION;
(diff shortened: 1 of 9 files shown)


cgalvan (approved):

Code/Framework/AzToolsFramework/Tests/PaintBrush/PaintBrushBaseTests.cpp
```diff
@@ -0,0 +1,249 @@
...
+    TEST_F(PaintBrushTestFixture, BeginAndEndBrushStrokeMustBePairedCorrectlyInsidePaintMode)
+    {
+        // BeginBrushStroke() / EndBrushStroke() can only be called in the correct order in pairs.
+        // They will error if they're called in the wrong order, more than once in a row, or outside of BeginPaintMode().
+
+        AzToolsFramework::PaintBrush paintBrush(EntityComponentIdPair);
+
+        // BeginBrushStroke without BeginPaintMode should assert.
+        AZ_TEST_START_TRACE_SUPPRESSION;
+        paintBrush.BeginBrushStroke(m_settings);
```
hultonha, line 95: suggestion: Enforcing this strict API usage does make a lot of sense but I wanted to mention one term I came across while looking at another library ([flecs](https://flecs.docsforge.com/master/manual/#idempotence)) called '[idempotence](https://en.wikipedia.org/wiki/Idempotence)'. It means invoking an operation once has the same effect as invoking an operation multiple times, which would mean if you call `begin` a second time before calling `end` it would basically be a noop and not be an error. Maybe this is something you definitely don't want but it just made me think of it and I wanted to share it just as food for thought as a slightly different philosophy (likely with its own downsides/drawbacks as well).

Code/Framework/AzToolsFramework/Tests/PaintBrush/PaintBrushPaintLocationTests.cpp
```diff
@@ -0,0 +1,339 @@
...
+                        AZ::Vector3(dirtyArea.GetMin().GetX(), dirtyArea.GetMax().GetY(), 0.0f),
+                        AZ::Vector3(dirtyArea.GetCenter().GetX(), dirtyArea.GetMax().GetY(), 0.0f),
+                        AZ::Vector3(dirtyArea.GetMax().GetX(), dirtyArea.GetMax().GetY(), 0.0f),
+                    };
+                    AZStd::vector<AZ::Vector3> validPoints;
+                    AZStd::vector<float> opacities;
+                    valueLookupFn(points, validPoints, opacities);
+
+                    // We should only have the 5 points along the + in validPoints.
+                    EXPECT_EQ(validPoints.size(), 5);
```
hultonha, line 85: uber nit: This check is technically redundant as `:testing::Pointwise(` will fail if the number of points doesn't match but this can stay just for readability if you'd prefer it

Code/Framework/AzToolsFramework/Tests/PaintBrush/PaintBrushPaintLocationTests.cpp
```diff
@@ -0,0 +1,339 @@
...
+        // Set the brush radius to 1 meter (diameter is 2 meters)
+        const float TestBrushRadius = 1.0f;
+        m_settings.SetSize(TestBrushRadius * 2.0f);
+
+        // The distance we expect to need to move to trigger another paint call is 50% of our brush size.
+        const float DistanceToTriggerSecondCall = TestBrushRadius * 2.0f * (TestDistancePercent / 100.0f);
+
+        // Choose a second brush center location that's just slightly under the threshold that should be needed to trigger a second
+        // OnPaint call.
+        const AZ::Vector3 tooSmallSecondLocation = TestBrushCenter + AZ::Vector3(DistanceToTriggerSecondCall - 0.01f, 0.0f, 0.0f);
```
hultonha, line 138: question: Could these be tied to constants somewhere as these values might change in future potentially? If you think this is unlikely it's fine to leave as-is

hultonha (approved):
These tests are a sight to behold! 😄 Awesome work @mbalfour-amzn. I only had a few minor nits/suggestions/questions, overall looks fantastic!

The one thing I wanted to check is it'd be good to try and intentionally 'break' some internal code and verify the tests fail when you expect them too if you haven't already. It's all too easy to have the tests pass even when they should fail. You may well have done this already but I just wanted to double check as it can be a nice thing to confirm.

Code/Framework/AzToolsFramework/Tests/PaintBrush/PaintBrushBaseTests.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+    TEST_F(PaintBrushTestFixture, BeginAndEndPaintModeMustBePairedCorrectly)
+    {
+        // BeginPaintMode() / EndPaintMode() can only be called in the correct order in pairs.
+        // They will error if they're called in the wrong order or more than once in a row.
+
+        AzToolsFramework::PaintBrush paintBrush(EntityComponentIdPair);
+
+        // End before Begin should assert.
+        AZ_TEST_START_TRACE_SUPPRESSION;
```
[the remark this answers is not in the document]
mbalfour-amzn: Hmmm, thanks for pointing me at it! I've got mixed feelings about using it. I like that you can make sure you're not getting a false positive from an unexpected assert, but I don't like that it's specifically testing the wording of the assert, since that's something that should be able to change without breaking unit tests.

My inclination is to _not_ use it here, since these are some pretty trivial assert tests that would be difficult to fail unexpectedly, but if you feel strongly about it I can switch them.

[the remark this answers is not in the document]
hultonha: It's totally fine to leave as-is I just wanted to mention it as an option (might be useful in another context in future if you want to check something specific). Agreed it's not perfect but for asserts that's about the only 'unique' property they have (the associated string/text).

Code/Framework/AzToolsFramework/Tests/PaintBrush/PaintBrushBaseTests.cpp
mbalfour-amzn: Oooh, this is one I feel really strongly about, so I'll skip the suggestion. One of my core design principles is "symmetry". I won't create a Begin() without an End(), a Startup() without a Shutdown(), etc. In the same vein, I do my best to ensure that people have paired, symmetric calls to my APIs. I've generally found it to be an anti-pattern to allow asymmetric calls, and it I believe it to be an overall disservice to the API user to allow this. Most of the time, from what I've seen, asymmetric calls are a result of incorrect API usage, they cause less predictable resource usage and lifetimes, and they're really hard to fix at the point that they start really causing problems.

There are definitely some exception cases every now and again where you need to allow for _nested_ Begin/End calls, but even in those cases I've still lightly enforced pairing with a counter. Though more often than not even in these cases you actually _can_ redesign to avoid nested calls by using keys or identifiers.

hultonha: Totally valid points and I tend to agree with you, just seeing the way this is structured made me think of the idempotence concept and I wanted to mention it just as a philosophical alternative. Please do leave as is, it's just interesting these different viewpoints exist out in the wild 🙂

Code/Framework/AzToolsFramework/Tests/PaintBrush/PaintBrushPaintLocationTests.cpp
mbalfour-amzn: Oooh, neat, didn't know that! I'll remove all of them except the one that's checking for 3 valid points from 4 input points, since that's worth documenting a little more explicitly.

Code/Framework/AzToolsFramework/Tests/PaintBrush/PaintBrushPaintLocationTests.cpp
mbalfour-amzn: Can you explain what you mean a bit more here? 

The way the paintbrush works is that it draws circles every N pixels, where N is determined by the radius of the circle and the Distance % setting. This test is verifying that if you've moved _less_ than (BrushSize * Distance%), then a circle won't get drawn, but if you move at least that amount, then it will. 

The only constant at this point in the test is the "- 0.01", which is just an arbitrary small amount. This test should work equally well if that's changed to "- 1" or "- 0.00001". (At some point float precision will come into play, but you get the idea)

I'm not sure what you're suggesting should be turned into constants outside of the test here...

mbalfour-amzn:
> The one thing I wanted to check is it'd be good to try and intentionally 'break' some internal code and verify the tests fail when you expect them too if you haven't already. It's all too easy to have the tests pass even when they should fail. You may well have done this already but I just wanted to double check as it can be a nice thing to confirm.

Oh no worries, I do this ALL the time both intentionally and unintentionally when writing tests. I don't trust my ability to write tests enough to just assume that a passing test is passing because it's correct. It's also why you see things like the paired "expect call" / "on call" constructs. If I only wrote "on call", the test could pass by never calling the "on call" mock handler, so I add the "expect call" to make sure that case doesn't happen. :)

mbalfour-amzn:
I'll address the nits / suggestions in my next PR, since I have another round of unit tests nearly ready to post.

Code/Framework/AzToolsFramework/Tests/PaintBrush/PaintBrushPaintLocationTests.cpp
hultonha: Ah okay understood - yes I really just meant the magic number here (`0.01`) that might be an important minimum that could be named and comes up as a tolerance in the main paintbrush code. If it's totally arbitrary then it's not a big deal, I just didn't want to miss the opportunity that something could be moved to a named constant if it made sense 🙂

mbalfour-amzn: Thx for explaining! Yep, it's a 100% totally arbitrary constant here, it has nothing to do with the main paintbrush code. Guess I should improve the comment here to make that more explicit. :)

PR #13175, opened 2022-11-14
Asset Processor - Metadata System

amzn-mike:
## What does this PR do?

Adds a metadata manager component available for use in tools applications which can read/write generic values to a metadata JSON file stored alongside a source asset.

## How was this PR tested?

Created and ran unit tests


Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/MetadataManager.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+#include <AzToolsFramework/Metadata/MetadataManager.h>
+
+namespace AzToolsFramework
+{
+    void MetadataManager::Reflect(AZ::ReflectContext* context)
+    {
+        if (auto* serializeContext = azrtti_cast<AZ::SerializeContext*>(context))
+        {
+            serializeContext->Class<MetadataManager>()->Version(1);
```
AMZN-stankowi, line 17: Should this version be set to MetadataVersion? Is there a reason you wouldn't want them in sync?

AMZN-stankowi (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/MetadataManager.cpp
amzn-mike: Interesting thought, I think maybe they should be separate though, because technically they're different serializations.  MetadataVersion is the version for the file format, this is the version for the component itself, which doesn't have any data (and probably won't...).

AMZN-stankowi: That makes sense, maybe add a comment in the code itself, too, so someone else doesn't think the same thing later?

Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/MetadataManager.h
```diff
@@ -0,0 +1,68 @@
...
+#pragma once
+
+#include <AzCore/Component/Component.h>
+#include <AzCore/Interface/Interface.h>
+#include <AzCore/IO/Path/Path.h>
+#include <AzCore/Serialization/Json/JsonUtils.h>
+#include <AzCore/IO/FileReader.h>
+#include <AzCore/Serialization/Json/JsonSerialization.h>
+#include <AzCore/Serialization/Json/StackedString.h>
+#include <AzCore/JSON/pointer.h>
```
jackalbe, line 18: can some of these header files be moved into the CPP file?

Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/MetadataManager.h
```diff
@@ -0,0 +1,68 @@
...
+
+        static void Reflect(AZ::ReflectContext* context);
+
+    protected:
+        void Activate() override{}
+        void Deactivate() override{}
+
+    public:
+        bool Get(AZ::IO::PathView file, AZStd::string_view key, void* outValue, AZ::Uuid typeId) override;
+        bool Set(AZ::IO::PathView file, AZStd::string_view key, const void* inValue, AZ::Uuid typeId) override;
```
jackalbe, line 63: Get() and Set() are just verbs. Consider adding a noun to the APIs like SetValue() or GetValue() or something that better describes what is being fetched or mutated.

lmbr-pip (commented):
Do we have a json schema or at least versioning system in place for the metadata file?

I believe this is part of the asset relocation work, so consider linking to RFC.

Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/MetadataManager.h
amzn-mike: Good catch, I had everything in the header to start with and forgot to the move them over.  Fixed!

Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/MetadataManager.cpp
```diff
@@ -0,0 +1,158 @@
...
+            // Save JSON to disk
+            auto saveResult = AZ::JsonSerializationUtils::WriteJsonFile(document, path.Native());
+
+            return saveResult.IsSuccess();
+        }
+
+        return false;
+    }
+
+    AZ::IO::Path MetadataManager::ToMetadataPath(AZ::IO::PathView file)
```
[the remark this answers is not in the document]
amzn-mike: PathView is just a string_view which is normally passed by copy

Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/MetadataManager.h
```diff
@@ -0,0 +1,68 @@
...
+        AZ_RTTI(IMetadataRequests, "{88EB4C71-F2A5-48C6-A08B-0D6A8EB08A74}");
+
+        virtual ~IMetadataRequests() = default;
+
+        //! Gets a value from the metadata file associated with the file.
+        //! @param file Absolute path to the file (or metadata file).
+        //! @param key JSONPath formatted key for the metadata value to read.  Ex: /Settings/Platform/pc.
+        //! @param typeId Type of the stored value and outValue.
+        //! @return True if metadata file and key exists and was successfully read, false otherwise.
+        virtual bool Get(AZ::IO::PathView file, AZStd::string_view key, void* outValue, AZ::Uuid typeId) = 0;
```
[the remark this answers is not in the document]
amzn-mike: These are modeled after the settings registry SetObject/GetObject functions, which also take a void* and a TypeId.  I noticed the registry has templated helper versions of these which avoids the need to call azrtti_typeid directly, so I've added those in.

The intent for this was that users would, in c++, have a concrete object type that they're saving/loading so they'd know what type they're working with.  We could always add additional APIs which allow reading a key back as a JSON object so the user can work with it manually, but I think we should wait until there's a usecase for that, as I'm doubting how useful that would be here.

amzn-mike:
> Do we have a json schema or at least versioning system in place for the metadata file?
> 
> I believe this is part of the asset relocation work, so consider linking to RFC.

@lmbr-pip A FileVersion field is added to every metadata file written out.  There's no code reading it currently as there's no versioning work to do yet, but if we make some change in the future we should be able to read that version field and do whatever conversion is needed on load.

Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/MetadataManager.h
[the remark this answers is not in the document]
amzn-mike: After thinking about this more, we need the ability to read the JSON object itself so users can handle reading old versions themselves.  I've updated the PR to add this as well as injecting a version field into the serialized type.

Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/MetadataManager.h
```diff
@@ -0,0 +1,101 @@
...
+        //! @param key JSONPath formatted key for the metadata value to read.  Ex: /Settings/Platform/pc.
+        //! @return True if metadata file and key exists and was successfully read, false otherwise.
+        virtual bool GetValueVersion(AZ::IO::PathView file, AZStd::string_view key, int& version) = 0;
+
+        //! Sets a value in the metadata file associated with the file.
+        //! @param file Absolute path to the file (or metadata file).
+        //! @param key JSONPath formatted key for the metadata value to write.  Ex: /Settings/Platform/pc.
+        //! @param typeId Type of the stored value and inValue.
+        //! @return True if the value is successfully saved to disk, false otherwise.
+        virtual bool SetValue(AZ::IO::PathView file, AZStd::string_view key, const void* inValue, AZ::Uuid typeId) = 0;
```
junbo75, line 58: I'm wondering whether the API should allow to set the value version as well. If all the key/value pairs share the same version, why don't we just keep the version of the metadata file itself?

amzn-mike: Discussed offline - There was confusion between the 2 versions, I've updated the comments to try and make it more clear.  The metadata file version is appended to the document as its own key.  Each key/value pair has its type's version appended to it to allow some rough version conversion.  The version isn't per field, so an update to a "type within a type" will not show a different version.

jackalbe: I read this "GetValueVersion()" and got the idea that the version was attached to the "Value" but if the value is Bool or Float how would that work? If the Value is a TypeT, then that sort of makes sense.

jackalbe (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/MetadataManager.h
amzn-mike: The version is only added for types which serialize as Json Objects.  So simple types like bool wouldn't have a version attached

PR #13180, opened 2022-11-14
Build fixes for Android Unit Tests

amzn-phist:
## What does this PR do?

Introduces an updated software stack for Android builds that uses recent versions of development software:
- JDK 11
- Android Gradle Plugin 7.3.1
- Gradle 7.5.1
- NDK 25
- Build-Tools 33.0.0

Fixes issues with building unit tests for Android:
- When supplying the `--unit-test` argument to `generate_android_project.py`, it wasn't properly specifying a CMake target, which resulted in building everything and ultimately would fail during APK packaging.
- Fixes the dependent targets that are generated when specifying a `TEST_SUITE` target.  LYWrappers was only including the AzTestRunner target and not the suite of test modules that make up the suite.
- 

## How was this PR tested?

Building Android Unit Tests locally and running them on a device.  The unit tests themselves aren't passing, but at least we can get logs and start down that path.


amzn-phist:
Sorry if this spammed a ton of people, let me fix it.

Issue #13080, opened 2022-11-09
Action Manager: An Editor crash can occur if levels are changed while the Paint Brush Settings tool is opened


jckand-amzn:
**Describe the bug**
An Editor crash can occur if levels are changed while the Paint Brush Settings tool is opened while the new ActionManager is enabled. Without the ActionManager, the user does not have the option to open a level while in Component Edit mode.

**Prerequisites**
1. Editor is launched with the new ActionManager enabled.

```
Editor.exe --regset=O3DE/ActionManager/EnableNewActionManager=true
```

**Steps to reproduce**
1. Open a new or existing level.
3. Create a new entity with an Image Gradient component and all dependencies.
4. Assign an asset to the Image Gradient component, and click the Edit button on the component card to bring up the Paint Brush Settings tool.
5. Click File > Open level and open any different level without saving any changes.

**Expected behavior**
Paint Brush Settings tool is closed without saving any changes.

**Actual behavior**
Paint Brush Settings tool is closed, however the Viewport remains in Component Edit mode, and the newly opened level cannot be interacted with. After a few seconds, the Editor crashes.

```
Call Stack Trace:
 7) 00007FFD4CFB02E7 (KERNELBASE) : UnhandledExceptionFilter
 6) 00007FFD4F495530 (ntdll) : memset
 5) 00007FFD4F47C876 (ntdll) : _C_specific_handler
 4) 00007FFD4F49241F (ntdll) : _chkstk
 3) 00007FFD4F4414A4 (ntdll) : RtlRaiseException
 2) 00007FFD4F490F4E (ntdll) : KiUserExceptionDispatcher
 1) D:\GitHub\o3de\Gems\GradientSignal\Code\Source\Editor\EditorImageGradientComponentMode.cpp (275) : GradientSignal::PaintBrushUndoBuffer::Changed
```

**Found in Branch**
development@ 1498390c98f7836c343130f7c323192151ce12bb

**Desktop/Device:**
 - Device: PC
 - OS: Windows 10 21H1
 - CPU: Ryzen 5950x
 - GPU: Nvidia 2070
 - Memory: 64GB

AMZN-daimini:
@mbalfour-amzn @hultonha Isn't paintbrush mode a Component Mode? If not, should it be?

hultonha:
[...]

PR #13187, opened 2022-11-15
Action Manager | Use ActionContextModes to support disabling actions when a Component Mode is started

AMZN-daimini:
## What does this PR do?

[GIF WIP...]

This PR adds a handler class to the Component Mode setup that flips the ActionContextMode from default (regular editing) to "AnyComponentMode" (as stated, used for any other component mode). This is the first step to migrate component mode action handling to use the new Action Manager system.

The changes include:
- Added `ComponentModeStarted` and `ComponentModeEnded`  to the `EditorComponentModeNotificationBus`. Used to notify systems when we're entering any component mode.
- Added a new handler class called `ComponentModeActionHandler` to the ComponentModeCollection. Right now, it only sets up to listen to `ActionManagerRegistrationNotificationBus` to do Action Manager registrations at startup, and to `EditorComponentModeNotificationBus` to ensure the ActionContextMode is switched correctly between `default` (for regular editing) and `AnyComponentMode` for any special Component Mode.
- All existing actions in the File menu were updated to only be available when the ActionContextMode is set to `default`. This will ensure that whenever the `ComponentModeActionHandler` flips the mode to something that isn't `default`, all of these actions will be inaccessible. Some of the actions were set to still be displayed even when disabled to mirror the behavior in the legacy systems.
- `ComponentModeActionHandler` adds a new action ("Done") to close the current Component Mode and return to the regular editing.

Notes:
- Some actions that were disabled in the current implementation were intentionally left enabled in the new one as they seemed to still be useful for user interaction without causing trouble. For example, access to viewport settings like helpers, debug text and camera position.
- In the previous system, most sub-menus were manually set to disabled during Component Mode. The current Action Manager implementation doesn't provide such functionality; so sub-menus are shown as enabled, but then all elements inside them are disabled. This sounds like a good standard to me so I'm not too concerned by the discrepancy, but I'll bring it up during UX reviews.

I will work on a follow-up PR as soon as this goes in, which will add:
- The remaining actions shared by all Component Modes that allow users to switch between the component modes on the current selection seamlessly;
- A system that retrieves all Component Modes registered at Editor startup and maps them to specific ActionContextModes automatically (which will remove the `AnyContextMode` label)
- A mechanism that allows systems to add actions to specific Component Modes, migrating the existing use cases to adopt this new setup.

## How was this PR tested?

Manual testing to verify the new implementation matches the requirements of the existing one.


AMZN-daimini:
Fixes #13080

hultonha:
@AMZN-daimini I haven't looked at the review yet but after reading the description and seeing the first bullet point, can you please use...

```
OnEditorModeActivated(...
    if (mode == AzToolsFramework::ViewportEditorMode::Component)
OnEditorModeDeactivated(...)

// see ViewportEditorModeTrackerNotificationBus.h for details
```

Instead of adding a new `ComponentModeStarted/Ended` event please - we're using that Bus for all Editor wide state changes including Component Mode (tagging @jonawals for visibility too as he as more info about it). Let me know if this will cause any issues. Thanks!

Code/Editor/Core/EditorActionsHandler.cpp
```diff
@@ -347,6 +367,9 @@ void EditorActionsHandler::OnActionRegistrationHook()
                 UpdateRecentFileActions();
             }
         );
+
+        // This action is only accessible outside of Component Modes
+        m_actionManagerInterface->AssignModeToAction(AzToolsFramework::DefaultActionContextModeIdentifier, actionIdentifier);
```
hultonha, line 372: question: Might it be possible to make this the default and people opt out of this behavior as opposed to the other way round? I guess it depends on if there are more things that should be accessed outside component modes or inside (my thought was always there's more things that are only accessible outside, but I could be wrong, I'd be interested to hear your thoughts).

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeActionHandler.cpp
```diff
@@ -0,0 +1,132 @@
...
+{
+    ComponentModeActionHandler::ComponentModeActionHandler()
+    {
+        if (IsNewActionManagerEnabled())
+        {
+            AzFramework::EntityContextId editorEntityContextId = AzFramework::EntityContextId::CreateNull();
+            EditorEntityContextRequestBus::BroadcastResult(editorEntityContextId, &EditorEntityContextRequests::GetEditorEntityContextId);
+
+            ActionManagerRegistrationNotificationBus::Handler::BusConnect();
+            ComponentModeFramework::EditorComponentModeNotificationBus::Handler::BusConnect(editorEntityContextId);
```
hultonha, line 33: update: This should instead likely be `ViewportEditorModeNotifications` and not specifically `EditorComponentModeNotificationBus`

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeActionHandler.cpp
```diff
@@ -0,0 +1,132 @@
...
+                actionProperties,
+                []
+                {
+                    ComponentModeFramework::ComponentModeSystemRequestBus::Broadcast(
+                        &ComponentModeFramework::ComponentModeSystemRequests::EndComponentMode);
+                }
+            );
+
+            // Only add this to the component modes, not default.
+            m_actionManagerInterface->AssignModeToAction("SomeMode", actionIdentifier);
```
hultonha, line 91: question: What does `"SomeMode"` mean here in this context?

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeActionHandler.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+            m_hotKeyManagerInterface->SetActionHotKey(actionIdentifier, "Esc");
```
hultonha, line 93: question: One thing I'd like to know is does this still allow actions to 'stack' - so in the example of the Spline Component, if you enter Component Mode and have no vertices selected, and press Esc, you leave Component Mode. However if you have a vertex selected, and press Esc, if will first deselect the Vertex, and then pressing Esc again will leave Component Mode. This might not be in this change but having the ability for shortcuts to dynamically change based on the current state of the mode is super important and I just want to double check we can still support that with the new system. Thanks!

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeActionHandler.cpp
```diff
@@ -0,0 +1,132 @@
...
+        m_menuManagerInterface = AZ::Interface<AzToolsFramework::MenuManagerInterface>::Get();
+        AZ_Assert(
+            m_menuManagerInterface,
+            "ComponentModeActionHandler - could not get MenuManagerInterface on ComponentModeActionHandler OnMenuBindingHook.");
+
+        m_menuManagerInterface->AddSeparatorToMenu(EditMenuIdentifier, 10000);
+        m_menuManagerInterface->AddActionToMenu(EditMenuIdentifier, "o3de.action.componentMode.end", 10001);
+    }
+
+    void ComponentModeActionHandler::ComponentModeStarted([[maybe_unused]] const AZ::Uuid& componentType)
```
hultonha, line 108: update: I believe this can be replaced with... `OnEditorModeActivated` on the `ViewportEditorModeNotificationsBus`

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeActionHandler.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+    void ComponentModeActionHandler::ActiveComponentModeChanged([[maybe_unused]] const AZ::Uuid& componentType)
+    {
+        m_actionManagerInterface->SetActiveActionContextMode(EditorMainWindowActionContextIdentifier, "AnyComponentMode");
+
+        // Update Component Mode Changed updater
+        m_actionManagerInterface->RegisterActionUpdater(ComponentModeChangedUpdaterIdentifier);
+    }
+
+    void ComponentModeActionHandler::ComponentModeEnded()
```
hultonha, line 124: update: Same as previous comment

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeActionHandler.h
```diff
@@ -0,0 +1,47 @@
...
+#include <AzToolsFramework/ActionManager/ActionManagerRegistrationNotificationBus.h>
+#include <AzToolsFramework/ComponentMode/EditorComponentModeBus.h>
+
+namespace AzToolsFramework
+{
+    class ActionManagerInterface;
+    class MenuManagerInterface;
+    class HotKeyManagerInterface;
+
+    class ComponentModeActionHandler
```
hultonha, line 22: update: It would definitely be useful to add an API comment for this new type

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeCollection.cpp
```diff
@@ -216,6 +216,12 @@ namespace AzToolsFramework
             m_componentModeState = ComponentModeState::Active;
             m_adding = false;
 
+            // notify other systems ComponentMode has started
+            EditorComponentModeNotificationBus::Event(
+                GetEntityContextId(),
+                &EditorComponentModeNotifications::ComponentModeStarted,
+                m_activeComponentTypes[m_selectedComponentModeIndex]);
+
```
hultonha, line 224: update: This can be removed, see `m_viewportEditorModeTracker->ActivateMode({ GetEntityContextId() }, ViewportEditorMode::Component);` below which is already sending the notification

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeCollection.cpp
```diff
@@ -295,6 +301,12 @@ namespace AzToolsFramework
             // wish to modify state to indicate this (e.g. appearance, functionality etc.)
             m_viewportEditorModeTracker->DeactivateMode({ GetEntityContextId() }, ViewportEditorMode::Component);
 
+            // notify other systems ComponentMode has ended
+            EditorComponentModeNotificationBus::Event(
+                GetEntityContextId(),
+                &EditorComponentModeNotifications::ComponentModeEnded
+            );
```
hultonha, line 308: update: See comment above and `m_viewportEditorModeTracker->DeactivateMode({ GetEntityContextId() }, ViewportEditorMode::Component);` call

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeCollection.cpp
```diff
@@ -461,7 +473,7 @@ namespace AzToolsFramework
 
                 if (m_selectedComponentModeIndex < m_activeComponentTypes.size())
                 {
-                    // notify other systems ComponentMode actions have changed
+                    // notify other systems ComponentMode has changed
```
hultonha, line 476: nit: I think here we could say `// notify other systems the active ComponentMode has changed` to be more precise (in the case we can have more than one enabled but not active at a time such as the Spline and Tube Shape)

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeCollection.h
```diff
@@ -116,6 +117,8 @@ namespace AzToolsFramework
             /// Editor (global) ComponentMode state - is ComponentMode active or not.
             ComponentModeState m_componentModeState = ComponentModeState::Stopped; 
             ViewportEditorModeTrackerInterface* m_viewportEditorModeTracker = nullptr; //!< Tracker for activating/deactivating viewport editor modes.
+
+            ComponentModeActionHandler m_componentModeActionHandler;
```
hultonha, line 121: nit: Brief member variable comment here would be useful

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/EditorComponentModeBus.h
```diff
@@ -241,10 +241,16 @@ namespace AzToolsFramework
...
 
+            /// Called when a Component Mode is started.
+            virtual void ComponentModeStarted(const AZ::Uuid& /*componentType*/) {}
+
             /// Called when Tab is pressed to cycle the 'selected' ComponentMode (which shortcuts/actions are active).
             /// Also called when directly selecting a Component in the EntityOutliner.
             virtual void ActiveComponentModeChanged(const AZ::Uuid& /*componentType*/) {}
 
+            /// Called when Component Mode is ended.
+            virtual void ComponentModeEnded() {}
```
hultonha, line 252: update: Please remove these calls and use the other notification bus mentioned earlier

Code/Framework/AzToolsFramework/AzToolsFramework/ViewportSelection/EditorTransformComponentSelection.cpp
```diff
@@ -2921,6 +2949,9 @@ namespace AzToolsFramework
                 }
             );
 
+            // This action is only accessible outside of Component Modes
+            m_actionManagerInterface->AssignModeToAction(DefaultActionContextModeIdentifier, actionIdentifier);
```
hultonha, line 2953: update: Again I'm wondering about this just being the default and we instead opt in to actions we do want in Component Mode instead - it feels like there would be less chance of error that way around

hultonha (commented):
Changes look good! I just have one requested update (see inline comments) and some suggestions/questions. Otherwise looking like an excellent step in the right direction. I know getting Component Mode moved over to the new system is a big challenge so awesome work getting this far 👍 Let me know if there's anything else I can help with.

AMZN-daimini:
> @AMZN-daimini I haven't looked at the review yet but after reading the description and seeing the first bullet point, can you please use...
> 
> ```
> OnEditorModeActivated(...
>     if (mode == AzToolsFramework::ViewportEditorMode::Component)
> OnEditorModeDeactivated(...)
> 
> // see ViewportEditorModeTrackerNotificationBus.h for details
> ```
> 
> Instead of adding a new `ComponentModeStarted/Ended` event please - we're using that Bus for all Editor wide state changes including Component Mode (tagging @jonawals for visibility too as he as more info about it). Let me know if this will cause any issues. Thanks!

Honestly, I avoided that cause it looked explicitly made for viewport visualization modes. I can change the code to use it, but it felt weird.

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeActionHandler.h
AMZN-daimini: Done!

hultonha:
> Honestly, I avoided that cause it looked explicitly made for viewport visualization modes. I can change the code to use it, but it felt weird.

@AMZN-daimini No it's definitely intended more for central editor state tracking and it's just the viewport that happens to use it at the moment - the naming probably should be updated to make it clearer though (might be worth @jonawals weighing in on this too if he has any opinions)

Code/Editor/Core/EditorActionsHandler.cpp
hultonha: Thanks for the feedback - totally fine to leave as-is and we can always revisit if needed. There are upsides and downsides to either approach and this can be useful in a number of situations. We were overly cautious with Component Mode as we had seen how bad things could get in the old system with lots of competing systems left 'on' inadvertently that would compete for inputs and logic and cause all kinds of weird bugs. This meant we maybe turned off more that was 'technically' allowed but it did mean Component Mode was more isolated by default. The drawback was some operations that might have been useful at times were off limits (our justification was as it was quick and easy to leave Component Mode it wasn't a big deal) but I can see it being an irritation for customers at times too.

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeActionHandler.cpp
hultonha: For this I'd listen to `ActiveComponentModeChanged` on the `EditorComponentNotificationBus`, you might need that and the other I mentioned but you shouldn't need to add a new notification (just listen for both of those)

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeActionHandler.cpp
hultonha: Okay thanks Danilo, using the Spline Component as an example of how we want things to work would be great (we have similar behavior in White Box too) where Esc acts as 'back' wherever you are in the current 'stack' of operations and it definitely matches with what uses expect should happen. Having 'Esc' always leave Component Mode actually feels horrible even if in theory it seems to make sense.

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeActionHandler.cpp
hultonha: Thanks for updating

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/EditorComponentModeBus.h
hultonha: Thanks!

Code/Framework/AzToolsFramework/AzToolsFramework/ViewportSelection/EditorTransformComponentSelection.cpp
hultonha: +1 yeah, maybe another convo to follow-up with UX about. In the short-term I'd be up for keeping things consistent with how they are now at the very least, and we can then relax what is available as a separate change so we're not changing multiple things at once

hultonha (approved):
Thanks for making the updates @AMZN-daimini, changes look great 👍

Code/Framework/AzToolsFramework/AzToolsFramework/ViewportSelection/EditorTransformComponentSelection.cpp
AMZN-daimini: Yeah I'm usually just trying to keep things as they are apart from minor things, and record what looks weird for future conversation.

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeActionHandler.cpp
AMZN-daimini: Forgive me, I'm a little confused now :D I have searched for `EditorComponentNotificationBus` in the whole solution and I get no hits. Do you mean `EditorComponentModeNotificationBus`? Cause if that's the case, current implementation is using that already (and I have reverted the addition of the new notifications)

gadams3 (approved):

cgalvan (approved):

PR #13210, opened 2022-11-15
Add support for reverting overrides on an entity

srikappa-amzn:
## What does this PR do?

This PR adds support for reverting overrides on an entity. Changes include:
1. Providing outliner menu option
2. PrefabOverrideHandler and PrefabOverridePublicHandler methods to support revert overrides
3. New Prefab::Link methods to add and remove overrides from a link
4. Undo/Redo support.
5. Unit tests

RFC: https://github.com/o3de/sig-content/issues/75
Here's a video of the feature in action:

https://user-images.githubusercontent.com/82230713/202017562-97c10a01-49aa-4ab8-ac29-deb4816e9e72.mp4

## How was this PR tested?

Tested manually in the editor. Also added new unit tests.


Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabIntegrationManager.cpp
```diff
@@ -59,6 +60,7 @@ namespace AzToolsFramework
         PrefabFocusInterface* PrefabIntegrationManager::s_prefabFocusInterface = nullptr;
         PrefabFocusPublicInterface* PrefabIntegrationManager::s_prefabFocusPublicInterface = nullptr;
         PrefabLoaderInterface* PrefabIntegrationManager::s_prefabLoaderInterface = nullptr;
+        PrefabOverridePublicInterface* PrefabIntegrationManager::s_prefabOverridePublicInterface = nullptr;
```
lumberyard-employee-dm, line 63: When are these static variables cleaned up?

Actually why are static variables being used here and not member variables of the PrefabIntegrationManager?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Link/Link.cpp
```diff
@@ -135,6 +135,16 @@ namespace AzToolsFramework
             return areOverridesPresent;
         }
 
+        AZ::Dom::DomPrefixTree<Link::PrefabOverrideMetadata> Link::RemoveOverrides(AZ::Dom::Path path)
+        {
+            return AZStd::move(m_linkPatchesTree.DetachSubTree(path));
```
lumberyard-employee-dm, line 140: The move inhibits RVO and is not needed.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabIntegrationManager.cpp
srikappa-amzn: > When are these static variables cleaned up?

I believe the the static variables get cleaned up when the application terminates. 

> Actually why are static variables being used here and not member variables of the PrefabIntegrationManager?

I've kept up with the existing pattern of declaring interface pointers static in the class. I'll try to make these non-static and see if anything breaks. I'm not sure if there is a particular reason any of these interfaces are declared static. Maybe @AMZN-daimini knows?

srikappa-amzn: I've made the interface to be a non-static member it works as expected. There are other interfaces in the class that are declared static as well that I'll discuss with @AMZN-daimini and modify in a later PR if needed

lumberyard-employee-dm (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabIntegrationManager.cpp
AMZN-daimini: I am trying to remember/dig out the reasoning behind why that pattern was established but I haven't really found a reason yet. I'll let you know if I figure it out.
Right now, though, I have no objections against turning them into member variables.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverrideHandler.cpp
```diff
@@ -26,5 +28,35 @@
             }
             return false;
         }
+
+        bool PrefabOverrideHandler::RevertOverrides(AZ::Dom::Path path, LinkId linkId) const
+        {
+            PrefabSystemComponentInterface* prefabSystemComponentInterface = AZ::Interface<PrefabSystemComponentInterface>::Get();
+            if (prefabSystemComponentInterface != nullptr)
```
AMZN-daimini, line 35: Since it seems to be used in multiple functions, could this be turned into a member and initialized once?

AMZN-daimini (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverrideHandler.cpp
```diff
@@ -26,5 +28,35 @@
...
+
+                    ScopedUndoBatch undoBatch("Revert Prefab Overrides");
+                    PrefabUndoRevertOverrides* state = new Prefab::PrefabUndoRevertOverrides("Capture Override SubTree");
+                    state->Capture(path, AZStd::move(subTree), linkId);
+                    state->SetParent(undoBatch.GetUndoBatch());
+                    
+                    link->get().UpdateTarget();
+                    prefabSystemComponentInterface->PropagateTemplateChanges(link->get().GetTargetTemplateId());
+                    AzToolsFramework::ToolsApplicationRequestBus::Broadcast(
+                        &AzToolsFramework::ToolsApplicationRequestBus::Events::ClearDirtyEntities);
```
forkercat, line 55: why do we need to clear dirty entities here?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverridePublicHandler.cpp
```diff
@@ -50,16 +70,15 @@ namespace AzToolsFramework
...
-                InstanceOptionalReference overriddenInstance = focusedInstance->get().FindNestedInstance(overriddenInstanceKey);
-                if (overriddenInstance.has_value())
+                AZStd::string_view topMostInstanceKey = absoluteEntityAliasPath[1].GetKey().GetStringView();
+                InstanceOptionalReference topMostInstance = focusedInstance->get().FindNestedInstance(topMostInstanceKey);
+                if (topMostInstance.has_value())
                 {
                     auto pathIterator = absoluteEntityAliasPath.begin() + 2;
-                    AZ::Dom::Path modifiedPath(pathIterator, absoluteEntityAliasPath.end());
-                    return m_prefabOverrideHandler.AreOverridesPresent(modifiedPath, overriddenInstance->get().GetLinkId());
+                    return AZStd::pair(AZ::Dom::Path(pathIterator, absoluteEntityAliasPath.end()), topMostInstance->get().GetLinkId());
```
forkercat, line 78: I think I tried doing the similar thing before, but eventually I ended up calling ClimbUp first and GetRelativePath from climbed instances and then retrieve the link id from the climbed instances.

And can this logic be moved inside InstanceToTemplatePropagator?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverridePublicInterface.h
```diff
@@ -25,6 +25,11 @@ namespace AzToolsFramework
             //! @param entityId The id of the entity to check for overrides.
             //! @return true if overrides are present on the given entity id.
             virtual bool AreOverridesPresent(AZ::EntityId entityId) = 0;
+
+            //! Revert overrides on a entity id. Returns false if no overrides are present on the entity.
```
forkercat, line 29: nit: an

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoRevertOverrides.h
```diff
@@ -0,0 +1,40 @@
...
+        AZ_CLASS_ALLOCATOR(PrefabUndoRevertOverrides, AZ::SystemAllocator, 0);
+
+        explicit PrefabUndoRevertOverrides(const AZStd::string& undoOperationName);
+
+        bool Changed() const override;
+        void Undo() override;
+        void Redo() override;
+
+        void Capture(
+            const AZ::Dom::Path& pathToSubTree, AZ::Dom::DomPrefixTree<Link::PrefabOverrideMetadata>&& m_overrideSubTree, LinkId linkId);
```
forkercat, line 31: maybe we can start using alias name PrefabOverridesPrefixTree for AZ::Dom::DomPrefixTree<Link::PrefabOverrideMetadata>?

forkercat (approved):

forkercat:
Looks good! From the video, one thing I would try to verify: after undo/redo hit save and check if the override patches in focused template file gets updated as expected.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverridePublicHandler.cpp
srikappa-amzn: I don't think this belongs in the InstanceToTemplatePropagator because it tries to do something very specific with the fetched path, i.e removing the first 2 path tokens and fetching the linkId from there. I think we should avoid polluting the InstanceToTemplatePropagator with a lot of small helper functions like this. If it turns out that other override workflows also need this, we can revisit this then. But for now, I think it's best to have util/helper functions as private members instead of public interface methods.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverrideHandler.cpp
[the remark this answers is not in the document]
srikappa-amzn: It is to prevent any dirty entities from entering the CreateUndosForDirtyEntities() function when the scoped undo node is destroyed. At the end of this function, the only thing that should be modifying the prefab templates is the RevertOverrides() code itself. 

Although, all of this is prior to this PR: https://github.com/o3de/o3de/pull/13127 . So I'll try to remove that and see if it is still needed.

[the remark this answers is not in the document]
srikappa-amzn: Removed this as it no longer looks like its is needed. Probably fixed by the above PR

srikappa-amzn:
> Looks good! From the video, one thing I would try to verify: after undo/redo hit save and check if the override patches in focused template file gets updated as expected.

Good callout! I was actually able to find a bug about not marking the template dirty with this testing.

PR #13213, opened 2022-11-15
[Terrain] PaintBrush bugfixes and code cleanup

mbalfour-amzn:
## What does this PR do?

A few small bugfixes and cleanups:
* Split the paint brush SubMode Cluster into a separate generic file so that they can be easily reused across any component using a paintbrush. (Also helps to reduce the testable surface area of EditorImageGradientComponentMode)
* Prevent spurious error message from printing when an Image Asset is cleared on an Image Gradient.
* Make the Image Gradient correctly register/unregister the ComponentMode only when it is visible and available so that the Component Mode Switcher only provides a switcher button to the Image Gradient Component Mode at the right times.
* Track whether or not we've made any changes while editing so that we can skip prompting the user for a save on exit if no changes were made. It's very naive tracking right now, so changing a pixel to the same value that it was still counts as a change.

## How was this PR tested?

Tried out each of the cases above in-Editor and made sure they worked.


cgalvan (approved):

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponent.cpp
```diff
@@ -372,7 +373,7 @@ namespace GradientSignal
 
     void EditorImageGradientComponent::EnableComponentMode()
     {
-        if (m_componentModeDelegate.AddedToComponentMode())
+        if (m_componentModeDelegate.IsConnected())
```
hultonha, line 376: Excellent you can use this new API now, hopefully it's doing what you need 👍

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponentMode.h
```diff
@@ -84,17 +82,17 @@ namespace GradientSignal
...
-        AzToolsFramework::ViewportUi::ClusterId m_paintBrushControlClusterId;
-        AzToolsFramework::ViewportUi::ButtonId m_paintModeButtonId;
-        AzToolsFramework::ViewportUi::ButtonId m_eyedropperModeButtonId;
-        AzToolsFramework::ViewportUi::ButtonId m_smoothModeButtonId;
+        //! Track whether or not anything has changed while editing. If not, then don't prompt to save the image at the end.
+        bool m_anyValuesChanged = false;
 
-        AZ::Event<AzToolsFramework::ViewportUi::ButtonId>::Handler m_buttonSelectionHandler;
+        //! The paint brush cluster that manages switching between paint/smooth/eyedropper modes
+        AzToolsFramework::PaintBrushSubModeCluster m_subModeCluster;
```
hultonha, line 96: Nice refactor moving this out to its own type 👍

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponent.cpp
mbalfour-amzn: Yep it's perfect, thanks!

hultonha: Awesome to hear 😄  Great work @amzn-ahmadkrm

PR #13215, opened 2022-11-15
Refactor ScriptCanvas_ChangingAssets_ComponentStable and remove Hydra dependancy

michleza:
Signed-off-by: Michael <michleza@amazon.com>

## What does this PR do?

- migrated test script's dependency away from hydra
- updated the constructor of the script canvas component class
- updated the sc component class setters to just perform their setting/initialization function.
- included the script canvas component class in the component validation file
- updated the component validation file to support script canvas component

## How was this PR tested?
python testrunner and O3DE python console


AutomatedTesting/Gem/PythonTests/scripting/scripting_utils/script_canvas_component.py
```diff
@@ -65,6 +61,22 @@ def create_new_entity_with_component(self, entity_name: str, sc_file_path: str,
         self.script_canvas_component = self.editor_entity.get_components_of_type([SCRIPT_CANVAS_UI])[0]
         self.script_canvas_component.set_component_property_value(SCRIPT_CANVAS_COMPONENT_PROPERTY_PATH, sourcehandle)
 
+    def change_component_graph_file(self, sc_file_path: str):
```
FuzzyCarterAWS, line 64: Rename to `set_componenet_graph_file`

AutomatedTesting/Gem/PythonTests/scripting/scripting_utils/script_canvas_component.py
```diff
@@ -36,11 +36,7 @@ class VariableState(Enum):
 
 class ScriptCanvasComponent:
```
FuzzyCarterAWS, line 37: Let's relocate this file to the `editor component behaviors` in _AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/_

Additionally, we may want to refactor to follow the get/set properties then add the custom validators.

AutomatedTesting/Gem/PythonTests/scripting/scripting_utils/script_canvas_component.py
```diff
@@ -65,6 +61,22 @@ def create_new_entity_with_component(self, entity_name: str, sc_file_path: str,
...
+        returns none
+        """
+        sourcehandle = scriptcanvas.SourceHandleFromPath(sc_file_path)
+        old_value = self.script_canvas_component.get_component_property_value(SCRIPT_CANVAS_COMPONENT_PROPERTY_PATH)
+        self.script_canvas_component.set_component_property_value(SCRIPT_CANVAS_COMPONENT_PROPERTY_PATH, sourcehandle)
+
+        set_value = self.script_canvas_component.get_component_property_value(SCRIPT_CANVAS_COMPONENT_PROPERTY_PATH)
+        assert set_value != old_value and set_value is not None, f"Graph file could not be set! {sc_file_path} not found."
+
+
```
FuzzyCarterAWS, line 79: Let's refactor this into using the model I've set in `editor component behaviors`

FuzzyCarterAWS (changes requested):
Let's refactor and relocate the `script_canvas_component` and impacted classes according to the pattern I've set in the code I recently merged to ` origin/development `

AutomatedTesting/Gem/PythonTests/scripting/scripting_utils/script_canvas_component.py
michleza: renamed this function. Also updated the script_canvas_component field to be a list since I looked ahead in the other tests and noticed a few that use multi-component entities.

AutomatedTesting/Gem/PythonTests/scripting/scripting_utils/script_canvas_component.py
michleza: updated the class to use the validation file

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_script_canvas.py
```diff
@@ -93,18 +104,19 @@ def set_variable_value(self, variable_name: str, variable_state: VariableState,
...
 
-        #validate the change
-        set_value = self.script_canvas_component.get_component_property_value(component_property_path)
-        assert set_value == variable_value, f"Component variable {variable_name} was not set properly"
+        from editor_python_test_tools.editor_component.editor_component_validation import (
+            validate_script_canvas_variable_changed)
 
-    def __construct_variable_component_property_path(self, variable_name: str, variable_state: VariableState) -> str:
+        validate_script_canvas_variable_changed(self.get_script_canvas_components, component_property_path,
+                                                variable_name, variable_value, component_index)
```
FuzzyCarterAWS, line 117: As discussed remove the validation from the setters and move them to the test steps.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_script_canvas.py
```diff
@@ -62,14 +59,27 @@ def create_new_entity_with_component(self, entity_name: str, sc_file_path: str,
...
-                                sc_component_index=0) -> None:
+        returns none
         """
-        Function for constructing the SCComponent object. This uses an existing entity and loads a new script canvas
-        file into the source handle value
+        from editor_python_test_tools.editor_component.editor_component_validation import (
+            validate_script_canvas_graph_file)
+
+        sourcehandle = scriptcanvas.SourceHandleFromPath(sc_file_path)
+        validate_script_canvas_graph_file(self.get_script_canvas_components, sourcehandle, component_index)
```
FuzzyCarterAWS, line 78: Validators do not belong in the setters and breaks the pattern set up in the the other Editor Component Behaviors model.

We should not have diverging patterns of development in these libraries.

FuzzyCarterAWS (changes requested):

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_script_canvas.py
michleza: I made an attempt at putting the validation calls in the test scripts. The setters now return some data that the client can use to call the validation functions

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
```diff
@@ -12,7 +12,9 @@
 from editor_python_test_tools.editor_entity_utils import EditorComponent
 from consts.general import ComponentPropertyVisibilityStates as PropertyVisibility
 from editor_python_test_tools.asset_utils import Asset
-
+from editor_python_test_tools.editor_component.editor_script_canvas import (ScriptCanvasComponent,
```
FuzzyCarterAWS, line 15: ScriptCanvasComponent & VariableState appear to be unused.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
```diff
@@ -215,3 +217,41 @@ def validate_asset_property(property_name: str, get_asset_value: typing.Callable
     assert expected_asset.id.is_equal(expected_asset.id, set_value), \
         f"Error: The {component_name}'s {property_name} property failed to properly set the mesh. Asset Id: " \
         f"{expected_asset.id} was expected but Asset Id: {set_value} was retrieved from \"{expected_asset.get_path()}\""
+
+
+def validate_script_canvas_graph_file(get_script_canvas_components: typing.Callable, source_handle: str, old_value: str,
```
FuzzyCarterAWS, line 222: Validators should be both a setter & a getter action and follow the patterns. The description even implies this, but is only validating that something has been set. This will also prevent needing to pass in the old value and setting up for testing the behavior outside of the validator. The validator for asset properties are probably the best to compare this particular validator to.

Validator expectations:

- Takes a getter and setter method in (these will hold the reference to the correct component since determining which component to validate should be done outside the validator)
- Takes the asset to set (Path in this case? Or you can pass a source handle if you prefer that method)
- Guard conditions executed
- Builds the asset correctly if using the path
- Captures the currently set asset
- Sets the Asset
- Validates the asset has been set.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
```diff
@@ -215,3 +217,41 @@ def validate_asset_property(property_name: str, get_asset_value: typing.Callable
...
+    assert source_handle is not None, "File not found!"
+
+    sc_components = get_script_canvas_components()
+
+    set_value = sc_components[component_index].get_component_property_value(
+        SCRIPT_CANVAS_COMPONENT_PROPERTY_PATH)
+    assert set_value != old_value and set_value is not None, f"Graph file could not be set!"
+
+
+def validate_script_canvas_variable_changed(get_script_canvas_components: typing.Callable, component_property_path: str,
```
FuzzyCarterAWS, line 243: Update according to the standards defined above.

Additionally, make sure values passed in are typed, even if they're set to any.

https://docs.python.org/3/library/typing.html#the-any-type

Lastly, let's get this function to have that backlog item referenced where we talked about how this currently only works for specific data types and we need to think about our approach outside those data types.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_script_canvas.py
```diff
@@ -38,12 +37,10 @@ class ScriptCanvasComponent:
 
     def __init__(self):
         self.editor_entity = None
-        self.script_canvas_component = None
+        self.script_canvas_components = []
```
FuzzyCarterAWS, line 40: The script canvas component should only handle a single instance of the component. Overloading it like this puts some awkward behavior into the class that belongs to the editor entity class and not the component class. If somebody wants to add multiple script canvas components, it should not be be the the responsibility of the component class to care and should only have a reference to itself.

Please refactor this class to be closer to` editor_physx_collider.`

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_script_canvas.py
```diff
@@ -62,28 +59,46 @@ def create_new_entity_with_component(self, entity_name: str, sc_file_path: str,
         self.editor_entity = EditorEntity.create_editor_entity_at(position, entity_name)
         self.editor_entity.add_components([SCRIPT_CANVAS_UI])
 
-        self.script_canvas_component = self.editor_entity.get_components_of_type([SCRIPT_CANVAS_UI])[0]
-        self.script_canvas_component.set_component_property_value(SCRIPT_CANVAS_COMPONENT_PROPERTY_PATH, sourcehandle)
+        self.script_canvas_components = self.editor_entity.get_components_of_type([SCRIPT_CANVAS_UI])
+        self.script_canvas_components[0].set_component_property_value(SCRIPT_CANVAS_COMPONENT_PROPERTY_PATH, sourcehandle)
+
+    def set_component_graph_file(self, sc_file_path: str, component_index=0) -> tuple:
```
FuzzyCarterAWS, line 65: `set_component_graph_file_from_path`

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_script_canvas.py
```diff
@@ -62,28 +59,46 @@ def create_new_entity_with_component(self, entity_name: str, sc_file_path: str,
...
+        returns the source handle from the file path and the previous component property value for validation in a tuple
+        """
+        sourcehandle = scriptcanvas.SourceHandleFromPath(sc_file_path)
+
+        old_component_property_value = self.script_canvas_components[component_index].get_component_property_value(SCRIPT_CANVAS_COMPONENT_PROPERTY_PATH)
+
+        self.script_canvas_components[component_index].set_component_property_value(SCRIPT_CANVAS_COMPONENT_PROPERTY_PATH,
+                                                                    sourcehandle)
+
+        return sourcehandle, old_component_property_value
```
FuzzyCarterAWS, line 81: Remove `old_componenet_property_value`, this belongs in a getter.

FuzzyCarterAWS (changes requested):

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
michleza: VariableState is now used by the validation file.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
michleza: I think the validation functions now statisfy these expectations. both validation functions now take a set/get function as their parameters. They capture the old value we are testing by using the get function then compare it against the new value that is changed with the set function.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_script_canvas.py
michleza: component class now only manages a single script canvas component reference.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_script_canvas.py
michleza: updated the function name and changed what it returns since we don't need the old data anymore.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
michleza: added the Any type and referenced the QtPy class and GH issues where this is being worked on.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
```diff
@@ -215,3 +216,52 @@ def validate_asset_property(property_name: str, get_asset_value: typing.Callable
...
+    we also only support basic types and need to update associated QtPy classes to allow exposing the basic types to
+    the component class. See github GH-13344. We currently only support strings and boolean variables.
+
+
+    """
+    Report.info(f"Validating Script Canvas component's variable was set.")
+
+    old_variable_value = get_variable_value(variable_name, variable_state)
+
+    assert old_variable_value is not None, "Path to variable was invalid! Check use/unused state or variable name"
```
FuzzyCarterAWS, line 260: What happens here if the variable value was unset before this?

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
```diff
@@ -215,3 +216,52 @@ def validate_asset_property(property_name: str, get_asset_value: typing.Callable
...
+
+    set_variable_value(variable_name, variable_state, variable_value)
+
+    new_variable_value = get_variable_value(variable_name, variable_state)
+
+    assert old_variable_value != new_variable_value, f"Component variable {variable_name} was not set properly"
```
FuzzyCarterAWS, line 266: We should validate that new_variable_value matches the expected variable value passed in. Right now, if I ran the function twice I'd get a failure because it appears that it didn't change.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
```diff
@@ -215,3 +216,52 @@ def validate_asset_property(property_name: str, get_asset_value: typing.Callable
...
+    """
+    Function for validating that a script canvas component variable can be changed.
+
+    get_variable_value: getter for variables on the script canvas component
+    set_variable_value: setter for variables on the script canvas component
+    variable_name: the name of the variable to change
+    variable_state: whether the variable is initialized or not within graph file(initialized by the script canvas editor).
+    variable_value: the value to set the variable to. has no rigid type since variables can be primitives or user created types.
+    we also only support basic types and need to update associated QtPy classes to allow exposing the basic types to
+    the component class. See github GH-13344. We currently only support strings and boolean variables.
```
FuzzyCarterAWS, line 252: Remove reference to QtPy, this is not related to the QtPy libraries and is part of the editor component behavior libs.

Additionally, move the notes about the GHI to a separate line below `variable_value` and make the GHI the start of the line. Let's also keep things uniform and use `o3de/o3de#13344` over `GH-13344` as that is not the standard way to reference a GHI issue according to the GH docs: 

image

FuzzyCarterAWS (changes requested):

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
```diff
@@ -7,12 +7,13 @@
 
 import azlmbr.asset as AzAsset
 import azlmbr.math as Math
-
+from typing import Any
```
FuzzyCarterAWS, line 10: Blank line separating import methodologies.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
```diff
@@ -215,3 +216,51 @@ def validate_asset_property(property_name: str, get_asset_value: typing.Callable
...
+    Function to validate the setting of a script canvas graph file to the file source field in the script canvas component
+
+    get_script_canvas_component_graph_file: getter for the script canvas component file source handle field
+    set_script_canvas_component_graph_file: setter for the script canvas component's file source field.
+    sc_file_path: the path on disk where the graph file exists
+
+    """
+    Report.info(f"Validating Script Canvas component's file source field can be set.")
+
+    old_value = get_script_canvas_component_graph_file(SCRIPT_CANVAS_COMPONENT_PROPERTY_PATH)
```
FuzzyCarterAWS, line 233: This probably shouldn't pass in anything?

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_script_canvas.py
```diff
@@ -37,53 +36,46 @@ class VariableState(Enum):
...
-        Returns None
+        returns None
         """
         sourcehandle = scriptcanvas.SourceHandleFromPath(sc_file_path)
 
-        self.editor_entity = EditorEntity.create_editor_entity_at(position, entity_name)
-        self.editor_entity.add_components([SCRIPT_CANVAS_UI])
-
-        self.script_canvas_component = self.editor_entity.get_components_of_type([SCRIPT_CANVAS_UI])[0]
+        self.script_canvas_component = editor_entity.add_component(SCRIPT_CANVAS_UI)
```
FuzzyCarterAWS, line 55: Unneeded. 

The pattern used here is for Editor Component Behavior classes to be created by the `EditorEntity` and the returned `EditorComponent` to be consumed and set in the initializer. We should not be doing it the other way around as the logic puts the onus of entity management on class that does not manipulate the entity directly.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_script_canvas.py
```diff
@@ -37,53 +36,46 @@ class VariableState(Enum):
 class ScriptCanvasComponent:
 
     def __init__(self):
-        self.editor_entity = None
         self.script_canvas_component = None
```
FuzzyCarterAWS, line 39: Update the initializer to     

```
def __init__(self, editor_entity: EditorEntity) -> None:
        self.component = editor_entity.add_component(SCRIPT_CANVAS_COMPONENT)
```

FuzzyCarterAWS (changes requested):

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
michleza: that space doesn't exist in my files.... 
I will move the import and resave to see if changes take.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
michleza: yep. it used to take something but I guess giving parameters to a function that doesn't need them isn't an issue in python.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
```diff
@@ -215,3 +216,51 @@ def validate_asset_property(property_name: str, get_asset_value: typing.Callable
...
         f"Error: The {component_name}'s {property_name} property failed to properly set the mesh. Asset Id: " \
         f"{expected_asset.id} was expected but Asset Id: {set_value} was retrieved from \"{expected_asset.get_path()}\""
+
+
+def validate_script_canvas_graph_file(get_script_canvas_component_graph_file: typing.Callable,
+                                      set_script_canvas_component_graph_file: typing.Callable, sc_file_path: str) -> None:
+    """
+    Function to validate the setting of a script canvas graph file to the file source field in the script canvas component
+
+    get_script_canvas_component_graph_file: getter for the script canvas component file source handle field
```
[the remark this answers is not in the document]
michleza: renamed both functions to be a bit shorter

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
```diff
@@ -215,3 +216,51 @@ def validate_asset_property(property_name: str, get_asset_value: typing.Callable
...
+    sc_file_path: the path on disk where the graph file exists
+
+    """
+    Report.info(f"Validating Script Canvas component's file source field can be set.")
+
+    old_value = get_script_canvas_component_graph_file(SCRIPT_CANVAS_COMPONENT_PROPERTY_PATH)
+
+    set_script_canvas_component_graph_file(sc_file_path)
+
+    set_value = get_script_canvas_component_graph_file(SCRIPT_CANVAS_COMPONENT_PROPERTY_PATH)
```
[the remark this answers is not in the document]
michleza: same as before

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_script_canvas.py
michleza: removed this function and updated the constructor to require an editor entity to get the component object.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_script_canvas.py
michleza: ended up doing this exact thing after acting on the previous comment.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
michleza: this depends on the variable type. a string would look like it's None but is actually a blank string ""

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
michleza: I will update this to check against expected (new) value instead since this use case makes sense

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
michleza: updated this function docstring to match this formatting.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
```diff
@@ -215,3 +214,46 @@ def validate_asset_property(property_name: str, get_asset_value: typing.Callable
...
         f"{expected_asset.id} was expected but Asset Id: {set_value} was retrieved from \"{expected_asset.get_path()}\""
+
+
+def validate_script_canvas_graph_file(get_script_canvas_graph_file: typing.Callable,
+                                      set_script_canvas_graph_file: typing.Callable, sc_file_path: str) -> None:
+    """
+    Function to validate the setting of a script canvas graph file to the file source field in the script canvas component
+
+    get_script_canvas_component_graph_file: getter for the script canvas component file source handle field
+    set_script_canvas_component_graph_file: setter for the script canvas component's file source field.
```
FuzzyCarterAWS, line 225: Update docstring references to variable names.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
```diff
@@ -215,3 +214,46 @@ def validate_asset_property(property_name: str, get_asset_value: typing.Callable
...
+
+    """
+    Report.info(f"Validating Script Canvas component's file source field can be set.")
+
+    old_value = get_script_canvas_graph_file()
+
+    set_script_canvas_graph_file(sc_file_path)
+
+    set_value = get_script_canvas_graph_file()
+    assert set_value != old_value and set_value is not None, f"Graph file could not be set!"
```
FuzzyCarterAWS, line 236: Add some parentheses around the two conditions to make it clear to a reader skimming through what the intent of the test conditions are.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
```diff
@@ -215,3 +214,46 @@ def validate_asset_property(property_name: str, get_asset_value: typing.Callable
...
+    variable_name: the name of the variable to change
+    variable_state: whether the variable is initialized or not within graph file(initialized by the script canvas editor).
+    variable_value: the value to set the variable to. has no rigid type since variables can be primitives or user created types.
+    o3de/o3de#13344. We currently only support strings and boolean variables.
+
+
+    """
+    Report.info(f"Validating Script Canvas component's variable was set.")
+
+    set_variable_value(variable_name, variable_state, expected_variable_value)
```
FuzzyCarterAWS, line 255: Let's add the data we're trying to manipulate to the Info here.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
```diff
@@ -215,3 +214,46 @@ def validate_asset_property(property_name: str, get_asset_value: typing.Callable
...
+
+    new_variable_value = get_variable_value(variable_name, variable_state)
+
+    assert expected_variable_value == new_variable_value, f"Component variable {variable_name} was not set properly"
```
FuzzyCarterAWS, line 259: Let's add the data that was attempted to set here too.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_script_canvas.py
```diff
@@ -0,0 +1,119 @@
...
+        self.script_canvas_component.set_component_property_value(Path.SCRIPT_CANVAS_ASSET_PROPERTY, sourcehandle)
+
+    def get_component_graph_file(self) -> str:
+        """
+        function for getting the component property value of the graph file in the script canvas component
+
+        returns the component property value of the script canvas source handle (string)
+        """
+
+        return self.script_canvas_component.get_component_property_value(Path.SCRIPT_CANVAS_COMPONENT_PROPERTY_PATH)
```
FuzzyCarterAWS, line 89: Missed updating the enum in refactor from `SCRIPT_CANVAS_COMPONENT_PROPERTY_PATH` to `SCRIPT_CANVAS_ASSET_PROPERTY`

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_script_canvas.py
```diff
@@ -0,0 +1,119 @@
...
+        component_property_path = Path.construct_variable_component_property_path(Path, variable_name, variable_state)
+
+        self.script_canvas_component.set_component_property_value(component_property_path, variable_value)
+
+        return component_property_path
+
+    def get_variable_value(self, variable_name: str, variable_state: VariableState):
+        """
+        Function for getting the value of an exposed variable off the script canvas component. The variable type
+        can vary but only 2 types are currently supported: String adn Boolean. See github GH-13344 for more info.
```
FuzzyCarterAWS, line 113: Update GHI reference to the o3de/o3de#13344 format and make it the start of a line discussing the issue so it's up front and center to somebody reading this docstring that there's an ongoing implementation issue.

AutomatedTesting/Gem/PythonTests/scripting/scripting_utils/scripting_tools.py
```diff
@@ -20,13 +20,11 @@
...
-                                                 VARIABLE_MANAGER_QT, NODE_INSPECTOR_QT, NODE_INSPECTOR_UI, SCRIPT_EVENT_UI,
-                                                 VARIABLE_PALETTE_QT, ADD_BUTTON_QT, VARIABLE_TYPES, EVENTS_QT, DEFAULT_SCRIPT_EVENT,
-                                                 SCRIPT_EVENT_FILE_PATH, PARAMETERS_QT, VARIABLE_MANAGER_QT, NODE_INSPECTOR_QT,
-                                                 NODE_INSPECTOR_UI, VARIABLE_PALETTE_QT, ADD_BUTTON_QT, VARIABLE_TYPES,
-                                                 SCRIPT_CANVAS_COMPONENT_PROPERTY_PATH, ENTITY_STATES)
+                                 TREE_VIEW_QT, SEARCH_FRAME_QT, SEARCH_FILTER_QT, SAVE_STRING, NAME_STRING,
+                                 SAVE_ASSET_AS, WAIT_TIME_3, NODE_INSPECTOR_TITLE_KEY, WAIT_FRAMES, SCRIPT_EVENT_UI,
+                                 EVENTS_QT, DEFAULT_SCRIPT_EVENT, PARAMETERS_QT, NODE_INSPECTOR_QT, NODE_INSPECTOR_UI,
+                                 VARIABLE_PALETTE_QT, ADD_BUTTON_QT, VARIABLE_TYPES,
+                                 SCRIPT_CANVAS_COMPONENT_PROPERTY_PATH, ENTITY_STATES)
```
FuzzyCarterAWS, line 27: Remove any unused consts. For example SCRIPT_CANVAS_COMPONENT_PROPERTY_PATH isn't used

FuzzyCarterAWS (approved):
Some minor things I caught that once they're fixed you're good to go from me!

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
michleza: updated the docstring

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
michleza: good call. parentheses added

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
michleza: Updated the report.info and the assert call to make them more informative. There's realistically 3 things that could go wrong when setting the variable (incorrect name, incorrect state or incorrect type being sent to it)

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_component_validation.py
michleza: updated this as well

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_script_canvas.py
michleza: fixed.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_script_canvas.py
michleza: updated this and the setter as well since clients using either should be aware of the current limitations

AutomatedTesting/Gem/PythonTests/scripting/scripting_utils/scripting_tools.py
michleza: checked these constants and removed the ones not being used

jckand-amzn (approved):
General comments:

1. Outside of variable/function names, should Script Canvas be capitalized as it's the proper name of a tool?
2. Params in your docstrings don't use reStructuredText formatting that allows them to show up under the Params section when using auto-complete/mouseover in an IDE like PyCharm, and they're inconsistent between files. That said, PEP guidelines differ from PyCharm here so this is really a nit:

https://peps.python.org/pep-0257/
https://www.jetbrains.com/help/pycharm/creating-documentation-comments.html#1ba61c27

AutomatedTesting/Gem/PythonTests/scripting/TestSuite_Periodic.py
```diff
@@ -434,11 +434,8 @@ def test_ScriptCanvasComponent_OnEntityActivatedDeactivated_PrintMessage(self, r
...
     def test_ScriptCanvas_ChangingAssets_ComponentStable(self, request, workspace, editor, launcher_platform):
         expected_lines = [
-            "Test Entity created",
             "Game Mode successfully entered",
             "Game Mode successfully exited",
-            "Script Canvas File Updated",
-            "Expected log lines were found",
         ]
         hydra.launch_and_validate_results(
             request,
```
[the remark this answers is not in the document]
michleza: I believe so but that batch framework doesn't appear to draw the UI, which many of our QtPy tests need in order to function. There have been discussions around this and it's not clear if the test framework is set up for what we need or not.

PR #13218, opened 2022-11-16
Disable Atom level loads test to investigate crash.

jromnoa:
Signed-off-by: NULL <jromnoa@amazon.com>

## What does this PR do?
Temporarily disables the level loads test for Atom null render while the underlying crash is investigated.

Relates to latest build failure on https://github.com/o3de/o3de/issues/11238 which is build https://jenkins.build.o3de.org/job/O3DE/job/development/3468/ 

## How was this PR tested?
Ran the tests locally after removing the problematic tests from CMakeLists.txt registration.

### DIFF
--- a/AutomatedTesting/Gem/PythonTests/Atom/CMakeLists.txt
+++ b/AutomatedTesting/Gem/PythonTests/Atom/CMakeLists.txt
@@ -7,19 +7,19 @@
 #
 
 if(PAL_TRAIT_BUILD_HOST_TOOLS AND PAL_TRAIT_BUILD_TESTS_SUPPORTED)
-    ly_add_pytest(
-        NAME AutomatedTesting::Atom_Main_Null_Render_00
-        TEST_SUITE main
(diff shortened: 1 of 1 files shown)


lumberyard-employee-dm:
Any reason not to move this to the Sandbox SUITE?

Kadino (approved):
Reasonable to disable during investigation.

jromnoa:
@lumberyard-employee-dm 
We shouldn't move it to sandbox suite because the test was a successful test: it found a crash. That means the test is useful. I just want to disable it so that the build pipeline for other unrelated changes isn't blocked on a crash in one specific level.

dmcdiarmid-ly (approved):

Kadino:
Sandbox and disabled are effectively the same effect from the perspective of AR. Benefit of Sandbox is to automatically collect more data from periodic (nightly-ish) runs, which isn't always necessary.

lumberyard-employee-dm (approved):

PR #13242, opened 2022-11-16
Test fixes found with the address sanitizer

burelc-amzn:
This fixes some minor issues found in AzCore as a result of running the AzCore tests with the address sanitizer enabled on Linux.

### DIFF
--- a/Code/Framework/AzCore/AzCore/Name/NameDictionary.cpp
+++ b/Code/Framework/AzCore/AzCore/Name/NameDictionary.cpp
@@ -106,19 +106,21 @@ namespace AZ
 
         [[maybe_unused]] bool leaksDetected = false;
 
-        for (const auto& keyValue : m_dictionary)
+        for (auto i = m_dictionary.begin(), last = m_dictionary.end(); i != last;)
         {
-            Internal::NameData* nameData = keyValue.second.m_nameData;
+            Internal::NameData* nameData = i->second.m_nameData;
             const int useCount = nameData->m_useCount;
 
             if (useCount == 0)
             {
+                i = m_dictionary.erase(i);
                 delete nameData;
             }
             else
             {
                 leaksDetected = true;
-                AZ_TracePrintf("NameDictionary", "\tLeaked Name [%3d reference(s)]: hash 0x%08X, '%.*s'\n", useCount, keyValue.first, AZ_STRING_ARG(nameData->GetName()));
+                AZ_TracePrintf("NameDictionary", "\tLeaked Name [%3d reference(s)]: hash 0x%08X, '%.*s'\n", useCount, i->first, AZ_STRING_ARG(nameData->GetName()));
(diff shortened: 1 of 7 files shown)


Code/Framework/AzCore/AzCore/Name/NameDictionary.cpp
```diff
@@ -106,19 +106,21 @@ namespace AZ
 
         [[maybe_unused]] bool leaksDetected = false;
 
-        for (const auto& keyValue : m_dictionary)
+        for (auto i = m_dictionary.begin(), last = m_dictionary.end(); i != last;)
```
spham-amzn, line 109: huh, I thought one of the benefits of iterators was that you can do the origin for loop

spham-amzn: oh I see, you need to remove the entry

Code/Framework/AzCore/AzCore/std/hash_table.h
```diff
@@ -592,7 +592,7 @@ namespace AZStd
             // insert [first, last) one at a time
             for (; first != last; ++first)
             {
-                const key_type& valueKey = Traits::key_from_value(*first);
+                const key_type valueKey = Traits::key_from_value(*first);
```
lumberyard-employee-dm, line 595: This can potentially make an expensive copy of the key, when it does return a reference, such as when it is an AZStd::string.

Instead of storing the key in a local variable, we should just perform the `operator*()` when it is actually needed in the call to `m_hasher` and `find_insert_position`

lumberyard-employee-dm (changes requested):

Code/Framework/AzCore/AzCore/std/hash_table.h
burelc-amzn: This is fixed in the latest revision

lumberyard-employee-dm (approved):

Code/Framework/AzCore/AzCore/std/hash_table.h
```diff
@@ -590,28 +590,31 @@ namespace AZStd
...
+                    m_data.m_list.push_front(*it);
                     bucket.second = m_data.m_list.begin();
                 }
                 else
                 {
-                    if (!find_insert_position(valueKey, m_keyEqual, iter, numElements, AZStd::integral_constant<bool, Traits::has_multi_elements>()))
+                    if (!find_insert_position(AZStd::forward<decltype(valueKey)>(valueKey), m_keyEqual, iter, numElements, AZStd::integral_constant<bool, Traits::has_multi_elements>()))
                     {
-                        continue;
+                        return;
```
byrcolin, line 609: Is this correct?

byrcolin (approved):

spham-amzn (approved):

PR #13244, opened 2022-11-16
Updates android project generation scripts

amzn-phist:
## What does this PR do?

Introduces an updated software stack for Android builds that uses recent versions of development software:
- JDK 11
- Android Gradle Plugin 7.3.1
- Gradle 7.5.1
- NDK 25
- Build-Tools 33.0.0

Fixes issues with building unit tests for Android:
- When supplying the `--unit-test` argument to `generate_android_project.py`, it wasn't properly specifying a CMake target, which resulted in building everything and ultimately would fail during APK packaging.
- Fixes the dependent targets that are generated when specifying a `TEST_SUITE` target.  LYWrappers was only including the AzTestRunner target and not the suite of test modules that make up the suite.

## How was this PR tested?

Building Android Unit Tests locally and running them on a device.  The unit tests themselves aren't passing, but at least we can get logs and start down that path.


cgalvan (approved):

amzn-changml:
Hey, I've got a PR to update the build nodes to JDK11: https://github.com/o3de/o3de/pull/13245

Do we want to update all Android dependencies to the levels you indicated or keep it at minimum supported levels to test regression?

amzn-changml (approved):

amzn-phist:
@amzn-changml 

Here's what I'm thinking...
The Android unit tests aren't fully passing, this merely gets them to a build-able state.  So we shouldn't try to immediately run them on the main build jobs.  The fact that the unit tests aren't passing is independent of the software being used to build, I've also ran them with the current software stack and get the same results.

As for the Jenkins config, I think we could start down a migration path to update to the software stack I listed and eventually get off JDK 8 altogether.  I've built locally with these versions, but I've been focused on Unit Tests only and haven't spent much time on the main "GameLauncher" builds.  Maybe we need sig-platform approval, but I think we can start working towards an upgrade.

lumberyard-employee-dm (approved):

moraaar (approved):
The change looks great. But **please hold this PR and do not submit it yet**. We are currently doing a QA pass on OpenXR and this is a highly impactful change for Android since it updates many software pieces it uses.

Please do not submit until the QA pass has finished.

Thanks

moraaar:
> The change looks great. But **please hold this PR and do not submit it yet**. We are currently doing a QA pass on OpenXR and this is a highly impactful change for Android since it updates many software pieces it uses.
> 
> Please do not submit until the QA pass has finished.
> 
> Thanks

After discussing this with @amzn-phist the minimum versions were not increased and therefore the current versions documented should still work. So this should not affect the QA pass. Thanks!

PR #13249, opened 2022-11-17
Fix GHI-13060 resource view cache crash

amzn-tommy:
## What does this PR do?

Instead of using intrusive_ptr_base, RHI::Object now implements its own addref() and release() functions to support intrusive_ptr. This allows ResourceView to implement a custom release function to fix a threading issue.

The threading issue is the same as was previously encountered in the InstanceDatabase, and the fix is the same. This change also includes new unit tests that stress the refcounting, which are based on the unit tests in the InstanceDatabase.

The problem with the existing ObjectDeleter, when combined with the ResourceView cache, is you can have the following
Thread A                                                         Thread B
create resourceview in cache
let RHI::Ptr go out of scope
refcount == 0, so start object deleter
                                                                       Find resourceview in cache
                                                                       increment refcount to 1
                                                                       let RHI::Ptr go out of scope
                                                                       refcount == 0, so start object deleter
object deleter calls resourceview->shutdown
remove resourceview from cache
delete resourceview
                                                                       object deleter calls resourceview->shutdown
                                                                       crash, because resourceview has already been deleted

Now, instead, ResourceView::release() will not try and delete a cached resourceview if it is not found in the cache, and EraseResourceView will only remove a resource if nothing came along in the meantime and incremented the refcount

## How was this PR tested?

A new unit test was added to test this change. The vegetation hydra tests, which originally uncovered the issue, were also tested with this change.


Gems/Atom/RHI/Code/Source/RHI/Object.cpp
```diff
@@ -28,5 +28,25 @@ namespace AZ
...
+        }
+
+        void Object::release() const
+        {
+            int useCount = --m_useCount;
+            AZ_Assert(useCount >= 0, "Releasing an already released object");
+
+            if (useCount == 0)
+            {
+                // Get a mutable pointer to shutdown the object before deleting
```
rgba16f, line 45: suggestion: set m_useCount to a negative value here, force any other accessing thread to assert.

Gems/Atom/RHI/Code/Source/RHI/ResourceView.cpp
```diff
@@ -30,16 +30,41 @@ namespace AZ
...
+
+            if (useCount == 0)
             {
-                ResourceInvalidateBus::Handler::BusDisconnect(m_resource.get());
-                ShutdownInternal();
-                
-                m_resource->EraseResourceView(this);
-                m_resource = nullptr;
-                DeviceObject::Shutdown();
+                if (IsInitialized())
```
rgba16f, line 47: Maybe also force m_useCount to negative here.

rgba16f (approved):

Gems/Atom/RHI/Code/Source/RHI/Object.cpp
amzn-tommy: It's still expected that after this point, another thread might come and grab a reference to the resource view from the cache, and increment it back to 1. The current code allows that to happen without needing to re-create it. Setting it to -1 here would entail checking for -1 in GetResourceView, and creating a new one in that case instead of using the one that's found, but I figure you might as well re-use what's there.

Though on the other hand, doing would allow this thread to just always delete it, knowing that if another thread tried to find it in the cache, that thread would have to replace it with a new one. So it would simplify the logic of this release a bit.

rgba16f: I don't see any extra steps here to validate that another thread hasn't inc'd the use count to keep the object around, so this thread is going to delete it. Better to force other threads to handle that, otherwise there is still a thin gap where 2 threads can disagree what state the object is in. If you try to set the value to -1 and fail because it's no longer zero you will have removed that gap.

Gems/Atom/RHI/Code/Source/RHI/ResourceView.cpp
```diff
@@ -30,16 +30,41 @@ namespace AZ
...
         }
 
-        void ResourceView::Shutdown()
+        void ResourceView::release() const
         {
-            if (IsInitialized())
+            // It is possible that some other thread, not us, will delete this ResourceView after we
+            // decrement m_useCount. For example, another thread could create and release an instance
+            // immediately after we decrement. So we copy the necessary data to the callstack before
+            // decrementing. This ensures the call to ReleaseInstance() below won't crash even if this
```
invertednormal, line 38: I don't see `ReleaseInstance()` below anywhere, do you mean `EraseResourceView()`?

invertednormal (approved):

Gems/Atom/RHI/Code/Include/Atom/RHI/ResourceView.h
```diff
@@ -48,10 +52,12 @@ namespace AZ
             ResultCode Init(const Resource& resource);
 
         private:
-            //////////////////////////////////////////////////////////////////////////
-            // RHI::Object
-            void Shutdown() override final;
-            //////////////////////////////////////////////////////////////////////////
+            ///////////////////////////////////////////////////////////////////
+            // IntrusivePtrCountPolicy template overrides
+            void release() const final;
```
moudgils, line 57: Add a comment explaining why we are overriding release as oppose to shutdown which is the case everywhere else.

moudgils (approved):
Nice Fix and thanks for adding all the test case

Gems/Atom/RHI/Code/Source/RHI/Object.cpp
amzn-tommy: I see what you're saying. I was conflating ResourceView::release() with this one. The extra step exists for ResourceViews in EraseResourceView. There is no extra step for objects in general. 

This is a good suggestion, as it will prevent other non-resource view systems from running into the same problem (or at least make it more obvious what the problem is), since they should hit the assert in add_ref if they ever try to increment the refcount after the object is on track to be deleted.

Gems/Atom/RHI/Code/Include/Atom/RHI/Object.h
```diff
@@ -51,12 +36,33 @@ namespace AZ
              */
             const Name& GetName() const;
 
+            /**
```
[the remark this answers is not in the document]
amzn-tommy: Done.

Gems/Atom/RHI/Code/Include/Atom/RHI/ResourceView.h
amzn-tommy: Done.

Gems/Atom/RHI/Code/Source/RHI/ResourceView.cpp
amzn-tommy: Here is where I don't think that would work, since another thread can always come along and increase the refcount, but that is expected behavior due to the ResourceView cache.

rgba16f (approved):

PR #13251, opened 2022-11-17
Add more comparison operators to Dom::Path and Dom::PathEntry

amzn-tmryan:
**Description**
Before this PR, `Dom::PathEntry` and `Dom::Path` didn't have their own `operator<` which made comparisons difficult when ordering `Dom::Path` objects.

This PR adds `operator<` to `Dom::PathEntry` and operators `<, >, <=, >=` to `Dom::Path`.

The `Dom::PathEntry` sort order:
- Considers `size_t` < than `AZ::Name`
- Compares `AZ::Name` vs `AZ::Name` lexicographically
- Returns true for matching `PathEntry` values

The `Dom::Path` operators:
- Utilizes `Dom::PathEntry` operators
- Handles compares for mixed value paths
- Returns true for matching `Path` objects

Example sort order:
```
/0
/1
/1/4
/0hello (because this is actually a key, not a size_t index)
/two
/two/three
/two/threes
```

**Testing**
- Newly added and old AzCore.Tests passed

### DIFF
--- a/Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
+++ b/Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
@@ -150,12 +150,12 @@ namespace AZ::Dom
 
     bool operator<(const AZ::Dom::PathEntry& entry, size_t index)
     {
-        return entry.IsIndex() && entry.GetIndex() < index;
+        return entry < AZStd::to_string(index);
     }
 
     bool operator<(size_t index, const AZ::Dom::PathEntry& entry)
     {
-        return !entry.IsIndex() || index < entry.GetIndex();
+        return AZStd::to_string(index) < entry;
     }
 
     bool operator<(const AZ::Dom::PathEntry& entry, const AZ::Name& key)
@@ -170,12 +170,26 @@ namespace AZ::Dom
 
     bool operator<(const AZ::Dom::PathEntry& entry, AZStd::string_view key)
     {
-        return entry.IsIndex() || entry.GetKey().GetStringView() < key;
+        if (entry.IsIndex())
+        {
+            return AZStd::to_string(entry.GetIndex()) < key;
+        }
+        else
+        {
+            return entry.GetKey().GetStringView() < key;
+        }
     }
 
     bool operator<(AZStd::string_view key, const AZ::Dom::PathEntry& entry)
     {
-        return entry.IsKey() && key < entry.GetKey().GetStringView();
+        if (entry.IsIndex())
+        {
+            return key < AZStd::to_string(entry.GetIndex());
+        }
+        else
+        {
+            return key < entry.GetKey().GetStringView();
+        }
     }
 } // namespace AZ::Dom
 
--- a/Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
+++ b/Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
@@ -51,10 +51,10 @@ namespace AZ::Dom
         return m_value == other.m_value;
     }
 
-    bool PathEntry::operator==(size_t value) const
+    bool PathEntry::operator==(size_t index) const
     {
         const size_t* internalValue = AZStd::get_if<size_t>(&m_value);
-        return internalValue != nullptr && *internalValue == value;
+        return internalValue != nullptr && *internalValue == index;
     }
 
     bool PathEntry::operator==(const AZ::Name& key) const
@@ -134,6 +134,48 @@ namespace AZ::Dom
                 }
             },
             m_value);
+    }
+
+    bool operator<(const AZ::Dom::PathEntry& lhs, const AZ::Dom::PathEntry& rhs)
+    {
+        if (lhs.IsIndex())
+        {
+            return lhs.GetIndex() < rhs;
+        }
+        else
+        {
+            return lhs.GetKey() < rhs;
+        }
+    }
+
+    bool operator<(const AZ::Dom::PathEntry& entry, size_t index)
+    {
+        return entry.IsIndex() && entry.GetIndex() < index;
+    }
+
+    bool operator<(size_t index, const AZ::Dom::PathEntry& entry)
+    {
+        return !entry.IsIndex() || index < entry.GetIndex();
+    }
+
+    bool operator<(const AZ::Dom::PathEntry& entry, const AZ::Name& key)
+    {
+        return entry < key.GetStringView();
+    }
+
+    bool operator<(const AZ::Name& key, const AZ::Dom::PathEntry& entry)
+    {
+        return key.GetStringView() < entry;
+    }
+
+    bool operator<(const AZ::Dom::PathEntry& entry, AZStd::string_view key)
+    {
+        return entry.IsIndex() || entry.GetKey().GetStringView() < key;
+    }
+
+    bool operator<(AZStd::string_view key, const AZ::Dom::PathEntry& entry)
+    {
+        return entry.IsKey() && key < entry.GetKey().GetStringView();
     }
 } // namespace AZ::Dom
 
@@ -225,6 +267,46 @@ namespace AZ::Dom
     bool Path::operator==(const Path& other) const
     {
         return m_entries == other.m_entries;
+    }
+
+    bool Path::operator<(const Path& other) const
+    {
+        auto ComparePathEntry = [](const PathEntry& lhs, const PathEntry& rhs)
+        {
+            return lhs == rhs;
+        };
+        // Find the PathEntry where the two Path objects differ, if any
+        auto [thisPathIter, otherPathIter] = AZStd::mismatch(
+            GetEntries().begin(), GetEntries().end(),
+            other.GetEntries().begin(), other.GetEntries().end(),
+            ComparePathEntry);
+
+        // If the paths differ then compare the differing PathEntries
+        if (thisPathIter != GetEntries().end() && otherPathIter != other.GetEntries().end())
+        {
+            return *thisPathIter < *otherPathIter;
+        }
+        // If one Path is a prefix of the other, or the paths are identical, then the shorter
+        // Path is the smaller Path
+        else
+        {
+            return other.GetEntries().size() > GetEntries().size();
+        }
+    }
+
+    bool Path::operator>(const Path& other) const
+    {
+        return other.operator<(*this);
+    }
+
+    bool Path::operator<=(const Path& other) const
+    {
+        return !other.operator<(*this);
+    }
+
+    bool Path::operator>=(const Path& other) const
+    {
+        return !operator<(other);
     }
 
     const Path::ContainerType& Path::GetEntries() const
--- a/Code/Framework/AzCore/AzCore/DOM/DomPath.h
+++ b/Code/Framework/AzCore/AzCore/DOM/DomPath.h
@@ -47,6 +47,21 @@ namespace AZ::Dom
         bool operator!=(const AZ::Name& key) const;
         bool operator!=(AZStd::string_view key) const;
 
+        //! The desired sort order of PathEntry first has index values sorted
+        //! by numerical value then key values sorted in lexicographic order.
+        //! For example:
+        //! 0
+        //! 10
+        //! 0hello (note that this is a key, not an index)
+        //! two
+        friend bool operator<(const PathEntry& lhs, const PathEntry& rhs);
+        friend bool operator<(const PathEntry& entry, size_t index);
+        friend bool operator<(size_t index, const PathEntry& entry);
+        friend bool operator<(const PathEntry& entry, const AZ::Name& key);
+        friend bool operator<(const AZ::Name& key, const PathEntry& entry);
+        friend bool operator<(const PathEntry& entry, AZStd::string_view key);
+        friend bool operator<(AZStd::string_view key, const PathEntry& entry);
+
         void SetEndOfArray();
 
         bool IsEndOfArray() const;
@@ -111,6 +126,19 @@ namespace AZ::Dom
         {
             return !operator==(rhs);
         }
+
+        //! The ordering of Path objects is based on PathEntry sort order.
+        //! An example sort would be:
+        //! /0/1/3
+        //! /1
+        //! /1/4
+        //! /apple
+        //! /two
+        //! /two/0hello
+        bool operator<(const Path&) const;
+        bool operator>(const Path&) const;
+        bool operator<=(const Path&) const;
+        bool operator>=(const Path&) const;
 
         const ContainerType& GetEntries() const;
         void Push(PathEntry entry);
--- a/Code/Framework/AzCore/AzCore/DOM/DomPath.h
+++ b/Code/Framework/AzCore/AzCore/DOM/DomPath.h
@@ -47,13 +47,7 @@ namespace AZ::Dom
         bool operator!=(const AZ::Name& key) const;
         bool operator!=(AZStd::string_view key) const;
 
-        //! The desired sort order of PathEntry first has index values sorted
-        //! by numerical value then key values sorted in lexicographic order.
-        //! For example:
-        //! 0
-        //! 10
-        //! 0hello (note that this is a key, not an index)
-        //! two
+        //! PathEnty objects are ordered lexicographically.
         friend bool operator<(const PathEntry& lhs, const PathEntry& rhs);
         friend bool operator<(const PathEntry& entry, size_t index);
         friend bool operator<(size_t index, const PathEntry& entry);
@@ -127,14 +121,11 @@ namespace AZ::Dom
             return !operator==(rhs);
         }
 
-        //! The ordering of Path objects is based on PathEntry sort order.
-        //! An example sort would be:
-        //! /0/1/3
-        //! /1
-        //! /1/4
-        //! /apple
-        //! /two
-        //! /two/0hello
+        //! Path objects are ordered lexicographically based on their full path. Note that
+        //! paths can contain a mix of interger or string value PathEntry components.
+        //! PathEntry objects with string values, as opposed to integer values, will not
+        //! necessarily sort as they appear in the DOM. Paths consisting only of
+        //! indexed PathEntry objects will be ordered as they are positioned in the DOM.
         bool operator<(const Path&) const;
         bool operator>(const Path&) const;
         bool operator<=(const Path&) const;
--- a/Code/Framework/AzCore/Tests/DOM/DomPathTests.cpp
+++ b/Code/Framework/AzCore/Tests/DOM/DomPathTests.cpp
@@ -153,37 +153,46 @@ namespace AZ::Dom::Tests
 
     TEST_F(DomPathTests, OperatorOverloads_LessThanGreaterThan)
     {
+        // Path - index and key tests
         EXPECT_TRUE(Path("/0/99") < Path("/0/a"));
         EXPECT_FALSE(Path("/0/a") < Path("/0/99"));
-
+        // Path - key and key tests
         EXPECT_FALSE(Path("/0/b") < Path("/0/aa"));
         EXPECT_TRUE(Path("/0/a") < Path("/0/aa"));
-
+        // Path - index and index tests
         EXPECT_TRUE(Path("/0/1") < Path("/0/2"));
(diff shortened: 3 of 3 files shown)


Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
```diff
@@ -89,6 +89,59 @@ namespace AZ::Dom
...
 
+    bool PathEntry::operator<(const PathEntry& other) const
+    {
+        const size_t* otherIndexValue = AZStd::get_if<size_t>(&other.m_value);
+        const AZ::Name* otherNameValue = AZStd::get_if<AZ::Name>(&other.m_value);
+        if (otherIndexValue)
+        {
+            return operator<(*otherIndexValue);
+        }
+        return operator<(*otherNameValue);
```
amzn-tmryan, line 100: Is there a better way to do this comparison?

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
```diff
@@ -89,6 +89,59 @@ namespace AZ::Dom
...
+        const AZ::Name* internalValue = AZStd::get_if<AZ::Name>(&m_value);
+        return internalValue != nullptr &&
+            AZStd::lexicographical_compare(
+                   internalValue->GetStringView().begin(),
+                   internalValue->GetStringView().end(),
+                   key.begin(),
+                   key.end());
+    }
+
+    bool PathEntry::operator>(const PathEntry& other) const
```
alexmontAmazon, line 125: couple things:
1) generally containers just use operator< for comparisons, so the operator> is largely unnecessary
2) the inverse of lessThan isn't greaterThan, it's greaterThanOrEqualTo, so this code is incorrect. Honestly, I would just remove these operator> altogether

amzn-tmryan: Makes sense. I'll remove the operator>. We can revisit it if needed later on.

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
lumberyard-employee-dm: Actually here is what can be done.
If the `m_value` variant contains a size_t, convert it to a string.
And then just use the `string_view` < operator

```suggestion
        // Used to store a string that has been from an index path entry
        // for long enough to be used in a string comparison 
        AZStd::fixed_string<32> thisIndexStorage;
        AZStd::fixed_string<32> otherIndexStorage;
        
        // References a view of this path entry as a string
        AZStd::string_view thisValueView;
        // References a view of the other path entry as a string
        AZStd::string_view otherValueView;
        
        // Get string_view for this PathEntry instance
        if (const size_t* thisIndexValue = AZStd::get_if<size_t>(&m_value);
            thisIndexValue != nullptr)
        {
            AZStd::to_string(thisIndexStorage, *thisIndexValue);
            thisValueView = thisIndexStorage;
        }
        else if (const AZ::Name* thisNameValue = AZStd::get_if<AZ::Name>(&m_value);
            thisNameValue != nullptr)
        {
            thisValueView = thisNameValue->GetStringView();
        }

        if (const size_t* otherIndexValue = AZStd::get_if<size_t>(&other.m_value);
            otherIndexValue != nullptr)
        {
            AZStd::to_string(otherIndexStorage, *otherIndexValue);
            otherValueView = otherIndexStorage;
        }
        else if (const AZ::Name* otherNameValue = AZStd::get_if<AZ::Name>(&other.m_value);
            otherNameValue != nullptr)
        {
            otherValueView = otherNameValue->GetStringView();
        }

        return thisValueView < otherValueView;
```

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
lumberyard-employee-dm: If you are going to implement the less than operator, than either it should implemented by itself or all four of the comparison operators should be implemented as well(operator<, operator>, operator<=, operator>=).

In no situation should only 2 or 3 out of the 4 comparison be implemented.
It would surprising to users of the class if they had `operator>` but not `operator<`= or `operator>=`

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
```diff
@@ -89,6 +89,59 @@ namespace AZ::Dom
...
+    bool PathEntry::operator<(AZStd::string_view key) const
+    {
+        const AZ::Name* internalValue = AZStd::get_if<AZ::Name>(&m_value);
+        return internalValue != nullptr &&
+            AZStd::lexicographical_compare(
+                   internalValue->GetStringView().begin(),
+                   internalValue->GetStringView().end(),
+                   key.begin(),
+                   key.end());
+    }
```
lumberyard-employee-dm, line 123: With the snippet in the comment I added above.
These overloads of operator< aren't needed.
The numeric index would be converted to a string value and the comparison will use lexical string comparison via the `string_view::operator<`

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
lumberyard-employee-dm: Also all the other operators should be implemented in terms of operator<
```
bool PathEntry::operator>(const PathEntry& other) const
{
    return other < *this;
}
bool PathEntry::operator<=(const PathEntry& other) const
{
    return !(other < *this);
}
bool PathEntry::operator>=(const PathEntry& other) const
{
    return *this < other;
}
```

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
```diff
@@ -227,6 +280,32 @@ namespace AZ::Dom
...
+        }
+        if (thisIter != GetEntries().end() && otherIter != other.GetEntries().end())
+        {
+            isLessThan = *thisIter < *otherIter;
+        }
+        else
+        {
+            isLessThan = other.GetEntries().size() > GetEntries().size();
+        }
+        return isLessThan;
```
lumberyard-employee-dm, line 301: This can use use [AZStd::mismatch](https://en.cppreference.com/w/cpp/algorithm/mismatch) for the comparison and then check if the `thisIter` is at the end

```suggestion
            auto ComparePathEntry = [](const PathEntry& lhs, const PathEntry& rhs)
            {
                return lhs < rhs;
            };
            auto [thisPathIter, otherPathIter] = AZStd::mismatch(GetEntries().begin(), GetEntries().end(),
                other.GetEntries().begin(), other.GetEntries().end(), ComparePathEntry);
                
            // If the thisPathIter isn't the sentinel iterator than one of it's path entry  was not less than the corresponding `other` path entry
            return thisPathIter != GetEntries().end();
```

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
amzn-tmryan: Working on adding the other operators (<=, >=) and fixing up some other issues I noticed now. Will look at your suggestions above and get another commit up.

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
amzn-tmryan: `return lhs == rhs` yields the same iterators as the current code, but after some testing it seems like we still have the two cases after finding the mismatch.

Changed this function to use AZStd::mismatch because using ::mismatch has clearer semantics, but the PR is still using the original two cases after finding the mismatch which cover Paths with differing entries as well as Paths where one is a prefix of the other and matching Paths.

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
amzn-tmryan: Turns out I will need the other overloads for comparing values to PathEntry objects. Going to leave them as is for now. 

I did fix the return of `PathEntry::operator<(AZStd::string_view key)` so it returns `internalValue == nullptr || ...` since we want strings to show up after numbers as it would lexicographically.

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
amzn-tmryan: After looking into what it would take to add (>, <=, >=) to PathEntry, I decided to leave the class as is with operator< only. If the need arises for the others then we can implement them.

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
amzn-tmryan: This is cool and clever, but it looks like I might need the other operator< overloads anyway for value comparisons like:
Path("0/1").GetEntries()[1] < "donut"
Path("0/3/1").GetEntries()[1] < "2"

I'll leave it as is for now since I do need those other operators anyway.

amzn-tmryan:
Addressed feedback and added some tests. Thanks @lumberyard-employee-dm and @alexmontAmazon!

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
```diff
@@ -227,6 +260,48 @@ namespace AZ::Dom
...
+    }
+
+    bool Path::operator>(const Path& other) const
+    {
+        return other.operator<(*this);
+    }
+
+    bool Path::operator<=(const Path& other) const
+    {
+        return operator<(other) || operator==(other);
```
alexmontAmazon, line 297: It's probably faster to do !operator> here

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
```diff
@@ -227,6 +260,48 @@ namespace AZ::Dom
...
+    }
+
+    bool Path::operator>=(const Path& other) const
+    {
+        return operator>(other) || operator==(other);
```
alexmontAmazon, line 302: and this should probably be !operator<

alexmontAmazon (approved):
looks good, just a couple of performance nits with not using the == operator for the <= and the >=

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
lumberyard-employee-dm: If the other overloads are going to remain, then this will need a friend functions as well where they are the first overload.
furthermore they should then be implemented using the string_view overload then.

The lexicalgraphical_compare is unnecessary as string_view comparison operators already performs a lexigraphical comparison.


    friend bool operator<(AZStd::string_view key, const PathEntry& entry)
    {
        const AZ::Name* internalValue = AZStd::get_if<AZ::Name>(&entry.m_value);
        return internalValue != nullptr &&
            AZStd::lexicographical_compare(
                   internalValue->GetStringView().begin(),
                   internalValue->GetStringView().end(),
                   key.begin(),
                   key.end());
    }

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
lumberyard-employee-dm, line 297: This shouldn't use `operator==` here at all.
```suggestion
        return !other.operator<(*this);
```

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
lumberyard-employee-dm: I agree, this should be 
```suggestion
        return !operator<(other);
```

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
```diff
@@ -227,6 +260,48 @@ namespace AZ::Dom
...
+        {
+            isLessThan = *thisPathIter < *otherPathIter;
+        }
+        // If one Path is a prefix of the other, or the paths are identical, then the shorter
+        // Path is the smaller Path
+        else
+        {
+            isLessThan = other.GetEntries().size() > GetEntries().size();
+        }
+        return isLessThan;
```
lumberyard-employee-dm, line 287: The isLessThan boolean isn't needed.
The if block and else blocks can just return the comparison result.
```suggestion
        bool isLessThan = false;
        // If the paths differ then compare the differing PathEntries
        if (thisPathIter != GetEntries().end() && otherPathIter != other.GetEntries().end())
        {
            return *thisPathIter < *otherPathIter;
        }
        // If one Path is a prefix of the other, or the paths are identical, then the shorter
        // Path is the smaller Path
        else
        {
            return GetEntries().size() < other.GetEntries().size();
        }
```

Code/Framework/AzCore/AzCore/DOM/DomPath.h
```diff
@@ -47,6 +47,11 @@ namespace AZ::Dom
         bool operator!=(const AZ::Name& key) const;
         bool operator!=(AZStd::string_view key) const;
 
+        bool operator<(const PathEntry& other) const;
+        bool operator<(size_t index) const;
+        bool operator<(const AZ::Name& key) const;
+        bool operator<(AZStd::string_view key) const;
```
lumberyard-employee-dm, line 53: The other functions need free functions equivalents where they are a the first parameter.
It actually would be more consistent to make all these functions friend free functions instead of members.

```suggestion
        friend bool operator<(const PathEntry& entry, size_t index);
        friend bool operator<(size_t index, const PathEntry& entry);
        friend bool operator<(const PathEntry& entry, const AZ::Name& key);
        friend bool operator<(const AZ::Name& key, const PathEntry& entry);
        friend bool operator<(const PathEntry& entry, AZStd::string_view key);
        friend bool operator<(AZStd::string_view key, const PathEntry& entry)
```

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
amzn-tmryan: Done.

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
amzn-tmryan: Nice, that way everything is using the operator<? Done.

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
amzn-tmryan: Done.

Code/Framework/AzCore/AzCore/DOM/DomPath.h
amzn-tmryan: Swapped the member `operator<` functions for a set of `operator<` friend free functions.

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
```diff
@@ -135,6 +135,58 @@ namespace AZ::Dom
...
+
+    bool operator<(const AZ::Dom::PathEntry& entry, size_t index)
+    {
+        return entry.IsIndex() && entry.GetIndex() < index;
+    }
+
+    bool operator<(size_t index, const AZ::Dom::PathEntry& entry)
+    {
+        return !entry.IsIndex() || index < entry.GetIndex();
+    }
```
lumberyard-employee-dm, line 159: The operation being performed here doesn't make any sense.

If the path entry isn't an index then the less than operator should be false in all cases

lumberyard-employee-dm (changes requested):

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
```diff
@@ -135,6 +135,58 @@ namespace AZ::Dom
...
+
+    bool operator<(const AZ::Dom::PathEntry& entry, AZStd::string_view key)
+    {
+        return entry.IsIndex() ||
+            AZStd::lexicographical_compare(
+                entry.GetKey().GetStringView().begin(),
+                entry.GetKey().GetStringView().end(),
+                key.begin(),
+                key.end());
```
lumberyard-employee-dm, line 178: This should use the string_view operator<
```suggestion
        return entry.IsKey() && entry.GetKey().GetStringView() < key;
```

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
```diff
@@ -135,6 +135,58 @@ namespace AZ::Dom
...
+    }
+
+    bool operator<(AZStd::string_view key, const AZ::Dom::PathEntry& entry)
+    {
+        return entry.IsKey() &&
+            AZStd::lexicographical_compare(
+                key.begin(),
+                key.end(),
+                entry.GetKey().GetStringView().begin(),
+                entry.GetKey().GetStringView().end());
```
lumberyard-employee-dm, line 188: repeat: This should use the string_view operator< free function instead of lexicographical_compare
```suggestion
                return entry.isKey() && key < entry.GetKey().GetStringView();
```

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
amzn-tmryan: `!entry.IsIndex() || index < entry.GetIndex()`
I was thinking it meant "return true if entry is a key, or return their comparison result if they are both indices".

My understanding is that we want non-index path entry values to be sorted after index path entries. Something like this:
```
/0
/1
/1/3
/0hello (because this is actually a key, not a size_t index)
/two
/two/three
```

With the code above, the comparison `1 < PathEntry("/hello")` would yield true because 1 is an index value and "hello" is a key value. So our sort order is:
```
/1
/hello
```

If we always return false when the path entry is not an index, then the sort order becomes:
```
/hello
/1
```

I should add an example sort to the description of this PR too so it's less unclear. And probably a comment inline somewhere.
@alexmontAmazon let me know if I misunderstood the desired sort order!

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
amzn-tmryan: I didn't realize that strings were compared lexicographically. I thought there was some difference in something like `"ac" < "b" == false`. I see now that they use https://en.cppreference.com/w/cpp/algorithm/lexicographical_compare. Will get that swapped out.

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
amzn-tmryan: Same as above. Swapping this out.

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
alexmontAmazon: I think it's reasonable to sort index entries before string entries, just so long as it's documented as such.

amzn-tmryan:
Removed use of implicit lexicographic compare since it's already done by string_view comparisons.

amzn-tmryan:
Added comments to the operator< for both Path and PathEntry so it's more clear what sorting objects of these types will do.

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
lumberyard-employee-dm: But in JSON indices and strings aren't treated differently.
They sort together as an index is just treated as a string.
And by string sorting logic the hex values of 0x30 - 0x39 [0-9] sorts before 0x41 - 0x5a [A-Z]

Code/Framework/AzCore/AzCore/DOM/DomPath.cpp
lumberyard-employee-dm: The [string_view](https://en.cppreference.com/w/cpp/string/basic_string_view/operator_cmp) and string C++ classe comparison operators are lexicographical.
What is isn't lexicographical is comparing of the raw `const char*` C type which is a numeric pointer check.

lumberyard-employee-dm (approved):

amzn-tmryan:
Just an update since this has been sitting here for a while. I'm going to change this PR so we are sorting as JSON would be sorted. Will get to this soon.

amzn-tmryan:
Closing this PR after some discussion. The need for it has passed and there isn't really a clear way to order PathEntry objects since they can contain one of two types that don't play well with each other semantically when ordered.

amzn-tmryan:
It was a mistake to close this PR. It's still needed by another upcoming PR.

PR #13254, opened 2022-11-17
Python script to generate shdervariantlist for each .shader in the engine

etherealchain:

## What does this PR do?

Addressing https://github.com/o3de/sig-graphics-audio/discussions/92
I'm not sure if the original function is still needed (manually generate shader variant list), so I keep the script and create a new one `GenerateShaderVariantListForAllShaders.py` to scan for all .shader files in the Gems folder, and generate .shadervariantlist for each of them.

The script can be executed by the Run Python Script function in ShaderManagementConsole.
image

Shader with actual material usage will generate a shadervariantlist, shader without material usage will do nothing.

## How was this PR tested?
Test locally on my PC.


gadams3 (approved):
This looks good to me. If I remember correctly from past discussions, one of the goals was to be able to run this from the main editor or other tools. Technically, you can do that by launching shader management console and passing the Python script path as a command line parameter. But if we want to run it without launching shader management console all of the utility functions from azlmbr.shadermanagementconsole would need to move to some RPI or common feature system component.

Gems/Atom/Tools/ShaderManagementConsole/Scripts/GenerateShaderVariantListForAllShaders.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def main():
+    print("==== Begin processing shader ==========================================================")
+    # Search all .shader in Engine Gems folder
+    search_and_save_shadervariant(azlmbr.paths.engroot + '/Gems')
+
+    # Search in project folder
+    search_and_save_shadervariant(azlmbr.paths.projectroot + '/Shaders')
+    search_and_save_shadervariant(azlmbr.paths.projectroot + '/Assets')
```
santorac, line 29: I believe the engine now supports having externally registered gems, outside the engine folder and outside the project folder. So it's not enough to just scan these three folders. A more reliable way to find all the shaders would be to call EnumerateAssets. I'm not sure whether that's exposed to python or not though. It looks like it would need to be added in AssetManagerComponent::Reflect https://github.com/o3de/o3de/blob/f4bda1feee338b0426f8ee7783cc5d903efc5b43/Code/Framework/AzCore/AzCore/Asset/AssetManagerComponent.cpp#L198

Gems/Atom/Tools/ShaderManagementConsole/Scripts/GenerateShaderVariantListForAllShaders.py
```diff
@@ -0,0 +1,44 @@
...
+import re
+import json
+import GenerateShaderVariantListUtil
+
+def search_and_save_shadervariant(searchPath):
+    for subdir, dirs, files in os.walk(searchPath):
+        for file in files:
+            filename = os.path.join(subdir, file)
+            if re.search('^.*\.shader$', filename):
+                shaderVariantList, defaultShaderVariantListFilePath = GenerateShaderVariantListUtil.create_shadervariantlist_for_shader(filename)
```
santorac, line 17: This process probably isn't the most efficient. This is looping through all the shaders, then searching for all materials that use that shader, load the material, and generate the variant list for that one shader. So if a material type has 10 shaders, then every material of that type will be loaded 10 times. 

Instead, it would probably be better to scan for all materials, load each material once, and then collect information about the shaders and variants in memory. Then in the end, dump out all the shader variant lists.

Gems/Atom/Tools/ShaderManagementConsole/Scripts/GenerateShaderVariantListForAllShaders.py
```diff
@@ -0,0 +1,44 @@
...
+                if shaderVariantList.shaderVariants:
+                    azlmbr.shader.SaveShaderVariantListSourceData(defaultShaderVariantListFilePath, shaderVariantList)
```
santorac, line 19: defaultShaderVariantListFilePath is (in most cases) not the right path. Instead of saving the .shadervariantlist next to the original shader in whatever gem it came from, we want to save the file in the active project's folder, because the list of variants for each project will be different. It's similar to how the SMC "Save As" dialog allows you to pick the original location or the project folder. The project folder should be the default.

Gems/Atom/Tools/ShaderManagementConsole/Scripts/GenerateShaderVariantListForAllShaders.py
etherealchain: I try to add `AssetCatalogRequests::EnumerateAssets` to BehaviorContext, but I got a warning
```
BehaviorMethod EnumerateAssets: Parameter at [0] index expects (AZStd::function<void>:{73369FC8-B59B-5B48-B61A-ECCBF94031C4}) for method but got type (<class 'function'>)
```
Looks like Python function didn't get translated to `AZStd::function<void>` correctly.
Do you know the right way to use it?

etherealchain: I am able to call the function in c++ and pass back the list to python.

PR #13258, opened 2022-11-17
Added PIX implementation for AZ_PROFILE_DATAPOINTs

SergeyAMZN:

## What does this PR do?

Adds PIX implementation for AZ_PROFILE_DATAPOINTs. 
Now we can plot data point metrics along with profiling information of the functions.

## How was this PR tested?

Manual testing.

pix1
pix2

### DIFF
--- a/Code/Framework/AzCore/AzCore/IO/Streamer/Statistics.cpp
+++ b/Code/Framework/AzCore/AzCore/IO/Streamer/Statistics.cpp
@@ -140,7 +140,7 @@ namespace AZ::IO
         [[maybe_unused]] double value)
     {
         AZ_PROFILE_DATAPOINT(AzCore, value,
-            AZStd::string::format(L"Streamer/%.*s/%.*s (Raw)",
+            AZStd::wstring::format(L"Streamer/%.*s/%.*s (Raw)",
             aznumeric_cast<int>(owner.size()), owner.data(),
             aznumeric_cast<int>(name.size()), name.data()).data());
     }
--- a/Code/Framework/AzCore/AzCore/IO/Streamer/Statistics.cpp
+++ b/Code/Framework/AzCore/AzCore/IO/Streamer/Statistics.cpp
@@ -140,9 +140,9 @@ namespace AZ::IO
         [[maybe_unused]] double value)
     {
         AZ_PROFILE_DATAPOINT(AzCore, value,
-            "Streamer/%.*s/%.*s (Raw)",
+            AZStd::string::format(L"Streamer/%.*s/%.*s (Raw)",
             aznumeric_cast<int>(owner.size()), owner.data(),
-            aznumeric_cast<int>(name.size()), name.data());
+            aznumeric_cast<int>(name.size()), name.data()).data());
     }
 
     AZStd::string_view Statistic::GetOwner() const
--- a/Gems/PhysX/Code/Source/Scene/PhysXScene.cpp
+++ b/Gems/PhysX/Code/Source/Scene/PhysXScene.cpp
@@ -1217,7 +1217,7 @@ namespace PhysX
 
     void PhysXScene::UpdateAzProfilerDataPoints()
     {
-#if !defined(_RELEASE)
+#if !defined(AZ_RELEASE_BUILD)
         if (!physx_profileSimulationDatapoints)
         {
             return;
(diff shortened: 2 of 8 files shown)


Gems/PhysX/Code/Source/Scene/PhysXScene.cpp
```diff
@@ -1212,15 +1217,8 @@ namespace PhysX
...
-        using physx::PxGeometryType;
-
-        bool isProfilingActive = false;
-        if (auto profilerSystem = AZ::Debug::ProfilerSystemInterface::Get(); profilerSystem)
-        {
-            isProfilingActive = profilerSystem->IsActive();
-        }
-
-        if (!isProfilingActive)
+#if !defined(_RELEASE)
```
moraaar, line 1220: Please use `AZ_RELEASE_BUILD` instead of `_RELEASE`.

Gems/PhysX/Code/Source/Scene/PhysXScene.cpp
```diff
@@ -1282,20 +1288,11 @@ namespace PhysX
...
-        AZ_PROFILE_DATAPOINT(Physics, stats.nbDiscreteContactPairsWithContacts, RootCategory, CollisionsSubCategory, "DiscreteContactPairsWithContacts");
-        AZ_PROFILE_DATAPOINT(Physics, stats.nbNewPairs, RootCategory, CollisionsSubCategory, "NewPairs");
-        AZ_PROFILE_DATAPOINT(Physics, stats.nbLostPairs, RootCategory, CollisionsSubCategory, "LostPairs");
-        AZ_PROFILE_DATAPOINT(Physics, stats.nbNewTouches, RootCategory, CollisionsSubCategory, "NewTouches");
-        AZ_PROFILE_DATAPOINT(Physics, stats.nbLostTouches, RootCategory, CollisionsSubCategory, "LostTouches");
-        AZ_PROFILE_DATAPOINT(Physics, stats.nbPartitions, RootCategory, CollisionsSubCategory, "Partitions");
+        AZ_PROFILE_DATAPOINT(Physics, ccdPairs, L"PhysX/Collisions/CCDPairs");
+        AZ_PROFILE_DATAPOINT(Physics, modifiedPairs, L"PhysX/Collisions/ModifiedPairs");
+        AZ_PROFILE_DATAPOINT(Physics, triggerPairs, L"PhysX/Collisions/TriggerPairs");
+#endif
```
moraaar, line 1295: [nit] Please include comment `#endif // !defined(AZ_RELEASE_BUILD)`

moraaar (approved):
LGTM but I'm not familiarized with the profiler code.

hultonha (approved):
LGTM! Nice work adding this 👍

Code/Framework/AzCore/AzCore/IO/Streamer/Statistics.cpp
```diff
@@ -140,9 +140,9 @@ namespace AZ::IO
         [[maybe_unused]] double value)
     {
         AZ_PROFILE_DATAPOINT(AzCore, value,
-            "Streamer/%.*s/%.*s (Raw)",
+            AZStd::string::format(L"Streamer/%.*s/%.*s (Raw)",
```
lumberyard-employee-dm, line 143: This should be `AZStd::wstring::format`

lumberyard-employee-dm (changes requested):

Code/Framework/AzCore/AzCore/IO/Streamer/Statistics.cpp
jeremyong-az: Good callout, I sort of assumed that there was a `wchar_t const*` overload or something. Is it worth a static assert in `AZStd::string::format` to make sure this error isn't made do you think?

SergeyAMZN: good point, thanks! fixed

Gems/PhysX/Code/Source/Scene/PhysXScene.cpp
SergeyAMZN: Updated, thanks! 
There are 300+ cases of using `_RELEASE` in the codebase but only 13 of `AZ_RELEASE_BUILD`..

lumberyard-employee-dm (approved):

Code/Framework/AzCore/AzCore/IO/Streamer/Statistics.cpp
lumberyard-employee-dm: A static_assert could be extra protection for the use to make sure they want are using the `AZStd::wstring` name when to make sure it is known that the function will return a **wstring** and not a  _string_

Gems/PhysX/Code/Source/Scene/PhysXScene.cpp
moraaar: Yeah... hopefully they'll be replaced with time bit by bit when code changes happen around them. I always encourage in PRs to use `AZ_RELEASE_BUILD`.

phuchau1989 (approved):

Issue #13155, opened 2022-11-11
Bug Report - Unexpected transform update for container entity in prefab template


forkercat:
**Describe the bug**

When we add/remove entity or prefab under a focused container entity, there will be an unexpected patch for updating the container entity's transform in the source template. The container entity's transform in source template is supposed to be empty (as it will be replaced by linkDom during template-template propagation).

So, if users continue saving the changes, that update will be saved to disk, then it would likely cause more weird issues in the future.

Note:
1. It is not obvious without looking into the DOM representation. Operations in Editor might work without any noticeable issues.
2. CreatePrefab/Instantiate prefab might also have this issue. Check PrefabUndoHelpers::UpdateEntity.

**Steps to reproduce**
1. Create an empty level
3. Create a prefab with one nested entity
4. Focus on the prefab
5. In source code, adds a breakpoint in PrefabUndoAddEntity.cpp#L63 to observe what are generated in `m_redoPatch`.
6. Add/Delete an entity **under the container entity**

**Actual behavior**

There are 3 patches. One extra patch to update the transform component.

![Create Entity](https://user-images.githubusercontent.com/11157226/201429042-239182c5-e8ca-4547-ae44-660f3bf487e0.png)

![RedoPatch Data](https://user-images.githubusercontent.com/11157226/201429056-d0a854e7-e1d2-49af-bfa7-2b70e351949b.png)

**Expected behavior**

There should only be one patch for adding the entity and one patch to update parent entity's child sort order.

**Found in Branch**

development https://github.com/o3de/o3de/commit/772758ec0e045fb8638c792eac7fb2fe24c3c798

**Desktop/Device (please complete the following information):**

* Device: PC
* OS: Windows 10
* CPU: AMD Ryzen 5950x
* GPU: NVIDIA RTX 3070
* Memory: 64 GB

**Additional Context**

For override editing, this won't be an issue because the parent entity that needs to be updated can't be the focused container entity.

forkercat:
Proposed Fix: Check if the parent is the focused container entity. If it is, we call InstanceDomGenerator's GenerateContainerEntityDomFromRootInstance or GenerateEntityDom to get the DOM; if it is not, it is safe to fetch from focused template.

I noticed this in unit testing. How might we avoid this kind of issues by verifying the patches in unit testing?

forkercat:
This PR resolves the issue for delete-workflow https://github.com/o3de/o3de/pull/13171

srikappa-amzn:
[...]

PR #13261, opened 2022-11-17
Prefab Bug Fix - Remove unexpected transform update of container entity for add-entity

forkercat:

## What does this PR do?

Fix: https://github.com/o3de/o3de/issues/13155

This PR addresses the issue of an unexpected patch for updating the parent container entity's transform in the source template. During add-entity operation, if the container entity is being focused and we fetch from focused template, the transform in the retrieved entity DOM will be empty. An unexpected patch will be generated to update the transform.

- Patch 1: Add new entity DOM
- Patch 2: Update parent's child entity order
- **Patch 3: Update parent entity's transform    <-- unexpected**

In this case, we need to change it to "fetch from root" such that the transform contains data and later patch generation would not generate an extra patch to update parent transform.

## How was this PR tested?

The recently added unit tests could not verify the redo/undo patches. With changes of connecting PrefabTestFixture to EditorRequestBus going in (https://github.com/o3de/o3de/pull/13211), we should be able to correctly capture the behavior of parent update patches and add some new validations for patch types in undo node unit tests.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoAddEntity.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoAddEntity.cpp
@@ -53,10 +53,10 @@ namespace AzToolsFramework
             // contain the transform data, which may generate unexpected patches to update transform.
             if (parentEntityId == focusedInstance.GetContainerEntityId())
             {
-                PrefabDom parentContainerEntityDom;
-                m_instanceDomGeneratorInterface->GenerateEntityDom(parentContainerEntityDom, parentEntity);
+                PrefabDom parentEntityDomBeforeAddingEntity;
+                m_instanceDomGeneratorInterface->GenerateEntityDom(parentEntityDomBeforeAddingEntity, parentEntity);
 
-                if (parentContainerEntityDom.IsNull())
+                if (parentEntityDomBeforeAddingEntity.IsNull())
                 {
                     AZ_Error("Prefab", false, "PrefabUndoAddEntity::Capture - "
                         "Cannot retrieve parent container entity DOM from root template via instance DOM generator.");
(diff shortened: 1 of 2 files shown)


Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoAddEntity.cpp
```diff
@@ -46,21 +47,50 @@ namespace AzToolsFramework
...
+            // not contain transform data seen from root, but after-state entity DOM from serialization always
+            // contain the transform data, which may generate unexpected patches to update transform.
+            if (parentEntityId == focusedInstance.GetContainerEntityId())
             {
-                // DOM value pointers can't be relied upon if the original DOM gets modified after pointer creation.
-                const PrefabDomValue* parentEntityDomInFocusedTemplate = entityPathInFocusedTemplate.Get(focusedTemplateDom);
-                AZ_Assert(parentEntityDomInFocusedTemplate,
-                    "Could not load parent entity's DOM from the focused template's DOM. "
-                    "Focused template id: '%llu'.", static_cast<AZ::u64>(m_templateId));
+                PrefabDom parentContainerEntityDom;
```
srikappa-amzn, line 56: nit: Should this be called parentEntityDomBeforeAddingEntity for consistency?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoAddEntity.cpp
```diff
@@ -46,21 +47,50 @@ namespace AzToolsFramework
...
+                m_instanceDomGeneratorInterface->GenerateEntityDom(parentContainerEntityDom, parentEntity);
```
srikappa-amzn, line 57: yeah...the names are definitely confusing. It took me a while to figure out that GenerateDomForEntity and GenerateEntityDom are actually 2 different functions doing 2 different things. Please add a task to rename these

srikappa-amzn (approved):
Thanks for proactively fixing this!

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoAddEntity.cpp
forkercat: Do you think we still need InstanceDomGenerator? I think these two functions fit better in InstanceToTemplatePropagator. And keeping them near GenerateDomForEntity and GenerateDomForInstance would be good for us to maintain and to be clear about their functionalities imo, or vice versa moving the other two to InstanceDomGenerator.

AMZN-daimini (approved):
Agree with the direction of adding some unit tests to prevent regressions as a follow up :)

Issue #12847, opened 2022-10-27
Multiplayer warning whenever an entity is removed (properly with MarkForRemoval)


AMZN-Olex:
**Describe the bug**
Trying to isolate the issue to help the community to reproduce it easily and increase chances for a fast fix. MultiplayerScriptingSample level spawns a projectile and removes them after a few seconds of flight time. For each removed projectiles from the server correctly, a warning is printed.

```
[Warning] (System) - Replicator for id 8303078270375559194 is null on remote host 127.0.0.1:33450. It may have already been deleted.
```

**Assets required**
o3de-multiplayersample new MultiplayerScriptingSample level https://github.com/aws-lumberyard-dev/o3de-multiplayersample/tree/Network/olex/lbc_10_2022

**Steps to reproduce**
1. Open MultiplayerScriptingSample  level
2. Enter game mode
3. Left click to fire
4. See warnings in the editor console

**Expected behavior**
No warnings occur when a multiplayer entity is removed correctly.

**Actual behavior**
```
[Warning] (System) - Replicator for id 8303078270375559194 is null on remote host 127.0.0.1:33450. It may have already been deleted.
[Warning] (System) - Replicator for id 8303078270375559195 is null on remote host 127.0.0.1:33450. It may have already been deleted.
[Warning] (System) - Replicator for id 8303078270375559196 is null on remote host 127.0.0.1:33450. It may have already been deleted.
[Warning] (System) - Replicator for id 8303078270375559197 is null on remote host 127.0.0.1:33450. It may have already been deleted.
[Warning] (System) - Replicator for id 8303078270375559198 is null on remote host 127.0.0.1:33450. It may have already been deleted.
[Warning] (System) - Replicator for id 8303078270375559199 is null on remote host 127.0.0.1:33450. It may have already been deleted.
[Warning] (System) - Replicator for id 8303078270375559200 is null on remote host 127.0.0.1:33450. It may have already been deleted.
[Warning] (System) - Replicator for id 8303078270375559201 is null on remote host 127.0.0.1:33450. It may have already been deleted.
[Warning] (System) - Replicator for id 8303078270375559202 is null on remote host 127.0.0.1:33450. It may have already been deleted.
```

**Found in Branch**
development

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Windows
 - Version 11

AMZN-puvvadar:
[...]

Issue #12137, opened 2022-09-22
Reliable RPC Isn't Received If Called OnActivate()


AMZN-Gene:
**Describe the bug**
Sending a reliable RPC from authority to client wont be received by the client if it's called on entity activation.

**Steps to reproduce**
1. Create a Multiplayer autocomponent with an RPC that's sent from Authority to Client
   Example, sending clients the current player id for them to display locally.
```
<RemoteProcedure Name="PlayerIdRPC" InvokeFrom="Authority" HandleOn="Client" IsPublic="true" IsReliable="true" GenerateEventBindings="false" Description="">
        <Param Type="int" Name="PlayerId" />
</RemoteProcedure>

```

**Expected behavior**
The RPC Handler is hit on the clients 

**Actual behavior**
The RPC is never received.

**Found in Branch**
development latest

lmbr-pip:
This maybe valid behaviour, may require a warning that its not callable in OnActivate. Entity activation is generally too early, may need more investigation.

lmbr-pip:
Or provide new "ready for network traffic" call.

AMZN-Gene:
1. Add a warning if calling RPC too soon
2. Create new a method called "ReadyForNetworkMessages()" so you know when it's okay

AMZN-Gene:
Verified, this _should_ be working.

AMZN-puvvadar:
> Verified, this _should_ be working.

I took a look at this and I have to disagree. RPC Handlers are bound by their `EntityReplicactor`. This occurs through `EntityReplicationManager` on an event based Tick in which new windows are processed. This essentially means that that RPC handler registration is deferred a tick. Consequently, RPCs should be not be invoked at Activation time.

A potential fix here would be to drive this behavior at the top of `NetBindComponent::Activate` so there's immediate availability. I have not investigated what the feasibility or impact of such a change would be however. Another would be to add a NetworkEntity specific lifecycle event for the ReplicationWIndow becoming ready.

AMZN-Gene:
> > Verified, this _should_ be working.
> 
> I took a look at this and I have to disagree. RPC Handlers are bound by their `EntityReplicactor`. This occurs through `EntityReplicationManager` on an event based Tick in which new windows are processed. This essentially means that that RPC handler registration is deferred a tick. Consequently, RPCs should be not be invoked at Activation time as handlers won't be bound.
> 
[...]

PR #13282, opened 2022-11-18
Synchronizing ReplicationWindow and EntityReplicationManager updates

AMZN-puvvadar:

## What does this PR do?

This PR addresses timing issues between the EntityReplicationManager and Replication Windows. The primary issue is the EntityReplicationManager consumes info from Replication Windows but the two update at different rates. This generally means the EntityReplicationManager is behind the data in the window.

This change moves window updates to be driven from EntityReplicationManager. It also has immediate handling for Activating/Deactivating entities when possible. One exception to this case is the window's controlled entity. In order for the window to function, the controlled entity needs to be activated at the window's construction time.

Related issues:
https://github.com/o3de/o3de/issues/12137
https://github.com/o3de/o3de/issues/12847

This also addresses a bug where if a packet failed to Handle, it would fall through to the next packet type vs returning a failure.

## How was this PR tested?

Breakpoint based timing tests and regression testing.

## Suggested future changes

1. Removing EntityReplicationManager tick entirely and only updating on demand. This requires a clear picture of how net entities are added throughout the simulation. This includes how starting simulation entities are added and how entities are added throughout the sim.
2. Adding a OnNetworkInitialized function to MultiplayerComponent and its descendants so users can take immediate action when a NetEntity is ready. This can be AZ::Event driven under the hood via codegen.


AMZN-Gene (approved):

Gems/Multiplayer/Code/Source/Debug/MultiplayerDebugSystemComponent.cpp
```diff
@@ -69,6 +69,10 @@ namespace Multiplayer
     {
 #ifdef IMGUI_ENABLED
         ImGui::ImGuiUpdateListenerBus::Handler::BusDisconnect();
+        m_auditTrailElems.clear();
```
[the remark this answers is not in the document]
AMZN-puvvadar: I would occasionally see issues around these deques during teardown so just cleaning them out here.

AMZN-Olex (approved):

Gems/Multiplayer/Code/Include/Multiplayer/ReplicationWindows/IReplicationWindow.h
```diff
@@ -50,6 +50,15 @@ namespace Multiplayer
         //! @param outNetworkRole output containing the network role of the requested entity if found
         virtual bool IsInWindow(const ConstNetworkEntityHandle& entityPtr, NetEntityRole& outNetworkRole) const = 0;
 
+        //! Adds an entity to the replication window's set
+        //! @param entity The entity to try adding
+        //! @return Whether the entity was able to be added
```
AMZN-Gene, line 55: I'm curious why adding might return false; adding an example would be nice. 
_"Each window is left to decide if it wishes to block an entity from being added; for example, ServerToClientWindow may wish to block entities outside of a client's awareness radius."_

AMZN-Gene (approved):

Gems/Multiplayer/Code/Include/Multiplayer/ReplicationWindows/IReplicationWindow.h
AMZN-puvvadar: It's really dependent on the implementation of the specific ReplicationWindow.

AMZN-Gene: On first read I thought a false would be because of an error. Maybe because the comment "was able to be added", I put a lot of weight on ABLE.  It sounded like "we're going to try to add, but something might go wrong". Only after reading the implementation did I see "oh, not adding is desired; a good thing"

phuchau1989 (approved):

PR #13287, opened 2022-11-18
Switch to reset deleted entity in prefab delete-entity operation

forkercat:

## What does this PR do?

This PR changes DeleteEntity call to reset() in prefab delete operation.

Difference: [ComponentApplication::DeleteEntity](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp#L1139) also signals "entity removed event".

Since not noticing any of this event handler being registered (i.e. empty), we discussed that we can try to switch to calling reset() to keep consistency with how it's deleting entities in nested instances `detachedInstance.reset()`.

**Note:** We can revert this commit later if we see new issue because of this change.

Before:

```cpp
AZ::ComponentApplicationBus::Broadcast(&AZ::ComponentApplicationRequests::DeleteEntity, entityId);
```

After:

```cpp
commonOwningInstance->get().DetachEntity(entityId).reset();
```

See previous discussion here: https://github.com/o3de/o3de/pull/13171#discussion_r1022152546

## How was this PR tested?

Manually tested delete operation. Passed tests.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
@@ -1408,7 +1408,6 @@ namespace AzToolsFramework
             {
                 const AZ::EntityId entityId = entity->GetId();
                 const AZStd::string nestedEntityAliasPath = m_instanceToTemplateInterface->GenerateEntityAliasPath(entityId);
-                detachedEntityAliasPaths.push_back(AZStd::move(nestedEntityAliasPath));
 
                 // Captures the parent entity that needs to be updated.
                 AZ::EntityId parentEntityId;
@@ -1419,7 +1418,9 @@ namespace AzToolsFramework
                     parentEntitiesToUpdate.insert(GetEntityById(parentEntityId));
                 }
 
-                commonOwningInstance->get().DetachEntity(entityId).reset();
+                commonOwningInstance->get().DetachEntity(entityId);
+
+                detachedEntityAliasPaths.push_back(AZStd::move(nestedEntityAliasPath));
             }
 
             // 3. Updates template DOM with undo/redo support.
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
@@ -1408,6 +1408,7 @@ namespace AzToolsFramework
             {
                 const AZ::EntityId entityId = entity->GetId();
                 const AZStd::string nestedEntityAliasPath = m_instanceToTemplateInterface->GenerateEntityAliasPath(entityId);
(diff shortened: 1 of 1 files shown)


Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -1418,10 +1419,7 @@ namespace AzToolsFramework
                     parentEntitiesToUpdate.insert(GetEntityById(parentEntityId));
                 }
 
-                commonOwningInstance->get().DetachEntity(entityId).release();
```
tkothadev, line 1421: What was the difference between `release` and `reset`?

forkercat: In this case, DetachEntity() returns a unique ptr which holds the ownership of the memory. `release` just releases the ownership but does not actually free the memory.

In the previous workflow, we then depend on ComponentApplication::DeleteEntity call to delete the entity object and free the memory. In the new changes, we just call reset by ourselves on the unique ptr. `reset` frees the memory.

forkercat: I think common use case is assign to a local variable after `release`, otherwise the memory is leaked (no longer being tracked).

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -1408,6 +1408,7 @@ namespace AzToolsFramework
             {
                 const AZ::EntityId entityId = entity->GetId();
                 const AZStd::string nestedEntityAliasPath = m_instanceToTemplateInterface->GenerateEntityAliasPath(entityId);
+                detachedEntityAliasPaths.push_back(AZStd::move(nestedEntityAliasPath));
```
srikappa-amzn, line 1411: nit: I think it's good to add this to the paths after the actual detaching happens. If this code changes in the future and we introduce an early return between this and the actual detach, it can cause issues.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -1418,10 +1419,7 @@ namespace AzToolsFramework
                     parentEntitiesToUpdate.insert(GetEntityById(parentEntityId));
                 }
 
-                commonOwningInstance->get().DetachEntity(entityId).release();
-                AZ::ComponentApplicationBus::Broadcast(&AZ::ComponentApplicationRequests::DeleteEntity, entityId);
-
-                detachedEntityAliasPaths.push_back(AZStd::move(nestedEntityAliasPath));
+                commonOwningInstance->get().DetachEntity(entityId).reset();
```
srikappa-amzn, line 1422: nit: Don't need to explicitly reset the entity as it'll be destroyed at the end of this scope

srikappa-amzn (approved):

PR #13288, opened 2022-11-18
Cached projected shadows

invertednormal:
## What does this PR do?

This PR add support for shadows which will only update when they detect a change somewhere in their field of view. The option must be turned on inside the light component. Detecting that something moved is done as part of the view frustum cull, so the cost on the CPU side remains roughly the same as normal shadows, but in cases where the shadow doesn't need to re-render, the rendering cost is saved. This feature lends itself to scenes that are largely static like arch-vis.

This feature should be able to be made even cheaper in the future once we support the concept of static meshes. That functionality would eliminate the need for static lights to check against static meshes most of the time.

image

## Details

Tracking whether meshes moved for a given view was done in an abstract way where views can have arbitrary bit flags stored on them in a 32 bit uint. The meaning of each flag is determined at runtime - a system can register some bit by name, for instance "MeshMoved", then get a bit position back in the flag array. Cullables can also use those flags and accumulate the OR of all cullable flags on a view. So if a mesh moves, it marks the "MeshMoved" bit in its cullable, which then is applied to any views that it's rendered in. The projected shadow map pass then checks this bit if it's a cached shadow to determine if it should re-render due to a mesh moving.

Certain meshes may need to be considered to be always moving, for instance if they are skinned or have vertex shader animation. Skinned meshes do this automatically, but regular meshes also allow you to mark them as always moving.

image

The one downside of this approach is that systems that don't use the standard culling need to make sure they set the bit manually themselves. Currently the only known system that does this is terrain, and there should be improvements to the culling system in the future that will make it flexible enough for terrain to use.

## Other stuff

This change also fixed a couple bugs. 
- Terrain was recalculating its candidate sectors to draw every frame even if nothing changed.
- ESM shadow passes could at times continue to calculate in the background even when the light was no longer using esm shadows.

## How was this PR tested?

Most testing was done in the loft scene in both DirectX and Vulkan. Terrain was also tested separately since it had a separate implementation.


Gems/AtomLyIntegration/CommonFeatures/Code/Source/Mesh/EditorMeshComponent.cpp
```diff
@@ -76,6 +76,7 @@ namespace AZ
                             ->DataElement(AZ::Edit::UIHandlers::Default, &MeshComponentConfig::m_sortKey, "Sort Key", "Transparent meshes are first drawn by sort key, then depth. Use this to force certain transparent meshes to draw before or after others.")
                             ->DataElement(AZ::Edit::UIHandlers::CheckBox, &MeshComponentConfig::m_excludeFromReflectionCubeMaps, "Exclude from reflection cubemaps", "Model will not be visible in baked reflection probe cubemaps")
                                 ->Attribute(AZ::Edit::Attributes::ChangeNotify, Edit::PropertyRefreshLevels::ValuesOnly)
+                            ->DataElement(AZ::Edit::UIHandlers::CheckBox, &MeshComponentConfig::m_isAlwaysDynamic, "Always Moving", "Forces this mesh to be considered to always be moving, even if the transform didn't update. Useful for meshes with vertex shader animation.")
```
amzn-tommy, line 79: Eventually it would be nice to be able to drive the always-dynamic behavior automatically via the material in some way. But I do think this will remain useful as a fallback for a user to be able to say 'hey, I know this needs to be re-rendered every frame, and the automatic systems aren't picking it up correctly'.

I do think this probably belongs a little further down in the UI below IsRayTracingEnabled, since it's a less-often used feature.

Gems/Atom/Feature/Common/Code/Include/Atom/Feature/Mesh/MeshFeatureProcessor.h
```diff
@@ -194,13 +199,15 @@ namespace AZ
             RPI::Cullable::LodConfiguration GetMeshLodConfiguration(const MeshHandle& meshHandle) const override;
 
             void SetExcludeFromReflectionCubeMaps(const MeshHandle& meshHandle, bool excludeFromReflectionCubeMaps) override;
+            void SetIsAlwaysDynamic(const MeshHandle& meshHandle, bool isAlwaysDynamic) override;
```
amzn-tommy, line 202: You're going to need to update MockMeshFeatureProcessor.h with these or some unit tests won't compile.

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/View.h
```diff
@@ -213,6 +229,9 @@ namespace AZ
 
             // Masked Occlusion Culling interface
             MaskedOcclusionCulling* m_maskedOcclusionCulling = nullptr;
+
+            AZStd::atomic_uint32_t m_andFlags{ 0xFFFFFFFF };
```
amzn-tommy, line 233: I think we would benefit from a document laying out how all these interrelated flags/masks are used if we don't already have one...

We have 
PipelineViewTag
DrawListMask/DrawListTags
UsageFlags
AndFlags
OrFlags

plus ShaderOptionFlags on the cullable. I might even be forgetting some.

It's pretty daunting to a new user.

amzn-tommy (commented):
Not many comments. Great work!

Gems/Atom/Feature/Common/Assets/Shaders/Shadow/ClearShadowmap.azsl
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#pragma once
```
jeremyong-az, line 9: Needed?

Gems/Atom/Feature/Common/Code/Source/CoreLights/ProjectedShadowmapsPass.cpp
```diff
@@ -130,9 +156,32 @@ namespace AZ
...
 
-                    pass->SetClearEnabled(!sliceIsCleared[origin.m_arraySlice]);
-                    sliceIsCleared[origin.m_arraySlice] = true;
+                    SliceInfo& sliceInfoItem = sliceInfo.at(origin.m_arraySlice);
+                    sliceInfoItem.m_shadowPasses.push_back(pass);
+                    sliceInfoItem.m_hasStaticShadows = sliceInfoItem.m_hasStaticShadows || it.m_isCached;
+                }
+            }
+
+            for (auto it : sliceInfo)
```
jeremyong-az, line 167: would you want `const auto&` here to avoid copying the vector of shadow pass pointers?

Gems/Atom/Feature/Common/Code/Source/CoreLights/ProjectedShadowmapsPass.cpp
```diff
@@ -210,6 +270,44 @@ namespace AZ
             return ShadowmapPass::CreateWithPassRequest(passName, passData);
         }
 
+        void ProjectedShadowmapsPass::CreateClearShadowDrawPacket()
+        {
+            // Force load of shader to clear shadow maps.
+            const AZStd::string clearShadowShaderFilePath = "Shaders/Shadow/ClearShadow.azshader";
```
jeremyong-az, line 276: I think I noticed that the clear shadow shader was doing a fullscreen triangle VS approach. It might be worth doing the clear with a render target clear since the hardware can do faster clears operating over the metadata (as opposed to per-pixel). Maybe there's a reason this isn't possible that I'm missing?

Gems/Atom/Feature/Common/Code/Source/CoreLights/ShadowmapPass.h
```diff
@@ -46,12 +47,24 @@ namespace AZ
             //! This enable/disable clearing of the image view.
             void SetClearEnabled(bool enabled);
 
+            //! Sets this shadow pass to only update its shadows when a mesh in view moves.
+            void SetIsStatic(bool isSatic);
```
jeremyong-az, line 51: `isStatic`

Gems/Atom/RPI/Code/Source/RPI.Public/View.cpp
```diff
@@ -125,6 +125,33 @@ namespace AZ
             m_drawListContext.AddDrawItem(drawListTag, drawItemProperties);
         }
 
+        void View::ApplyFlags(uint32_t flags)
```
jeremyong-az, line 128: Definitely worth documenting how these flags are used I think - without seeing this PR or searching to eventually find the various "caster moved" bits it's not easy to see 😉

jeremyong-az (approved):

Gems/Atom/Feature/Common/Code/Source/CoreLights/ShadowmapPass.cpp
```diff
@@ -140,10 +157,62 @@ namespace AZ
...
+                {
+                    // Only decrement startIndex if startIndex is greater than 0. This means that RasterPass will submit one less
+                    // draw item for the first batch since the clear draw was submitted in this batch.
+                    --startIndex; 
+                }
+                // RasterPass's draw item indices need to be adjusted to make sure it stays inside its expected bounds.
+                --endIndex;
+                ++offset;
+            }
+            Base::SubmitDrawItems(context, startIndex, endIndex, offset);
```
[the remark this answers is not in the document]
invertednormal: This is the result of some strict validation that is done on the command list where you need to pass the index of the thing you're submitting purely for validation purposes.

RasterPass handles most of the submission. Previously it was basically impossible to add your own draw items in a sub class, but I made changes to make this possible.  `BuildCommandListInternal()` now calls `SubmitDrawItems()` which can be overridden.

So, the clear item _is_ being submitted here as the first item - `commandList->Submit(*m_clearShadowDrawPacket->GetDrawItem(0).m_item, 0);` However, the indices need to be adjusted for the base class so that it submits the correct items with the correct indices and doesn't go out of bounds (ShadowmapPass sets the estimated item count to +1 when there's a clear draw, which the raster pass isn't aware of).

Longer term I think we should relax the validation to make this less ugly. There used to be no validation at all, and it was easy for people to not realize that `BuildCommandListInternal()` could be called multiple times per pass with different submit ranges. However, I think just making sure that you submit the correct number of items should be enough, making sure to index them all seems overkill.

Gems/AtomLyIntegration/CommonFeatures/Code/Source/CoreLights/EditorAreaLightComponent.cpp
```diff
@@ -172,16 +172,24 @@ namespace AZ
                             ->Attribute(Edit::Attributes::Visibility, &AreaLightComponentConfig::SupportsShadows)
                             ->Attribute(Edit::Attributes::ReadOnly, &AreaLightComponentConfig::IsEsmDisabled)
                         ->DataElement(
-                            Edit::UIHandlers::Slider, &AreaLightComponentConfig::m_normalShadowBias, "Normal Shadow Bias",
+                            Edit::UIHandlers::Slider, &AreaLightComponentConfig::m_normalShadowBias, "Normal shadow bias",
```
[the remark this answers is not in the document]
invertednormal: Ugh, I better check. Good callout.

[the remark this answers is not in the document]
invertednormal: Needed one update, thankfully all the python scripts reference one file for all the constants `AutomatedTesting\Gem\PythonTests\Atom\atom_utils\atom_constants.py`

Gems/AtomLyIntegration/CommonFeatures/Code/Source/Mesh/EditorMeshComponent.cpp
invertednormal: I agree about the material thing - was actually chatting with Brendan about that. I think that'll just take a bit more work to figure out the right way to do though. I need to chat with Chris about that while he's still here. 

As for where the property shows up in the UI, I'll move it down to just above LOD stuff.

Gems/Atom/Feature/Common/Assets/Shaders/Shadow/ClearShadowmap.azsl
invertednormal: over-aggressive copy-pasting of copyright header :D. Will fix.

Gems/Atom/Feature/Common/Code/Source/CoreLights/ProjectedShadowmapsPass.cpp
invertednormal: Doing the shader approach because certain platforms *cough* metal *cough* can't do clears of regions of render targets. Since the shadows use an atlas, it's possible some items in the atlas are cached and some aren't, so only certain viewports need to be cleared. While there is support for this kind of thing in DirectX and Vulkan, we haven't plumbed it into our RHI, and it's not support in metal at all.

invertednormal: I do regular full clears in cases where an atlas has no cached shadows, so the extra cost is only when there's a mix.

Gems/Atom/Feature/Common/Code/Source/CoreLights/ProjectedShadowmapsPass.cpp
invertednormal: Whew. The danger of auto. Good catch.

Gems/Atom/RPI/Code/Source/RPI.Public/View.cpp
invertednormal: yeah, keeping things really decoupled can be a two-edged sword. Makes it really easy for other system to inject their own bits for whatever they might want to track, but it also makes it harder for someone reading the code to figure out how it all fits together.

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/View.h
invertednormal: Yeah it's a lot, but I don't think there's a lot of interactions, they're used in different places for different things. But I agree some documentation with regards to the flags on the cullable and how they relate to views would be worth doing, probably once the culling system has been through its refactoring.

Gems/AtomLyIntegration/EMotionFXAtom/Code/Source/AtomActorInstance.cpp
```diff
@@ -652,6 +652,8 @@ namespace AZ::Render
 
             m_meshHandle = AZStd::make_shared<MeshFeatureProcessorInterface::MeshHandle>(
                 m_meshFeatureProcessor->AcquireMesh(meshDescriptor, materials));
+
+            m_meshFeatureProcessor->SetIsAlwaysDynamic(*m_meshHandle, true);
```
amzn-tommy, line 656: nit: If you add m_isAlwaysDynamic to the MeshHandleDescriptor, then you can set that on the descriptor without needing to make a separate call.

Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
```diff
@@ -533,6 +542,24 @@ namespace AZ
             }
         }
 
+        void MeshFeatureProcessor::SetIsAlwaysDynamic(const MeshHandle & meshHandle, bool isAlwaysDynamic)
+        {
+            if (meshHandle.IsValid())
+            {
+                meshHandle->m_isAlwaysDynamic = isAlwaysDynamic;
```
amzn-tommy, line 549: You should update ModelDataInstance::MeshLoader::OnAssetReady to set m_isAlwaysDynamic in the
`if (m_parent->m_descriptor.m_requiresCloneCallback &&
                m_parent->m_descriptor.m_requiresCloneCallback(modelAsset))
            {`
condition. That way cloth is set to be dynamic.

amzn-tommy (approved):

Issue #12932, opened 2022-11-01
IMGUI Unusable: Mouse Doesn't Work


AMZN-Gene:
**Describe the bug**
The mouse cannot select IMGUI menu items

**Steps to reproduce**
1. Open a game or serve launcer
2. Press HOME key to open IMGUI
3. Try to select an IMGUI option

**Expected behavior**
I should be able to open the IMGUI menus and select buttons and such

**Actual behavior**
IMGUI is unusable. The mouse is stuck on the top left of the screen and I cannot select IMGUI menu items

**Screenshots/Video**
**Notice my Windows mouse is moving around, but the app mouse is stuck in upper-left-hand corner**
https://user-images.githubusercontent.com/32776221/199294797-4ec62a89-f702-47a4-b0e8-cd286ef44c62.mp4

**Found in Branch**
Development Friday October 28 2022

**Desktop/Device (please complete the following information):**
Windows

AMZN-Gene:
Setting to critical, not blocker because I think most IMGUI options are possible via cvars?

galibzon:
@santorac Did some work around ImGui mouse a few weeks ago. Will assign to him to see if he can identify an easy fix.

AMZN-puvvadar:
@galibzon Is there any update on this? This renders certain ImGui based tools completely unusable.

lmbr-pip:
Can you use arrow keys + space bar to navigate? Can you change behaviour by toggling input mode? See https://github.com/o3de/o3de.org/pull/2074

AMZN-puvvadar:
> Can you use arrow keys + space bar to navigate? Can you change behaviour by toggling input mode? See [o3de/o3de.org#2074](https://github.com/o3de/o3de.org/pull/2074)

You can use arrows + space to navigate and action some of the menus but many of the windows that are then opened are not operable.

santorac:
I've been doing a "git bisect" to track this down. I have to step away from this for a few days, though. Here's what I have so far:
```
D:\o3de>git bisect log
# bad: [c59c38219a33913b87753f57019009976da85607] Merge pull request #13170 from aws-lumberyard-dev/jckand/AddEntityTestFix
# good: [a7baa576b413e24c6503e2aab45cc89947b7318d] Merge pull request #11124 from aws-lumberyard-dev/Atom/santorac/FixImGuiManagerMouseInput
git bisect start 'development' 'a7baa576b413e24c6503e2aab45cc89947b7318d'
[...]

PR #12609, opened 2022-10-14
Improved CPU performance by almost 13ms (37.48ms -> 24.74ms) on the RosCon level

antonmic:
Improved CPU performance with the following optimizations:
- Removed a redundant call to UpdateVisibility at the end of the frame (~3ms)
- Switched culling to use jobs instead of task graph (~2.5ms)
- Switch the tools render pipeline to be off when created since it gets manually turned on and then off when it needs to do a capture (~0.5ms)
- Switch ImGui update to be multithreaded (up to 5ms when CPU profiler is running)

Those were the big ones if I'm not forgetting any. Other changes here are small things like removing perf markers that are too small to show up in profiling, adding perf markers for tasks that are definitely big enough to profile,

PR #13292, opened 2022-11-19
Fixed IMGUI Unusable: Mouse Doesn't Work in Game Launcher

santorac:
## What does this PR do?

Fixes https://github.com/o3de/o3de/issues/12932

The problem was introduced at https://github.com/o3de/o3de/pull/12609 which moved ImGuiManager rendering to a job thread, as an optimization. I simply reverted that part of the changes, as there isn't time for a thorough investigation and proper fix, and I'm concerned there might be more threading issues yet to be uncovered. If we really want ImGuiManager to render on a separate thread, someone else will have to do a deep dive.

## How was this PR tested?

The original change is reported to have saved 5ms, but only when the ImGui CPU profiler tool was open, and on a specific level. I checked the CPU profiler performance in Editor.exe before and after the change. When I tested using the default scene, the main thread took about 2.75 ms before my change and 3.25ms after.

I tested Editor.exe and the GameLauncher to make sure the mouse worked.

### DIFF
--- a/Gems/AtomLyIntegration/ImguiAtom/Code/Source/ImguiAtomSystemComponent.cpp
+++ b/Gems/AtomLyIntegration/ImguiAtom/Code/Source/ImguiAtomSystemComponent.cpp
@@ -95,13 +95,6 @@ namespace AZ
 #endif
         }
 
-        void ImguiAtomSystemComponent::WaitForRender()
-        {
-#if defined(IMGUI_ENABLED)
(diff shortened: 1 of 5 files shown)


rgba16f (approved):

AMZN-Gene (approved):

moraaar (approved):
Great work finding what caused the issue @santorac.

AMZN-puvvadar (approved):

cgalvan (approved):

PR #13299, opened 2022-11-20
Draw icon for add entity as override

michabr:
## What does this PR do?

Displays a "+" icon next to entities that have been added as a prefab override.

## How was this PR tested?

Manual testing in the editor.

OutlinerAddEntityOverride


tkothadev:
Should the icon be a `+` or a `*`? I'm thinking asterisk would better illustrate its nature as an overriden prefab.

AMZN-daimini:
> Should the icon be a `+` or a `*`? I'm thinking asterisk would better illustrate its nature as an overriden prefab.

The circle represents an entity that exists in the prefab but has property overrides; the + represents an entity that does not exist in the prefab but is added by one of the ancestors as an override.
We already use the asterisk to signify "unsaved" so I'm not sure if re-using it here would cause confusion to be honest.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Link/Link.cpp
```diff
@@ -135,6 +135,24 @@ namespace AzToolsFramework
             return areOverridesPresent;
         }
 
+        PrefabDomConstReference Link::FindOverridePatch(
+            AZ::Dom::Path path,
+            AZ::Dom::PrefixTreeTraversalFlags prefixTreeTraversalFlags)
```
forkercat, line 140: looks like it can be const.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverrideHandler.cpp
```diff
@@ -60,5 +81,44 @@
...
             return false;
         }
-    }
+
+        AZStd::optional<EntityOverrideType> PrefabOverrideHandler::GetOverrideType(AZ::Dom::Path path, LinkId linkId) const
+        {
+            AZStd::optional<EntityOverrideType> overrideType = {};
+
+            PrefabSystemComponentInterface* prefabSystemComponentInterface = AZ::Interface<PrefabSystemComponentInterface>::Get();
+            if (prefabSystemComponentInterface != nullptr)
```
forkercat, line 90: nit: Can reuse `m_prefabSystemComponentInterface`

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverrideHandler.cpp
```diff
@@ -8,13 +8,34 @@
...
+            // an entity object or adding/removing an entity property
+            bool IsDirectEntityPatch(PrefabDomConstReference patch)
+            {
+                auto pathMember = patch->get().FindMember("path");
+                if (pathMember != patch->get().MemberEnd())
+                {
+                    AZ::Dom::Path path(pathMember->value.GetString());
+                    if (path.Size() > 1)
+                    {
+                        return (path[path.Size() - 2] == PrefabDomUtils::EntitiesName);
```
forkercat, line 31: Would be helpful to have some comment.
Example direct entity patch path: `/Entities/Entity_[123213123123]`

forkercat (approved):
Left some comments. Nice work!

gadams3 (approved):

rainbj:
Looks great in the outliner :)

Code/Framework/AzQtComponents/AzQtComponents/Images/resources.qrc
```diff
@@ -16,6 +16,7 @@
         <file alias="entity_editoronly.svg">Entity/entity_editoronly.svg</file>
         <file alias="entity_notactive.svg">Entity/entity_notactive.svg</file>
         <file alias="entity_overridden.svg">Entity/entity_overridden.svg</file>
+        <file alias="entity_override_add.svg">Entity/entity_override_add.svg</file>
```
srikappa-amzn, line 19: nit: entity_added_as_override might be a more clearer name.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverridePublicInterface.h
```diff
@@ -26,6 +27,12 @@ namespace AzToolsFramework
             //! @return true if overrides are present on the given entity id.
             virtual bool AreOverridesPresent(AZ::EntityId entityId) = 0;
 
+            //! Gets the override type on the given entity id. The prefab that creates the overrides is identified
+            //! by the class implmenting this interface based on certain selections in the editor. eg: the prefab currently being edited.
+            //! @param entityId The id of the entity for which to get the override type.
+            //! @return an override type if an override exists on the given entity id.
+            virtual AZStd::optional<EntityOverrideType> GetOverrideType(AZ::EntityId entityId) = 0;
```
srikappa-amzn, line 34: I think this should just be called 'OverrideType' since it'll be possible to create a 'AddPrefab' override in the future too.

Also, 'RemovePrefab' should be one of the types supported as it is already a supported operation

srikappa-amzn (changes requested):

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabUiHandler.h
```diff
@@ -89,8 +91,9 @@ namespace AzToolsFramework
         QString m_prefabEditCloseIconPath = QString(":/Entity/prefab_edit_close.svg");
 
         inline static const QColor s_overrideIconBackgroundColor = QColor("#444444");
-        inline static const QPoint s_overrideIconOffset = QPoint(10, 10);
-        inline static const int s_overrideIconSize = 5;
-        QIcon s_overrideIcon = QIcon(QString(":/Entity/entity_overridden.svg"));
+        inline static const QPoint s_overrideIconOffset = QPoint(12, 12);
+        inline static const int s_overrideIconSize = 8;
+        QIcon m_overrideIconReplace = QIcon(QString(":/Entity/entity_overridden.svg"));
```
srikappa-amzn, line 96: nit: svg can be renamed to entity_modified_as_override to keep the naming consistent.
Similarly, the icon names can be m_editEntityOverrideIcon and m_addEntityOverrideIcon

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Link/Link.cpp
```diff
@@ -135,6 +135,24 @@ namespace AzToolsFramework
...
+        {
+            PrefabDomConstReference overridePatch = {};
+            auto visitorFn = [&overridePatch](AZ::Dom::Path, const PrefabOverrideMetadata& overrideData)
+            {
+                overridePatch = overrideData.m_patch;
+
+                // We just need to get one override at the provided path.
+                // Return false here so that we don't keep looking for all patches at the path.
+                return false;
```
srikappa-amzn, line 149: This will make sure that the function stops visiting other nodes in the tree if it encounters at least 1 patch. This will work for something like 'AreOverridesPresent' . But for something like 'FindOverridePatches', this should probably do something like GetLinkPatches() where it'll fetch all overrides at a given path.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverrideHandler.cpp
srikappa-amzn, line 31: This will map any override that has the second to last word as 'Entities'. If a component in the future chooses to serialize on of its fields to 'Entities', we will get a false positive. 

Instead of doing it this way, a better way imo would be to do something like:
```
if (get overrides present at path /PathFromFocusedPrefab/Entities/TheCurrentEntityAlias)
{
   if(only 1 override is present)
   {
       if( type is 'add')
       {
          // show the 'plus' icon
       }
       else
       {
           // show 'dot' icon
       }
   }
   else
   {
       // show 'dot' icon
   }
}

```

Feel free to reach out to me if you have any questions on the above approach

srikappa-amzn (changes requested):

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverrideHandler.cpp
michabr: Added comment.

michabr:
> @michabr This looks great, based on the image above! :)
> 
> Question. I noticed there are some random white lines below each row. Is that just the image, or is that actually in the editor?

I'm not sure where those white lines came from. They're not actually in the editor. :)

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Link/Link.cpp
michabr: If there is an "add" or "remove" override present, won't there just be that one patch anyway (other patches get merged into the one)?

michabr: I'll change to GetLinkPatches and later check number of patches anyway to address your other comment...

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverrideHandler.cpp
michabr: Even with only one override present with an "add" or "remove" operation, it still doesn't determine whether we're adding an entity or editing an entity.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Link/Link.cpp
```diff
@@ -135,6 +135,18 @@ namespace AzToolsFramework
             return areOverridesPresent;
         }
 
+        PrefabDomConstReference Link::GetOverridePatch(AZ::Dom::Path path) const
```
srikappa-amzn, line 138: nit: GetOverridePatchAtExactPath might be a clearer name that indicates that the patch returned is from the exact path and not any child paths below

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverrideHandler.cpp
```diff
@@ -8,13 +8,36 @@
...
 
  namespace AzToolsFramework
 {
     namespace Prefab
     {
+        namespace Internal
+        {
+            // Helper to differentiate between adding/removing an entire
+            // entity object or adding/removing a property within an entity
+            bool IsDirectEntityPatch(PrefabDomConstReference patch)
```
srikappa-amzn, line 23: This can be removed now.

srikappa-amzn (approved):

PR #13302, opened 2022-11-21
Added render pipeline selection to Material Editor.

santorac:
## What does this PR do?

The common viewport widget used in Material Editor now includes a combo box for selecting from the available render pipelines. Most of the plumbing is very similar to the controls for lighting preset or model preset.

I had to split up a few functions to support AssetId overloads, so that EntityPreviewViewportScene can track AssetIds. (Using paths would have been too error prone).

image

## How was this PR tested?

In Material Editor, tried switching between all the available pipelines with the MaterialPipelineTest material types, on dx12 and vulkan.

Here is how the existing material types behave (as long as it doesn't crash, we shouldn't care at this time, as these will eventually be ported to the new material pipeline system)...
StandardPBR: Works on Main, LowEnd, and Multiview pipelines. The surface is black on the prototype Deferred pipeline.
EnhancedPBR: Works on Main only. The surface is black on the LowEnd and Deferred. The surface is missing on the Multiveiw pipelines.
Skin: Same as EnhancedPBR.
StandardMultilayerPBR: Same as EnhancedPBR.
Eye: Same as EnhancedPBR.

I discovered a couple issues during testing, which I verified were occurring prior to my changes:
https://github.com/o3de/o3de/issues/13311
https://github.com/o3de/o3de/issues/13312


Gems/Atom/Tools/AtomToolsFramework/Code/Include/AtomToolsFramework/EntityPreviewViewport/EntityPreviewViewportScene.h
```diff
@@ -27,20 +27,35 @@ namespace AtomToolsFramework
...
+
+        RenderPipelineMap::iterator AddRenderPipeline(const AZ::Data::AssetId& pipelineAssetId);
+
         const AZ::Crc32 m_toolId = {};
         AZ::RPI::ScenePtr m_scene;
         AZStd::shared_ptr<AzFramework::Scene> m_frameworkScene;
-        AZ::RPI::RenderPipelinePtr m_renderPipeline;
+
+        RenderPipelineMap m_renderPipelines;
+        RenderPipelineMap::iterator m_activeRenderPipeline;
```
gadams3, line 55: It is probably best to avoid storing iterators as members because they can be invalidated whenever the container is mutated. Instead, you could just copy the pair or store the active asset ID and pipeline pointer in their own members.

gadams3 (commented):
This looks fancy! How does this affect pre-existing materials and material canvas? Looks like we'll need some upgrades. Main comments revolve around storing the iterator for the pipeline descriptor map, which is probably not safe. 

I also posted this PR last night as a follow up to the one where I added the future watchers to enumerate all of the files in the background. One of us will need to resolve or pool in the others changes. https://github.com/o3de/o3de/pull/13300

Gems/Atom/RPI/Code/Include/Atom/RPI.Reflect/System/RenderPipelineDescriptor.h
```diff
@@ -24,6 +24,8 @@ namespace AZ
             AZ_TYPE_INFO(RenderPipelineDescriptor, "{B1A5CF41-AC8D-440E-A1E9-3544D7F3445B}");
             static void Reflect(AZ::ReflectContext* context);
 
+            static constexpr AZStd::string_view Extension = "RenderPipeline.azasset";
```
jeremyong-az, line 27: Personally, I'm not a fan of the coercion from a statically embedded `const char*` string literal to a string view (I believe this is one of several `string_view` design flaws). This usage implies that users of the string view may depend on the presence of a `NUL` terminator, which is an implicit compatibility burden and future footgun (what happens when someone passes in a `string_view` which isn't nul-terminated?). Even though a `char*` isn't necessarily nul-terminated, this is at the very least a well-established convention and communicates an expectation to the users of the type. Prefer the very not fancy `const char*` literal (I'm aware this pattern is abused in other places in the codebase, it's something we should root out over time IMO)

Gems/Atom/Tools/AtomToolsFramework/Code/Source/EntityPreviewViewport/EntityPreviewViewportScene.cpp
```diff
@@ -101,13 +91,96 @@ namespace AtomToolsFramework
         }
     }
 
+    EntityPreviewViewportScene::RenderPipelineMap::iterator EntityPreviewViewportScene::AddRenderPipeline(const AZ::Data::AssetId& pipelineAssetId)
+    {
+        // Load the render pipeline asset
+        AZStd::optional<AZ::RPI::RenderPipelineDescriptor> pipelineDesc =
+            AZ::RPI::GetRenderPipelineDescriptorFromAsset(pipelineAssetId, m_viewportIdSuffix);
+        AZ_Assert(pipelineDesc.has_value(), "Invalid render pipeline descriptor from asset %s", pipelineAssetId.ToFixedString().c_str());
```
jeremyong-az, line 99: I wonder if it's worth trying to recover a bit more gracefully in this case, is it possible that pipelines are edited/moved while this tool is running for example?

Gems/Atom/Tools/AtomToolsFramework/Code/Source/EntityPreviewViewport/EntityPreviewViewportSettingsSystem.cpp
```diff
@@ -272,6 +272,54 @@ namespace AtomToolsFramework
...
     }
 
+    bool EntityPreviewViewportSettingsSystem::LoadRenderPipeline(const AZStd::string& path)
+    {
+        const auto& pathWithAlias = GetPathWithAlias(path);
+        const auto& pathWithoutAlias = GetPathWithoutAlias(path);
+        auto cacheItr = m_renderPipelineDescriptorCache.find(pathWithAlias);
+        if (cacheItr != m_renderPipelineDescriptorCache.end())
+        {
+            SetSettingsValue("/O3DE/AtomToolsFramework/EntityPreviewViewportSettings/RenderPipelinePath", pathWithAlias);
```
jeremyong-az, line 282: Nit: the setreg key is repeated a few times here, might be useful to have as a constant to avoid typos/errors later

jeremyong-az (commented):
Pretty excited for this!

Gems/Atom/Tools/AtomToolsFramework/Code/Include/AtomToolsFramework/EntityPreviewViewport/EntityPreviewViewportScene.h
santorac: In this case it's safe because every time the map is mutated we update m_activeRenderPipeline. But I'll change it anyway to get rid of the "funny smell".

Gems/Atom/RPI/Code/Include/Atom/RPI.Reflect/System/RenderPipelineDescriptor.h
santorac: I just copied and pasted from elsewhere, but I can update this and the original.

santorac: Oh wow, I found 161 hits of this pattern and only 13 of them are in Atom. I'll just update the couple ones I coped from.

Gems/Atom/Tools/AtomToolsFramework/Code/Source/EntityPreviewViewport/EntityPreviewViewportScene.cpp
santorac: Sure I can improve this a bit.

Gems/Atom/Tools/AtomToolsFramework/Code/Source/EntityPreviewViewport/EntityPreviewViewportSettingsSystem.cpp
santorac: There is a lot of duplication throughout this file, I was on the fence about how much I want to improve it right now. @gadams3 what do you think?

Gems/Atom/RPI/Code/Include/Atom/RPI.Reflect/System/RenderPipelineDescriptor.h
gadams3: I have historically not been a fan of using string_view globally for all of the reasons you mentioned. However, it has worked beautifully and optimally in the few cases that I used it recently, comparing file extensions at the end of strings. Otherwise, I have avoided it because it requires additional checks and special handling everywhere it gets used as opposed to const AZStD::string&. I can see the benefits if you want the interface of a string without the allocations or to be able to treat different types of strings uniformly.

Gems/Atom/Tools/AtomToolsFramework/Code/Source/EntityPreviewViewport/EntityPreviewViewportSettingsSystem.cpp
gadams3: You're welcome to do it if you want to. Otherwise I can do it later. This class was originally created specifically for material editor lighting and model presets prior to the document system being generalized. I would like to move lighting presets to their own tool or move a lot of this class to use the document system instead. The viewport settings panel would be replaced by the document inspector. Model presets would go away and we would just reference model assets. The viewport would just listen for asset changes or settings changes. Most of this class would disappear and the rest could live in the settings registry. All of this is in line with several GHI bugs and feature requests.

santorac: I'll just leave it then, since you have some plans to move all this around anyway.

Gems/Atom/RPI/Code/Include/Atom/RPI.Reflect/System/RenderPipelineDescriptor.h
santorac: At some point we might want to look into using AZStd::fixed_string for static consts, to get the best of both worlds.

gadams3 (approved):

gadams3 (approved):

jeremyong-az (approved):

PR #13305, opened 2022-11-21
Prevent user from loading levels with invalid names.

### LINKED ISSUE #6879: It is possible to edit level name so it contains invalid characters
**Describe the bug**
The project's folder name can be changed in the file explorer even with special characters. The level can be loaded and saved afterwards without any restrictions regarding it's name.

**Steps to reproduce**
1. Create a level.
2. Close the Editor.
3. Change the level folder's name in the file explorer.
4. Launch the Editor.
5. Load the level.
6. Navigate to File > Save As and verify that the invalid characters can be used.

**Expected behavior**
There is a prompt about invalid characters during level creation.
A level containing special characters in it's name is not selectable in the Open a level window.

**Actual behavior**
Levels with special characters can be selected, loaded and saved later on, without restrictions regarding it's name.

**Screenshots/Video**

https://user-images.githubusercontent.com/86952009/149342192-dc8f921f-061c-40f1-a061-c0cdd35f0b93.mp4

**Found in Branch**
Development (https://github.com/o3de/o3de/commit/e224fc2ee5ec2a12e75a10acae268b7b38ae3a32)

**Desktop/Device:**
- Device: PC
- OS: Windows
- Version 10
- CPU Intel Core i5-9600k
- GPU Nvidia GeForce RTX 2060
- Memory 16GB

sphrose:

This PR fixes #6879. Formerly, level prefabs containing illegal characters still showed in the level selector, and could be loaded. The new behavior is to hide them in the welcome screen, and disable the open button in the level select dialog.

## How was this PR tested?

Manual testing was performed.


Code/Editor/LevelFileDialog.cpp
```diff
@@ -268,6 +268,12 @@ void CLevelFileDialog::OnTreeSelectionChanged()
     if (!indexes.isEmpty())
     {
         ui->nameLineEdit->setText(NameForIndex(indexes.first()));
+
+        if (m_bOpenDialog)
+        {
+            // Do not allow the user to open a level that haas illegal characters in the name.
```
rgba16f, line 274: speeling error: haas

Code/Editor/LevelFileDialog.cpp
```diff
@@ -432,6 +438,12 @@ bool CLevelFileDialog::ValidateSaveLevelPath(QString& errorMessage) const
         return false;
     }
 
+    if (!ui->nameLineEdit->hasAcceptableInput())
+    {
+        errorMessage = tr("The level name contains illegal characters.");
```
rgba16f, line 443: Can we include the level name in the error message? and potentially what the illegal character was, sometimes it can be hard to tell from the ui.

Code/Editor/WelcomeScreen/WelcomeScreenDialog.cpp
```diff
@@ -162,6 +162,24 @@ bool WelcomeScreenDialog::eventFilter(QObject *watched, QEvent *event)
...
 
+bool WelcomeScreenDialog::IsValidLevelName(const QString& path)
+{
+    QStringList pathParts = Path::SplitIntoSegments(path);
+
+    QString levelName = pathParts.at(pathParts.size() - 1);
+    if (!levelName.endsWith(".prefab", Qt::CaseInsensitive))
+    {
+        return true;
+    }
```
cgalvan, line 173: Is this meant to return `false` if it doesn't end in `.prefab` ?

Also, this will need to account for if prefabs are enable/disabled. We still need to allow `.ly` legacy level names if prefabs are disabled.

cgalvan (changes requested):

Code/Editor/LevelFileDialog.cpp
sphrose: Fixed.

Code/Editor/WelcomeScreen/WelcomeScreenDialog.cpp
sphrose: I was just skipping the check for non prefab levels, but it makes sense to put that in as well, I'll get that done.

Code/Editor/LevelFileDialog.cpp
sphrose: Yes, although at the moment this message never gets displayed, I'll refactor it so it does.

rgba16f (approved):

hultonha:
One question I have about this is would it not be better to stop users from ever saving a level with invalid characters in it? What happens if a user saves their level with an odd name and then they can never load it again? Just checking we're definitely solving the right problem

sphrose:
We are stopping them from doing that, they can't use an illegal name from inside the editor. The issue comes if they rename the level in explorer etc.

hultonha:
> @hultonha We are stopping them from doing that, they can't use an illegal name from inside the editor. The issue comes if they rename the level in explorer etc.

@sphrose Okay excellent thanks 👍

lsemp3d (approved):

Code/Editor/WelcomeScreen/WelcomeScreenDialog.cpp
sphrose: @cgalvan Could you check you're happy with this change when you get a minute please?

cgalvan (approved):

PR #13314, opened 2022-11-21
Fix path width calculation in BreadCrumbs

MiKom:
BreadCrumbs::fillPath used a crude approximation for how much space will the path string take based on font metrics of a plain text string of the path. This patch fixes a bunch of issues with this:

The actual html that was put into the QLabel was very different. It used different spacing, could add icons, etc.

In some situations it could cause a crash because the html inserted into the label could be longer that the available space. This caused a resize that again, caused the re-calculation of the path. This time there was enough space so a path portion was added again. This led to infinite recursion and stack overflow.

Instead, the path label has the horizontal policy set to "Ignore". We ourselves make sure to use the available space in the best way.

The calculation of the path width was also improved. Now it properly takes the separators ("  >  ") and icons into account. Thus, the OptimalPathWidth config option became redundant.

QFontMetrics::width was replaced with the non-deprecated QFontMetrics::horizontalAdvance.

The showing/hiding of the menu button is now handled by ::resizeEvent(). We first calculate if the whole path will fit, and if not, we show it.

Some additional improvements:
 - configuration variables are now static so they have internal linkage, no need for them to be external
 - Rich text setting is explicitly set on the label, without it, there could be a situation where autodetection didn't work
 - any operation that can influence the total width of the path now calls updateGeometry() to make sure that layout is re-calculated

## What does this PR do?

Fixes various problem with the calculation of breadcrumbs path for the purpose of showing only the part that fits. Now the breadcrumbs predictably show only the part of the path that fits:

https://user-images.githubusercontent.com/104767/203140259-c2bf3a11-04bb-429f-a816-b4cdfd87f6ff.mp4

## How was this PR tested?

Manually in all occurences of breadcrumbs in the editor.

### DIFF
--- a/Code/Framework/AzQtComponents/AzQtComponents/Components/Widgets/BreadCrumbs.cpp
+++ b/Code/Framework/AzQtComponents/AzQtComponents/Components/Widgets/BreadCrumbs.cpp
@@ -6,31 +6,40 @@
  *
  */
 
+#include <AzCore/std/algorithm.h>
+#include <AzCore/std/numeric.h>
 #include <AzQtComponents/Components/Widgets/BreadCrumbs.h>
 #include <AzQtComponents/Components/ConfigHelpers.h>
 #include <AzQtComponents/Components/Style.h>
 
 AZ_PUSH_DISABLE_WARNING(4244 4251, "-Wunknown-warning-option") // 4251: 'QLayoutItem::align': class 'QFlags<Qt::AlignmentFlag>' needs to have dll-interface to be used by clients of class 'QLayoutItem'
-#include <QHBoxLayout>
-#include <QLabel>
-#include <QMenu>
-#include <QResizeEvent>
-#include <QSettings>
-#include <QToolBar>
+#include <QtMath>
 #include <QToolButton>
+#include <QToolBar>
+#include <QSettings>
+#include <QResizeEvent>
+#include <QMenu>
+#include <QLabel>
+#include <QHBoxLayout>
 AZ_POP_DISABLE_WARNING
 
 namespace AzQtComponents
 {
-    const QChar g_separator = '/';
-    const QChar g_windowsSeparator = '\\';
-    const QString g_labelName = QStringLiteral("BreadCrumbLabel");
-    const QString g_buttonName = QStringLiteral("MenuButton");
+    static const QChar g_separator = '/';
+    static const QChar g_windowsSeparator = '\\';
+    static const QString g_labelName = QStringLiteral("BreadCrumbLabel");
+    static const QString g_buttonName = QStringLiteral("MenuButton");
+    // Separator is two non-breaking spaces, Right-Pointing Angle Quotation Mark (U+203A) and two more
+    // non-breaking spaces
+    static const QString g_plainTextSeparator = QStringLiteral("\u00a0\u00a0\u203a\u00a0\u00a0");
+    static constexpr int g_iconWidth = 16;
 
     BreadCrumbs::BreadCrumbs(QWidget* parent)
         : QWidget(parent)
         , m_config(defaultConfig())
     {
+        // WARNING: If you add any any new widget to the layout, make sure that it's accounted for in ::sizeHint() and ::fillLabel()
+        
         // create the layout
         QHBoxLayout* boxLayout = new QHBoxLayout(this);
         boxLayout->setContentsMargins(0, 0, 0, 0);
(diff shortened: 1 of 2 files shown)


AMZN-daimini (approved):

Code/Framework/AzQtComponents/AzQtComponents/Components/Widgets/BreadCrumbs.cpp
```diff
@@ -134,6 +150,37 @@ namespace AzQtComponents
...
     }
 
+    QSize BreadCrumbs::sizeHint() const
+    {
+        const QFontMetrics fm(m_label->font());
+        const qreal separatorWidth = fm.horizontalAdvance(g_plainTextSeparator);
+        // Icon adds the icon itself plus two spaces, see generateIconHtml()
+        const qreal iconWidth = g_iconWidth + fm.horizontalAdvance("\u00a0\u00a0");
+
+        // TODO(Qt 5.15.2): replace with:
```
cgalvan, line 160: Does this mean replace once we move to a version >= 5.15.2? Our current binary package is incorrectly labeled as 5.15.2, but it's actually a branch of 5.15.1:

https://github.com/o3de/qt5/tree/5.15.1-o3de

cgalvan (approved):

Code/Framework/AzQtComponents/AzQtComponents/Components/Widgets/BreadCrumbs.cpp
MiKom: Yes. I'll create a separate task for it.

MiKom:
Changed the algorithms from `std::` to `AZStd::`

Code/Framework/AzQtComponents/AzQtComponents/Components/Widgets/BreadCrumbs.cpp
MiKom: https://github.com/o3de/o3de/issues/13336 - task is created. However, it would probably be good to have a tag that will denote todos to be done when the move to the actual 5.15.2 is made.

burelc-amzn: The Qt documentation states that `QStringView::split` was [introduced in Qt 6.0](https://doc.qt.io/qt-6/qstringview.html#split:~:text=be%20dangling%20pointers.-,This%20function%20was%20introduced%20in%20Qt%206.0.,-template%20%3Ctypename%20Container). So I think this can't be done until we move to Qt 6.

Code/Framework/AzQtComponents/AzQtComponents/Components/Widgets/BreadCrumbs.cpp
```diff
@@ -134,6 +152,37 @@ namespace AzQtComponents
...
+        );
+
+        // separators are there only if path has more than one segment
+        const qreal separatorsOnlyWidth = fullPath.size() > 1 ? (fullPath.size() - 1) * separatorWidth : .0;
+
+        const auto numIcons = AZStd::count_if(m_currentPathIcons.cbegin(), m_currentPathIcons.cend(), [this](const QString& iconPath){
+            // assume that icon will be there if there's a path for a given icon or we have a default one
+            return !iconPath.isEmpty() || !m_defaultIcon.isEmpty();
+        });
+
```
burelc-amzn, line 180: Because `m_defaultIcon` doesn't change during the iteration, you can factor it out of the `count_if` call:
```suggestion
        const auto numIcons = !m_defaultIcon.isEmpty() ? m_currentPathIcons.size() : 
            AZStd::count_if(m_currentPathIcons.cbegin(), m_currentPathIcons.cend(), [](const QString& iconPath){
                // assume that icon will be there if there's a path for a given icon or we have a default one
                return !iconPath.isEmpty();
        });
```

Code/Framework/AzQtComponents/AzQtComponents/Components/Widgets/BreadCrumbs.cpp
```diff
@@ -291,31 +349,56 @@ namespace AzQtComponents
...
+        if (m_truncatedPaths.isEmpty())
         {
+            m_label->clear();
+            return;
+        }
+        else
+        {
+            // last section is not clickable
             plainTextPath = m_truncatedPaths.takeLast();
         }
```
burelc-amzn, line 361: Nit: Don't use `else` after `return`
```suggestion
        if (m_truncatedPaths.isEmpty())
        {
            m_label->clear();
            return;
        }

        // last section is not clickable
        plainTextPath = m_truncatedPaths.takeLast();
```

Code/Framework/AzQtComponents/AzQtComponents/Components/Widgets/BreadCrumbs.cpp
MiKom: done

Code/Framework/AzQtComponents/AzQtComponents/Components/Widgets/BreadCrumbs.cpp
MiKom: done

MiKom:
@burelc-amzn Thanks for feedback! Patch updated

MiKom:
@jjjoness Hey. Could you please kick off the build again? This test runs fine on Windows on my machine. Just in case, I also rebased the code on top of the current development branch.

hultonha:
Just started another build @MiKom 👍

jjjoness:
Thanks Tom, missed the message.

PR #13319, opened 2022-11-21
Prefab - Add new unit tests for delete undo nodes

forkercat:

## What does this PR do?

This PR adds 4 new unit tests for delete undo nodes:
1. PrefabUndoDeleteEntity (for source editing)
2. PrefabUndoInstanceLink (for source editing)
3. PrefabUndoDeleteAsOverride

Follow-up for: https://github.com/o3de/o3de/pull/13171

#### Changes:

PrefabUndoDeleteTests
1. PrefabUndoDeleteTests_DeleteEntity
1. PrefabUndoDeleteTests_DeleteNestedInstance
1. PrefabUndoDeleteTests_DeleteEntityAsOverride
1. PrefabUndoDeleteTests_DeleteInstanceAsOverride

PrefabTestFixture
1. Add helper functions for finding entity alias and nested instance alias in owning instance
1. Add helper functions for validating entity/instance existence
1. Move setting registry interface to PrefabTestFixture

_New helper functions should be generic and can be reused in other prefab unit tests, so I put them in PrefabTestFixture._

## How was this PR tested?

Passed all AzToolsFramework unit tests.


Code/Framework/AzToolsFramework/Tests/Prefab/PrefabUndoDeleteTests.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+        AZ::EntityId tireEntityId = CreateEditorEntity(tireEntityName, GetRootContainerEntityId());
+        AZ::EntityId carContainerId = CreateEditorPrefab(carPrefabFilepath, { tireEntityId });
+        
+        EntityAlias tireEntityAlias = FindEntityAliasInInstance(tireEntityName, carContainerId);
+        ASSERT_FALSE(tireEntityAlias.empty());
+
+        AZ::EntityId secondCarContainerId = InstantiateEditorPrefab(carPrefabFilepath, GetRootContainerEntityId());
+
+        auto focusedCarInstance = m_instanceEntityMapperInterface->FindOwningInstance(carContainerId);
```
srikappa-amzn, line 42: If you don't need to explicitly set focus for these tests, then I suggest you call this 'firstCarInstance' since its confusing to know whether the focus is actually set or not.

Code/Framework/AzToolsFramework/Tests/Prefab/PrefabUndoDeleteTests.cpp
```diff
@@ -0,0 +1,222 @@
...
+#include <Prefab/PrefabTestFixture.h>
+
+#include <AzToolsFramework/Prefab/Undo/PrefabUndoDelete.h>
+#include <AzToolsFramework/Prefab/Undo/PrefabUndoDeleteAsOverride.h>
+
+namespace UnitTest
+{
+    using PrefabUndoDeleteTests = PrefabTestFixture;
+
+    TEST_F(PrefabUndoDeleteTests, PrefabUndoDeleteTests_DeleteEntity)
```
srikappa-amzn, line 18: For the sake of covering all bases, maybe we should also add a 'PrefabUndoDeleteTests_DeleteNestedInstance' test case?

srikappa-amzn (commented):
Great job on the tests and the new helper functions! They are easy to read and follow : )

One thing to be cautious about is the size of PrefabTestFixture class. It can easily grow in size if we keep adding a lot of helper functions required for different tests. I suggest considering putting these helper functions in a child test fixture unless you are sure that future tests also will need them.

forkercat:
> One thing to be cautious about is the size of PrefabTestFixture class. It can easily grow in size if we keep adding a lot of helper functions required for different tests. I suggest considering putting these helper functions in a child test fixture unless you are sure that future tests also will need them.

I will re-evaluate if it is needed to have the Validate functions in PrefabTestFixture.

Code/Framework/AzToolsFramework/Tests/Prefab/PrefabUndoDeleteTests.cpp
forkercat: There is this inconsistency in the current workflow that DeleteNestedInstance depends on PrefabUndoInstanceLink to remove the link rather than on new undo nodes.

But I think you're right and it makes sense to have another undo test here for that call. And I will add some comment to describe the inconsistency for reference.

Code/Framework/AzToolsFramework/Tests/Prefab/PrefabUndoDeleteTests.cpp
```diff
@@ -0,0 +1,307 @@
...
+        // Create an undo node
+        PrefabUndoDeleteEntity undoDeleteNode("Undo Deleting Entity");
+
+        EntityOptionalReference parentEntityToUpdate = firstCarInstance->get().GetContainerEntity();
+        ASSERT_TRUE(parentEntityToUpdate.has_value());
+
+        AZ::EntityId firstTireEntityId = firstCarInstance->get().GetEntityId(tireEntityAlias);
+        ASSERT_TRUE(firstTireEntityId.IsValid());
+
+        AZStd::string firstEngineEntityAliasPath = m_instanceToTemplateInterface->GenerateEntityAliasPath(firstTireEntityId);
```
michabr, line 62: Should this be firstTireEntityAliasPath?

michabr (commented):
I'm curious, where does the current focused prefab get set in these tests?

forkercat:
> I'm curious, where does the current focused prefab get set in these tests?

For these undo tests, we don't set the focused instance, so the actual focused instance is always the level. It is okay because it currently does not use any Get*FromFocused functions.


In undo nodes, I think it is a good design to pass in focused instance without the dependency on the logic to get it from the focus handler.

michabr (approved):

PR #13320, opened 2022-11-21
[Terrain] Fix paint/eyedropper with auto scale setting

mbalfour-amzn:
## What does this PR do?

When the Image Gradient has the "Auto Scale" setting enabled, the values in the image are supposed to auto-scale so that the minimum value scales down to 0.0 and the maximum value scales to 1.0. However, the paintbrush wasn't refreshing the scaling values, so they would continue to use whatever scale values the image was loaded at. If you saved and reloaded the image, it would then suddenly get a new scale range.

With these changes, the scale is refreshed while painting. The "Auto Scale" setting really _shouldn't_ be used with painting, since it's pretty jarring, but at least it now works correctly and consistently.

Also, the eyedropper tool was reading from scaled and smoothed image values, instead of the raw values, so it was picking up incorrect values from the image. This has been fixed so that it reads the raw values.

## How was this PR tested?

Tried painting and using the eyedropper with Auto-scale enabled.

With auto-scale on, the eyedropper still gets the correct value ( < 100%) and painting higher and lower values causes everything to scale immediately.

https://user-images.githubusercontent.com/82224783/203165354-8ab36020-d3a2-4560-81ca-940576928f31.mp4


cgalvan (approved):

Gems/GradientSignal/Code/Source/Components/ImageGradientComponent.cpp
```diff
@@ -1072,57 +1053,68 @@ namespace GradientSignal
         LmbrCentral::DependencyNotificationBus::Event(GetEntityId(), &LmbrCentral::DependencyNotificationBus::Events::OnCompositionChanged);
     }
 
-    void ImageGradientComponent::SetPixelValueByPosition(const AZ::Vector3& position, float value)
+    PixelIndex ImageGradientComponent::GetPixelIndexForPositionInternal(const AZ::Vector3& position) const
     {
-        SetPixelValuesByPosition(AZStd::span<const AZ::Vector3>(&position, 1), AZStd::span<float>(&value, 1));
-    }
+        const auto& width = m_imageDescriptor.m_size.m_width;
+        const auto& height = m_imageDescriptor.m_size.m_height;
```
hultonha, line 1059: uber nit: Do these need to be references when they types are just built-in?

Gems/GradientSignal/Code/Include/GradientSignal/Components/ImageGradientComponent.h
```diff
@@ -186,6 +186,7 @@ namespace GradientSignal
         void GetValuesInternal(SamplingType samplingType, AZStd::span<const AZ::Vector3> positions, AZStd::span<float> outValues) const;
         float GetValueFromImageData(SamplingType samplingType, const AZ::Vector3& uvw, float defaultValue) const;
         float GetPixelValue(AZ::u32 x, AZ::u32 y) const;
+        float InvertHeightAndGetPixelValue(AZ::u32 x, AZ::u32 y) const;
```
hultonha, line 189: suggestion: An API comment for this call might be useful to make it clear what it's doing - it might be clearer to just say `GetInvertedPixelValue` perhaps? Or maybe there's more to it than that?

hultonha (approved):
LGTM just one small suggestion and a nit

Gems/GradientSignal/Code/Include/GradientSignal/Components/ImageGradientComponent.h
mbalfour-amzn: The method is private to the component, but I'll add an API comment. I can't use "GetInvertedPixelValue" because that makes it sound like the pixel value is getting inverted, not the Y component for lookup. 

The backstory here is that there's a tiny chunk of code in GetPixelValue() that I wanted to avoid duplicating multiple times in this file. But half the places that use it need to invert the Y value for lookup and half don't because they've already got an inverted Y. So I split this into two methods, one that inverts the Y first and calls GetPixelValue(), and then just GetPixelValue() itself.

I also considered "InvertYAndGetPixelValue" but that just hurts my eyes a little bit, I find it harder to read that "Y" quickly next to the capital A. I'd be willing to switch to this though if you think it's more clear.

Gems/GradientSignal/Code/Source/Components/ImageGradientComponent.cpp
mbalfour-amzn: Changed.

Gems/GradientSignal/Code/Include/GradientSignal/Components/ImageGradientComponent.h
mbalfour-amzn: Changed name and added API comments.

hultonha: Thanks for the description and adding the comment @mbalfour-amzn, much appreciated!

PR #13331, opened 2022-11-22
Action Manager | Migrate Spline and Tube Shape Component Modes

AMZN-daimini:
Reimplementation of Component Mode custom actions via the new Action Manager.

Right now it only extends to Spline and Tube Shape components, other existing modes will be a follow up (to avoid blowing up the scope of this PR even more)

Some of the changes
- A couple modifications to SerializeContext to allow classes with non-default constructors to be registered
- Registering all ComponentMode classes to RTTI and SerializeContext so that it's possible to loop through them programmatically
- Adding an API to retrieve the ComponentMode Type Uuid from the Component Type
- Testing a setup to register the Component Mode actions at Editor startup and have them dynamically show up correctly when Component Mode is changed.

Known issue:
- Shortcut for Component Mode Actions don't seem to work (will file a bug for that)

ComponentModeActionManager
(Note Resize Radii doesn't work in this gif but is fixed in the current PR)


Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelection.h
```diff
@@ -338,6 +353,11 @@ namespace AzToolsFramework
         virtual void ShowVertexDeletionWarning();
 
     private:
+        // EditorVertexSelectionRequestBus overrides ...
+        virtual void DuplicateSelectedVertices() override;
```
jonawals, line 357: Nit: you can remove the `virtual` qualifiers from these inherited methods.

Gems/LmbrCentral/Code/Source/Shape/EditorTubeShapeComponentMode.cpp
```diff
@@ -52,6 +58,85 @@ namespace LmbrCentral
...
+            actionProperties.m_description = s_resetRadiiDesc;
+            actionProperties.m_category = "Tube Shape";
+
+            actionManagerInterface->RegisterAction(
+                EditorMainWindowActionContextIdentifier,
+                actionIdentifier,
+                actionProperties,
+                []
+                {
+                    /*
```
jonawals, line 91: Nit: I presume this dead code will be removed but just in case... ;p

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelection.cpp
```diff
@@ -57,13 +60,105 @@ namespace std
...
+    }
+
+    void EditorVertexSelection::BindEditorVertexSelectionActionsToMenus()
+    {
+        auto menuManagerInterface = AZ::Interface<AzToolsFramework::MenuManagerInterface>::Get();
+        AZ_Assert(
+            menuManagerInterface,
+            "EditorVertexSelection - could not get MenuManagerInterface on BindActionsToMenus.");
+
+        menuManagerInterface->AddActionToMenu(EditMenuIdentifier, "o3de.action.vertexSelection.duplicate", 6000);
```
hultonha, line 157: suggestion: Might be good to move these (`"o3de.action.vertexSelection.duplicate"`) to constants at the top of the file (so they aren't repeated here and in `RegisterEditorVertexSelectionActions`)

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelection.cpp
```diff
@@ -845,6 +940,24 @@ namespace AzToolsFramework
...
+    void EditorVertexSelectionVariable<Vertex>::DeleteSelectedVertices()
+    {
+        DestroySelected();
+    }
+
+    template<typename Vertex>
+    void EditorVertexSelectionVariable<Vertex>::ClearVertexSelection()
+    {
+        EditorVertexSelectionBase<Vertex>::ClearSelected();
+    }
```
hultonha, line 959: question: Just curious why we need to add these aliases? 🤔 Is it purely for readability?

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelection.cpp
```diff
@@ -1098,6 +1229,18 @@ namespace AzToolsFramework
...
+    template<typename Vertex>
+    EditorVertexSelectionVariable<Vertex>::EditorVertexSelectionVariable()
+    {
+        EditorVertexSelectionRequestBus::Handler::BusConnect();
+    }
+
+    template<typename Vertex>
+    EditorVertexSelectionVariable<Vertex>::~EditorVertexSelectionVariable()
+    {
+        EditorVertexSelectionRequestBus::Handler::BusDisconnect();
```
hultonha, line 1241: question: This must have been being connected before? 🤔 Were these moved?

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelection.h
```diff
@@ -20,6 +21,14 @@
 
 namespace AzToolsFramework
 {
+    //! Registers the Actions provided by the EditorVertexSelection while it is active to the Action Manager.
+    //! e.g. Vertex deletion, duplication etc.
+    struct EditorVertexSelection
```
hultonha, line 26: suggestion: It might be better to call this something Action related, so maybe `EditorVertexSelectionActions`?

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelection.h
```diff
@@ -291,16 +300,20 @@ namespace AzToolsFramework
...
 
-        EditorVertexSelectionFixed() = default;
-        EditorVertexSelectionFixed(EditorVertexSelectionFixed&&) = default;
-        EditorVertexSelectionFixed& operator=(EditorVertexSelectionFixed&&) = default;
+        EditorVertexSelectionFixed();
+        ~EditorVertexSelectionFixed();
 
     private:
+        // EditorVertexSelectionRequestBus overrides ...
+        virtual void ClearVertexSelection() override;
```
hultonha, line 315: update: We shouldn't add these here - `EditorVertexSelectionFixed` explicitly does not implement this bus as you cannot change the number of vertices that are stored. This should only be implemented in `EditorVertexSelectionVariable`

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelectionBus.h
```diff
@@ -0,0 +1,43 @@
...
+    {
+    public:
+        virtual void DuplicateSelectedVertices() {};
+        virtual void DeleteSelectedVertices() {};
+        virtual void ClearVertexSelection() = 0;
+
+    protected:
+        ~EditorVertexSelectionRequests() = default;
+    };
+    using EditorVertexSelectionRequestBus = AZ::EBus<EditorVertexSelectionRequests>;
```
hultonha, line 26: update: I think we should be able to use the already existing buses for this - please see `VertexContainerInterface` for the existing buses (`FixedVerticesRequestBus` and `VariableVerticesRequestBus`). This new type should be removed.

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelectionBus.h
```diff
@@ -0,0 +1,43 @@
...
+
+    //! Bus to be notified about changes in Editor Vertex Selection state.
+    class EditorVertexSelectionNotifications : public AZ::EBusTraits
```
hultonha, line 29: question: I'm not 100% sure this is needed, it would also be useful to pass the new vertex index here which might be useful if we do need it... I'm just wondering why this is needed now and not before?

Code/Framework/AzToolsFramework/AzToolsFramework/ViewportSelection/EditorTransformComponentSelection.cpp
```diff
@@ -3037,6 +3038,9 @@ namespace AzToolsFramework
             // Update when the transform mode changes.
             m_actionManagerInterface->AddActionToUpdater(TransformModeChangedUpdaterIdentifier, actionIdentifier);
 
+            // This action is only accessible outside of Component Modes
+            m_actionManagerInterface->AssignModeToAction(DefaultActionContextModeIdentifier, actionIdentifier);
```
hultonha, line 3042: suggestion: I wonder if wrapping this in a function with a friendlier name might make this easier to use and understand, e.g....

```
// This action is only accessible outside of Component Modes
void MakeActionUnavilableInComponentMode(
    ActionManagerInterface* actionManagerInterface, const ActionIdentifier actionIdentifier)
{
    actionManagerInterface->AssignModeToAction(DefaultActionContextModeIdentifier, actionIdentifier);
}
```

Code/Framework/AzToolsFramework/Tests/ComponentModeTestDoubles.cpp
```diff
@@ -244,5 +254,10 @@ namespace AzToolsFramework
             return "OverrideMouseInteraction Edit Mode";
         }
 
+        AZ::Uuid OverrideMouseInteractionComponentMode::GetComponentModeType() const
+        {
+            return azrtti_typeid<OverrideMouseInteractionComponentMode>();
+        }
```
hultonha, line 260: update: As this is a breaking change, we should also make a note to include a deprecation notice (or provide a default implementation) so that clients who might have implemented their own Component Modes know how to upgrade.

Gems/LmbrCentral/Code/Source/Shape/EditorSplineComponentMode.cpp
```diff
@@ -35,6 +36,40 @@ namespace LmbrCentral
         m_vertexSelection.Destroy();
     }
 
+    void EditorSplineComponentMode::Reflect(AZ::ReflectContext* context)
+    {
+        if (auto serializeContext = azrtti_cast<AZ::SerializeContext*>(context))
+        {
+            serializeContext->Class<EditorSplineComponentMode, EditorBaseComponentMode>(AZ::Internal::NullFactory::GetInstance());
```
hultonha, line 43: update: Not sure what `AZ::Internal::NullFactory::GetInstance()` is for but it would be hard for a user to know to use this, maybe we should make a templated helper to raise the abstraction level of this?

Gems/LmbrCentral/Code/Source/Shape/EditorSplineComponentMode.cpp
```diff
@@ -35,6 +36,40 @@ namespace LmbrCentral
...
+        auto actionManagerInterface = AZ::Interface<AzToolsFramework::ActionManagerInterface>::Get();
+        AZ_Assert(actionManagerInterface, "EditorVertexSelection - could not get ActionManagerInterface on RegisterActions.");
+
+        AZ::SerializeContext* serializeContext = nullptr;
+        AZ::ComponentApplicationBus::BroadcastResult(serializeContext, &AZ::ComponentApplicationRequests::GetSerializeContext);
+        
+        AZStd::string modeIdentifier = AZStd::string::format(
+            "o3de.context.mode.%s", serializeContext->FindClassData(azrtti_typeid<EditorSplineComponentMode>())->m_name);
+
+        actionManagerInterface->AssignModeToAction(modeIdentifier, "o3de.action.vertexSelection.duplicate");
```
hultonha, line 63: suggestion: The identifiers should likely be moved to constants as they crop up in a few places

Gems/LmbrCentral/Code/Source/Shape/EditorTubeShapeComponentMode.cpp
```diff
@@ -52,6 +58,85 @@ namespace LmbrCentral
...
+
+    void EditorTubeShapeComponentMode::RegisterActions()
+    {
+        auto actionManagerInterface = AZ::Interface<AzToolsFramework::ActionManagerInterface>::Get();
+        AZ_Assert(actionManagerInterface, "EditorTubeShapeComponentMode - could not get ActionManagerInterface on RegisterActions.");
+
+        auto hotKeyManagerInterface = AZ::Interface<AzToolsFramework::HotKeyManagerInterface>::Get();
+        AZ_Assert(hotKeyManagerInterface, "EditorTubeShapeComponentMode - could not get HotKeyManagerInterface on RegisterActions.");
+
+        // Duplicate Vertices
```
hultonha, line 77: update: This comment looks incorrect

hultonha (commented):
This looks like a great start, I have a few comments/questions/suggestions but this does 100% seem to be heading in the right direction. Great work @AMZN-daimini 👍

Code/Framework/AzCore/AzCore/Serialization/SerializeContext.h
```diff
@@ -1514,12 +1514,15 @@ namespace AZ
                 }
                 case AZStd::any::Action::Construct:
                 {
-                    // Default construct the ValueType object
-                    // This occurs in the case where a Copy and Move action is invoked
-                    void* ptr = AZStd::any_cast<void>(dest);
-                    if (ptr)
+                    if constexpr (AZStd::is_default_constructible_v<ValueType>)
```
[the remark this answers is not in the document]
AMZN-daimini: This is actually a result of a chat we had last week so he's aware :)

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelection.cpp
AMZN-daimini: I created the bus in this PR so it did not exist before.

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelectionBus.h
AMZN-daimini: Oh this is a stub, I wanted to disable the vertex actions if no vertex was selected... but I noticed it's not a current feature so I stopped early. I will remove it from this PR.

Code/Framework/AzCore/AzCore/Serialization/SerializeContext.h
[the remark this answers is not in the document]
lumberyard-employee-dm: I am aware.

Code/Framework/AzToolsFramework/AzToolsFramework/ViewportSelection/EditorTransformComponentSelection.cpp
AMZN-daimini: I will mark it down and do a pass (since many actions had this added in previous PRs)

Gems/LmbrCentral/Code/Source/Shape/EditorTubeShapeComponentMode.cpp
AMZN-daimini: Yup I need to replace it with a bus call, incoming.

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelectionBus.h
AMZN-daimini: I have checked those buses and I'm not sure how they can be used without an intermediary. May need some assistance here.

hultonha: Sure thing, give me a ping later and we can take a look 👍

AMZN-daimini:
@hultonha I think I have resolved everything but the bus discussion. Let me know if we can arrange a call on Friday to discuss this :)

hultonha:
> @hultonha I think I have resolved everything but the bus discussion. Let me know if we can arrange a call on Friday to discuss this :)

Sounds good thanks @AMZN-daimini, let's catch-up on Friday 👍

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/EditorBaseComponentMode.cpp
```diff
@@ -103,6 +111,15 @@ namespace AzToolsFramework
             return PopulateActionsImpl();
         }
 
+        AZ::Uuid EditorBaseComponentMode::GetComponentModeType() const
+        {
+            AZ_Assert(
+                false,
+                "Classes derived from EditorBaseComponentMode need to override this function and return their typeid."
+                "Example: \"return azrtti_typeid<DerivedComponentMode>();\"");
+            return AZ::Uuid();
```
hultonha, line 120: update: I think there's a `CreateNull()` function you can call here instead which I think should be preferred vs a default constructed one

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/EditorBaseComponentMode.h
```diff
@@ -41,6 +44,21 @@ namespace AzToolsFramework
...
+
+            //! Bind actions to appear in this component mode.
+            //! This is provided as guidance but it will not be called automatically;
+            //! you will need a system component to call it in the OnActionContextModeBindingHook handler.
+            static void BindActionsToModes() {}
+
+            //! Bind actions to appear in menus.
+            //! This is provided as guidance but it will not be called automatically;
+            //! you will need a system component to call it in the OnMenuBindingHook handler.
+            static void BindActionsToMenus() {}
```
hultonha, line 60: suggestion: This is the type of thing we should definitely make a note to document in the API for Component Mode docs when they get ported to O3DE

Gems/LmbrCentral/Code/Source/Shape/EditorTubeShapeComponentMode.cpp
AMZN-daimini: Fixed

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeCollection.cpp
```diff
@@ -210,6 +212,15 @@ namespace AzToolsFramework
             return InComponentMode() ? m_activeComponentTypes : AZStd::vector<AZ::Uuid>{};
         }
 
+        void ComponentModeCollection::EnumerateActiveComponents(
+            AZStd::function<void(const AZ::EntityComponentIdPair&, const AZ::Uuid&)> handler) const
+        {
+            for (size_t i = 0; i < m_activeComponentModes.size(); ++i)
+            {
+                handler(m_activeComponentModes[i].first, m_activeComponentModes[i].second);
+            }
```
hultonha, line 221: question: Is there any reason why we can't use a range based for loop here?

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelection.h
```diff
@@ -291,16 +300,20 @@ namespace AzToolsFramework
     //! EditorVertexSelectionFixed provides selection and editing for a fixed length number of
     //! vertices. New vertices cannot be inserted/added or removed.
     template<typename Vertex>
-    class EditorVertexSelectionFixed : public EditorVertexSelectionBase<Vertex>
+    class EditorVertexSelectionFixed
+        : public EditorVertexSelectionBase<Vertex>
+        , private AzToolsFramework::EditorVertexSelectionRequestBus::Handler
```
hultonha, line 305: update: Here we should just implement `EditorVertexSelectionFixedRequestBus`

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelection.h
```diff
@@ -291,16 +300,20 @@ namespace AzToolsFramework
...
     {
     public:
         AZ_CLASS_ALLOCATOR_DECL
 
-        EditorVertexSelectionFixed() = default;
-        EditorVertexSelectionFixed(EditorVertexSelectionFixed&&) = default;
-        EditorVertexSelectionFixed& operator=(EditorVertexSelectionFixed&&) = default;
```
hultonha, line 301: question: Why were these special members removed? I believe it's safe to add them back (by providing a user declared destructor these move operations are implicitly disabled which is why you need to explicitly add them as was done before).

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelection.h
```diff
@@ -291,16 +300,20 @@ namespace AzToolsFramework
...
+        EditorVertexSelectionFixed(const AZ::EntityComponentIdPair& entityComponentIdPair);
```
hultonha, line 310: update: Should add `explicit` here for single argument constructor

Gems/LmbrCentral/Code/Source/Shape/EditorTubeShapeComponentMode.cpp
```diff
@@ -268,6 +350,27 @@ namespace LmbrCentral
         ContainerChanged();
     }
 
+    void EditorTubeShapeComponentMode::ResetRadii()
```
hultonha, line 353: question: I take it the code in `PopulateActionsImpl` in this file will be removed when the feature flag gets removed and all the new code is turned on by default? (might be worth maybe moving this to a function to avoid the duplication in both places but it's not a huge deal).

hultonha (commented):
Changes are looking really solid. I still have some suggestions about the Editor Vertex Selection fixed/variable setup (ping me if you want to discuss) and some smaller nits but overall really strong and very nearly there! Excellent work getting this resolved 👍

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeCollection.cpp
AMZN-daimini: No, I had a different implementation before but I ended up refactoring it to support multi-entity editing. Will fix.

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeCollection.h
```diff
@@ -107,6 +108,7 @@ namespace AzToolsFramework
             bool ActiveComponentModeChanged(const AZ::Uuid& previousComponentType);
 
             AZStd::vector<AZ::Uuid> m_activeComponentTypes; ///< What types of ComponentMode are currently active.
+            AZStd::vector<AZStd::pair<AZ::EntityComponentIdPair, AZ::Uuid>> m_activeComponentModes; ///< What Entity Components have a ComponentMode currently active.
```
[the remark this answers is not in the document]
AMZN-daimini: That was the original plan, but I used it to replace `m_activeComponentTypes` and was going down a weird rabbit hole lol. Will refactor!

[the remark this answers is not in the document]
AMZN-daimini: Yes and no. `m_activeComponentTypes` will only store the types of the component modes, `m_activeComponentModes` keeps track of all components with active component modes. In the case of multi-entity editing, the latter will have more entries since more component modes are created for a single type.
I originally was going for just expanding `m_activeComponentTypes` but it was bloating up the scope significantly, so yes we could refactori this setup but I would probably want a better view of the whole architecture before touching it :)

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelection.cpp
```diff
@@ -57,13 +61,131 @@ namespace std
...
+                EditorMainWindowActionContextIdentifier,
+                actionIdentifier,
+                actionProperties,
+                []
+                {
+                    auto componentModeCollectionInterface = AZ::Interface<ComponentModeCollectionInterface>::Get();
+                    AZ_Assert(componentModeCollectionInterface, "Could not retrieve component mode collection.");
+
+                    componentModeCollectionInterface->EnumerateActiveComponents(
+                        [](const AZ::EntityComponentIdPair& entityComponentIdPair, const AZ::Uuid&)
```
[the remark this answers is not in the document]
AMZN-daimini: So, technically this function should check whether the active component is of the correct type before blasting off the bus call. We skip it here since that bus simply won't be handled for an `EntityComponentIdPair` pointing to a component of the wrong type.

You're correct that we currently don't use it, but I'm expecting it to be useful in many cases. We can remove it and re-introduce it down the line if it actually becomes needed.

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelection.cpp
```diff
@@ -57,13 +61,131 @@ namespace std
...
+                        {
+                            EditorVertexSelectionRequestBus::Event(
+                                entityComponentIdPair, &EditorVertexSelectionRequests::DeleteSelectedVertices);
+                        }
+                    );
+                }
+            );
+
+            hotKeyManagerInterface->SetActionHotKey(actionIdentifier, "Delete");
```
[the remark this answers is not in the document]
AMZN-daimini: This is the hotkey definition, so `Delete` here is the keyboard button

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelection.cpp
```diff
@@ -57,13 +61,131 @@ namespace std
...
+                        [](const AZ::EntityComponentIdPair& entityComponentIdPair, const AZ::Uuid&)
+                        {
+                            EditorVertexSelectionRequestBus::Event(
+                                entityComponentIdPair, &EditorVertexSelectionRequests::ClearVertexSelection);
+                        }
+                    );
+                }
+            );
+
+            hotKeyManagerInterface->SetActionHotKey(actionIdentifier, "Esc");
```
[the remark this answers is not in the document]
AMZN-daimini: Qt has implicit conversion from string to `QKeySequence`

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelection.cpp
```diff
@@ -1061,6 +1201,24 @@ namespace AzToolsFramework
         }
     }
 
+    template<typename Vertex>
+    EditorVertexSelectionFixed<Vertex>::EditorVertexSelectionFixed(const AZ::EntityComponentIdPair& entityComponentIdPair)
+    {
+        EditorVertexSelectionRequestBus::Handler::BusConnect(entityComponentIdPair);
+    }
```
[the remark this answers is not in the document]
AMZN-daimini: This clears the **selection**, not the vertices. The function was already implemented.
Ping me if I'm missing something though :D

Gems/LmbrCentral/Code/Source/Shape/EditorTubeShapeComponentMode.cpp
AMZN-daimini: Correct. I will try to remove the duplication if it's not too complex.

AMZN-daimini: Removed duplicate code

Code/Framework/AzToolsFramework/AzToolsFramework/API/ComponentModeCollectionInterface.h
```diff
@@ -22,11 +22,12 @@ namespace AzToolsFramework
...
 
-        //! Retrieves the list of all Component types (usually one).
+        //! Retrieves the list of all component types (usually one).
         //! @note If called outside of component mode, an empty vector will be returned.
         virtual AZStd::vector<AZ::Uuid> GetComponentTypes() const = 0;
 
-        //! Calls the handler function for each ComponentMode that is currently active.
-        virtual void EnumerateActiveComponents(AZStd::function<void(const AZ::EntityComponentIdPair&, const AZ::Uuid&)> handler) const = 0;
+        //! Calls the handler function for each component mode that is currently active.
+        using ActiveComponentModeCB = AZStd::function<void(const AZ::EntityComponentIdPair&, const AZ::Uuid&)>;
```
hultonha, line 30: standards/nit: This should ideally be `Cb` not `CB` to be consistent with the coding standards (see https://github.com/o3de/sig-core/blob/main/governance/Coding-Standards-and-Style-Guide.md#abbreviations)

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeCollection.h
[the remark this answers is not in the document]
hultonha: Okay awesome thanks!

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeCollection.cpp
hultonha: Okay nice thanks

Code/Framework/AzToolsFramework/AzToolsFramework/API/ComponentModeCollectionInterface.h
```diff
@@ -25,5 +25,8 @@ namespace AzToolsFramework
         //! Retrieves the list of all Component types (usually one).
         //! @note If called outside of component mode, an empty vector will be returned.
         virtual AZStd::vector<AZ::Uuid> GetComponentTypes() const = 0;
+
+        //! Calls the handler function for each ComponentMode that is currently active.
+        virtual void EnumerateActiveComponents(AZStd::function<void(const AZ::EntityComponentIdPair&, const AZ::Uuid&)> handler) const = 0;
```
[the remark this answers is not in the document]
hultonha: I mentioned about CB -> Cb but if you want to leave as-is that's okay

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeCollection.h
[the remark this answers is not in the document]
hultonha: Okay understood thanks! Follow-up makes sense! 😅

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelection.cpp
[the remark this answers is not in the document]
hultonha: Okay understood, better to have it and not need it than need it and not have it 😝

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeCollection.cpp
```diff
@@ -140,7 +140,7 @@ namespace AzToolsFramework
                 m_viewportUiHandlers.emplace_back(componentType);
             }
 
-            m_activeComponentModes.push_back(AZStd::make_pair(entityComponentIdPair, componentType));
+            m_activeComponentModes.emplace_back(ComponentModeItem{ entityComponentIdPair, componentType });
```
hultonha, line 143: update: You should be able to omit `ComponentModeItem` and just do...

```
m_activeComponentModes.emplace_back(entityComponentIdPair, componentType);
```

The arguments will then be forwarded to the constructor (to have this work you might need to add a constructor for `ComponentModeItem` but it's likely worth it

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelection.cpp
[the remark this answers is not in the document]
hultonha: Cool got you, thanks for clarifying

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelection.cpp
```diff
@@ -1262,6 +1432,12 @@ namespace AzToolsFramework
     {
         AZ_PROFILE_FUNCTION(AzToolsFramework);
 
+        // if no vertex is selected, bail
```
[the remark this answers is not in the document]
hultonha: Thanks!

Gems/LmbrCentral/Code/Source/Shape/EditorTubeShapeComponentMode.cpp
hultonha: Awesome thanks! 🙂

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeCollection.h
```diff
@@ -106,7 +107,14 @@ namespace AzToolsFramework
             // Internal helper used by Select[|Prev|Next]ActiveComponentMode
             bool ActiveComponentModeChanged(const AZ::Uuid& previousComponentType);
 
+            struct ComponentModeItem
```
hultonha, line 110: suggestion: For `emplace_back` to work as I mentioned having a constructor for this with the two arguments as parameters would be great (plus a defaulted constructor too)

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelection.cpp
```diff
@@ -57,13 +61,131 @@ namespace std
...
+                []
+                {
+                    auto componentModeCollectionInterface = AZ::Interface<ComponentModeCollectionInterface>::Get();
+                    AZ_Assert(componentModeCollectionInterface, "Could not retrieve component mode collection.");
+
+                    componentModeCollectionInterface->EnumerateActiveComponents(
+                        [](const AZ::EntityComponentIdPair& entityComponentIdPair, const AZ::Uuid&)
+                        {
+                            EditorVertexSelectionVariableRequestBus::Event(
+                                entityComponentIdPair, &EditorVertexSelectionVariableRequests::ClearVertexSelection);
```
hultonha, line 167: it still would be nice if this was in 'Fixed' and we implemented it here but as it's unused at the moment this is okay (we used to have components that used it that were removed but we should try and maintain the functionality)

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelection.h
```diff
@@ -313,18 +323,24 @@ namespace AzToolsFramework
...
-    class EditorVertexSelectionVariable : public EditorVertexSelectionBase<Vertex>
+    class EditorVertexSelectionVariable
+        : public EditorVertexSelectionBase<Vertex>
+        , private AzToolsFramework::EditorVertexSelectionVariableRequestBus::Handler
     {
     public:
         AZ_CLASS_ALLOCATOR_DECL
 
-        EditorVertexSelectionVariable() = default;
-        EditorVertexSelectionVariable(EditorVertexSelectionVariable&&) = default;
```
hultonha, line 322: update: Please can we add these back

hultonha (approved):
LGTM! Let's get this in 🙂 Awesome work @AMZN-daimini

Code/Framework/AzToolsFramework/AzToolsFramework/API/ComponentModeCollectionInterface.h
AMZN-daimini: The rest of the codebase uses `CB` so I'd like to conform to that right now. I see your point though :)

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelection.h
AMZN-daimini: Removing these actually broke tests, so I had to restore them. Thanks for pointing it out :D

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelection.cpp
```diff
@@ -1098,6 +1238,21 @@ namespace AzToolsFramework
...
+        EditorVertexSelectionVariableRequestBus::Handler::BusConnect(entityComponentIdPair);
+    }
+
+    template<typename Vertex>
+    EditorVertexSelectionVariable<Vertex>::~EditorVertexSelectionVariable()
+    {
+        if (EditorVertexSelectionVariableRequestBus::Handler::BusIsConnected())
+        {
+            EditorVertexSelectionVariableRequestBus::Handler::BusDisconnect();
+        }
```
cgalvan, line 1253: Nit: The `BusDisconnect` method already checks if its connected first, so I think you can skip that check here:

```
if (BusIsConnected())
{
    BusType::DisconnectInternal(*context, m_node);
}
```

cgalvan (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelection.cpp
AMZN-daimini: Got it, I will honestly move on with it if it passes AR, but will keep this in mind for a follow-up!

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelection.h
hultonha: Nice one thanks!

PR #13333, opened 2022-11-22
Fleshed out the material pipeline design.

santorac:
## What does this PR do?

- Rearrange the overall design to be more in line with what @antonmic was envisioning with his material shader refactor. The vertex structs and "Evaluate" functions are mostly defined in the material code now, although we are keeping the lighting code defined in the material pipeline.
- Added as much of the core material type functionality as possible from StandardPBR, EnhancedPBR, and BasePBR into the test material types. This mostly just meant copying and pasting the property layout from the core material types, and copying #include statements from their main azsli files.
- Updated the MaterialTypeBuilder to simplify the way it generates the intermediate azsl file, by using a MATERIAL_TYPE_AZSLI_FILE_PATH macro to include the material type shader code into the middle of the template azsli file. 
- Added DepthPass_VertexData and DepthPass_VertexEval files, similar to BasePBR_VertexData and BasePBR_VertexEval.
- Reduced error message spam in MaterialTypeSourceData.
-  Moved material property name checks around to clarify some error reporting.
- Removed a scarcely used and unnecessary DefaultObjectSrg.azsli.
- Added an example of a material that is the very minimum to run a material pipeline.

## How was this PR tested?

In AtomSampleViewer there are materials for each of the "MaterialPipelineTest" material types. I opened each of these in Material Editor and switched to each render pipeline, checking the results looked correct.
AtomSampleViewer full test suite showed no new screenshot failures on vulkan.


antonmic (approved):

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/LowEndPipeline/ForwardPass_StandardLighting_B.azsli
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#include <Atom/RPI/ShaderResourceGroups/DefaultDrawSrg.azsli>
+#include <Atom/Features/Pipeline/Forward/ForwardPassSrg.azsli>
+
+// TODO(MaterialPipeline): I don't like how this file #includes something from BasePBR. I'd rather it include a file
+// called StandardPBR_LightingData.azsli which just #includes the BasePBR file. Otherwise this looks suspcious like
```
jeremyong-az, line 13: s/suspcious/suspicious

FWIW, it looks suspicious perhaps because we are already familiar with what "Base" and "Standard" PBR mean in the context of our material system, but it's not unnatural for a more advanced lighting model to simply include and extend functionality from a simpler lighting model. I'll leave it to you if you think you'd rather keep this TODO though.

Gems/Atom/RPI/Code/Include/Atom/RPI.Edit/Material/MaterialPipelineSourceData.h
```diff
@@ -63,7 +64,7 @@ namespace AZStd
     {
         size_t operator()(const AZ::RPI::MaterialPipelineSourceData::ShaderTemplate& value) const
         {
-            return hash<AZStd::string>()(value.m_shader) ^ hash<AZStd::string>()(value.m_azsli);
+            return hash<AZStd::string>()(value.m_shader) ^ hash<AZStd::string>()(value.m_preMaterialAzsli) ^ hash<AZStd::string>()(value.m_postMaterialAzsli);
```
jeremyong-az, line 67: Nit: can leave as is but using `xor` as a hash combining operator can get you into trouble, so it'd be nice if we had a more "official" mechanism for this (if only to avoid the cognitive overhead of needing to determine each time we review the code if the hash combination is safe). In particular, `xor` is a symmetric operation, so if we exchanged the values of any of these fields, the resulting hash would be the same (again, doesn't matter here, but this is just an example of one of the pitfalls to be aware of).

Gems/Atom/RPI/Code/Source/RPI.Builders/Model/MaterialAssetBuilderComponent.cpp
```diff
@@ -81,36 +81,36 @@ namespace AZ
...
-
-                // Note that without the normal job dependencies, materials may not load or reload consistently or in sync with shader and
-                // material type changes without manually monitoring and handling dependency changes.
-                jobDependency.m_type = MaterialBuilderUtils::BuildersShouldFinalizeMaterialAssets()
-                    ? AssetBuilderSDK::JobDependencyType::Order
-                    : AssetBuilderSDK::JobDependencyType::OrderOnce;
-
-                jobDependencyList.push_back(jobDependency);
-            }
+            //if (conversionEnabled && !materialTypePath.empty())
```
jeremyong-az, line 84: Was this code intended to be removed?

Gems/Atom/RPI/Code/Source/RPI.Edit/Material/MaterialUtils.cpp
```diff
@@ -226,6 +227,40 @@ namespace AZ
                 // the string (for the extension) we can assume it's an Image file path.
                 return value.Is<AZStd::string>() && AzFramework::StringFunc::Contains(value.GetValue<AZStd::string>(), ".");
             }
+
+            bool IsValidName(AZStd::string_view name)
+            {
+                // Checks for a c++ style identifier
+                return AZStd::regex_match(name.begin(), name.end(), AZStd::regex("^[a-zA-Z_][a-zA-Z0-9_]*$"));
```
jeremyong-az, line 234: In some regex engines, it's important to cache the state machine constructed from parsing the regex string. Is that the case here?

Gems/Atom/RPI/Code/Source/RPI.Builders/Model/MaterialAssetBuilderComponent.cpp
santorac: Ugh, thank you. I kept accidentally staging this file and then unstaging later.

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/MainPipeline/MainPipeline.materialpipeline
```diff
@@ -2,23 +2,28 @@
...
             "shader": "../Common/DepthPass.shader.template",
-            "azsli": "../Common/DepthPass.azsli"
+            "preMaterialAzsli": "../Common/DepthPass_A.azsli",
+            "postMaterialAzsli": "../Common/DepthPass_B.azsli"
         },
         {
             "shader": "../Common/ShadowmapPass.shader.template",
-            "azsli": "../Common/ShadowmapPass.azsli"
+            "preMaterialAzsli": "../Common/DepthPass_A.azsli",
+            "postMaterialAzsli": "../Common/DepthPass_B.azsli"
```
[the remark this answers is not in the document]
santorac: I found a way to merge the preMaterialAzsli and postMaterialAzsli into a single file, and still put the material type source code in the middle.

Gems/Atom/RPI/Code/Include/Atom/RPI.Edit/Material/MaterialPipelineSourceData.h
santorac: I reverted this change.

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/LowEndPipeline/ForwardPass_StandardLighting_B.azsli
santorac: At some point I expect to rename these files anyway because "StandardPBR" is the name of a material type, but these are just different lighting model code, so should be named like "StandardLightingModel". I'll address the TODO in some way at that point.

Gems/Atom/RPI/Code/Source/RPI.Edit/Material/MaterialUtils.cpp
santorac: I searched the code for other uses of regex_match and they're all doing the same thing.

jeremyong-az: Looking at the code of `AZStd::regex`, it does appear to compile the regex every single time, so if we're doing this everywhere that's unfortunate. In this case the string is short and the parsing time probably isn't too bad, but maybe something to be aware of.

PR #13335, opened 2022-11-22
Multitest framework refactor

evanchia-ly-sdets:
## What does this PR do?
No functionality change. Refactors some of the complicated nested functions to be protected for maintainability.

## How was this PR tested?
Tested by running smoke and main suite locally.


Tools/LyTestTools/ly_test_tools/o3de/multi_test_framework.py
```diff
@@ -361,6 +363,122 @@ class MultiTestCollector(pytest.Class):
...
+
+            return spec_impl
+
+        @staticmethod
+        def _make_single_run(inner_test_spec: AbstractTestBase) -> typing.Callable:
+            """
+            Prepares a test run to be collected as a SingleTest. Single tests are ran independently of each other as
+            opposed to a SharedTest. This also includes setup() and teardown() functions if applicable.
+
+            :param inner_test_spec: The type of test (i.e. SingleTest, SharedTest)
```
Kadino, line 391: Given the param, I assume this allows non-SingleTest tests to be run as if they were a SingleTest. Might be worth explicitly noting in the docstring, if that's the case.

...if it's not the case, then the param docstring is incorrect! Also, this appears to slightly incorrect, as it expects the actual test object that will be run, not only its type.

Kadino (approved):
A nice reduction to the nested anonymous function complexity. Plus some basic tests of the plumbing!

Minor suggestion for docstrings, which hopefully isn't a functional issue:

AMZN-scspaldi (approved):
Looks good and appreciate the detail on adding extra tests.

Tools/LyTestTools/ly_test_tools/o3de/multi_test_framework.py
evanchia-ly-sdets: Fixed docstring to be more explicit

PR #13345, opened 2022-11-22
Asset Processor - UUID manager

amzn-mike:
## What does this PR do?

Adds a UUID Manager to the AP which will handle all UUID requests inside AP as well as generating UUIDs and storing them in metadata files.

Example Metadata File:

```
{
    "FileVersion": 1,
    "UUID": {
        "uuid": "{2A03492B-0E3F-4499-883B-A984107A348C}",
        "legacyUuids": [
            "{2C05FD2D-2C0D-5C82-BAFE-84D335697E67}",
            "{E4B0737D-256B-5768-A4C5-021DE37FE970}"
        ],
        "originalPath": "Icons/animation/animation_aimpose.png",
        "creationUnixEpochMS": 1669159808591,
        "__version": 0
    }
}
```

## How was this PR tested?

Added and ran unit tests.  The system is currently non-functional when actually running AP as none of the existing calls to the UUID API have been migrated to use the UUID manager.  That work will be done in a later PR.


Code/Tools/AssetProcessor/native/utilities/UuidManager.h
```diff
@@ -0,0 +1,112 @@
+/*
```
AMZN-stankowi, line 1: We don't want generation of UUIDs & meta data files outside the Asset Processor, correct? I assume that's why you put these files into the Asset Processor directly, instead of the tools framework.

AMZN-stankowi (approved):

Code/Tools/AssetProcessor/native/utilities/UuidManager.h
amzn-mike: Sort of.  This class has a cache and is relying on the file monitor to clear it, and the file monitor only exists in AP.  I think we probably want to expose an API for requesting UUIDs (and probably creating them) at the tools level but I'm not sure what the design for that will be yet, it might either just not use a cache or maybe it'll interface with the AP over network messages.

Code/Tools/AssetProcessor/native/tests/UuidManagerTests.cpp
```diff
@@ -0,0 +1,344 @@
...
+    {
+        static constexpr const char* TestFile = "mockfile.txt";
+        auto uuid = m_uuidInterface->GetUuid(TestFile);
+
+        EXPECT_FALSE(uuid.IsNull());
+        // Make sure a metadata file was created
+        EXPECT_TRUE(AZ::IO::FileIOBase::GetInstance()->Exists(AZStd::string::format("%s%s", TestFile, AzToolsFramework::MetadataManager::MetadataFileExtension).c_str()));
+    }
+
+    TEST_F(UuidManagerTests, GetUuidTwice_ReturnsSameUuid)
```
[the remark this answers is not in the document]
amzn-mike: Good idea, added

Code/Tools/AssetProcessor/native/utilities/UuidManager.h
```diff
@@ -0,0 +1,105 @@
...
+    struct IUuidRequests
+    {
+        AZ_RTTI(IUuidRequests, "{4EA7E0F6-CB4E-4F9C-ADBC-807676D51772}");
+
+        virtual ~IUuidRequests() = default;
+
+        //! Gets the canonical UUID for a given source asset.  A new metadata file may be created with a randomly generated UUID if
+        //! generation is enabled for the file type.  Otherwise the UUID returned will follow the legacy system based on the file path.
+        virtual AZ::Uuid GetUuid(const SourceAssetReference& sourceAsset) = 0;
+        //! Gets the set of legacy UUIDs for the given source asset.  These are all the UUIDs that may have been used to reference this asset based on previous generation methods.
```
jackalbe, line 33: does "legacy UUIDs" equal "previous UUIDs"? If so, can the term be switched to be previous UUIDs?

Code/Tools/AssetProcessor/native/utilities/UuidManager.h
```diff
@@ -0,0 +1,105 @@
...
+{
+    struct IUuidRequests
+    {
+        AZ_RTTI(IUuidRequests, "{4EA7E0F6-CB4E-4F9C-ADBC-807676D51772}");
+
+        virtual ~IUuidRequests() = default;
+
+        //! Gets the canonical UUID for a given source asset.  A new metadata file may be created with a randomly generated UUID if
+        //! generation is enabled for the file type.  Otherwise the UUID returned will follow the legacy system based on the file path.
+        virtual AZ::Uuid GetUuid(const SourceAssetReference& sourceAsset) = 0;
```
jackalbe, line 32: This API seems to only take in absolute paths to source assets, will there be a way to request UUIDs using relative paths to source assets? 

Possible API: "GetScanFoldersForRelativeSourceAssetPath(path)" or something?

Code/Tools/AssetProcessor/native/utilities/UuidManager.h
amzn-mike: Maybe? Right now what this means is "the old, deprecated, do-not-generate-new-references-using-this-algorithm set of UUIDs which previously might have been used to refer to this asset".  This is stored in a JSON format that anyone can modify with external tools, or may be updated in the future to include more things, so the list might expand beyond those 2 UUIDs we have currently.

I kind of prefer calling them legacy though because it does imply "these are old and all references using them should move away from them and use the canonical UUID".

Is there a reason to prefer the term "previous"?

Code/Tools/AssetProcessor/native/utilities/UuidManager.h
amzn-mike: I don't believe so, no.  This API is AP-specific right now (not sure yet what we'll be exposing to AzToolsFramework), so there's no need for that.  But we generally want to move away from accepting relative paths (by themselves) for assets, since without a known scanfolder, that involves the override system which is usually not intentional.  The override system isn't compatible with the metadata system anyway (for a given asset type).

Is there a usecase you're concerned about?

jackalbe: right now the procedural prefab system is referring to the assets (AzToolsFramework::Prefab::PrefabDomUtils::Internal::FixUpInvalidAssets) by hints to fix up the asset IDs 

As soon as the asset sub-id becomes {source UUID}."my_model.azmodel" then this will become moot, but the use of relative paths will need to persist for the foreseeable future; 

That being mentioned, the system could use a "go through the scan folders + this relative path" to make candidate SourceAssetReference sourceAsset or something;

amzn-mike: Ok, that should be fine as-is.  This is all AP internal code and that fixup happens in AssetBuilder.  This is basically adding support for replacing `CreateSafeSourceUUIDFromName` calls inside AP

jackalbe (approved):

junbo75 (approved):

PR #13346, opened 2022-11-22
Update the comment in FileWatcherUnitTest for the reason to disable the tests

junbo75:
Signed-off-by: Junbo Liang <68558268+junbo75@users.noreply.github.com>

## What does this PR do?

Update the comment in FileWatcherUnitTest for the reason to disable the tests

## How was this PR tested?

No logic changes in the test

### DIFF
--- a/Code/Tools/AssetProcessor/native/unittests/FileWatcherUnitTests.cpp
+++ b/Code/Tools/AssetProcessor/native/unittests/FileWatcherUnitTests.cpp
@@ -31,9 +31,10 @@ void FileWatcherUnitTest::TearDown()
     UnitTest::AssetProcessorUnitTestBase::TearDown();
 }
 
-// File operations on Linux behave differently on Linux than Windows.
-// The system under test doesn't yet handle Linux's file operations, which surfaced as this test arbitrarily passing and failing.
-// This test is disabled on Linux until the system under test can handle Linux file system behavior correctly.
+// Linux has a maximum number of watches allowed by the system, which is set by fs.inotify.max_user_watches.
+// FileWatcher will fail to watch any more folders once it hits this limit, which causes the test failures on Linux.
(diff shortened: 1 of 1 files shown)


AMZN-stankowi (approved):

cgalvan (approved):

amzn-mike (approved):

lmbr-pip:
Is there a way we can link this to the `FileWatcher: Unable to initialize inotify, file monitoring will not be available: Too many open files` error?

https://github.com/o3de/o3de/blob/757b544be06b30fd70ab664c314ef49f25eacb77/Code/Tools/AssetProcessor/Platform/Linux/native/FileWatcher/FileWatcher_linux.cpp#L35

We do highlight setting this value here:https://github.com/o3de/o3de/blob/757b544be06b30fd70ab664c314ef49f25eacb77/Code/Tools/AssetProcessor/Platform/Linux/native/FileWatcher/FileWatcher_linux.cpp#L69

But if we get the unable to initialize inotify message, theres not the same actionable error message so comment is hard to discover

jackalbe (approved):

junbo75:
> Is there a way we can link this to the `FileWatcher: Unable to initialize inotify, file monitoring will not be available: Too many open files` error?
> 
> https://github.com/o3de/o3de/blob/757b544be06b30fd70ab664c314ef49f25eacb77/Code/Tools/AssetProcessor/Platform/Linux/native/FileWatcher/FileWatcher_linux.cpp#L35
> 
> We do highlight setting this value here:
> 
> https://github.com/o3de/o3de/blob/757b544be06b30fd70ab664c314ef49f25eacb77/Code/Tools/AssetProcessor/Platform/Linux/native/FileWatcher/FileWatcher_linux.cpp#L69
> 
> But if we get the unable to initialize inotify message, theres not the same actionable error message so comment is hard to discover

The problem is that the base class of this test has an error absorber defined which hide the warning messages. I have cut a separate GHI to handle the error absorber problem.

PR #13347, opened 2022-11-22
Material Canvas: Fixing graph canvas crashes on shut down

gadams3:
## What does this PR do?

The first issue was improper widget cleanup in recently added code for the color picker button on the graph canvas vector property display class.

The second issue was caused by trying to access clipboard mime data before the application exits. In this case, I clear the clipboard when the application is destroyed. This prevents the crash from happening but the clipboard data should be managed by Qt.

Set up other widget parents to make sure they were cleaned up correctly while debugging the previously mentioned problems.

Resolves https://github.com/o3de/o3de/issues/12152

## How was this PR tested?

Manually and rapidly creating, destroying, cutting, copying, pasting, duplicating several notes, shutting down confirm exceptions or crashes no longer occur.

Being able to shut down cleanly with correct return codes will unlock automated test creation.

### DIFF
--- a/Gems/GraphCanvas/Code/Source/Components/NodePropertyDisplays/VectorNodePropertyDisplay.cpp
+++ b/Gems/GraphCanvas/Code/Source/Components/NodePropertyDisplays/VectorNodePropertyDisplay.cpp
@@ -355,14 +355,14 @@ namespace GraphCanvas
         {
             m_proxyWidget = new QGraphicsProxyWidget();
 
-            QWidget* widgetContainer  = new QWidget(nullptr);
-            QHBoxLayout* layout = new QHBoxLayout(widgetContainer);
-            widgetContainer->setSizePolicy(QSizePolicy::MinimumExpanding, QSizePolicy::Fixed);
-            
+            m_widgetContainer = new QWidget();
+            QHBoxLayout* layout = new QHBoxLayout(m_widgetContainer);
+            m_widgetContainer->setSizePolicy(QSizePolicy::MinimumExpanding, QSizePolicy::Fixed);
+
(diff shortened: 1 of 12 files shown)


cgalvan (approved):

lsemp3d (approved):

Gems/GraphCanvas/Code/Source/Components/NodePropertyDisplays/VectorNodePropertyDisplay.cpp
```diff
@@ -355,14 +355,14 @@ namespace GraphCanvas
...
-            
+            m_widgetContainer = new QWidget();
+            QHBoxLayout* layout = new QHBoxLayout(m_widgetContainer);
+            m_widgetContainer->setSizePolicy(QSizePolicy::MinimumExpanding, QSizePolicy::Fixed);
+
             layout->setAlignment(Qt::AlignLeft);
             layout->setContentsMargins(0, 0, 0, 0);
 
-            m_button = new QToolButton();
+            m_button = new QToolButton(m_widgetContainer);
```
onecent1101, line 365: if it is a best practice to assign parent object while creating qt object, we should document it somewhere so others can follow the same standard to avoid this issue happens again

onecent1101 (approved):

Gems/GraphCanvas/Code/Source/Components/NodePropertyDisplays/VectorNodePropertyDisplay.cpp
gadams3: Normally I do. This code was from graph canvas or script canvas. It was not the direct cause of the issues i was having. This update was a precaution.

PR #13358, opened 2022-11-23
Add position and normal reconstruction helper functions.

jonawals:
Signed-off-by: John <jonawals@amazon.com>

## What does this PR do?

This PR adds helper functions to reconstruct the view/world position and normal of fragments for Editor Mode Visual Feedback effects that require this information.

## How was this PR tested?

This PR has been tested manually in the editor.


hultonha (approved):
Change looks okay to me but I'm not a shader expert I'm afraid. Is this code used anywhere yet or is the plan to add usage in a separate PR? Thanks!

Gems/AtomLyIntegration/EditorModeFeedback/Assets/Shaders/EditorModeCommon.azsli
```diff
@@ -76,4 +82,61 @@ partial ShaderResourceGroup PassSrg : SRG_PerPass_WithFallback
     {
         return CalculateOutputColor(inColor, finalEffect, PassSrg::CalculateFinalBlendAmount(t));
     }
-}
\ No newline at end of file
+}
+
+//! Calculates the view position and normal from the fragment position.
+void ComputeViewNormalAndPosition(VSOutput IN, out float3 outPositionWS, out float3 outNormalWS)
```
moraaar, line 88: It would be a lot cheaper if the vertex shader used just outputs the normal and the position in the `VSOuput`, rather that recalculate everything in the pixel shader with these helpers. Is that not a possibility?

moraaar: I see. Thanks. I'll let the the guys from sig-graphics to take a look and give the second approval.

jeremyong-az: > It would be a lot cheaper if the vertex shader used just outputs the normal and the position in the VSOuput, rather that recalculate everything in the pixel shader with these helpers. Is that not a possibility

This isn't really true universally. Forwarding entries from VS to PS via ISBE (inter-stage buffer entries) allocates from LDS (local data share) storage which has a direct impact on occupancy (similar effect to allocating registers from the RF). In addition, vertices that are occluded are always fully shaded, as opposed to pixels that can be rejected with early-Z. The choice to compute data at vertex rate and pixel rate is, as a result, a more nuanced decision that depends on the geometry and rest of the shader itself. These routines should be orthogonal to that decision.

BTW, I could have sworn these routines already existed, will do some digging.

jeremyong-az: Oh, is this code just duplicated from the fullscreen shadow code implementation?

jeremyong-az (changes requested):
https://github.com/o3de/o3de/pull/13358/files#r1030686411

Gems/AtomLyIntegration/EditorModeFeedback/Assets/Shaders/EditorModeCommon.azsli
jonawals: Yeah, it's mostly a cannibalization of some existing code in O3DE. Both versions have their own coupling to their use cases, a generic version could quite easily be refactored and used in both, though 👍

jonawals: The shadow mapping impl has conditional compilation for both regular and compute shaders so it's not going to be able to be separated out into a generic util. I'm gonna stick my refactored version into  `ScreenSpaceUtil.azsli`, though.

Gems/Atom/Feature/Common/Assets/ShaderLib/Atom/Features/ScreenSpace/ScreenSpaceUtil.azsli
```diff
@@ -45,3 +45,62 @@ float4 WorldPositionFromDepthBufferMS(const Texture2DMS<float2> depthStencilText
     depthStencilTexture.GetDimensions(dimensions.x, dimensions.y, sampleCount);
     return WorldPositionFromDepthBuffer(dimensions, zDepth, screenCoords);
 }
+
+//! Calculates the view position and normal from the fragment position.
+void ViewPositionAndNormalFromDepth(const Texture2DMS<float2> depthStencilTexture, const VSOutput IN, out float3 outPositionWS, out float3 outNormalWS)
```
jeremyong-az, line 50: Thanks for the change, but I'm not seeing a corresponding change for the fullscreen shadow code to actually use this. (reminder, shader code needs the exact same discipline as normal C++ code and test code).

Issue #13191, opened 2022-11-15
Deselecting an asset with dependencies during renaming without any changes made causes an "Asset renamed" window to pop up


LB-ArturZieba:
**Describe the bug**
If a file or a folder that contains files with dependencies is deselected during renaming, without making any changes, a window will pop up informing the user about the asset rename.

**Steps to reproduce**
1. In the Asset Browser select a file or a folder that contains files with a dependency (for example the project's folder).
2. Right-click the file/folder from step 1 and select Rename Folder from the context menu.
3. Do not change anything and click out of the name selection.

**Expected behavior**
Nothing pops up.

**Actual behavior**
Pop up informing that the asset has been renamed pops up.

**Video**

https://user-images.githubusercontent.com/86952082/201920490-9f5b1a7e-1e0c-475e-b6e4-e4b01c5f7013.mp4

**Found in Branch**
Development (ba92844)

**Desktop**
 - Device: PC
 - OS: Windows
 - Version 11
 - CPU AMD Ryzen 5 3600
 - GPU NVIDIA GeForce RTX 2060 SUPER
 - Memory 16GB

LB-ArturZieba:
Issue verified as fixed on the Development (ecda842) build.

PR #13366, opened 2022-11-23
Deselecting an asset with dependencies during renaming without any changes made causes an "Asset renamed" window to pop up

jjjoness:
Signed-off-by: John Jones-Steele <82226755+jjjoness@users.noreply.github.com>

Fixes bug #13191

Old behavior: If an asset was deselected while renaming and the name hadn't changed it still tried to rename it.
New behavior:If an asset was deselected while renaming and the name hadn't changed it early exits from the rename task.

## How was this PR tested?

Tested in Asset Browser.


amzn-tmryan (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Views/AssetBrowserTreeView.cpp
```diff
@@ -702,7 +702,10 @@ namespace AzToolsFramework
                 toPath.ReplaceFilename(newVal.toStdString().c_str());
                 toPath.ReplaceExtension(extension);
             }
-
+            if (fromPath == toPath)
+            {
+                return;
+            }
```
gadams3, line 708: nit - worth adding a comment to explain why the check and early return is needed.

gadams3 (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Views/AssetBrowserTreeView.cpp
jjjoness: Will do.

PR #13370, opened 2022-11-23
Metadata Manager - Fix API returning json Value which references deallocated memory

amzn-mike:
## What does this PR do?

Fix MetadataManager returning rapidjson::Value which pointed to deallocated memory.

Replaced Value with Document outValue since Document owns its own memory whereas Value does not

## How was this PR tested?

Ran unit tests


Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/MetadataManager.cpp
```diff
@@ -91,13 +91,13 @@ namespace AzToolsFramework
             return false;
         }
 
-        outValue = *value;
+        outValue.CopyFrom(*value, document.GetAllocator());
```
lawsonamzn, line 94: It probably wont affect anything but it may be worth doing
```
outValue = rapidjson_ly::Document(); 
```

Before calling copyfrom.  Because copyfrom doesn't clear memory.

lawsonamzn (approved):

lawsonamzn (approved):

lumberyard-employee-dm (approved):

Code/Framework/AzToolsFramework/Tests/MetadataManagerTests.cpp
```diff
@@ -310,7 +310,7 @@ namespace UnitTest
         EXPECT_TRUE(m_metadata->GetValueVersion("mockfile", "/Test", version));
         EXPECT_EQ(version, 1);
 
-        rapidjson_ly::Value inValue;
+        rapidjson_ly::Document inValue;
         EXPECT_TRUE(m_metadata->GetJsonValue("mockfile", "/Test", inValue));
```
burelc-amzn, line 314: This should probably use `ASSERT_TRUE`. Continuing to access members of the document if `GetJsonValue` failed will probably return fail conditions, and the verifications done further in this test will surely fail.

Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/MetadataManager.h
```diff
@@ -90,7 +90,7 @@ namespace AzToolsFramework
 
     public:
         bool GetValue(AZ::IO::PathView file, AZStd::string_view key, void* outValue, AZ::Uuid typeId) override;
-        bool GetJsonValue(AZ::IO::PathView file, AZStd::string_view key, rapidjson_ly::Value& outValue) override;
+        bool GetJsonValue(AZ::IO::PathView file, AZStd::string_view key, rapidjson_ly::Document& outValue) override;
```
burelc-amzn, line 93: Is this method misnamed now? It seems to make sense to pass a `Value&` out-parameter to `GetJsonValue()`, but less so to pass a `Document&`. But maybe the purpose of the method didn't change, so the name still is appropriate?

lawsonamzn: Maybe its misnamed.  Ideally we'd potentially start moving to AZ::Dom, in which case, Value won't matter.  in rapidjson, Document is a value, because it derives from value, but it gets weird...

amzn-mike: How about just GetJson?

PR #13381, opened 2022-11-24
Fix for new Mac Xcode 14.1 64bit to 32bit truncation errors

rgba16f:
On Mac long is a 64 bit type, on windows it's a 32 bit type, most of the errors are from assuming long can silently cast to a 32 bit int.
Signed-off-by: rgba16f <82187279+rgba16f@users.noreply.github.com>

### DIFF
--- a/Code/Editor/PythonEditorFuncs.cpp
+++ b/Code/Editor/PythonEditorFuncs.cpp
@@ -61,7 +61,7 @@ namespace
         }
         else if (pCVar->GetType() == CVAR_INT)
         {
-            PySetCVarFromInt(pName, std::stol(pValue));
+            PySetCVarFromInt(pName, static_cast<int>(std::stol(pValue)));
         }
         else if (pCVar->GetType() == CVAR_FLOAT)
         {
--- a/Code/Editor/QtUI/ColorButton_mac.mm
+++ b/Code/Editor/QtUI/ColorButton_mac.mm
@@ -19,7 +19,7 @@
 void macRaiseWindowDelayed(QWidget* window)
 {
     NSWindow* nativeWindow = [(NSView*)(window->winId()) window];
-    CGWindowLevel level = nativeWindow.level;
+    CGWindowLevel level = static_cast<CGWindowLevel>(nativeWindow.level);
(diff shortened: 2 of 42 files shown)


micronAMZN (approved):
I have a successful configure/build on an M1 Max Mac Studio. Asset Processor is getting hung up on some (mainly .exr) assets, but the editor did launch, even though it was waiting endlessly on AP and I couldn't actually do anything with it.

Total build time, just over 14 mins. Pretty great.

hultonha:
Much as these warnings/errors are undoubtedly irritating, the warning is trying to say that there could be a potential bug here where information could be lost. Adding `static_cast` in every case might not always be the right thing to do as we're then just ignoring the potential problem. I know it's likely safe in most cases to add the cast but it might be worth doing an audit to see if any types should be updated to `uint64_t` for example, or changed to be `int` instead of `long` if a 32 bit width was intended (I know this can spiral quickly so if it seems like that's happening it might not be worth it, but I just wanted to mention it as something to consider).

rgba16f:
> Much as these warnings/errors are undoubtedly irritating, the warning is trying to say that there could be a potential bug here where information could be lost. Adding `static_cast` in every case might not always be the right thing to do as we're then just ignoring the potential problem. I know it's likely safe in most cases to add the cast but it might be worth doing an audit to see if any types should be updated to `uint64_t` for example, or changed to be `int` instead of `long` if a 32 bit width was intended (I know this can spiral quickly so if it seems like that's happening it might not be worth it, but I just wanted to mention it as something to consider).

I agree, but it seemed like the best first step was to identify all the places where we were doing the unsafe behavior. This is a short term fix until all the respective code owners can look at their code and make the correct fixes.

hultonha:
> _I agree, but it seemed like the best first step was to identify all the places where we were doing the unsafe behavior. This is a short term fix until all the respective code owners can look at their code and make the correct fixes._

Okay understood, are there any follow-up GHIs to track the various areas to review? If there's any sig-simulation or viewport related files I'll make a note of them to review 👍 Thanks!

lumberyard-employee-dm (approved):

Issue #13235, opened 2022-11-16
Incorrect rendering of search highlight in anim graph node palette window


MiKom:
## Steps to reproduce
1. Open animation editor
2. Open some anim graph asset (e.g. AutomatedTesting\Animations\AnimationEditorFiles\sample0.animgraph)
3. Open the node palette window (View → Palette window)
4. Enter some search term (e.g. `ect`)

## Expected result
Highlights of the text match in the tree are semi-transparent

## Actual result
Highlights of the text matches are opaque:

![Image](https://user-images.githubusercontent.com/104767/202231579-7ff93fe9-8cf0-48e6-a65e-79b1f04e3456.png)

FuzzyCarterAWS:
Verified that this is an issue for me too. This wasn't present before the widget was converted.

MiKom:
Yup. I finally found the root cause today. I'll work on a solution tomorrow.

PR #13384, opened 2022-11-24
Fix style application in EMotionFX

MiKom:
For some reason the Editor.qss stylesheet is not applied to the NodePaletteWidget when it's inserted into the dock widget in EMotionFX.

This widget is later reparented to the chain originating in the editor main window so in theory it should be applied and the palette for the tree view delegate should be proper.

As a workaround, we apply the style directly.

Fixes: #13235

## What does this PR do?

Fixes the search highlight in Node Palette window in EMotionFX

## How was this PR tested?

Manually

### DIFF
--- a/Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/NodePaletteWidget.cpp
+++ b/Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/NodePaletteWidget.cpp
@@ -20,6 +20,7 @@
 #include <GraphCanvas/Widgets/NodePalette/TreeItems/NodePaletteTreeItem.h>
 #include <AzQtComponents/Utilities/Conversions.h>
 #include "AnimGraphPlugin.h"
+#include <AzQtComponents/Components/StyleManager.h>
 #include <QVBoxLayout>
 #include <QIcon>
 #include <QAction>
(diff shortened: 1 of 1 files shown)


moraaar (approved):

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/NodePaletteWidget.cpp
```diff
@@ -88,6 +89,9 @@ namespace EMStudio
         m_palette->SetupNodePalette(config);
         m_palette->hide();
         m_layout->addWidget(m_palette);
+        // TODO(#13382): Investigate why we need to apply the style here even though it's applied globally
```
hultonha, line 92: nit: I know this is super minor but could you please update the 'TODO' to instead say `GHI-13382` instead please

```
// GHI-13382 - Investigate why we need to apply the style here even though it's applied globally
```

hultonha (approved):
Change looks good to me, I just have a minor nit and it would be good to have at least one of the rest of the Editor team take a quick look from a Qt perspective (added as reviewers)

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/NodePaletteWidget.cpp
MiKom: Sure.

MiKom: Do we want to retain the `TODO` so it shows up in tools or just the `// GHI-13382 ...`?

hultonha: Just `GHI-13382`

MiKom: Done

PR #13389, opened 2022-11-25
Variable Rate Shading support for Atom_RHI

### LINKED ISSUE #12877: [RHI] Add support for Variable Rate Shading
Traditionally, the density of fragment shader invocations for a given triangle is constant across the framebuffer. Due to performance constraints, a graphics renderer can't always afford to deliver the same level of quality to every part of its output image. Variable-rate shading—or coarse pixel shading—is a mechanism that lets you allocate rendering performance/power at rates that vary across your rendered image.

In some cases, the shading rate can be reduced with little or no reduction in perceptible output quality; leading to a performance improvement that's essentially free of charge.

One of the most common bottlenecks for graphic applications is the cost of shader invocations done during a frame. As resolutions continue to increase, this problem is exacerbated and any saving in this area becomes even more important. This is specially true for VR platforms, where resolutions double and framerate demands are high. Shading millions of pixels at high speed is crucial for maintaining a great experience and avoiding VR sickness.

The solution will require to add support to the RHI for Variable-rate shading. This way developers can take advantage of it and improve performance by reducing the number of shading invocations.

- The simple alternative is reducing the size of the render targets used and upscale before presenting to the screen. This would decrease the bandwidth and the number of shading invocations, but at a cost of quality through the whole screen. With this approach we cannot focus the quality in areas that are important (like the center in VR content).

- Another approach is using a Fixed Foveated Rendering (FFR) provided by the platform. Certain VR devices (e.g. Quest 2) provides fixed areas where the quality is automatically decrease. These areas cannot be change and they only work if the app is rendering directly to the swapchain images. Currently, the O3D3 pipelines do not render to the swapchain, but to intermediate textures and at the content is copy to the swapchain as the last command. This approach is only limited to certain VR devices.

**Additional context**
Please refer to the proposed [RFC](https://github.com/o3de/sig-graphics-audio/issues/87)

akioCL:
- Add support on the RHI for an agnostic way to handle Variable Rate Shading. 
- Add members to the DeviceFeatures to signal if the feature is supported by the device. Also add the limits used byt the Variable Rate Shading feature.
- Add a new attachment type called "ShadingRate" so it can be used during scope declaration.
- Add support for Variable Rate Shading on Vulkan using VK_KHR_fragment_shading_rate and VK_EXT_fragment_density_map. On Vulkan, there's two extension that support it. The VK_EXT_fragment_density_map extension uses the VK_KHR_create_renderpass2 extension as a dependency, so some refactor was done in the way we build the Vulkan's renderpass to support the standard and the renderpass2 way of creating renderpasses. Extra information was needed for the RenderAttachmentLayout due to the requirements of the renderpass2 extension.
- Created a new Conversion.h file for the Atom_RHI_Vulkan.Private library and moved the functions that depended on it from Atom_RHI_Vulkan.Reflect to the new file. This way we prevent a circular dependency.
- Removed some unnecessary copies for device creation when dealing with extensions.
-  Add support for VRS on DX12. Add check to see if the feature is supported by the device. Add a function to get the value needed for populating a shading rate image.

Resolves #12877

## What does this PR do?

Implementation of Variable Rate Shading

https://github.com/o3de/sig-graphics-audio/issues/87

## How was this PR tested?

Run the new Variable Rate Shading sample from AtomSampleViewer
[PR](https://github.com/o3de/o3de-atom-sampleviewer/pull/555) for the sample


Gems/Atom/RHI/Vulkan/Code/Source/RHI/RenderPass.cpp
```diff
@@ -16,6 +16,356 @@ namespace AZ
...
+
+            using RenderPassResult = AZStd::tuple<VkResult, VkRenderPass>;
+
+            template<
+                typename VkRenderPassCreateInfoTraits,
+                typename VkAttachmentDescriptionTraits,
+                typename VkAttachmentReferenceTraits,
+                typename VkSubpassDescriptionTraits,
+                typename VkSubpassDependencyTraits>
+            struct VkRenderpassBuilder
```
moudgils, line 41: This struct needs a comment explaining why it is setup this way and what functionality it applies. I am assuming that the only new changes here are related to ShadingRate and everything else was copied over from before?

Gems/Atom/RHI/Vulkan/Code/Source/RHI/RenderPass.cpp
```diff
@@ -16,6 +16,356 @@ namespace AZ
...
+                using VkSubpassDependencyType = typename VkSubpassDependencyTraits::value_type;
+
+                struct SubpassInfo
+                {
+                    AZStd::array<AZStd::vector<VkAttachmentReferenceType>, AttachmentTypeCount> m_attachmentReferences;
+                    AZStd::vector<uint32_t> m_preserveAttachments;
+                    VkFragmentShadingRateAttachmentInfoKHR m_shadingRateAttachmentExtension;
+                };
+
+                RenderPassResult Build(const RenderPass::Descriptor& descriptor)
```
moudgils, line 62: Would be good to add a comment above each method since they dont exist in the header file

Gems/Atom/RHI/Vulkan/Code/Source/RHI/RenderPass.cpp
```diff
@@ -16,6 +16,356 @@ namespace AZ
...
+                            createInfo.pNext = &fdmAttachmentCreateInfo;
+                        }
+                    }
+
+                    return Create<VkRenderPassCreateInfoType>(createInfo);
+                }
+
+            private:
+
+                // Classes used for handeling setting a member that may not exist.
```
moudgils, line 110: handling

moudgils (commented):
Given the size of the PR are you able to add more details in the PR description based on the changes here. It will just help anyone not familiar with RHI to understand the changes.

Gems/Atom/RHI/Code/Tests/Device.h
```diff
@@ -65,6 +65,11 @@ namespace UnitTest
         AZ::RHI::ResourceMemoryRequirements GetResourceMemoryRequirements([[maybe_unused]] const AZ::RHI::BufferDescriptor& descriptor) { return AZ::RHI::ResourceMemoryRequirements{}; };
 
         void ObjectCollectionNotify(AZ::RHI::ObjectCollectorNotifyFunction notifyFunction) override {}
+
+        AZ::RHI::ShadingRateImageValue ConvertShadingRate(AZ::RHI::ShadingRate rate) override
```
jeremyong-az, line 69: Input rate unused?

Gems/Atom/RHI/DX12/Code/Source/RHI/CommandList.cpp
```diff
@@ -707,6 +750,31 @@ namespace AZ
...
             }
+
+#ifdef AZ_DX12_VRS_SUPPORT
+            if (RHI::CheckBitsAll(GetDevice().GetFeatures().m_shadingRateTypeMask, RHI::ShadingRateTypeFlags::PerImage))
+            {
+                auto commandList5 = DX12ResourceCast<ID3D12GraphicsCommandList5>(GetCommandList());
+                AZ_Assert(commandList5, "Failed to cast command list to ID3D12GraphicsCommandList5");
+                if (commandList5)
+                {
+                    if (shadingRateAttachment)
```
jeremyong-az, line 761: I suggest checking for a shading rate attachment first, before doing the COM interface query and asserts and capability checking (since in most cases here, the SR attachment won't be active)

rgba16f (approved):
Just a few nits, should probably fix the ShaderRageMode though

Gems/Atom/RHI/Vulkan/Code/Source/RHI/Device.cpp
```diff
@@ -829,6 +847,95 @@ namespace AZ
...
+            {
+                // Fragment sizes are encoded in a single texel as follows:
+                // size(w) = 2^((texel/4) & 3)
+                // size(h) = 2^(texel & 3)
+
+                uint8_t rate_w, rate_h;
+                switch (rate)
+                {
+                case RHI::ShadingRate::Rate1x1:
+                    rate_w = rate_h = 0;
```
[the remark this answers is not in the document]
jeremyong-az: Yea, I know the encoding uses a mask as you pointed out, I just meant that in code, I could see a person being confused if they looked at `rate_w` and it was `0` (since this is not actually the rate). I'd suggest a name like `encoded_rate_w` maybe? That said, not a huge deal at all

Gems/Atom/RHI/DX12/Code/Source/RHI/CommandList.cpp
```diff
@@ -612,6 +625,35 @@ namespace AZ
...
+            {
+                return;
+            }
+
+#ifdef AZ_DX12_VRS_SUPPORT
+            auto& device = static_cast<Device&>(GetDevice());
+            if (RHI::CheckBitsAll(device.GetFeatures().m_shadingRateTypeMask, RHI::ShadingRateTypeFlags::PerDraw))
+            {
+                AZStd::array<D3D12_SHADING_RATE_COMBINER, RHI::ShadingRateCombinators::array_size> d3d12Combinators;
+                for (int i = 0; i < m_state.m_shadingRateState.m_shadingRateCombinators.size(); ++i)
```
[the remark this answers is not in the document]
jeremyong-az: Thinking about it, it feels like the feature check really should only happen when we try to assign any shading rate combinators in the first place. Here, I would expect the feature check to be an assert right?

jeremyong-az (approved):
Thanks just a couple minor comments but looking good!

Gems/Atom/RHI/Vulkan/Code/Source/RHI/RenderPass.cpp
```diff
@@ -16,6 +16,387 @@ namespace AZ
...
+                    structure.pNext = next;
+                }
+
+                template<typename StructureType>
+                void SetNext([[maybe_unused]] StructureType& structure, [[maybe_unused]] const void* next, general_) const
+                {
+                }
+
+                const Device& m_device;
+                const RenderPass::Descriptor* m_descriptor;
```
moraaar, line 373: [minor] Initialize `m_descriptor` to nullptr.

Gems/Atom/RHI/Vulkan/Code/Source/RHI/RenderPass.cpp
```diff
@@ -16,6 +16,387 @@ namespace AZ
...
+                // Holds the information about a subpass
+                struct SubpassInfo
+                {
+                    AZStd::array<AZStd::vector<VkAttachmentReferenceType>, AttachmentTypeCount> m_attachmentReferences;
+                    AZStd::vector<uint32_t> m_preserveAttachments;
+                    // Used only if the pass uses fragment rate attachment.
+                    VkFragmentShadingRateAttachmentInfoKHR m_shadingRateAttachmentExtension;
+                };
+
+                RenderPassResult Build(const RenderPass::Descriptor& descriptor)
```
moraaar, line 75: Just an opinion. This class will be more readable and easy to see its interface if the functions implementations would be move outside the class.

Gems/Atom/RHI/Vulkan/Code/Source/RHI/RenderPass.cpp
```diff
@@ -16,6 +16,387 @@ namespace AZ
...
+                    createInfo.pDependencies = subpassDependencies.empty() ? nullptr : subpassDependencies.data();
+
+                    // Fragment shade attachments are declared at a renderpass level (same for all subpasses), so we need to
+                    // check if we have one as part of the renderpass declaration. We check if the first subpass contains the shading rate attachment,
+                    // and use that one for the whole renderpass. If more than one is found, we raise an assert because
+                    // there can only be one fragment attachment per renderpass.
+                    VkRenderPassFragmentDensityMapCreateInfoEXT fdmAttachmentCreateInfo;
+                    if (m_device.GetImageShadingRateMode() == Device::ShadingRateImageMode::DensityMap)
+                    {
+                        auto& shadingRateAttachmentRefList = subpassInfo.front()
```
moraaar, line 106: Maybe assert that `subpassInfo`  is not empty before accessing its front?

Gems/Atom/RHI/Vulkan/Code/Source/RHI/Scope.cpp
```diff
@@ -23,7 +23,7 @@
 #include <RHI/SwapChain.h>
 #include <RHI/Framebuffer.h>
 #include <RHI/Device.h>
-#include <Atom/RHI.Reflect/Vulkan/Conversion.h>
```
moraaar, line 26: `Conversion.h` needs to be exposed in `Atom/RHI.Reflect/Vulkan` because it needs to be accessible outside Atom by other gems which use Vulkan's `RHI.Reflect` target. [OpenXRVk](https://github.com/o3de/o3de-extras/blob/6de5c12c77a5cdabdbf402a4ab1a08766afe9aa3/Gems/OpenXRVk/Code/Source/OpenXRVkSwapChain.cpp#L9) gem uses it. Please make sure this doesn't break OpenXRVk.

moraaar: Thanks for confirming, I can see that `ConvertFormat` is still part of `Atom/RHI.Reflect/Vulkan/Conversion.h`.

moraaar (approved):

Issue #13327, opened 2022-11-22
Component Mode Switcher - Undoing during Component Mode will duplicate the switcher button


AMZN-daimini:
**Describe the bug**
Undoing an operation during Component Mode duplicates the button for that mode in the Switcher.

**Steps to reproduce**
1. Create an entity with Spline and Tube Shape components
2. Enter Component Mode for the Tube Shape component
3. Change the radius of one of the points
4. Use the "Reset Radii" action (R)
5. Undo the interaction (Ctrl+Z)

**Expected behavior**
The state of the Editor is restored to the same as before using the "Reset Radii" action

**Actual behavior**
The undo operation works but the Tube Shape Component appears twice in the Switcher.

**Screenshots/Video**
![UndoSwitcher](https://user-images.githubusercontent.com/82231674/203213500-1bb99158-1c35-4084-8f26-4075b841fd39.gif)

**Found in Branch**
Latest development

PR #13391, opened 2022-11-25
Bug Fix: Component mode switcher duplicates switcher component icon on undo while in component mode

amzn-ahmadkrm:

## What does this PR do?

Fixes https://github.com/o3de/o3de/issues/13327

The problem was that the switcher duplicated the active component mode icon when it was in component mode and the user carried out a component editing action then pressed undo. On Undo, all the components disconnect from the component mode delegate and reconnect. This switcher uses this notification to update add/remove components, but when the Editor is in component mode there is the "active" button which doesn't get affected leading to the icon being duplicated.

The switcher now checks if it is in component mode before adding / removing components to itself under the assumption that the Entity composition cannot change in component mode.

## How was this PR tested?

Unit testing:

Before:
VsDebugConsole_yIaz6HprOD

After:
VsDebugConsole_TFPOOEGPTp


moraaar (approved):
LGTM

hultonha (approved):
Thanks for jumping on this Ahmad and great work adding a test too! 🥳

PR #13403, opened 2022-11-25
Toggle Filter For Project and Source Assets

### LINKED ISSUE #12874: Asset Browser  - Hide and show engine assets
As a user, I need the ability to occasionally turn on (toggle switch) and see my engine assets from inside the Asset Browser. By default, we should hide any assets and folder if they can't be used (in their current state) within their project. These files can make the Asset browser feel very noisey for our customers, and there is very little value for customers seeing this data on a daily basis. We should also set this setting to be the default for most people.

 

The specific files or folder we can hide from user are as followed.
 * intermediate assets,
 * etc.

 
image

image

dtamkin1:

## What does this PR do?

Added a toggle button to filter out assets that are not project and source assets. Resolves #12874. 
This filter is added behind the WIP Asset Browser flag.
The Tree View and List View options will be removed in the next PR, as per discussion with @rainbj

Editor_sVaGKjnWTk

## How was this PR tested?

Locally with the asset browser


Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Search/Filter.cpp
```diff
@@ -404,6 +404,30 @@ namespace AzToolsFramework
...
+        QString AssetPathFilter::GetNameInternal() const
+        {
+            return m_assetPath;
+        }
+
+        bool AssetPathFilter::MatchInternal(const AssetBrowserEntry* entry) const
+        {
+            QStringList projectDirectory = m_assetPath.split("\\");
+            QString entryPath = QString::fromUtf8(entry->GetFullPath().data());
+            return StringMatch(projectDirectory.last(), entryPath);
```
cgalvan, line 428: The `StringMatch` method only checks if text contains what is being searched for, but we can be more explicit than that. We should instead check if the entryPath begins with the project directory path, which will ensure that the entry exists in a sub-directory of the intended search path. Instead of splitting by `\\` which would only work on windows, you can use the AZ::IO::Path to make sure the paths you are comparing have the same path separators.

There is even a `IsRelativeTo` on `AZ::IO::BasicPath` that looks like it might do the trick as well.

cgalvan (changes requested):

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Search/Filter.cpp
dtamkin1: Thanks for the feedback, I wasn't sure what the best way to filter for path descendants was. Fixed it with the `IsRelativeTo` method!

cgalvan (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Search/Filter.cpp
```diff
@@ -404,6 +404,29 @@ namespace AzToolsFramework
...
         }
 
+        //////////////////////////////////////////////////////////////////////////
+        // AssetPathFilter
+        //////////////////////////////////////////////////////////////////////////
+        AssetPathFilter::AssetPathFilter()
+        {
+        }
+
+        void AssetPathFilter::SetAssetPath(AZ::IO::Path path)
```
amzn-tmryan, line 414: nit: Might be able to pass by `const AZ::IO::Path&`.

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Search/SearchWidget.cpp
```diff
@@ -159,6 +164,27 @@ namespace AzToolsFramework
...
+            sourceFilter->SetName("Source");
+            sourceFilter->SetEntryType(AssetBrowserEntry::AssetEntryType::Source);
+            m_projectSourceFilter->AddFilter(FilterConstType(sourceFilter));
+
+            auto pathFilter = new AssetPathFilter();
+            pathFilter->SetAssetPath(AZStd::string_view{ AZ::Utils::GetProjectPath() });
+            m_projectSourceFilter->AddFilter(FilterConstType(pathFilter));
+        }
+
+        void SearchWidget::FilterProjectSourceAssets()
```
amzn-tmryan, line 178: nit: This function reads like it sometimes removes the filter instead of performing a filter. Would a name like `ToggleProjectSourceAssetsFilter` be correct here, or is removing the filter considered a type of filtering?

amzn-tmryan (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Search/Filter.cpp
dtamkin1: Thanks, I'll try that out!

PR #13405, opened 2022-11-25
Material Canvas: Fixing creation and serialization of node groups, bookmarks, comments

gadams3:
## What does this PR do?

- Implemented construct presets for material canvas graph view. The toolbar actions for creating groups of nodes are only functional if presets are set up.
- Graph model controller was not sending all of the requests and notifications needed to fully initialize a new graph, update bookmarks, and allow groups to resolve and re add contained nodes.
- Graph model controller was serializing “other” redundant metadata and then attempts to capture information about comments and groups. The information is already captured as part of the scene metadata when it is serialized.
- Graph model controller does not currently clear metadata from nodes that have been deleted. This may be a requirement for landscape canvas and will need to be addressed so excess data is not retained or serialized between sessions.

## How was this PR tested?

Verified that node groups can now be created in material canvas from the toolbar.
Verified that bookmarks and node groups are serialized with material graph file and restored on reload.
Verified that the collapsed state of note groups is correctly serialized and restored.
Verified that material graphs continue to compile correctly with instruction and input nodes wrapped in node groups.

image
image

### DIFF
--- a/Gems/Atom/Tools/MaterialCanvas/Code/Source/Window/MaterialCanvasGraphView.cpp
+++ b/Gems/Atom/Tools/MaterialCanvas/Code/Source/Window/MaterialCanvasGraphView.cpp
@@ -11,14 +11,51 @@
 
 namespace MaterialCanvas
 {
+    void MaterialCanvasGraphConstructPresets::InitializeConstructType(GraphCanvas::ConstructType constructType)
+    {
+        if (constructType == GraphCanvas::ConstructType::NodeGroup)
+        {
+            auto presetBucket = static_cast<GraphCanvas::NodeGroupPresetBucket*>(ModPresetBucket(GraphCanvas::ConstructType::NodeGroup));
+            if (presetBucket)
+            {
+                presetBucket->ClearPresets();
+
+                const AZStd::map<AZStd::string, AZ::Color> defaultPresetGroups = {
+                    { "Logic", AZ::Color(0.188f, 0.972f, 0.243f, 1.0f) },
+                    { "Function", AZ::Color(0.396f, 0.788f, 0.788f, 1.0f) },
+                    { "Output", AZ::Color(0.866f, 0.498f, 0.427f, 1.0f) },
+                    { "Input", AZ::Color(0.396f, 0.788f, 0.549f, 1.0f) } };
+
+                for (const auto& defaultPresetPair : defaultPresetGroups)
+                {
+                    if (auto preset = presetBucket->CreateNewPreset(defaultPresetPair.first))
+                    {
+                        const auto& presetSaveData = preset->GetPresetData();
+                        if (auto saveData = presetSaveData.FindSaveDataAs<GraphCanvas::CommentNodeTextSaveData>())
+                        {
+                            saveData->m_backgroundColor = defaultPresetPair.second;
+                        }
+                    }
+                }
+            }
+        }
+        else if (constructType == GraphCanvas::ConstructType::CommentNode)
+        {
+            auto presetBucket = static_cast<GraphCanvas::CommentPresetBucket*>(ModPresetBucket(GraphCanvas::ConstructType::CommentNode));
+            if (presetBucket)
+            {
+                presetBucket->ClearPresets();
+            }
+        }
+    }
+
     MaterialCanvasGraphView::MaterialCanvasGraphView(
-        const AZ::Crc32& toolId,
-        const AZ::Uuid& documentId,
-        const AtomToolsFramework::GraphViewConfig& graphViewConfig,
-        QWidget* parent)
+        const AZ::Crc32& toolId, const AZ::Uuid& documentId, const AtomToolsFramework::GraphViewConfig& graphViewConfig, QWidget* parent)
         : AtomToolsFramework::GraphView(toolId, GraphCanvas::GraphId(), graphViewConfig, parent)
         , m_documentId(documentId)
     {
+        m_materialCanvasGraphConstructPresets.SetEditorId(m_toolId);
+
         AtomToolsFramework::AtomToolsDocumentNotificationBus::Handler::BusConnect(m_toolId);
         OnDocumentOpened(m_documentId);
         m_openedBefore = false;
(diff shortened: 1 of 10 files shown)


Gems/GraphModel/Code/Source/Integration/GraphController.cpp
```diff
@@ -1484,64 +1483,25 @@ namespace GraphModelIntegration
...
-            else
-            {
-                container = mapIter->second;
-            }
-
+            auto container = AZStd::make_shared<GraphCanvas::EntitySaveDataContainer>();
             GraphCanvas::EntitySaveDataRequestBus::Event(
-                graphCanvasElement, &GraphCanvas::EntitySaveDataRequests::WriteSaveData, (*container));
+                graphCanvasElement, &GraphCanvas::EntitySaveDataRequests::WriteSaveData, *container);
+            graphCanvasMetadata->m_sceneMetadata = container;
```
cgalvan, line 1504: Previously the comment/group data went into other, but it sounds like you've discovered it's been going into the scene element as well. The `graphCanvasElement` used to be different for the comment/group, does that path still happen but its just being duplicated by the scene's `graphCanvasElement` storing that info as well?

cgalvan (approved):

Gems/GraphModel/Code/Source/Integration/GraphController.cpp
gadams3: void SceneComponent::WriteSaveData(EntitySaveDataContainer& saveDataContainer) const seems to be responsible for recording all of that data, which is very heavyweight because it's serializing components for everything not represented by graph model nodes. There is an equivalent function for restoring the graph data. The graph controller was also writing out the data for the other group but it was restoring it with an invalid/default entity ID. So I don't think the data was getting applied.

Whenever landscape canvas closes a graph, it manually sends a notification to mark the graph canvas scene dirty. That will record the data for the scene and all of the comments, bookmarks, etc. I added the same notification to the material canvas document class whenever the graph is loaded or modified and the state recorded for undo/redo.

After that,  groups, comments, and bookmarks were being restored but the association between groups and the contained nodes were not. That last part required additional notifications that signal the beginning and end of opening a graph. The funny thing is that those notifications just send other notifications. It took a lot of digging. I'll do more thorough testing and landscape canvas but the AR tests have passed.

Gems/Atom/Tools/MaterialCanvas/Code/Source/Window/MaterialCanvasGraphView.cpp
```diff
@@ -11,14 +11,51 @@
...
+    void MaterialCanvasGraphConstructPresets::InitializeConstructType(GraphCanvas::ConstructType constructType)
+    {
+        if (constructType == GraphCanvas::ConstructType::NodeGroup)
+        {
+            auto presetBucket = static_cast<GraphCanvas::NodeGroupPresetBucket*>(ModPresetBucket(GraphCanvas::ConstructType::NodeGroup));
+            if (presetBucket)
+            {
+                presetBucket->ClearPresets();
+
+                const AZStd::map<AZStd::string, AZ::Color> defaultPresetGroups = {
```
jeremyong-az, line 23: We're these colors derived from something somewhere, or is there a notion of "theming" that our UI supports at all? I'm wondering if we should considering setting things up so that UI folks can make changes to things like this in data without needing a code check-in in the future. (doesn't have to be this PR, but a general "plan" for how to extract form from function to that end might be useful)

Gems/Atom/Tools/MaterialCanvas/Code/Source/Window/MaterialCanvasGraphView.h
```diff
@@ -12,10 +12,24 @@
...
 {
+    //! The implementation of the graph view requires construct presets in order to be able to create node groups and comment blocks.
+    class MaterialCanvasGraphConstructPresets : public GraphCanvas::EditorConstructPresets
+    {
+    public:
+        AZ_RTTI(MaterialCanvasGraphConstructPresets, "{8E349BC8-1D8B-4A1B-8DE0-FFD61438DBBD}", GraphCanvas::EditorConstructPresets);
+        AZ_CLASS_ALLOCATOR(MaterialCanvasGraphConstructPresets, AZ::SystemAllocator, 0);
+
+        MaterialCanvasGraphConstructPresets() = default;
+        ~MaterialCanvasGraphConstructPresets() = default;
```
jeremyong-az, line 29: I believe that a non-virtual destructor in a derived class won't be called if `delete` is invoked through the class base pointer. This is a stateless class so it doesn't matter here, but usually things like this make me nervous since it might explode someday, perhaps consider adding the `override` postfix keyword?

Gems/GraphCanvas/Code/StaticLib/GraphCanvas/Types/ConstructPresets.cpp
```diff
@@ -509,15 +509,15 @@ namespace GraphCanvas
 
     void EditorConstructPresets::SetDefaultPreset(ConstructType constructType, int presetIndex)
     {
-        ConstructTypePresetBucket* typeBucket = ModPresetBuckket(constructType);
+        ConstructTypePresetBucket* typeBucket = ModPresetBucket(constructType);
```
jeremyong-az, line 512: 😜

jeremyong-az (approved):

Gems/Atom/Tools/MaterialCanvas/Code/Source/Window/MaterialCanvasGraphView.cpp
gadams3: These colors were copied from script canvas. I'm preparing a follow up PR that passes in this list of names and colors and allows the user to configure them from the settings dialog. Either way, these are just the defaults and the user can create new presets based on other groups, comments, nodes in the scene.

Gems/Atom/Tools/MaterialCanvas/Code/Source/Window/MaterialCanvasGraphView.h
gadams3: Adjusting follow up PR

PR #13412, opened 2022-11-28
[Terrain] PaintBrush code reorganization

mbalfour-amzn:
## What does this PR do?

This reorganizes the paintbrush code so that the core paintbrush logic is available for use at runtime. The reorganization also formalizes a split between the core logic and paint brush settings (of which multiple can exist), and the tool-side Global Paint Brush Settings in which we only have one global copy of settings for use within the Editor. We already only had one global copy of tool-side settings, the naming and code split just reinforces this distinction.

This PR will be followed up with a second one that changes the Image Gradient usage of the paintbrush so that it will support using the paintbrush both in the Editor and at runtime through Script Canvas / Lua.

The changes below all boil down to the following:
* Core PaintBrush logic and unit tests moved from AzToolsFramework to AzFramework
* Tool-side global settings logic renamed from "PaintBrushSettings" to "GlobalPaintBrushSettings"
* PaintBrushSettings exposed to Behavior Context

The one slightly awkward piece is that the EditContext reflection for PaintBrushSettings needs to live at the AzToolsFramework level still, because it relies on color configuration for the ColorPicker widget, and the configuration & widget are only defined at the AzToolsFramework level.

## How was this PR tested?

* Verified that the unit tests still run and work correctly.
* Manually verified that painting in the Editor still works correctly.
* Tested the paintbrush code at runtime with local changes that will be in an upcoming PR.


Gems/ScriptCanvas/Assets/TranslationAssets/Classes/PaintBrushSettings.names
```diff
@@ -0,0 +1,366 @@
...
+        {
+            "base": "PaintBrushSettings",
+            "context": "BehaviorClass",
+            "variant": "",
+            "details": {
+                "name": "Paint Brush Settings"
+            },
+            "methods": [
+                {
+                    "base": "GetsmoothingRadius",
```
cgalvan, line 12: Is the casing of `GetsmoothingRadius` and below methods intentional?

cgalvan (approved):

Gems/ScriptCanvas/Assets/TranslationAssets/Classes/PaintBrushSettings.names
mbalfour-amzn: The base values are just set to whatever the "generate translation" in script canvas produces into the file. I manually edited the "name" values (which are the display values) to look correct. I have no idea what the base values are used for, I assume they're just an internal key and don't really matter too much what they look like.

hultonha (approved):

PR #13413, opened 2022-11-28
Jenkinsfile fixes for external repos

amzn-changml:
## What does this PR do?

- Add new behavior to the Jenkinsfile to allow for pulling the bootstrap scripts from o3de to the o3de-extras or project repos
- Some cleanup on LFS behavior
- Validates the workspace availability before the build
- Consolidating the script paths to a single variable

## How was this PR tested?

Tested in a sandbox Jenkins environment.

### DIFF
--- a/scripts/build/Jenkins/Jenkinsfile
+++ b/scripts/build/Jenkins/Jenkinsfile
@@ -10,9 +10,10 @@
 import groovy.json.JsonOutput
 import java.util.regex.Pattern
 
-PIPELINE_CONFIG_FILE = 'scripts/build/Jenkins/o3de.json'
-INCREMENTAL_BUILD_SCRIPT_PATH = 'scripts/build/bootstrap/incremental_build_util.py'
-EBS_SNAPSHOT_SCRIPT_PATH = 'scripts/build/tools/ebs_snapshot.py'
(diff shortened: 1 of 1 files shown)


cgalvan (approved):

scripts/build/Jenkins/Jenkinsfile
```diff
@@ -414,7 +405,13 @@ def PreBuildCommonSteps(Map pipelineConfig, String snapshot, String repositoryNa
         }
     }
 
+    def workspace_retry = 5
```
brianherrera, line 408: Is this issue limited to the workspace directory specifically or do we need to add this retry for other subdirectories created using dir().

gncesar (approved):
Looking good

byrcolin (approved):

scripts/build/Jenkins/Jenkinsfile
amzn-changml: Just the root workspace specifically. It seems the issue is that the EBS volume may not be immediately available for writes once mounted. Once this is available, then the rest of the subfolders can be created.

brianherrera (approved):

PR #13421, opened 2022-11-28
Allow for Optional Gem annotations

amzn-phist:
## What does this PR do?

This allows for gems listed in `project.json`'s `"gem_names"` array to be marked as optional.  This change helps projects that might want to soft-enable external Gems.  You can list the Gems as optional, and if the optional gems aren't found by CMake, configuration will still succeed.  External Gems that are optional can also be listed permanently in the `enabled_gems.cmake` file in the project, they will still retain their optional property.

This is an example of how it might look in `project.json`:
```
...
    "gem_names": [
        {
            "name": "MyExternalGem",
            "optional": true
        }
    ]
...
```

## How was this PR tested?

This was tested against a project (Atom Sample Viewer) with external Gems from o3de-extras repo.
- With the external gems NOT registered to my manifest, configuration succeeds and the project builds successfully, but does not build or load the gems.
- With the external gems registered to my manifest, configuration succeeds and the project builds and loads the gems.

### DIFF
--- a/cmake/Subdirectories.cmake
+++ b/cmake/Subdirectories.cmake
@@ -155,10 +155,17 @@ function(query_gem_paths_from_external_subdirs output_gem_dirs gem_names registe
     if (gem_names)
         foreach(gem_name IN LISTS gem_names)
             unset(gem_path)
+            # Use the ERROR_VARIABLE to catch the common case when it's a simple string and not a json type.
+            string(JSON json_type ERROR_VARIABLE json_error TYPE ${gem_name})
+            set(gem_optional FALSE)
+            if(${json_type} STREQUAL "OBJECT")
+                string(JSON gem_optional GET ${gem_name} "optional")
+                string(JSON gem_name GET ${gem_name} "name")    # overwrite the gem_name variable
+            endif()
             o3de_find_gem_with_registered_external_subdirs(${gem_name} gem_path "${registered_external_subdirs}")
             if (gem_path)
                 list(APPEND gem_dirs ${gem_path})
-            else()
+            elseif(NOT gem_optional)
                 list(JOIN registered_external_subdirs "\n" external_subdirs_formatted)
                 message(SEND_ERROR "The gem \"${gem_name}\""
                 " could not be found in any gem.json from the following list of registered external subdirectories:\n"
--- a/cmake/Subdirectories.cmake
+++ b/cmake/Subdirectories.cmake
@@ -155,13 +155,7 @@ function(query_gem_paths_from_external_subdirs output_gem_dirs gem_names registe
     if (gem_names)
         foreach(gem_name IN LISTS gem_names)
             unset(gem_path)
-            # Use the ERROR_VARIABLE to catch the common case when it's a simple string and not a json type.
-            string(JSON json_type ERROR_VARIABLE json_error TYPE ${gem_name})
-            set(gem_optional FALSE)
-            if(${json_type} STREQUAL "OBJECT")
-                string(JSON gem_optional GET ${gem_name} "optional")
-                string(JSON gem_name GET ${gem_name} "name")    # overwrite the gem_name variable
-            endif()
+            get_property(gem_optional GLOBAL PROPERTY ${gem_name}_OPTIONAL)
             o3de_find_gem_with_registered_external_subdirs(${gem_name} gem_path "${registered_external_subdirs}")
             if (gem_path)
                 list(APPEND gem_dirs ${gem_path})
@@ -259,7 +253,23 @@ endfunction()
 function(get_all_external_subdirectories_for_o3de_object output_subdirs object_type object_name object_path object_json_filename)
     # Append the gems referenced by name from "gem_names" field in the <object>.json
     # These gems are registered in the users o3de_manifest.json
-    o3de_read_json_array(gem_names ${object_path}/${object_json_filename} "gem_names")
+    o3de_read_json_array(initial_gem_names ${object_path}/${object_json_filename} "gem_names")
+    set(gem_names "")
+    foreach(gem_name IN LISTS initial_gem_names)
+        # Use the ERROR_VARIABLE to catch the common case when it's a simple string and not a json type.
+        string(JSON json_type ERROR_VARIABLE json_error TYPE ${gem_name})
+        set(gem_optional FALSE)
+        if(${json_type} STREQUAL "OBJECT")
+            string(JSON gem_optional GET ${gem_name} "optional")
+            string(JSON gem_name GET ${gem_name} "name")
+        endif()
+
+        # Set a global "optional" property on the gem name
+        set_property(GLOBAL PROPERTY "${gem_name}_OPTIONAL" ${gem_optional})
+        # Build the gem_names list with extracted names
+        list(APPEND gem_names ${gem_name})
+    endforeach()
+
     add_registered_gems_to_external_subdirs(object_gem_reference_dirs "${gem_names}")
     list(APPEND subdirs_for_object ${object_gem_reference_dirs})
 
--- a/scripts/o3de/o3de/engine_properties.py
+++ b/scripts/o3de/o3de/engine_properties.py
@@ -33,7 +33,7 @@ def _edit_gem_names(engine_json: dict,
     if delete_gem_names:
         removal_list = delete_gem_names.split() if isinstance(delete_gem_names, str) else delete_gem_names
         if 'gem_names' in engine_json:
-            def in_list(gem: str or list, remove_list: list) -> bool:
+            def in_list(gem: str or dict, remove_list: list) -> bool:
                 if isinstance(gem, dict):
                     return gem['name'] in remove_list
                 else:
--- a/scripts/o3de/o3de/engine_properties.py
+++ b/scripts/o3de/o3de/engine_properties.py
@@ -24,6 +24,15 @@ def _edit_gem_names(engine_json: dict,
                     delete_gem_names: str or list = None,
                     replace_gem_names: str or list = None,
                     is_optional: bool = False):
+    """
+    Edits the current list of gems and returns an updated list.
+    :param engine_json: The project json data
+    :param new_gem_names: Any gem names to be added to the list
+    :param delete_gem_names: Any gem names to be removed from the list
+    :param replace_gem_names: A list of gem names that will completely replace the current list
+    :param is_optional: Only applies to new_gem_names, when true will add an 'optional' property to the gem(s)
+    :return: The modified list of gems (a json blob)
+    """
     if new_gem_names:
         add_list = new_gem_names.split() if isinstance(new_gem_names, str) else new_gem_names
         if is_optional:
@@ -35,7 +44,7 @@ def _edit_gem_names(engine_json: dict,
         if 'gem_names' in engine_json:
             def in_list(gem: str or dict, remove_list: list) -> bool:
                 if isinstance(gem, dict):
-                    return gem['name'] in remove_list
+                    return gem.get('name', '') in remove_list
                 else:
                     return gem in remove_list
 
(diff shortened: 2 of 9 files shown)


cmake/Subdirectories.cmake
```diff
@@ -155,10 +155,17 @@ function(query_gem_paths_from_external_subdirs output_gem_dirs gem_names registe
     if (gem_names)
         foreach(gem_name IN LISTS gem_names)
             unset(gem_path)
+            # Use the ERROR_VARIABLE to catch the common case when it's a simple string and not a json type.
+            string(JSON json_type ERROR_VARIABLE json_error TYPE ${gem_name})
+            set(gem_optional FALSE)
+            if(${json_type} STREQUAL "OBJECT")
+                string(JSON gem_optional GET ${gem_name} "optional")
+                string(JSON gem_name GET ${gem_name} "name")    # overwrite the gem_name variable
+            endif()
```
lumberyard-employee-dm, line 164: We are going to have to move this extraction of the gem name entry that is an object to [get_all_external_subdirectories_for_o3de_object](https://github.com/o3de/o3de/blob/c7d31e7562c5156e11b96782a9eab931f8fcb5d7/cmake/Subdirectories.cmake#L260-L262) function below, because the `gem_names` array is iterated in that function and appended to a CMAKE property name.

What we could do there is after the `o3de_read_json_array` call to read the each gem_name field from the project.json, we can then create a "mapping" in CMake using a property that contains the gem_name in it to map to the optional value,

i.e
```cmake
o3de_read_json_array(gem_names ${object_path}/${object_json_filename} "gem_names")
foreach(gem_name IN LISTS gem_names)
    # Use the ERROR_VARIABLE to catch the common case when it's a simple string and not a json type.
    string(JSON json_type ERROR_VARIABLE json_error TYPE ${gem_name})
    set(gem_optional FALSE)
    if(${json_type} STREQUAL "OBJECT")
        string(JSON gem_optional GET ${gem_name} "optional")
        string(JSON gem_name GET ${gem_name} "name")    # overwrite the gem_name variable
    endif()

    # Set a global cmake property that stores a mapping of gem_name to optional value
    set_property(GLOBAL PROPERTY "${gem_name}_OPTIONAL" ${gem_optional})
endforeach()
```

Then in this `query_gem_paths_from_external_subdirs` function, we can just query the optional state using get_property
```cmake
    get_property(gem_optional GLOBAL PROPERTY ${gem_name}_OPTIONAL)
```

scripts/o3de/o3de/engine_properties.py
```diff
@@ -18,25 +18,35 @@
...
+                if isinstance(gem, dict):
+                    return gem['name'] in remove_list
+                else:
+                    return gem in remove_list
+
+            engine_json['gem_names'] = [gem for gem in engine_json['gem_names'] if not in_list(gem, removal_list)]
+
     if replace_gem_names:
         tag_list = replace_gem_names.split() if isinstance(replace_gem_names, str) else replace_gem_names
         engine_json['gem_names'] = tag_list
```
lumberyard-employee-dm, line 47: We also need to support the `is_optional` field in the replace_gem_names list as well

scripts/o3de/o3de/utils.py
```diff
@@ -319,3 +323,29 @@ def find_ancestor_dir_containing_file(target_file_name: pathlib.PurePath, start_
...
+    Adds names to a dict, and when a collision occurs, eject the existing one in favor of the new one.
+    This is because when adding gems the list is extended, so the override will come last.
+    :param gems: The original list of gems, strings or small dicts (json objects)
+    :return: A new list with duplicate gem entries removed
+    """
+    new_list = []
+    names = {}
+    for gem in gems:
+        if isinstance(gem, dict):
+            if gem['name'] not in names:
```
lumberyard-employee-dm, line 340: It is safer to use `gem.get('name', '')` just in case the the user has a project.json with a "gem_names" object that doesn't have a "names" field in it.
i.e
```json
"gem_names": [
    "MyGem",
    {
        "name": "YourGem":
    },
    {
        "name_field_mismatch": "BadFieldGem":
    },
]
```
In that case the python code would raise an exception.

scripts/o3de/tests/test_utils.py
```diff
@@ -37,3 +39,16 @@ def test_validate_identifier(value, expected_result):
...
     result = utils.validate_uuid4(value)
     assert result == expected_result
+
+
+@pytest.mark.parametrize(
+    "in_list, out_list", [
+        pytest.param(['A', 'B', 'C'], ['A', 'B', 'C']),
+        pytest.param(['A', 'B', 'C', 'A', 'C'], ['A', 'B', 'C']),
+        pytest.param(['A', {'name': 'A', 'optional': True}], [{'name': 'A', 'optional': True}]),
+        pytest.param([{'name': 'A', 'optional': True}, 'A'], ['A']),
```
lumberyard-employee-dm, line 49: Can you add a test parameters for when the optional field is missing or for when the name field is missing as well?

```suggestion
        pytest.param(['A', 'B', 'C'], ['A', 'B', 'C']),
        pytest.param(['A', 'B', 'C', 'A', 'C'], ['A', 'B', 'C']),
        pytest.param(['A', {'name': 'A', 'optional': True}], [{'name': 'A', 'optional': True}]),
        pytest.param([{'name': 'A', 'optional': True}, 'A'], ['A']),
        pytest.param([{'name': 'A'}], [{'name': 'A'}]),
        pytest.param([{'optional': False}], []),
```

scripts/o3de/o3de/engine_properties.py
amzn-phist: I was thinking that it wouldn't be common to make all the gems optional if replacing the list, which is what this would do.  I'd expect users to replace the list of gems then later do `enable-gem --optional ...` to a smaller set of gems.  What do you think?

scripts/o3de/o3de/utils.py
amzn-phist: Alright, I can do that.  It would address the added test you recommended below.

scripts/o3de/o3de/engine_properties.py
lumberyard-employee-dm: I don't think it would be common, but it would be surprising to the user that the `--optional` doesn't work when replacing a list of gems though

cmake/Subdirectories.cmake
amzn-phist: Addressed with latest commit.

scripts/o3de/o3de/engine_properties.py
amzn-phist: I just looked.  There's no code path that gets there from `enable-gem` sub-command, which is where the `--optional` parameter is added.  The code that it executes is:

```
        ret_val = project_properties.edit_project_props(project_path, new_gem_names=gem_json_data['gem_name'],
                                                                                    is_optional=optional)
```

Enable-gem doesn't have a call to `edit_project_props` that sets `replace_gem_names`, and the `--optional` flag isn't attached to any other user-facing sub-command.

amzn-phist: I can update both `_edit_gem_names` functions (in project_properties and engine_properties) and add docstring documentation.  That seems like the best place for this information since this function is very internal.

AMZN-alexpete (commented):
Cool to see support for optional gems!  One thing I'd like your thoughts on are how this will affect the plans for versioning - namely the `gem_name` field was planned on being renamed as `gem_dependencies` and contain versioning info as part of [versioning all the things (tm)](https://github.com/o3de/sig-core/issues/44).  With this new `gem_names` format we can add a "version" field, but it is no longer backwards compatible with the old gem.json format nor would it have the nice similarities to python requirements files. 

We'd be moving from using this style in `engine.json`, `project.json` and `gem.json`
```
   "gem_dependencies": [
        "GemA == 1.2",
        "GemB ~= 1.0"
    ],
```

to having `project.json` have a different format like this:
```
    "gem_dependencies": [
        {
            "name": "GemA",
            "optional": false,
            "version":"==1.2"
        },
        {
            "name": "GemB",
            "optional": true,
            "version":"~=1.0"
        },     
    ]
```
I suppose we might go with a different name since they're not necessarily dependencies.  Although, I wonder if there is a version specifier we could use to represent optional i.e. `GemA===optional-1.2` or `GemA?=1.2`

Another option would be to use a second field - one for gems dependencies and one for optional gems, I'm guessing you already considered this and have some thoughts on it.

scripts/o3de/o3de/project_properties.py
```diff
@@ -31,22 +31,31 @@ def get_project_props(name: str = None, path: pathlib.Path = None) -> dict:
 def _edit_gem_names(proj_json: dict,
                     new_gem_names: str or list = None,
                     delete_gem_names: str or list = None,
-                    replace_gem_names: str or list = None):
+                    replace_gem_names: str or list = None,
+                    is_optional: bool = False):
```
AMZN-alexpete, line 35: Did you test activating/deactivating Gems in the Project Manager after making this change? Would be good as a sanity check to make sure nothing is broken there.

amzn-phist: Good idea, I'll try it and update when I inevitably find an error ;)

amzn-phist:
@AMZN-alexpete Good comments!  I didn't mean to step on toes with this design.  As you stated, another option is to make a list on the side like `"optional_gems": [ "GemA", "GemB" ]`, then we could have the flat(ter) list like you posted with the `<name><relation><version>` format.  Either of the two forms you posted would work, or a hybrid.  If you're decorating the "name" string with the version info, couldn't it also be like this?

```
"gem_dependencies": [
    {
        "name": "GemA==1.2",
        "optional": true,
        "futureMetaData": "someValue"
        ...
    }
]
```

Hopefully this change isn't too pervasive and it's a two-way door decision.  Could easily be reworked into another format when the versioning data starts to come together.

scripts/o3de/o3de/engine_properties.py
lumberyard-employee-dm: But the user can invoke the `edit-project-properties` and `edit-engine-properties` command without going through enable-gem

amzn-phist:
Conducting testing with Project Manager, disabling and enabling Gems, and may have ran into a general issue.  It reproduces for me with non-external, non-optional Gems as well.  Investigating...

scripts/o3de/o3de/project_properties.py
amzn-phist: I ran into an issue that I commented about below, but other than that, enabling/disabling is working.

scripts/o3de/o3de/engine_properties.py
amzn-phist: The `--optional` flag is not exposed to the `edit-project-properties` or `edit-engine-properties` sub-commands.  Do you want these to be added?

lumberyard-employee-dm: It doesn't need to be now.
If at some point in the future we want to expose it we can do so then.

lumberyard-employee-dm (approved):

amzn-phist:
Bump

AMZN-alexpete:
Have you tested if the o3de CLI can add/remove gems from a project.json with the old format after this change? Does it update the `gem_names` format from old to new? I'm also noticing it looks like the changes affect `engine.json` also - are there any similar issues if an `engine.json` has the old format?

amzn-phist:
I've tested with CLI.  The old format (str) is still valid and will be used by default.  Only with `enable-gem` sub-command with `--optional` flag will it update an entry with the new format.  If a gem is not optional, it will write the str name only.  I changed the code that writes to engine.json only for consistency, but there's still no `--optional` flag to pass from CLI which will write that file.  Since those two functions are so similar, I could consolidate it and move that to the utils.py file if you want.

AMZN-alexpete (approved):

PR #13433, opened 2022-11-29
Ensure camera inputs are cleared when the viewport loses focus

hultonha:
Signed-off-by: Tom Hulton-Harrop <82228511+hultonha@users.noreply.github.com>

## What does this PR do?

A bug was discovered where if moving the camera using the WASD (or ED) keys, if the user clicked outside of the viewport while this was happening (keys were held) the input would get 'stuck' on as the key up event would be missed as the viewport was no longer in focus.

To avoid this we now detect when the viewport for the camera receiving the input loses focus and reset all camera inputs to ensure nothing is still running.

## How was this PR tested?

Both manually in the Editor and also a new test was created (`CameraSystemStopsMovingWhenViewportLosesFocus`)

### Before

```
[ RUN      ] ModularViewportCameraControllerFixture.CameraSystemStopsMovingWhenViewportLosesFocus
D:/o3de/Code/Editor/Lib/Tests/test_ModularViewportCameraController.cpp(531): error: Value of: lastCameraTranslation
Expected: is close (X: -10, Y: 0, Z: 0)
  Actual: (X: -20, Y: 0, Z: 0) (of type class AZ::Vector3)
[  FAILED  ] ModularViewportCameraControllerFixture.CameraSystemStopsMovingWhenViewportLosesFocus (4 ms)
```

### After

```
[ RUN      ] ModularViewportCameraControllerFixture.CameraSystemStopsMovingWhenViewportLosesFocus
[       OK ] ModularViewportCameraControllerFixture.CameraSystemStopsMovingWhenViewportLosesFocus (118 ms)
```

### DIFF
--- a/Code/Editor/EditorModularViewportCameraComposerBus.h
+++ b/Code/Editor/EditorModularViewportCameraComposerBus.h
@@ -27,5 +27,5 @@ namespace SandboxEditor
     };
 
     using EditorModularViewportCameraComposerNotificationBus =
-        AZ::EBus<EditorModularViewportCameraComposerNotifications, AzToolsFramework::ViewportInteraction::ViewportEBusTraits>;
+        AZ::EBus<EditorModularViewportCameraComposerNotifications, AzToolsFramework::ViewportInteraction::ViewportNotificationsEBusTraits>;
 } // namespace SandboxEditor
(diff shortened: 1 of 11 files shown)


hultonha:
> LGTM, although curious as to the move away from `unique_ptr` to raw pointers?

Ah this is because Qt manages the lifetimes of widgets in an interesting way. When you assign a widget as a parent, it will then be responsible for the lifetime of the widget and will clean it up. If we keep using unique pointers then they both try and delete it which can cause problems, so I decided to just new and delete the root widget and have the other widget be cleaned up by it.

cgalvan (approved):

PR #13440, opened 2022-11-29
DPE: Handle entity change undo events in ComponentAdapter

amzn-tmryan:
**Description**
Before this PR, changes made internally to a prefab weren't being propagated to other instances of the same prefab in a scene when the DPE EntityInspector was being used.

This PR adds handler code to the DPEComponentAdapter similar to what is done in the RPE's EntityPropertyEditor for creation and tracking of undo nodes. This allows the necessary events to reach the prefab system interface, triggering the propagation of those changes to other instances of the same prefab.

**Testing**
- Verified behavior with the RPE enabled seems consistent with DPE-enabled behavior in the EntityInspector regarding prefab change propagation and undoing those changes

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DPEComponentAdapter.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DPEComponentAdapter.cpp
@@ -7,6 +7,7 @@
  */
 
 #include <AzToolsFramework/UI/DocumentPropertyEditor/DPEComponentAdapter.h>
+
 #include <QtCore/QTimer>
 
 namespace AZ::DocumentPropertyEditor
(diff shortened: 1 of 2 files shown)


cgalvan (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DPEComponentAdapter.cpp
```diff
@@ -87,4 +88,54 @@ namespace AZ::DocumentPropertyEditor
...
+    {
+        auto handlePropertyEditorChanged = [&]([[maybe_unused]] const Dom::Value& valueFromEditor, Nodes::ValueChangeType changeType)
+        {
+            switch (changeType)
+            {
+            case Nodes::ValueChangeType::InProgressEdit:
+                if (m_componentInstance)
+                {
+                    const AZ::EntityId& entityId = m_componentInstance->GetEntityId();
+                    if (entityId.IsValid())
```
srikappa-amzn, line 101: Is this even possible?

srikappa-amzn (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DPEComponentAdapter.cpp
amzn-tmryan: Do you mean is it possible for the EntityId to be invalid? If so, I'm not entirely sure but it seemed like a good thing to check.

amzn-tmryan: Actually, I see in EntityId Component::GetEntityId() const that it will return the InvalidEntityId if its m_entity is nulltpr.

srikappa-amzn: Yes, I'm sure it can be invalid logically. I'm just wondering whether it is possible for a component adapter's component to have an invalid entity id once it is created. It just seems like an impossible scenario. But the logic can stay as an additional guardrail

dtamkin1 (approved):
Looks good!

alexmontAmazon (approved):

PR #13448, opened 2022-11-29
Material Canvas: Moving graph view settings to their own class and adding to settings dialog

gadams3:
## What does this PR do?

•	Renamed graph view config to graph view settings
•	Move graph view settings from the graph view class into the new settings class
•	Added serialization and reflection for all of the graph view settings
•	Updated the atom tools framework settings dialog so that it’s not tied directly to the dynamic property classes, the internal inspector is exposed, and any kind of data And widgets can be added to its inspector.
•	Broke apart the application class initialization into multiple functions

## How was this PR tested?

These changes are currently isolated to material canvas. 
Confirmed that pre-existing settings continue to function.
Confirmed that graph view settings changes take effect immediately. However, those that affect the visual appearance of the graph will require the graph to be refreshed.
Confirmed that graph view settings are loaded from the settings registry on startup.
Confirmed that graph your settings are saved to the settings registry on shutdown.


Gems/Atom/Tools/AtomToolsFramework/Code/Include/AtomToolsFramework/GraphView/GraphViewConstructPresets.h
```diff
@@ -0,0 +1,34 @@
...
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#pragma once
+
+#if !defined(Q_MOC_RUN)
+#include <GraphCanvas/Types/ConstructPresets.h>
```
jeremyong-az, line 12: Should the `Q_MOC_RUN` guard be placed in `ConstructPresets.h` instead to protect against misuse for other would-be users of that header?

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Document/AtomToolsDocumentMainWindow.cpp
```diff
@@ -325,45 +325,50 @@ namespace AtomToolsFramework
...
+              CreateSettingsPropertyValue(
+                  "/O3DE/AtomToolsFramework/AtomToolsDocumentSystem/AutoSaveEnabled",
+                  "Enable Auto Save",
+                  "Automatically save documents after they are modified",
+                  false),
+              CreateSettingsPropertyValue(
+                  "/O3DE/AtomToolsFramework/AtomToolsDocumentSystem/AutoSaveInterval",
+                  "Auto Save Interval",
+                  "How often (in milliseconds) auto save occurs",
+                  aznumeric_cast<AZ::s64>(250)) });
```
jeremyong-az, line 364: If you want, I think `250ll` expresses the same thing

Gems/Atom/Tools/AtomToolsFramework/Code/Source/GraphView/GraphView.cpp
```diff
@@ -47,11 +47,11 @@ namespace AtomToolsFramework
     GraphView::GraphView(
         const AZ::Crc32& toolId,
         const GraphCanvas::GraphId& activeGraphId,
-        const GraphViewConfig& graphViewConfig,
+        AZStd::shared_ptr<GraphViewSettings> graphViewSettingsPtr,
         QWidget* parent)
         : QWidget(parent)
         , m_toolId(toolId)
-        , m_graphViewConfig(graphViewConfig)
+        , m_graphViewSettingsPtr(graphViewSettingsPtr)
```
jeremyong-az, line 54: Since you've already made a copy of the shared pointer, it's safe to `move` it here (a modest efficiency gain).

Gems/Atom/Tools/MaterialCanvas/Code/Source/Window/MaterialCanvasGraphView.cpp
```diff
@@ -11,51 +11,14 @@
...
-            }
-        }
-    }
-
     MaterialCanvasGraphView::MaterialCanvasGraphView(
-        const AZ::Crc32& toolId, const AZ::Uuid& documentId, const AtomToolsFramework::GraphViewConfig& graphViewConfig, QWidget* parent)
-        : AtomToolsFramework::GraphView(toolId, GraphCanvas::GraphId(), graphViewConfig, parent)
+        const AZ::Crc32& toolId,
+        const AZ::Uuid& documentId,
+        const AZStd::shared_ptr<AtomToolsFramework::GraphViewSettings>& graphViewSettingsPtr,
```
jeremyong-az, line 17: If the intention is to pass `GraphViewSettings` without sharing or granting ownership, I suggest accepting a `const GraphViewSettings&` reference instead, or if this is a nullable parameter, just `const GraphViewSettings`. If however, this constructor intends to share ownership, passing `shared_ptr<GraphViewSettings>` by value is preferred, or if ownership is to be taken, an r-value `shared_ptr<GraphViewSettings>&&` instead. Finally, passing the shared pointer by reference really only makes sense if the function intends to manipulate the ownership of the caller (e.g. cause the caller to release its reference) or query the state of the shared pointer itself.

jeremyong-az (approved):

Gems/Atom/Tools/MaterialCanvas/Code/Source/Window/MaterialCanvasGraphView.cpp
gadams3: Based on this feedback, passing by value sounds like the most appropriate option. Ownership is being shared to keep the instance alive and connected to the bus without having to use a weak pointer or check for validity. Most of the graph view settings members that were added should be accessible through the bus interface. Only the handful of members that were originally in the graph view config are excluded from that interface. I could make all of the members private and flesh out the interface for the additional members.

cgalvan (approved):

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Document/AtomToolsDocumentMainWindow.cpp
gadams3: s64 is required by the settings registry and defined differently on different platforms. Probably something that should be brought to the attention of core if there's a distinction between the types.

PR #13471, opened 2022-11-30
Fixed Linux file watch issue with AP messaging test

AMZN-stankowi:
## What does this PR do?

Before this change: This test required the asset cache folder for the active project to exist and have an asset database in it. As a side effect of this, it added every file in the cache to the file watch, because it was booting up so much of the core asset processor logic. The result was, on Linux, this test would fail if the maximum number of file watches available was too low.

With this change: Less of the asset processor is initialized to run this test, while still keeping enough functional for all messages under test to function correctly. The main message test that fails without the stubbed out file watcher is addPairFunc(new RequestAssetStatus(), new ResponseAssetStatus());

## How was this PR tested?

Ran locally on Windows + Linux. Ran with and without a populated asset cache folder. Ran on Linux with a low file watcher count that failed on the old version of this test, to verify the new one works.


amzn-mike (approved):

Code/Tools/AssetProcessor/native/tests/AssetProcessorMessagesTests.cpp
```diff
@@ -71,6 +85,22 @@ namespace AssetProcessorMessagesTests
...
+                AZ::Utils::LoadObjectFromBuffer<AzFramework::AssetSystem::BaseAssetProcessorMessage>(
+                    payload.constData(), payload.size(), serializeContext)
+            };
+            NetworkRequestID key(connId, serial);
+            int fenceFileId = 0;
+            m_pendingFenceRequestMap[fenceFileId] = AZStd::move(AssetRequestHandler::RequestInfo(key, AZStd::move(message), platform));
+
+            OnFenceFileDetected(fenceFileId);
+        }
+
```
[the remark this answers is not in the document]
AMZN-stankowi: Requires fencing is actually set to false on that message, but the message was failing to second in this test if the fencing didn't occur.

Code/Tools/AssetProcessor/native/AssetManager/AssetRequestHandler.h
```diff
@@ -249,7 +249,7 @@ namespace AssetProcessor
         // Invokes the appropriate handler and returns true if the message should be deleted by the caller and false if the request handler is responsible for deleting the message
         virtual bool InvokeHandler(MessageData<AzFramework::AssetSystem::BaseAssetProcessorMessage> message);
 
-    private:
+    protected:
```
lmbr-pip, line 252: Seems dangerous to make all these methods protected. Do all of them need to have weaker access control? You also have made class members vars protected here, which really breaks encapsulation.

Class also looks strange with two large protected blocks, separated by one public method

Code/Tools/AssetProcessor/assetprocessor_static_files.cmake
```diff
@@ -50,6 +50,7 @@ set(FILES
     native/FileProcessor/FileProcessor.h
     native/FileWatcher/FileWatcher.cpp
     native/FileWatcher/FileWatcher.h
+    native/FileWatcher/FileWatcherBase.h
```
burelc-amzn, line 53: If you intend to merge this PR with a merge commit, and not a squash, then this commit isn't really correct. `FileWatcherBase.h` was added in 75aaa3c5fd78f5cc015a83e76cc7ff9c84015ded, but not added to the `assetprocessor_static_files.cmake` until this later commit.

Code/Tools/AssetProcessor/native/AssetManager/AssetRequestHandler.h
AMZN-stankowi: I've updated this to move just what I needed to protected.

Code/Tools/AssetProcessor/native/unittests/FileWatcherUnitTests.cpp
```diff
@@ -6,6 +6,7 @@
  *
  */
 
+#include <native/FileWatcher/FileWatcher.h>
```
junbo75, line 9: Just curious: Any reason to include this new header when there's no other change in this test file?

junbo75 (approved):

Code/Tools/AssetProcessor/native/unittests/FileWatcherUnitTests.cpp
AMZN-stankowi: Before it was getting FileWatcher.h included because it included ApplicationManagerBase.h which included FileWatcher.h.

Now, ApplicationManagerBase.h includes FileWatcherBase.h, so FileWatcher.h isn't included automatically.

FileWatcherUnitTests.cpp is instantiating a FileWatcher (not base) here, so it needs the include:
https://github.com/o3de/o3de/blob/development/Code/Tools/AssetProcessor/native/unittests/FileWatcherUnitTests.cpp#L22

PR #13477, opened 2022-12-01
Replace redundant `AZ::Debug::DebugAllocator` with `AZStd::stateless_allocator`

burelc-amzn:
These two allocators serve the same purpose, the former can be removed in favor of the latter.


cgalvan (approved):

Code/Framework/AzCore/AzCore/Memory/PoolAllocator.cpp
```diff
@@ -29,6 +28,29 @@
...
 #define POOL_ALLOCATION_MAX_ALLOCATION_SIZE size_t{512}
 
+namespace
+{
+    struct Magic32
+    {
+        static const AZ::u32 m_defValue = 0xfeedf00d;
+        AZ_FORCE_INLINE Magic32()
+        {
+            m_value = (m_defValue ^ (AZ::u32)((AZStd::size_t)this));
```
lumberyard-employee-dm, line 38: The `this` pointer should be cast using `uintptr_t` as that is the appropriate type for casting between pointer and integer types.
`size_t` is not an appropriate type.
```suggestion
            m_value = (m_defValue ^ AZ::u32(reinterpret_cast<AZStd::uintptr_t>(this));
```

PR #13480, opened 2022-12-01
add AzQtComponents::InputDialog to allow validator to be used.

### LINKED ISSUE #11688: Editor Layout: Saving a custom layout with a "/" in the name causes issues
**Describe the bug**
Saving a custom layout with a "/" in the name causes issues. If text is added after the "/", the layout will simply not save. If a layout is saved with a name like "Test/", it will save properly, but will be unable to be Deleted.

**Steps to reproduce**
1. Launch the Editor.
2. Resize a few panes/open a few tools to make a custom layout.
3. Click View > Layouts > Save Layouts and name the new layout "Test/Test"
4. Attempt to reload the newly saved layout.
5. Click View > Layouts > Save Layouts and name the new layout "Test/"
6. Attempt to Delete the newly saved layout.

**Expected behavior**
Layouts save without issue with special characters, or alert the user that the name can't be used.

**Actual behavior**
Custom layouts aren't saved if text is added after a "/" in the layout name, and cannot be deleted if "/" is used with text after it.

**Found in Branch**
stabilization/2210@ d3760da0ab375a21d67a3be5ecefbedad2440d27

**Desktop/Device:**
 - Device: PC
 - OS: Windows 10 21H1
 - CPU: Ryzen 5950x
 - GPU: Nvidia 2070
 - Memory: 64GB

sphrose:

## What does this PR do?

This PR fixes #11688. It adds an AzQtComponents::InputDialog class which expands QInputDialog by allowing a text validator to be added to the text input. It then used this to prevent the user from adding illegal characters, e.g. '/' to layout filenames.

It also fixes a discrepancy between naming and renaming a layout, where naming forced the name to lower case, whereas renaming did not. Layout names are now always lower case.

## How was this PR tested?

Testing performed manually.

### DIFF
--- a/Code/Framework/AzQtComponents/AzQtComponents/Components/InputDialog.cpp
+++ b/Code/Framework/AzQtComponents/AzQtComponents/Components/InputDialog.cpp
@@ -32,6 +32,8 @@ namespace AzQtComponents
 
     void InputDialog::setValidator(QValidator* validator)
     {
+        m_validator = validator;
+
         AttemptAssignValidator();
     }
 
--- a/Code/Framework/AzQtComponents/AzQtComponents/Components/InputDialog.cpp
+++ b/Code/Framework/AzQtComponents/AzQtComponents/Components/InputDialog.cpp
@@ -20,11 +20,9 @@ namespace AzQtComponents
     {
     }
 
-    void InputDialog::setValidator(QValidator* validator)
+    void InputDialog::AttemptAssignValidator()
     {
-        m_validator = validator;
-
-        // If this is called after show has been called, the QLineEdit will be available to patch.
+        // This will succeed only if show() has happened, which sets up the UI.
         QLineEdit* lineEdit = this->findChild<QLineEdit*>();
         if (lineEdit)
         {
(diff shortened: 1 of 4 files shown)


Code/Framework/AzQtComponents/AzQtComponents/Components/InputDialog.cpp
```diff
@@ -0,0 +1,89 @@
...
+    void InputDialog::setValidator(QValidator* validator)
+    {
+        m_validator = validator;
+
+        // If this is called after show has been called, the QLineEdit will be available to patch.
+        QLineEdit* lineEdit = this->findChild<QLineEdit*>();
+        if (lineEdit)
+        {
+            lineEdit->setValidator(m_validator);
+        }
```
cgalvan, line 32: Nit: Could move this logic to a private helper method since we do the same logic on `show()`

cgalvan (approved):
Thanks for doing this!

Code/Framework/AzQtComponents/AzQtComponents/Components/InputDialog.cpp
sphrose: That's done. I've also removed the assignment being conditional on the validator existing to allow the user to remove a validator if they want to.

gadams3 (approved):

PR #13483, opened 2022-12-01
BugFix - Move back override subtree in prefab override revert redo

forkercat:

## What does this PR do?

This PR adds the missing "move operation" in prefab override revert undo node. Without moving back, the subtree stored in the undo node will become empty in subsequent undo/redo operations.

Before this fix, issue occurs when hitting "revert", undo, redo, and undo. In redo, the override subtree is not populated correctly, then in the second undo, the override subtree is empty.

## How was this PR tested?

1. Verified in editor
2. Update existing unit test to cover this fix. Verified that it failed before the fix and passed after the fix.


srikappa-amzn (approved):
Good catch!

michabr (approved):

PR #13490, opened 2022-12-01
EMFX: AnimGraph profiler to mainline

amzn-rhhong:
## What does this PR do?

Merge the Anim Graph profiler code to mainline.
Code is already reviewed in the animation/animgraphprofiler branch. Feedback are still welcomed.

https://user-images.githubusercontent.com/69218254/205183390-7bed1611-45a4-4d99-bc6f-473d6e269455.mp4


ParasiteSt:
@amzn-rhhong This is more of a UI question, is there a better location for the debug information (maybe right above or below the node)? I like seeing it with each node to know if goes with that said node, but it seems like the debug info really overlays all other information on the node, which would end up making it hard to know what nodes debug info you are looking at if the graph gets complex.

hultonha:
> @amzn-rhhong This is more of a UI question, is there a better location for the debug information (maybe right above or below the node)? I like seeing it with each node to know it goes with that said node, but it seems like the debug info really overlays all other information on the node, which would end up making it hard to know what nodes debug info you are looking at if the graph gets complex.

Good suggestion @ParasiteSt but might we be able to do this as a fast-follow PR to get this functionality in and then make the UX change as a separate polish PR after? Thanks for the idea, it would be good to loop in @o3de/sig-ui-ux for their thoughts too.

hultonha:
@amzn-rhhong It looks like DCO is failing, you may need to follow the new DCO verification steps to get it resolved

Gems/EMotionFX/Code/EMotionFX/Source/AnimGraphNode.cpp
```diff
@@ -2537,4 +2571,53 @@ namespace EMotionFX
...
+        , m_animGraphInstance(animGraphInstance)
+        , m_profileMode(mode)
+        , m_isIncomingNode(incomingNode)
+        , m_startPoint(AZStd::chrono::system_clock::now())
+    {
+    }
+
+    AnimGraphNode::ProfileSection::~ProfileSection()
+    {
+        if (m_node->GetProfileMode() & static_cast<AZ::u8>(m_profileMode))
```
hultonha, line 2606: uber nice: It's good to be explicit here and make this a bool check by doing...

```
if ((m_node->GetProfileMode() & static_cast<AZ::u8>(m_profileMode)) != 0)
```

Gems/EMotionFX/Code/EMotionFX/Source/AnimGraphNode.cpp
```diff
@@ -2537,4 +2571,53 @@ namespace EMotionFX
...
+        {
+            AZStd::chrono::nanoseconds duration = AZStd::chrono::system_clock::now() - m_startPoint;
+            if (AnimGraphNodeData* pUniqueData = m_node->FindOrCreateUniqueNodeData(m_animGraphInstance))
```
hultonha, line 2609: nit/standards: Is it possible to do a pass on the naming to remove the hungarian notation we don't usually use (so this just becomes `uniqueData` (not `pUniqueData`)

Gems/EMotionFX/Code/EMotionFX/Source/AnimGraphNode.h
```diff
@@ -943,7 +944,46 @@ namespace EMotionFX
...
+        {
+            Update = 1 << 0,
+            TopDown = 1 << 1,
+            PostUpdate = 1 << 2,
+            Output = 1 << 3
+        };
+
+        AZ::u8 m_profileMode = 0;
+        virtual void SetProfileMode(AZ::u8 profileMode) { m_profileMode = profileMode; };
+        virtual AZ::u8 GetProfileMode() { return m_profileMode; };
```
hultonha, line 958: nit: This can be `const`

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/BlendGraphViewWidget.cpp
```diff
@@ -565,6 +594,55 @@ namespace EMStudio
...
+        case VISUALIZATION_PROFILING_TOPDOWN:
+            displayFlags = AnimGraphPlugin::DISPLAYFLAG_PROFILING_TOPDOWN;
+            break;
+        case VISUALIZATION_PROFILING_POSTUPDATE:
+            displayFlags = AnimGraphPlugin::DISPLAYFLAG_PROFILING_POSTUPDATE;
+            break;
+        case VISUALIZATION_PROFILING_OUTPUT:
+            displayFlags = AnimGraphPlugin::DISPLAYFLAG_PROFILING_OUTPUT;
+            break;
+        default:
```
hultonha, line 626: suggestion: minor defensive programming but it's good form to add a `break` after the `default` case just in case someone adds something after this in future - not 100% necessary but it's good hygiene

ParasiteSt:
@hultonha that is a great idea as well. This functionality should be a priority over ux changes. Also having thoughts from the sig-ui-ux would be the best approach to make sure it is a comfortable change that most developers will like, and not made for one.

hultonha (approved):
Overall change looks great, I just have a few minor suggestions and would like to follow-up on the best usage of the new define just to make it consistent and easy to disable for release builds

Gems/EMotionFX/Code/CMakeLists.txt
```diff
@@ -40,6 +40,7 @@ ly_add_target(
     COMPILE_DEFINITIONS
         PUBLIC
             EMFX_DEVELOPMENT_BUILD
+            EMFX_ANIMGRAPH_PROFILER_ENABLED
```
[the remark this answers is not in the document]
amzn-rhhong: I thought about that as well. But I personally think the EMFX_DEVELOPMENT_BUILD define is too general. For example, user could want to still use the analysisor code to observe the performance of a non-development build.

Gems/EMotionFX/Code/EMotionFX/Source/AnimGraphNode.cpp
amzn-rhhong: good call

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/BlendGraphViewWidget.cpp
```diff
@@ -565,6 +594,55 @@ namespace EMStudio
...
     }
 
+#if defined(EMFX_ANIMGRAPH_PROFILER_ENABLED)
+    void BlendGraphViewWidget::AddProfilingAction(const char* actionName, EOptionFlag optionFlag)
+    {
+        m_actions[optionFlag] = new QAction(
+            tr(actionName),
+            this);
+        m_actions[optionFlag]->setCheckable(true);
+        connect(m_actions[optionFlag], &QAction::triggered, this, [=]([[maybe_unused]] bool checked) { OnDisplayProfiling(optionFlag); });
```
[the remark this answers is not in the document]
amzn-rhhong: good call

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/BlendGraphViewWidget.cpp
amzn-rhhong: who would be that insane to add more condition code after default?

amzn-rhhong:
@ParasiteSt In the blend tree view it doesn't block the name of the node
image
The problem is in the state machine view i think. I can try to move it down a few pixel.

AMZN-alexpete (approved):
Awesome :smile: my only suggestion - likely for a future changelist- would be to support changing the display mode to show some average over a time period (1sec), weighted average or max or something in case the numbers in the display are changing so fast it's hard to read.

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/AnimGraph/BlendGraphViewWidget.cpp
hultonha: You never know 😉

Gems/EMotionFX/Code/CMakeLists.txt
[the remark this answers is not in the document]
hultonha: What would be your suggestion for the best way to fix this @lumberyard-employee-dm? We'd essentially like to have this on/enabled by default in Debug and Profile configurations but disabled in Release. It would be possible for users to set it on in Release if they really wanted it, but it shouldn't be on by default. I believe @lawsonamzn suggested a generator expression might be the best way to handle this? Something like...

```
$<$<NOT:$<CONFIG:release>>:EMFX_ANIMGRAPH_PROFILER_ENABLED>
```

PR #13493, opened 2022-12-02
Material Canvas: Separated MaterialCanvasDocument into MaterialGraphCompiler and a configurable GraphDocument

gadams3:
## What does this PR do?

This change splits the material canvas document class into two distinct pieces. The first piece is a general purpose, configurable graph document class for integrating with graph model. That was moved to atom tools framework for now. The second part is the material graph compiler, which transforms the graph into all of the generated shader and material assets. It was almost an even split across the middle of the original file.

In addition to that, all of the “experimental” labels were changed to “preview”.

Lastly, a notification was added to the RHI factory manager so material canvas could be informed when the RHI factory was set up before assets load.

## How was this PR tested?

Confirmed that the new graph document and view classes can load, display, and edit graphs.
Confirmed that material canvas graphs continue to be converted into the expected materials and shaders.


galibzon (approved):

Gems/Atom/Tools/MaterialCanvas/Code/Source/Window/MaterialCanvasMainWindow.h
```diff
@@ -27,7 +27,7 @@
 
 namespace MaterialCanvas
 {
-    //! MaterialCanvasMainWindow
+    //! MaterialCanvasMainWindow Creates and manages all of the craft canvas and viewport related docked windows for Material Canvas.
```
santorac, line 30:
```suggestion
    //! MaterialCanvasMainWindow creates and manages all of the graph canvas and viewport related docked windows for Material Canvas. 
```

santorac (approved):

cgalvan (approved):

PR #13515, opened 2022-12-02
Fix entity selection when parent instance is the root

### LINKED ISSUE #13415: Prefabs: With overrides enabled, selecting an entity in the Viewport selects its root entity
**Describe the bug**
With overrides enabled, selecting an entity in the Viewport selects its root entity.

**Assets required**
[!Important!] Please provide sample assets needed to reproduce the issue, either as an [attachment](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/attaching-files) or as a link to a public asset in Github.

**Steps to reproduce**
1. Launch the Editor with prefab overrides enabled:
```
Editor.exe --regset=O3DE/Preferences/Prefabs/EnableOverridesUx=true
```
2. Create a new level.
3. Create an entity hierarchy in the level with a parent, child, and grandchild.
4. Move the entities so they're not on top of each other.
5. In the Viewport, select the grandchild entity.

**Expected behavior**
Grandchild entity is selected.

**Actual behavior**
The Parent entity is selected.

**Screenshots/Video**
https://user-images.githubusercontent.com/82226555/204357919-76388673-aa25-46e7-970f-94ffe591c0d9.mp4

**Found in Branch**
development@ 4f2ce1cda6d8fd986c7fcae9c0d6e2eaded453c8

**Additional context**
1. This does not occur if the user drags in the Viewport to select rather than directly clicking the entity.
2. This does not occur if the user selects entities in the Entity Outliner.

michabr:

## What does this PR do?

This PR fixes some of the entity selection logic for when prefab override UX is enabled. It was not allowing descendant entities to be selected in the viewport for entities whose owning instance is the root (level).

Fixes #13415.

## How was this PR tested?

Manual testing in Editor.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/ContainerEntity/ContainerEntitySystemComponent.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/ContainerEntity/ContainerEntitySystemComponent.cpp
@@ -128,36 +128,32 @@ namespace AzToolsFramework
                 return entityId;
             }
 
-            // Return the highest closed container, or the entity if none is found.
-            AZ::EntityId secondLastOpenContainerBeforeSelection = entityId;
-            AZ::EntityId lastOpenContainerBeforeSelection = entityId;
+            // Return the highest unselected ancestor container of the entity, or the entity if none is found.
+            AZ::EntityId highestUnselectedAncestorContainer = entityId;
 
             // Skip the queried entity, as we only want to check its ancestors.
             AZ::TransformBus::EventResult(entityId, entityId, &AZ::TransformBus::Events::GetParentId);
 
-            // Assumption - there are no closed containers. This is due to the way PrefabEditScope::NESTED_INSTANCES works,
-            // since we open all descendant containers of the focused prefab.
-            // If this is no longer the case, this code will need a rewrite.
-
-            // Go up the hierarchy until you hit the root
+            // Go up the hierarchy until you hit the root or a selected container
             while (entityId.IsValid())
             {
-                if (IsContainer(entityId) &&
-                    AZStd::find(selectedEntities.begin(), selectedEntities.end(), entityId) != selectedEntities.end())
+                if (IsContainer(entityId))
                 {
-                    return lastOpenContainerBeforeSelection;
-                }
-                else
-                {
-                    secondLastOpenContainerBeforeSelection = lastOpenContainerBeforeSelection;
-                    lastOpenContainerBeforeSelection = entityId;
+                    if (AZStd::find(selectedEntities.begin(), selectedEntities.end(), entityId) != selectedEntities.end())
+                    {
+                        // Found a selected container
+                        break;
+                    }
+                    else
+                    {
+                        highestUnselectedAncestorContainer = entityId;
+                    }
                 }
 
                 AZ::TransformBus::EventResult(entityId, entityId, &AZ::TransformBus::Events::GetParentId);
             }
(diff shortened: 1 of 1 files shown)


dtamkin1 (approved):

lsemp3d (approved):

AMZN-daimini (approved):
Looks good, thanks for fixing this!

hultonha:
Any chance we could add a test for this at the same time @michabr? 😁

michabr:
> Any chance we could add a test for this at the same time @michabr? 😁

@hultonha OK I'll look into adding a test now before submitting.

Code/Framework/AzToolsFramework/AzToolsFramework/ContainerEntity/ContainerEntitySystemComponent.cpp
```diff
@@ -128,36 +128,32 @@ namespace AzToolsFramework
...
-            AZ::EntityId lastOpenContainerBeforeSelection = entityId;
+            // Return the highest unselected ancestor container of the entity, or the entity if none is found.
+            AZ::EntityId highestUnselectedAncestorContainer = entityId;
 
             // Skip the queried entity, as we only want to check its ancestors.
             AZ::TransformBus::EventResult(entityId, entityId, &AZ::TransformBus::Events::GetParentId);
 
-            // Assumption - there are no closed containers. This is due to the way PrefabEditScope::NESTED_INSTANCES works,
-            // since we open all descendant containers of the focused prefab.
-            // If this is no longer the case, this code will need a rewrite.
```
forkercat, line 140: question: Since we are removing these comments, would it be good to update the API comment to describe the "no closed container" case? It looks like if the override flag is on, the lines after 160 won't be executed at all.

```cpp
//! Detects if one of the ancestors of entityId is a closed container entity.
//! @return The highest closed entity container id if any, or entityId otherwise.
virtual AZ::EntityId FindHighestSelectableEntity(AZ::EntityId entityId) const = 0;
```

forkercat (approved):
Thanks for fixing this!

hultonha:
> > Any chance we could add a test for this at the same time @michabr? 😁
> 
> @hultonha OK I'll look into adding a test now before submitting.

Amazing thank you, let me know if you run into any problems trying to set it up and I'll see if I can help. Writing tests can sometimes be more of a pain than fixing the bug but long term they're incredible valuable to help ensure the software is continuing to do what we think it should be doing. Thanks for taking a look!

michabr:
> 

It looks like there's already a test for this but it didn't catch the issue because PrefabOverridesEnabled is false in the tests. I'm going to create a new ticket/PR to change the logic so that it doesn't rely on that flag.

Code/Framework/AzToolsFramework/AzToolsFramework/ContainerEntity/ContainerEntitySystemComponent.cpp
michabr: I'm making a follow up ticket to change the logic so that there's no branching on PrefabsOverridesEnabled flag.

hultonha:
> > 
> 
> It looks like there's already a test for this but it didn't catch the issue because PrefabOverridesEnabled is false in the tests. I'm going to create a new ticket/PR to change the logic so that it doesn't rely on that flag.

Okay cool good to hear, if you can link the ticket here once it's created that'd be great! Thanks for investigating @michabr :+1:

PR #13519, opened 2022-12-03
Action Manager - Migrate the Viewport ToolBar to use the new system (Part 2)

AMZN-daimini:
## What does this PR do?

Migrates the remaining menus for the Viewport ToolBar to the new Action Manager.
ViewportToolBar

This changes the UI for some of the menus to standardize it.

## How was this PR tested?

Manual testing. Verified settings are updated correctly in both directions (if a setting is changed in the Editor Preferences, it gets updated correctly in this menu too and viceversa).


Code/Editor/EditorViewportSettings.cpp
```diff
@@ -288,6 +297,9 @@ namespace SandboxEditor
     void SetCameraSpeedScale(float speedScale)
     {
         AzToolsFramework::SetRegistry(CameraSpeedScaleSetting, speedScale);
+
+        AzToolsFramework::ViewportInteraction::ViewportSettingsNotificationBus::Broadcast(
+            &AzToolsFramework::ViewportInteraction::ViewportSettingNotifications::OnCameraSpeedScaleChanged, speedScale);
```
hultonha, line 302: question/suggestion: I don't think this should need to be added if we use an event like everything else does (see top of file), I know it's a bit more typing but at least things are then consistent for now, instead of mixing how notifications are sent?

Code/Editor/ViewPane.cpp
```diff
@@ -51,13 +53,26 @@
...
 static constexpr AZStd::string_view ViewportTopToolBarIdentifier = "o3de.toolbar.viewport.top";
 
+static const std::pair<int, int> ViewportRatios[] = { { 16, 9 }, { 16, 10 }, { 4, 3 }, { 5, 4 } };
+static const size_t ViewportRatiosCount = sizeof(ViewportRatios) / sizeof(ViewportRatios[0]);
+static const std::pair<int, int> ViewportResolutions[] =
+    { { 1280, 720 }, { 1920, 1080 }, { 2560, 1440 }, { 2048, 858 }, { 1998, 1080 }, { 3480, 2160 } };
+static const size_t ViewportResolutionsCount = sizeof(ViewportResolutions) / sizeof(ViewportResolutions[0]);
+
+
+
```
hultonha, line 75: uber nit: Update 3 lines break to just 1

Code/Editor/ViewPane.cpp
```diff
@@ -397,19 +465,115 @@ void CLayoutViewPane::OnActionRegistrationHook()
...
+        AZStd::string actionIdentifier = AZStd::string::format("o3de.action.viewport.size.ratio[%i:%i]", width, height);
+        AzToolsFramework::ActionProperties actionProperties;
+        actionProperties.m_name = AZStd::string::format("%i:%i", width, height);
+        actionProperties.m_category = "Viewport Size Ratio";
+
+        m_actionManagerInterface->RegisterAction(
+            EditorMainWindowActionContextIdentifier,
+            actionIdentifier,
+            actionProperties,
+            [&, w = width, h = height]
```
hultonha, line 489: nit: Should we capture `this` here instead of `&`?

Code/Editor/ViewPane.cpp
```diff
@@ -397,19 +465,115 @@ void CLayoutViewPane::OnActionRegistrationHook()
...
+        constexpr AZStd::string_view actionIdentifier = "o3de.action.viewport.size.resolution.custom";
+        AzToolsFramework::ActionProperties actionProperties;
+        actionProperties.m_name = "Custom...";
+        actionProperties.m_category = "Viewport Size Resolution";
+
+        m_actionManagerInterface->RegisterAction(
+            EditorMainWindowActionContextIdentifier,
+            actionIdentifier,
+            actionProperties,
+            [&]
```
hultonha, line 554: nit: It may only be necessary to capture `this` here potentially

Code/Editor/ViewPane.cpp
```diff
@@ -424,18 +588,68 @@ void CLayoutViewPane::OnMenuBindingHook()
...
+    {
+        m_menuManagerInterface->AddSubMenuToMenu(ViewportSizeMenuIdentifier, ViewportSizeRatioMenuIdentifier, 100);
+        {
+            for (size_t i = 0; i < ViewportRatiosCount; ++i)
+            {
+                int width = ViewportRatios[i].first;
+                int height = ViewportRatios[i].second;
+                AZStd::string actionIdentifier = AZStd::string::format("o3de.action.viewport.size.ratio[%i:%i]", width, height);
+
+                m_menuManagerInterface->AddActionToMenu(ViewportSizeRatioMenuIdentifier, actionIdentifier, 100 * (aznumeric_cast<int>(i) + 1));
```
hultonha, line 602: suggestion: In this function is it worth moving the `100`s to a constant so it's easier to know what it is referring to (be easier to change in one place in future too then 🙂)

Code/Editor/EditorViewportSettings.cpp
AMZN-daimini: I have transitioned this to use AZ::Event.

Code/Editor/Core/Widgets/ViewportSettingsWidgets.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+#include <Core/Widgets/ViewportSettingsWidgets.h>
+#include <EditorViewportSettings.h>
+
+#include <QHBoxLayout>
+
+constexpr int FieldMargins = 17;
+constexpr int SpinBoxWidth = 64;
+
+PropertyInputDoubleWidget::PropertyInputDoubleWidget()
```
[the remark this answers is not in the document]
hultonha: Awesome thank you!

Code/Editor/EditorViewportSettings.cpp
hultonha: Perfect thanks!

hultonha:
> > @AMZN-daimini I recently fixed some issues with focus when changing camera speed scale and fov (as well as grid and angle snapping). I just wanted to check how those drop down menus work. Does the menu immediately disappear when changing the value? As we actually made this not happen in the current version so it's easier to continue editing the value. Is there a chance of including a video showing changing the speed scale and then pressing enter to confirm? Thanks!
> 
> Thanks for the nudge, I thought it was working fine but in one of the menus it actually would trigger the action next to the submenu on Enter. I have added some code to stop propagation and it works as expected now.

Perfect thank you! 😍

PR #13526, opened 2022-12-03
Material Canvas: Small change starting file for graph related utility functions

gadams3:
## What does this PR do?

Creating holding pin for useful utility functions not exclusive to material canvas

## How was this PR tested?

Just moving one function and removing another. Testing that graph document still displayed correctly in inspector And instructions were still generated in execution order.

### DIFF
--- a/Gems/Atom/Tools/AtomToolsFramework/Code/Include/AtomToolsFramework/Graph/GraphUtil.h
+++ b/Gems/Atom/Tools/AtomToolsFramework/Code/Include/AtomToolsFramework/Graph/GraphUtil.h
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#pragma once
+
+#include <AzCore/std/sort.h>
+#include <AzCore/std/tuple.h>
+#include <GraphModel/Model/Connection.h>
+#include <GraphModel/Model/Node.h>
+
+namespace AtomToolsFramework
+{
+    // Sort a container of nodes by depth, considering the number and state of input and output slots and connections, for consistent
+    // display and execution order. The function is templatized to account for different container types and constness. 
+    template<typename NodeContainer>
+    void SortNodesInExecutionOrder(NodeContainer& nodes)
+    {
+        using NodeValueType = typename NodeContainer::value_type;
+        using NodeValueTypeRef = typename NodeContainer::const_reference;
+
+        // Pre-calculate and cache sorting scores for all nodes to avoid reprocessing during the sort
+        AZStd::map<NodeValueType, AZStd::tuple<bool, bool, uint32_t>> nodeScoreTable;
+        for (NodeValueTypeRef node : nodes)
+        {
+            nodeScoreTable.emplace(node, AZStd::make_tuple(node->HasInputSlots(), !node->HasOutputSlots(), node->GetMaxInputDepth()));
+        }
+
+        AZStd::stable_sort(nodes.begin(), nodes.end(), [&](NodeValueTypeRef nodeA, NodeValueTypeRef nodeB) {
+            return nodeScoreTable[nodeA] < nodeScoreTable[nodeB];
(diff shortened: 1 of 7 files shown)


cgalvan (approved):

Gems/Atom/Tools/AtomToolsFramework/Code/Include/AtomToolsFramework/Graph/GraphUtil.h
```diff
@@ -0,0 +1,37 @@
...
+#include <AzCore/std/tuple.h>
+#include <GraphModel/Model/Connection.h>
+#include <GraphModel/Model/Node.h>
+
+namespace AtomToolsFramework
+{
+    // Sort a container of nodes by depth, considering the number and state of input and output slots and connections, for consistent
+    // display and execution order. The function is templatized to account for different container types and constness. 
+    template<typename NodeContainer>
+    void SortNodesInExecutionOrder(NodeContainer& nodes)
```
jeremyong-az, line 21: I know we do this all over the codebase, but this actually exceeds my bar for "too big to be in an inlined template function." Why not have the `NodeValueType` here inherit from a base class providing `HasInputSlots`, `HasOutputSlots`, and `GetMaxInputDepth`?

Gems/Atom/Tools/AtomToolsFramework/Code/Include/AtomToolsFramework/Graph/GraphUtil.h
```diff
@@ -0,0 +1,37 @@
...
+        using NodeValueTypeRef = typename NodeContainer::const_reference;
+
+        // Pre-calculate and cache sorting scores for all nodes to avoid reprocessing during the sort
+        AZStd::map<NodeValueType, AZStd::tuple<bool, bool, uint32_t>> nodeScoreTable;
+        for (NodeValueTypeRef node : nodes)
+        {
+            nodeScoreTable.emplace(node, AZStd::make_tuple(node->HasInputSlots(), !node->HasOutputSlots(), node->GetMaxInputDepth()));
+        }
+
+        AZStd::stable_sort(nodes.begin(), nodes.end(), [&](NodeValueTypeRef nodeA, NodeValueTypeRef nodeB) {
```
jeremyong-az, line 33: Q: does this need to be a stable sort?

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Graph/GraphUtil.cpp
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#include <AtomToolsFramework/Graph/GraphUtil.h>
+
+namespace AtomToolsFramework
```
jeremyong-az, line 11: What is this file for?

Gems/Atom/Tools/AtomToolsFramework/Code/Include/AtomToolsFramework/Graph/GraphUtil.h
gadams3: The node class is the same regardless of the template type. The template was previously needed because the container type and constness would change. I will see if there's another way to organize it to shrink the template function.

Gems/Atom/Tools/AtomToolsFramework/Code/Include/AtomToolsFramework/Graph/GraphUtil.h
gadams3: It could probably just be sort. I went through several iterations before ending up with a mostly simple depth based sort that resulted in deterministic instruction order. One of the iterations did multiple passes with different criteria. That required the sort to be stable. This probably does not.

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Graph/GraphUtil.cpp
gadams3: Placeholder for other shared graph utility functions as I extract them from material canvas

Issue #7986, opened 2022-03-01
Remove all elements button does not change the cursor's appearance when hovered over


LB-ArturZieba:
**Describe the bug**
When hovering over the "Remove all elements" button, the cursor does not change it's appearance, which is inconsistent with "Remove this element" button hover cursor behavior.

**Steps to reproduce**
1. Create an entity and add a component that has an array property (e.g. Compound Shape or Tag).
2. Add a child element to that array property and unfold it.
3. Hover over the "Remove all elements" and "Remove this element" buttons.

**Expected behavior**
When hovered over either the "Remove all elements" or "Remove this element" button the cursor changes it's appearance.

**Actual behavior**
When hovered over the "Remove all elements" button the cursor does not change it's appearance.

**Video**
https://user-images.githubusercontent.com/86952082/156187801-83967f4e-abd7-4393-ade4-472c1a6b4449.mp4

**Found in Branch**
Development (f74dca6)

**Desktop**
 - Device: PC
 - OS: Linux
 - Version: Ubuntu 20.04.3 LTS
 - CPU AMD Ryzen 5 3600
 - GPU NVIDIA GeForce RTX 2060 SUPER
 - Memory 16GB

PR #13528, opened 2022-12-04
Fix to cursor's appearance not changing when hovering over remove all elements and add new child element buttons

EyeAmfour:
## What does this PR do?

This change fixes the bug described in [Issue #7986](https://github.com/o3de/o3de/issues/7986). Previously, the cursor stayed as the default pointer when hovering over the "Remove all elements" and "Add new child element" buttons. Now, the cursor's appearance is set to the same "Grab_release.svg" image that is used for the "Remove this element" button.

### Cursor Appearance Before Changes
image
image
image
image

### Cursor Appearance After Changes
image
image
image
image

## How was this PR tested?

Manually tested in development. Was able to view changes successfully, as displayed above.

Signed-off-by: EyeAmfour <fruitloopsgo@gmail.com>


Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyRowWidget.cpp
```diff
@@ -368,8 +371,6 @@ namespace AzToolsFramework
             delete m_containerAddButton;
         }
 
-        this->unsetCursor();
```
cgalvan, line 371: I think we need to leave this call to `unsetCursor` because the property row widgets don't get deleted, but instead get returned to a pool of widgets to be re-used.

EyeAmfour: Leaving that call, at least where it was, is what was preventing the cursor from changing on the two buttons (in addition to the lack of setting the cursor). If it is necessary to have `unsetCursor` called then I can look into moving it above the initialization of the widgets. I tried moving it below all widget initialization and it prevents the cursor from changing over any of the buttons.

cgalvan: Yeah try moving it up to the top of the `Initialize` method, below this block:

```
m_initialized = true;
m_hadChildren = false;
m_parentRow = pParent;
m_isMultiSizeContainer = false;
m_isSelected = false;
...
```

EyeAmfour: @cgalvan That seemed to do the trick. Committed and pushed the change.

cgalvan (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyRowWidget.cpp
cgalvan: Great, thanks! I'll kick off an AR run now.

phuchau1989 (approved):

gadams3 (approved):

amzn-tmryan (approved):

lsemp3d (approved):

dtamkin1 (approved):

cgalvan:
Hi @EyeAmfour, I'm sorry the AR run has failed a few times, I don't believe the error is due to your change. I've tried a few different levels incremental vs. clean builds that haven't worked yet, but have just kicked off a new run that should completely rebuild the workspace, which should hopefully resolve the issue.

If it still fails, can you try pulling the latest from development into your branch? No need to touch it yet though, just wait until we get the results from this current run and then we can go from there!

Thank you for your patience.

cgalvan:
@EyeAmfour Unfortunately it looks like AR still failed, but if you pull latest development into your branch I can kick off a new AR run and that should resolve it! All you need to do is the following from your branch:

```
git fetch upstream
git pull
git push origin
```

hultonha:
Hooray! 😄 Great to see this finally merged, awesome work @EyeAmfour and @cgalvan (I'm sure the next one will be easier! 😝)

Issue #13050, opened 2022-11-07
Feature Request: Material Canvas Compiling Visual Feedback


ParasiteSt:
**Is your feature request related to a problem? Please describe.**
When working in the Material Canvas I am always having to wait and check the AP to see if it has finished compiling to know if what I did is working correctly or not.

**Describe the solution you'd like**
Having a visual icon or text would allow people to see that the material is compiling and has compiled once it has finished. This will make it easier to stay in one window and not have to bounce around to make sure things are compiling correctly.

**Describe alternatives you've considered**
I have not had a good alternative yet, if others have one I would love to hear about it.

**Additional context**
If you look at other visual shader editors like Unreal's once you hit the compile button it shows in the viewport that the shaders or compiling, and once they hit 0 your shader is updated. Having this visual feedback lets the developer know how long it should take, once its finished, etc.

gadams3:
Thank you for the feedback. This is on the list. Are you also aware of options in the tools settings menu that enable faster iteration/preview times by disabling shader compilation for inactive platforms? Could you please also provide info on your hardware and how long it takes to see updates with that setting enabled or disabled?

ParasiteSt:
Thanks for the assurance its on the list @gadams3 . I did a few small tests, and compilation times without this feature toggled on is around 12.8 - 13.06 seconds, and with the feature toggled on it is around 4.6 -4.84 seconds.
Without it toggled on ;
![image](https://user-images.githubusercontent.com/100494538/201410458-a6454d4a-8a70-4124-a7b7-00bf62a86c70.png)
With it toggled on ;
![image](https://user-images.githubusercontent.com/100494538/201410673-2f871fea-1684-4bce-8d08-931030b8d4c5.png)


Hardware is Intel i5-8400, 16gb ram, Nvidia GTX 1050 Ti.

PR #13532, opened 2022-12-05
Material Canvas: Change material graph compiler to check and report asset job status\

gadams3:
## What does this PR do?

• This change updates the material graph compiler to check the status of all asset jobs created from generated files. The graph compiler checks the status of each file on system tick and sends the status messages to the application main windows status bar.
• Main windows status bar functions were moved to the main window request bus so they could be accessed from the graph compiler. This might be inverted so that the graph compiler sends out the status messages for any observer to handle.
• Changed the status message functions to use the AZStd string and updated all the call sites. This might be cleaner with overloads for AZStd string and QString.
• Updated the viewport to clear the material whenever graph compilation starts or fails. Added options to toggle this behavior for users that would rather not have animations or other effects interrupted.

Resolves https://github.com/o3de/o3de/issues/13050

## How was this PR tested?

Manually tested and confirmed that material canvas still generates materials and shaders from open graph documents.
Manually tested and confirmed that material canvas still updates the viewport after all of the materials and shaders have finished being processed by the AP.
Manually tested and confirmed that status messages are reported to the main windows taskbar when material graph compilation starts, finishes, or fails.


Gems/Atom/Tools/MaterialCanvas/Code/Source/Document/MaterialGraphCompiler.cpp
```diff
@@ -1122,37 +1176,80 @@ namespace MaterialCanvas
...
     }
 
-    void MaterialGraphCompiler::BuildSlotValueTable() const
+    AZStd::string MaterialGraphCompiler::GetGraphPath() const
+    {
+        AZStd::string absolutePath;
+        AtomToolsFramework::AtomToolsDocumentRequestBus::EventResult(
+            absolutePath, m_documentId, &AtomToolsFramework::AtomToolsDocumentRequestBus::Events::GetAbsolutePath);
+
+        if (absolutePath.ends_with(".materialgraph"))
```
cgalvan, line 1244: Nit: Can we use `MaterialGraphExtensionWithDot` here so we don't need to duplicate the extension string?

cgalvan (approved):

Gems/Atom/Tools/MaterialCanvas/Code/Source/Document/MaterialGraphCompiler.cpp
```diff
@@ -1122,37 +1176,80 @@ namespace MaterialCanvas
...
         m_slotValueTable.clear();
         m_generatedFiles.clear();
-        AZ_TracePrintf_IfTrue("MaterialGraphCompiler", IsCompileLoggingEnabled(), "Compile graph failed.\n");
-        MaterialGraphCompilerNotificationBus::Event(m_toolId, &MaterialGraphCompilerNotificationBus::Events::OnCompileGraphFailed, m_documentId);
+        AZ_TracePrintf_IfTrue("MaterialGraphCompiler", IsCompileLoggingEnabled(), "Compiling failed.\n");
+
+        AtomToolsFramework::AtomToolsMainWindowRequestBus::Event(
+            m_toolId,
+            &AtomToolsFramework::AtomToolsMainWindowRequestBus::Events::SetStatusError,
+            AZStd::string::format("Compiling %s (Failed)", GetGraphPath().c_str()));
```
santorac, line 1215: Wherever possible, I think I'd rather have the same message be used in both the status bar and in the log. That way if I see a message in the status bar I can search the log for that message, and see if there is other relevant information around that event.

jeremyong-az (approved):

Gems/Atom/Tools/MaterialCanvas/Code/Source/Document/MaterialGraphCompiler.cpp
gadams3: I had to think about that for a minute because of where I put the string definitions. This should be accessible from here but that's a weird dependency. I should find a better commissary place to move them to be accessible by all.

PR #13540, opened 2022-12-05
Made RenderPipeline filter by material pipeline tag

santorac:
## Background

The material pipeline system that we are working toward (https://github.com/o3de/sig-graphics-audio/blob/main/rfcs/MaterialPipelineAbstraction.md) is supposed to hide the render pipeline details, including the list of specific shader assets, from the material type. Each material pipeline will need its own material functors that are responsible for enabling and disabling specific shaders, based on configuration that comes from the material type.

So far, the MaterialTypeAsset has had a flat list of all possible shaders for all possible pipelines, and shaders were filtered into appropriate passes by the drawListTag. Here are some examples of the shaders we have:

| Shader File | Draw List Tag | Notes |
| --- | --- | --- |
| Common\_Depth.shader | depth | runs in every render pipeline because every render pipeline has a "depth" pass, if enabled |
| Common\_Depth\_WithPS.shader | depth | also runs in every render pipeline, if enabled. |
| MainPipeline\_ForwardPass.shader  | forward | that works in the MainPipeline |
| LowEndPipeline\_ForwardPass.shader  | forwardLowEnd | works in the LowEndPipeline |
| DeferredPipeline\_DeferredMaterialPass.shader | deferredMaterial | works in the prototype DeferredPipeline |
    
Notice that there are two different depth pass shaders available, but only one of them should be used, depending on each material's configuration. What if one material pipeline's shader-enable functor wants to enable the Common_Depth.shader, but another pipeline's shader-enable functor wants to enable the Common_Depth_WithPS.shader? (There will be other use cases, this is just one example). The only way we can achieve this is by making the MaterialTypeAsset differentiate between the each material pipeline's shaders and configure them separately with separate functors. 

## What does this PR do?

In this PR we're making MaterialTypeAsset separate the shaders for each material pipeline, and filter them into the appropriate corresponding render pipeline at runtime. (A subsequent PR will add the support for separate per-pipeline functors).

- The RenderPipelineDescriptor can include the name of a material pipeline to indicate which material pipeline's draw items it can process. This name is included in the RenderPipeline's DrawFilterMask. Then each DrawItem can opt-in to the filtering by providing a DrawFilterTag to force it into a sepecific pipeline. 
- The MaterialTypeAsset and other material classes now use a map of ShaderCollection objects instead of just one, for each material pipeline. There's also general ShaderCollection (using an empty material pipeline name) that works like before, so that old-style material types are supported.
- I had to change the way MaterialTypeAsset was tracking the MaterialSrg and ObjectSrg. It was using indexes into the shader lists and was going to be too clunky if it also had to track a material piepline. Keeping references to the ShaderAssets is much cleaner.
- Even though this PR does not introduce material pipeline functors, it does include a bit of groundwork for them in MaterialTypeSourceData and MaterialTypeAssetCreator. This helps clarify where the system design is headed.
- Note there was already a good amount of plumbing in the DrawPacket, DrawItem, and related classes for DrawFilterMask that was not being used. I just had to move some things around a little bit.

Now, the generated intermediate .materialtype file looks like this:
```
{
    "propertyLayout": {
    ...
    },
    "materialPipelines": {
        "DeferredPipeline": {
            "shaders": [
                {
                    "file": "materialpipelinetest_basic_common_depthpass.shader"
                },
                {
                    "file": "materialpipelinetest_basic_common_shadowmappass.shader"
                },
                {
                    "file": "materialpipelinetest_basic_deferredpipeline_deferredmaterialpass.shader"
                }
            ]
        },
        "LowEndPipeline": {
            "shaders": [
                {
                    "file": "materialpipelinetest_basic_common_depthpass.shader"
                },
                {
                    "file": "materialpipelinetest_basic_common_shadowmappass.shader"
                },
                {
                    "file": "materialpipelinetest_basic_lowendpipeline_forwardpass_baselighting.shader"
                }
            ]
        },
        "MainPipeline": {
            "shaders": [
                {
                    "file": "materialpipelinetest_basic_common_depthpass.shader"
                },
                {
                    "file": "materialpipelinetest_basic_common_shadowmappass.shader"
                },
                {
                    "file": "materialpipelinetest_basic_mainpipeline_forwardpass_baselighting.shader"
                }
            ]
        },
        "MultiViewPipeline": {
            "shaders": [
                {
                    "file": "materialpipelinetest_basic_multiviewpipeline_forwardpass_baselighting.shader"
                }
            ]
        }
    },
    ...
}
```

## How was this PR tested?

- Opened original material types in Material Editor.
- In Material Editor, cycled through each render pipeline for each MaterialPipelineTest material.
- Saw thumbnails working in Material Editor for both original material types and MaterialPipelineTest material types.
- Toggled Use Texture properties in Material Editor.
- Tested ASV DynamicDraw sample, and toggled each setting.
- Edge highlighting of selected entities in Editor.exe still works.
- Monochromatic mode for prefab editing still works.
- Make sure Material Canvas graph and preview is still working.
- Atom RPI unit tests.


Gems/Atom/RPI/Code/Source/RPI.Reflect/Material/MaterialFunctor.cpp
```diff
@@ -33,46 +33,60 @@ namespace AZ
...
+            static ShaderCollection EmptyShaderCollection;
 
+            auto iter = m_allShaderCollections->find(MaterialPipelineNameCommon);
+            if (iter != m_allShaderCollections->end())
+            {
+                m_commonShaderCollection = &iter->second;
+            }
+            else
+            {
+                m_commonShaderCollection = &EmptyShaderCollection;
```
jeremyong-az, line 57: This gets stored as a mutable pointer. Are you worried about the collection being modified here? (the static empty shader collection will be shared across all `MaterialFunctor` instances)

jeremyong-az (approved):

gadams3 (approved):

Gems/Atom/RPI/Code/Source/RPI.Reflect/Material/MaterialFunctor.cpp
santorac: I'm making a _lot_ of code changes for this feature, I think putting it all together in one PR would be too much for reviewers.

PR #13547, opened 2022-12-05
bootstrap setreg files to use the proper platform folders

byrcolin:
Change/add the bootstrap setreg files to use the proper platform folders instead of using the old PlatformGet() function. Once no longer used we can deprecate PlatformGet()

Signed-off-by: byrcolin <byrcolin@amazon.com>


lumberyard-employee-dm (approved):

moraaar (approved):
Will the old way still work? If not, then please update the following [android document](https://github.com/o3de/sig-platform/blob/main/testing-guidance/workflow-tests/android/android-lkg-workflows.md), table entry `Assets by VFS` > point 2.

moraaar:
In addition, please check `PlatformGet` is still not used in o3de-extras and asv.

I remember we have this one in OpenXRVk gem: https://github.com/o3de/o3de-extras/blob/development/Gems/OpenXRVk/Code/Source/OpenXRVkSwapChain.cpp#L163-L169

If you want to remove that one too the following file needs to be updated `https://github.com/o3de/o3de-extras/blob/development/Projects/OpenXRTest/Registry/OpenXR.setreg` and PALify that value with the android version.

Lastly mentions to `android_ViewResolutionScale` would need to be updated in this document https://github.com/o3de/o3de-extras/wiki/Build-and-Run-OpenXR-in-O3DE

amzn-phist (approved):
It's a good change.  I agree to be a bit more cautious before removing `PlatformGet` completely though, but for sure put it on a deprecation path.

PR #13550, opened 2022-12-06
name is only mean default folder

byrcolin:
If a name is only given instead of a path then the default folder is assumed

### DIFF
--- a/scripts/o3de/o3de/engine_template.py
+++ b/scripts/o3de/o3de/engine_template.py
@@ -489,6 +489,7 @@ def _instantiate_template(template_json_data: dict,
 
 def create_template(source_path: pathlib.Path,
                     template_path: pathlib.Path,
+                    template_name: str = None,
                     source_name: str = None,
                     source_restricted_path: pathlib.Path = None,
                     source_restricted_name: str = None,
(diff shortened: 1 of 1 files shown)


scripts/o3de/o3de/engine_template.py
```diff
@@ -1633,16 +1647,19 @@ def create_project(project_path: pathlib.Path,
...
         default_projects_folder = manifest.get_registered(default_folder='projects')
-        new_project_path = default_projects_folder / project_path
-        logger.info(f'Project Path {project_path} is not a full path, we must assume its relative'
+        new_project_path = default_projects_folder / project_name
+        logger.info(f'Only Project Name {project_name} specified, we must assume its relative'
                     f' to default projects path = {new_project_path}')
         project_path = new_project_path
+        # project name is now the last component of the project_path, so if someone enters some/path/toproject
+        # the destination is <default folder>/some/path/toproject which would make the name of the gem "toproject"
+        project_name = os.path.basename(project_path)
```
lumberyard-employee-dm, line 1661: The project_name shouldn't be set here as the user supplied via the command line option.
This code is within the if block when `project_name` is valid and therefore we shouldn't override the value.

lumberyard-employee-dm (approved):

PR #9569, opened 2022-05-16
Fixed UIElement ordering issue

cgalvan:
Fixes #9478 

Fixed issue when trying to use an `UIElement` without a backing data element getting an incorrect/random order in the RPE (usually getting pinned to the top regardless of expected order).

The logic that was iterating through the children on the node in the `InstanceDataHierarchy` didn't handle the case where a child was found that had no edit data (which occurs on anything that is trying to be reflected from an `AZ::Component` because it has an `Id` parameter that is serialized but not exposed to the `EditContext`). So the logic wouldn't move forward past this child, which would result (usually) in any `UIElement` being pinned to the top, since it would never be able to correctly find the child index of the child that came before the `UIElement`.

Made the following changes:
- Logic in `InstanceDataHierarchy` remains the same with the addition of it skipping over elements that don't have any edit data
- Updated unit test that was previously checking `UIElement` ordering to derive from `AZ::Component` in order to expose the ordering issue
- Updated RPE page in the control gallery to showcase using `UIElements` without a data element (without this change, the `UIElements` would show up at the top)
- Updated a few places in the code-base that were using the workaround of having a dummy variable be serialized (but never actually used) in order to control the ordering of a button to instead just use a `UIElement` with no data element, and verified the buttons still show up in the expected order
- Fixed Atom automated test to actually click the "Add Material Component" button that was previously relying on the workaround by changing the boolean data element value

UIElementsCorrectOrder

Signed-off-by: Chris Galvan <chgalvan@amazon.com>

PR #13557, opened 2022-12-06
Update Rigid Body CCD UI to reflect the global CCD property

amzn-ahmadkrm:

## What does this PR do?

Currently, the user can try to enable CCD in the Rigid body component not knowing that it disabled globally and wonder why nothing is changing. This PR updates the UI so that it is clear when the global CCD setting is enabled or disabled.

Editor_4OhPzrH5Iu
.

https://user-images.githubusercontent.com/105638312/205929996-e731de7b-0902-4bce-87eb-b159c858d471.mp4

## How was this PR tested?

Manual testing and unit test for checking if the button should be visible based on scene configuration.


amzn-ahmadkrm:
Tagging @amzn-leenguy for UI input

Gems/PhysX/Code/Source/EditorRigidBodyComponent.cpp
```diff
@@ -188,18 +190,29 @@ namespace PhysX
...
+                        ->Attribute(AZ::Edit::Attributes::ReadOnly, &EditorRigidBodyComponent::IsSceneCCDDisabled)
                     ->DataElement(AZ::Edit::UIHandlers::Default, &AzPhysics::RigidBodyConfiguration::m_ccdFrictionEnabled,
                         "CCD friction", "When active, friction is applied when continuous collision detection (CCD) collisions are resolved.")
                         ->Attribute(AZ::Edit::Attributes::Visibility, &AzPhysics::RigidBodyConfiguration::IsCCDEnabled)
+                        ->Attribute(AZ::Edit::Attributes::ReadOnly, &EditorRigidBodyComponent::IsSceneCCDDisabled)
+                    ->DataElement(
+                        AZ::Edit::UIHandlers::Button,
+                        &AzPhysics::RigidBodyConfiguration::m_configButton,
+                        "",
+                        "Click here to open the PhysX Configuration window, Enable global CCD to enable component CCD editing.")
```
greerdv, line 212: nitpick - I would suggest either changing the capital E in Enable to lower case, or changing the comma preceding it to a full stop.

Gems/PhysX/Code/Source/EditorRigidBodyComponent.cpp
```diff
@@ -302,6 +331,11 @@ namespace PhysX
     {
         m_debugDisplayDataChangeHandler.Disconnect();
 
+        if (m_sceneConfigChangedHandler.IsConnected())
+        {
+            m_sceneConfigChangedHandler.Disconnect();
```
greerdv, line 336: Not sure how it works with events, but I think with buses you can safely call Disconnect even if you're not connected, so you don't need to wrap it in a check to see if it's connected. Maybe worth checking if this is necessary for events.

greerdv (approved):
Looks good to me, nice work!

moraaar (approved):
Nice improvement! Good job!

Gems/PhysX/Code/Source/EditorRigidBodyComponent.cpp
hultonha: That's correct, the `IsConnected` check happens already in `Disconnect` itself

Gems/PhysX/Code/Source/EditorRigidBodyComponent.h
```diff
@@ -92,6 +92,8 @@ namespace PhysX
         void BuildGameEntity(AZ::Entity* gameEntity) override;
 
         const AzPhysics::RigidBody* GetRigidBody() const;
+        const bool IsSceneCCDDisabled() const;
+        void OpenPhysXConfigurationPane() const;
```
hultonha, line 96: update: `const bool` -> `bool` and also should `OpenPhysXConfigurationPane` be `private`? Does anything else need this outside the class?

Gems/PhysX/Code/Tests/RigidBodyComponentTests.cpp
```diff
@@ -231,4 +231,26 @@ namespace PhysXEditorTests
         PhysX::EditorColliderComponentRequestBus::Event(idPair, &PhysX::EditorColliderComponentRequests::SetCylinderSubdivisionCount, subdivisionsTooLarge);
         EXPECT_EQ(expectedError.GetExpectedWarningCount(), 2);
     }
+
+    TEST_F(PhysXEditorFixture, EditorRigidBodyComponent_PhysXConfigButtonVisibilityFunctionUpdatesOnSceneConfigurationChange)
```
hultonha, line 235: Great work adding this test 👍

hultonha (approved):
Looks good just a few mostly minor comments

Code/Framework/AzFramework/AzFramework/Physics/Configuration/RigidBodyConfiguration.h
```diff
@@ -57,6 +57,7 @@ namespace AzPhysics
         bool m_ccdEnabled = false; //!< Whether continuous collision detection is enabled.
         float m_ccdMinAdvanceCoefficient = 0.15f; //!< Coefficient affecting how granularly time is subdivided in CCD.
         bool m_ccdFrictionEnabled = false; //!< Whether friction is applied when resolving CCD collisions.
+        bool m_configButton = false; // required for button positioning in edit context, otherwise unused
```
cgalvan, line 60: I actually fixed the `UIElement` ordering issue earlier this year: https://github.com/o3de/o3de/pull/9569

So we don't need to do the workaround anymore by having an unused data element, so this can be removed and below in the EditContext you can just do:

`->UIElement(AZ::Edit::UIHandlers::Button, ...`

cgalvan (changes requested):

Gems/PhysX/Code/Source/EditorRigidBodyComponent.cpp
```diff
@@ -317,6 +351,21 @@ namespace PhysX
         }
     }
 
+    const bool EditorRigidBodyComponent::IsSceneCCDDisabled() const
```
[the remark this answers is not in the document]
amzn-ahmadkrm: I agree but I can't seem to use a negative with the Attributes in Reflect(), if I use !IsSceneCCDEnabled then it won't change the visibility

Gems/PhysX/Code/Source/EditorRigidBodyComponent.h
amzn-ahmadkrm: OpenPhysXConfigurationPane is defined in EditorRigidBodyComponent and is being accessed by EditorRigidBodyConfiguration, I could move it there but it doesn't really make sense there

Code/Framework/AzFramework/AzFramework/Physics/Configuration/RigidBodyConfiguration.h
amzn-ahmadkrm: I tried that first and just again now and the button doesn't appear in the right place: 
image

moraaar (approved):

Code/Framework/AzFramework/AzFramework/Physics/Configuration/RigidBodyConfiguration.h
hultonha: Any suggestions @cgalvan, it seems this may have regressed? Can we keep the data element for now?

Gems/PhysX/Code/Source/EditorRigidBodyComponent.cpp
```diff
@@ -188,18 +190,29 @@ namespace PhysX
...
                     ->DataElement(AZ::Edit::UIHandlers::Default, &AzPhysics::RigidBodyConfiguration::m_ccdFrictionEnabled,
                         "CCD friction", "When active, friction is applied when continuous collision detection (CCD) collisions are resolved.")
                         ->Attribute(AZ::Edit::Attributes::Visibility, &AzPhysics::RigidBodyConfiguration::IsCCDEnabled)
+                        ->Attribute(AZ::Edit::Attributes::ReadOnly, &EditorRigidBodyComponent::IsSceneCCDDisabled)
+                    ->DataElement(
+                        AZ::Edit::UIHandlers::Button,
+                        &AzPhysics::RigidBodyConfiguration::m_configButton,
+                        "",
+                        "Click here to open the PhysX Configuration window, Enable global CCD to enable component CCD editing.")
+                        ->Attribute(AZ::Edit::Attributes::ButtonText, "Open PhysX Configuration to Enable CCD")
```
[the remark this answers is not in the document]
hultonha: @amzn-ahmadkrm did you spot this one?

Gems/PhysX/Code/Source/EditorRigidBodyComponent.cpp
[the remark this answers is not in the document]
hultonha: Hmm I see... can you maybe have a public function that's called `IsSceneCcdEnabled`, and then make another private one which is called `IsSceneCcdDisabled` which just calls `!IsSceneCcdEnabled()`. This makes other calling code simple but handles the annoying internal case with the EditContext? It might also be possible to use a lambda instead maybe? So something like....

```
[this]
{
    return !IsSceneCcdEnabled();
}
```

hultonha (approved):

Code/Framework/AzFramework/AzFramework/Physics/Configuration/RigidBodyConfiguration.h
cgalvan: Weird, I see the same issue locally for this component, it works fine in other components in sub-groups so I'm not sure what the issue is, I'll remove my change request. It's probably not worth investigating further with the efforts on the DPE.

cgalvan (approved):

PR #13567, opened 2022-12-06
Gem, project and engine compatibility checks

AMZN-alexpete:
## What does this PR do?

Initial version functions to support versioning in `engine.json`, `project.json` and `gem.json`

Relevant RFC https://github.com/o3de/sig-core/issues/44

CLI change summary:
- `enable_gem` command now checks if the project is missing gem dependencies or engines when  a user attempts to enable a gem.  
  - If the gem specifies gem dependencies and the project is missing a gem dependency or has a version that is not known compatible, the command fails.
  -  If the gem specifies compatible engines or engine api dependencies and the project uses an engine that is not known compatible, or if engine api dependencies are missing or wrong versions, the command fails.  
  - The command can be force to succeed using the `--force` parameter
  - The command can be used to check dependencies only using the `--check` parameter
- `gem_properties` command now lets the user specify gem `version` and `compatible engines`
- `project_properties` command now lets the user specify project `version` and `compatible engines`
- `register` command now updates the `project.json` with the `engine_version` when registering a project path
- `engine_properties` command now lets the user specify `display_version` and `api_versions`
 
Test change summary:
- `test_enable_gem` tests the result of enabling a gem with various gem and engine dependencies
- `test_gem_properties` tests modification of `gem_version` and `compatible_engines` fields
- `test_project_properties` tests modification of `project_version` and `compatible_engines` fields
- `test_register` tests project registration with various gem and engine dependencies
- `test_engine_properties` tests modification of `gem_names`, `display_version` and `api_versions` fields

## How was this PR tested?

- Unit tests have been added or updated to cover the changes
- Tested CLI editing of gem, project and engine properties, registering a project with compatibility settings with an engine and enabling a gem with compatibility settings with a project


scripts/o3de/o3de/utils.py
```diff
@@ -346,3 +365,97 @@ def remove_gem_duplicates(gems: list) -> list:
...
+
+    regex_str = r"(?P<object_name>(.*?))(?P<version_specifier>((~=|==|!=|<=|>=|<|>|===)(\s*\S+)+))"
+    regex = re.compile(r"^\s*" + regex_str + r"\s*$", re.VERBOSE | re.IGNORECASE)
+    match = regex.search(input)
+
+    if not match:
+        raise Exception(f"Invalid name and/or version specifier {input}, expected <name><version specifiers> e.g. o3de==1.2.3")
+
+    if not match.group("object_name"):
+        raise Exception(f"Invalid or missing name {input}, expected <name><version specifiers> e.g. o3de==1.2.3")
```
lumberyard-employee-dm, line 444: We should make an specific exception class for this.
We should narrow our exception specifications to only catch the exceptions we expect.
The ones that aren't expected we want to surface loudly so that they can be fixed.
For example if the regex compile function raised an exception after an update.

scripts/o3de/o3de/utils.py
```diff
@@ -109,6 +112,22 @@ def validate_identifier(identifier: str) -> bool:
     return True
 
 
+def validate_version_specifier(version_specifier:str) -> bool:
+    try:
+        get_object_name_and_version_specifier(version_specifier)
+    except Exception as e:
+        return False
```
lumberyard-employee-dm, line 119: This is dangerous as it swallows all exceptions, even ones we didn't expect that could hide bugs in the code

scripts/o3de/o3de/utils.py
```diff
@@ -346,3 +365,97 @@ def remove_gem_duplicates(gems: list) -> list:
...
+
+
+def get_object_name_and_optional_version_specifier(input:str):
+    """
+    Returns an object name and optional version specifier 
+    :param input: The input string
+    """
+    try:
+        return get_object_name_and_version_specifier(input)
+    except Exception as e:
```
lumberyard-employee-dm, line 460: repeat: We should define our own custom exception class here.

It is as simple as deriving from [Exception](https://stackoverflow.com/a/1319675)

scripts/o3de/o3de/compatibility.py
```diff
@@ -0,0 +1,295 @@
...
+    """
+    Returns True if the project is compatible with the provided engine data.
+    :param project_path: path to the project
+    :param project_json_data: project json data dictionary
+    :param engine_json_data: engine json data dictionary
+    """
+    if not o3de_object_is_compatible_with_engine(project_json_data, engine_json_data):
+        logger.error(f'{project_json_data["project_name"]} is not known to be compatible with the '
+                        f'engine {engine_json_data["engine_name"]}{engine_json_data.get("version","")} and requires the --force parameter '
+                        'to register.')
```
lumberyard-employee-dm, line 30: What does the "registration" have to do with checking that the engine is compatible?

This part of the error message seems that it should in register.py

scripts/o3de/o3de/compatibility.py
```diff
@@ -0,0 +1,295 @@
...
+        return False
+
+    # verify project -> gem -> engine compatibility
+    enabled_gems_file = cmake.get_enabled_gem_cmake_file(project_path=project_path)
+    gem_names = cmake.get_enabled_gems(enabled_gems_file)
```
lumberyard-employee-dm, line 35: We also need to grab the list of gem names from the "gem_names" key in the project.json

scripts/o3de/o3de/compatibility.py
```diff
@@ -0,0 +1,295 @@
...
+
+    # it's much more efficient to get all gem data once than to query them by name one by one
+    gem_paths = manifest.get_all_gems(project_path)
+    gems_json_data = {}
+    for gem_path in gem_paths:
+        gem_path = pathlib.Path(gem_path).resolve()
```
lumberyard-employee-dm, line 41: The gem_paths should already be [resolved](https://github.com/o3de/o3de/blob/development/scripts/o3de/o3de/manifest.py#L219-L221) .
So the call `.resolve` isn't needed.
```suggestion
        gem_path = pathlib.Path(gem_path)
```

scripts/o3de/o3de/compatibility.py
```diff
@@ -0,0 +1,295 @@
...
+        gem_json_data = manifest.get_gem_json_data(gem_path=gem_path)
+        # only store the json data if it's a gem we're looking for and has compatibility data
+        if gem_json_data['gem_name'] in gem_names and\
+            (gem_json_data.get('compatible_engines') or gem_json_data.get('engine_api_dependencies')):
+            gems_json_data[gem_json_data['gem_name']] = gem_json_data
```
lumberyard-employee-dm, line 46: Can we give the variable `gems_json_data` a different name to avoid readability concerns with it being unrelated to the `gem_json_data` variable?

How about `gem_name_to_json_dict`?

lumberyard-employee-dm, line 46: Can extract gem_name to a local variable before this logic?
```suggestion
        candidate_gem_name = gem_json_data['gem_name']
        if candidate_gem_name in gem_names and\
            (gem_json_data.get('compatible_engines') or gem_json_data.get('engine_api_dependencies')):
            gems_json_data[candidate_gem_name] = gem_json_data
```

scripts/o3de/o3de/compatibility.py
lumberyard-employee-dm: Also can we name the `gem_names` variable `active_gem_names`.

It gets a bit confusing later on when searching for an active gem inside the list of registered gem names

scripts/o3de/o3de/compatibility.py
```diff
@@ -0,0 +1,295 @@
...
+
+    incompatible_gem_found = False
+    for gem_name in gem_names:
+        if gem_name not in gems_json_data:
+            continue
+        if not gem_engine_compatible(gems_json_data[gem_name], engine_json_data, gem_paths):
+            incompatible_gem_found = True
+            logger.error(f'{gem_name} is not known to be compatible with the '
+                            f'engine {engine_json_data["engine_name"]} and requires the --force '
+                            'parameter to register.')
```
lumberyard-employee-dm, line 56: Again this logger message about registration should be part of the register.py script and not in the compatiblity utility code

scripts/o3de/o3de/compatibility.py
```diff
@@ -0,0 +1,295 @@
...
+    # opening many gem.json files to get version information
+    gem_dependencies = gem_json_data.get('dependencies','')
+    if not gem_dependencies:
+        return True
+
+    incompatible_gem_version_specifiers = None
+
+    if not gem_paths:
+        gem_paths = manifest.get_engine_gems()
+        gem_paths.extend(manifest.get_project_gems(project_path))
```
lumberyard-employee-dm, line 79: Shouldn't this check if project is not none since it is optional?

scripts/o3de/o3de/compatibility.py
```diff
@@ -0,0 +1,295 @@
...
+        project_path, gem_dependencies, gem_paths)
+    if incompatible_gem_version_specifiers:
+        if check:
+            logger.info(f'{gem_json_data["gem_name"]} is not known to be compatible with the '
+            ' following gems used by this project and requires the --force parameter to enable.'
+            f'{incompatible_gem_version_specifiers}')
+        else:
+            logger.error(f'{gem_json_data["gem_name"]} is not known to be compatible with the '
+            ' following gems used by this project and requires the --force parameter to enable.'
+            f'{incompatible_gem_version_specifiers}')
```
lumberyard-employee-dm, line 93: So it seems like this message should be part of enable gem, since that that is where the activation code is located at.

scripts/o3de/o3de/compatibility.py
```diff
@@ -0,0 +1,295 @@
...
+        return False
+
+    engine_name = project_json_data.get('engine','')
+    if engine_name:
+        engine_json_data = manifest.get_engine_json_data(engine_name=engine_name)
+    else:
+        engine_json_data = manifest.get_engine_json_data(engine_path=manifest.get_this_engine_path())
+    if not engine_json_data:
+        logger.error(f'Failed to load engine.json data based on the engine field in project.json or detect the engine from the current folder')
+        return False
```
lumberyard-employee-dm, line 118: Actually the [get_project_engine_path](https://github.com/o3de/o3de/blob/development/scripts/o3de/o3de/manifest.py#L309-L311) should be used here to get the path to the engine the project is actually using and then manifest.get_engine_json_data should be called with that path.

scripts/o3de/o3de/compatibility.py
```diff
@@ -0,0 +1,295 @@
...
+    Returns True if the engine is compatible with the provided object's compatible_engines and engine_api_versions information.
+    :param object_json_data: the json data for the object 
+    :param engine_json_data: the json data for the engine
+    """
+    if (object_json_data.get('compatible_engines') or object_json_data.get('engine_api_dependencies')):
+        return engine_is_compatible(
+            engine_json_data['engine_name'], engine_json_data.get('version',''), 
+            object_json_data.get('compatible_engines',''), object_json_data.get('engine_api_dependencies',''),
+            engine_json_data)
+    return True
```
lumberyard-employee-dm, line 165: This can be a single return statement. Since it is only boolean checks.
```suggestion
    return object_json_data.get('compatible_engines') or object_json_data.get('engine_api_dependencies') \
        engine_is_compatible(
            engine_json_data['engine_name'], engine_json_data.get('version',''), 
            object_json_data.get('compatible_engines',''), object_json_data.get('engine_api_dependencies',''),
            engine_json_data)
```

scripts/o3de/o3de/compatibility.py
```diff
@@ -0,0 +1,295 @@
...
+
+def engine_is_compatible(engine_name:str, engine_version:str, compatible_engines:list, engine_api_version_specifiers:list, engine_json_data:dict or None) -> bool:
+    """
+    Returns True if the engine is compatible with the provided compatible_engines and engine_api_versions information.
+    If a compatible_engine entry only has an engine name, it is assumed compatible with every engine version with that name.
+    If an engine_api_version entry only has an api name, it is assumed compatible with every engine api version with that name.
+    :param engine_name: the engine name
```
lumberyard-employee-dm, line 172: Can you specify that if engine_name is `None`, the this engine containing the o3de scripts will be used?

lumberyard-employee-dm: Or better yet, make the engine_json_data not optional and remove the `engine_name` and `engine_version` parameter.

There is no need for those parameters when the json data is supplied.

scripts/o3de/o3de/compatibility.py
```diff
@@ -0,0 +1,295 @@
...
+        # assume not compatible if no engine api version information is available
+        return False
+
+    if engine_api_version_specifiers:
+        for api_version_specifier in engine_api_version_specifiers:
+            api_name, unused_version_specifiers = utils.get_object_name_and_optional_version_specifier(api_version_specifier)
+            if has_compatible_version([api_version_specifier], api_name, engine_api_versions.get(api_name,'')):
+                return True
+
+    return False
```
lumberyard-employee-dm, line 214: The if compatible_engines block shouldn't `return True` here as the engine api version might not match.

What we should do is store a boolean that indicates that the engine compatibility check has succeeded so far.
```suggestion
    is_compatible = not compatible_engines
    if compatible_engines:
        if engine_version and has_compatible_version(compatible_engines, engine_name, engine_version):
            is_compatible = True
        elif not engine_version and has_compatible_name(compatible_engines, engine_name):
            # assume an engine with no version is compatible
            is_compatible = True

    if is_compatible:
        engine_api_versions = engine_json_data.get('api_versions','')
        if engine_api_version_specifiers and not engine_api_versions:
            # assume not compatible if no engine api version information is available
            return False

        if engine_api_version_specifiers:
            for api_version_specifier in engine_api_version_specifiers:
                api_name, unused_version_specifiers = utils.get_object_name_and_optional_version_specifier(api_version_specifier)
                if has_compatible_version([api_version_specifier], api_name, engine_api_versions.get(api_name,'')):
                    return True

    return False
```

scripts/o3de/o3de/compatibility.py
AMZN-alexpete: Ah yes, I moved this logic from registery.py to here and didn't update the error logging correctly.  Will address.

AMZN-alexpete: I think I'm going to refactor this code to return incompatible objects instead of True/False so I can provide better logging higher up

scripts/o3de/o3de/compatibility.py
AMZN-alexpete: My thinking here was that if an engineer lists some engine in the `compatible_engines` list then it should be considered compatible even if there is some `api` incompatibility - so the fields could be used as an OR instead of AND.  An engine is compatible if it satisfies the `compatible_engines` OR `engine_api_dependencies` OR both.

scripts/o3de/o3de/compatibility.py
```diff
@@ -106,127 +87,104 @@ def gem_project_compatible(gem_json_data:dict, project_path:pathlib.Path, gem_pa
...
             # assume an engine with no version is compatible
-            return True
+            return set()
 
-    engine_api_versions = engine_json_data.get('api_versions','')
-    if engine_api_version_specifiers and not engine_api_versions:
-        # assume not compatible if no engine api version information is available
-        return False
+        # object is not known compatible with this engine
+        incompatible_objects.add(f"{engine_name} {engine_version} does not match any version specifiers in the list of compatible engines: {compatible_engines}")
```
lumberyard-employee-dm, line 151: Is this meant to be a string or the actual incompatible object?

scripts/o3de/o3de/enable_gem.py
```diff
@@ -202,6 +228,11 @@ def add_parser_args(parser):
     parser.add_argument('-egf', '--enabled-gem-file', type=pathlib.Path, required=False,
                         help='The cmake enabled_gem file in which the gem names are specified.'
                         'If not specified it will assume enabled_gems.cmake')
+    group = parser.add_mutually_exclusive_group(required=False)
+    group.add_argument('-f', '--force', required=False, action='store_true', default=False,
+                       help='Bypass version compatibility checks')
+    group.add_argument('-c', '--check', required=False, action='store_true', default=False,
+                       help='Checks if this gem can be enabled for the specified project, but does not actually enable.')
```
lumberyard-employee-dm, line 235: The check option is similar to the same vein of a [dry-run option](https://dpb.bitbucket.io/always-use-dry-run-options-if-possible.html)

scripts/o3de/o3de/enable_gem.py
```diff
@@ -102,6 +105,27 @@ def enable_gem_in_project(gem_name: str = None,
     # Convert each path to pathlib.Path object and filter out duplicates using dict.fromkeys
     buildable_gems = list(dict.fromkeys(map(lambda gem_path_string: pathlib.Path(gem_path_string), buildable_gems)))
 
+    # check compatibility
+    if force:
+        logger.warning(f'Bypassing version compatibility check for {gem_json_data["gem_name"]}.')
+    else:
+        incompatible_objects = compatibility.get_gem_project_incompatible_objects(gem_json_data, project_path, check=check, gem_paths=buildable_gems)
+        if check:
```
lumberyard-employee-dm, line 113: The force option warning shouldn't occur if the `check` option has been specified.

scripts/o3de/o3de/enable_gem.py
```diff
@@ -102,6 +105,27 @@ def enable_gem_in_project(gem_name: str = None,
...
+            if incompatible_objects:
+                logger.info(f'{gem_json_data["gem_name"]} is not known compatible with the '
+                    'following objects/APIs and requires the --force parameter to register:')
+                for element in incompatible_objects:
+                    logger.info(f'  {element}')
+            else:
+                logger.info(f'{gem_json_data["gem_name"]} is compatible with this project')
+            return 0
```
lumberyard-employee-dm, line 121: I say we should still return non-zero if the check option finds incompatible objects.

That way it can be used to chain operations in shell scripting.
i.e
```bash
~/o3de$ if scripts/o3de.sh enable-gem --check -gn Multiplayer; then
     # do stuff
fi
```

scripts/o3de/o3de/project_properties.py
```diff
@@ -104,24 +109,43 @@ def edit_project_props(proj_path: pathlib.Path = None,
...
-                else:
-                    logger.warning(f'{tag} not found in user_tags for removal.')
-        else:
-            logger.warning(f'user_tags property not found for removal of {delete_tags}.')
-    if replace_tags:
-        tag_list = replace_tags.split() if isinstance(replace_tags, str) else replace_tags
-        proj_json['user_tags'] = tag_list
-    # Update the gem_names field in the project.json
-    _edit_gem_names(proj_json, new_gem_names, delete_gem_names, replace_gem_names, is_optional)
+    if new_version:
```
lumberyard-employee-dm, line 112: Should we update this to use the same `if isinstance(new_version, str)` check as the gem_properties.py?

scripts/o3de/o3de/project_properties.py
```diff
@@ -104,24 +109,43 @@ def edit_project_props(proj_path: pathlib.Path = None,
...
+
+    if new_gem_names or delete_gem_names or replace_gem_names != None:
+        _edit_gem_names(proj_json, new_gem_names, delete_gem_names, replace_gem_names, is_optional)
+
+
+    def valid_specifier(version_specifier_list):
+        if version_specifier_list and not utils.validate_version_specifier_list(version_specifier_list):
+            logger.error(f'Version specifiers must be in the format <name><version specifiers>. e.g. name==1.2.3 \n {version_specifier_list}')
+            return False
+        return True
```
lumberyard-employee-dm, line 128: Twice is a trend with this valid_specifier function.
Should it be in validation.py?

scripts/o3de/o3de/project_properties.py
```diff
@@ -81,7 +79,14 @@ def edit_project_props(proj_path: pathlib.Path = None,
...
                        new_engine_name: str or list = None,
-                       is_optional: bool = False
+                       new_compatible_engines: str or list = None,
+                       delete_compatible_engines: str or list = None,
+                       replace_compatible_engines: str or list = None,
+                       new_version: str = None,
+                       is_optional: bool = False,
+                       new_engine_api_dependencies: list or str = None,
+                       delete_engine_api_dependencies: list or str = None,
+                       replace_engine_api_dependencies: list or str = None
```
lumberyard-employee-dm, line 89: I wish we had pydoc for these arguments.

@amzn-phist  Such as the user doesn't know that the `is_optional` parameter is for setting optional gem dependencies into the "gem_names" field.
The name for the parameter `is_optional` doesn't really provide context without any documentation.
The user could read the method signature and ask the question "What is optional?"

scripts/o3de/o3de/register.py
```diff
@@ -398,12 +397,41 @@ def register_gem_path(json_data: dict,
                       gem_path: pathlib.Path,
                       remove: bool = False,
                       engine_path: pathlib.Path = None,
-                      project_path:  pathlib.Path = None) -> int:
+                      project_path:  pathlib.Path = None,
+                      force: bool = False) -> int:
```
lumberyard-employee-dm, line 401: Should we add check options for the `register_gem_path` and `register_project_path` functions as well?

scripts/o3de/o3de/utils.py
```diff
@@ -346,3 +374,102 @@ def remove_gem_duplicates(gems: list) -> list:
...
+    
+    if new_values:
+        new_values = new_values.split() if isinstance(new_values, str) else new_values
+        new_values = dict(entry.split('=') for entry in new_values if '=' in entry)
+        if new_values:
+            existing_values.update(new_values)
+    
+    if remove_values:
+        remove_values = remove_values.split() if isinstance(remove_values, str) else remove_values
+        [existing_values.pop(key) for key in remove_values]
```
lumberyard-employee-dm, line 403: Shouldn't we also split the key in remove_values on the `=` character to be consistent with the replace_values and new_values list?

In case the user specifies a command such as `o3de.py edit-gem-properties --remove-compatible-engines o3de=1.0`

scripts/o3de/o3de/utils.py
```diff
@@ -346,3 +374,102 @@ def remove_gem_duplicates(gems: list) -> list:
...
+        o3de>=1.2.3
+        o3de-sdk==1.2.3,~=2.3.4
+    :param input the string in the form <name><version specifier(s)>
+
+    return an engine name and a version specifier or raises an exception if input is invalid
+    """
+
+    regex_str = r"(?P<object_name>(.*?))(?P<version_specifier>((~=|==|!=|<=|>=|<|>|===)(\s*\S+)+))"
+    regex = re.compile(r"^\s*" + regex_str + r"\s*$", re.VERBOSE | re.IGNORECASE)
+    match = regex.search(input)
```
lumberyard-employee-dm, line 452: Instead of needing to match for whitespace, around the object name, version pair, it would be better to strip it from the input line and simplify the regular expression.
Also the [VERBOSE](https://docs.python.org/3/library/re.html#re.X) option isn't needed here since the regular expression doesn't contain any comments
```suggestion
    regex = re.compile(regex_str, re.IGNORECASE)
    match = regex.search(input.strip())
```

scripts/o3de/o3de/utils.py
```diff
@@ -346,3 +374,102 @@ def remove_gem_duplicates(gems: list) -> list:
...
+class InvalidObjectNameException(Exception):
+    pass
+
+def get_object_name_and_version_specifier(input:str) -> (str, str) or None:
+    """
+    Get the object name and version specifier from a string in the form <name><version specifier(s)>
+    Valid input examples:
+        o3de>=1.2.3
+        o3de-sdk==1.2.3,~=2.3.4
+    :param input the string in the form <name><version specifier(s)>
```
lumberyard-employee-dm, line 445: Can you add a comment here indicating that the version specifier is not just the number "1.2.3" but includes the relational operator such as ">=1.2.3" is the version specifier.

scripts/o3de/tests/test_enable_gem.py
```diff
@@ -164,3 +190,152 @@ def add_gem_dependency(enable_gem_cmake_file: pathlib.Path, gem_name: str):
...
+        pytest.param('', [], {}, False, False, [], None, 'o3de-test', '', None, 0),
+        # passes when a compatible engine version is provided
+        pytest.param('1.2.3', [], {}, False, False, ['o3de-test>=1.0.0'], [], 'o3de-test', '1.0.0', {}, 0),
+        pytest.param('1.2.3', [], {}, False, False, ['o3de-test>=1.0.0, <2.3.4'], [], 'o3de-test', '2.0.1', {}, 0),
+        pytest.param('1.2.3', [], {}, False, False, ['o3de-test'], [], 'o3de-test', '', {}, 0),
+        pytest.param('1.2.3', [], {}, False, False, ['o3de-test'], [], 'o3de-test', '1.2.3', {}, 0),
+        pytest.param('1.2.3', [], {}, False, False, ['o3de-test==1.2.3'], [], 'o3de-test', '', {}, 1),
+        pytest.param('1.2.3', [], {}, False, False, ['o3de-test==1.2.3'], [], 'o3de-test', None, {}, 1),
+        # fails when no compatible engine version is found
+        pytest.param('1.2.3', [], {}, False, False, ['o3de-test<1.0.0'], [], 'o3de-test', '1.0.0', {}, 1),
```
lumberyard-employee-dm, line 208: The comment should be moved above these parameters since these are expected fail due to not finding a compatible engine version.

scripts/o3de/tests/test_enable_gem.py
```diff
@@ -164,3 +190,152 @@ def add_gem_dependency(enable_gem_cmake_file: pathlib.Path, gem_name: str):
...
+        pytest.param('1.2.3', [], {}, False, False, ['o3de-test==1.0.0'], ['api~=1.0'], 'o3de-test', '2.0.0', {'api':'1.9.2'}, 0),
+        ],
+    )
+    def test_enable_gem_checks_engine_compatibility(self, gem_version, gem_dependencies, 
+                                                    gem_names_and_versions, check, force, compatible_engines, 
+                                                    engine_api_dependencies, test_engine_name, test_engine_version, 
+                                                    test_engine_api_versions, expected_result):
+        gem_registered_with_project = True
+        gem_registered_with_engine = False
+        project_path = pathlib.PurePath('TestProject')
```
lumberyard-employee-dm, line 251: There is still the case where the gem is registered with the user's global manifest (`~/.o3de_manifest.json`) instead of a project or engine that should be checked as well.

scripts/o3de/tests/test_enable_gem.py
```diff
@@ -164,3 +190,152 @@ def add_gem_dependency(enable_gem_cmake_file: pathlib.Path, gem_name: str):
...
+                patch('o3de.manifest.get_engine_json_data', side_effect=get_engine_json_data) as get_engine_json_data_patch,\
+                patch('o3de.manifest.get_gem_json_data', side_effect=get_gem_json_data) as get_gem_json_data_patch,\
+                patch('o3de.manifest.get_project_json_data', side_effect=get_project_json_data) as get_gem_json_data_patch,\
+                patch('o3de.manifest.get_project_gems', side_effect=get_project_gems) as get_project_gems_patch,\
+                patch('o3de.manifest.get_engine_gems', side_effect=get_engine_gems) as get_engine_gems_patch,\
+                patch('o3de.cmake.add_gem_dependency', side_effect=add_gem_dependency) as add_gem_dependency_patch,\
+                patch('o3de.validation.valid_o3de_gem_json', return_value=True) as valid_gem_json_patch:
+
+            self.enable_gem.project_data.pop('gem_names', None)
+            result = enable_gem.enable_gem_in_project(gem_path=gem_path, project_path=project_path, force=force, check=check)
```
lumberyard-employee-dm, line 333: Does this work when the optional gem flag is supplied?

scripts/o3de/tests/test_register.py
```diff
@@ -248,3 +251,158 @@ def find_ancestor_dir(target_file_name: pathlib.PurePath, start_path: pathlib.Pa
...
+            pytest.param('o3de3', '2.3.4', None, { 'gem1':'' }, None, None, ['gem1'], ['o3de3==1.2.3'], None, None, None,  False, 1),
+            pytest.param('o3de4', '1.2.3', None, { 'gem1':'' }, "", "", ['gem1'], ['o3de1'], None, None, None, False, 1),
+            # passes when forced
+            pytest.param('o3de5', '1.2.3', None, { 'gem1':'' }, None, None, ['gem1'], ['o3de1'], None, None, None, True, 0),
+            # passes when compatible_engines has match
+            pytest.param('o3de6', '0.0.0', None, { 'gem1':'' }, None, None, ['gem1'], ['o3de6'], None, None, None, False, 0),
+            pytest.param('o3de7', '1.2.3', None, { 'gem1':'' }, None, None, ['gem1'], ['o3de7>=1.2.3','o3de-sdk==2.3.4'], None, None, None, False, 0),
+            # fails when uses gem that is not known compatible with version 1.2.3 
+            pytest.param('o3de8', '1.2.3', None, { 'gem1':'' }, None, None, ['gem1'], ['o3de8'], None, ['o3de==2.3.4'], None, False, 1),
+            # passes when uses gem that is known compatible with version 1.2.3
```
lumberyard-employee-dm, line 285: nitpick: extra word "uses"
```suggestion
            # fails when gem is used that is not known compatible with version 1.2.3 
            pytest.param('o3de8', '1.2.3', None, { 'gem1':'' }, None, None, ['gem1'], ['o3de8'], None, ['o3de==2.3.4'], None, False, 1),
            # passes when gem is used that is known compatible with version 1.2.3 
```

scripts/o3de/tests/test_register.py
```diff
@@ -248,3 +251,158 @@ def find_ancestor_dir(target_file_name: pathlib.PurePath, start_path: pathlib.Pa
...
+            # passes when uses gem that is known compatible with version 1.2.3 
+            pytest.param('o3de9', '1.2.3', None, { 'gem1':'' }, None, None, ['gem1'], ['o3de9'], None, ['o3de9==1.2.3'], None, False, 0),
+            # passes when compatible engine not found but compatible api found
+            pytest.param('o3de10', '1.2.3', {'api':'1.2.3'}, { 'gem1':'' }, "", "", ['gem1'], ['o3de10==2.3.4'], ['api==1.2.3'], None, None, False, 0),
+            # passes when compatible engine found and no compatible api found
+            pytest.param('o3de11', '1.2.3', {'api':'2.3.4'}, { 'gem1':'' }, "", "", ['gem1'], ['o3de11==1.2.3'], ['api==1.2.3'], None, None, False, 0),
+            # fails when no compatible engine or api found
+            pytest.param('o3de12', '1.2.3', {'api':'2.3.4'}, { 'gem1':'' }, None, None, ['gem1'], ['o3de12==3.4.5'], ['api==4.5.6'], None, None, False, 1),
+            # passes when uses gem with compatible engine and api found
+            pytest.param('o3de13', '1.2.3', {'api':'2.3.4'}, { 'gem1':'' }, None, None, ['gem1'], None, None, ['o3de13>=1.0.0'], ['api~=1.2.3'] , False, 0),
```
lumberyard-employee-dm, line 294: In this case the API is not compatible due to being a different version

scripts/o3de/o3de/register.py
```diff
@@ -408,17 +408,28 @@ def register_gem_path(json_data: dict,
...
 
-        if project_path and not compatibility.gem_project_compatible(gem_json_data, project_path):
-            return 1
+        if project_path:
+            incompatible_objects = compatibility.get_gem_project_incompatible_objects(gem_json_data, project_path)
+            if incompatible_objects: 
+                logger.error(f'{gem_json_data["gem_name"]} is not known compatible with the '
+                    'following objects/APIs and requires the --force parameter to register:')
+                for element in incompatible_objects:
+                    logger.error(f'  {element}')
```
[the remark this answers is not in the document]
AMZN-alexpete: I was using that initially, but thought a new line per comment was more human-readable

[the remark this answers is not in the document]
lumberyard-employee-dm: You can use [str.join](https://docs.python.org/3/library/stdtypes.html?highlight=setdefault#str.join) to join with a newline

scripts/o3de/o3de/compatibility.py
AMZN-alexpete: each incompatible object is a human readable string that should help the user determine why something isn't compatible.  For the engine, I chose to add a bit more help text to give the user as much information as possible.

scripts/o3de/o3de/enable_gem.py
AMZN-alexpete: I'd be fine switching the arg name to `--dry-run` if that's more standard - thoughts?

scripts/o3de/o3de/enable_gem.py
AMZN-alexpete: `check` and `force` are set as mutually exclusive (add_mutually_exclusive_group) so we shouldn't have both set at the same time.

scripts/o3de/o3de/project_properties.py
AMZN-alexpete: good catch

scripts/o3de/o3de/register.py
AMZN-alexpete: a `check`/`dry-run` option would probably be good in the future for register in general - at least for anything that may fail due to compatibility.

scripts/o3de/o3de/project_properties.py
lumberyard-employee-dm: Thanks @AMZN-alexpete

scripts/o3de/o3de/utils.py
AMZN-alexpete: a valid version specifier looks like `o3de==1.0.0` or `o3de~=1.2.3`, a single `=` would not be valid

scripts/o3de/tests/test_enable_gem.py
AMZN-alexpete: This brings up a good question on what the default behavior should be if you try to add an optional gem that is incompatible.  For now one option we could take would be to have the operation fail if you try to add an optional incompatible gem so you see the error message - you can always use `force` to enable it despite the incompatibility.  Thoughts @lumberyard-employee-dm ?

scripts/o3de/tests/test_register.py
AMZN-alexpete: This is another situation where I wanted to let the developer allow a match in `compatible_engine` or `engine_api_dependencies` - this could be a fairly common case where a gem owner wants to say their gem is compatible with an engine version or some API version - if we require both then the developer would need to be sure to include all API versions for that API from every compatible_engine entry.  In the case above they would need to add a version specifier to include the newer API version like so:
```json
'api~=1.2.3',
'api==2.3.4'
```
or similar

scripts/o3de/tests/test_enable_gem.py
lumberyard-employee-dm: That sounds fine with me for checking incompatibility.
I see the common use case as the user adding the optional gem would have the gem available on their machine, but inspecting others users to have the gem available.
Therefore checking for incompatibility when the `--is-optional-gem` flag is supplied makes sense.

Now on to clarify what I was actually asking.
I was wondering if the "gem_names" entry was actually updated correctly with the JSON object that specifies optional when checking for incompatiblity.
i.e
```json
"gem_names" : [
    "atom=1.2.3",
    {
        "name": "openxr=2.3.4",
        "optional": true
    }
]
```

scripts/o3de/o3de/enable_gem.py
lumberyard-employee-dm: Thanks for making the options mutually exclusive.

scripts/o3de/o3de/enable_gem.py
lumberyard-employee-dm: Yeah --dry-run is normally the more standard option.
`git` commands have a lot of dry run options, [rsync](https://linux.die.net/man/1/rsync) has one

scripts/o3de/tests/test_engine_properties.py
```diff
@@ -64,9 +92,18 @@ def save_o3de_manifest(new_engine_data: dict, engine_path: pathlib.Path) -> bool
...
             assert result == expected_result
-            if engine_new_name:
-                assert self.engine_json.data.get('engine_name', '') == engine_new_name
-            if engine_version:
-                assert self.engine_json.data.get('O3DEVersion', '') == engine_version
+            if result == 0:
+                if engine_new_name:
+                    assert self.engine_json.data.get('engine_name', '') == engine_new_name
+                if engine_version:
+                    assert self.engine_json.data.get('O3DEVersion', '') == engine_version
```
lumberyard-employee-dm, line 103: Do we still need to check the O3DEVersion in test since it is deprecated?

scripts/o3de/o3de/compatibility.py
lumberyard-employee-dm: So it is pretty much the error part of a structure like an `AZ::Outcome<void, AZStd::string>`

scripts/o3de/o3de/enable_gem.py
AMZN-alexpete: I went ahead and switched the argument name to `dry-run`

scripts/o3de/tests/test_enable_gem.py
AMZN-alexpete: I added unit tests to verify you get a dictionary object with `name` and `optional` when adding an optional gem

scripts/o3de/tests/test_engine_properties.py
AMZN-alexpete: I left it in for now, but will deprecate (with a warning first) then remove it when we add version info to the main `engine.json` and core `gem.json` files.

AMZN-Phil (approved):

AMZN-alexpete:
Running on Jenkins revealed an important bug - compatibility checks  were not using all gem dependencies - so if a gem depended on a gem like `Atom_RPI` the compatibility check would fail because the engine does not list that gem in the immediate list of `external_subdirectories`, you need to recurse those to pull in all registered gems and then remove any that are not in the project's dependency tree based on the gems activated in the project.  I've fixed the issue but may add a unit test or two to make sure we're testing those 2 issues if code changes are made.

Also, I re-wired some of the code to load all gem json data into a cache because continually looking up gem json data by name gets really slow (many seconds vs ~150ms + 4k using a simple cache of 137 gems)

scripts/o3de/o3de/manifest.py
```diff
@@ -357,7 +358,17 @@ def get_gem_external_subdirectories(gem_path: pathlib.Path, visited_gem_paths: l
...
+        # Use the cache 
+        if gem_path in gems_json_data_by_path:
+            gem_object = gems_json_data_by_path[gem_path]
+        else:
+            gem_object = get_gem_json_data(gem_path=gem_path)
+            # store the value even if its None so we don't open the file again
+            gems_json_data_by_path[gem_path] = gem_object
+    else:
+        gem_object = get_gem_json_data(gem_path=gem_path)
+
```
lumberyard-employee-dm, line 371: Thanks for adding the caching mechanism to the gem's external subdirectories.

AMZN-alexpete: Good idea.  At first I just tried setting the default arg of `get_gem_external_subdirectories` to be `dict()` instead of `None` but that caused the tests to fail because apparently that data is persisted across calls to the top level function which was unexpected and could be a source of subtle bugs in Project Manager, so I resorted to passing an empty `dict()` object to the function at the call sites.

lumberyard-employee-dm: Yeah, that is because python is more of a pass by name paradigm.
So setting a default value of `gems_json_data_by_path = {}` binds a reference to the dictionary of `{}` to `gems_json_data_by_path` that gets used every time the function is invoked without a parameter

scripts/o3de/o3de/manifest.py
```diff
@@ -393,12 +404,89 @@ def get_all_projects() -> list:
...
+    gem_names_to_remove = [gem_name for gem_name in gems_json_data_by_name if gem_name not in gem_names_to_keep]
+    for gem_name in gem_names_to_remove:
+        del gems_json_data_by_name[gem_name]
+
+
+def get_gems_json_data_by_name( engine_path:pathlib.Path = None, 
+                                project_path: pathlib.Path = None, 
+                                external_subdirectories: list = list(),
+                                include_manifest_gems: bool = False,
+                                include_engine_gems: bool = False) -> dict:
```
lumberyard-employee-dm, line 445: nitpick: extra space after left parenthesis

scripts/o3de/o3de/utils.py
```diff
@@ -483,3 +483,36 @@ def get_object_name_and_optional_version_specifier(input:str):
...
+    keys = list(input.keys())
+    for key in keys:
+        value = input[key]
+
+        # if the value is invalid just remove it
+        if value == None:
+            del input[key]
+            continue
+
+        # include the key we're removing?
```
lumberyard-employee-dm, line 510: What is the question mark for?

scripts/o3de/o3de/utils.py
```diff
@@ -483,3 +483,36 @@ def get_object_name_and_optional_version_specifier(input:str):
...
+
+def replace_dict_keys_with_value_key(input:dict, value_key:str, replaced_key_name:str = None):
+    """
+    Takes a dictionary of dictionaries and replaces the keys with the value of 
+    a specific value key.
+    For example, if you have a dictionary of gem_paths->gem_json_data, this function can be used
+    to convert the dictionary so the keys are gem names instead of paths (gem_name->gem_json_data)
+    :param input: A dictionary of key->value pairs where every value is a dictionary that has a value_key
+    :param value_key: The value's key to replace the current key with
+    :param replaced_key_name: (Optional) A key name under which to store the replaced key in value
```
lumberyard-employee-dm, line 496: Isn't this just a [flatten](https://stackoverflow.com/questions/6027558/flatten-nested-dictionaries-compressing-keys) function?

scripts/o3de/tests/test_enable_gem.py
```diff
@@ -150,10 +150,18 @@ def load_o3de_manifest(manifest_path: pathlib.Path = None) -> dict:
                 return json.loads(TEST_O3DE_MANIFEST_JSON_PAYLOAD)
             return None
 
+        def get_gems_json_data_by_name( engine_path:pathlib.Path = None, 
+                                        project_path: pathlib.Path = None, 
+                                        external_subdirectories: list = list(),
+                                        include_manifest_gems: bool = False,
+                                        include_engine_gems: bool = False) -> dict:
+            return {}
```
lumberyard-employee-dm, line 157: nitpick repeat: extra space between `(` and `engine_path`

scripts/o3de/tests/test_enable_gem.py
```diff
@@ -127,15 +150,28 @@ def load_o3de_manifest(manifest_path: pathlib.Path = None) -> dict:
                 return json.loads(TEST_O3DE_MANIFEST_JSON_PAYLOAD)
             return None
 
+        def get_gems_json_data_by_name( engine_path:pathlib.Path = None,
```
lumberyard-employee-dm, line 153: nitpick: extra space before `engine_path` string

scripts/o3de/tests/test_project_properties.py
```diff
@@ -37,7 +45,7 @@
 '''
 
 
-@pytest.fixture(scope='class')
+@pytest.fixture(scope='function')
```
lumberyard-employee-dm, line 48: I assuming we had to change the scope to function because subsequent test was modifying, the project json daa in a way that would cause further test to fail?

According to the pytest documentation, 'function'; is the default scope and therefore we can remove it https://docs.pytest.org/en/6.2.x/fixture.html#fixture-scopes

scripts/o3de/o3de/project_properties.py
```diff
@@ -32,18 +30,18 @@ def _edit_gem_names(proj_json: dict,
                     new_gem_names: str or list = None,
                     delete_gem_names: str or list = None,
                     replace_gem_names: str or list = None,
-                    is_optional: bool = False):
+                    is_optional_gem: bool = False):
```
tkothadev, line 33: styling question/nitpick: is it necessary to add the gem keyword here? I'm assuming that the function name _edit_gem_names provides the sufficient context.

scripts/o3de/o3de/utils.py
```diff
@@ -109,6 +112,27 @@ def validate_identifier(identifier: str) -> bool:
...
+
+def validate_version_specifier_list(version_specifiers:str or list) -> bool:
+    version_specifier_list = version_specifiers.split() if isinstance(version_specifiers, str) else version_specifiers
+    if not version_specifier_list:
+        logger.error(f'Version specifiers must be in the format <name><version specifiers>. e.g. name==1.2.3')
+        return False
+
+    for version_specifier in version_specifier_list:
+        if not validate_version_specifier(version_specifier):
+            logger.error(f'Version specifiers must be in the format <name><version specifiers>. e.g. name==1.2.3 \n {version_specifier}')
```
tkothadev, line 131: should it be `<name>==<version specifiers>`?

scripts/o3de/o3de/utils.py
AMZN-alexpete: I made it a question because it's optional, will update to remove ambiguity

scripts/o3de/o3de/utils.py
AMZN-alexpete: I don't think so, the goal is not to flatten into a single dict, but to preserve the nested dict structure and replace the key in the parent dict with a value in the child.  I couldn't think of a good name for this function.  It's basically for re-keying the parent dict on `gem_name` instead of `path`, but I made it a utility function since it's seemed like a generic thing.

scripts/o3de/tests/test_project_properties.py
AMZN-alexpete: You are correct, persisting data was causing tests to fail that shouldn't

scripts/o3de/o3de/project_properties.py
AMZN-alexpete: this was changed to be consistent with the other places `is_optional` was changed to `is_optional_gem`

scripts/o3de/o3de/utils.py
lumberyard-employee-dm: So it is re-key with value function?

lumberyard-employee-dm (approved):

scripts/o3de/o3de/compatibility.py
```diff
@@ -0,0 +1,245 @@
...
+logging.basicConfig(format=utils.LOG_FORMAT)
+
+def get_project_engine_incompatible_objects(project_path:pathlib.Path, engine_path:pathlib.Path = None) -> set:
+    """
+    Returns any incompatible objects for this project and engine.
+    :param project_path: Path to the project
+    :param engine_path: Optional path to the engine. If not specified, the current engine path is used
+    """
+    # use the specified engine path NOT necessarily engine the project is registered to
+    engine_path = engine_path if engine_path else manifest.get_this_engine_path()
```
spham-amzn, line 27: FYI: You can also do 
```suggestion
    engine_path = engine_path or manifest.get_this_engine_path()
```

scripts/o3de/o3de/compatibility.py
```diff
@@ -0,0 +1,245 @@
...
+    engine_json_data = manifest.get_engine_json_data(engine_path=engine_path)
+    if not engine_json_data:
+        logger.error(f'Failed to load engine.json data needed for compatibility check from {engine_path}. '
+            'Please verify the path is correct, the file exist and is formatted correctly.')
```
spham-amzn, line 31:
```suggestion
            'Please verify the path is correct, the file exists and is formatted correctly.')
```

scripts/o3de/o3de/compatibility.py
```diff
@@ -0,0 +1,245 @@
...
+        return set('engine.json (missing)')
+
+    project_json_data = manifest.get_project_json_data(project_path=project_path)
+    if not project_json_data:
+        logger.error(f'Failed to load project.json data needed for compatibility check from {project_path}. '
+            'Please verify the path is correct, the file exist and is formatted correctly.')
```
spham-amzn, line 37:
```suggestion
            'Please verify the path is correct, the file exists and is formatted correctly.')
```

scripts/o3de/o3de/compatibility.py
```diff
@@ -0,0 +1,245 @@
...
+    Returns any incompatible objects for this gem and project.
+    :param gem_json_data: gem json data dictionary
+    :param project_path: path to the project
+    :param all_gems_json_data: optional dictionary containing data for all gems to use in compatibility checks. 
+    If not provided, uses all gems from the manifest, engine and project.
+    """
+    # early out if this project has no assigned engine
+    engine_path = manifest.get_project_engine_path(project_path=project_path)
+    if not engine_path:
+        return set()
```
spham-amzn, line 87: Should we also warn the caller of this fact? (project has no assigned engine)

scripts/o3de/o3de/compatibility.py
```diff
@@ -0,0 +1,245 @@
...
+def get_incompatible_objects_for_engine(object_json_data:dict, engine_json_data:dict) -> set:
+    """
+    Returns any incompatible objects for this object and engine.
+    If a compatible_engine entry only has an engine name, it is assumed compatible with every engine version with that name.
+    If an engine_api_version entry only has an api name, it is assumed compatible with every engine api version with that name.
+    :param object_json_data: the json data for the object 
+    :param engine_json_data: the json data for the engine
+    """
+    compatible_engines = object_json_data.get('compatible_engines','')
+    engine_api_version_specifiers = object_json_data.get('engine_api_dependencies','')
```
spham-amzn, line 130: FYI: By default, ```get``` will return None if it can't locate from a key, so no need to set the second argument
```suggestion
    compatible_engines = object_json_data.get('compatible_engines')
    engine_api_version_specifiers = object_json_data.get('engine_api_dependencies')
```

scripts/o3de/o3de/compatibility.py
```diff
@@ -0,0 +1,245 @@
...
+    returns True if the object_name matches an entry in the name_and_version_specifier_list that has no version,
+    or if the object_name matches and the object_version is compatible with the version specifier.
+    :param name_and_version_specifier_list: a list of names and (optional)version specifiers
+    :param object_name: the object name
+    :param object_version: the object version
+    """
+    try:
+        version = Version(object_version)
+    except InvalidVersion as e:
+        return False
```
spham-amzn, line 230: If the version string cannot be parsed, we should at least log a warning and the version string that can't be parsed

scripts/o3de/o3de/enable_gem.py
```diff
@@ -96,17 +100,45 @@ def enable_gem_in_project(gem_name: str = None,
...
-    # Before adding the gem_dependency check if the project is registered in either the project or engine manifest
+    # Before adding the gem_dependency check if the gem is registered in either the project or engine manifest
     buildable_gems = manifest.get_engine_gems()
     buildable_gems.extend(manifest.get_project_gems(project_path))
     # Convert each path to pathlib.Path object and filter out duplicates using dict.fromkeys
     buildable_gems = list(dict.fromkeys(map(lambda gem_path_string: pathlib.Path(gem_path_string), buildable_gems)))
 
+    # check compatibility
+    if force:
+        logger.warning(f'Bypassing version compatibility check for {gem_json_data["gem_name"]}.')
```
spham-amzn, line 111: IMHO, if 'force' was explicitly passed in, then this should not be a warning but an info

scripts/o3de/o3de/project_properties.py
```diff
@@ -81,47 +79,89 @@ def edit_project_props(proj_path: pathlib.Path = None,
...
     if not proj_json:
         return 1
-    if new_name:
+    if isinstance(new_name, str):
         if not utils.validate_identifier(new_name):
             logger.error(f'Project name must be fewer than 64 characters, contain only alphanumeric, "_" or "-" characters, and start with a letter.  {new_name}')
             return 1
         proj_json['project_name'] = new_name
-    if new_id:
+    if isinstance(new_id, str):
```
spham-amzn, line 126: ```new_id``` (and other arguments) are tagged as ```str``` in the argument list. Just wondering why we are checking the instancetype here instead of just relying on it not being Null or empty?

AMZN-alexpete: we want to allow the user to set the id field to an empty string, and this is basically following the pattern we use in other python files allowing that.

scripts/o3de/o3de/register.py
```diff
@@ -917,8 +982,11 @@ def add_parser_args(parser):
     parser.add_argument('-r', '--remove', action='store_true', required=False,
                         default=False,
                         help='Remove entry.')
-    parser.add_argument('-f', '--force', action='store_true', default=False,
+    group = parser.add_mutually_exclusive_group(required=False)
+    group.add_argument('-f', '--force', action='store_true', default=False,
                         help='For the update of the registration field being modified.')
```
spham-amzn, line 987: No part of the change, but it looks like this was meant to read
```suggestion
                        help='Force the update of the registration field being modified.')
```

spham-amzn (approved):
Looks good, with minor comments

lumberyard-employee-dm (approved):
Looks good to me as well.

PR #13572, opened 2022-12-06
Diffuse probe grid shape translation offset support

greerdv:
## What does this PR do?
Handles shape translation offsets with diffuse probe grids. Also fixes a couple of issues:
 - the box shape UI was not updating after changes to the dimensions during validation of the dimensions in the probe grid
 - entity rotation was not being ignored for axis-aligned box shapes
 
## How was this PR tested?
Manual testing in the editor.


moraaar (approved):
LGTM but I'm not much familiarized with this code :)

Gems/LmbrCentral/Code/Source/Shape/AxisAlignedBoxShape.h
```diff
@@ -40,5 +40,8 @@ namespace LmbrCentral
 
         // AZ::TransformNotificationBus::Handler
         void OnTransformChanged(const AZ::Transform& local, const AZ::Transform& world) override;
+
+        // BoxShapeComponentRequestBus::Handler overrides ...
+        bool IsTypeAxisAligned() override;
```
hultonha, line 45: suggestion: Is it worth making this `const` (I know it's a bus call but nowadays it's legal to do this and arguably still useful - if people do want to cache something they can always fallback to `mutable`)

hultonha (approved):
LGTM but likely be useful to get a review from someone from @o3de/sig-graphics-audio

dmcdiarmid-ly (approved):

PR #13578, opened 2022-12-07
Block override editing for instantiate, create and detach prefab

forkercat:

## What does this PR do?

This PR blocks override editing for some prefab-related operations as they are currently not supported. It will show an info dialog to users and direct them to perform a prefab edit instead. But the PR even hides the options in context menu so users won't be able to do so.

Block_Dialog

Related Workflows:
1. Instantiate Prefab (including Procedural Prefab)
2. Create Prefab
3. Detach Prefab

Current State: Users can perform the operations but they are essentially prefab-editing.
New State: Users cannot perform the operations in the very first place.

## How was this PR tested?

Manually tested in editor:
- [x] Prefab edit workflows are not impacted.
- [x] Override edit workflows are affected.
- [x] Make sure it also works when a nested prefab is being focused (rather than level).

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
@@ -106,7 +106,8 @@ namespace AzToolsFramework
             }
 
             // Block creating prefab as override since it is not supported.
-            if (commonRootEntityOwningInstance.has_value() && IsOverrideEditing(commonRootEntityOwningInstance->get()))
+            if (commonRootEntityOwningInstance.has_value() &&
+                !m_prefabFocusHandler.IsOwningPrefabBeingFocused(commonRootEntityOwningInstance->get().GetContainerEntityId()))
             {
                 return AZ::Failure<AZStd::string>("Creating prefab as an override edit is currently not supported.\n"
                     "To perform a prefab edit, please first enter Prefab Edit Mode on the direct owning prefab.");
@@ -462,7 +463,7 @@ namespace AzToolsFramework
             }
 
             // Block instantiating prefab as override since it is not supported.
-            if (IsOverrideEditing(instanceToParentUnder->get()))
+            if (!m_prefabFocusHandler.IsOwningPrefabBeingFocused(instanceToParentUnder->get().GetContainerEntityId()))
             {
                 return AZ::Failure<AZStd::string>("Instantiating prefab as an override edit is currently not supported.\n"
                     "To perform a prefab edit, please first enter Prefab Edit Mode on the direct owning prefab.");
@@ -615,16 +616,6 @@ namespace AzToolsFramework
             }
 
             return false;
-        }
-
-        bool PrefabPublicHandler::IsOverrideEditing(const Instance& owningInstance) const
-        {
-            AzFramework::EntityContextId editorEntityContextId = AzFramework::EntityContextId::CreateNull();
-            EditorEntityContextRequestBus::BroadcastResult(editorEntityContextId, &EditorEntityContextRequests::GetEditorEntityContextId);
-            InstanceOptionalReference focusedInstance = m_prefabFocusHandler.GetFocusedPrefabInstance(editorEntityContextId);
-            AZ_Assert(focusedInstance.has_value(), "PrefabPublicHandler::IsOverrideEditing - The focused instance is null.");
-
-            return &(focusedInstance->get()) != &owningInstance;
         }
 
         void PrefabPublicHandler::CreateLink(
@@ -1515,7 +1506,7 @@ namespace AzToolsFramework
                 const auto parentTemplateId = parentInstance.GetTemplateId();
 
                 // Block detaching prefab as override since it is not supported.
-                if (IsOverrideEditing(parentInstance))
+                if (!m_prefabFocusHandler.IsOwningPrefabBeingFocused(parentInstance.GetContainerEntityId()))
                 {
                     return AZ::Failure(AZStd::string("Detaching prefab as an override edit is currently not supported.\n"
                         "To perform a prefab edit, please first enter Prefab Edit Mode on the direct owning prefab."));
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
@@ -619,9 +619,9 @@ namespace AzToolsFramework
 
         bool PrefabPublicHandler::IsOverrideEditing(const Instance& owningInstance) const
         {
-            auto editorEntityContextId = AzFramework::EntityContextId::CreateNull();
+            AzFramework::EntityContextId editorEntityContextId = AzFramework::EntityContextId::CreateNull();
             EditorEntityContextRequestBus::BroadcastResult(editorEntityContextId, &EditorEntityContextRequests::GetEditorEntityContextId);
-            auto focusedInstance = m_prefabFocusHandler.GetFocusedPrefabInstance(editorEntityContextId);
+            InstanceOptionalReference focusedInstance = m_prefabFocusHandler.GetFocusedPrefabInstance(editorEntityContextId);
             AZ_Assert(focusedInstance.has_value(), "PrefabPublicHandler::IsOverrideEditing - The focused instance is null.");
 
             return &(focusedInstance->get()) != &owningInstance;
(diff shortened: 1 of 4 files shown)


forkercat:
The AR fails because in `create_linear_nested_prefabs` in prefab automated test, it currently depends on "Current State" workflow mentioned in the description. Looks like it can be fixed by focusing on owning instance first before creation.

jckand-amzn (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -603,6 +617,16 @@ namespace AzToolsFramework
             return false;
         }
 
+        bool PrefabPublicHandler::IsOverrideEditing(const Instance& owningInstance) const
+        {
+            auto editorEntityContextId = AzFramework::EntityContextId::CreateNull();
```
raktalen, line 622: What is the reasoning behind the use of auto here ?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -603,6 +617,16 @@ namespace AzToolsFramework
...
+            EditorEntityContextRequestBus::BroadcastResult(editorEntityContextId, &EditorEntityContextRequests::GetEditorEntityContextId);
+            auto focusedInstance = m_prefabFocusHandler.GetFocusedPrefabInstance(editorEntityContextId);
```
raktalen, line 624: What is the reasoning behind the use of auto here ?

forkercat: imo I tended not to use `auto` unless I am seeing either of these:
1. Long type name like iterator or `AZStd::unordered_map<AZStd::string, AZStd::vector<AZ::EntityId>>`
2. The whole line is so long and not easy to read.

But for these two lines here I would change not to use `auto`. Thanks for pointing this out. I think the reason I used here is I was copying this code from elsewhere in the class to keep some style consistency.

If you check other places in the code we frequently use `auto` for 'InstanceOptionalReference' type. But yeah I agree that we may avoid this in the future and probably create a backlog task to clean this up.

raktalen: Might be worth starting right now and not adding auto's to your code, these kinds of backlog tasks have a tendency to never get prioritized if some work hasn't already been done

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -105,6 +105,13 @@ namespace AzToolsFramework
                 return AZ::Failure(findCommonRootOutcome.TakeError());
             }
 
+            // Block creating prefab as override since it is not supported.
+            if (commonRootEntityOwningInstance.has_value() && IsOverrideEditing(commonRootEntityOwningInstance->get()))
+            {
+                return AZ::Failure<AZStd::string>("Creating prefab as an override edit is currently not supported.\n"
```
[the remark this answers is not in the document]
forkercat: There are many places looking like this. I wonder how we could improve this or how it could be cached off.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.h
```diff
@@ -171,6 +171,11 @@ namespace AzToolsFramework
             bool IsCyclicalDependencyFound(
                 InstanceOptionalConstReference instance, const AZStd::unordered_set<AZ::IO::Path>& templateSourcePaths);
 
+            //! Checks if it is an override edit on the focused instance.
+            //! \param The given owning instance to be checked.
+            //! \return True if it is an override editing.
+            bool IsOverrideEditing(const Instance& owningInstance) const;
```
srikappa-amzn, line 177: IsFocusedInstance might be a better name for this and I think PrefabFocusHandler is a better place for this function.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -603,6 +617,16 @@ namespace AzToolsFramework
...
         }
 
+        bool PrefabPublicHandler::IsOverrideEditing(const Instance& owningInstance) const
+        {
+            AzFramework::EntityContextId editorEntityContextId = AzFramework::EntityContextId::CreateNull();
+            EditorEntityContextRequestBus::BroadcastResult(editorEntityContextId, &EditorEntityContextRequests::GetEditorEntityContextId);
+            InstanceOptionalReference focusedInstance = m_prefabFocusHandler.GetFocusedPrefabInstance(editorEntityContextId);
+            AZ_Assert(focusedInstance.has_value(), "PrefabPublicHandler::IsOverrideEditing - The focused instance is null.");
+
+            return &(focusedInstance->get()) != &owningInstance;
```
srikappa-amzn, line 627: If the prefab you passed is not the owning instance, it doesn't mean that it'll be an override if you modify it. The instance can be outside of the focus hierarchy too. I think if you rename this function to IsFocusedInstance(), it'll be more accurate to what you are doing here.

srikappa-amzn (changes requested):
It's good that you are blocking the calls in public handler since that'll block anyone calling these functions from the interface or the public request bus like automated tests. But we can also go a step further and not show these menu options to the user in the first place by hiding them in PrefabIntegrationManager

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -603,6 +617,16 @@ namespace AzToolsFramework
...
         }
 
+        bool PrefabPublicHandler::IsOverrideEditing(const Instance& owningInstance) const
+        {
+            auto editorEntityContextId = AzFramework::EntityContextId::CreateNull();
+            EditorEntityContextRequestBus::BroadcastResult(editorEntityContextId, &EditorEntityContextRequests::GetEditorEntityContextId);
+            auto focusedInstance = m_prefabFocusHandler.GetFocusedPrefabInstance(editorEntityContextId);
+            AZ_Assert(focusedInstance.has_value(), "PrefabPublicHandler::IsOverrideEditing - The focused instance is null.");
+
+            return &(focusedInstance->get()) != &owningInstance;
```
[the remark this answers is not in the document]
raktalen: I'd suggest looking into == overloading or writing an is equal function that compares id's. In this case a shallow compare using just a memory address may work , but in general a well defined identification and comparison schema would be preferable  over memory addresses being compared.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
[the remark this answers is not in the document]
raktalen: Even if we don't cache the Failure object itself, the string should be made a const cached precreated value, that way if the string is ever updated then the person making such a change does not have to go find every location that string is being created. Instead the change has to be made only once.

Sidenote, in general, passing a string literal could cause a lot of allocations but it's possible that AZStd::Outcome has been written in such a way that along with compiler optimizations this problem is less egregious so we can dispense with caching the Failure object.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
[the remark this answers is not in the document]
forkercat: Sounds like a good idea. yeah I'll try out these 2 options and put up a follow-up PR with some clean-ups.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
forkercat: I think I can reuse this function without creating a new API.

```cpp
bool PrefabFocusHandler::IsOwningPrefabBeingFocused(AZ::EntityId entityId) const;
```

forkercat:
Discussed offline. For blocking "Create-Prefab" in PrefabIntegrationManager, I will put up a check on all the selected entities rather than on just the first one.

This functionality might be helpful if we want to have a similar block for "Duplicate" workflow later.

srikappa-amzn (approved):

PR #13592, opened 2022-12-07
Expose Cry CVars to IConsole system

alexmontAmazon:
## What does this PR do?

## How was this PR tested?


gadams3 (approved):

lumberyard-employee-dm (approved):

Code/Editor/Controls/ConsoleSCB.cpp
```diff
@@ -1096,6 +1128,40 @@ AZ::ConsoleCommandInvokedEvent::Handler ConsoleVariableEditor::m_commandInvokedH
...
+                const bool handled =
+                    (SetCVarFromConsoleCommand<AZ::s8, CVAR_INT>(changedCVar, azConsoleCommand) ||
+                     SetCVarFromConsoleCommand<AZ::s16, CVAR_INT>(changedCVar, azConsoleCommand) ||
+                     SetCVarFromConsoleCommand<AZ::s32, CVAR_INT>(changedCVar, azConsoleCommand) ||
+                     SetCVarFromConsoleCommand<AZ::s64, CVAR_INT>(changedCVar, azConsoleCommand) ||
+                     SetCVarFromConsoleCommand<long, CVAR_INT>(changedCVar, azConsoleCommand) ||
+                     SetCVarFromConsoleCommand<unsigned long, CVAR_INT>(changedCVar, azConsoleCommand) ||
+                     SetCVarFromConsoleCommand<float, CVAR_FLOAT>(changedCVar, azConsoleCommand) ||
+                     SetCVarFromConsoleCommand<double, CVAR_FLOAT>(changedCVar, azConsoleCommand) ||
+                     SetCVarFromConsoleCommand<AZStd::string, CVAR_STRING>(changedCVar, azConsoleCommand));
```
lumberyard-employee-dm, line 1148: You should add a case for AZ::CVarFixedString as well.

Code/Editor/Controls/ConsoleSCB.cpp
```diff
@@ -1103,7 +1103,11 @@ static bool SetCVarFromConsoleCommand(ICVar* cvar, AZ::ConsoleFunctorBase* conso
         {
             if constexpr (expectedCvarType != CVAR_STRING)
             {
-                cvar->Set(AZStd::to_string(value).c_str());
+                auto stringified = AZStd::to_string(value);
+                if (!stringified.empty())
+                {
+                    cvar->Set(stringified.c_str());
+                }
```
lumberyard-employee-dm, line 1113: There is no need to convert the string to a string. It is already a string.
```suggestion
                auto stringified = AZStd::to_string(value);
                if (!stringified.empty())
                {
                    cvar->Set(value.c_str());
                }
```

alexmontAmazon: this handles the case when the cvar type is actually a string, but the expected type (and therefore the T type) was not. So, although the cvar is a string, "value" is not

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/SortAdapter.cpp
```diff
@@ -38,6 +39,69 @@ namespace AZ::DocumentPropertyEditor
...
+        Dom::Path mappedPath;
+        const auto* currNode = m_rootNode.get();
+
+        // go through each entry in the path, looking for an existing mapping at the give index
+        for (const auto& pathEntry : path)
+        {
+            if (pathEntry.IsIndex())
+            {
+                bool found = false;
+                for (auto element : AZStd::views::zip(currNode->m_indexSortedChildren, currNode->m_adapterSortedChildren))
```
lumberyard-employee-dm, line 67: You can use a structured binding here
```suggestion
                for (auto [sortIndex, adapterIndex] : AZStd::views::zip(currNode->m_indexSortedChildren, currNode->m_adapterSortedChildren))
```

Code/Editor/Controls/ConsoleSCB.cpp
alexmontAmazon: done, thanks

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/SortAdapter.cpp
alexmontAmazon: done, thanks!

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/CvarAdapter.cpp
```diff
@@ -21,10 +21,38 @@ namespace AZ::DocumentPropertyEditor
             , m_commandInvokedHandler(
                   [adapter, this](AZStd::string_view command, const AZ::ConsoleCommandContainer&, AZ::ConsoleFunctorFlags, AZ::ConsoleInvokedFrom)
                   {
-                      (void)command;
-                      if (!m_performingCommand)
+                      // look up the command being performed and rebuild its editor
```
lumberyard-employee-dm, line 24: Well technically this after the command has been performed.
You could you add that to the comment so that readers are clear that the console command has finished executing at this point.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.h
```diff
@@ -234,14 +230,15 @@ namespace AzToolsFramework
...
-            return m_rowPool;
+            return static_cast<AZ::InstancePoolManager*>(AZ::Interface<AZ::InstancePoolManagerInterface>::Get())->GetPool<DPERowWidget>();
         }
 
-        auto GetLabelPool()
+        static auto GetLabelPool()
         {
-            return m_labelPool;
+            return static_cast<AZ::InstancePoolManager*>(AZ::Interface<AZ::InstancePoolManagerInterface>::Get())
+                ->GetPool<AzQtComponents::ElidingLabel>();
```
lumberyard-employee-dm, line 241: To be honest here.
InstancePoolManagerInterface isn't actually providing an interface since it has zero functions and is useless.

This code has a smell because of the need for a static_cast to access the `AZ::InstancePoolManager*`.

The better solution is to remove the `InstancePoolManagerInterface`, default the constructor and destructor of the `InstancePoolManager` class and then make a type alias for `InstancePoolManagerRegistrar` that points to `typename AZ::Interface<InstancePoolManager>::Registrar`
This code would then just directly access the InstancePool vfrom the AZ::InstancePoolManager that is registered with the global AZ::Interface.

Here is a link to the implementation change need for the the [InstancePoolManager](https://github.com/o3de/o3de/issues/14109)

This code should then update to the following
```c++
        static auto GetRowPool()
        {
            return AZ::Interface<AZ::InstancePoolInterfaceManager>::Get()->GetPool<DPERowWidget>();
        }

        static auto GetLabelPool()
        {
            return AZ::Interface<AZ::InstancePoolInterfaceManager>::Get()->GetPool<AzQtComponents::ElidingLabel>();
        }
```

hultonha:
question: Do we really want to be exposing Cry CVars at all and moving things to use the new(er) AZ_CVAR system? I just worry this is increasing not reducing technical debt, maybe there's a better way to wrap this if we can't update it all right now. Maybe @kberg-amzn might have a suggestion? I could be wrong but just wanted to sanity check as the more of this we can ideally remove the better.

alexmontAmazon:
Hi @hultonha @kberg-amzn, I worked with @lumberyard-employee-dm to expose the Cry CVars to the new system so that we could have a unified place to query all CVar values. This allows the Editor to query the new system and still get the (unfortunately important) existing Cry CVars as well as those registered with the new system. We can always phase out the Cry vars (or transition them to the new system) one by one, but I don't think there's a better option. Would you prefer that I shelve that part of the code and have my tool query both the old and the new system? We can't call this new editor a replacement for the old system if it doesn't include the Cry vars, because it would be less functional than the old tool.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/CvarAdapter.cpp
```diff
@@ -142,7 +197,7 @@ namespace AZ::DocumentPropertyEditor
         bool BuildEditorForCvar(AdapterBuilder& builder, ConsoleFunctorBase* functor)
         {
             // Check all possible type IDs for our supported types and try to create a property editor
-            return TryBuildNumericEditor<int8_t, int16_t, int32_t, int64_t, uint8_t, uint16_t, uint32_t, uint64_t, float, double>(
+            return TryBuildNumericEditor<AZ::s8, AZ::u8, AZ::s16, AZ::u16, AZ::s32, AZ::u32, AZ::s64, AZ::u64, int64_t, size_t, float, double>(
```
lumberyard-employee-dm, line 200: `int64_t` and `size_t` will not compile on Linux.
This should be `long` and `unsigned long`  as `AZ::s64` is always `long long` and `AZ::u64` is always `unsigned long long`.

lumberyard-employee-dm:
> Hi @hultonha @kberg-amzn, I worked with @lumberyard-employee-dm to expose the Cry CVars to the new system so that we could have a unified place to query all CVar values. This allows the Editor to query the new system and still get the (unfortunately important) existing Cry CVars as well as those registered with the new system. We can always phase out the Cry vars (or transition them to the new system) one by one, but I don't think there's a better option. Would you prefer that I shelve that part of the code and have my tool query both the old and the new system? We can't call this new editor a replacement for the old system if it doesn't include the Cry vars, because it would be less functional than the old tool.

Well this is used to forward the Cry Console variable updates through the AZ Console CommandInvokedEvent to provide a simplified implementation of getting any CVar updates for the legacy and new Console Editor.

This approach allows an iterative removal of Cry Vars or updates to Cry Vars to use the `AZ_CVAR` macro, while still allowing Cry variables to be edited through Console Editor.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/CvarAdapter.cpp
alexmontAmazon: fixed, thanks!

hultonha:
> > Hi @hultonha @kberg-amzn, I worked with @lumberyard-employee-dm to expose the Cry CVars to the new system so that we could have a unified place to query all CVar values. This allows the Editor to query the new system and still get the (unfortunately important) existing Cry CVars as well as those registered with the new system. We can always phase out the Cry vars (or transition them to the new system) one by one, but I don't think there's a better option. Would you prefer that I shelve that part of the code and have my tool query both the old and the new system? We can't call this new editor a replacement for the old system if it doesn't include the Cry vars, because it would be less functional than the old tool.
> 
> Well this is used to forward the Cry Console variable updates through the AZ Console CommandInvokedEvent to provide a simplified implementation of getting any CVar updates for the legacy and new Console Editor.
> 
> This approach allows an iterative removal of Cry Vars or updates to Cry Vars to use the `AZ_CVAR` macro, while still allowing Cry variables to be edited through Console Editor.

Thanks very much for the reply @alexmontAmazon and @lumberyard-employee-dm. Okay awesome, this approach does totally make sense. I think the PR title just caught be off guard and made me panic slightly and there was unfortunately no context or information in the description which is often very useful to help frame the change without having to dig into the code to see what's going on. If this acts as a shim and actually makes old CVars easier to transition one by one (which we should definitely aim to do if we can) then that's a positive step in the right direction. Thanks for clarifying and I'd recommend in future to add additional context/details in the PR description to help avoid any confusion. Thanks!

Code/Framework/AzCore/AzCore/Console/ConsoleFunctor.h
```diff
@@ -119,6 +119,18 @@ namespace AZ
         friend class Console;
     };
 
+    template<class T, class = void>
+    struct ConsoleCommandMemberFunctorSignature
```
hultonha, line 123: question: This has nothing to do with the actual changes sorry but just to check is this `class = void` usage a SFINAE trick to be used in conjunction with the definition below? I still struggle remembering how all these template tricks work 😅

Code/Framework/AzCore/AzCore/Console/ConsoleFunctor.h
```diff
@@ -127,9 +139,13 @@ namespace AZ
     {
     public:
 
-        using MemberFunctorSignature = void(_TYPE::*)(const ConsoleCommandContainer&);
+        using MemberFunctorSignature = typename ConsoleCommandMemberFunctorSignature<_TYPE>::type;
         using RawFunctorSignature = void(*)(_TYPE&, const ConsoleCommandContainer&);
-        using FunctorUnion = AZStd::variant<RawFunctorSignature, MemberFunctorSignature>;
+
+        using FunctorUnion = AZStd::conditional_t<
```
hultonha, line 145: suggestion: Maybe a comment here would be useful to explain what this alias is doing and why it's needed

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/CvarAdapter.cpp
```diff
@@ -18,7 +18,45 @@ namespace AZ::DocumentPropertyEditor
...
+                                      // found the command at column 0, the editor must be at column 1
+                                      path.Push(rowIndex);
+                                      path.Push(1);
+                                      break;
+                                  }
+                              }
+
+                              if (!path.IsEmpty())
+                              {
+                                  // rebuild and replace the editor with one with the correct value
```
hultonha, line 49: nit: editor with one with -> editor one with

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/CvarAdapter.cpp
```diff
@@ -142,7 +197,7 @@ namespace AZ::DocumentPropertyEditor
         bool BuildEditorForCvar(AdapterBuilder& builder, ConsoleFunctorBase* functor)
         {
             // Check all possible type IDs for our supported types and try to create a property editor
-            return TryBuildNumericEditor<int8_t, int16_t, int32_t, int64_t, uint8_t, uint16_t, uint32_t, uint64_t, float, double>(
+            return TryBuildNumericEditor<AZ::s8, AZ::u8, AZ::s16, AZ::u16, AZ::s32, AZ::u32, AZ::s64, AZ::u64, long, unsigned long, float, double>(
```
hultonha, line 200: question: Is it preferred to use the AZ aliases? I thought using the standard names was actually preferred these days? (those in `<cstdint>`)

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/FilterAdapter.h
```diff
@@ -21,6 +21,10 @@ namespace AZ::DocumentPropertyEditor
 
         void SetIncludeAllMatchDescendants(bool includeAll);
 
+        // MetaAdapter overrides
```
hultonha, line 24: uber nit: For consistency missing the ellipsis at the end (`// MetaAdapter overrides ...`)

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/SortAdapter.cpp
```diff
@@ -21,7 +22,7 @@ namespace AZ::DocumentPropertyEditor
             m_sortActive = active;
             if (active)
             {
-                // TODO: This method should genenerate patches to/from source once custom patching is implemented
+                // TODO: This method should generate patches to/from source once custom patching is implemented
```
hultonha, line 25: suggestion: While we're here is it worth updating this TODO to a GHI if this issue is being tracked somewhere?

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/SortAdapter.cpp
```diff
@@ -38,6 +39,60 @@ namespace AZ::DocumentPropertyEditor
         }
     }
 
+    Dom::Path RowSortAdapter::MapPath(const Dom::Path& path, bool mapToSource)
```
hultonha, line 42: comment: Not a huge fan of `bool` params as the callsites can be cryptic and the complexity of the function immediately increases. Consider at least making this an enum if possible (you can ignore but bool params are just the worst 😅 I'm now convinced of it 😝)

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/SortAdapter.cpp
```diff
@@ -38,6 +39,60 @@ namespace AZ::DocumentPropertyEditor
...
+            else
+            {
+                // RowSortAdapter only affect index entries, pass other types (like attribute strings) through
+                mappedPath.Push(pathEntry);
+            }
+        }
+        return mappedPath;
+    }
+
+    Dom::Path RowSortAdapter::MapFromSourcePath(const Dom::Path& sourcePath)
```
hultonha, line 86: question: Could this and the function it calls be `const`? It doesn't _look_ like any member data is mutated.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/SortAdapter.cpp
```diff
@@ -60,16 +115,58 @@ namespace AZ::DocumentPropertyEditor
...
-    void RowSortAdapter::HandleDomChange([[maybe_unused]] const Dom::Patch& patch)
+    void RowSortAdapter::HandleDomChange(const Dom::Patch& patch)
     {
         if (m_sortActive)
         {
-            // TODO: handle changes upstream by making  custom patches. This is not currently
-            // possible, because a sort change is mostly move operations, and the DPE doesn't handle moves yet.
-            HandleReset();
+            bool needsReset = false;
+            //const auto& sourceContents = m_sourceAdapter->GetContents();
```
hultonha, line 126: update: Looks like this line should be removed

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/SortAdapter.cpp
```diff
@@ -60,16 +115,58 @@ namespace AZ::DocumentPropertyEditor
...
+
+            Dom::Patch outgoingPatch;
+            for (auto operationIterator = patch.begin(), endIterator = patch.end(); !needsReset && operationIterator != endIterator;
+                 ++operationIterator)
```
hultonha, line 130: suggestion: Any reason why we can't use a range-based for loop here? I see we have the additional check, but we could check `needsReset` at the start of the loop potentially? Not a huge deal it's just iterators do usually imply something more involved is going on like elements being inserted/removed

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/SortAdapter.cpp
```diff
@@ -117,11 +214,12 @@ namespace AZ::DocumentPropertyEditor
 
     Dom::Value RowSortAdapter::GetSortedValue(const Dom::Value& sourceValue, SortInfoNode* sortInfo)
     {
-        // copy the entire Dom::Value, but this is cheap, thanks to its copy-on-write semantics
+        // copy the entire Dom::Value and sort in place. Note that this is cheap, thanks to its copy-on-write semantics
         auto sortedValue = sourceValue;
 
         AZ_Assert(sortInfo->m_indexSortedChildren.size() == sortInfo->m_adapterSortedChildren.size(), "child lists must be the same size!");
 
+        // iterate children in model index order, copying from the adapter sorted value into the current position
```
hultonha, line 222: Thanks for improving and adding to these comments 👍

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/SortAdapter.cpp
```diff
@@ -132,10 +230,19 @@ namespace AZ::DocumentPropertyEditor
...
+            // do the actual copy into the current position
             sortedValue[nextIndex] = sortedChild;
         }
 
         return sortedValue;
     }
 
+    void RowSortAdapter::ResortRow(Dom::Path sortedPath)
+    {
+        // TODO once the DPE supports move operations:
```
hultonha, line 243: suggestion: Rather than adding a TODO it might be better to refer to a GHI (or Jira Id) to know how this is tracked.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ValueStringSort.cpp
```diff
@@ -38,11 +38,7 @@ namespace AZ::DocumentPropertyEditor
 
     RowSortAdapter::SortInfoNode* ValueStringSort::NewSortInfoNode() const
```
hultonha, line 39: suggestion: Might be out of scope of this PR, but as this is returning a pointer that is owning, it might be nice to update this to a `unique_ptr` to indicate the transfer of ownership and ensure people can't forget to delete it.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyColorCtrl.cpp
```diff
@@ -246,9 +246,11 @@ namespace AzToolsFramework
     {
         // Update the color but don't need to update the dialog color since
         // this signal came from the dialog
-        SetColor(color, false);
-
-        emit valueChanged(m_color);
+        if (m_color != color)
+        {
+            SetColor(color, false);
```
hultonha, line 251: comment: Why `bool` parameters suck - there's no way to know really what `false` here means without going to look at the implementation or API interface docs

Code/Legacy/CrySystem/SystemInit.cpp
```diff
@@ -617,15 +617,21 @@ class AzConsoleToCryConsoleBinder final
...
-        if (functor->GetTypeId() != AZ::TypeId::CreateNull())
-        {
-            AZ::CVarFixedString value;
-            functor->GetValue(value);
-            gEnv->pConsole->RegisterString(functor->GetName(), value.c_str(), cryFlags, functor->GetDesc(), AzConsoleToCryConsoleBinder::OnVarChanged);
-        }
-        else
+
+        // only add CVar versions if they are not already present
+        //if (!gEnv->pConsole->GetCVar(functor->GetName()))
```
hultonha, line 622: update: Should this code now get deleted?

Code/Legacy/CrySystem/XConsole.cpp
```diff
@@ -452,9 +455,21 @@ void CXConsole::RegisterVar(ICVar* pCVar, ConsoleVarFunc pChangeFunc)
     }
 
     ConsoleVariablesMapItor::value_type value = ConsoleVariablesMapItor::value_type(pCVar->GetName(), pCVar);
-
     m_mapVariables.insert(value);
 
+    auto consoleInterface = AZ::Interface<AZ::IConsole>::Get();
+    if (consoleInterface != nullptr)
```
hultonha, line 461: nit: Could reduce to...

```
if (auto consoleInterface = AZ::Interface<AZ::IConsole>::Get())
{
    ...
```

This makes `consoleInterface` harder to use incorrectly too

hultonha (approved):
Overall looks good. I have a few suggestions/comments but the general functionality seems very useful 👍

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/CvarAdapter.cpp
lumberyard-employee-dm: There is an issue with using `int64_t` and `uint64_t` having a different type on Linux(`long`, `unsigned long) vs Windows(`long long`, `unsigned long long`).
Therefore in order to support all the existing numeric types that Cry Variables uses with the old CVAR system, we need to specify the "other" `long` or `long long` type on the platform to make sure all types are accounted for.
As the AZ s64 and u64 aliases are always `long long`, `unsigned long long` on all of O3DE platforms, then only `long`, `unsigned long` needs to be accounted for.

Code/Framework/AzCore/AzCore/Console/ConsoleFunctor.h
lumberyard-employee-dm: Yeah, that could be done.
The FunctorUnion alias is avoiding the issue with two of the exact types being added as alternatives to a variant.
That causes ambiguity errors when initializing the variants without using the explicit index of the alternative to use.

So the solution is to collapse the variant to a single alternative as there isn't a need to have two alternatives in this case.

Code/Framework/AzCore/AzCore/Console/ConsoleFunctor.h
lumberyard-employee-dm: Yes, what is occurring here is that the base using the ConsoleCommandFunctorSignature template with a single parameter would first use the base template to find what the parameters should be.
i.e `ConsoleCommandMemberFunctorSignature<int>` complete type is in `ConsoleCommandMemberFunctorSignature<int, void>`.
As the template parameters were viable in the base template, specializations are then examined to see if there more specialized template. The specializations that result in an `<int, void>` parameter after argument substitution are the viable candidates. 
The specialization of struct `ConsoleCommandMemberFunctorSignature <T, AZStd::enable_if_t<AZStd::is_class_v<T>>>`, uses `std::enable_if_t` which exposes a type alias to `void`, if the condition succeeds, therefore allowing it to fulfill the requirement of `ConsoleCommandMemberFunctorSignature<T, void>` when the condition succeeds(i.e T is class).
If T is not a class, then `std::enable_if_t` does not expose a type alias, result in a substitution error removing the specialization from being a candidate.

hultonha: I see! 😅 Thank you for the detailed explanation @lumberyard-employee-dm, much appreciated (would love you to write a tutorial we could share on this at some point! 😝)

Code/Framework/AzCore/AzCore/Console/ConsoleFunctor.h
hultonha: I see interesting, yeah as it's a bit of a strange case a comment could be helpful

Code/Editor/Controls/ConsoleSCB.cpp
```diff
@@ -1113,23 +1113,93 @@ void ConsoleVariableModel::ClearModifiedRows()
...
+                     SetCVarFromConsoleCommand<AZ::s64, CVAR_INT>(changedCVar, azConsoleCommand) ||
+                     SetCVarFromConsoleCommand<AZ::u8, CVAR_INT>(changedCVar, azConsoleCommand) ||
+                     SetCVarFromConsoleCommand<AZ::u16, CVAR_INT>(changedCVar, azConsoleCommand) ||
+                     SetCVarFromConsoleCommand<AZ::u32, CVAR_INT>(changedCVar, azConsoleCommand) ||
+                     SetCVarFromConsoleCommand<AZ::u64, CVAR_INT>(changedCVar, azConsoleCommand) ||
+                     SetCVarFromConsoleCommand<long, CVAR_INT>(changedCVar, azConsoleCommand) ||
+                     SetCVarFromConsoleCommand<unsigned long, CVAR_INT>(changedCVar, azConsoleCommand) ||
+                     SetCVarFromConsoleCommand<float, CVAR_FLOAT>(changedCVar, azConsoleCommand) ||
+                     SetCVarFromConsoleCommand<double, CVAR_FLOAT>(changedCVar, azConsoleCommand)
+                     );
```
lumberyard-employee-dm, line 1194: This is missing the `bool` to `CVAR_INT` case.

lumberyard-employee-dm (approved):

Code/Legacy/CrySystem/XConsole.cpp
alexmontAmazon: Good point, but I think it's considered better style to use the [if initializer](https://www.tutorialspoint.com/cplusplus17-if-statement-with-initializer) now, so I'll use that. Thanks!

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/CvarAdapter.cpp
alexmontAmazon: thanks, fixed the comment :)

Code/Framework/AzCore/AzCore/Console/ConsoleFunctor.h
alexmontAmazon: yep, added a comment, thanks!

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/CvarAdapter.cpp
alexmontAmazon: fixed, thanks

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/FilterAdapter.h
alexmontAmazon: fixed? Well, at least changed now :P

Code/Editor/Controls/ConsoleSCB.cpp
alexmontAmazon: good catch, thanks! Fixed now

PR #13595, opened 2022-12-07
Fixing natvis for AZ::Outcome<void, *> overload

lumberyard-employee-dm:
The `m_storage->m_value` member isn't available when the value_type is void.

## How was this PR tested?

Validated the debugger watch window was able to properly visualize an `AZ::Outcome<void, fixed_string<256>`

### DIFF
--- a/Code/Framework/AzCore/Platform/Common/VisualStudio/AzCore/Natvis/azcore.natvis
+++ b/Code/Framework/AzCore/Platform/Common/VisualStudio/AzCore/Natvis/azcore.natvis
@@ -656,6 +656,18 @@
     </Expand>
   </Type>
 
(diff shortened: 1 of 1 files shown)


lumberyard-employee-dm:
> Change looks good, if it's possible to include a screenshot of what things look like in the debugger before/after that would be really cool too but not essential. Thanks!

Here is the debugger visualization `Outcome<void, *>` output before
image


Here the the debugger visualization after the changes
image

cgalvan (approved):

gadams3 (approved):

hultonha:
Awesome thanks @lumberyard-employee-dm! 🙂

PR #13597, opened 2022-12-08
Implemented MaterialEditor & Canvas menu items with ActionManager

lsemp3d:
## What does this PR do?

Migrated the Material Canvas & Material Editor main menu bar items to use the Action Manager

## How was this PR tested?

Built the editor, launched both tools with the menu option and with the key shortcut

Signed-off-by: Luis Sempé <58790905+lsemp3d@users.noreply.github.com>

### DIFF
--- a/Gems/AtomLyIntegration/CommonFeatures/Code/Source/Material/EditorMaterialSystemComponent.cpp
+++ b/Gems/AtomLyIntegration/CommonFeatures/Code/Source/Material/EditorMaterialSystemComponent.cpp
@@ -22,6 +22,9 @@
 #include <AzCore/StringFunc/StringFunc.h>
 #include <AzCore/Utils/Utils.h>
 #include <AzFramework/Application/Application.h>
+#include <AzToolsFramework/ActionManager/Action/ActionManagerInterface.h>
+#include <AzToolsFramework/ActionManager/HotKey/HotKeyManager.h>
+#include <AzToolsFramework/ActionManager/Menu/MenuManagerInterface.h>
 #include <AzToolsFramework/API/EditorAssetSystemAPI.h>
 #include <AzToolsFramework/API/ViewPaneOptions.h>
 #include <Editor/LyViewPaneNames.h>
@@ -107,6 +110,7 @@ namespace AZ
             AzToolsFramework::EditorEvents::Bus::Handler::BusConnect();
             AzToolsFramework::ToolsApplicationNotificationBus::Handler::BusConnect();
             AzFramework::AssetCatalogEventBus::Handler::BusConnect();
+            AzToolsFramework::ActionManagerRegistrationNotificationBus::Handler::BusConnect();
 
             // All material previews use the same model and lighting preset assets
             // models/sphere.azmodel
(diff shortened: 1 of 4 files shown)


Gems/AtomLyIntegration/CommonFeatures/Code/Source/Material/EditorMaterialSystemComponent.h
```diff
@@ -100,6 +102,9 @@ namespace AZ
             // AzToolsFramework::ToolsApplicationNotificationBus::Handler overrides...
             void AfterEntitySelectionChanged(const AzToolsFramework::EntityIdList& newlySelectedEntities, const AzToolsFramework::EntityIdList&) override;
 
+            // AzToolsFramework::ActionManagerRegistrationNotificationBus::Handler...
+            void OnMenuBindingHook() override;
+
             void PurgePreviews();
 
             QAction* m_openMaterialEditorAction = nullptr;
```
gadams3, line 110: Since launching these tools is using the action manager now, can the action pointers be removed?

Gems/AtomLyIntegration/CommonFeatures/Code/Source/Material/EditorMaterialSystemComponent.cpp
```diff
@@ -428,6 +395,55 @@ namespace AZ
...
+            actionProperties.m_iconPath = ":/Menu/material_editor.svg";
+            auto outcome = actionManagerInterface->RegisterAction(ActionContext, MaterialEditorAction, actionProperties, [this]()
+                {
+                    OpenMaterialEditor("");
+                });
+            AZ_Assert(outcome.IsSuccess(), "Failed to RegisterAction %s", MaterialEditorAction);
+
+            outcome = menuManagerInterface->AddActionToMenu(ToolsMenuIdentifier, MaterialEditorAction, 100);
+            AZ_Assert(outcome.IsSuccess(), "Failed to AddAction %s to Menu %s", MaterialEditorAction, ToolsMenuIdentifier);
+
```
gadams3, line 424: This probably warrants a separate discussion but it seems like everything except for RegisterAction can be moved to data files or the settings registry then loaded into all of the necessary managers from a central location. I think that would shrink the code footprint and other checks wherever actions are being registered Adding action manager utility functions that wrap all of the boilerplate might also have the same impact on code footprint.

gadams3 (approved):

Gems/AtomLyIntegration/CommonFeatures/Code/Source/Material/EditorMaterialSystemComponent.h
lsemp3d: Yes, I did in a follow up commit :) good catch!

Gems/AtomLyIntegration/CommonFeatures/Code/Source/Material/EditorMaterialSystemComponent.cpp
lsemp3d: Agreed, there is a task on the Action Manager side to centralize the global identifiers, currently they're not accessible.

AMZN-daimini (approved):

cgalvan (approved):

PR #13611, opened 2022-12-08
[Terrain] Updating terrain to use new bindless system

invertednormal:
## What does this PR do?

This PR updates the terrain system to use the new bindless system. While terrain has always used bindless, the new system considerably simplifies things by taking care of ensuring every texture that exists is available in a bindless array. This resulted in a lot of code removal, but there were a few changes worth pointing out.
- The new bindless system required texture indices to be updated from 16 to 32 bits
- There was an assert in the AZSL compiler incorrectly triggering on unbounded buffers. It has been removed.

## How was this PR tested?

Several terrains were tested to make sure they look the same. Clipmaps were also checked to make sure the bindless texture lookups also worked in compute.


Gems/Terrain/Code/Source/TerrainRenderer/TerrainDetailMaterialManager.cpp
```diff
@@ -190,39 +186,13 @@ namespace Terrain
...
-            };
-            
-            checkRemoveImage(shaderData.m_colorImageIndex);
-            checkRemoveImage(shaderData.m_normalImageIndex);
-            checkRemoveImage(shaderData.m_roughnessImageIndex);
-            checkRemoveImage(shaderData.m_metalnessImageIndex);
-            checkRemoveImage(shaderData.m_specularF0ImageIndex);
-            checkRemoveImage(shaderData.m_occlusionImageIndex);
-            checkRemoveImage(shaderData.m_heightImageIndex);
-        }
```
rgba16f, line 214: Do this still need to do a reference decrement on the images? Or am I misunderstanding this function?

Gems/Terrain/Assets/Shaders/Terrain/TerrainMacroHelpers.azsli
```diff
@@ -61,13 +61,13 @@ void SampleMacroTexture(float2 worldPosition, float2 positionDdx, float2 positio
         float2 ddx_macroUv = positionDdx / macroUvSize;
         float2 ddy_macroUv = positionDdx / macroUvSize;
 
-        if (macroMaterialData.m_colorMapId != 0xFFFF)
+        if (macroMaterialData.m_colorMapId != 0xFFFFFFFF)
```
rgba16f, line 64: nit: magic number, could we get a shader side define for InvalidIndex

Gems/Terrain/Code/Source/TerrainRenderer/TerrainDetailMaterialManager.cpp
invertednormal: The old function removed the image from the bindless array that terrain controlled, but the new system will just drop the image automatically whenever the image view dies, so terrain doesn't need to do anything.

Gems/Terrain/Code/Source/TerrainRenderer/TerrainMacroMaterialManager.h
```diff
@@ -43,7 +40,7 @@ namespace Terrain
 
     private:
 
-        static constexpr auto InvalidImageIndex = AZ::Render::BindlessImageArrayHandler::InvalidImageIndex;
+        static constexpr auto InvalidImageIndex = AZStd::numeric_limits<uint32_t>::max();
```
rgba16f, line 43: The named constant seemed better, why switch to manually calculating it?

rgba16f (approved):

Gems/Terrain/Code/Source/TerrainRenderer/TerrainMacroMaterialManager.h
rgba16f: nm, The change is more obvious in the side by side view.

invertednormal: It was just "named" because it was referring to a different place where it was actually defined, but `BindlessImageArrayHandler` no longer exists. I think using limits here is fine.

Gems/Terrain/Assets/Shaders/Terrain/TerrainMacroHelpers.azsli
invertednormal: Hm, that's probably worth changing if AR fails, otherwise I'll do it in my next PR.

mbalfour-amzn (approved):

gadams3 (approved):

PR #13613, opened 2022-12-08
Look Adjustment for Prefab Override

forkercat:

## What does this PR do?

This PR makes 2 adjustments to prefab override visualization in Entity Outliner. We discussed that these two adjustments would help users feel more consistent during override editing.
1. Override icon color: orange -> blue
2. Add the blue prefab capsule on Level prefab

**Old Look vs New Look:**

 
_(The divider was removed. See the new screenshot posted below)_

Demo video: [prefab_override_look_adjustment.mp4](https://user-images.githubusercontent.com/11157226/206564331-aba229b3-1fa0-411d-8952-6c157ae812a5.mp4)

**Changes:**
1. Update LevelRootUiHandler to display the capsule when UX flag is on (code reference: PrefabUiHandler.cpp)
2. Add constants file to manage common UI numbers
4. Small tweaks to remove gray lines between hide/lock columns (see CarPrefab in above screenshot)
   5. Update: Revert this fix for non-level rows
6. SVG change: override icon color
7. SVG change: level icon viewbox (make it smaller for better look)

@forkercat 
1. Want to know if there is other way to change the Level icon indent rather than tweaking the SVG file. Looks like it is not easy to do in code. @AMZN-daimini 
2. I feel like the divider looks weird in the video. Maybe it is because it is too close to the capsule. But would it be good to remove it as the indent already shows parent-child relationship? @rainbj 

## How was this PR tested?

Manually tested in editor.
In game mode, the hierarchy looks good and the look is unchanged.


rainbj:
Hi @forkercat  
 All the changes you have made look great! I do have one question about prefab edit mode and the capsule around the level. When editing a prefab under the level, and the level is no longer in focus (a child of the level). What visualizations do we currently have on the level capsule?  My assumption (because there is no image included on the ticket of this state) is that is would change to the dark blue and turn disabled looking like the other prefabs in the level to match the visual feedback everywhere else? See my example below

Artboard 1

As for your second question about the line. Because we want to avoid having a container box around the entire Outliner showing ownership of the level object to all of it's children. We wanted to minimize the visual queues to something simple. We wanted to use a divider line to help users distinguish the ownership clauses for level prefabs that is different from all other prefabs in any level. This will also help for drag and drop, operations. We want to show users that this line doesn't allow users to drag any entities or prefabs outside the level container. Adjusting the space or the line weight is totally doable, but we do need some visual queue here that distinguishes a level prefab from a normal prefab.

forkercat:
> All the changes you have made look great! I do have one question about prefab edit mode and the capsule around the level. When editing a prefab under the level, and the level is no longer in focus (a child of the level). What visualizations do we currently have on the level capsule? My assumption (because there is no image included on the ticket of this state) is that is would change to the dark blue and turn disabled looking like the other prefabs in the level to match the visual feedback everywhere else? See my example below

Discussed offline that this is the case. The level capsule would change to the disabled state. This can be seen in the demo video.

> Adjusting the space or the line weight is totally doable, but we do need some visual queue here that distinguishes a level prefab from a normal prefab.

Discussion in progress. Update: TBA

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/LevelRootUiHandler.h
```diff
@@ -42,8 +42,11 @@ namespace AzToolsFramework
...
 
-        static constexpr int s_levelRootBorderThickness = 1;
-        static const QColor s_levelRootBorderColor;
-        static const QString s_levelRootIconPath;
+        QString m_levelRootIconPath = QString(":/Level/level.svg");
+        QColor m_levelRootBorderColor = QColor("#656565");
+
+        QColor m_prefabCapsuleColor = QColor("#1E252F");
+        QColor m_prefabCapsuleDisabledColor = QColor("#35383C");
+        QColor m_prefabCapsuleEditColor = QColor("#4A90E2");
```
michabr, line 50: Any reason to remove the const on these?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/Constants.h
```diff
@@ -0,0 +1,22 @@
...
+namespace AzToolsFramework
+{
+    namespace PrefabUIConstants
+    {
+        static const int prefabCapsuleRadius = 6;
+        static const int prefabBorderThickness = 2;
+        static const int prefabFileNameFontSize = 10;
+        static const int prefabEditIconSize = 16;
+
+        static const int levelDividerThickness = 1;
```
michabr, line 20: Do we need these to be static if they're in a namespace?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/LevelRootUiHandler.cpp
```diff
@@ -79,16 +85,76 @@ namespace AzToolsFramework
...
+            {
+                backgroundColor = m_prefabCapsuleEditColor;
+            }
+            else if (!(option.state & QStyle::State_Enabled))
+            {
+                backgroundColor = m_prefabCapsuleDisabledColor;
+            }
+
+            QPainterPath backgroundPath;
+            backgroundPath.setFillRule(Qt::WindingFill);
```
michabr, line 113: Suggestion: Move backgroundPath further down to where it's used.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/LevelRootUiHandler.cpp
```diff
@@ -79,16 +85,76 @@ namespace AzToolsFramework
...
+
+            QRect tempRect = option.rect;
```
michabr, line 115: Suggestion: Give tempRect a more meaningful name.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/LevelRootUiHandler.cpp
```diff
@@ -79,16 +85,76 @@ namespace AzToolsFramework
...
+            tempRect.setTop(tempRect.top() + 2);
+            tempRect.setHeight(tempRect.height() - 2); // tweaking capsule height
+
+            if (isFirstColumn)
+            {
+                tempRect.setLeft(tempRect.left() + 1);
+                tempRect.setWidth(tempRect.width() + 1); // remove the gray line between columns
+            }
+
+            if (isLastColumn)
```
michabr, line 125: Can do an "else if" here. Also could move to inside the "if (isFirstColumn || isLastColumn)".

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/LevelRootUiHandler.cpp
```diff
@@ -79,16 +85,76 @@ namespace AzToolsFramework
...
+            if (isFirstColumn || isLastColumn)
+            {
+                // Rounded rect to have rounded borders on top.
+                backgroundPath.addRoundedRect(tempRect, PrefabUIConstants::prefabCapsuleRadius, PrefabUIConstants::prefabCapsuleRadius);
+
+                // Regular rect, half height, to square the opposite border
+                QRect squareRect = tempRect;
+                if (isFirstColumn)
+                {
+                    squareRect.setLeft(tempRect.left() + (tempRect.width() / 2));
```
michabr, line 140: Can use moveLeft.

michabr (approved):
Looks nice. Just left some super minor questions/comments.

forkercat:
Managed to (1) increase the row height for the level prefab and (2) decrease the capsule height. This introduces some padding between the level and divider.

forkercat:
Previously, I thought it would be good to have this new adjustment showing up only when override UX flag is on, so the blue capsule could clearly indicate that "override is on!".

But atm I would tend not to put this logic under the override ux flag. In other words, users can still see the blue capsule (with height adjustment) even the flag is off.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/LevelRootUiHandler.cpp
forkercat: Can just move it down few lines before `if (isFirstColumn || isLastColumn)`.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/LevelRootUiHandler.h
forkercat: The reason I changed it is I was trying to keep consistency between PrefabUiHandler and LevelRootUiHandler.

I was confused by two styles in two classes. We can decide one to go with.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/Constants.h
forkercat: No, it's not needed. They are implicitly static!

https://stackoverflow.com/a/177781/5284451

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/LevelRootUiHandler.cpp
```diff
@@ -79,16 +85,75 @@ namespace AzToolsFramework
...
+            QRect columnRect = option.rect;
+            columnRect.setTop(columnRect.top() + 2);
+            columnRect.setHeight(columnRect.height() - 4); // tweaking capsule height
+
+            if (isFirstColumn || isLastColumn)
+            {
+                if (isFirstColumn)
+                {
+                    columnRect.setLeft(columnRect.left() + 1);
+                    columnRect.setWidth(columnRect.width() + 1); // remove the gray line between columns
```
AMZN-daimini, line 124: The grey line between columns only happened on specific screen resolutions/environments, so I'd either specify that or remove the comment

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/LevelRootUiHandler.h
AMZN-daimini: The reason the members are not const in `PrefabUiHandler` is that the Ui Handler for `procprefabs` inherits from the prefab one and changes the colors.

AMZN-daimini (approved):

rainbj (approved):
Approved from a UX perspective. I cannot approve from a code perspective.

PR #13618, opened 2022-12-08
Asset Processor - Fix FromAssetId being set incorrectly on Job Dependencies

amzn-mike:
## What does this PR do?

Fixes FromAssetId not being set correctly for Job Dependencies

## How was this PR tested?

Updated and ran unit test.  Also verified in asset database.


AMZN-stankowi (approved):

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
```diff
@@ -4259,7 +4259,7 @@ namespace AssetProcessor
                         AssetBuilderSDK::SourceFileDependency::SourceFileDependencyType::Wildcards
                         ? SourceFileDependencyEntry::DEP_SourceLikeMatch
                         : JobDependencyType,
-                        !entry.m_sourceFileInfo.m_uuid.IsNull(),
+                        !jobDependency.m_jobDependency.m_sourceFile.m_sourceFileDependencyUUID.IsNull(),
```
lmbr-pip, line 4262: Q: Why do we set as bool when the underlying type is bit flags as AZ::u32?

https://github.com/o3de/o3de/blob/f4ed5e8efd39a4b93e8f7e1eed445dcf33f1923c/Code/Framework/AzToolsFramework/AzToolsFramework/AssetDatabase/AssetDatabaseConnection.h#L218

Code/Tools/AssetProcessor/native/tests/assetmanager/AssetProcessorManagerTest.cpp
```diff
@@ -3168,6 +3168,45 @@ TEST_F(SourceFileDependenciesTest, UpdateSourceFileDependenciesDatabase_BasicTes
     EXPECT_NE(deps.find(m_dependsOnFile2_Source.toUtf8().constData()), deps.end());
     EXPECT_NE(deps.find(m_dependsOnFile1_Job.toUtf8().constData()), deps.end());
     EXPECT_NE(deps.find(m_dependsOnFile2_Job.toUtf8().constData()), deps.end());
+
+    AssetProcessor::SourceAssetReference sourceAsset(m_absPath);
+    auto uuid = AssetUtilities::CreateSafeSourceUUIDFromName(sourceAsset.RelativePath().c_str());
+    AZStd::vector<AzToolsFramework::AssetDatabase::SourceFileDependencyEntry> dependencyEntry;
```
jackalbe, line 3174: if it would be easy to split out this test to a separate unit test then I would suggest that instead of extending this unit test function

jackalbe (approved):

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
amzn-mike: I'm not actually sure, I know the database doesn't support bools, its possible while this API was being developed it was originally a bool and then got switched since the DB doesn't support bool.  I think we could probably change this constructor to accept a bool without an issue though.

Code/Tools/AssetProcessor/native/tests/assetmanager/AssetProcessorManagerTest.cpp
amzn-mike: Sure, updated

PR #13622, opened 2022-12-09
Fix crash when focusing on instance that was deleted as an override

### LINKED ISSUE #13549: Prefab Overrides: Focusing on a deleted instance can crash the editor
**Describe the bug**

If there is a deleted prefab instance as override and we focus on it, the editor crashes.

**Assets required**

[LevelAndPrefabs.zip](https://github.com/o3de/o3de/files/10159051/LevelAndPrefabs.zip)

**Steps to reproduce**
1. Create a Car prefab
2. Create a SuperCar prefab
3. Instantiate Car prefab under SuperCar prefab
4. Create the scene with a SuperCar prefab
5. (Focusing on the Level)
6. Delete Car prefab as an override in SuperCar
7. Focus on the SuperCar, then focus on the Car

**Expected behavior**

Successfully focus on the Car prefab.

**Actual behavior**

Editor crashed.

**Screenshots/Video**

image

Note: When focusing on the SuperCar, the Car prefab should show up so we can focus on it.

**Found in Branch**

Development https://github.com/o3de/o3de/commit/2e3a0c5b21ce90f3a250395acfa5d2ed46b79e12

* Device: PC
* OS: Windows 10
* CPU: AMD Ryzen 5950x
* GPU: NVIDIA RTX 3070
* Memory: 64 GB

**Additional context**

<details>
  <summary>CLICK ME to see the call stack</summary>

```
> EditorLib.dll!rapidjson_ly::GenericValue<rapidjson_ly::UTF8<char>,rapidjson_ly::CrtAllocator>::GenericValue<rapidjson_ly::UTF8<char>,rapidjson_ly::CrtAllocator><rapidjson_ly::CrtAllocator>(const rapidjson_ly::GenericValue<rapidjson_ly::UTF8<char>,rapidjson_ly::CrtAllocator> & rhs, rapidjson_ly::CrtAllocator & allocator, bool copyConstStrings) Line 743	C++
  [Inline Frame] EditorLib.dll!rapidjson_ly::GenericValue<rapidjson_ly::UTF8<char>,rapidjson_ly::CrtAllocator>::CopyFrom(const rapidjson_ly::GenericValue<rapidjson_ly::UTF8<char>,rapidjson_ly::CrtAllocator> &) Line 981	C++
  EditorLib.dll!AzToolsFramework::Prefab::InstanceDomGenerator::UpdateContainerEntityInDomFromRoot(rapidjson_ly::GenericDocument<rapidjson_ly::UTF8<char>,rapidjson_ly::CrtAllocator,rapidjson_ly::CrtAllocator> & instanceDom, const AzToolsFramework::Prefab::Instance & instance) Line 232	C++
  EditorLib.dll!AzToolsFramework::Prefab::InstanceDomGenerator::GenerateInstanceDom(rapidjson_ly::GenericDocument<rapidjson_ly::UTF8<char>,rapidjson_ly::CrtAllocator,rapidjson_ly::CrtAllocator> & instanceDom, const AzToolsFramework::Prefab::Instance & instance) Line 119	C++
  EditorLib.dll!AzToolsFramework::Prefab::InstanceUpdateExecutor::UpdateTemplateInstancesInQueue() Line 225	C++
  [Inline Frame] EditorLib.dll!AZStd::Internal::INVOKE(void(AZ::SystemTickEvents::*)() &) Line 177	C++
  [Inline Frame] EditorLib.dll!AZStd::invoke(void(AZ::SystemTickEvents::*)() &) Line 54	C++
  [Inline Frame] EditorLib.dll!AZ::EBusEventProcessingPolicy::Call(void(AZ::SystemTickEvents::*)() &) Line 437	C++
  EditorLib.dll!AZ::Internal::EBusContainer<AZ::SystemTickEvents,AZ::SystemTickEvents,0,1>::Dispatcher<AZ::EBus<AZ::SystemTickEvents,AZ::SystemTickEvents>>::Broadcast<void (__cdecl AZ::SystemTickEvents::*)(void)>(void(AZ::SystemTickEvents::*)() && func) Line 1360	C++
  EditorLib.dll!AZ::ComponentApplication::TickSystem() Line 1487	C++
  EditorLib.dll!CCryEditApp::IdleProcessing(bool bBackgroundUpdate) Line 2339	C++
  EditorLib.dll!CCryEditApp::OnIdle(long lCount) Line 2249	C++
  EditorLib.dll!Editor::EditorQtApplication::maybeProcessIdle() Line 312	C++
  [Inline Frame] Qt5Core.dll!QtPrivate::QSlotObjectBase::call(QObject *) Line 398	C++
  Qt5Core.dll!QSingleShotTimer::timerEvent(QTimerEvent * __formal) Line 329	C++
  Qt5Core.dll!QObject::event(QEvent * e) Line 1341	C++
  Qt5Widgets.dll!QApplicationPrivate::notify_helper(QObject * receiver, QEvent * e) Line 3632	C++
  Qt5Widgets.dll!QApplication::notify(QObject * receiver, QEvent * e) Line 3584	C++
  Qt5Core.dll!QCoreApplication::notifyInternal2(QObject * receiver, QEvent * event) Line 1063	C++
  Qt5Core.dll!QEventDispatcherWin32Private::sendTimerEvent(int timerId) Line 442	C++
  Qt5Core.dll!QEventDispatcherWin32::event(QEvent * e) Line 1067	C++
  Qt5Widgets.dll!QApplicationPrivate::notify_helper(QObject * receiver, QEvent * e) Line 3632	C++
  Qt5Widgets.dll!QApplication::notify(QObject * receiver, QEvent * e) Line 3584	C++
  Qt5Core.dll!QCoreApplication::notifyInternal2(QObject * receiver, QEvent * event) Line 1063	C++
  [Inline Frame] Qt5Core.dll!QCoreApplication::sendEvent(QObject *) Line 1458	C++
  Qt5Core.dll!QCoreApplicationPrivate::sendPostedEvents(QObject * receiver, int event_type, QThreadData * data) Line 1817	C++
  [External Code]	
  Qt5Core.dll!QEventDispatcherWin32::processEvents(QFlags<enum QEventLoop::ProcessEventsFlag> flags) Line 530	C++
  [External Code]	
  [Inline Frame] Qt5Core.dll!QEventLoop::processEvents(QFlags<enum QEventLoop::ProcessEventsFlag>) Line 139	C++
  Qt5Core.dll!QEventLoop::exec(QFlags<enum QEventLoop::ProcessEventsFlag> flags) Line 232	C++
  Qt5Core.dll!QCoreApplication::exec() Line 1371	C++
  EditorLib.dll!CryEditMain(int argc, char * * argv) Line 3997	C++
  [Inline Frame] Editor.exe!AZStd::Internal::INVOKE(int(*)(int, char * *) &) Line 208	C++
  [Inline Frame] Editor.exe!AZStd::invoke(int(*)(int, char * *) &) Line 54	C++
  Editor.exe!main(int argc, char * * argv) Line 32	C++
  Editor.exe!WinMain(HINSTANCE__ * __formal, HINSTANCE__ * __formal, char * __formal, int __formal) Line 97	C++
  [External Code]	
```

</details>

michabr:

## What does this PR do?

Fixes a crash due to not finding an entity dom in the root dom because the instance was deleted as an override in root prefab.
Fixes #13549.

Notice that I also changed an assert to a warning. This needs some more investigation to see if we should remove this warning completely or handle it a different way. This warning happens when you focus on an instance and then press ESC, but the root has deleted the previously focused instance as an override.

## How was this PR tested?

Manual testing in editor.


Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceDomGenerator.cpp
```diff
@@ -200,11 +204,12 @@ namespace AzToolsFramework
...
-        void InstanceDomGenerator::UpdateContainerEntityInDomFromRoot(PrefabDom& instanceDom, const Instance& instance) const
+        void InstanceDomGenerator::UpdateContainerEntityInDomFromOwningInstance(PrefabDom& instanceDom, const Instance& instance) const
         {
             // TODO: Modifies the function so it updates the transform only.
 
-            InstanceClimbUpResult climbUpResult = PrefabInstanceUtils::ClimbUpToTargetOrRootInstance(instance, nullptr);
+            const Instance* targetInstance = instance.HasParentInstance() ? &(instance.GetParentInstance()->get()) : nullptr;
+            InstanceClimbUpResult climbUpResult = PrefabInstanceUtils::ClimbUpToTargetOrRootInstance(instance, targetInstance);
             const Instance* rootInstancePtr = climbUpResult.m_reachedInstance;
             AZStd::string relativePathFromRoot = PrefabInstanceUtils::GetRelativePathFromClimbedInstances(climbUpResult.m_climbedInstances);
```
forkercat, line 214: Now it is not necessarily to be the root instance. Would be good to rename it to `targetInstancePtr`.

forkercat (commented):
The fix looks good. Does it pass all prefab tests?

I may want to test it in editor for double check. Hopefully we are not missing something.

```cpp
const Instance* targetInstance = instance.HasParentInstance() ? &(instance.GetParentInstance()->get()) : nullptr;
```

Tagging @AMZN-daimini for visibility

forkercat:
This "seen from root" logic is needed because we're currently not doing like what Unity does to have a separate hierarchy when entering Prefab Edit Mode.

```
Level
| SuperCar
  | Car      (deleted as override on level)
```

The crash happens when we enter focus on Car. Atm of switching focus from SuperCar to Car, InstanceUpdateExecutor first updates the SuperCar template and then the Car template. Does the crash happen in the second update?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceDomGenerator.cpp
```diff
@@ -200,11 +204,12 @@ namespace AzToolsFramework
...
         }
 
-        void InstanceDomGenerator::UpdateContainerEntityInDomFromRoot(PrefabDom& instanceDom, const Instance& instance) const
+        void InstanceDomGenerator::UpdateContainerEntityInDomFromOwningInstance(PrefabDom& instanceDom, const Instance& instance) const
         {
             // TODO: Modifies the function so it updates the transform only.
 
-            InstanceClimbUpResult climbUpResult = PrefabInstanceUtils::ClimbUpToTargetOrRootInstance(instance, nullptr);
+            const Instance* targetInstance = instance.HasParentInstance() ? &(instance.GetParentInstance()->get()) : nullptr;
+            InstanceClimbUpResult climbUpResult = PrefabInstanceUtils::ClimbUpToTargetOrRootInstance(instance, targetInstance);
```
srikappa-amzn, line 212: I'm not sure I follow the reasoning behind this change here. It is my understanding that we were previously always fetching the container entity dom 'as seen' by the root dom because we want to account for all the overrides that can possibly be applied on it from any of its ancestors. If it cannot be found in the root prefab, then I'd assume that we would at least try to go 'as close' to the root while climbing up the hierarchy. But the new logic stops at the immediate parent instance.

michabr:
> This "seen from root" logic is needed because we're currently not doing like what Unity does to have a separate hierarchy when entering Prefab Edit Mode.
> 
> ```
> Level
> | SuperCar
>   | Car      (deleted as override on level)
> ```
> 
> The crash happens when we enter focus on Car. Atm of switching focus from SuperCar to Car, InstanceUpdateExecutor first updates the SuperCar template (previously focused) and then the Car template. Does the crash happen in the second update where it is calling `m_instanceDomGenerator->GenerateInstanceDom(Car)` (when it updates Car container entity's transform)?

@forkercat the crash happens in GenerateInstanceDom:
`containerEntityDom.CopyFrom(*domPathToContainerEntity.Get(rootTemplateDom), instanceDom.GetAllocator());`
because domPathToContainerEntity.Get(rootTemplateDom) is null since the root dom has deleted the Car entity as an override.

phuchau1989 (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceDomGenerator.cpp
michabr: @srikappa-amzn we can change the logic to find the closest container to the root instead of using the immediate parent. Initially I was thinking this is just to get the correct parent so it didn't matter, but if we're copying the whole transform component maybe we need to preserve the other transform values.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceDomGenerator.cpp
```diff
@@ -200,38 +202,57 @@ namespace AzToolsFramework
...
-            PrefabDomPath domPathToContainerEntity(pathToContainerEntity.c_str());
-
-            // Retrieves the container entity DOM of the root template DOM.
+            // Get the instance's container entity DOM from the top-most ancestor that has it.
+            // An ancestor DOM may not have the data if it has deleted the instance as an override
             PrefabDom containerEntityDom;
+
+            AZStd::vector<const Instance*>& climbedInstances = climbUpResult.m_climbedInstances;
+            const Instance* topInstance = climbUpResult.m_reachedInstance;
+            for (auto instanceIter = climbedInstances.crbegin(); instanceIter != climbedInstances.crend(); ++instanceIter)
```
forkercat, line 223: Just want to double check I understand it right.
```cpp
const Instance* topInstance = climbUpResult.m_reachedInstance;
```
At this line, topInstance will always be initialized as the root instance?

```cpp
for (auto instanceIter = climbedInstances.crbegin(); instanceIter != climbedInstances.crend(); 
     ++instanceIter) {}
```
It usually just stops in the first instance (i.e. root instance) and if it could not find the DOM it will end up fetching the container entity DOM from itself?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceDomGenerator.cpp
```diff
@@ -200,38 +202,57 @@ namespace AzToolsFramework
...
             {
+                // Create a path from the current top instance to the container entity of the given instance.
+                AZStd::string relativePathFromTopInstance =
+                    PrefabInstanceUtils::GetRelativePathBetweenInstances(*topInstance, instance);
+                const AZStd::string pathToContainerEntity =
+                    AZStd::string::format("%s/%s", relativePathFromTopInstance.c_str(), PrefabDomUtils::ContainerEntityName);
+                PrefabDomPath domPathToContainerEntity(pathToContainerEntity.c_str());
+                const PrefabDom& topInstanceTemplateDom =
+                    m_prefabSystemComponentInterface->FindTemplateDom((*topInstance).GetTemplateId());
```
forkercat, line 232: nit: Looks like we can use AZ::DOM::Path to dynamically update the path, so we don't need to call `PrefabInstanceUtils::GetRelativePathBetweenInstances` for multiple times. I'm okay with the current version.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceDomGenerator.cpp
```diff
@@ -200,38 +202,57 @@ namespace AzToolsFramework
...
-            containerEntityPath.Set(instanceDom, containerEntityDom, instanceDom.GetAllocator());
+            AZ_Assert(containerEntityDom.IsObject(), "Prefab - InstanceDomGenerator::UpdateContainerEntityInDomFromHighestAncestor - "
+                "Couldn't find container entity DOM in any ancestor.");
+            if (containerEntityDom.IsObject())
+            {
+                // Sets the container entity DOM in the given instance DOM.
+                const AZStd::string containerEntityName = AZStd::string::format("/%s", PrefabDomUtils::ContainerEntityName);
+                PrefabDomPath containerEntityPath(containerEntityName.c_str());
+                containerEntityPath.Set(instanceDom, containerEntityDom, instanceDom.GetAllocator());
+            }
```
forkercat, line 255: Looks like it is not optimized.

```cpp
PrefabDom containerEntityDom;   // has its own allocator
...
containerEntityDom.CopyFrom(*containerEntityDomPtr, instanceDom.GetAllocator()); // 1st copy
...
containerEntityPath.Set(instanceDom, containerEntityDom, instanceDom.GetAllocator()); // 2nd copy
```

You can probably do this:

```cpp
PrefabDom containerEntityDom(&(instanceDom.GetAllocator()));   // uses instanceDom's allocator
...
containerEntityDom.CopyFrom(*containerEntityDomPtr, containerEntityDom.GetAllocator()); // 1st copy
...
containerEntityPath.Set(instanceDom, containerEntityDom.Move()); // Move -> no copy
```

Alternatively, if the "Set" logic can be put right before `break`, we can do this:

```cpp
if (containerEntityDomPtr)
{
    PrefabDomValue containerEntityDom(*containerEntityDomPtr, 
        &(instanceDom.GetAllocator()));   // uses instanceDom's allocator -> 1st copy
    containerEntityPath.Set(instanceDom, containerEntityDom.Move()); // Move -> no copy

    // Or using one line
    containerEntityPath.Set(*containerEntityDomPtr, instanceDom.GetAllocator());  // 1st copy
}
```

Let me know if you have any question!

Reference: https://github.com/o3de/o3de/pull/13151#discussion_r1020537422

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceDomGenerator.cpp
```diff
@@ -200,38 +202,57 @@ namespace AzToolsFramework
...
         }
 
         void InstanceDomGenerator::GenerateContainerEntityDomFromRootInstance(PrefabDom& containerEntityDom, const Instance& instance) const
```
forkercat, line 258: I think we probably need to do similar changes to this function.

forkercat (commented):
The changes look good overall! I left some questions and comments.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceDomGenerator.cpp
michabr: I looked into using AZ::DOM::Path but it doesn't look like we'd gain much benefit at this time as it uses an AZStd::vector internally. How were you thinking it would be used?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceDomGenerator.cpp
michabr: @forkercat I updated this and rearranged so that code can be shared with GenerateContainerEntityDomFromRootInstance.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceDomGenerator.cpp
michabr: Yes, good call. I tested this scenario and it also crashed. I've updated the PR to cover this.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceDomGenerator.cpp
michabr: Actually after the change to share code, it will und up fetching from itself. However "UpdateContainerEntityInDom" already returns early if reached instance is itself because it doesn't have to do anything in that case.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceDomGenerator.cpp
forkercat:
```
Level                   <-- topInstance
| SuperCar
  | Car
    | Wheel
      | TireEntity
```
My thought was, in the above example, we iterate from Level to Wheel. The path would change in each iteration:
1. /Instances/Instance_[SuperCar]/Insances/Instance_[Wheel]/Entities/Entity_[TireEntity]
2. /Instances/Instance_[Wheel]/Entities/Entity_[TireEntity]
3. /Entities/Entity_[TireEntity]

In each iteration, the path can be reconstructed by:
```cpp
// AZ::Dom::Path
path = AZ::Dom::Path(path.begin() + 2, path.end());  // remove the leading 2 entries
```

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceDomGenerator.cpp
```diff
@@ -200,64 +202,102 @@ namespace AzToolsFramework
...
+                        PrefabInstanceUtils::GetRelativePathBetweenInstances(*topInstance, *bottomInstance);
+                    const AZStd::string pathToContainerEntity =
+                        AZStd::string::format("%s/%s", relativePathFromTopInstance.c_str(), PrefabDomUtils::ContainerEntityName);
+                    PrefabDomPath domPathToContainerEntity(pathToContainerEntity.c_str());
+                    const PrefabDom& topInstanceTemplateDom =
+                        m_prefabSystemComponentInterface->FindTemplateDom((*topInstance).GetTemplateId());
+                    // DOM reference can't be relied upon if the original DOM gets modified after reference creation.
+                    // This scope is added to limit their usage and ensure DOM is not modified when it is being used.
+                    {
+                        const auto* containerEntityDomInTopInstanceDom = domPathToContainerEntity.Get(topInstanceTemplateDom);
```
forkercat, line 271: nit: name can be more compact

```cpp
const PrefabDomValue* containerEntityDomFromTop = domPathToContainerEntity.Get(topInstanceTemplateDom);
```

forkercat (approved):
lgtm. Thanks for refactoring the generator!

I think at some time later we should add unit tests for this class.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceDomGenerator.h
```diff
@@ -51,16 +52,22 @@ namespace AzToolsFramework
...
 
-            //! Given an instance, it walks up to the root instance in the hierarchy and populates the container entity dom as seen
-            //! from the root template DOM.
-            //! @param[out] containerEntityDom The DOM of the container entity as seen from the root.
+            //! Given an instance and a provided DOM, populates the DOM with the container entity DOM as seen
+            //! from the template DOM of the highest ancestor that contains the data, or from the instance if it's the root.
+            //! @param[out] containerEntityDom The DOM of the container entity as seen from the highest ancestor or root.
             //! @param instance The given instance object.
-            void GenerateContainerEntityDomFromRootInstance(PrefabDom& containerEntityDom, const Instance& instance) const;
+            void GenerateContainerEntityDomFromHighestAncestorOrSelf(PrefabDom& containerEntityDom, const Instance& instance) const;
```
srikappa-amzn, line 64: nit: is the 'OrSelf' part necessary? I think it's kind of obvious and also follows consistent naming with the UpdateContainerEntityInDomFromHighestAncestor() function too

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceDomGenerator.cpp
```diff
@@ -200,64 +202,102 @@ namespace AzToolsFramework
...
+                            break;
+                        }
+                    }
+
+                    topInstance = *instanceIter;
+                }
+            }
 
-            // Retrieves the container entity DOM of the root template DOM.
+            if (!containerEntityDom.IsObject())
```
srikappa-amzn, line 283: this can be an else or 'else if' as this can only happen if it didn't enter the other if block

srikappa-amzn (approved):

srikappa-amzn:
> lgtm. Thanks for refactoring the generator!
> 
> I think at some time later we should add unit tests for this class.

+1 on the need for tests. While reviewing the code, I was running through different use cases in my mind to check that the logic is fool proof. That is an indication that we should be writing tests for logic like this where it is difficult to cover all possible cases in our mind.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceDomGenerator.h
michabr: UpdateContainerEntityInDomFromHighestAncestor does an early return if the instance is the root, so that's why I differentiated between them.

From what I gathered, these two methods are used for different purposes:
UpdateContainerEntityInDomFromHighestAncestor is used when propagating changes.
GenerateContainerEntityDomFromHighestAncestorOrSelf is used in undo/redo nodes to capture "before" entity states.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceDomGenerator.cpp
michabr: I'll add more descriptive comments, but this is not set up as an "else" in case the container entity DOM wasn't found in any of its ancestors.

Issue #13434, opened 2022-11-29
Prefab Overrides: Adding an entity as a prefab override throws a warning to the Console


jckand-amzn:
**Describe the bug**
Adding an entity as a prefab override throws a warning to the Console.

```
[Warning] (Prefab) - Source template from link with id '4' shouldn't contains key 'LinkId'
```

**Steps to reproduce**
1. Launch the Editor with prefab overrides enabled:
```
Editor.exe --regset=O3DE/Preferences/Prefabs/EnableOverridesUx=true
```
2. Create a new level.
3. Create an entity hierarchy in the level with a parent, child, and grandchild.
4. Create a prefab from the entity hierarchy.
5. Click the prefab drop-down arrow to expand the hierarchy.
6. Right-click the child entity and Create entity.

**Expected behavior**
A new entity is added as an override.

**Actual behavior**
A new entity is added as an override, however a warning throws to the Console.

**Found in Branch**
development

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
b6d77a9bb2f029008865035ec18669205a5c1272

forkercat:
Where is the warning?
```cpp
// Remove the link ids if present in the DOMs. We don't want any overrides to be created on top of linkIds because
// linkIds are not persistent and will be created dynamically when prefabs are loaded into the editor.
instanceDom.RemoveMember(PrefabDomUtils::LinkIdName);
if (sourceDom.HasMember(PrefabDomUtils::LinkIdName))
{
    AZ_Warning("Prefab", false, "Source template from link with id '%llu' shouldn't contains key 'LinkId'",
        static_cast<AZ::u64>(linkId));
    sourceDom.RemoveMember(PrefabDomUtils::LinkIdName);
}
```

The link id warning shows up because the source template DOM has a link id field. The id is shown valid but it is actually an invalid id.

forkercat:
Notice this warning in create-prefab operation as well. I will put up a similar PR to fix it.

PR #13629, opened 2022-12-09
Prefab | Remove link id in source template

forkercat:

## What does this PR do?

Fix https://github.com/o3de/o3de/issues/13434

```
[Warning] (Prefab) - Source template from link with id '4' shouldn't contains key 'LinkId'
```

The cause is described in the issue. This PR removes the invalid link id stored in source template DOM. The link id was created in prefab loader when the template is sanitized.

It is okay that we remove it as when comparing between source template DOM and linked instance DOM, we already make sure link id is removed.

## How was this PR tested?

Manually tested and the warning is gone. Passed all tests.


cgalvan (approved):

srikappa-amzn (approved):
Nice!! That warning was starting to get annoying.

PR #13630, opened 2022-12-09
Updated Google Benchmark to 1.7.0

SergeyAMZN:
## What does this PR do?

Upgrade Google Benchmark version used in O3DE to 1.7.0

## How was this PR tested?

Unit tests, running benchmarks.


Code/Framework/AzCore/AzCore/azcoretestcommon_files.cmake
```diff
@@ -7,6 +7,7 @@
 #
 
 set(FILES
+    UnitTest/DLLTestVirtualClass.h
```
lumberyard-employee-dm, line 10: This should be in `Tests/azcoretestdll_files.cmake` file.

The `azcoretestcommon_files.cmake` file is used by the AzCoreTestCommon target which is not a dependency of AzCoreTestDLL.

SergeyAMZN: Updated, thanks!

lumberyard-employee-dm (approved):

hultonha (approved):

PR #13632, opened 2022-12-09
Material pipeline now supports functors and optional shaders

santorac:
## What does this PR do?
Material pipelines can now provide their own layout of internal properties along with material functors that get included in the MaterialTypeAsset. These are used to allow material types to control shader configuration without knowing about the shaders. Using this feature, the test material types now support transparency, optional shadow casting, and double-sided rendering.

- Code changes
  - MaterialPipelineSourceData (.materialpipeline file) has a new section for defining a layout of internal material properties.
  - MaterialTypeSourceData (.materialtype file), MaterialTypeAsset, and Material can have internal material properties and a functor list for each material pipeline.
  - Added a new property connection type "InternalProperty" that can connect a main material property to an internal material property, to pass settings directly to the material pipeline(s).
  - Updated MaterialFunctor and LuaMaterialFunctor with new functions for setting internal material property values, similar to the "InternalProperty" connection.
  - Added new plumbing in MaterialFunctor and LuaMaterialFunctor to support the material pipeline context, which can only read from internal material properties and then configure shaders.
  - I removed EditContext and BehaviorContext reflection for MaterialPipelineSourceData because I realized we weren't doing this reflection for MaterialTypeSourceData either, so I'd rather not maintain this code that isn't need at this time.

image

- Data changes
  - The material pipelines now decleare internal properties
    - castShadows
    - isTransparent
    - isTintedTransparent
    - doubleSided
  - Important new files
    - The material pipelines have new shader template files (.shader.template and .azsli) for transparent shaders, but only for the standard and enhanced lighting models (BasePBR and the basic lighting model does not support transparency, there is no alpha in the surface struct). These are very similar to the forward pass shader templates. The main differences are the "FORCE_OPAQUE" flag, DepthStencilState/Depth/writeMask, GlobalTargetBlendState, and DrawList. (There's lots of duplicate configuration here, I'll refactor later after the dust settles).
    - DoubleSided.lua sets the cull mode based on the "doubleSided" internal property
    - ShaderEnable.lua enables/disables shaders based on internal properties for shadow and transparency
    - MaterialPipelineTest_HandleOpacityMode.lua connects opacity mode properties to the material pipelines.
    - MaterialPipelineTest_GeneralCommonPropertyGroup.json connects castShadows and doubleSided properties to the material pipelines.

- Fixed issues
  - Minor code cleanup in MaterialPropertySourceData from a prior PR.
  - Fixed MaterialTypeBuilder to report dependencies on .materialpipeline files even if they cannot be loaded, so they will be reprocessed when the user fixes them.
  - Fixed several issues where azsl files were using #ifdef or #ifndef instead of #if, which can give an incorrect result if you do "#define FORCE_OPAQUE 0".
  - Fixed an issue where the "BlendStates" field in .shader files were ignored by getting stomped with the default values.
  - Fixed an issue where .shader file warnings would break CreateJobs. Now CreateJobs will pass but ProcessJobs will still fail, which will give a better overall experience because dependencies are reported.
  - Fixed the issue where dependencies were not being reported on the material type's shader code file. I just updated the regex (which is already a bit hacky anyway) to treat "#define MATERIAL_TYPE_AZSLI_FILE_PATH" as an include statement for dependency tracking purposes.
  - Fixed a bug where isFrontFacting was not applied to flip normals when detail mapping is enabled.

## How was this PR tested?

- Opened each of my "MaterialPipelineTest_" material types in material editor and switched to each render pipeline, including the proof-of-concept deferred pipeline.
- Toggled the double-sided flag and saw the correct response.
- Opened MaterialPipelineTest_Standard.material in Editor.exe and edited its properties using the Material Component instance editor. Turned off the directional light so only IBL is applied, toggled clear coat off/on, toggled "Forward Pass IBL Specular", and saw similar behavior to StandardPBR. I did find a bug here, but it exists in the development branch too. See known issues.
- Atom RPI unit tests
- New unit tests
- Atom Sample Viewer full test suite vulkan
- Editor Mode Visual Feedback still works, outlining and grayscale background when editing prefabs
- Did a bunch of testing of material hot reloading in Material Editor. I did find problems, listed below.

Known issues:
- Using separate shader assets for transparent and tintedTransparent, with just a different blend state, results in compiling the same shader code twice. There are ways we can remove this redundancy later, but I don't want to use the render state overrides because of this RFC: https://github.com/o3de/sig-graphics-audio/issues/74
- MultiViewPipeline transparency doesn't work right. It looks similar regardless of whether I use StandardPBR or MaterialPipelineTest_Standard, so the problem is outside the scope of the current task.
- Material hot reload is failing in Material Editor, and this is happening on the development branch too. https://github.com/o3de/o3de/issues/13667 
- There is some inconsistency in the way IBL is applied after toggling clear coat, not introduced by this PR. See https://github.com/o3de/o3de/issues/13668


Gems/Atom/Feature/Common/Assets/Materials/Pipelines/LowEndPipeline/TintedTransparent_StandardLighting.shader.template
```diff
@@ -0,0 +1,52 @@
+{
+    "Source" : "INSERT_AZSL_PATH_HERE",
+
+    "DepthStencilState" :
+    {
+        "Depth" :
+        {
+            "Enable" : true,
+            "CompareFunc" : "GreaterEqual",
+            "writeMask": "Zero"
```
santorac, line 10: Note to reviewers: writeMask is different between the **[tinted]transparent** and **opaque** .shader.template files

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/LowEndPipeline/TintedTransparent_StandardLighting.shader.template
```diff
@@ -0,0 +1,52 @@
...
+          "type": "Vertex"
+        },
+        {
+          "name": "PixelShader",
+          "type": "Fragment"
+        }
+      ]
+    },
+
+    "DrawList" : "transparent"
```
santorac, line 50: Note to reviewers: DrawList is different between the **[tinted]transparent** and **opaque** .shader.template files

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/LowEndPipeline/TintedTransparent_StandardLighting.shader.template
```diff
@@ -0,0 +1,52 @@
...
+            {
+                "Func" : "Always",
+                "DepthFailOp" : "Keep",
+                "FailOp" : "Keep",
+                "PassOp" : "Replace"
+            }
+        }
+    },
+    
+    "GlobalTargetBlendState" :
```
santorac, line 27: Note to reviewers: GlobalTargetBlendState is different between the **[tinted]transparent** and **opaque** .shader.template files

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/LowEndPipeline/TintedTransparent_StandardLighting.shader.template
```diff
@@ -0,0 +1,52 @@
...
+                "PassOp" : "Replace"
+            }
+        }
+    },
+    
+    "GlobalTargetBlendState" : 
+    {
+        "Enable" : true,
+        "BlendSource" : "One",
+        "BlendDest" : "ColorSource1",
```
santorac, line 31: Note to reviewers: BlendDest is different between the **transparent** and **_tinted_ transparent** .shader.template files

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/LowEndPipeline/Transparent_StandardLighting.azsli
```diff
@@ -0,0 +1,42 @@
...
+ */
+ 
+#define MATERIALPIPELINE_SHADER_HAS_PIXEL_STAGE 1
+
+// TODO(MaterialPipeline): In the end, after all refactoring and rearranging of our shader code is done, some of these flags might become unnecessary and should be re-evaluated.
+// Some of these setting used to be in ShaderQualityOptions.azsli as a basic/limited solution for multiple pipelines.
+#define FORCE_IBL_IN_FORWARD_PASS       1
+#define ENABLE_AREA_LIGHT_VALIDATION    0
+#define OUTPUT_DEPTH 0
+#define FORCE_OPAQUE 0
```
santorac, line 16: Note to reviewers: FORCE_OPAQUE is different between the **[tinted]transparent** and **opaque** template azsli files.

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Material/Material.h
```diff
@@ -165,9 +173,18 @@ namespace AZ
             bool TryApplyPropertyConnectionToShaderEnable(
                 const MaterialPropertyValue & value,
                 const MaterialPropertyOutputId & connection);
```
santorac, line 175: Note to reviewers: the above TryApply functions are new too, but I forgot to remove the declarations before committing an earlier PR. The definitions in cpp are new in this PR.

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Material/Material.h
```diff
@@ -165,9 +173,18 @@ namespace AZ
...
+            bool TryApplyPropertyConnectionToInternalProperty(
+                const MaterialPropertyValue & value,
+                const MaterialPropertyOutputId & connection);
```
santorac, line 178: Note to reviewers: all of the above TryApply functions are new, not just the last one. I forgot to remove the declarations before committing an earlier PR, so they were just undefined before. The definitions in cpp are new in this PR.

jeremyong-az (approved):

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/Common/DoubleSided.lua
```diff
@@ -0,0 +1,29 @@
+--------------------------------------------------------------------------------------
+--
+-- Copyright (c) Contributors to the Open 3D Engine Project.
+-- For complete copyright and license terms please see the LICENSE at the root of this distribution.
+--
+-- SPDX-License-Identifier: Apache-2.0 OR MIT
+--
```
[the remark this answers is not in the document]
santorac: I have no idea

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/LowEndPipeline/Transparent_StandardLighting.shader.template
```diff
@@ -0,0 +1,52 @@
...
+            {
+                "Func" : "Always",
+                "DepthFailOp" : "Keep",
+                "FailOp" : "Keep",
+                "PassOp" : "Replace"
+            }
+        }
+    },
+    
+    "GlobalTargetBlendState" :
```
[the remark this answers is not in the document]
santorac: Probably not.

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/MultiView/Transparent_StandardLighting.azsli
```diff
@@ -0,0 +1,49 @@
...
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+ 
+#define MATERIALPIPELINE_SHADER_HAS_PIXEL_STAGE 1
+
+// TODO(MaterialPipeline): In the end, after all refactoring and rearranging of our shader code is done, some of these flags might become unnecessary and should be re-evaluated.
+// Some of these setting used to be in ShaderQualityOptions.azsli as a basic/limited solution for multiple pipelines.
+// TODO(MaterialPipeline), instead of UNIFIED_FORWARD_OUTPUT, fork ForwardPassOutput.azsli
+#define UNIFIED_FORWARD_OUTPUT          1
```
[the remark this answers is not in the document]
santorac: I plan to rearrange code in these template files after migrating the core material types to the material pipeline system. Right now it's difficult to clean this up because I have to maintain the old material types together with the new ones.

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderBuilderUtility.cpp
```diff
@@ -850,6 +856,8 @@ namespace AZ
 
             IncludedFilesParser::IncludedFilesParser()
             {
+                // TODO(MaterialPipeline): This regex is not sufficient to find the material type source code file, because it is included through
+                // a MATERIAL_TYPE_AZSLI_FILE_PATH macro.
                 AZStd::regex regex(R"(#\s*include\s+[<|"]([\w|/|\\|\.|\-|\:]+)[>|"])", AZStd::regex::ECMAScript);
                 m_includeRegex.swap(regex);
```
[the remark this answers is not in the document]
santorac: See my update below 1dd768d9d5b50d08435f16f53761baa9858fbf90

PR #13636, opened 2022-12-09
update sphere and capsule lights to support shape translation offset

greerdv:

## What does this PR do?
Supports shape translation offsets in sphere and capsule lights

## How was this PR tested?
Manual testing in the editor (tested light appeared in expected location, debug draw, visibility bounds, combinations of translation offsets with entity position, rotation and scale).


Gems/AtomLyIntegration/CommonFeatures/Code/Source/CoreLights/LightDelegateBase.h
```diff
@@ -82,12 +82,16 @@ namespace AZ
             // TransformNotificationBus::Handler overrides ...
             void OnTransformChanged(const Transform& local, const Transform& world) override;
 
+            LmbrCentral::ShapeComponentRequests* m_shapeBus = nullptr;
```
hultonha, line 85: suggestion: It might be an idea to leave this private and then add an accessor for it that's protected to limit derived classes being able to modify it unnecessarily

Gems/AtomLyIntegration/CommonFeatures/Code/Source/CoreLights/LightDelegateBase.h
```diff
@@ -82,12 +82,16 @@ namespace AZ
...
+
         private:
+            // Computes overall effect transform, taking shape translation offsets into account if applicable
+            AZ::Transform ComputeOverallTransform(const Transform& entityTransform);
```
hultonha, line 89: suggestion: Maybe have an overload of this with no parmeters that calls the TransformBus internally so the caller doesn't have to do that and pass it in (but if you have it cached you can use this overload which will be faster).

hultonha (commented):
LGTM, I just have a couple minor suggestions, I can add a approval after a code owner for these files is able to weigh in. Nice work!

dmcdiarmid-ly (approved):

PR #13640, opened 2022-12-10
DomPrefixTree | Update APIs of attaching subtrees

forkercat:

## What does this PR do?

This PR updates APIs of attaching/adding subtrees to a DOM prefix tree.

**Change 1:** Rename `AttachSubTree()` to `OverwritePath()` such that it clearly suggests it is "overwrite/replacing" operation
**Change 2:** Add `shouldCreateNodes` flag such that it can create missing nodes if they do not exist for the path

**Note:** `OverwritePath()` can only succeeds if the path is an exact match (including parent entries and node entry) unless `shouldCreateNodes` is on.

Currently, the existing API `AttachSubTree()` would fail if the parent entries in the path do not exist. For example:
1. AttachSubTree("/foo/bar", subtree) returns _false_ for an empty tree since parent entry 'foo' does not exist.
2. AttachSubTree("/foo/bar/baz", subtree) returns _false_ for an empty tree OR a tree with a node at "/foo".

What's Next: `OverwritePath()` with the flag on will be used in adding a prefab override subtree at non-existing path.

fyi, This is the PR that introduced `AttachSubTree()`: https://github.com/o3de/o3de/pull/13158

## How was this PR tested?

Passed tests and new unit tests.


Code/Framework/AzCore/AzCore/DOM/DomPrefixTree.inl
```diff
@@ -91,23 +91,17 @@ namespace AZ::Dom
...
         for (size_t i = 0; i < entriesToIterate; ++i)
         {
             const PathEntry& entry = path[i];
-            auto nodeIt = node->m_values.find(entry);
-            if (nodeIt == node->m_values.end())
-            {
-                return false;
-            }
-            node = &nodeIt->second;
+            node = &node->m_values[entry]; // get or create an entry in this node
```
lumberyard-employee-dm, line 101: Actually I think we should have separate API similar to std::(unordered_)map:: [insert_or_assign](https://en.cppreference.com/w/cpp/container/map/insert_or_assign) function which always updates an entry, but returns a bool indicating whether it is a new entry.

The existing function being able to not attach if there is not an existing sub-tree seems fine

Code/Framework/AzCore/AzCore/DOM/DomPrefixTree.h
```diff
@@ -129,11 +129,18 @@ namespace AZ::Dom
         //! @return The DomPrefixTree that is detached.
         DomPrefixTree<T> DetachSubTree(const Path& path);
 
-        //! Attaches a subtree provided at the node that matches the provided path. Attaching will overwrite the node at the path.
+        //! Attaches a subtree at the provided path. Attaching will overwrite the subtree at the path.
```
srikappa-amzn, line 132: nit: Probably good to also add a comment that says it'll return false if the path couldn't be reached in the tree from the root.

Code/Framework/AzCore/Tests/DOM/DomPrefixTreeTests.cpp
```diff
@@ -255,37 +255,101 @@ namespace AZ::Dom::Tests
...
+    TEST_F(DomPrefixTreeTests, AttachSubTreeAddsNewNodeAtExistingPath)
     {
         DomPrefixTree<int> tree;
         tree.SetValue(Path("/foo"), 40);
 
         DomPrefixTree<int> treeToAttach;
         treeToAttach.SetValue(Path(), 80);
         treeToAttach.SetValue(Path("/1"), 120);
-        tree.AttachSubTree(Path("/foo/0"), AZStd::move(treeToAttach));
+        bool result = tree.AttachSubTree(Path("/foo/0"), AZStd::move(treeToAttach));
```
srikappa-amzn, line 266: Doesn't this also mean that you are inserting a new node? If we are going to split the apis as 'attach' and 'attachorinsert', I feel like AttachSubTree should be called ReplaceSubTree and only be allowed to replace nodes in the tree.

Code/Framework/AzCore/Tests/DOM/DomPrefixTreeTests.cpp
```diff
@@ -255,37 +255,101 @@ namespace AZ::Dom::Tests
...
+        DomPrefixTree<int> tree;
+        tree.SetValue(Path("/foo"), 40);
+
+        DomPrefixTree<int> treeToAttach;
+        treeToAttach.SetValue(Path(), 80);
+        treeToAttach.SetValue(Path("/1"), 120);
+        bool isInserted = tree.InsertOrAttachSubTree(Path("/foo/bar"), AZStd::move(treeToAttach)); // Path '/foo' exists
+
+        // Validate that the subtree is attached and the new values is present at the paths in the original tree.
+        EXPECT_FALSE(isInserted);
```
srikappa-amzn, line 315: Building on my other comment, this should also be treated as an 'insertion' because 'bar' node didn't exist in the tree before

Code/Framework/AzCore/Tests/DOM/DomPrefixTreeTests.cpp
forkercat: Actually I feel like using "Attach" and "Insert" together is ambiguous. In the current revision, they can be used to add/replace node.

After some consideration about your suggestion, I would prefer these APIs:
1. **ReplaceNodeAtPath** - Returns true if there is a node that exists already and false if not.
2. **InsertOrReplaceNodeAtPath** - Returns true if it is an insertion and false if it is a replacement. Only insertion adds new node. Replace requires an exact match.
3. **ReplaceSubTree** - same
4. **InsertOrReplaceSubTree** - same

Replace an existing node can also replace any descendants nodes if they exist. For insertion, it is always insert a new node at the exact path that matches. Descendant nodes are either inserted or replaced.

forkercat:
**Discussed offline. Here is the update:**
1. Change "Attach" to "Move" to clearly indicate it is a move operation. It does not alter the logic for the existing API.
2. Add a new API called "MoveSubTreeAndCreateParents" that can create parent entries in the given path if they do not pre-exist.

forkercat:
**Update:**
1. Change to use `OverwritePath` name to indicate "overwrite/replace"
2. Instead of having two functions, use one and add `createParentsIfNeeded = false`

```cpp
bool OverwritePath(const Path& path, DomPrefixTree&& subtree, bool createParentsIfNeeded = false);
bool OverwritePath(const Path& path, Node&& newNode, bool createParentsIfNeeded = false);
```

lawsonamzn (approved):
the API LGTM.

Code/Framework/AzCore/AzCore/DOM/DomPrefixTree.h
```diff
@@ -161,11 +169,18 @@ namespace AZ::Dom
...
-        //! @param path The path which corresponds to the node at which the subtree should be attached.
-        //! @param node The node to be attached.
-        //! @return True if the node was attached successfully. Return false otherwise.
-        bool AttachNodeAtPath(const Path& path, Node&& node);
+        //! Moves a node to the provided path. It will overwrite the node at the path and children nodes.
+        //! The node would be moved only if all parent path entries of the provided path already exist.
+        //! @param path The path which corresponds to where the node should be moved to.
+        //! @param nodeToMove The node to move.
+        //! @return True if the node was moved successfully. Return false otherwise.
+        bool MoveNodeAtPath(const Path& path, Node&& nodeToMove);
```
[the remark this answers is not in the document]
forkercat: Oh this is from the previous version. I replaced them with OverwritePath.

lumberyard-employee-dm (approved):
I am good with the API changes as well.

The test are good as well

Code/Framework/AzCore/AzCore/DOM/DomPrefixTree.inl
```diff
@@ -91,22 +91,21 @@ namespace AZ::Dom
...
+            if (!createParentsIfNeeded && node->m_values.find(entry) == node->m_values.end())
             {
                 return false;
             }
-            node = &nodeIt->second;
+            node = &node->m_values[entry]; // get or create an entry in this node
         }
 
-        node->m_values[path[path.Size() - 1]] = AZStd::move(nodeToAttach);
+        node->m_values[path[path.Size() - 1]] = AZStd::move(newNode);
```
srikappa-amzn, line 108: This can also create a new node when 'createParentsIfNeeded' is false. That can be confusing imo. We can't overwrite something that didn't exist in the first place. While we are here cleaning up this logic, might as well cover all bases.

I recommend returning false here if node->m_values[path[path.Size() - 1]]  is empty.

Code/Framework/AzCore/AzCore/DOM/DomPrefixTree.h
```diff
@@ -129,11 +129,14 @@ namespace AZ::Dom
...
-        //! @param subTree The subtree to attach at the provided path.
-        //! @return True if the subTree was attached successfully. Return false otherwise.
-        bool AttachSubTree(const Path& path, DomPrefixTree&& subTree);
+        //! Overwrites the subtree at the path and replaces it with the given subtree.
+        //! Old node and its children at the path would be destroyed. The new subtree will be moved to the path.
+        //! @param path The path which corresponds to the root node to which the subtree would be overwritten.
+        //! @param subtree The subtree to move to the provided path.
+        //! @param createParentsIfNeeded Create parent entries for the provided path if the flag is on.
+        //! @return True if the subtree was overwritten successfully, false if createParentsIfNeeded is off and no target path exists.
+        bool OverwritePath(const Path& path, DomPrefixTree&& subtree, bool createParentsIfNeeded = false);
```
srikappa-amzn, line 138: nit : can make the bool name compact by calling it 'shouldCreateParents' or 'createParents'

Code/Framework/AzCore/Tests/DOM/DomPrefixTreeTests.cpp
```diff
@@ -255,35 +255,67 @@ namespace AZ::Dom::Tests
...
-        treeToAttach.SetValue(Path("/1"), 120);
-        tree.AttachSubTree(Path("/foo/0"), AZStd::move(treeToAttach));
+        DomPrefixTree<int> subtree;
+        subtree.SetValue(Path(), 80);
+        subtree.SetValue(Path("/1"), 120);
+        bool isOverwritten = tree.OverwritePath(Path("/foo/0"), AZStd::move(subtree));
 
-        // Validate that the new values is present at the paths in the original tree.
+        // Validate that the subtree is overwritten and new values are present at the paths in the original tree.
+        EXPECT_TRUE(isOverwritten);
```
srikappa-amzn, line 269: This is the use case I'm referring to. 'foo/0' didn't exist in the the tree before. So overwriting the path and returning true for this function would be confusing.

forkercat:
**Update:**
1. Returns true only when the path matches exactly (parent entries + node entry) or if `shouldCreateNodes` is true
2. Add two new tests to cover that

lawsonamzn (approved):
LGTM at this point.  At some point we gotta get it in and try it out and if it needs further changes they can be made.

lumberyard-employee-dm (approved):

Code/Framework/AzCore/AzCore/DOM/DomPrefixTree.h
```diff
@@ -133,9 +133,9 @@ namespace AZ::Dom
         //! Old node and its children at the path would be destroyed. The new subtree will be moved to the path.
         //! @param path The path which corresponds to the root node to which the subtree would be overwritten.
         //! @param subtree The subtree to move to the provided path.
-        //! @param createParentsIfNeeded Create parent entries for the provided path if the flag is on.
-        //! @return True if the subtree was overwritten successfully, false if createParentsIfNeeded is off and no target path exists.
-        bool OverwritePath(const Path& path, DomPrefixTree&& subtree, bool createParentsIfNeeded = false);
+        //! @param shouldCreateNodes Create all nodes for the provided path if the flag is on.
```
srikappa-amzn, line 136: nit: We will be creating nodes only if they are missing. Would be good to clarify that in the comments

srikappa-amzn (approved):

lawsonamzn (approved):

forkercat:
Will check in when dev is green next week

PR #13642, opened 2022-12-10
Introduction of GameplayControllerComponent test CharacterController_GamePlayComponent_GravitySetsWhileMoving

FuzzyCarterAWS:
## What does this PR do?
Introduces unit test CharacterController_GamePlayComponent_GravitySetsWhileMoving to interact with the GameplayControllerComponent

## How was this PR tested?

Unit Tests passed locally

```
[----------] 1 test from PhysXDefaultWorldTest
[ RUN      ] PhysXDefaultWorldTest.CharacterController_GamePlayComponent_GravitySetsWhileMoving
[       OK ] PhysXDefaultWorldTest.CharacterController_GamePlayComponent_GravitySetsWhileMoving (2 ms)
[----------] 1 test from PhysXDefaultWorldTest (3 ms total)

[==========] 1 test from 1 test case ran. (156 ms total)
[  PASSED  ] 1 test.
```


Gems/PhysX/Code/Source/PhysXCharacters/Components/CharacterGameplayComponent.cpp
```diff
@@ -66,9 +66,9 @@ namespace PhysX
     {
     }
 
-    CharacterGameplayComponent::CharacterGameplayComponent(const CharacterGameplayConfiguration& config)
-        : m_gravityMultiplier(config.m_gravityMultiplier)
-        , m_groundDetectionBoxHeight(config.m_groundDetectionBoxHeight)
+    CharacterGameplayComponent::CharacterGameplayComponent(AZStd::unique_ptr<CharacterGameplayConfiguration> config)
+        : m_gravityMultiplier(config->m_gravityMultiplier)
+        , m_groundDetectionBoxHeight(config->m_groundDetectionBoxHeight)
```
hultonha, line 71: update: I don't think we want to update this interface if we can help it, passing a `unique_pointer` makes less sense to me here. What might be better is adding a `const` accessor to `CharacterGameplayComponent` to retrieve the current configuration if you need to check its state.

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
```diff
@@ -159,6 +159,115 @@ namespace PhysX
...
+        AzFramework::TransformComponent* transform = nullptr;
+
+        float expected_gravity = 1.5f;
+
+        // Create character
+        auto gameplayEntity = AZStd::make_unique<AZ::Entity>("GameplayEntity");
+        {
+            auto characterConfiguration = AZStd::make_unique<Physics::CharacterConfiguration>();
+            auto characterShapeConfiguration = AZStd::make_unique<Physics::CapsuleShapeConfiguration>();
+            auto characterGameplayConfiguration = AZStd::make_unique<PhysX::CharacterGameplayConfiguration>();
```
hultonha, line 175: update: These don't need to be pointers, they can all just be created on the stack

```
auto characterConfiguration = Physics::CharacterConfiguration();
auto characterShapeConfiguration = Physics::CapsuleShapeConfiguration();
auto characterGameplayConfiguration = PhysX::CharacterGameplayConfiguration();
```

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
```diff
@@ -159,6 +159,115 @@ namespace PhysX
...
+            {
+                physXSystem->Simulate(timeStep);
+                totalTime += timeStep;
+            }
+        }
+
+        AZ::Transform endTransform = transform->GetWorldTM();
+
+        EXPECT_FALSE(endTransform.IsClose(startTransform));
+        EXPECT_NEAR(expected_gravity, gameplayComponent->GetGravityMultiplier(), AZ::Constants::Tolerance);
```
hultonha, line 212: update: Please use google test matchers (see this PR for examples - https://github.com/o3de/o3de/pull/13643). They give much better info/feedback when the test fails (not just true/false)

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
```diff
@@ -159,6 +159,115 @@ namespace PhysX
         }
     }
 
+    TEST_F(PhysXDefaultWorldTest, CharacterController_GamePlayComponenet_GravityMovesEntity)
```
hultonha, line 162: suggestion/nit: Test name could instead be.. `EntityFallsUnderGravity`

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
```diff
@@ -159,6 +159,115 @@ namespace PhysX
...
+                physXSystem->Simulate(timeStep);
+                totalTime += timeStep;
+
+                EXPECT_NEAR(expected_gravity + i, gameplayComponent->GetGravityMultiplier(), AZ::Constants::Tolerance);
+            }
+        }
+
+        AZ::Transform endTransform = transform->GetWorldTM();
+
+        EXPECT_FALSE(endTransform.IsClose(startTransform));
```
hultonha, line 268: update: Same comment about matchers

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
```diff
@@ -159,6 +159,115 @@ namespace PhysX
...
+            characterShapeConfiguration->m_radius = 0.5f;
+
+            characterGameplayConfiguration->m_gravityMultiplier = 1.5f;
+            characterGameplayConfiguration->m_groundDetectionBoxHeight = 0.05f;
+
+            transform = gameplayEntity->CreateComponent<AzFramework::TransformComponent>();
+            characterComponent = gameplayEntity->CreateComponent<CharacterControllerComponent>(
+                AZStd::move(characterConfiguration), AZStd::move(characterShapeConfiguration));
+            gameplayComponent = gameplayEntity->CreateComponent<CharacterGameplayComponent>(AZStd::move(characterGameplayConfiguration));
+
```
hultonha, line 240: update: This code all looks nearly identical in both tests, could we extract it to a utility function (or move to the fixture setup) to avoid so much duplication? Thanks!

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
```diff
@@ -159,6 +159,115 @@ namespace PhysX
...
+        // Let scene run for a few moments so the entity can be manipulated by gravity from the gameplay component
+        auto startTransform = transform->GetWorldTM();
+
+        int duration = 10;
+        float totalTime = 0.0f;
+        float timeStep = AzPhysics::SystemConfiguration::DefaultFixedTimestep;
+
+        if (auto* physXSystem = GetPhysXSystem())
+        {
+
```
hultonha, line 255: uber nit: Please avoid additional line breaks when opening/closing braces

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
FuzzyCarterAWS: Updated

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
FuzzyCarterAWS: Updated

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
FuzzyCarterAWS: Updated

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
FuzzyCarterAWS: Related to above comments/questions.

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
FuzzyCarterAWS: Leftovers from when I was prototyping.

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
FuzzyCarterAWS: Will tackle this once I get clarity on above questions.

greerdv: +1

Gems/PhysX/Code/Source/PhysXCharacters/Components/CharacterGameplayComponent.cpp
greerdv: I think CharacterControllerComponent is actually the one which should probably be updated (but that's out of scope for this work). It's still using the old pattern where we used to have shared pointers everywhere.

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
```diff
@@ -159,6 +159,115 @@ namespace PhysX
...
+            transform->SetWorldTM(AZ::Transform::Identity());
+        }
+        gameplayEntity->Init();
+        gameplayEntity->Activate();
+
+        // Let scene run for a few moments so the entity can be manipulated by gravity from the gameplay component
+        auto startTransform = transform->GetWorldTM();
+
+        int duration = 10;
```
greerdv, line 196: nitpick - I would suggest calling this timeStepCount or something rather than duration. Duration sounds like it might be a time in e.g. seconds.

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
```diff
@@ -159,6 +159,115 @@ namespace PhysX
...
+        float totalTime = 0.0f;
```
greerdv, line 197: Do you ever actually use this?

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
```diff
@@ -159,6 +159,115 @@ namespace PhysX
...
+        float timeStep = AzPhysics::SystemConfiguration::DefaultFixedTimestep;
+
+        if (auto* physXSystem = GetPhysXSystem())
```
greerdv, line 200: If you're not using totalTime, I thought we already had a helper function for this. If there isn't a helper function we should definitely add one.

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
```diff
@@ -159,6 +159,115 @@ namespace PhysX
...
+            for (int i = 0; i < duration; i++)
+            {
+                physXSystem->Simulate(timeStep);
+                totalTime += timeStep;
+            }
+        }
+
+        AZ::Transform endTransform = transform->GetWorldTM();            
+
+        EXPECT_THAT(endTransform, testing::Not(UnitTest::IsClose(startTransform)));
```
greerdv, line 211: I'd prefer to check for something more specific than just not being close - this could mean it was 1cm away from where it started or 1km. Should be possible to figure out how far it should have fallen in the number of timesteps that have passed, and you can always use a fairly big tolerance.

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
```diff
@@ -159,6 +159,115 @@ namespace PhysX
         }
     }
 
+    TEST_F(PhysXDefaultWorldTest, CharacterController_GamePlayComponenet_EntityFallsUnderGravity)
+    {
+        PhysX::CharacterControllerComponent* characterComponent = nullptr;
+        PhysX::CharacterGameplayComponent* gameplayComponent = nullptr;
+        AzFramework::TransformComponent* transform = nullptr;
+
+        float expectedGravity = 1.5f;
```
greerdv, line 168: nitpick - I would call this expectedGravityMultiplier, otherwise it might read like this value is the actual acceleration.

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
```diff
@@ -159,6 +159,115 @@ namespace PhysX
...
+            {
+                physXSystem->Simulate(timeStep);
+                totalTime += timeStep;
+            }
+        }
+
+        AZ::Transform endTransform = transform->GetWorldTM();            
+
+        EXPECT_THAT(endTransform, testing::Not(UnitTest::IsClose(startTransform)));
+        EXPECT_THAT(expectedGravity, testing::FloatEq(gameplayComponent->GetGravityMultiplier()));
```
greerdv, line 212: I'm not sure it makes sense to test this here as well, it means the test is doing two things. If a lot of the body can be moved to utility functions, I would suggest making this a separate test.

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
```diff
@@ -159,6 +159,115 @@ namespace PhysX
         }
     }
 
+    TEST_F(PhysXDefaultWorldTest, CharacterController_GamePlayComponenet_EntityFallsUnderGravity)
```
greerdv, line 162: nitpick - typo in Component (here and other places)

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
```diff
@@ -159,6 +159,115 @@ namespace PhysX
...
+            transform->SetWorldTM(AZ::Transform::Identity());
+        }
+        gameplayEntity->Init();
+        gameplayEntity->Activate();
+
+        // Let scene run for a few moments so the entity can be manipulated by gravity from the gameplay component
+        auto startTransform = transform->GetWorldTM();
+
+        int duration = 10;
+        float totalTime = 0.0f;
```
greerdv, line 250: This doesn't look used again.

Gems/PhysX/Code/Source/PhysXCharacters/Components/CharacterGameplayComponent.cpp
hultonha: +1 to Dave's point, I think if you can just change this back and everything works @FuzzyCarterAWS I'd leave as is. For a `const` accessor I just meant something like...

```
const CharacterGameplayConfiguration& GetCharacterGameplayConfiguration() const
{
    return m_characterGameplayConfiguration;
}
```

If we're not storing it we might not be able to do that exactly and would need to return the equivalent for `m_gravityMultiplier` and `m_groundDetectionBoxHeight` but if they aren't needed I wouldn't bother.

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
hultonha: If you want to have a quick chat about this it might be quicker to resolve, let me know if my comments above have cleared things up. Thanks!

Gems/PhysX/Code/Source/PhysXCharacters/Components/CharacterGameplayComponent.cpp
FuzzyCarterAWS: Reverted and got things working.

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
FuzzyCarterAWS: Updated

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
FuzzyCarterAWS: Going to refactor this into either a helper or a test basis.

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
FuzzyCarterAWS: Going to use this in a gravity calculation.

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
FuzzyCarterAWS: Going to use this in a gravity calculation.

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
FuzzyCarterAWS: @greerdv 

Trying to figure out how to calculate the gravity here.

Currently stuck on trying to figure out how to use transforms.

Is there an easy way to get the X,Y,Z positions or a Vector3?

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
FuzzyCarterAWS: Updated

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
FuzzyCarterAWS: Fixed

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
FuzzyCarterAWS: Removed

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
FuzzyCarterAWS: @greerdv I've added a gravity calculation in, but now I'm getting failures when testing against my end Position and my expected position.

I'm going to have to dig deeper into this unless you can see anything that I'm doing incorrectly here.

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
FuzzyCarterAWS: I will extract this into its own test.

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
hultonha: Nice one thanks! 👍

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
hultonha: Yeah good plan, having a follow-up task to update these to the same style (removing the pointers) would be great. Thanks!

Gems/PhysX/Code/Source/PhysXCharacters/Components/CharacterGameplayComponent.cpp
hultonha: Sounds great @FuzzyCarterAWS, if you don't need one no need to add it. Thanks!

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
FuzzyCarterAWS: Backlogged on internal tracking tool. Will create GHIs when I start working on these after beginning of year planning finishes.

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
FuzzyCarterAWS: Extracting this out into the backlog on internal tracking too. I will create GHIs when work resumes on this after beginning of the year planning finishes.

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
FuzzyCarterAWS: Backlogged into internal tracking and simplified scope of this test into a single test.

Gems/PhysX/Code/Tests/PhysXTestCommon.cpp
```diff
@@ -63,6 +64,16 @@ namespace PhysX
             }
         }
 
+        float CalculateZFromFalling(float gravityMultipler, float totalTimeSteps, float startingZPosition)
+        {
+            // Calculate distance fallen (d = 0.5 * g * t^2)
+            float gravity = startingZPosition * gravityMultipler;
+            float time = totalTimeSteps * AzPhysics::SystemConfiguration::DefaultFixedTimestep;
+
+            return 0.5f * (gravity) * powf(time, 2);
```
hultonha, line 73: uber nit: Change `powf(time, 2)` to just `time * time` (it's also good to prefer `AZStd::pow` if you do need to use it

Gems/PhysX/Code/Tests/PhysXTestCommon.h
```diff
@@ -48,6 +48,9 @@ namespace PhysX
 
         void ResetPhysXSystem();
 
+        // Physics Math Helpers
```
hultonha, line 51: suggestion: A clearer comment for this function might be useful as opposed to a section comment (e.g. how to use it, what it does)

FuzzyCarterAWS: Noted in internal backlog ticket and removing from this PR since this code is not used for the remaining test for the focus of this PR.

GHIs will be created this week.

Gems/PhysX/Code/Tests/PhysXTestCommon.cpp
FuzzyCarterAWS: Noted in internal backlog ticket and removing from this PR since this code is not used for the remaining test for the focus of this PR.

GHIs will be created this week.

Gems/PhysX/Code/Tests/PhysXTestCommon.cpp
```diff
@@ -63,6 +64,16 @@ namespace PhysX
             }
         }
 
+        float CalculateZFromFalling(float gravityMultipler, float totalTimeSteps, float startingZPosition)
+        {
+            // Calculate distance fallen (d = 0.5 * g * t^2)
+            float gravity = startingZPosition * gravityMultipler;
```
greerdv, line 70: g in that equation is the acceleration due to gravity, which you can get from the GetGravity method in the scene interface (or in the tests you can just assume it's 0, 0, -10). startingZPosition should not affect the distance fallen. 

It won't apply to character controllers, but one nuance to be aware of for rigid bodies is that they have some linear damping by default, which means they fall a bit slower than you would expect just from gravity.

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
```diff
@@ -159,6 +159,56 @@ namespace PhysX
...
 
+    TEST_F(PhysXDefaultWorldTest, CharacterController_GamePlayComponent_GravitySetsWhileMoving)
+    {
+        PhysX::CharacterControllerComponent* characterComponent = nullptr;
+        PhysX::CharacterGameplayComponent* gameplayComponent = nullptr;
+        AzFramework::TransformComponent* transform = nullptr;
+
+        float expectedGravityMultiplier = 2.5f;
+
+        // Create character
```
greerdv, line 170: If the plan is to write more character controller tests later, this block would be great to split out as a helper function.

greerdv (approved):
Looks good to me once you remove the currently unused code from this change.

Gems/PhysX/Code/Tests/PhysXTestCommon.cpp
FuzzyCarterAWS: I had already removed these from my local but forgot to push them up. I'll add this feedback to the notes of the backlog item.

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
FuzzyCarterAWS: I had already removed these from my local but forgot to push them up. I'll add this feedback to the notes of the backlog item.

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
```diff
@@ -159,6 +159,56 @@ namespace PhysX
...
+                AZStd::move(characterConfiguration), AZStd::move(characterShapeConfiguration));
+            gameplayComponent = gameplayEntity->CreateComponent<CharacterGameplayComponent>(characterGameplayConfiguration);
+
+            transform->SetWorldTM(AZ::Transform::Identity());
+        }
+        gameplayEntity->Init();
+        gameplayEntity->Activate();
+
+        // Let scene run for a few moments so the entity can be manipulated by gravity from the gameplay component
+        auto startTransform = transform->GetWorldTM();
```
amzn-ahmadkrm, line 195: question: It doesn't look like this variable is being used anywhere, can it be removed?

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
```diff
@@ -159,6 +159,56 @@ namespace PhysX
...
+            characterShapeConfiguration->m_height = 1.5f;
+            characterShapeConfiguration->m_radius = 0.5f;
+
+            PhysX::CharacterGameplayConfiguration characterGameplayConfiguration;
+
+            characterGameplayConfiguration.m_gravityMultiplier = expectedGravityMultiplier;
+            characterGameplayConfiguration.m_groundDetectionBoxHeight = 0.05f;
+
+            transform = gameplayEntity->CreateComponent<AzFramework::TransformComponent>();
+            characterComponent = gameplayEntity->CreateComponent<CharacterControllerComponent>(
```
amzn-ahmadkrm, line 185: question: Is this used anywhere? If not can this also be removed?

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
```diff
@@ -159,6 +159,56 @@ namespace PhysX
         }
     }
 
+    TEST_F(PhysXDefaultWorldTest, CharacterController_GamePlayComponent_GravitySetsWhileMoving)
+    {
+        PhysX::CharacterControllerComponent* characterComponent = nullptr;
+        PhysX::CharacterGameplayComponent* gameplayComponent = nullptr;
+        AzFramework::TransformComponent* transform = nullptr;
```
amzn-ahmadkrm, line 166: suggestion: `transformComponent` instead of `transform`

amzn-ahmadkrm (approved):
Had some minor comments but overall LGTM 👍

hultonha (approved):
Changes look good, +1 to @amzn-ahmadkrm's suggestions if you have time to make those small fixes. Let me know if you need me to kick off a build. Thanks!

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
FuzzyCarterAWS: Was used in a previous version of this test and missed the cleanup when usage was removed. Thanks for catching this.

Removed.

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
FuzzyCarterAWS: Updated

Gems/PhysX/Code/Tests/CharacterControllerTests.cpp
FuzzyCarterAWS: characterComponent was not used elsewhere and was acting as an object catch for the created component that is required for the Game Play Component. Since we're not using it, removed the variable catch but leaving in the generation of the component.

PR #13643, opened 2022-12-10
Math test matchers

greerdv:
## What does this PR do?
Rewrites AzCore/Math tests to use matchers as much as possible, to provide more useful output if the tests fail.

## How was this PR tested?
Ran AzCore tests.

### DIFF
--- a/Code/Framework/AzCore/Tests/Math/AabbTests.cpp
+++ b/Code/Framework/AzCore/Tests/Math/AabbTests.cpp
@@ -30,32 +30,32 @@ namespace UnitTest
         Aabb aabb = Aabb::CreateFromPoint(Vector3(0.0f));
         EXPECT_TRUE(aabb.IsValid());
         EXPECT_TRUE(aabb.IsFinite());
-        EXPECT_TRUE(aabb.GetMin().IsClose(aabb.GetMax()));
+        EXPECT_THAT(aabb.GetMin(), IsClose(aabb.GetMax()));
     }
 
     TEST(MATH_Aabb, TestCreateFromMinMax)
     {
         Aabb aabb = Aabb::CreateFromMinMax(Vector3(0.0f), Vector3(1.0f));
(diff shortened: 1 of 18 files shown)


hultonha (approved):
A thing of beauty 😍 Awesome work @greerdv!

One quick question is did you test the output of at least one test failure per file? I think we need the include for the printers to get helpful output which can sometimes be missing (and you then just get the memory address printed). Thanks!

hultonha:
> > A thing of beauty 😍 Awesome work @greerdv!
> > One quick question is did you test the output of at least one test failure per file? I think we need the include for the printers to get helpful output which can sometimes be missing (and you then just get the memory address printed). Thanks!
> 
> I didn't do that but I have seen some tests which did have matchers fail in the past and they did print nicely.
> 
> We have all these in MathTestHelpers.h, I think that should cover pretty much all of the types here.
> 
> ```
> std::ostream& operator<<(std::ostream& os, const Vector2& vec);
> std::ostream& operator<<(std::ostream& os, const Vector3& vec);
> std::ostream& operator<<(std::ostream& os, const Vector4& vec);
> std::ostream& operator<<(std::ostream& os, const Aabb& aabb);
> std::ostream& operator<<(std::ostream& os, const Quaternion& quat);
> std::ostream& operator<<(std::ostream& os, const Transform& transform);
> std::ostream& operator<<(std::ostream& os, const Color& transform);
> std::ostream& operator<<(std::ostream& os, const Sphere& sphere);
> std::ostream& operator<<(std::ostream& os, const Matrix3x3& mat);
> std::ostream& operator<<(std::ostream& os, const Matrix3x4& mat);
> std::ostream& operator<<(std::ostream& os, const Matrix4x4& mat);
> ```

Cool good to know, yes most types are covered it's just if the include for those declarations isn't present the they won't be printed nicely. It's not a big problem to fix in future anyway just something to be aware of

PR #13650, opened 2022-12-12
[Terrain] Misc gradient / terrain bugfixes

### LINKED ISSUE #13573: Gradient Paint Brush: Allow switching paintbrush modes directly from the tool
The user can easily switch between paintbrush modes via the Viewport widget, however the drop-down to switch to the same modes on the actual paint brush tool is read-only.

Allow for switching the drop-down menu for paint modes directly on the paintbrush tool.

Using the widget that allows the same behavior, though this it wasn't apparent to me as a new user that those buttons switched between modes.

### LINKED ISSUE #13574: Gradient Paint Brush: Highlight the current paintbrush mode in the Viewport widget
Currently, the only indicator that you're in a specific paint mode is via the read-only drop-down menu on the Paint Brush Settings tool window.

Highlight the currently selected paint mode in the Viewport widget, much like the Transform widgets work.

mbalfour-amzn:
## What does this PR do?

This fixes a handful of bugs that were discovered while testing terrain painting:
* Paintbrush mode is now selectable via the dropdown in the paintbrush settings, instead of being readonly. (closes #13573 )
* Paintbrush mode now correctly highlights the active mode icon in the viewport. (closes #13574 )
* The default image gradient save path is now the root project path, not the bin folder that the Editor was run from.
* The Terrain Mesh Manager now refreshes its settings correctly when the Terrain World Renderer settings change. Previously, it was recalculating LODs based on stale settings.
* Image pixels per meter calculations now correctly account for tiling, frequency zoom, and scale. This fixed issues where the wrong pixels were changed when any of these were set to non-default values.

There are still more painting / refresh issues related to non-default Gradient Transform settings that will get addressed separately.

## How was this PR tested?

Manually tested in the Editor.


Gems/GradientSignal/Code/Source/Components/ImageGradientComponent.cpp
```diff
@@ -1082,17 +1082,25 @@ namespace GradientSignal
...
+            const AZ::Vector3 transformScale = m_gradientTransform.GetScale();
+            const float transformZoom = m_gradientTransform.GetFrequencyZoom();
+
+            // Get the local bounds for the gradient. This determines the total number of meters in each direction that the image
+            // repeats are mapped onto.
+            const AZ::Aabb localBounds = m_gradientTransform.GetBounds();
+
+            const AZ::Vector2 boundsMeters(localBounds.GetExtents());
+            const AZ::Vector2 imagePixelsInBounds(width * GetTilingX(), height * GetTilingY());
+            return (imagePixelsInBounds * transformZoom) / (boundsMeters * AZ::Vector2(transformScale));
```
mbalfour-amzn, line 1103: There were 2 bugs here - the original code should've multipled by tiling factors, not divided by them. But it also needed to account for the frequency zoom & scale factors.

Gems/Terrain/Code/Source/TerrainRenderer/TerrainMeshManager.cpp
```diff
@@ -84,12 +84,15 @@ namespace Terrain
...
+        bool requireRebuild = m_config.CheckWouldRequireRebuild(config);
+
+        m_config = config;
+
+        if (requireRebuild)
         {
             m_rebuildSectors = true;
             OnTerrainDataChanged(AZ::Aabb::CreateNull(), TerrainDataChangedMask::HeightData);
         }
-        m_config = config;
```
mbalfour-amzn, line 96: The issue here was that the OnTerrainDataChanged() call uses m_config to rebuild the sectors, so m_config needs to get set with the new values *before* calling OnTerrainDataChanged().

cgalvan (approved):

dtamkin1 (approved):

tjmichaels (approved):

PR #13651, opened 2022-12-12
Add filewatcher filtering

lawsonamzn:
## What does this PR do?
Adds the ability to filter filewatcher, so it can exclude cache and other rapidly changing folders that are not relevant to asset processing.  This is necessary because the cache and user folders now live in ProjectFolder and create a lot of spam.

This is even more relevant to Linux, which uses 1 watch handle (default limit 8096) for every single sub folder (and in our case, file) in there, quickly blowing past the limit.   After this change, AutomatedTesting only uses about 1k handles.

After this change, the compile log is actually a lot cleaner, its no longer triggering itself on every change to log files in the user folder.

This was causing jenkins failures due to running out of file handles on linux and general flakiness of these tests. 

I moved the tests that spam a ton of files to periodic suite.   This one creates thousands of files and could be risky for SSD to run them over and over.

## How was this PR tested?

As part of this, I fixed up the unit tests and other problems/spams I got during the file watcher tests as well which were flaky before.  They should be good now -   There are so many gotchas in this system.  To verify I ran the tests on repeat-shuffle x1000 on google test on linux (WSL), linux (real), windows, and mac.  The tests no longer use sleeps, they use events that occur as fast as possible (such as dropping a file in the file system and waiting for it to appear on the watcher).


Code/Framework/AzFramework/Platform/Linux/AzFramework/Application/Application_Linux.cpp
```diff
@@ -27,11 +27,14 @@ namespace AzFramework
...
         int get_limit_result = getrlimit(RLIMIT_NOFILE, &currentLimit);
         AZ_Warning("Application", get_limit_result == 0, "Unable to read current ulimit open file limits");
-        if ((get_limit_result == 0) && (currentLimit.rlim_cur < g_minimumOpenFileHandles || currentLimit.rlim_max < g_minimumOpenFileHandles))
+        // non-privileged systems may not update the maximum hard cap, only the current limit and it cannot exceed the max:
+        if ((get_limit_result == 0) && (currentLimit.rlim_cur < g_minimumOpenFileHandles) && ( currentLimit.rlim_max > 0))
         {
             rlimit newLimit;
-            newLimit.rlim_cur = g_minimumOpenFileHandles; // Soft Limit
-            newLimit.rlim_max = g_minimumOpenFileHandles; // Hard Limit
+            newLimit.rlim_cur = AZ::GetMin(g_minimumOpenFileHandles, currentLimit.rlim_max); // Soft Limit
```
lumberyard-employee-dm, line 34: The `g_minimumOpenFileHandles` value is set to 65536 which is requires 3 bytes to represent and therefore can only be stored in an `int`. Should this value be 65535?

lawsonamzn: I'm not sure what you mean, on linux we use constexpr rlim_t g_minimumOpenFileHandles = 65536L;
rlim_t is unsigned int,.  
```
constexpr rlim_t g_minimumOpenFileHandles = 65536L;
```
rlimit struct also uses rlim_t, and 65536 fits easily into unsigned int...

phuchau1989 (approved):

Code/Tools/AssetProcessor/Platform/Linux/native/FileWatcher/FileWatcher_linux.cpp
```diff
@@ -34,38 +71,70 @@ bool FileWatcher::PlatformImplementation::Initialize()
...
-    m_inotifyHandle = -1;
+void FileWatcher::PlatformImplementation::Finalize()
+{
+    // This happens after the watch thread has already joined.
+    
+    // note from the man-page of inotify " all associated watches are automatically freed." - no need to call inotify_rm_watch on each handle.
+    if (m_wakeThreadHandle >= 0)
+    {
+        close(m_wakeThreadHandle);
+    }
```
lumberyard-employee-dm, line 107: For the safety should we reset m_wakeThreadHandle back to `-1`?

Code/Tools/AssetProcessor/Platform/Linux/native/FileWatcher/FileWatcher_linux.cpp
```diff
@@ -95,37 +164,103 @@ void FileWatcher::PlatformImplementation::AddWatchFolder(QString folder, bool re
...
         (recursive ? QDirIterator::Subdirectories : QDirIterator::NoIteratorFlags) | QDirIterator::FollowSymlinks);
 
     while (dirIter.hasNext())
     {
-        QString dirName = dirIter.next();
-        int theErrno;
-        if (!TryToWatch(dirName, &theErrno))
+        QString dirPath = dirIter.next();
+
+        if (source.IsExcluded(dirPath))
```
lumberyard-employee-dm, line 192: Is dirPath a fullpath or just a path segment.
i.e Is it `/home/lawsonn/o3de/AutomatedTesting/Assets/textures` or is it `textures`?

If is the latter, won't files with the same name, but in different levels of the folder hierarchy be seen as already being watched?

Such as
`/home/lawsonn/o3de/AutomatedTesting/Assets/rendering/textures`
`/home/lawsonn/o3de/AutomatedTesting/Assets/audio/textures`

Code/Tools/AssetProcessor/Platform/Linux/native/FileWatcher/FileWatcher_linux.cpp
lawsonamzn: I'll add that.

Code/Framework/AzFramework/Platform/Linux/AzFramework/Application/Application_Linux.cpp
lumberyard-employee-dm: I was thinking the limit was an unsigned short max.

Code/Tools/AssetProcessor/Platform/Linux/native/FileWatcher/FileWatcher_linux.cpp
```diff
@@ -154,90 +288,178 @@ bool FileWatcher::PlatformStart()
...
             break;
         }
-        if (!bytesRead)
+
+        ssize_t bytesRead = 0;
+        if ((fds[1].revents & POLLIN) && (!m_shutdownThreadSignal))
         {
-            continue;
+            bytesRead = ::read(m_platformImpl->m_inotifyHandle, eventBuffer, s_inotifyReadBufferSize);
+            if (bytesRead < 0)
```
lumberyard-employee-dm, line 350: Do we need to memset the `eventBuffer` at all?
Since we are going to cast the `&eventBuffer[index]` to an `inotify_event*`, we shouldn't need to zero out the buffer.

Code/Tools/AssetProcessor/Platform/Linux/native/FileWatcher/FileWatcher_linux.cpp
lawsonamzn: its always a full path.

Code/Tools/AssetProcessor/Platform/Linux/native/FileWatcher/FileWatcher_linux.cpp
lawsonamzn: I did it for pure safety here, as the 'self events' from the OS (as in, 'self was deleted') don't actually write to the name ptr or name buffer at all, so if someone tries to printf it, it printfs fragments of garbage.

lumberyard-employee-dm: Ok, so that is why.
Thanks for the explanation.

Code/Framework/AzFramework/Platform/Linux/AzFramework/Application/Application_Linux.cpp
lawsonamzn:
```
struct rlimit {
    rlim_t rlim_cur;  /* Soft limit */
    rlim_t rlim_max;  /* Hard limit (ceiling for rlim_cur) */
};
```
and rlim_t is unsigned int or even a unsigned int64 in linux.  From the docs:
```
The glibc getrlimit() and setrlimit() wrapper functions use a
       64-bit rlim_t data type, even on 32-bit platforms.  However, the
       rlim_t data type used in the getrlimit() and setrlimit() system
       calls is a (32-bit) unsigned long.  Furthermore, in Linux, the
       kernel represents resource limits on 32-bit platforms as unsigned
       long.  
```

lumberyard-employee-dm: Ok, then I am guessing we set `65536L` as a soft max then.

lawsonamzn: yes, we can't modify the hard max unless we have admin so this code should be setting the soft max to 65536, and leaving the hard max at wahtever the hard max was before.

Code/Tools/AssetProcessor/Platform/Linux/native/FileWatcher/FileWatcher_linux.cpp
```diff
@@ -154,90 +289,178 @@ bool FileWatcher::PlatformStart()
...
-                        // Maintain the recursive nature of that root.
-                        const bool shouldAddFolder = (found == end(m_folderWatchRoots)) ? true : found->m_recursive;
-
-                        if (shouldAddFolder)
+                    bool isChildOfRootFolder = found != end(m_folderWatchRoots);
+                    bool isChildOfRecursiveRootFolder = isChildOfRootFolder && found->m_recursive;
+                    
+                    if (event->mask & IN_ISDIR)
+                    {
+                        // for directories, we only only care about create or delete, not modify
```
lumberyard-employee-dm, line 394: nitpick: typo double "only"
```suggestion
                        // for directories, we only care about create or delete, not modify
```

phuchau1989 (approved):

junbo75 (commented):
Nit: May worth renaming FileWatcher_mac.h or FileWatcher_macos.cpp to keep consistent with the platform name (either mac or macos).

Code/Tools/AssetProcessor/Platform/Linux/native/FileWatcher/FileWatcher_linux.cpp
```diff
@@ -78,15 +147,15 @@ bool FileWatcher::PlatformImplementation::TryToWatch(const QString &pathStr, int
         }
         return false;
     }
-    {
-        QMutexLocker lock{&m_handleToFolderMapLock};
-        m_handleToFolderMap[watchHandle] = pathStr;
-    }
+
+    m_handleToFolderMap[watchHandle] = pathStr;
```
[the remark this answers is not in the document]
lawsonamzn: its only ever called on one thread a time.    During init, before it starts watching, its called from the main thread.  Once init is called, its only ever called by the watcher thread.

AMZN-stankowi (approved):

Code/Tools/AssetProcessor/Platform/Linux/native/FileWatcher/FileWatcher_linux.h
```diff
@@ -11,22 +11,30 @@
...
 public:
     bool Initialize();
+    void CloseMainWatchHandle();
     void Finalize();
-    void AddWatchFolder(QString folder, bool recursive);
+    void AddWatchFolder(QString folder, bool recursive, FileWatcher& source, bool notifyFiles);
     void RemoveWatchFolder(int watchHandle);
 
     //! Try to watch the given path.
     //! @param errnoPtr If provided, gets set to the errno right after the inotify_add_watch() call.
```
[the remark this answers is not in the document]
lawsonamzn: I'll update either the code, or the comments.

Code/Tools/AssetProcessor/Platform/Mac/native/FileWatcher/FileWatcher_macos.cpp
```diff
@@ -87,62 +98,98 @@ void FileWatcher::WatchFolderLoop()
...
 
+void FileWatcher::PlatformImplementation::ConsumeEvents(size_t numEvents, const char **filePaths, const FSEventStreamEventFlags eventFlags[], const FSEventStreamEventId eventIds[])
+{
     for (int i = 0; i < numEvents; ++i)
     {
         const QFileInfo fileInfo(QDir::cleanPath(filePaths[i]));
         const QString fileAndPath = fileInfo.absoluteFilePath();
+        
+        // avoid repeats
+        if (m_lastSeenEventId >= eventIds[i])
```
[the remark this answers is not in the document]
lawsonamzn: From what I observed, macos event IDs depict the order the event occurred, but can arrive OUT of order when it slips backwards.

Eg, this is a valid sequence you can see if you print out what the OS is sending us in the callback:

```
1 2 3 4 5 6     3 4 5 6   5 6    6 6 6 6 7
```
The repeated 3,4,5,6 are the exact same events as the 3,4,5,6 that came in before, just sent to us again.
This allows me to ignore events that we have seen before.  They are not new events, just repeated notification of events we have already "consumed".

Code/Tools/AssetProcessor/native/FileWatcher/FileWatcher.cpp
```diff
@@ -188,12 +211,97 @@ bool FileWatcher::Filter(QString path, const WatchRoot& watchRoot)
...
+    // (a handle) and there are limited handles available, so excluding entire folder trees that we know we don't care about
+    // is valuable to safe resources even if it costs more in the listener.
+    // 
+    // It is up to each implementation to make use of the list of excludes to best optimize itself for performance.  Even if the
+    // implementation does absolutely nothing with this exclude list, this class itself will still filter out excludes before
+    // forwarding the file events to actual handlers.
+    //
+    // To strike a balance here, add just a few hand-picked exclusions that tend to contain a lot of unnecessary folders:
+    //    * Everything in the cache EXCEPT for the "Intermediate Assets" and "fence" folders (filtering out fence will deadlock!)
+    //    * Project/build/*    (case insensitive)
```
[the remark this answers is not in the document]
lawsonamzn: I actually started with that sort of thing ( I tried adding all the AP ignores) but the problem is that the watcher is very very sensitive to any slow down, we want only to absorb the absolute minimum of events here, and excluding everything at the watcher level isn't actually useful since the other parts of the program still re-run exclusions on anything already excluded here.   The exclusions here are mainly to keep the # of watched folders down.  I picked fence, user, and cache becuase htey provide thousands of excludes in just 1 exclude rule.

amzn-mike (approved):

Code/Tools/AssetProcessor/Platform/Mac/native/FileWatcher/FileWatcher_macos.cpp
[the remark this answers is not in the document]
junbo75: Got you. Thanks for the explanation, Nick.

Code/Tools/AssetProcessor/Platform/Linux/native/FileWatcher/FileWatcher_linux.h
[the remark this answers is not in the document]
lawsonamzn: it wasn't used.  I'll update.

Code/Tools/AssetProcessor/Platform/Linux/native/FileWatcher/FileWatcher_linux.cpp
```diff
@@ -34,38 +71,71 @@ bool FileWatcher::PlatformImplementation::Initialize()
...
+    {
+        close(m_wakeThreadHandle);
+        m_wakeThreadHandle = -1;
+    }
+    m_handleToFolderMap.clear();
+    m_alreadyNotifiedCreate.clear();
 }
 
-bool FileWatcher::PlatformImplementation::TryToWatch(const QString &pathStr, int* errnoPtr)
+bool FileWatcher::PlatformImplementation::TryToWatch(const QString &pathStr, int* errnoPtr, FileWatcher& source)
```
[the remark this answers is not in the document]
lawsonamzn: It wasn't used, I'lll remove it.

Code/Tools/AssetProcessor/Platform/Linux/native/FileWatcher/FileWatcher_linux.cpp
```diff
@@ -34,38 +71,71 @@ bool FileWatcher::PlatformImplementation::Initialize()
...
+            DEBUG_FILEWATCHER("Not adding an additional file watch for %s - it is not a directory s\n", pathStr.toUtf8().constData());
+            return false;
+        }
+        if (err == EEXIST)
+        {
+            // the dir already has a watch handle on it that belongs to m_inotifyHandle.
+            // avoid duplicating watches.  This is also okay to absorb but we do return true to indicate
+            // that we successfully have a watch on it.
+            DEBUG_FILEWATCHER("Not adding an additional file watch for %s - already exists\n", pathStr.toUtf8().constData());
+            return false;
```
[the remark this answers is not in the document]
lawsonamzn: I'll update the comment.  The behavior is correct - that is, if a watch is already established on that folder, it will have already established watches on its subtree.

Code/Tools/AssetProcessor/native/FileWatcher/FileWatcher.cpp
```diff
@@ -188,12 +211,97 @@ bool FileWatcher::Filter(QString path, const WatchRoot& watchRoot)
...
+        AZStd::string exclusion = AZStd::string::format(filterString, sanitizedCacheFolderString.toUtf8().constData(), intermediates);
+        AddExclusion(AssetBuilderSDK::FilePatternMatcher(exclusion, AssetBuilderSDK::AssetBuilderPattern::Regex));
+    }
+
+    if (!projectRootPath.isEmpty())
+    {
+        // These are not regexes, so do not need sanitation.  Files can't use special characters like * or ? from globs anyway.
+        AZStd::string userPath = QDir::toNativeSeparators(QDir(projectRootPath).absoluteFilePath("user/*")).toUtf8().constData();
+        AZStd::string buildPath = QDir::toNativeSeparators(QDir(projectRootPath).absoluteFilePath("build/*")).toUtf8().constData();
+        AZStd::string gemCodePath = QDir::toNativeSeparators(QDir(projectRootPath).absoluteFilePath("gem/code/*")).toUtf8().constData();
```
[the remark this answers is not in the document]
lawsonamzn: This is a QT signal/slot class already full of Qt-isms, so I used Qt.  It'd be the only actual use of AZ::IO::Path in the entire file.

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
```diff
@@ -1373,7 +1373,59 @@ namespace AssetProcessor
...
+                QString sanitizedFilePath = AssetUtilities::NormalizeFilePath(QString::fromUtf8(affectedSourceFile.c_str())).toUtf8().constData();
+                if (IsInCacheFolder(sanitizedFilePath.toUtf8().constData()))
+                {
+                    // this is a silent notification, that is, it makes it into the log but not
+                    // into any ui display or console.
+                    AZ_TracePrintf(AssetProcessor::DebugChannel,
+                        "File \"%s\" \n"
+                        "Builder UUID \"%s\" \n"
+                        "requested reprocess \"%s\" \n"
+                        "but that file lives in the Cache folder.  \nIgnored.",
```
lumberyard-employee-dm, line 1386: A trailing linefeed is needed here as `AZ_TracePrintf` doesn't automatically append a newline after the message.

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
```diff
@@ -1373,7 +1373,59 @@ namespace AssetProcessor
...
+                {
+                    // this is probably a mistake by either the builder author, or asset author, so
+                    // it gets surfaced to the UI.
+                    AZ_Warning(
+                        AssetProcessor::DebugChannel,
+                        false,
+                        "File \"%s\" \n"
+                        "Builder UUID \"%s\" \n"
+                        "requested reprocess \"%s\" \n"
+                        "That file does not live in any folder monitored by AP. \nIgnored.",
```
lumberyard-employee-dm, line 1405: repeat: a newline should be added after "Ignored".
```suggestion
                        "That file does not live in any folder monitored by AP. \nIgnored.\n",
```

amzn-mike (approved):

junbo75 (approved):

lumberyard-employee-dm (approved):

Code/Tools/AssetProcessor/native/unittests/FileWatcherUnitTests.cpp
```diff
@@ -5,352 +5,789 @@
...
+        return m_assetRootPath;
+    }
 
-    UnitTest::AssetProcessorUnitTestBase::TearDown();
-}
+    void FileWatcherUnitTest::CreateFenceFile()
+    {
+        AZ::IO::Path fencePath = AZ::IO::Path(GetFenceFolder().toUtf8().constData()) / "__fence__";
+        AZStd::string fenceString; 
+        ASSERT_TRUE(AZ::IO::CreateTempFileName(fencePath.String().c_str(), fenceString));
```
lumberyard-employee-dm, line 158: You can call `.c_str()` directly on the AZ::IO::Path instance, but what is better than invoking AZ::IO::CreateTempFileName is to use an [AZ::Test::ScopedAutoTempDirectory](https://github.com/o3de/o3de/blob/development/Code/Framework/AzTest/AzTest/Utils.h#L73-L87), which creates a temporary directory that will be cleaned up at the end of the test.

PR #13653, opened 2022-12-12
Migrate script canvas tests from hydra to multitest

michleza:
## What does this PR do?
- Updates script canvas test files to remove hydra referances
- Test files now uses script canvas component class and editor entity
- Test files load a test helper base level instead of a custom level
- Test files' style also update to follow other tests in the script canvas test suite
- Updated periodic test suite to include test files in the multitest section. 

Tests updated:
ScriptCanvas_TwoEntities_UseSimultaneously
ScriptCanvas_TwoComponents_InteractSuccessfully
ScriptCanvasComponent_OnEntityActivatedDeactivated_PrintMessage
ScriptEvents_Default_SendReceiveSuccessfully
ScriptEvents_HappyPath_SendReceiveAcrossMultiple 
ScriptEvents_ReturnSetType_Successfully

## How was this PR tested?
CLI py test runner and o3de python console.


AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_script_canvas.py
```diff
@@ -88,7 +88,7 @@ def get_component_graph_file(self) -> str:
 
     def set_variable_value(self, variable_name: str, variable_state: VariableState, variable_value) -> str:
         """
-        o3de/o3de#13344. We currently only support strings and boolean variables.
+        o3de/o3de#13344. We currently only support strings, entity.id and boolean variables.
         Function for setting a value to a variable on the script canvas component.
 
         Note: The variable must be exposed to the component. this can be done through the script canvas node inspector.
```
FuzzyCarterAWS, line 94: Put the GHI below the description but above the note so that way the autocomplete will give the relevant information of the usage.

AutomatedTesting/Gem/PythonTests/scripting/ScriptCanvasComponent_OnEntityActivatedDeactivated_PrintMessage.py
```diff
@@ -5,47 +5,52 @@
...
+    from editor_python_test_tools.editor_component.editor_script_canvas import ScriptCanvasComponent
+    from editor_python_test_tools.editor_component.editor_script_canvas import VariableState
+    from editor_python_test_tools.utils import TestHelper as TestHelper
+    from scripting_utils.scripting_constants import (WAIT_TIME_3)
+
+    position = math.Vector3(512.0, 512.0, 32.0)
+
+    #create entity, give it a script canvas component and set its active status
+    editor_entity = EditorEntity.create_editor_entity_at(position, entity_dictionary["name"])
+    editor_entity_sc_component = ScriptCanvasComponent(editor_entity)
```
FuzzyCarterAWS, line 34: rename to `scriptcanvas_component`

AutomatedTesting/Gem/PythonTests/scripting/ScriptCanvasComponent_OnEntityActivatedDeactivated_PrintMessage.py
```diff
@@ -5,47 +5,52 @@
...
     lines_found          = ("Successfully found expected prints",    "Failed to find expected prints")
     game_mode_exited     = ("Successfully exited game mode", "Failed to exit game mode")
 # fmt: on
 
-class ScriptCanvasComponent_OnEntityActivatedDeactivated_PrintMessage():
+
+def setup_level_entities(entity_dictionary):
+    """
+    Helper function for setting up the 3 entities and their script canvas components needed for this test
+
```
FuzzyCarterAWS, line 20: Add an example format for the dictionary.

AutomatedTesting/Gem/PythonTests/scripting/ScriptCanvasComponent_OnEntityActivatedDeactivated_PrintMessage.py
```diff
@@ -58,101 +63,74 @@ class ScriptCanvasComponent_OnEntityActivatedDeactivated_PrintMessage():
...
+    activated_dict = {
+        "name": "ActivationTest",
+        "status": "inactive",
+        "path": os.path.join(paths.projectroot, "ScriptCanvas", "OnEntityActivatedScripts", "activator.scriptcanvas"),
+    }
+    deactivated_dict = {
+        "name": "DeactivationTest",
+        "status": "active",
+        "path": os.path.join(paths.projectroot, "ScriptCanvas", "OnEntityActivatedScripts", "deactivator.scriptcanvas"),
+    }
```
FuzzyCarterAWS, line 98: These might be a good candidate to be tuples instead of dictionaries since they are only 1 entry deep. This will allow for tuple unpacking in the helper function for cleaner code.

Also, if we remove these from the test function to elsewhere in the file, it'll help with readability of test portion of the file.

AutomatedTesting/Gem/PythonTests/scripting/ScriptCanvasComponent_OnEntityActivatedDeactivated_PrintMessage.py
```diff
@@ -58,101 +63,74 @@ class ScriptCanvasComponent_OnEntityActivatedDeactivated_PrintMessage():
...
+    # Preconditions
+    general.idle_enable(True)
+
+    # 1) Create temp level
+    TestHelper.open_level("", "Base")
+
+    # 2) create all the entities we need for the test
+    setup_level_entities(activated_dict)
+    setup_level_entities(deactivated_dict)
+    setup_level_entities(controller_dict)
```
FuzzyCarterAWS, line 111: Extract this into a function that takes a list of entities to setup and pass it all three objects as a list.

`setup_all_level_entities([activated, deactivated, controller])`

AutomatedTesting/Gem/PythonTests/scripting/ScriptCanvas_TwoEntities_UseSimultaneously.py
```diff
@@ -40,50 +40,44 @@ def ScriptCanvas_TwoEntities_UseSimultaneously():
...
+    from editor_python_test_tools.utils import Tracer
     import azlmbr.legacy.general as general
     import azlmbr.math as math
-    import azlmbr.asset as asset
-    import azlmbr.bus as bus
 
-    LEVEL_NAME = "tmp_level"
+    TEST_ENTITY_NAME_1 = "test_entity_1"
+    TEST_ENTITY_NAME_2 = "test_entity_2"
     ASSET_PATH = os.path.join("scriptcanvas", "T92563191_test.scriptcanvas")
```
FuzzyCarterAWS, line 52: Why not create a helper that allows for returning the expected scriptcanvas director + file?

FuzzyCarterAWS (changes requested):

AutomatedTesting/Gem/PythonTests/scripting/ScriptCanvasComponent_OnEntityActivatedDeactivated_PrintMessage.py
FuzzyCarterAWS: Better solution, no new function needed. 

`setup_level_entities(for enetiy in [activated_dict, deactivated_dict, controller_dict])`

AutomatedTesting/Gem/PythonTests/scripting/ScriptCanvasComponent_OnEntityActivatedDeactivated_PrintMessage.py
FuzzyCarterAWS: Additionally, a more meaningful name would be *_entity instead of dict.

michleza: I personally don't like the use of dictionaries here but I'm having trouble seeing how tuples would be better. If I remember tuples are a collection type that have a format like a list or array with the benefit of not having to contain objects of the same type; `{"hello, 40, True, } `. It's not clear to me why we would use a tuple here or how to effectively use it.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_script_canvas.py
michleza: reorganized this per request.

AutomatedTesting/Gem/PythonTests/scripting/ScriptCanvasComponent_OnEntityActivatedDeactivated_PrintMessage.py
michleza: I'm going to rename this in all the scripts related to this PR

AutomatedTesting/Gem/PythonTests/scripting/ScriptCanvasComponent_OnEntityActivatedDeactivated_PrintMessage.py
michleza: elaborated on the dictionary format in the comment

AutomatedTesting/Gem/PythonTests/scripting/ScriptCanvasComponent_OnEntityActivatedDeactivated_PrintMessage.py
michleza: did something similar to this. I'm not huge fan of a for loop as an argument but the loop is in the function and the input parameter is the list now.

AutomatedTesting/Gem/PythonTests/scripting/ScriptCanvasComponent_OnEntityActivatedDeactivated_PrintMessage.py
```diff
@@ -5,47 +5,58 @@
...
-    deactivated_exists   = ("Successfully found deactivated entity", "Failed to find deactivated entity")
-    start_states_correct = ("Start states set up successfully",      "Start states set up incorrectly")
     game_mode_entered    = ("Successfully entered game mode",         "Failed to enter game mode")
     lines_found          = ("Successfully found expected prints",    "Failed to find expected prints")
     game_mode_exited     = ("Successfully exited game mode", "Failed to exit game mode")
 # fmt: on
 
-class ScriptCanvasComponent_OnEntityActivatedDeactivated_PrintMessage():
+
+def setup_level_entities(dictionaries):
```
FuzzyCarterAWS, line 17: rename dictionaries to entities

AutomatedTesting/Gem/PythonTests/scripting/ScriptCanvasComponent_OnEntityActivatedDeactivated_PrintMessage.py
```diff
@@ -5,47 +5,58 @@
...
+    import scripting_utils.scripting_tools as scripting_tools
+    from editor_python_test_tools.editor_entity_utils import EditorEntity
+    from editor_python_test_tools.editor_component.editor_script_canvas import ScriptCanvasComponent
+    from editor_python_test_tools.editor_component.editor_script_canvas import VariableState
+    from editor_python_test_tools.utils import TestHelper as TestHelper
+    from scripting_utils.scripting_constants import (WAIT_TIME_3)
+
+    position = math.Vector3(512.0, 512.0, 32.0)
+
+    for entity_dictionary in dictionaries:
```
FuzzyCarterAWS, line 36: rename entity_dictionary to entity

AutomatedTesting/Gem/PythonTests/scripting/ScriptCanvasComponent_OnEntityActivatedDeactivated_PrintMessage.py
FuzzyCarterAWS: When you unpack the dictionaries in the `setup_level_entities` method, you can assign them to meaningful names.

EG: `name, status, path = entity` then use those directly instead of having mystery strings accessing the dictionary. 

Another alternative would be to use a [dataclass](https://docs.python.org/3/library/dataclasses.html).

Lastly, when building these, since you're not using these values anywhere else, you can listify them on generation.

Below is an example of combining a things together using the dataclass variant.

```suggestion
SCRIPT_CANVAS_DEFAULT_LOCATION= os.path.join(paths.projectroot, "ScriptCanvas")

def test_file_path(file_name: str) -> os.Pathlike:
    return os.path.join(SCRIPT_CANVAS_DEFAULT_LOCATION, "OnEntityActivatedScripts", file_name)
    
@dataclass
class testEntity:
    def __init__(self, name: str, status: str, path: os.Pathlike):
        self.name = name
        self.status = status
        self.path = path

test_entities = [
    testEntity("Controller", "active", test_file_path("controller.scriptcanvas"),
    testEntity("ActivationTest", "inactive", test_file_path("activator.scriptcanvas"),
    testEntity("DeactivationTest", "active", test_file_path("deactivator.scriptcanvas")
]

setup_level_entities(test_entities)
```

AutomatedTesting/Gem/PythonTests/scripting/ScriptCanvas_TwoEntities_UseSimultaneously.py
FuzzyCarterAWS, line 52: Does this need the project root as part of the path?

AutomatedTesting/Gem/PythonTests/scripting/scripting_utils/scripting_constants.py
```diff
@@ -71,6 +71,7 @@
 SAVE_STRING = "Save"
 NAME_STRING = "Name"
 WAIT_FRAMES = 200
+WAIT_TIME_1 = 1
 WAIT_TIME_3 = 3
 WAIT_TIME_5 = 5
```
FuzzyCarterAWS, line 76: Backlog: We should deprecate the timing scripting constants for the timing constants in `AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/consts/general.py`

FuzzyCarterAWS (changes requested):

AutomatedTesting/Gem/PythonTests/scripting/ScriptCanvasComponent_OnEntityActivatedDeactivated_PrintMessage.py
michleza: took this example and implemented it. There's a small issue with the way test runner handles imports so the os and paths imports need to be within the class constructor but otherwise this works.

AutomatedTesting/Gem/PythonTests/scripting/ScriptCanvasComponent_OnEntityActivatedDeactivated_PrintMessage.py
michleza: renamed the container as it now contains the test entities objects

AutomatedTesting/Gem/PythonTests/scripting/ScriptCanvasComponent_OnEntityActivatedDeactivated_PrintMessage.py
michleza: renamed

AutomatedTesting/Gem/PythonTests/scripting/ScriptCanvas_TwoEntities_UseSimultaneously.py
michleza: nope. not for this file. It's in automatedtesting/scriptcanvas

AutomatedTesting/Gem/PythonTests/scripting/scripting_utils/scripting_constants.py
michleza: i agree. something this generic seems like it shouldn't live in script canvas land. we should talk about planning something like this. It may have to way until after holiday season though

AutomatedTesting/Gem/PythonTests/scripting/ScriptCanvas_TwoEntities_UseSimultaneously.py
michleza: if the path structure was more complicated I would do this but here it's just two separate strings.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_script_canvas.py
```diff
@@ -88,8 +88,9 @@ def get_component_graph_file(self) -> str:
 
     def set_variable_value(self, variable_name: str, variable_state: VariableState, variable_value) -> str:
         """
-        o3de/o3de#13344. We currently only support strings and boolean variables.
+        We currently only support strings, entity.id and boolean variables.
         Function for setting a value to a variable on the script canvas component.
+        o3de/o3de#13344.
```
FuzzyCarterAWS, line 93: I think you mis-understood my previous feedback.

Ideally we should do the following for a docstring.

```
"""
Description of the function
issue links
notes
params
"""
```
```suggestion
        Function for setting a value to a variable on the script canvas component.
        o3de/o3de#13344: Only supports strings, entity.id, and boolean variables.
```

AutomatedTesting/Gem/PythonTests/scripting/ScriptCanvasComponent_OnEntityActivatedDeactivated_PrintMessage.py
```diff
@@ -5,154 +5,134 @@
...
-     7) Validate Print message
+        import os
+        import azlmbr.paths as paths
 
-    Note:
-     - This test file must be called from the Open 3D Engine Editor command terminal
-     - Any passed and failed tests are written to the Editor.log file.
-            Parsing the file or running a log_monitor are required to observe the test results.
+        self.name = name
+        self.status = status
```
FuzzyCarterAWS, line 23: entity_state is probably better here than status

FuzzyCarterAWS (approved):
Looks great. Approved after two small changes that don't impact how things work.

* Variable name change
* Comment structure update.

AutomatedTesting/Gem/PythonTests/EditorPythonTestTools/editor_python_test_tools/editor_component/editor_script_canvas.py
michleza: oh I see. updated

AutomatedTesting/Gem/PythonTests/scripting/ScriptCanvasComponent_OnEntityActivatedDeactivated_PrintMessage.py
michleza: agreed. updated variable name

FuzzyCarterAWS:
Final hydra deprecation of Scripting tests.

* Updates script canvas test files to remove hydra referances
* Test files now uses script canvas component class and editor entity
* Test files load a test helper base level instead of a custom level
* Test files' style also update to follow other tests in the script canvas test suite
* Updated periodic test suite to include test files in the multitest section.

Tests updated:

- ScriptCanvas_TwoEntities_UseSimultaneously
- ScriptCanvas_TwoComponents_InteractSuccessfully
- ScriptCanvasComponent_OnEntityActivatedDeactivated_PrintMessage
- ScriptEvents_Default_SendReceiveSuccessfully
- ScriptEvents_HappyPath_SendReceiveAcrossMultiple
- ScriptEvents_ReturnSetType_Successfully

PR #13664, opened 2022-12-12
Asset Processor - Hook up Uuid Manager to AP

amzn-mike:
## What does this PR do?

Added 2 new APIs for AssetUtils `GetSourceUuid` and `GetLegacySourceUuids` which call into the Uuid Manager.

Replaced all existing calls to `CreateSafeSourceUUIDFromName` with these 2 new calls and updated unit tests.

Uuid manager now requires that files exist on disk for these calls to succeed, as it makes no sense (**in AP**, this may be different for the public API later) to make a metadata file for a file which does not exist.

Note that the uuid manager is still effectively disabled and will return the same relative-path based UUIDs as before since there are no enabled file types yet.

## How was this PR tested?

Updated and added new unit tests and ran them.
Ran AP and reprocessed some files.
Started up editor and opened a level.


Code/Framework/AzFramework/AzFramework/Asset/AssetRegistry.cpp
```diff
@@ -139,25 +139,18 @@ namespace AzFramework
...
-                    m_realAssetIdToLegacyAssetIdMap.erase(rangeItr);
-                    break;
-                }
-            }
-
-            m_legacyAssetIdToRealAssetId.erase(itr);
+            m_legacyAssetIdToRealAssetId.erase(rangeItr->second);
         }
+
+        // Erase the readId -> legacyId lookup
```
jackalbe, line 152: "read Id" or "real Id" ?

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
```diff
@@ -1481,15 +1485,26 @@ namespace AssetProcessor
...
+
+                for (const auto& legacyUuid : legacySourceUuids)
                 {
-                    message.m_legacyAssetIds.push_back(legacySourceAssetId);
+                    AZ::Data::AssetId legacySourceAssetId(legacyUuid, newProduct.m_subID);
+
+                    if (legacySourceAssetId != assetId)
+                    {
+                        legacySourceAssetIds.emplace(legacySourceAssetId);
+                        message.m_legacyAssetIds.push_back(legacySourceAssetId);
```
jackalbe, line 1500: Should these lines be reversed to avoid a copy?

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
```diff
@@ -4646,7 +4668,31 @@ namespace AssetProcessor
...
+            }
+            else
+            {
+                if (AZ::IO::FileIOBase::GetInstance()->Exists(sourceAsset.AbsolutePath().c_str()))
+                {
+                    uuid = AssetUtilities::GetSourceUuid(sourceAsset);
+                }
+                else
+                {
+                    AZ_TracePrintf(
```
jackalbe, line 4687: This seems like an AZ_Warning since unexpected behavior might happen due to "may be incomplete" 
Will this happen often?

AMZN-stankowi (approved):

jackalbe (approved):

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
amzn-mike: A copy of what?  We're inserting the assetId into the tracking set and the message vector.

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
```diff
@@ -2918,9 +2925,24 @@ namespace AssetProcessor
...
+                            sourceAssetReference.AbsolutePath().c_str());
+                    }
+                    else
+                    {
+                        AzToolsFramework::AssetSystem::SourceFileNotificationMessage message(
+                            AZ::OSString(sourceAssetReference.RelativePath().c_str()),
+                            AZ::OSString(scanFolderInfo->ScanPath().toUtf8().constData()),
+                            AzToolsFramework::AssetSystem::SourceFileNotificationMessage::FileRemoved,
+                            sourceDatabaseEntry.m_sourceGuid);
+                        EBUS_EVENT(AssetProcessor::ConnectionBus, Send, 0, message);
```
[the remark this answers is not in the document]
amzn-mike: Updated

Code/Framework/AzFramework/AzFramework/Asset/AssetRegistry.cpp
amzn-mike: Updated

Code/Tools/AssetProcessor/native/tests/UnitTestUtilities.cpp
```diff
@@ -145,4 +145,76 @@ namespace UnitTests
             m_matcherBuilderPatterns.push_back(patternMatcher);
         }
     }
-}
+
+    TraceBusErrorChecker::TraceBusErrorChecker()
+    {
+        BusConnect();
```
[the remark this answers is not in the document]
amzn-mike: Yeah I thought about it but figured since it's a very specific purpose class I could get away with it, but since you mentioned it I'll update it

junbo75 (approved):

Code/Tools/AssetProcessor/native/AssetManager/AssetCatalog.cpp
```diff
@@ -469,80 +462,148 @@ namespace AssetProcessor
...
+                    if (!fileStateInterface)
+                    {
+                        AZ_Assert(false, "Programmer Error - IFileStateRequests interface is not available");
+                        return false;
+                    }
+
+                    const bool fileExists = fileStateInterface->Exists(sourceAsset.AbsolutePath().c_str());
+
+                    // Only try to update for files which actually exist
+                    if (fileExists)
```
jackalbe, line 494: This feels funny. If this is adding an entry to a source file, then why continue if it does not exist? Should it be removed from the DB instead?

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
```diff
@@ -194,11 +194,6 @@ namespace AssetProcessor
                  (status == AssetProcessor::AssetScanningStatus::Stopped))
         {
             AssetProcessor::StatsCapture::EndCaptureStat("AssetScanning");
-            // place a message in the queue that will cause us to transition
-            // into a "no longer scanning" state and then continue with the next phase
-            // we place this at the end of the queue rather than calling it immediately, becuase
-            // other messages may still be in the queue such as the incoming file list.
-            QMetaObject::invokeMethod(this, "FinishAssetScan", Qt::QueuedConnection);
```
jackalbe, line 201: This seems like an unrelated change, or did this cause issues with the new logic?

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
jackalbe: legacySourceAssetId could be moved to legacySourceAssetIds after a push_back with message, or do we want a "Construct and insert element" here?

Code/Tools/AssetProcessor/native/AssetManager/AssetCatalog.cpp
amzn-mike: Good question.  This runs really early on, before AP has really done any work.  AP will eventually get around to detecting deleted files and doing all the needed clean up and it actually needs these entries so it can find the files in the cache which need to be deleted.  So we do need to leave the entry until AP is ready to do the cleanup.

AMZN-stankowi (approved):

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
amzn-mike: This is part of the optimization for the catalog setup.  Since it's using the file state cache now, I had to reorganize some of the startup stuff, which included moving this line of code.  Previously the completion of the asset scan caused AP to immediately move to the next stage, but now AP collects the scan results and waits for catalog & builder init to finish before moving on to actually process the scan results.

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
```diff
@@ -3324,6 +3350,19 @@ namespace AssetProcessor
         }
 
         AssetProcessor::StatsCapture::EndCaptureStat("InitialFileAssessment");
+
+        // place a message in the queue that will cause us to transition
+        // into a "no longer scanning" state and then continue with the next phase
+        // we place this at the end of the queue rather than calling it immediately, becuase
+        // other messages may still be in the queue such as the incoming file list.
+        QMetaObject::invokeMethod(this, "FinishAssetScan", Qt::QueuedConnection);
```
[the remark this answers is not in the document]
amzn-mike: Pretty much, see above comment

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
amzn-mike: Ah I see.  Sure, I've updated it to move the assetId into the last container.

PR #13628, opened 2022-12-09
Add an option for instant focus entity and go to position for Editor Camera

hultonha:
Signed-off-by: Tom Hulton-Harrop <82228511+hultonha@users.noreply.github.com>

## What does this PR do?

This PR adds an option to allow the user to have the camera update its transform immediately when focusing on an Entity ('Z') or using the __Go To Position__ dialog box.

This is something that was requested a while ago by @benblack75 and I finally found time to add it 😅

This is also a very important option for those users who are more sensitive to motion sickness and find the current interpolation of the camera uncomfortable.

## How was this PR tested?

Manually in the editor and also more unit tests have been added (see testing section below).

### Before

https://user-images.githubusercontent.com/82228511/206748879-86bb73f7-a5d1-481a-995f-e653dbd37bdb.mp4

### After

https://user-images.githubusercontent.com/82228511/206748922-f5bc1d8d-304d-4ffa-aca8-ea6dcb9cd6f7.mp4

### Tests

```bash
Note: Google Test filter = *EditorCameraFixture*
[==========] Running 14 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 14 tests from EditorCameraFixture
[ RUN      ] EditorCameraFixture.ModularViewportCameraControllerReferenceFrameUpdatedWhenViewportEntityisChanged
[       OK ] EditorCameraFixture.ModularViewportCameraControllerReferenceFrameUpdatedWhenViewportEntityisChanged (114 ms)
[ RUN      ] EditorCameraFixture.TrackingTransformIsTrueAfterTransformIsTracked
[       OK ] EditorCameraFixture.TrackingTransformIsTrueAfterTransformIsTracked (98 ms)
[ RUN      ] EditorCameraFixture.TrackingTransformIsFalseAfterTransformIsStoppedBeingTracked
[       OK ] EditorCameraFixture.TrackingTransformIsFalseAfterTransformIsStoppedBeingTracked (94 ms)
[ RUN      ] EditorCameraFixture.InterpolateToTransform
[       OK ] EditorCameraFixture.InterpolateToTransform (94 ms)
[ RUN      ] EditorCameraFixture.InterpolateToTransformWithReferenceSpaceSet
[       OK ] EditorCameraFixture.InterpolateToTransformWithReferenceSpaceSet (102 ms)
[ RUN      ] EditorCameraFixture.BeginningCameraInterpolationReturnsTrue
[...]

PR #13669, opened 2022-12-13
Improvements to camera interpolation

hultonha:
## What does this PR do?

This PR improves the current interpolation support for the camera (and adds 'Go to instantly' for View Bookmarks which were missed in the previous PR). An interpolation time can now be set by the user if they wish to tweak how long or short a camera transition is. There are also some subtle improvements to how interpolations can be started/stopped (A camera cannot interpolate to where it already is, if we've started a transition, we don't want to be able to start it again going to the same place but we do want to be able to 'interrupt' the transition by setting a new transform that's different to the first. We also clear the end transform when completed so it's simple to animate to that transform again after the camera moves).

Relates to https://github.com/o3de/o3de/pull/13628

image

## How was this PR tested?

The PR was tested manually in the Editor and some additional unit/integration tests have also been added to verify the behavior.

### Tests

```bash
[==========] Running 18 tests from 1 test case.
[----------] 18 tests from EditorCameraFixture
[ RUN      ] EditorCameraFixture.ModularViewportCameraControllerReferenceFrameUpdatedWhenViewportEntityisChanged
[       OK ] EditorCameraFixture.ModularViewportCameraControllerReferenceFrameUpdatedWhenViewportEntityisChanged (127 ms)
[ RUN      ] EditorCameraFixture.TrackingTransformIsTrueAfterTransformIsTracked
[       OK ] EditorCameraFixture.TrackingTransformIsTrueAfterTransformIsTracked (97 ms)
[ RUN      ] EditorCameraFixture.TrackingTransformIsFalseAfterTransformIsStoppedBeingTracked
[       OK ] EditorCameraFixture.TrackingTransformIsFalseAfterTransformIsStoppedBeingTracked (100 ms)
[ RUN      ] EditorCameraFixture.InterpolateToTransform
[       OK ] EditorCameraFixture.InterpolateToTransform (95 ms)
[ RUN      ] EditorCameraFixture.InterpolateToTransformWithReferenceSpaceSet
[       OK ] EditorCameraFixture.InterpolateToTransformWithReferenceSpaceSet (96 ms)
[ RUN      ] EditorCameraFixture.BeginningCameraInterpolationReturnsTrue
[       OK ] EditorCameraFixture.BeginningCameraInterpolationReturnsTrue (93 ms)
[ RUN      ] EditorCameraFixture.CameraInterpolationIsNotInterruptedIfGoingToTheSameTransform
[       OK ] EditorCameraFixture.CameraInterpolationIsNotInterruptedIfGoingToTheSameTransform (93 ms)
[ RUN      ] EditorCameraFixture.CameraInterpolationCanBeInterruptedIfGoingToDifferentTransform
[       OK ] EditorCameraFixture.CameraInterpolationCanBeInterruptedIfGoingToDifferentTransform (95 ms)
[ RUN      ] EditorCameraFixture.CameraInterpolationCanBeginAfterAnInterpolationCompletes
[       OK ] EditorCameraFixture.CameraInterpolationCanBeginAfterAnInterpolationCompletes (96 ms)
[ RUN      ] EditorCameraFixture.CameraCannotInterpolateToSamePositionWithoutMovingFromIt
[       OK ] EditorCameraFixture.CameraCannotInterpolateToSamePositionWithoutMovingFromIt (93 ms)
[ RUN      ] EditorCameraFixture.CameraInterpolatesToTransformWhenGoToPositionInstantlyIsOff
[       OK ] EditorCameraFixture.CameraInterpolatesToTransformWhenGoToPositionInstantlyIsOff (95 ms)
[ RUN      ] EditorCameraFixture.CameraChangesImmediatelyWhenGoToPositionInstantlyIsOn
[       OK ] EditorCameraFixture.CameraChangesImmediatelyWhenGoToPositionInstantlyIsOn (97 ms)
[ RUN      ] EditorCameraFixture.CameraInterpolatesToTransform
[       OK ] EditorCameraFixture.CameraInterpolatesToTransform (95 ms)
[ RUN      ] EditorCameraFixture.CameraInterpolatesToTransformWithZeroDurationAfterUpdate
[       OK ] EditorCameraFixture.CameraInterpolatesToTransformWithZeroDurationAfterUpdate (97 ms)
[ RUN      ] EditorCameraFixture.CameraDoesNotInterpolateToTransformWithZeroDurationAndNoUpdate
[       OK ] EditorCameraFixture.CameraDoesNotInterpolateToTransformWithZeroDurationAndNoUpdate (95 ms)
[ RUN      ] EditorCameraFixture.CameraSettingsRegistryValuesCanBeModified
[       OK ] EditorCameraFixture.CameraSettingsRegistryValuesCanBeModified (97 ms)
[ RUN      ] EditorCameraFixture.CameraSettingsRegistryValuesCanBeReset
[       OK ] EditorCameraFixture.CameraSettingsRegistryValuesCanBeReset (96 ms)
[ RUN      ] EditorCameraFixture.CameraSettingsRegistryInputValuesCanBeReset
[       OK ] EditorCameraFixture.CameraSettingsRegistryInputValuesCanBeReset (98 ms)
[----------] 18 tests from EditorCameraFixture (1776 ms total)

[==========] 18 tests from 1 test case ran. (1778 ms total)
[  PASSED  ] 18 tests.
```

Also updated camera settings menu (tagging @mbalfour-amzn for visibility)

image

### DIFF
--- a/Code/Editor/Lib/Tests/Camera/test_EditorCamera.cpp
+++ b/Code/Editor/Lib/Tests/Camera/test_EditorCamera.cpp
@@ -296,7 +296,7 @@ namespace UnitTest
             secondTransformToInterpolateTo,
             InterpolateToTransformDuration);
 
-        bool interpolating = true;
+        bool interpolating = false;
         AtomToolsFramework::ModularViewportCameraControllerRequestBus::EventResult(
             interpolating, TestViewportId, &AtomToolsFramework::ModularViewportCameraControllerRequestBus::Events::IsInterpolating);
 
@@ -710,7 +710,7 @@ namespace UnitTest
 
         {
             const float existingCameraGoToPositionDuration = SandboxEditor::CameraGoToPositionDuration();
-            const float expectedCameraGoToPositionDuration = !existingCameraGoToPositionDuration;
+            const float expectedCameraGoToPositionDuration = existingCameraGoToPositionDuration + 1.0f;
             SandboxEditor::SetCameraGoToPositionDuration(expectedCameraGoToPositionDuration);
             const float actualCameraGoToPositionDuration = SandboxEditor::CameraGoToPositionDuration();
             EXPECT_THAT(expectedCameraGoToPositionDuration, Eq(actualCameraGoToPositionDuration));


Code/Editor/EditorPreferencesPageViewportCamera.cpp
```diff
@@ -185,11 +186,19 @@ void CEditorPreferencesPage_ViewportCamera::CameraMovementSettings::Reflect(AZ::
...
+                "Camera Go To Position Duration",
+                "Time it takes for the camera to interpolate to a given position")
+            ->Attribute(AZ::Edit::Attributes::Visibility, &CameraMovementSettings::GoToPositionDurationVisibility)
+            ->Attribute(AZ::Edit::Attributes::Min, minValue)
             ->DataElement(
                 AZ::Edit::UIHandlers::CheckBox,
                 &CameraMovementSettings::m_goToPositionInstantly,
                 "Camera Go To Position Instantly",
                 "Camera will instantly go to the set position and won't interpolate there")
+            ->Attribute(AZ::Edit::Attributes::ChangeNotify, AZ::Edit::PropertyRefreshLevels::EntireTree)
```
mbalfour-amzn, line 201: Suggestion: It would be nice to add a comment here to explain why the EntireTree refresh is needed. (I'm assuming it's to refresh the visibility of the Go To Position Duration?

Suggestion: If you make the duration readonly instead of hidden, it might be a little bit less jarring of a change (items won't reposition on the screen) and as a bonus you only have to refresh AttributesAndValues, not EntireTree.

Code/Editor/EditorPreferencesPageViewportCamera.cpp
```diff
@@ -81,6 +81,7 @@ void CEditorPreferencesPage_ViewportCamera::CameraMovementSettings::Reflect(AZ::
         ->Field("DefaultOrbitDistance", &CameraMovementSettings::m_defaultOrbitDistance)
         ->Field("SpeedScale", &CameraMovementSettings::m_speedScale)
         ->Field("GoToPositionInstantly", &CameraMovementSettings::m_goToPositionInstantly)
+        ->Field("GoToPositionDuration", &CameraMovementSettings::m_gotoPositionDuration)
```
mbalfour-amzn, line 84: Nit: Seems like this should be "m_goToPositionDuration" (capital T) for consistency with m_goToPositionInstantly.

Code/Editor/EditorViewportCamera.h
```diff
@@ -60,21 +60,22 @@ namespace SandboxEditor
     //! @param pitch Amount of pitch in radians.
     //! @param yaw Amount of yaw in radians.
     SANDBOX_API void InterpolateViewportCameraToTransform(
-        AzFramework::ViewportId viewportId, const AZ::Vector3& position, float pitch, float yaw);
+        AzFramework::ViewportId viewportId, const AZ::Vector3& position, float pitch, float yaw, float duration);
```
mbalfour-amzn, line 63: Repeated: The comments above all these APIs should be updated to include the new parameter.

Code/Editor/EditorViewportCamera.h
```diff
@@ -86,21 +87,23 @@ namespace SandboxEditor
...
-    SANDBOX_API void HandleViewportCameraTransitionFromSetting(AzFramework::ViewportId viewportId, const AZ::Vector3& position, float pitch, float yaw);
+    SANDBOX_API void HandleViewportCameraTransitionFromSetting(
+        AzFramework::ViewportId viewportId, const AZ::Vector3& position, float pitch, float yaw, float duration);
 
     //! Will call either Set or Interpolate camera transform depending on user setting.
     //! @param position The new position of the camera in world space.
     //! @param pitch Amount of pitch in radians.
     //! @param yaw Amount of yaw in radians.
-    SANDBOX_API void HandleDefaultViewportCameraTransitionFromSetting(const AZ::Vector3& position, float pitch, float yaw);
+    SANDBOX_API void HandleDefaultViewportCameraTransitionFromSetting(const AZ::Vector3& position, float pitch, float yaw, float duration);
```
mbalfour-amzn, line 97: Suggestion: Since it seems like all the uses of this API consist of passing in the value read in from the settings, maybe the API should internally look up that setting instead of exposing duration.

Gems/Atom/Tools/AtomToolsFramework/Code/Include/AtomToolsFramework/Viewport/ModularViewportCameraController.h
```diff
@@ -150,9 +150,10 @@ namespace AtomToolsFramework
         struct CameraAnimation
         {
             //! The transform of the camera at the start of the animation.
-            AZ::Transform m_transformStart = AZ::Transform::CreateIdentity();
-            AZ::Transform m_transformEnd = AZ::Transform::CreateIdentity(); //!< The transform of the camera at the end of the animation.
+            AZ::Transform m_transformStart;
+            AZ::Transform m_transformEnd; //!< The transform of the camera at the end of the animation.
             float m_time = 0.0f; //!< The interpolation amount between the start and end transforms (in the range 0.0 - 1.0).
+            float m_duration = 1.0f; //!< The length of the animation (how long it takes to completed).
```
mbalfour-amzn, line 156: Nit: Would be nice to either name the variable or list in the comments what unit the duration is in. Milliseconds? Seconds? Ticks?

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Viewport/ModularViewportCameraController.cpp
```diff
@@ -285,18 +291,24 @@ namespace AtomToolsFramework
...
+        const auto& currentCameraTransform = CombinedCameraTransform();
+
+        // ensure the transform we're interpolating to isn't the same as our current transform
+        // and the transform we're setting isn't the same as one previously set
+        if (!currentCameraTransform.IsClose(worldFromLocal) &&
+            (!m_cameraAnimation.has_value() || !worldFromLocal.IsClose(m_cameraAnimation->m_transformEnd)))
         {
             m_cameraMode = CameraMode::Animation;
-            m_cameraAnimation = CameraAnimation{ CombinedCameraTransform(), worldFromLocal, 0.0f };
+            m_cameraAnimation = CameraAnimation{ currentCameraTransform, worldFromLocal, 0.0f, duration };
```
mbalfour-amzn, line 311: If the new transform is close, shouldn't it still set it as the new transform immediately instead of ignoring it? It seems like this could lead to an edge case behavior where tiny intentional changes just get dropped.

Code/Editor/EditorPreferencesPageViewportCamera.cpp
hultonha: Good spot, will fix (fixed up a few other places too)

Code/Editor/EditorViewportCamera.h
hultonha: Good call will add 👍

Code/Editor/EditorViewportCamera.h
hultonha: Yeah that's not a bad plan, keeps that interface a little simpler, I'll remove it.

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Viewport/ModularViewportCameraController.cpp
hultonha: I do not believe so no. If you're attempting to interpolate to where you already are, before what would happen is the 'animation' would play, but you don't move (as it's in the same place) which feels really weird if you then try and move the camera (this is easy to do when quickly moving between view bookmarks).

We use an optional so `cameraAnimation` can be fully 'unset' when not in use - this is to stop users spamming 'Go to position' when initially first moving to the new location. I did add several tests to replicate some of the edge cases, let me know if there's another you think is worth adding (we can hop on a call to discuss if needed). Thanks!

mbalfour-amzn: I guess what I was thinking was more of this:
1. If an animation is currently playing, and a new endpoint is requested that IsClose() to the current endpoint, then the current one should get adjusted to the new endpoint without restarting the animation.
2. If no animation is currently playing, and a new endpoint is requested that IsClose() to the current location, it would just immediately jump to the new value.
This way it would preserve the new request, even if it's really close to the previous one, without introducing the restarting animation problem that you're describing. But it's also super super nitpicky and not something that I was overly concerned about, just figured it was worth asking.

mbalfour-amzn (approved):

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Viewport/ModularViewportCameraController.cpp
hultonha: I guess my feeling with point 1 is, if `IsClose()` is true, then it'd be equivalent to (as far is the user is concerned) to updating it and not starting the animation again, so I personally don't think it's necessary to update it.

Also with point 2, if the requested location is `IsClose` to the current position, there's no point in jumping there as you're already there, so we just skip doing that too 😝

So I don't think there is an edge case that would result in an observable negative outcome, we're just talking about slightly different approaches reaching the same end.

Code/Editor/Lib/Tests/Camera/test_EditorCamera.cpp
```diff
@@ -584,6 +707,14 @@ namespace UnitTest
             const bool actualCameraGoToPositionInstantly = SandboxEditor::CameraGoToPositionInstantlyEnabled();
             EXPECT_THAT(expectedCameraGoToPositionInstantly, Eq(actualCameraGoToPositionInstantly));
         }
+
+        {
+            const float existingCameraGoToPositionDuration = SandboxEditor::CameraGoToPositionDuration();
+            const float expectedCameraGoToPositionDuration = !existingCameraGoToPositionDuration;
```
greerdv, line 713: Using ! with a float reads a bit weird to me.

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Viewport/ModularViewportCameraController.cpp
```diff
@@ -254,18 +254,24 @@ namespace AtomToolsFramework
...
             {
                 return t * t * t * (t * (t * 6.0f - 15.0f) + 10.0f);
             };
 
-            m_cameraAnimation.m_time = AZ::GetClamp(
-                m_cameraAnimation.m_time +
-                    (event.m_deltaTime.count() / ModularViewportCameraControllerRequests::InterpolateToTransformDuration),
-                0.0f,
-                1.0f);
+            if (m_cameraAnimation->m_duration == 0.0f)
```
greerdv, line 264: I'm always a bit wary about exact equality tests with floats.

greerdv (approved):

Code/Editor/Lib/Tests/Camera/test_EditorCamera.cpp
```diff
@@ -264,6 +271,47 @@ namespace UnitTest
...
+
+        bool nextInterpolationBegan = false;
+        AtomToolsFramework::ModularViewportCameraControllerRequestBus::EventResult(
+            nextInterpolationBegan,
+            TestViewportId,
+            &AtomToolsFramework::ModularViewportCameraControllerRequestBus::Events::InterpolateToTransform,
+            secondTransformToInterpolateTo,
+            InterpolateToTransformDuration);
+
+        bool interpolating = true;
```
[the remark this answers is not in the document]
hultonha: Good call will fix (likely a copy-pasta fail sorry)

Code/Editor/Lib/Tests/Camera/test_EditorCamera.cpp
```diff
@@ -288,18 +335,54 @@ namespace UnitTest
...
 
         // Then
         EXPECT_THAT(initialInterpolationBegan, ::testing::IsTrue());
         EXPECT_THAT(interpolating, ::testing::IsFalse());
         EXPECT_THAT(nextInterpolationBegan, ::testing::IsTrue());
     }
 
+    TEST_F(EditorCameraFixture, CameraCannotInterpolateToSamePositionWithoutMovingFromIt)
+    {
+        const auto& transformToInterpolateTo = AZ::Transform::CreateTranslation(AZ::Vector3(10.0f, 10.0f, 10.0f));
```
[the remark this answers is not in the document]
hultonha: That's true but I don't think doing both hurts, I'd prefer to leave as-is for now as this the editor behavior I wanted to check (I could maybe add both in future)

Code/Editor/Lib/Tests/Camera/test_EditorCamera.cpp
hultonha: It should be ` + 1.0f` like the others (just to show it's changed) - so crazy this compile without issue 😅

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Viewport/ModularViewportCameraController.cpp
hultonha: I did consider this, but as it's only guarding against a divide by zero, I think the exact check is okay here? 🤔 If you think very small values could still cause a problem I can change it but I explicitly thought in this case it was correct/safe

greerdv: If it's just to avoid a DBZ I guess it's fine. I did just briefly wonder if -0.0f was considered equal to 0.0f but did a quick check and apparently it is.

PR #13675, opened 2022-12-13
Shape translation offset support for reflection probe component

greerdv:
## What does this PR do?
Adds shape translation offset support for reflection probe component.

## How was this PR tested?
Manual testing in editor.


hultonha (approved):
LGTM! :+1:

(if it's possible to add any simple tests, even as a follow-up I'd be supportive - but we could leave that for another day)

dmcdiarmid-ly (approved):

PR #13678, opened 2022-12-13
Fixes for Android Unit Tests

amzn-phist:
## What does this PR do?

Fixes some issues that were preventing Android Unit Tests from being run.
Fixes an issue with dynamic modules on Android.  When the module name is changed to a real filename via adding the `AZ_TRAIT_OS_DYNAMIC_LIBRARY_PREFIX` and `AZ_TRAIT_OS_DYNAMIC_LIBRARY_EXTENSION`, write this change back to the `DynamicModuleHandle`.

## How was this PR tested?

Running `AzTestRunner` on various modules (AzCore, FastNoise, GradientSignal, Terrain, ...).  Some modules still have test errors, others now pass, but they can all run their tests now.


Code/Framework/AzCore/Tests/AZStd/Hashed.cpp
```diff
@@ -1420,7 +1420,11 @@ namespace UnitTest
     TYPED_TEST_CASE(HashedSetDifferentAllocatorFixture, SetTemplateConfigs);
 
 #if GTEST_HAS_DEATH_TEST
+#if AZ_TRAIT_DISABLE_FAILED_DEATH_TESTS
+    TYPED_TEST(HashedSetDifferentAllocatorFixture, DISABLED_InsertNodeHandleWithDifferentAllocatorsLogsTraceMessages)
```
lumberyard-employee-dm, line 1424: Do you know what the issue with the Google Death test are?
Is it not correctly causing a non-zero exit code?

lumberyard-employee-dm (approved):

Code/Framework/AzCore/Tests/AZStd/Hashed.cpp
amzn-phist: I'll have to run them again, but encountering one would prevent running the remaining tests.

lumberyard-employee-dm: Ok then it is probably an issue with the Android runtime then, where it maybe having issues forking processes in an isolated manner..

cgalvan (approved):

Code/Framework/AzCore/Tests/AZStd/Hashed.cpp
amzn-phist: It does log a warning about forking process with threads, but the test itself is aborting while operating on a null `FILE*`.  Needs some investigation.

moraaar (commented):
This couldn't have come in a better moment. Thanks @amzn-phist !

PR #13679, opened 2022-12-13
Move AutomatedTesting::Atom_Main_Null_Render_02 test to periodic suite temporarily.

jromnoa:
## What does this PR do?
1. This PR moves the AutomatedTesting::Atom_Main_Null_Render_02 test to periodic suite since it is intermittently failing in AR.
2. Some example builds demonstrating this failure below:
https://jenkins.build.o3de.org/blue/rest/organizations/jenkins/pipelines/O3DE/pipelines/development/runs/3600/nodes/876/log/
https://jenkins.build.o3de.org/blue/rest/organizations/jenkins/pipelines/O3DE/pipelines/development/runs/3630/nodes/1047/log/
https://jenkins.build.o3de.org/blue/rest/organizations/jenkins/pipelines/O3DE/pipelines/development/runs/3595/nodes/865/log/

## How was this PR tested?
There was no need to test this as it is simply moving suites from main suite to periodic suite plus the issue has a low repro rate (0% locally so far and 1/30 or so on the AR runs).

### DIFF
--- a/AutomatedTesting/Gem/PythonTests/Atom/CMakeLists.txt
+++ b/AutomatedTesting/Gem/PythonTests/Atom/CMakeLists.txt
@@ -35,7 +35,7 @@ if(PAL_TRAIT_BUILD_HOST_TOOLS AND PAL_TRAIT_BUILD_TESTS_SUPPORTED)
     )
     ly_add_pytest(
         NAME AutomatedTesting::Atom_Main_Null_Render_02
-        TEST_SUITE main
+        TEST_SUITE periodic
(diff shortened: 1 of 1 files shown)


jromnoa:
I have an internal ticket to investigate this intermittent failure next sprint, but for now it should be moved to periodic as in the last 60 or so runs its failed 3 times. It also has failed in PR branches not just development with the same type of errors.

smurly (approved):
this represents 1/3 of null render atom component tests so we should prioritize re-enabling this suite.

NeilGitGud (approved):

PR #13680, opened 2022-12-13
Fix clang compile and add extra flags feature

lawsonamzn:
## What does this PR do?

Fixes compilation errors with clang, and adds a feature mainly meant for clang tools like the clang build analyzer and others (which require adding flags to compiler).

It also fixes the OpenEXR package to no longer set "/EHsc" (a MSVC-only flag) on Windows, instead, it sets it only when in MSVC.

## How was this PR tested?

Building this with MSVC, as well as building it with clang on windows.

to build it on windows with clang, make sure llvm clang is installed and on the path, and set the CMAKE_C_COMPILER and CMAKE_CXX_COMPILER  variables in cmake to both be clang++

You need to be inside a visual studio command prompt to do this, since some tools still invoke 'link' from vs.


lawsonamzn:
Note that the associated change to the 3p package is in https://github.com/o3de/3p-package-source/pull/172

cgalvan (approved):

hultonha (approved):
Nice! 😄

PR #13684, opened 2022-12-13
Exposing PrefabOverridePublicInterface methods to behavior context

raktalen:
Signed-off-by: Nikhil Raktale <raktalen@amazon.com>

## What does this PR do?
Exposing a few methods, specifically 
PrefabOverridePublicInterface::AreOverridesPresent and PrefabOverridePublicInterface::RevertOverrides
to the behavior context

## How was this PR tested?

The code builds without errors and all override functionality is preserved. Checked the python symbols file and made sure that the exposed methods do show up there.


Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverridePublicHandler.h
```diff
@@ -19,12 +19,14 @@ namespace AzToolsFramework
         class PrefabFocusInterface;
         class PrefabSystemComponentInterface;
 
-        class PrefabOverridePublicHandler : private PrefabOverridePublicInterface
+        class PrefabOverridePublicHandler : public PrefabOverridePublicRequestBus::Handler
         {
```
forkercat, line 23: curious why we can remove `private PrefabOverridePublicInterface` and it still works?

forkercat (approved):

michabr (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverridePublicHandler.h
raktalen: The handler is templated off of the Interface

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverridePublicHandler.cpp
```diff
@@ -34,6 +35,20 @@ namespace AzToolsFramework
         PrefabOverridePublicHandler::~PrefabOverridePublicHandler()
         {
             AZ::Interface<PrefabOverridePublicInterface>::Unregister(this);
+            PrefabOverridePublicRequestBus::Handler::BusDisconnect();
```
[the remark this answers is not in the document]
raktalen: The convention to disconnect in the reverse order was created in order to combat Order of operation issues with one system coming online after the other and buses becoming ready to respond in that order. There is no dependency relationship between Unregister and bus dc for the same interface, technically both are invalid at the exact same time, but there is no way to ensure atomicity of that operation.

srikappa-amzn (approved):
Just curious. If the function `AZStd::optional<OverrideType> GetOverrideType(AZ::EntityId entityId)` is exposed to behavior context too, will python functions  be able to fetch the return value since its an optional custom enum? If not, we should probably move it from this interface to a different one and make a rule that all functions in this public interface should be scriptable for consistency

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverridePublicHandler.h
lsemp3d: There's some info here with regards to choosing private/public EBus implementations:

https://www.o3de.org/docs/user-guide/programming/components/entity-system-pg-components-ebuses-best-practices/#making-functions-public-or-protected

lawsonamzn:
You may have to rebase onto development to get this fixed, I had similar issues in other PRs.

Issue #13646, opened 2022-12-12
Editor: A bunch of entity created with just transform were created when dragging an FBX contains skinned mesh to the main level. 


amzn-rhhong:
**Describe the bug**
I tried to drag an fbx to the main editor. The newly created entity contains many transform object. It doesn't really make any sense.

**Assets required**
[YellowHood.zip](https://github.com/o3de/o3de/files/10204687/YellowHood.zip)

**Steps to reproduce**
1. Create an empty level
2. Drag the above fbx to main level

**Expected behavior**
One entity created, with an actor component and a transform component

**Actual behavior**
Many random transform objects are created.

**Screenshots/Video**

https://user-images.githubusercontent.com/69218254/206964210-a8bd878f-3550-46be-b7b9-bb8bcbae2bfc.mp4

AMZN-stankowi:
This is behaving currently as designed (but the work isn't final yet).

When you drag and drop a root FBX object, it instantiates the associated procedural prefab and then immediately detaches it.

Right now the default procedural prefabs are only built to work with FBX files that contain static geometry. If you use a default procedural prefab with an FBX containing a skinned mesh, the default prefab tends to create an entity per node in the skeleton.

This will feel better when default procedural prefabs support skinned meshes.

hultonha:
@AMZN-stankowi Can we please change the defaults so proc prefabs are disabled by default for actors and animations. It would be good to just disable this for now until we can revisit it. Thanks!

AMZN-stankowi:
@hultonha we need to have farther discussion on this. I started a thread on another too. There are complications in doing so, so we'll need to talk through those before committing.

hultonha:
Okay understood, I'd like to know what those downsides are and see if we can mitigate them. From the reports we've got it sounds like disabling proc prefabs for actors is still a net positive. Thanks!

PR #13701, opened 2022-12-14
Auto generated mesh groups are now marked read only

AMZN-stankowi:
## What does this PR do?

Mesh groups that are automatically generated and not manually added are now marked read only.

image

## How was this PR tested?

Manual testing: Loading several different scene files in the scene settings UI, verified read only worked as expected.
Ran the existing automated FBX tests + scene settings UI tests, they all passed.
Ran the new scene settings UI test, it passed.


lmbr-pip:
@AMZN-stankowi Is this for https://github.com/o3de/o3de/issues/13646 ?

AMZN-stankowi:
@lmbr-pip loosely, this will make it more obvious where those mesh groups come from, but it won't make the drag and drop itself less confusing

Code/Tools/SceneAPI/SceneCore/DataTypes/Rules/IReadOnlyRule.h
```diff
@@ -0,0 +1,30 @@
...
+#include <SceneAPI/SceneCore/DataTypes/Rules/IRule.h>
+
+namespace AZ
+{
+    namespace SceneAPI
+    {
+        namespace DataTypes
+        {
+            // Disables UI interaction for the node and children of the node with this rule.
+            class IReadOnlyRule : public IRule
```
amzn-rhhong, line 21: I'm not really a fan of creating a inheriting class between IRules and derived class (e.g ISkinRules) just to modify the tool tips. 
What about just add the modifyToolTips function to IRules and let the derived class handle the actual part of tool tip modification?

amzn-rhhong:
With this change I think it's even more crucial to add instructions on how to turn off the proc prefab in the tool tips.
Also maybe more UX inputs on this. Personally I think we could do something like this:
ui mockup procprefab

Code/Tools/SceneAPI/SceneCore/DataTypes/Rules/IReadOnlyRule.h
AMZN-stankowi: I originally didn't have an IReadOnlyRule class, but it caused some serialization and linker issues due to how the systems are setup. SceneAPI / SceneCore are divided into a few pieces that make it tough to get everything linking in the right places.

amzn-mike (approved):

Code/Tools/SceneAPI/SceneCore/DataTypes/Rules/IReadOnlyRule.h
amzn-rhhong: Ok I think I understand the code a little better now. It's not a "ReadOnly" Rule - instead it's basically a rule that makes whatever container holds it to become read only, correct?
Maybe there could be a better name for this? UnmodifiableRule?

Code/Tools/SceneAPI/SceneCore/DataTypes/Rules/IReadOnlyRule.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
```
junbo75, line 1: Nit: May worth moving this line after the copyright comment.

junbo75 (approved):

AMZN-stankowi:
Converting back to a draft while I work through an automated test. I'm going to be updating the branch with some WIP changes before going on holiday break.

Code/Tools/SceneAPI/SceneCore/DataTypes/Rules/IReadOnlyRule.h
AMZN-stankowi: Done, and I updated a bunch of other ones to prevent this copy & paste error from happening again.

Code/Tools/SceneAPI/SceneCore/DataTypes/Rules/IReadOnlyRule.h
AMZN-stankowi: Sure, I've renamed it to that. I didn't rename it everywhere because it's using the ReadOnly system in the reflected property editor, but I renamed the rule.

AMZN-stankowi:
> With this change I think it's even more crucial to add instructions on how to turn off the proc prefab in the tool tips. Also maybe more UX inputs on this. Personally I think we could do something like this: ui mockup procprefab

I've updated the description to mention how to remove the group, and I've removed the delete button from the disabled mesh groups.

amzn-rhhong (approved):

PR #13704, opened 2022-12-14
[Draft PR] Experimental work-in-progress to reduce memory allocations around network serialization buffers

AMZN-Olex:
## What does this PR do?

This is a work in progress attempt to reduce memory allocations of large replication buffers (16k bytes each) inside NetworkEntityUpdateMessage:
```cpp
m_data = AZStd::make_unique<AzNetworking::PacketEncodingBuffer>();
```

Can be turned on and off with bg_useBufferPoolForNetworkEntityUpdateMessages.

## How was this PR tested?

Currently measuring performance and testing with o3de-multiplayersample.


kberg-amzn (commented):
Interesting.. If we ever multithreaded sends again, would this be thread safe?
Here's just a quick idea, what if each replicator? maybe? .. unsure where the most appropriate layer might be.. owned a PacketEncodingBuffer explicitly meant only for moving entity update data around.

PR #13705, opened 2022-12-14
Remove the revert overrides outliner option

michabr:

## What does this PR do?

This PR removes the context menu option of reverting overrides on an entity in the outliner. The main reason is the added complexity caused by the sort order component which currently sits on a child's parent. So reverting overrides on a parent that has a child added as an override, or reverting overrides on an added child would cause the sort order component of the parent to be in an invalid state.

## How was this PR tested?

Manual testing in editor.


forkercat (approved):

michabr:
It was decided to leave this option available.

Issue #10588, opened 2022-07-08
Update PyTest and custom plugins


Kadino:
**Is your feature request related to a problem? Please describe.**
A recent open source maintenance issue with the atomicwrites library has forced PyTest to remove it as a dependency:  https://github.com/pytest-dev/pytest/pull/10115

Breakage to O3DE's installation mechanism has been patched via https://github.com/o3de/o3de/pull/10586 however it would still be appropriate to update PyTest and to reduce dependency on atomicwrites.

**Describe the solution you'd like**
PyTest is updated to the latest version, along with any required updates to O3DE's custom plugins and configuration files. Remove any dependencies no longer required by pytest, such as the module 'py'

**Describe alternatives you've considered**
N/A

**Additional context**
Pytest version is managed by https://github.com/o3de/o3de/blob/development/python/requirements.txt
Plugins exist in https://github.com/o3de/o3de/tree/development/Tools/LyTestTools
Configuration files include https://github.com/o3de/o3de/search?l=INI&q=pytest
Changes to PyTest's API may need to be reflected in https://github.com/o3de/o3de/blob/development/cmake/LYTestWrappers.cmake

Kadino:
Moved to critical as there is a security flaw which needs patching.

Kadino:
Confirmed custom plugins do in fact fail after upgrading. The majority of failures are emitted from the EditorTest plugin.

Kadino:
Isolated the single plugin failure to unexpectedly encountering attribute `__test__` set explicitly to `False` for class EditorTest

Kadino:
There's another underlying failure, which is likely from an early-out on an unexpected node subclass in https://github.com/o3de/o3de/blob/development/Tools/LyTestTools/ly_test_tools/_internal/pytest_plugin/multi_testing.py#L47

Kadino:
Still don't quite have this narrowed down. Perhaps the pytest_pycollect_makeitem custom registration is being silently aborted by pluggy due to an interface change?

Issue #12685, opened 2022-10-19
Bug Report: Py library is insecure and needs patching


lmbr-pip:
**Describe the bug**
Py versions <= 1.11.0

> The py library through 1.11.0 for Python allows remote attackers to conduct a ReDoS (Regular expression Denial of Service) attack via a Subversion repository with crafted info data, because the InfoSvnCommand argument is mishandled.

CVSS score of 5.3 puts this as major.

**Assets required**
None

**Steps to reproduce**
1. Go to 'requirements.txt': https://github.com/o3de/o3de/blob/development/python/requirements.txt#L222
2. Look at the version of Py in use
3. Look at CVE for PY: https://cwe.mitre.org/data/definitions/1333.html

**Expected behavior**
PY is patched once a new version is available

**Actual behavior**
Py is insecure. No patched version exists

**Screenshots/Video**
If applicable, add screenshots and/or a video to help explain your problem.

**Found in Branch**
Dev

**Desktop/Device (please complete the following information):**
 - Device: [e.g. PC, Mac, iPhone, Samsung] 
 - OS: [e.g. Windows, macOS, iOS, Android]
 - Version [e.g. 10, Monterey, Oreo]
 - CPU [e.g. Intel I9-9900k , Ryzen 5900x, ]
 - GPU [AMD 6800 XT, NVidia RTX 3090]
 - Memory [e.g. 16GB]

**Additional context**

lmbr-pip:
Assigning to SIG/security to identify owners of Py and confirm issue details.

dshmz:
SIG-Build checked this issue and Py module is not owned by SIG-Build, we need to figure out which SIG owns it. It should go to the overall triage to get the assignment to correct SIG.

lmbr-pip:
SIG/testing touched this last (see https://github.com/o3de/o3de/pull/7155). Routing to SIG/testing for input.

Kadino:
I'll look into why this is required. Preferably the package can be removed, as no secure version exists

Kadino:
The only direct dependency is pytest, which is a fairly critical python dependency:
```bash
python -m pipdeptree -r -p py
py==1.11.0
  - pytest==6.2.5 [requires: py>=1.8.2]
```

Confirmed owned by @o3de/sig-testing

Kadino:
[...]

PR #13707, opened 2022-12-14
Update PyTyest, custom plugin, and remove unused dependency

Kadino:
Signed-off-by: sweeneys <sweeneys@amazon.com>

## What does this PR do?

Resolves https://github.com/o3de/o3de/issues/10588 and https://github.com/o3de/o3de/issues/12685 by updating to the latest version of PyTest. Also includes minor updates to make it easier to run tests locally.

## How was this PR tested?

Tests pass when running the following command with CWD in o3de/build:
`ctest -C profile -L "(SUITE_smoke|SUITE_main)" --no-tests=error --parallel 12 --output-on-failure`

### DIFF
--- a/Tools/LyTestTools/tests/integ/__init__.py
+++ b/Tools/LyTestTools/tests/integ/__init__.py
@@ -1,6 +1,10 @@
-"""
+r"""
 Copyright (c) Contributors to the Open 3D Engine Project.
 For complete copyright and license terms please see the LICENSE at the root of this distribution.
 
 SPDX-License-Identifier: Apache-2.0 OR MIT
+
+Tests in this directory require setting the `build-directory` parameter, examples:
+python -m pytest --build-directory="c:\git\o3de\build\bin\profile"
+python -m pytest --build-directory="~/git/o3de/build/bin/profile"
 """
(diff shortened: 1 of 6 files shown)


python/requirements.txt
```diff
@@ -219,10 +219,6 @@ psutil==5.8.0 \
     --hash=sha256:f4634b033faf0d968bb9220dd1c793b897ab7f1189956e1aa9eae752527127d3 \
     --hash=sha256:fcc01e900c1d7bee2a37e5d6e4f9194760a93597c97fee89c4ae51701de03563
     # via requirements.txt
-py==1.11.0 \
```
evanchia-ly-sdets, line 222: Was this intended to be removed?

Kadino: Yes, as linked as a security ticket.

Kadino:
Marking PR as draft as it appears that it requires a Jenkins clean-build (and possibly additional steps) to prove functionality.

Tools/LyTestTools/tests/integ/__init__.py
```diff
@@ -1,6 +1,10 @@
-"""
+r"""
```
smurly, line 1: why are you prepending an r to this docstring block? I'm guessing this is something I don't know.

evanchia-ly-sdets (approved):

Tools/LyTestTools/tests/integ/__init__.py
Kadino: It makes it a raw string. In this case it make sure it contains literal backslashes instead of using backslash as an escape character. I wanted the string below to be copyable from the file into a Windows CLI, while also not making garbled text in the docstring.

smurly (approved):

PR #13719, opened 2022-12-15
Updated var names for Templates/ScriptCanvasNode and added a required step to the README

HillinessStudios:
Updated var names for Templates/ScriptCanvasNode and added a required step to the README

## What does this PR do?

The templates CMakeList references $GemName, but a newly created Gem uses $gem_name in its Code/CMakeList though.

To make on boarding easier the required add directory directive to the Code/CMakeList was added to the README


lumberyard-employee-dm:
@HillinessStudios Looks like your commit is missing the Developer Certificate of Origin signoff.
You can sign your commit by using following the steps in the [Details](https://github.com/o3de/o3de/pull/13719/checks?check_run_id=10124912405) link in the actions section of the PR.

Since you only have a single commit in this case, you can amend the commit with a signoff using `git commit --amend -s` to amend the commit with your signoff and then force push the changes back to your branch.

Templates/ScriptCanvasNode/Template/README.md
```diff
@@ -31,6 +31,10 @@ scripts\o3de register -esgp Gems\${GemName} -es Gems\${GemName}\Code\Source\${Na
 
 Configure & Generate your projects
 
+To your YourGem Code\CMakeList add your MyScriptCanvasNodes directory by appending
+add_subdirectory(Source/MyScriptCanvasNodes)
```
lumberyard-employee-dm, line 35: This step is not needed.

Once the folder container containing the ScriptCanvas Node directory is registered with the Gem in it's gem.json file via the "external_subdirectories" key, then add_subdirectory will be called on it automatically via the subdirectory visitation logic in [Subdirectories.cmake](https://github.com/o3de/o3de/blob/development/cmake/Subdirectories.cmake#L327-L356) when that gem is active.

The Atom gem has an [example](https://github.com/o3de/o3de/blob/development/Gems/Atom/gem.json#L19-L32) of visiting sub-gems and regular external_subdirectories.
A "regular" external_subdirectory is just a directory with a CMakeLists.txt in it, but no `gem.json`. 
A "Gem" is an external_subdirectory with a `gem.json` in it.

Any external_subdirectories registered with the current project, the engine being used or any gem's being used by the current project will have recursively have all of it's children "external_subdirectories" entries gathered and then `add_subdirectory` invoked on each path.

lumberyard-employee-dm (changes requested):

Templates/ScriptCanvasNode/Template/README.md
lsemp3d: You're right, I think it was carryover from the first iteration that I missed cleaning. Thanks for the update!

lsemp3d (approved):

Templates/ScriptCanvasNode/Template/README.md
HillinessStudios: @lumberyard-employee-dm the external sub directories apparently doesn't pass on the gem_name variable, at least it produces build errors for me "The target .Private.Object was not found, if the gem was not created using"

Which is unfortunate since therefor we need to pass the -r ${GemName} YourGem to the create-from-template
While with the manual addition of the add_subdirectory this is working.
Since the bit that gets replaced has the format ${something} it is prone to user error due to shell substitution. 

I have some further changes locally which I will push w/ further feedback from you and the amended commit to satisfy the Certificate of Origin.

lumberyard-employee-dm: In that case if the gem name is needed it can be queried from the nearest ancestor gem.json using the [o3de_find_ancestor_gem_root](https://github.com/o3de/o3de/blob/development/cmake/Gems.cmake#L18-L23) method.

That can be added to the template CMakeLists.txt to query the gem name and therefore the need to specify `-r ${GemName} <YourGem>` argument isn't needed.

lumberyard-employee-dm: Try adding this code to the top of the CMakeLists.txt in the template
```
# Query the gem name from the gem.json file if possible
# otherwise fallback to using ${GemName}
o3de_find_ancestor_gem_root(gem_path gem_name "${CMAKE_CURRENT_SOURCE_DIR}")
if (NOT gem_name)
    set(gem_name "${GemName}")
endif()
```

HillinessStudios:
@lumberyard-employee-dm so I force pushed w/ -s for signed off and added the o3de_find_ancestor_gem_root you suggested, also changed the wording for the readme. I also added the template to the Templates register so one is able to use -tn; I hope that omitting it wasn't intentional by @lsemp3d 

The fix to the AutoGen jinja technically doesn't belong to these changes but toying around w/ the ScriptCanvasNode I observed that having more than one return value was broken so this got bundled in. We can do a separate PR for this change if you like.

lsemp3d:
> @lumberyard-employee-dm so I force pushed w/ -s for signed off and added the o3de_find_ancestor_gem_root you suggested, also changed the wording for the readme. I also added the template to the Templates register so one is able to use -tn; I hope that omitting it wasn't intentional by @lsemp3d 
> 
> The fix to the AutoGen jinja technically doesn't belong to these changes but toying around w/ the ScriptCanvasNode I observed that having more than one return value was broken so this got bundled in. We can do a separate PR for this change if you like.

It was not intentional, thanks for these fixes!

lumberyard-employee-dm (approved):
Looks good to me.

lumberyard-employee-dm:
@HillinessStudios It looks like the build has failed to some Asset processor failures.
Can you trying updating your branch to latest development?
The build can be kicked off again afterwards.

HillinessStudios:
pulled latest development as requested

cgalvan (approved):

cgalvan:
A new build has been kicked off.

PR #13721, opened 2022-12-15
Removed the loading of the legacy `system_<platform>_<asset_platform>.cfg` files

lumberyard-employee-dm:
The settings registry has replaced the loading of platform specific setreg files using the Platform Abstraction Layer to determine platform specific folders.
This is detailed in the documentation
[here](https://www.o3de.org/docs/user-guide/settings/developer-documentation/#specialization-system-in-depth)

The settings registry can also issue console commands which is described on the [Issue Console
Commands](https://www.o3de.org/docs/user-guide/settings/issue-az-console-commands/) page


cgalvan (approved):

amzn-phist (approved):
Question, are all the settings being deleted already set up somewhere in `.setreg` files?  If not, do they need to be added?

lumberyard-employee-dm:
> Question, are all the settings being deleted already set up somewhere in `.setreg` files? If not, do they need to be added?

- The `log_includeTime` CVAR would need it's default updated to 1
- The `sys_float_exceptions` CVAR would need it's default updated to 0
- The `log_RemoteConsoleAllowedAddresses` CVAR would need it's default updated to "127.0.0.1"
- The `sys_localization_format` CVAR would need it's default updated to 0
- The `r_MaxWidth` and `r_MaxHeight` CVAR doesn't exist
- The `r_ShadersAllowCompilation` CVAR doesn't exist
- The `r_WidthAndHeightAsFractionOfScreenSize` CVAR doesn't exist
- The `sys_affinity` CVAR doesn't exist
- The `sys_physics_CPU` CVAR doesn't exist
- The `r_fullscreen` setting needs to be set to 1 on iOS and can be left as the default 0 for other platforms
- The `sys_PakLogInvalidFileAccess` needs to be set to 1 for iOS and Android
- The `log_spamdelay` CVAR is commented out.
- The `r_displayInfo` is set to 3 for Linux, but it is fine to use 1 as the default
- The `r_width` and `r_height` default to 1280x720 on Windows, Linux, MacOS, but the default values for them are 1920x1080
- 
-

ParasiteSt:
Question, do developers with an already created project have any additional work to do when this is merged? Should the developers switch these values on their previously created levels, or will the be updated automatically?

lumberyard-employee-dm:
> Question, do developers with an already created project have any additional work to do when this is merged? Should the developers switch these values on their previously created levels, or will the be updated automatically?

I am adding settings registry files that invoke these console commands in the `<engine-root>/Registry` folder.

cgalvan (approved):

cgalvan (approved):

lawsonamzn (approved):
Looks good to me.  Might save a little bit of startup time.

Tools/LyTestTools/ly_test_tools/o3de/settings.py
```diff
@@ -31,36 +31,13 @@ def __init__(self, temp_path, resource_locator):
     def get_temp_path(self):
         return self._temp_path
 
-    def modify_asset_processor_setting(self, setting, value):
```
evanchia-ly-sdets, line 34: Was this intended to be removed?

evanchia-ly-sdets (changes requested):

Tools/LyTestTools/ly_test_tools/o3de/settings.py
lumberyard-employee-dm: Yes, there are no calls to this method and it would have been incorrect to call this method since the asset processor file is no longer an .ini style file, but a setreg file([AssetProcessorPlatformConfig.setreg](https://github.com/o3de/o3de/tree/development/Registry)).

evanchia-ly-sdets (approved):

PR #13737, opened 2022-12-16
Jjjoness/create add button in the header of asset browser#13652

### LINKED ISSUE #13652: Create an Add new button in the header of Asset Browser
As a user, I need the ability to "add new" from the main navigation bar instead of just right click behavior. Currently, we already have the ability to add new via the right click menu. However, this is a hidden action. Right click should be a secondary behavior for add new instead of the primary. Primary will be this new "Add new" button.

The add new button should apply to the selected folder the user has selected. So this means if a user clicks onto a directory and then hits add new. The add new will be applied to that folder.

If no folder is selected, we should ask the user where they would like this file to be added to. Maybe via selector menu.

Image

jjjoness:
## What does this PR do?

This adds the missing + button to the Asset Browser. Functionally, this works the same as the context menu "Create".

Fixes #13652

## How was this PR tested?

Checked by creating new files in the Asset Browser.


Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.ui
```diff
@@ -82,6 +82,47 @@
...
+          </widget>
+         </item>
+         <item>
+          <widget class="QToolButton" name="m_createButton">
+           <property name="text">
+            <string>...</string>
+           </property>
+           <property name="icon">
+            <iconset>
+             <normaloff>:/stylesheet/img/UI20/plus.svg</normaloff>:/stylesheet/img/UI20/plus.svg</iconset>
```
cgalvan, line 115: What does this look like compared to our `img/UI20/Add.svg`. And if it is different, it seems like we should have UX investigate having a consistent plus sign image. @rainbj @amzn-leenguy

cgalvan (changes requested):

Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
```diff
@@ -115,6 +115,14 @@ AzAssetBrowserWindow::AzAssetBrowserWindow(QWidget* parent)
         m_tableModel->setDynamicSortFilter(true);
         m_ui->m_assetBrowserTableViewWidget->setModel(m_tableModel.data());
 
+        m_createMenu = new QMenu("Create Menu", this);
+        m_ui->m_createButton->setMenu(m_createMenu);
+        m_ui->m_createButton->setEnabled(true);
+        m_ui->m_createButton->setAutoRaise(true);
+        m_ui->m_createButton->setPopupMode(QToolButton::InstantPopup);
+
+        connect(m_createMenu, &QMenu::aboutToShow, this, [this] { AddCreateMenu(); });
```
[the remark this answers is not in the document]
jjjoness: I had originally some initialization in here but when I deleted it forgot this was a slot! Chjanged.

Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
```diff
@@ -209,6 +219,63 @@ AzAssetBrowserWindow::~AzAssetBrowserWindow()
...
+
+    if (entry->GetEntryType() == AssetBrowserEntry::AssetEntryType::Product)
+    {
+        entry = entry->GetParent();
+        if (!entry)
+        {
+            return;
+        }
+    }
+    fullFilePath = entry->GetFullPath();
```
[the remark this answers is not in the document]
jjjoness: Done

Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.ui
jjjoness: It is the same. Deleted the Plus and replaced with Add.

Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
```diff
@@ -115,6 +115,14 @@ AzAssetBrowserWindow::AzAssetBrowserWindow(QWidget* parent)
         m_tableModel->setDynamicSortFilter(true);
         m_ui->m_assetBrowserTableViewWidget->setModel(m_tableModel.data());
 
+        m_createMenu = new QMenu("Create Menu", this);
```
AMZN-daimini, line 118: nit: "Create Menu" may be easy to misunderstand, maybe it's better to label this "Create File Menu"?

AMZN-daimini (approved):

Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
jjjoness: @AMZN-daimini I'll change it to "Create New Asset Menu" to make it even easier😀

dtamkin1 (approved):

PR #13742, opened 2022-12-16
Half Float support plus other gpu perf related optimizations for mobile/VR pipeline.

moudgils:
Signed-off-by: moudgils <47460854+moudgils@users.noreply.github.com>

## What does this PR do?

Resolution - 1440 x 1584 across both eyes (75% of native resolution)
OpenXrTest level - Runs at ~35fps with these changes. Previously it was running at ~25fps

- Half float support - This was done via introducing a new keyword called 'real' and PAL-ifying it across different platforms
	- Half float support is enabled for android and ios
	- Most of the relevant material types like BasePBR, StandardPBR, EnhancedStandardPBR were converted over to use half floats. Any shaders that were not converted over will show precision loss warnings during compilation especially if they call into functions that switched over to use real keyword. This should be fine until we convert over all the shaders in future
	- Positions and UVs still use full float
- Removing all branches related to shader debugging code by default for the MultiView pipeline
- Disable reflection probe parallax correction for MultiView pipeline
- Disable reflection probe blending for MultiView pipeline
- Disable specular anti aliasing for MultiView pipeline
- Fixed shader pipeline to ensure that the correct AZSLHeader is pre-appended.
- Fixed crash when trying to run openxr with DX12 rhi
- Fixed a crash related to RenderDoc gpu capture when openxr is enabled
- Various other cleanup tasks

## How was this PR tested?

Extensively tested ASV and OpenXrTest on PC + Quest2 (Vr + non VR mode). Ios was not tested as Mac AP is not working at the moment although this work as no metal specific work so in theory ios should just work. 

image


Gems/Atom/Feature/Common/Assets/ShaderLib/Atom/Features/BlendUtility.azsli
```diff
@@ -45,10 +45,13 @@ float3 ApplyTextureBlend(float3 color, float3 blendColor, float factor, TextureB
     {
         // [GFX TODO][ATOM-3237] See if we can avoid these extra color space conversions...
         // To get blend behavior that mimics photoshop, convert to sRGB
-        float3 blendColorSRGB = TransformColor(blendColor, ColorSpaceId::ACEScg, ColorSpaceId::SRGB);
-        float3 colorSRGB = TransformColor(color, ColorSpaceId::ACEScg, ColorSpaceId::SRGB);
+        //Acessg -> srgb
+        real3 blendColorSRGB = AcesCg_To_LinearSrgb(blendColor);
+        blendColorSRGB = LinearSrgb_To_Srgb(blendColorSRGB);
```
jeremyong-az, line 50: For blending, don't we want to stay in linear SRGB?

Gems/Atom/Feature/Common/Assets/ShaderLib/Atom/Features/PBR/LightingUtils.azsli
```diff
@@ -11,13 +11,17 @@
 #include <Atom/RPI/Math.azsli>
 
 //! Transform cubemap coordinates to the expected cubemap orientation
+half3 GetCubemapCoords(half3 original)
+{
+    return half3(-original.x, original.z, -original.y);
```
jeremyong-az, line 16: Should this function accept `real` types?

Gems/Atom/Feature/Common/Assets/ShaderLib/Atom/Features/PBR/Surfaces/StandardSurface.azsli
```diff
@@ -79,15 +83,15 @@ class Surface
...
+    const real screenVariance = 0.25;
+    const real varianceThresh = 0.18;
 
     // Specular Anti-Aliasing
-    float3 dndu = ddx_fine( normal );
-    float3 dndv = ddy_fine( normal );
-    float variance = screenVariance * (dot( dndu , dndu ) + dot( dndv , dndv ));
-    float kernelRoughnessA2 = min(2.0 * variance , varianceThresh );
-    float filteredRoughnessA2 = saturate ( roughnessA2 + kernelRoughnessA2 );
+    real3 dndu = real3(ddx_fine( float3(normal) ));
```
jeremyong-az, line 90: Note that down/upcasting to and from `half` precision floats will reduce perf. In this case you might be better off leaving `normal` as a `float3`.

Gems/Atom/Feature/Common/Assets/ShaderLib/Atom/Features/ParallaxMapping.azsli
```diff
@@ -399,33 +399,33 @@ struct PixelDepthOffset
...
-    float scaleZ = length(objectToWorldMatrix[2].xyz);
-    worldOffset *= float3(scaleX, scaleY, scaleZ);
+    real scaleX = length(objectToWorldMatrix[0].xyz);
+    real scaleY = length(objectToWorldMatrix[1].xyz);
+    real scaleZ = length(objectToWorldMatrix[2].xyz);
+    worldOffset *= real3(scaleX, scaleY, scaleZ);
 
-    float3 worldOffsetPosition = posWS + worldOffset;
-    float4 clipOffsetPosition = mul(viewProjectionMatrix, float4(worldOffsetPosition, 1.0));
+    float3 worldOffsetPosition = posWS + float3(worldOffset);
```
jeremyong-az, line 427: Should WPO be `real3` also to avoid the casts?

Gems/Atom/Feature/Common/Assets/Shaders/Materials/StandardPBR/StandardPBR_PixelGeometryEval.azsli
```diff
@@ -45,20 +45,28 @@ PixelGeometryData EvaluatePixelGeometry_StandardPBR(
...
 {
-    float4x4 objectToWorld = ObjectSrg::GetWorldMatrix();
-    float3x3 objectToWorldIT = ObjectSrg::GetWorldMatrixInverseTranspose();
+    real4x4 objectToWorld = real4x4(ObjectSrg::GetWorldMatrix());
+    real3x3 objectToWorldIT = real3x3(ObjectSrg::GetWorldMatrixInverseTranspose());
 
-    float3 vertexNormal, vertexTangent, vertexBitangent;
-    ConstructTBN(IN.normal, IN.tangent, objectToWorld, objectToWorldIT, vertexNormal, vertexTangent, vertexBitangent);
+    real3 vertexNormal, vertexTangent, vertexBitangent;
+    ConstructTBN(real3(IN.normal), real4(IN.tangent), objectToWorld, objectToWorldIT, vertexNormal, vertexTangent, vertexBitangent);
```
jeremyong-az, line 60: Should `ConstructTBN` accept `real` types instead?

Gems/Atom/Feature/Common/Assets/ShaderLib/Atom/Features/PBR/Decals.azsli
```diff
@@ -108,10 +112,10 @@ void ApplyDecal(uint currDecalIndex, inout Surface surface)
         
         const float decalAttenuation = GetDecalAttenuation(surface.normal, decalRot[2], decal.m_angleAttenuation);
         const float albedoOpacity = baseMap.a * decal.m_opacity * decalAttenuation;
-        surface.albedo = lerp(surface.albedo, baseMap.rgb, albedoOpacity);      
+        surface.albedo = real3(lerp(surface.albedo, baseMap.rgb, albedoOpacity));
```
moudgils, line 115: I will convert this function over to use real in the next commit

Gems/Atom/Feature/Common/Assets/ShaderLib/Atom/Features/BlendUtility.azsli
```diff
@@ -45,10 +45,13 @@ float3 ApplyTextureBlend(float3 color, float3 blendColor, float factor, TextureB
     {
         // [GFX TODO][ATOM-3237] See if we can avoid these extra color space conversions...
         // To get blend behavior that mimics photoshop, convert to sRGB
-        float3 blendColorSRGB = TransformColor(blendColor, ColorSpaceId::ACEScg, ColorSpaceId::SRGB);
-        float3 colorSRGB = TransformColor(color, ColorSpaceId::ACEScg, ColorSpaceId::SRGB);
+        //Acessg -> srgb
+        real3 blendColorSRGB = AcesCg_To_LinearSrgb(blendColor);
```
moudgils, line 49: I made this change so we dont have to incur the cost of all the branches within TransformColor. Iacessgd a comment and fix up the typo related to acessg.

Gems/Atom/Feature/Common/Assets/ShaderLib/Atom/Features/BlendUtility.azsli
moudgils: I didnt want to change the existing behavior in this case. This could be a separate change in future.

Gems/Atom/Feature/Common/Assets/ShaderLib/Atom/Features/PBR/LightingUtils.azsli
moudgils: Yes I could. This was part of my attempt to satisfy precision loss warnings coming from other android shaders that were not converted over to use real keyword. In the end I gave up on that effort and basically the plan is to just live with those warnings until we convert over all the other relevant shaders.

Gems/Atom/Asset/Shader/Code/Source/Editor/AzslShaderBuilderSystemComponent.cpp
```diff
@@ -83,7 +83,7 @@ namespace AZ
             // Register Shader Asset Builder
             AssetBuilderSDK::AssetBuilderDesc shaderAssetBuilderDescriptor;
             shaderAssetBuilderDescriptor.m_name = "Shader Asset Builder";
-            shaderAssetBuilderDescriptor.m_version = 118; // material pipeline MATERIAL_TYPE_AZSLI_FILE_PATH trick
+            shaderAssetBuilderDescriptor.m_version = 119; // Half float support
```
AMZN-alexpete, line 86: Can you bump the minor version in `Gems/Atom/gem.json` to reflect this change - or the major version if this is a breaking API change.

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/MultiView/ForwardPass_BaseLighting.azsli
```diff
@@ -14,17 +14,24 @@
...
-#define ENABLE_SHADER_DEBUGGING 1
+#define ENABLE_SHADER_DEBUGGING 0
 #define ENABLE_CLEAR_COAT 0
 #define ENABLE_SHADOWS 0
 #define ENABLE_LIGHT_CULLING 0
 #define ENABLE_AREA_LIGHTS 0
 #define ENABLE_PHYSICAL_SKY 0
 #define ENABLE_MERGE_FILMIC_TONEMAP 1
+#define ENABLE_DECALS              0
+#define ENABLE_SHADER_DEBUGGING           0
```
AMZN-alexpete, line 30: this is in here twice (line 22)
```suggestion
```

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/MultiView/ForwardPass_StandardLighting.azsli
```diff
@@ -11,21 +11,27 @@
...
+#define ENABLE_SHADER_DEBUGGING 0
 #define ENABLE_CLEAR_COAT 0
 #define ENABLE_SHADOWS 0
 #define ENABLE_LIGHT_CULLING 0
 #define ENABLE_AREA_LIGHTS 0
+#define ENABLE_AREA_LIGHT_VALIDATION    0
 #define ENABLE_PHYSICAL_SKY 0
 #define ENABLE_MERGE_FILMIC_TONEMAP 1
+#define ENABLE_DECALS 0
+#define ENABLE_SHADER_DEBUGGING 0
```
AMZN-alexpete, line 31: another duplicate
```suggestion

```

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/MultiView/ForwardPass_StandardLighting_CustomZ.azsli
```diff
@@ -19,13 +19,18 @@
...
-#define ENABLE_CLEAR_COAT 1
+#define ENABLE_SHADER_DEBUGGING 0
+#define ENABLE_CLEAR_COAT 0
 #define ENABLE_SHADOWS 0
 #define ENABLE_LIGHT_CULLING 0
 #define ENABLE_AREA_LIGHTS 0
 #define ENABLE_PHYSICAL_SKY 0
 #define ENABLE_MERGE_FILMIC_TONEMAP 1
+#define ENABLE_DECALS 0
+#define ENABLE_SHADER_DEBUGGING 0
```
AMZN-alexpete, line 30:
```suggestion
```

Gems/Atom/Feature/Common/Assets/Materials/Pipelines/MultiView/Transparent_StandardLighting.azsli
```diff
@@ -18,13 +18,18 @@
...
-#define ENABLE_CLEAR_COAT 1
+#define ENABLE_SHADER_DEBUGGING 0
+#define ENABLE_CLEAR_COAT 0
 #define ENABLE_SHADOWS 0
 #define ENABLE_LIGHT_CULLING 0
 #define ENABLE_AREA_LIGHTS 0
 #define ENABLE_PHYSICAL_SKY 0
 #define ENABLE_MERGE_FILMIC_TONEMAP 1
+#define ENABLE_DECALS              0
+#define ENABLE_SHADER_DEBUGGING           0
```
AMZN-alexpete, line 29:
```suggestion
```

Gems/Atom/Feature/Common/Assets/ShaderLib/Atom/Features/ColorManagement/TransformColor.azsli
```diff
@@ -30,13 +30,27 @@ enum class ColorSpaceId
...
+}
+
+real3 TransformColor(in real3 color, ColorSpaceId fromColorSpace, ColorSpaceId toColorSpace)
 {
     if(fromColorSpace == toColorSpace)
     {
         return color;
     }
-    
+
```
AMZN-alexpete, line 53: nit some whitespace got added
```suggestion

```

Gems/Atom/Feature/Common/Assets/ShaderLib/Atom/Features/PBR/Lighting/StandardLighting.azsli
```diff
@@ -21,26 +21,26 @@
...
     {
         // Attenuate diffuse term by clear coat's fresnel term to account for energy loss
-        float HdotV = saturate(dot(normalize(dirToLight + lightingData.dirToCamera), lightingData.dirToCamera));
+        real HdotV = saturate(dot(normalize(dirToLight + lightingData.dirToCamera), lightingData.dirToCamera));
         diffuse *= 1.0 - (FresnelSchlick(HdotV, 0.04) * surface.clearCoat.factor);
     }
-#endif
-
+ #endif
+
```
AMZN-alexpete, line 36: nit: whitespace got added
```suggestion
#endif

```

moudgils:
> In general looks good, but big question is - what about backwards compatibility? - is this going to break existing shaders or shaders that pull in any Atom `.azsli` files? If so then we definitely want to bump the major version of the affected Atom `gem.json` and in the `engine.json`

The shaders are backwards compatible. I have bumped the minor version of Atom gem to be 0.1.0

Gems/Atom/RPI/Code/Source/RPI.Public/RPISystem.cpp
```diff
@@ -83,12 +84,17 @@ namespace AZ
             if (m_xrSystem)
             {
                 AZ::RHI::ResultCode resultCode = m_xrSystem->InitInstance();
-                // Fail result code can happen if no compatible device is attached. UnRegister xr system if that happens
-                if (resultCode == AZ::RHI::ResultCode::Fail)
+                // Check if the result code is a Success. Failure can happen if no compatible device is attached.
+                // Check if the active rhi backend is not vulkan
+                // UnRegister xr system if any of the criteria is not met
+                if (resultCode == AZ::RHI::ResultCode::Fail || !AZ::RHI::IsVulkanRHI())
```
invertednormal, line 90: if resultCode is success, but the RHI is not vulkan, won't this silently fail?

invertednormal (approved):
Looks good - will be interesting to try halfs out on higher end platforms and see if we can get a performance bump.

I think long term most variables will be ok to either be half everywhere or need to be float everywhere, only a few will need to mutate depending on platform, which is the support that 'real' provides, but this is a good start to unblock mobile and not affect the shaders on desktop.

Two main issues I expect you'll have with halfs:
- Some of the numbers involved with GGX calculations for very smooth surfaces get quite tiny, like `roughnessA2`, which may require floats to work correctly.
- Any kind of world coordinate math will likely need full float - we run into problems with that even on normal floats once you get a few km from origin. It's likely that halves will start to break at 10s or 100s of meters from origin.

Gems/Atom/Asset/Shader/Code/AZSL/Platform/Linux/Vulkan/AzslcHeader.azsli
```diff
@@ -14,3 +14,11 @@
...
 static const float4 s_AzslDebugColor = float4(165.0 / 255.0, 30.0 / 255.0, 36.0 / 255.0, 1);
 
+// Full floats for Linux
+#define real float
+#define real2 float2
+#define real3 float3
+#define real4 float4
+#define real3x3 float3x3
+#define real3x4 float3x4
+#define real4x4 float4x4
```
siliconvoodoo, line 24: why not `typedef`s (or `typealias`. aszsl form that works the same as `using`)

PR #13746, opened 2022-12-16
Asset Processor - Do not mark a file as failed in batch if it succeeds later

amzn-mike:
## What does this PR do?

Quick fix to avoid asset processor batch reporting a failure which is later resolved.

This can occur with intermediate assets which have already been processed in a previous run.  When a dependency is updated that causes both the source and produced intermediate to reprocess, the source can output an updated intermediate before the intermediate job has read the existing file, causing the job to attempt to load a newly added dependency which has not finished processing yet.  The job will be count as a failure in this case and cause APB to return an error code, even though the job later re-runs and succeeds.

## How was this PR tested?

None, this is a simple speculative fix for a race condition observed on Jenkins.


Code/Tools/AssetProcessor/native/utilities/BatchApplicationManager.cpp
```diff
@@ -123,6 +123,9 @@ void BatchApplicationManager::MakeActivationConnections()
         {
             m_processedAssetCount++;
 
+            // If a file fails and later succeeds, don't count it as a failure
```
santorac, line 126: Maybe expand your comment to explicitly note that this is to avoid failing jenkins jobs when intermediate assets cause transient failures.

lmbr-pip (approved):

AMZN-stankowi (approved):

lmbr-pip:
Do we need to cover with any testing?

amzn-mike:
> Do we need to cover with any testing?

@lmbr-pip I looked at this for a while but I don't think there's a good, stable way of testing this unless I either make sleep-loops (which we know are not reliable) or introduce a signal just for the purpose of testing.  It's a pretty minor change and if it doesn't work, Jenkins will pretty much let us know with a job failure.

PR #13617, opened 2022-12-08
Shorten AWSI python test project name

allisaurus:
## What does this PR do?

Related to https://github.com/o3de/o3de/issues/12487

Reduces generated resource name to be <140 chars (138, to be exact) to fix validation errors like seen in: 
https://jenkins-pipeline.agscollab.com/blue/organizations/jenkins/O3DE-LY-Fork-development_periodic-incremental-daily-internal/detail/O3DE-LY-Fork-development_periodic-incremental-daily-internal/575/pipeline/1915

PR #13747, opened 2022-12-16
Shorten AWSI python test project name formulation

allisaurus:
## What does this PR do?

Attempts to fix periodic build errors in the AWS python tests. Subsumes https://github.com/o3de/o3de/pull/13617

## How was this PR tested?

Deploys and tests pass for both Windows & Linux in forked [Jenkins job](https://jenkins-pipeline.agscollab.com/job/O3DE-Fork-dev_periodic-incremental-daily-internal/7/).

### DIFF
--- a/AutomatedTesting/Gem/PythonTests/AWS/common/constants.py
+++ b/AutomatedTesting/Gem/PythonTests/AWS/common/constants.py
@@ -12,7 +12,7 @@ import platform
 ASSUME_ROLE_ARN = os.environ.get('ASSUME_ROLE_ARN', 'arn:aws:iam::645075835648:role/o3de-automation-tests')
 # Name of the AWS project deployed by the CDK applications
 AWS_PROJECT_NAME = os.environ.get('O3DE_AWS_PROJECT_NAME').upper() if os.environ.get('O3DE_AWS_PROJECT_NAME') else \
-    (os.environ.get('BRANCH_NAME', '') + '-' + os.environ.get('PIPELINE_NAME', '') + '-' + platform.system()).upper()
+    (os.environ.get('BRANCH_NAME', '') + '-' + os.environ.get('PIPELINE_NAME', '')[:25] + '-' + platform.system()).upper()
 # Region for the existing CloudFormation stacks used by the automation tests
 AWS_REGION = os.environ.get('O3DE_AWS_DEPLOY_REGION', 'us-east-1')
 # Name of the default resource mapping config file used by the automation tests
--- a/AutomatedTesting/Gem/PythonTests/AWS/common/constants.py
+++ b/AutomatedTesting/Gem/PythonTests/AWS/common/constants.py
@@ -8,11 +8,15 @@ SPDX-License-Identifier: Apache-2.0 OR MIT
 import os
 import platform
 
+# To avoid resource name length issues, potentially verbose pipeline names are capped at 25 chars.
+# TODO: consolidate project name formulation for tests and deploy scripts to same place.
+MAX_PIPELINE_NAME_LENGTH = 25
+
 # ARN of the IAM role to assume for retrieving temporary AWS credentials
 ASSUME_ROLE_ARN = os.environ.get('ASSUME_ROLE_ARN', 'arn:aws:iam::645075835648:role/o3de-automation-tests')
 # Name of the AWS project deployed by the CDK applications
 AWS_PROJECT_NAME = os.environ.get('O3DE_AWS_PROJECT_NAME').upper() if os.environ.get('O3DE_AWS_PROJECT_NAME') else \
-    (os.environ.get('BRANCH_NAME', '') + '-' + os.environ.get('PIPELINE_NAME', '')[:25] + '-' + platform.system()).upper()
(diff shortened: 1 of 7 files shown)


scripts/build/Platform/Linux/build_config.json
```diff
@@ -393,7 +393,7 @@
     "COMMAND": "deploy_cdk_applications.sh",
     "PARAMETERS": {
       "NVM_VERSION": "v0.39.1",
-      "CDK_VERSION": "1.179.0",
+      "CDK_VERSION": "1.183.0",
```
lmbr-pip, line 396: IMHO we should let these float and allow the python applications to decide (https://peps.python.org/pep-0440/#compatible-release). 

We should be usng pyenv to reduce impact between tests.

Tying to a fixed version is just brittle.

allisaurus: Totally agree we should stop pinning to a version, however I wanted to get this project name formulation change out in isolation first, and we know it will fail on the current version so it is bumped here. Would prefer to change the logic that sets and uses the `CDK_VERSION` in a separate PR so it's easier to isolate what's wrong if problems arise.

AutomatedTesting/Gem/PythonTests/AWS/common/constants.py
```diff
@@ -12,7 +12,7 @@
 ASSUME_ROLE_ARN = os.environ.get('ASSUME_ROLE_ARN', 'arn:aws:iam::645075835648:role/o3de-automation-tests')
 # Name of the AWS project deployed by the CDK applications
 AWS_PROJECT_NAME = os.environ.get('O3DE_AWS_PROJECT_NAME').upper() if os.environ.get('O3DE_AWS_PROJECT_NAME') else \
-    (os.environ.get('BRANCH_NAME', '') + '-' + os.environ.get('PIPELINE_NAME', '') + '-' + platform.system()).upper()
+    (os.environ.get('BRANCH_NAME', '') + '-' + os.environ.get('PIPELINE_NAME', '')[:25] + '-' + platform.system()).upper()
```
lmbr-pip, line 15: nit: 25 is magic number, consider at least referencing in comments why 25

lmbr-pip (approved):
Only concern here is that we are proliferating the 25 truncation to shell and other commands.

In the short term, link/provide context with comments as its too easy to forget to update a .sh or .cmd file here.

In the long term, truncation should be handled in just one place and thats the source of truth.

scripts/build/Platform/Linux/build_config.json
amzn-changml: Note, the versions denoted here are tied to the NPM CDK toolkit, not the Python package. 

It's fine to let the version float, though it also means NodeJS has to be maintained as well, as the latest version of the toolkit does not guarantee backwards compatibility with older versions of NodeJS. NVM acts like pyenv in this case for version pinning

AutomatedTesting/Gem/PythonTests/AWS/common/constants.py
amzn-changml: Would there be a issue here if the branch name also exceeds the character limit?

AutomatedTesting/Gem/PythonTests/AWS/common/constants.py
```diff
@@ -8,11 +8,15 @@
...
+# To avoid resource name length issues, potentially verbose pipeline names are capped at 25 chars.
+# TODO: consolidate project name formulation for tests and deploy scripts to same place.
+MAX_PIPELINE_NAME_LENGTH = 25
+
 # ARN of the IAM role to assume for retrieving temporary AWS credentials
 ASSUME_ROLE_ARN = os.environ.get('ASSUME_ROLE_ARN', 'arn:aws:iam::645075835648:role/o3de-automation-tests')
 # Name of the AWS project deployed by the CDK applications
 AWS_PROJECT_NAME = os.environ.get('O3DE_AWS_PROJECT_NAME').upper() if os.environ.get('O3DE_AWS_PROJECT_NAME') else \
-    (os.environ.get('BRANCH_NAME', '') + '-' + os.environ.get('PIPELINE_NAME', '') + '-' + platform.system()).upper()
+    (os.environ.get('BRANCH_NAME', '') + '-' + os.environ.get('PIPELINE_NAME', '')[:MAX_PIPELINE_NAME_LENGTH] + '-' + platform.system()).upper()
```
amzn-changml, line 19: Would there also be an issue if the branch name also reaches the char limit?

amzn-changml (approved):
LGTM, with some comments

junbo75 (approved):

shirangj (approved):

AutomatedTesting/Gem/PythonTests/AWS/common/constants.py
allisaurus: Yes, the ceiling is a 40 char total length which would include everything (branch, project, platform). Capping what is currently the most verbose parameter (Pipeline) is just the most impactful single preventative measure we could add.

As @lmbr-pip says above, long-term there's more work to be done here, including consolidating how we're formulating the project names to one place. At that point we should probably enforce a formula that caps each argument so some portion of the allowed length.

scripts/build/Platform/Linux/build_config.json
allisaurus: Thanks for pointing this out @amzn-changml , we'll need to keep that in mind during subsequent changes to how to how we version these.

PR #13751, opened 2022-12-16
Change the screenshot python tests to use outcome as return value.

jiaweig-amzn:
## What does this PR do?

Screenshot APIs (including taking screenshot/image comparison/path building) now return AZ::Outcome

## How was this PR tested?

Tested ASV for lua (it should not be affected, for stability)
Tested python hydra tests in Atom (they are skipped, force enabled for testing purpose.)

### DIFF
--- a/Gems/Atom/Feature/Common/Code/Source/FrameCaptureSystemComponent.h
+++ b/Gems/Atom/Feature/Common/Code/Source/FrameCaptureSystemComponent.h
@@ -41,12 +41,12 @@ namespace AZ
 
             // FrameCaptureRequestBus overrides ...
             bool CanCapture() const override;
-            AZ::Outcome<FrameCaptureId, FrameCaptureId> CaptureScreenshot(const AZStd::string& filePath) override;
-            AZ::Outcome<FrameCaptureId, FrameCaptureId> CaptureScreenshotForWindow(const AZStd::string& filePath, AzFramework::NativeWindowHandle windowHandle) override;
-            AZ::Outcome<FrameCaptureId, FrameCaptureId> CaptureScreenshotWithPreview(const AZStd::string& outputFilePath) override;
-            AZ::Outcome<FrameCaptureId, FrameCaptureId> CapturePassAttachment(const AZStd::vector<AZStd::string>& passHierarchy, const AZStd::string& slotName, const AZStd::string& outputFilePath,
+            AZ::Outcome<FrameCaptureId, FrameCaptureError> CaptureScreenshot(const AZStd::string& filePath) override;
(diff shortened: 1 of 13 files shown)


Gems/Atom/Feature/Common/Code/Source/FrameCaptureSystemComponent.h
```diff
@@ -41,23 +41,23 @@ namespace AZ
...
-            FrameCaptureId CapturePassAttachment(const AZStd::vector<AZStd::string>& passHierarchy, const AZStd::string& slotName, const AZStd::string& outputFilePath,
+            AZ::Outcome<FrameCaptureId, FrameCaptureId> CaptureScreenshot(const AZStd::string& filePath) override;
+            AZ::Outcome<FrameCaptureId, FrameCaptureId> CaptureScreenshotForWindow(const AZStd::string& filePath, AzFramework::NativeWindowHandle windowHandle) override;
+            AZ::Outcome<FrameCaptureId, FrameCaptureId> CaptureScreenshotWithPreview(const AZStd::string& outputFilePath) override;
+            AZ::Outcome<FrameCaptureId, FrameCaptureId> CapturePassAttachment(const AZStd::vector<AZStd::string>& passHierarchy, const AZStd::string& slotName, const AZStd::string& outputFilePath,
                 RPI::PassAttachmentReadbackOption option) override;
-            FrameCaptureId CapturePassAttachmentWithCallback(const AZStd::vector<AZStd::string>& passHierarchy, const AZStd::string& slotName
+            AZ::Outcome<FrameCaptureId, FrameCaptureId> CapturePassAttachmentWithCallback(const AZStd::vector<AZStd::string>& passHierarchy, const AZStd::string& slotName
```
rgba16f, line 49: This is a breaking change for existing external users of the api, it probably needs to go through a deprecation cycle before it can be changed like this.

gadams3 (commented):
Why change this to an outcome at all? Isn't the -1 sufficient as an invalid value? Using an outcome might be more useful if the second parameter was a string or reason why the operation failed. Since it's just another frame capture ID, seems unnecessary IMHO.

jiaweig-amzn:
> Why change this to an outcome at all? Isn't the -1 sufficient as an invalid value? Using an outcome might be more useful if the second parameter was a string or reason why the operation failed. Since it's just another frame capture ID, seems unnecessary IMHO.

I'll expand and work on a error class for both capture and comparison. The major things we are addressing are:
a consistent usage for behavior context used for testing
richer debug info

In fact, we already have
        AZ_ENUM_CLASS(FrameCaptureResult,
            None,
            Success,
            FileWriteError,
            InvalidArgument,
            UnsupportedFormat,
            InternalError
        )
But it is only used inside C++ code.

gadams3 (approved):

jiaweig-amzn:
Due to some other code refactoring, it will be a mess merging the code on top of this PR. Restart a new one:
https://github.com/o3de/o3de/pull/14444

PR #13752, opened 2022-12-16
Add a mutex to MaterialAssignmentBus…

amzn-tommy:
…since MeshFeatureProcessor connects to it from multiple threads

Signed-off-by: Tommy Walton <waltont@amazon.com>

## What does this PR do?

Allows the bus to be connected to from multiple threads without asserting.

## How was this PR tested?

Tested using the ROSConDemo, and the assert no longer occurs


gadams3 (approved):
I understand why the mutex is necessary. However, this is a very high traffic bus that should be removed. Instead, the changes that it's monitoring should be accumulated and sent as a single notification in the material component tick. Then the mesh component can listen and tell the mesh feature processor to update the ray tracing data. This is on my list have things to do.

cgalvan (approved):

PR #13755, opened 2022-12-17
Make it so ENABLE_COMPILED_TASK_GRAPH_EVENT_TRACKING is only defined …

amzn-tommy:
…in debug builds

Signed-off-by: Tommy Walton <waltont@amazon.com>

## What does this PR do?

As it stands, ENABLE_COMPILED_TASK_GRAPH_EVENT_TRACKING is always defined, so it is always enabled despite surrounding code with 
#ifdef ENABLE_COMPILED_TASK_GRAPH_EVENT_TRACKING

## How was this PR tested?

Put a break point in some of the code that was surrounded by #ifdef ENABLE_COMPILED_TASK_GRAPH_EVENT_TRACKING. It was hit before in a profile build, now that breakpoint isn't hit any more.


rgba16f (approved):

rgba16f:
This is probably why Anton found TaskGraph was slower than the old job system

amzn-tommy:
> This is probably why Anton found TaskGraph was slower than the old job system

I didn't notice an improvement with this fix, but I wasn't looking broadly at the whole frame. I was looking specifically at a bubble I encountered, thinking this was the cause of the bubble, but it turned out this didn't fix the issue I saw.

Anton never confirmed why the job system was faster for him, but I suspect it was mainly load balancing. 

At the time he was testing, the culling workload for each task was highly varied. So you could end up with a worst case scenario of the round-robin scheduling putting two long-running tasks on ThreadA, then every other thread finishes their short tasks and ends up waiting for the thread with the long tasks. In the case of the job system, with its work stealing behavior, if another thread finishes its short workload while ThreadA is still on its first job/task, the now available thread would take the second job/task and start processing it.

That being said, Anton did the work to make those tasks more balanced, so that shouldn't happen any more (for culling). I suggested he re-test with the TaskGraph after that was done, but don't recall if he got around to that or not.

lumberyard-employee-dm (approved):

Issue #11401, opened 2022-08-18
Memory keeps on spiking upon modifying many entities in the level


srikappa-amzn:
**Describe the bug**
In a large level with many entities, modifying many entities seems to keep on spiking memory without ever coming down.

**Assets required**
[!Important!] Please provide sample assets needed to reproduce the issue, either as an [attachment](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/attaching-files) or as a link to a public asset in Github.

**Steps to reproduce**
1. Download the attached zip below and put it under the levels folder in your project
2. Launch the editor using visual studio and open the level
3. Select any one of the entities named '1k entity' and click the lock icon
4. Notice the spike in memory usage in Visual studio
5. Keep clicking the lock icon and notice the increase in memory usage for each click.

**Expected behavior**
Memory doesn't spike

**Actual behavior**
It spikes for each click

**Found in Branch**
development

**Desktop/Device (please complete the following information):**
Device: PC
OS: Windows
Version 10
CPU Intel(R) Xeon(R) W-2155 CPU @ 3.30GHz
GPU NVidia GeForce RTX Super 2700
Memory 64GB

**Additional context**
[testEntityLockOnLargeLevels.zip](https://github.com/o3de/o3de/files/9378156/testEntityLockOnLargeLevels.zip)

LB-KatarzynaDylska:
@srikappa-amzn 
It has been observed on Development (23cc830) build that if all the entities are in prefabs (including the nested single entities), locking them results in a crash after about 4 minutes of unsuccessful loading. This was the case on a machine with 16GB of RAM. Should this crash be logged as a separate issue?

Video of the level with entities not in prefabs:

https://user-images.githubusercontent.com/99388484/185929782-a53930f1-cec1-4f57-821d-a5d2cef54564.mp4

Video of the level with entities in prefabs:

https://user-images.githubusercontent.com/99388484/185930269-1c64b3fe-2f7b-4c21-a940-96cdc2279f5f.mp4

monroegm:
@lawsonamzn Is this going to be fixed by your performance improvements?

lawsonamzn:
This may have already been remedied in development - I'll check.

lawsonamzn:
This still occurs.  It isn't a deadlock but its a o(N^y) operation where N is the number of entities and y is the number of levels in the dom :(

lawsonamzn:
[...]

PR #13756, opened 2022-12-17
Prefab Overrides | New edit-entity undo node for improving performance

forkercat:

## What does this PR do?

This PR introduces a new edit-entity undo node for override editing. This node is a more selective version of avoid comparing instance DOMs to generate patches. As a result, it would increase the editor performance for override editing as mentioned in https://github.com/o3de/o3de/issues/11401.

This PR depends on https://github.com/o3de/o3de/pull/13640 to add override subtree successfully.

| Performance Improvement | Base | This PR | Change % |
| :---    |  ---: | ---: | ---: |
| Edit entity in 10k-entity prefab (Single-Edit) |  4,000 ms |  2,700 ms | -32.5% |
| Toggle lock on 1k-entity prefab (Multi-Edit) |  56,000 ms |  8,000 ms | -85.7% |

#### Major Changes
1. New undo node for edit-entity override that supports batch processing, merging override patches to add-entity patch. It can even be used for add-entity override workflow in the future.
2. Add patch index counter and functionality to generate subtree in Link class.
3. Update edit-entity workflow to partially adopt this undo node

#### What's Next

1. Further improvement in edit-entity workflow:
    1. Batch processing dirty entities in edit-entity workflow (improvement is noted in the GitHub issue)
    1. Can use this undo node for reparent editing as well
1. Can use this undo node for add-entity workflow (improvement is not obvious though)

## How was this PR tested?

- [x] Manually testing in editor.
- [x] Passed all tests.
- [x] Ran performance check. See improvement.
- [ ] Add some new unit tests for the new undo node

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Link/Link.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Link/Link.cpp
@@ -37,6 +37,7 @@ namespace AzToolsFramework
             , m_instanceName(AZStd::move(other.m_instanceName))
             , m_prefabSystemComponentInterface(AZStd::move(other.m_prefabSystemComponentInterface))
             , m_linkPatchesTree(AZStd::move(other.m_linkPatchesTree))
+            , m_patchIndexCounter(AZStd::move(other.m_patchIndexCounter))
         {
             other.m_prefabSystemComponentInterface = nullptr;
         }
@@ -56,6 +57,7 @@ namespace AzToolsFramework
                     "It is a requirement for the Link class. Check that it is being correctly initialized.");
                 other.m_prefabSystemComponentInterface = nullptr;
                 m_linkPatchesTree = AZStd::move(other.m_linkPatchesTree);
+                m_patchIndexCounter = AZStd::move(other.m_patchIndexCounter);
             }
             return *this;
         }
@@ -283,6 +285,35 @@ namespace AzToolsFramework
             }
         }
 
+        PrefabOverridePrefixTree Link::GenerateOverrideSubTreeForEntity(
+            const PrefabDomValue& subpathPatches, const AZStd::string& pathToEntity)
+        {
+            PrefabOverridePrefixTree prefixTree;
+            if (subpathPatches.IsArray())
+            {
+                rapidjson::GenericArray patchesArray = subpathPatches.GetArray();
+                for (rapidjson::SizeType i = 0; i < patchesArray.Size(); i++)
+                {
+                    PrefabDom patchEntry;
+                    patchEntry.CopyFrom(patchesArray[i], patchEntry.GetAllocator());
+
+                    auto pathIter = patchEntry.FindMember("path");
+                    if (pathIter != patchEntry.MemberEnd())
+                    {
+                        AZStd::string subPath = pathIter->value.GetString(); // eg: '/Components/Component_[123]/Child Sort Order/0'
+                        AZStd::string fullPath = pathToEntity + subPath;
+                        pathIter->value.SetString(fullPath.c_str(), static_cast<rapidjson::SizeType>(fullPath.length()),
+                            patchEntry.GetAllocator());
+
+                        PrefabOverrideMetadata overrideMetadata(AZStd::move(patchEntry), m_patchIndexCounter++);
+                        // Subpath is used to build the subtree within an entity scope.
+                        prefixTree.SetValue(AZ::Dom::Path(subPath), AZStd::move(overrideMetadata));
+                    }
+                }
+            }
+            return prefixTree;
+        }
+
         void Link::RebuildLinkPatchesTree(const PrefabDomValue& patches)
         {
             m_linkPatchesTree.Clear();
@@ -298,7 +329,7 @@ namespace AzToolsFramework
                     if (path != patchEntry.MemberEnd())
                     {
                         AZ::Dom::Path domPath(path->value.GetString());
-                        PrefabOverrideMetadata overrideMetadata(AZStd::move(patchEntry), i);
+                        PrefabOverrideMetadata overrideMetadata(AZStd::move(patchEntry), m_patchIndexCounter++);
                         m_linkPatchesTree.SetValue(domPath, AZStd::move(overrideMetadata));
                     }
                 }
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Link/Link.h
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Link/Link.h
@@ -153,6 +153,14 @@ namespace AzToolsFramework
              */
             void AddLinkIdToInstanceDom(PrefabDomValue& instanceDomValue);
 
+            //! Generates a DOM prefix subtree that contains the provided override patches for an entity.
+            //! Paths in patches should be subpath from an entity's perspective, eg: /Components/Component_[123].
+            //! @param subpathPatches The override patches for an entity that include subpaths.
+            //! @param pathToEntity The override path to an entity that will be used to construct patch paths.
+            //! @return DOM prefix tree that contains override patches.
+            AZ::Dom::DomPrefixTree<PrefabOverrideMetadata> GenerateOverrideSubTreeForEntity(
+                const PrefabDomValue& subpathPatches, const AZStd::string& pathToEntity);
+
         private:
 
             /**
@@ -185,7 +193,11 @@ namespace AzToolsFramework
             // Name of the nested instance of target Template.
             AZStd::string m_instanceName;
 
+            // Identifier for the link.
             LinkId m_id = InvalidLinkId;
+
+            // Index counter for generating patches.
+            AZ::u32 m_patchIndexCounter = 0u;
 
             PrefabSystemComponentInterface* m_prefabSystemComponentInterface = nullptr;
         };
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
@@ -894,6 +894,7 @@ namespace AzToolsFramework
 
                 if (isInFocusTree && !isOwnedByFocusedPrefabInstance)
                 {
+                    // Override Editing
                     if (isNewParentOwnedByDifferentInstance)
                     {
                         Internal_HandleInstanceChange(parentUndoBatch, entity, beforeParentId, afterParentId);
@@ -913,15 +914,20 @@ namespace AzToolsFramework
                     }
                     else
                     {
-                        PrefabDom newPatch;
-                        m_instanceToTemplateInterface->GeneratePatch(newPatch, beforeState, afterState);
-                        LinkId linkId = m_prefabFocusInterface->AppendPathFromFocusedInstanceToPatchPaths(newPatch, entityId);
+                        AzFramework::EntityContextId editorEntityContextId = AzFramework::EntityContextId::CreateNull();
+                        EditorEntityContextRequestBus::BroadcastResult(
+                            editorEntityContextId, &EditorEntityContextRequests::GetEditorEntityContextId);
+                        InstanceOptionalReference focusedInstance = m_prefabFocusInterface->GetFocusedPrefabInstance(
+                            editorEntityContextId);
 
-                        Internal_HandleContainerOverride(parentUndoBatch, entityId, newPatch, linkId);
+                        AZ::Entity* entityToUpdate = GetEntityById(entityId);
+                        PrefabUndoHelpers::UpdateEntityListAsOverride(
+                            { entityToUpdate }, owningInstance->get(), focusedInstance->get(), parentUndoBatch);
                     }
                 }
                 else
                 {
+                    // Source Template Editing
                     Internal_HandleEntityChange(parentUndoBatch, entityId, beforeState, afterState);
 
                     if (isNewParentOwnedByDifferentInstance)
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabUndoHelpers.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabUndoHelpers.cpp
@@ -13,6 +13,7 @@
 #include <AzToolsFramework/Prefab/Undo/PrefabUndoAddEntityAsOverride.h>
 #include <AzToolsFramework/Prefab/Undo/PrefabUndoDelete.h>
 #include <AzToolsFramework/Prefab/Undo/PrefabUndoDeleteAsOverride.h>
+#include <AzToolsFramework/Prefab/Undo/PrefabUndoEntityOverrides.h>
 
 namespace AzToolsFramework
 {
@@ -101,6 +102,18 @@ namespace AzToolsFramework
                 PrefabUndoEntityUpdate* state = aznew PrefabUndoEntityUpdate("Undo Updating Entity");
                 state->SetParent(undoBatch);
                 state->Capture(entityDomBeforeUpdatingEntity, entityDomAfterUpdatingEntity, entityId);
+                state->Redo();
+            }
+
+            void UpdateEntityListAsOverride(
+                const AZStd::vector<const AZ::Entity*>& entityList,
+                Instance& owningInstance,
+                const Instance& focusedInstance,
+                UndoSystem::URSequencePoint* undoBatch)
+            {
+                PrefabUndoEntityOverrides* state = aznew PrefabUndoEntityOverrides("Undo Updating Entity List As Override");
+                state->SetParent(undoBatch);
+                state->Capture(entityList, owningInstance, focusedInstance);
                 state->Redo();
             }
 
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabUndoHelpers.h
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabUndoHelpers.h
@@ -60,6 +60,17 @@ namespace AzToolsFramework
                 AZ::EntityId entityId,
                 UndoSystem::URSequencePoint* undoBatch);
 
+            //! Helper function for updating entities as overrides to focused template with undo-redo support.
+            //! @param entityList Entity list for entities to be updated in focused template.
+            //! @param owningInstance The common owning prefab instance of all inputs.
+            //! @param focusedInstance The current focused prefab instance.
+            //! @param undoBatch The undo batch node to register the update-entity undo node to.
+            void UpdateEntityListAsOverride(
+                const AZStd::vector<const AZ::Entity*>& entityList,
+                Instance& owningInstance,
+                const Instance& focusedInstance,
+                UndoSystem::URSequencePoint* undoBatch);
+
             //! Helper function for deleting entities and update parents to prefab template with undo-redo support.
             //! Note: It includes updating relevant parent entities.
             //! @param entityAliasPathList The alias path list for entities that will be removed.
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoEntityOverrides.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoEntityOverrides.cpp
@@ -1,0 +1,145 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#include <AzToolsFramework/Prefab/Instance/InstanceToTemplateInterface.h>
+#include <AzToolsFramework/Prefab/Link/Link.h>
+#include <AzToolsFramework/Prefab/Undo/PrefabUndoEntityOverrides.h>
+#include <AzToolsFramework/Prefab/Undo/PrefabUndoUtils.h>
+#include <AzToolsFramework/Prefab/PrefabInstanceUtils.h>
+#include <AzToolsFramework/Prefab/PrefabSystemComponentInterface.h>
+
+namespace AzToolsFramework
+{
+    namespace Prefab
+    {
+        PrefabUndoEntityOverrides::PrefabUndoEntityOverrides(const AZStd::string& undoOperationName)
+            : UndoSystem::URSequencePoint(undoOperationName)
+            , m_linkId(InvalidLinkId)
+            , m_changed(true)
+        {
+            m_prefabSystemComponentInterface = AZ::Interface<PrefabSystemComponentInterface>::Get();
+            AZ_Assert(m_prefabSystemComponentInterface, "PrefabUndoEntityOverrides - Failed to grab prefab system component interface.");
+
+            m_instanceToTemplateInterface = AZ::Interface<InstanceToTemplateInterface>::Get();
+            AZ_Assert(m_instanceToTemplateInterface, "PrefabUndoEntityOverrides - Could not get instance to template interface.");
+        }
+
+        bool PrefabUndoEntityOverrides::Changed() const
+        {
+            return m_changed;
+        }
+
+        void PrefabUndoEntityOverrides::Capture(
+            const AZStd::vector<const AZ::Entity*>& entityList,
+            Instance& owningInstance,
+            const Instance& focusedInstance)
+        {
+            InstanceClimbUpResult climbUpResult = PrefabInstanceUtils::ClimbUpToTargetOrRootInstance(owningInstance, &focusedInstance);
+            AZ_Assert(climbUpResult.m_isTargetInstanceReached, "PrefabUndoEntityOverrides::Capture - "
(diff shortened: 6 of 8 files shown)


lawsonamzn (approved):
When I profiled this in a big level, this particular part was the thing making it take potentially hours to do a change.  This needs testing though - unit, or manual, etc.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -913,15 +914,20 @@ namespace AzToolsFramework
                     }
                     else
                     {
-                        PrefabDom newPatch;
-                        m_instanceToTemplateInterface->GeneratePatch(newPatch, beforeState, afterState);
-                        LinkId linkId = m_prefabFocusInterface->AppendPathFromFocusedInstanceToPatchPaths(newPatch, entityId);
-
-                        Internal_HandleContainerOverride(parentUndoBatch, entityId, newPatch, linkId);
```
srikappa-amzn, line 920: Lol, so the name of the function before was 'handleContainerOverride' for all type of overrides. Looks like we never changed it when override capability is expanded.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabUndoHelpers.h
```diff
@@ -60,6 +60,17 @@ namespace AzToolsFramework
                 AZ::EntityId entityId,
                 UndoSystem::URSequencePoint* undoBatch);
 
+            //! Helper function for updating entities as overrides to focused template with undo-redo support.
+            //! @param entityList Entity list for entities to be updated in focused template.
+            //! @param owningInstance The common owning prefab instance of all inputs.
+            //! @param focusedInstance The current focused prefab instance.
+            //! @param undoBatch The undo batch node to register the update-entity undo node to.
+            void UpdateEntityListAsOverride(
```
srikappa-amzn, line 68: nit: override's' as there will most likely be one per each entity. Also 'UpdateEntitiesAsOverrides' might be a clearer name.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
forkercat: Yeah. I'd like to refactor/rework this function in a follow-up PR. The code/workflow needs to be improved to clear confusion and there should be more documentation.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoEntityOverrides.cpp
```diff
@@ -0,0 +1,145 @@
...
+            PrefabDomReference cachedOwningInstanceDom = owningInstance.GetCachedInstanceDom();
+
+            const AZStd::string overridePatchPathToOwningInstance = PrefabInstanceUtils::GetRelativePathFromClimbedInstances(
+                climbUpResult.m_climbedInstances, true); // skip top instance
+
+            for (const AZ::Entity* entity : entityList)
+            {
+                if (!entity)
+                {
+                    AZ_Error("Prefab", false, "PrefabUndoEntityOverrides::Capture - Found a null entity. Skipping it.");
```
srikappa-amzn, line 66: warning might be more appropriate here since this won't be a user generated error and the user can't do anything to fix it

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoEntityOverrides.cpp
```diff
@@ -0,0 +1,145 @@
...
+                    continue;
+                }
+
+                // Hierarchy Example: Level (focused) -> Car (top) -> ... -> Wheel (owning) -> Tire (override)
+                // Path seen from focused: /Instances/Instance_[Car]/.../Instances/Instance_[Wheel]/Entities/Entity_[Tire]
+                //                                   (top instance)               (owning instance)         (entity to update)
+                // Note: Override patch path is the path seen from top without top instance alias path.
```
srikappa-amzn, line 73: I recommend reserving comments like these to the tests and keep the actual code clean and self explanatory with better names and smaller comments. Otherwise, I'm afraid this would become the norm across the prefabs code in multiple classes

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoEntityOverrides.cpp
```diff
@@ -0,0 +1,145 @@
...
+                    const PrefabDomValue* entityDomInTopTemplate = entityDomPathFromTopInstance.Get(topTemplateDom);
+
+                    if (entityDomInTopTemplate)
+                    {
+                        m_instanceToTemplateInterface->GeneratePatch(overridePatches, *entityDomInTopTemplate, entityDomAfterUpdate);
+                    }
+                    else
+                    {
+                        // Merge patches to an entity DOM as add-entity override.
+                        // Path is empty because there is no subpath for add-entity override patch like "/Components/Component_[123]".
```
srikappa-amzn, line 99: comment is not correct because it'll enter the else path if it can't find the entity in the dom and entity path stops at '/Instances/..../Entities/someAlias'

srikappa-amzn: Also, when does this even happen?(Because the add entity nodes are supposed to take care of additions right? )

Regardless, a unit test for this case would be great.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Link/Link.cpp
```diff
@@ -282,6 +284,35 @@ namespace AzToolsFramework
...
+
+                    auto pathIter = patchEntry.FindMember("path");
+                    if (pathIter != patchEntry.MemberEnd())
+                    {
+                        AZStd::string subPath = pathIter->value.GetString(); // eg: '/Components/Component_[123]/Child Sort Order/0'
+                        AZStd::string fullPath = pathToEntity + subPath;
+                        pathIter->value.SetString(fullPath.c_str(), static_cast<rapidjson::SizeType>(fullPath.length()),
+                            patchEntry.GetAllocator());
+
+                        PrefabOverrideMetadata overrideMetadata(AZStd::move(patchEntry), m_patchIndexCounter++);
```
srikappa-amzn, line 307: By incrementing the patch counter here in the hopes of making an additional call later to use the same indices feels error prone to me. For example, what happens when you generate the subtree but then have some issues adding it back to the tree under the link object? In such cases, you would have incremented the counter unnecessarily. 

Instead of doing this way, maybe we can remove the overrides at the path first, then generate and insert at the same time to the tree?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoEntityOverrides.cpp
```diff
@@ -0,0 +1,145 @@
...
+        void PrefabUndoEntityOverrides::UpdateLink()
+        {
+            LinkReference link = m_prefabSystemComponentInterface->FindLink(m_linkId);
+            if (link.has_value())
+            {
+                for (auto& [pathToSubTree, overrideSubTree] : m_overrideSubTrees)
+                {
+                    PrefabOverridePrefixTree prevSubTree = AZStd::move(link->get().RemoveOverrides(pathToSubTree));
+                    link->get().AddOverrides(pathToSubTree, AZStd::move(overrideSubTree));
+                    overrideSubTree = AZStd::move(prevSubTree);
```
srikappa-amzn, line 136: It took me a while to understand that you are using the same map to store 'before' trees when redoing and 'after' states when undoing. Maybe add a comment in the header file to explain the usage of the map

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoEntityOverrides.cpp
forkercat: Yeah. Good callout. I will keep the descriptive comments and remove others.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoEntityOverrides.cpp
```diff
@@ -0,0 +1,145 @@
...
+        void PrefabUndoEntityOverrides::Capture(
+            const AZStd::vector<const AZ::Entity*>& entityList,
+            Instance& owningInstance,
+            const Instance& focusedInstance)
+        {
+            InstanceClimbUpResult climbUpResult = PrefabInstanceUtils::ClimbUpToTargetOrRootInstance(owningInstance, &focusedInstance);
+            AZ_Assert(climbUpResult.m_isTargetInstanceReached, "PrefabUndoEntityOverrides::Capture - "
+                "Owning prefab instance should be a descendant of the focused prefab instance.");
+
+            m_linkId = climbUpResult.m_climbedInstances.back()->GetLinkId();
```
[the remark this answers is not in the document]
forkercat: Yes! I'll add the assertion.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoEntityOverrides.cpp
forkercat: Yeah. `m_overrideSubTrees` would be confusing. Better to rename it to `m_prevSubTrees` and add comments.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -913,15 +914,20 @@ namespace AzToolsFramework
...
+                        AzFramework::EntityContextId editorEntityContextId = AzFramework::EntityContextId::CreateNull();
+                        EditorEntityContextRequestBus::BroadcastResult(
+                            editorEntityContextId, &EditorEntityContextRequests::GetEditorEntityContextId);
+                        InstanceOptionalReference focusedInstance = m_prefabFocusInterface->GetFocusedPrefabInstance(
+                            editorEntityContextId);
+
+                        AZ::Entity* entityToUpdate = GetEntityById(entityId);
+                        PrefabUndoHelpers::UpdateEntityListAsOverride(
+                            { entityToUpdate }, owningInstance->get(), focusedInstance->get(), parentUndoBatch);
```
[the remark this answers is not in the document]
forkercat: I didn't quite follow. Which 'else' did you refer to? Do you mean line 900?

If yes: I thought about doing that as well and I remembered I tested it and it worked. But I removed the changes and prefer doing it in a next PR.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoEntityOverrides.cpp
forkercat: > Also, when does this even happen?(Because the add entity nodes are supposed to take care of additions right? )
> Regardless, a unit test for this case would be great.

So, this "else" branch is for covering the case that there is already an add-entity DOM as override, but the users continue editing the entity (like a component update) so these new override patches should be "merged" into that add-entity DOM.

Yeah a unit test for this is definitely needed.

forkercat: > comment is not correct because it'll enter the else path if it can't find the entity in the dom and entity path stops at '/Instances/..../Entities/someAlias'

I think you're referring to the entity path but my comment is referring to the empty path "" below. I need to update the comment for clearer description.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
michabr: Basically when we're override editing, and isNewParentOwnedByDifferentInstance is false, it doesn't seem like we need beforeState and afterState.

forkercat:
Update: Added unit tests for new undo node and non-override edit node

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Link/Link.cpp
```diff
@@ -282,6 +284,35 @@ namespace AzToolsFramework
...
+                    PrefabDom patchEntry;
+                    patchEntry.CopyFrom(patchesArray[i], patchEntry.GetAllocator());
+
+                    auto pathIter = patchEntry.FindMember("path");
+                    if (pathIter != patchEntry.MemberEnd())
+                    {
+                        AZStd::string subPath = pathIter->value.GetString(); // eg: '/Components/Component_[123]/Child Sort Order/0'
+                        AZStd::string fullPath = pathToEntity + subPath;
+                        pathIter->value.SetString(fullPath.c_str(), static_cast<rapidjson::SizeType>(fullPath.length()),
+                            patchEntry.GetAllocator());
```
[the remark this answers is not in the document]
forkercat: @srikappa-amzn I think the purpose I exposed the subtree API is I want to manage **subtree states** in undo node. The moment it generates the subtree it needs to finalize the JSON patch stored in the subtree; in other words, the indices stored in the patches need to be set up correctly. To be able to use the right indices, I changed the function from util to non-static member to understand the internal details of Link class.

So if I change to `AddOverrides(patchArray)`, it means it will manage **patch arrays** in undo node rather than subtree states.

I am still investigating how to improve this along with the indexing concern mentioned below.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Link/Link.cpp
forkercat: Yeah I feel it would be error prone as well. I used to add some logic to recalibrate index counter during tree rebuild but later changed to just increase the counter always.

Taking this as an example:
```
Index   0  1   2  3  4  5  6  |  7  8  9  10  11  12  13  14  15  |  16  17  18  19  20
Patch   [existing overrides]     [existing + new overrides]          [future overrides]
                                 (edit-entity node)                  (follow-up node)
```

(for the edit-entity node) When every of its undo/redo operations happens, the indices 7-15 won't conflict with other indices like 0-6 and 16-20. I am worried as you said if there is some issue generating the subtree that may break the consistency. But it looks like as we always increase the index counter, the future indices should always be greater than the one in the node and it won't cause any issue?

And note that the indices are automatically recalibrated (removing unused index gaps) every time we reload the level.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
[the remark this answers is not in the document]
forkercat: Yes! I actually noticed that as well. Maybe it can be improved by doing branching earlier in this function before generating the DOMs.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Link/Link.cpp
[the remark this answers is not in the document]
srikappa-amzn: I didn't mean to change the implementation detail. I just think that a public function in the link class is the wrong place for this function to be in. If you look at this function, all it does is increment m_patchIndexCounter. If that's the case, we might as well create a function called SetPatchIndexCounter() and call just that and move everything else to a static member function in PrefabUndoEntityOverrides.cpp

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Link/Link.cpp
srikappa-amzn: Had offline discussion about this. There are two options here to avoid the disconnect between the time we update the index and when we actually insert patches into the tree.

option 1 : make m_patchIndexCounter an atomic variable so that this won't become an issue in the future due to asynchronous requests
option 2 : Instead of adding the patches in the Redo() function, do it in the Capture() function itself and skip the first redo. By doing so, you'll do both the patch index increment and patch addition in the same function.

forkercat: > option 2 : Instead of adding the patches in the Redo() function, do it in the Capture() function itself and skip the first redo. By doing so, you'll do both the patch index increment and patch addition in the same function.

I am experimenting with option 2 with a function called `AddOverrides(path, patches)` that is called in Capture. Looks like it is working as expected.

Code/Framework/AzToolsFramework/Tests/Prefab/PrefabUndoEditEntityTests.cpp
```diff
@@ -0,0 +1,211 @@
...
+#include <AzToolsFramework/Entity/EditorEntityHelpers.h>
+#include <AzToolsFramework/Prefab/Undo/PrefabUndo.h>
+#include <AzToolsFramework/Prefab/Undo/PrefabUndoEntityOverrides.h>
+#include <Prefab/PrefabTestComponent.h>
+
+namespace UnitTest
+{
+    using PrefabUndoEditEntityTests = PrefabTestFixture;
+
+    TEST_F(PrefabUndoEditEntityTests, PrefabUndoEditEntityTests_EditEntity)
```
srikappa-amzn, line 20: Google test library appends the class name to the test name internally. So, the final name it'll generate for this test is PrefabUndoEditEntityTests_PrefabUndoEditEntityTests_EditEntity. The 'PrefabUndoEditEntityTests_' prefix to these tests can be removed as having underscores can cause a few issues in the tests. See this comment more information on this: https://github.com/o3de/o3de/pull/13926#discussion_r1063789238

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Link/Link.h
```diff
@@ -89,7 +89,7 @@ namespace AzToolsFramework
             void SetSourceTemplateId(TemplateId id);
             void SetTargetTemplateId(TemplateId id);
             void SetLinkDom(const PrefabDomValue& linkDom);
-            void AddPatchesToLink(const PrefabDom& patches);
+            void SetLinkPatches(const PrefabDom& patches);
```
srikappa-amzn, line 92: nice! This is more accurate.

michabr (approved):
Looks good. Left some minor comments. Could you also add a ticket to prevent existing entity overrides indices from changing when a new override is added?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoEntityOverrides.cpp
```diff
@@ -0,0 +1,172 @@
...
+                        continue;
+                    }
+                }
+
+                // Remove the subtree and cache the subtree for undo.
+                AZ::Dom::Path entityDomPath(entityPathFromTopInstance.c_str());
+                m_subtreeStates[entityDomPath] = AZStd::move(link->get().RemoveOverrides(entityDomPath));
+
+                // Redo - Add the override patches to the tree.
+                link->get().AddOverrides(entityDomPath, overridePatchesWithSubpaths);
```
srikappa-amzn, line 126: Rather than making the link class append the path to each patch, we can do that during the addition of patch itself in line 111. This will also make the public function in the link class more generic and reusable for other use cases. I recommend that since its a good practice to keep the base api classes as minimalistic and generic as possible and leave custom logic to classes like this. Please update function comments too if you do that.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabUndoHelpers.cpp
```diff
@@ -104,6 +105,17 @@ namespace AzToolsFramework
...
 
+            void UpdateEntitiesAsOverrides(
+                const AZStd::vector<const AZ::Entity*>& entityList,
+                Instance& owningInstance,
+                const Instance& focusedInstance,
+                UndoSystem::URSequencePoint* undoBatch)
+            {
+                PrefabUndoEntityOverrides* state = aznew PrefabUndoEntityOverrides("Undo Updating Entity List As Override");
+                state->SetParent(undoBatch);
+                state->CaptureAndRedo(entityList, owningInstance, focusedInstance);
```
srikappa-amzn, line 116: nit: maybe add a comment here or in the undo class as to why this undo node does both at the same time as opposed to other nodes.

srikappa-amzn (approved):
Looks good!

Left one comment to simplify the AddOverrides method even further.

Code/Framework/AzToolsFramework/Tests/Prefab/PrefabUndoEditEntityTests.cpp
forkercat: GP. I'll also create a task to clean up test names like these.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -913,15 +914,20 @@ namespace AzToolsFramework
                     }
                     else
                     {
-                        PrefabDom newPatch;
-                        m_instanceToTemplateInterface->GeneratePatch(newPatch, beforeState, afterState);
-                        LinkId linkId = m_prefabFocusInterface->AppendPathFromFocusedInstanceToPatchPaths(newPatch, entityId);
-
-                        Internal_HandleContainerOverride(parentUndoBatch, entityId, newPatch, linkId);
+                        AzFramework::EntityContextId editorEntityContextId = AzFramework::EntityContextId::CreateNull();
+                        EditorEntityContextRequestBus::BroadcastResult(
```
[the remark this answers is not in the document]
forkercat: It can be changed to call once for a bunch of dirty entities. But I think it might not be expensive as it didn't show obvious slowdown in profiling.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoEntityOverrides.cpp
```diff
@@ -0,0 +1,154 @@
...
+#include <AzToolsFramework/Prefab/PrefabSystemComponentInterface.h>
+
+namespace AzToolsFramework
+{
+    namespace Prefab
+    {
+        PrefabUndoEntityOverrides::PrefabUndoEntityOverrides(const AZStd::string& undoOperationName)
+            : UndoSystem::URSequencePoint(undoOperationName)
+            , m_linkId(InvalidLinkId)
+            , m_changed(true)
```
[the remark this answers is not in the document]
forkercat: Looks like it can be removed and have `Changed()` always return true.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoEntityOverrides.cpp
```diff
@@ -0,0 +1,154 @@
...
+            PrefabDomReference cachedOwningInstanceDom = owningInstance.GetCachedInstanceDom();
+
+            const AZStd::string overridePatchPathToOwningInstance = PrefabInstanceUtils::GetRelativePathFromClimbedInstances(
+                climbUpResult.m_climbedInstances, true); // skip top instance
+
+            for (const AZ::Entity* entity : entityList)
+            {
+                if (!entity)
+                {
+                    AZ_Warning("Prefab", false, "PrefabUndoEntityOverrides::Capture - Found a null entity. Skipping it.");
```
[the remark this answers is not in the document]
forkercat: Discussed offline. Warning is more appropriate here.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Link/Link.cpp
```diff
@@ -293,11 +300,38 @@ namespace AzToolsFramework
...
 
-                    auto path = patchEntry.FindMember("path");
-                    if (path != patchEntry.MemberEnd())
+                    auto pathIter = patchEntry.FindMember("path");
+                    if (pathIter != patchEntry.MemberEnd())
                     {
-                        AZ::Dom::Path domPath(path->value.GetString());
-                        PrefabOverrideMetadata overrideMetadata(AZStd::move(patchEntry), i);
+                        AZ::Dom::Path domPath(pathIter->value.GetString());
+                        PrefabOverrideMetadata overrideMetadata(AZStd::move(patchEntry), m_patchIndexCounter++);
```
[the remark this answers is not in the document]
forkercat: I think it can be reset to 0. I'll do some testing to see if there could be any issue.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoEntityOverrides.cpp
forkercat: Good callout! The previous workflow was for GenerateSubtree. After changing to AddOverrides(patches), it can be simplified a lot.

PR #13765, opened 2022-12-19
Materials reference the original material type

santorac:
## What does this PR do?

Materials reference the original material type even when intermediate material types are involved.

This allows Material Editor to create new materials based on the original .materialtype file and both Material Editor and Material Builder understand the indirect reference to the final MaterialTypeAsset.

Also fixed a bug where child materials weren't always reprocessed when they should be, if a .materialtype is changed. See https://github.com/o3de/o3de/issues/13766

## How was this PR tested?
AtomSampleViewer material screenshot test script
Created new materials in material editor, using both StandardPBR and the new pipeline based equivalent material type. Also a child material. These are processed successfully by the AP and opened successfully my Material Editor.
Tested removing and adding the pipeline test .materialtype files, saw the corresponding .material files fail, and then replaced the missing .materialtype files and saw the .material files pass again.

The only issue is that some materials will now report a warning about a missing job dependency, like this:
```
No job was found to match the job dependency criteria declared by file D:/o3de/AtomSampleViewer/Assets/NewStandardPipelineChild.material. (File: D:/o3de/Gems/Atom/TestData/TestData/Materials/Types/MaterialPipelineTest_Standard.materialtype, JobKey: Material Type Builder (Final Stage), Platform: pc)
```
Everything should still function properly. I think we can have a separate PR to address this issue. (We could use a wildcard dependency on like "StandardPBR*.materialtype", but I'd rather not do that as it would pick up other material types beyond just "StandardPBR.materialtype" and "standardpbr_generated.materialtype". Also that might not work on Linux because of the difference in case between the source file and the intermediate file. A cleaner and simpler option that I think would work would be to use the same job key for both stages of the material builder instead of separate "Material Type Builder (Pipeline Stage)" and "Material Type Builder (Final Stage)" keys.

### DIFF
--- a/Gems/Atom/RPI/Code/Include/Atom/RPI.Edit/Material/MaterialTypeSourceData.h
+++ b/Gems/Atom/RPI/Code/Include/Atom/RPI.Edit/Material/MaterialTypeSourceData.h
@@ -38,6 +38,8 @@ namespace AZ
 
             static constexpr const char Extension[] = "materialtype";
 
+            static constexpr AZ::u32 IntermediateMaterialTypeSubId = 0;
+
             static void Reflect(ReflectContext* context);
 
             //! The .materialtype file has two slightly different formats, in most cases users will want to author content in the abstract format, which is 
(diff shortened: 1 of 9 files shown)


galibzon (approved):

gadams3 (approved):
Should we expect these kinds of patterns and explicit handling anytime intermediate assets are used? It's probably worth mentioning any observations or opportunities to the group responsible for intermediate assets.

amzn-mike (approved):

PR #13767, opened 2022-12-19
Potential fix for ShapeIntersection unit test on Android (Scalar)

moraaar:

## What does this PR do?

Fix ShapeIntersection test on Android (Scalar) by addressing float overflows in Overlaps().

The trick of separating `0.5f * (GetMax() - GetMin())` into two separate multiplies before the subtraction `(0.5f * aabb.GetMax()) - (0.5f * aabb.GetMin())` to avoid overflowing when using -FLT_MAX and FLT_MAX doesn't work on Android, it gets optimized away and therefore overflowing.

## How was this PR tested?

Before:
````
[ RUN      ] MATH_ShapeIntersection.Test
E:/Amazon/GitHub/o3de/Code/Framework/AzCore/Tests/Math/ShapeIntersectionTests.cpp:80: Failure
Value of: AZ::ShapeIntersection::Overlaps(frustum, maxSizeAabb)
  Actual: false
Expected: true
[  FAILED  ] MATH_ShapeIntersection.Test (0 ms)
````

After:
````
[ RUN      ] MATH_ShapeIntersection.Test
[       OK ] MATH_ShapeIntersection.Test (0 ms)
````

### DIFF
--- a/Code/Framework/AzCore/AzCore/Math/ShapeIntersection.inl
+++ b/Code/Framework/AzCore/AzCore/Math/ShapeIntersection.inl
@@ -178,15 +178,33 @@ namespace AZ
             //The AABB is not overlapping if it is fully behind any of the planes, otherwise it is overlapping.
             const Vector3 center = aabb.GetCenter();
 
-            //If the AABB contains FLT_MAX at either (or both) extremes, it would be easy to overflow here by using "0.5f * GetExtents()"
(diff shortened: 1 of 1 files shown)


moraaar:
After discussing this and other potential ways to address this issues, it was concluded that it'd be better to not trying to workaround --ffast-math by doing some less performant solution without being absolutely necessary and since the default SIMD set in Android is NEON now and it passes, it will be left as is for the time being. If in the future this test starts failing in the same spot we can reevaluate options.

https://github.com/o3de/o3de/pull/13783

PR #13780, opened 2022-12-19
External repo update to JenkinsFile

byrcolin:
Currently the way AR works is that o3de and o3de-extras are always pulled an linked, depending on which repo the commit is for one repo will be pulled at the branch and the other development branch.

This change will make it so that each repo declares which other repos are pulled and what branch. The .automatedtesting.json will now have another section "REPOS". o3de's .automatedtesting.json will now include o3de-extras like so:

"REPOS": [
        {
            "NAME": "o3de-extras",
            "URL": "https://github.com/o3de/o3de-extras.git",
            "BRANCH": "development"
        }
    ]
Now when AR runs it will still always pull o3de as it is the engine, if the commit is for the engine it pulls that branch, if not it pulls development. Then it loads o3de's .automatedtesting.json finds it should pull o3de-extras at development branch. If we want to make any number of other repos canonical, we would only need to add it to the REPOS list for the o3de. These REPOS are recursive, so in the above scenario if o3de-extras had REPOS in its .automatedtesting.json they would also be pulled and linked. This should make any repo that has a dependency on any other repo to be able to run AR. This also means that the branch repo could be ANY repo. So our sample projects can also take advantage of this and don't need their own JenkinsFiles. They just need to point to the o3de JenkinsFile and everything should work for them.
This change then makes 3 scenarios:
1. The commit is for the core engine repo (o3de/o3de)
2. The commit is for a dependent repo (like o3de/o3de-extras)
3. The commit is for a non-dependent repo (everything else, like o3de-multiplayersample)

If the commit is for the core engine repo: it pulls the branch of the engine, which also pulls o3de-extras development branch and runs AR.
If the commit is for a dependent repo (in this case the only dependent repo is o3de-extras but in the future there may be more): it pulls o3de engine development branch, which also pulls o3de-extras development branch, finds the commit is not for the engine but a branch of o3de-extras, pulls that branch and runs AR. So even though the extras branch was already pulled at development, it will then pull the needed o3de-extras branch.
If the commit is for a non-dependent repo: it will pull the o3de engine development, which also pulls o3de-extras development. finds that the commit is not for the engine or a dependent repo, but a branch of yet another repo and since the commit is not the engine it will pull that repo at that branch and run AR.


amzn-changml (commented):
Looks mostly reasonable. Some questions about our existing official repos structure and some best practices with this new logic:

.automatedtesting.json
```diff
@@ -0,0 +1,21 @@
...
+    "REGISTER_RESTRICTED": [
+    ],
+    "ENABLE_GEMS": [
+    ],
+    "REPOS": [
+        {
+            "NAME": "o3de-extras",
+            "URL": "https://github.com/o3de/o3de-extras.git",
+            "BRANCH": "development"
+        }
```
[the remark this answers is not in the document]
byrcolin: Yes, we would no longer hard code any other repo than o3de/o3de. Every repo will list their dependencies. We will treat o3de-extras as a dependency of o3de as it is consider a canonical repo of o3de.

.automatedtesting.json
```diff
@@ -0,0 +1,21 @@
+{
+    "REGISTER_ENGINES": [
+    ],
+    "REGISTER_PROJECTS": [
+    ],
```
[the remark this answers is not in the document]
byrcolin: This is o3de-extras .automatedtesting.json
```
{
    "REGISTER_ENGINES": [
    ],
    "REGISTER_PROJECTS": [
        "Projects/OpenXRTest"
    ],
    "REGISTER_GEMS": [
        "Gems/XR",
        "Gems/OpenXRVk"
    ],
    "REGISTER_TEMPLATES": [
        "Templates/Multiplayer"
    ],
    "REGISTER_RESTRICTED": [
    ],
    "ENABLE_GEMS": [
        "Gems/XR",
        "Gems/OpenXRVk"
    ],
    "REPOS": [
        {
            "NAME": "o3de",
            "URL": "https://github.com/o3de/o3de.git",
            "BRANCH": "development"
        }
    ]
}
```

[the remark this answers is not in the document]
byrcolin: While it is debatable that extras would need to specify the engine as a dependency as it is always assumed to be part of any ar run, this was used to capture the case of cyclical dependencies. The final version of extras will probably not have a reference to o3de, so i left it in there for reference.

scripts/build/Jenkins/Jenkinsfile
```diff
@@ -384,6 +360,190 @@ def HandleDriveMount(String snapshot, String repositoryName, String projectName,
...
+                    if(fileExists("Templates")) {
+                        templatesExist = true
+                    }
+                    if(fileExists("Restricted")) {
+                        restrictedExist = true
+                    }
+
+                    //If this is not a standard composite repo, then it wont have any composite dirs, then we know
+                    //its a singular object repo, see what kind it is.
+                    if(!enginesExist && !projectsExist && !gemsExist && !templatesExist && !restrictedExist) {
```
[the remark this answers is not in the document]
byrcolin: No, the repo astom-sampleviewer is conformant because it is a singular repo that has children, like the engine and is not a compound repo like extras. In singular repos only the top level object needs top be linked and all children are known by the parent. In a compund repo like extras there is no parent object, as the repo.json is not a parent.

scripts/build/Jenkins/Jenkinsfile
```diff
@@ -384,6 +360,190 @@ def HandleDriveMount(String snapshot, String repositoryName, String projectName,
...
+            }
+        }
+    }
+
+    //check out all repos
+    if(controlFile.REPOS) {
+        echo 'REPOS'
+        controlFile.REPOS.each { repo ->
+            //stop cyclical repos
+            if(PROCESSED_REPOS.contains(repo.NAME)) {
```
amzn-changml, line 420: As discussed offline: We may need to verify this test against workspaces that are restored from snapshot. The o3de and possibly the o3de-extras folder will already exist at development on first run

.automatedtesting.json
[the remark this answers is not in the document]
lumberyard-employee-dm: Should we add this file to the .CODEOWNERS file under the sig-build-reviewers and sig-build-maintainers?

scripts/build/Jenkins/Jenkinsfile
```diff
@@ -21,12 +21,9 @@ PIPELINE_TIMEOUT = 60
 
 EMPTY_JSON = readJSON text: '{}'
 
-ENGINE_DEFAULT_ORGANIZATION_NAME = 'o3de'
 ENGINE_REPOSITORY_NAME = 'o3de'
 ENGINE_DEVELOPMENT_BRANCH = 'development'
-EXTRAS_DEFAULT_ORGANIZATION_NAME = 'o3de'
-EXTRAS_REPOSITORY_NAME = 'o3de-extras'
-EXTRAS_DEVELOPMENT_BRANCH = 'development'
+ENGINE_URL = "https://github.com/o3de/o3de.git"
```
brianherrera, line 26: If we hardcode the url here instead of loading it from scm.userRemoteConfigs.url, will this affect builds running from a branch in a fork? Meaning will the engine repo still come from o3de/o3de.git instead of fork/o3de.git

scripts/build/Jenkins/Jenkinsfile
```diff
@@ -384,6 +358,208 @@ def HandleDriveMount(String snapshot, String repositoryName, String projectName,
...
+        }
+        if(templatesExist) {
+            PlatformSh("scripts/o3de${ext} register --all-templates-path ${workspace}/${repoName}/Templates || ver>nul", "Registering all templates in ${workspace}/${repoName}/Templates")
+        }
+        if(restrictedExist) {
+            PlatformSh("scripts/o3de${ext} register --all-restricted-path ${workspace}/${repoName}/Restricted || ver>nul", "Registering all restricted in ${workspace}/${repoName}/Restricted")
+        }
+        //enable all gems in this repo
+        if(gemsExist) {
+            PlatformSh("scripts/o3de${ext} enable-gem --all-gem-paths ${workspace}/${repoName}/Gems --project-name AutomatedTesting || ver>nul", "Enabling all gems in AutomatedTesting")
```
brianherrera, line 449: Can this `enable-gem` command be combined with the `if(gemsExist)` block above?

scripts/build/Jenkins/Jenkinsfile
```diff
@@ -401,29 +577,27 @@ def PreBuildCommonSteps(Map pipelineConfig, String snapshot, String repositoryNa
     // we need to cleanup. Once all branches take this relocation, we can remove this
     if(env.CLEAN_WORKSPACE?.toBoolean() || fileExists("${workspace}/.git")) {
         if(fileExists(workspace)) {
+            echo 'Clean workspace...'
             PlatformRmDir(workspace)
         }
     }
 
-    def workspace_retry = 5
```
brianherrera, line 408: @amzn-changml discovered an issue where EBS volume may not be immediately available for writes once mounted which is the reason I think the retry exists. Is this issue resolved?

scripts/build/Jenkins/Jenkinsfile
byrcolin: ENGINE_URL is only used if the commit repo name is NOT o3de, in which case it pulls the development branch of o3de/o3de. If the commit repo is */o3de, meaning o3de/o3de or aws-lumberyard-dev/o3de or apmg/o3de, then it will use it.

scripts/build/Jenkins/Jenkinsfile
byrcolin: I don't think so. We want the scripts to run from dir(ENGINE_REPOSITORY_NAME) but the decision to run is based on if a different folder/file exists. The determination of if a file or folder exists is the fileExists() function, and it depends on the current folder so I have to be in that dir to make the determination but I need to be in the engine dir in order to run the command, so i have to save off the determination, change dir, execute based on the determination. I don't know of any way to call fileExists() or PlatformSh() from a dir outside the current dir... This logic is simple enough and we know this works, so we should just keep it.

scripts/build/Jenkins/Jenkinsfile
byrcolin: I removed the retry because it was causing my builds to fail.

scripts/build/Jenkins/Jenkinsfile
byrcolin: I'll try it

brianherrera: I'm fine to leave as-is if it keeps the jenkinsfile logic simplified.

scripts/build/Jenkins/Jenkinsfile
brianherrera: Got it. We can remove it if this is causing issues with the updated folder setup.

scripts/build/Jenkins/Jenkinsfile
byrcolin: we also need to test 2 more cases:
1. singular repo with no .automatedtesting.json
2. a compound repo with no .automatedtesting.json

scripts/build/Jenkins/Jenkinsfile
byrcolin: updated: needs to be tested though.

scripts/build/Jenkins/Jenkinsfile
amzn-changml: I curious on what the failure was here, but fine with removing it if the prebuild setup is implicitly looking for the presence of the workspace folder.

scripts/build/Jenkins/Jenkinsfile
byrcolin: done

aws-lumberyard-dev (dismissed):

amzn-changml (approved):
Looks reasonable. As discussed offline, will need to prove it out with external projects

brianherrera (approved):

PR #13787, opened 2022-12-20
Adding actuation to PhysX joints

michalpelka:
## What does this PR do?

This PR introduces actuation to O3DE PhysX joints

## How was this PR tested?

Manually using this test project https://github.com/RobotecAI/o3de-physics-test-scene/pull/2.


Gems/PhysX/Code/Editor/EditorJointConfiguration.cpp
```diff
@@ -512,4 +514,37 @@ namespace PhysX
...
+
+    void EditorJointMotorConfig::Reflect(AZ::ReflectContext* context)
+    {
+        if (auto* serializeContext = azrtti_cast<AZ::SerializeContext*>(context))
+        {
+            serializeContext->Class<EditorJointMotorConfig>()
+                    ->Version(1)
+                    ->Field("Use Motor", &EditorJointMotorConfig::m_useMotor)
+                    ->Field("Force Limit", &EditorJointMotorConfig::m_driveForceLimit)
+                    ->Field("Gear Ratio", &EditorJointMotorConfig::m_gearRatio)
```
hultonha, line 526: nit: This is super minor but I think for serialized fields (`SerializeContext` not `EditContext` we prefer no spaces), so these would be `UseMotor`, `ForceLimit` etc...

Gems/PhysX/Code/Editor/EditorJointConfiguration.cpp
```diff
@@ -512,4 +514,37 @@ namespace PhysX
...
+                        ->DataElement(0, &PhysX::EditorJointMotorConfig::m_driveForceLimit, "Force Limit Value"
+                                , "Sets force limit value")
+                        ;
+            }
+        }
+    }
+
+    JointMotorProperties EditorJointMotorConfig::ToGameTimeConfig() const{
+        return JointMotorProperties(m_useMotor, m_gearRatio, m_driveForceLimit);
+    }
```
hultonha, line 549: nit/standards: Might be worth running clang-format over this code to make sure braces appear on a new line etc...

Gems/PhysX/Code/Include/PhysX/Joint/Configuration/PhysXJointConfiguration.h
```diff
@@ -84,6 +84,22 @@ namespace PhysX
         float m_tolerance = 0.1f; ///< Distance from the joint at which limits becomes enforced. Used if limit is hard.
     };
 
+    struct JointMotorProperties
```
hultonha, line 87: question: This looks very similar to `EditorJointMotorConfig`, could they share the same internal struct or could we just pick one? Just curious why we need both and there's a chance they could get out of sync (the naming variable names are also super similar but not quite the same which could cause confusion).

Gems/PhysX/Code/Include/PhysX/Joint/PhysXJointBus.h
```diff
@@ -0,0 +1,52 @@
...
+        static constexpr AZ::EBusAddressPolicy AddressPolicy = AZ::EBusAddressPolicy::ById;
+        using BusIdType = AZ::EntityId;
+
+        virtual ~JointInterface() = default;
+
+        //! Returns current position from joint
+        virtual float GetPosition() = 0;
+
+        //! Returns velocity of joint
+        virtual float GetVelocity() = 0;
```
hultonha, line 33: suggestion: For these getters, even though this is intended to be used as a bus, it's likely a good idea to mark them `const` (if we do want to cache something we can always rely in `mutable` in the implementation class)

Gems/PhysX/Code/Include/PhysX/Joint/PhysXJointBus.h
```diff
@@ -0,0 +1,52 @@
...
+#include <AzCore/Component/EntityId.h>
+#include <AzCore/EBus/EBus.h>
+#include <AzCore/Interface/Interface.h>
+#include <AzCore/std/utils.h>
+#include <AzCore/Math/Transform.h>
+
+namespace PhysX
+{
+    //! Interface to communicate with PhysX joint's motor.
+    class JointInterface : public AZ::EBusTraits
```
hultonha, line 21: update: Please name this `JointInterfaceRequests` (or maybe just `JointRequests`) to be more consistent with other EBus naming across the engine

Gems/PhysX/Code/Include/PhysX/Joint/PhysXJointBus.h
```diff
@@ -0,0 +1,52 @@
...
+    {
+    public:
+        static constexpr AZ::EBusAddressPolicy AddressPolicy = AZ::EBusAddressPolicy::ById;
+        using BusIdType = AZ::EntityId;
```
hultonha, line 25: question: It might be worth double checking if we want this to be addressed by EntityComponentIdPair instead of EntityId (this would make it possible to have more than one joint on an entity if we wanted that). That might not be something that comes up but I just wanted to mention it in case.

Gems/PhysX/Code/Include/PhysX/Joint/PhysXJointBus.h
```diff
@@ -0,0 +1,52 @@
...
+        virtual AZStd::pair<float, float> GetLimits() = 0;
+
+        //! Sets maximum motor force.
+        //! @param force in Newtons or Newton-meters.
+        virtual void SetMaximumForce(float force) = 0;
+
+    };
+
+    using JointInterfaceRequestBus = AZ::EBus<JointInterface>;
+} // namespace ROS2
```
hultonha, line 52: uber nit: missing EOL character (most editors not have a setting to add this automatically on save - VSCode definitely does if you're using this)

Gems/PhysX/Code/Editor/EditorJointConfiguration.h
```diff
@@ -71,6 +71,24 @@ namespace PhysX
         float m_limitNegative = -45.0f;
     };
 
+    /// Pair (angles) limits for joints.
+    class EditorJointMotorConfig
```
hultonha, line 75: nit: I believe most classes use the full `Configuration` as opposed to the abbreviated `Config`, might be nice to change this just to be consistent (if there's already a mix elsewhere feel free to ignore this)

Gems/PhysX/Code/Source/HingeJointComponent.cpp
```diff
@@ -85,4 +88,60 @@ namespace PhysX
             m_jointSceneOwner = leadFollowerInfo.m_followerBody->m_sceneOwner;
         }
     }
+
+    // AZ::Component
```
hultonha, line 92: nit: This comment can be removed from the .cpp

Gems/PhysX/Code/Source/HingeJointComponent.cpp
```diff
@@ -85,4 +88,60 @@ namespace PhysX
...
+    void HingeJointComponent::Activate() {
+        JointComponent::Activate();
+        PhysX::JointInterfaceRequestBus::Handler::BusConnect(this->GetEntityId());
+    }
+
+    void HingeJointComponent::Deactivate() {
+        JointComponent::Deactivate();
+        PhysX::JointInterfaceRequestBus::Handler::BusDisconnect();
+    }
```
hultonha, line 101: nit: Braces again (clang-format should fix this)

Also it's usually good practice to disconnect and deactivate in the reverse order, so this would be...

```
PhysX::JointInterfaceRequestBus::Handler::BusDisconnect();
JointComponent::Deactivate();
```

instead, thanks!

Gems/PhysX/Code/Source/HingeJointComponent.cpp
```diff
@@ -85,4 +88,60 @@ namespace PhysX
...
+
+    physx::PxRevoluteJoint* GetPhysXNativeRevoluteJoint(const AzPhysics::JointHandle& joint_handle,
```
hultonha, line 103: nit/standards: We don't use underscores unfortunately so variable names would be `jointHandle` and `jointSceneOwner` etc...

Gems/PhysX/Code/Source/HingeJointComponent.cpp
```diff
@@ -85,4 +88,60 @@ namespace PhysX
...
+        return native;
+    }
+
+    float HingeJointComponent::GetPosition(){
+        return GetPhysXNativeRevoluteJoint(m_jointSceneOwner,m_jointHandle)->getAngle();
+    }
+
+    float HingeJointComponent::GetVelocity() {
+        return GetPhysXNativeRevoluteJoint(m_jointSceneOwner,m_jointHandle)->getVelocity();
+    }
```
hultonha, line 122: nit (repeated): Braces on new line (I won't bother mentioning it again)

Gems/PhysX/Code/Source/HingeJointComponent.cpp
```diff
@@ -85,4 +88,60 @@ namespace PhysX
...
+        const auto type = joint->GetNativeType();
+        AZ_Assert(type == PhysX::NativeTypeIdentifiers::HingeJoint, "It is not PhysXHingeJoint");
+        physx::PxJoint *native_joint = static_cast<physx::PxJoint*>(joint->GetNativePointer());
+        physx::PxRevoluteJoint *native =native_joint->is<physx::PxRevoluteJoint>();
+        AZ_Assert(native, "It is not PxRevoluteJoint");
+        return native;
+    }
+
+    float HingeJointComponent::GetPosition(){
+        return GetPhysXNativeRevoluteJoint(m_jointSceneOwner,m_jointHandle)->getAngle();
```
hultonha, line 117: suggestion: We might want to cache the value returned be `GetPhysXNativeRevoluteJoint` so we don't have to invoke it for every `Get` call (not critical but might be a performance win depending on how expensive `GetJointFromHandle` is.

Gems/PhysX/Code/Source/HingeJointComponent.cpp
```diff
@@ -85,4 +88,60 @@ namespace PhysX
...
+    }
+
+
+    AZStd::pair<float, float> HingeJointComponent::GetLimits(){
+        auto limit = GetPhysXNativeRevoluteJoint(m_jointSceneOwner,m_jointHandle)->getLimit();
+        return AZStd::pair<float, float>(limit.lower,limit.upper);
+    }
+
+    AZ::Transform HingeJointComponent::GetTransform(){
+        auto t =GetPhysXNativeRevoluteJoint(m_jointSceneOwner,m_jointHandle)->getRelativeTransform();
```
hultonha, line 131: nit/standards: We tend to avoid single letter variable names so this would be better as `transform` (or my preferred `worldFromLocal` to show the spaces the transform is moving from/to)

Gems/PhysX/Code/Source/HingeJointComponent.h
```diff
@@ -24,9 +26,24 @@ namespace PhysX
         HingeJointComponent(
             const JointComponentConfiguration& configuration, 
             const JointGenericProperties& genericProperties,
-            const JointLimitProperties& limitProperties);
+            const JointLimitProperties& limitProperties,
+            const JointMotorProperties& motorProperties = JointMotorProperties());
         ~HingeJointComponent() = default;
 
+        // JointInterfaceRequestBus::Handler
```
hultonha, line 33: nit: The preferred style for overrides is `// JointInterfaceRequestBus overrides ...` etc (super minor but would be great to update if possible)

Gems/PhysX/Code/Source/HingeJointComponent.h
```diff
@@ -24,9 +26,24 @@ namespace PhysX
...
+        float GetPosition() override;
+        float GetVelocity() override;
+        AZ::Transform GetTransform() override;
+
+        // JointMotorRequestBus ::Handler
```
hultonha, line 38: update: Is this correct? It seems like this should be grouped with the member functions from `JointInterfaceRequestBus`?

Gems/PhysX/Code/Source/Joint/Configuration/PhysXJointConfiguration.cpp
```diff
@@ -35,6 +35,14 @@ namespace PhysX
 
     }
 
+    JointMotorProperties::JointMotorProperties(bool enabled, float gear_ratio, float force_limit)
```
hultonha, line 38: nit: Parameter names underscore naming -> `gearRatio`, `forceLimit`

Gems/PhysX/Code/Source/Joint/PhysXJoint.cpp
```diff
@@ -133,6 +133,7 @@ namespace PhysX
         : PhysXJoint(sceneHandle, parentBodyHandle, childBodyHandle)
     {
         m_pxJoint = Utils::PxJointFactories::CreatePxPrismaticJoint(configuration, sceneHandle, parentBodyHandle, childBodyHandle);
+        AZ_Printf("PhysXPrismaticJoint::PhysXPrismaticJoint", " PhysXPrismaticJoint %d", m_pxJoint.get());
```
hultonha, line 136: update: Might want to add a reminder to remove this before moving out of draft status

Gems/PhysX/Code/Source/Joint/PhysXJointUtils.cpp
```diff
@@ -209,6 +209,42 @@ namespace PhysX::Utils
         nativeJoint->setPrismaticJointFlag(physx::PxPrismaticJointFlag::eLIMIT_ENABLED, true);
     }
 
+    void InitializePrismaticLimitD6Properties(const JointLimitProperties& properties, physx::PxD6Joint* nativeJoint)
+    {
+
+        if (!nativeJoint)
+        {
+            return;
+        }
```
hultonha, line 218: suggestion: As this could fail we might want to return a result from this function (could be `bool` or an `AZ::Outcome` that might be more useful to the caller)

Gems/PhysX/Code/Source/JointComponent.h
```diff
@@ -73,8 +79,8 @@ namespace PhysX
         };
 
         // AZ::Component
-        void Activate() override;
-        void Deactivate() override;
+        virtual void Activate() override;
+        virtual void Deactivate() override;
```
hultonha, line 83: nit: `virtual` keyword can be removed here as `override` is present

Gems/PhysX/Code/Source/PrismaticJointComponent.cpp
```diff
@@ -84,4 +87,83 @@ namespace PhysX
...
+
+    // AZ::Component
+    void PrismaticJointComponent::Activate() {
+        JointComponent::Activate();
+        PhysX::JointInterfaceRequestBus::Handler::BusConnect(this->GetEntityId());
+    };
+
+    void PrismaticJointComponent::Deactivate() {
+        JointComponent::Deactivate();
+        PhysX::JointInterfaceRequestBus::Handler::BusDisconnect();
```
hultonha, line 99: update: Disconnect order (see earlier comment)

Gems/PhysX/Code/Source/PrismaticJointComponent.cpp
```diff
@@ -84,4 +87,83 @@ namespace PhysX
...
+    };
+
+    physx::PxD6Joint* GetPhysXD6Joint(const AzPhysics::JointHandle& joint_handle,
+                                                        const AzPhysics::SceneHandle& joint_scene_owner)
+    {
```
hultonha, line 104: nit: Indentation looks a bit out here

Gems/PhysX/Code/Source/PrismaticJointComponent.cpp
```diff
@@ -84,4 +87,83 @@ namespace PhysX
...
+    };
+
+
+    AZStd::pair<float, float> PrismaticJointComponent::GetLimits(){
+        auto* native = GetPhysXD6Joint(m_jointSceneOwner,m_jointHandle);
+        if (native){
+            auto limits = native->getLinearLimit(physx::PxD6Axis::eX);
+            return AZStd::pair<float, float>(limits.lower,limits.upper);
+        }
+        return AZStd::pair<float, float> (0.f,0.f);
```
hultonha, line 139: uber nit: Could use `AZStd::make_pair` here and avoid the need for template args

(this actually should work without as-is due to CTAD in C++17 I believe, but I could be wrong 🤔)

Gems/PhysX/Code/Source/PrismaticJointComponent.h
```diff
@@ -27,9 +29,24 @@ namespace PhysX
...
         ~PrismaticJointComponent() = default;
 
+        // JointInterfaceRequestBus::Handler
+        float GetPosition() override;
+        float GetVelocity() override;
+        AZ::Transform GetTransform() override;
+
+        // JointMotorRequestBus ::Handler
+        void SetVelocity(float velocity) override;
+        void SetMaximumForce(float force) override;
```
hultonha, line 43: update: API comment block seems incorrect here again (should just be `JointInterfaceRequestBus overrides ...`)

hultonha (commented):
Overall this looks like a great change to me, my feedback was mostly around superficial things. I'l love to hear from @greerdv and @SergeyAMZN but from what I can tell this looks like a great addition. Thanks for adding! 😄

Gems/PhysX/Code/Editor/EditorJointConfiguration.h
moraaar: In this file there is `EditorJointLimitConfig` and `EditorJointLimitPairConfig`, so having Config here is fine to be consistent.

Gems/PhysX/Code/Editor/EditorJointConfiguration.h
```diff
@@ -71,6 +71,24 @@ namespace PhysX
...
+        static const float MaxMotorForce;
+
+        AZ_CLASS_ALLOCATOR(EditorJointMotorConfig, AZ::SystemAllocator, 0);
+        AZ_TYPE_INFO(EditorJointMotorConfig, "{8EADE1F7-40F5-44B1-8389-FDBB972E49C6}");
+        static void Reflect(AZ::ReflectContext* context);
+
+        JointMotorProperties ToGameTimeConfig() const;
+
+        bool m_useMotor{false}; ///< True if joint uses motor.
+        float m_gearRatio{1}; ///< Gear ratio.
```
moraaar, line 87: nit: please use `1.0f` to match the type.

Gems/PhysX/Code/Include/PhysX/Joint/Configuration/PhysXJointConfiguration.h
```diff
@@ -84,6 +84,22 @@ namespace PhysX
...
+    {
+        AZ_CLASS_ALLOCATOR(JointMotorProperties, AZ::SystemAllocator, 0);
+        AZ_TYPE_INFO(JointMotorProperties, "{9CF35393-82CD-4726-B387-96F6381046B3}");
+        static void Reflect(AZ::ReflectContext* context);
+
+        JointMotorProperties() = default;
+        JointMotorProperties(
+            bool enabled, float gear_ratio, float force_limit);
```
moraaar, line 95: [standard] name convention for parameters:
gear_ratio -> gearRatio
force_limit -> forceLimit

Gems/PhysX/Code/Include/PhysX/Joint/Configuration/PhysXJointConfiguration.h
```diff
@@ -84,6 +84,22 @@ namespace PhysX
...
+
+        bool m_enabled = false; ///< Enables joint actuation.
+        float m_gearRatio = 1.0f; ///< Gear ratio.
+        float m_forceLimit = 100000.0f; ///< Force limit.
```
moraaar, line 99: nit: Please expand the comments a bit more or just remove them. If it's the same as the variable name it doesn't provide useful information.

Gems/PhysX/Code/Include/PhysX/Joint/PhysXJointBus.h
moraaar: I vote for `JointRequests`.

Gems/PhysX/Code/Include/PhysX/Joint/PhysXJointBus.h
```diff
@@ -0,0 +1,52 @@
...
+    class JointInterface : public AZ::EBusTraits
+    {
+    public:
+        static constexpr AZ::EBusAddressPolicy AddressPolicy = AZ::EBusAddressPolicy::ById;
+        using BusIdType = AZ::EntityId;
+
+        virtual ~JointInterface() = default;
+
+        //! Returns current position from joint
+        virtual float GetPosition() = 0;
```
moraaar, line 30: How come `GetPosition` only returns a single float? Shouldn't return a AZ::Vector3? What kind of position is this?

Gems/PhysX/Code/Include/PhysX/Joint/PhysXJointBus.h
moraaar, line 33: Similar question for velocity? How come it's only single float, could it be Vector3?

Gems/PhysX/Code/Include/PhysX/Joint/PhysXJointBus.h
```diff
@@ -0,0 +1,52 @@
...
+
+        //! Returns velocity of joint
+        virtual float GetVelocity() = 0;
+
+        //! Sets drive velocity.
+        //! @param velocity velocity in meters per second or radians per second
```
moraaar, line 36: >velocity in meters per second or radians per second

A bit confusing that can be parameter passed can be either meters per second or radians per second, shouldn't just we choose one?

Gems/PhysX/Code/Include/PhysX/Joint/PhysXJointBus.h
```diff
@@ -0,0 +1,52 @@
...
+        virtual void SetVelocity(float velocity) = 0;
+
+        //! Gets local transformation of joint.
+        virtual AZ::Transform GetTransform() = 0;
+
+        //! Gets limits on active joint's axis.
+        virtual AZStd::pair<float, float> GetLimits() = 0;
+
+        //! Sets maximum motor force.
+        //! @param force in Newtons or Newton-meters.
```
moraaar, line 46: Same question. Newtons and Newtown-meters different? if so it might be clearer just use one.

Gems/PhysX/Code/Source/HingeJointComponent.cpp
```diff
@@ -85,4 +88,60 @@ namespace PhysX
             m_jointSceneOwner = leadFollowerInfo.m_followerBody->m_sceneOwner;
         }
     }
+
+    // AZ::Component
+    void HingeJointComponent::Activate() {
+        JointComponent::Activate();
+        PhysX::JointInterfaceRequestBus::Handler::BusConnect(this->GetEntityId());
```
moraaar, line 95: nit: `this->` shouldn't be necessary: `this->GetEntityId()` -> `GetEntityId()`

Gems/PhysX/Code/Source/HingeJointComponent.cpp
```diff
@@ -85,4 +88,60 @@ namespace PhysX
...
+    void HingeJointComponent::Deactivate() {
+        JointComponent::Deactivate();
+        PhysX::JointInterfaceRequestBus::Handler::BusDisconnect();
+    }
+
+    physx::PxRevoluteJoint* GetPhysXNativeRevoluteJoint(const AzPhysics::JointHandle& joint_handle,
+                                                        const AzPhysics::SceneHandle& joint_scene_owner)
+    {
+        auto* sceneInterface = AZ::Interface<AzPhysics::SceneInterface>::Get();
+        const auto joint = sceneInterface->GetJointFromHandle(joint_handle,joint_scene_owner);
```
moraaar, line 107: [standard] Needs a space after the comma in parameters (plus the name convention that Tom already mentioned)

`(joint_handle,joint_scene_owner)` -> `(jointHandle, jointSceneOwner)`

Gems/PhysX/Code/Source/HingeJointComponent.cpp
```diff
@@ -85,4 +88,60 @@ namespace PhysX
...
+        const auto type = joint->GetNativeType();
+        AZ_Assert(type == PhysX::NativeTypeIdentifiers::HingeJoint, "It is not PhysXHingeJoint");
+        physx::PxJoint *native_joint = static_cast<physx::PxJoint*>(joint->GetNativePointer());
+        physx::PxRevoluteJoint *native =native_joint->is<physx::PxRevoluteJoint>();
```
moraaar, line 111: [standard] * is part of the type so we put it beside the type, not the variable name.

Also missing an space after =

`physx::PxJoint* nativeJoint = static_cast...`
`physx::PxRevoluteJoint* native = nativeJoint...`

Gems/PhysX/Code/Source/HingeJointComponent.cpp
moraaar: [standard] space after the comma in parameters.`(m_jointSceneOwner, m_jointHandle)`

Same in other places. (I won't mention it again, please review the PR having this into consideration)

Gems/PhysX/Code/Source/HingeJointComponent.cpp
moraaar, line 131: [standard] space after the =

Same in other places. (I won't mention it again, please review the PR having this into consideration)

Gems/PhysX/Code/Source/Joint/Configuration/PhysXJointConfiguration.cpp
```diff
@@ -242,7 +251,21 @@ namespace PhysX
...
     }
+
+    void JointMotorProperties::Reflect(AZ::ReflectContext* context)
+    {
+        if (auto* serializeContext = azrtti_cast<AZ::SerializeContext*>(context))
+        {
+            serializeContext->Class<JointMotorProperties>()
+                    ->Version(1)
+                    ->Field("Enabled", &JointMotorProperties::m_enabled)
+                    ->Field("Gear ration", &JointMotorProperties::m_gearRatio)
```
moraaar, line 266: typo: "ration" -> "ratio"

Gems/PhysX/Code/Source/PrismaticJointComponent.cpp
```diff
@@ -73,6 +74,8 @@ namespace PhysX
 
         configuration.m_genericProperties = m_genericProperties;
         configuration.m_limitProperties = m_limits;
+        configuration.m_motorProperties = m_motor;
+        AZ_Printf("PrismaticJointComponent", "m_motor %d %f \n", m_motor.m_enabled, m_motor.m_forceLimit);
```
moraaar, line 78: Remember to remove the `AZ_Printf`.

Gems/PhysX/Code/Source/PrismaticJointComponent.cpp
```diff
@@ -84,4 +87,83 @@ namespace PhysX
             m_jointSceneOwner = leadFollowerInfo.m_followerBody->m_sceneOwner;
         }
     }
+
+    // AZ::Component
+    void PrismaticJointComponent::Activate() {
+        JointComponent::Activate();
+        PhysX::JointInterfaceRequestBus::Handler::BusConnect(this->GetEntityId());
```
moraaar, line 94: `this->` is not necessary.

Gems/PhysX/Code/Source/Joint/PhysXJointUtils.cpp
michalpelka: I agree, but I've tried to blend it into the codebase here - the same sanity check is other `Initialize****LimitProperties` functions. I think there should assertion in line 214. 
I can add it here and other sibling functions as well. What do you think @hultonha ?

Gems/PhysX/Code/Source/PrismaticJointComponent.cpp
```diff
@@ -84,4 +87,83 @@ namespace PhysX
...
+            physx::PxD6Joint *native_d6 = static_cast<physx::PxD6Joint*>(joint->GetNativePointer());
+            return native_d6;
+        }
+        return nullptr;
+    }
+
+    float PrismaticJointComponent::GetPosition(){
+        auto* native = GetPhysXD6Joint(m_jointSceneOwner,m_jointHandle);
+        if (native){
+            return native->getRelativeTransform().p.x;
```
moraaar, line 119: Why only X component of the position?

Gems/PhysX/Code/Source/PrismaticJointComponent.cpp
```diff
@@ -84,4 +87,83 @@ namespace PhysX
...
+        }
+        return 0.f;
+    };
+
+    float PrismaticJointComponent::GetVelocity() {
+        auto* native = GetPhysXD6Joint(m_jointSceneOwner,m_jointHandle);
+        if (native){
+            return native->getRelativeLinearVelocity().x;
```
moraaar, line 127: Why only X component of the velocity?

Gems/PhysX/Code/Source/PrismaticJointComponent.cpp
```diff
@@ -84,4 +87,83 @@ namespace PhysX
...
+            auto limits = native->getLinearLimit(physx::PxD6Axis::eX);
+            return AZStd::pair<float, float>(limits.lower,limits.upper);
+        }
+        return AZStd::pair<float, float> (0.f,0.f);
+    }
+
+    AZ::Transform PrismaticJointComponent::GetTransform(){
+        auto* native = GetPhysXD6Joint(m_jointSceneOwner,m_jointHandle);
+        if (native){
+            auto t = native->getRelativeTransform();
```
moraaar, line 145: `t` can be const.

Gems/PhysX/Code/Source/PrismaticJointComponent.cpp
```diff
@@ -84,4 +87,83 @@ namespace PhysX
...
+           native->setDriveVelocity({velocity,0,0},{0,0,0},true);
+        }
+    };
+
+    void PrismaticJointComponent::SetMaximumForce(float force)
+    {
+        auto* native = GetPhysXD6Joint(m_jointSceneOwner,m_jointHandle);
+        if (native){
+
+            physx::PxD6JointDrive drive (0,PX_MAX_F32,force,true);
```
moraaar, line 165: `drive` can be const.

Gems/PhysX/Code/Source/Joint/PhysXJointUtils.cpp
hultonha: Hmm good idea, if this really never should be null an assert is probably the way to go as then something else has gone horribly wrong somewhere and we shouldn't fail gracefully.

Gems/PhysX/Code/Source/PrismaticJointComponent.cpp
michalpelka: Added explanation to code.

Gems/PhysX/Code/Source/PrismaticJointComponent.cpp
greerdv: prismatic joints can only move along the x-axis

Gems/PhysX/Code/Include/PhysX/Joint/PhysXJointBus.h
greerdv: Yeah, I would avoid having Interface in the name because it might get confusing with AZ::Interface. If it's specifically for joint drives, it might be good to go with something like JointDriveRequests.

Gems/PhysX/Code/Include/PhysX/Joint/PhysXJointBus.h
michalpelka: Changed to address AZ::EntityComponentIdPair.

Gems/PhysX/Code/Source/HingeJointComponent.cpp
michalpelka: I've reworked this code, this function is a protected method now, without parameter. It getter to PhysX native joint type with cache.

Gems/PhysX/Code/Source/HingeJointComponent.cpp
michalpelka: I applied clang format here.

Gems/PhysX/Code/Source/HingeJointComponent.cpp
michalpelka: I've reworked this code, this function is a protected method now, without a parameter. It is getter to PhysX native joint type with cache.

Gems/PhysX/Code/Source/HingeJointComponent.cpp
michalpelka: applied standard

Gems/PhysX/Code/Source/HingeJointComponent.h
michalpelka: Change to recommended block comments.

Gems/PhysX/Code/Source/Joint/Configuration/PhysXJointConfiguration.cpp
michalpelka: adjusted to code standard.

Gems/PhysX/Code/Source/Joint/PhysXJoint.cpp
michalpelka: Removed.

Gems/PhysX/Code/Source/JointComponent.h
michalpelka: Done

Gems/PhysX/Code/physx_files.cmake
```diff
@@ -104,6 +104,7 @@ set(FILES
     Include/PhysX/Debug/PhysXDebugConfiguration.h
     Include/PhysX/Debug/PhysXDebugInterface.h
     Include/PhysX/Configuration/PhysXConfiguration.h
+    Include/PhysX/Joint/JointRequestsBus.h
```
moraaar, line 107: I would actually keep the filename as it was `PhysXJointBus.h` for 2 reasons:
- This header can potentially include more buses related to Joints in the future, it's quite common that a include is called `FeatureBus.h` and the inside it has `FeatureRequestsBus` and `FeatureNotificationBus`.
- All the other joint files have the PhysX prefix, so for consistency I'd keep it.

Gems/PhysX/Code/Source/PrismaticJointComponent.cpp
michalpelka: Done

Gems/PhysX/Code/Source/PrismaticJointComponent.h
michalpelka: Adjusted to block comments here.

Gems/PhysX/Code/Include/PhysX/Joint/PhysXJointBus.h
moraaar:
````
        //! It is a relative position of the entity in the direction of the free axis of the joint.
        //! For a hinge joint it is a twist angle (in radians), for the prismatic joint it is travel (in meters).
````

Awesome! thanks for the comment, it's very clear now.

File: Gems/PhysX/Code/Editor/EditorJointConfiguration.h (hunk 25 above)
Comment (michalpelka) [reply Gems/PhysX/Code/Editor/EditorJointConfiguration.h]: Ok.

File: Gems/PhysX/Code/Include/PhysX/Joint/PhysXJointBus.h (hunk 4 above)
Comment (moraaar) [reply Gems/PhysX/Code/Include/PhysX/Joint/PhysXJointBus.h]:
````
        //! It is a relative velocity of the entity in the direction of the free axis of the joint.
        //! For a hinge joint rotation velocity in radians per second. For prismatic joint it is linear velocity in
````
Awesome! thanks for the comment, it's very clear now.

File: Gems/PhysX/Code/Include/PhysX/Joint/PhysXJointBus.h (hunk 29 above)
Comment (moraaar) [reply Gems/PhysX/Code/Include/PhysX/Joint/PhysXJointBus.h]: Thanks for clarifying it in the comment!

File: Gems/PhysX/Code/Include/PhysX/Joint/PhysXJointBus.h (hunk 30 above)
Comment (moraaar) [reply Gems/PhysX/Code/Include/PhysX/Joint/PhysXJointBus.h]: Thanks for clarifying it in the comment!

File: Gems/PhysX/Code/Source/HingeJointComponent.cpp (hunk 10 above)
Comment (michalpelka) [reply Gems/PhysX/Code/Source/HingeJointComponent.cpp]: Adjusted order.

File: Gems/PhysX/Code/Source/PrismaticJointComponent.cpp (hunk 22 above)
Comment (michalpelka) [reply Gems/PhysX/Code/Source/PrismaticJointComponent.cpp]: Reworked this function and fixed formatting.

File: Gems/PhysX/Code/Source/PrismaticJointComponent.cpp (hunk 37 above)
Comment (moraaar) [reply Gems/PhysX/Code/Source/PrismaticJointComponent.cpp:129]:
````
            // Underlying PhysX joint is D6, but it simulates PhysXPrismatic joint.
            // The D6 joint has only X-axis unlocked, so report only X travel.
````

Thanks for the comment, it makes it super clear now!

File: Gems/PhysX/Code/Source/PrismaticJointComponent.cpp (hunk 38 above)
Comment (moraaar) [reply Gems/PhysX/Code/Source/PrismaticJointComponent.cpp:141]:
````
            // Undelying PhysX joint is D6, but it simulates PhysXPrismatic joint.
            // The D6 joint has only X-axis unlocked, so report only X velocity.
````

Thanks for the comment, it makes it super clear now!

File: Gems/PhysX/Code/Include/PhysX/Joint/JointRequestsBus.h [hunk 42]
```diff
@@ -0,0 +1,55 @@
...
+
+        //! Gets limits on active joint's axis.
+        virtual AZStd::pair<float, float> GetLimits() const = 0;
+
+        //! Sets maximum motor force.
+        //! @param force in Newtons (for prismatic joint) or Newton-meters (for hinge joint).
+        virtual void SetMaximumForce(float force) = 0;
+    };
+
+    using JointInterfaceRequestBus = AZ::EBus<JointRequests>;
```
moraaar, line 54: [Suggestion] `JointInterfaceRequestBus` -> `JointRequestBus`.

Gems/PhysX/Code/Source/HingeJointComponent.cpp
```diff
@@ -85,4 +88,69 @@ namespace PhysX
...
+        if (m_nativeJoint)
+        {
+            return m_nativeJoint;
+        }
+        auto* sceneInterface = AZ::Interface<AzPhysics::SceneInterface>::Get();
+        AZ_Assert(sceneInterface, "No sceneInterface");
+        const auto joint = sceneInterface->GetJointFromHandle(m_jointSceneOwner, m_jointHandle);
+        const auto type = joint->GetNativeType();
+        AZ_Assert(type == PhysX::NativeTypeIdentifiers::HingeJoint, "It is not PhysXHingeJoint");
+        physx::PxJoint* native_joint = static_cast<physx::PxJoint*>(joint->GetNativePointer());
```
moraaar, line 116: `native_joint` -> `nativeJoint`

Gems/PhysX/Code/Source/PrismaticJointComponent.cpp
```diff
@@ -84,4 +86,104 @@ namespace PhysX
...
+        {
+            return m_native;
+        }
+        auto* sceneInterface = AZ::Interface<AzPhysics::SceneInterface>::Get();
+        AZ_Assert(sceneInterface, "No sceneInterface");
+        const auto* joint =  sceneInterface->GetJointFromHandle(m_jointSceneOwner,m_jointHandle);
+        physx::PxJoint* native = static_cast<physx::PxJoint*>(joint->GetNativePointer());
+        if (native && native->is<physx::PxD6Joint>())
+        {
+            physx::PxD6Joint* native_d6 = static_cast<physx::PxD6Joint*>(joint->GetNativePointer());
```
moraaar, line 115: `native_d6` -> `nativeD6`

moraaar (approved):
Minor additional changes. But it's looking very good. Thanks for this contribution!

Gems/PhysX/Code/physx_files.cmake
michalpelka: Ok

Gems/PhysX/Code/Include/PhysX/Joint/PhysXJointBus.h
michalpelka: Set to "JointRequests "

Gems/PhysX/Code/Include/PhysX/Joint/Configuration/PhysXJointConfiguration.h
michalpelka: Removed EditorJointMotorConfig class.

Gems/PhysX/Code/Include/PhysX/Joint/PhysXJointBus.h
michalpelka: JointRequests will be.

Gems/PhysX/Code/Editor/EditorJointConfiguration.cpp
michalpelka: This function was removed.

Gems/PhysX/Code/Source/Joint/PhysXJointUtils.cpp
michalpelka: Added assert in existing code as well.

Gems/PhysX/Code/Source/HingeJointComponent.cpp
michalpelka: Done.

Gems/PhysX/Code/Editor/EditorJointConfiguration.cpp
michalpelka: Done, I also removed gearRatio - it is no supported by D6. 
We can reintroduce it.

michalpelka:
Thanks, @moraaar @hultonha for reviewing! Applied and submitted an updated version. I also adjusted https://github.com/RobotecAI/o3de-physics-test-scene/pull/2 to those changes.

michalpelka:
Rebased.

pawelbudziszewski:
It seems that limits of the prismatic joint have stopped working. 

https://user-images.githubusercontent.com/31925544/210001537-0c0bcad9-cb65-444c-a4cf-f6636bbe61e4.mp4

michalpelka:
Thanks, @pawelbudziszewski, good catch - I fixed it with the latest commit. 
Please test it against this branch - https://github.com/michalpelka/o3de/tree/add_physx_joint_control_old . It is rebased on top of the older o3de development.

Gems/PhysX/Code/Source/PrismaticJointComponent.cpp
```diff
@@ -84,4 +86,104 @@ namespace PhysX
...
+        return AZStd::make_pair(0.f, 0.f);
+    }
+
+    AZ::Transform PrismaticJointComponent::GetTransform() const
+    {
+        auto* native = GetPhysXD6Joint();
+        if (native)
+        {
+            const auto worldFromLocal = native->getRelativeTransform();
+            return AZ::Transform(
```
SergeyAMZN, line 163: See if `return PxMathConvert(worldFromLocal);` works here.
There are a bunch of utility conversion functions in Gems\PhysX\Code\Include\PhysX\MathConversion.h

Gems/PhysX/Code/Source/PrismaticJointComponent.h
```diff
@@ -8,6 +8,8 @@
 #pragma once
 
 #include <AzCore/Component/Component.h>
+#include <AzCore/Component/TickBus.h>
```
SergeyAMZN, line 11: Probably don't need to include TickBus.h?

SergeyAMZN (approved):
Looks good!

Gems/PhysX/Code/Source/PrismaticJointComponent.h
```diff
@@ -27,11 +30,34 @@ namespace PhysX
         PrismaticJointComponent(
             const JointComponentConfiguration& configuration,
             const JointGenericProperties& genericProperties,
-            const JointLimitProperties& limitProperties);
+            const JointLimitProperties& limitProperties,
+            const JointMotorProperties& motorProperties);
         ~PrismaticJointComponent() = default;
 
+        ///////////////////////////////////////////////////////////////////////////////////
```
hultonha, line 37: nit: ////////// lines can be removed

hultonha (approved):
LGTM! I just have few more minor bits of feedback but overall looks awesome, great work @michalpelka 😄

Gems/PhysX/Code/Source/Joint/PhysXJointUtils.cpp
```diff
@@ -379,6 +418,13 @@ namespace PhysX::Utils
                 configuration.m_genericProperties,
                 static_cast<physx::PxJoint*>(joint));
 
+            if (configuration.m_motorProperties.m_useMotor)
+            {
+                joint->setRevoluteJointFlag(physx::PxRevoluteJointFlag::eDRIVE_ENABLED, true);
+                joint->setDriveVelocity(0);
```
greerdv, line 424: nitpick - I would suggest writing 0.0f so the type is clear

Gems/PhysX/Code/Source/Joint/PhysXJointUtils.cpp
```diff
@@ -396,24 +442,48 @@ namespace PhysX::Utils
...
+                    PHYSX_SCENE_READ_LOCK(actorData.childActor->getScene());
+                    joint_d6 = physx::PxD6JointCreate(
+                        PxGetPhysics(),
+                        actorData.parentActor,
+                        PxMathConvert(parentLocalTM),
+                        actorData.childActor,
+                        PxMathConvert(childLocalTM));
+                }
+                InitializePrismaticLimitD6Properties(configuration.m_limitProperties, joint_d6);
+                physx::PxD6JointDrive drive(0, PX_MAX_F32, configuration.m_motorProperties.m_driveForceLimit, true);
```
greerdv, line 465: same nitpick about making the float type clear, here and line 467

Gems/PhysX/Code/Source/PrismaticJointComponent.cpp
```diff
@@ -84,4 +86,104 @@ namespace PhysX
...
+        }
+        return AZ::Transform();
+    };
+
+    void PrismaticJointComponent::SetVelocity(float velocity)
+    {
+        auto* native = GetPhysXD6Joint();
+        if (native)
+        {
+            native->setDriveVelocity({ velocity, 0, 0 }, { 0, 0, 0 }, true);
```
greerdv, line 176: same nitpick about making the float type clear (also line 185)

Gems/PhysX/Code/Include/PhysX/Joint/PhysXJointRequestsBus.h
```diff
@@ -0,0 +1,55 @@
...
+
+        //! Gets limits on active joint's axis.
+        virtual AZStd::pair<float, float> GetLimits() const = 0;
+
+        //! Sets maximum motor force.
+        //! @param force in Newtons (for prismatic joint) or Newton-meters (for hinge joint).
+        virtual void SetMaximumForce(float force) = 0;
+    };
+
+    using JointInterfaceRequestBus = AZ::EBus<JointRequests>;
```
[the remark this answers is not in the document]
greerdv: Yeah, we're (mostly) pretty consistent about bus naming, so I would definitely +1 this. I would also consider changing the naming to something a bit more specific like JointDriveRequests / JointDriveRequestBus.

greerdv (approved):
Looks good to me once the bus is renamed, everything else is just nitpicks. Great work, thanks for adding this improvement!

Gems/PhysX/Code/Include/PhysX/Joint/PhysXJointRequestsBus.h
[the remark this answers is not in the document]
michalpelka: Done.

Gems/PhysX/Code/Source/PrismaticJointComponent.cpp
```diff
@@ -84,4 +86,104 @@ namespace PhysX
...
+        JointComponent::Activate();
+    };
+
+    void PrismaticJointComponent::Deactivate()
+    {
+        PhysX::JointInterfaceRequestBus::Handler::BusDisconnect();
+        JointComponent::Deactivate();
+    };
+
+    physx::PxD6Joint* PrismaticJointComponent::GetPhysXD6Joint() const
```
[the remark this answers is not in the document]
michalpelka: Agreed, but it has some side effects-  GetPhysXD6Joint is not called from cont methods (Physx::JointRequest::GetVelocity). Instead, m_native pointer is set in `InitNativeJoint`

Gems/PhysX/Code/Source/HingeJointComponent.cpp
```diff
@@ -85,4 +88,72 @@ namespace PhysX
...
+    {
+        auto limit = GetPhysXNativeRevoluteJoint()->getLimit();
+        return AZStd::pair<float, float>(limit.lower, limit.upper);
+    }
+
+    AZ::Transform HingeJointComponent::GetTransform() const
+    {
+        const auto worldFromLocal = GetPhysXNativeRevoluteJoint()->getRelativeTransform();
+        return AZ::Transform(
+            AZ::Vector3{ worldFromLocal.p.x, worldFromLocal.p.y, worldFromLocal.p.z },
```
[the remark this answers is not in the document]
michalpelka: Totally agree. Using PxMathConvert now.

cgalvan (approved):

Gems/PhysX/Code/Source/PrismaticJointComponent.cpp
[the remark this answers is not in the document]
hultonha: Ah I see, so should we just remove the `const` here then?

Gems/PhysX/Code/Include/PhysX/Joint/PhysXJointRequestsBus.h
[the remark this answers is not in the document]
hultonha: Thanks!

Gems/PhysX/Code/Include/PhysX/Joint/PhysXJointRequestsBus.h
```diff
@@ -0,0 +1,55 @@
...
+namespace PhysX
+{
+    //! Interface to communicate with PhysX joint's motor.
+    class JointRequests : public AZ::EBusTraits
+    {
+    public:
+        static constexpr AZ::EBusAddressPolicy AddressPolicy = AZ::EBusAddressPolicy::ById;
+        using BusIdType = AZ::EntityComponentIdPair;
+
+        virtual ~JointRequests() = default;
```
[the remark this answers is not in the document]
hultonha: No problem! 🙂

Gems/PhysX/Code/Source/PrismaticJointComponent.h
hultonha: Cheers!

hultonha:
Looks great @michalpelka, I'll kick off a build now 👍

Issue #7572, opened 2022-02-11
Follow up on BoxManipulatorRequestBus deprecation


greerdv:
Remove the deprecated functions from Code\Framework\AzToolsFramework\AzToolsFramework\Manipulators\BoxManipulatorRequestBus.h after the required deprecation period has passed.

greerdv:
Done in #13790

PR #7463, opened 2022-02-07
Fixing various bugs with collider and box shape manipulators with non-uniform scale

greerdv:
Putting up a review for people to start looking at, but I still want to add a bunch more unit tests.

PR #13790, opened 2022-12-20
remove deprecated functions from BoxManipulatorRequestBus

greerdv:

## What does this PR do?
Removes deprecated functions from BoxManipulatorRequestBus (#7572). The functions were marked with O3DE_DEPRECATION_NOTICE in #7463, prior to the 2210 release. The bus has very light usage, so moving directly to deleting the functions rather than marking with AZ_DEPRECATED.

## How was this PR tested?
Ran LmbrCentral and PhysX tests and editor tests.


moraaar (approved):

hultonha (approved):
Good find! Thanks for updating this @greerdv 🙂

PR #13792, opened 2022-12-20
Fix RegressionTestForSpecificSegmentsAndTriangles math unit test on Android

moraaar:

## What does this PR do?

Disabling -ffast-math for 'IntersectSegment.cpp' file because its "Watertight Ray/Triangle Intersection" algorithm breaks with fast-math optimizations.

Change `RegressionTestForSpecificSegmentsAndTriangles` test to use google test parametrization so each case entry is reported as a test.

Remove `AZ_TRAIT_DISABLE_FAILED_MATH_TESTS` trait macro since that was the last test failing.

## How was this PR tested?

`RegressionTestForSpecificSegmentsAndTriangles` tests pass on Android:
````
Note: Google Test filter = MATH_IntersectSegmentTriangleTest*
[==========] Running 13 tests from 2 test cases.
[----------] 2 tests from MATH_IntersectSegmentTriangleTest
[ RUN      ] MATH_IntersectSegmentTriangleTest.SegmentThroughSphereCenterHasCorrectIntersectionCount
[       OK ] MATH_IntersectSegmentTriangleTest.SegmentThroughSphereCenterHasCorrectIntersectionCount (1423 ms)
[ RUN      ] MATH_IntersectSegmentTriangleTest.CompareSegmentTriangleIntersectionMethods
[       OK ] MATH_IntersectSegmentTriangleTest.CompareSegmentTriangleIntersectionMethods (1 ms)
[----------] 2 tests from MATH_IntersectSegmentTriangleTest (1424 ms total)

[----------] 11 tests from MATH_IntersectSegmentTriangleTest/RayTriangleTests
[ RUN      ] MATH_IntersectSegmentTriangleTest/RayTriangleTests.RegressionTestForSpecificSegmentsAndTriangles/0
[       OK ] MATH_IntersectSegmentTriangleTest/RayTriangleTests.RegressionTestForSpecificSegmentsAndTriangles/0 (0 ms)
[ RUN      ] MATH_IntersectSegmentTriangleTest/RayTriangleTests.RegressionTestForSpecificSegmentsAndTriangles/1
[       OK ] MATH_IntersectSegmentTriangleTest/RayTriangleTests.RegressionTestForSpecificSegmentsAndTriangles/1 (0 ms)
[ RUN      ] MATH_IntersectSegmentTriangleTest/RayTriangleTests.RegressionTestForSpecificSegmentsAndTriangles/2
[       OK ] MATH_IntersectSegmentTriangleTest/RayTriangleTests.RegressionTestForSpecificSegmentsAndTriangles/2 (0 ms)
[ RUN      ] MATH_IntersectSegmentTriangleTest/RayTriangleTests.RegressionTestForSpecificSegmentsAndTriangles/3
[       OK ] MATH_IntersectSegmentTriangleTest/RayTriangleTests.RegressionTestForSpecificSegmentsAndTriangles/3 (0 ms)
[ RUN      ] MATH_IntersectSegmentTriangleTest/RayTriangleTests.RegressionTestForSpecificSegmentsAndTriangles/4
[       OK ] MATH_IntersectSegmentTriangleTest/RayTriangleTests.RegressionTestForSpecificSegmentsAndTriangles/4 (0 ms)
[ RUN      ] MATH_IntersectSegmentTriangleTest/RayTriangleTests.RegressionTestForSpecificSegmentsAndTriangles/5
[       OK ] MATH_IntersectSegmentTriangleTest/RayTriangleTests.RegressionTestForSpecificSegmentsAndTriangles/5 (0 ms)
[ RUN      ] MATH_IntersectSegmentTriangleTest/RayTriangleTests.RegressionTestForSpecificSegmentsAndTriangles/6
[       OK ] MATH_IntersectSegmentTriangleTest/RayTriangleTests.RegressionTestForSpecificSegmentsAndTriangles/6 (0 ms)
[ RUN      ] MATH_IntersectSegmentTriangleTest/RayTriangleTests.RegressionTestForSpecificSegmentsAndTriangles/7
[       OK ] MATH_IntersectSegmentTriangleTest/RayTriangleTests.RegressionTestForSpecificSegmentsAndTriangles/7 (0 ms)
[ RUN      ] MATH_IntersectSegmentTriangleTest/RayTriangleTests.RegressionTestForSpecificSegmentsAndTriangles/8
[       OK ] MATH_IntersectSegmentTriangleTest/RayTriangleTests.RegressionTestForSpecificSegmentsAndTriangles/8 (0 ms)
[ RUN      ] MATH_IntersectSegmentTriangleTest/RayTriangleTests.RegressionTestForSpecificSegmentsAndTriangles/9
[       OK ] MATH_IntersectSegmentTriangleTest/RayTriangleTests.RegressionTestForSpecificSegmentsAndTriangles/9 (0 ms)
[ RUN      ] MATH_IntersectSegmentTriangleTest/RayTriangleTests.RegressionTestForSpecificSegmentsAndTriangles/10
[       OK ] MATH_IntersectSegmentTriangleTest/RayTriangleTests.RegressionTestForSpecificSegmentsAndTriangles/10 (0 ms)
[----------
] 11 tests from MATH_IntersectSegmentTriangleTest/RayTriangleTests (0 ms total)

[==========] 13 tests from 2 test cases ran. (1424 ms total)
[  PASSED  ] 13 tests.
````

### CODE REVIEW

File: Code/Framework/AzCore/Platform/Android/platform_android_files.cmake [hunk 1]
```diff
@@ -88,3 +88,12 @@ if (LY_TEST_PROJECT)
         VALUES LY_NO_ASSETS
     )
 endif()
+
+# Disabling -ffast-math for 'IntersectSegment.cpp' file only.
+# It has an implementation of the "Watertight Ray/Triangle Intersection"
+# algorithm, which gets broken by the fast-math optimizations.
+ly_add_source_properties(
+    SOURCES ${CMAKE_CURRENT_LIST_DIR}/../../AzCore/Math/IntersectSegment.cpp
```
hultonha, line 96: suggestion: Rather than using a relative path with (../../) could you instead use `CMAKE_SOURCE_DIR` and specify the path from the root of the repo (maybe @lumberyard-employee-dm could confirm if that's the best CMake variable to use).

Code/Framework/AzCore/Tests/Math/IntersectionTests.cpp
```diff
@@ -496,109 +496,109 @@ namespace UnitTest
...
+        // This failure came from a vegetation system raycast to a triangle in a sphere model.
         {
-            AZ::Vector3 m_rayStart;
-            AZ::Vector3 m_rayEnd;
-            AZ::Vector3 m_triVerts[3];
-            bool m_shouldHit;
-            float m_hitDistance;
-            AZ::Vector3 m_hitNormal;
-        };
+            { 0.0, 0.16, 1.01948643 }, { 0.0, 0.16, -0.0194873810 }, // segment
```
hultonha, line 513: nit: While you're here could you add `f` to the floating point params to avoid the implicit double conversions please

hultonha (approved):
LGTM I just have a couple minor suggestions

Code/Framework/AzCore/Tests/Math/IntersectionTests.cpp
moraaar: I'm `99.9e+36`% sure that's not going to break anything. But the last couple of days I've been fighting with precision issues witnessing very minor changes making these tests fail, so I'm going to play it safe and not change that for now.

Code/Framework/AzCore/Tests/Math/IntersectionTests.cpp
```diff
@@ -496,109 +496,109 @@ namespace UnitTest
...
         {
-            // This failure came from a vegetation system raycast to a triangle in a sphere model.
-            {
-                { 0.0, 0.16, 1.01948643 }, { 0.0, 0.16, -0.0194873810 }, // segment
-                { { 0.0294727981, 0.18608, 0.9605 }, { 0.0, 0.18842, 0.9605 }, { 0.0, 0.15215, 0.9741 } }, // triangle
-                true, 0.0465169623, { 0.0278645027, 0.350958735, 0.935976326 }  // expected results
-            },
+            { 962.503540, -788.401978, -320.390747 }, { -838.229004, 555.508301, 748.038086 }, // segment
+            { { -353.655212, 579.209229, -827.134216 }, { 358.185913, 346.845093, 729.328979 }, { 222.018555, -358.014160, -4.521179 } }, // triangle
+            true, 0.399751157, { 0.899101138, 0.220195562, -0.378326714 } // expected results
```
Kadino, line 523: aside: Packing expected results into a test fixture for a test can turn into a maintenance anti-pattern. Doesn't look like it's currently creating a problem, but be wary about the value of reuse versus any penalties to simplicity and readability.

Kadino (approved):
Looks good from a testing standpoint. I generally don't trust fastmath, but SIG-Core may have other opinions as far as where speed vs. accuracy is concerned.

Code/Framework/AzCore/Tests/Math/IntersectionTests.cpp
mbalfour-amzn: Since I'm the one who wrote these tests, I'll justify them. :) 

Our previous line/triangle intersection algorithms had precision issues where they could very occasionally fail with tiny triangles, rays that intersect edges or vertices, etc. These cases were extremely hard to catch, and I'm not sure how to create simple reproductions. To verify that the new algorithm didn't have these problems, I introduced a few sets of tests in here. One test fires a set of deterministic rays through the center of a highly-tesselated icosphere and verifies that there are always exactly 2 hits. One test runs all three line/triangle algorithms across a small set of deterministically-random rays and triangles and verifies that all three algorithms match in their results. Finally, this test takes a specific set of test cases that were discovered to fail with the previous algorithms and verifies them against the new algorithm to ensure that they succeed. This particular case was generated from running the three-algorithm-comparison test 10 million times and capturing some of the failing results that show up when you run that many combinations. (Other cases in this test were caught from failed Editor raycasts and failed vegetation system raycasts that occurred "in the wild".)

I would argue that this specific test has just proven its worth, because it did exactly what it's supposed to do and caught failures introduced into the system via the fast-math flag. By having a successful positive failure and no false failures yet, it feels pretty worthwhile to keep around and makes me super happy that I added it in the first place. :) 

I also 100% recognize that when it _does_ fail, there's no good way of debugging what's going on because of the mysteriousness of the numbers. If you can find a way to improve the understandability here without harming the reliability of the test, I'm all for it!

Code/Framework/AzCore/Tests/Math/IntersectionTests.cpp
mbalfour-amzn: FWIW, there is no true implicit double conversion here. The "conversion" takes place at compile time, the compiler produces the same results whether or not the f is there (see godbolt to verify). I'm not opposed to someone adding the f, but it seems like purely a style choice. 

If we were going to modify these constants, my preference would be that we go a step further and actually write out the full exact values that are stored in the float instead of these lightly-rounded values, so that way anyone who wants to try doing the math by hand would be starting from the exact same numbers that the computer is using. That feels pretty far outside the range of this PR though.

mbalfour-amzn (approved):

Code/Framework/AzCore/Tests/Math/IntersectionTests.cpp
Kadino: Definitely agree that the caught failure has proved useful! As far as making the tests more linear, it would probably involve deconvoluting the different expectations into more-specific tests. Since they're all segment-to-triangle tests, it would probably need to be related to the edge case (results) it is trying to verify. Then each specific test could more quickly inform what failed.

Since I can't think of a better way to classify triangle data or edge cases, another thing that could help is to pack a string descriptor of each case along with the data. This can add context via `EXPECT_EQ(test.m_shouldHit, result) << test.m_caseDescriptor;` to ramp up whoever comes in to debug it. It would also document why each set of data contains what it does.

kberg-amzn: I definitely like taking actual failure cases from actual production situations, and encoding them as tests. Especially for mathematical operations such as this. That said, it is a lot of magic numbers so lots of comments are encouraged.

kberg-amzn (approved):

PR #13799, opened 2022-12-21
Fix PhysX unit tests in android

moraaar:
## What does this PR do?

- Reduced max number of scenes from 256 to 64 because for each scene PhysX SDK calls `pthread_key_create` and the max number of keys on android (`PTHREAD_KEYS_MAX`) is 128. 64 should be enough on any platform.
- Moving ragdoll configuration xml data to a header file since data files are not available in android unit tests.
- Missed read lock in `RigidBody::SetKinematic` function
- Avoid resolving contacts between two kinematic objects, which is invalid.

## How was this PR tested?

PhysX unit tests pass on Android (debug and profile).
PhysX unit tests pass on Windows (debug and profile).

````
[==========] 607 tests from 26 test cases ran. (37337 ms total)
[  PASSED  ] 607 tests.
````

### CODE REVIEW

File: Gems/PhysX/Code/Source/Collision.cpp [hunk 1]
```diff
@@ -46,6 +46,12 @@ namespace PhysX
                 physx::PxPairFlag::eNOTIFY_TOUCH_LOST |
                 physx::PxPairFlag::eNOTIFY_CONTACT_POINTS;
 
+            // resolving contacts between two kinematic objects is invalid
+            if (physx::PxFilterObjectIsKinematic(attributes0) && physx::PxFilterObjectIsKinematic(attributes1))
+            {
+                pairFlags &= ~physx::PxPairFlag::eSOLVE_CONTACT;
```
SergeyAMZN, line 52: Nice, I had the same change done locally!
`pairFlags.clear(physx::PxPairFlag::eSOLVE_CONTACT);` would be a bit more readable 

Could you also add the same code to `DefaultFilterShaderCCD` too?

moraaar: Cool! Thanks! Will do.

moraaar: Done!

Code/Framework/AzFramework/AzFramework/Physics/Common/PhysicsTypes.h
```diff
@@ -24,11 +24,12 @@ namespace Physics
 
 namespace AzPhysics
 {
-    //! Default Scene Names and Crc32
-    static constexpr const char* DefaultPhysicsSceneName = "DefaultScene";
-    static constexpr const AZ::Crc32 DefaultPhysicsSceneId = AZ_CRC_CE(DefaultPhysicsSceneName);
-    static constexpr const char* EditorPhysicsSceneName = "EditorScene";
-    static constexpr const AZ::Crc32 EditorPhysicsSceneId = AZ_CRC_CE(EditorPhysicsSceneName);
+    //! Maximum number of scenes, Default Scene Names and Crc32
```
greerdv, line 27: I think this doxygen markup will only be applied to the first line below. If you wrap the lines in //! @{ and //! @} I think the markup will apply to all 5 lines.

Code/Framework/AzFramework/AzFramework/Physics/Common/PhysicsTypes.h
```diff
@@ -24,11 +24,12 @@ namespace Physics
...
+    constexpr AZ::u32 MaxNumberOfScenes = 64;
```
greerdv, line 28: Is it worth adding a comment explaining the limitation on Android which limits this?

Gems/PhysX/Code/Source/Shape.cpp
```diff
@@ -354,8 +354,12 @@ namespace PhysX
         const physx::PxHitFlags hitFlags = SceneQueryHelpers::GetPxHitFlags(worldSpaceRequest.m_hitFlags);
 
         physx::PxRaycastHit hitInfo;
-        const bool hit = physx::PxGeometryQuery::raycast(start, unitDir, m_pxShape->getGeometry().any(), pose,
-                                                         worldSpaceRequest.m_distance, hitFlags, maxHits, &hitInfo);
+        bool hit;
```
greerdv, line 357: I bet hultonha would suggest an immediately invoked lambda here so you can make hit const.

greerdv (approved):
Nice!

SergeyAMZN (approved):

Code/Framework/AzFramework/AzFramework/Physics/Common/PhysicsTypes.h
moraaar: Ahh, you're right. Not worth another AR though. I'll do the change in the future.

Code/Framework/AzFramework/AzFramework/Physics/Common/PhysicsTypes.h
moraaar: I don't think so, very platform specific...

Gems/PhysX/Code/Source/Shape.cpp
moraaar: hahaha, I want to keep this one simple because I have to resolve this file now in the PhysX5Support branch.

hultonha: 😂 I would have said that, there's also the `AZStd::invoke` pattern which can be used to make the callsite clearer if you prefer, but this is okay for now (might want to initialize `bool hit` to a known value first though)

hultonha (approved):
Great to see! 😄

PR #13802, opened 2022-12-21
Add support to optionally enable PhysX 5.1 in PhysX gem

SergeyAMZN:
## What does this PR do?

Add support to optionally enable PhysX 5.1 in PhysX gem via `-DAZ_USE_PHYSX5=ON` passed into CMake project generation command. 

RFC: https://github.com/o3de/sig-simulation/issues/31

## How was this PR tested?

QA pass, unit tests.


Gems/PhysX/Code/Source/Common/PhysXSceneQueryHelpers.h
```diff
@@ -11,6 +11,7 @@
 #include<AzFramework/Physics/Common/PhysicsSceneQueries.h>
 
 //physx sdk includes
+#include <PxPhysicsVersion.h>
```
moraaar, line 14: This include, which I guess defines `PX_PHYSICS_VERSION_MAJOR` is only included here in `PhysXSceneQueryHelpers.h`, but the macro is used in many different cpps. Is it gurantee that all the cpps using the macro reaches `PhysXSceneQueryHelpers.h`? otherwise for some cpps `#if (PX_PHYSICS_VERSION_MAJOR == 5)` might be false and others true.

Maybe `#include <PxPhysicsVersion.h>` could be put in some basic header inside PhysX Gem to guarantee that?

cmake/3rdParty/Platform/Android/BuiltInPackages_android.cmake
```diff
@@ -32,3 +31,9 @@ ly_associate_package(PACKAGE_NAME OpenSSL-1.1.1o-rev2-android        TARGETS Ope
 ly_associate_package(PACKAGE_NAME zlib-1.2.11-rev5-android           TARGETS ZLIB            PACKAGE_HASH 73c9e88892c237a3fc6eafc04268ccd9d479e6d55f9df2ed58b236c8f9cf2cae)
 ly_associate_package(PACKAGE_NAME lz4-1.9.3-vcpkg-rev4-android       TARGETS lz4             PACKAGE_HASH f5b22642d218dbbb442cae61e469e5b241c4740acd258c3e8678e60dec61ea93)
 
+set(AZ_USE_PHYSX5 OFF CACHE BOOL "When ON PhysX Gem will use PhysX 5 SDK")
```
moraaar, line 34: Since Android uses the script `generate_android_project.py` on top of CMake, currently there is no way to pass `-DAZ_USE_PHYSX5=ON` option to enable PhysX 5 on Android. We have to add an input parameter to `generate_android_project.py` and modify the scripts to be able to use PhysX 5 on Android.

Gems/PhysX/Code/Source/Common/PhysXSceneQueryHelpers.h
SergeyAMZN: Good idea to try a non-unity build, all working fine with no issues.

There's no single basic header that is included everywhere (e.g. a precompiled header) but any code actually using PhysX SDK will have `PxPhysicsAPI.h` included which does include `PxPhysicsVersion.h` internally.

lawsonamzn (approved):

Code/Framework/AzFramework/AzFramework/Physics/Common/PhysicsSceneQueries.h
```diff
@@ -71,12 +71,12 @@ namespace AzPhysics
             //! Performance hint flag for ShapeCasts when it is known upfront there's no initial overlap.
             //! NOTE: using this flag may cause undefined results if shapes are initially overlapping.
             AssumeNoInitialOverlap = (1 << 4),
-            MeshMultiple = (1 << 5), //!< Report all hits for meshes rather than just the first. Not applicable to ShapeCast queries.
+            AnyHit = (1 << 5), //!< Report any first hit. Used for geometries that contain more than one primitive. For meshes,
+                               //!< if neither MeshMultiple nor AnyHit is specified, a single closest hit will be reported.
+            MeshMultiple = (1 << 6), //!< Report all hits for meshes rather than just the first. Not applicable to ShapeCast queries.
```
hultonha, line 76: question: As `MeshMultiple` has changed from `(1 << 5)` to `(1 << 6)`, is there any chance this could impact or break any serialized data? Do we need to think of any kind of upgrade/conversion path or is this relatively safe you think?

Code/Framework/AzFramework/AzFramework/Physics/Common/PhysicsSceneQueries.h
```diff
@@ -71,12 +71,12 @@ namespace AzPhysics
...
             //! Report any first hit for meshes. If neither MeshMultiple nor MeshAny is specified,
             //! a single closest hit will be reported for meshes.
-            MeshAny = (1 << 6),
-            //! Report hits with back faces of mesh triangles. Also report hits for raycast
-            //! originating on mesh surface and facing away from the surface normal. Not applicable to ShapeCast queries.
+            MeshAny = AnyHit, //!< @deprecated Deprecated, please use AnyHit instead.
```
hultonha, line 79: update: Could you please create a deprecation issue to track this (there's a template here - https://github.com/o3de/o3de/issues/new?assignees=&labels=kind%2Fdeprecation&template=deprecation_template.md&title=Deprecation+Notice). Then add the GHI here and we can then actually delete this after the next release goes out.

Gems/PhysX/Code/Source/BaseColliderComponent.cpp
```diff
@@ -64,13 +64,20 @@ namespace PhysX
 
             auto pxShape = static_cast<physx::PxShape*>(shapes[0]->GetNativePointer());
             physx::PxTransform pxWorldTransform = PxMathConvert(m_worldTransform);
-            physx::PxBounds3 bounds = physx::PxGeometryQuery::getWorldBounds(pxShape->getGeometry().any(),
+
+#if (PX_PHYSICS_VERSION_MAJOR == 5)
+            const physx::PxGeometry& pxShapeGeom = pxShape->getGeometry();
+#else
+            const physx::PxGeometry& pxShapeGeom = pxShape->getGeometry().any();
+#endif
```
hultonha, line 72: nit: I believe we generally like to include closing comments for defines (so this would have `#endif // PX_PHYSICS_VERSION_MAJOR == 5`)

Gems/PhysX/Code/Source/Common/PhysXSceneQueryHelpers.cpp
```diff
@@ -128,20 +149,30 @@ namespace PhysX
...
+            physx::PxHitFlags pxHitFlags; // PxFlags is initialized to 0 by default
+            const auto ConvertToPxFlags = [&pxHitFlags, hitFlags](AzPhysics::SceneQuery::HitFlags azHitFlagCheck, physx::PxHitFlag::Enum pxHitFlagSet)
+            {
+                if (azHitFlagCheck == (hitFlags & azHitFlagCheck))
+                {
+                    pxHitFlags |= pxHitFlagSet;
+                }
+            };
+
+            ConvertToPxFlags(AzPhysics::SceneQuery::HitFlags::Position, physx::PxHitFlag::ePOSITION);
```
hultonha, line 161: Nice improvement! 😄

Gems/PhysX/Code/Source/Common/PhysXSceneQueryHelpers.cpp
```diff
@@ -278,12 +314,25 @@ namespace PhysX
             return physx::PxQueryHitType::eNONE;
         }
 
+        // @deprecated Unused, we're only pre-filtering at this time
```
hultonha, line 317: update: Same comment before about the deprecation notice (could be the same one for all related code to make things easier to track)

Gems/PhysX/Code/Source/Pipeline/MeshExporter.cpp
```diff
@@ -136,7 +136,8 @@ namespace PhysX
             AZ::SerializeContext* serializeContext = azrtti_cast<AZ::SerializeContext*>(context);
             if (serializeContext)
             {
-                serializeContext->Class<MeshExporter, AZ::SceneAPI::SceneCore::ExportingComponent>()->Version(5);
+                serializeContext->Class<MeshExporter, AZ::SceneAPI::SceneCore::ExportingComponent>()
```
hultonha, line 139: question: Should this only shift by 1 or 2 depending on if it's defined or not so we don't jump a large number of versions, won't the version number now be 37 (5 + 32)?

cmake/3rdParty/Platform/Mac/BuiltInPackages_mac.cmake
```diff
@@ -47,3 +46,10 @@ ly_associate_package(PACKAGE_NAME lz4-1.9.3-vcpkg-rev4-mac
 ly_associate_package(PACKAGE_NAME azslc-1.8.9-rev1-mac                              TARGETS azslc                       PACKAGE_HASH ccadb9bb90db18cfb31d416f1da2a9429266f22c7375fd0acf5aed299d68a7fb)
 ly_associate_package(PACKAGE_NAME SQLite-3.37.2-rev2-mac                            TARGETS SQLite                      PACKAGE_HASH b7d9abdb68045003e030e1a9a805db1aefa5e8fde6dccfbb4fab3a06249a41fc)
 ly_associate_package(PACKAGE_NAME AwsIotDeviceSdkCpp-1.15.2-rev2-mac                TARGETS AwsIotDeviceSdkCpp          PACKAGE_HASH 4854edb7b88fa6437b4e69e87d0ee111a25313ac2a2db5bb2f8b674ba0974f95)
+
+set(AZ_USE_PHYSX5 OFF CACHE BOOL "When ON PhysX Gem will use PhysX 5 SDK")
```
hultonha, line 50: nit/minor: Feel free to ignore but CMake has the `option` command (instead of set) to make this sort of thing simpler (see https://cmake.org/cmake/help/latest/command/option.html)

hultonha (commented):
Overall looks good but I have a couple of questions before adding an approval. Great progress!

Gems/PhysX/Code/Source/BaseColliderComponent.cpp
lumberyard-employee-dm: This is not needed for simple cases where the `#endif` is close to the `#if`: https://github.com/o3de/sig-core/blob/main/governance/Coding-Standards-and-Style-Guide.md#comments-after-preprocessor-statements

cmake/3rdParty/Platform/iOS/BuiltInPackages_ios.cmake
```diff
@@ -31,3 +30,10 @@ ly_associate_package(PACKAGE_NAME libsamplerate-0.2.1-rev2-ios   TARGETS libsamp
 ly_associate_package(PACKAGE_NAME OpenSSL-1.1.1o-rev1-ios        TARGETS OpenSSL         PACKAGE_HASH 1325bbb2dcc97fe33d8db71e4b0a61e95e9e77558e3ce6767a1b5d330240f78f)
 ly_associate_package(PACKAGE_NAME zlib-1.2.11-rev5-ios           TARGETS ZLIB            PACKAGE_HASH c7f10b4d0fe63192054d926f53b08e852cdf472bc2b18e2f7be5aecac1869f7f)
 ly_associate_package(PACKAGE_NAME lz4-1.9.3-vcpkg-rev4-ios       TARGETS lz4             PACKAGE_HASH 588ea05739caa9231a9a17a1e8cf64c5b9a265e16528bc05420af7e2534e86c1)
+
+set(AZ_USE_PHYSX5 OFF CACHE BOOL "When ON PhysX Gem will use PhysX 5 SDK")
+if(AZ_USE_PHYSX5)
+    ly_associate_package(PACKAGE_NAME PhysX-5.1.1-rev1-ios           TARGETS PhysX       PACKAGE_HASH 3495e0bc1404a369fb6f4880ffc525c0a9d860867f5c524b33887cfd08f5c376)
+else()
+    ly_associate_package(PACKAGE_NAME PhysX-4.1.2.29882248-rev5-ios  TARGETS PhysX       PACKAGE_HASH 4a5e38b385837248590018eb133444b4e440190414e6756191200a10c8fa5615)
+endif()
```
lumberyard-employee-dm, line 39: These ly_associate_package calls can be moved to the PhysX gem itself, so that if the PhysX gem isn't part of the project, it can skip downloading the libraries.

Realistically though, my suggested change will not make a difference since the PhysX gem is part of the engine via the `engine.json`, but if the gem was ever not associated with the engine, but instead a project, then it would become applicable

lumberyard-employee-dm (commented):
How long are we planning to support PhysX 4 after PhysX 5 support added?

Because all the split `#if (PX_PHYSICS_VERSION_MAJOR == 5)` defines should go away when PhysX 4 support is dropped.

I think it would be good to perhaps add an issue for the SIG-Simulation group to track how long PhysX 4 support would last

cmake/3rdParty/Platform/Android/BuiltInPackages_android.cmake
lumberyard-employee-dm: There is actually a way to pass input parameters to CMake via `generate_android_project.py`
The [--extra-cmake-configure-args](https://github.com/o3de/o3de/blob/development/cmake/Tools/Platform/Android/generate_android_project.py#L242-L244) argument can be used to supply arguments to the CMake.

hultonha:
> How long are we planning to support PhysX 4 after PhysX 5 support added?
> 
> Because all the split `#if (PX_PHYSICS_VERSION_MAJOR == 5)` defines should go away when PhysX 4 support is dropped.
> 
> I think it would be good to perhaps add an issue for the SIG-Simulation group to track how long PhysX 4 support would last

+1 a tracking issue for PhysX 4 deprecation would be good... I don't think we have a firm ETA at this time but we want to ensure performance is still solid across mobile platforms that we're less sure of with PhysX 5. If all concerns are addressed there, then it's not unreasonable to assume PhysX 4 would be removed in development after the next release, but we're not committing to that at this stage, so it may be around longer.

Code/Framework/AzFramework/AzFramework/Physics/Common/PhysicsSceneQueries.h
SergeyAMZN: These are not serialized anywhere. 
Obviously technically external code could keep the hit flags saved somewhere in the database or file, but that'd be outside of our reflection system.

cmake/3rdParty/Platform/iOS/BuiltInPackages_ios.cmake
SergeyAMZN: That's a good idea, when it comes to supporting multiple default physics backends we won't have to have PhysX enabled. 
At the moment we can have more than one physics engine simulating, but the actual API backend is always PhysX.

cmake/3rdParty/Platform/Mac/BuiltInPackages_mac.cmake
SergeyAMZN: nice, looks like `option` would work too! 
I couldn't find any single use of `option()` in our cmake files though, so probably best to keep `set` for consistency?

lumberyard-employee-dm:
> > How long are we planning to support PhysX 4 after PhysX 5 support added?
> > Because all the split `#if (PX_PHYSICS_VERSION_MAJOR == 5)` defines should go away when PhysX 4 support is dropped.
> > I think it would be good to perhaps add an issue for the SIG-Simulation group to track how long PhysX 4 support would last
> 
> +1 a tracking issue for PhysX 4 deprecation would be good... I don't think we have a firm ETA at this time but we want to ensure performance is still solid across mobile platforms that we're less sure of with PhysX 5. If all concerns are addressed there, then it's not unreasonable to assume PhysX 4 would be removed in development after the next release, but we're not committing to that at this stage, so it may be around longer.

Makes sense that we don't have a firm estimated time of removal at this point. But making sure we have an issue that documents an intent for removing PhysX 4 at the very least allows O3DE to indicate that it will not support PhysX 4 in perpetuity.

Code/Framework/AzFramework/AzFramework/Physics/Common/PhysicsSceneQueries.h
hultonha: Okay understood thanks for clarifying 👍

Gems/PhysX/Code/Source/Pipeline/MeshExporter.cpp
hultonha: Okay perfect, makes sense, thanks for the explanation

cmake/3rdParty/Platform/Mac/BuiltInPackages_mac.cmake
hultonha: Ah yeah you could be right, no problem 👍

hultonha (approved):

lumberyard-employee-dm (approved):

Gems/PhysX/Code/Source/Common/PhysXSceneQueryHelpers.cpp
SergeyAMZN: Removed this comment since we can't actually delete/deprecate this function until the base class function is removed from PhysX SDK

SergeyAMZN:
Deprecation issue for SIG-SIMULATION: https://github.com/o3de/o3de/issues/13897

hultonha:
> Deprecation issue for SIG-SIMULATION: #13897

Awesome work @SergeyAMZN! Thanks very much 🙂

Gems/PhysX/Code/Tests/PhysXForceRegionTest.cpp
```diff
@@ -317,15 +317,15 @@ namespace PhysX
     TEST_F(PhysXForceRegionTest, ForceRegion_SimpleDragForce_EntityVelocitySpecificValue)
     {
         AZ::Vector3 entityVelocity = TestForceVolume<BoxColliderComponent>(GetTestSceneHandle(), SimpleDragForce);
-        EXPECT_GT(entityVelocity.GetZ(), -12.65f); // Falling velocity should be slower than free fall velocity, which is -12.65.
+        EXPECT_TRUE(entityVelocity.GetZ() > -12.66f); // Falling velocity should be slower than free fall velocity, which is -12.65 (+-0.01).
```
lumberyard-employee-dm, line 320: This is wrong, we should go from an EXPECT_GT to an EXPECT_TRUE when comparing to floats.

Before when this test failed it would output the `entityVelocity.GetZ()` value.

Now when it fails it just outputs that the expression "entityVelocity.GetZ() > -12.66f" has failed without being able to resolve the variable values.

```suggestion
        EXPECT_GT(entityVelocity.GetZ(), -12.66f); // Falling velocity should be slower than free fall velocity, which is -12.65 (+-0.01).
```

lawsonamzn: You can also add any amount of additional context you want since the EXPECT_xxxx macros return a stream that will be printed to the console, ie
```
EXPECT_xxxx(a, b) << "The acutal value of blah is " << blah; 
```

SergeyAMZN: nice spot! 
I actually made them to be EXPECT_GT/LT exactly for that reason, unit test was failing but no visibility what values were: https://github.com/o3de/o3de/pull/13802/commits/a84ea6aa584a829d5edfd633043d49fca02a701e

Must be switching between branches locally caused the bad merge. 

Thanks!

lumberyard-employee-dm (approved):

Issue #13720, opened 2022-12-15
Intermittent failures from pytest during xml creation


Kadino:
**Describe the bug**
Intermittent failures are continuing to emit from Pytest due to this unsafe call https://github.com/pytest-dev/pytest/blob/7.2.x/src/_pytest/junitxml.py#L649

**Failed Jenkins Job Information:**
https://jenkins.build.o3de.org/blue/organizations/jenkins/O3DE/detail/development/3649/pipeline/899

```
  File "D:\workspace\o3de\python\runtime\python-3.10.5-rev1-windows\python\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "D:\workspace\o3de\python\runtime\python-3.10.5-rev1-windows\python\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "D:\workspace\o3de\python\runtime\python-3.10.5-rev1-windows\python\lib\site-packages\pytest\__main__.py", line 5, in <module>
    raise SystemExit(pytest.console_main())
  File "D:\workspace\o3de\python\runtime\python-3.10.5-rev1-windows\python\lib\site-packages\_pytest\config\__init__.py", line 185, in console_main
    code = main()
  File "D:\workspace\o3de\python\runtime\python-3.10.5-rev1-windows\python\lib\site-packages\_pytest\config\__init__.py", line 162, in main
    ret: Union[ExitCode, int] = config.hook.pytest_cmdline_main(
  File "D:\workspace\o3de\python\runtime\python-3.10.5-rev1-windows\python\lib\site-packages\pluggy\hooks.py", line 286, in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
  File "D:\workspace\o3de\python\runtime\python-3.10.5-rev1-windows\python\lib\site-packages\pluggy\manager.py", line 93, in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
  File "D:\workspace\o3de\python\runtime\python-3.10.5-rev1-windows\python\lib\site-packages\pluggy\manager.py", line 84, in <lambda>
    self._inner_hookexec = lambda hook, methods, kwargs: hook.multicall(
  File "D:\workspace\o3de\python\runtime\python-3.10.5-rev1-windows\python\lib\site-packages\pluggy\callers.py", line 208, in _multicall
    return outcome.get_result()
[...]

PR #13809, opened 2022-12-21
Monkeypatch pytest xml file race condition

Kadino:
Signed-off-by: sweeneys <sweeneys@amazon.com>

## What does this PR do?

Attempts a new workaround to resolve a filesystem race condition in pytest https://github.com/o3de/o3de/issues/13720 . The race condition most frequently appears when running multiple instances of pytest in parallel, or when invoking pytest inside another instance of pytest.

The change copies existing pytest code from https://github.com/pytest-dev/pytest/blob/7.2.x/src/_pytest/junitxml.py#L646 and modifies a single line to include `exist_ok=True`. This notably does not work around another edge case where the user requests creating a file at `grandparent/parent/file.xml` but a non-directory file already exists at `grandparent/parent` . That case does not appear to be reproducing in O3DE, should be very uncommon, and would likely require renaming the user's requested file path.

This fix was hooked into the pytest plugin for configuration because it cannot be done in an autouse pytest fixture. Unfortunately all fixtures are instantiated after pytest initializes a LogXML object.

A better fix would be to directly fix pytest itself. After proving stability, this fix will be recommended to the pytest maintainers.

## How was this PR tested?

Executed the unit tests of LyTestTools with the --junitxml flag set, to make sure the patched method is being called. Manually set a breakpoint and manually recreated the race condition of the file structure changing between `os.path.isdir` and `os.makedirs`.

Will use Jenkins runs to collect additional data on whether we continue to encounter race conditions here.

### DIFF
--- a/Tools/LyTestTools/ly_test_tools/_internal/pytest_plugin/test_tools_fixtures.py
+++ b/Tools/LyTestTools/ly_test_tools/_internal/pytest_plugin/test_tools_fixtures.py
@@ -468,4 +468,4 @@ def monkeypatch_pytest():
             logfile.write(ET.tostring(testsuites, encoding="unicode"))
 
     # override
-    _pytest.junitxml.LogXML.pytest_sessionfinish = types.MethodType(pytest_sessionfinish, _pytest.junitxml.LogXML)
+    _pytest.junitxml.LogXML.pytest_sessionfinish = pytest_sessionfinish
--- a/Tools/LyTestTools/ly_test_tools/_internal/pytest_plugin/test_tools_fixtures.py
+++ b/Tools/LyTestTools/ly_test_tools/_internal/pytest_plugin/test_tools_fixtures.py
@@ -55,6 +55,9 @@ def pytest_configure(config):
     ly_test_tools._internal.pytest_plugin.build_directory = _get_build_directory(config)
     ly_test_tools._internal.pytest_plugin.output_path = _get_output_path(config)
 
+    # patch to work around TODO GHI
+    monkeypatch_pytest()
+
 
 def _get_build_directory(config):
     """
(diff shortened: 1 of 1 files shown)


evanchia-ly-sdets (approved):
LGTM, but I was wondering how we could prevent this from breaking if pytest ever updates their function?

Kadino:
> I was wondering how we could prevent this from breaking if pytest ever updates their function?

We'd have to co-maintain this monkeypatch when we upgrade pytest versions. At least until we get the pytest maintainers to merge a fix. I want a little more data to prove it works before submitting them a bug and/or PR.

Tools/LyTestTools/ly_test_tools/_internal/pytest_plugin/test_tools_fixtures.py
```diff
@@ -55,6 +55,9 @@ def pytest_configure(config):
     ly_test_tools._internal.pytest_plugin.build_directory = _get_build_directory(config)
     ly_test_tools._internal.pytest_plugin.output_path = _get_output_path(config)
 
+    # patch to work around TODO GHI
```
Kadino, line 58: Oops, left myself a TODO in which I intended to link to a GHI in the pytest project.

evanchia-ly-sdets:
> > I was wondering how we could prevent this from breaking if pytest ever updates their function?
> 
> We'd have to co-maintain this monkeypatch when we upgrade pytest versions. At least until we get the pytest maintainers to merge a fix _**and**_ we update to consume their fix. I want a little more data to prove it works before submitting them a bug and/or PR.

Could we perhaps comment which pytest version we're on to make sure we're up to date on co-maintaining or something of that effect?

Kadino:
> Could we perhaps comment which pytest version we're on to make sure we're up to date on co-maintaining or something of that effect?
Line 431

Kadino:
Appears to have created an issue with cached python files. Testing a different type of fix, but this monkeypatch approach would not be suitable if these problems persist.

Kadino:
Clean AR build proved second monkeypatch approach works. Have not found additional race conditions in local testing.

Kadino:
Additional local testing reproduced no other race conditions, so I proposed a PR to the pytest maintainers. This monkeypatch can be removed after that code gets packaged, released, and o3de updates its requirements.txt

AMZN-scspaldi (approved):

PR #13814, opened 2022-12-22
Update to use certifi certificates when opening URLs

AMZN-Phil:
This makes this process more consistent across devices/OSs and fixes some issues some users were seeing with certain https addresses.

Signed-off-by: AMZN-Phil <pconroy@amazon.com>

### DIFF
--- a/scripts/o3de/o3de/utils.py
+++ b/scripts/o3de/o3de/utils.py
@@ -11,9 +11,11 @@ This file contains utility functions
 import argparse
 import sys
 import uuid
+import certifi
 import os
 import pathlib
 import shutil
+import ssl
 import urllib.request
 import logging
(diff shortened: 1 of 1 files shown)


lawsonamzn:
Looks correct, but does this actually fix the problem (were we able to verify it was broken before and that this fixed it?)

nick-l-o3de (approved):

AMZN-Phil:
I could repro the same error that was seen locally on my Mac and was fixed by this change. I can't be certain this was the same root cause that caused the issue that was seen on Windows though.

AMZN-alexpete (approved):
@amzn-tommy if you have a chance could you try this change and see if it fixes the issue for you?

PR #13827, opened 2022-12-22
Fix validation error with depth/stencil RT in bindless pool

akioCL:

## What does this PR do?

Fix a validation error on Vulkan and Metal when incorrectly adding a depth/stencil rendertarget in the bindless pool

## How was this PR tested?

Run ASV bindless sample


moudgils (approved):
Thanks for fixing this. I totally didn't test properly when I made this change haha.

rgba16f (approved):

PR #13828, opened 2022-12-22
Includes CTest in top level CMakeLists to create the CTest config file

evanchia-ly-sdets:

## What does this PR do?

Adds the include(CTest) option to the top level CMakeLists file to create the CTest DartConfiguration file. No actual config is being added, this change is merely to remove the "Cannot find file: ~/o3de/build/windows/DartConfiguration.tcl" warning.

## How was this PR tested?

Tested by configuring locally and running CTest

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,7 @@ endif()
 ################################################################################
 # Initialize
 ################################################################################
+include(CTest)
 include(cmake/GeneralSettings.cmake)
 include(cmake/FileUtil.cmake)
 include(cmake/PAL.cmake)


smurly (approved):

Kadino:
Where does this warning emit from, the Jenkins build scripts? Perhaps a more appropriate change would be to remove extraneous flag `-T Test` from the build scripts?

evanchia-ly-sdets:
> Where does this warning emit from, the Jenkins build scripts? Perhaps a more appropriate change would be to remove extraneous flag `-T Test` from the build scripts?
> 
> We don't use CMake's Dart server for reporting, and it'd be good to avoid extra steps where possible.

The warning emits any time the "-T Test" option is used from the CTest CLI (even locally). The option is used to generate test result xmls and the junit-xml option was proved to not be sufficient as it provides different (less) information.

phuchau1989 (approved):

Kadino (approved):
This appears to be an acceptable workaround, but missing info from junitxml may be worth cutting a bug to the CMake maintainers

Issue #13830, opened 2022-12-23
When using lambdas with Attributes in the EditContext, the lambda signature requires a void* handler


mbalfour-amzn:
**Describe the bug**
When trying to use a lambda with an Attribute in an EditContext, the expectation is that it would have a signature that receives a typed instance of the class, similar to passing a member function in. Instead, the lambda is required to receive a void*.

**Steps to reproduce**
1. Change the code for the AttributeData_WithCallableType_Succeeds in AzCore/Tests/Serialization.cpp to the following. The only change to the method is changing `AZ::AttributeInvoker invoker` to `AZ::AttributeReader invoker`:
```
    TEST_F(Serialization, AttributeData_WithCallableType_Succeeds)
    {
        constexpr AZ::Crc32 invokableCrc = AZ_CRC_CE("Invokable");
        constexpr AZ::Crc32 nonInvokableCrc = AZ_CRC_CE("NonInvokable");
        auto ReadFloat = [](SerializeTestClasses::BaseNoRtti* instance) -> float
        {
            auto noRttiInstance = instance;
            if (!noRttiInstance)
            {
                ADD_FAILURE() << "BaseNoRtti instance object should not be nullptr";
                return 0.0f;
            }
            EXPECT_FALSE(noRttiInstance->m_data);
            return 2.0f;
        };

        m_serializeContext->Class<SerializeTestClasses::BaseNoRtti>()
            ->Attribute(invokableCrc, ReadFloat)
            ->Attribute(nonInvokableCrc, 4.0f)
            ;

        SerializeTestClasses::BaseNoRtti baseNoRttiInstance;
        baseNoRttiInstance.Set();
        auto classData = m_serializeContext->FindClassData(azrtti_typeid<SerializeTestClasses::BaseNoRtti>());
        ASSERT_NE(nullptr, classData);
        AZ::Attribute* attribute = AZ::FindAttribute(invokableCrc, classData->m_attributes);
        ASSERT_NE(nullptr, attribute);
        AZ::AttributeReader invoker(&baseNoRttiInstance, attribute);
        float value = 0;
        EXPECT_TRUE(invoker.Read<float>(value));
        EXPECT_FLOAT_EQ(2.0f, value);

        AZ::Attribute* nonInvokeAttribute = AZ::FindAttribute(nonInvokableCrc, classData->m_attributes);
        ASSERT_NE(nullptr, nonInvokeAttribute);
        invoker = { &baseNoRttiInstance, nonInvokeAttribute };
[...]

PR #13831, opened 2022-12-23
[Terrain] Fix paint brush settings when used with a color.

mbalfour-amzn:
## What does this PR do?

This fixes two problems discovered while testing the Paint Brush Settings with color painting:
1. The lambda for setting the color space wasn't getting called due to the wrong function signature. (See #13830 ) This was causing the color field to display the colors in the wrong color space.
2. By setting the color field to readonly when using the eyedropper, the color is greyed out, making it hard to see the color the eyedropper is picking. There's no harm in leaving these fields writeable, so this removes the readonly flags on them.

## How was this PR tested?

I used these changes with local modifications for doing color painting and verified that the color picker is much more usable.

### DIFF
--- a/Code/Framework/AzFramework/AzFramework/PaintBrush/PaintBrushSettings.cpp
+++ b/Code/Framework/AzFramework/AzFramework/PaintBrush/PaintBrushSettings.cpp
@@ -54,23 +54,6 @@ namespace AzFramework
         }
     }
 
-    // The following settings are visible but read-only when in the Eyedropper mode
-
-    bool PaintBrushSettings::GetColorReadOnly() const
-    {
-        return (m_brushMode == PaintBrushMode::Eyedropper);
-    }
-
-    bool PaintBrushSettings::GetIntensityReadOnly() const
-    {
-        return (m_brushMode == PaintBrushMode::Eyedropper);
-    }
-
-    bool PaintBrushSettings::GetOpacityReadOnly() const
-    {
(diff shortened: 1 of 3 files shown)


greerdv (approved):

AMZN-daimini (approved):
> By setting the color field to readonly when using the eyedropper, the color is greyed out, making it hard to see the color the eyedropper is picking. There's no harm in leaving these fields writeable, so this removes the readonly flags on them.

If possible, I would also raise this as a separate issue to UX. I think the UI should still show the color for color properties in some way even when the field is disabled. It could make sense to fix that aspect for other use cases :)

dtamkin1 (approved):

mbalfour-amzn:
> > By setting the color field to readonly when using the eyedropper, the color is greyed out, making it hard to see the color the eyedropper is picking. There's no harm in leaving these fields writeable, so this removes the readonly flags on them.
> 
> If possible, I would also raise this as a separate issue to UX. I think the UI should still show the color for color properties in some way even when the field is disabled. It could make sense to fix that aspect for other use cases :)

Fair point, I'll raise the question. There are some merits both to greying it out as well as to leaving it colored when it's set to readonly / disabled, it's a good question to bring to UX.

mbalfour-amzn:
Discussed the color picker with UX, and submitted the change as a request on top of a related bug that I discovered while discussing this with UX. :)
https://github.com/o3de/o3de/issues/13833

Issue #13304, opened 2022-11-21
`test_FindCodeowners_ThisCWD_FindForO3DE` test fails if the path to the O3DE engine does not contain "o3de"


burelc-amzn:
**Describe the bug**
The test `test_FindCodeowners_ThisCWD_FindForO3DE` is trying to ensure that the `.github/CODEOWNERS` file  at the top level of the repository is not moved. It does this by checking to see if the absolute path to the test contains the string "o3de". Locally on my Linux machine, I clone O3DE to `~/ws/main`, so this test constantly fails. Because this test is part of the `LyTestTools_Integ_Codeowners.smoke::TEST_RUN` test in CTest, and part of the smoke suite, it is part of the normal test run. I expect to be able to run all the same tests locally as run on the build machines, and get the same results.

**Steps to reproduce**
1. Build O3DE in a path that does not contain the string "o3de".
2. Run `ctest --verbose -R LyTestTools_Integ_Codeowners.smoke::TEST_RUN`

**Expected behavior**
Test passes

**Actual behavior**
Test fails

**Found in Branch**
`development` at 958dcac745

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Linux
 - Version: Ubuntu 20.04 LTS

PR #13835, opened 2022-12-24
Changed codeowners test to not fail if o3de text is not in path.

AMZN-scspaldi:
Signed-off-by: scspaldi <scspaldi@amazon.com>

## What does this PR do?
https://github.com/o3de/o3de/issues/13304
Fixes a test to make it not rely on "o3de" being in the file path unnecessarily.

## How was this PR tested?
Ran the test locally.

### DIFF
--- a/Tools/LyTestTools/tests/integ/test_o3de_codeowners.py
+++ b/Tools/LyTestTools/tests/integ/test_o3de_codeowners.py
@@ -14,7 +14,7 @@ import ly_test_tools.cli.codeowners_hint as hint
 def test_FindCodeowners_ThisCWD_FindForO3DE():
     codeowners_path = hint.find_github_codeowners(os.path.abspath(__file__))
     assert codeowners_path
-    assert "o3de" in str(codeowners_path), "codeowners file has been tampered with, or test unexpectedly migrated"
+    assert ".github" in str(codeowners_path), "codeowners file has been tampered with, or test unexpectedly migrated"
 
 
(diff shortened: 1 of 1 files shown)


Kadino (approved):
While it's still possible that the user has locally tampered with their o3de repo by adding extra .github folders (or injecting other repos), that should be less common than renaming the o3de repo's root folder. And it should be easy for humans to catch such changes from being submitted in pull requests.

Tools/LyTestTools/tests/integ/test_o3de_codeowners.py
```diff
@@ -14,7 +14,7 @@
 def test_FindCodeowners_ThisCWD_FindForO3DE():
     codeowners_path = hint.find_github_codeowners(os.path.abspath(__file__))
     assert codeowners_path
-    assert "o3de" in str(codeowners_path), "codeowners file has been tampered with, or test unexpectedly migrated"
+    assert ".github" in str(codeowners_path), "codeowners file has been tampered with, or test unexpectedly migrated"
```
evanchia-ly-sdets, line 17: Checking for .github in the path is essentially checking if the path exists which it already does above

AMZN-scspaldi: While the function above does currently assume that the path contains .github, this is not a requirement of a codeowners file, and this line in the test ensures that even if the above function changes that the codeowners file maintains its current location.

evanchia-ly-sdets (approved):

PR #13839, opened 2022-12-27
EMFX: Child motion set are not showing up correctly in the actor component

amzn-rhhong:
Signed-off-by: Roman Hong <rhhong@amazon.com>

The child motion set are not showing up correctly in the actor component
image


cgalvan (approved):

Gems/EMotionFX/Code/Include/Integration/AnimGraphComponentBus.h
```diff
@@ -191,6 +191,9 @@ namespace EMotionFX
             /// Making a request to desync from the anim graph to its leader graph
             /// \param leaderEntityId - the entity id of another anim graph.
             virtual void DesyncAnimGraph(AZ::EntityId leaderEntityId) = 0;
+
+            /// Set the name of the active motion set.
+            virtual void SetActiveMotionSet(const char* activeMotionsetName) = 0;
```
hultonha, line 196: nit: `MotionsetName` -> `MotionSetName` (also should the function be called `SetActiveMotionSetName`?

Gems/EMotionFX/Code/Source/Integration/Components/AnimGraphComponent.h
```diff
@@ -138,6 +138,7 @@ namespace EMotionFX
             bool GetVisualizeEnabled() override;
             void SyncAnimGraph(AZ::EntityId leaderEntityId) override;
             void DesyncAnimGraph(AZ::EntityId leaderEntityId) override;
+            void SetActiveMotionSet(const char* activeMotionsetName) override;
```
hultonha, line 141: nit: Name of parameter and function (mentioned above)

Gems/EMotionFX/Code/Source/Integration/Components/AnimGraphComponent.h
```diff
@@ -159,6 +160,7 @@ namespace EMotionFX
             void UpdateActorExternal(float deltatime) override;
             void SetNetworkRandomSeed(AZ::u64 seed) override;
             AZ::u64 GetNetworkRandomSeed() const override;
+
```
hultonha, line 163: uber nit: Could just revert this new line break addition

hultonha (approved):
Change looks good I just had a minor suggestion about some naming

Gems/EMotionFX/Code/Include/Integration/AnimGraphComponentBus.h
hultonha: @amzn-rhhong did you see my comment about the name of the function? As this isn't actually changing the active motion set but the name of the action motion set, should this instead be called `SetActionMotionSetName`? If you think it's unnecessary it doesn't have to change but it may be clearer at the callsite if called this (so it doesn't seem like we're using the string as a key to lookup a new motion set and set it as the active one).

amzn-rhhong: This will change the active motion set, not change the name of the activated motion set. I think the function name is fine, but I understand it's a little confusing.

hultonha: Ah okay understood, thanks for clarifying! 👍

PR #13850, opened 2022-12-28
Add opt in OnNetworkActivated event for MultiplayerComponents

AMZN-puvvadar:

## What does this PR do?

This change adds OnNetworkActivated, an event for MultiplayerComponents that is invoked

1. After RPCs are bound and the net entity is attached
2. If the MultiplayerComponent's networked activated event handler is attached to its corresponding NetBindComponent.

This addresses desired functionality in https://github.com/o3de/o3de/issues/12137

## How was this PR tested?

See included unit test


Gems/Multiplayer/Code/Source/NetworkEntity/EntityReplication/EntityReplicator.cpp
```diff
@@ -172,6 +172,7 @@ namespace Multiplayer
             m_onEntityStopHandler.Disconnect();
             netBindComponent->AddEntityStopEventHandler(m_onEntityStopHandler);
             AttachRPCHandlers();
+            netBindComponent->NetworkActivated();
```
lmbr-pip, line 175: nit: It seems strange that the entity replicator "owns" telling an entity that the network is activated. Are there other places this event could live or is there a clear reason why this is owned by EntityReplicators?

Gems/Multiplayer/Code/Include/Multiplayer/Components/MultiplayerComponent.h
```diff
@@ -67,6 +67,10 @@ namespace Multiplayer
         NetworkEntityHandle GetEntityHandle();
         void MarkDirty();
 
+        //! Invoked when the NetworkEntity is attached and has RPCs bound via m_networkActivatedHandler
+        //! if m_networkActivatedHandler is registered via NetBindComponent::AddNetworkActivatedEventHandler 
+        virtual void OnNetworkActivated(){};
```
lmbr-pip, line 72: IHMO It would be better to write this in a less technical framing. Why do I as a MultiplayerComponent author care about this event? What can I do once this event is received?

lmbr-pip (commented):
LG overall.

Gems/Multiplayer/Code/Source/NetworkEntity/EntityReplication/EntityReplicator.cpp
AMZN-puvvadar: The specific thing we care about is AttachRPCHandlers right above it. That function is what actually makes RPC dispatch functional for a networked entity.

Gems/Multiplayer/Code/Include/Multiplayer/Components/MultiplayerComponent.h
AMZN-puvvadar: I wanted to be technical here so users can understand why overriding OnNetworkActivated by itself is not sufficient. I can add a little more description here.

Gems/Multiplayer/Code/Include/Multiplayer/Components/MultiplayerComponent.h
```diff
@@ -67,6 +67,11 @@ namespace Multiplayer
         NetworkEntityHandle GetEntityHandle();
         void MarkDirty();
 
+        //! Override to run component logic when the NetworkEntity has completed network activation
+        //! Invoked when the NetworkEntity is attached and has RPCs bound via m_networkActivatedHandler
+        //! Requires m_networkActivatedHandler be registered via NetBindComponent::AddNetworkActivatedEventHandler
```
AMZN-Gene, line 72: +1 for explaining how to use it

AMZN-Gene (approved):

byrcolin (approved):

PR #13851, opened 2022-12-28
Add new AZ_Trace and AZ_Info macros

lmbr-pip:
Signed-off-by: Pip Potter <61438964+lmbr-pip@users.noreply.github.com>

## What does this PR do?

See https://github.com/o3de/sig-core/issues/47 for context.

* Adds two new logging levels: trace and debug (which aligns with ILogger).
     * Move to enum class to avoid name conflict with `AZ::Debug::Trace` class
     * Added `debug` even though its not used, to align with other levels ILogger and common logging standards. Otherwise you could not do not simple < or > equal tests if we added it in later. 
* Add two new macros `AZ_Info` and `AZ_Trace` and a single use `AZ_TraceOnce`
* Marks `AZ_TracePrintf` and `AZ_TracePrintfOnce` as deprecated

## How was this PR tested?

* Ran Windows Editor
* Tested that trace level statements are ignored when `bg_loglevel` is left at default.
* Tested that trace level statements are outputted when `bg_loglevel` dropped to `trace`


Code/Framework/AzCore/AzCore/Debug/Trace.h
```diff
@@ -319,22 +313,50 @@ namespace AZ
...
+    if(AZ::Debug::Trace::Instance().IsTraceLoggingEnabledForLevel(AZ::Debug::LogLevel::Trace))                     \
+    {                                                                                                              \
+        AZ::Debug::Trace::Instance().Printf(window, __VA_ARGS__);                                                  \
+    }
+
+    //! The AZ_TraceOnce macro output the result of the format string only once for each use of the macro
+    //! It does not take into account the result of the format string to determine whether to output the string or not
+    //! What this means is that if the formatting results in different output string result only the first result
+    //! will ever be output
+    #define AZ_TraceOnce(window, ...) \
```
amzn-mike, line 341: Do we need an InfoOnce as well?

Code/Framework/AzCore/AzCore/Debug/Trace.h
```diff
@@ -319,22 +313,50 @@ namespace AZ
...
+    if(AZ::Debug::Trace::Instance().IsTraceLoggingEnabledForLevel(AZ::Debug::LogLevel::Info))                      \
     {                                                                                                              \
         AZ::Debug::Trace::Instance().Printf(window, __VA_ARGS__);                                                  \
     }
 
+    #define AZ_Trace(window, ...)                                                                                  \
+    if(AZ::Debug::Trace::Instance().IsTraceLoggingEnabledForLevel(AZ::Debug::LogLevel::Trace))                     \
+    {                                                                                                              \
+        AZ::Debug::Trace::Instance().Printf(window, __VA_ARGS__);                                                  \
+    }
```
amzn-mike, line 335: This is going to continue forward with having to put `\n` at the end of every Info/Trace call, which isn't required for Warning/Error/Assert.  I think we should probably take this opportunity to make these two calls to Printf always include a `\n` (probably add another parameter to Printf) so they're consistent with the other 3 APIs.

lumberyard-employee-dm: Make the change to add a the newline is not so simple ....

The macro itself would need to create a local fixed_string_buffer to format the `__VA_ARGS__` into before calling Printf.
The easiest solution would be to add a `PrintfWithNewline` function that does so in the context of a function.

lumberyard-employee-dm (approved):

cgalvan (approved):

lsemp3d (approved):

lumberyard-employee-dm:
So the release build is failing due to missing empty versions of the new macros in the `!AZ_ENABLE_TRACING` [block](https://github.com/o3de/o3de/blob/cf92cebc5891d4e24e71d1361ba4c5aa04c5c0d3/Code/Framework/AzCore/AzCore/Debug/Trace.h#L393-L405)

lumberyard-employee-dm:
Thanks @AMZN-alexpete for adding the macros in release configurations!

lumberyard-employee-dm (approved):

rgba16f (approved):

hultonha (approved):

PR #13739, opened 2022-12-16
Fix Android math unit tests (Scalar and NEON) and enable NEON by default

moraaar:
## What does this PR do?

This PR fixes math unit tests on Android using both Scalar and NEON instructions.

There are only 2 tests disabled (Android only) that will be fixed later: `MATH_IntersectSegmentTriangleTest.RegressionTestForSpecificSegmentsAndTriangles` and `MATH_ShapeIntersection.Test`

Changes:
- Fixed NEON Simd Vec3 comparison functions by filtering out the last element. Thanks @greerdv for figuring out the filtering operations!
- Enabled NEON by default on Android.
- Fixed and unified `FromVecX`, `ToVecX`, `Div` and `Reciprocal` operations to make sure there are no unexpected values in their simd data.
- Added additional Simd unit tests for `FromVecX`, `ToVecX`, `TestSplatFirst|Second|Third|Fourth` operations.
- Added AZStd functions for `isinf`, `isnan` and `azisfinite` and provide custom implementation when --fast-math flag is used.
- Use the right tolerance when necessary for tests where NEON precision is specially bad. In other cases the tolerance was just slightly increased.

## How was this PR tested?

Checked that math unit tests pass on:
- Android (Scalar and NEON)
- Windows (Scalar and SSE)
- iOS (NEON)
- Mac (SSE)
- Linux (SSE)

PR #13852, opened 2022-12-29
Fixed Matrix3x4::GetInverseFull() and FBX processing.

santorac:
## What does this PR do?

Fixed Matrix3x4::GetInverseFull() and FBX processing.
This bug was introduced at https://github.com/o3de/o3de/pull/13739 It messed up the normal vectors on some models, which caused many AtomSampleViewer screenshot tests to fail. The tests are passing again.

## How was this PR tested?

AtomSampleViewer full test suite on vulkan.

### DIFF
--- a/Code/Framework/AzCore/AzCore/Math/Matrix3x4.cpp
+++ b/Code/Framework/AzCore/AzCore/Math/Matrix3x4.cpp
@@ -433,7 +433,7 @@ namespace AZ
         // calculate the determinant
         const float determinant = result.m_rows[0].Dot3(GetColumn(0));
 
-        if (!AZ::IsClose(determinant, 0.0f))
+        if (!AZ::IsClose(determinant, 0.0f, Constants::FloatEpsilon))
         {
             float determinantInv = 1.0f / determinant;
             result.m_rows[0] *= determinantInv;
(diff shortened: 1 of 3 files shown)


gadams3 (approved):

lmbr-pip (approved):

jeremyong-az (approved):

jeremyong-az:
Thanks for the fix. Looking a bit closer, the `AZ::IsClose` function seems *really dangerous* to the point where I simply don't think it should be used. A default tolerance of `0.001` is a pretty bizarre choice and we may want to look into forcing the user to specify a tolerance in the future and possibly providing a separate function for more common cases where we want to measure float similarity within X ULPs or something. (cc @moraaar)

jeremyong-az:
One thing I noticed btw is that `Matrix4x4` and `Matrix3x3` don't have the associated `IsClose` change and instead checks if the matrix is exactly singular. It probably makes sense to compare those determinants with a relaxed `1e-7` threshold check similar to the `3x4` later.

moraaar:
Thanks for fixing the issue. It's great you've added the specific case as a unit test. I run the math tests on Android and it passes.

>A default tolerance of 0.001 is a pretty bizarre choice
Yeah, that seems like a quite high default number for `IsClose`, not sure where it comes from or if there is a valid reason behind it. It can be tricky to change it as it might affect other unit tests, but it's a valid observation.

PR #13857, opened 2022-12-29
[Color Picker] Misc fixes to the Color Picker control

mbalfour-amzn:
## What does this PR do?

This fixes a few bugs with the Color Picker widget:
1. The ColorEditorConfiguration has a setting for RGB vs RGBA, but the setting was ignored when the configuration was set via the RPE. It is now used correctly.
2. The ColorPicker control had inconsistent alpha initial values in several places, including the alpha slider and the RGBA edit control. Because there are lots of places that only emit "changed" signals if a value has changed from its previous value, this caused problems with alpha because setting the color to an alpha of 1 matched the default value in one place, preventing a "changed" signal, but didn't match the default values everywhere else, so they didn't refresh. The fixes involved making the different alpha values initialize to the same consistent value and to make sure that all the widgets initialized themselves regardless of the value.

## How was this PR tested?

Opened the color picker dialog box and verified that it shows RGBA when the configuration requests it and shows a non-zero alpha slider by default.

Before:
image

After:
image


Code/Framework/AzQtComponents/AzQtComponents/Components/Widgets/ColorPicker/ColorHexEdit.cpp
```diff
@@ -61,7 +61,7 @@ ColorHexEdit::ColorHexEdit(QWidget* parent)
     , m_red(0)
     , m_green(0)
     , m_blue(0)
-    , m_alpha(0)
+    , m_alpha(1)
```
AMZN-daimini, line 64: `qreal` is a typedef for `double`, so these definitions should be doubles:
```
    , m_red(0.0)
    , m_green(0.0)
    , m_blue(0.0)
    , m_alpha(1.0)
```

AMZN-daimini (approved):

Code/Framework/AzQtComponents/AzQtComponents/Components/Widgets/ColorPicker/ColorHexEdit.cpp
mbalfour-amzn: Since I already started the AR, I'll either adjust this if I need to resubmit or save it for a future submission. 

The compiler will generate the same code for both, but I totally agree it's more pleasing to the eye to have the constants in the correct type. :)

amzn-tmryan (approved):

PR #13858, opened 2022-12-29
[Terrain] Refactors that will help with better color painting support

mbalfour-amzn:
## What does this PR do?

1. Moved brush mode and color mode from PaintBrushSettings to GlobalPaintBrushSettings. This eliminates some problems with using PaintBrushSettings from Script Canvas, because it was preventing the user from setting color and smoothing settings.
2. Reduced some complexity caused by trying to hook the Editor paintbrush directly to a wrapped runtime component, and instead now just use pass-through calls from the Editor component to the runtime component. This makes the logic a bit more straightforward, and had the side benefit of improving dirty image tracking across multiple edits.
3. Fixed a small bug where the opacity slider wasn't getting refreshed when the alpha value was changed in the color picker.
4. Fixed bugs in GradientPreviewer where a serialized bounds entity field was getting overwritten on Activate/Deactivate. This was causing issues with undo/redo, and it also caused the field to get cleared erroneously if the component was ever accidentally deactivated (for example, by removing a dependent component).
5. Fixed issue in EditorGradientTransformComponent where it would over-aggressively mark its entity as dirty on every painting change, even if nothing in its configuration actually changed.

There are a lot of files touched, but not many changes. Most of the changes boil down to the following:
* Moved brush mode and brush color mode from the AzFramework to the AzToolsFramework namespace.
* Made all the edit context fields visible by default for PaintBrushSettings.
* Moved imageIsModified tracking from the EditorImageGradientComponentMode to ImageGradientComponent
* Changed EditorImageGradientComponent from using an extra bus and some weird shenanigans to hook up the Editor paintbrush to instead just using a more simple pass-through call mechanism. 
* Isolated bugfixes in GradientPreviewer and EditorGradientTransformComponent.

## How was this PR tested?

1. Painted on image gradients and verified that the behavior didn't change, except for a dirty image will now prompt for saving across multiple sessions if it hasn't been saved, instead of only for a single session.
2. Added PaintBrushSettings into a Script Canvas script and verified that color and smoothness settings are now visible and settable, and the brush mode (which is an Editor-only thing) is no longer exposed.


Code/Framework/AzFramework/AzFramework/PaintBrush/PaintBrushSettings.cpp
```diff
@@ -19,7 +19,6 @@ namespace AzFramework
         {
             serializeContext->Class<PaintBrushSettings>()
                 ->Version(6)
```
AMZN-daimini, line 21: Should this removal prompt a version change?

Code/Framework/AzFramework/AzFramework/PaintBrush/PaintBrushSettings.cpp
```diff
@@ -58,51 +57,51 @@ namespace AzFramework
 
     bool PaintBrushSettings::GetSizeVisibility() const
     {
-        return (m_brushMode != PaintBrushMode::Eyedropper);
+        return true;
     }
```
AMZN-daimini, line 61: (repeated) since all of these visibility settings seem to be conditional, should they all default to `false`? Or some of them?

AMZN-daimini (approved):

Code/Framework/AzFramework/AzFramework/PaintBrush/PaintBrushSettings.cpp
mbalfour-amzn: Probably, though it's semi-irrelevant. The version number is only used by slices, not prefabs, and removing fields generally "just works" anyways. I can bump it just for cleanliness either in this change though if I need to resubmit anything or in my next change.

Code/Framework/AzFramework/AzFramework/PaintBrush/PaintBrushSettings.cpp
mbalfour-amzn: No, they actually very much _need_ to default to true in the PaintBrushSettings, and then use the conditional logic in the overrides in GlobalPaintBrushSettings.

The issue that this fixes is that the GlobalPaintBrushSettings, which are the ones used in the Editor from the paintbrush tool, show and hide various fields based on whether you're using the paintbrush, eyedropper, or smoothing tool. However, if you try to use PaintBrushSettings from ScriptCanvas, the SC API for painting exposes those tools as separate APIs, not separate modes, so you need to be able to edit _all_ the settings for the PaintBrushSettings if you create an SC variable containing one of these. What was happening is that with the conditionals on here, you could never see or edit the color field or the smoothing values from SC, because it would only show them if the brush mode changed which literally couldn't happen in the SC use case.

AMZN-daimini: > No, they actually very much _need_ to default to true in the PaintBrushSettings, and then use the conditional logic in the overrides in GlobalPaintBrushSettings.
> 
> The issue that this fixes is that the GlobalPaintBrushSettings, which are the ones used in the Editor from the paintbrush tool, show and hide various fields based on whether you're using the paintbrush, eyedropper, or smoothing tool. However, if you try to use PaintBrushSettings from ScriptCanvas, the SC API for painting exposes those tools as separate APIs, not separate modes, so you need to be able to edit _all_ the settings for the PaintBrushSettings if you create an SC variable containing one of these. What was happening is that with the conditionals on here, you could never see or edit the color field or the smoothing values from SC, because it would only show them if the brush mode changed which literally couldn't happen in the SC use case.

Oh the visibility actually prevents the fields from being edited programmatically? :/ Not sure that makes a lot of sense, but I understand now.

mbalfour-amzn: It doesn't prevent it from being edited programmatically, but if you create an SC variable and set it to "initialize from component", the edit context for the structure gets displayed on the component card in the entity inspector. _That's_ the point at which the visibility affects what you can and can't initialize.

mbalfour-amzn: image

The circled fields weren't showing up before, and the Brush Mode *was* showing up even though it's meaningless in this context.

Code/Framework/AzFramework/AzFramework/PaintBrush/PaintBrushSettings.cpp
mbalfour-amzn: Bumped!

amzn-mike (approved):

phuchau1989 (approved):

PR #13860, opened 2022-12-29
AssetProcessor: Fix new startup scan skip mode, update UX and rename tools to settings

lmbr-pip:
Signed-off-by: Pip Potter <61438964+lmbr-pip@users.noreply.github.com>

## What does this PR do?

Will consider splitting into seperate PRs

1. Fixes new startup scan mode so AP correct signals that its Idle.  Ensures that `m_isAssetProcessorReady` flag is correctly set when entering Idle after skipping startup scan.
3. Update UX for startup scan mode to new language discussed with sig/ui-ux
4. Rename Tools tab -> settings
5. Fix missing or incorrect virtuals noticed in code

Screenshot 2022-12-29 125957

## How was this PR tested?

* Ran all unit tests
* Ran AP with new startup scan and ensured that once AP goes Idle, Editor gets a correct response from `RequestAssetProcessorStatus`

### DIFF
--- a/Code/Tools/AssetProcessor/native/ui/MainWindow.ui
+++ b/Code/Tools/AssetProcessor/native/ui/MainWindow.ui
@@ -1464,10 +1464,10 @@
              </layout>
             </widget>
            </widget>
-          </item>
+          </item>dis
          </layout>
         </widget>
-        <widget class="QWidget" name="ToolsPage">
+        <widget class="QWidget" name="SettingsPage">
          <layout class="QVBoxLayout" name="verticalLayout_3">
           <item>
            <layout class="QVBoxLayout" name="fullScanLayout">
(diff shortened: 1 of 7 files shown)


Code/Tools/AssetProcessor/native/ui/MainWindow.ui
```diff
@@ -1464,18 +1464,18 @@
              </layout>
             </widget>
            </widget>
-          </item>
+          </item>dis
```
amzn-mike, line 1467: What happened here?

amzn-mike (approved):

Code/Tools/AssetProcessor/native/ui/MainWindow.ui
lmbr-pip: Good catch, not sure where that came from (assume some accidental keystrokes during shutdown of VisualStudio and committing the file). Will fix.

Code/Tools/AssetProcessor/native/utilities/ApplicationManagerBase.cpp
```diff
@@ -774,7 +774,7 @@ void ApplicationManagerBase::InitAssetRequestHandler(AssetProcessor::AssetReques
         {
             if (!m_assetProcessorManagerIsReady)
             {
-                if (m_remainingAPMJobs == newNum)
+                if ((m_remainingAPMJobs == newNum) && m_remainingAPMJobs)
```
lmbr-pip, line 777: This is the fix for skipping startup scan. Without this change, when we move to Idle, this signal is called with newNum == 0 and m_remainingAPMJobs never changed from zero. 

This is the sole place in the code that changes the state of  `m_assetProcessorManagerIsReady` which is one of the three things that must be true to signal back to the Editor that AP has completed enough work to allow launch to continue.

amzn-mike (approved):

junbo75 (approved):

AMZN-stankowi (approved):

Issue #12870, opened 2022-10-27
Asset Browser - Multi browser experience


lsemp3d:
As a user: I need the ability to explore, view, and manage my asset for more than one Asset Browser window. Imagine your experience when managing your personal assets on Mac or PC (Windows explorer experience). 

 
 * As a user, I want to right-click on Asset browser and open up a new window
 * As a user, I want to right-click on a folder or tab and open up a new window in that specific location
 * As a user, I want the ability to dock both asset browsers windows side by side.
 * As a user, I want the ability to copy past OR click drag assets from one Asset Browser window to another.

 

![image](https://user-images.githubusercontent.com/58790905/198414105-590674d3-675c-4300-af79-efa1e0c5927f.png)

PR #13862, opened 2022-12-30
show secondary asset browser from context menu

Be-ing:
This adds a secondary asset browser, which is not shown in the Tools menu, that can be opened by right clicking assets in the asset browser.

This partially fixes https://github.com/o3de/o3de/issues/12870. Dragging and dropping between the two asset browsers is not implemented yet.

Signed-off-by: Be Wilson <be.wilson@kdab.com>

https://user-images.githubusercontent.com/9455094/210030961-69a04fe2-c962-41f6-889e-383f00c5f3bb.mp4

### DIFF
--- a/Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
+++ b/Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
@@ -10,6 +10,8 @@
 
 #include "AzAssetBrowserWindow.h"
 
+#include <AssetBrowser/Views/AssetBrowserTreeView.h>
+
 // AzToolsFramework
 #include <AzCore/Console/IConsole.h>
 #include <AzToolsFramework/API/ToolsApplicationAPI.h>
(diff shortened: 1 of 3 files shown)


dtamkin1 (approved):

Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
```diff
@@ -310,6 +312,10 @@ void AzAssetBrowserWindow::RegisterViewClass()
     AzToolsFramework::ViewPaneOptions options;
     options.preferedDockingArea = Qt::LeftDockWidgetArea;
     AzToolsFramework::RegisterViewPane<AzAssetBrowserWindow>(LyViewPane::AssetBrowser, LyViewPane::CategoryTools, options);
+
+    options.showInMenu = false;
+    const QString name = QString("%1 2").arg(LyViewPane::AssetBrowser);
+    AzToolsFramework::RegisterViewPane<AzAssetBrowserWindow>(qPrintable(name), LyViewPane::CategoryTools, options);
```
cgalvan, line 318: Does the design intend to only have 2 asset browsers, or is the user going to be able to open any number of asset browsers? If the latter, we wouldn't need to register an additional one it could just be registered as a multi-view pane?

cgalvan: Cool makes sense.

My only other comment is that in the mockups the 2 is surrounded by parentheses so it would be `Asset Browser (2)`, did that get changed in an offline discussion or should that be updated here?

cgalvan (approved):

Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
cgalvan: Thanks! Looks good, a new AR build is in progress.

PR #13884, opened 2023-01-03
{ghi13575} fixing the python life cycle in ScriptProcessorRule

jackalbe:
## What does this PR do?

This fixes the Python shutdown issues that the Asset Builder is seeing when the Script Rule behavior uses a stale interface pointer. This first checks to see if the script interface is valid before shutting down Python.

## How was this PR tested?

Added ScriptProcessorRule_EditorPythonEventsInterface_RunsWithWhenCleared to regress test that the Deactivate() does not use a stale pointer.

### DIFF
--- a/Code/Tools/SceneAPI/SceneData/Tests/SceneManifest/SceneManifestRuleTests.cpp
+++ b/Code/Tools/SceneAPI/SceneData/Tests/SceneManifest/SceneManifestRuleTests.cpp
@@ -139,17 +139,18 @@ namespace AZ
                 m_jsonSystemComponent->Reflect(m_jsonRegistrationContext.get());
 
                 m_data.reset(new DataMembers);
+                m_data->m_settings = AZStd::make_unique<AZ::NiceSettingsRegistrySimpleMock>();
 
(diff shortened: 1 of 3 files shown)


cgalvan (approved):

AMZN-stankowi (approved):

Code/Tools/SceneAPI/SceneData/Tests/SceneManifest/SceneManifestRuleTests.cpp
```diff
@@ -502,5 +507,38 @@ namespace AZ
             scriptProcessorRuleBehavior.Deactivate();
             EXPECT_TRUE(executeByFilenameCalled);
         }
+
+        TEST_F(SceneManifest_JSON, ScriptProcessorRule_EditorPythonEventsInterface_RunsWithWhenCleared)
```
junbo75, line 511: Nit: The test name is a little confusing. Do you mean ..._RunsWithEditorPythonEventsInterfaceCleared?

junbo75 (approved):

PR #13890, opened 2023-01-04
Make CCD read only when kinematic is enabled and vice versa

amzn-ahmadkrm:

## What does this PR do?

This PR updates the UI to disallow the user to set a Rigid Body Component to Kinematic mode while CCD is enabled and vice versa, these are not compatible in PhysX and results in a warning. This PR makes it so it doesn't get to the point of the PhysX warning and informs the user that only one setting can be used at once.

devenv_C9X5GRLxQv
devenv_LlneAyonsK

## How was this PR tested?

Manual testing.

### DIFF
--- a/Code/Framework/AzFramework/AzFramework/Physics/Configuration/RigidBodyConfiguration.cpp
+++ b/Code/Framework/AzFramework/AzFramework/Physics/Configuration/RigidBodyConfiguration.cpp
@@ -105,13 +105,23 @@ namespace AzPhysics
         }
     }
 
-    const AZStd::string kinematicDescription =
+    constexpr AZStd::string_view kinematicDescription =
         "When active, the rigid body is not affected by gravity or other forces and is moved by script. ";
 
-    const AZStd::string ccdDescription =
+    constexpr AZStd::string_view kinematicDescriptionReadOnly =
+        "When active, the rigid body is not affected by gravity or other forces and is moved by script. "
+        "<b>The rigid body cannot be set as Kinematic if CCD is enabled, disable CCD to allow changes to be made.</b>";
+
+    constexpr AZStd::string_view ccdDescription =
         "When active, the rigid body has continuous collision detection (CCD). Use this to ensure accurate "
         "collision detection, particularly for fast moving rigid bodies. CCD must be activated in the global PhysX "
         "configuration. ";
+
+    constexpr AZStd::string_view ccdDescriptionReadOnly =
+        "When active, the rigid body has continuous collision detection (CCD). Use this to ensure accurate "
+        "collision detection, particularly for fast moving rigid bodies. CCD must be activated in the global PhysX "
+        "configuration. <b>CCD cannot be enabled if the rigid body is kinematic, set the rigid body as non-kinematic"
+        "to allow changes to be made.</b>";
 
     AZ_CLASS_ALLOCATOR_IMPL(RigidBodyConfiguration, AZ::SystemAllocator, 0);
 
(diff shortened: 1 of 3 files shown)


Code/Framework/AzFramework/AzFramework/Physics/Configuration/RigidBodyConfiguration.cpp
```diff
@@ -184,6 +193,35 @@ namespace AzPhysics
...
+
+    AZStd::string RigidBodyConfiguration::GetKinematicTooltip() const
+    {
+        if (m_ccdEnabled)
+        {
+            return kinematicDescription +
+                "<b>The rigid body cannot be set as Kinematic if CCD is enabled, disable CCD to allow changes to be made.</b>";
+        }
+        return kinematicDescription;
+    }
```
AMZN-daimini, line 223: This setup forces the system to allocate a new string every time the function is called. Since the outcomes of the function are known at compile time, it may be simpler to change this so that there is no memory allocation involved:

_(I'll write examples for just the kinematic part for clarity)_

1. Define all possible strings as `constexpr AZStd::string_view` constants:
```
    constexpr AZStd::string_view kinematicDescription =
        "When active, the rigid body is not affected by gravity or other forces and is moved by script. ";
    const AZStd::string kinematicDescriptionReadOnly =
        "When active, the rigid body is not affected by gravity or other forces and is moved by script. <b>The rigid body cannot be set as Kinematic if CCD is enabled, disable CCD to allow changes to be made.</b>"";
```

or even better, if the compiler lets you do it:

```
    constexpr AZStd::string_view kinematicDescription =
        "When active, the rigid body is not affected by gravity or other forces and is moved by script. ";
    const AZStd::string kinematicDescriptionReadOnly =
       kinematicDescription + "<b>The rigid body cannot be set as Kinematic if CCD is enabled, disable CCD to allow changes to be made.</b>"";
```

2. Change the function signature to return a `const AZStd::string_view`
```
const AZStd::string_view RigidBodyConfiguration::GetKinematicTooltip() const
```

3. Just return one string or the other in the function body
```
return (m_ccdEnabled) ? kinematicDescriptionReadOnly : kinematicDescription;
```

I'm writing this off the top of my mind, so the exact types may need to be adjusted a bit (maybe a string instead of string view, or it may be even simpler to just use `const char*`) but I think it conveys the base idea :)

amzn-ahmadkrm: Thank you for this :) I changed it to a string view but no the compiler didn't let me concatenate string views :( 👎

Code/Framework/AzFramework/AzFramework/Physics/Configuration/RigidBodyConfiguration.cpp
```diff
@@ -104,6 +105,24 @@ namespace AzPhysics
         }
     }
 
+    constexpr AZStd::string_view kinematicDescription =
```
greerdv, line 108: I think the coding standards have upper camel case for constant names like this https://github.com/o3de/sig-core/blob/main/governance/Coding-Standards-and-Style-Guide.md#constant-names (e.g. KinematicDescription rather than kinematicDescription).

Code/Framework/AzFramework/AzFramework/Physics/Configuration/RigidBodyConfiguration.cpp
```diff
@@ -184,6 +203,25 @@ namespace AzPhysics
...
+        if (auto* physicsSystem = AZ::Interface<AzPhysics::SystemInterface>::Get())
+        {
+            return !physicsSystem->GetDefaultSceneConfiguration().m_enableCcd || m_kinematic;
+        }
+        return m_kinematic;
+    }
+
+    AZStd::string_view RigidBodyConfiguration::GetCcdTooltip() const
+    {
+        return (m_kinematic) ? ccdDescriptionReadOnly : ccdDescription;
```
greerdv, line 217: nitpick - not a big deal but I think the brackets are redundant here and on line 222

greerdv (approved):
Looks good, nice work

Code/Framework/AzFramework/AzFramework/Physics/Configuration/RigidBodyConfiguration.cpp
AMZN-daimini: Makes sense now that I think about it again lol. One could replace it with a macro I guess, but given it's that simple I wouldn't overcomplicate it.

AMZN-daimini (approved):

Code/Framework/AzFramework/AzFramework/Physics/Configuration/RigidBodyConfiguration.cpp
AMZN-daimini: I usually add them as well cause they help me with readability, but yes they are not needed.

Code/Framework/AzFramework/AzFramework/Physics/Configuration/RigidBodyConfiguration.h
```diff
@@ -36,6 +36,9 @@ namespace AzPhysics
         void SetMassComputeFlags(MassComputeFlags flags);
 
         bool IsCcdEnabled() const;
+        bool CcdReadOnly() const;
+        AZStd::string_view GetCcdTooltip() const;
+        AZStd::string_view GetKinematicTooltip() const;
```
hultonha, line 41: I saw @AMZN-daimini's earlier comments and as the strings are known at compile time (constant) then returning a `string_view` is safe but it's definitely something to be careful about because if this code ever changes there's the risk of a dangling reference. Unless it's performance critical return by value is often safer but in this case I totally understand how it's justified.

hultonha (approved):
Really nice change, and this is definitely nicer than having the other option switch back and forward, the additional tooltip is a nice addition! 🙂

moraaar (approved):

Issue #10673, opened 2022-07-14
Migrate Multiplayer automation tests back from Sandox


lmbr-pip:
**Describe the bug**
All the Multiplayer tests were sandboxed due to instability in https://github.com/o3de/o3de/pull/10434. This issue is a placeholder to stabilize and migrate tests back to main AR runs.

**Assets required**
None

**Steps to reproduce**
1. Look at https://github.com/o3de/o3de/pull/10434
2. Notice all the tests have been sandboxed

**Expected behavior**
Critical networking tests are running in AR.

**Actual behavior**
Tests are not running AR leading to potentially ungated PRs that could destabilize networking/Multiplayer.

**Screenshots/Video**
If applicable, add screenshots and/or a video to help explain your problem.

**Found in Branch**
Name of or link to the branch where the issue occurs.

**Desktop/Device (please complete the following information):**
 - Device: [e.g. PC, Mac, iPhone, Samsung] 
 - OS: [e.g. Windows, macOS, iOS, Android]
 - Version [e.g. 10, Bug Sur, Oreo]
 - CPU [e.g. Intel I9-9900k , Ryzen 5900x, ]
 - GPU [AMD 6800 XT, NVidia RTX 3090]
 - Memory [e.g. 16GB]

**Additional context**

lmbr-pip:
@robhaco Assigning you to own ensuring tests are migrated back in by SIG-Network contributors. We should track this as an open issue until tests are re-enabled or we have new agreed test coverage plan for AR.

AMZN-Gene:
This has cropped up again. The Walk the Engine (WTE) build failed to launch the server when using CTRL+G in editor. A mainline test would have caught that.

AMZN-Gene:
Reopening, the last PR was a partial fix. It brought back 1 test (simple connection test) which doesn't rely on any scripting.

PR #13898, opened 2023-01-04
Multiplayer: Reintroducing Simple Connection Editor Test

AMZN-Gene:
Re-enable simple connection editor test now that the editor doesn't timeout as quickly and fail on slower machines.
It doesn't require any reading of logs or running scripts so if this fails it'll help narrow down the issue to being related to multiplayer.

Partial address #10673

## How was this PR tested?
Run pytest many times
`..\python\python.cmd -m pytest --build-directory .\build\windows\bin\profile\ .\Gem\PythonTests\Multiplayer\TestSuite_Main.py::TestAutomation`


cgalvan (approved):

byrcolin (approved):

AMZN-Gene:
Looks like this is still timing out. Increasing connection time helped to connect, but still taking too long. The time stamps in the logs are all the same, so hard to tell what's actually taking too long. Will address once I'm able to get a valid timestamp.
https://github.com/o3de/o3de/issues/13913

Issue #13803, opened 2022-12-21
Bug Report: Ubuntu 22.04 - AssetProcessor can't process assets because all AssetBuilders die


galibzon:
**Describe the bug**
The AP fails to process assets because all AssetBuilders die immediately.

**Assets required**
N/A

**Steps to reproduce**
Get latest "development" branch and try to compile at least with AutomatedTesting project.

**Expected behavior**
The AP builds all assets

**Actual behavior**
The AP fails as all AssetBuilders fail to start.

The first apparent bug is this error:
```
AssetBuilder: Module: Attempting to load module:/mnt/data/GIT/o3de/build/bin/profile/libSceneData.so
AssetBuilder: Module: Success!
AssetBuilder: Module: Attempting to load module:/mnt/data/GIT/o3de/build/bin/profile/libSceneBuilder.so
AssetBuilder: Module: Success!
AssetBuilder: C: [Trace] = /mnt/data/GIT/o3de/Gems/QtForPython/Code/Source/Platform/Linux/InitializeEmbeddedPyside2.h(46): 'static void *QtForPython::InitializeEmbeddedPyside2::LoadModule(const char *)'
AssetBuilder: C: [Type] = Trace::Error
AssetBuilder: E: QtForPython: Unable to load python library libpython3.10.so for Pyside2: libpython3.10.so: cannot open shared object file: No such file or directory
AssetBuilder: Module: Attempting to load module:/mnt/data/GIT/o3de/build/bin/profile/libSceneCore.so
AssetBuilder: Module: Success!
```
And later all AssetBuilders fail to start:
```
BuilderManager: Builder 28AE731F67574EEEAA26E79F88FD959F connected, connId: 1
BuilderManager: Builder started successfully
AssetProcessor: Scanning file system for changes...
FileWatcher: Using 1180 file watch handles.
AssetProcessor: File Change Monitoring started.
AssetProcessor: File system scan done.
AssetProcessor: Received 7327 files from the scanner.  Assessing...
AssetProcessor: Initial Scan complete, checking for missing files...
BuilderManager: Builder F6409D05E0314EE1B233EBE67AA92C8A connected, connId: 2
BuilderManager: Lost connection to builder F6409D05E0314EE1B233EBE67AA92C8A
BuilderManager: Builder 601524CCE7B04340AD106793CEC4A1EF connected, connId: 3
BuilderManager: Lost connection to builder 601524CCE7B04340AD106793CEC4A1EF
[...]

PR #13902, opened 2023-01-04
Restore rapidjson fixes

lawsonamzn:
## What does this PR do?

The fixes to the memory allocators in rapidjson caused crashes on linux which turned out to be from linker choosing symbols from the assimp static library which used a different version of the rapidjson library than the rest of the engine.

Because of those crashes, the changes were reverted.  

This PR reverts the revert, and then adds one additional change which fixes the problem, by putting o3de's rapidjson into a namespace (One which it used to be in, so as to retain backward compat).  This fixes the problem, as the linker will only use the o3de rapidjson symbols that come from that namespace instead of matching them with other versions of rapidjson embedded in other third party static libraries.

## How was it tested?
Tested manually on linux and windows by compiling all assets with it.

## Other notes
Note that it contains 2 commits - the first commit was the already approved (and then reverted) changes to rapidjson.  
The second commit is the fixup that moves it into its own namespace and aliases that namespace with the old namespace to make it backward compatible.  

This is basically a "proper" fix for https://github.com/o3de/o3de/issues/13803 that does not revert the rapidjson changes (since they are still valuable to have and prevent other problems, including memory usage)

### DIFF
--- a/Code/Framework/AzCore/AzCore/JSON/rapidjson.h
+++ b/Code/Framework/AzCore/AzCore/JSON/rapidjson.h
@@ -8,52 +8,97 @@
 
 #pragma once
 
-#include <AzCore/Memory/SystemAllocator.h>
+#include <AzCore/base.h>
 
-// Skip the error as we are including the recommended way.
-#define RAPIDJSON_SKIP_AZCORE_ERROR
-
-#if defined(AZ_ENABLE_TRACING)
-#   define RAPIDJSON_ASSERT(x) AZ_Assert(x, "Assert[rapidjson]: " #x)
-#else
-#   define RAPIDJSON_ASSERT(x) (void)(0)
+#if defined(RAPIDJSON_RAPIDJSON_H_)
+// if this happens, someone has alrady included the default rapidjson.h already, which is a bad idea as it won't
+// include any customizations such as memory management.
+#error vanilla rapidjson was included before including <AzCore/JSON/rapidjson.h>
 #endif
-#define RAPIDJSON_STATIC_ASSERT(x) static_assert(x, "Assert[rapidjson]: " #x)
-#define RAPIDJSON_HAS_CXX11_TYPETRAITS 1
-#define RAPIDJSON_HAS_CXX11_RVALUE_REFS 1
 
-namespace rapidjson_ly_internal
-{
-    template<typename T>
-    void Delete(T* instance)
+// Only override RAPIDJSON_ customization macros which are not already defined.
+// This allows the user to pre-define these before including O3DE if they want to use default values
+// or their own values.
+
+#if !defined(RAPIDJSON_ASSERT)
+#   if defined(AZ_ENABLE_TRACING)
+#       define RAPIDJSON_ASSERT(x) AZ_Assert(x, "Assert[rapidjson]: " #x)
+#   else
+#       define RAPIDJSON_ASSERT(x) (void)(0)
+#   endif
+#endif
+
+#if !defined(RAPIDJSON_STATIC_ASSERT)
+   #define RAPIDJSON_STATIC_ASSERT(x) static_assert(x, "Assert[rapidjson]: " #x)
+#endif
+
+#if !defined(RAPIDJSON_HAS_CXX11_TYPETRAITS)
+#   define RAPIDJSON_HAS_CXX11_TYPETRAITS 1
+#endif
+
+#if !defined(RAPIDJSON_HAS_CXX11_RVALUE_REFS)
+#   define RAPIDJSON_HAS_CXX11_RVALUE_REFS 1
+#endif
+
+// note for future maintainers:
+// RapidJSON allows you to customize the default allocator type for documents by overriding with RAPIDJSON_DEFAULT_ALLOCATOR.
+// However, customizing the default type of rapidjson::Document and rapidjson::Value in that way will cause incompatibilities
+// with external software libraries that use unmodified RapidJSON.   It is not thus recommended
+// to actually set the value of RAPIDJSON_DEFAULT_ALLOCATOR or redefine what a rapidjson::Document itself is.
+// The other way to get a 'custom' allocator installed in RapidJSON is to make your own Document/Value/Pointer type, since it's templated
+// on allocator type - for example, implement your own allocator and do
+// using MyDocument = rapidjson::GenericDocument<rapidjson::UTF8<char>, MyAllocator, MyAllocator> and then use MyDocument in your code
+// instead of RapidJSON::Document.  Note GenericDocument<> customized must use customized GenericValue<> and GenericPointer<>
+// or else you haven't acutally plugged into all the places where memory is allocated and will lose some.
+// 
+// Unfortunately, rapidjson::Pointer can't deal with these customized values or documents due to a bug in the code where functions
+// like Set/Get/Create use the default template parameter instead of deducing it:
+// ie, they are set up like somefunction<T>(GenericValue<char> val) which will only accept default GenericValue<char, CrtAllocator>.
+// correct form: somefunction<T>(GenericValue<char, T::AllocatorType> val, which would allow MyValue<char, MyAllocator> to be passed in.
+// thus, it is not recommended to use customized documents, values, or pointers either, (until that bug is fixed).
+
+// this leaves just one avenue for customization of memory of Rapidjson - overriding the macros RAPIDJSON_xxxx (new,delete, malloc, realloc..)
+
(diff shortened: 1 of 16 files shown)


Code/Framework/AzCore/AzCore/JSON/rapidjson.h
```diff
@@ -8,52 +8,97 @@
 
 #pragma once
 
-#include <AzCore/Memory/SystemAllocator.h>
+#include <AzCore/base.h>
 
-// Skip the error as we are including the recommended way.
-#define RAPIDJSON_SKIP_AZCORE_ERROR
+#if defined(RAPIDJSON_RAPIDJSON_H_)
+// if this happens, someone has alrady included the default rapidjson.h already, which is a bad idea as it won't
```
AMZN-stankowi, line 14: nitpick typo: alrady -> already

Code/Framework/AzCore/AzCore/JSON/rapidjson.h
```diff
@@ -8,52 +8,97 @@
...
+// 
+// Unfortunately, rapidjson::Pointer can't deal with these customized values or documents due to a bug in the code where functions
+// like Set/Get/Create use the default template parameter instead of deducing it:
+// ie, they are set up like somefunction<T>(GenericValue<char> val) which will only accept default GenericValue<char, CrtAllocator>.
+// correct form: somefunction<T>(GenericValue<char, T::AllocatorType> val, which would allow MyValue<char, MyAllocator> to be passed in.
+// thus, it is not recommended to use customized documents, values, or pointers either, (until that bug is fixed).
+
+// this leaves just one avenue for customization of memory of Rapidjson - overriding the macros RAPIDJSON_xxxx (new,delete, malloc, realloc..)
+
+// forward all allocations to our own allocators.  But preferrably without acutally having to drag our allocators into this header:
```
AMZN-stankowi, line 62: Nitpick typo: acutally -> actually

AMZN-stankowi (approved):

lawsonamzn:
In case anyone is interested in the gritty details, what was happening here on linux was that 3rd party library assimp.a was in the linker before azcore.a

Because both had their own embedded rapidjson in the same namespace, this meant that the symbol 
```
rapidjson::PoolAllocator<rapidjson::CrtAllocator>::Malloc(...)
``` 
existed in both libraries.   which means that the linker would choose the assimp.a one over the azcore.a one when it went hunting to resolve it.

Unfortunately the one for assimp had a different version of rapidjson, specifically, the code generated to deal with the PoolAllocator's struct layout was not the same.  This means that when the assimp Malloc function was used on an o3de allocator's struct, it would dereference the Chunk Size field as if it were a pointer, ie, ultimately trying to dereference the literal value (65536 + 8) and lead to a crash.

Code/Framework/AzCore/AzCore/JSON/rapidjson.h
lawsonamzn: I have decided to install a spell checker extension. :)

Code/Framework/AzCore/AzCore/JSON/rapidjson.h
```diff
@@ -86,13 +131,25 @@ namespace rapidjson_ly_internal
     #endif
 #endif
 
+// Push all of rapidjson into a different namespace (rapidjson_ly for backward compatibility, as this is what
+// it used to be at some point).  
+// The risk of using the default namespace is that during library search, the linker might find other versions
+// of rapidjson from other 3rd party libraries that have the same mangled function names and use those when 
+// linking the code in.  The problem with that is different versions of rapidjson have different struct layouts
+// and thus even though the signature of the function is the same, the actual implementation is not compatible,
+// leading to mystery crashes.
```
AMZN-alexpete, line 140: excellent comments

Code/Framework/AzCore/AzCore/JSON/rapidjson.h
```diff
@@ -86,13 +131,25 @@ namespace rapidjson_ly_internal
...
+// in the same compile unit somehow.
+namespace rapidjson = rapidjson_ly;
+
 #if AZ_TRAIT_JSON_CLANG_IGNORE_UNKNOWN_WARNING && defined(AZ_COMPILER_CLANG)
 #pragma clang diagnostic pop
-#endif
-
-// Allow our existing code to continue use "rapidjson::"
-#define rapidjson RAPIDJSON_NAMESPACE
+#endif
```
AMZN-alexpete, line 155: linux would appreciate a newline character at the end of this file

Code/Framework/AzCore/Platform/Common/VisualStudio/AzCore/Natvis/rapidjson.natvis
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <AutoVisualizer xmlns="http://schemas.microsoft.com/vstudio/debugger/natvis/2010">
 	<!-- rapidjson::GenericValue - basic support -->
-	<Type Name="rapidjson_ly::GenericValue&lt;*,*&gt;">
+	<Type Name="rapidjson::GenericValue&lt;*,*&gt;">
```
AMZN-alexpete, line 4: if we reverted to using the `rapidjson_ly` namespace, do we not want to use it here?

Code/Framework/AzCore/AzCore/JSON/rapidjson.h
lawsonamzn: weird.  There is one in my editor.  I'll add an extra.
image

Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/MetadataManager.cpp
```diff
@@ -28,7 +28,7 @@ namespace AzToolsFramework
 
     bool MetadataManager::GetValue(AZ::IO::PathView file, AZStd::string_view key, void* outValue, AZ::Uuid typeId)
     {
-        rapidjson_ly::Document value;
+        rapidjson::Document value;
```
AMZN-alexpete, line 31: ditto here - does your change above not revert to using the `rapidjson_ly` namespace?

Code/Framework/AzCore/Platform/Common/VisualStudio/AzCore/Natvis/rapidjson.natvis
AMZN-alexpete: nvm, I see we alias it using `namespace rapidjson = rapidjson_ly;`

Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/MetadataManager.cpp
AMZN-alexpete: nvm I see the alias `namespace rapidjson = rapidjson_ly;`

AMZN-alexpete (approved):
Just one missing newline, otherwise looks good.

Code/Framework/AzCore/Platform/Common/VisualStudio/AzCore/Natvis/rapidjson.natvis
lawsonamzn: Hmm, this makes me think it might be better to use 
```
#define rapidjson rapidjson_ly
```
instead of 'using'.  With 'using' its a type alias and its unclear if the debugger / natvis can tell. 

I'll do a quick experiemnt and see!

Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/MetadataManager.cpp
lawsonamzn: It aliases rapidjson to rapidjson_ly actually
```
namespace rapidjson = rapidjson_ly;
```

So in our code, either works.

burelc-amzn (approved):

Code/Framework/AzCore/Platform/Common/VisualStudio/AzCore/Natvis/rapidjson.natvis
lawsonamzn: Confirmed as broken - the debugger shows it as the underlying type (rapidjson_ly) but the natvis system isn't smart enough to understand that rapidjson == rapidjson_ly..  I'll fix the natvis file

lawsonamzn:
I rebased onto upstream/development to deal with the conflict.  The rest is unchanged - see the four changes 
(Base, The Fix, Spelling Fix, natvis fix)

lumberyard-employee-dm (approved):

junbo75 (approved):

PR #13920, opened 2023-01-05
Added an option to enable parallel OnPreRender processing in Multiplayer system

SergeyAMZN:
Added an option to enable parallel OnPreRender processing in Multiplayer system

## What does this PR do?

PreRender event notification is processed in Job threads now allowing to parallelize and boost server performance. 

## How was this PR tested?

Manual testing in Multiplayer sample with 100+ characters and rigid bodies.

preRenderParallel

TickVisibleParallel


Gems/EMotionFX/Code/Include/Integration/AnimGraphNetworkingBus.h
```diff
@@ -36,6 +36,8 @@ namespace EMotionFX
         virtual void UpdateActorExternal(float deltatime) = 0;
         virtual void SetNetworkRandomSeed(AZ::u64 seed) = 0;
         virtual AZ::u64 GetNetworkRandomSeed() const = 0;
+        virtual void SetActorThreadIndex(AZ::u32 threadIndex) = 0;
```
AMZN-Olex, line 39: Is this used in this PR? I don't see where.

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
```diff
@@ -114,6 +117,9 @@ namespace Multiplayer
     AZ_CVAR(AZ::TimeMs, bg_captureTransportPeriod, AZ::TimeMs{1000}, nullptr, AZ::ConsoleFunctorFlags::DontReplicate,
         "How often in milliseconds to record transport metrics.");
 
+    AZ_CVAR(bool, bg_parallelNotifyPreRender, true, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "bg_useJobsNotifyPreRender");
```
lmbr-pip, line 120: nit: could we add a meaningful description to this cvar? How will others know how to use this setting?

AMZN-Olex: We should default it to false as well for safety and turn it on in mps depot for testing.

Gems/EMotionFX/Code/Source/Integration/Components/AnimGraphComponent.cpp
```diff
@@ -433,6 +433,24 @@ namespace EMotionFX
...
+        }
+
+        AZ::u32 AnimGraphComponent::GetActorThreadIndex() const
+        {
+            if (m_actorInstance)
+            {
+                return m_actorInstance->GetThreadIndex();
+            }
+
+            return 0;
```
hultonha, line 451: question: Is 0 ever a valid thread index? Should we instead return a nullopt here or some invalid value?

hultonha (commented):
LGTM, once the updates @lmbr-pip and @AMZN-Olex suggested are implemented will definitely approve, nice work!

Gems/EMotionFX/Code/Include/Integration/AnimGraphNetworkingBus.h
SergeyAMZN: It's used in `NetworkAnimationComponent` which is in MP Sample, so has to be a separate PR..
Seems all of these functions are used only in that component

Gems/EMotionFX/Code/Source/Integration/Components/AnimGraphComponent.cpp
SergeyAMZN: Seems that's the default value in : 
`static ActorInstance* Create(Actor* actor, AZ::Entity* entity = nullptr, uint32 threadIndex = 0);`

It's not actually used anywhere, added more for it to be paired with the setter

hultonha: Okay cool understood, thanks for confirming (maybe could include a comment to say `// default thread index` perhaps?)

hultonha (approved):
Thanks for making the updates, LGTM (a before/after capture and performance numbers would still be super cool to see if possible). Thanks!

Gems/EMotionFX/Code/Include/Integration/AnimGraphNetworkingBus.h
SergeyAMZN: https://github.com/o3de/o3de-multiplayersample/pull/212

SergeyAMZN:
> Thanks for making the updates, LGTM (a before/after capture and performance numbers would still be super cool to see if possible). Thanks!

Thanks!
Yes, you can see it in the description above 17ms -> 3.5ms

TickVisibleParallel

Gems/EMotionFX/Code/Source/Integration/Components/AnimGraphComponent.cpp
amzn-rhhong: Some question:
1. Does this actually work setting the threadIndex directly hence bypasses the multiThreadScheduler in emfx? Sounds a little too good to be true.. Perhaps the networking system in MP is calling those in a different multi threaded job system?
2. Shouldn't this be part of actor component instead of anim graph component, since it's directly changing the actor instance?
3. We might want to have this on the editor component as well (actor component vs editor actor component)

SergeyAMZN: 1. MultiThreadScheduler isn't used here since in multiplayer each actor is updated separately via `AnimGraphComponent::UpdateActorExternal`. 

2. Similarly to how `AnimGraphComponent::UpdateActorExternal` is called and changing the actor instance, we do the same with thread index. 
If you feel `UpdateActorExternal` should be removed from `AnimGraphComponent` to something like `ActorComponent::UpdateExternal`, then of course we can move `SetThreadIndex` too. 
Or in fact maybe just pass the thread index as an argument to `UpdateActorExternal`?

3. if we UpdateActorExternal in the Editor, could have it too. Otherwise if there's no manual update, we can simply use MultiThreadScheduler which does set thread index internally.

amzn-rhhong: Regarding 2 I definitely think they should be part of the actor component.

amzn-rhhong (approved):

PR #13922, opened 2023-01-05
Add the second mode to BC1 runtime decompression.

mbalfour-amzn:
## What does this PR do?

BC1 compression has two modes - if color0 > color1, then the 4 colors are color0, color1, 1/3 lerp, 2/3 lerp. However, there's a second mode as well, where if color0 <= color1, then the 4 colors should be color0, color1, 1/2 lerp, and transparent black.

(See https://learn.microsoft.com/en-us/windows/win32/direct3d10/d3d10-graphics-programming-guide-resources-block-compression for documentation on the two modes)

This fix adds the second mode.

## How was this PR tested?

Loaded a BC1-compressed image that was using both modes into an image gradient and verified that the results looked much more correct.

### DIFF
--- a/Gems/Atom/RPI/Code/Include/Atom/RPI.Public/BlockCompression.h
+++ b/Gems/Atom/RPI/Code/Include/Atom/RPI.Public/BlockCompression.h
@@ -20,8 +20,8 @@ namespace AZ
         // The 2-bit indices represent the following:
         // 00 - use color0
         // 01 - use color1
-        // 10 - use 2/3 color0 and 1/3 color1
-        // 11 - use 1/3 color0 and 2/3 color1
+        // 10 - if color0 > color1, use 2/3 color0 and 1/3 color1, else use 1/2 color0 and 1/2 color1
+        // 11 - if color0 > color1, use 1/3 color0 and 2/3 color1, else use transparent black
         struct BC1Block
         {
             uint16_t m_color0;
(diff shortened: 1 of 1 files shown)


Gems/Atom/RPI/Code/Include/Atom/RPI.Public/BlockCompression.h
```diff
@@ -78,11 +78,13 @@ namespace AZ
                     return extractColor(m_color1);
                     break;
                 case 2:
-                    // Index 2 uses 2/3 of color 0 and 1/3 of color 1
-                    return extractColor(m_color0).Lerp(extractColor(m_color1), (1.0f / 3.0f));
+                    // Index 2 either uses 2/3 of color 0 and 1/3 of color 1 or 1/2 of color 0 and 1/2 of color 1
+                    return (m_color0 > m_color1) ? extractColor(m_color0).Lerp(extractColor(m_color1), (1.0f / 3.0f))
```
rgba16f, line 82: is it worth doing the color0 > color1 check at block init rather than for every pixel decode?

mbalfour-amzn: There is no actual block init to do it at, the struct doesn't get constructed / destroyed, it gets overlaid on top of existing data.

mbalfour-amzn: The usage in RPIUtils.cpp looks like this:
```
                case AZ::RHI::Format::BC1_UNORM:
                    {
                        auto actualMem = reinterpret_cast<const BC1Block*>(mem);
                        return actualMem[indices.first].GetBlockColor(indices.second).GetElement(componentIndex);
                    }
```

rgba16f (approved):

VickyAtAZ (approved):

Issue #13873, opened 2023-01-03
Frequent Editor Crash using Component Mode (White Box)


AMZN-Gene:
**Describe the bug**
The editor commonly crashes when using the whitebox tool (and likely any component that has an alternative "component mode")

**Steps to reproduce**
1. Open a level
2. Create a white box
3. Enter white-box component editor mode and edit the white box
4. Hit escape to exit component mode
5. Select another entity 

**Expected behavior**
Editor doesn't crash

**Actual behavior**
Editor crashes
`
>	EditorLib.dll!AzToolsFramework::ComponentModeFramework::ComponentModeSwitcher::RemoveNonCommonComponents(const AZ::Entity & entity) Line 339	C++
 	EditorLib.dll!AzToolsFramework::ComponentModeFramework::ComponentModeSwitcher::UpdateSwitcher(const AZStd::vector<AZ::EntityId,AZStd::allocator> & newlySelectedEntityIds, const AZStd::vector<AZ::EntityId,AZStd::allocator> & newlyDeselectedEntityIds) Line 143	C++
 	[Inline Frame] EditorLib.dll!AZStd::Internal::INVOKE(void(AzToolsFramework::ToolsApplicationEvents::*)(const AZStd::vector<AZ::EntityId,AZStd::allocator> &, const AZStd::vector<AZ::EntityId,AZStd::allocator> &) &) Line 177	C++
 	[Inline Frame] EditorLib.dll!AZStd::invoke(void(AzToolsFramework::ToolsApplicationEvents::*)(const AZStd::vector<AZ::EntityId,AZStd::allocator> &, const AZStd::vector<AZ::EntityId,AZStd::allocator> &) &) Line 54	C++
 	[Inline Frame] EditorLib.dll!AZ::EBusEventProcessingPolicy::Call(void(AzToolsFramework::ToolsApplicationEvents::*)(const AZStd::vector<AZ::EntityId,AZStd::allocator> &, const AZStd::vector<AZ::EntityId,AZStd::allocator> &) &) Line 437	C++
[...]

PR #13923, opened 2023-01-05
Bug fix: Switcher crash on undo + select other component

amzn-ahmadkrm:

## What does this PR do?
Fixes issue: https://github.com/o3de/o3de/issues/13873

The problem was that when a user hits Undo the component is destroyed and recreated but the pointer does not update. The crash happens when it was trying to access the pointer when another entity was added to the selection. This fix stops using a pointer for the component and uses FindComponent instead.

## How was this PR tested?

Manual testing.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeSwitcher.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeSwitcher.cpp
@@ -22,9 +22,7 @@ namespace AzToolsFramework::ComponentModeFramework
     ComponentData::ComponentData(AZ::EntityComponentIdPair pairId)
         : m_pairId(pairId)
     {
-        AZ::Entity* entity = AzToolsFramework::GetEntityById(pairId.GetEntityId());
-
-        AZ::Component* component = entity->FindComponent(pairId.GetComponentId());
+        AZ::Component* component = GetComponent();
         m_componentName = AzToolsFramework::GetFriendlyComponentName(component);
 
         AzToolsFramework::EditorRequestBus::BroadcastResult(
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeSwitcher.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeSwitcher.cpp
@@ -22,7 +22,7 @@ namespace AzToolsFramework::ComponentModeFramework
     ComponentData::ComponentData(AZ::EntityComponentIdPair pairId)
         : m_pairId(pairId)
     {
-        AZ::Component* component = GetComponent();
+        AZ::Component* component = FindComponent();
         m_componentName = AzToolsFramework::GetFriendlyComponentName(component);
 
         AzToolsFramework::EditorRequestBus::BroadcastResult(
@@ -32,7 +32,7 @@ namespace AzToolsFramework::ComponentModeFramework
             component);
     }
 
-    AZ::Component* ComponentData::GetComponent() const
+    AZ::Component* ComponentData::FindComponent() const
     {
         AZ::Entity* entity = AzToolsFramework::GetEntityById(m_pairId.GetEntityId());
         return entity->FindComponent(m_pairId.GetComponentId());
@@ -183,7 +183,7 @@ namespace AzToolsFramework::ComponentModeFramework
                 m_addedComponents,
                 [newComponentData](const ComponentData& componentInfo)
                 {
-                    return componentInfo.GetComponent()->RTTI_GetType() == newComponentData.GetComponent()->RTTI_GetType();
+                    return componentInfo.FindComponent()->RTTI_GetType() == newComponentData.FindComponent()->RTTI_GetType();
                 });
             componentDataIt == m_addedComponents.end())
         {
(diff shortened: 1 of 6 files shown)


Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeSwitcher.cpp
```diff
@@ -22,17 +22,20 @@ namespace AzToolsFramework::ComponentModeFramework
...
-            m_component->GetUnderlyingComponentType(),
-            m_component);
+            component->GetUnderlyingComponentType(),
+            component);
+    }
+
+    AZ::Component* ComponentData::GetComponent() const
+    {
+        AZ::Entity* entity = AzToolsFramework::GetEntityById(m_pairId.GetEntityId());
+        return entity->FindComponent(m_pairId.GetComponentId());
```
cgalvan, line 38: Nit: Maybe add an assert here to validate the `entity`

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeSwitcher.cpp
```diff
@@ -135,7 +138,7 @@ namespace AzToolsFramework::ComponentModeFramework
                 {
                     // if two or more entities are selected, ensure the only components
                     // that remain on the switcher are common to all entities
-                    if (selectedEntityIds.size() > 1 && !m_addedComponents.empty())
+                    if (selectedEntityIds.size() > 1)
```
cgalvan, line 141: Does this `!m_addedComponents.empty()` check still need to go away?

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeSwitcher.h
```diff
@@ -34,11 +34,13 @@ namespace AzToolsFramework
 
             // Member variables
             AZ::EntityComponentIdPair m_pairId; //!< Id of entity component pair.
-            AZ::Entity* m_entity = nullptr; //!< Pointer to entity associated with pairId.
-            AZ::Component* m_component = nullptr; //!< Pointer to component associated with pairId.
+            //AZ::Entity* m_entity = nullptr; //!< Pointer to entity associated with pairId.
+            //AZ::Component* m_component = nullptr; //!< Pointer to component associated with pairId.
```
cgalvan, line 38: Please remove these instead of commenting them out

cgalvan (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeSwitcher.cpp
amzn-ahmadkrm: The function AzToolsFramework::GetEntityById does the assert, it's from EdtitorEntityHelpers

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeSwitcher.cpp
amzn-ahmadkrm: I removed this because in the case where an entity has no components on the switcher and another entity is selected it doesn't go through the block to find which components are common between them.

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeSwitcher.h
amzn-ahmadkrm: Thank you for this :)

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeSwitcher.cpp
```diff
@@ -22,17 +22,20 @@ namespace AzToolsFramework::ComponentModeFramework
...
         AzToolsFramework::EditorRequestBus::BroadcastResult(
             m_iconPath,
             &AzToolsFramework::EditorRequestBus::Events::GetComponentEditorIcon,
-            m_component->GetUnderlyingComponentType(),
-            m_component);
+            component->GetUnderlyingComponentType(),
+            component);
+    }
+
+    AZ::Component* ComponentData::GetComponent() const
```
AMZN-Gene, line 35: Suggestion: Change method name to FindComponent()
Getters() generally imply a little/no performance cost method of returning a cached value.

AMZN-Gene (approved):

mbalfour-amzn (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeSwitcher.h
```diff
@@ -34,11 +34,11 @@ namespace AzToolsFramework
 
             // Member variables
             AZ::EntityComponentIdPair m_pairId; //!< Id of entity component pair.
-            AZ::Entity* m_entity = nullptr; //!< Pointer to entity associated with pairId.
-            AZ::Component* m_component = nullptr; //!< Pointer to component associated with pairId.
             AZStd::string m_componentName; //!< Friendly name of component.
             AZStd::string m_iconPath; //!< Path of component icon.
             ViewportUi::ButtonId m_buttonId; //!< Button Id of switcher component.
+
+            AZ::Component* FindComponent() const;
```
hultonha, line 41: nit: It's usually good to return a `const` pointer from a const member function (or have an overload so you have...)

```c++
AZ::Component* FindComponent();
const AZ::Component* FindComponent() const;
```

If you just need the `const` one then don't bother with the non-const overload.

This is referred to as 'const broken', this page has a nice overview - https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/c++/const.md#classes-of-const-in_correctness

hultonha (approved):

hultonha (approved):
Nice! And great tidy-up changes to make things const correct 🙂

Issue #13930, opened 2023-01-05
Several Atom automated test are intermittently failing on Linux.


jromnoa:
**Describe the bug**
There has been a recent history in the past 50 runs of AR where some Atom tests are intermittently failing, but only on Linux. Sufficient testing time has not been spent on root causing the issues on Linux since intermittent failures are difficult and timely to reproduce, but important to catch and fix. On Windows, the tests appear to be passing without any issues and the problematic tests before have been disabled until they can be fixed (see https://github.com/o3de/o3de/issues/13819 and https://github.com/o3de/o3de/issues/13818 for the tests that were causing intermittent failures and disabled until the fix can be applied).

**Assets required**
N/A

**Steps to reproduce**
Review the logs in AR, for an example of the latest Atom intermittent test failure on Linux see here: https://jenkins.build.o3de.org/blue/rest/organizations/jenkins/pipelines/O3DE/pipelines/development/runs/3705/nodes/987/log/ 
The level loads test can be observed failing on Linux as well, so also will be disabled while these issues are investigated.

1. Have a Linux machine setup and built with all assets processed for the AutomatedTesting project.
2. Open a terminal window and use any of the following command 100+ times until a failure occurs:
- For **TestSuite_Main_Null_Render_Component_02.py**
```py
/home/new_user/git/o3de/python/python.sh -m pytest -vv -s --build-directory="/home/new_user/git/o3de/build/linux/bin/profile" /home/new_user/git/o3de/AutomatedTesting/Gem/PythonTests/Atom/TestSuite_Main_Null_Render_Component_02.py
```
- For **TestSuite_Main_Null_Render_Component_03.py**
```py
/home/new_user/git/o3de/python/python.sh -m pytest -vv -s --build-directory="/home/new_user/git/o3de/build/linux/bin/profile" /home/new_user/git/o3de/AutomatedTesting/Gem/PythonTests/Atom/TestSuite_Main_Null_Render_Component_03.py
```
- For **TestSuite_Periodic_Null_Render_Level_Loads.py**
```py
/home/new_user/git/o3de/python/python.sh -m pytest -vv -s --build-directory="/home/new_user/git/o3de/build/linux/bin/profile" /home/new_user/git/o3de/AutomatedTesting/Gem/PythonTests/Atom/TestSuite_Periodic_Null_Render_Level_Loads.py
```

**Expected behavior**
The tests run and pass without issue.

**Actual behavior**
[...]

Issue #13819, opened 2022-12-22
Hair component python automated tests for Atom can fail due to an assert in C:\git\o3de\Code\Framework\AzFramework\AzFramework/Asset/GenericAssetHandler.h


jromnoa:
**Describe the bug**
1. The main suite null render 02 tests for Atom had to be disabled due to an intermittent failure occurring every ~20 or so builds: https://github.com/o3de/o3de/pull/13679
2. An investigation was done to root cause the intermittent failure and the Mesh component along with the Hair component portions of the test were the main culprits. They both were sandboxed.
3. This ticket is to address the bug with the Hair component.
4. The Mesh component ticket is here: https://github.com/o3de/o3de/issues/13818 

**Assets required**
No special assets required just need AutomatedTesting project built and the assets all finished processing before running the test.

**Steps to reproduce**
1. After building the AutomatedTesting project and processing all of the assets, run the AutomatedTesting::Atom_Main_Null_Render_02 test 20+ times until the test fails due to an assert:
`C:\git\o3de\python\runtime\python-3.10.5-rev1-windows\python\python.exe -m pytest -vv -s --build-directory="C:\git\o3de\build\bin\profile" C:\git\o3de\AutomatedTesting\Gem\PythonTests\Atom\TestSuite_Main_Null_Render_Component_02.py`
2. Observe the output after each test run.
3. NOTE: After a PR update to move the test to sandbox, you may need to use this command instead:
`C:\git\o3de\python\runtime\python-3.10.5-rev1-windows\python\python.exe -m pytest -vv -s --build-directory="C:\git\o3de\build\bin\profile" C:\git\o3de\AutomatedTesting\Gem\PythonTests\Atom\TestSuite_Sandbox.py -k AtomEditorComponents_HairAdded`

**Expected behavior**
The Atom_Main_Null_Render_02 test passes when run.

**Actual behavior**
1/20 times on average, the Atom_Main_Null_Render_02 test fails due to an assert:
```
<14:37:24> (System) -
==================================================================
<14:37:24> (System) - Trace::Assert
[...]

Issue #13818, opened 2022-12-22
Mesh component python automated tests for Atom can fail due to an assert in C:\git\o3de\Code\Framework\AzCore\AzCore/EBus/Internal/CallstackEntry.h


jromnoa:
**Describe the bug**
1. The main suite null render 02 tests for Atom had to be disabled due to an intermittent failure occurring every ~20 or so builds: https://github.com/o3de/o3de/pull/13679
2. An investigation was done to root cause the intermittent failure and the Mesh component along with the Hair component portions of the test were the main culprits. They both were sandboxed as part of 
3. This ticket is to address the bug with the Mesh component.

**Assets required**
No special assets required just need AutomatedTesting project built and the assets all finished processing before running the test.

**Steps to reproduce**
1. After building the AutomatedTesting project and processing all of the assets, run the AutomatedTesting::Atom_Main_Null_Render_02 test 20+ times until the test fails due to an assert:
`C:\git\o3de\python\runtime\python-3.10.5-rev1-windows\python\python.exe -m pytest -vv -s --build-directory="C:\git\o3de\build\bin\profile" C:\git\o3de\AutomatedTesting\Gem\PythonTests\Atom\TestSuite_Main_Null_Render_Component_02.py`
2. Observe the output after each test run.

**Expected behavior**
The Atom_Main_Null_Render_02 test passes when run.

**Actual behavior**
1/20 times on average, the Atom_Main_Null_Render_02 test fails due to an assert:
```
==================================================================
<14:37:36> (System) - Trace::Assert
 C:\git\o3de\Code\Framework\AzCore\AzCore/EBus/Internal/CallstackEntry.h(86): (35476) '__cdecl AZ::Internal::CallstackEntry<class AZ::SystemTickEvents,class AZ::SystemTickEvents>::CallstackEntry(class AZ::EBus<class AZ::SystemTickEvents,class AZ::SystemTickEvents>::Context *,const struct AZ::NullBusId *)'
<14:37:36> (System) - Bus const char *__cdecl AZ::EBus<class AZ::SystemTickEvents,class AZ::SystemTickEvents>::GetName(void) has multiple threads in its callstack records. Configure MutexType on the bus, or don't send to it from multiple threads
<14:37:36> (System) - ------------------------------------------------
[...]

PR #13931, opened 2023-01-05
Skip some failing Atom Linux tests while bug is investigated (Windows now passes).

jromnoa:
## What does this PR do?
1. This PR is to temporarily disable some Atom tests on Linux since they intermittent fail.
2. It also re-enables the level loads test on Windows since it now appears to be only failing on Linux.
3. The tests do not intermittently fail on Windows (thoroughly tested already and the problematic Windows tests have bugs now in https://github.com/o3de/o3de/issues/13818 + https://github.com/o3de/o3de/issues/13819).

## How was this PR tested?
1. No testing needed, this is just re-enabling the level loads test for Windows and disabling 3 different test files for Linux so the failure can be root caused.
2. See this main GHI tracking ticket for all updates on these tests for Linux: https://github.com/o3de/o3de/issues/13930

### DIFF
--- a/AutomatedTesting/Gem/PythonTests/Atom/TestSuite_Main_Null_Render_Component_02.py
+++ b/AutomatedTesting/Gem/PythonTests/Atom/TestSuite_Main_Null_Render_Component_02.py
@@ -4,16 +4,15 @@ For complete copyright and license terms please see the LICENSE at the root of t
 
 SPDX-License-Identifier: Apache-2.0 OR MIT
 """
-import sys
-
 import pytest
 
+import ly_test_tools
 from ly_test_tools.o3de.editor_test import EditorBatchedTest, EditorTestSuite
 
 
 @pytest.mark.parametrize("project", ["AutomatedTesting"])
 @pytest.mark.parametrize("launcher_platform", ['windows_editor'])
-@pytest.mark.skipif(sys.platform.startswith("linux"), reason="https://github.com/o3de/o3de/issues/13930")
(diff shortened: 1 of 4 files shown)


galibzon (approved):

AutomatedTesting/Gem/PythonTests/Atom/TestSuite_Main_Null_Render_Component_02.py
```diff
@@ -4,13 +4,16 @@
...
+import sys
+
 import pytest
 
 from ly_test_tools.o3de.editor_test import EditorBatchedTest, EditorTestSuite
 
 
 @pytest.mark.parametrize("project", ["AutomatedTesting"])
 @pytest.mark.parametrize("launcher_platform", ['windows_editor'])
+@pytest.mark.skipif(sys.platform.startswith("linux"), reason="https://github.com/o3de/o3de/issues/13930")
```
smurly, line 16: in other suites we use `@pytest.mark.skipif(ly_test_tools.LINUX, reason="https://github.com/o3de/o3de/issues/13930")` along with it's `import ly_test_tools`

jromnoa: oops, well im going to leave it as is for now but will remember to use this method instead in the future like in the other suites

jromnoa: reason being is i dont want the pipeline to get blocked by one of these intermittent failures

cgalvan (approved):

smurly (approved):

PR #13957, opened 2023-01-05
{lyn16847} adding import skeleton logic into scene pipeline

jackalbe:
## What does this PR do?

The Helios team has already updated the AssImp library to 5.2.5 so that the scene pipeline but we are now turning on loading the “skeleton” flag. This will require a version bump to reprocess all the source scene (e.g. FBX and GLTF) files. This means that the next time a user fetches/pulls code then all the source scene files will be processed again. This should not cause any problems other than a longer load time to get back into the Editor (to finish Asset Processor).

### DIFF
--- a/Code/Tools/SceneAPI/SDKWrapper/AssImpSceneWrapper.cpp
+++ b/Code/Tools/SceneAPI/SDKWrapper/AssImpSceneWrapper.cpp
@@ -6,6 +6,7 @@
  *
  */
 #include <AzCore/Debug/Trace.h>
+#include <AzCore/Settings/SettingsRegistry.h>
 #include <AzToolsFramework/Debug/TraceContext.h>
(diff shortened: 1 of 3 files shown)


junbo75 (approved):

cgalvan (approved):

AMZN-stankowi:
Should this be hidden behind a settings registry flag that's disabled by default, until we do a thorough testing pass on this feature?

jackalbe:
> Should this be hidden behind a settings registry flag that's disabled by default, until we do a thorough testing pass on this feature?

That sounds good to me!

AMZN-stankowi (approved):

junbo75 (approved):

cgalvan (approved):

hultonha (approved):

PR #13964, opened 2023-01-06
DPE: Add support for GenericValueList attribute

amzn-tmryan:
**Description**
This PR adds support to the DPE for the GenericValueList attribute which is used to specify a vector of data and its descriptions for use in a property editor combobox. It also unifies the EnumValue/EnumValues and GenericValue/GenericValueList attribute usage within DPE related code.

Changes of note:
- Added `GenericValueListAttributeDefinition<T>` which handles extracting data from attribute nodes and creation of attribute nodes using previously extracted data
- Added virtual function `PropertyHandlerBase::RegisterWithPropertySystem` which allows derived property handler classes to register their own attribute elements (Nodes, AttributeDefinitions, etc) with the `PropertyEditorSystem`
- Changed `PropertyEditorSystem`'s attribute metadata container to use an `unordered_multimap` for storage of registered `AttributeDefintionInterface`; this fixed an issue with only one attribute definition to be stored per attribute per node name
- `PropertyEditorAPI_Internals::SetValueFromDom` is now using `PropertyEditorSystem::EnumerateRegisteredAttributes` to find the correct `AttributeDefintionInterface` amongst the templated definitions per attribute name that have been registered
- Removed `EnumValuesAttributeDefinition` and converted `EnumValueAttributeDefinition` into `GenericValueAttributeDefinition` to cut down on the special cases needed to support legacy attributes

**Testing**
- Added cases to the DPE debug standalone for the various ways that EnumValues can be reflected
- Verified the reflected values using this attribute in the DPE debug standalone app work properly
- Verified assets that can open in the AssetEditor that have comboboxes reflected with GenericValueList/EnumValues/EnumValue are all showing correctly when the DPE is enabled
- Verified that performance with and without this PR + its 2 dependent PRs is very close to the same: 
  - Used the Very Sleep tool to gather performance data while performing roughly the same steps of opening 3 different types of files, all of which contain comboboxes, in the AssetEditor and then opening the CVar editor
  - Time taken to complete DPE and DOM related functions was very close to the same with and without changes (some exclusive times were higher before the changes which is probably coincidental)
  - Total time (inclusive) taken to generate contents from the Adapter and set DOM values in the DPE view was similar both with and without changes

Signed-off-by: amzn-tmryan [104796591+amzn-tmryan@users.noreply.github.com](mailto:104796591+amzn-tmryan@users.noreply.github.com)


Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/DocumentSchema.h
```diff
@@ -312,6 +312,74 @@ namespace AZ::DocumentPropertyEditor
         AZStd::optional<EnumValuesContainer> DomToValue(const Dom::Value& value) const override;
     };
 
+    template<typename GenericValueType>
+    class GenericValueListAttributeDefinition final
+        : public AttributeDefinition<AZStd::vector<AZStd::pair<GenericValueType, AZStd::string>>>
```
lsemp3d, line 317: Perhaps alias this bit: `AZStd::vector<AZStd::pair<GenericValueType, AZStd::string>>` it's difficult to read

lsemp3d (approved):

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/DocumentSchema.h
amzn-tmryan: I was originally aliasing this stuff before the `EnumValue` PR, but when we have to use templating for the AttributeDefinitions that need to support multiple types (int, unsigned int, etc) the alias won't work. It's because the alias would need the template as well.

If there is a way to do that then I'll definitely get this changed here, and for similar types.

amzn-tmryan: I'll check this out in the morning and update the PR if I can get it working.

amzn-tmryan: I tried the approach above but the code gets more complicated and difficult to read. If this were lower level code then I think it would look normal (it's often harder to read) but I think I'll leave this alone for now.

alexmontAmazon (approved):

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/DocumentSchema.h
lumberyard-employee-dm: You could go with an alias called ComboBoxContainer for the AZStd::vector
```c++
template <typename T>
using ComboBoxContainer = AZStd::vector<AZStd::pair<T, AZStd::string>;
```

amzn-tmryan:
Unified EnumValues and GenericValueList as well as EnumValue and GenericValue inside of the DPE DOM.

This PR now requires PR https://github.com/o3de/o3de/pull/14155 to merge before it will pass AR.

cgalvan (approved):

amzn-tmryan:
Rebased onto latest development, updated some comments/warning messages, and integrated with latest updates to the non-void* instance AttributeReader changes.

PR #13966, opened 2023-01-06
Default prefab adding prefab adds mesh group

AMZN-stankowi:
## What does this PR do?

Without this change, when you add a default prefab group to a scene manifest in the scene settings UI, it didn't automatically update the interface to show the read-only mesh groups used by that.

With this change, the read only mesh groups are added when you add a default prefab group.

The mesh groups are actually added when processing the scene manifest based on the default prefab group being enabled, but they're useful in the UI to provide a breadcrumb trail back to why product assets are generated.

image

## How was this PR tested?

Manual testing, plus new automated test


Gems/Prefab/PrefabBuilder/PrefabGroup/PrefabGroupBehavior.cpp
```diff
@@ -215,6 +214,28 @@ namespace AZ::SceneAPI::Behaviors
         return Events::ProcessingResult::Success;
     }
 
+    Events::ProcessingResult PrefabGroupBehavior::ExportEventHandler::UpdateManifest(
```
jackalbe, line 217: so the logic is now "create the prefab group" but not the mesh groups for the prefab until InitializeObject() ?

Gems/Prefab/PrefabBuilder/PrefabGroup/PrefabGroupBehavior.cpp
```diff
@@ -123,12 +126,10 @@ namespace AZ::SceneAPI::Behaviors
...
 
-    Events::ProcessingResult PrefabGroupBehavior::ExportEventHandler::UpdateManifest(
+    Events::ProcessingResult PrefabGroupBehavior::ExportEventHandler::UpdateSceneForPrefabGroup(
         Containers::Scene& scene,
-        ManifestAction action,
-        [[maybe_unused]] RequestingApplication requester)
-    {
-        // this toggle makes constructing default mesh groups and a prefab optional
+        ManifestAction action)
+    { // this toggle makes constructing default mesh groups and a prefab optional
```
jackalbe, line 132: this comment does not make sense to me... what "toggle" is it referring to?

Gems/Prefab/PrefabBuilder/PrefabGroup/PrefabGroupBus.h
```diff
@@ -49,6 +49,7 @@ namespace AZ::SceneAPI
         using ManifestUpdates = AZStd::vector<AZStd::shared_ptr<DataTypes::IManifestObject>>;
 
         virtual AZStd::optional<ManifestUpdates> GeneratePrefabGroupManifestUpdates(const Scene& scene) const = 0;
+        virtual AZStd::vector<AZStd::shared_ptr<DataTypes::IManifestObject>> GeneratePrefabMeshGroups(const Scene& scene) const = 0;
```
jackalbe, line 52: I wonder is this should be called "Generate**Default**PrefabMeshGroups" since it is going to make all those default_XYZ mesh group names

Gems/Prefab/PrefabBuilder/PrefabGroup/PrefabGroupBehavior.cpp
AMZN-stankowi: I didn't change the existing logic. The mesh groups were previously only created when the manifest was created.

Now, in the UI, they get created also when you add a default prefab if there was none previously. It's purely visual for the UI, they were getting created before when the scene settings file was processed after saving, you just didn't see them in the UI until you closed/reopened it.

Gems/Prefab/PrefabBuilder/PrefabGroup/PrefabGroupBehavior.cpp
AMZN-stankowi: that comment was already there, it just ended up moved. I deleted the comment.

Gems/Prefab/PrefabBuilder/PrefabGroup/PrefabGroupBus.h
AMZN-stankowi: Sure no problem, done

cgalvan (approved):

Code/Tools/SceneAPI/SceneUI/SceneWidgets/ManifestWidgetPage.cpp
```diff
@@ -469,6 +469,35 @@ namespace AZ
...
+                for (auto& object : objects)
+                {
+                    if (!SupportsType(object))
+                    {
+                        continue;
+                    }
+                    if (!manifest.AddEntry(object))
+                    {
+                        AZ_Assert(false, "Unable to add new object to manifest.");
+                    }
```
jackalbe, line 495: If AddObjects can be executed from a Python script, then this AZ_Assert() will crash the Editor based on a script error, right? Is that the behavior we want here?

junbo75 (approved):

Code/Tools/SceneAPI/SceneUI/SceneWidgets/ManifestWidgetPage.cpp
AMZN-stankowi: It won't crash, but the whole tool will be in an unstable state with information not in the manifest that is in the widget page.

I'll downgrade it to an error, but if this happens, then things are broken and you'll want to re-start your Editor or find an engineer to debug it.

PR #13969, opened 2023-01-09
Action Manager | Reflect White Box Component Mode to work correctly with the new setup

AMZN-daimini:
## What does this PR do?

Reflects the White Box Component Mode to Serialize Context:
This allows it to be picked up by the new setup that allows Component Modes to register actions and switch them correctly in the file menu of the Editor using the new Action Manager.

This mode has 3 sub-modes, which required some extra setup (the White Box modes registers 3 additional Action Context modes and does the switching internally).

I also replaced the logic to change between the Transform types since menu items piggybacked off the Viewport UI.

## How was this PR tested?

Manual Testing


cgalvan (approved):

Gems/WhiteBox/Code/Source/EditorWhiteBoxComponentMode.cpp
```diff
@@ -303,6 +356,20 @@ namespace WhiteBox
         m_intersectionAndRenderData = {};
         m_currentSubMode = SubMode::Default;
         SetViewportUiClusterActiveButton(m_modeSelectionClusterId, m_defaultModeButtonId);
+
+        // Change sub-mode to default at the next frame to go after the automated mode switching in ComponentModeActionHandler.
+        QTimer::singleShot(
+            0,
```
hultonha, line 362: update: I believe it's preferred to use the TickBus next frame call instead of `QTimer::singleShot` (I don't remember the syntax off the top of my head but I know @mbalfour-amzn has suggested this in the past). I think that might be nicer here (and could be easier to test if we want to as well)

Gems/WhiteBox/Code/Source/EditorWhiteBoxComponentMode.cpp
```diff
@@ -311,6 +378,12 @@ namespace WhiteBox
         m_intersectionAndRenderData = {};
         m_currentSubMode = SubMode::EdgeRestore;
         SetViewportUiClusterActiveButton(m_modeSelectionClusterId, m_edgeRestoreModeButtonId);
+
+        // Set the Action Context Mode
```
hultonha, line 382: nit: Should we move this comment down to line 386?

Gems/WhiteBox/Code/Source/EditorWhiteBoxComponentMode.cpp
```diff
@@ -319,6 +392,12 @@ namespace WhiteBox
         m_intersectionAndRenderData = {};
         m_currentSubMode = SubMode::Transform;
         SetViewportUiClusterActiveButton(m_modeSelectionClusterId, m_transformModeButtonId);
+
+        // Set the Action Context Mode
```
hultonha, line 396: nit: Same here, move comment to 400?

Gems/WhiteBox/Code/Source/SubComponentModes/EditorWhiteBoxDefaultMode.cpp
```diff
@@ -44,6 +48,10 @@ namespace WhiteBox
     static const char* const HideEdgeUndoRedoDesc = "Hide an edge to merge two connected polygons together";
     static const char* const HideVertexUndoRedoDesc = "Hide a vertex to merge two connected edges together";
 
+    static constexpr AZStd::string_view EditorMainWindowActionContextIdentifier = "o3de.context.editor.mainwindow";
```
hultonha, line 51: question: Does this constant/string not exist elsewhere in some more central place? Do we need to create it again here?

hultonha: Same for some of the other strings below too

Gems/WhiteBox/Code/Source/SubComponentModes/EditorWhiteBoxDefaultMode.cpp
```diff
@@ -132,6 +148,166 @@ namespace WhiteBox
...
+                            EditorWhiteBoxDefaultModeRequestBus::Event(
+                                entityComponentIdPair, &EditorWhiteBoxDefaultModeRequests::HideSelectedEdge);
+                        }
+                    );
+                }
+            );
+
+            actionManagerInterface->InstallEnabledStateCallback(
+                actionIdentifier,
+                [&]()
```
hultonha, line 196: suggestion: It's generally preferred to capture variables explicitly instead of a blanket reference capture, does this add too much noise or is it doable here? Might be nice to capture this and whatever other local variables are needed (if any)

Gems/WhiteBox/Code/Source/SubComponentModes/EditorWhiteBoxDefaultMode.cpp
```diff
@@ -132,6 +148,166 @@ namespace WhiteBox
...
+
+                    componentModeCollectionInterface->EnumerateActiveComponents(
+                        [&isEdgeSelected](const AZ::EntityComponentIdPair& entityComponentIdPair, const AZ::Uuid&)
+                        {
+                            WhiteBox::Api::EdgeHandles handles;
+
+                            EditorWhiteBoxDefaultModeRequestBus::EventResult(
+                                handles, entityComponentIdPair, &EditorWhiteBoxDefaultModeRequests::SelectedEdgeHandles);
+
+                            if (handles.size() > 0)
```
hultonha, line 212: nit: Can we update this to `(!handles.empty())`

Gems/WhiteBox/Code/Source/SubComponentModes/EditorWhiteBoxDefaultMode.cpp
```diff
@@ -132,6 +148,166 @@ namespace WhiteBox
...
+                            EditorWhiteBoxDefaultModeRequestBus::Event(
+                                entityComponentIdPair, &EditorWhiteBoxDefaultModeRequests::HideSelectedVertex);
+                        }
+                    );
+                }
+            );
+
+            actionManagerInterface->InstallEnabledStateCallback(
+                actionIdentifier,
+                [&]()
```
hultonha, line 256: nit: Same comment as before about `&` capture

Gems/WhiteBox/Code/Source/SubComponentModes/EditorWhiteBoxDefaultMode.cpp
```diff
@@ -132,6 +148,166 @@ namespace WhiteBox
...
+                        {
+                            WhiteBox::Api::EdgeHandles edgeHandles;
+                            EditorWhiteBoxDefaultModeRequestBus::EventResult(
+                                edgeHandles, entityComponentIdPair, &EditorWhiteBoxDefaultModeRequests::SelectedEdgeHandles);
+
+                            WhiteBox::Api::VertexHandles vertexHandles;
+                            EditorWhiteBoxDefaultModeRequestBus::EventResult(
+                                vertexHandles, entityComponentIdPair, &EditorWhiteBoxDefaultModeRequests::SelectedVertexHandles);
+
+                            if (edgeHandles.size() == 0 && vertexHandles.size() > 0)
```
hultonha, line 275: nit: `!empty` change again instead of `size > 0`

Gems/WhiteBox/Code/Source/SubComponentModes/EditorWhiteBoxDefaultMode.cpp
```diff
@@ -157,17 +333,12 @@ namespace WhiteBox
                     .SetTip(HideEdgeDesc)
                     .SetEntityComponentIdPair(entityComponentIdPair)
                     .SetCallback(
-                        [entityComponentIdPair, modifier]()
+                        [&]()
```
hultonha, line 336: nit: Prefer `[this]` here

Gems/WhiteBox/Code/Source/SubComponentModes/EditorWhiteBoxDefaultMode.cpp
```diff
@@ -180,17 +351,12 @@ namespace WhiteBox
                     .SetTip(HideVertexDesc)
                     .SetEntityComponentIdPair(entityComponentIdPair)
                     .SetCallback(
-                        [entityComponentIdPair, modifier2]()
+                        [&]()
```
hultonha, line 354: update: Prefer `[this]` here again

Gems/WhiteBox/Code/Source/SubComponentModes/EditorWhiteBoxDefaultModeBus.h
```diff
@@ -60,6 +60,10 @@ namespace WhiteBox
         //! Returns the hovered polygon handle.
         //! @note If no polygon is currently being hovered then an empty PolygonHandle is returned.
         virtual Api::PolygonHandle HoveredPolygonHandle() const = 0;
+        //! Hide the selected Edge.
+        virtual void HideSelectedEdge() const = 0;
+        //! Hide the selected Vertex.
+        virtual void HideSelectedVertex() const = 0;
```
hultonha, line 66: update: Should these be `const`? It seems to me like this is a mutable operation, they probably shouldn't be `const` 🤔

Gems/WhiteBox/Code/Source/SubComponentModes/EditorWhiteBoxEdgeRestoreMode.cpp
```diff
@@ -33,6 +34,26 @@ namespace WhiteBox
         // noop
     }
 
+    void EdgeRestoreMode::RegisterActionUpdaters()
+    {
+    }
+
+    void EdgeRestoreMode::RegisterActions()
+    {
+    }
```
hultonha, line 43: question: Do we need these if they're not doing anything? Or are they part of an interface we have to include?

Gems/WhiteBox/Code/Source/SubComponentModes/EditorWhiteBoxTransformMode.cpp
```diff
@@ -125,6 +134,139 @@ namespace WhiteBox
...
+            AzToolsFramework::ActionProperties actionProperties;
+            actionProperties.m_name = SwitchToTranslationModeTile;
+            actionProperties.m_description = SwitchToTranslationModeDesc;
+            actionProperties.m_category = "White Box Component Mode - Transform";
+
+            actionManagerInterface->RegisterAction(
+                EditorMainWindowActionContextIdentifier,
+                actionIdentifier,
+                actionProperties,
+                [&]
```
hultonha, line 165: update: Won't mention again but same comment about this likely being able to switch to `[this]` and any other values that need to be captured explicitly

Gems/WhiteBox/Code/Source/SubComponentModes/EditorWhiteBoxTransformModeBus.h
```diff
@@ -0,0 +1,35 @@
...
+namespace WhiteBox
+{
+    enum class TransformType
+    {
+        Translation,
+        Rotation,
+        Scale
+    };
+
+    //! Request bus for White Box ComponentMode operations while in 'default' mode.
```
hultonha, line 23: update: Should this comment be updated?

Gems/WhiteBox/Code/Source/SubComponentModes/EditorWhiteBoxTransformModeBus.h
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+#pragma once
+
+#include <AzCore/Component/ComponentBus.h>
+#include <WhiteBox/WhiteBoxToolApi.h>
+
+namespace WhiteBox
+{
+    enum class TransformType
```
hultonha, line 16: suggestion: API comment for this type would be useful

hultonha (commented):
Changes look good overall but I have a few minor suggestions/comments, thanks!

Gems/WhiteBox/Code/Source/EditorWhiteBoxComponentMode.cpp
mbalfour-amzn: i did suggest it in the past, though turns out they both have the same flaw. ideally, i'd prefer us to find patterns that didn't require us to do "one tick delays" quite as often in the UI. 

the specific problem i ran into is that if you have any code that brings up a dialog box or anything, Qt will continue ticking inside the dialog box, inside of the main engine tick. The thing I discovered when looking further is that this actually makes the main engine continue ticking reentrantly as well, which is pretty awkward. So either way, you can have these "one tick delays" actually fire _sort of_ in the same tick that they're created in with the right circumstances.

ideally, we'd want code like this to be more deterministically hooked up to the exact event or circumstance that they need to have completed before running the next stage of logic. so for example, if it needs to run after "automated mode switching", it would queue an action to run on an automated mode switch, not on a tick.

hultonha: `QueueFunction` I think was the name, but @mbalfour-amzn does make a good point above

AMZN-daimini: I agree that this should not be necessary. If this same issue pops up in another one of the migrations, I'll try to refactor this so that the timeout isn't necessary.

Gems/WhiteBox/Code/Source/SubComponentModes/EditorWhiteBoxDefaultMode.cpp
AMZN-daimini: I have a task set up to centralize these in one or more headers, but since I'm working on these migrations and sometimes they get parallelized I'm gonna do that pass all at once at the end

Gems/WhiteBox/Code/Source/SubComponentModes/EditorWhiteBoxDefaultMode.cpp
AMZN-daimini: It was actually unnecessary, fixed.

Gems/WhiteBox/Code/Source/SubComponentModes/EditorWhiteBoxDefaultModeBus.h
AMZN-daimini: You are correct, fixing.

Gems/WhiteBox/Code/Source/SubComponentModes/EditorWhiteBoxEdgeRestoreMode.cpp
AMZN-daimini: They are not necessary, but I'm leaving them around for consistency since there's nothing enforcing them at the moment. I'll try to refactor this to make it more streamlined if possible.

Gems/WhiteBox/Code/Source/SubComponentModes/EditorWhiteBoxTransformMode.cpp
AMZN-daimini: These were unnecessary again, removed.

Gems/WhiteBox/Code/Source/SubComponentModes/EditorWhiteBoxTransformModeBus.h
AMZN-daimini: This was pre-existing, I just moved it :) Adding a comment.

Gems/WhiteBox/Code/Source/SubComponentModes/EditorWhiteBoxDefaultMode.cpp
hultonha: Okay got you makes sense, thanks for clarifying

Gems/WhiteBox/Code/Source/SubComponentModes/EditorWhiteBoxDefaultMode.cpp
hultonha: Thanks!

Gems/WhiteBox/Code/Source/SubComponentModes/EditorWhiteBoxDefaultMode.cpp
hultonha: Cheers!

Gems/WhiteBox/Code/Source/SubComponentModes/EditorWhiteBoxDefaultModeBus.h
hultonha: Nice one thanks!

Gems/WhiteBox/Code/Source/SubComponentModes/EditorWhiteBoxEdgeRestoreMode.cpp
hultonha: Okay makes sense for consistency, thanks

Gems/WhiteBox/Code/Source/SubComponentModes/EditorWhiteBoxTransformModeBus.h
hultonha: Awesome thanks!

Gems/WhiteBox/Code/Source/EditorWhiteBoxComponentMode.cpp
```diff
@@ -16,18 +16,27 @@
 #include <AzCore/Settings/SettingsRegistry.h>
 #include <AzCore/std/sort.h>
 
+#include <AzToolsFramework/ActionManager/Action/ActionManagerInterface.h>
+#include <AzToolsFramework/ActionManager/Menu/MenuManagerInterface.h>
+#include <AzToolsFramework/ActionManager/HotKey/HotKeyManagerInterface.h>
```
lsemp3d, line 21: Should we consider a single include that pulls these? `#include <AzToolsFramework/ActionManager/ActionManager.h>`

PR #13971, opened 2023-01-09
Removing outdated pinned entity test

jckand-amzn:
Signed-off-by: jckand-amzn <82226555+jckand-amzn@users.noreply.github.com>

## What does this PR do?

Removes test_GradientPreviewSettings_DefaultPinnedEntityIsSelf. A recent logic change no longer populates the EntityId field this test used to validate that the preview is set to self. A new unit test will be written to validate the changed behavior. This test is no longer necessary.

## How was this PR tested?

Ran updated Periodic test suite without issue.


gadams3 (approved):

michleza:
Do we have documentation or style guides for python tests someplace? I see a note about python tests in the docs repo but I'm wondering if we make a note somewhere to avoid including tests like this. Also if we ever audit our automated tests to remove 'is a regression/unit test' scripts such as this one.

michleza (approved):
Do we have documentation or style guides for python tests someplace? I see a note about python tests in the docs repo but I'm wondering if we make a note somewhere to avoid including tests like this. Also if we ever audit our automated tests to remove 'is a regression/unit test' scripts such as this one.

amzn-tmryan (approved):

PR #13975, opened 2023-01-09
Removed the broken seed builder.

AMZN-stankowi:
This was used originally to facilitate asset relocation: It checked if any seed lists referenced the asset to be relocated. However, it didn't work in practice. Often times asset seed files use the same relative file name for the default seed system, which does not play well with the asset system. If multiple source assets exist at the same relative path to the scan folder root for the scan folder they are in, the Asset Processor selects only one of them to be allowed to exist, which means that most seed lists (the default seeds) ended up invisible to this system. Testing it farther locally, it wasn't actually generating dependencies correctly, the jobs weren't running correctly, and the asset relocation wasn't catching these references in the seed list file to the file I attempted to relocate.

To address this, we'll need to re-evalute and likely need to follow a different approach for resolving references from seeds to relocated assets, with the current relocation system.

Also note that the new relocation system (see RFC https://github.com/o3de/sig-content/issues/77) will likely require us to revisit this entire asset relocation system in the asset processor batch. We may discard it, if we find the new asset relocation system doesn't need any additional tooling.

## What does this PR do?

Removes the asset builder for seed list files, because it wasn't functional and was slowing down asset processing.

## How was this PR tested?

Tested locally. Before removing this I:
1. Looked in the asset database, and verified that seeds weren't showing up as source assets, and weren't showing up with associated jobs. At first I thought they weren't even running at all, but I found record of them running in the metrics table.
2. Did some asset relocation tests locally, setup with seed files pointed at the files I was relocating. I verified it wasn't catching the seed list files at all. I also made sure the system itself worked, by having another reference that did get picked up correctly.
3. Looked at the source control history of the creation of these files, from pre-O3DE times, and verified it was built just for this situation: To help asset relocation find these references.


lsemp3d (approved):

jackalbe (approved):

lumberyard-employee-dm (approved):

lawsonamzn (approved):
I guess the "seed builder worker" is hte only instance of a Dependency Builder?

AMZN-stankowi:
> I guess the "seed builder worker" is hte only instance of a Dependency Builder?

Yeah. There's a separate Copy Dependency Builder, but it doesn't use the Dependency Builder.

junbo75 (approved):

cgalvan (approved):

PR #13978, opened 2023-01-10
Prefab - Fix cached instance DOM update in undo delete

forkercat:

## What does this PR do?

This PR changes to update cached instance DOM in owning instance rather than the focused instance for deletion overrides.

We should update the cached instance DOM that is stored in the owning instance. Otherwise, the Entity 1 container entity would be unnecessarily reloaded due to sort order update.

## How was this PR tested?

Passed all tests.


amzn-ahmadkrm (approved):

forkercat:
I'm still investigating if the cached focused instance DOM needs to be updated as well. The reloading, in the above example, on the level instance did occur if we don't do the update.

srikappa-amzn (approved):
good catch!

PR #13983, opened 2023-01-10
[Terrain] Add mip safeguards and warnings when painting

mbalfour-amzn:
## What does this PR do?

This adds some safety guard rails when creating the modification buffer for Terrain Macro Material painting. Previously, the code assumed that mip 0 would be loaded if the asset is loaded. While this *should* be a safe assumption since we aren't directly using the StreamingImagePool, and we're loading the asset in a way that should cause it to load all mips, this PR adds some safety checks to doubly-ensure that we're creating the modification buffer off of the highest loaded mip and that we'll emit a warning if that mip is anything but mip 0.

## How was this PR tested?

Adjusted the streaming image memory budgets and tried painting. In all cases, mip 0 has remained loaded (as expected), so the safety checks haven't been fully exercised, but the modification buffer creation has remained stable throughout my testing.


rgba16f (approved):

invertednormal (approved):

Gems/Terrain/Code/Source/TerrainRenderer/Components/TerrainMacroMaterialComponent.cpp
```diff
@@ -465,19 +465,54 @@ namespace Terrain
...
+
+                // The tail mip chain is always loaded but can't be queried the same way as the other mip chain assets,
+                // so we'll just assume that it's loaded and return its highest mip level.
+                // Note that this could still potentially be mip 0 if the image is extremely small.
+                return aznumeric_cast<uint32_t>(m_configuration.m_macroColorAsset->GetMipLevel(numMipChains - 1));
+            }
+        }
+
+        // No loaded mip level found.
+        return AZStd::numeric_limits<uint32_t>::max();
```
jiaweig-amzn, line 494: It's better to define the invalid mip level somewhere.

jiaweig-amzn (approved):

PR #13995, opened 2023-01-11
use the python provided with o3de

kh-huawei:
Signed-off-by: Karl Haubenwallner <karl.haubenwallner@huawei.com>

## What does this PR do?

set the `Python_ROOT_DIR` to point towards the python-version provided with o3de.
Otherwise cmake finds the python installed on the PC first, which has a potentially different version.

## How was this PR tested?

The editor compiles and starts successfully

### DIFF
--- a/cmake/LYPython.cmake
+++ b/cmake/LYPython.cmake
@@ -226,7 +226,7 @@ ly_set(LY_PACKAGE_VALIDATE_PACKAGE ${temp_LY_PACKAGE_VALIDATE_PACKAGE})
 
 if (NOT CMAKE_SCRIPT_MODE_FILE)
     # if we're in script mode, we dont want to actually try to find package or anything else
-
+    set (Python_ROOT_DIR ${LY_ROOT_FOLDER}/python/runtime/${LY_PYTHON_PACKAGE_NAME}/python)
     find_package(Python ${LY_PYTHON_VERSION} REQUIRED)
 
     # note - if you want to use a normal python via FindPython instead of the LY package above,


lawsonamzn (approved):

cmake/LYPython.cmake
```diff
@@ -226,7 +226,7 @@ ly_set(LY_PACKAGE_VALIDATE_PACKAGE ${temp_LY_PACKAGE_VALIDATE_PACKAGE})
 
 if (NOT CMAKE_SCRIPT_MODE_FILE)
     # if we're in script mode, we dont want to actually try to find package or anything else
-
+    set (Python_ROOT_DIR ${LY_ROOT_FOLDER}/python/runtime/${LY_PYTHON_PACKAGE_NAME}/python)
```
lumberyard-employee-dm, line 229: @kh-huawei I am assuming the version of python you have installed matches the current `LY_PYTHON_VERSION` value which is 3.10.5?

cgalvan (approved):

cmake/LYPython.cmake
kh-huawei: My local python-versions  as of now are 3.11.1 and 3.10.7, and none of them are in my PATH (i.e. only installed in AppData/Local/...).

But upon some further testing I am not able to reproduce the issue that caused me to add this line.
I suspect this was a transient issue that occurred during the python upgrade, since back then cmake did find the wrong python at one point.

But since i can't reproduce the issue now on my end, should i close this MR?

lawsonamzn:
The way our scripts work is that they all explicitly run the python exe from the runtime folder, and python itself will use the python dlls and packages adjacent to its executable, so I'm not sure how it came to be that your build ran the wrong python unless someone had used some sort of operating system level override like PATH (or linux, LD_LIBRARY_PATH).

kh-huawei:
> The way our scripts work is that they all explicitly run the python exe from the runtime folder, and python itself will use the python dlls and packages adjacent to its executable, so I'm not sure how it came to be that your build ran the wrong python unless someone had used some sort of operating system level override like PATH (or linux, LD_LIBRARY_PATH).

Hmm.

The LY_PYTHON_CMD always points to `${LY_ROOT_FOLDER}/python/runtime/python-3.10.5-rev1-windows/python/python.exe`, so i'm assuming that's the executable that was actually used.

But the `find_package(Python ...)` - call did return a different python version which would lead to the case where the wrong value for `${Python_BINFOLDER}` is appended to the `CMAKE_PROGRAM_PATH`.

With this change i try to avoid that by setting the `Python_ROOT_DIR` hint before the `find_package()` - call.

But since i can't really reproduce the issue anymore: Is it possible this behaviour was just an edge case, where the old path was still in the cmake cache or something?

cmake/LYPython.cmake
lumberyard-employee-dm: > > The way our scripts work is that they all explicitly run the python exe from the runtime folder, and python itself will use the python dlls and packages adjacent to its executable, so I'm not sure how it came to be that your build ran the wrong python unless someone had used some sort of operating system level override like PATH (or linux, LD_LIBRARY_PATH).
> 
> Hmm.
> 
> The LY_PYTHON_CMD always points to `${LY_ROOT_FOLDER}/python/runtime/python-3.10.5-rev1-windows/python/python.exe`, so i'm assuming that's the executable that was actually used.
> 
> But the `find_package(Python ...)` - call did return a different python version which would lead to the case where the wrong value for `${Python_BINFOLDER}` is appended to the `CMAKE_PROGRAM_PATH`.
> 
> With this change i try to avoid that by setting the `Python_ROOT_DIR` hint before the `find_package()` - call.
> 
> But since i can't really reproduce the issue anymore: Is it possible this behaviour was just an edge case, where the old path was still in the cmake cache or something?



> > The way our scripts work is that they all explicitly run the python exe from the runtime folder, and python itself will use the python dlls and packages adjacent to its executable, so I'm not sure how it came to be that your build ran the wrong python unless someone had used some sort of operating system level override like PATH (or linux, LD_LIBRARY_PATH).
> 
> Hmm.
> 
> The LY_PYTHON_CMD always points to `${LY_ROOT_FOLDER}/python/runtime/python-3.10.5-rev1-windows/python/python.exe`, so i'm assuming that's the executable that was actually used.
> 
> But the `find_package(Python ...)` - call did return a different python version which would lead to the case where the wrong value for `${Python_BINFOLDER}` is appended to the `CMAKE_PROGRAM_PATH`.
> 
> With this change i try to avoid that by setting the `Python_ROOT_DIR` hint before the `find_package()` - call.
> 
> But since i can't really reproduce the issue anymore: Is it possible this behaviour was just an edge case, where the old path was still in the cmake cache or something?



> > The way our scripts work is that they all explicitly run the python exe from the runtime folder, and python itself will use the python dlls and packages adjacent to its executable, so I'm not sure how it came to be that your build ran the wrong python unless someone had used some sort of operating system level override like PATH (or linux, LD_LIBRARY_PATH).
> 
> Hmm.
> 
> The LY_PYTHON_CMD always points to `${LY_ROOT_FOLDER}/python/runtime/python-3.10.5-rev1-windows/python/python.exe`, so i'm assuming that's the executable that was actually used.
> 
> But the `find_package(Python ...)` - call did return a different python version which would lead to the case where the wrong value for `${Python_BINFOLDER}` is appended to the `CMAKE_PROGRAM_PATH`.
> 
> With this change i try to avoid that by setting the `Python_ROOT_DIR` hint before the `find_package()` - call.
> 
> But since i can't really reproduce the issue anymore: Is it possible this behaviour was just an edge case, where the old path was still in the cmake cache or something?

Could it have been possible that at the time the issue occurred the O3DE Python version matched the System Python version?

lumberyard-employee-dm:
> > The way our scripts work is that they all explicitly run the python exe from the runtime folder, and python itself will use the python dlls and packages adjacent to its executable, so I'm not sure how it came to be that your build ran the wrong python unless someone had used some sort of operating system level override like PATH (or linux, LD_LIBRARY_PATH).
> 
> Hmm.
> 
> The LY_PYTHON_CMD always points to `${LY_ROOT_FOLDER}/python/runtime/python-3.10.5-rev1-windows/python/python.exe`, so i'm assuming that's the executable that was actually used.
> 
> But the `find_package(Python ...)` - call did return a different python version which would lead to the case where the wrong value for `${Python_BINFOLDER}` is appended to the `CMAKE_PROGRAM_PATH`.
> 
> With this change i try to avoid that by setting the `Python_ROOT_DIR` hint before the `find_package()` - call.
> 
> But since i can't really reproduce the issue anymore: Is it possible this behaviour was just an edge case, where the old path was still in the cmake cache or something?



> > The way our scripts work is that they all explicitly run the python exe from the runtime folder, and python itself will use the python dlls and packages adjacent to its executable, so I'm not sure how it came to be that your build ran the wrong python unless someone had used some sort of operating system level override like PATH (or linux, LD_LIBRARY_PATH).
> 
> Hmm.
> 
> The LY_PYTHON_CMD always points to `${LY_ROOT_FOLDER}/python/runtime/python-3.10.5-rev1-windows/python/python.exe`, so i'm assuming that's the executable that was actually used.
> 
> But the `find_package(Python ...)` - call did return a different python version which would lead to the case where the wrong value for `${Python_BINFOLDER}` is appended to the `CMAKE_PROGRAM_PATH`.
> 
> With this change i try to avoid that by setting the `Python_ROOT_DIR` hint before the `find_package()` - call.
> 
> But since i can't really reproduce the issue anymore: Is it possible this behaviour was just an edge case, where the old path was still in the cmake cache or something?



> > The way our scripts work is that they all explicitly run the python exe from the runtime folder, and python itself will use the python dlls and packages adjacent to its executable, so I'm not sure how it came to be that your build ran the wrong python unless someone had used some sort of operating system level override like PATH (or linux, LD_LIBRARY_PATH).
> 
> Hmm.
> 
> The LY_PYTHON_CMD always points to `${LY_ROOT_FOLDER}/python/runtime/python-3.10.5-rev1-windows/python/python.exe`, so i'm assuming that's the executable that was actually used.
> 
> But the `find_package(Python ...)` - call did return a different python version which would lead to the case where the wrong value for `${Python_BINFOLDER}` is appended to the `CMAKE_PROGRAM_PATH`.
> 
> With this change i try to avoid that by setting the `Python_ROOT_DIR` hint before the `find_package()` - call.
> 
> But since i can't really reproduce the issue anymore: Is it possible this behaviour was just an edge case, where the old path was still in the cmake cache or something?

Could it have been possible that at the time the issue occurred the O3DE Python version matched the System Python version?

The find_package script shouldn't try to locate the system python. I wounder if the find_package call should supply the [EXACT](https://cmake.org/cmake/help/latest/command/find_package.html#basic-signature) parameter.

kh-huawei:
> Could it have been possible that at the time the issue occurred the O3DE Python version matched the System Python version?

That is entirely possible.

lawsonamzn:
It might be worth looking into find_package and seeing if theres a way to enforce it only looking in specified folders.

cmake/LYPython.cmake
rgba16f: Please do close it if the change is no longer required, we get an email listing all the Pull Requests that are currently open.

PR #13998, opened 2023-01-11
add GetMeshScene() to the MeshFeatureProcessorInterface

kh-huawei:
Signed-off-by: Karl Haubenwallner <karl.haubenwallner@huawei.com>

## What does this PR do?

Adds the interface-function `GetMeshScene()` to the MeshFeatureProcessorInterface, to get access to the `RPI::Scene` of a `MeshHandle`.

This is somewhat of a workaround when broadcasting `MeshHandleStateRequestBus::Events::GetMeshHandle()`, since that Ebus does not offer a way to limit which mesh-handles are returned, which means you will also get mesh-handles from the Preview-Scene, created and stored by a different MeshFeatureProcessor.

## How was this PR tested?

Windows & Vulkan. Explicitly collecting all mesh-handles in a component and sorting out mesh-handles from the Preview-Scene.

### DIFF
--- a/Gems/Atom/Feature/Common/Code/Include/Atom/Feature/Mesh/MeshFeatureProcessor.h
+++ b/Gems/Atom/Feature/Common/Code/Include/Atom/Feature/Mesh/MeshFeatureProcessor.h
@@ -216,6 +216,7 @@ namespace AZ
             bool GetRayTracingEnabled(const MeshHandle& meshHandle) const override;
             void SetVisible(const MeshHandle& meshHandle, bool visible) override;
             bool GetVisible(const MeshHandle& meshHandle) const override;
+            RPI::Scene* GetMeshScene(const MeshHandle& meshHandle) const override;
             void SetUseForwardPassIblSpecular(const MeshHandle& meshHandle, bool useForwardPassIblSpecular) override;
 
             RHI::Ptr <FlagRegistry> GetShaderOptionFlagRegistry();
--- a/Gems/Atom/Feature/Common/Code/Include/Atom/Feature/Mesh/MeshFeatureProcessorInterface.h
+++ b/Gems/Atom/Feature/Common/Code/Include/Atom/Feature/Mesh/MeshFeatureProcessorInterface.h
@@ -133,6 +133,9 @@ namespace AZ
             //! Returns the visibility state of the mesh.
             //! This only refers to whether or not the mesh has been explicitly hidden, and is not related to view frustum visibility.
             virtual bool GetVisible(const MeshHandle& meshHandle) const = 0;
+            //! Returns the scene of the MeshHandle. It's possible this is a different scene than the Parent-Scene of the
+            //! MeshFeatureProcessor.
+            virtual RPI::Scene* GetMeshScene(const MeshHandle& meshHandle) const = 0;
             //! Sets the mesh to render IBL specular in the forward pass.
             virtual void SetUseForwardPassIblSpecular(const MeshHandle& meshHandle, bool useForwardPassIblSpecular) = 0;
         };
--- a/Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
@@ -655,6 +655,15 @@ namespace AZ
             }
         }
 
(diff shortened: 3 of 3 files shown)


invertednormal:
As is, I don't think this change makes sense - a MeshFeatureProcessor (and therefor every handle in it) belongs to a single RPI::Scene, so requesting the scene for a _specific_ handle doesn't make sense - they're all the same.

gadams3 (approved):

kh-huawei:
I'm using `BroadcastResult` to get all mesh handles:         
```
using MeshHandle = AZ::Render::MeshFeatureProcessorInterface::MeshHandle;
AZ::EBusAggregateResults<const MeshHandle*> handles;
AZ::Render::MeshHandleStateRequestBus::BroadcastResult(handles, &AZ::Render::MeshHandleStateRequestBus::Events::GetMeshHandle);
```
Which is quite convenient, but there i definitely get all handles of all scenes.
Preselecting the Entities would work but this seems more efficient, at least if the only other scene is the PreviewScene.

invertednormal:
It looks like the MeshHandleStateRequestBus is already addressed by ID based on the entity ID, so we can't easily limit the scope to the scene ID instead. The problem with doing a bus call then is that it will always invoke every single instance of the bus, which could be thousands spread across many scenes. 

Another problem with the proposed approach is it I think it's making the assumption that you can pass any mesh handle to any mesh feature processor, which is not safe. I'm guessing that you're getting the correct feature processor for your scene then checking every handle returned from the bus with that feature processor, but some of those mesh handles might not even belong to the feature processor - it only works because the only thing you're doing is checking the handle itself for some private data.

Would something like a visitor pattern on the mesh feature processor work? It could go through all of the mesh handles owned by that feature processor and call some arbitrary function that you provide with the handle.

kh-huawei:
> Another problem with the proposed approach is it I think it's making the assumption that you can pass any mesh handle to any mesh feature processor, which is not safe. I'm guessing that you're getting the correct feature processor for your scene then checking every handle returned from the bus with that feature processor, but some of those mesh handles might not even belong to the feature processor - it only works because the only thing you're doing is checking the handle itself for some private data.
> 

I agree. I did find out that this can happen rather late, unfortunately.

> Would something like a visitor pattern on the mesh feature processor work? It could go through all of the mesh handles owned by that feature processor and call some arbitrary function that you provide with the handle.

That would work, yes. 
But for my use-case (creating custom draw-items for all meshes for a rasterpass that only exists if my Gem is active), new Events `OnMeshAdded()` and `OnMeshRemoved()` (or so) that i can subscribe to would work even better, but that would require further reaching changes, and some refactoring from my side.

invertednormal:
> That would work, yes. But for my use-case (creating custom draw-items for all meshes for a rasterpass that only exists if my Gem is active), new Events `OnMeshAdded()` and `OnMeshRemoved()` (or so) that i can subscribe to would work even better, but that would require further reaching changes, and some refactoring from my side.

I think I'd be ok with some `AZ::Event`s on the `MeshFeatureProcessor` for those notifications, which shouldn't be too hard to do by itself. You'd have to check if the model was actually ready when `OnMeshAdded()` is called, and listen for updates if it's not (using `ConnectModelChangeEventHandler()`.

amzn-tommy:
> 

The only way to pass MeshHandles around at the moment is to pass around a pointer to a MeshHandle, just as the MeshHandleState busses do. Are you working in the context of o3de or Atom as a standalone renderer? For O3DE, the MeshHandleStateNotificationBus is already going to send an event whenever a mesh handle is acquired, only it's sent by the component instead of the MeshFeatureProcessor (making it an O3DE only event, not something that would be sent in something like AtomSampleViewer which uses Atom as a standalone renderer).

The only caveat with using the MeshHandleStateNotificationBus or MeshHandleStateRequestBus is that they are addressed by entityId, so you need to already know which entityId(s) you are interested in. If you are interested in a particular scene, you can use IVisibilitySystem::FindVisibilityScene to get the scene, then IVisibilitySystem::EnumerateNoCull with a callback to get all the entities in a scene. You can cast the VisibilityEntry->m_userData to an Entity*, and get the EntityId from there. See MultiplayerDebugHierarchyReporter::CollectHierarchyRoots as an example of how to do this (although that example uses Enumerate instead of EnumerateNoCull. Use Enumerate if you want to only get entities where the Octree node overlaps the view frustum).

From there, you could connect to the MeshHandleStateNotificationBus or call the MeshHandleStateRequestBus as needed for each entity.

PR #14005, opened 2023-01-11
Simplify screenshot code

jiaweig-amzn:

## What does this PR do?

Move duplicated code into a function. Add a middle internal function for capture as well.
This will help modifying the following API change work.

## How was this PR tested?
Tested in ASV screenshot tests.


Gems/Atom/Feature/Common/Code/Source/FrameCaptureSystemComponent.cpp
```diff
@@ -389,94 +389,81 @@ namespace AZ
...
         }
 
-        FrameCaptureId FrameCaptureSystemComponent::CaptureScreenshotForWindow(const AZStd::string& filePath, AzFramework::NativeWindowHandle windowHandle)
+        FrameCaptureSystemComponent::CaptureHandle FrameCaptureSystemComponent::ScreenshotPreparation(
+            const AZStd::string& imagePath, AZ::RPI::AttachmentReadback::CallbackFunction callbackFunction)
         {
             if (!CanCapture())
             {
-                return InvalidFrameCaptureId;
+                AZ_Error("FrameCaptureSystemComponent", false, "Frame capture is not avaible.");
```
VickyAtAZ, line 397: I am not sure if we want to report an error here. It seems the CanCapture() checks if it's null renderer. And we may want this behavior and doesn't want to treat it as an error.

VickyAtAZ (approved):
I only have a question.

Gems/Atom/Feature/Common/Code/Source/FrameCaptureSystemComponent.cpp
rgba16f: +1, not an error to fail a capture here.

Gems/Atom/Feature/Common/Code/Source/FrameCaptureSystemComponent.cpp
```diff
@@ -509,50 +496,54 @@ namespace AZ
...
+            if (captureHandle.IsNull())
+            {
+                return InvalidFrameCaptureId;
+            }
+            AZStd::scoped_lock<CaptureHandle> scope_lock(captureHandle);
+            CaptureState* captureState = captureHandle.GetCaptureState();
+
+            // Find SwapChainPass for the window handle
+            RPI::SwapChainPass* pass = AZ::RPI::PassSystemInterface::Get()->FindSwapChainPass(windowHandle);
+            if (!pass)
```
rgba16f, line 511: This check could be done before the call to ScreenshotPreparation, would make the logic easier as you don't have to return a capture back to idle.

Gems/Atom/Feature/Common/Code/Source/FrameCaptureSystemComponent.cpp
jiaweig-amzn: I'll remove it. But is there a reason to capture a frame while it is running a null renderer?

Gems/Atom/Feature/Common/Code/Source/FrameCaptureSystemComponent.cpp
```diff
@@ -389,94 +389,80 @@ namespace AZ
...
+            }
+            else
+            {
+                capture->m_readback->SetCallback(
+                    AZStd::bind(&FrameCaptureSystemComponent::CaptureAttachmentCallback, this, AZStd::placeholders::_1));
+            }
+            if (!imagePath.empty())
+            {
+                capture->m_outputFilePath = ResolvePath(imagePath);
             }
```
rgba16f, line 436: We don't support supplying an output filepath if a custom callback function is supplied, so this block should probably be moved up inside the `callbackFunction == nullptr` block. And possibly an assert added to the `callbackFunction != nullptr` block

lsemp3d (approved):

Gems/Atom/Feature/Common/Code/Source/FrameCaptureSystemComponent.cpp
```diff
@@ -389,94 +389,80 @@ namespace AZ
...
-            {
-                AZ_Assert(false, "Failed to capture attachment since the readback is not ready");
-                m_idleCaptures.push_back(captureHandle);
-                return InvalidFrameCaptureId;
-            }
-            capture->m_readback->SetUserIdentifier(captureHandle.GetCaptureStateIndex());
-            capture->m_readback->SetCallback(AZStd::bind(&FrameCaptureSystemComponent::CaptureAttachmentCallback, this, AZStd::placeholders::_1));
-
-            capture->m_outputFilePath = ResolvePath(outputFilePath);
+            CaptureState* captureState = captureHandle.GetCaptureState();
```
rgba16f, line 465: I think these can be moved later in the function so it's not holding the lock as long.

Gems/Atom/Feature/Common/Code/Source/FrameCaptureSystemComponent.cpp
```diff
@@ -497,7 +483,7 @@ namespace AZ
                 return InvalidFrameCaptureId;
             }
 
-            bool result = previewPass->ReadbackOutput(capture->m_readback);
+            bool result = previewPass->ReadbackOutput(captureState->m_readback);
```
rgba16f, line 486: This looks to be the only place that uses the capture state in this function

Gems/Atom/Feature/Common/Code/Source/FrameCaptureSystemComponent.h
```diff
@@ -102,11 +102,20 @@ namespace AZ
             };
             friend CaptureHandle;
 
-            CaptureHandle  InternalCapturePassAttachment(const AZStd::vector<AZStd::string>& passHierarchy, 
-                                                         const AZStd::string& slotName, 
-                                                         const AZStd::string& outputFilePath,
-                                                         RPI::PassAttachmentReadbackOption option,
-                                                         AZ::RPI::AttachmentReadback::CallbackFunction callbackFunction);
+            FrameCaptureSystemComponent::CaptureHandle ScreenshotPreparation(
```
rgba16f, line 105: nit: you can trim the `FrameCaptureSystemComponent::`

Gems/Atom/Feature/Common/Code/Source/FrameCaptureSystemComponent.cpp
jiaweig-amzn: Good point. Swapped the order.

Gems/Atom/Feature/Common/Code/Source/FrameCaptureSystemComponent.cpp
jiaweig-amzn: There are 3 places using this actually.
One for the swapchain readback and the other 2 for pass attachment readback.
`pass->ReadbackSwapChain(captureState->m_readback);`

Previewpass can be treated as a special pass attachment capture. I am also looking into have them more generalized so that it can be simplified further.

Gems/Atom/Feature/Common/Code/Source/FrameCaptureSystemComponent.cpp
```diff
@@ -584,8 +574,8 @@ namespace AZ
             if (captureHandle.IsValid())
             {
                 AZStd::scoped_lock<CaptureHandle> scope_lock(captureHandle);
-                CaptureState* capture = captureHandle.GetCaptureState();
-                if (!capture)
+                CaptureState* captureState = captureHandle.GetCaptureState();
+                if (!captureState)
                 {
                     return InvalidFrameCaptureId;
                 }
```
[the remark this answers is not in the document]
jiaweig-amzn: Changed to asset and swapped the if block.

Gems/Atom/Feature/Common/Code/Source/FrameCaptureSystemComponent.cpp
```diff
@@ -389,115 +389,113 @@ namespace AZ
...
-            capture->m_readback->SetCallback(AZStd::bind(&FrameCaptureSystemComponent::CaptureAttachmentCallback, this, AZStd::placeholders::_1));
+            else
+            {
+                capture->m_readback->SetCallback(
+                    AZStd::bind(&FrameCaptureSystemComponent::CaptureAttachmentCallback, this, AZStd::placeholders::_1));
 
-            capture->m_outputFilePath = ResolvePath(filePath);
-            pass->ReadbackSwapChain(capture->m_readback);
-            m_inProgressCaptures.push_back(captureHandle);
+                AZ_Assert(!imagePath.empty(), "The image path must present if the callback is not assigned.");
```
rgba16f, line 445: slightly incorrect grammer, should be "image path must be provided"

rgba16f (approved):

PR #14010, opened 2023-01-11
added logic to skip proc prfab with actor scenes

jackalbe:
## What does this PR do?

This sets a flag to skip creating a procedural prefab for scenes that would create actor assets.

This is done since the procedural prefabs logic does not create Actor or Motion components.

The /O3DE/Preferences/Prefabs/IgnoreActors settings registry key was added to feature toggle this logic.

## How was this PR tested?

This was manually tested with the settings registry flag toggled to TRUE and FALSE which gave the expected behavior.


cgalvan (approved):

AMZN-stankowi (commented):
It would be good to see a few automated tests around this: 1) Verify that when the setreg is set, and a scene file has an actor, it doesn't generate a default prefab. 2) Verify that went the setreg is set, and a scene file does not have an actor, it still generates a default prefab. 3) Verify that when the setreg is not set, and a scene file has an actor, it generates a default prefab.

amzn-rhhong (approved):
LGTM. Slight concern with a string based approach maybe a problem to maintain later.

Code/Tools/SceneAPI/SceneCore/Events/GraphMetaInfoBus.h
```diff
@@ -67,6 +67,9 @@ namespace AZ
 
                 // Converts the virtual type hashed name into a readable name.
                 SCENE_CORE_API virtual void GetVirtualTypeName([[maybe_unused]] AZStd::string& name, [[maybe_unused]] Crc32 type) {}
+
+                // Provides the policies that will be applied to the scene from the asset builders.
+                SCENE_CORE_API virtual void GetAppliedPolicyNames([[maybe_unused]] const Containers::Scene& scene, [[maybe_unused]] AZStd::set<AZStd::string>& appliedPolicySet) const {}
```
hultonha, line 72: nit/standards: Potentially would consider breaking the line after the first paren (as the line length is 186 and we're technically meant to have a limit of 140)

hultonha, line 72: suggestion: Also we could rename `appliedPolicySet` to `appliedPolicies` so we don't encode the type in the parameter name

Gems/Prefab/PrefabBuilder/PrefabGroup/PrefabGroupBehavior.cpp
```diff
@@ -128,15 +129,30 @@ namespace AZ::SceneAPI::Behaviors
         ManifestAction action,
         [[maybe_unused]] RequestingApplication requester)
     {
-        // this toggle makes constructing default mesh groups and a prefab optional
+        using namespace SceneAPI::Events;
```
hultonha, line 132: suggestion: It's generally recommended/preferred to about `using namespace` directives (due to the name lookup rule issues it can cause) and instead use a `using` declaration such as `using SceneAPI::Events;` (or whatever type(s) is/are needed (or use a namespace alias if that makes things cleaner)

Gems/Prefab/PrefabBuilder/PrefabGroup/PrefabGroupBehavior.cpp
```diff
@@ -128,15 +129,30 @@ namespace AZ::SceneAPI::Behaviors
...
                 return Events::ProcessingResult::Ignored;
             }
+
+            // do not make a Prefab Group if the animation policy will be applied (if ignore actors is set)
+            bool ignoreActors = true;
+            settingsRegistry->Get(ignoreActors, s_PrefabGroupBehaviorIgnoreActorsKey);
+            if (ignoreActors)
+            {
+                AZStd::set<AZStd::string> appliedPolicySet;
+                GraphMetaInfoBus::Broadcast(&GraphMetaInfoBus::Events::GetAppliedPolicyNames, scene, appliedPolicySet);
```
hultonha, line 150: suggestion/question: If we returned the set instead of using an out param then we could call `BroadcastResult` which might make the callsite a bit clearer?

Gems/Prefab/PrefabBuilder/PrefabGroup/PrefabGroupBehavior.cpp
```diff
@@ -128,15 +129,30 @@ namespace AZ::SceneAPI::Behaviors
...
+                if (appliedPolicySet.find("ActorGroupBehavior") != appliedPolicySet.end())
```
hultonha, line 151: suggestion: I believe we have `.contains` now which is little more concise as we don't need to use the iterator here

Gems/Prefab/PrefabBuilder/PrefabGroup/tests/PrefabBehaviorTests.cpp
```diff
@@ -410,4 +428,79 @@ namespace UnitTest
         EXPECT_EQ(result, ProcessingResult::Ignored);
         EXPECT_EQ(scene->GetManifest().GetEntryCount(), 0);
     }
+
+    TEST_F(PrefabBehaviorTests, PrefabBehavior_IgnoreActors_ToggleTrueReturnsIgnored)
+    {
+        using namespace AZ::SceneAPI;
+        using namespace AZ::SceneAPI::Events;
```
hultonha, line 435: nit: Same comment before about using alternatives to `using namespace`

Gems/Prefab/PrefabBuilder/PrefabGroup/tests/PrefabBehaviorTests.cpp
```diff
@@ -410,4 +428,79 @@ namespace UnitTest
         EXPECT_EQ(result, ProcessingResult::Ignored);
         EXPECT_EQ(scene->GetManifest().GetEntryCount(), 0);
     }
+
+    TEST_F(PrefabBehaviorTests, PrefabBehavior_IgnoreActors_ToggleTrueReturnsIgnored)
```
hultonha, line 432: Awesome work adding the tests! Super helpful to detect the logic is working as expected

hultonha (approved):
I have some minor nits but nothing blocking, feel free to take or leave. One thing that would be super helpful for context would be a before/after video of showing what difference having this setting on causes (when dragging an actor from Asset Browser to the scene). Otherwise looks great! 👍 Thank you very much for adding this workaround, we really appreciate it.

PR #14012, opened 2023-01-11
The scene settings tool now always forces the scene file to re-process on save, even if the file didn't change.

AMZN-stankowi:
The scene settings tool now always forces the scene file to re-process on save, even if the file didn't change.

This is because it's common for content creators to assume this will be the case, and they'll save with no changes to try and alleviate problems with stale content. It's unlikely this would fix their stale content issues, especially around hot reload not working, but it's no longer mis-leading in the tool, which made it seem like the file was reprocessing.

## What does this PR do?

Adds a new message to the asset processor that clears out the fingerprint/hash for files & jobs, so that the file will re-build the next time it changes (or AP is launched) even if the file didn't change in a meaningful way.

## How was this PR tested?

Manual testing, automated tests WIP


Code/Framework/AzToolsFramework/AzToolsFramework/API/EditorAssetSystemAPI.h
```diff
@@ -127,6 +127,14 @@ namespace AzToolsFramework
...
             */
             virtual bool GetAssetsProducedBySourceUUID(const AZ::Uuid& sourceUuid, AZStd::vector<AZ::Data::AssetInfo>& productsAssetInfo) = 0;
+
+            /** This will cause the file to reprocess the next time it changes, even if it's identical to what it was before.
+            * This is useful for in-editor tools that save source files. A content creator expects that, every time they save a file,
+            * the file will be processed, even if nothing has actually changed, so they will sometimes save a file specifically to
+            * force the asset to reprocess.
+            * @return true if the fingerprint was cleared, false if not.
+            */
+            virtual bool ClearFingerprintForAsset(const AZStd::string& sourcePath) = 0;
```
jackalbe, line 137: this might be good to plug into the AP as well to "reprocess" the asset, right?

AMZN-stankowi: The AP already has an existing, different workflow for reprocessing assets. Right click on an asset and you can reprocess it.

This is different because the Editor was sending the command, it wasn't happening in AP directly, and in this case I didn't want to immediately reprocess, I wanted to flag the file in a way it would reprocess the next time it saved, to avoid race conditions if I requested a reprocess after saving the file, to avoid having to have the Editor code check if the fingerprint changed or not.

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
```diff
@@ -357,6 +357,47 @@ namespace AssetProcessor
...
+        {
+            sourceAsset = SourceAssetReference(request.m_searchTerm.c_str());
+        }
+        else
+        {
+            QString absolutePath = m_platformConfig->FindFirstMatchingFile(request.m_searchTerm.c_str());
+
+            if (absolutePath.isEmpty())
+            {
+                return;
```
jackalbe, line 382: should also be "response.m_isSuccess = false;" ?

Code/Tools/AssetProcessor/native/utilities/ApplicationManagerBase.cpp
```diff
@@ -751,6 +751,7 @@ void ApplicationManagerBase::InitAssetRequestHandler(AssetProcessor::AssetReques
 
     if (router)
     {
+        router->RegisterQueuedCallbackHandler(GetAssetProcessorManager(), &AssetProcessorManager::ProcessFingerprintClearRequest);
```
jackalbe, line 754: do we have docs about how to add new AP packets?

AMZN-stankowi: No, I just followed along with existing code. It's reasonably self documenting.

Code/Tools/AssetProcessor/native/unittests/AssetProcessorManagerUnitTests.cpp
```diff
@@ -2963,4 +2963,134 @@ namespace AssetProcessor
...
+        QCoreApplication::processEvents(QEventLoop::AllEvents);
+
+        AssetProcessorManagerUnitTestUtils::SortAssetToProcessResultList(m_processResults);
+    }
+    #if 1
+    // This test verifies the fingerprint clearing command causes assets to reprocess.
+    // It does this by processing assets once to setup. Then processes them a second time, to verify
+    // they don't reprocess due to identical fingerprints. After that, it clears the fingerprint from the DB,
+    // and processes a final time, to verify the assets process because of the fingerprint change.
+    TEST_F(AssetProcessorManagerUnitTests, FingerprintClearRequest_ClearsFingerprints_AssetsReprocess)
```
jackalbe, line 2983: will this unit test be safe to run on an Android and/or iOS?

jackalbe (approved):

Code/Tools/AssetProcessor/native/unittests/AssetProcessorManagerUnitTests.cpp
AMZN-stankowi: We don't run the asset processor on Android/iOS yet. I don't think we have plans to, and if someone wanted to do so, it would take a lot of work to make the tests work on that platform.

Code/Tools/AssetProcessor/native/unittests/AssetProcessorManagerUnitTests.cpp
```diff
@@ -2963,4 +2963,134 @@ namespace AssetProcessor
...
+    {
+        m_processResults.clear();
+        QMetaObject::invokeMethod(
+            m_assetProcessorManager.get(), "AssessModifiedFile", Qt::QueuedConnection, Q_ARG(QString, assetToProcess));
+        EXPECT_TRUE(BlockUntil(m_idling, 5000));
+        QCoreApplication::processEvents(QEventLoop::AllEvents);
+
+        AssetProcessorManagerUnitTestUtils::SortAssetToProcessResultList(m_processResults);
+    }
+    #if 1
```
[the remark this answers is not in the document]
AMZN-stankowi: Fixed

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
AMZN-stankowi: Good catch, fixed

Code/Tools/AssetProcessor/native/unittests/AssetProcessorManagerUnitTests.cpp
```diff
@@ -2963,4 +2963,133 @@ namespace AssetProcessor
...
+        AzToolsFramework::AssetDatabase::FileDatabaseEntry file;
+        file.m_scanFolderPK = scanFolderInfo->ScanFolderID();
+        file.m_fileName = fileName.toUtf8().constData();
+        file.m_isFolder = false;
+
+        // Create a scoped asset database connection, so it's only kept as long as it's needed.
+        {
+            AssetDatabaseConnection connection;
+            EXPECT_TRUE(connection.OpenDatabase());
+            bool entryAlreadyExists;
```
amzn-mike, line 3028: Might be best to init this variable

amzn-mike (approved):
I think there's a unit test with all the network messages in it, would it be easy enough to add this to it, to verify the new message is hooked up right?

amzn-mike (approved):

PR #14013, opened 2023-01-11
Multi-threading updates from server to clients

AMZN-Olex:
## What does this PR do?

Improves the performance of the server by updating connections on separates threads.
Off by default, controlled with sv_multithreadedConnectionUpdates cvar.

## How was this PR tested?

- Tested that the workflow works in the Editor
- Tested with a standalone server and 3 standalone clients
- [IN PROGRESS] Stress test and soak testing now


Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
```diff
@@ -585,6 +569,74 @@ namespace Multiplayer
...
+
+                    job->SetDependent(&jobCompletion);
+                    job->Start();
+                };
+
+                m_networkInterface->GetConnectionSet().VisitConnections(sendNetworkUpdates);
+                jobCompletion.StartAndWaitForCompletion();
+            }
+        }
+        else // On clients (including the Editor) run in a single threaded mode to avoid issues in UI asset loading
```
AMZN-alexpete, line 623: can you provide more details on the UI asset loading issues?

Code/Framework/AzNetworking/AzNetworking/UdpTransport/UdpConnection.cpp
```diff
@@ -88,12 +88,14 @@ namespace AzNetworking
 
     bool UdpConnection::SendReliablePacket(const IPacket& packet)
     {
+        AZStd::lock_guard lock(m_sendPacketMutex);
```
lmbr-pip, line 91: nit: Do we need to guard the both the sequence id and send steps. Maybe useful to cover what the lock is for here.

Looks like we're guarding both. Assume its because SendPacket is not multithread aware but curious as to why.

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
```diff
@@ -115,6 +117,9 @@ namespace Multiplayer
     AZ_CVAR(AZ::TimeMs, bg_captureTransportPeriod, AZ::TimeMs{1000}, nullptr, AZ::ConsoleFunctorFlags::DontReplicate,
         "How often in milliseconds to record transport metrics.");
 
+    AZ_CVAR(bool, sv_multithreadedConnectionUpdates, true, nullptr, AZ::ConsoleFunctorFlags::DontReplicate,
```
lmbr-pip, line 120: How does the dev control # of threads? Whats the impact of the rest of application performance?

lmbr-pip (commented):
Do we have before and after performance numbers? Do we have a sense of CPU/thread usage post change?

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
AMZN-Olex: By the default the number of threads are set by JobManagerComponent for the entire system and shared among all jobs that use the default Job Context.

AMZN-Olex: The exact number depends on the system and the configuration:
```cpp
    void JobManagerComponent::Activate()
...
        int numberOfWorkerThreads = m_numberOfWorkerThreads;
        if (numberOfWorkerThreads <= 0) // spawn default number of threads
        {
        #if (AZ_TRAIT_THREAD_NUM_JOB_MANAGER_WORKER_THREADS)
            numberOfWorkerThreads = AZ_TRAIT_THREAD_NUM_JOB_MANAGER_WORKER_THREADS;
        #else
            uint32_t scaledHardwareThreads = Threading::CalcNumWorkerThreads(cl_jobThreadsConcurrencyRatio, cl_jobThreadsMinNumber, 0, cl_jobThreadsNumReserved);
            numberOfWorkerThreads = AZ::GetMin(static_cast<unsigned int>(desc.m_workerThreads.capacity()), scaledHardwareThreads);
        #endif // (AZ_TRAIT_THREAD_NUM_JOB_MANAGER_WORKER_THREADS)
        }
```

Code/Framework/AzNetworking/AzNetworking/UdpTransport/UdpConnection.cpp
AMZN-Olex: Correct, SendPacket is not multi-threaded, so we need to mutex the whole call here. Sequence is is managed inside by UdpNetworkInterface.

AMZN-puvvadar: Are we scoped locking for each packet send? What's the long term feasibility of scoped locking once for a batch of packets?

Code/Framework/AzNetworking/AzNetworking/DataStructures/TimeoutQueue.h
```diff
@@ -86,6 +86,10 @@ namespace AzNetworking
         TimeoutId        m_nextTimeoutId = TimeoutId{ 0 };
         TimeoutItemMap   m_timeoutItemMap;
         TimeoutItemQueue m_timeoutItemQueue;
+
+        // TimeoutQueue is a shared resource among connections. A mutex (or a read-write sync object) is required with multi-threaded sends.
+        // See @sv_multithreadedConnectionUpdates cvar.
+        AZStd::mutex m_mutex;
```
AMZN-puvvadar, line 92: This works for now but it feels like a lockless scheme could be devised here. Maybe a bucket per thread or similar division to eliminate contention.

AMZN-Olex: Agreed, definitely revisit in the future.

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
AMZN-Olex: I tried getting a repro for this but now client connections dont even connect to the server if they were updated from a thread. So there is something else blocking this. But this isn't necessary for performance work.

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
```diff
@@ -634,6 +620,72 @@ namespace Multiplayer
         stats.RecordFrameTime(AZ::TimeUs{ duration.count() });
     }
 
+    void MultiplayerSystemComponent::UpdatedMetricsConnectionCount()
+    {
+        MultiplayerStats& stats = GetStats();
+        auto sendNetworkUpdates = [&stats](IConnection& connection)
```
AMZN-alexpete, line 626:
```suggestion
        auto updateConnectionStats = [&stats](IConnection& connection)
```
`sendNetworkUpdates` doesn't seem appropriate now that it isn't sending any updates in this lambda

Gems/Multiplayer/Code/Tests/MultiplayerSystemTests.cpp
```diff
@@ -74,12 +77,20 @@ namespace Multiplayer
                 {
                     TestEndpointDisconnectedEvent(value);
                 });
-            m_mpComponent->AddEndpointDisconnectedHandler(m_endpointDisconnectedHandler);
+            m_mpComponent->AddEndpointDisconnectedHandler(m_endpointDisconnectedHandler);
```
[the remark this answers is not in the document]
AMZN-Olex: Done

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
AMZN-Olex: Good point, updated

Gems/Multiplayer/Code/Tests/MultiplayerSystemTests.cpp
```diff
@@ -77,10 +80,18 @@ namespace Multiplayer
                 });
             m_mpComponent->AddEndpointDisconnectedHandler(m_endpointDisconnectedHandler);
             m_mpComponent->Activate();
+
+            m_systemEntity = AZStd::make_unique<AZ::Entity>();
+            m_systemEntity->CreateComponent<AZ::JobManagerComponent>(); // Needed by Job system when @sv_multithreadedConnectionUpdates is on.
+            m_systemEntity->Init();
+            m_systemEntity->Activate();
```
lumberyard-employee-dm, line 87: Well this isn't really the SystemEntity since it doesn't have EntityID 0.
This is just an entity that contains the JobManagerComponent

AMZN-Olex: Last I looked, a system entity had an ID of 0. Has that changed?

lumberyard-employee-dm: The System Entity does have an entity ID of zero, but the Entity class default constructor does [NOT generate](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Component/Entity.cpp#L82-L85) an Entity with ID 0.
It generates a psuedo random ID by combining the process ID + the machine ID

AMZN-Olex: Oh, good catch! Fixed.

mbalfour-amzn (approved):

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
```diff
@@ -126,6 +128,8 @@ namespace Multiplayer
     AZ_CVAR(AZ::TimeMs, bg_captureTransportPeriod, AZ::TimeMs{1000}, nullptr, AZ::ConsoleFunctorFlags::DontReplicate,
         "How often in milliseconds to record transport metrics.");
 
+    AZ_CVAR(bool, sv_multithreadedConnectionUpdates, false, nullptr, AZ::ConsoleFunctorFlags::DontReplicate,
```
AMZN-Gene, line 131: Maybe update MultiplayerSample setreg to keep this enabled? Disabled cvars tend to wilt over time.

Issue #13891, opened 2023-01-04
Material Canvas: Connecting Refraction Vector node to Experimental PBR node causes .materialgraph to fail in Asset Processor


LB-ArturZieba:
**Describe the bug**
If a Refraction Vector node is connected directly or indirectly (for example by pathing it through a different node) to the Experimental PBR node, the .materialgraph will fail processing in the Asset Processor. Please refer to the **Video and Asset Processor log** section for full log.

This issue occurs on Linux as well as Windows (DX12 RHI).

**Steps to reproduce**
1. Go to Tools > Material Canvas (Preview).
2. In the Material Canvas, right-click on the graph, find the Experimental PBR node and select it.
3. Add a Refraction Vector node the same way.
4. Connect the Refraction Vector's *Value* output pin to a compatible input pin on the Experimental PBR node (for example *Base color*).

**Expected behavior**
Asset Processor successfully process the .materialgraph asset.

**Actual behavior**
Asset Processor fails to process the .materialgraph asset.

**Video and Asset Processor log**

https://user-images.githubusercontent.com/86952082/210551148-68f589cb-515e-4e9e-b369-59d344126220.mp4

[untitled_ForwardPass.shader-2392495840-14111.log](https://github.com/o3de/o3de/files/10343831/untitled_ForwardPass.shader-2392495840-14111.log)

**Found in Branch**
Development (9849ab8)

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
9849ab86323026e59189aa38b5451ef03bd36d44

**Desktop**
 - Device: PC
 - OS: Linux
 - Version: Ubuntu 20.04.3 LTS
 - CPU AMD Ryzen 5 3600
 - GPU NVIDIA GeForce RTX 2060 SUPER
 - Memory 16GB

LB-KatarzynaDylska:
Issue verified as fixed in Development (ecf542f) build. All the mentioned nodes have been deprecated.

Issue #13876, opened 2023-01-03
Material Canvas: Some nodes lack tooltips in Node Palette


LB-MateuszSzweda:
**Describe the bug**
5 of Material Canvas nodes lack descriptions when the mouse cursor is hoovered over them in the Node Palette. These nodes are:

**Material Outputs > Experimental PBR
Math Functions > Step
Math Functions > Transform2
Math Functions > Transform3
Math Functions > Transform4
Utilities > Node Group**

Please note that this issue occurs on both dx12 and Vulkan enabled Material Canvas.

For further information please refer to the video file attached below.

**Steps to reproduce**
1. Open the Material Canvas.
2. Hoover the mouse cursor over **Node Group**, **Transform2**, **Transform3**, **Transform4** , **Step** or **Experimental PBR**
node in Node Palette.

**Expected behavior**
Nodes display their descriptions when the mouse cursor hoovers over them.

**Actual behavior**
**Node Group**, **Transform2**, **Transform3**, **Transform4** , **Step** or **Experimental PBR** nodes do not
display descriptions when the mouse cursor is hoovered over them.

**Screenshots/Video**

https://user-images.githubusercontent.com/86953456/210373250-f72e32f6-2848-4e82-8070-e91de70be633.mp4

**Found in Branch**
Development

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
https://github.com/o3de/o3de/commit/8af2ef552ccf8a0e5e0724b02d25121654c6a146

**Desktop/Device:**
- Device: PC
- OS: Windows
- Version: 10 Pro
- CPU: AMD Ryzen 9 3900X
- GPU: NVIDIA GeForce RTX 2060 SUPER
- Memory: 16GB

LB-MateuszSzweda:
Issue verified as fixed in Development (2bd2ade) .

PR #14018, opened 2023-01-12
Material Canvas: Shader options support and replaced experimental PBR with node templates derived from base and standard PBR

gadams3:
## What does this PR do?

Implemented settings and support for specifying shader options in no dance slots.

Started by attempting to update experimental PBR to support options and transparency. Experimental PBR was originally based off of minimal PBR, which was only meant to be test data and is not fully featured. It did not implement all of the necessary shader or functor support to manage enabling and disabling shaders, switching between draw lists. While experimental PBR was upgraded to have shader options, it was simpler to remove transparency and create new templates based off of base and standard PBR.

The base PBR node and template is directly derived from and references base PBR shader code. The vertex, pixel, and surface evaluation logic as well as the SRG and material type are replaced by the code and SRG generated by material canvas.

The standard PBR version does the exact same thing. It currently does not enable parallax or displacement because the functions that implemented need to be refactored to allow supplying a direct depth value instead of expecting it to be sampled from textures, as was the case with the normal standard PBR material.

Both require custom shaders four depth, shadows, and motion in order to support vertex deformations from the graph on all passes. The motion vector code still needs to be updated to support deformations.

This change also fixes a syntax error inside the blank node configuration template that prevented creating new blank node types in the editor.

This change also removes the refraction node which was implemented using an intrinsic function that is no longer supported.

Fixed dynamic node to register property slots before others.
Added support for enum values to dynamic node.
Fixed dynamic property to support combo boxes with string lists.
Added missing pragmas to shader include files.
Removed default values from existing node configurations to reduce file footprint.

Current issues:
I will get one of the other rendering engineers to look into a problem with the new node shader code that appears to be rendering a wire frame version of the model on top of itself. There may be some other setting or pass affecting things. I’ll continue to investigate with pix after material pipeline changes have made it to development.

There are currently 14 shader config files generated from the standard PBR template. Some of those files may no longer be needed after fully integrating the material pipeline changes. We want to keep this from ballooning as people create more materials and shaders with material canvas. There are multiple options to resolve this. If material graph compilation is moved to its own builder, then the shader source files can be generated into the intermediate asset folder instead of dumping them into the source folder with the graph. Regardless of that, we should consider updating the material type to allow embedding the shader source data config directly inside the material type. That option would help with packaging and make it easier to view all of the shader settings in one file. One last option would require general support for zip or pack files to contain all of the generated data.

https://github.com/o3de/sig-graphics-audio/issues/51
Resolves https://github.com/o3de/o3de/issues/13891
Resolves https://github.com/o3de/o3de/issues/13876

Most of the changes are data files. Feel free to filter out node data changes and everything except the C++ related files unless reviewing shader code.
image

## How was this PR tested?

https://user-images.githubusercontent.com/82461473/212164895-f9251c63-59ed-4608-909e-544ccded0320.mp4


Gems/GraphModel/Code/Source/Model/Slot.cpp
```diff
@@ -495,6 +508,11 @@ namespace GraphModel
         return m_slotDefinition->GetDescription();
     }
 
+    const AZStd::vector<AZStd::string>& Slot::GetEnumValues() const
+    {
+        return m_slotDefinition->GetEnumValues();
+    }
+
```
cgalvan, line 515: Nice!

cgalvan (approved):
GraphModel and other C++ changes LGTM! Will differ to atom folks for the shader changes.

Gems/GraphModel/Code/Source/Model/Slot.cpp
gadams3: Thanks! I will also follow up and expose the enum values so they can show up on slots. The graph controller is setting up most of the supported types except for those that require additional data like this one.

Gems/Atom/Tools/MaterialCanvas/Code/Source/Document/MaterialGraphCompiler.cpp
```diff
@@ -422,9 +424,11 @@ namespace MaterialCanvas
         switch (score)
         {
         case 4:
-            return ConvertToVector<AZ::Vector4>(slotValue);
+            // Skipping color to vector conversions that they export as the correct type with the material type.
+            return slotValue.is<AZ::Color>() ? slotValue : ConvertToVector<AZ::Vector4>(slotValue);
         case 3:
-            return ConvertToVector<AZ::Vector3>(slotValue);
+            // Skipping color to vector conversions that they export as the correct type with the material type.
+            return slotValue.is<AZ::Color>() ? slotValue : ConvertToVector<AZ::Vector3>(slotValue);
```
invertednormal, line 431: nit: The comments reads strangely, is it supposed to be "...so that they..." ?

invertednormal (approved):
Checked everything except .lua .py and .materialgraph* files. Looks good, just a couple small nits.

Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/GraphData/Nodes/MaterialOutputs/StandardPBR/MaterialGraphName_DepthPassTransparentMax.shader
```diff
@@ -0,0 +1,24 @@
+{
+    "Source" : "./MaterialGraphName_Depth.azsl",
+
+    "RasterState": { "CullMode": "None" },
+
+    "DepthStencilState" : { 
+        // Note that we assuming reversed depth buffer, which normally means we 
+        // are rendering with GreaterEqual. But in our case we want to render the maximum (furthest) values from the camera.
```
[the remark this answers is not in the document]
gadams3: Updated the comments

Gems/Atom/Tools/MaterialCanvas/Code/Source/Document/MaterialGraphCompiler.cpp
gadams3: Comments updated

Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/GraphData/Nodes/MaterialOutputs/StandardPBR/MaterialGraphName_MeshMotionVector.azsl
```diff
@@ -0,0 +1,82 @@
+/*
```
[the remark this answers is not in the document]
gadams3: The files are currently identical. I copied them from the original motion vector shader and made some changes. I think I'll need a way to apply offsets and transformations from material canvas to the current and previous positions. If both AZSL files end up being the same after that then I can move them to a common folder. I'll do the same with the functors now that I'm no longer copying them when generating new material types.

Issue #13981, opened 2023-01-10
CCD and Kinematic settings can be turned on at the same time


LB-KacperMetelski:
**Describe the bug**
CCD and Kinematic settings can be turned on simultaneously and can't be turned off afterwards. The following error is printed in console afterwards:
> [Error] (PhysX) - PxErrorCode 4: RigidBody::setRigidBodyFlag: kinematic bodies with CCD enabled are not supported! CCD will be ignored. (line 476 in /tmp/tmpy49ykj1y/buildtrees/physx/src/38fe6fd5fa-764b391545.clean/physx/source/physx/src/NpRigidBodyTemplate.h)

This issue is related to the following commit - [a6943b6779bccd8f8b5a285e965f63307124279a](https://github.com/o3de/o3de/commit/a6943b6779bccd8f8b5a285e965f63307124279a)

**Steps to reproduce**
1. Create an entity with the PhysX Rigid Body Component.
2. Turn Kinematic ON.
3. Turn the CCD ON.

**Expected behavior**
CCD setting becomes read only when Kinematic is enabled and vice versa.

**Actual behavior**
An error occurs. Both CCD and Kinematic settings cannot be disabled anymore.

**Screenshots/Video**
https://user-images.githubusercontent.com/110163473/211572915-c6d1ad73-d2dd-4713-a74b-15f0863b3e5b.mp4

**Found in Branch**
Development

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
446de8a

**Desktop/Device (please complete the following information):**
- Device: PC
- OS: Linux
- Version: Ubuntu 20.04.5 LTS
- CPU: i7-17000F
- GPU: RTX-3070 MSI
- Memory: 32GB

AMZN-daimini:
@amzn-ahmadkrm FYI :)

LB-KacperMetelski:
Issue doesn't occur as of commit 681d577; CCD and Kinematic mode can't be enabled at the same time.

PR #14024, opened 2023-01-12
Widget for Kinematic property

amzn-ahmadkrm:

## What does this PR do?

This PR adds in a dialog box to the Type setting of rigid bodies, the dialog helps users understand the difference between the two movement types Dynamic and Kinematic.

An original UX goal was to have the combo box disabled so the user has to go through the dialog to change the type but this could be tedious for the user to open if they want to quickly switch between types. If there was an option to acknowledge the dialog and then have the combo box unlocked it would be preferable.

https://user-images.githubusercontent.com/105638312/213267912-a5d1b51c-0cf6-4a08-a9f8-149a1c5b9ee0.mp4

This PR also fixes https://github.com/o3de/o3de/issues/13981.

## How was this PR tested?

This PR was tested manually.


Gems/PhysX/Code/Editor/ComboBoxEditButtonPair.cpp
```diff
@@ -29,6 +29,25 @@ namespace PhysX
 
             layout->addWidget(m_comboBox);
             layout->addWidget(m_editButton);
+
+            connect(m_comboBox, SIGNAL(currentIndexChanged(int)), this, SLOT(onChildComboBoxValueChange(int)));
```
cgalvan, line 33: Can use the updated syntax here:
```suggestion
            connect(m_comboBox, &QComboBox::currentIndexChanged, this, &ComboBoxEditButtonPair::onChildComboBoxValueChange);
```

Gems/PhysX/Code/Editor/ComboBoxEditButtonPair.cpp
```diff
@@ -29,6 +29,25 @@ namespace PhysX
...
+        void ComboBoxEditButtonPair::setValue(bool value) 
+        {
+            m_comboBox->blockSignals(true);
+            m_comboBox->setCurrentIndex(value ? 1 : 0);
+            m_comboBox->blockSignals(false);
+        }
+
+        void ComboBoxEditButtonPair::onChildComboBoxValueChange(int value)
+        {
+            emit valueChanged(value == 0 ? false : true);
```
cgalvan, line 50: This logic is referring to the value, but it looks like it's actually operating on the index instead. Should either be made more generic to handle any value, or should be changed to `CurrentIndex()` / `SetCurrentIndex()`

Gems/PhysX/Code/Editor/ComboBoxEditButtonPair.h
```diff
@@ -21,11 +21,23 @@ namespace PhysX
         class ComboBoxEditButtonPair
```
cgalvan, line 21: Maybe outside the scope here, but this class is actually duplicated in the `NvCloth` gem. We should either move it to `AzQtComponents` so it can have a single class, or could remove it all-together and just update the existing combobox handler to add support for an edit icon/action.

Gems/PhysX/Code/Editor/KinematicDescriptionDialog.h
```diff
@@ -0,0 +1,40 @@
...
+#pragma once
+
+#if !defined(Q_MOC_RUN)
+#include <QDialog>
+#include <AzFramework/Physics/Configuration/RigidBodyConfiguration.h>
+#endif
+
+class QTreeView;
+
+namespace Ui {
```
cgalvan, line 18: Nit: `{` on newline

Gems/PhysX/Code/Editor/KinematicDescriptionDialog.h
```diff
@@ -0,0 +1,40 @@
...
+class KinematicDescriptionDialog
+    : public QDialog
+{
+    Q_OBJECT
+
+public:
+    KinematicDescriptionDialog(bool kinematicSetting, QWidget* parent = nullptr);
+    ~KinematicDescriptionDialog();
+
+    bool GetResult() const { return m_kinematicSetting; }
```
cgalvan, line 31: Would be cleaner to move this to the cpp instead.

Gems/PhysX/Code/Editor/KinematicWidget.cpp
```diff
@@ -0,0 +1,99 @@
...
+namespace PhysX
+{
+    namespace Editor
+    {
+        AZ::u32 KinematicWidget::GetHandlerName() const
+        {
+            return Physics::Edit::KinematicSelector;
+        }
+
+        QWidget* KinematicWidget::CreateGUI(QWidget* parent)
```
cgalvan, line 24: As mentioned above, if we updated the existing combo box handler to support an edit button, then we wouldn't need to introduce an entirely new handler.

cgalvan (changes requested):

Gems/PhysX/Code/Editor/KinematicDescriptionDialog.cpp
```diff
@@ -0,0 +1,117 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#include "KinematicDescriptionDialog.h"
```
moraaar, line 9: Does it work using `<>` instead of `""` to rely on the cmake paths rather than local paths?

Gems/PhysX/Code/Editor/KinematicDescriptionDialog.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+AZ_PUSH_DISABLE_DLL_EXPORT_MEMBER_WARNING
+#include <Editor/ui_KinematicDescriptionDialog.h>
+AZ_POP_DISABLE_DLL_EXPORT_MEMBER_WARNING
+
+static constexpr AZStd::string_view dynamicDescription = "With <b>Dynamic</b> rigid bodies, you can physics forces "
```
moraaar, line 15: ` you can physics forces` -> ` you can use physics forces`

Gems/PhysX/Code/Editor/KinematicDescriptionDialog.cpp
```diff
@@ -0,0 +1,117 @@
...
+                                                  "(gravity, collision, etc.) to control the movement and "
+                                                  "position of the object.";
+
+static constexpr AZStd::string_view kinematicDescription = "With <b>Kinematic</b> rigid bodies, you can use Transform to "
```
moraaar, line 19: All these variables are global. The name convention is that they start with upper case: `KinematicDescription `. 

Same with the other variables.

Gems/PhysX/Code/Editor/KinematicDescriptionDialog.cpp
```diff
@@ -0,0 +1,117 @@
...
+                                                    "control the movement and position of the object. ";
+
+static constexpr AZStd::string_view moveWithCode = "Move with code";
+static constexpr AZStd::string_view moveManually = "Move manually";
+static constexpr AZStd::string_view collisions = "Collisions";
+static constexpr AZStd::string_view gravity = "Gravity";
```
moraaar, line 25: I'd recommend to add `Text` at the end of the variables' name so it's clear what the variable contains, for example `gravity` -> `gravityText`. Same with `moveWithCode`, `moveManually` and `collisions`.

Gems/PhysX/Code/Editor/KinematicDescriptionDialog.h
```diff
@@ -0,0 +1,40 @@
...
+
+private:
+    void OnButtonClicked();
+    void InitializeButtons();
+    void UpdateDialogText();
+
+    QScopedPointer<Ui::KinematicDescriptionDialog> m_ui;
+    bool m_kinematicSetting;
```
moraaar, line 39: Please assign a default value here to bool.

Gems/PhysX/Code/Editor/KinematicDescriptionDialog.cpp
```diff
@@ -0,0 +1,117 @@
...
+}
+
+void KinematicDescriptionDialog::InitializeButtons()
+{
+    connect(m_ui->dynamicRadioButton, &QPushButton::clicked, this, &KinematicDescriptionDialog::OnButtonClicked);
+    connect(m_ui->kinematicRadioButton, &QPushButton::clicked, this, &KinematicDescriptionDialog::OnButtonClicked);
+    connect(m_ui->dynamicBox, &QGroupBox::clicked, this, &KinematicDescriptionDialog::OnButtonClicked);
+    connect(m_ui->kinematicBox, &QGroupBox::clicked, this, &KinematicDescriptionDialog::OnButtonClicked);
+
+    m_kinematicSetting ? m_ui->kinematicRadioButton->setChecked(true) : m_ui->dynamicRadioButton->setChecked(true);
```
moraaar, line 54: I find it odd to use ternary operator for branching without being assigned to something, so it reads like something it's missing and leaves it wondering if it's intentional of if `setChecked` actually returns something but it's just an error not assign it to a variable. For these cases I would just use a regular `if` block.

````
if (m_kinematicSetting )
{
    m_ui->kinematicRadioButton->setChecked(true);
}
else
{
    m_ui->dynamicRadioButton->setChecked(true);
}
````

File: Gems/PhysX/Code/Editor/KinematicDescriptionDialog.cpp [hunk 13]
```diff
@@ -0,0 +1,117 @@
...
+
+    // running OnButtonClicked manually to get initial border highlight
+    OnButtonClicked();
+}
+
+void KinematicDescriptionDialog::OnButtonClicked()
+{
+    QString boxStyleSheet = QString("background-color: rgb(51, 51, 51);");
+    QString boxBorderStyleSheet = QString(" border: 1px solid rgb(30, 112, 235);");
```
moraaar, line 63: `boxStyleSheet ` and `boxBorderStyleSheet` cand be const.

moraaar (commented):
Excellent work! I have some feedback, but nothing major.

Code/Framework/AzFramework/AzFramework/Physics/Configuration/RigidBodyConfiguration.cpp
```diff
@@ -105,22 +105,22 @@ namespace AzPhysics
         }
     }
 
-    constexpr AZStd::string_view KinematicDescription =
-        "When active, the rigid body is not affected by gravity or other forces and is moved by script. ";
+    static constexpr AZStd::string_view KinematicDescription =
+        "Determines how the movement / position of the rigid body is controlled.";
```
hultonha, line 109: uber nit: I'd omit the space around the `/` character (so just `movement/position`) - repeated below

Gems/PhysX/Code/Editor/ComboBoxEditButtonPair.cpp
hultonha: IIRC it needs the `QOverload` expression to work too

Gems/PhysX/Code/Editor/ComboBoxEditButtonPair.cpp
hultonha: +1 to @moraaar's comment - could you clarify what you mean @cgalvan? Do the `CurrentIndex/SetCurrentIndex` calls already exist?

Gems/PhysX/Code/Editor/ComboBoxEditButtonPair.h
hultonha: Might be nice to do as a follow-up PR I'd say

Gems/PhysX/Code/Editor/ComboBoxEditButtonPair.h
```diff
@@ -21,11 +21,23 @@ namespace PhysX
...
+            bool value() const;
+
+        signals:
+            void valueChanged(bool newValue);
+
+        public slots:
+            void setValue(bool val);
+
+        protected slots:
+            void onChildComboBoxValueChange(int value);
```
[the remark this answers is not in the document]
hultonha: Just doubly checking are these overrides too btw (from Qt functions that already exist?) in that case we should add `override` specifier after the member function

Gems/PhysX/Code/Editor/KinematicDescriptionDialog.cpp
hultonha: +1

Gems/PhysX/Code/Editor/KinematicDescriptionDialog.cpp
hultonha: +1

Gems/PhysX/Code/Editor/KinematicDescriptionDialog.cpp
hultonha: or `Description` or `Name` works too, not sure what we use most often but being a little more explicit is definitely a good idea

Gems/PhysX/Code/Editor/KinematicDescriptionDialog.cpp
```diff
@@ -0,0 +1,117 @@
...
+static constexpr AZStd::string_view moveManually = "Move manually";
+static constexpr AZStd::string_view collisions = "Collisions";
+static constexpr AZStd::string_view gravity = "Gravity";
+
+static constexpr AZStd::string_view impactIcon = "<img src=\":/stylesheet/img/16x16/impact.svg\"/>";
+static constexpr AZStd::string_view gravityIcon = "<img src=\":/stylesheet/img/16x16/gravity.svg\"/>";
+static constexpr AZStd::string_view moveManuallyIcon = "<img src=\":/stylesheet/img/16x16/move_manually.svg\"/>";
+static constexpr AZStd::string_view moveWithCodeIcon = "<img src=\":/stylesheet/img/16x16/move_with_code.svg\"/>";
+
+KinematicDescriptionDialog::KinematicDescriptionDialog(bool kinematicSetting, QWidget* parent)
```
hultonha, line 32: update: It looks like all this code has not been added to the PhysX namespace, please make sure this is all nested inside `namespace PhysX`

Gems/PhysX/Code/Editor/KinematicDescriptionDialog.h
hultonha: +1

Gems/PhysX/Code/Editor/KinematicWidget.cpp
hultonha: As we took a while to get here I'd rather add this as-is for now and then see if we can refactor/unify after if possible

hultonha (commented):
Changes look great! I have a few comments but nothing major, mostly small things. Great work @amzn-ahmadkrm! 🥳

Gems/PhysX/Code/Editor/KinematicDescriptionDialog.cpp
```diff
@@ -0,0 +1,117 @@
...
+void KinematicDescriptionDialog::OnButtonClicked()
+{
+    QString boxStyleSheet = QString("background-color: rgb(51, 51, 51);");
+    QString boxBorderStyleSheet = QString(" border: 1px solid rgb(30, 112, 235);");
+
+    if (m_ui->dynamicRadioButton->isChecked())
+    {
+        m_ui->dynamicBox->setStyleSheet(boxStyleSheet + boxBorderStyleSheet);
+        m_ui->kinematicBox->setStyleSheet(boxStyleSheet + " border: none;");
+        m_kinematicSetting = false;
```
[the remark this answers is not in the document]
amzn-ahmadkrm: The acted upon "box" is different in each branch, I feel like it would be more convoluted to make a whole function for this right?

moraaar (approved):

Code/Framework/AzFramework/AzFramework/Physics/PropertyTypes.h
```diff
@@ -15,6 +15,7 @@ namespace Physics
     namespace Edit
     {
         const static AZ::Crc32 CollisionLayerSelector = AZ_CRC_CE("CollisionLayerSelector");
+        const static AZ::Crc32 KinematicSelector = AZ_CRC_CE("KinematicSelector");
         const static AZ::Crc32 CollisionGroupSelector = AZ_CRC_CE("CollisionGroupSelector");
```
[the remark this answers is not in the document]
hultonha: Thanks!

Gems/PhysX/Code/Editor/ComboBoxEditButtonPair.cpp
```diff
@@ -42,3 +61,5 @@ namespace PhysX
         }
     }
 }
```
hultonha, line 63: uber nit: Could we add closing namespaces here

```c++
    } // namespace Editor
} // namespace PhysX
```

Gems/PhysX/Code/Editor/KinematicDescriptionDialog.cpp
```diff
@@ -0,0 +1,136 @@
...
+                m_ui->invalidIcon2->setText(MoveManuallyIcon);
+            }
+        }
+
+        bool KinematicDescriptionDialog::GetResult() const
+        {
+            return m_kinematicSetting;
+        }
+    }
+}
```
hultonha, line 134: uber nit: Closing namespaces would be great here too (clang-format will add these automatically by the way which is great for new files)

Gems/PhysX/Code/Editor/KinematicDescriptionDialog.h
```diff
@@ -0,0 +1,47 @@
...
+        private:
+            void OnButtonClicked();
+            void InitializeButtons();
+            void UpdateDialogText();
+
+            QScopedPointer<Ui::KinematicDescriptionDialog> m_ui;
+            bool m_kinematicSetting = false;
+        };
+    }
+}
```
hultonha, line 46: nit: closing namespaces again sorry

Gems/PhysX/Code/Editor/KinematicDescriptionDialog.h
```diff
@@ -0,0 +1,47 @@
...
+namespace Ui
+{
+    class KinematicDescriptionDialog;
+}
+
+namespace PhysX
+{
+    namespace Editor
+    {
+        class KinematicDescriptionDialog : public QDialog
```
hultonha, line 27: suggestion: It would be nice to add a brief API comment for this type explaining what it does and why it's needed to help future maintainers

hultonha (approved):
I only have a couple of tiny bits of feedback now, otherwise looks great, thanks for making the updates @amzn-ahmadkrm 👍

Gems/PhysX/Code/Editor/ComboBoxEditButtonPair.cpp
cgalvan: What I mean is that this logic is assuming the combo box will only ever have 2 values, so it's a lot of extra boilerplate for this class `ComboBoxEditButtonPair` where we could instead just add support to the generic combo box handler for an edit attribute (I mentioned this in another comment as well).

cgalvan (approved):
Had an offline conversation with Ahmad. There will be a separate follow-up PR to extend the edit button action support to the base combo box handler similar to other fields so that we can remove the `ComboBoxEditButtonPair` class entirely.

hultonha:
> Had an offline conversation with Ahmad. There will be a separate follow-up PR to extend the edit button action support to the base combo box handler similar to other fields so that we can remove the `ComboBoxEditButtonPair` class entirely.

That's great to hear @cgalvan and sounds like a good plan! 👍 Thanks for the update

Issue #13115, opened 2022-11-10
Atom: Editor can crash on startup in ShaderResourceGroupLayout.cpp


jckand-amzn:
**Describe the bug**
Editor can crash on startup in ShaderResourceGroupLayout.cpp. This has caused a Python failure in the Dynamic Vegetation automated test test_PrefabInstanceSpawner_Embedded_E2E_Editor.

https://jenkins.build.o3de.org/job/O3DE/job/development/3427/testReport/junit/AutomatedTesting.Gem.PythonTests.largeworlds.dyn_veg.TestSuite_Main/TestAutomation/Build___Windows__profile_pipe____test_cpu_profile_results___test_PrefabInstanceSpawner_Embedded_E2E_Editor_windows_AutomatedTesting_windows_editor_/

```
[2022-11-10T08:39:55.351Z] ================================== FAILURES ===================================

[2022-11-10T08:39:55.351Z] _ TestAutomation.test_PrefabInstanceSpawner_Embedded_E2E_Editor[windows-AutomatedTesting-windows_editor] _

[2022-11-10T08:39:55.351Z] d:\workspace\o3de\tools\lytesttools\ly_test_tools\o3de\multi_test_framework.py:431: in single_run

[2022-11-10T08:39:55.351Z]     self._run_single_test(request, workspace, collected_test_data, inner_test_spec)

[2022-11-10T08:39:55.351Z] d:\workspace\o3de\tools\lytesttools\ly_test_tools\o3de\multi_test_framework.py:794: in _run_single_test

[2022-11-10T08:39:55.351Z]     self._report_result(test_name, test_result)

[2022-11-10T08:39:55.351Z] d:\workspace\o3de\tools\lytesttools\ly_test_tools\o3de\multi_test_framework.py:1319: in _report_result

[2022-11-10T08:39:55.351Z]     pytest.fail(error_str)

[2022-11-10T08:39:55.351Z] E   Failed: Test test_PrefabInstanceSpawner_Embedded_E2E_Editor:

[2022-11-10T08:39:55.351Z] E   Test CRASHED, return code 0xc0000005

[2022-11-10T08:39:55.351Z] E   ---------------

[2022-11-10T08:39:55.351Z] E   |  Stacktrace |

[2022-11-10T08:39:55.351Z] E   ---------------

[2022-11-10T08:39:55.351Z] E   Logged at Thu Nov 10 08:37:53 2022

[...]

Issue #6139, opened 2021-12-03
The Editor crashes upon launching


jborut:
**Describe the bug**
The project (with default gems) created with Project Manager (GUI) can't be opened. Editor crashes with an error message saying that Shaders/SimpleTextured.azshader, Fonts/default.font, and Fonts/default-ui.font can't be found.

![image](https://user-images.githubusercontent.com/31534934/144654509-f46d1c49-e7d7-44eb-b7a2-f57ce795d37f.png)

**Steps to reproduce**
1. Install O3DE with windows installer
2. Open O3DE Project Manager
3. Click Create new Project
4. Click Create (with default gems selection)
5. Once the project is created, build it
6. When the project is built, click Open Editor

**Expected behavior**
When I click Open Editor on a project in Project Manager, Editor should open.

**Actual behavior**
When I click Open Editor on a project in Project Manager, the above described error message pops up and editor doesn't open.

**Found in Branch**
Release 21.11.1

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Windows
 - Version 10
 - CPU Ryzen 5900x
 - GPU NVidia 1050Ti
 - Memory 32GB

**Additional context**
This problem is probably related to this https://github.com/o3de/o3de/issues/3955

stephanrb3:
I'm running into the same problem.

WayneKoorts:
Same problem, I just downloaded this for the first time and set up a new project with all defaults (except project location), built the project from the launcher but can't launch the editor because of this error.

lumberyard-employee-dm:
I have a fix for this issue up for PR now https://github.com/o3de/o3de/pull/6234

wangbq96:
I cloned and built the latest source code but still encountered this problem. 
```
Logged at Monday, January 10, 2022 20:12:51
FileVersion: 0.0.0.0
ProductVersion: 0.0.0.0
LogFile: @log@/Editor.log
ProjectDir: C:\Users\boquan\O3DE\Projects\NewProject
Executable: Editor.exe (debug: no
Exception Code: 0xC0000005
Exception Addr: 0x0033:0x00007FFC2ABC8811
Exception Module: 
Exception Description: EXCEPTION_ACCESS_VIOLATION, Attempt to read from address 0x00000448

The memory could not be "read"

Call Stack Trace:
25) 00007FFCC034B7F7 (KERNELBASE) : UnhandledExceptionFilter
24) 00007FFCC29551B0 (ntdll) : memset
[...]

PR #14027, opened 2023-01-12
GHI-13115 Atom: Editor can crash on startup in ShaderResourceGroupLayout.cpp

VickyAtAZ:
## What does this PR do?

Changes include:
- Move RPISystem::InitializeSystemAssets to after all system components are activated.
- Add initialize dependency of RPISystem to PrevewRendererSystemComponent.
- Move BootstrapSystemComponent's initalize function to after all system components are activated. 
 
The RPISystem initialization with assets was relying on "CriticalAssetsCompiled" event. If by any chance the event wasn't triggered, it may lead to the crash in the RHI. The changes in this commit remove that dependency. Instead, it adds the initialization all system components are activated and when the AZ::TickBus::ExecuteQueuedEvents() is called.

https://github.com/o3de/o3de/issues/13115
https://github.com/o3de/o3de/issues/6139

## How was this PR tested?
Tested with Editor (AutomatedTesting); AutomatedTesting.GameLauncher; ASV standalone; Material Editor

### DIFF
--- a/Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
+++ b/Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
@@ -219,24 +219,17 @@ namespace AZ
                 Render::Bootstrap::DefaultWindowBus::Handler::BusConnect();
                 Render::Bootstrap::RequestBus::Handler::BusConnect();
 
-                // If the settings registry isn't available, something earlier in startup will report that failure.
-                if (auto settingsRegistry = AZ::SettingsRegistry::Get(); settingsRegistry != nullptr)
-                {
-                    // Automatically register the event if it's not registered, because
-                    // this system is initialized before the settings registry has loaded the event list.
-                    AZ::ComponentApplicationLifecycle::RegisterHandler(
-                        *settingsRegistry, m_componentApplicationLifecycleHandler,
-                        [this](const AZ::SettingsRegistryInterface::NotifyEventArgs&)
+                // delay one frame for Initialize which asset system is ready by then
+                AZ::TickBus::QueueFunction(
+                    [this]()
+                    {
+                        Initialize();
+                        if (m_nativeWindow)
                         {
-                            Initialize();
-                            if (m_nativeWindow)
-                            {
-                                // wait until swapchain has been created before setting fullscreen state
-                                m_nativeWindow->SetFullScreenState(r_fullscreen);
-                            }
-                        },
-                        "CriticalAssetsCompiled");
-                }
+                            // wait until swapchain has been created before setting fullscreen state
+                            m_nativeWindow->SetFullScreenState(r_fullscreen);
+                        }
+                    });
             }
 
             void BootstrapSystemComponent::Deactivate()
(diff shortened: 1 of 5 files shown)


Gems/Atom/Tools/AtomToolsFramework/Code/Source/PreviewRenderer/PreviewRendererSystemComponent.cpp
```diff
@@ -36,6 +36,11 @@ namespace AtomToolsFramework
     {
         provided.push_back(AZ_CRC_CE("PreviewRendererSystem"));
     }
+    
+    void PreviewRendererSystemComponent::GetDependentServices(AZ::ComponentDescriptor::DependencyArrayType& dependent)
+    {
+        dependent.push_back(AZ_CRC("RPISystem", 0xf2add773));
```
gadams3, line 42:
```suggestion
        dependent.push_back(AZ_CRC_CE("RPISystem"));
```

gadams3 (approved):
Please also test the material editor

Gems/Atom/RPI/Code/Source/RPI.Public/RPISystem.cpp
```diff
@@ -379,7 +379,6 @@ namespace AZ
         {
             if (m_systemAssetsInitialized)
             {
-                AZ_Warning("RPISystem", false , "InitializeSystemAssets should only be called once'");
```
jiaweig-amzn, line 382: Can I ask why is it removed? Delaying the initialization doesn't seem to affect the warning.

jiaweig-amzn (approved):

Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
```diff
@@ -219,24 +219,17 @@ namespace AZ
...
-                                // wait until swapchain has been created before setting fullscreen state
-                                m_nativeWindow->SetFullScreenState(r_fullscreen);
-                            }
-                        },
-                        "CriticalAssetsCompiled");
-                }
+                            // wait until swapchain has been created before setting fullscreen state
+                            m_nativeWindow->SetFullScreenState(r_fullscreen);
+                        }
+                    });
```
rgba16f, line 232: Does this still work if the Cache is empty before you start the system?

VickyAtAZ:
> Please also test the material editor

It was tested with material editor
"Tested with Editor (AutomatedTesting); AutomatedTesting.GameLauncher; ASV standalone; Material Editor"

Gems/Atom/RPI/Code/Source/RPI.Public/RPISystem.cpp
VickyAtAZ: I think it's okay to have the function be called more than once and it shouldn't be warning since it's handled gracefully. 
Currently the material editor manually call this function manually and it's earlier than the queued function RPISystemComponent did. so it would be called twice. and there is no harm the function is called twice.

Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
VickyAtAZ: Yes it will still work. 
The  CCryEditApp::ConnectToAssetProcessor() and  CCryEditApp::CompileCriticalAssets() is called after system components activated but before handle the queued functions.

rgba16f (approved):

rgba16f (approved):

PR #14028, opened 2023-01-12
Allow adjusting scan folder priority for project-relative Gems

### LINKED ISSUE #8480: Gems: Assets paths for gems local to a project are incorrect
**Describe the bug**
Assets paths for gems local to a project are incorrect

**Steps to reproduce**
1. Create an external gem: `scripts\o3de.bat create-gem -gp C:\TestGem`
2. Add a mesh in the test gem Assets folder, for example TestMesh.fbx
3. Enable the gem for your project: `scripts\o3de enable-gem -gp C:\TestGem -pp C:\TestProject`
4. Put TestMesh in a scene of TestProject and save
5. Move `C:\TestGem` folder to `C:\TestProject\Gems\TestGem` to make it local to your project
6. Edit `C:\TestProject\project.json` to remove `"TestGem"` from `gem_names` and add `"Gems/TestGem"` to `external_subdirectories`
7. Delete `C:\TestProject\Cache` folder
8. Re-open the scene with TestMesh. It failed to load the mesh since the asset path was previously testmesh.azmodel and now it's gems/testgem/assets/testmesh.azmodel

**Expected behavior**
The `Assets` folder of local gems should be automatically added to the AssetProcessor ScanFolders like for gems external to the project.

**Actual behavior**
The gems assets paths are not the same if the gem is local or external to the project.
This cause issues for assets referenced in the scenes but also make the gem unusable if the code of the gem load assets bundle with it (shaders for example).

This is caused by the file `Registry/AssetProcessorPlatformConfig.setreg` in the engine which adds the project root folder at order 0:
```
"ScanFolder Project/Assets": {
    "watch": "@PROJECTROOT@",
    "display": "@PROJECTNAME@",
    "recursive": 1,
    "order": 0
},
```
It is possible to workaround this by adding a setreg file in the Registry folder of the project which adds the `Assets` folder of the gem with an order -1:
```
{
    "Amazon": {
        "AssetProcessor": {
            "Settings": {
                "ScanFolder TestGem Assets": {
                    "watch": "@PROJECTROOT@/Gems/TestGem/Assets",
                    "recursive": 1,
                    "order": -1
                }
            }
        }
    }
}
```
But it would probably be better if this could work without this.

**Found in Branch**
commit 9e96dcf on development branch
also on commit c1d7294 on main branch

amzn-phist:
## What does this PR do?

It may be desirable to have the scan folders of any project-relative Gems (Gems that reside within the project folder) be prioritized before the main Project scan folder.  By default, the project scan folder is priority 0, and all Gem scan folders start incrementing from starting value of 100, so Gem scan folders are always at a lower priority than the project.

This introduces two settings registry keys that can be used to tweak relative priorities:
`"/Amazon/AssetProcessor/Settings/GemScanFolderStartingPriorityOrder": 100`
This setting will define the starting priority order for all Gems.  The first Gem scan folder will be assigned 101 and increase value from there.

`"/Amazon/AssetProcessor/Settings/ProjectRelativeGemsScanFolderPriority": "none"`
This setting determines how you want project-relative Gem scan folders to be prioritized against the main project scan folder.
`"none"` (default): This means don't do anything special.  All project-relative Gem scan folders will be assigned priority order using the 'running value' that starts at the value defined in the above setting.
`"lower"`: All project-relative Gem scan folders will have a priority that is lower than the project.  Users could configure the priority order of the project's main scan folder to be lower than all Gems, but having a setting of "lower" here would force project-relative Gem scan folders to be lower than the project.
`"higher"`: All project-relative Gem scan folders will have a priority that is higher than the project.  Again, by default Gems have lower priority than the project, but having a setting of "higher" here would force project-relative Gem scan folders to be higher than the project.

fixes #8480 

## How was this PR tested?

Followed instructions from the GHI.  Was able to tweak relative priorities and get the assets in project-relative Gems to scan and be picked up.  More testing ongoing.

### DIFF
--- a/Code/Tools/AssetProcessor/native/utilities/PlatformConfiguration.cpp
+++ b/Code/Tools/AssetProcessor/native/utilities/PlatformConfiguration.cpp
@@ -1611,7 +1611,7 @@ namespace AssetProcessor
         auto* fileStateInterface = AZ::Interface<AssetProcessor::IFileStateRequests>::Get();
 
         // Only compute the intermediate assets folder path if we are going to search for and skip it.
-       
+
         if (skipIntermediateScanFolder)
         {
             if (m_intermediateAssetScanFolderId == -1)
@@ -1619,13 +1619,13 @@ namespace AssetProcessor
                 CacheIntermediateAssetsScanFolderId();
             }
         }
-        
+
         QString absolutePath; // avoid allocating memory repeatedly here by reusing absolutePath each scan folder.
         absolutePath.reserve(AZ_MAX_PATH_LEN);
 
         QFileInfo details(relativeName); // note that this does not actually hit the actual storage medium until you query something
         bool isAbsolute = details.isAbsolute(); // note that this looks at the file name string only, it does not hit storage.
-        
+
         for (int pathIdx = 0; pathIdx < m_scanFolders.size(); ++pathIdx)
         {
             const AssetProcessor::ScanFolderInfo& scanFolderInfo = m_scanFolders[pathIdx];
@@ -1879,16 +1879,11 @@ namespace AssetProcessor
 
     void PlatformConfiguration::AddGemScanFolders(const AZStd::vector<AzFramework::GemInfo>& gemInfoList)
     {
-        // Determine if a Gem is from the project.
-        // Cross-check gem paths against visited project gem paths.
-        // If the gem is project-relative, make adjustments to it's priority order.
-        // Registry Settings:
+        // Determine if a Gem is relative to the project. Cross-check gem paths against visited project gem paths.
+        // If the gem is project-relative, make adjustments to it's priority order based on registry settings:
         // /Amazon/AssetProcessor/Settings/GemScanFolderStartingPriorityOrder
-        //      The starting gem priority order
         // /Amazon/AssetProcessor/Settings/ProjectRelativeGemsScanFolderPriority
-        //      "none" - any project Gem scan folder priority will be incremented from the "GemScanFolderStartingPriorityOrder" value.
-        //      "lower" - each project Gem scan folder will be set to lower priority (higher numeric value) than the project scan folder.
-        //      "higher" - each project Gem scan folder will be set to higher priority (lower numeric value) than the project scan folder.
+        // See <o3de-root>/Registry/AssetProcessorPlatformConfig.setreg for more information.
 
         AZ::s64 gemStartingOrder = 100;
         AZStd::string projectGemPrioritySetting{};
--- a/Code/Tools/AssetProcessor/native/utilities/PlatformConfiguration.cpp
+++ b/Code/Tools/AssetProcessor/native/utilities/PlatformConfiguration.cpp
@@ -619,6 +619,9 @@ namespace AssetProcessor
 
     const char AssetConfigPlatformDir[] = "AssetProcessorConfig/";
     const char AssetProcessorPlatformConfigFileName[] = "AssetProcessorPlatformConfig.ini";
+    constexpr const char* ProjectScanFolderKey = "Project/Assets";
+    constexpr const char* GemStartingPriorityOrderKey = "/GemScanFolderStartingPriorityOrder";
+    constexpr const char* ProjectRelativeGemPriorityKey = "/ProjectRelativeGemsScanFolderPriority";
 
     PlatformConfiguration::PlatformConfiguration(QObject* pParent)
         : QObject(pParent)
@@ -1325,7 +1328,7 @@ namespace AssetProcessor
     {
         auto mainProjectScanFolder = FindScanFolder([](const AssetProcessor::ScanFolderInfo& scanFolderInfo) -> bool
             {
-                return scanFolderInfo.GetPortableKey() == "Project/Assets";
+                return scanFolderInfo.GetPortableKey() == ProjectScanFolderKey;
             });
         if (mainProjectScanFolder)
         {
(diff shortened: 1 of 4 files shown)


Code/Tools/AssetProcessor/native/utilities/PlatformConfiguration.cpp
```diff
@@ -1873,6 +1934,9 @@ namespace AssetProcessor
             {
                 const AZ::IO::Path& absoluteSourcePath = gemElement.m_absoluteSourcePaths[sourcePathIndex];
                 QString gemAbsolutePath = QString::fromUtf8(absoluteSourcePath.c_str(), aznumeric_cast<int>(absoluteSourcePath.Native().size())); // this is an absolute path!
+
+                bool isProjectGem = AZStd::find(projectGemPaths.begin(), projectGemPaths.end(), absoluteSourcePath) != projectGemPaths.end();
```
lumberyard-employee-dm, line 1938: You can use `AZStd::ranges::contain` here.
```suggestion
                const bool isProjectGem = AZStd::ranges::contain(projectGemPaths, absoluteSourcePath);
```

Code/Tools/AssetProcessor/native/utilities/PlatformConfiguration.cpp
```diff
@@ -1326,6 +1321,19 @@ namespace AssetProcessor
         }
     }
 
+    int PlatformConfiguration::GetProjectScanFolderOrder() const
+    {
+        auto mainProjectScanFolder = FindScanFolder([](const AssetProcessor::ScanFolderInfo& scanFolderInfo) -> bool
+            {
+                return scanFolderInfo.GetPortableKey() == "Project/Assets";
```
lumberyard-employee-dm, line 1328: Can you make the "Project/Asset" into a `constexpr const char*` constant?

lumberyard-employee-dm: Or if you want to get really "meta", make the ProjectScanFolder key a setting itself.

Code/Tools/AssetProcessor/native/utilities/PlatformConfiguration.cpp
```diff
@@ -1446,17 +1454,20 @@ namespace AssetProcessor
         return m_scanFolders[index];
     }
 
+    const AssetProcessor::ScanFolderInfo* PlatformConfiguration::FindScanFolder(
+        AZStd::function<bool(const AssetProcessor::ScanFolderInfo&)> predicate) const
+    {
+        const AssetProcessor::ScanFolderInfo* result = AZStd::ranges::find_if(m_scanFolders, predicate);
```
lumberyard-employee-dm, line 1460: It would be safer to use `auto` here actually.
Or barring that `typename AZStd::vector<AssetPRocessor::ScanFolderInfo>::const_iterator`

While the iterator type of m_scanFolders is an `AssetProcessor::ScanFolderInfo*` we shouldn't direct access it, as it would inhibit changing the `vector` to a `set`, `deque` or `list` in the future.

Code/Tools/AssetProcessor/native/utilities/PlatformConfiguration.cpp
lumberyard-employee-dm: Now that I think about this.
Instead of gathering the list of Project Gems, we can just check if the gem path is relative to the project root.

```suggestion
                const bool isProjectGem = gemAbsolutePath.IsRelativeTo(projectPath);
```

Code/Tools/AssetProcessor/native/utilities/PlatformConfiguration.cpp
```diff
@@ -1863,7 +1874,57 @@ namespace AssetProcessor
...
+            settingsRegistry->Get(projectGemPrioritySetting,
+                AZ::SettingsRegistryInterface::FixedValueString(AssetProcessorSettingsKey) + "/ProjectRelativeGemsScanFolderPriority");
+            AZStd::to_lower(projectGemPrioritySetting.begin(), projectGemPrioritySetting.end());
+
+            auto projectGemCallback = [&projectGemPaths]([[maybe_unused]] AZStd::string_view manifestObjectKey,
+                [[maybe_unused]] AZStd::string_view manifestObjectName, AZStd::string_view manifestRootPath)
+            {
+                projectGemPaths.push_back(AZ::IO::Path{ manifestRootPath });
+            };
+            AZ::SettingsRegistryMergeUtils::VisitProjectGems(*settingsRegistry, projectGemCallback);
```
lumberyard-employee-dm, line 1905: Instead of gathering a list of project gem paths from the SettingsRegistry, we can go with the cheaper option of just querying the project path.
```
AZ::IO::FixedMaxPath projectPath = AZ::Utils::GetProjectPath();
```

Registry/AssetProcessorPlatformConfig.setreg
```diff
@@ -155,6 +154,16 @@
                     "order": 40000
                 },
 
+                // Configurable starting prioirity for all Gem scan folders to use.
+                // Each Gem scan folder added will increment this.
+                "GemScanFolderStartingPriorityOrder": 100,
```
lumberyard-employee-dm, line 159: Can you update this comment to mention that due to incrementing this value, every successive gem has a lower priority to each other?
```suggestion
                // Configurable starting prioirity for all Gem scan folders to use.
                // Each Gem scan folder added will increment this.
                // NOTE: Each successive gem priority will be lower than the last
                "GemScanFolderStartingPriorityOrder": 100,
```

lumberyard-employee-dm (commented):
This looks good, just a few suggestion to make this even more data driven and to simplify the detection a gem being underneath the project.

Code/Tools/AssetProcessor/native/utilities/PlatformConfiguration.cpp
amzn-phist: I went with a string constant.

Code/Tools/AssetProcessor/native/utilities/PlatformConfiguration.cpp
amzn-phist: Fixed.

Code/Tools/AssetProcessor/native/utilities/PlatformConfiguration.cpp
amzn-phist: Simplified to this and used `IsRelativeTo` to check.

Registry/AssetProcessorPlatformConfig.setreg
amzn-phist: Done.

lumberyard-employee-dm (approved):

byrcolin (approved):

PR #14030, opened 2023-01-12
Add AZ_CLASS_ALLOCATOR macros to subclasses as appropriate

burelc-amzn:
This adds AZ_CLASS_ALLOCATOR macros to subclasses if:
* they did not have it before,
* they directly have data members,
* one of their subclasses did use the AZ_CLASS_ALLOCATOR

In addition, the `operator new()` method that is created by the AZ_CLASS_ALLOCATOR checks that the size of the allocation being requested matches exactly the size of the class that used the macro.

The AZ_CLASS_ALLOCATOR macro hard-codes the alignment used in the call
to the underlying allocator's allocate method to the alignment of the class that has used it: https://github.com/o3de/o3de/blob/446de8ac0fb3d5e803a7bf32f33fd04f50328de6/Code/Framework/AzCore/AzCore/Memory/Memory.h#L205-L209

If a derived class increases the alignment requirement,
`Base::operator new(size_t)` has no idea, and supplies the alignment of
Base to the allocate function. But only up to a point: if the alignment
of Derived is greater than `__STDCPP_DEFAULT_NEW_ALIGNMENT__` (which is
16), then the operator new overload that takes both a size and alignment
come into play. At that point, the correct alignment is provided to the
function as a parameter, and everything works again.

Attempting to store a misaligned address to a type that requires a
higher alignment may crash. But that only manifests itself when
alignof(Base) is 8 and alignof(Derived) is 16.

This would be causing problems all over the place, if it weren't for the
Serialize::InstanceFactory: https://github.com/o3de/o3de/blob/446de8ac0fb3d5e803a7bf32f33fd04f50328de6/Code/Framework/AzCore/AzCore/Serialization/SerializeContext.h#L1714-L1743
The InstanceFactory ignores inherited
`operator new()` methods and checks if a type T directly has used the
`AZ_CLASS_ALLOCATOR` macro. If it has, it uses new to create the instance
(aznew is now synonymous with new). If it didn't, it uses the real
alignment requirement of T, but forces the instance to be allocated with
the SystemAllocator. This gets around the incorrect alignment hard-coded
in Base::operator new(size), but if Base is configured to use a
different allocator, it would cause its own set of problems when delete
derived; is called.

I found this in action in the ConvertEmissiveUnitFunctorSourceData
class. Its AZ::Vector2 instance increases its alignment requirement to
16, when the base class MaterialFunctorSourceData's alignment is 8. This
means that simply writing new ConvertEmissiveUnitFunctorSourceData() may
crash. In practice though, nothing does this, because all the instances
are made by the Serialize::InstanceFactory.

This fixes the above problem by adding the AZ_CLASS_ALLOCATOR macro to
subclasses that don't have it. These were identified with the following
matcher to clang-query:

```
let hasOverloadedNew hasMethod(cxxMethodDecl(hasName("operator new"), parameterCountIs(1), anyOf(isExpandedFromMacro("AZ_CLASS_ALLOCATOR"), isExpandedFromMacro("AZ_CLASS_ALLOCATOR_DECL"))))
match cxxRecordDecl(isDefinition(), unless(hasOverloadedNew), has(fieldDecl()), hasAnyBase(hasType(cxxRecordDecl( hasOverloadedNew ))))
```


Gems/EMotionFX/Code/Source/Integration/System/AnimationModule.cpp
```diff
@@ -51,6 +51,7 @@ namespace EMotionFX
             : public CryHooksModule
         {
         public:
+            AZ_CLASS_ALLOCATOR(EMotionFXIntegrationModule, AZ::SystemAllocator)
```
amzn-phist, line 54: General question, why do most pass a `0` as third parameter, but some don't?

amzn-phist (approved):

Gems/EMotionFX/Code/Source/Integration/System/AnimationModule.cpp
burelc-amzn: The `AZ_CLASS_ALLOCATOR` macro used to take 3 parameters, but due to other changes the 3rd parameter stopped being used. It was made an optional parameter in https://github.com/o3de/o3de/commit/2e59035cfa79fe959acc272bbefb61ab09d23cd4 (line 246 of Memory.h)

Code/Framework/AzToolsFramework/AzToolsFramework/UI/LegacyFramework/Core/EditorFrameworkApplication.h
```diff
@@ -51,6 +51,7 @@ namespace LegacyFramework
     {
         /// Create application, if systemEntityFileName is NULL, we will create with default settings.
     public:
+        AZ_CLASS_ALLOCATOR(Application, AZ::SystemAllocator, 0);
```
spham-amzn, line 54: Question: How does the allocator resolve situations where the class name is duplicated in another namespace?

burelc-amzn: The `AZ_CLASS_ALLOCATOR` macro expands to a bunch of overloads of `operator new()` and `operator delete()`. In this case, it would expand to:
```cpp
namespace LegacyFramework
{
// ... snip ...
    class Application
        : public AZ::ComponentApplication
        , protected FrameworkApplicationMessages::Handler
        , protected CoreMessageBus::Handler
    {
    public:
        //AZ_CLASS_ALLOCATOR(Application, AZ::SystemAllocator, 0);
        // the above macro will add, for example:
        AZ_FORCE_INLINE void* operator new(std::size_t size) {
            AZ_Assert(size == sizeof(Application), "Size mismatch! Did you forget to declare the macro in derived class? Size: %d sizeof(%s): %d", size, "Application", sizeof(Application));
            AZ_Warning(0, true/*false*/, "Make sure you use aznew, offers better tracking! (%s)", "Application");
            return AZ::AllocatorInstance< AZ::SystemAllocator >::Get().allocate(size, AZStd::alignment_of< Application >::value);
        }
```

At this point, the normal C++ name lookup rules apply. `AZStd::alignment_of< Application >::value` will look in the current namespace (`LegacyFramework` in this case), but not in others.

spham-amzn: Thanks for the education

cgalvan (approved):

Gems/Atom/RHI/Code/Include/Atom/RHI.Reflect/BufferPoolDescriptor.h
```diff
@@ -23,6 +23,7 @@ namespace AZ
         public:
             virtual ~BufferPoolDescriptor() = default;
             AZ_RTTI(BufferPoolDescriptor, "{53074556-41D1-4246-8BF0-E5B096997C65}", ResourcePoolDescriptor);
+            AZ_CLASS_ALLOCATOR(BufferPoolDescriptor, SystemAllocator, 0);
```
spham-amzn, line 26: nit: Can this be moved to above AZ_RTTI? So far its been consistently declared before AZ_RTTI (if any) to this point. There are also a few others in the PR that are not in this order.

Gems/EMotionFX/Code/Source/Integration/System/AnimationModule.cpp
spham-amzn: If its not used, isn't it worth it to go back and remove it from the other files in this PR? I can see copy/paste situations where the unused parameter keeps getting perpetuated, and these are all new changes.

Gems/PhysX/Code/Source/HeightfieldCollider.h
```diff
@@ -122,6 +122,7 @@ namespace PhysX
         class HeightfieldUpdateJobContext : public AZ::JobContext
         {
         public:
+            AZ_CLASS_ALLOCATOR(HeightfieldUpdateJobContext, AZ::ThreadPoolAllocator, 0)
```
byrcolin, line 125: Id the  AZ::ThreadPoolAllocator correct here?

spham-amzn (approved):
LGMT, just minor comments

Gems/EMotionFX/Code/EMotionFX/Source/Importer/ChunkProcessors.cpp
```diff
@@ -69,6 +69,7 @@
 namespace EMotionFX
 {
     AZ_CLASS_ALLOCATOR_IMPL(SharedData, ImporterAllocator, 0)
+    AZ_CLASS_ALLOCATOR_IMPL(SharedHelperData, ImporterAllocator)
```
byrcolin, line 72: Should this be AZ_CLASS_ALLOCATOR_IMPL(SharedHelperData, ImporterAllocator, 0) ?

byrcolin (approved):
only 2 questions below, looks good.

Gems/EMotionFX/Code/Source/Integration/System/AnimationModule.cpp
lumberyard-employee-dm: Yeah, I address with @spham-amzn, if new version of the macro are being added, then they should all be added in a consistent manner.

Gems/Atom/RHI/Metal/Code/Include/Atom/RHI.Reflect/Metal/BufferPoolDescriptor.h
```diff
@@ -18,6 +18,7 @@ namespace AZ
             : public RHI::BufferPoolDescriptor
         {
         public:
+            AZ_CLASS_ALLOCATOR(BufferPoolDescriptor, SystemAllocator, 0);
```
lumberyard-employee-dm, line 21: Some of these calls are missing the AZ::SystemAllocator.
I know they are in the AZ namespace, but other calls added the "AZ" as well.

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponentMode.cpp
```diff
@@ -28,6 +28,8 @@
 
 namespace GradientSignal
 {
+    AZ_CLASS_ALLOCATOR_IMPL(EditorImageGradientComponentMode, AZ::SystemAllocator)
```
lumberyard-employee-dm, line 31: Any reason that the combination of the  EditorImageGradientComponentMode.cpp/.h files are using the `AZ_CLASS_ALLOCATOR_DECL` and `AZ_CLASS_ALLOCATOR_IMPL` paradigm?

Gems/PhysX/Code/Source/HeightfieldCollider.h
lumberyard-employee-dm: Yes, the JobContext base class uses [AZ::ThreadPoolAllocator](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Jobs/JobContext.h#L24)

lumberyard-employee-dm (approved):

Gems/Atom/RHI/Metal/Code/Include/Atom/RHI.Reflect/Metal/BufferPoolDescriptor.h
burelc-amzn: To make sure I was using the same allocator as the parent class, I used a macro in vim to add the `AZ_CLASS_ALLOCATOR` line. The macro did this:
* Go to the subclass's base class's definition
* Search for its `AZ_CLASS_ALLOCATOR` line, and copy it
* Go back to the subclass, paste the `AZ_CLASS_ALLOCATOR` line from the base class
* Replace the base class's name in the new macro with the subclass's name

The result of this is that if the base class spelled the allocator as `SystemAllocator` instead of `AZ::SystemAllocator`, then that's how it is spelled in the subclass's use as well. For this specific instance, the base class goes through `BufferPoolDescriptor` to `ResourcePoolDescriptor`:
https://github.com/o3de/o3de/blob/1c914f819c473ef4492360b02424e531727e846b/Gems/Atom/RHI/Code/Include/Atom/RHI.Reflect/ResourcePoolDescriptor.h#L24

So it is consistent in that it's spelled the same as the base class, rather than consistent in that a given allocator is always spelled the same way.

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponentMode.cpp
burelc-amzn: Because its base class does:
https://github.com/o3de/o3de/blob/1c914f819c473ef4492360b02424e531727e846b/Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/EditorBaseComponentMode.h#L21-L26

Gems/Atom/RHI/Code/Include/Atom/RHI.Reflect/BufferPoolDescriptor.h
burelc-amzn: Fixed.

Gems/EMotionFX/Code/EMotionFX/Source/Importer/ChunkProcessors.cpp
burelc-amzn: No, in fact the third parameter to this macro can be removed altogether (it is `0` in all cases and isn't used anymore). I've updated this PR to not add the third parameter to these new uses of the macro, we can make a separate pass to remove the third parameter from the existing usages later.

Gems/EMotionFX/Code/Source/Integration/System/AnimationModule.cpp
burelc-amzn: I went ahead and removed the third parameter from all the uses that are being added by this PR.

PR #14031, opened 2023-01-12
Fix RELEASE build issues introduced by Client/Server split

AMZN-puvvadar:
## What does this PR do?

This change addresses build issues in RELEASE builds introduced with the Multiplayer Client Server split change. The specific issue here is that `AZ_DECLARE_MODULE_CLASS` functions differently in RELEASE builds.

The issue here is for any Gem that specifies different targets for the Client, Server or Unified aliases. In this case, Multiplayer specifies `Multiplayer.Client`, `Multiplayer.Server` and `Multiplayer` respectively. In RELEASE, this requires module declarations for `Gem_Multiplayer_Client`, `Gem_Multiplayer_Server` and `Gem_Multiplayer`. This is not an issue outside of RELEASE builds.

There were two potential fixes here.

1. Modify build instructions so that the convention broadly used by split Gems would be honored by default. I did not want to go this route as I have concerns it'd be prone to error.
2. Update affected Module cases in RELEASE mode only with the necessary additional defines.

This change also corrects Multiplayer_ScriptCanvas's module name and adds a Multiplayer.Debug.Client target to untangle the UnifiedLauncher build from Multiplayer.Client.

## How was this PR tested?

I built RELEASE builds of MultiplayerSample's GameLauncher, ServerLauncher and UnifiedLauncher.

### DIFF
--- a/AutomatedTesting/Gem/Code/CMakeLists.txt
+++ b/AutomatedTesting/Gem/Code/CMakeLists.txt
@@ -41,6 +41,8 @@ ly_create_alias(NAME AutomatedTesting.Builders NAMESPACE Gem TARGETS Gem::Automa
 ly_create_alias(NAME AutomatedTesting.Tools    NAMESPACE Gem TARGETS Gem::AutomatedTesting)
 ly_create_alias(NAME AutomatedTesting.Clients  NAMESPACE Gem TARGETS Gem::AutomatedTesting)
 ly_create_alias(NAME AutomatedTesting.Servers  NAMESPACE Gem TARGETS Gem::AutomatedTesting)
+ly_create_alias(NAME AutomatedTesting.Unified  NAMESPACE Gem TARGETS Gem::AutomatedTesting)
+
 
 ################################################################################
 # Gem dependencies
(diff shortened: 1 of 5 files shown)


AMZN-Olex (approved):

AMZN-Gene (approved):

cgalvan (approved):

lumberyard-employee-dm (approved):
Should we add have a Unified Server Gem template that adds the AZ_DECLARE_MODULE_CLASS calls for both clients and servers?

AMZN-puvvadar:
> Should we add have a Unified Server Gem template that adds the AZ_DECLARE_MODULE_CLASS calls for both clients and servers?

That's probably the best way to go about handling this so I'd say yeah.

moraaar (approved):
Approved

> Update affected Module cases...

Can you quickly mention what are the conditions for a Module to be affected by this? Why does only Multiplayer gem need this change?

Thanks

AMZN-puvvadar:
> Approved
> 
> > Update affected Module cases...
> 
> Can you quickly mention what are the conditions for a Module to be affected by this? Why does only Multiplayer gem need this change?
> 
> Thanks

I've updated the description with more detail.

Issue #14029, opened 2023-01-12
[Component Mode] Entity Picker leads to a crash when used in Component Mode


mbalfour-amzn:
**Describe the bug**
While in Component Mode, if the component being edited has an Entity Picker field accessible on it, attempting to use the Entity Picker will lead to unpredictable behavior and ultimately a crash.

**Steps to reproduce**
1. Open the Editor with a new or existing level.
2. Create a new entity with an Image Gradient, a Gradient Transform Modifier, and a Box Shape.
3. On the Image Gradient, select any available existing image for the Image Asset.
4. On the Image Gradient, press the Edit button to start the paintbrush component mode.
5. On the Image Gradient, press the Entity Picker icon next to the "Pin Preview to Entity" field.
6. Press Esc to exit entity picking.

**Expected behavior**
The user is able to pick an entity, despite being in component mode.

**Actual behavior**
After step 5, the component mode partially ends, though the Entity Outliner remains greyed out and the "Edit" button still says "Done" but does nothing when pressed.

After step 6, the Editor crashes.

The callstack of the crash is as follows:
```
 	Qt5Core.dll!00007ff837f3a279()	Unknown
>	[Inline Frame] EditorLib.dll!QObject::findChildren(const QString &) Line 184	C++
 	EditorLib.dll!ShortcutDispatcher::shortcutFilter(QObject * obj, QShortcutEvent * shortcutEvent) Line 394	C++
 	[External Code]	
 	EditorLib.dll!CryEditMain(int argc, char * * argv) Line 3997	C++
 	[Inline Frame] Editor.exe!AZStd::Internal::INVOKE(int(*)(int, char * *) &) Line 208	C++
 	[Inline Frame] Editor.exe!AZStd::invoke(int(*)(int, char * *) &) Line 54	C++
 	Editor.exe!main(int argc, char * * argv) Line 32	C++
 	Editor.exe!WinMain(HINSTANCE__ * __formal, HINSTANCE__ * __formal, char * __formal, int __formal) Line 97	C++
 	[External Code]	
```

**Screenshots/Video**
https://user-images.githubusercontent.com/82224783/212152251-a2ae1463-49cc-4b5e-aee7-a72e6f37a1e8.mp4

**Found in Branch**
development

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
5779140a90977e46e33a1db51ff241652294e341

**Additional context**
I'm currently trying out potential solutions to this problem.

[...]

PR #14032, opened 2023-01-12
[Component Mode] Support entity picker in paintbrush component mode

mbalfour-amzn:
## What does this PR do?

PR to support the Entity Picker while remaining in Component Mode. This is necessary if we have any entity fields that remain active and accessible in the Entity Inspector or other windows while Component Mode is active. (See #14029 for example)

Basically, instead of switching the Viewport Mode from "Default Selection" (which contains Component Mode) to "Pick Selection" (which ends Component Mode), this creates a new helper class that lets "Pick Selection" functionality get used from a component mode without actually changing the mode.

This provides a way to make entity picking work while in component mode, but it has a couple points worth discussing:
* Each component mode needs to separately implement the functionality. This is the most flexible, because different component modes may need to do different things to "pause" themselves while entity picking is active. But it also means that each component mode can separately cause the entity picker to at least partially break if they _don't_ implement it.
* It's not really clear whether or not "Pick Selection" truly belongs as a replacement mode for "Default Selection". Of the four modes that exist (Default, Focus, Component, Pick), Pick is the only one that is a replacement instead of being integrated into Default. And since we need Picking while Default/Focus/Component are running, it might make more sense to include it as a part of Default instead of making it a replacement.

The PaintBrushManipulator ought to get refactored into a PaintBrushBaseComponentMode class instead that ImageGradient / TerrainMacroMaterial component modes inherit from, so that the manipulator code gets stripped back down to what's truly just a manipulator again, but that can be done in a separate PR.

There are a couple other small unrelated bugfixes in here that were discovered while testing this:
* TerrainMacroMaterial needed an allocator definition. New changes in the dev branch were causing this to assert.
* One of the unit tests had an incorrect name, describing the wrong expected result.

## How was this PR tested?

Tried out entity picking from both inside and outside the Image Gradient component mode.


Code/Editor/Plugins/ComponentEntityEditorPlugin/UI/Outliner/OutlinerWidget.cpp
```diff
@@ -1087,6 +1088,13 @@ void OutlinerWidget::SelectSliceRoot()
 
 void OutlinerWidget::OnEntityPickModeStarted()
 {
+    // If we're in component mode, it's possible the outliner isn't currently enabled. If so,
+    // make sure to enable it while in pick mode.
+    if (!isEnabled())
```
hultonha, line 1093: suggestion: Might it be worth adding a check (or assert) to check we're in Component Mode at this point and we're not disabled for another reason?

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
```diff
@@ -301,6 +319,20 @@ namespace AzToolsFramework
...
+                "Stop Entity Pick Mode",
+                "Exit out of the entity pick mode",
+                []()
+                {
+                    AzToolsFramework::EditorPickModeRequestBus::Broadcast(&AzToolsFramework::EditorPickModeRequests::StopEntityPickMode);
+                });
+
+            backAction.SetEntityComponentIdPair(m_ownerEntityComponentId);
+            return { backAction };
+        }
```
hultonha, line 334: Cool to see you could handle specifying the new shortcut in this way :+1: (tagging @AMZN-daimini for visibility as this will be one more thing to account for in the Action Manager updates)

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.h
```diff
@@ -97,5 +103,7 @@ namespace AzToolsFramework
         AZ::EntityComponentIdPair m_ownerEntityComponentId;
 
         AzFramework::PaintBrush m_paintBrush;
+
+        AZStd::unique_ptr<EditorPickEntitySelectionHelper> m_pickEntitySelectionMode;
```
hultonha, line 107: suggestion: You could potentially use `AZStd::optional` here instead of making this a pointer, it would work almost exactly the same it's just you don't then have the pointer indirection as the instance in a normal class member

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Outliner/EntityOutlinerWidget.cpp
```diff
@@ -918,6 +919,13 @@ namespace AzToolsFramework
 
     void EntityOutlinerWidget::OnEntityPickModeStarted()
     {
+        // If we're in component mode, it's possible the outliner isn't currently enabled. If so,
+        // make sure to enable it while in pick mode.
+        if (!isEnabled())
```
hultonha, line 924: suggestion: Same as before, might be worth adding an explicit check or assert we're definitely in Component Mode as this hasn't been disabled for some other reason

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyEntityIdCtrl.cpp
```diff
@@ -117,15 +117,21 @@ namespace AzToolsFramework
...
-            // replace the default input handler with one specific for dealing with
-            // entity selection in the viewport
-
-            EditorInteractionSystemViewportSelectionRequestBus::Event(
-                GetEntityContextId(), &EditorInteractionSystemViewportSelection::SetHandler,
-                [](const EditorVisibleEntityDataCacheInterface* entityDataCache, ViewportEditorModeTrackerInterface* viewportEditorModeTracker)
+            // Replace the default input handler with one specific for dealing with entity selection in the viewport,
+            // but only if we're not in component mode.
+            // If we're in component mode, we should continue to use the default handler and let the component mode handling
+            // manage the entity picking.
```
hultonha, line 123: Thanks for adding the helpful comment

Code/Framework/AzToolsFramework/AzToolsFramework/ViewportSelection/EditorInteractionSystemComponent.cpp
```diff
@@ -56,6 +56,12 @@ namespace AzToolsFramework
         return m_interactionRequests->InternalHandleMouseManipulatorInteraction(mouseInteraction);
     }
 
+    const EditorVisibleEntityDataCacheInterface* EditorInteractionSystemComponent::GetEntityDataCache() const
+    {
+        return m_entityDataCache.get();
+    }
+
+
```
hultonha, line 64: uber nit: Additional line break, this class just usually has one between member functions in the .cpp

Code/Framework/AzToolsFramework/AzToolsFramework/ViewportSelection/EditorInteractionSystemViewportSelectionRequestBus.h
```diff
@@ -42,6 +42,11 @@ namespace AzToolsFramework
     class EditorInteractionSystemViewportSelection : public ViewportInteraction::InternalViewportSelectionRequests
     {
     public:
+        //! \ref GetEntityDataCache enables callers to get access to the current entity data cache.
+        //! No assumptions should be made about the lifetime of this pointer, it may get destroyed
+        //! and recreated every time the Editor interaction handler is changed.
```
hultonha, line 47: Excellent API comment thanks!

Code/Framework/AzToolsFramework/AzToolsFramework/ViewportSelection/EditorPickEntitySelection.cpp
```diff
@@ -14,17 +14,18 @@
 
 namespace AzToolsFramework
 {
+    AZ_CLASS_ALLOCATOR_IMPL(EditorPickEntitySelectionHelper, AZ::SystemAllocator, 0)
     AZ_CLASS_ALLOCATOR_IMPL(EditorPickEntitySelection, AZ::SystemAllocator, 0)
 
-    EditorPickEntitySelection::EditorPickEntitySelection(
+    EditorPickEntitySelectionHelper::EditorPickEntitySelectionHelper(
```
hultonha, line 20: thought: I wonder if these changes will actually make this type easier to test too which is really nice (not something you have to do now but it'd be great in future to have some sanity checks this code is doing what we expect)

hultonha (approved):
I really like these changes for what it's worth, I know it might not be perfect but it definitely moves us in the right direction and future refactors will be easier from this point onward. Awesome work @mbalfour-amzn 🥳

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.h
mbalfour-amzn: It still _looks_ like you have a pointer indirection though... Happy to change it, it's really half dozen of one vs the other to me.

cgalvan (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
AMZN-daimini: We were chatting about this yesterday, we may need to introduce a priority system in the HotKey Manager for situations like this

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.h
hultonha: Yeah it's not a huge deal just in certain situations can be a performance win

hultonha (approved):

PR #14033, opened 2023-01-12
Prefab | Rename prefab DOM generator functions

forkercat:

## What does this PR do?

This PR renames the prefab entity/instance DOM generator functions to reduce confusion.

Before:

```cpp
InstanceToTemplatePropagator::GenerateDomForInstance();
InstanceToTemplatePropagator::GenerateDomForEntity();
InstanceDomGenerator::GenerateInstanceDom();
InstanceDomGenerator::GenerateEntityDom();
```

After:

```cpp
InstanceToTemplatePropagator::GenerateInstanceDomBySerializing();
InstanceToTemplatePropagator::GenerateEntityDomBySerializing();
InstanceDomGenerator::GetInstanceDomFromTemplate();
InstanceDomGenerator::GetEntityDomFromTemplate();
```

## How was this PR tested?

Passed tests. The PR does not alter implementation logic.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceDomGenerator.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceDomGenerator.cpp
@@ -54,15 +54,15 @@ namespace AzToolsFramework
             AZ::Interface<InstanceDomGeneratorInterface>::Unregister(this);
         }
 
-        void InstanceDomGenerator::GenerateInstanceDomFromTemplate(PrefabDom& instanceDom, const Instance& instance) const
+        void InstanceDomGenerator::GetInstanceDomFromTemplate(PrefabDom& instanceDom, const Instance& instance) const
         {
-            AZ_Assert(instanceDom.IsNull(), "GenerateInstanceDomFromTemplate must be called with an empty instance to generate into.");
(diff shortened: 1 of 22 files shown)


srikappa-amzn (approved):
'Generate' can indicate creation of something new. I recommend changing 2 of them to
GetInstanceDomFromTemplate() and GetEntityDomFromTemplate

cgalvan (approved):

michabr (commented):
I'm glad this is happening. For InstanceDomGenerator, we could avoid using "Generate" and use "Compose" instead, but then the class would still have "Generate" in the name...
ComposeInstanceDomFromTemplates
ComposeEntityDomFromTemplates

FYI the "after" description in the PR is mixed up.

michabr:
Just noticed names were changed to "GetDom". That works too.

michabr:
One more thought. Would it make sense to move GenerateInstanceDomBySerializing, GenerateEntityDomBySerializing to InstanceDomGenerator? Not necessary but it might make sense since we have a dedicated "InstanceDomGenerator" interface.

michabr (approved):

forkercat:
> One more thought. Would it make sense to move GenerateInstanceDomBySerializing, GenerateEntityDomBySerializing to InstanceDomGenerator? Not necessary but it might make sense since we have a dedicated "InstanceDomGenerator" interface.

Yeah that's a good point. I discussed with @srikappa-amzn about having that move as well. We can even change "InstanceDomGenerator" to "PrefabTemplateDomGenerator" so it better fits what functions it has.

I would leave the changes in a next PR.

PR #14038, opened 2023-01-13
Monolithic Asset Bundle Runtime fixes

### LINKED ISSUE #13473: Monolithics builds are broken
**Describe the bug**
After updating to the latest version of O3DE a few weeks ago, our release version (a monolithic build) was broken because it couldn't load any asset (including critical ones) from its .pak files.

We tracked down the issue, and it turns out it was because the engine tried to register the .pak manifests before the `BundlingSystemComponent` activated (and registered to the `AZ::IO::ArchiveNotificationBus`).

Essentially, `Archive::OnSystemEntityActivated()` is triggered before system entities are activated.

This is because of this change: https://github.com/o3de/o3de/commit/0f02a6485a9af08c962304b37e8b0b633428cb51

which added those lines:
https://github.com/o3de/o3de/blob/92278197f7567cd63c92560658252bd47b150862/Code/Framework/AzCore/AzCore/Settings/SettingsRegistryImpl.cpp#L1509-L1513

This causes an issue because of the way the engine starts, in ComponentApplication.cpp:
https://github.com/o3de/o3de/blob/92278197f7567cd63c92560658252bd47b150862/Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp#L751-L764

As you can see, settings are merged before creating the system components (which is normal), this will load the "boostrap.game.release.setreg" file which contains all .setreg settings, including those:
https://github.com/o3de/o3de/blob/92278197f7567cd63c92560658252bd47b150862/Registry/application_lifecycle_events.setreg#L9-L15

Before https://github.com/o3de/o3de/commit/0f02a6485a9af08c962304b37e8b0b633428cb51 this triggered an event telling that the root key changed, not triggering bundle loading.

However, since this commit triggers a notification for every merged key, it will trigger an event telling `/O3DE/Application/LifecycleEvents/SystemComponentsActivated` changed, which will cause https://github.com/o3de/o3de/blob/92278197f7567cd63c92560658252bd47b150862/Code/Framework/AzFramework/AzFramework/Archive/Archive.cpp#L369-L378 to trigger, calling `Archive::OnSystemEntityActivated` too early (before system components are even created).

This empties `m_archivesWithCatalogsToLoad`, even though they couldn't load, which prevent them to load correctly later.

With bundles not being loaded, critical assets can't load and we got this error at startup:
image

Commenting out those lines fixes the issue:
https://github.com/o3de/o3de/blob/92278197f7567cd63c92560658252bd47b150862/Code/Framework/AzCore/AzCore/Settings/SettingsRegistryImpl.cpp#L1509-L1513

92278197f7567cd63c92560658252bd47b150862

 - Device: PC
 - OS: Windows
 - Version 11
 - CPU: AMD R9 5950X
 - GPU: RTX 2080 Ti
 - Memory: 64GB

lumberyard-employee-dm:
Updated the Archive system handling of OnSystemActivated lifecycle event to ONLY prune the list of archive files with asset bundle catalogs to load, if the files were opened as an archive successfully.

Updated the Component Application Lifecycle event to signal, using a different section of the Settings Registry which is never loaded from a file and is transient in memory.

See the [setregbuilder](https://github.com/o3de/o3de/blob/development/Registry/setregbuilder.assetprocessor.setreg#L17-L30) settings registry file.

fixes #13473

## How was this PR tested?

Built and Launched the Windows GameLauncher Monolithic build.

### DIFF
--- a/Code/Framework/AzCore/AzCore/Component/ComponentApplicationLifecycle.cpp
+++ b/Code/Framework/AzCore/AzCore/Component/ComponentApplicationLifecycle.cpp
@@ -33,10 +33,11 @@ namespace AZ::ComponentApplicationLifecycle
                 " or in *.setreg within the project", AZ_STRING_ARG(eventName), AZ_STRING_ARG(ApplicationLifecycleEventRegistrationKey));
             return false;
         }
-        auto eventRegistrationKey = FixedValueString::format("%.*s/%.*s", AZ_STRING_ARG(ApplicationLifecycleEventRegistrationKey),
+        // The Settings Registry key used to signal the event is a transient runtime key which is separate from the registration key
+        auto eventSignalKey = FixedValueString::format("%.*s/%.*s", AZ_STRING_ARG(ApplicationLifecycleEventSignalKey),
             AZ_STRING_ARG(eventName));
 
-        return settingsRegistry.MergeSettings(eventValue, Format::JsonMergePatch, eventRegistrationKey);
+        return settingsRegistry.MergeSettings(eventValue, Format::JsonMergePatch, eventSignalKey);
     }
 
     bool RegisterEvent(AZ::SettingsRegistryInterface& settingsRegistry, AZStd::string_view eventName)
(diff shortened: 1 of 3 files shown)


cgalvan (approved):
Thanks for fixing this!

lawsonamzn (approved):

tkothadev (approved):

AMZN-stankowi (approved):
How possible would it be to write an automated test that fails without these changes, and passes with them?

lumberyard-employee-dm:
> How possible would it be to write an automated test that fails without these changes, and passes with them?

I could write a test that attempts to open an archive file with a bundle manifest within it from the `OnSystemActivated` event and then create that archive file with a bundle manifest and invoke the `OnSystemActivated` lifecycle event a second time.

lawsonamzn (approved):

PR #14040, opened 2023-01-13
Set the Dispatch args in the constructor instead of LoadShader.

msat-huawei:
Signed-off-by: Martin Sattlecker <martin.sattlecker@huawei.com>

## What does this PR do?

The Dispatch Arguments in the Compute Shader were overwritten when the Shader was reloaded. This was unexpected, as we use Indirect Dispatch, and only set the Dispatch Arguments in the `BuildInternal` function. Not every frame.

We now load the `PassData` Dispatch Arguments in the constructor. Only the data from the Shader is updated in the LoadShader function.

## How was this PR tested?

Windows & Vulkan. Tested with direct and indirect dispatch and hot reloading.

### DIFF
--- a/Gems/Atom/RPI/Code/Source/RPI.Public/Pass/ComputePass.cpp
+++ b/Gems/Atom/RPI/Code/Source/RPI.Public/Pass/ComputePass.cpp
@@ -44,6 +44,20 @@ namespace AZ
             : RenderPass(descriptor)
             , m_passDescriptor(descriptor)
         {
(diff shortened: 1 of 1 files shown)


rgba16f (approved):

invertednormal (approved):

rgba16f:
Automated Review run started: https://jenkins.build.o3de.org/job/O3DE/view/change-requests/job/PR-14040/1/

moudgils (approved):

Issue #14052, opened 2023-01-13
Unable to test Jenkins pipeline config changes in a branch


brianherrera:
**Describe the bug**
Changes made to build config files in scripts/build/ are not picked up by Jenkins when running build in a branch. 

There is a bug in the checkout step for the bootstrap scripts. The check will always fail and force the pipeline to pull the bootstrap scripts from the O3DE development branch. 

**Steps to reproduce**
1. Create a branch off O3DE development
2. Make an update to a build config file under scripts/build (e.g. scripts/build/Platform/Windows/pipeline.json)
3. Commit and push changes. 
4. Run a build in Jenkins
5. Build config files are loaded from the development branch instead.  

**Expected behavior**
Branch build correctly loads build config files from the current branch.

**Actual behavior**
Build config files are always loaded from the development branch.

PR #14054, opened 2023-01-13
Update Jenkinsfile to checkout bootstrap scripts from current branch

brianherrera:
## What does this PR do?

Fixes https://github.com/o3de/o3de/issues/14052

This change fixes an issue introduced in the Jenkisfile that prevents users from testing build config changes in a branch. The bootstrap files are always loaded from the development branch since the check will fail on new branches. 

This change moves the checkout step to the original location. The bootstrap scripts will still be pulled from the O3DE development branch if they are missing. 

## How was this PR tested?

Performed the following tests in a branch:
- Updated one of the build config files (e.g. scripts/build/Platform/Windows/pipeline.json). Jenkins successfully picked up the change from the branch. 
- Ran a build with the scripts/build folder missing. Bootstrap files successfully loaded from the development branch.

### DIFF
--- a/scripts/build/Jenkins/Jenkinsfile
+++ b/scripts/build/Jenkins/Jenkinsfile
@@ -1152,14 +1152,13 @@ try {
 
                         echo "Running repository: \"${repositoryName}\", pipeline: \"${pipelineName}\", branch: \"${branchName}\", CHANGE_ID: \"${env.CHANGE_ID}\", GIT_COMMMIT: \"${scm.GIT_COMMIT}\"..."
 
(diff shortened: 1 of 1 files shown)


cgalvan (approved):

amzn-changml (approved):

brianherrera:
Found another issue when running the AR. I'll also need to add `scm.userRemoteConfigs` as a parameter on the initial checkout, otherwise Jenkins will not find any commits to run. When ran on an external repo the follow up bootstrap checkout will still use the default engine repo config. 

Successfully tested using the replay option. I'll make the commits now.

PR #14055, opened 2023-01-13
Adding metrics scraper script

evanchia-ly-sdets:

## What does this PR do?
Adds a metrics scraping script that creates csv files to be uploaded. Most portions were taken from the existing MARS metrics scraping script. 

## How was this PR tested?
Tested by running locally.


AMZN-scspaldi (approved):

scripts/metrics/common/config.py
```diff
@@ -0,0 +1,79 @@
...
+        target = config
+
+        while len(path_elements) > 0:
+            key = path_elements.pop(0)
+
+            if isinstance(target, dict) and key in target:
+                if len(path_elements) == 0:
+                    if isinstance(target[key], (int, float, str, bool)):
+                        try:
+                            target[key] = type(target[key])(value)
```
Kadino, line 47: Why is it necessary to coerce this value to be of the same type as the previous value?

scripts/metrics/common/__init__.py
```diff
@@ -0,0 +1,8 @@
+"""
+Copyright (c) Contributors to the Open 3D Engine Project.
+For complete copyright and license terms please see the LICENSE at the root of this distribution.
+
+SPDX-License-Identifier: Apache-2.0 OR MIT
+
+Helper file for assisting in building workspaces and setting up LTT with the current Lumberyard environment.
```
Kadino, line 7: Each file contains this invalid docstring

scripts/metrics/common/logging.py
```diff
@@ -0,0 +1,65 @@
+"""
```
Kadino, line 1: Why does this metrics tool need a custom logging setup managed in its own module? Why not configure with logging.basicConfig directly in the metrics Main?

scripts/metrics/test_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,198 @@
...
+now = datetime.datetime.now()
+MONTH = now.month
+if MONTH < 10:
+    MONTH = f"0{now.month}"
+DAY = now.day
+if DAY < 10:
+    DAY = f"0{now.day}"
+DEFAULT_CSV_FILE = f"{now.year}_{MONTH}_{DAY}.csv"
+
+# Setup logging.
```
Kadino, line 31: Logging should not be configured at the top of a script like this, as this will run during import and hijack any existing logging settings of another module that (accidentally) imports this module. Instead, have this called in the main() method or the "__main__" hook.

scripts/metrics/test_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,198 @@
+"""
```
Kadino, line 1: Given that there will be multiple types/formats of files to parse, I assume separate scripts will handle other files, and suggest this is a `ctest_xml_to_csv` script.

Kadino (changes requested):

scripts/metrics/common/config.py
```diff
@@ -0,0 +1,79 @@
...
+    global config
+
+    if override is None:
+        override = {}
+
+    try:
+        with open(os.path.join(sys.path[0], local_path), "r") as file:
+            config = yaml.safe_load(file)
+    except FileNotFoundError:
+        raise ConfigExn(f"Failed to load configuration from local path '{local_path}'") from None
```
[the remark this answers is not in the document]
evanchia-ly-sdets: fixed

scripts/metrics/common/config.py
evanchia-ly-sdets: This should prevent any string/int/float conversions from the config override

scripts/metrics/common/__init__.py
evanchia-ly-sdets: Fixed

scripts/metrics/common/logging.py
evanchia-ly-sdets: Scalability/standardization if other metrics scripts will be separate from the current

scripts/metrics/test_metrics_xml_to_csv.py
evanchia-ly-sdets: I haven't made the decision yet to split the metrics script. I will rename the file according when that's decided

scripts/metrics/test_metrics_xml_to_csv.py
evanchia-ly-sdets: Fixed

scripts/metrics/common/logging.py
```diff
@@ -0,0 +1,65 @@
...
+Standardized logging util file for metrics scripts
+"""
+
+import enum
+import logging
+import sys
+
+
+class Level(enum.IntEnum):
+    CRITICAL = 50
```
lmbr-pip, line 16: Why reinvent log levels when logging has defined log levels?

scripts/metrics/common/logging.py
```diff
@@ -0,0 +1,65 @@
...
+    ERROR = 40
+    WARNING = 30
+    INFO = 20
+    DEBUG = 10
+    NOTSET = 0
+
+
+class Formatter(logging.Formatter):
```
lmbr-pip, line 24: All public classes should have comment to role/responsibility IMHO

lmbr-pip (commented):
Consider renaming `test_metrics_xml_to_csv` as it looks like a test file, consider adding unit tests

scripts/metrics/common/logging.py
evanchia-ly-sdets: This is using the same values as the defined log levels, but just a nicer enum to reference them. Was unsure if accessing the raw values would be preferable.

Kadino (commented):
(oops, forgot to hit submit on these comments last week!)

scripts/metrics/common/config.py
```diff
@@ -0,0 +1,79 @@
...
+
+config = {}
+
+
+class ConfigExn(MetricsExn):
+    pass
+
+
+def load_config(local_path, override=None):
+    global config
```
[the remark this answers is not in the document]
evanchia-ly-sdets: I prefer that the read_config() uses only the same global singleton to prevent the mishap of having multiple/different config objects. We can pass the config object around in the script if it's strongly preferred, but imo this way is cleaner.

scripts/metrics/common/config.py
```diff
@@ -0,0 +1,79 @@
...
+
+    if override is None:
+        override = {}
+
+    try:
+        with open(os.path.join(sys.path[0], local_path), "r") as file:
+            config = yaml.safe_load(file)
+    except FileNotFoundError as exc:
+        raise ConfigExn(f"Failed to load configuration from local path '{local_path}'") from exc
```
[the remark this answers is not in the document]
evanchia-ly-sdets: Fixed

scripts/metrics/common/config.py
```diff
@@ -0,0 +1,79 @@
...
+
+    for path, value in override:
+        path_elements = path.split(".")
+        target = config
+
+        while len(path_elements) > 0:
+            key = path_elements.pop(0)
+
+            if isinstance(target, dict) and key in target:
```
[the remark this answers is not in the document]
evanchia-ly-sdets: This is primarily used to for both local testing + being automation accessible. It's simple for a local user to set their own configuration while testing. The override is used for pipelines to paramterize different configuration (such as jenkins base urls).

Since this is meant to override existing keys, this should only be updating and not adding.

I'm unsure what you mean by this. I think there is some confusion that doc strings should clarify. It's kind of like multi layered keys with a value i.e.
Key1.Key2 = value
Key1.Key3 = value

scripts/metrics/common/config.py
```diff
@@ -0,0 +1,79 @@
...
+from common.exception import MetricsExn
+
+config = {}
+
+
+class ConfigExn(MetricsExn):
+    pass
+
+
+def load_config(local_path, override=None):
```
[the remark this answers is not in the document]
evanchia-ly-sdets: Added

scripts/metrics/common/config.py
```diff
@@ -0,0 +1,79 @@
...
+                            ) from exc
+                else:
+                    target = target[key]
+            else:
+                raise ConfigExn(f"Invalid override configuration parameter '{path}'")
+
+
+def read_config(path, source=None):
+    if source is None:
+        source = config
```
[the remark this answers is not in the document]
evanchia-ly-sdets: Source is the config source which can be the config file or overridden

scripts/metrics/config/config.yml
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+#
+# SPDX-License-Identifier: Apache-2.0 OR MIT
```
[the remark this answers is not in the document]
evanchia-ly-sdets: Added

scripts/metrics/test_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,204 @@
...
+        writer.writeheader()
+
+        # Parse CTest xml and write to csv file
+        parse_ctest_xml_to_csv(file_path, writer)
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description='This script assumes that a CTest xml file has been produced via the -T Test CTest option. The xml '
+                    'file will be parsed and write to a csv file.')
```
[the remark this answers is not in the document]
evanchia-ly-sdets: Yes, added clarification

scripts/metrics/test_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,204 @@
+"""
```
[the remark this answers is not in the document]
evanchia-ly-sdets: Fixed

scripts/metrics/common/config.py
```diff
@@ -0,0 +1,79 @@
...
+
+class ConfigExn(MetricsExn):
+    pass
+
+
+def load_config(local_path, override=None):
+    global config
+
+    if override is None:
+        override = {}
```
spham-amzn, line 28: Instead of doing this, why not just set ```{}``` as the default value if it is not supplied and remove this check. Assigning argument variables is not a good idea in any case

scripts/metrics/common/logging.py
```diff
@@ -0,0 +1,65 @@
...
+    def __init__(self):
+        super().__init__("%(asctime)s  %(name)-40s  %(message)s")
+
+    def format(self, record):
+        return "{} {}".format(Formatter.get_level_acronym(record.levelno), super().format(record))
+
+    @classmethod
+    def get_level_acronym(cls, level):
+        if level >= Level.CRITICAL:
```
spham-amzn, line 33: Why is ```>=``` comparisons being used here instead of ```==```? Are these level ranges?

scripts/metrics/common/config.py
evanchia-ly-sdets: It's best practice to use None as the default arg because the default is only evaluated once at function definition. Multiple uses of the function may result in incorrect behavior.

scripts/metrics/common/logging.py
evanchia-ly-sdets: I don't have a strong reason. I initially thought there might be more logging values for the levels that we didn't include in our enums, but everything seems 1 for 1 so we could probably just check equals to.

scripts/metrics/common/config.py
spham-amzn: Alternatively then, you can do 
```
override = {} if override is None else override
```

scripts/metrics/common/logging.py
spham-amzn: I would make them ```==``` checks, and default to maybe the ```f"?({level})``` or something like that to indicate its a level that is not matched

scripts/metrics/test_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,204 @@
...
+Copyright (c) Contributors to the Open 3D Engine Project.
```
spham-amzn, line 2: Is this a test tool or a test for the actual metrics scraper?

scripts/metrics/common/logging.py
evanchia-ly-sdets: Fixed

scripts/metrics/metrics_xml_to_csv.py
```diff
@@ -18,20 +18,26 @@
...
-    now = datetime.datetime.now()
+    now = now_wrap()
     month = now.month
     if month < 10:
         month = f"0{now.month}"
     day = now.day
     if day < 10:
         day = f"0{now.day}"
-    DEFAULT_CSV_FILE = f"{now.year}_{month}_{day}.csv"
+    return f"{now.year}_{month}_{day}.csv"
```
spham-amzn, line 40: Since you are updating this file, this block can be simplified to:
```
return f"{now.year}_{now.month:02d}_{now.day:02d}"
```

evanchia-ly-sdets: Oh neat, never knew that thanks

AMZN-scspaldi (approved):
Reapproving.

scripts/metrics/common/config.py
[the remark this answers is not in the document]
Kadino: I don't see why it needs to thread through this data-transformer script.
1. If I'm going to add data to a file, then I'll the data to the source of truth. Users can directly modify an XML, or directly modify a CSV after this performs transformation.
2. If I'm going to expect/query different data, then I'll access different data. Tests can directly feed in a different CSV data.

scripts/metrics/common/config.py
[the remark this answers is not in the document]
Kadino: I think it's cleaner not to provide an optional config file with overrides, and simply make a data transformer that expects a JUnit-style XML and condenses the data into CSVs with named columns.

scripts/metrics/common/config.py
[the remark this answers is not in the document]
evanchia-ly-sdets: Removed config util file

scripts/metrics/common/config.py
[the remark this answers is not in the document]
evanchia-ly-sdets: Removed config util file

scripts/metrics/common/config.py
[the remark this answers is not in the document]
Kadino: What I'm getting at here is essentially the https://en.wikipedia.org/wiki/Single-responsibility_principle and https://en.wikipedia.org/wiki/Interface_segregation_principle of SOLID

scripts/metrics/common/logging.py
evanchia-ly-sdets: I'm already working on a script which I want to use the same logging format. I can remove the enums if you want though.

scripts/metrics/metrics_xml_to_csv.py
```diff
@@ -0,0 +1,187 @@
...
+    'duration_seconds'
+]
+
+
+def main():
+    # Parse args
+    args = parse_args()
+
+    # Load global configuration.
+    logger.info("Loading configuration")
```
Kadino, line 53: Appears to no longer be a configuration to load

scripts/metrics/metrics_xml_to_csv.py
```diff
@@ -0,0 +1,187 @@
...
+
+    # Construct the full path to the xml file
+    file_path = _get_test_xml_path(args.build_folder, args.ctest_log)
+
+    # Create csv files
```
Kadino, line 58: minor typo: appears to generate only one file

scripts/metrics/metrics_xml_to_csv.py
```diff
@@ -0,0 +1,187 @@
...
+    with open(args.csv_file, 'w', encoding='UTF8', newline='') as csv_file:
```
Kadino, line 59: `'w'` will overwrite a file that already exists. Should this instead emit an error if a CSV file already existed at that location? May be appropriate to warn the user that the file got overwritten.

scripts/metrics/metrics_xml_to_csv.py
```diff
@@ -0,0 +1,187 @@
+"""
+Copyright (c) Contributors to the Open 3D Engine Project.
+For complete copyright and license terms please see the LICENSE at the root of this distribution.
+
+SPDX-License-Identifier: Apache-2.0 OR MIT
+
+Scrapes metrics from xml files and creates csv formatted files.
```
Kadino, line 7: Notably scrapes from a CTest XML, and a different scraper will need to be used for extracting metrics from other XML files. Should probably also change the file name to ctest_metrics_xml[...]

scripts/metrics/ctest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,186 @@
...
+
+from common import logging
+
+DEFAULT_CTEST_LOG_FILENAME = 'Test.xml'
+TAG_FILE = 'TAG'
+TESTING_DIR = 'Testing'
+
+def now_wrap():
+    """
+    Wrapper function for now() for unit testing purposes.
```
Kadino, line 24: Why was this necessary? Can you not directly mock datetime.datetime.now in the tests?

scripts/metrics/ctest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,186 @@
...
+        # Parse CTest xml and write to csv file
+        parse_ctest_xml_to_csv(file_path, writer)
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description='This script assumes that a CTest xml file has been produced via the -T Test CTest option. The file'
+                    'should exist inside of the build directory. The xml file will be parsed and write to a csv file.')
+    parser.add_argument(
+        'build_folder',
```
Kadino, line 71: Argument appears [required](https://docs.python.org/3/library/argparse.html#required) for the script to function

Kadino (commented):
Looks good outside of these two last snags:

scripts/metrics/ctest_metrics_xml_to_csv.py
evanchia-ly-sdets: Correct, it is not mockable

scripts/metrics/ctest_metrics_xml_to_csv.py
evanchia-ly-sdets: A positional arg is already required by default

Kadino (approved):

scripts/metrics/ctest_metrics_xml_to_csv.py
Kadino: https://docs.python.org/3/library/unittest.mock-examples.html#partial-mocking

evanchia-ly-sdets: Fixed

PR #14058, opened 2023-01-13
Errors in python packages will now be a build failure

lawsonamzn:
## What does this PR do?

This adds checks to the get_python and lypython.cmake files which will cause cmake or the command prompt to fail immediately and return a non zero exit code.

Continuing while the python package configuration is bad causes hard-to-attribute failures.

## How was this PR tested?

Windows and linux testing.

Tested the following cases on each
* No python installed
* Python installed but no packages
* Python and packages installed

Tested the above 3 against the matrix of
* Via CMake
* Via get_python
* Via CMake + an invaild package -> error
* Via get_python + an invalid package -> error

### DIFF
--- a/cmake/LYPython.cmake
+++ b/cmake/LYPython.cmake
@@ -92,8 +92,8 @@ function(update_pip_requirements requirements_file_path unique_name)
     message(VERBOSE "pip output: ${PIP_OUT}")
 
     if (NOT ${PIP_RESULT} EQUAL 0)
-        message(CHECK_FAIL "Failed to fetch / update python dependencies: ${PIP_OUT} - use CMAKE_MESSAGE_LOG_LEVEL to VERBOSE for more information")
-        message(FATAL_ERROR "some python dependencies failed to install/download.  This will likely cause errors further down the line, stopping.")
+        message(CHECK_FAIL "Failed to fetch / update python dependencies from ${requirements_file_path}\nPip install log:\n${PIP_OUT}")
+        message(FATAL_ERROR "The above failure will cause errors later - stopping now.  Check the output log (above) for details.")
     else()
         string(FIND "${PIP_OUT}" "Installing collected packages" NEW_PACKAGES_INSTALLED)
 
--- a/cmake/LYPython.cmake
+++ b/cmake/LYPython.cmake
@@ -93,6 +93,7 @@ function(update_pip_requirements requirements_file_path unique_name)
 
     if (NOT ${PIP_RESULT} EQUAL 0)
         message(CHECK_FAIL "Failed to fetch / update python dependencies: ${PIP_OUT} - use CMAKE_MESSAGE_LOG_LEVEL to VERBOSE for more information")
+        message(FATAL_ERROR "some python dependencies failed to install/download.  This will likely cause errors further down the line, stopping.")
     else()
         string(FIND "${PIP_OUT}" "Installing collected packages" NEW_PACKAGES_INSTALLED)
 
@@ -193,6 +194,7 @@ function(ly_pip_install_local_package_editable package_folder_path pip_package_n
 
     if (NOT ${PIP_RESULT} EQUAL 0)
         message(CHECK_FAIL "Failed to install ${package_folder_path}: ${PIP_OUT} - use CMAKE_MESSAGE_LOG_LEVEL to VERBOSE for more information")
+        message(FATAL_ERROR "Failure to install a python package will likely cause errors further down the line, stopping!")
     else()
         file(TOUCH ${stamp_file})
     endif()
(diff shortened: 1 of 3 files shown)


lawsonamzn:
Note that I'm not usually happy with the idea of fatal-erroring out early, but in this case, if you specify invalid packages in your requirements.txt, it will now error and stop.

Previously, it would warn and continue, and this lead to really weird compile or build or run probblems further down the line, as various packages were still the old insecure version, missing, or wrong API.

cmake/LYPython.cmake
```diff
@@ -93,6 +93,7 @@ function(update_pip_requirements requirements_file_path unique_name)
 
     if (NOT ${PIP_RESULT} EQUAL 0)
         message(CHECK_FAIL "Failed to fetch / update python dependencies: ${PIP_OUT} - use CMAKE_MESSAGE_LOG_LEVEL to VERBOSE for more information")
+        message(FATAL_ERROR "some python dependencies failed to install/download.  This will likely cause errors further down the line, stopping.")
```
lsemp3d, line 96: is it possible to know which dependencies failed? is there a log where someone could go and inspect? can we point them there in this error message?

lsemp3d (approved):

cmake/LYPython.cmake
lawsonamzn: Yes, the line above it shows the full log from running PIP which reveal's the error. ( ${PIP_OUT} )

amzn-changml (approved):
LGTM. Thanks for working on this!

cmake/LYPython.cmake
lawsonamzn: I'll update the message to try to make it clearer.

cgalvan (approved):

lawsonamzn:
I updated the messages.  

Adding bog-us-package (version required == 1.2.3) yields the following output

Cmake version (appears in cmake gui)
```
Python: Getting/Checking packages listed in D:/o3de_repos/o3de/python/requirements.txt
Python: Getting/Checking packages listed in D:/o3de_repos/o3de/python/requirements.txt - Failed to fetch / update python dependencies from D:/o3de_repos/o3de/python/requirements.txt
Pip install log:
Requirement already satisfied: atomicwrites==1.4.1 in d:\o3de_repos\o3de\python\runtime\python-3.10.5-rev1-windows\python\lib\site-packages (from -r D:/o3de_repos/o3de/python/requirements.txt (line 1)) (1.4.1)
...
... (lots of 'already installed' lines appear here)
...
ERROR: Could not find a version that satisfies the requirement bog-us-package==1.2.3 (from versions: none)
ERROR: No matching distribution found for bog-us-package==1.2.3

CMake Error at cmake/LYPython.cmake:96 (message):
  The above failure will cause errors later - stopping now.  Check the output
  log (above) for details.
Call Stack (most recent call first):
  cmake/LYPython.cmake:269 (update_pip_requirements)
  CMakeLists.txt:44 (include)
Configuring incomplete, errors occurred!
See also "D:/o3de_repos/o3de/build/windows_vs2019/CMakeFiles/CMakeOutput.log".
See also "D:/o3de_repos/o3de/build/windows_vs2019/CMakeFiles/CMakeError.log".
```

command line version (get_python.bat)
```
calling PIP to install requirements..
...
... (lots of 'already installed' lines appear here)
...
ERROR: Could not find a version that satisfies the requirement bog-us-package==1.2.3 (from versions: none)
ERROR: No matching distribution found for bog-us-package==1.2.3
Failed to install the packages listed in D:\o3de_repos\o3de\python\requirements.txt.  Check the log above
```

lumberyard-employee-dm (approved):

python/get_python.bat
```diff
@@ -59,8 +57,19 @@ if ERRORLEVEL 1 (
     EXIT /b 1
 )
 
+:install_packages
 echo calling PIP to install requirements...
 call "%CMD_DIR%\pip.cmd" install -r "%CMD_DIR%/requirements.txt" --disable-pip-version-check --no-warn-script-location
+if ERRORLEVEL 1 (
```
Kadino, line 63: Pip could return [other error codes](https://github.com/pypa/pip/blob/main/src/pip/_internal/cli/status_codes.py) so this may want to use:
`if ERRORLEVEL NEQ 0`

Kadino (approved):
A definite improvement!

python/get_python.bat
lawsonamzn:
```
if ERRORLEVEL 1
```
actually means 1 or greater.  Its true for 1, 2, 3,... MAX_INT

lawsonamzn:
failed due to unrelated AR Issue.

PR #14070, opened 2023-01-16
Action Manager | Reflect Editor Image Gradient Component Mode to work correctly with the new setup

AMZN-daimini:

## What does this PR do?
Reflects the Editor Image Gradient Component Mode to Serialize Context:
This allows it to be picked up by the new setup that allows Component Modes to register actions and switch them correctly in the file menu of the Editor using the new Action Manager.

## How was this PR tested?
Manual testing.


hultonha (approved):
LGTM! Though would likely be good to have @mbalfour-amzn have a quick look over the changes too. Nice work! 👍

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponentMode.cpp
```diff
@@ -141,6 +147,44 @@ namespace GradientSignal
...
+        actionManagerInterface->AssignModeToAction(modeIdentifier, "o3de.action.paintBrushManipulator.decreaseHardness");
+    }
+
+    void EditorImageGradientComponentMode::BindActionsToMenus()
+    {
+        auto menuManagerInterface = AZ::Interface<AzToolsFramework::MenuManagerInterface>::Get();
+        AZ_Assert(menuManagerInterface, "EditorImageGradientComponentMode - could not get MenuManagerInterface on BindActionsToMenus.");
+
+        menuManagerInterface->AddActionToMenu(EditMenuIdentifier, "o3de.action.paintBrushManipulator.increaseSize", 6000);
+        menuManagerInterface->AddActionToMenu(EditMenuIdentifier, "o3de.action.paintBrushManipulator.decreaseSize", 6001);
```
mbalfour-amzn, line 183: The paintbrush manipulation stuff is used for more than just the Image Gradient. It's also used for the Terrain Macro Material, and _could_ be used by more things in other gems in the future. Seems like the registration for it should be handled somewhere in AzToolsFramework, not tied to the Image Gradient.

mbalfour-amzn (changes requested):

cgalvan (approved):
Approved pending the feedback discussion w/ mbalfour

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
```diff
@@ -343,13 +347,106 @@ namespace AzToolsFramework
...
                 .SetCallback(
-                    [this]()
+                    []()
                     {
                         AdjustHardnessPercent(-ed_paintBrushHardnessPercentAdjustAmount);
                     }),
         };
     }
 
+    void PaintBrushManipulator::RegisterActions()
```
mbalfour-amzn, line 357: Please add some comments explaining why this exists and how it's different from PopulateActionsImpl(). It's very confusing from looking at the code why we have two different but similar methods that are dealing with Actions.

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
```diff
@@ -343,13 +347,106 @@ namespace AzToolsFramework
...
+                actionIdentifier,
+                actionProperties,
+                []
+                {
+                    AdjustSize(ed_paintBrushSizeAdjustAmount);
+                }
+            );
+
+            hotKeyManagerInterface->SetActionHotKey(actionIdentifier, "]");
+        }
```
mbalfour-amzn, line 384: Would it be possible to create a helper function that takes in the array of ActionOverride objects and turns them into whatever this needs? I'm concerned about all the duplicated boilerplate that this is introducing. It feels like it will be extremely easy to get the two sets of actions out of sync with each other.

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponentMode.cpp
```diff
@@ -28,6 +32,8 @@
 
 namespace GradientSignal
 {
+    static constexpr AZStd::string_view EditMenuIdentifier = "o3de.menu.editor.edit";
```
mbalfour-amzn, line 35: Does this same string get hard-coded everywhere that something interacts with the edit menu?

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponentMode.cpp
mbalfour-amzn: The main issue that I've got here is that paintbrush consumers shouldn't need to know that the paintbrush has hotkeys for controlling size. If more hotkeys get added in the future for controlling opacity, etc, it should be in one place (the paintbrush), not in every consumer.

The way I was dealing with this from a component mode point of view is that I created methods on the paint brush manipulator to do all the registration, and just had pass-through calls that each component mode needed to make.

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponentMode.cpp
```diff
@@ -141,6 +147,44 @@ namespace GradientSignal
...
+    void EditorImageGradientComponentMode::BindActionsToMenus()
+    {
+        auto menuManagerInterface = AZ::Interface<AzToolsFramework::MenuManagerInterface>::Get();
+        AZ_Assert(menuManagerInterface, "EditorImageGradientComponentMode - could not get MenuManagerInterface on BindActionsToMenus.");
+
+        menuManagerInterface->AddActionToMenu(EditMenuIdentifier, "o3de.action.paintBrushManipulator.increaseSize", 6000);
+        menuManagerInterface->AddActionToMenu(EditMenuIdentifier, "o3de.action.paintBrushManipulator.decreaseSize", 6001);
+        menuManagerInterface->AddActionToMenu(EditMenuIdentifier, "o3de.action.paintBrushManipulator.increaseHardness", 6002);
+        menuManagerInterface->AddActionToMenu(EditMenuIdentifier, "o3de.action.paintBrushManipulator.decreaseHardness", 6003);
+    }
```
mbalfour-amzn, line 186: What is the significance of the 6000-6003 constants?

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponentMode.h
```diff
@@ -25,9 +25,17 @@ namespace GradientSignal
         , private ImageGradientModificationNotificationBus::Handler
     {
     public:
+        AZ_RTTI(EditorImageGradientComponentMode, "{49957D52-F1C3-4C34-AA84-7661BC418AB2}", EditorBaseComponentMode)
+
         EditorImageGradientComponentMode(const AZ::EntityComponentIdPair& entityComponentIdPair, AZ::Uuid componentType);
         ~EditorImageGradientComponentMode() override;
 
+        static void Reflect(AZ::ReflectContext* context);
```
mbalfour-amzn, line 33: Just curious, why is the reflection needed?

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
AMZN-daimini: The legacy ones will be removed as soon as we enable the new system by default and start red-coding it. So the intention is to make this transition as quick as possible.

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponentMode.cpp
AMZN-daimini: I have a follow-up PR ready to consolidate all these identifiers in headers, I had to wait for all migrations to be done before submitting it cause I was doing them in parallel :)

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponentMode.cpp
AMZN-daimini: They're just sorting keys. I am preparing the documentation and that will go over it.

They could hypothetically all be the same number, but having them separate allows external Gems to add things in-between if necessary.

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
mbalfour-amzn: Ohhh. I didn't realize one was replacing the other, I thought these existed in parallel for some reason. Thanks for the explanation!

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponentMode.h
AMZN-daimini: The new Component Mode action handler gets a list of all component modes to iterate over from the reflection system, so we don't have to add registrations for them in a single place.

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
AMZN-daimini: To answer the actual question... not at runtime. The legacy functions are members and require a component mode to be instantiated to be run, while the new ones are static and triggered at startup.

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponentMode.cpp
AMZN-daimini: That is a good point. I can make a helper so that every component mode can just pass their identifier up to the PaintBrushManipulator and the helper takes care of adding the actions to their mode for them.

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponentMode.cpp
mbalfour-amzn: Wouldn't they need to increment by more than 1 to fit something in-between? 

Please at least make it/them named constants. For example, `int baseSortKey = 6000; // arbitrary starting sort number` and then use `baseSortKey += 10; // leave space between entries in case something wants to add menu items in-between` or something just so that these are a little less magical.

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/PaintBrushManipulator.cpp
```diff
@@ -343,13 +349,131 @@ namespace AzToolsFramework
...
+        actionManagerInterface->AssignModeToAction(modeIdentifier, "o3de.action.paintBrushManipulator.increaseHardness");
+        actionManagerInterface->AssignModeToAction(modeIdentifier, "o3de.action.paintBrushManipulator.decreaseHardness");
+    }
+
+    void PaintBrushManipulator::BindActionsToMenus()
+    {
+        auto menuManagerInterface = AZ::Interface<AzToolsFramework::MenuManagerInterface>::Get();
+        AZ_Assert(menuManagerInterface, "PaintBrushManipulator - could not get MenuManagerInterface on BindActionsToMenus.");
+
+        int baseSortKey = 6000; // arbitrary starting sort number
```
mbalfour-amzn, line 468: Nit: Since you're not modifying the value, you could make it const or constexpr.

Code/Framework/AzToolsFramework/AzToolsFramework/Viewport/ViewBookmarkSystemComponent.h
```diff
@@ -9,13 +9,16 @@
...
 #include <Viewport/LocalViewBookmarkLoader.h>
 
 namespace AzToolsFramework
 {
     //! @class ViewBookmarkSystemComponent
     //! @brief System Component that holds functionality for the ViewBookmarks
-    class ViewBookmarkSystemComponent final : public AZ::Component
+    class ViewBookmarkSystemComponent final
+        : public AZ::Component
+        , public AzToolsFramework::ActionManagerRegistrationNotificationBus::Handler
```
mbalfour-amzn, line 21: This seems like it belongs either on GlobalPaintBrushSettingsSystemComponent (since that's at least a part of the paintbrush system), or on ToolsApplication, since it's reflecting a lot of the paintbrush stuff directly. There's no connection that I'm aware of between ViewBookmark and PaintBrush, so this feels pretty out-of-place the way it is.

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponentMode.cpp
```diff
@@ -143,6 +144,33 @@ namespace GradientSignal
...
     }
 
+    void EditorImageGradientComponentMode::Reflect(AZ::ReflectContext* context)
+    {
+        AzToolsFramework::ComponentModeFramework::ReflectEditorBaseComponentModeDescendant<EditorImageGradientComponentMode>(context);
+    }
+
+    void EditorImageGradientComponentMode::RegisterActions()
+    {
+    }
```
mbalfour-amzn, line 154: Suggestion: Maybe add a comment to this and BindActionsToMenus() to explain why they're empty but still exist.

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponentMode.cpp
```diff
@@ -143,6 +144,33 @@ namespace GradientSignal
...
+    {
+        auto actionManagerInterface = AZ::Interface<AzToolsFramework::ActionManagerInterface>::Get();
+        AZ_Assert(actionManagerInterface, "EditorImageGradientComponentMode - could not get ActionManagerInterface on BindActionsToModes.");
+
+        AZ::SerializeContext* serializeContext = nullptr;
+        AZ::ComponentApplicationBus::BroadcastResult(serializeContext, &AZ::ComponentApplicationRequests::GetSerializeContext);
+
+        AZStd::string modeIdentifier = AZStd::string::format(
+            "o3de.context.mode.%s", serializeContext->FindClassData(azrtti_typeid<EditorImageGradientComponentMode>())->m_name);
+
```
mbalfour-amzn, line 166: Suggestion: Consider moving all this boilerplate down into PaintBrushManipulator::BindActionsToMode() and just pass down `azrtti_typeid<EditorImageGradientComponentMode>()` as the method input.

mbalfour-amzn (changes requested):

Gems/GradientSignal/Code/Source/Editor/EditorImageGradientComponentMode.cpp
AMZN-daimini: Done!

mbalfour-amzn (approved):

Issue #13994, opened 2023-01-11
Cylinder colliders do not work with PhysX 5.1 on prefabs created with PhysX


michalpelka:
**Describe the bug**
Thanks for the introduction PhysX5.1. I started testing against our project, but I suffer from some issues with colliders.
It seems that PhysX shape collider cannot create / cook mesh for the cylinder when we compile with '-DAZ_USE_PHYSX5=ON'.

**Assets required**
No asset required.

**Steps to reproduce**
1. Compile O3DE project without `-DAZ_USE_PHYSX5=ON`
2. Create a new level
3. Create entity, add PhysX collider, set Shape to the cylinder
4. Add rigid body with default parameters to a created entity
7. Compile O3DE project with `-DAZ_USE_PHYSX5=ON`
8. Open Level 
9. Start simulation

**Expected behavior**
Collider should collide.

**Actual behavior**
The cylinder collider does not collide. PhysX Debug Gem shows no cooked mesh.

**Screenshots/Video**
I've attached:
 - PVD Log for PhysX4 and 5 : [PrefabInPhysX.zip](https://github.com/o3de/o3de/files/10391854/PrefabInPhysX.zip)
 - Here is prefab : [Prefab](https://gist.github.com/michalpelka/171fde426b562f60f5ef7961a1895c82)
**Found in Branch**
development

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
8952835571e4875c20590d88ef48f9067fc525e8

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Ubuntu
 - Version 22.04
 - CPU Intel
 - GPU Nvidia
 - Memory 64 GB

**Additional context**
Note that re-adding a PhysX Shape collider fixes the issue.
This issue can suffer everywhere mesh needs to be recooked.

PR #14071, opened 2023-01-16
Fix for cylinder collider to support switch between PhysX 4 and 5

SergeyAMZN:

## What does this PR do?

Cook physics geometry when building game entity to make it re-cook by the Asset Processor
#13994 13994

Removed old unused version converters of Editor physics components since they don't work for prefabs and there are no cases when editor components are in the ObjectStream

## How was this PR tested?

Manual testing


Gems/PhysX/Code/Source/EditorRigidBodyComponent.cpp
```diff
@@ -132,7 +132,7 @@ namespace PhysX
         if (serializeContext)
         {
             serializeContext->Class<EditorRigidBodyConfiguration, AzPhysics::RigidBodyConfiguration>()
-                ->Version(2, &ClassConverters::EditorRigidBodyConfigVersionConverter)
+                ->Version(2)
```
hultonha, line 135: question: In the other case where we removed the version converted the number was dropped by 1 but is the same here. Is it worth being consistent and either keeping it the same or dropping by one (or incrementing?)

hultonha (approved):
Change looks good I just had one quick question. Thanks for fixing this (tagging @michalpelka for visibility)

moraaar (approved):

Gems/PhysX/Code/Source/EditorRigidBodyComponent.cpp
SergeyAMZN: Changed the version to 1, thanks!

cgalvan (approved):

PR #14075, opened 2023-01-17
Removed temporary workaround on Linux that limited the PhysX simulation to one thread

moraaar:

## What does this PR do?

Previously using a PhysXCpuDispatcher that allowed more threads resulted in internal PhysX mutex asserting and the app crashing.

This assert happened back in Lumberyard days when Linux platform was still in very early stages and only supported the dedicated server. Now that Linux platform is much more mature this thread limitation can be removed.

## How was this PR tested?

Using a Linux machine I run a level with physics components using the Editor, GameLauncher and ServerLauncher with AutomatedTesting project. Testing with both debug and profile and PhysX 4 and PhysX 5. They all run successfully without PhysX asserts.


greerdv (approved):
Nice!

PR #14079, opened 2023-01-17
Multi gpu rhi multi device resources

jhmueller-huawei:
Please see https://github.com/o3de/sig-graphics-audio/discussions/32#discussioncomment-4706904


rgba16f:
Last commit has a DCO issue, should be fixed with `git commit --amend --no-edit --signoff` and a force push

rgba16f:
😭image

Gems/Atom/Feature/Common/Code/Source/MorphTargets/MorphTargetDispatchItem.h
```diff
@@ -58,7 +58,7 @@ namespace AZ
 
             bool Init();
 
-            const RHI::DispatchItem& GetRHIDispatchItem() const;
+            const RHI::DeviceDispatchItem& GetRHIDispatchItem() const;
```
rgba16f: should this be GetDeviceDispatchItem?

moudgils (commented):
Overall great work. Lots of stuff happening in this PR. I will quickly point out high level points but we should dig into them in detail in a meeting. 

- Overall the design looks good to me. Based on the original design I had posted it seems like we went with the option where we are now expanding RHI to hold resources for N devices. Using DX12 as an example it seems like we split the association between RHI::XXX and DX::XXX and now RHI::XXX can own N DX12::XXX where N is the number of devices. Even though it does make everything more complex it works and looks good to me. In the end a feature like this will eventually bring complexity (be it in RHI or RPI). I have a some suggestions below based on a quick pass of the PR. It may also be a good idea to do a RFC that can contain UML diagrams with the proposed changes to help other engineers better understand the changes. 

Suggestions for the PR
- I see lots of leftover unused code for XXX objects compared to DeviceXXX objects (for example BufferPool vs . Probably resulting from copy paste of the XXX object in question. Lets do a deeper pass and remove anything that is not being used.

- I see a lot of objects where creating a DeviceXXX version didnt make sense to me. We could probably get away with expanding the original (i.e XXX) class. An example could be DispatchItem vs DeviceDispatchItem.

- It may make sense to add MultiDevice before all the objects that inherit from RHI::MultiDeviceObject. As an example rename RHI::Resource to RHI::MultiDeviceResource or RH::ResourceView to RHI::MultiDeviceResourceView. It will make understanding the relationships a lot better. 

- I recommend moving this PR to a branch within o3de repo so other people can access it and provide assistance with future steps. 

- Virtual device support - Something easy to add would be virtual device support. Maybe something for future. But it should be easy to create multiple instances of the same device and use that for stuff like debugging multiple RHIs (dx12 and vk) using split screen mode. 

- Testing - Given the nature of this PR will require every type of testing that we can imagine. Test each backend, test each platform, test various type of content (ASV, Loft, etc). Unit tests on all platforms. Test, test, test. 

- Metal backend support - It sounds like we just made basic changes for MEtal but really this code may not even compile for Mac or ios. This again will require testing. 

- Unit tests - We should look into adding unit tests for this work. 

- Performance tests - We will need to do cpu perf testing to make sure we are not introducing any regression if running on one device. 

Looking forward - Based on these RHI changes I was just thinking out aloud. Even though we only have one framegraph my design mentioned here https://github.com/o3de/sig-graphics-audio/discussions/32 would still work. We could have multiple RPI pipelines that could run on different devices (device index is data driven from pass files). These pipelines will automatically get flattened into one RHI framegraph and since each scope knows which device it is for the draw packets will get encoded with the correct device. What is missing (unless I missed it) is supporting and using different commandlists for different devices. So in order for this to work you will need to create different commandlists for different devices and then scopes will get dispatched to different devices. Anyways, we can discuss this further in the meeting. Excited to see what this becomes in the end :)

VickyAtAZ (commented):
I think we should provide functions in CommandList which allow user still using the old API to submit draw item instead of device draw item. This is applying to all the CommandList functions, they should keep all the old APIs. 

With this change, The change required in RPI and FeatureCommon gem can be down to minimum or none. 

Also, we would like to see a list of classes which are not derived from RHI::DeviceObject but have been added 'Device' prefix in this PR.

rgba16f:
Should we close out this PR now? I believe the feature is being worked on in another PR?

jhmueller-huawei:
Yes, sure!

PR #14084, opened 2023-01-17
Multiplayer Editor Test: Wait 100 Seconds for Server to Start

### LINKED ISSUE #11635: Multiplayer Editor 5 Second Timeout Too Short on Slower Machines
**Describe the bug**
The 5 seconds time out when connecting the Editor to Server wasn't enough time to allow the server to connect.

**Steps to reproduce**
On a slow computer...
1. Open Editor
2. Open a Multiplayer level
3. Press CTRL+G
4. Notice the time it takes to connect 
     _Attempting to connect to server in order to send level data. Attempt 4 of 5_

**Expected behavior**
The server connects  

**Actual behavior**
editorsv_max_connection_attempts is too short and times out

**Found in Branch**
Development 8/29/2022

**Additional context**
It's possible the server just crashes or there's a bug that causes connection to not occur, so there _should_ be a time out. However, we probably want to increase the time between connect attempts so that relatively fast machines connect on the 1st or 2nd attempt at most. And increase the editorsv_max_connection_attempts for slower machines.

We could also leave the connections attempts infinite since users can exit anytime; seeing the attempts is enough to let them know the editor did not hang. Keep in mind if the server crashes we should already be detected that, so this would only occur should an actual bug occur and the server fail to connect even after it's loaded without crashing

AMZN-Gene:
Increasing multiplayer editor connections attempts to 15 tries. That's 106 seconds total. 1+1+2+3+4...+14

Note that this change doesn't cause the test to actually take 10x longer. It will still proceed as soon as it receives a connection. It should just give it a bit more of a margin for when the machine is busy.

## What does this PR do?
Fixes #11635

### DIFF
--- a/Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
+++ b/Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
@@ -59,7 +59,7 @@ namespace Multiplayer
     AZ_CVAR(AZ::CVarFixedString, editorsv_serveraddr, AZ::CVarFixedString(LocalHost), nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "The address of the server to connect to");
     AZ_CVAR(AZ::CVarFixedString, editorsv_rhi_override, "", nullptr, AZ::ConsoleFunctorFlags::DontReplicate,
         "Override the default rendering hardware interface (rhi) when launching the Editor server. For example, you may be running an Editor using 'dx12', but want to launch a headless server using 'null'. If empty the server will launch using the same rhi as the Editor.");
-    AZ_CVAR(uint16_t, editorsv_max_connection_attempts, 15, nullptr, AZ::ConsoleFunctorFlags::DontReplicate,
+    AZ_CVAR(uint16_t, editorsv_max_connection_attempts, 5, nullptr, AZ::ConsoleFunctorFlags::DontReplicate,
         "The maximum times the editor will attempt to connect to the server. Time between attempts is increased based on the number of failed attempts.");
 
     AZ_CVAR(bool, editorsv_print_server_logs, true, nullptr, AZ::ConsoleFunctorFlags::DontReplicate,
--- a/Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
+++ b/Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
@@ -59,7 +59,7 @@ namespace Multiplayer
     AZ_CVAR(AZ::CVarFixedString, editorsv_serveraddr, AZ::CVarFixedString(LocalHost), nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "The address of the server to connect to");
     AZ_CVAR(AZ::CVarFixedString, editorsv_rhi_override, "", nullptr, AZ::ConsoleFunctorFlags::DontReplicate,
         "Override the default rendering hardware interface (rhi) when launching the Editor server. For example, you may be running an Editor using 'dx12', but want to launch a headless server using 'null'. If empty the server will launch using the same rhi as the Editor.");
-    AZ_CVAR(uint16_t, editorsv_max_connection_attempts, 5, nullptr, AZ::ConsoleFunctorFlags::DontReplicate,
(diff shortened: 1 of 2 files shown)


AMZN-puvvadar (approved):
Approving as a means to unblock the AR pipeline but I'm not a big fan of a 106s timeout. Is there a PR planned to make this callback based verse timer?

AMZN-Gene:
> Approving as a means to unblock the AR pipeline but I'm not a big fan of a 106s timeout. Is there a PR planned to make this callback based verse timer?
Yes! See https://github.com/o3de/o3de/issues/14086

lawsonamzn (approved):
I'm okay with this.   Looking at some AR runs, when it was succeeding it was making it with only a second or two of margin in some cases.   That's not a lot of margin for something that could take a little bit longer just because some cron job is happening in the background at random or the network is a little congested on the virtual hdd due to data center or random program updates or something.

Note that this shouldn't cause the test to actually take 10x longer.   It will still proceed as soon as it receives a connection.  It should just give it a bit more of a margin for when the machine is busy.

burelc-amzn (approved):

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
```diff
@@ -59,7 +59,7 @@ namespace Multiplayer
     AZ_CVAR(AZ::CVarFixedString, editorsv_serveraddr, AZ::CVarFixedString(LocalHost), nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "The address of the server to connect to");
     AZ_CVAR(AZ::CVarFixedString, editorsv_rhi_override, "", nullptr, AZ::ConsoleFunctorFlags::DontReplicate,
         "Override the default rendering hardware interface (rhi) when launching the Editor server. For example, you may be running an Editor using 'dx12', but want to launch a headless server using 'null'. If empty the server will launch using the same rhi as the Editor.");
-    AZ_CVAR(uint16_t, editorsv_max_connection_attempts, 5, nullptr, AZ::ConsoleFunctorFlags::DontReplicate,
+    AZ_CVAR(uint16_t, editorsv_max_connection_attempts, 15, nullptr, AZ::ConsoleFunctorFlags::DontReplicate,
```
lumberyard-employee-dm, line 62: How about we set this CVar value through the command line that launches the test?

lumberyard-employee-dm: @jromnoa Would you happen to know how to inject a command line argument to an Editor Python test.
Looking at the git blame, it looks like you added support for an [extra_cmdline_args](https://github.com/o3de/o3de/blame/development/Tools/LyTestTools/ly_test_tools/o3de/multi_test_framework.py#L826-L828) option?

Specifically, the [test_Multiplayer_BasicConnectivity_Connects](https://github.com/o3de/o3de/blob/development/AutomatedTesting/Gem/PythonTests/Multiplayer/TestSuite_Main.py#L38-L39) test would need the command line argument injected.

AMZN-Gene: Yeah, adding it in pytest, or even an AutomatedTesting setreg? I like your suggesting for setting it right inside of pytest, I'm sure I've done that before... searching for the the method now.

AMZN-Gene: How's this? It a little different than setting the cvar via commandline, and instead just setting it before the test runs. Thoughts?

lawsonamzn (approved):

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
lumberyard-employee-dm: That is good with me.

AMZN-alexpete (approved):

AMZN-Gene:
> Note that this shouldn't cause the test to actually take 10x longer. It will still proceed as soon as it receives a connection. It should just give it a bit more of a margin for when the machine is busy.

Good to call out. Yes, this doesn't force the test to wait 10x longer, it only gives much more margin for waiting, but the test will still proceed as soon as the connection occurs. I've added that info to the PR description.

PR #14088, opened 2023-01-17
Removed toggle for default procedural prefabs

AMZN-stankowi:
## What does this PR do?

The toggle for default procedural prefabs for scene settings wasn't fully functional, and would cause unexpected behavior.

Also, it functionally duplicated the existence of the prefab group at all. So instead of having two ways to disable default prefabs on a scene file (toggle or remove the group), it seemed cleaner to just remove the toggle, and let the existence of the group control default prefab generation.

**Before this change:**
image

**After this change:**
image

## How was this PR tested?

Manual testing + verified existing automated tests (that used the group removal and not toggle) still pass


cgalvan (approved):

Gems/Prefab/PrefabBuilder/PrefabGroup/PrefabGroup.cpp
```diff
@@ -126,27 +116,31 @@ namespace AZ::SceneAPI::SceneData
                 ->Version(1);
 
             serializeContext->Class<PrefabGroup, DataTypes::IPrefabGroup>()
-                ->Version(2) // added createProceduralPrefab
```
jackalbe, line 129: hmm, this seems to remove the option to not generate a proc prefab per scene asset... is that desired?

jackalbe (changes requested):

Gems/Prefab/PrefabBuilder/PrefabGroup/PrefabGroup.cpp
AMZN-stankowi: The toggle just did double duty over the group itself. With and without this change, you could turn off default prefabs by removing the change.

Before this change, the toggle and removing the group both turned off default prefabs, but the toggle wasn't as functional as removing the group. So this simplifies and consolidates the workflow to just removing the group.

jackalbe (approved):

PR #12344, opened 2022-09-30
[Impactful Change] Multiplayer CompileDef based Client/Server split

AMZN-puvvadar:
## What does this PR do?

This change enables the usage of macros (AZ_TRAIT_CLIENT and AZ_TRAIT_SERVER) to conditionally exclude logic in the Multiplayer Gem such that ServerLaunchers can be built without Client specific code and vice versa. The macro usage approach _also_ allows for a Unified mode and adds a UnifiedLauncher in which all logic is included. This is important for tools and to support Client as a Host based Launchers and allows developers to easily build all target types they may require.

These changes are based on the proposal in this RFC: https://github.com/o3de/sig-network/issues/77

Specifically:
1. Updates LyAutoGen such that multiple targets can specify the same codegen rules. This allows usage of the same ruleset/inputs with different compile flags. This does result in more generated output though this should not impact exe size. This change is **Impactful** as it affects the output path of generated code. It does not however, affect include paths used in code which keeps in code includes the same even with the split.
2. Split Multiplayer into several targets

- Common: All code in Multiplayer that is not split or does not depend on split code
- Client: Common + split code compiled for client usage
- Server: Common + split code compiled for server usage
- Unified: All code, intended for client host usage

3. Expected usage of AZ_TRAIT_CLIENT and AZ_TRAIT_SERVER to test changes and preview how split Multiplayer logic should be authored.
4. Treats Authority to Server NetworkInputs as always non scriptable. I did not want to have platform specific behavior context reflection and it felt unlikely that a user would want to script Auth to Server.
5. Adds a UnifiedLauncher target for users that may want a build featuring Client and Server logic (Client as a Host builds). All Gems have had their cmakes updated appropriately This particular change is **Impactful**.

Note that the intent of ServerLauncher in this prototype is to move it towards being headless.

Feedback on this approach is appreciated. This works but can we do better?
1. Can the cmake approach be simplified at all?
2. Is there a better approach with respect to CodeGen?
3. Is the usability of the Multiplayer API 'good'? Is the tradeoff of security vs usability acceptable as proposed?

## What should reviewers be looking at?
1. CMake, LauncherUnified and project Template changes to add the UnifiedLauncher
2. Changes to the Multiplayer Gem
3. Changes to the Multiplayer Scripting sub Gem. It uses the Multiplayer.Common target as it does not require MP components (which motivates most split logic).
[...]

PR #14091, opened 2023-01-18
Add UnifiedMultiplayerGem Template

AMZN-puvvadar:

## What does this PR do?

This change adds a template for Gems that would depend on the Multiplayer Gem and would require splitting Client and Server logic.

See https://github.com/o3de/o3de/pull/12344 for context

This use case is fairly niche but handles non-obvious or error-prone steps like setting up the Gem's CMake and adding Module declarations for Client, Server and Unified (which do need to be distinct in Release builds).

## How was this PR tested?

This PR was tested by

1. Running `scripts\o3de.bat create-gem -gp Gems\MultiplayerTemplateTest -tp Templates\UnifiedMultiplayerGem`
2. Configuring
3. Building the Client, Server and Unified targets of the Gem


lumberyard-employee-dm (commented):
You are going to need to add an add_subdirectory call for this template to the [Templates/CMakeLists.txt](https://github.com/o3de/o3de/blob/development/Templates/CMakeLists.txt) to make sure it gets copied into the SDK layout

Templates/UnifiedMultiplayerGem/Template/Code/Source/Clients/${Name}Module.cpp
```diff
@@ -0,0 +1,30 @@
...
+    class ${SanitizedCppName}Module
+        : public ${SanitizedCppName}ModuleInterface
+    {
+    public:
+        AZ_RTTI(${SanitizedCppName}Module, "${ModuleClassId}", ${SanitizedCppName}ModuleInterface);
+        AZ_CLASS_ALLOCATOR(${SanitizedCppName}Module, AZ::SystemAllocator, 0);
+    };
+}// namespace ${SanitizedCppName}
+
+#if defined(AZ_MONOLITHIC_BUILD)
```
lumberyard-employee-dm, line 26: Should this folder for Client files be renamed `Unified`?

Templates/UnifiedMultiplayerGem/template.json
```diff
@@ -0,0 +1,261 @@
...
+            "dir": "Code/Source/Clients"
+        },
+        {
+            "dir": "Code/Source/Tools"
+        },
+        {
+            "dir": "Code/Tests"
+        },
+        {
+            "dir": "Code/Tests/Clients"
```
lumberyard-employee-dm, line 252: Since this is for a unified client server gem, how about renaming the "Clients" folder to "Unified"?

AMZN-Gene:
Does this gem support multiplayer auto-components? Didn't see any auto-component AUTOGEN_RULES rules in the cmake file.

Could we sneak in an ExampleMultiplayerComponent.Autocomponent.xml/h/cpp? 
Would love to have that since I think a gem with MP components would typically be why someone uses this template.
Game projects could use this gem template as a go to place for storing all their multiplayer components.

```
<?xml version="1.0"?>

<Component
    Name="ExampleMultiplayerComponent" 
    Namespace="${GemName}" 
    OverrideComponent="true" 
    OverrideController="true" 
    OverrideInclude="Source/Components/ExampleMultiplayerComponent.h"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xsi:noNamespaceSchemaLocation="MultiplayerAutoComponentSchema.xsd">

    <ComponentRelation Constraint="Required" HasController="true" Name="NetworkTransformComponent" Namespace="Multiplayer" Include="Multiplayer/Components/NetworkTransformComponent.h" />

    <NetworkInput Type="float" Name="MyInput" Init="1.0f" />

    <NetworkProperty Type="float"  Name="MyNetworkProperty" Init="0.0f" Container="Object" ReplicateFrom="Authority" ReplicateTo="Client" IsPublic="false" IsRewindable="false" IsPredictable="false" ExposeToEditor="false" GenerateEventBindings="false" Description="" />

    <RemoteProcedure Name="MyRPC" InvokeFrom="Authority" HandleOn="Client" IsPublic="false" IsReliable="false" GenerateEventBindings="false" Description="" >
        <Param Type="float" Name="MyRPCParam" />
    </RemoteProcedure>

    <ArchetypeProperty Type="AZStd::string" Name="MyArchetype" Init="" Container="Array" Count="4" ExposeToEditor="true" Description="" />
</Component>
```

Templates/UnifiedMultiplayerGem/template.json
AMZN-puvvadar: Sure, I wasn't actually sure what the motivation for Clients was. I like the idea of Unified here.

AMZN-puvvadar:
> Does this gem support multiplayer auto-components? Didn't see any auto-component AUTOGEN_RULES rules in the cmake file.

The template does not but since it pulls Multiplayer in it could. I didn't want to assume this would be an MP Component Gem.

Templates/UnifiedMultiplayerGem/template.json
lumberyard-employee-dm: Clients was meant to represent the runtime code which is not tooling code.

It is meant to map pretty closely to the .Clients target.

AMZN-Gene:
> > Does this gem support multiplayer auto-components? Didn't see any auto-component AUTOGEN_RULES rules in the cmake file.
> 
> The template does not but since it pulls Multiplayer in it could. I didn't want to assume this would be an MP Component Gem.

I might just be misunderstanding the typical use-case (you mention it being niche above). Could you provide a README.md on who would want to use this type of Gem template and a command-line example for using it with o3de.bat

AMZN-puvvadar:
> > > Does this gem support multiplayer auto-components? Didn't see any auto-component AUTOGEN_RULES rules in the cmake file.
> > 
> > 
> > The template does not but since it pulls Multiplayer in it could. I didn't want to assume this would be an MP Component Gem.
> 
> I might just be misunderstanding the typical use-case (you mention it being niche above). Could you provide a README.md on who would want to use this type of Gem template and a command-line example for using it with o3de.bat.
> 
> For example, here's the script canvas node library template README.md. I found it useful https://github.com/o3de/o3de/blob/development/Templates/ScriptCanvasNode/Template/README.md

Sure I can add a README. The template summary describes it as well but more instruction doesn't hurt. The gist is this is a template for a Multiplayer dependent Gem where the user wants to do something different on the Client and Server. This could range from including different libraries, using different cmake lists or selectively compiling logic. It could be for anything from debug tooling, library integrations or component libraries as you suggested. What makes it niche is the dependency on Multiplayer.

lumberyard-employee-dm (approved):

AMZN-Gene:
> > > > Does this gem support multiplayer auto-components? Didn't see any auto-component AUTOGEN_RULES rules in the cmake file.
> > > 
> > > 
> > > The template does not but since it pulls Multiplayer in it could. I didn't want to assume this would be an MP Component Gem.
> > 
> > 
> > I might just be misunderstanding the typical use-case (you mention it being niche above). Could you provide a README.md on who would want to use this type of Gem template and a command-line example for using it with o3de.bat.
> > For example, here's the script canvas node library template README.md. I found it useful https://github.com/o3de/o3de/blob/development/Templates/ScriptCanvasNode/Template/README.md
> 
> Sure I can add a README. The template summary describes it as well but more instruction doesn't hurt. The gist is this is a template for a Multiplayer dependent Gem where the user wants to do something different on the Client and Server. This could range from including different libraries, using different cmake lists or selectively compiling logic. It could be for anything from debug tooling, library integrations or component libraries as you suggested. What makes it niche is the dependency on Multiplayer.

Ah, I totally forgot to the description in the template.json. That works! So to understand, the default gem template was also updated when client/server split occurred, but it didn't break out the logic for client and server, ie: the Default Gem template produces a gem which will just produce the usual output (ie MyGem.dll and MyGem.editor.dll). This PR is a gem that will break out client and server (IE: MyGem.dll, MyGem.editor.dll, MyGem.client.dll, and Gem.server.dll)?

Templates/UnifiedMultiplayerGem/Template/Code/CMakeLists.txt
```diff
@@ -0,0 +1,298 @@
...
+    BUILD_DEPENDENCIES
+        PUBLIC
+            Gem::${gem_name}.API
+        PRIVATE
+            Gem::${gem_name}.Unified.Private.Object
+)
+
+# By default, we will specify that the above target ${gem_name} would be used by
+# Client and Server type targets when this gem is enabled.  If you don't want it
+# active in Clients or Servers by default, delete one of both of the following lines:
```
AMZN-Gene, line 164: Does this sentence need to be updated/removed?

AMZN-Gene:
If this is producing a MyGem.server.dll and MyGem.client.dll I might suggest adding a little code snippet somewhere demonstrating to developers how to make their logic end up in one place or the other. Something simple for users to see... like a AZPrint message in the SystemComponent::Activate that changes based on if def AZ_TRAIT_CLIENT or if def AZ_TRAIT_SERVER.

AMZN-Gene (approved):

Templates/UnifiedMultiplayerGem/Template/Code/CMakeLists.txt
AMZN-puvvadar: It can stay. It's valid to want Client and Unified but not Server for example.

cgalvan (approved):

lumberyard-employee-dm (approved):

PR #14098, opened 2023-01-18
EditContext EnumValue -> GenericValue wrapper

lumberyard-employee-dm:
## What does this PR do?

Adding shim to EditContext Attribute to add a GenericValueList attribute whenever an EnumValues attribute is reflected.

It does this by wrapping the data value or function that is used to obtain a `vector<EnumConstants<EnumType>>` to return a 
`vector<pair<EnumType, string>`

From @amzn-tmryan Added a shim to the `EnumAttribute` method to make a `GenericValue` value attribute for each legacy `EnumValue` attribute, so that enum value attributes can be read using the same API reading GenericValues.

## How was this PR tested?

Tested it as part of the PropertyEditor DPE attribute work by @amzn-tmryan

### DIFF
--- a/Code/Framework/AzCore/AzCore/Serialization/EditContext.h
+++ b/Code/Framework/AzCore/AzCore/Serialization/EditContext.h
@@ -788,30 +788,6 @@ namespace AZ
     //=========================================================================
     // Attribute
     //=========================================================================
-    template<class T>
-    EditContext::ClassBuilder*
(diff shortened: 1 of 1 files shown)


cgalvan (approved):
Thanks!

amzn-phist (approved):

amzn-tmryan (approved):

Issue #13660, opened 2022-12-12
[Terrain] Terrain Macro Material sometimes uses wrong mipmap


mbalfour-amzn:
**Describe the bug**
The Terrain Macro Material component intermittently uses the wrong mipmap for an image.

**Assets required**
![color_noise](https://user-images.githubusercontent.com/82224783/207152487-8f65ccb4-1513-4881-9994-63f0e6c77251.png)

**Steps to reproduce**
1. Create a new level.
2. Save the attached image as "color_noise.png" in the level.
3. Add Terrain World and Terrain World Renderer components to the level.
4. Delete the Grid, Shader Ball, and Ground components.
5. Create a entity at (0, 0, 0) with Terrain Layer Spawner, an Axis Aligned Box Shape of (256, 256, 100) and a Terrain Macro Material.
6. Turn off "Filled" on the Axis Aligned Box Shape.
7. Set the Color Texture to color_noise.png.
8. Notice how the noise is blurry on the texture.
![image](https://user-images.githubusercontent.com/82224783/207153207-fe3bd9b3-f80c-4a56-8178-48bbf5c8cea4.png)

9. Duplicate the entity and move it to (256, 0, 0).
10. Notice how the noise is now crisp on both entities.
![image](https://user-images.githubusercontent.com/82224783/207153378-8ab8181d-fb48-4fe3-b2ed-f8af16a29bcc.png)

11. Click on the "eyeball" icon next to both entities in the Entity Outliner once to hide them both, then again to show them both.
![image](https://user-images.githubusercontent.com/82224783/207153692-3c703394-0d09-47ad-bbf8-7e57cc060e54.png)

**Expected behavior**
Both entities should have a crisp noise texture on steps 8-11.

**Actual behavior**
The mipmap selection seems to differ on the entities intermittently.

**Found in Branch**
development

mbalfour-amzn:
This might be related to the recent bindless image changes. It appears that the TerrainMacroMaterialManager is storing a colorMapId and normalMapId in the shader buffers that refer to specific mip levels, but it isn't getting any notifications that it needs to change these Id values as mips get asynchronously loaded and evicted.

invertednormal:
This same issue happens with detail materials, agree it something related to bindless, not macro materials specifically.

mbalfour-amzn:
Also, asset reloads (such as modifying a macro material source image outside the Editor while the Editor is running) causes a "GPU device has been removed" crash.

[...]

PR #14102, opened 2023-01-18
Fixing ImageView and BufferView to have consistent bindless IDs after invalidation.

invertednormal:
## What does this PR do?

Previously, when an ImageView or BufferView was invalidated, the old location on the bindless array was deallocated and a new one was created. This meant that it was possible for the bindless ID of the view to change under the hood, which could cause crashes in anything using that view in a bindless way. This updates DirectX, Vulkan, and Metal to replace the old location instead of allocating a new location when a view is being invalidated and reinitialized.

This also fixes BufferView invalidation in Vulkan which was previously just shutting down the BufferView instead of reinitializing it. Making it reinitialize keeps the behavior consistent with the way both ImageView on Vulkan and BufferView on Dx12 work.

The root issue this PR is attempting to address is from #13660 where as mips streamed in for terrain detail materials (which use bindless), the image would not update (dx12) or it would lead to a crash (vulkan).

Fixes: #13660

Tested the full ASV test suite. Also tested with bindless terrain locally where the mip issue was originally occurring.

### DIFF
--- a/Gems/Atom/RHI/Code/Include/Atom/RHI/BufferView.h
+++ b/Gems/Atom/RHI/Code/Include/Atom/RHI/BufferView.h
@@ -25,6 +25,8 @@ namespace AZ
             AZ_RTTI(BufferView, "{3012F770-1DD7-4CEC-A5D0-E2FC807548C1}", ResourceView);
             virtual ~BufferView() = default;
 
+            static constexpr uint32_t InvalidBindlessIndex = 0xFFFFFFFF;
+
             //! Initializes the buffer view with the provided buffer and view descriptor.
             ResultCode Init(const Buffer& buffer, const BufferViewDescriptor& viewDescriptor);
 
@@ -46,12 +48,12 @@ namespace AZ
 
             virtual uint32_t GetBindlessReadIndex() const
             {
-                return 0xffffffff;
+                return InvalidBindlessIndex;
             }
 
             virtual uint32_t GetBindlessReadWriteIndex() const
             {
-                return 0xffffffff;
+                return InvalidBindlessIndex;
             }
 
         protected:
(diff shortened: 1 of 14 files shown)


Gems/Atom/RHI/Vulkan/Code/Source/RHI/BufferView.cpp
```diff
@@ -80,30 +80,47 @@ namespace AZ
...
 
         RHI::ResultCode BufferView::InvalidateInternal()
+        {
+            ReleaseView();
+            RHI::ResultCode initResult = InitInternal(GetDevice(), GetResource());
+            if (initResult != RHI::ResultCode::Success)
+            {
+                ReleaseBindlessIndices();
+            }
+            return initResult;
```
moudgils, line 90: Is it possible to make a similar change for metal too as this issue will exist on metal as well. I can help with testing.

invertednormal: Sure - I'll look into metal, no way to compile that on my end though so it's a little tricky.

Gems/Atom/RHI/Code/Include/Atom/RHI/BufferView.h
```diff
@@ -25,6 +25,8 @@ namespace AZ
             AZ_RTTI(BufferView, "{3012F770-1DD7-4CEC-A5D0-E2FC807548C1}", ResourceView);
             virtual ~BufferView() = default;
 
+            static constexpr uint32_t InvalidBindlessIndex = 0xFFFFFFFF;
```
moudgils, line 28: Maybe easier to move this variable to Limits.h so we dont have multiple definitions

Gems/Atom/RHI/DX12/Code/Source/RHI/DescriptorContext.cpp
```diff
@@ -310,7 +320,16 @@ namespace AZ
                 }
             }
             CopyDescriptor(unorderedAccessViewClear, unorderedAccessView);
-            staticView = AllocateStaticDescriptor(unorderedAccessDescriptor);
+
+            if (staticView.m_index == DescriptorHandle::NullIndex)
```
moudgils, line 324: same comment as above

moudgils (approved):

rgba16f (approved):

Gems/Atom/RHI/Code/Include/Atom/RHI/BufferView.h
invertednormal: It probably does make more sense to merge these two into one common place. There's not really a section in Limits for this kind of thing, but one could be added. We could also add it to the common ResourceView. Holding off on this for now since it's a minor change and I'd rather not re-run AR for it.

PR #14103, opened 2023-01-18
Re-factor bus binding strings inside atom_tools_utils.py

jromnoa:
## What does this PR do?
Re-factors the hard-coded string values for bus calls into constants stored on a class object. These class objects match the BehaviorContext bindings visible in the C++ code. There are probably some more, but this covers the existing tests + adds some extra ones since we'll probably need them in the near future.

## How was this PR tested?
Unable to run local tests yet but will run them once I resolve an ACME issue on the threadripper. These code changes are just moving strings into class objects and replacing the old hard-coded strings with these new strings inside the new class object so it should be fine.


FuzzyCarterAWS (approved):
This looks great.

As a styling choice, I'd rename each of these container classes from `SomeContainerBusStrings` to `SomeContainerBusEvents` to driver better readability and understanding to the classes at a glance.

smurly (approved):

PR #14108, opened 2023-01-19
Removing actors from reflection cubemap images.

### LINKED ISSUE #14106: Reflection Probes: Remove Actors by Default
Character actors will appear in reflection probes when generated at editor time or runtime. This is undesirable since characters are moving/dynamic objects and their locations in world won't match the static reflection probes. They should be excluded from reflection probes.

image

Update Actor Component so that actors cull themselves from the reflection probe pass using.

See _MeshFeatureProcessor::SetExcludeFromReflectionCubeMaps_ 

If it doesn't affect performance, maybe make this an Actor component flag "Appears In Reflection Probes" instead of hardcoding actors to always be culled? Maybe some cases it's okay, for example big statues which only rarely move, or NPCs that have subtle animations, maybe it's desirable for them to be in reflections?

invertednormal:
## What does this PR do?

This PR updates `AtomActorInstance` so that the meshes it uses aren't reflected in cubemaps (since they are dynamic) . This change also moves `m_excludeFromReflectionCubemaps` from the `ModelDataInstance` to the `MeshHandleDescriptor` so it can be set on `AcquireMesh()`.

Fixes #14106

This change also exposes `SetExcludeFromReflectionCubeMaps()` and `GetExcludeFromReflectionCubeMaps()` to the `MeshComponentBus` (thanks @CyaNinix for the idea)

## How was this PR tested?

Tested in Editor to ensure a generated cubmap no longer shows an actor in the reflection

image

### DIFF
--- a/Gems/Atom/Feature/Common/Code/Include/Atom/Feature/Mesh/MeshFeatureProcessor.h
+++ b/Gems/Atom/Feature/Common/Code/Include/Atom/Feature/Mesh/MeshFeatureProcessor.h
@@ -210,6 +210,7 @@ namespace AZ
             RPI::Cullable::LodConfiguration GetMeshLodConfiguration(const MeshHandle& meshHandle) const override;
 
             void SetExcludeFromReflectionCubeMaps(const MeshHandle& meshHandle, bool excludeFromReflectionCubeMaps) override;
+            bool GetExcludeFromReflectionCubeMaps(const MeshHandle& meshHandle) const override;
             void SetIsAlwaysDynamic(const MeshHandle& meshHandle, bool isAlwaysDynamic) override;
             bool GetIsAlwaysDynamic(const MeshHandle& meshHandle) const override;
             void SetRayTracingEnabled(const MeshHandle& meshHandle, bool rayTracingEnabled) override;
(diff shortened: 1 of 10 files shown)


amzn-tommy (approved):
Now that this is in the descriptor, you should set this in MeshComponentController.cpp before calling AcquireMesh, and remove the subsequent call to SetExcludeFromReflectionCubeMaps.

Same change should apply to IsAlwaysDynamic, while you're in there.

AMZN-alexpete (approved):

AMZN-alexpete:
@CyaNinix @CyaSid @CyaMika will this change help exclude actors from your dynamic cubemaps?

AMZN-Gene (approved):

qcouillard:
Hi @AMZN-alexpete 
Yes, totally, thank you.
We also made something in local to allow that, we have added a Bus method to set this boolean and add it to ActorComponents, with option in Editor component in case.
I think that both will complete each other and @CyaNinix will soon make a PR for it.

CyaNinix:
Here's my PR: https://github.com/o3de/o3de/pull/14119

This allows to selectively exclude actors from cubemap reflections (and could be on by default).

AMZN-alexpete (approved):

invertednormal:
@CyaNinix I like your additions to the mesh controller bus, that was definitely an oversight when this property was added. Mind if I bring those changes over to this PR?

I'm on the fence about adding the exclude from cubemap option to the actor component - It's hard to think of good use cases for that, and once we add an editor property it becomes very hard to change or remove it later (an upgrade path would need to be provided for the component which is doable, but a pain).

PR #14112, opened 2023-01-19
gem.json and project.json version updates

AMZN-alexpete:
## What does this PR do?

In an effort to secure world peace I present this finely crafted set of changes.

Though these changes may seem related to https://github.com/o3de/sig-core/issues/44 they are in-fact designed to inspire a generation and bring about version 1.0.0 of ... something ... as they same in marines, "semverfy"

- Updated gem.json files to have a version field set to "0.0.0"
- Updated project.json files to have a version field set to "1.0.0"
- Updated templates and O3DE CLI to allow setting these values upon creation

## How was this PR tested?

- create a gem and project and examined the gem.json and project.json output to confirm the default versions were used
- ran updated O3DE CLI unit tests


rgba16f (approved):

Templates/DefaultProject/Template/project.json
```diff
@@ -1,5 +1,6 @@
 {
     "project_name": "${Name}",
+    "version": "1.0.0",
```
lumberyard-employee-dm, line 3: What is the rationale for the Projects, not have support for a a ${version} placeholder?

scripts/o3de/o3de/engine_template.py
```diff
@@ -2625,6 +2633,8 @@ def add_args(subparsers) -> None:
     create_project_subparser.add_argument('--no-register', action='store_true', default=False,
                                           help='If the project template is instantiated successfully, it will not register the'
                                                ' project with the global or engine manifest file.')
+    create_project_subparser.add_argument('-v', '--version', type=str, required=False,
+                                          help='An optional version. Defaults to 1.0.0')
```
lumberyard-employee-dm, line 2637: If the default is 1.0.0, why isn't it set here?

lumberyard-employee-dm: Also `-v` is normally using for a `verbose` option as it relates to logginc.

AMZN-alexpete: It could be, I mostly tried to keep the defaults in the same location - which are currently in the `create_project` function.

Templates/DefaultProject/Template/project.json
AMZN-alexpete: Good catch!  I forgot to update these.

lumberyard-employee-dm (approved):

cgalvan (approved):

PR #14113, opened 2023-01-19
Removed unused files at the root the O3DE repo.

lumberyard-employee-dm:
The editor.cfg file has been replaced with the o3de.editor.setreg file in the `<engine-root>/Registry` directory.
Its settings can be overridden by a project or gem setreg file.

## How was this PR tested?

Launched and loaded an a level in the Editor with no issues.


hultonha (approved):
LGTM but quick question.

I still locally use `user.cfg` for some custom settings that are quick to enable/disable (various AZ_CVARS for example). Will that will work after this change or not? It is so much more convenient editing the .ini file than a JSON file which is why I'd be sorry to lose it (I know maybe in future we've talked about yaml support to make editing things easier, but not sure when that's going to happen). Thanks!

cgalvan (approved):

hultonha:
> > LGTM but quick question.
> > I still locally use `user.cfg` for some custom settings that are quick to enable/disable (various AZ_CVARS for example). Will that will work after this change or not? It is so much more convenient editing the .ini file than a JSON file which is why I'd be sorry to lose it (I know maybe in future we've talked about yaml support to make editing things easier, but not sure when that's going to happen). Thanks!
> 
> Windows style INI files are still support via the Console API [ExecuteConfigFile](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Console/Console.cpp#L116-L164) method. The secret however is that cfg settings are stored into the SettingRegistry and then get executed by the Notification Handler that the AZ::Console is hooked to listen for settings changes underneath the [/Amazon/AzCore/Runtime/ConsoleCommandKey](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Console/Console.cpp#L617-L621).
> 
> As you can see from this change, no code was actually changed, the legacy Cry code is still trying to load an editor.cfg file from [CryEdit.cpp](https://github.com/o3de/o3de/blob/development/Code/Editor/CryEdit.cpp#L1264-L1266).
> 
> I will however point out that the Windows style-INI system has limitations in O3DE. Unlike the Settings Registry settings there is no clear documentation in how settings are overwritten. Such as the Engine Registry settings are loaded first and then the Gem Registry settings are merged overtop of that and then the Project Registry settings are merged overtop of that and then the per-user non-source controled Project User Registry settings are merged overtop of that.
> 
> Second all the settings in the .cfg files have to correspond to existing CVars declared in code. Those settings can't be random key, value pairs.

Got you, that's helpful to know, thank you for the explanation @lumberyard-employee-dm, much appreciated 👍

burelc-amzn (approved):

lmbr-pip:
Do we need to do the same with client.cfg and server.cfg? ie https://github.com/o3de/o3de-multiplayersample

Is the recommendation to move away from .cfg

lumberyard-employee-dm:
> Do we need to do the same with client.cfg and server.cfg? ie https://github.com/o3de/o3de-multiplayersample
> 
> Is the recommendation to move away from .cfg
I think the `client.cfg` and `server.cfg` files are OK with being within a project.
The o3de repo root should avoid having loose files, as when we create an SDK layout, we should treat modifications to the engine as immutable.
For a project this can be left up to project owner.

I would recommend however, not having configuration files within the root of the project however.
For example I would recommend that at the very least the cfg files are in the `<project-root>/User` directory probably under some `Config` directory

PR #14116, opened 2023-01-19
Added an explicit binary directory to the mono-default configure CMake

lumberyard-employee-dm:
Presets.

Previously these presets did not specify a binary directory, therefore using these presets would default to using the binary directory of the first inherited preset.

For example windows this would result in the monolithic build being placed in the same folder as non-monolithic builds(`build/windows`), because the windows_unity target has a binary directory set to `build/windows`. his makes it so that by default the monolithic binary directory is a different directory than the non-monolithic binary directory. This can overridden via the `-B <binary-dir>` option when cmake configuring.

For more information see the
[configure_preset](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html?highlight=presets#configure-preset) inherits entry documentation for inheritance of presets work.

## How was this PR tested?

Configure using the Engine using following command and verified that the binary directory was `<engine-root>/build/windows_mono`
```cmake
cmake --preset windows-mono-default -DLY_PROJECTS=AutomatedTesting
```

### DIFF
--- a/cmake/Platform/Android/CMakePresets.json
+++ b/cmake/Platform/Android/CMakePresets.json
@@ -25,6 +25,7 @@
       "displayName": "Android Monolithic with Unity and Ninja",
       "description": "Configures Android to build the Monolithic permutation using Ninja with Unity builds",
       "toolchainFile": "${sourceDir}/cmake/Platform/Android/Toolchain_android.cmake",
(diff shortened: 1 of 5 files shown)


amzn-changml (approved):

cgalvan (approved):
Thanks for doing this!

burelc-amzn (approved):

PR #14119, opened 2023-01-19
Allow to exclude actors from reflections

CyaNinix:
## What does this PR do?

This PR allows to excludes actors from cubemap reflections, adding the same option from mesh components to actor component:
image

## How was this PR tested?

In action:
image


Gems/EMotionFX/Code/Source/Integration/Editor/Components/EditorActorComponent.cpp
```diff
@@ -139,6 +140,9 @@ namespace EMotionFX
                         ->Attribute(AZ::Edit::Attributes::ChangeNotify, &EditorActorComponent::OnSkinningMethodChanged)
                         ->EnumAttribute(SkinningMethod::DualQuat, "Dual quat skinning")
                         ->EnumAttribute(SkinningMethod::Linear, "Linear skinning")
+                        ->DataElement(AZ::Edit::UIHandlers::CheckBox, &EditorActorComponent::m_excludeFromReflectionCubeMaps, "Exclude from reflection cubemaps", "Actor will not be visible in baked reflection probe cubemaps")
```
amzn-tommy, line 143: Do you have a use-case where you do want to enable actors in the reflection probes, or where you want to enable/disable via the MeshComponentBus? If not, then I'd say let's keep it simple and just always disable actors in reflection probes via https://github.com/o3de/o3de/pull/14108. We can always revive this change if the need arises.

If there is a use case, then I think we should ping sig-ui-ux for suggestions on where to put this attribute. I imagine we want to add a new group for render settings, put the group at the bottom of all the Actor component attributes, and have it collapsed by default, then put this setting in that group. It's something that most people will never touch, so we probably don't want it right at the surface of the UI. And with a render settings group, we can then add other mesh component settings like 'Use Forward Pass IBL Specular', or 'Use raytracing' (once we support that for skinned meshes).

AMZN-alexpete: The only use case I can think of is when you use an actor for something that has a subtle or inactive animation like vegetation - not sure how common it is but we were going to use animated trees in the robotics demo to simulate wind in a pinch.

CyaQcouillard: Little bit like trees, we have a case where we could want to enable reflection probes on actors, entities like an audience.
In our case, we have an audience, so a group of many stuck entities together which can have a great influence in reflection probes, but in same time they have animations (to have a little bit of life) and so are actors.
I think this case works in both case precomputed and dynamic reflection probes, because the audience position doesn't change at runtime (for us), so can be taken into account for probes computation in editor without creating too big inconsistency in game.
Naturally dynamic case is more complex and will depend on what designers and developers want.
For example, in case of a First Person game (like exploration game) we could want to exclude actors for probes to save performances, but at the same time keep them "on" for the player character to have a good effect when the player come near a reflective surface.
Or more original, a game with vampires (no reflection) and humans (reflection) ^^.

invertednormal: So I checked in #14108, so that Actors are by default hidden now, however the question of adding a control to make it optional for Actors remains.

Personally, I've come around to being ok with adding this control, but agree with @amzn-tommy that we should bury it in a collapsed "Advanced" section or something similar, and also the "Exclude from cube maps" control would default to on for Actors.

Pinging the ux sig for recommendations. @o3de/sig-ui-ux

invertednormal: Another thing to keep in mind, if the UX for the Actor Component changes, then the docs will need to be updated as well - https://www.o3de.org/docs/user-guide/components/reference/animation/actor/

hultonha:
Tagging @amzn-rhhong for visibility as well

rainbj:
vampire mode?

Gems/EMotionFX/Code/Source/Integration/Editor/Components/EditorActorComponent.cpp
amzn-tommy: Trees and crowds seem like plausible use cases for now (given a lack of built in vegetation bending for static meshes or bespoke crowd animation system), so adding a control to the UI makes sense to me. It just shouldn't be front and center.

Gems/EMotionFX/Code/Source/Integration/Components/ActorComponent.cpp
```diff
@@ -156,6 +156,7 @@ namespace EMotionFX
                     ->Field("BoundingBoxConfig", &Configuration::m_bboxConfig)
                     ->Field("ForceJointsUpdateOOV", &Configuration::m_forceUpdateJointsOOV)
                     ->Field("RenderFlags", &Configuration::m_renderFlags)
+                    ->Field("ExcludeFromReflectionCubeMaps", &Configuration::m_excludeFromReflectionCubeMaps)
```
AMZN-alexpete, line 159: Let's bump the Version to 6 on line 149 to reflect this new field being added

Gems/EMotionFX/Code/Source/Integration/Components/ActorComponent.h
```diff
@@ -94,6 +94,8 @@ namespace EMotionFX
                 bool m_forceUpdateJointsOOV = false;
                 BoundingBoxConfiguration m_bboxConfig; ///< Configuration for bounding box type and updates
 
+                bool m_excludeFromReflectionCubeMaps = false;
```
AMZN-alexpete, line 97:
```suggestion
                bool m_excludeFromReflectionCubeMaps = true;
```
From the discussion it sounds like we want to default this to exclude actors from reflections cube maps by default.

Gems/EMotionFX/Code/Source/Integration/Editor/Components/EditorActorComponent.cpp
```diff
@@ -60,6 +60,7 @@ namespace EMotionFX
                     ->Field("UpdateJointTransformsWhenOutOfView", &EditorActorComponent::m_forceUpdateJointsOOV)
                     ->Field("LodLevel", &EditorActorComponent::m_lodLevel)
                     ->Field("BBoxConfig", &EditorActorComponent::m_bboxConfig)
+                    ->Field("ExcludeFromReflectionCubeMaps", &EditorActorComponent::m_excludeFromReflectionCubeMaps)
```
AMZN-alexpete, line 63: Lets bump the Version number to 5 on line 50 to reflect this new field being added

Gems/EMotionFX/Code/Source/Integration/Editor/Components/EditorActorComponent.cpp
```diff
@@ -188,6 +192,7 @@ namespace EMotionFX
             , m_attachmentJointIndex(0)
             , m_lodLevel(0)
             , m_actorAsset(AZ::Data::AssetLoadBehavior::NoLoad)
+            , m_excludeFromReflectionCubeMaps(false)
```
AMZN-alexpete, line 195:
```suggestion
            , m_excludeFromReflectionCubeMaps(true)
```
It sounds like we want to default to exclude actors, which appears to be the case since this change:
https://github.com/o3de/o3de/pull/14108/files#diff-489dc962e9ade1ed57ce21c859910ec56a83c0192caf65b937d2de6d91d0404eR671

Gems/AtomLyIntegration/EMotionFXAtom/Code/Source/AtomActorInstance.h
```diff
@@ -162,6 +162,7 @@ namespace AZ
             void SetExcludeFromReflectionCubeMaps(bool excludeFromReflectionCubeMaps) override;
             bool GetExcludeFromReflectionCubeMaps() const override;
             // GetWorldBounds/GetLocalBounds already overridden by BoundsRequestBus::Handler
+            void SetExcludeFromReflectionCubeMaps(bool excludeFromReflectionCubeMaps) override;
```
AMZN-alexpete, line 165: This function is now declared twice
```suggestion

```

Gems/AtomLyIntegration/CommonFeatures/Code/Source/Mesh/MeshComponentController.h
```diff
@@ -97,6 +97,7 @@ namespace AZ
             void SetModelAssetPath(const AZStd::string& modelAssetPath) override;
             AZStd::string GetModelAssetPath() const override;
             AZ::Data::Instance<RPI::Model> GetModel() const override;
+            void SetExcludeFromReflectionCubeMaps(bool excludeFromReflectionCubeMaps) override;
```
AMZN-alexpete, line 100: This function already exists on line 132
```suggestion
```

Gems/AtomLyIntegration/CommonFeatures/Code/Include/AtomLyIntegration/CommonFeatures/Mesh/MeshComponentBus.h
```diff
@@ -96,6 +96,9 @@ namespace AZ
 
             //! Returns the axis-aligned bounding box in model space.
             virtual AZ::Aabb GetLocalBounds() = 0;
+
+            //! Sets the option to exclude this mesh from baked reflection probe cubemaps
+            virtual void SetExcludeFromReflectionCubeMaps(bool excludeFromReflectionCubeMaps) = 0;
```
AMZN-alexpete, line 101: This is already defined on line 90
```suggestion
```

Gems/AtomLyIntegration/CommonFeatures/Code/Source/Mesh/MeshComponentController.cpp
```diff
@@ -524,6 +524,15 @@ namespace AZ
...
         }
+
+        void MeshComponentController::SetExcludeFromReflectionCubeMaps(bool excludeFromReflectionCubeMaps)
+        {
+            m_configuration.m_excludeFromReflectionCubeMaps = excludeFromReflectionCubeMaps;
+            if (m_meshFeatureProcessor)
+            {
+                m_meshFeatureProcessor->SetExcludeFromReflectionCubeMaps(m_meshHandle, excludeFromReflectionCubeMaps);
+            }
+        }
```
AMZN-alexpete, line 535: this is already defined on line 662
```suggestion
```

Gems/EMotionFX/Code/Source/Integration/Components/ActorComponent.cpp
```diff
@@ -155,6 +155,7 @@ namespace EMotionFX
...
```
AMZN-alexpete, line 158: can you bump the Version number on line 149 from 6 to 7 please?

AMZN-alexpete (approved):
Looks good!

Gems/EMotionFX/Code/Source/Integration/Components/ActorComponent.cpp
amzn-tommy: FWIW, when adding members, you do not need to bump the version number. Unless you know differently.

amzn-tommy:
> Yes, even if the discusion is in pause since a long time. But I think that UX for Actor Component subject go a little bit far from the original PR subject. The original purpuse is juste to user allow the exclusion of actors for relfection pass and maybe the UX question can be treat in another Issues or Discusion

You are adding a new element to the user interface, so I don't think getting feedback from #sig-ui-ux is straying too far from the original PR. You can create a GHI if you want to follow up, but then you'll have to update the documentation a second time to reflect the new change (if it changes). IMO it's less work to ping https://discord.com/channels/805939474655346758/816043761687068712 with a quick question up front rather than handling it later.

Gems/EMotionFX/Code/Source/Integration/Editor/Components/EditorActorComponent.cpp
amzn-tommy: Looks like this is already under a group. Apologies for missing that the first time around. There is no indentation in the code where the group starts, so it was not obvious.

AMZN-alexpete:
@CyaQcouillard  the latest commit could not be built due to this
```

[2023-04-03T19:02:51.606Z] D:/workspace/o3de/Gems/EMotionFX/Code/Tests/RenderBackendManagerTests.cpp(99): error C2259: 'EMotionFX::Integration::TestRenderActorInstance': cannot instantiate abstract class [D:\workspace\o3de\build\windows\External\EMotionFX-78fb240e\Code\EMotionFX.Tests.vcxproj]

[2023-04-03T19:02:51.606Z]                   return aznew TestRenderActorInstance(entityId, actorInstance, asset, skinningMethod, worldTransform);

[2023-04-03T19:02:51.606Z]   D:/workspace/o3de/Gems/EMotionFX/Code/Tests/RenderBackendManagerTests.cpp(44): note: see declaration of 'EMotionFX::Integration::TestRenderActorInstance'

[2023-04-03T19:02:51.606Z]           class TestRenderActorInstance

[2023-04-03T19:02:51.606Z]   D:/workspace/o3de/Gems/EMotionFX/Code/Tests/RenderBackendManagerTests.cpp(99): note: due to following members:

[2023-04-03T19:02:51.606Z]                   return aznew TestRenderActorInstance(entityId, actorInstance, asset, skinningMethod, worldTransform);

[2023-04-03T19:02:51.606Z]   D:/workspace/o3de/Gems/EMotionFX/Code/Tests/RenderBackendManagerTests.cpp(99): note: 'void EMotionFX::Integration::RenderActorInstance::SetExcludeFromReflectionCubeMaps(bool)': is abstract

[2023-04-03T19:02:51.606Z]                   return aznew TestRenderActorInstance(entityId, actorInstance, asset, skinningMethod, worldTransform);

[2023-04-03T19:02:51.606Z]   D:\workspace\o3de\Gems\EMotionFX\Code\Source\Integration/Rendering/RenderActorInstance.h(50): note: see declaration of 'EMotionFX::Integration::RenderActorInstance::SetExcludeFromReflectionCubeMaps'

[2023-04-03T19:02:51.606Z]               virtual void SetExcludeFromReflectionCubeMaps(bool excludeFromReflectionCubeMaps) = 0;
```
I think you just need to add the method to the test class or mock to fix it.

PR #14134, opened 2023-01-19
when the register show is given the project name it should return the prokect path

byrcolin:
This can be used via cli to get the project path from the project name


lumberyard-employee-dm (changes requested):
How will uses print out all the gems associated with the project then if they use the verbose option?

scripts/o3de/o3de/print_registration.py
```diff
@@ -305,6 +305,11 @@ def print_repos(verbose: int) -> int:
 
 def register_show(verbose: int, project_path: pathlib.Path = None, project_name: str = None) -> int:
     json_data = manifest.load_o3de_manifest()
+    if project_name:
+        project_path = get_project_path(project_path, project_name)
+        print(project_path.as_posix())
```
spham-amzn, line 310: How does this protect against an invalid ```project_name``` ?

PR #14147, opened 2023-01-20
Action Manager | Refactor Identifier definition to make them easier to reuse.

AMZN-daimini:
## What does this PR do?
Introduces headers to define identifiers for Editor Action Contexts, Action Updaters, Menus and ToolBars so that they can be reused easily in C++.

I have tried to limit the proliferation of different headers and avoided creating one for action identifiers at this moment. If the need arises, I will work on a different PR.

## How was this PR tested?
Built and tested manually.


cgalvan (approved):
Nice!

gadams3 (approved):

lsemp3d (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Editor/ActionManagerIdentifiers/EditorActionUpdaterIdentifiers.h
```diff
@@ -0,0 +1,28 @@
...
+ *
+ */
+
+#pragma once
+
+#include <AzCore/std/string/string_view.h>
+
+namespace EditorActionUpdater
+{
+    static constexpr AZStd::string_view AngleSnappingStateChangedUpdaterIdentifier = "o3de.updater.onAngleSnappingStateChanged";
```
hultonha, line 15: update: I believe you can omit `static` here (and `inline` should be preferred if anything but I believe `constexpr` already implies this

Code/Framework/AzToolsFramework/AzToolsFramework/Editor/ActionManagerIdentifiers/EditorContextIdentifiers.h
```diff
@@ -0,0 +1,16 @@
...
+ *
+ */
+
+#pragma once
+
+#include <AzCore/std/string/string_view.h>
+
+namespace EditorActionContext
+{
+    static constexpr AZStd::string_view MainWindowContextIdentifier = "o3de.context.editor.mainwindow";
```
hultonha, line 15: update: Remove `static`

Code/Framework/AzToolsFramework/AzToolsFramework/Editor/ActionManagerIdentifiers/EditorMenuIdentifiers.h
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+#pragma once
+
+#include <AzCore/std/string/string_view.h>
+
+namespace EditorMenu
+{
+    // Editor main window top MenuBar
+    static constexpr AZStd::string_view EditorMainWindowMenuBarIdentifier = "o3de.menubar.editor.mainwindow";
```
hultonha, line 16: update: Remove `static` (see earlier comment) - this is fine in .cpp files but should be avoided in .h files

hultonha (commented):
Great to see this change but I think we should update the use of `static` in the headers. Other than that LGTM

Code/Framework/AzToolsFramework/AzToolsFramework/Editor/ActionManagerIdentifiers/EditorActionUpdaterIdentifiers.h
AMZN-daimini: Good point, will do.

AMZN-daimini: Looks like `constexpr` does not imply `inline` for non-static variables. Would it make sense to write `inline constexpr`?

AMZN-daimini: (Watched some videos and read some docs, it makes sense to use `inline constexpr` for variables in header space. Doing that!)

Issue #14146, opened 2023-01-20
New Action Manager doesn't support "Esc" to exit entity picking while in paintbrush mode


mbalfour-amzn:
**Describe the bug**
When using the paintbrush component mode, if the new Action Manager has been enabled, "Esc" doesn't work to exit entity picking.

**Steps to reproduce**
1. In user/registry/editorpreferences.setreg, enable the new Action Manager:
```
    "O3DE": {
        "ActionManager": {
            "EnableNewActionManager": true
	},
```
2. Open the Editor with a new or existing level.
3. Create a new entity with an Image Gradient, a Gradient Transform Modifier, and a Box Shape.
4. On the Image Gradient, select any available existing image for the Image Asset.
5. On the Image Gradient, press the Edit button to start the paintbrush component mode.
6. On the Image Gradient, press the Entity Picker icon next to the "Pin Preview to Entity" field.
7. Press Esc to exit entity picking.
8. Press Esc to exit component mode.

**Expected behavior**
The first "Esc" press exits entity picking mode and returns to painting mode in the viewport. The second "Esc" exits the component mode entirely.

**Actual behavior**
Pressing "Esc" does nothing.

**Found in Branch**
development

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
31b8100ecaebaa9ad4f65e67e403afdd458afb17

**Additional context**
This works correctly when the new Action Manager is not enabled.

mbalfour-amzn:
This is related to #13467 though even after that is fixed, this bug should still remain, because there's currently no code in PaintBrushManipulator::RegisterActions() to override the Esc key while picking entities. In contrast, in PaintBrushManipulator::PopulateActionsImpl() there is some logic at the top to override it while in entity-picking mode. The logic doesn't transfer over as-is because PopulateActionsImpl() is called during component mode and can refresh the actions, but RegisterActions() doesn't have an equivalent concept of refreshing actions.

hultonha:
Assigning to @AMZN-daimini to take a look at. Let me know if you have any questions. Thanks!

PR #14149, opened 2023-01-20
WIP: Ambiguous shortcut handling for new action manager

cgalvan:
## What does this PR do?

WIP: Handling ambiguous shortcuts with the new action manager. Specifically, when a parent and child (or N-child) have an action with the same shortcut, Qt hits an ambiguous shortcut error and doesn't know which action to trigger so it does nothing.

## How was this PR tested?

Signed-off-by: Chris Galvan <chgalvan@amazon.com>

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/ActionManager/Action/ActionManager.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/ActionManager/Action/ActionManager.cpp
@@ -6,12 +6,104 @@
  *
  */
 
-#include <AzToolsFramework/ActionManager/Action/ActionManager.h>
+#include <QApplication>
+#include <QtGui/private/qguiapplication_p.h>
+#include <QShortcutEvent>
+#include <QTimer>
 
+#include <AzQtComponents/Components/WindowDecorationWrapper.h>
+
+#include <AzToolsFramework/ActionManager/Action/ActionManager.h>
 #include <AzToolsFramework/ActionManager/Menu/MenuManagerInterface.h>
 
 namespace AzToolsFramework
 {
+    //! This class is used to install an event filter on each widget that the action contexts are registered to in order to properly handle
+    //! ambiguous shorcuts.
+    class ActionContextWidgetWatcher : public QObject
+    {
+    public:
+        explicit ActionContextWidgetWatcher(QObject* parent = nullptr)
+            : QObject(parent)
+        {
+        }
+
+        bool eventFilter(QObject* watched, QEvent* event) override
+        {
+            switch (event->type())
+            {
+            case QEvent::Shortcut:
+            {
+                QShortcutEvent* shortcutEvent = static_cast<QShortcutEvent*>(event);
+                if (shortcutEvent->isAmbiguous())
+                {
+                    QWidget* watchedWidget = qobject_cast<QWidget*>(watched);
+                    for (QAction* action : watchedWidget->actions())
+                    {
+                        if (action->shortcut() == shortcutEvent->key())
+                        {
+                            action->trigger();
+                            return true;
+                        }
+                    }
+                }
+                break;
+            }
+            case QEvent::ShortcutOverride:
+            {
+                QKeyEvent* keyEvent = static_cast<QKeyEvent*>(event);
+                int keyInt = keyEvent->key();
+                Qt::KeyboardModifiers modifiers = keyEvent->modifiers();
+                if (modifiers & Qt::ShiftModifier)
+                {
+                    keyInt += Qt::SHIFT;
+                }
+                if (modifiers & Qt::ControlModifier)
+                {
+                    keyInt += Qt::CTRL;
+                }
+                if (modifiers & Qt::AltModifier)
+                {
+                    keyInt += Qt::ALT;
+                }
+                if (modifiers & Qt::MetaModifier)
+                {
+                    keyInt += Qt::META;
+                }
+
+                QKeySequence keySequence(keyInt);
+
+                // FIXME: Originally it looked like we could check this shortcutMap to determine if the ShortcutOverride
+                // was ambiguous, but unfortunately that doesn't seem to be the case :/
+                auto globalShortcutMap = &QGuiApplicationPrivate::instance()->shortcutMap;
+                bool isAmbiguous = globalShortcutMap->hasShortcutForKeySequence(keySequence);
+                if (isAmbiguous)
+                {
+                    QWidget* watchedWidget = qobject_cast<QWidget*>(watched);
+                    for (QAction* action : watchedWidget->actions())
+                    {
+                        if (action->shortcut() == keySequence)
+                        {
+                            // TODO: If we delay the action then we can properly stop the ShortcutOverride from
+                            // continuing on to the parent. For some reason if we trigger it during this flow,
+                            // the parent action's shortcut will get triggered also.
+                            QTimer::singleShot(0, this, [action] {
+                                action->trigger();
+                            });
+
+                            return true;
+                        }
+                    }
+                }
+                break;
+            }
+            default:
+                break;
+            }
+            return false;
+        }
+    };
+
     ActionManager::ActionManager()
     {
         AZ::Interface<ActionManagerInterface>::Register(this);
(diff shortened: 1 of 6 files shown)


Code/Framework/AzToolsFramework/AzToolsFramework/ActionManager/Action/ActionManager.cpp
```diff
@@ -6,12 +6,104 @@
...
+                }
+
+                QKeySequence keySequence(keyInt);
+
+                // FIXME: Originally it looked like we could check this shortcutMap to determine if the ShortcutOverride
+                // was ambiguous, but unfortunately that doesn't seem to be the case :/
+                auto globalShortcutMap = &QGuiApplicationPrivate::instance()->shortcutMap;
+                bool isAmbiguous = globalShortcutMap->hasShortcutForKeySequence(keySequence);
+                if (isAmbiguous)
+                {
```
AMZN-daimini, line 81: If you need this information for actions defined in the Action Manager... 

The Action Manager holds a pointer to all QActions in the corresponding EditorAction internal class instance, so it should be possible to either retrieve this information or create a cache somewhere pretty easily.

The `HotKeyMapping` class in the HotKeyManager also holds a map between shortcut and action identifier. We may have to create a map per Action Context instead of having all in one place like it is right now, but that may also be a simple way to bypass this issue.

Of course it's harder in the current situation where we're also taking actions that are not defined in the Action Manager into account.

AMZN-daimini (commented):
Nice work, I'll get back to this later today and actually build it and play around with it, but for now I only have a few random thoughts that may help :)

hultonha:
Tagging @mbalfour-amzn who might want to check this out too - see this issue for an example - https://github.com/o3de/o3de/issues/14146

mbalfour-amzn:
> Tagging @mbalfour-amzn who might want to check this out too - see this issue for an example - #14146

Yeah, #14146 is related to this, but isn't fixed by it I don't think. The issue is that the component mode overrides are changing from being called during component mode to being called at system startup, so any state-dependent overrides (like Esc in spline vertex editing or entity picking during paintbrush) need some sort of additional mechanism to control when their override kicks in that's separate from component mode lifetime.

hultonha:
> > Tagging @mbalfour-amzn who might want to check this out too - see this issue for an example - #14146
> 
> Yeah, #14146 is related to this, but isn't fixed by it I don't think. The issue is that the component mode overrides are changing from being called during component mode to being called at system startup, so any state-dependent overrides (like Esc in spline vertex editing or entity picking during paintbrush) need some sort of additional mechanism to control when their override kicks in that's separate from component mode lifetime.

Understood, thanks @mbalfour-amzn. This is definitely a really important bit of functionality we need to get working with the new Action Manager. Ideally an action like 'Delete Vertex' should not show (or at least be greyed out) when no vertex is actually selected.

Code/Framework/AzToolsFramework/AzToolsFramework/ActionManager/Action/ActionManager.cpp
cgalvan: Yeah it would be nice to just lookup in our own internal cache, but certainly made trickier with our current environment where we have actions registered outside of the Action Manager. I'll keep digging.

hultonha:
> > > > Tagging @mbalfour-amzn who might want to check this out too - see this issue for an example - #14146
> > > 
> > > 
> > > Yeah, #14146 is related to this, but isn't fixed by it I don't think. The issue is that the component mode overrides are changing from being called during component mode to being called at system startup, so any state-dependent overrides (like Esc in spline vertex editing or entity picking during paintbrush) need some sort of additional mechanism to control when their override kicks in that's separate from component mode lifetime.
> > 
> > 
> > Understood, thanks @mbalfour-amzn. This is definitely a really important bit of functionality we need to get working with the new Action Manager. Ideally an action like 'Delete Vertex' should not show (or at least be greyed out) when no vertex is actually selected.
> 
> Yeah I don't believe my fix will address this issue. The new action manager does have support for controlling when the actions are enabled/disabled though, so it sounds like we just need to use that in a separate PR.

Okay good to know! Thanks for confirming :+1:

Code/Framework/AzToolsFramework/AzToolsFramework/ActionManager/Action/ActionManager.cpp
```diff
@@ -6,12 +6,99 @@
...
+        {
+        }
+
+        bool eventFilter(QObject* watched, QEvent* event) override
+        {
+            switch (event->type())
+            {
+            case QEvent::Shortcut:
+            {
+                QShortcutEvent* shortcutEvent = static_cast<QShortcutEvent*>(event);
```
hultonha, line 36: uber nit: With `static_cast` as the type is already on the right of the expression we can just use `auto` on the left to reduce redundancy of the type info (repeated below)

Code/Framework/AzToolsFramework/AzToolsFramework/ActionManager/Action/ActionManager.cpp
```diff
@@ -6,12 +6,99 @@
...
+                            return true;
+                        }
+                    }
+                }
+                break;
+            }
+            case QEvent::ShortcutOverride:
+            {
+                QKeyEvent* keyEvent = static_cast<QKeyEvent*>(event);
+                int keyInt = keyEvent->key();
```
hultonha, line 54: suggestion/nit: Could this be called `keyIdentifier` perhaps or `keyCode` as opposed to `keyInt`?

hultonha (approved):
@AMZN-daimini is the expert here but the changes look good to me overall. One thing that would be cool though is is it possible to replicate this problem in a test (either integration or automation) and then prove this code handles and fixes the problem? It just feels like having some verification in place would be really useful for a system that is very complex and could be inadvertently broken over time with updates/refactorings. Thanks!

AMZN-daimini:
> > > Tagging @mbalfour-amzn who might want to check this out too - see this issue for an example - #14146
> > 
> > 
> > Yeah, #14146 is related to this, but isn't fixed by it I don't think. The issue is that the component mode overrides are changing from being called during component mode to being called at system startup, so any state-dependent overrides (like Esc in spline vertex editing or entity picking during paintbrush) need some sort of additional mechanism to control when their override kicks in that's separate from component mode lifetime.
> 
> Understood, thanks @mbalfour-amzn. This is definitely a really important bit of functionality we need to get working with the new Action Manager. Ideally an action like 'Delete Vertex' should not show (or at least be greyed out) when no vertex is actually selected.

That specific setup (not showing "Delete Vertex" when unavailable) is already working. I remember hooking things up during migration, but if I missed it in one Component Mode it can definitely be fixed. For other cases, I will look into the possibilities and determine whether the intent can be achieved with what we have or we may need an additional system (shortcut priorities? multiple enabled state callbacks?)

AMZN-daimini (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/ActionManager/Action/ActionManager.cpp
cgalvan: Good call, will fix!

Code/Framework/AzToolsFramework/AzToolsFramework/ActionManager/Action/ActionManager.cpp
cgalvan: I like `keyCode` I'll go with that, thanks!

cgalvan:
> @AMZN-daimini is the expert here but the changes look good to me overall. One thing that would be cool though is is it possible to replicate this problem in a test (either integration or automation) and then prove this code handles and fixes the problem? It just feels like having some verification in place would be really useful for a system that is very complex and could be inadvertently broken over time with updates/refactorings. Thanks!

I should be able to add a test for this! I'll do that now.

Code/Framework/AzToolsFramework/AzToolsFramework/ActionManager/Action/ActionManager.cpp
```diff
@@ -6,12 +6,99 @@
...
+                if (shortcutEvent->isAmbiguous())
+                {
+                    QWidget* watchedWidget = qobject_cast<QWidget*>(watched);
+                    for (QAction* action : watchedWidget->actions())
+                    {
+                        if (action->shortcut() == shortcutEvent->key())
+                        {
+                            action->trigger();
+                            return true;
+                        }
```
AMZN-daimini, line 46: I realized it on a call with @cgalvan, but I'll point it out here too: this code should check if the action is enabled before triggering it :)

Code/Framework/AzToolsFramework/AzToolsFramework/ActionManager/Action/ActionManager.cpp
```diff
@@ -6,12 +6,99 @@
...
+                    {
+                        if (action->shortcut() == keySequence)
+                        {
+                            action->trigger();
+
+                            // We need to accept the event in addition to return true on this event filter
+                            // to ensure the event doesn't get propagated to any parent widgets.
+                            event->accept();
+                            return true;
+                        }
```
AMZN-daimini, line 90: Same as above

AMZN-daimini (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/ActionManager/Action/ActionManager.cpp
cgalvan: Good catch thank you!

hultonha:
> > @AMZN-daimini is the expert here but the changes look good to me overall. One thing that would be cool though is is it possible to replicate this problem in a test (either integration or automation) and then prove this code handles and fixes the problem? It just feels like having some verification in place would be really useful for a system that is very complex and could be inadvertently broken over time with updates/refactorings. Thanks!
> 
> I should be able to add a test for this! I'll do that now.

That would be epic! 😄 Thank you kind sir 🙂

AMZN-daimini:
(Merged for cgalvan to mitigate timezone differences as previously arranged, since I need this work as a base for a different PR :) )

PR #14155, opened 2023-01-20
AttributeInvocable non-void instance call wrapper

### LINKED ISSUE #13830: When using lambdas with Attributes in the EditContext, the lambda signature requires a void* handler

lumberyard-employee-dm:
## What does this PR do?

Added the GetVoidInstanceAttributeInvocable function which returns a smart pointer that either wraps the existing attribute instance in a deleter that is a no-op or if the Attribute is of type `AttributeInvocable` wrap that wraps a functor stored inside of it with another function that can call the original with the first argument replaced with a void*.

It is the responsibility of the caller to verify that correct instance type is passed supplied when using the wrapped attribute with an AttributeReader. This is due to needing to perform static_cast from a `void*` instance to the original first argument type in order to make the function call.

fixes #13830

Fixed the creation of the GenericValueList attribute in the EditContext whenever an EnumValues attribute is reflected.
It now properly replicates an `EnumValues` attribute that stores data in an `AZStd::vector<Edit::EnumConstant<EnumType>` to a GenericValueList attribute that stores the data in an `AZStd::vector<AZStd::pair<EnumType, AZStd::string>>`

This allow all querying of the value, description pairs for a combobox to be serviced through a Unified API.
This is done by just using the `GenericValueList` attribute.
The Legacy ReflectedPropertyEditor still uses `EnumValues` attribute for querying the list of (enum value -> enum description) entries.

## How was this PR tested?

Added a UnitTest to validate that the GetUnsafeAttributeReader was able to call a lambda function that accepted a class type as the first parameter.

### DIFF
--- a/Code/Framework/AzCore/AzCore/RTTI/AttributeReader.h
+++ b/Code/Framework/AzCore/AzCore/RTTI/AttributeReader.h
@@ -165,27 +165,34 @@ namespace AZ
         template <class T, class U, typename... Args>
         bool Read(U& value, Args&&... args)
         {
-            return Internal::AttributeReader<T, U, Args...>::Read(value, m_attribute, m_instancePointer, AZStd::forward<Args>(args)...);
+            return Internal::AttributeReader<T, U, Args...>::Read(value, m_attribute, m_instance, AZStd::forward<Args>(args)...);
         }
 
         template <typename RetType, typename... Args>
         bool Invoke(Args&&... args)
         {
-            return Internal::AttributeInvoke<RetType>(m_attribute, m_instancePointer, AZStd::forward<Args>(args)...);
+            return Internal::AttributeInvoke<RetType>(m_attribute, m_instance, AZStd::forward<Args>(args)...);
         }
 
         // ------------ private implementation ---------------------------------
-        AttributeInvoker(InstType* instancePointer, Attribute* attrib)
-            : m_instancePointer(instancePointer)
+        AttributeInvoker(const InstType& instance, Attribute* attrib)
+            : m_instance(instance)
+            , m_attribute(attrib)
+        {
+        }
+
+        AttributeInvoker(InstType&& instance, Attribute* attrib)
+            : m_instance(AZStd::move(instance))
             , m_attribute(attrib)
         {
         }
 
         Attribute* GetAttribute() const { return m_attribute; }
-        InstType* GetInstancePointer() const { return m_instancePointer; }
+        InstType& GetInstance() { return m_instance; }
+        const InstType& GetInstance() const { return m_instance; }
 
     private:
-        InstType* m_instancePointer;
+        InstType m_instance;
         Attribute* m_attribute;
     };
 
@@ -194,13 +201,13 @@ namespace AZ
     // For Attributes that stores raw or free function such as AZ::AttributeData and AZ::AttributeFunction
     // this class works fine as the void pointer instance isn't used, but for Attributes that works
     class AttributeReader
-        : public AttributeInvoker<void>
+        : public AttributeInvoker<void*>
     {
     public:
         AZ_CLASS_ALLOCATOR(AttributeReader, AZ::SystemAllocator, 0);
 
-        // Bring in AttributeInvoker<void> constructors into scope
-        using AttributeInvoker<void>::AttributeInvoker;
+        // Bring in AttributeInvoker<void*> constructors into scope
+        using AttributeInvoker<void*>::AttributeInvoker;
     };
 
     using AZ::Internal::AttributeRead;
--- a/Code/Framework/AzCore/AzCore/RTTI/AttributeReader.h
+++ b/Code/Framework/AzCore/AzCore/RTTI/AttributeReader.h
@@ -5,8 +5,7 @@
  * SPDX-License-Identifier: Apache-2.0 OR MIT
  *
  */
-#ifndef AZCORE_ATTRIBUTE_READER_H
-#define AZCORE_ATTRIBUTE_READER_H
+#pragma once
 
 #include <AzCore/RTTI/ReflectContext.h>
 #include <AzCore/std/string/string.h>
@@ -36,7 +35,7 @@ namespace AZ
                 value = static_cast<DestType>(invocable->operator()(AZStd::forward<InstType>(instance), AZStd::forward<Args>(args)...));
                 return true;
             }
-            // try a type with an invocable function that takes no arguments
+            // try a type with an invocable function that takes no "this" instance
             if (auto invocable = azrtti_cast<AttributeInvocable<AttrType(Args...)>*>(attr); invocable != nullptr)
             {
                 value = static_cast<DestType>(invocable->operator()(AZStd::forward<Args>(args)...));
@@ -61,7 +60,7 @@ namespace AZ
                 invocable->operator()(AZStd::forward<InstType>(instance), AZStd::forward<Args>(args)...);
                 return true;
             }
-            // try a type with an invocable function that takes no arguments
+            // try a type with an invocable function that takes no "this" instance
             if (auto invocable = azrtti_cast<AttributeInvocable<RetType(Args...)>*>(attr); invocable != nullptr)
             {
                 invocable->operator()(AZStd::forward<Args>(args)...);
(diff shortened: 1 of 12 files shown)


amzn-tmryan (approved):

cgalvan (approved):

Code/Framework/AzCore/Tests/Serialization.cpp
```diff
@@ -5712,6 +5712,39 @@ namespace UnitTest
         EXPECT_FLOAT_EQ(4.0f, value);
     }
 
+    TEST_F(Serialization, AttributeInvocable_UsingVoidPointerInstance_Succeeds)
+    {
+        constexpr AZ::Crc32 invokableCrc = AZ_CRC_CE("Invokable");
```
burelc-amzn, line 5717: `constexpr` variables are still stack variables, there's nothing here that is forcing this to be evaluated in a `constexpr` context (like a `static_assert`). This should be `static constexpr` instead.

Code/Framework/AzCore/AzCore/RTTI/AttributeReader.h
```diff
@@ -165,27 +177,53 @@ namespace AZ
...
-            return Internal::AttributeInvoke<RetType>(m_attribute, m_instancePointer, AZStd::forward<Args>(args)...);
+            using DecayInstType = AZStd::remove_cvref_t<InstType>;
+            if constexpr (AZStd::is_pointer_v<DecayInstType> && AZStd::is_void_v<AZStd::remove_pointer_t<DecayInstType>>)
+            {
+                // Forward the void* as an rvalue so it is not a void*&
+                return Internal::AttributeInvoke<RetType>(m_attribute, static_cast<DecayInstType>(m_instance), AZStd::forward<Args>(args...));
+            }
+            else
+            {
+                return Internal::AttributeInvoke<RetType>(m_attribute, m_instance, AZStd::forward<Args>(args...));
```
burelc-amzn, line 204: These lines have the ellipsis in the wrong place:
```suggestion
                return Internal::AttributeInvoke<RetType>(m_attribute, static_cast<DecayInstType>(m_instance), AZStd::forward<Args>(args)...);
            }
            else
            {
                return Internal::AttributeInvoke<RetType>(m_attribute, m_instance, AZStd::forward<Args>(args)...);
```

burelc-amzn, line 204: These lines have the ellipsis in the wrong place:
```suggestion
                return Internal::AttributeInvoke<RetType>(m_attribute, static_cast<DecayInstType>(m_instance), AZStd::forward<Args>(args)...);
            }
            else
            {
                return Internal::AttributeInvoke<RetType>(m_attribute, m_instance, AZStd::forward<Args>(args)...);
```

Code/Framework/AzCore/Tests/Serialization.cpp
lumberyard-employee-dm: No actually, this variable is being evaluated compile time by MSVC, GCC and Clang due to the `constexpr` keyword, now it is being placed on the stack however, because there is no const.
https://godbolt.org/z/GqjY44KMM

Code/Framework/AzCore/AzCore/RTTI/AttributeReader.h
lumberyard-employee-dm: I'll fix this.

burelc-amzn (approved):

Code/Framework/AzCore/Tests/Serialization.cpp
```diff
@@ -5712,6 +5712,39 @@ namespace UnitTest
         EXPECT_FLOAT_EQ(4.0f, value);
     }
 
+    TEST_F(Serialization, AttributeInvocable_UsingVoidPointerInstance_Succeeds)
+    {
+        static constexpr AZ::Crc32 invokableCrc = AZ_CRC_CE("Invokable");
```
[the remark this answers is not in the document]
spham-amzn: You are both correct! https://wikidiff.com/invokable/invocable

spham-amzn (approved):

spham-amzn:
Is this compilable with gcc/g++ ?

lumberyard-employee-dm:
> Is this compilable with gcc/g++ ?

Yes it is.

Code/Framework/AzCore/AzCore/RTTI/ReflectContext.h
```diff
@@ -132,8 +134,87 @@ namespace AZ
 
         friend OnDemandReflectionOwner;
     };
-
+}
     // Attributes to be used by reflection contexts
+    // Forward declare UnsafeAttributeInvoker here so that it can be used as return value for
+    // GetUnsafeAttributeInvoker function below, but it is implemented in AttributeReader.h to avoid
+    // a circular include
```
amzn-tmryan, line 141: Comment might be stale.

amzn-tmryan (approved):

Code/Framework/AzCore/AzCore/RTTI/AttributeReader.h
```diff
@@ -72,54 +84,54 @@ namespace AZ
             {
                 if constexpr (AZStd::is_integral_v<DestType> && !AZStd::is_same_v<DestType, bool>)
                 {
-                    if (AttributeRead<bool, DestType, Args...>(value, attr, instance, args ...))
+                    if (AttributeRead<bool, DestType, Args...>(value, attr, AZStd::forward<InstType>(instance), args...))
```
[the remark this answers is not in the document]
lumberyard-employee-dm: Arguments should not be forwarded.
The parameters are `const&` here.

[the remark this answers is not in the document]
lumberyard-employee-dm: You know what, I see don't a reason why the arugments here need to be a `const&` reference since only once of the Attribute functions calls will eventually be invoked.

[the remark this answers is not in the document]
lumberyard-employee-dm: Actually I realize the problem now.
The `AttributeFunction` types have an specific signature.
i.e `AttributeFunction<int(void*, AZStd::string)>`

The problem is forwarding the arguments pass the reference type for the arguments to the AttributeRead function, where the cast would fail.
i.e `AttributeFunction<int(void*, AZStd::string&)>` can't cast an `AttributeFunction<int(void*, AZStd::string)>` with the **&** as they are two completely unrelated types

PR #14173, opened 2023-01-23
Enable gems with found implicit gem dependencies

### LINKED ISSUE #14154: New enable-gem functionality does not automatically add dependencies and fails
**Describe the bug**
In the Project Manager, gem dependencies are added for you when you activate a gem, but with the new compatibility checks they are not added.

**Steps to reproduce**
1. In the project manager activate the DccScriptingInterface gem
2. Click Save

**Expected behavior**
The DccScriptingInterface gem is enabled

**Actual behavior**
The gem is not actually active must be force activated from the CLI

**Found in Branch**
development

c4f0c392168b05eea307719c05bee041a90d2024

AMZN-alexpete:

## What does this PR do?

1. Fixes #14154 (enable-gem functionality does not automatically add dependencies and fails)
2. Fixes bad logic in `get_gem_project_incompatible_objects` that was causing unit tests to pass when they shouldn't
3. Adds 2 unit tests that check implicit dependencies
4. Adds infinite recursion protection to `get_incompatible_gem_version_specifiers` and makes it check dependencies recursively

## How was this PR tested?

- Updated unit tests, new unit tests
- Tested adding implicit dependencies (existing, missing and recursive) to misc gem.json files and used --dry-run to see if enable_gem returned correct output.

### DIFF
--- a/scripts/o3de/o3de/compatibility.py
+++ b/scripts/o3de/o3de/compatibility.py
@@ -70,7 +70,7 @@ def get_incompatible_gem_dependencies(gem_json_data:dict, all_gems_json_data:dic
     if not gem_dependencies:
         return set()
 
-    return get_incompatible_gem_version_specifiers(gem_dependencies, all_gems_json_data, set())
+    return get_incompatible_gem_version_specifiers(gem_dependencies, all_gems_json_data, checked_specifiers=set())
 
 
 def get_gem_project_incompatible_objects(gem_json_data:dict, project_path:pathlib.Path, all_gems_json_data:dict = None) -> set:
@@ -170,7 +170,7 @@ def get_incompatible_objects_for_engine(object_json_data:dict, engine_json_data:
     return incompatible_objects
 
 
-def get_incompatible_gem_version_specifiers(gem_version_specifier_list:list, all_gems_json_data:dict, checked_specifiers:set() = None) -> set:
+def get_incompatible_gem_version_specifiers(gem_version_specifier_list:list, all_gems_json_data:dict, checked_specifiers:set) -> set:
     """
     Returns a set of gem version specifiers that are not compatible with the gem's provided
     If a gem_version_specifier_list entry only has a gem name, it is assumed compatible with every gem version with that name.
--- a/scripts/o3de/o3de/compatibility.py
+++ b/scripts/o3de/o3de/compatibility.py
@@ -70,7 +70,7 @@ def get_incompatible_gem_dependencies(gem_json_data:dict, all_gems_json_data:dic
     if not gem_dependencies:
         return set()
 
-    return get_incompatible_gem_version_specifiers(gem_dependencies, all_gems_json_data)
+    return get_incompatible_gem_version_specifiers(gem_dependencies, all_gems_json_data, set())
 
 
 def get_gem_project_incompatible_objects(gem_json_data:dict, project_path:pathlib.Path, all_gems_json_data:dict = None) -> set:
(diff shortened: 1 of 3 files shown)


scripts/o3de/o3de/enable_gem.py
```diff
@@ -120,8 +120,8 @@ def enable_gem_in_project(gem_name: str = None,
             # it takes about 150ms to populate this structure with 137 gems, 4696 bytes in total
             all_gems_json_data = manifest.get_gems_json_data_by_name(project_path=project_path, include_manifest_gems=True, include_engine_gems=True)
 
-            # remove any gems that are not active or dependencies
-            manifest.remove_non_dependency_gem_json_data(enabled_gem_names, all_gems_json_data)
+            # we don't remove gems that are not active or dependencies
+            # because they will be implicitely found and activated via cmake
```
spham-amzn, line 124: nit: This comment may be mis-interpreted to describe the ```incompatible_objects=``` line below it. Suggest making a note: here instead 
```suggestion
            # Note: we don't remove gems that are not active or dependencies
            #       because they will be implicitely found and activated via cmake 
```

scripts/o3de/o3de/compatibility.py
```diff
@@ -70,7 +70,7 @@ def get_incompatible_gem_dependencies(gem_json_data:dict, all_gems_json_data:dic
     if not gem_dependencies:
         return set()
 
-    return get_incompatible_gem_version_specifiers(gem_dependencies, all_gems_json_data)
+    return get_incompatible_gem_version_specifiers(gem_dependencies, all_gems_json_data, set())
```
spham-amzn, line 73: Is the empty ```set()``` argument needed here if ```checked_specifiers``` already defaults to ```None```? You could also do named arguments here as well
```suggestion
    return get_incompatible_gem_version_specifiers(gem_version_specifier_list=gem_dependencies, 
                                        all_gems_json_data=all_gems_json_data)
```

scripts/o3de/o3de/compatibility.py
```diff
@@ -185,21 +185,38 @@ def get_incompatible_gem_version_specifiers(gem_version_specifier_list:list, all
...
     for gem_version_specifier in gem_version_specifier_list:
+        if gem_version_specifier in checked_specifiers:
+            continue
+
+        checked_specifiers.add(gem_version_specifier)
+
         gem_name, version_specifier = utils.get_object_name_and_optional_version_specifier(gem_version_specifier)
         if not gem_name in all_gems_json_data:
             incompatible_gem_version_specifiers.add(f"{gem_version_specifier} (missing dependency)")
             continue
```
lumberyard-employee-dm, line 206: What is the error message around if a dependent gem is missing an dependency.

Is it is something such as "Atom could be activated as one of its dependencies("Atom_Feature_Common") has a missing dependency on "ImGui"?

AMZN-alexpete: Hmmm, it just shows the missing dependency - but it doesn't state what gem it is a dependency of - e.g.
Given:
`Gem A` which depends on `Gem B` which depends on `Gem C`
When:
`Gem C` is missing, you just see the error that `Gem C` is missing, but no that it is a dependency of `Gem B`

AMZN-alexpete: I've updated the output to include gem names. e.g.
GemA is missing the depencency GemB>1.2.3
GemB depends on GemC==2.3.4 but GemC version 3.2.1 was found

lumberyard-employee-dm (approved):

AMZN-Phil (approved):

PR #14178, opened 2023-01-23
Reduce MPS connectivity errors

mbalfour-amzn:
## What does this PR do?

The Multiplayer Sample project was hitting issues trying to send a non-trivial level from the Editor to an Editor server. This PR reduces those errors and makes them more explicit. This area of code would benefit from a larger refactor, but this at least temporarily improves the stability until more far-reaching improvements can be implemented.

There were two specific problems getting hit:
1. Atom recently changed the timing of its initialization (See #14027 ), which meant that its initialization was happening after MultiplayerEditorConnection::ActivateDedicatedEditorServer() was called. The issue with this is that MultiplayerEditorConnection polls for its TCP packets from the main tick loop, so if anything causes an excessive delay on the main loop, it won't process packets as fast. Atom's initialization performs a lot of blocking shader loads, which can take several seconds. 
2. On the Editor side, the MultiplayerEditorSystemComponent blindly sent all the packets for a level to the Editor server at once, without checking for errors. If MultiplayerEditorConnection gets bogged down (which it did), the send buffer would fill up and packets would get dropped. This left the Editor server with gaps in the level data that could eventually lead to serialization errors and crashes.

The fixes are as follows:
1. Added a new "CommandLineProcessed" event that occurs even later in engine initialization than the previously-used "CriticalAssetsCompiled". This once again defers the MultiplayerEditorConnection activation until after the Atom initialization is complete.
2. Added error checking and retries to MultiplayerEditorSystemComponent so that it's a little more tolerant to errors, and will directly assert in the case that the retries are exceeded, so that there's a more obvious cause of failure than serialization errors.

Longer-term, MultiplayerEditorConnection should move its packet processing off the main thread so that it isn't as reliant on the speed of the tick bus, and MultiplayerEditorSystemComponent should either add significantly more robust error-handling or change the level-sending methodology entirely (possibly using a temp file or something) so that it's less prone to failure.

## How was this PR tested?

Ran the Multiplayer Sample project level in the Editor and no longer hit failures due to buffer overflow errors.


Code/Framework/AzNetworking/AzNetworking/TcpTransport/TcpConnection.cpp
```diff
@@ -328,8 +328,16 @@ namespace AzNetworking
...
+            // Ring buffer is full, try sending existing data then reserving the ring buffer memory again.
+            UpdateSend();
+            dstData = reinterpret_cast<uint8_t*>(m_sendRingbuffer.ReserveBlockForWrite(headerSize + payloadSize));
+
+            // Still failed to get ring buffer memory, this packet can't be sent.
+            if (dstData == nullptr)
+            {
+                AZLOG_ERROR("Send ringbuffer full, dropped packet");
+                return false;
+            }
```
cgalvan, line 340: As you pointed out in your description, it'd be nice to have some kind of exponential backoff retry logic here, but this is a good improvement.

cgalvan (approved):

Code/Framework/AzNetworking/AzNetworking/TcpTransport/TcpConnection.cpp
AMZN-puvvadar: I like the idea of force sending to clear the buffer but I don't think this is the place to do it. We're seeing this happen due to an exceptional case of the Editor flooding the connection with level data for a massive level. I think the thing to do here is check failure on `SendReliablePacket` (which the return false here will prop up to) and then call `AZ::Interface<AzNetworking::INetworking>::Get()->ForceUpdate();` to flush.

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
```diff
@@ -410,7 +411,21 @@ namespace Multiplayer
...
+            bool packetSent = false;
+            static constexpr int MaxRetries = 20;
+            static constexpr int MillisecondDelayPerRetry = 100;
+            int numRetries = 0;
+            while (!packetSent && (numRetries < MaxRetries))
+            {
+                packetSent = connection->SendReliablePacket(editorServerLevelDataPacket);
+                if (!packetSent)
+                {
+                    AZStd::this_thread::sleep_for(AZStd::chrono::milliseconds(MillisecondDelayPerRetry));
```
AMZN-puvvadar, line 424: This would be a good spot to add `AZ::Interface<AzNetworking::INetworking>::Get()->ForceUpdate();`. You could check that the send buffer has enough space and go from there. Just be sure to pad that check since SendReliablePacket will have to add extra for the packet header.

AMZN-puvvadar:
So the issue here is less the Editor getting bogged down but more that a TcpConnection in AzNetworking has a max size to its send buffer and MultiplayerEditor doesn't particularly know what to do when the level exceeds that size. The instinct to flush outstanding data is correct. Presently, when we're sending a massive amount of data the caller is responsible for flushing as needed.

I also agree with the changes around initialization timing. I'd definitely like to see those submitted.

Code/Framework/AzNetworking/AzNetworking/TcpTransport/TcpConnection.cpp
mbalfour-amzn: > So the issue here is less the Editor getting bogged down but more that a TcpConnection in AzNetworking has a max size to its send buffer and MultiplayerEditor doesn't particularly know what to do when the level exceeds that size. The instinct to flush outstanding data is correct. Presently, when we're sending a massive amount of data the caller is responsible for flushing as needed.
> 
> I also agree with the changes around initialization timing. I'd definitely like to see those submitted.

It's hard to separate the two concerns with the way the system works right now though. Both the send and the receive buffers have maximum sizes, and a part of why the send buffer is getting filled up is precisely _because_ the receiving side wasn't processing fast enough due to getting bogged down in Atom initialization. Flushing doesn't solve the problem by itself, because all the sends would still fail due to the receive buffer being full.

mbalfour-amzn:
> So the issue here is less the Editor getting bogged down but more that a TcpConnection in AzNetworking has a max size to its send buffer and MultiplayerEditor doesn't particularly know what to do when the level exceeds that size. The instinct to flush outstanding data is correct. Presently, when we're sending a massive amount of data the caller is responsible for flushing as needed.
> 
> I also agree with the changes around initialization timing. I'd definitely like to see those submitted.

It's hard to separate the two concerns with the way the system works right now though. Both the send and the receive buffers have maximum sizes, and a part of why the send buffer is getting filled up is precisely _because_ the receiving side wasn't processing fast enough due to getting bogged down in Atom initialization. Flushing doesn't solve the problem by itself, because all the sends would still fail due to the receive buffer being full.

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
mbalfour-amzn: That's a great idea as a short-term solution, but I still think it needs to get addressed more holistically and at a lower level. Otherwise, everyone who calls SendReliablePacket will need to write the same error-checking and error-handling code on the off chance that their call is the one that causes the buffer to overflow. Perhaps we need an API like "ReliablySendReliablePacket". :P

Code/Framework/AzNetworking/AzNetworking/TcpTransport/TcpConnection.cpp
mbalfour-amzn: I agree that the calling code needs to check for errors and handle them, but I disagree that it's _fully_ the responsibility of the caller. This particular problem is caused by the choice of using a ringbuffer implementation inside of TcpConnection instead of infinitely-growing buffers, shared-ptr management, or other strategies, so it seems user-unfriendly to push that knowledge outside of the API to force every caller to deal with that choice.

That said, I can accept that the extra call to UpdateSend() here isn't particularly helpful, since in the case that the send buffer is full because the receiver isn't listening, calling it just one more time instead of inside of some sort of a loop likely isn't going to help. I'll remove it from here, but I think the overall error-handling strategy should be improved in a way that isn't burdensome on every caller.

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
mbalfour-amzn: Side note - there's no obvious way to check the send buffer size without exposing more implementation details into the interface, but it gets caught indirectly anyways through the failure in SendReliablePacket.

Registry/application_lifecycle_events.setreg
```diff
@@ -24,7 +24,8 @@
                 "FileIOAvailable": {},
                 "FileIOUnavailable": {},
                 "LegacySystemInterfaceCreated": {},
-                "CriticalAssetsCompiled": {}
+                "CriticalAssetsCompiled": {},
+                "CommandLineProcessed": {}
```
lumberyard-employee-dm, line 28: I would claim this should be called "LegacyCommandLine" processed, since it is sent after CrySystem has processed the Command Line.
The ComponentApplication processes the command line in it's [constructor](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp#L407-L408)

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
AMZN-puvvadar: Agreed, it's worth a longer term rethink.

Code/Framework/AzNetworking/AzNetworking/TcpTransport/TcpConnection.cpp
AMZN-puvvadar: Yeah I agree, there needs to be a bit of a rethink here systemically. Definitely we should get a longer term issue cut to sig-network to rethink both the memory scheme and the error reporting mechanisms here. 

> > So the issue here is less the Editor getting bogged down but more that a TcpConnection in AzNetworking has a max size to its send buffer and MultiplayerEditor doesn't particularly know what to do when the level exceeds that size. The instinct to flush outstanding data is correct. Presently, when we're sending a massive amount of data the caller is responsible for flushing as needed.
> > I also agree with the changes around initialization timing. I'd definitely like to see those submitted.
> 
> It's hard to separate the two concerns with the way the system works right now though. Both the send and the receive buffers have maximum sizes, and a part of why the send buffer is getting filled up is precisely _because_ the receiving side wasn't processing fast enough due to getting bogged down in Atom initialization. Flushing doesn't solve the problem by itself, because all the sends would still fail due to the receive buffer being full.

The Receive side is interesting. I did some testing and it seems to never fill with your changes. It also seems to be the case that UpdateRecv reads a maximum of the maximum packet size off the buffer before proceeding to deserialize and flush all data on the receive buffer. The whole thing need re-evaluation.

mbalfour-amzn: If you want to see it fill, change "LegacyCommandLineProcessed" back to "CriticalAssetsCompiled" in MultiplayerEditorConnection. This will cause it to once again announce that the server is ready to receieve packets _before_ Atom initializes, so it will start receiving packets just before getting several seconds of blocking loads that prevent it from processing packets.

AMZN-puvvadar (approved):
These changes seem to get us past the issue at hand. I'll go ahead and open a task on sig-network related to the issues around Send/Recv.

Issue #14121, opened 2023-01-19
Material Canvas crashes after adding Standard PBR or Base PBR nodes to graph


LB-KacperMetelski:
**Describe the bug**
Adding the Standard PBR or Base PBR nodes to the Material Graph Document freezes the Canvas, which afterwards leads to the crash upon any interaction within the Material Canvas.

After this issue occurs, the untitled.material file is being saved on the disk and searching it (i.e. in Asset Browser) causes the Editor to crash.

**Steps to reproduce**
1. Launch the Material Canvas.
2. Add a Standard or Base PBR Node.
3. Click anywhere on the Canvas repeatedly.

**Expected behavior**
All of the nodes are added correctly without causing any stability issues.

**Actual behavior**
The Material Canvas crashes.

**Screenshots/Video**
[2023-01-19 12-11-40.mp4
](https://user-images.githubusercontent.com/110163473/213465990-ab727efe-96a1-4158-8b83-55a9d38619c2.mp4)

**Found in Branch**
Development

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
c601866

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Windows
 - Version: 11 22H2
 - CPU: I7-11700f
 - GPU: Geforce RTX3070 MSI
 - Memory: 32GB
 
 **Additional context**
[MaterialCanvas.log](https://github.com/o3de/o3de/files/10457642/MaterialCanvas.log)

LB-JakubSkoczylas:
Hi, I was able to add both PBR nodes and no crash occurred on Development (ecf542f). After adding the PBR node, my PC froze for a few seconds. Without clicking anything, the AP processed the new assets and I was able to work with the Material Canvas. However after adding Base PBR node to the canvas that already had the Standard PBR, a new assets were created and failed in the Asset Processor. I am not sure if these fails are somehow connected to the freeze and crash, so would you like me to open a new ticket for this issue?

**My Device:**
Device: PC
OS: Linux
Version Ubuntu 20.04.3 LTS
CPU AMD Ryzen 5 3600
GPU Nvidia Geforce RTX 2060 SUPER
Memory 16GB

Failed asset logs: 
[untitled_6_DepthPass_WithPS.shader-295053385-36668.log](https://github.com/o3de/o3de/files/10466474/untitled_6_DepthPass_WithPS.shader-295053385-36668.log)
[...]

PR #14180, opened 2023-01-24
Material Canvas: Fixing base and standard PBR material templates

gadams3:
## What does this PR do?

This change fixes and cleans up all of the shaders provided with the current material canvas templates. When they were last submitted with support for transparency, everything appeared to work correctly on the tested pipelines but the depth shader was spamming warnings and asserts about requiring an SRG that was not bound. This turned out to be the pass SRG, which is not used in the depth shader code but was referenced in the included, common lighting code. The resulting log spam eventually caused a significant frame rate drop and potential crash.

Resolves https://github.com/o3de/o3de/issues/14121 and similar issues

Created material graphs using both templates.
Configured inputs upon created graphs to test transparency options and alpha.
Tested generated materials on all available pipelines through material canvas and material editor.
Confirmed that there were no errors reported by the AP for the templates or the generated assets.
Confirmed that spam no longer appeared or caused the drop in frame rate in material canvas and material edito.

### DIFF
--- a/Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/GraphData/Nodes/MaterialOutputs/BasePBR/MaterialGraphName_DepthPass.shader
+++ b/Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/GraphData/Nodes/MaterialOutputs/BasePBR/MaterialGraphName_DepthPass.shader
@@ -12,10 +12,6 @@
         {
           "name": "MainVS",
           "type": "Vertex"
-        },
-        {
-          "name": "MainPS",
-          "type": "Fragment"
         }
       ]
     },
--- a/Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/GraphData/Nodes/MaterialOutputs/BasePBR/MaterialGraphName_LowEndForward.shader
+++ b/Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/GraphData/Nodes/MaterialOutputs/BasePBR/MaterialGraphName_LowEndForward.shader
@@ -1,10 +1,4 @@
 {
-    // Note: "LowEnd" shaders are for supporting the low end pipeline
-    // These shaders can be safely added to materials without incurring additional runtime draw
-    // items as draw items for shaders are only created if the scene has a pass with a matching
-    // DrawListTag. If your pipeline doesn't have a "lowEndForward" DrawListTag, no draw items
-    // for this shader will be created.
-
     "Source" : "./MaterialGraphName_Forward.azsl",
 
     "Definitions": ["QUALITY_LOW_END_TIER1=1", "LOW_END_FORWARD=1"],
--- a/Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/GraphData/Nodes/MaterialOutputs/StandardPBR/MaterialGraphName_SurfaceEval.azsli
+++ b/Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/GraphData/Nodes/MaterialOutputs/StandardPBR/MaterialGraphName_SurfaceEval.azsli
@@ -79,7 +79,7 @@ Surface EvaluateSurface_MaterialGraphName(
     surface.specularOcclusion = inSpecularOcclusion;
     surface.opacityAffectsSpecularFactor = o_opacity_affects_specular_factor;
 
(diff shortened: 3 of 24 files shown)


galibzon (approved):

Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/GraphData/Nodes/MaterialOutputs/BasePBR/MaterialGraphName_LowEndForward.shader
```diff
@@ -1,10 +1,4 @@
 {
-    // Note: "LowEnd" shaders are for supporting the low end pipeline
-    // These shaders can be safely added to materials without incurring additional runtime draw
-    // items as draw items for shaders are only created if the scene has a pass with a matching
-    // DrawListTag. If your pipeline doesn't have a "lowEndForward" DrawListTag, no draw items
-    // for this shader will be created.
-
```
rgba16f, line 7: why remove this comment?

Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/GraphData/Nodes/MaterialOutputs/BasePBR/MaterialGraphName_DepthPass.shader
```diff
@@ -12,10 +12,6 @@
         {
           "name": "MainVS",
           "type": "Vertex"
-        },
-        {
-          "name": "MainPS",
-          "type": "Fragment"
```
rgba16f, line 18: Why remove this?

Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/GraphData/Nodes/MaterialOutputs/BasePBR/MaterialGraphName_LowEndForward.shader
gadams3: The comments were copied from and still exist in the original files for each of these material types and will change once the material pipeline branch is integrated.

Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/GraphData/Nodes/MaterialOutputs/BasePBR/MaterialGraphName_DepthPass.shader
gadams3: Because this version of the depth shader does not use the pixel shader. There is a separate depth and shadow shader, which uses the same source file and includes the pixel shader.

Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/GraphData/Nodes/MaterialOutputs/StandardPBR/MaterialGraphName_SurfaceEval.azsli
```diff
@@ -79,7 +79,7 @@ Surface EvaluateSurface_MaterialGraphName(
     surface.specularOcclusion = inSpecularOcclusion;
     surface.opacityAffectsSpecularFactor = o_opacity_affects_specular_factor;
 
-    CheckClipping(surface.alpha, 0.0);
+    CheckClipping(surface.alpha, 0.5);
```
moudgils, line 82: Should this be hardcoded like this? Does this value need to match the depth pass?

rgba16f (approved):

Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/GraphData/Nodes/MaterialOutputs/StandardPBR/MaterialGraphName_SurfaceEval.azsli
gadams3: For now, this PR is resolving a regression. The value will be configurable after I surface it on the material canvas node for this type. There is an existing task to rework opacity factor. In the hardcoded material types, the value for opacity factor came from a fixed material SRG, with a default value of 0.5. It was used inside GetAlphaAndClip. I will be noting and exposing more parameters as I come across them. I want to do it in such a way where I'm not littering all of the code with material canvas code generation blocks, which I will also be simplifying.

PR #14182, opened 2023-01-24
Fixing potential crash that can happen with cached shadows

invertednormal:
## What does this PR do?

Previously, `ProjectedShadowmapsPass` expected that `FrameBeginInternal()` will always be called before `BuildInternal()` however this is not always true. That caused `m_clearShadowDrawPacket` to sometimes not be initialized when it was needed. This PR changes it so that we no longer rely on `FrameBeginInternal()` and simply make sure the draw packet is built if needed right before its used.

In addition, there seem to be circumstances where `BuildInternal()` can be called and the children should update even if `m_updateChildren` isn't set to true. This PR changes it so the children always rebuild if `BuildInternal()` is called, which forces cached shadows to re-draw.

There is still a bug where it's possible for cached shadows to become corrupted, at least in editor mode. It seems to be related to the Editor Mode selection overlay injecting draw packets into the pipeline. I'm still investigating that, but reasoned it was best to get a fix for the crash bug in immediately while I search for the rarer and less fatal shadow corruption bug.

## How was this PR tested?

Tested using the multi camera project from ROSCon

### DIFF
--- a/Gems/Atom/Feature/Common/Code/Source/CoreLights/ProjectedShadowmapsPass.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/CoreLights/ProjectedShadowmapsPass.cpp
@@ -164,6 +164,12 @@ namespace AZ
                 }
             }
 
+            if (!m_clearShadowDrawPacket)
+            {
+                CreateClearShadowDrawPacket();
+            }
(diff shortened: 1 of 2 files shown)


rgba16f (approved):

gadams3 (approved):

hultonha (approved):
LGTM! And thanks for the thorough description too 👍

cgalvan (approved):

amzn-tommy (approved):

PR #14183, opened 2023-01-24
Fixes an issue where the engine cmake version wasn't pulled from engine.json

AMZN-alexpete:

## What does this PR do?

The `O3DE_ENGINE_VERSION_STRING` and friends were ignoring the values in `engine.json`.  This PR updates `Version.cmake` to look in the `engine.json` file for version, build and name values to use if no environment var overrides are set.

This also sets `O3DE_VERSION_MAJOR`, `O3DE_VERSION_MINOR` and `O3DE_VERSION_PATCH` which are currently unused but part of the [engine versioning RFC](https://github.com/o3de/sig-core/issues/44).  api_versions will be added in a later PR

**IMPORTANT:  This changes the output name of the installer file to be `o3de_X.Y.Z_installer.exe` for development installers - it looks like we name our nightly installer `o3de_installer.exe` so it should not break any download links, but worth checking as I'm not 100% of that logic.**  Based on[ this CMake code](https://github.com/o3de/o3de/blob/c577d3fd6534a1a222a6746f40b7e17f17136513/cmake/Platform/Common/PackagingPostBuild_common.cmake#L91), it looks like the version is stripped automatically from the latest filename 

## How was this PR tested?

Configured, created an installer and ran it, verifying the path now uses the engine version in engine.json by default:
image

Verified the path for the installer uses the engine_display_version if not "00.00".


cgalvan (approved):

cmake/Version.cmake
```diff
@@ -6,22 +6,62 @@
...
-
-if(NOT "$ENV{O3DE_DISPLAY_VERSION}" STREQUAL "")
-    set(O3DE_DISPLAY_VERSION_STRING "$ENV{O3DE_DISPLAY_VERSION}")
-endif()
-
-if(NOT "$ENV{O3DE_BUILD_VERSION}" STREQUAL "")
-    set(O3DE_BUILD_VERSION "$ENV{O3DE_BUILD_VERSION}")
-endif()
+
+get_filename_component(ENGINE_JSON_PATH "${CMAKE_CURRENT_LIST_DIR}/../engine.json" ABSOLUTE)
```
lumberyard-employee-dm, line 14: It would be better to use `${LY_ROOT_FOLDER}/engine.json` here to locate the engine.json

cmake/Version.cmake
```diff
@@ -6,22 +6,62 @@
...
+# \arg:output_value - name of output variable to set 
+# \arg:key - name of field in engine.json 
+# \arg:env_value - name of environment variable to use as override 
+# \arg:default_value - value to use if neither environment var or engine.json value found 
+function(o3de_read_engine_default output_value env_value key default_value description)
+    # use Environment override if provided
+    if(NOT ${env_value} STREQUAL "")
+        set(${output_value} ${env_value} CACHE STRING ${description})
+        return()
+    endif()
```
lumberyard-employee-dm, line 32: Checking that the ENV variable isn't an empty string is fine for the version fields later on in this file.
But if there is ever a time when the environment variable can exist, but be empty, then it would be better to use the cmake [if(DEFINED ENV{var})](https://cmake.org/cmake/help/latest/command/if.html?highlight=#existence-checks) check

cmake/Version.cmake
```diff
@@ -6,22 +6,62 @@
...
+
+    # pull the value from engine.json if available
+    string(JSON value ERROR_VARIABLE manifest_json_error GET ${manifest_json_data} ${key})
+    if(manifest_json_error)
+        message(WARNING "Failed to read ${key} from \"${ENGINE_JSON_PATH}\" : ${manifest_json_error}")
```
lumberyard-employee-dm, line 37: This could use the [o3de_read_json_key](https://github.com/o3de/o3de/blob/development/cmake/O3DEJson.cmake#L51-L59) function which will also output the warning

cmake/Version.cmake
AMZN-alexpete: I was using that originally but noticed it gets set later on `GeneralSettings.cmake` and wasn't sure if we wanted to include that so early or if that might mess up project based vs engine based setups.  I'll take a closer look at that and do some testing.

cmake/Version.cmake
AMZN-alexpete: I was using that originally, but that would open and parse the `engine.json` for every field we check and I wanted to keep this code fast because it gets run with ever cmake op.

lumberyard-employee-dm: Then opening the engine.json file outside of any function with the name `manifest_json_data` is problem then as that is a common variable name that could be used to overwrite the engine json data.
If anything the engine.json data that is read could be stored in a cmake GLOBAL property. so it could be reused later

lumberyard-employee-dm (approved):

AMZN-Phil (approved):

amzn-changml (approved):
Seems reasonable. As long as env vars are prioritized as you've done, this shouldn't affect CD workflows (pending testing)

AMZN-alexpete:
The latest change adds an `INSTALL_` prefix to the CMake vars that can be set by the user and if the user does not set that value, it will default to the engine version e.g. `O3DE_INSTALL_VERSION_STRING` defaults to `O3DE_VERSION_STRING`.  This change was made so we can always pick up changes made in `engine.json` unless the user (or Jenkins) specifically sets the value to use for the INSTALL target.

cmake/Version.cmake
```diff
@@ -8,20 +8,100 @@
...
+#! o3de_read_engine_default: Read a field from engine.json or use the default if not found
+# \arg:output_value - name of output variable to set 
+# \arg:key - name of field in engine.json 
+# \arg:default_value - value to use if neither environment var or engine.json value found 
+macro(o3de_read_engine_default output_value key default_value)
+    get_property(engine_json_data GLOBAL PROPERTY O3DE_ENGINE_JSON_DATA)
+    string(JSON tmp_value ERROR_VARIABLE manifest_json_error GET ${engine_json_data} ${key})
+
+    # unset engine_json_data because we're in a macro
+    unset(${engine_json_data})
```
lumberyard-employee-dm, line 26: This shouldn't use curly braces here as the the `engine_json_data` is still the name of the CMake variable, even if it is in macro.
The difference is there is no scoping within he macro so engine_json_data is available to caller
```suggestion
    # unset engine_json_data because we're in a macro
    unset($engine_json_data)
```

cmake/Version.cmake
```diff
@@ -8,20 +8,100 @@
...
+    if(manifest_json_error)
+        message(WARNING "Failed to read ${key} from \"${LY_ROOT_FOLDER}/engine.json\" : ${manifest_json_error}")
+        set(${tmp_value} ${default_value})
+    endif()
+
+    ly_set(${output_value} ${tmp_value})
+    set_property(GLOBAL PROPERTY ${output_value} ${tmp_value})
+
+    # unset tmp_value because we're in a macro
+    unset(${tmp_value})
```
lumberyard-employee-dm, line 37: `tmp_value` is the name of the variable and therefore it needs to be supplied to unset.
```suggestion
    # unset tmp_value because we're in a macro
    unset(tmp_value)
```

cmake/Version.cmake
```diff
@@ -8,20 +8,100 @@
...
+macro(o3de_read_engine_default output_value key default_value)
+    get_property(engine_json_data GLOBAL PROPERTY O3DE_ENGINE_JSON_DATA)
+    string(JSON tmp_value ERROR_VARIABLE manifest_json_error GET ${engine_json_data} ${key})
+
+    # unset engine_json_data because we're in a macro
+    unset(${engine_json_data})
+
+    if(manifest_json_error)
+        message(WARNING "Failed to read ${key} from \"${LY_ROOT_FOLDER}/engine.json\" : ${manifest_json_error}")
+        set(${tmp_value} ${default_value})
```
lumberyard-employee-dm, line 30: repeat: This should reference tmp_value which is the name of the variable, not `${tmp_value}` which is the value of the variable.
```suggestion
        set(tmp_value ${default_value})
```

cmake/Version.cmake
```diff
@@ -8,20 +8,100 @@
...
+    endif()
+
+    ly_set(${output_value} ${tmp_value})
+    set_property(GLOBAL PROPERTY ${output_value} ${tmp_value})
```
lumberyard-employee-dm, line 34: This logic right here... **Is perfectly fine**.

The `output_value` variable stores the name of the variable supplied for that macro parameter.
Therefore `${output_value}` is name of the variable that was passed in.
FYI using `${${output_value}}` would retrieve the referenced name variable value.

cmake/Version.cmake
```diff
@@ -8,20 +8,100 @@
...
+# \arg:output_value - name of output variable to set 
+# \arg:env_var - name of environment variable to use 
+# \arg:default_value - value to use if environment var is empty
+macro(o3de_set_install_default output_value env_var default_value description)
+    set(${output_value} ${default_value} CACHE STRING ${description})
+    if(NOT ${env_var} STREQUAL "")
+        set(${output_value} ${env_var})
+    endif()
+    string(CONFIGURE ${${output_value}} ${output_value} @ONLY)
+endmacro()
```
lumberyard-employee-dm, line 51: This would be a good to move to `LySet.cmake` since this is general purpose function.
I don't have a short name that explains all what the function does, but how about
`o3de_set_in_env_or_cache_and_substitute_var_refs`

AMZN-alexpete: would you recommend I use `ly_set` instead of `set` on line 48?

lumberyard-employee-dm: Not really..., Because that would result in different expectations which scope the variable is set in.
For example this function being called in the Version.cmake file would result in the values being set in the engine(`<engine-root>`) and project(`<project-root>`) scopes in a project-centric workflow and only the engine scope(`<engine-root>` in an engine-centric flow.

What this means is that in an engine-centric with the path to the AutomatedTesting project set via `LY_PROJECTS=./AutomatedTesting`, the AutomatedTesting project will not have access to the variable set by this function.
However if the AutomatedTesting project was run in a project workflow(i.e you run CMake configure from inside the AutomatedTestingProject, then it would have access to that variable).

To avoid that inconsistency, I think it is better to just set the value in the current scope and then let the caller deal with setting it in any other scopes.

lumberyard-employee-dm (approved):

lumberyard-employee-dm (approved):

PR #14184, opened 2023-01-24
Preserve entity order in newly created prefab

### LINKED ISSUE #12603: Bug Report - Create prefab does not preserve the order of entities and nested prefabs
**Describe the bug**

Create Prefab operation does not preserve the order of top level (i.e. direct) entities and nested prefabs. The new order will always be entities first and then nested prefabs. See the example below about what top level means.

```
// Before
Level
|-- Wheel.prefab        (top level container entity)
|-- Engine              (top level entity)
    |-- Dummy

// Create a Car prefab
Level
|-- Car.prefab
    |-- Wheel.prefab
    |-- Engine
        |-- Dummy
```

**Steps to reproduce**
1. Open an empty level
2. Create a Wheel entity and create a Wheel prefab from it
3. Create an Engine entity
4. Create a Car prefab out of Wheel prefab and Engine entity
5. Save the prefab
6. Double click the Car prefab to enter edit mode to see its children

**Expected behavior**

```
Level
|-- Car.prefab
    |-- Wheel.prefab
    |-- Engine
```

**Actual behavior**

```
Level
|-- Car.prefab
    |-- Engine
    |-- Wheel.prefab
```

**Screenshots/Video**

Will add if needed

**Found in Branch**

Development

* Device: PC
* OS: Windows 10
* CPU: AMD Ryzen 5950x
* GPU: NVIDIA RTX 3070
* Memory: 64 GB

michabr:

## What does this PR do?

Rearrange the "Create prefab" logic a bit so the original entity order remains intact when a prefab is created out of those entities. Before this PR, top-level container entities always appeared before regular entities.

Fixes #12603
 
## How was this PR tested?

Tested different hierarchies with a mix of entities and nested instances in Editor.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
@@ -260,7 +260,7 @@ namespace AzToolsFramework
                     newContainerEntityDomInitialStateWithTransform, *newContainerEntity);
 
                 // Helper function to create a link between a nested instance and the newly created instance
-                auto createLinkForNestedInstanceFunc = [&](Instance& nestedInstance)
+                auto createLinkForNestedInstance = [&](Instance& nestedInstance)
                 {
                     EntityOptionalReference nestedInstanceContainerEntity = nestedInstance.GetContainerEntity();
                     AZ_Assert(
@@ -294,7 +294,7 @@ namespace AzToolsFramework
                         AZ_Assert(topLevelInstance.has_value(), "Invalid nested instance found in the new prefab created.");
 
                         // Create a link for the nested instance
-                        createLinkForNestedInstanceFunc(topLevelInstance->get());
+                        createLinkForNestedInstance(topLevelInstance->get());
 
                         // Update the instance link with the new parent transform data
                         PrefabDom nestedContainerEntityDomBefore;
@@ -333,7 +333,7 @@ namespace AzToolsFramework
                 EditorEntityContextNotificationBus::Broadcast(&EditorEntityContextNotification::SetForceAddEntitiesToBackFlag, false);
 
                 // Set up remaining links for the nested instances that are not at the top level. This is done without undo/redo support.
-                instanceToCreate->get().GetNestedInstances([topLevelEntities, createLinkForNestedInstanceFunc](AZStd::unique_ptr<Instance>& nestedInstance) {
+                instanceToCreate->get().GetNestedInstances([topLevelEntities, createLinkForNestedInstance](AZStd::unique_ptr<Instance>& nestedInstance) {
                     AZ_Assert(nestedInstance, "Invalid nested instance found in the new prefab created.");
 
                     EntityOptionalReference nestedInstanceContainerEntity = nestedInstance->GetContainerEntity();
@@ -343,7 +343,7 @@ namespace AzToolsFramework
                     // Create a link if this container entity is not a top-level entity. Otherwise, the link has already been created
                     if (AZStd::find(topLevelEntities.begin(), topLevelEntities.end(), &nestedInstanceContainerEntity->get()) == topLevelEntities.end())
                     {
-                        createLinkForNestedInstanceFunc(*nestedInstance);
+                        createLinkForNestedInstance(*nestedInstance);
                     }
                 });
 
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
@@ -270,9 +270,10 @@ namespace AzToolsFramework
                     PrefabDom previousPatch;
 
                     // Retrieve the previous patch if it exists
-                    if (nestedInstanceLinkPatchesMap.contains(&nestedInstance))
+                    if (auto nestedInstanceIt = nestedInstanceLinkPatchesMap.find(&nestedInstance);
+                        nestedInstanceIt != nestedInstanceLinkPatchesMap.end())
                     {
-                        previousPatch.Swap(nestedInstanceLinkPatchesMap[&nestedInstance]);
+                        previousPatch.Swap(nestedInstanceIt->second);
                         UpdateLinkPatchesWithNewEntityAliases(previousPatch, oldEntityAliases, instanceToCreate->get());
                     }
 
@@ -289,12 +290,12 @@ namespace AzToolsFramework
 
                     if (IsInstanceContainerEntity(topLevelEntityId))
                     {
-                        InstanceOptionalReference nestedInstance = m_instanceEntityMapperInterface->FindOwningInstance(topLevelEntityId);
+                        InstanceOptionalReference topLevelInstance = m_instanceEntityMapperInterface->FindOwningInstance(topLevelEntityId);
 
-                        AZ_Assert(nestedInstance.has_value(), "Invalid nested instance found in the new prefab created.");
+                        AZ_Assert(topLevelInstance.has_value(), "Invalid nested instance found in the new prefab created.");
 
                         // Create a link for the nested instance
-                        createLinkForNestedInstanceFunc(nestedInstance->get());
+                        createLinkForNestedInstanceFunc(topLevelInstance->get());
 
                         // Update the instance link with the new parent transform data
                         PrefabDom nestedContainerEntityDomBefore;
@@ -313,7 +314,7 @@ namespace AzToolsFramework
                         // We won't parent this undo node to the undo batch so that the newly created template and link will remain
                         // unaffected by undo actions. This is needed so that any future instantiations of the template will work.
                         PrefabUndoUpdateLink linkUpdate = PrefabUndoUpdateLink(AZStd::to_string(static_cast<AZ::u64>(topLevelEntityId)));
-                        linkUpdate.Capture(reparentPatch, nestedInstance->get().GetLinkId());
+                        linkUpdate.Capture(reparentPatch, topLevelInstance->get().GetLinkId());
                         linkUpdate.Redo();
                     }
                     else
@@ -332,8 +333,8 @@ namespace AzToolsFramework
                 }
                 EditorEntityContextNotificationBus::Broadcast(&EditorEntityContextNotification::SetForceAddEntitiesToBackFlag, false);
 
-                // Set up remaining links between the created prefab instance and its nested instances without undo/redo support.
-                instanceToCreate->get().GetNestedInstances([&](AZStd::unique_ptr<Instance>& nestedInstance) {
+                // Set up remaining links for the nested instances that are not at the top level. This is done without undo/redo support.
+                instanceToCreate->get().GetNestedInstances([topLevelEntities, createLinkForNestedInstanceFunc](AZStd::unique_ptr<Instance>& nestedInstance) {
                     AZ_Assert(nestedInstance, "Invalid nested instance found in the new prefab created.");
 
                     EntityOptionalReference nestedInstanceContainerEntity = nestedInstance->GetContainerEntity();
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
@@ -259,34 +259,10 @@ namespace AzToolsFramework
                 m_instanceToTemplateInterface->GenerateEntityDomBySerializing(
                     newContainerEntityDomInitialStateWithTransform, *newContainerEntity);
 
-                // Parent the non-container top level entities to the new container entity.
-                // It also patches the entity DOMs to the new template without undo/redo support.
-                EditorEntityContextNotificationBus::Broadcast(&EditorEntityContextNotification::SetForceAddEntitiesToBackFlag, true);
-                for (AZ::Entity* topLevelEntity : topLevelEntities)
+                // Helper function to create a link between a nested instance and the newly created instance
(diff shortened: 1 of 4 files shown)


hultonha:
I know I'm not a code owner here so you can disregard this but I'd strongly recommend/advise looking into the possibility of adding a test for this to verify the functionality now (and more importantly in future). This feels pretty fundamental and we want to know if anything changes that could break this. I realize it's not always easy but remember the time we spend writing tests (and making writing tests easier) is an investment that pays off massively in future when we don't have these horrible regressions we than have to spend days hunting down 🙂

srikappa-amzn:
> I know I'm not a code owner here so you can disregard this but I'd strongly recommend/advise looking into the possibility of adding a test for this to verify the functionality now (and more importantly in future). This feels pretty fundamental and we want to know if anything changes that could break this. I realize it's not always easy but remember the time we spend writing tests (and making writing tests easier) is an investment that pays off massively in future when we don't have these horrible regressions we than have to spend days hunting down 🙂

I agree. A test would be great here to catch regressions in the future. Particularly because the CreatePrefab code has become so big since it was first written and it continues to grow.

Code/Framework/AzToolsFramework/Tests/Prefab/PrefabCreateInMemoryTests.cpp
```diff
@@ -0,0 +1,61 @@
...
+ *
+ */
+
+#include <AzToolsFramework/Entity/EditorEntityHelpers.h>
+
+#include <Prefab/PrefabTestFixture.h>
+
+namespace UnitTest
+{
+    using PrefabCreateInMemoryTest = PrefabTestFixture;
```
forkercat, line 15: Should it be PrefabCreateTests?

Would like to have a follow-up task to put all the tests for public API under "public" folder and all undo tests under "undo" folder. 😄

Code/Framework/AzToolsFramework/Tests/Prefab/PrefabCreateInMemoryTests.cpp
```diff
@@ -0,0 +1,61 @@
...
+        auto carInstance = m_instanceEntityMapperInterface->FindOwningInstance(carContainerId);
+        ASSERT_TRUE(carInstance.has_value());
+        ASSERT_TRUE(carInstance->get().GetContainerEntityId() == carContainerId);
+
+        // Validate the child entity order of the car instance
+        AzToolsFramework::EntityOrderArray entityOrderArray = AzToolsFramework::GetEntityChildOrder(carContainerId);
+        EXPECT_TRUE(entityOrderArray.size() == 3);
+        AZStd::string entityName;
+        AZ::ComponentApplicationBus::BroadcastResult(entityName, &AZ::ComponentApplicationRequests::GetEntityName, entityOrderArray[0]);
+        EXPECT_TRUE(entityName == engineEntityName);
```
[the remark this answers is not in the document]
forkercat: Good to know!

Code/Framework/AzToolsFramework/Tests/Prefab/PrefabCreateInMemoryTests.cpp
```diff
@@ -0,0 +1,61 @@
...
+        // Create the Wheel instance after the Engine entity
+        AZ::EntityId tireEntityId = CreateEditorEntityUnderRoot(tireEntityName);
+        AZ::EntityId wheelContainerId = CreateEditorPrefab(wheelPrefabFilepath, { tireEntityId });
+        // Create the Battery entity after the Wheel instance
+        AZ::EntityId batteryEntityId = CreateEditorEntityUnderRoot(batteryEntityName);
+        // Create a Car prefab from the entities and isntance
+        AZ::EntityId carContainerId = CreateEditorPrefab(carPrefabFilepath, { engineEntityId, batteryEntityId, wheelContainerId });
+        auto carInstance = m_instanceEntityMapperInterface->FindOwningInstance(carContainerId);
+        ASSERT_TRUE(carInstance.has_value());
+        ASSERT_TRUE(carInstance->get().GetContainerEntityId() == carContainerId);
```
forkercat, line 48: nit: can do `ASSERT_EQ`

Code/Framework/AzToolsFramework/Tests/Prefab/PrefabCreateInMemoryTests.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+        // Validate the child entity order of the car instance
+        AzToolsFramework::EntityOrderArray entityOrderArray = AzToolsFramework::GetEntityChildOrder(carContainerId);
+        EXPECT_TRUE(entityOrderArray.size() == 3);
+        AZStd::string entityName;
+        AZ::ComponentApplicationBus::BroadcastResult(entityName, &AZ::ComponentApplicationRequests::GetEntityName, entityOrderArray[0]);
```
forkercat, line 54: Learn new thing every day! I would probably do something similar to validate order in "detach prefab". Currently, the order is not covered in auto and unit tests.

The existing approach seems better. Just curious if we can also retrieve the sort order component and do the validation?

forkercat (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -259,74 +259,92 @@ namespace AzToolsFramework
...
+                            topLevelEntityDomWithParentInTransform, *topLevelEntity);
+
+                        PrefabDomPath entityAliasDomPath(m_instanceToTemplateInterface->GenerateEntityAliasPath(topLevelEntityId).c_str());
+                        entityAliasDomPath.Set(newInstanceTemplateDom, topLevelEntityDomWithParentInTransform.Move());
+                    }
+                }
+                EditorEntityContextNotificationBus::Broadcast(&EditorEntityContextNotification::SetForceAddEntitiesToBackFlag, false);
+
+                // Set up remaining links between the created prefab instance and its nested instances without undo/redo support.
+                instanceToCreate->get().GetNestedInstances([&](AZStd::unique_ptr<Instance>& nestedInstance) {
```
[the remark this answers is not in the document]
forkercat: Good to know!

Code/Framework/AzToolsFramework/Tests/Prefab/PrefabCreateInMemoryTests.cpp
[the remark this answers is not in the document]
forkercat: Good to know!

Code/Framework/AzToolsFramework/Tests/Prefab/PrefabCreateInMemoryTests.cpp
michabr: Yes, we could also do the validation using the sort order component, but since there may be changes coming to the sort order component soon, I decided to use this API that's less likely to change.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -259,74 +259,92 @@ namespace AzToolsFramework
...
 
                     AZ::EntityId nestedInstanceContainerEntityId = nestedInstanceContainerEntity->get().GetId();
                     PrefabDom previousPatch;
 
                     // Retrieve the previous patch if it exists
-                    if (nestedInstanceLinkPatchesMap.contains(nestedInstance.get()))
+                    if (nestedInstanceLinkPatchesMap.contains(&nestedInstance))
                     {
-                        previousPatch.Swap(nestedInstanceLinkPatchesMap[nestedInstance.get()]);
+                        previousPatch.Swap(nestedInstanceLinkPatchesMap[&nestedInstance]);
```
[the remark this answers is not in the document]
michabr: Thanks for pointing that out! Changed.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -259,74 +259,93 @@ namespace AzToolsFramework
                 m_instanceToTemplateInterface->GenerateEntityDomBySerializing(
                     newContainerEntityDomInitialStateWithTransform, *newContainerEntity);
 
-                // Parent the non-container top level entities to the new container entity.
-                // It also patches the entity DOMs to the new template without undo/redo support.
-                EditorEntityContextNotificationBus::Broadcast(&EditorEntityContextNotification::SetForceAddEntitiesToBackFlag, true);
-                for (AZ::Entity* topLevelEntity : topLevelEntities)
+                // Helper function to create a link between a nested instance and the newly created instance
+                auto createLinkForNestedInstanceFunc = [&](Instance& nestedInstance)
```
srikappa-amzn, line 263: nit: I recommend renaming it to 'createLinkForNestedInstance' to avoid using half names.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -259,74 +259,93 @@ namespace AzToolsFramework
...
+                            topLevelEntityDomWithParentInTransform, *topLevelEntity);
+
+                        PrefabDomPath entityAliasDomPath(m_instanceToTemplateInterface->GenerateEntityAliasPath(topLevelEntityId).c_str());
+                        entityAliasDomPath.Set(newInstanceTemplateDom, topLevelEntityDomWithParentInTransform.Move());
+                    }
+                }
+                EditorEntityContextNotificationBus::Broadcast(&EditorEntityContextNotification::SetForceAddEntitiesToBackFlag, false);
+
+                // Set up remaining links for the nested instances that are not at the top level. This is done without undo/redo support.
+                instanceToCreate->get().GetNestedInstances([topLevelEntities, createLinkForNestedInstanceFunc](AZStd::unique_ptr<Instance>& nestedInstance) {
```
srikappa-amzn, line 337: I assume this piece of code handles case like below when a prefab is not a toplevel entity but a child of another entity? If that's the case, it would be good to add another unit test to cover this logic as it won't be covered with the test you added below

- entity1 (select)
  - prefab1(select)
- entity2 (select)

srikappa-amzn (approved):

srikappa-amzn:
> Great to see the tests added! Thanks for doing that 👍 Last question is have we run any performance regression tests against this just to ensure this hasn't slowed anything down too much. I just know these kinds of regressions are all to easy to create so if we have any large scenes we can test these operations on it would definitely be worth checking before/after to make sure there hasn't been a noticeable increase. Thanks!

This PR simply rearranges existing code rather than adding any new logic. As such, I don't believe this can cause performance regressions. On a related note, we do have[ benchmarks around creation of prefabs](https://github.com/o3de/o3de/blob/development/Code/Framework/AzToolsFramework/Tests/Prefab/Benchmark/PrefabCreateBenchmarks.cpp) but they only test the PrefabSystemComponent's CreatePrefab() function. They don't test the PrefabPublicHandler's CreatePrefab() function which is a wrapper around the previous one. It might be a good idea to add a github issue to update the benchmarks to use the PrefabPublicHandler's version so that we can mimic the editor behavior better

hultonha:
> > Great to see the tests added! Thanks for doing that 👍 Last question is have we run any performance regression tests against this just to ensure this hasn't slowed anything down too much. I just know these kinds of regressions are all to easy to create so if we have any large scenes we can test these operations on it would definitely be worth checking before/after to make sure there hasn't been a noticeable increase. Thanks!
> 
> This PR simply rearranges existing code rather than adding any new logic. As such, I don't believe this can cause performance regressions. On a related note, we do have[ benchmarks around creation of prefabs](https://github.com/o3de/o3de/blob/development/Code/Framework/AzToolsFramework/Tests/Prefab/Benchmark/PrefabCreateBenchmarks.cpp) but they only test the PrefabSystemComponent's CreatePrefab() function. They don't test the PrefabPublicHandler's CreatePrefab() function which is a wrapper around the previous one. It might be a good idea to add a github issue to update the benchmarks to use the PrefabPublicHandler's version so that we can mimic the editor behavior better

Okay good to know, I just wanted to be sure 😅 Yeah that might be nice, it's definitely a useful thing to have long term to build confidence in our changes not having any unintended consequences (which is all too easy 🙈)

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
michabr: Actually thinking about it more, the entity order only applies to top-level entities, so prefab1 wouldn't be reparented in this case.

srikappa-amzn: Yes, it shouldn't be reparented, it should retain its original parent. It'll be good to validate that still happens though. Because the code highlighted is responsible for creating a link between 'prefab1' and the new prefab that is being created now. So, a new link will be created for 'prefab1' but it should retain its old entity parent. It will be a good test case to have because of the way the code is split in the CreatePrefab() function

forkercat: +1. I also think we can, probably in a follow-up PR, add one more test to validate parent entities of non-top-level entities. 

For 'prefab1', we can also add these to validate the change in its owning instance.
```cpp
ValidateNestedInstanceNotUnderInstance(<previous parent instance>, prefab1InstanceAlias);
ValidateNestedInstanceUnderInstance(<newly created prefab instance>, prefab1InstanceAlias);
```

michabr: I agree we need to add more "create prefab" tests. I will create a separate ticket for that.

PR #14195, opened 2023-01-24
Jointname modification

danielemorra98:
## What does this PR do?

It adds a new label to physX joint that specifies the joint's name. It is useful in many robotics application, such as the control of a robotic manipulator using [MoveIt](https://moveit.picknik.ai/humble/index.html) through the [ROS2Gem](https://github.com/o3de/o3de-extras/tree/development/Gems/ROS2)  

## How was this PR tested?

Tested on my workstation device:
OS: Linux
Version: Focal - 20.04
CPU: 11th Gen Intel Core i9-11900K
GPU: NVIDIA RTX A5000
Memory: 64 Gb

### DIFF
--- a/Gems/PhysX/Code/Editor/EditorJointConfiguration.cpp
+++ b/Gems/PhysX/Code/Editor/EditorJointConfiguration.cpp
@@ -378,7 +378,7 @@ namespace PhysX
             flags |= JointGenericProperties::GenericJointFlag::SelfCollide;
         }
 
-        return JointGenericProperties(flags, m_forceMax, m_torqueMax, m_jointname);
+        return JointGenericProperties(flags, m_forceMax, m_torqueMax, m_jointName);
     }
 
     JointComponentConfiguration EditorJointConfig::ToGameTimeConfig() const
--- a/Gems/PhysX/Code/Editor/EditorJointConfiguration.cpp
+++ b/Gems/PhysX/Code/Editor/EditorJointConfiguration.cpp
@@ -333,7 +333,7 @@ namespace PhysX
                     ->Attribute(AZ::Edit::Attributes::ReadOnly, &EditorJointConfig::IsInComponentMode)
                     ->DataElement(AZ::Edit::UIHandlers::Default, &EditorJointConfig::m_jointname 
                         , "Joint name"
-                        , "Usually coming from URDF")
+                        , "Name used to easily identify the joint")
                     ->DataElement(0, &PhysX::EditorJointConfig::m_selectLeadOnSnap, "Select Lead on Snap"
                         , "Select lead entity on snap to position in component mode.")
                     ->DataElement(0, &PhysX::EditorJointConfig::m_breakable
--- a/Gems/PhysX/Code/Editor/EditorJointConfiguration.cpp
+++ b/Gems/PhysX/Code/Editor/EditorJointConfiguration.cpp
@@ -298,6 +298,7 @@ namespace PhysX
                 ->Field("Display Debug", &EditorJointConfig::m_displayJointSetup)
                 ->Field("Select Lead on Snap", &EditorJointConfig::m_selectLeadOnSnap)
                 ->Field("Self Collide", &EditorJointConfig::m_selfCollide)
+                // ->Field("Joint Name", &EditorJointConfig::m_jointname)
                 ;
 
             if (auto* editContext = serializeContext->GetEditContext())
@@ -374,7 +375,7 @@ namespace PhysX
             flags |= JointGenericProperties::GenericJointFlag::SelfCollide;
         }
 
-        return JointGenericProperties(flags, m_forceMax, m_torqueMax);
+        return JointGenericProperties(flags, m_forceMax, m_torqueMax, m_jointname);
     }
 
     JointComponentConfiguration EditorJointConfig::ToGameTimeConfig() const
--- a/Gems/PhysX/Code/Editor/EditorJointConfiguration.cpp
+++ b/Gems/PhysX/Code/Editor/EditorJointConfiguration.cpp
@@ -298,7 +298,7 @@ namespace PhysX
                 ->Field("Display Debug", &EditorJointConfig::m_displayJointSetup)
                 ->Field("Select Lead on Snap", &EditorJointConfig::m_selectLeadOnSnap)
                 ->Field("Self Collide", &EditorJointConfig::m_selfCollide)
-                // ->Field("Joint Name", &EditorJointConfig::m_jointname)
+                ->Field("Joint Name", &EditorJointConfig::m_jointname)
                 ;
 
             if (auto* editContext = serializeContext->GetEditContext())
@@ -331,6 +331,9 @@ namespace PhysX
                         "Select = Show setup display when entity is selected."
                         "Always = Always show setup display.")
                     ->Attribute(AZ::Edit::Attributes::ReadOnly, &EditorJointConfig::IsInComponentMode)
+                    ->DataElement(AZ::Edit::UIHandlers::Default, &EditorJointConfig::m_jointname 
+                        , "Joint name"
+                        , "Usually coming from URDF")
                     ->DataElement(0, &PhysX::EditorJointConfig::m_selectLeadOnSnap, "Select Lead on Snap"
                         , "Select lead entity on snap to position in component mode.")
                     ->DataElement(0, &PhysX::EditorJointConfig::m_breakable
--- a/Gems/PhysX/Code/Editor/EditorJointConfiguration.cpp
+++ b/Gems/PhysX/Code/Editor/EditorJointConfiguration.cpp
@@ -298,7 +298,7 @@ namespace PhysX
                 ->Field("Display Debug", &EditorJointConfig::m_displayJointSetup)
                 ->Field("Select Lead on Snap", &EditorJointConfig::m_selectLeadOnSnap)
                 ->Field("Self Collide", &EditorJointConfig::m_selfCollide)
-                ->Field("Joint Name", &EditorJointConfig::m_jointname)
+                ->Field("Joint Name", &EditorJointConfig::m_jointName)
                 ;
 
             if (auto* editContext = serializeContext->GetEditContext())
@@ -331,7 +331,7 @@ namespace PhysX
                         "Select = Show setup display when entity is selected."
                         "Always = Always show setup display.")
                     ->Attribute(AZ::Edit::Attributes::ReadOnly, &EditorJointConfig::IsInComponentMode)
-                    ->DataElement(AZ::Edit::UIHandlers::Default, &EditorJointConfig::m_jointname 
+                    ->DataElement(AZ::Edit::UIHandlers::Default, &EditorJointConfig::m_jointName 
                         , "Joint name"
                         , "Name used to easily identify the joint")
                     ->DataElement(0, &PhysX::EditorJointConfig::m_selectLeadOnSnap, "Select Lead on Snap"
--- a/Gems/PhysX/Code/Editor/EditorJointConfiguration.h
+++ b/Gems/PhysX/Code/Editor/EditorJointConfiguration.h
@@ -147,7 +147,7 @@ namespace PhysX
         float m_forceMax = 1.0f;
         float m_torqueMax = 1.0f;
         
-        AZStd::string m_jointname{""};
+        AZStd::string m_jointName{""};
 
         AZ::Vector3 m_localPosition = AZ::Vector3::CreateZero();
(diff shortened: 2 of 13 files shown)


Gems/PhysX/Code/Editor/EditorJointConfiguration.cpp
```diff
@@ -330,6 +331,9 @@ namespace PhysX
                         "Select = Show setup display when entity is selected."
                         "Always = Always show setup display.")
                     ->Attribute(AZ::Edit::Attributes::ReadOnly, &EditorJointConfig::IsInComponentMode)
+                    ->DataElement(AZ::Edit::UIHandlers::Default, &EditorJointConfig::m_jointname 
+                        , "Joint name"
+                        , "Usually coming from URDF")
```
lgleim, line 336: The comment could be more descriptive. Given that "Joint name" is already quite generic, it might be hard for users to derive what this name is actually used for/by.

lgleim (approved):
LGTM. The description of the "Joint name" UI field could use some more context about what the joint name is actually used for though.

Gems/PhysX/Code/Editor/EditorJointConfiguration.cpp
danielemorra98: I agree, maybe it is better not to erase it from the `editContext`. Finally it is going to be used only by developers.

moraaar, line 336: >"Usually coming from URDF"

That is very specific to be added into description of the name. It'd need to be kept generic: "Name used to easily identify the joint" or something in those lines.

Gems/PhysX/Code/Editor/EditorJointConfiguration.h
```diff
@@ -146,6 +146,8 @@ namespace PhysX
 
         float m_forceMax = 1.0f;
         float m_torqueMax = 1.0f;
+        
+        AZStd::string m_jointname{""};
```
moraaar, line 150: [standard] name convention for member variables: `m_jointname` > `m_jointName`

Gems/PhysX/Code/Include/PhysX/Joint/Configuration/PhysXJointConfiguration.h
```diff
@@ -52,7 +52,7 @@ namespace PhysX
         static void Reflect(AZ::ReflectContext* context);
 
         JointGenericProperties() = default;
-        JointGenericProperties(GenericJointFlag flags, float forceMax, float torqueMax);
+        JointGenericProperties(GenericJointFlag flags, float forceMax, float torqueMax, AZStd::string joint_name);
```
moraaar, line 55: [standard] `joint_name` > `jointName`

Gems/PhysX/Code/Source/HingeJointComponent.cpp
```diff
@@ -120,6 +120,7 @@ namespace PhysX
         return native;
     }
 
+
```
moraaar, line 123: The only change in this file is this unnecessary extra line, please revert.

Gems/PhysX/Code/Include/PhysX/Joint/Configuration/PhysXJointConfiguration.h
```diff
@@ -61,6 +61,7 @@ namespace PhysX
         GenericJointFlag m_flags = GenericJointFlag::None;
         float m_forceMax = 1.0f; ///< Max force joint can tolerate before breaking.
         float m_torqueMax = 1.0f; ///< Max torque joint can tolerate before breaking.
+        AZStd::string m_jointname{""};  // joint_name (useful for some ROS2 application)
```
moraaar, line 64: [standard] `m_jointname`> `m_jointName`

I also wouldn't mention other gems in here in the comment, please remove the comment, it's pretty self explanatory that it's a joint name.

Gems/PhysX/Code/Source/Joint/Configuration/PhysXJointConfiguration.cpp
```diff
@@ -13,12 +13,12 @@
 
 namespace PhysX
 {
-    JointGenericProperties::JointGenericProperties(GenericJointFlag flags, float forceMax, float torqueMax)
+    JointGenericProperties::JointGenericProperties(GenericJointFlag flags, float forceMax, float torqueMax, AZStd::string joint_name)
```
moraaar, line 16: `joint_name` > `jointName`

Gems/PhysX/Code/Source/Joint/Configuration/PhysXJointConfiguration.cpp
```diff
@@ -13,12 +13,12 @@
...
         : m_flags(flags)
         , m_forceMax(forceMax)
         , m_torqueMax(torqueMax)
     {
-
+        m_jointname = joint_name;
```
moraaar, line 21: The initialization of `m_jointName` could be done like the other members, and since the string parameter will contain a copy (since it's not passed as reference) it can move the string to avoid another copy.

````
    JointGenericProperties::JointGenericProperties(GenericJointFlag flags, float forceMax, float torqueMax, AZStd::string jointName)
        : m_flags(flags)
        , m_forceMax(forceMax)
        , m_torqueMax(torqueMax)
        , m_jointName(AZStd::move(jointName))
    {
    }
````

File: Gems/PhysX/Code/Source/System/PhysXSystem.cpp [hunk 8]
```diff
@@ -455,13 +455,7 @@ namespace PhysX
         m_physXSdk.m_cooking = PxCreateCooking(PX_PHYSICS_VERSION, *m_physXSdk.m_foundation, cookingParams);
 
         // Set up CPU dispatcher
-#if defined(AZ_PLATFORM_LINUX)
-        // Temporary workaround for linux. At the moment using AzPhysXCpuDispatcher results in an assert at
-        // PhysX mutex indicating it must be unlocked only by the thread that has already acquired lock.
-        m_cpuDispatcher = physx::PxDefaultCpuDispatcherCreate(0);
-#else
         m_cpuDispatcher = PhysXCpuDispatcherCreate();
-#endif
```
moraaar, line 464: This change is not related to this PR and it's already present in `development` branch. Please make sure to merge the latest from `development` branch to this branch before proceeding.

moraaar:
I forgot, please also confirm you've built and run PhysX runtime (PhysX.Tests) and editor (PhysX.Editor.Tests) unit tests and that they all compile and pass.

greerdv:
Would this be just for display/debugging purposes or would you ever want to do things like search for a joint by name? If so, could it be worth looking at AZ::Name (which hashes names to make comparisons faster)?

moraaar:
Thanks for covering my feedback. I've triggered the AR build, in the meantime @danielemorra98 if you could please answer @greerdv question and mine that'd be great. Thanks.

danielemorra98:
> Would this be just for display/debugging purposes or would you ever want to do things like search for a joint by name? If so, could it be worth looking at AZ::Name (which hashes names to make comparisons faster)?

@greerdv thanks for the suggestion. I have still some doubts on how I can get the reference Component/Entity associated with a specific AZ::Name. 
Let's say I want to implement a Controller for a manipulator. I would like to retrieve all the jointComponents which stand in the robotic arm tree. I looked at the code but I have no hints on how to do that with AZ::Name methods. How would I do that?

danielemorra98:
> I forgot, please also confirm you've built and run PhysX runtime (PhysX.Tests) and editor (PhysX.Editor.Tests) unit tests and that they all compile and pass.

Currently doing it. I will let you know as soon as possible.

greerdv:
> 

I don't think the Name system gives you any functionality like that (it just hashes names and has a system to prevent hash collisions). So if you wanted to do something like that I think you'd still need something like a map between Names and e.g. EntityId or EntityComponentIdPair. The advantage of using Name would just be that lookup in that map should be faster vs if the key was a string. Any other logic that needs to do comparisons would also be faster. Anyway, it's just a suggestion, and how worthwhile it is depends on how frequent the lookup/comparison use case is going to be.

danielemorra98:
`PhysX.Tests` and `PhysX.Editor.Tests` built. I do not know if it is enough or there are some way to run them (I did not find any executables related, only .so library)

moraaar:
> `PhysX.Tests` and `PhysX.Editor.Tests` built. I do not know if it is enough or there are some way to run them (I did not find any executables related, only .so library)

Thanks for building them, that's the first step done :) 

You can run the unit tests using `AzTestRunner` executable, you have to build it using `AzTestRunner` target and it will be in the `bin/profile` folder of your build folder.

Then to run the tests of a specific gem you use the following command:

````
>AzTestRunner /<build-folder-path>/bin/profile/PhysX.Tests.Gem.so AzRunUnitTests
>AzTestRunner /<build-folder-path>/bin/profile/PhysX.EditorTests.Gem.so AzRunUnitTests
````

That should work on Linux. On windows with Visual Studio you can just run the `PhysX.Tests` and `PhysX.Editor.Tests` project and it handles AzTestRunner for you, on Linux you have to call AzTestRunner directly.

It's good practice to build and run the unit tests of a gem when modifying it (and add more unit tests to cover the modification when applies).

**Comment by moraaar** [comment]
Using `ctest` (part of CMake), as Dave suggested, should work as well. I normally use AzTestRunner directly, but there is nothing wrong using `ctest`.

**Comment by greerdv** [comment]
The ctest way does have a slight advantage that you can run all the relevant tests in parallel with one command

ctest -j16 -C profile --output-on-failure -R Gem::PhysX*

**Comment by hultonha** [comment]
Tagging @michalpelka, @adamdbrw and @pawelbudziszewski for visibility as well as this may be useful for them to know about.

File: Gems/PhysX/Code/Editor/EditorJointConfiguration.h [hunk 9]
```diff
@@ -146,6 +146,8 @@ namespace PhysX
 
         float m_forceMax = 1.0f;
         float m_torqueMax = 1.0f;
+        
+        AZStd::string m_jointName{""};
```
hultonha, line 150: nit: An `AZStd::string` has a default constructor and will be initialized to a known state (empty string) so the `{""}` is superfluous here and could be removed.

moraaar (approved):
[AR build passed](https://jenkins.build.o3de.org/blue/organizations/jenkins/O3DE/detail/PR-14195/1/pipeline/297/). Approved.

Gems/PhysX/Code/Include/PhysX/Joint/Configuration/PhysXJointConfiguration.h
```diff
@@ -61,6 +61,7 @@ namespace PhysX
         GenericJointFlag m_flags = GenericJointFlag::None;
         float m_forceMax = 1.0f; ///< Max force joint can tolerate before breaking.
         float m_torqueMax = 1.0f; ///< Max torque joint can tolerate before breaking.
+        AZStd::string m_jointName{""};
```
hultonha, line 64: nit: Same comment about `{""}`

Gems/PhysX/Code/Source/Joint/Configuration/PhysXJointConfiguration.cpp
```diff
@@ -13,12 +13,13 @@
...
 {
-    JointGenericProperties::JointGenericProperties(GenericJointFlag flags, float forceMax, float torqueMax)
+    JointGenericProperties::JointGenericProperties(GenericJointFlag flags, float forceMax, float torqueMax, AZStd::string jointName)
         : m_flags(flags)
         , m_forceMax(forceMax)
         , m_torqueMax(torqueMax)
+        , m_jointName(AZStd::move(jointName))
     {
-
+
```
hultonha, line 22: uber nit: Can we please remove the trailing white space if possible please 😅

Gems/PhysX/Code/Source/JointComponent.h
```diff
@@ -64,6 +64,8 @@ namespace PhysX
             const JointLimitProperties& limitProperties,
             const JointMotorProperties& motorProperties);
 
+        AZStd::string GetJointName() const;
```
hultonha, line 67: suggestion: We could have this return a `const AZStd::string&` as we're not making this virtual and might have to deal with users returning a temporary. Returning by reference will be faster in this case.

hultonha (approved):
LGTM 👍 I just have a few small minor comments, thank you for the contribution! 😄

lgleim:
I just had a longer review discussion with Daniele and it might actually make more sense to maintain Joint name metadata within the ROS2 Gem instead of here. Especially, since the name may not be the only type of metadata that needs to be managed. 

Do applications outside of robotics benefit from being able to name joints? 

I will convert this PR to draft until this question has been clarified. Please excuse the overhead.

hultonha:
No problem, thanks for the update @lgleim

Issue #13832, opened 2022-12-23
Crash Saving a new Prefab after deleting an old one with the same name


amzn-tommy:
**Describe the bug**
Trying to create a new prefab and delete an old one with the same name leads to a crash. The fix could be as simple as a null check, but would like someone more familiar with the system to weigh in on if there is something deeper that needs to be addressed.

**Assets required**

**Steps to reproduce**
1. Open any level in the editor
2. In the main editor viewport, right-click and choose "create prefab"
3. Give it a file name like myprefab and save it. This will create a prefab out of the currently selected entity
4. ctrl+z to undo the prefab creation and make the entity a regular entity again
5. In the main viewport, righ-click->create entity
6. In the entity outliner, select the new entity and right click->save as prefab
7. Try to save it as myprefab, and see a warning that it will break other prefabs
8. Ignore the warning, because you know that one was just now created by mistake, and nothing is using it or inheriting from it
9. Delete myprefab.prefab via the windows file dialog
10. Save the new prefab as myprefab.prefab

**Expected behavior**
No crash, new prefab saves successfully, original prefab is deleted

**Actual behavior**
Crash with the following callstack:
>	EditorLib.dll!AzToolsFramework::Prefab::Template::IsValid() Line 88	C++
 	EditorLib.dll!AzToolsFramework::Prefab::PrefabSystemComponent::FindTemplateDom(unsigned __int64 templateId) Line 505	C++
 	EditorLib.dll!AzToolsFramework::Prefab::PrefabPublicHandler::CreatePrefabInMemory(const AZStd::vector<AZ::EntityId,AZStd::allocator> & entityIds, AZ::IO::PathView filePath) Line 246	C++
 	EditorLib.dll!AzToolsFramework::Prefab::PrefabPublicHandler::CreatePrefabAndSaveToDisk(const AZStd::vector<AZ::EntityId,AZStd::allocator> & entityIds, AZ::IO::PathView filePath) Line 390	C++
 	ComponentEntityEditorPlugin.dll!AzToolsFramework::Prefab::PrefabIntegrationManager::ContextMenu_CreatePrefab(AZStd::vector<AZ::EntityId,AZStd::allocator> selectedEntities) Line 790	C++
 	[Inline Frame] ComponentEntityEditorPlugin.dll!AzToolsFramework::Prefab::PrefabIntegrationManager::PopulateEditorGlobalContextMenu::__l76::<lambda_6>::operator()() Line 499	C++
[...]

PR #14197, opened 2023-01-24
Enable hot reloading to detect prefab deletions

srikappa-amzn:
## What does this PR do?

Fixes https://github.com/o3de/o3de/issues/13832 by enabling hot reloading for prefab deletions

Here is the toast and warning that shows up when a prefab is deleted but there are instances of it in the level currently opened in the O3DE editor

toast-for-prefab-deletions

## Are any alternatives considered?
Yes. Another idea was prototyped but discarded since it has shortcomings. The idea and the pros and cons can be found here : https://github.com/o3de/o3de/pull/14196

## How was this PR tested?

Manually. Integration tests need to be added later as testing this would require mimicking file deletions and dialog clicks
Test cases performed manually:

1. Deleting a prefab removes the template from memory if no instances using it are found
2. Deleting a prefab shows a toast and warning indicating the user about deletion if at least 1 instance is found.


Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabSaveLoadHandler.cpp
```diff
@@ -20,10 +20,14 @@
...
 #include <AzToolsFramework/AssetBrowser/Entries/SourceAssetBrowserEntry.h>
 #include <AzToolsFramework/AssetBrowser/Entries/AssetBrowserEntryUtils.h>
+#include <AzToolsFramework/Entity/EditorEntityContextBus.h>
+#include <AzToolsFramework/Entity/PrefabEditorEntityOwnershipInterface.h>
 #include <AzToolsFramework/Prefab/Instance/InstanceEntityMapperInterface.h>
+#include <AzToolsFramework/Prefab/PrefabFocusPublicInterface.h>
 #include <AzToolsFramework/Prefab/PrefabLoaderInterface.h>
 #include <AzToolsFramework/Prefab/PrefabPublicInterface.h>
 #include <AzToolsFramework/Prefab/PrefabSystemComponentInterface.h>
+#include <AzToolsFramework/Prefab/Instance/TemplateInstanceMapperInterface.h>
```
forkercat, line 30: nit: include order. Should be on line 26?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabSaveLoadHandler.cpp
```diff
@@ -111,17 +115,40 @@ namespace AzToolsFramework
...
+            }
+
+            m_prefabFocusPublicInterface = AZ::Interface<PrefabFocusPublicInterface>::Get();
+            if (m_prefabFocusPublicInterface == nullptr)
+            {
+                AZ_Assert(false, "PrefabSaveHandler - Could not get PrefabFocusPublicInterface on construction.");
+                return;
+            }
+            
+
```
forkercat, line 136: nit: extra line

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabSaveLoadHandler.h
```diff
@@ -114,6 +122,14 @@ namespace AzToolsFramework
                 const AZStd::string& prefabName, const AZStd::string& targetDirectory, AZStd::string& suggestedFullPath);
 
         private:
+
+            //! AssetSystemBus notification handlers
+            //! @{
+            //! Called by the AssetProcessor when the source file has been removed.
+            void SourceFileRemoved(AZStd::string relativePath, AZStd::string scanFolder, AZ::Uuid sourceUUID) override;
+            //! @}
+
```
forkercat, line 131: nit: extra line

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabFocusHandler.h
```diff
@@ -65,6 +64,7 @@ namespace AzToolsFramework::Prefab
         PrefabFocusOperationResult FocusOnPathIndex(AzFramework::EntityContextId entityContextId, int index) override;
         PrefabFocusOperationResult SetOwningPrefabInstanceOpenState(AZ::EntityId entityId, bool openState) override;
         AZ::EntityId GetFocusedPrefabContainerEntityId(AzFramework::EntityContextId entityContextId) const override;
+        TemplateId GetFocusedPrefabTemplateId(AzFramework::EntityContextId entityContextId) const override;
```
forkercat, line 67: Would be good to include why the focus API changes in description.

In addition, should `GetFocusedPrefabInstance()` be moved to public interface as well?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabSaveLoadHandler.cpp
```diff
@@ -993,5 +1020,59 @@ namespace AzToolsFramework
...
+                    // There are no instances mapping to this template. So it can be removed from memory.
+                    s_prefabSystemComponentInterface->RemoveTemplate(loadedTemplateId);
+                }
+                else
+                {
+                    TemplateId rootTemplateId = m_prefabEditorEntityOwnershipInterface->GetRootPrefabTemplateId();
+                    if (rootTemplateId == loadedTemplateId)
+                    {
+                        // Deletion of the current level that the user has opened needs to be handled differently. Skip for now
+                        return;
```
forkercat, line 1045: question: After return, would it crash the editor?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabSaveLoadHandler.cpp
```diff
@@ -993,5 +1020,59 @@ namespace AzToolsFramework
...
+                            QString(),
+                            QString("Prefab '%1' has been deleted from the file system. But we detected %2 instance(s) of it "
+                                    "in the level. Would you like to delete them ?")
+                                .arg(relativePath.c_str())
+                                .arg(instancesMappedToTemplate->get().size()),
+                            QMessageBox::Yes | QMessageBox::No);
+                        switch (button)
+                        {
+                        case QMessageBox::Yes:
+                            s_prefabSystemComponentInterface->RemoveTemplate(loadedTemplateId);
```
forkercat, line 1069: Is this enough for removing nested instances in outliner? How does the nested instance get removed from outliner?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabSaveLoadHandler.cpp
```diff
@@ -993,5 +1020,59 @@ namespace AzToolsFramework
...
+                        if (m_prefabFocusPublicInterface->GetFocusedPrefabTemplateId(editorEntityContextId) == loadedTemplateId)
+                        {
+                            m_prefabFocusPublicInterface->FocusOnParentOfFocusedPrefab(editorEntityContextId);
+                        }
+
+                        auto button = QMessageBox::question(
+                            AzToolsFramework::GetActiveWindow(),
+                            QString(),
+                            QString("Prefab '%1' has been deleted from the file system. But we detected %2 instance(s) of it "
+                                    "in the level. Would you like to delete them ?")
```
forkercat, line 1062: If there are no instances instantiated in Outliner, the dialog would still show up with "0 instance(s)"?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabSaveLoadHandler.cpp
srikappa-amzn: No it doesn't. It's the same behavior as current development. If level prefab or other prefabs within it are deleted, the loaded level stays as-is.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabSaveLoadHandler.cpp
srikappa-amzn: No. When there are 0 instances, we will remove the template from memory without notifying the user. It's handled here : https://github.com/o3de/o3de/pull/14197/files#diff-d5729d094928c4642299dddca98bba2de13aec0c7220ead69fb2f42f949fb751R1034

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabSaveLoadHandler.cpp
srikappa-amzn: Yes. Because RemoveTemplate() will also remove any links from the template to other dependent templates. I added an additional statement to trigger propagation on those dependent templates, which will eventually put the outliner in the right state : https://github.com/o3de/o3de/pull/14197/files#diff-ae6dbd26ad65eaa6de3465864f940853c6f5842fc3b9f68557b5b70600db1d48R1211

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabFocusHandler.h
srikappa-amzn: It is my understanding that the PrefabFocusPublicRequestBus should be callable via scripting. So, putting the GetFocusedPrefabInstance function there would violate that rule as it returns an optional instance reference that can't be easily read in python as far as I'm aware. @AMZN-daimini , correct me if I'm wrong.

As for moving GetFocusedPrefabTemplateId() to public interface, the reasoning is the same. It adds additional capability for scripts to query template id which they can use to query prefab system later if needed. Plus I didn't want the PrefabSaveLoadHandler to be using both the public and the nonpublic interfaces at the same time.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabSaveLoadHandler.cpp
forkercat: Cool. I missed that line. Thanks!

AMZN-daimini:
Description doesn't really go over what happens if the user clicks "No". Will they keep the template in memory, but then be redirected to a "Save As..." dialog when trying to save it? In that case, it would take us closer to the original intent with prefabs (as in, you can create prefabs completely in-memory and only specify a file path when saving) but it also means that there's a lot more UX that we haven't had so far that needs to be explored.

(Still have to check the code so maybe I will find some of the answers there)

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabFocusHandler.h
AMZN-daimini: Correct, the idea was to only use TemplateIds and InstanceReferences in the internal interfaces/buses, and use EntityIds in the public interfaces. This both to hide the Prefab internals from client systems, and to avoid having to expose those to Python scripting.

srikappa-amzn:
> Description doesn't really go over what happens if the user clicks "No". Will they keep the template in memory, but then be redirected to a "Save As..." dialog when trying to save it? In that case, it would take us closer to the original intent with prefabs (as in, you can create prefabs completely in-memory and only specify a file path when saving) but it also means that there's a lot more UX that we haven't had so far that needs to be explored.
> 
> (Still have to check the code so maybe I will find some of the answers there)
> 
> EDIT: No I don't think I found answers, so my questions are still in place :D

When user clicks 'No', the dialog is simply closed without any action. This is what currently happens in development as well. I've considered marking the template as dirty. But when closing the level, if user saves dirty prefabs again, then it would be saved back to the file. I'm not sure if that's the desired behavior always. For example, a prefab is deliberately deleted in your project but one of the team members have a level that uses it. Should they be given the ability to add it back? I think that question requires some ux investigation and we can probably keep the behavior as is and revisit it later once we get feedback from users. The chances of deletions happening without the user knowing would be rare imo btw. So there is also a chance here that we maybe preemptively solving all cases without it being a real issue.

forkercat (approved):

forkercat:
One use case just came to my mind. Will it show up all dialogs **one by one** if the users delete multiple prefab files at once?

srikappa-amzn:
> One use case just came to my mind. Will it show up all dialogs **one by one** if the users delete multiple prefab files at once?

Yes, it will. It needs to be improved in next iterations to queue up deletions and show a different dialog including everything in the queue

AMZN-daimini:
> > Description doesn't really go over what happens if the user clicks "No". Will they keep the template in memory, but then be redirected to a "Save As..." dialog when trying to save it? In that case, it would take us closer to the original intent with prefabs (as in, you can create prefabs completely in-memory and only specify a file path when saving) but it also means that there's a lot more UX that we haven't had so far that needs to be explored.
> > (Still have to check the code so maybe I will find some of the answers there)
> > EDIT: No I don't think I found answers, so my questions are still in place :D
> 
> When user clicks 'No', the dialog is simply closed without any action. This is what currently happens in development as well(Not the dialog part, but the part where deletes doesn't affect loaded prefab instances). I've considered marking the template as dirty. But when closing the level, if user saves dirty prefabs again, then it would be saved back to the file. I'm not sure if that's the desired behavior always. For example, a prefab is deliberately deleted in your project but one of the team members have a level that uses it. Should they be given the ability to add it back? I think that question requires some ux investigation and we can probably keep the behavior as is and revisit it later once we get feedback from users. The chances of deletions happening without the user knowing would be rare imo btw. So there is also a chance here that we maybe preemptively solving all cases without it being a real issue.

Yeah you're pretty much nailing the UX issue - we don't support purely in-memory prefabs (even when creating a new one, we need a file path before moving on) but this change puts us in the position of having one.

I'm not sure this PR's current approach works, cause saving the level and restarting the Editor either selecting No results in a broken level being loaded.

We could identify a path as to how this should work so that it can be consistent everywhere. At a first glance, I can see two approaches:

1. We start working towards supporting in-memory templates with no filepath. In that case, clicking No would mark the template dirty and strip it of the filepath. Next time the user tries to save that template, no filepath is found, so the user is prompted to pick one via the "Save As..." dialog.
2. We don't support in-memory templates. Deleting a prefab and selecting "No" simply file marks the template as dirty, saving the template in any way just re-creates the file exactly where it was.

I think 1. would be the best path long term, as it puts us closer to most other applications where creating a new document doesn't automatically save it to disk but only does it when the user intentionally presses Ctrl+S. The problem with this approach is that we use filepaths to identify nested prefabs in the serialized templates, so this kind of behavior would require to also invest in more stable ways to relocate prefab files without other files losing their dependency and breaking. That seems to be something we would want either way though, especially with the current efforts with allowing asset relocation in the Asset Browser without breaking dependencies, so maybe it makes sense to push in that direction?

...didn't mean to write this much :D
TLDR: I'm not sure the current approach in this PR for the "No" code path makes sense, but I wouldn't hold the PR as long as a discussion is started about the implications long term.

srikappa-amzn:
> > > Description doesn't really go over what happens if the user clicks "No". Will they keep the template in memory, but then be redirected to a "Save As..." dialog when trying to save it? In that case, it would take us closer to the original intent with prefabs (as in, you can create prefabs completely in-memory and only specify a file path when saving) but it also means that there's a lot more UX that we haven't had so far that needs to be explored.
> > > (Still have to check the code so maybe I will find some of the answers there)
> > > EDIT: No I don't think I found answers, so my questions are still in place :D
> > 
> > 
> > When user clicks 'No', the dialog is simply closed without any action. This is what currently happens in development as well(Not the dialog part, but the part where deletes doesn't affect loaded prefab instances). I've considered marking the template as dirty. But when closing the level, if user saves dirty prefabs again, then it would be saved back to the file. I'm not sure if that's the desired behavior always. For example, a prefab is deliberately deleted in your project but one of the team members have a level that uses it. Should they be given the ability to add it back? I think that question requires some ux investigation and we can probably keep the behavior as is and revisit it later once we get feedback from users. The chances of deletions happening without the user knowing would be rare imo btw. So there is also a chance here that we maybe preemptively solving all cases without it being a real issue.
> 
> Yeah you're pretty much nailing the UX issue - we don't support purely in-memory prefabs (even when creating a new one, we need a file path before moving on) but this change puts us in the position of having one.
> 
> I'm not sure this PR's current approach works, cause saving the level and restarting the Editor either selecting No results in a broken level being loaded.
> 
> We could identify a path as to how this should work so that it can be consistent everywhere. At a first glance, I can see two approaches:
> 
> 1. We start working towards supporting in-memory templates with no filepath. In that case, clicking No would mark the template dirty and strip it of the filepath. Next time the user tries to save that template, no filepath is found, so the user is prompted to pick one via the "Save As..." dialog.
> 2. We don't support in-memory templates. Deleting a prefab and selecting "No" simply file marks the template as dirty, saving the template in any way just re-creates the file exactly where it was.
> 
> I think 1. would be the best path long term, as it puts us closer to most other applications where creating a new document doesn't automatically save it to disk but only does it when the user intentionally presses Ctrl+S. The problem with this approach is that we use filepaths to identify nested prefabs in the serialized templates, so this kind of behavior would require to also invest in more stable ways to relocate prefab files without other files losing their dependency and breaking. That seems to be something we would want either way though, especially with the current efforts with allowing asset relocation in the Asset Browser without breaking dependencies, so maybe it makes sense to push in that direction?
> 
> ...didn't mean to write this much :D TLDR: I'm not sure the current approach in this PR for the "No" code path makes sense, but I wouldn't hold the PR as long as a discussion is started about the implications long term.

Thanks for the explaining the thought process in detail!  After reading this, I agree that the 'No' use case does have these problems right now:

- It's not obvious from the dialog what clicking 'No' means.
  -  So, for now, I'll expand the dialog text description with what it means. In the future, maybe we can show another popup dialog if ux agrees that says something like "If you click 'no', that the next time you load this level, you will not be able to load these problems. To prevent that, you can save these prefabs"
- The only reason the prefab system should even provide a 'no' option is to give the users an option to save the prefabs back again in cases like accidental deletes
  - So, it makes sense that when user clicks 'no', they at least show up as 'dirty' templates so that they can be saved via the outliner and be shown in the dirty prefab dialog when they try to close the level. I'll update the PR with that change. It does open up the door for users to accidentally save it back to file. But at some point, we need to trust the users to know what they are doing. So, it should be fine to mark it dirty imo.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabSystemComponent.cpp
```diff
@@ -46,18 +46,10 @@ namespace AzToolsFramework
...
             AZ::SystemTickBus::Handler::BusConnect();
-            if (AzToolsFramework::Prefab::IsHotReloadingEnabled())
-            {
-                AzToolsFramework::AssetSystemBus::Handler::BusConnect();
-            }
         }
 
         void PrefabSystemComponent::Deactivate()
         {
-            if (AzToolsFramework::Prefab::IsHotReloadingEnabled())
```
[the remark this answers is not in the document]
srikappa-amzn: It keeps the bus handler always active to listen to source file deletions. But we are enabling the hot reloading only when the deleted source file is a prefab and there are instances of it in the current level. So, rather than opening the floodgates to detect every prefab change on file, we will be carefully enabling hot reloading for only those cases that are causing issues to users.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabSaveLoadHandler.h
```diff
@@ -60,6 +68,7 @@ namespace AzToolsFramework
             : public QObject
             , public AzQtComponents::DragAndDropEventsBus::Handler
             , public AzQtComponents::DragAndDropItemViewEventsBus::Handler
+            , private AzToolsFramework::AssetSystemBus::Handler
```
michabr, line 71: The PrefabSaveHandler class seems specific to Qt/UI events. Would it make more sense to move the prefab asset change handling to a different class?

michabr (commented):
Does this also fix the case mentioned in the ticket where a user right clicks on an entity and selects "Create Prefab...", deletes a prefab file from the "Save File" dialog that pops up, but then uses the same file name as the deleted one?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabSaveLoadHandler.cpp
```diff
@@ -993,5 +1027,65 @@ namespace AzToolsFramework
 
             return AZStd::move(unsavedPrefabsContainer);
         }
+
+        void PrefabSaveHandler::SourceFileRemoved(
+            AZStd::string relativePath, AZStd::string scanFolder, [[maybe_unused]] AZ::Uuid sourceUUID)
+        {
+            // This gets triggered for every source file. We only need source files that are prefabs and are loaded in the current level.
+            TemplateId loadedTemplateId = s_prefabSystemComponentInterface->GetTemplateIdFromFilePath(relativePath.c_str());
```
michabr, line 1035: Can we check whether the relativePath has a ".prefab" extension or some way to early return if the source file is not a prefab file?

srikappa-amzn: I've considered doing that but the cost of fetching the extension from an AZStd::string would be higher than looking it up in a map that contains less than 100 entries (for a large level) in a map. So I'm directly looking it up in the map instead.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceDomGenerator.cpp
```diff
@@ -114,21 +114,27 @@ namespace AzToolsFramework
             // in focused instance DOM with the one seen by the root.
             else if (PrefabInstanceUtils::IsDescendantInstance(focusedInstance->get(), instance))
             {
-                // use instanceDom's allocator, because ultimately we will be setting this data back into
-                // instanceDom with move semantics.
-                PrefabDom focusedTemplateDomCopy(&instanceDom.GetAllocator());
+                // The template of the focused instance can be deleted in cases like hot reloading via file removal.
+                // So, check for a valid template id here before accessing its DOM
+                TemplateReference focusedTemplate= m_prefabSystemComponentInterface->FindTemplate(focusedInstance->get().GetTemplateId());
```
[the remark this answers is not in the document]
srikappa-amzn: You are correct. This doesn't happen anymore. So I'll remove the comment. I think it is still a valid check to have since other use cases in the future might run into this. I'll add an assert as this shouldn't happen.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabSaveLoadHandler.h
michabr: OK. Perhaps we can change the comment describing the PrefabSaveLoadHandler class for now then.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabSaveLoadHandler.cpp
michabr: OK, fair enough if we're not anticipating too many templates loaded at a time.

michabr (approved):
Might be worth checking with UX on the wording of the dialog box.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabSaveLoadHandler.h
srikappa-amzn: I think the PrefabSaveLoadHandler file needs to be refactored a bit because the file name is PrefabSaveLoadHandler but the class name is PrefabSaveHandler. And its also not clear whats the difference between PrefabSaveLoadHandler and PrefabLoader. I'll do the refactoring as part of another PR

micronAMZN:
We were discussing this on another channel, here's a suggestion for text of the current solution:

Suggest changing title bar text to something more appropriate such as:

* Missing prefab file(s)
* Prefab file(s) not found

Suggest using Warning icon or Error icon if it’s warranted rather than the Question icon.

Dialog text suggestion:

`Assets\Prefabs\Bush.prefab` could not be located on disk, but X instance(s) of it exist in the level.
Would you like to delete the instances of the prefab from the level?

If you choose Cancel, you must take one of the following actions to avoid loss of prefab data:

* Restore the prefab file on disk. Select an instance and choose *Save prefab to file*.
* Detach the prefab instances.
* Delete the prefab instances

Button text suggestions:

* Delete
* Cancel

My understanding is that this solution is for the case where a prefab is in memory, instanced in an active level, and the prefab file is deleted from disk. For this scenario, I don't think the user should be prompted to delete the prefab instances. I suggest offering to attempt to save or detach prefabs to avoid data loss. Here's a mock-up.

missing-prefab-file

For the scenario where a level is being loaded and prefabs are missing, I suggest a warning and the option to either delete prefab instances or replace them with another prefab. Another mock-up:

loading-missing-prefab-file

srikappa-amzn:
> We were discussing this on another channel. I have a concern about prompting users to delete data by default and have another suggestion. First, here's a suggestion for text of the current solution:
> 
> delete-missing-prefab-file
> 
> Suggest changing title bar text to something more appropriate such as:
> 
> * Missing prefab file(s)
> * Prefab file(s) not found
> 
> Suggest using Warning icon or Error icon if it’s warranted rather than the Question icon.
> 
> Dialog text suggestion:
> 
> 'Assets\Prefabs\Bush.prefab' could not be located on disk, but N instance(s) of it exist in the level.
> 
> Would you like to delete the instances of the prefab from the level?
> 
> If you choose Cancel, you must take one of the following actions:
> 
> * Restore the prefab file on disk. Select an instance and choose **Save prefab to file**.
> * Detach the prefab instances.
> * Delete the prefab instances.
> 
> Button text suggestions:
> 
> * Delete
> * Cancel
> 
> My understanding is that this solution is for the case where a prefab is in memory, instanced in an active level, and the prefab file is deleted from disk. For this scenario, I don't think the user should be prompted to delete the prefab instances. I suggest offering to attempt to save or detach prefabs to avoid data loss. Here's a mock-up.
> 
> missing-prefab-file
> 
> For the scenario where a level is being loaded and prefabs are missing, I suggest a warning and the option to either delete prefab instances or replace them with another prefab. Another mock-up:
> 
> loading-missing-prefab-file

This PR doesn't handle the case of missing nested prefab use case. So, that needs to be handled in an other change. Regarding providing detach as the option instead of delete, here are my thoughts:

If users don't care about the deleted prefab, then users would need to find all those instances in the outliner and manually delete them after detaching. For example, what if the deleted prefab has 1000 entities in it and the prefab is under 20 other prefabs, based on the above approach, after detaching users would need to search for the entity by name and delete all of those entities manually. 

One other concern with using 'detach' as the default option is that a prefab can be made up of 100s of entities and other prefabs. So, detaching the prefabs and spilling all its children in various places in the outliner would be very jarring to the user imo. At least with delete, there's a direct correlation that the user can make : "Prefab x is deleted, 'n' instances of x are removed from the outliner/scene"


Plus we also need to consider the implications of detaching prefabs with nested prefabs. When a prefab containing nested prefabs is detached, the owning of nested prefabs get transferred the parent of the detached prefab. Consider this example:


- Level
  - car.prefab
    - engine entitty
    - body.prefab
  - car.prefab
    - engine entitty
    - body.prefab
  - car.prefab
    - engine entitty
    - body.prefab
  - car.prefab
    - engine entitty
    - body.prefab


Now, if car gets deleted on file and we detach, the new state would look like


- Level
  - car entity
    - engine entitty
    - body.prefab
  - car entity
    - engine entitty
    - body.prefab
  - car entity
    - engine entitty
    - body.prefab
  - car entity
    - engine entitty
    - body.prefab

But when you save that level to disk, your level file would look like
```
{
   "containerentity":{},
   "entities":{...},
   "instances":{
        "body1":{},
        "body2":{},
        "body3":{},
        "body4":{}
    }

}
```


Imagine, if you have 20 other prefabs under the car. Then you would be linking all of those to the level now that car is detached.
At least with delete, it is easier to explain why a certain level/prefab looks different now : "Oh, there are no cars in the level because cars were deleted"
But if you detach and add new entities and dependencies (nested prefabs) on the level, then that may not sit well with users. We'd be assuming too much and modifying user data based on those assumptions and adding dependencies on user files without them knowing

All of this is to say that in terms of both code maintainability and user file maintainability, detach would make things harder than delete.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabSaveLoadHandler.h
forkercat: It would be good to something else rather than `PrefabSaveLoadHandler` since it is handling multiple functionalities except save/load. Currently, it also handles drag-and-drop logic. Alternatively, we should break it down into different classes before it gets assigned more responsibility. 😀

srikappa-amzn:
After discussing with UX (@rainbj ) and docs (@micronAMZN ), we have reached a consensus that a toast followed by a warning message is a less intrusive and less destructive way to handle these hot reloading cases. I've updated the PR and description section with how the new toast dialog looks like. 

@forkercat and @michabr , let me know if you have any final comments/concerns. Otherwise, I'll be merging this once AR goes green.

forkercat (approved):
lgtm

PR #14198, opened 2023-01-24
Asset Processor - Intermediate UUID Metadata

amzn-mike:
## What does this PR do?

- Adds support to AzToolsFramework for generating a UUID and storing it in the metadata file.
- Updates UUID generation scheme for intermediate assets to be based on `HASH({SourceUUID}:{BuilderUUID}:{SubId})` if the direct source asset is enabled for metadata UUIDs.
- Stores intermediate asset UUID in metadata file alongside the intermediate asset.

Intermediates generated from sources which are not using metadata UUIDs yet will continue to use the path-based UUID system.

## How was this PR tested?

Manual testing.
Added and ran unit tests.


Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/MetadataManager.cpp
```diff
@@ -24,6 +25,8 @@ namespace AzToolsFramework
             // Note the version here is not the same as the MetadataFile version, since this is a separate serialization
             serializeContext->Class<MetadataManager, AZ::Component>()->Version(1);
         }
+
+        UuidEntry::Reflect(context);
```
AMZN-stankowi, line 29: Why are you calling UuidEntry::Reflect from within MetadataManager::Reflect, instead of in the function that calls MetadataManager::Reflect?

Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/UuidUtils.cpp
```diff
@@ -0,0 +1,56 @@
...
+        {
+            AZ_Assert(metadataInterface, "Programmer Error - IMetadataRequests interface is not available");
+            return false;
+        }
+
+        AzToolsFramework::UuidEntry entry;
+
+        if (metadataInterface->GetValue(absoluteFilePath, UuidKey, entry) && !entry.m_uuid.IsNull())
+        {
+            return false;
```
AMZN-stankowi, line 37: Should there be a warning, error, or log here, or is it that handled in the call to GetValue? The site calling CreateSourceUuid probably can't handle it because it won't know if the false return is due to GetValue failing, the metadataInterface not being available, or SetValue failing.

Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/UuidUtils.cpp
```diff
@@ -0,0 +1,56 @@
...
+    AZ::Uuid UuidUtilComponent::CreateSourceUuid(AZ::IO::PathView absoluteFilePath)
+    {
+        auto uuid = AZ::Uuid::CreateRandom();
+
+        if(CreateSourceUuid(absoluteFilePath, uuid))
+        {
+            return uuid;
+        }
+
+        return {};
```
AMZN-stankowi, line 54: Should there be an error or warning here, if the call above failed?

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
```diff
@@ -4089,6 +4090,27 @@ namespace AssetProcessor
                         newJob.m_scanFolder = scanFolder;
                         newJob.m_checkServer = jobDescriptor.m_checkServer;
 
+                        auto* uuidInterface = AZ::Interface<AssetProcessor::IUuidRequests>::Get();
+
+                        AZ_Assert(uuidInterface, "Programmer Error - IUuidRequests interface is not available.");
```
AMZN-stankowi, line 4095: Earlier in this PR in UuidUtilComponent::CreateSourceUuid, when an interface wasn't available, you checked for null and returned, here you aren't. I think checking for null + returning is usually safer than just crashing.

AMZN-stankowi (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/UuidUtils.cpp
junbo75, line 37: Curious: Not quite sure how we want to handle this case. Do we want to override the old value and throw a warning/error here?

Code/Tools/AssetProcessor/native/AssetManager/ProductAsset.cpp
```diff
@@ -63,6 +64,9 @@ namespace AssetProcessor
             wasRemoved = AZ::IO::SystemFile::Delete(m_absolutePath.c_str());
         }
 
+        // Try to delete the metadata file too if one exists
```
junbo75, line 67: Nit: May worth checking for the return value to make sure that the file is deleted successfully.

Code/Tools/AssetProcessor/native/resourcecompiler/rcjob.cpp
```diff
@@ -828,9 +831,28 @@ namespace AssetProcessor
...
+                    // The assumption for the UUID generated below is that the source UUID will not change.  A type which has no metadata
+                    // file currently may be updated later to have a metadata file, which would break that assumption.  In that case, stick
+                    // with the default path-based UUID.
+                    if (!params.m_topLevelSourceUuid.IsNull())
+                    {
+                        // Generate a UUID for the intermediate as:
+                        // SourceUuid:BuilderUuid:SubId
+                        auto uuid = AZ::Uuid::CreateName(AZStd::string::format(
+                            "%s:%s:%d",
+                            params.m_topLevelSourceUuid.ToFixedString().c_str(),
```
junbo75, line 847: Should we use the top level source UUID or the UUID of the direct source here? For example, consider a processing chain like this: source asset A -> intermediate asset B -> intermediate asset C -> product asset B. The UUID of intermediate asset C should be SourceUuidOfA:FirstBuilderUuid:SecondBuilderUuid:SubIdOfC instead of SourceUuidOfA:SecondBuilderUuid:SubIdOfC.

Code/Tools/AssetProcessor/native/resourcecompiler/rcjob.cpp
```diff
@@ -876,6 +898,20 @@ namespace AssetProcessor
...
+        if (!uuidInterface)
+        {
+            AZ_Assert(false, "Programmer Error - IUuidUtil interface is not available");
+            return false;
+        }
+
+        // Go through all the intermediate products and output the assigned UUID
+        for (auto [intermediateProduct, uuid] : intermediateOutputPaths)
+        {
+            uuidInterface->CreateSourceUuid(intermediateProduct.toUtf8().constData(), uuid);
```
junbo75, line 912: Nit: May worth checking the return value here for potential errors?

Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/MetadataManager.cpp
amzn-mike: I guess it doesn't make sense to have it here, but this reflect is called as part of the component setup stuff (GetRequiredComponents) so I can't reflect non-components there.  But this is required by the UuidUtilComponent so Ill move the reflect into that call instead

Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/UuidUtils.cpp
amzn-mike: I have another PR coming up which will overhaul the error handling

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
amzn-mike: Updated

Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/UuidEntry.h
```diff
@@ -0,0 +1,30 @@
...
+ *
+ */
+
+#pragma once
+
+#include <AzCore/Serialization/SerializeContext.h>
+
+namespace AzToolsFramework
+{
+    struct UuidEntry
```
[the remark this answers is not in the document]
amzn-mike: Added

Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/UuidUtils.cpp
amzn-mike: I think it's best to not overwrite the old values because there shouldn't be a use case for programmatically changing the UUID since that'll break any references.

Code/Tools/AssetProcessor/native/AssetManager/ProductAsset.cpp
amzn-mike: Added

Code/Tools/AssetProcessor/native/resourcecompiler/rcjob.cpp
amzn-mike: Added

Code/Tools/AssetProcessor/native/resourcecompiler/rcjob.cpp
amzn-mike: If we include the UUIDs of all the builders in the chain it would make it difficult for users (asset builders) to generate this UUID.  Having the last builder + subID should be easy enough as those should be knowable.  As for the UUID of the top level source vs the direct source... I originally thought the top level would be easier to figure out for a builder trying to create a reference, but now I'm questioning that.  We might need to spend some time to figure out which is best here

amzn-mike: Ok so I spent some time thinking about this and I think using the top level source is not what we want here.   Here's the scenario I'm thinking about:

- FBX file generates material intermediate
- TXT file generates FBX file intermediate which generates material intermediate

Python file is attached to FBX file and wants to set up a reference to the material file.  If the ID is generated from the top level source, it will be difficult for the python file to find out the TXT file asset ID.  If it's generated from the FBX file (the direct source) it should be easy to look up and calculate.

Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/UuidEntry.h
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+#pragma once
+
+#include <AzCore/Serialization/SerializeContext.h>
+
+namespace AzToolsFramework
+{
+    //! Structure used to store UUID information for an asset in the metadata file
+    struct UuidEntry
```
AMZN-alexpete, line 16: suggestion: call this `MetaUuid` instead of `UuidEntry` so it is more clear when devs search for the term "uuid" that this is for meta files specifically

Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/UuidUtils.h
```diff
@@ -0,0 +1,50 @@
...
+
+#pragma once
+
+#include <AzCore/Component/Component.h>
+#include <AzCore/Interface/Interface.h>
+#include <AzCore/Serialization/SerializeContext.h>
+
+namespace AzToolsFramework
+{
+    struct IUuidUtil
```
AMZN-alexpete, line 17: first time looking at this so pardon my ignorance - why make a separate interace + component for this class instead of adding these functions to the existing `IMetadataRequests` interface and `MetadataManager`?

Code/Framework/AzToolsFramework/Tests/Metadata/UuidUtilTests.cpp
```diff
@@ -0,0 +1,215 @@
...
+#include <AzCore/UnitTest/Mocks/MockFileIOBase.h>
+#include <AzCore/Component/ComponentApplication.h>
+#include <AzCore/Utils/Utils.h>
+#include <AzToolsFramework/Metadata/UuidUtils.h>
+
+namespace UnitTest
+{
+    struct UuidUtilTests
+        : LeakDetectionFixture
+        , AZ::ComponentApplicationBus::Handler
```
AMZN-alexpete, line 23: could this be simpler if you used `UnitTest::MockComponentApplication` instead - like is used [here](https://github.com/o3de/o3de/blob/development/Code/Editor/Lib/Tests/test_AzAssetBrowserRequestHandler.cpp#L136)?

Code/Framework/AzToolsFramework/Tests/Metadata/UuidUtilTests.cpp
```diff
@@ -0,0 +1,215 @@
...
+        AZStd::unordered_map<AZ::IO::HandleType, AZStd::string> m_mockFiles;
+        AzToolsFramework::MetadataManager m_manager;
+        AzToolsFramework::UuidUtilComponent m_uuidUtil;
+    };
+
+    TEST_F(UuidUtilTests, CreateSourceUuid_Random_ReturnsRandomUuid)
+    {
+        auto* utilInterface = AZ::Interface<AzToolsFramework::IUuidUtil>::Get();
+
+        ASSERT_TRUE(utilInterface);
```
AMZN-alexpete, line 174: this happens in every test - consider moving it to Setup() and making it a member

Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/UuidEntry.h
amzn-mike: Good suggestion, I'll update that

Code/Framework/AzToolsFramework/AzToolsFramework/Metadata/UuidUtils.h
amzn-mike: IMetadataRequests is meant to handle metadata files in a generic way, it isn't supposed to have any UUID-specific function (it can be used to store any kind of data meant to be associated with an asset).  UuidUtil is meant to build on top of the metadata system to provide an API for assigning UUIDs to an asset.

Code/Framework/AzToolsFramework/Tests/Metadata/UuidUtilTests.cpp
amzn-mike: I didn't know that existed, thanks for pointing that out, Ill update to use that instead

Code/Framework/AzToolsFramework/Tests/Metadata/UuidUtilTests.cpp
amzn-mike: Will do.

AMZN-alexpete (approved):

PR #14199, opened 2023-01-24
Add picking for PhysX colliders in the viewport

amzn-ahmadkrm:

## What does this PR do?
This PR adds the ability to pick entities that have just a collider, i.e. without a mesh or shape attached. Previously if entities did not have these and just a collider it could not be picked.

This PR also fixes an issue where an entity could not be picked if it had a collider and a rigid body component together.

## How was this PR tested?

Created unit tests for the changes made:.
Before:
VsDebugConsole_CNJCFDQYsm
After:
VsDebugConsole_V8Mna6FVoZ


Gems/PhysX/Code/Source/EditorColliderComponent.cpp
```diff
@@ -1587,6 +1589,34 @@ namespace PhysX
...
+    bool EditorColliderComponent::EditorSelectionIntersectRayViewport(
+        [[maybe_unused]] const AzFramework::ViewportInfo& viewportInfo, const AZ::Vector3& src, const AZ::Vector3& dir, float& distance)
+    {
+        AzPhysics::RayCastRequest request;
+        request.m_direction = dir;
+        request.m_distance = distance;
+        request.m_start = src;
+
+        auto hit = RayCast(request);
+        if (hit)
```
hultonha, line 1611: uber nit: Could use an if initializer here 😛

hultonha (commented):
This change looks pretty great to me! 😅 Does this fix all cases of picking colliders (asset/box/capsule/sphere)? If so I'd say move this out of draft! 🙂 Would be good to just hear a bit more about how it was tested. Cool we can just reuse the `RayCast` function 🥳 (tagging @greerdv for visibility)

Gems/PhysX/Code/Tests/EditorColliderTests.cpp
```diff
@@ -0,0 +1,74 @@
...
+        struct TestData
+        {
+            const AZStd::vector<AZ::Vector2> polygonHShape = { AZ::Vector2(0.0f, 0.0f), AZ::Vector2(0.0f, 3.0f), AZ::Vector2(1.0f, 3.0f),
+                                                               AZ::Vector2(1.0f, 2.0f), AZ::Vector2(2.0f, 2.0f), AZ::Vector2(2.0f, 3.0f),
+                                                               AZ::Vector2(3.0f, 3.0f), AZ::Vector2(3.0f, 0.0f), AZ::Vector2(2.0f, 0.0f),
+                                                               AZ::Vector2(2.0f, 1.0f), AZ::Vector2(1.0f, 1.0f), AZ::Vector2(1.0f, 0.0f) };
+        };
+    } // namespace
+
+    TEST_F(PhysXEditorFixture, EditorShapeColliderComponent_ShapeDependencySatisfied_EntityIsValid)
```
hultonha, line 46: update: Looks like the test name needs to be updated 😉 Looks like a great start though! Give me or @greerdv/@jonawals a ping when this is ready for review 👍

greerdv:
I'm guessing the test is still a bit WIP, but everything else looks good to me.

Code/Framework/AzToolsFramework/Tests/Viewport/ViewportUiManagerTests.h
```diff
@@ -0,0 +1,99 @@
...
+    using ButtonId = AzToolsFramework::ViewportUi::ButtonId;
+
+    // child class of ViewportUiManager which exposes the protected button group and viewport display
+    class ViewportUiManagerTestable : public AzToolsFramework::ViewportUi::ViewportUiManager
+    {
+    public:
+        ViewportUiManagerTestable() = default;
+        ~ViewportUiManagerTestable() override = default;
+
+        const AZStd::unordered_map<AzToolsFramework::ViewportUi::ClusterId, AZStd::shared_ptr<ButtonGroup>>& GetClusterMap()
```
hultonha, line 33: nit: This function can also be made `const` as it returns a `const` reference

Gems/PhysX/Code/Source/EditorColliderComponent.cpp
```diff
@@ -457,6 +458,7 @@ namespace PhysX
         AZ::Data::AssetBus::Handler::BusDisconnect();
         m_nonUniformScaleChangedHandler.Disconnect();
         AzFramework::BoundsRequestBus::Handler::BusDisconnect();
+        AzToolsFramework::EditorComponentSelectionRequestsBus::Handler::BusDisconnect();
```
hultonha, line 461: uber nit: It's always good to connect/disconnect in the reverse order so I'd move this just above the `BoundRequestBus` disconnect

Code/Framework/AzToolsFramework/Tests/Viewport/ViewportUiManagerTests.h
```diff
@@ -0,0 +1,99 @@
...
+
+        ViewportUiManagerTestable* GetViewportManager()
+        {
+            return m_viewportManager.get();
+        }
+
+        QWidget* GetMockRenderOverlay()
+        {
+            return m_mockRenderOverlay.get();
+        }
```
moraaar, line 72: Suggestion: Please move the implementations to the `ViewportUiManagerTests.cpp`

Gems/PhysX/Code/Tests/EditorColliderComponentTests.cpp
```diff
@@ -0,0 +1,138 @@
...
+         auto clickPos1 = AzFramework::WorldToScreen(AZ::Vector3(7.5f, 12.4f, 10.0f), m_cameraState);
+         auto selectedEntities = ClickAndGetSelectedEntities(clickPos1);
+
+         EXPECT_THAT(selectedEntities.size(), testing::Eq(0));
+
+         // Then when a user clicks inside the collider it should be selected
+         auto clickPos2 = AzFramework::WorldToScreen(AZ::Vector3(7.5f, 12.6f, 10.0f), m_cameraState);
+         selectedEntities = ClickAndGetSelectedEntities(clickPos2);
+
+         EXPECT_THAT(selectedEntities.size(), testing::Eq(1));
```
hultonha, line 73: suggestion: Might be worth checking the entity id itself definitely matches too 🙂 (repeated)

Code/Framework/AzToolsFramework/Tests/Viewport/ViewportUiManagerTests.h
```diff
@@ -0,0 +1,99 @@
...
+
+        void SetUp() override
+        {
+            m_viewportManagerWrapper.Create();
+        }
+
+        void TearDown() override
+        {
+            m_viewportManagerWrapper.Destroy();
+        }
```
moraaar, line 97: Minor: Inside a class block first go the functions and last the members. Please reorder. 

Also, does `m_viewportManagerWrapper` need to be public? Can it be private or protected?

Gems/PhysX/Code/Tests/PhysXColliderComponentModeTests.h
```diff
@@ -0,0 +1,90 @@
...
+#include <AzToolsFramework/ViewportSelection/EditorDefaultSelection.h>
+#include <AzToolsFramework/ViewportSelection/EditorInteractionSystemViewportSelectionRequestBus.h>
+#include <AzToolsFramework/ViewportSelection/EditorSelectionUtil.h>
+#include <AzToolsFramework/ViewportUi/ViewportUiManager.h>
+#include <EditorColliderComponent.h>
+#include <Tests/Viewport/ViewportUiManagerTests.h>
+
+namespace UnitTest
+{
+    class PhysXColliderComponentModeTest : public ToolsApplicationFixture<false>
```
hultonha, line 29: nit: Could you add a quick comment for what the template is for please

Gems/PhysX/Code/Tests/PhysXColliderComponentModeTests.h
```diff
@@ -0,0 +1,90 @@
...
+    {
+    protected:
+        using EntityPtr = AZ::Entity*;
```
hultonha, line 32: nit: probably okay to just remove this and use `AZ::Entity*` directly

hultonha (approved):
This set of changes look excellent @amzn-ahmadkrm, great work on adding the tests too, I just have a few minor suggestions. Awesome to see (and thanks for the great description too! 👍)

Gems/PhysX/Code/Tests/PhysXColliderComponentModeTests.h
```diff
@@ -0,0 +1,90 @@
...
+
+        AZ::ComponentId m_colliderComponentId;
```
moraaar, line 34: Same: Members at the end after the functions.

Gems/PhysX/Code/Tests/PhysXColliderComponentModeTests.h
```diff
@@ -0,0 +1,90 @@
...
+
+            entity->Activate();
+
+            AzToolsFramework::SelectEntity(entityId);
+
+            return entity;
+        }
+
+        // Needed to support ViewportUi request calls.
+        ViewportManagerWrapper m_viewportManagerWrapper;
```
moraaar, line 56: Same: members at the end after the functions.

Gems/PhysX/Code/Tests/PhysXColliderComponentModeTests.h
```diff
@@ -0,0 +1,90 @@
...
+
+        void SetUpEditorFixtureImpl() override
+        {
+            m_viewportManagerWrapper.Create();
+        }
+        void TearDownEditorFixtureImpl() override
+        {
+            m_viewportManagerWrapper.Destroy();
+        }
```
moraaar, line 65: Please move implementation to `PhysXColliderComponentModeTests.cpp`

Gems/PhysX/Code/Tests/PhysXColliderComponentModeTests.h
```diff
@@ -0,0 +1,90 @@
...
+        void SetupTransform(const AZ::Quaternion& rotation, const AZ::Vector3& translation, float uniformScale);
+        void SetupCollider(
+            const Physics::ShapeConfiguration& shapeConfiguration,
+            const AZ::Quaternion& colliderRotation,
+            const AZ::Vector3& colliderOffset);
+        void SetupNonUniformScale(const AZ::Vector3& nonUniformScale);
+        void EnterColliderSubMode(PhysX::ColliderComponentModeRequests::SubMode subMode);
+
+        AZ::Entity* m_entity = nullptr;
+        AZ::EntityComponentIdPair m_idPair;
```
moraaar, line 85: Are these class members needed to be public?

Gems/PhysX/Code/Tests/EditorColliderComponentTests.cpp
```diff
@@ -0,0 +1,138 @@
...
+    public:
+        void SetUpEditorFixtureImpl() override;
+        //! Clicks at a given position and returns the entities that are selected.
+        AzToolsFramework::EntityIdList ClickAndGetSelectedEntities(AzFramework::ScreenPoint screenPoint);
+
+        const float m_uniformScale = 1.0f;
+        const AZ::Quaternion m_shapeRotation = AZ::Quaternion(0.0f, 0, 0, 1);
+        const AZ::Quaternion m_entityRotation = AZ::Quaternion(0, 0, 0, 1);
+        const AZ::Vector3 m_shapeOffset = AZ::Vector3(0, 0, 0);
+        const AZ::Vector3 m_entityTranslation = AZ::Vector3(5.0f, 15.0f, 10.0f);
```
moraaar, line 25: Are these needed to be public?

It's also odd that they are all `const`, why is that? Do they change per instance or can they be static?

Gems/PhysX/Code/Source/EditorRigidBodyComponent.cpp
```diff
@@ -308,6 +308,7 @@ namespace PhysX
         AZ::TransformNotificationBus::Handler::BusConnect(entityId);
         Physics::ColliderComponentEventBus::Handler::BusConnect(entityId);
         AzFramework::BoundsRequestBus::Handler::BusConnect(entityId);
+        AzToolsFramework::EditorComponentSelectionRequestsBus::Handler::BusConnect(entityId);
```
moraaar, line 311: Both `EditorColliderComponent` and `EditorRigidBodyComponent` connect to the `zToolsFramework::EditorComponentSelectionRequestsBus` using the same entity id. An entity can have both components, so both components are handling the bus for the same entity. Is that intentional? How that does work when a caller calls a EditorComponentSelectionRequestsBus function for `entityId`? How is it controller which one responds? Does it need an entity-component pair so the caller can make sure to call the one it ones?

Code/Framework/AzToolsFramework/Tests/Viewport/ViewportUiManagerTests.h
amzn-ahmadkrm: I moved it to a header because `ComponentModeTests.h` was including it as a cpp and I wanted to to include `ComponentModeTests.h` somewhere else but it wasn't compiling

Gems/PhysX/Code/Tests/PhysXColliderComponentModeTests.h
amzn-ahmadkrm: what would be the best way to include this file as a cpp?

Gems/PhysX/Code/Tests/PhysXColliderComponentModeTests.h
amzn-ahmadkrm: I didn't make this file, just moved the the class into a header, not too sure what the template is for

Gems/PhysX/Code/Source/EditorColliderComponent.cpp
amzn-ahmadkrm: good spot! I thought I did but misplaced it by a line :P

Gems/PhysX/Code/Tests/PhysXColliderComponentModeTests.h
hultonha: Just move the implementation -  I would argue these smaller functions are okay to leave in the grand scheme of things, but maybe the larger ones, leave the interface here, and then put the body in the .cpp

```
// .h

...
void SetUpEditorFixtureImpl() override;

// .cpp

void <typename>::SetUpEditorFixtureImpl()
{
    // implementation
}
```

Gems/PhysX/Code/Source/EditorRigidBodyComponent.cpp
amzn-ahmadkrm: It is intentional, the bus doesn't receive any calls if it is only connected to one of the components strangely which I can investigate in a follow up PR

Code/Framework/AzToolsFramework/Tests/Viewport/ViewportUiManagerTests.h
moraaar: It should work to do

`ViewportUiManagerTests.h`
````
        void Create();
        void Destroy();
````

`ViewportUiManagerTests.cpp`
````
        void ViewportManagerWrapper::Create()
        {
            m_viewportManager = AZStd::make_unique<ViewportUiManagerTestable>();
            m_viewportManager->ConnectViewportUiBus(AzToolsFramework::ViewportUi::DefaultViewportId);
            m_mockRenderOverlay = AZStd::make_unique<QWidget>();
            m_parentWidget = AZStd::make_unique<QWidget>();
            m_viewportManager->InitializeViewportUi(m_parentWidget.get(), m_mockRenderOverlay.get());
        }

        void ViewportManagerWrapper::Destroy()
        {
            m_viewportManager->DisconnectViewportUiBus();
            m_viewportManager.reset();
            m_mockRenderOverlay.reset();
            m_parentWidget.reset();
        }
````

If that doesn't work then it'd need to be investigated, there might be some other issue somewhere.

File: Gems/PhysX/Code/Tests/EditorColliderComponentTests.cpp [hunk 16]
```diff
@@ -0,0 +1,142 @@
...
+#include <Tests/PhysXColliderComponentmodeTests.h>
+#include <Source/EditorRigidBodyComponent.h>
+
+namespace UnitTest
+{
+    static const float uniformScale = 1.0f;
+    static const AZ::Quaternion shapeRotation = AZ::Quaternion(0.0f, 0, 0, 1);
+    static const AZ::Quaternion entityRotation = AZ::Quaternion(0, 0, 0, 1);
+    static const AZ::Vector3 shapeOffset = AZ::Vector3(0, 0, 0);
+    static const AZ::Vector3 entityTranslation = AZ::Vector3(5.0f, 15.0f, 10.0f);
```
moraaar, line 18: [standard] global const variables start with upper case: 

`uniformScale` > `UniformScale`
`shapeRotation` > `ShapeRotation`
etc.

Gems/PhysX/Code/Source/EditorRigidBodyComponent.cpp
hultonha: This is actually okay but requires the caller to use `AZ::EBusReduceResult` (see `CalculateEntityLocalBoundsUnion` as an example - this is what we do with components with world bounds as multiple components might provide an Aabb of different sizes and we need to return the aggregate/union of the bounds. So there's nothing technically incorrect with this approach.

hultonha: Usage looks like this:

```c++
AZ::EBusReduceResult<AZ::Aabb, AabbUnionAggregator> aabbResult(AZ::Aabb::CreateNull());
BoundsRequestBus::EventResult(aabbResult, entity->GetId(), &BoundsRequestBus::Events::GetLocalBounds);
```

moraaar: That's cool, I wasn't aware of that. Thanks for sharing @hultonha !

Code/Framework/AzToolsFramework/Tests/Viewport/ViewportUiManagerTests.h
```diff
@@ -0,0 +1,100 @@
...
+            m_viewportManager->DisconnectViewportUiBus();
+            m_viewportManager.reset();
+            m_mockRenderOverlay.reset();
+            m_parentWidget.reset();
+        }
+
+        ViewportUiManagerTestable* GetViewportManager()
+        {
+            return m_viewportManager.get();
+        }
```
spham-amzn, line 67: This is exposing a pointer to an internal protected member of the base class, is this necessary? Is this necessary?

Code/Framework/AzToolsFramework/Tests/Viewport/ViewportUiManagerTests.h
amzn-ahmadkrm: oh my bad I misinterpreted it as move the whole thing back to the cpp, not the implementation!

Gems/PhysX/Code/Source/EditorRigidBodyComponent.cpp
amzn-ahmadkrm: This is actually already implemented in `PickEntity` in `AzToolsFramework/ViewportSelection/EditorSelectionUtil.cpp`

moraaar (approved):

Gems/PhysX/Code/Source/EditorRigidBodyComponent.cpp
hultonha: Cool no problem @moraaar and thanks for the info @amzn-ahmadkrm 👍

Gems/PhysX/Code/Tests/PhysXColliderComponentModeTests.h
hultonha: Hmm okay no worries, I'll have a check later 😝 (it's possible @jonawals might know)

Gems/PhysX/Code/Tests/EditorColliderComponentTests.cpp
hultonha: Thanks!

hultonha:
Temporarily moving to Draft while the author is away - will merge on their return

PR #14201, opened 2023-01-24
Material Canvas: Restoring graph model caching and minor graph canvas optimizations

gadams3:
## What does this PR do?

Material canvas does extensive traversing and sorting of nodes for populating the inspector and ordering instructions to generate code. Most accessor functions require looking up data on the graph object, which contains all of the node and connection data, which can be expensive.

This PR reimplements caching that saves parent nodes, connections, and data for some other recursive functions needed during sorting. The data is also protected using mutexes because a subsequent change makes material canvas sort and generate data asynchronously.

There are also minor changes to graph canvas to reserve space and avoid copying data in some places. During a stress test that loads a large graph of hundreds of nodes, it looks like most of the time is going to looking up and applying styles to graph canvas objects. Looking up matching styles is expensive because it copies containers of selector objects that are cloned and reallocated every time the search is done per object. This could be optimized by creating a set of functions that pass in the entity ID to be compared instead of copying the containers.

## How was this PR tested?

Tested loading and editing large graph with hundreds of nodes

### DIFF
--- a/Gems/GraphCanvas/Code/Source/Components/StylingComponent.cpp
+++ b/Gems/GraphCanvas/Code/Source/Components/StylingComponent.cpp
@@ -137,13 +137,14 @@ namespace GraphCanvas
     {
         Styling::SelectorVector selectors = m_coreSelectors;
 
-        QGraphicsItem* root = nullptr;
-        SceneMemberUIRequestBus::EventResult(root, GetEntityId(), &SceneMemberUIRequests::GetRootGraphicsItem);
-
+        selectors.reserve(selectors.size() + m_dynamicSelectors.size());
         for (auto& mapPair : m_dynamicSelectors)
         {
             selectors.emplace_back(mapPair.second);
         }
+
+        QGraphicsItem* root = nullptr;
+        SceneMemberUIRequestBus::EventResult(root, GetEntityId(), &SceneMemberUIRequests::GetRootGraphicsItem);
 
         if (!root)
         {
@@ -170,10 +171,11 @@ namespace GraphCanvas
 
         // TODO collapsed and highlighted
 
-        for (auto& mapPair : m_dynamicSelectors)
-        {
-            selectors.emplace_back(mapPair.second);
-        }
+        // Why is this repeated?
+        //for (auto& mapPair : m_dynamicSelectors)
+        //{
+        //    selectors.emplace_back(mapPair.second);
+        //}
 
         return selectors;
     }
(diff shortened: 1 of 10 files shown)


Gems/GraphCanvas/Code/Source/Components/StylingComponent.cpp
```diff
@@ -170,10 +171,11 @@ namespace GraphCanvas
...
 
-        for (auto& mapPair : m_dynamicSelectors)
-        {
-            selectors.emplace_back(mapPair.second);
-        }
+        // Why is this repeated?
+        //for (auto& mapPair : m_dynamicSelectors)
+        //{
+        //    selectors.emplace_back(mapPair.second);
+        //}
```
cgalvan, line 178: Seems like this should be safe to delete

Gems/GraphModel/Code/Source/Model/Node.cpp
```diff
@@ -339,45 +358,47 @@ namespace GraphModel
...
-            return slot->GetSlotDirection() == GraphModel::SlotDirection::Output && !slot->GetConnections().empty();
+            AZ_Assert(slot->GetSlotDirection() == GraphModel::SlotDirection::Output, "Slots in this container must be output slots.");
+            return !slot->GetConnections().empty();
         });
     }
 
     bool Node::HasInputConnectionFromNode(ConstNodePtr node) const
     {
-        return AZStd::any_of(m_allSlots.begin(), m_allSlots.end(), [&](const auto& slotPair) {
+         return AZStd::any_of(m_inputDataSlots.begin(), m_inputDataSlots.end(), [&](const auto& slotPair) {
```
cgalvan, line 379: Nit:  Is this an extra whitespace or was it the wrong tab spacing before?

cgalvan (approved):

Gems/GraphCanvas/Code/Source/Components/StylingComponent.cpp
```diff
@@ -137,14 +137,15 @@ namespace GraphCanvas
     {
         Styling::SelectorVector selectors = m_coreSelectors;
 
-        QGraphicsItem* root = nullptr;
-        SceneMemberUIRequestBus::EventResult(root, GetEntityId(), &SceneMemberUIRequests::GetRootGraphicsItem);
-
+        selectors.reserve(selectors.size() + m_dynamicSelectors.size());
```
jiaweig-amzn, line 140: There are other selectors after dynamic ones. Reserve the exact size may trigger resizing at:
`selectors.emplace_back(m_hoveredSelector);`
Not sure how many other selectors there are, maybe reserve space for all of them?

jiaweig-amzn (approved):

phuchau1989 (approved):

Issue #13847, opened 2022-12-28
Material Canvas crashes when materialtype is being generated


LB-KonradJuszczak:
**Describe the bug**
Material Canvas crashes when materialtype is being generated. The issue is inconsistent but occurs more often at the beginning of the Material Canvas session and when the Float input nodes are used.
Please note that the issue occurs with DX12 and Vulkan.

**Steps to reproduce**
1. Launch Material Canvas.
2. Add Float2 input node.
3. Add Experimental PBR node.
4. Connect color pins.

**Expected behavior**
Material Canvas remains stable and the material is generated without issues.

**Actual behavior**
Material Canvas crashes.

**Screenshots/Video**

[Call_Stack.txt](https://github.com/o3de/o3de/files/10313875/Call_Stack.txt)

**Found in Branch**
Development
**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
(https://github.com/o3de/o3de/commit/55856fbb15081d199b0d0b666882bb5edec339e6)

**Desktop/Device:**
- Device: PC
- OS: Windows
- Version: 10
- CPU: AMD Ryzen 5 3600
- GPU: NVIDIA GeForce RTX 2060 SUPERS
- Memory: 16 GB

LB-KacperMetelski:
Issue verified as fixed in commit 0a66664.

Issue #13840, opened 2022-12-27
Material Canvas: Material Canvas crashes if connected input node's text field is filled with number as first character


LB-BartoszDudzinski:
**Describe the bug**
The Material Canvas crashes after filling, with a number as a first character, a text field of any valid and connected to the output node.

**Steps to reproduce**
1. Launch the Material Canvas.
2. Add the Bool Input and the Experimental PBR nodes to the graph.
3. Connect them.
4. In the "Name" text field type any number.
5. Press LMB somewhere outside of the node.

**Expected behavior**
The Editor remains stable.

**Actual behavior**
The Editor crashes.

**Call stack trace**
```
MaterialCanvas.exe!AZStd::basic_string<char,AZStd::char_traits<char>,AZStd::allocator>::assign(const AZStd::basic_string<char,AZStd::char_traits<char>,AZStd::allocator> & rhs, unsigned __int64 rhsOffset, unsigned __int64 count) Line 378	C++
 	[Inline Frame] MaterialCanvas.exe!AZStd::basic_string<char,AZStd::char_traits<char>,AZStd::allocator>::assign(const AZStd::basic_string<char,AZStd::char_traits<char>,AZStd::allocator> &) Line 328	C++
 	[Inline Frame] MaterialCanvas.exe!AZStd::basic_string<char,AZStd::char_traits<char>,AZStd::allocator>::operator=(const AZStd::basic_string<char,AZStd::char_traits<char>,AZStd::allocator> &) Line 200	C++
 	MaterialCanvas.exe!MaterialCanvas::MaterialGraphCompiler::BuildMaterialTypeFromTemplate(AZStd::shared_ptr<GraphModel::Node const> templateNode, const AZStd::vector<AZStd::shared_ptr<GraphModel::Node const>,AZStd::allocator> & instructionNodes, const AZStd::basic_string<char,AZStd::char_traits<char>,AZStd::allocator> & templateInputPath, const AZStd::basic_string<char,AZStd::char_traits<char>,AZStd::allocator> & templateOutputPath) Line 1040	C++
 	MaterialCanvas.exe!MaterialCanvas::MaterialGraphCompiler::CompileGraph() Line 292	C++
 	MaterialCanvas.exe!AtomToolsFramework::GraphCompiler::OnSystemTick() Line 116	C++
 	[Inline Frame] MaterialCanvas.exe!AZStd::Internal::INVOKE(void(AZ::SystemTickEvents::*)() &) Line 177	C++
[...]

PR #14206, opened 2023-01-25
Material Canvas: Using jobs to parallelize work sorting and compiling graph data

gadams3:
## What does this PR do?

- This change is intended to help improve load times, iteration times, and make the editing experience smoother by changing graph operations that block the main thread to use jobs.
- Reversed direction with some previous changes that completely split graph compilation from the graph document class. Implementation of graph compiler is still separate but the graph document can use a factory function to create and manage the lifetime of a compiler. This is much easier to manage and understand.
- The graph compiler can be used on a graph without needing the document object or system. For the graph compiler to be able to operate independently on a worker thread, a copy of the graph and data is fed into it when initiating a new compile.
- Notifications sent from the worker thread are queued and sent on tick.
- Resolves other crashes with better error handling

Resolves https://github.com/o3de/o3de/issues/13847
Resolves https://github.com/o3de/o3de/issues/13840

## How was this PR tested?

Still testing to determine the cost or benefit of using algorithms like parallel for each.


gadams3:
Material canvas load times and usability are impacted by allocator logging. Changes in this PR https://github.com/o3de/o3de/pull/14213 disable that recording for the main editor and launcher. Those changes will need to be mirrored in material canvas, shader management console, and other tools. While testing, disabling the allocator recording resulted in an additional 4X performance gain and eliminated all of the noticeable stalls.

galibzon (approved):

amzn-mike (approved):
EditorAssetSystemAPI changes look good to me

PR #14213, opened 2023-01-25
Disable memory tracking by default in the Launcher and Editor

burelc-amzn:
This disables the memory tracking by default in the Editor and the Launcher. Tracking memory in those scenarios is too expensive to be on by default.


lawsonamzn (approved):

lawsonamzn:
LGTM - I'd definitely recommend breakpoint testing though, I remember last time I looked in here disabling them was harder than expected since some things used to inherit defaults not directly connected to the create struct.

Code/Editor/EditorToolsApplication.cpp
```diff
@@ -105,7 +105,17 @@ namespace EditorInternal
...
         // Must be done before creating QApplication, otherwise asserts when we alloc
-        AzToolsFramework::ToolsApplication::Start({}, params);
+        AzToolsFramework::ToolsApplication::Start({
+                /* .m_useExistingAllocator = */ false,
+                /* .m_allocationRecordsSaveNames = */ false,
+                /* .m_allocationRecordsAttemptDecodeImmediately = */ false,
+                /* .m_autoIntegrityCheck = */ false,
+                /* .m_markUnallocatedMemory = */ false,
+                /* .m_doNotUsePools = */ false,
+                /* .m_enableScriptReflection = */ false,
```
mbalfour-amzn, line 115: Should "enableScriptReflection" be enabled?

burelc-amzn: Fixed.

mbalfour-amzn (approved):

spham-amzn (approved):

cgalvan (approved):

gadams3:
We also need to make these changes to other tools, especially material canvas and shader management console. I believe this is affecting material canvas load times because graph canvas does a lot of inefficient container copying and reallocating while creating graph UI and applying style sheets. If it needs to be done in that many places should the default settings change?

https://github.com/o3de/o3de/blob/development/Gems/Atom/Tools/MaterialCanvas/Code/Source/main.cpp
https://github.com/o3de/o3de/blob/development/Gems/Atom/Tools/MaterialCanvas/Code/Source/main.cpp
https://github.com/o3de/o3de/blob/development/Gems/Atom/Tools/PassCanvas/Code/Source/main.cpp
https://github.com/o3de/o3de/blob/development/Gems/Atom/Tools/ShaderManagementConsole/Code/Source/main.cpp

lawsonamzn:
It may be easier to change the systemallocator.cpp and osallocator.cpp's constructors to not auto-enable profiling ?

gadams3:
> It may be easier to change the systemallocator.cpp and osallocator.cpp's constructors to not auto-enable profiling ? That would then affect all projects.

That's my recommendation. Disable it by default. Disabling the setting locally resulted in a ~4X improvement in the cases I was testing. But it also identified things that need to be optimized in graph canvas that shouldn't be excessively cloning containers of smart pointers while applying style sheets on graph objects.

lawsonamzn:
> > It may be easier to change the systemallocator.cpp and osallocator.cpp's constructors to not auto-enable profiling ? That would then affect all projects.
> 
> That's my recommendation. Disable it by default. Disabling the setting locally resulted in a ~4X improvement in the cases I was testing. But it also identified things that need to be optimized in graph canvas that shouldn't be excessively cloning containers of smart pointers while applying style sheets on graph objects.

yes, I just mean that its a smaller change to just cut those lines out
https://github.com/o3de/o3de/blob/560e148d6bd5dc221cf6598f3ed033d7a1c4a0fb/Code/Framework/AzCore/AzCore/Memory/SystemAllocator.cpp#L32
and the same kind of thing in OSAllocator, than the original change proposed here.

Then only those would need to change.

lawsonamzn:
Did not mean to close :)

SergeyAMZN (approved):

burelc-amzn:
I'll open a new PR for the revised approach, based on the suggestions here to disable tracking in the OSAllocator and SystemAllocator by default.

PR #14216, opened 2023-01-25
Add more comments to Enum.h to provide extra usage hints.

mbalfour-amzn:
## What does this PR do?

Expands the comments in Enum.h to include info about using AZ_TYPE_INFO_SPECIALIZE as well. This just helps make it more explicit that using the AZ_ENUM* macros make the enums serializable and provides helpers for counting and printing, but doesn't give them unique UUIDs. To get unique UUIDs, you also need to use AZ_TYPE_INFO_SPECIALIZE.

## How was this PR tested?

Recompiled and ran. It's just comment changes, so there's nothing really to test.

### DIFF
--- a/Code/Framework/AzCore/AzCore/Preprocessor/Enum.h
+++ b/Code/Framework/AzCore/AzCore/Preprocessor/Enum.h
@@ -31,6 +31,20 @@
 //!             Ex. AZ_CLASS_ENUM(E, (a, 1), (b, 2), (c, 2) )
 //!             Invoking either ToString(E::b) or ToString(E::c) will result in the string "b" being returned
 //! limitation: maximum of 125 enumerators
+//!
+//! By default, all enums declared with these macros will be assigned a default UUID due to the default enum specialization in TypeInfo.h.
+//! To create a unique UUID, you must use the AZ_TYPE_INFO_SPECIALIZE macro as well. This macro must be declared in the AZ namespace.
+//! Ex:
+//! namespace MyNamespace
+//! {
+//!     AZ_ENUM(MyCoolEnum, A, B);
+//! }
+//! 
+//! namespace AZ
+//! {
+//!     AZ_TYPE_INFO_SPECIALIZE(MyNamespace::MyCoolEnum, "{<SomeGuid>}");
+//! }
+//! 
 #define AZ_ENUM(EnumTypeName, ...)                                                      MAKE_REFLECTABLE_ENUM_UNSCOPED(EnumTypeName, __VA_ARGS__)
 
 //! Generate a decorated enumeration as AZ_ENUM, but with a specific underlying type.


lumberyard-employee-dm (approved):
The comments are good.

Alternatively, we could add a combined `AZ_ENUM_*` + `AZ_TYPE_INFO_SPECIALIZE` macro that performs the declaration of the actual Enum type, adds the helper functions for reflection and introspection and specializes the `AzTypeInfo` for the Enum

We add a macro such as 
```c++
//! Combines AZ_ENUM_CLASS_WITH_UNDERLYING_TYPE for declaring an enum +
//! AZ_TYPE_INFO for associating an Uuid with the C++ enum type
//! NOTE: Macro can only be used at global namespace scope
#define AZ_ENUM_CLASS_WITH_UNDERLYING_TYPE_AND_TYPEINFO(EnumTypeName, EnumUnderlyingType, EnumTypeUuid, CPlusPlusNamespace, ... ) \
namespace CPlusPlusNamespace \
{ \
    AZ_ENUM_CLASS_WITH_UNDERLYING_TYPE(EnumTypeName, EnumUnderlyingType, __VA__ARGS); \
} \
namespace AZ \
{ \
    AZ_TYPE_INFO_SPECIALIZE(CPlusPlusNamespace::EnumTypeName, EnumTypeUuid); \
}
```

gadams3 (approved):

amzn-ahmadkrm (approved):

mbalfour-amzn:
Discussed further with @lumberyard-employee-dm offline. I'll skip the extra macros for now, they warrant further thought since they bring up more issues:
* It's probably not great to have macros that put all enums into a single global namespace
* It would be nice if AZ_TYPE_INFO_SPECIALIZE didn't require that it was in the AZ namespace
* It might make sense to remove the partial specialization for enum types so that all reflected enums are _required_ to have unique UUIDs or else cause compile errors.

Just for a bit more context on that last point, the reason I'm adding these comments is because I was tracking down the cause of the following runtime error:
```
BehaviorContext: Trace::Error
 D:/github/o3de/Code/Framework/AzCore/AzCore/RTTI/BehaviorContext.cpp(86): 'bool __cdecl AZ::ValidateAzEventDescription(const class AZ::BehaviorContext &,const class AZ::BehaviorMethod &)'
BehaviorContext: AzEventDescription only contains names for 1 parameters, while the AZ::Event(RPC_OnEffect Notify Event) accepts 0 parameters
```

This error came about because the RPC used an AZ::Event<MyEnum>, which had the same RTTI result as AZ::Event<> due to MyEnum having the default UUID of 0, which led to ValidateAzEventDescription detecting the event as having no parameters instead of 1 parameter.

If instead MyEnum caused a compile-time error due to a missing type ID, the root cause would have been more obvious.

cgalvan (approved):

hultonha (approved):
This is awesome, thanks for adding this @mbalfour-amzn 😄

PR #14218, opened 2023-01-25
Performance improvement: Only adding non-pivot nodes to the bone list so they are checked once instead of against every anim.

AMZN-stankowi:

## What does this PR do?

I was profiling a slow to load animation in the scene settings tool. I saw IsPivotNode was a hot spot. Looking at the callstack, it was re-checking if the bone name was not a pivot node for every bone, for every animation to load.

This change has it only add non-pivot bones to the list to check, so it's only checked once during import.

This reduced the load time in the scene settings tool for this file from 10 seconds to 5 seconds.

Without this change:
20% of the time is spent in IsPivotNode.
image

With this change:
2.4% of the time is spent in IsPivotNode.
image

## How was this PR tested?

Manually, with the profiler in Visual Studio.


mbalfour-amzn (approved):

Code/Tools/SceneAPI/SceneBuilder/Importers/AssImpAnimationImporter.cpp
```diff
@@ -457,7 +457,10 @@ namespace AZ
                     {
                         aiBone* bone = mesh->mBones[boneIndex];
 
-                        boneList.insert(bone->mName.C_Str());
+                        if (!IsPivotNode(bone->mName))
```
lmbr-pip, line 460: nit: any danger the bone has a null ptr for a name?

lmbr-pip (approved):

jackalbe (approved):

Code/Tools/SceneAPI/SceneBuilder/Importers/AssImpAnimationImporter.cpp
AMZN-stankowi: I've added some null checks + error messages

cgalvan (approved):

amzn-mike (approved):

PR #14227, opened 2023-01-26
Turn off profiling by default for the System and OS allocators

burelc-amzn:
The purpose of enabling allocation profiling by default for these
allocators is to capture allocations that are made before `main()` starts.
However, allocation tracking gets too expensive for normal use, so it needs
to be opt-in for code paths that expect it (like the tests).

The tests must now explicitly turn on and off tracking. This means that
global static variables that are made by the test modules will be omitted
from the leak detection, because those allocations happen before the test
can enable tracking.

The AllocatorManager gains a new boolean flag to control if newly-created
allocators should have allocation tracking enabled by default. It is off by
default, and the tests turn it on.

Branched off of #14213.


galibzon (approved):

mbalfour-amzn (approved):

Code/Framework/AzCore/AzCore/Memory/SimpleSchemaAllocator.h
```diff
@@ -53,7 +52,14 @@ namespace AZ
...
         AllocatorDebugConfig GetDebugConfig() override
         {
-            return AllocatorDebugConfig();
+            return {
+#if defined(__cpp_designated_initializers)
+                .m_excludeFromDebugging = !ProfileAllocations,
+#else
+                /* .m_stackRecordLevels = */ 0,
+                /* .m_excludeFromDebugging = */ !ProfileAllocations,
+#endif
```
hultonha, line 61: question: Is it really worth adding this define now without us supporting C++20 yet? I agree they're cool and it would be great if we could use them but in the vast majority of cases we cannot

hultonha (approved):

lawsonamzn:
This kinda seems excessive?  As far as I understand it:

```
allocationRecords = (true/false) 
```
Controls whether or not to create and reserve the memory for the book keeping records that are necessary to track memory.
* False:  we do not create any allocation records structures and cannot track memory no matter what the value of the other variables are.
* True:  its possible to track memory, but this doesn't control whether we do so or not.

```
recordingMode = xxxx
```
Controls the type of records made, if records are made.
Record_no_records turns the system off and makes it so that no allocations are tracked even if the system is told to do so.  Other values control how much information is recorded - FULL means always capture stack, for example.  
This does not control whether or not to track memory - just what to track, when tracking is on.  If allocationRecords is false, there is no place to put the records, so this would have no effect.

```SetProfilingEnabled(true/false)```
This is the main on/off switch for tracking.  If this is false, no records are produced regardless of the other two options.  If this is true, it requires the other two options to be suitable (allocationRecords must be true, and recording mode must not be "do not make records").

Assuming my understanding of the above is correct (I could be wrong, correct me!), it seems like the only thing that needs to be done is to switch profiling mode to false, which it is by default, to stop using up cpu cycles and memory during allocations.

File: Code/Framework/AzCore/AzCore/Memory/SimpleSchemaAllocator.h (hunk 1 above)
Comment (burelc-amzn) [reply Code/Framework/AzCore/AzCore/Memory/SimpleSchemaAllocator.h:61]: I just want to say that I want to set the `m_excludeFromDebugging` member differently from its default value. In C++17 I have no way to express that. With this, when compiled with C++20, it will get the default value for `m_stackRecordLevels` from the member's declaration.

Comment (lumberyard-employee-dm) [reply Code/Framework/AzCore/AzCore/Memory/SimpleSchemaAllocator.h:61]: You could just set after it is assigned to a variable...

File: Code/Framework/AzCore/AzCore/Memory/OSAllocator.cpp [hunk 2]
```diff
@@ -14,9 +14,6 @@ namespace AZ
     {
         Create();
         PostCreate();
-#if defined(AZ_ENABLE_TRACING)
-        SetProfilingActive(true);
-#endif
```
lumberyard-employee-dm, line 19: The code for the non-allocating [Windows INI-style config parser](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Settings/ConfigParser.h#L139-L149) is in development.
We could just have this be set in a settings file that gets defaults this to false.

burelc-amzn:
Re: https://github.com/o3de/o3de/pull/14227#issuecomment-1405237252 @lawsonamzn 

Yes, we don't need to remove the `SetProfilingEnabled()` method in order to remove the `SetProfilingActive()` calls in the `OSAllocator` and `SystemAllocator` (but we do also need the calls to enable tracking in the test fixtures as well). But I find the current options to be redundant. I have to set `SetProfilingEnabled(true)` and `records->SetMode(something other than RECORD_NO_RECORDS)`. Why have two calls to turn on allocation tracking, when everywhere that we enable it, we call both? It seems to me that `RECORD_NO_RECORDS` doubles as `SetProfilingEnabled(false)`. But I'm fine with putting it back, if it's better to have separate controls over what type of tracking is done from if tracking is enabled.

To note though, the allocator's `AllocationRecords` instance is made depending on its `DebugConfig`, independent of those other options:
https://github.com/o3de/o3de/blob/cca0eb1b4e5b820f64c9b68655e206f9e74afd48/Code/Framework/AzCore/AzCore/Memory/AllocatorBase.cpp#L176-L181

Code/Framework/AzCore/AzCore/Memory/OSAllocator.cpp
lumberyard-employee-dm: You could use the following logic to read in a profiling enabled setting without hitting dynamic memory.
```c++
        AZ::IO::SystemFileStream configStream("<PathToConfigFile>", AZ::IO::OpenMode::ModeRead);
        bool setProfilingEnabled = false;
        auto parseConfigEntry = [&setProfilingEnabled](const AZ::Settings::ConfigParserSettings::ConfigEntry& configEntry)
        {
            constexpr bool caseSensitiveCmp = false;
            if (AZ::StringFunc::Equal(configEntry.m_keyValuePair.m_key, "profilingenabled", caseSensitiveCmp))
            {
                setProfilingEnabled = true;
            }

            return true;
        };

        AZ::Settings::ConfigParserSettings parserSettings{ parseConfigEntry };

        [[maybe_unused]] auto parseOutcome = AZ::Settings::ParseConfigFile(configStream, parserSettings);
        AZ_Assert(parseOutcome, "Error parsing Config File %s:\n%s", m_configStream.GetFilename(), parseOutcome.error().c_str());
```

lawsonamzn:
> Re: [#14227 (comment)](https://github.com/o3de/o3de/pull/14227#issuecomment-1405237252) @lawsonamzn
> 
> Yes, we don't need to remove the `SetProfilingEnabled()` method in order to remove the `SetProfilingActive()` calls in the `OSAllocator` and `SystemAllocator` (but we do also need the calls to enable tracking in the test fixtures as well). But I find the current options to be redundant. I have to set `SetProfilingEnabled(true)` and `records->SetMode(something other than RECORD_NO_RECORDS)`. Why have two calls to turn on allocation tracking, when everywhere that we enable it, we call both? It seems to me that `RECORD_NO_RECORDS` doubles as `SetProfilingEnabled(false)`. But I'm fine with putting it back, if it's better to have separate controls over what type of tracking is done from if tracking is enabled.
> 
> To note though, the allocator's `AllocationRecords` instance is made depending on its `DebugConfig`, independent of those other options:
> 
> https://github.com/o3de/o3de/blob/cca0eb1b4e5b820f64c9b68655e206f9e74afd48/Code/Framework/AzCore/AzCore/Memory/AllocatorBase.cpp#L176-L181

Yes.  note that the "profiling active" thing is fairly new historically.  The OG AZcore only had "records on or off" + "mode for the records if on".

IMO it would have been even better if there was just one setting (mode for the records, which includes off).  The thing is, some of this is compile time  and some of it is runtime.    

But we ship our engine as pre-built binaries.  We want people to be able to profile memory without recompiling the engine.  This means in the longer term we need our profile and debug builds to potentially have the *capability* to generate and store allocation records, but the spigot turned off until it is turned on (without a recompile).


For now, I think we should do the smallest possible ninja slice we can that will stop the slowdown in profile, and then release that.  Then figure out if there's a way to configure it so that the systems that are necessary to create and record records exist but the tracking isn't on by default.   This will help us not have to recompile for tests, too.

Code/Framework/AzCore/AzCore/Memory/OSAllocator.cpp
lumberyard-employee-dm: It should be possible to even load the config file at static initializing time as well.
Such as the first time the allocator is created on demand

burelc-amzn: I like this idea, but I'm going to leave it for a separate change.

lawsonamzn (approved):

Kadino (approved):
No testing concerns with opting out by default. Reducing overhead is a win. In general I think profiling markers should disabled unless the user explicitly opts in.

lawsonamzn (approved):
Looks like a comprehensive change here.

burelc-amzn:
> Looks like a comprehensive change here.

@lawsonamzn I did a rebase and force push, since the last time you approved it only commits b190da5c5a53627f9acd7a4e4955d508beed5d51 and 6a1c0fabfdbda9aa7fabd52301773b0af88634ec are new

Code/Framework/AzCore/Tests/Memory/AllocatorBenchmarks.cpp
```diff
@@ -259,6 +261,7 @@ namespace Benchmark
     public:
         void Benchmark(benchmark::State& state)
         {
+            m_allocator = AZStd::make_unique<TestAllocatorType>();
```
lumberyard-employee-dm, line 264: It seems that for consistency, this RecordedAllocationBenchmarkFixture class should override `SetUp` and `TearDown` and create the allocator the same way it is done as in the `AllocatorBenchmarkFixture`

Code/Framework/AzCore/AzCore/Memory/OSAllocator.cpp
lumberyard-employee-dm: Future Note: Alternatively, we can even read these a CFG file from the user home directory as well by using the [AZ::Utils::GetHomeDirectory](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Utils/Utils.h#L108-L111) method, as that can also be accessed during static initialization.

Code/Framework/AzCore/Tests/Memory/AllocatorBenchmarks.cpp
burelc-amzn: Fixed.

lawsonamzn (approved):

SergeyAMZN (approved):

PR #14244, opened 2023-01-26
Defer actor creation to the next tick to avoid creation during render ticks

mbalfour-amzn:
## What does this PR do?

When loading a level with EMFX actors in it, it's possible for the timing to work out where the EMFX actor is getting created while in the middle of a render tick. Specifically the callstack ends up looking roughly like the following (this has been truncated to show only the most relevant functions for this specific bug):

     Atom_RPI.Editor.dll!AZ::RPI::RPISystem::RenderTick() Line 317    C++
...
     Atom_RPI.Editor.dll!AZ::RPI::PassTree::BuildPasses() Line 76    C++
...
     Atom_RPI.Editor.dll!AZ::RPI::FullscreenTrianglePass::Create(const AZ::RPI::PassDescriptor & descriptor) Line 33    C++
...
     Atom_RPI.Editor.dll!AZ::RPI::FullscreenTrianglePass::LoadShader() Line 89    C++
     Atom_RPI.Editor.dll!AZ::RPI::FindShaderAsset(AZ::Data::AssetId shaderAssetId, const AZStd::basic_string<char,AZStd::char_traits<char>,AZStd::allocator> & shaderFilePath) Line 587    C++
     Atom_RPI.Editor.dll!AZ::Data::Asset<AZ::RPI::ShaderAsset>::BlockUntilLoadComplete() Line 1153    C++
     Atom_RPI.Editor.dll!AZ::Data::AssetManager::BlockUntilLoadComplete(const AZ::Data::Asset<AZ::Data::AssetData> & asset) Line 849    C++
...
     Atom_RPI.Editor.dll!AZ::Data::AssetManager::DispatchEvents() Line 591    C++
...
     EMotionFX.Editor.dll!EMotionFX::Integration::EditorActorComponent::OnAssetReady(AZ::Data::Asset<AZ::Data::AssetData> asset) Line 545    C++
...
     EMotionFX_Atom.Editor.dll!AZ::Render::AtomActorInstance::CreateRenderProxy(const AZStd::unordered_map<AZ::Render::MaterialAssignmentId,AZ::Render::MaterialAssignment,AZStd::hash<AZ::Render::MaterialAssignmentId>,AZStd::equal_to<AZ::Render::MaterialAssignmentId>,AZStd::allocator> & materials) Line 673    C++
...
>    Atom_Feature_Common.Editor.dll!AZ::Render::TransformServiceFeatureProcessor::ReserveObjectId() Line 175    C++

This leads to the error "Transform data cannot be written to during this phase".

The solution to this problem is to defer the creation of the Actor instance until the next OnTick() call. Unlike OnAssetReady(), OnTick() should never get called from within the render tick, so it should always be safe to create the instance there.

Longer-term, if Atom removes all uses of BlockUntilLoadComplete(), this problem should go away because OnAssetReady() would no longer get called in the middle of the render tick. If that ever occurs, this workaround could get removed.

## How was this PR tested?

Manually ran the scenario that was causing the error to trigger and verified that it no longer triggers.


amzn-rhhong:
What about the actor component (the runtime version of the component)? Won't there be the same problem?

mbalfour-amzn:
> What about the actor component (the runtime version of the component)? Won't there be the same problem?

I'll add it there too. I just don't have a solid repro for the runtime version of the problem to be able to 100% verify it's fixed.

cgalvan (approved):

amzn-rhhong (approved):
LGTM.

Gems/EMotionFX/Code/Source/Integration/Editor/Components/EditorActorComponent.cpp
```diff
@@ -541,7 +541,12 @@ namespace EMotionFX
             m_actorAsset = asset;
             AZ_Assert(m_actorAsset.IsReady() && m_actorAsset->GetActor(), "Actor asset should be loaded and actor valid.");
 
-            CheckActorCreation();
+            // We'll defer actor creation until the next tick on the tick bus. This is because OnAssetReady() can sometimes get
+            // triggered while in the middle of the render tick, since the rendering system sometimes contains blocking loads
+            // which will still process any pending OnAssetReady() commands while waiting. If that occurs, the actor creation
+            // would generate errors from trying to create a rendering actor while in the middle of processing the rendering data.
+            // We can avoid the problem by just always waiting until the next tick to create the actor.
+            m_processLoadedAsset = true;
```
AMZN-alexpete, line 549: Another option is to enqueue it on the tick bus rather than having a member var, but it would be bad if you enqueue it multiple times, so probably this is safer since I don't think we have an EnqueueOnce() function.

mbalfour-amzn: There are a couple of reasons not to do it that way:
1. This component literally already responds to OnTick(), so there's no value in enqueueing a second method onto the tick bus for the same component that could fire in either order. It's just more overhead and confusion for no gain.
2. (The main reason) If the component gets destroyed before the next tick, it's actually really hard to write a function that correctly does what it needs to for the component without getting tripped up by the lifetime issue, since it can't use a "this" pointer inside of it. On the flip side, as you can see in the PR, it's super _easy_ to get the lifetime management correct by having the component attach/unattach to the TickBus, because then it simply won't run the queued processing if the component gets destroyed before the queued work needs to happen.

AMZN-alexpete (approved):

moraaar (approved):
The change looks good. 

**Please check the EMotionFX unit tests still pass**, if there are unit tests checking the actor is correctly created then those tests might need to be updated to use the tick bus to be able to fully initialize the actor and check it's correctly created.

hultonha (approved):
Change looks good! (also a good reminder that it'd be great to refactor `ActorComponent` and `EditorActorComponent` to both use a shared `Actor` class that both the editor and runtime components can use... but that's a whole other story 😅

mbalfour-amzn:
> The change looks good.
> 
> **Please check the EMotionFX unit tests still pass**, if there are unit tests checking the actor is correctly created then those tests might need to be updated to use the tick bus to be able to fully initialize the actor and check it's correctly created.

Good call, one unit test broke. Fixed it. :)

PR #14247, opened 2023-01-27
Switch to using a persistent resource for the projected shadow map atlas

invertednormal:
## What does this PR do?

Cached shadows were implemented with the expectation that the shadow maps were persistent resources, however they were not. The fact that they've mostly worked thus far is surprising, but it _is_ possible for the cached shadows to be overwritten by other passes since they're transient resources. This PR updates the resource to be persistent so that the memory won't be stomped on by some other pass.

## How was this PR tested?

Tested using a test level with multiple cameras. Also tested in loft which uses many cached shadows.

### DIFF
--- a/Gems/Atom/Feature/Common/Code/Source/CoreLights/ProjectedShadowmapsPass.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/CoreLights/ProjectedShadowmapsPass.cpp
@@ -9,11 +9,14 @@
 #include <Atom/RHI/DrawListTagRegistry.h>
 #include <Atom/RHI/DrawPacketBuilder.h>
 #include <Atom/RHI/RHISystemInterface.h>
+#include <Atom/RPI.Public/Image/AttachmentImagePool.h>
+#include <Atom/RPI.Public/Image/ImageSystemInterface.h>
 #include <Atom/RPI.Public/Pass/PassAttachment.h>
 #include <Atom/RPI.Public/Shader/Shader.h>
 #include <Atom/RPI.Public/Scene.h>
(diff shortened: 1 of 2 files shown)


Gems/Atom/Feature/Common/Code/Source/CoreLights/ProjectedShadowmapsPass.cpp
```diff
@@ -212,22 +215,48 @@ namespace AZ
...
+            if (attachment->m_descriptor.m_type != RHI::AttachmentType::Image)
+            {
+                AZ_Assert(false, "[ProjectedShadowmapsPass %s] requires an image attachment", GetPathName().GetCStr());
+                return;
+            }
 
             RHI::ImageDescriptor& imageDescriptor = attachment->m_descriptor.m_image;
             const uint32_t shadowmapWidth = static_cast<uint32_t>(m_atlas.GetBaseShadowmapSize());
             imageDescriptor.m_size = RHI::Size(shadowmapWidth, shadowmapWidth, 1);
+            imageDescriptor.m_format = RHI::Format::D32_FLOAT;
```
VickyAtAZ, line 233: I guess the pass system wont read the format from .pass file if the attachment is imported

VickyAtAZ (approved):

Gems/Atom/Feature/Common/Code/Source/CoreLights/ProjectedShadowmapsPass.cpp
invertednormal: I decided to just move that to the cpp since most of the rest of the descriptor is defined there anyway.

rgba16f (approved):

moudgils (approved):

PR #14249, opened 2023-01-27
Extend ComboBox handlers to support Edit Button

amzn-ahmadkrm:
## What does this PR do?

This PR extends existing ComboBox handlers to support an "Edit Button" so that the `ComboBoxEditButtonPair` class can be removed as they only added the button and it was also duplicated in a couple of places. With this change `KinematicWidget` was removed as the logic for the edit button could be handed off directly to`PropertyBoolComboBoxCtrl`.

Affected components: 
  - Cloth
  - Collider
  - Rigid Body

These components relied on `ComboBoxEditButtonPair` as the UI handler but now use their default handlers, `PropertyStringComboBoxCtrl`, `PropertyEnumComboBoxCtrl` and `PropertyBoolComboBoxCtrl` respectively. 

Their behaviors should stay the same and have so far with manual testing. 

## How was this PR tested?

Manually tested. The  PhysX and NvCloth unit tests and their Editor counterpart tests were run and passed, and all the edit box functionalities behave as they did prior to the change.


Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyBoolComboBoxCtrl.cpp
```diff
@@ -71,14 +77,42 @@ namespace AzToolsFramework
...
+    }
+
+    void PropertyBoolComboBoxCtrl::OnEditButtonClicked()
+    {
+        if (m_editButtonCallback)
+        {
+            // intValue refers to the index value the comboBox will be set to on the return of the invoked callback
+            auto intValue = m_editButtonCallback->Invoke(nullptr, m_pComboBox->currentIndex());
+
+            if (intValue == 0 || intValue == 1)
```
hultonha, line 110: question: Is it possible to use an actual `bool` instead of `int` for this? It feels like it'd simplify things if possible

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyBoolComboBoxCtrl.hxx
```diff
@@ -52,7 +58,11 @@ namespace AzToolsFramework
         void onChildComboBoxValueChange(int value);
 
     private:
+        virtual void OnEditButtonClicked();
```
hultonha, line 61: question: Does this definitely need to be virtual? Are we expecting clients will need to override this and change the behavior?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyEnumComboBoxCtrl.hxx
```diff
@@ -58,7 +62,8 @@ namespace AzToolsFramework
         void onChildComboBoxValueChange(int comboBoxIndex);
 
     private:
-        QComboBox* m_pComboBox;
+        QComboBox* m_pComboBox = nullptr;
+        QToolButton* m_editButton = nullptr;
         AZStd::vector< AZStd::pair<AZ::s64, AZStd::string>  > m_enumValues;
```
hultonha, line 67: uber nit: Can we remove the white space around `> >` here (that's an old C++ thing when compilers used to think `>>` in this context was a shift operator 😅 Shows how old some of this code is! 😝)

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyEnumComboBoxCtrl.hxx
```diff
@@ -179,10 +184,10 @@ namespace AzToolsFramework
...
                 }
             }
-            else
-            {
-                GenericComboBoxHandler<ValueType>::ConsumeAttribute(GUI, attrib, attrValue, debugName);
-            }
+        else
+        {
+            GenericComboBoxHandler<ValueType>::ConsumeAttribute(GUI, attrib, attrValue, debugName);
+        }
```
hultonha, line 190: update: It looks like the indentation is off here a little bit

Gems/PhysX/Code/Source/EditorRigidBodyComponent.cpp
```diff
@@ -126,6 +127,30 @@ namespace PhysX
             &AzToolsFramework::EditorRequests::OpenViewPane, LyViewPane::PhysXConfigurationEditor);
     }
 
+    static int OnEditButtonClicked(int index)
```
hultonha, line 130: question: This change looks good but again was just curious if we can have this return a `bool`? Do we need the additional `-1` to detect it failed? I'm just wondering if we could simplify this or of there's a reason we have to use an `int`. Otherwise looks good. Thanks!

hultonha (commented):
Changes overall look really nice! Awesome to have streamlined this to use one shared widget. Would it be possible to add some screen shots of the updated widgets just to confirm they're all working as expected. It'd also be great to have @cgalvan and @moraaar review these changes. Great work! 🥳

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyBoolComboBoxCtrl.hxx
amzn-ahmadkrm: It doesn't need to be no, I thought that the user might use that to override the click logic but it doesn't necessarily work with how everything is set up

Gems/PhysX/Code/Source/EditorRigidBodyComponent.cpp
amzn-ahmadkrm: Changed to bool :)

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyBoolComboBoxCtrl.hxx
hultonha: Yeah I'd just remove it to remove confusion, this seems like the main thing you'd ever really want to do. Thanks!

Gems/PhysX/Code/Source/EditorRigidBodyComponent.cpp
hultonha: Awesome thanks!

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyBoolComboBoxCtrl.hxx
```diff
@@ -15,6 +15,7 @@
 #include <AzCore/base.h>
 #include <AzCore/Memory/SystemAllocator.h>
 #include <QtWidgets/QWidget>
+#include <QtWidgets/QToolButton>
```
cgalvan, line 18: Nit: This can just be `#include <QToolButton>`

cgalvan (approved):
Awesome, thanks for doing this!

Gems/NvCloth/Code/Source/Editor/MeshNodeHandler.h
```diff
@@ -54,8 +54,9 @@ namespace NvCloth
             bool ReadValuesIntoGUI(size_t index, widget_t* GUI, const property_t& instance, AzToolsFramework::InstanceDataNode* node) override;
 
         private:
-            void OnEditButtonClicked(widget_t* GUI);
+            AZ::EntityId m_entityId;
 
+            void OnEditButtonClicked();
             AZ::Data::Asset<AZ::Data::AssetData> GetMeshAsset(const AZ::EntityId entityId) const;
```
moraaar, line 60: nitpick: functions go first and last members.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyBoolComboBoxCtrl.hxx
```diff
@@ -42,6 +43,11 @@ namespace AzToolsFramework
         QWidget* GetLastInTabOrder();
         void UpdateTabOrder();
 
+        QComboBox* GetComboBox();
+        QToolButton* GetEditButton();
+
+        void SetEditButtonCallBack(AZ::Edit::AttributeFunction<bool(bool)>* function);
```
moraaar, line 49: nit: It'd be nice to use an alias for `AZ::Edit::AttributeFunction<bool(bool)>`, it'd make it more readable in the places it's used.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyBoolComboBoxCtrl.cpp
```diff
@@ -93,16 +123,53 @@ namespace AzToolsFramework
...
+            bool visible;
+            if (attrValue->Read<bool>(visible))
+            {
+                GUI->GetEditButton()->setVisible(visible);
+            }
+        }
+        else if (attrib == AZ_CRC_CE("SetTrueLabel"))
+        {
+            const char* label;
+            attrValue->Read<const char*>(label);
```
moraaar, line 139: Are you sure that `<const char*>` is valid usage for `Read`. There is no other place in the code using in like that, all the other places use `<AZStd::string>`. 

Is it safe to use `<const char*>`? Is the pointer that it returns still valid?

Same for `SetFalseLabel` and `EditButtonToolTip`.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyBoolComboBoxCtrl.cpp
```diff
@@ -93,16 +123,53 @@ namespace AzToolsFramework
...
+            GUI->GetComboBox()->setItemText(1, label);
+        }
+        else if (attrib == AZ_CRC_CE("SetFalseLabel"))
+        {
+            const char* label;
+            attrValue->Read<const char*>(label);
```
moraaar, line 145: Missing `if` check like the others?

````
if (attrValue->Read<const char*>(label))
{
    ....
}
````

Same for `SetTrueLabel`

**Review by moraaar — COMMENTED** [review]
Thanks for going the extra mile to implement this! I left some feedback.

Also, it'd be nice if the PR description would be explicit and say what classes this change affects (`PropertyBoolComboBoxCtrl`, `PropertyEnumComboBoxCtrl`, `PropertyStringComboBoxCtrl`).

Finally, `manual testing` is very vague and it doesn't give any information about what tests you did to validate this change:
- Were unit tests of NvCloth and PhysX (specially the Editor ones) built and run?
- What tests did you do to validate change in NvCloth? Did you test `MeshNodeHandler` in different scenarios when it's enabled/disabled (did the edit button behave correctly)?

File: Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyBoolComboBoxCtrl.hxx [hunk 11]
```diff
@@ -25,7 +25,7 @@ class QPushButton;
 
 namespace AzToolsFramework
 {
-    //just a test to see how it would work to pop a dialog
+    using ComboBoxSelectFunc = AZ::Edit::AttributeFunction<bool(bool)>*;
```
moraaar, line 28: To be more accurate with the naming please remove the pointer from the type:

````
using ComboBoxSelectFunc = AZ::Edit::AttributeFunction<bool(bool)>;
`````

This represents correctly it's the function type, then all the places where a pointer to the type is used please use `ComboBoxSelectFunc*`.

This is the same done in other places like in `PropertyAssetCtrl.hxx`:

````
        using EditCallbackType = AZ::Edit::AttributeFunction<void(const AZ::Data::AssetId&, const AZ::Data::AssetType&)>;
        using ClearCallbackType = AZ::Edit::AttributeFunction<void()>;

        EditCallbackType* m_editNotifyCallback = nullptr;
        ClearCallbackType* m_clearNotifyCallback = nullptr;
````

File: Gems/AtomLyIntegration/CommonFeatures/Assets/Objects/Shaderball/shaderball_default_1m.fbx.assetinfo [hunk 12]
```diff
@@ -0,0 +1,732 @@
+{
```
hultonha, line 1: update: It looks like this should be removed from this PR! 😅

hultonha (commented):
Updates look good, though it looks like one erroneous file has made it to the PR and should be removed. I +1 @moraaar's comment too. After those changes it should be good to go!

moraaar (approved):

Gems/PhysX/Code/Editor/CollisionGroupWidget.h
```diff
@@ -22,7 +21,7 @@ namespace PhysX
     {
         class CollisionGroupWidget
             : public QObject
-            , public AzToolsFramework::PropertyHandler<AzPhysics::CollisionGroups::Id, ComboBoxEditButtonPair>
+            , public AzToolsFramework::PropertyHandler<AzPhysics::CollisionGroups::Id, AzToolsFramework::PropertyEnumComboBoxCtrl>
```
amzn-tmryan, line 24: Would this work as a `AzToolsFramework::GenericComboBoxCtrl` instead?

Gems/PhysX/Code/Editor/CollisionLayerWidget.h
```diff
@@ -12,7 +12,7 @@
 #include <AzCore/Asset/AssetCommon.h>
 #include <AzToolsFramework/UI/PropertyEditor/PropertyEditorAPI.h>
 #include <AzFramework/Physics/Collision/CollisionLayers.h>
-#include <Editor/ComboBoxEditButtonPair.h>
+#include <AzToolsFramework/UI/PropertyEditor/PropertyEnumComboBoxCtrl.hxx>
```
amzn-tmryan, line 15: Same here. Could this be a `AzToolsFramework::GenericComboBoxCtrl`?

Gems/PhysX/Code/Editor/CollisionLayerWidget.h
```diff
@@ -21,7 +21,7 @@ namespace PhysX
     {
         class CollisionLayerWidget
             : public QObject
-            , public AzToolsFramework::PropertyHandler<AzPhysics::CollisionLayer, ComboBoxEditButtonPair>
+            , public AzToolsFramework::PropertyHandler<AzPhysics::CollisionLayer, AzToolsFramework::PropertyEnumComboBoxCtrl>
```
amzn-tmryan, line 24: Same. Could this be a `AzToolsFramework::GenericComboBoxCtrl`?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyBoolComboBoxCtrl.hxx
```diff
@@ -24,7 +25,7 @@ class QPushButton;
 
 namespace AzToolsFramework
 {
-    //just a test to see how it would work to pop a dialog
+    using ComboBoxSelectFunc = AZ::Edit::AttributeFunction<bool(bool)>;
```
lumberyard-employee-dm, line 28: We shouldn't be caching attributes locally.
Instead the value contained within the Attribute can be read from it.
Furthermore what should be read is an `AZStd::function<bool(bool>` from the Attribute as that would allow functors and lambdas to be used for the combo box select function.

amzn-tmryan (commented):
I have some questions/concerns about these changes. Sorry for being late to the party; I just saw this PR.

- How do you get the pairs added to the combobox through to the reflection system? I don't see where that's done; though, maybe I'm misunderstanding the point of editing the comboboxes.

- Could the changes to PropertyEnumComboBoxCtrl be made to GenericComboBoxCtrl instead? There is some overlap between the EnumValue and GenericValue attributes in the reflection system and I'm hoping they can be unified at some point in the future.

- Related to the point above, GenericValue is currently supported by DPE based editors instead of EnumValue. This is because GenericValue can cover any data that could be reflected with an EnumValue attribute. Using GenericValue could make this easier to support as we continue to transition editors to the DPE.

Please let me know if there were technical reasons for doing it the way you did, or if you have concerns.

lumberyard-employee-dm (changes requested):
The Edit Button changes aren't doing anything for the Enum and String Combo Box Handler types.
It also is missing for the [GenericCombo Box Ctrl](https://github.com/o3de/o3de/blob/development/Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/GenericComboBoxCtrl.h#L55-L96) which can work with any C++ type that supports a conversion to a string.

Finally conversions to the actual type for enums, AZStd::strings and Generic `T` types would require an Edit Button Callback that takes into account the case when the value **cannot** be converted to be stored in the underlying type.
For example if the user was trying to set  a value "-256" was set for an `int8_t` or `uint8_t` [Combo Box](https://github.com/o3de/o3de/blob/development/Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyEnumComboBoxCtrl.hxx#L205-L224), the callback function should be able to reject that that edit.

At the very least the callback function should return an `AZ::Outcome<T, ErrorString>` that allows the implementer to reject an edit and to provide reason why the rejection occurs.
For example the [IEventLogger API](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Metrics/IEventLogger.h#L368-L370) functions return an Outcome that provides an error string as to why the operation has failed.
That same pattern can be followed here.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyBoolComboBoxCtrl.hxx
lumberyard-employee-dm: Also API documentation as to what value the user can except to be passed into the function and the what the user should return needs to be added.
Such as
```c++
//! Type Alias for the Error String type that is used when an Edit is rejected
//! Uses AZStd::fixed_string to avoid dynamic memory allocations
using EditRejectString = AZStd::fixed_string<256>;

//! Outcome that returns the value that the ComboBox control should set on Success
//! On Failure a rejection string with the reason for failure is returned
using EditResultOutcome = AZ::Outcome<bool, EditRejectString>;

//! Callback to implement that determines the new value to update the Combo Box Ctrl with
//! or to provide the Combo Box Ctrl system with an error message as to why the new value could not be set
//! i.e "new value not convertible to type", "new value is out of range", etc...
//! @param currentValue The current value stored in the combo box
//! @return Outcome which should store the new value to set in the combo box control or a string supplying the reason why the edit has been rejected.
using EditButtonCallback = AZStd::function<EditResultOutcome(bool currentValue)>;
```

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/GenericComboBoxCtrl.inl
```diff
@@ -18,6 +18,10 @@ AZ_POP_DISABLE_WARNING
 #include <AzToolsFramework/UI/PropertyEditor/PropertyQTConstants.h>
 #include <AzToolsFramework/UI/PropertyEditor/DHQComboBox.hxx>
 
+
+#pragma optimize("", off)
```
kberg-amzn, line 22: Was this just for debugging?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyBoolComboBoxCtrl.cpp
```diff
@@ -13,73 +13,38 @@ AZ_PUSH_DISABLE_WARNING(4251, "-Wunknown-warning-option") // 4251: 'QLayoutItem:
 #include <QtWidgets/QHBoxLayout>
 AZ_POP_DISABLE_WARNING
 
-//just a test to see how it would work to pop a dialog
+#pragma optimize("", off)
```
kberg-amzn, line 16: Same here.. lots of pragma's that seem like you probably didn't mean to check them in.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/GenericComboBoxCtrl.inl
```diff
@@ -202,6 +216,39 @@ namespace AzToolsFramework
...
+    {
+        m_editButtonCallback = function;
+    }
+
+    template<typename T>
+    void GenericComboBoxCtrl<T>::OnEditButtonClicked()
+    {
+        if (m_editButtonCallback)
+        {
+            // A successful outcome returns a bool that determines the new value of the comboBox
```
lumberyard-employee-dm, line 242: The comment is not quite correct.
The successful outcome in this case returns the new value to set

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/GenericComboBoxCtrl.inl
```diff
@@ -202,6 +216,39 @@ namespace AzToolsFramework
...
+    template<typename T>
+    QToolButton* GenericComboBoxCtrl<T>::GetEditButton()
+    {
+        return m_editButton;
+    }
+
+    template<typename T>
+    void GenericComboBoxCtrl<T>::SetEditButtonCallBack(GenericEditButtonCallback<T> function)
+    {
+        m_editButtonCallback = function;
```
lumberyard-employee-dm, line 234: The function should be moved into the m_editButtonCallback to avoid a copy operation

`m_editButtonCallback = AZStd::move(function);`

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/GenericComboBoxCtrl.inl
```diff
@@ -363,4 +431,6 @@ namespace AzToolsFramework
             genericGUI->m_postChangeNotifyCB->Invoke(notifyInstance, oldValue);
         }
     }
-}
+} // namespace AzToolsFramework
+#pragma optimize("", on)
+#pragma inline_depth()
```
lumberyard-employee-dm, line 436: As kberg-amzn has mentioned, these pragmas need to be removed.

I believe the validator should catch these issues.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyStringComboBoxCtrl.cpp
```diff
@@ -12,170 +12,71 @@ AZ_PUSH_DISABLE_WARNING(4251, "-Wunknown-warning-option") // 'QLayoutItem::align
...
         for (size_t valIndex = 0; valIndex < vals.size(); valIndex++)
         {
-            m_values.push_back(vals[valIndex]);
-            if (vals[valIndex] != "---")
+            auto value = vals[valIndex];
+            if (value != "---")
             {
-                m_pComboBox->addItem(vals[valIndex].c_str());
+                const AZStd::pair<AZStd::string, AZStd::string> valueToAdd = AZStd::pair(value.c_str(), value.c_str());
+                ComboBoxBase::addElement(valueToAdd);
```
lumberyard-employee-dm, line 60: The `value.c_str()` is a anti-optimization that is unnecessary.
The vals vector is already an `AZStd::string`, no need to grab the `const char*` from it, where it would have to recalculate the string length to create an AZStd::string
```c++
const AZStd::pair<AZStd::string, AZStd::string> valueToAdd{ value, value };
```

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyBoolComboBoxCtrl.hxx
```diff
@@ -25,44 +26,38 @@ class QPushButton;
...
+    //! Outcome that returns the value that the ComboBox control should set on Success
+    //! On Failure a rejection string with the reason for failure is returned
+    using EditResultOutcome = AZ::Outcome<bool, EditRejectString>;
+
+    //! Callback to implement that determines the new value to update the Combo Box Ctrl with
+    //! or to provide the Combo Box Ctrl system with an error message as to why the new value could not be set
+    //! i.e "new value not convertible to type", "new value is out of range", etc...
+    //! @param currentValue The current value stored in the combo box
+    //! @return Outcome which should store the new value to set in the combo box control or a string supplying the reason why the edit has been rejected.
+    using EditButtonCallback = AZStd::function<EditResultOutcome(bool currentValue)>;
```
lumberyard-employee-dm, line 42: These could be aliased to the `Generic*` type aliases.

i.e
```c++
using EditRejectString = GenericEditRejectString;
using EditResultOutcome = GenericEditResultOutcome<bool>
using EditButtonCallback = GenericEditButtonCallback<bool>;
```

lumberyard-employee-dm: The same can be done to the PropertyStringComboBoxCtrl as well.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/GenericComboBoxCtrl.inl
amzn-ahmadkrm: Hey yes, sorry I should've changed it to draft I was still working on this!

lumberyard-employee-dm (approved):

amzn-tmryan (approved):

Gems/PhysX/Code/Editor/CollisionLayerWidget.h
amzn-ahmadkrm: Hey thank you for your feedback :) I moved everything to the GenericComboBoxCtrl

hultonha (approved):
Great work getting these changes in @amzn-ahmadkrm and thank you very much for your help @lumberyard-employee-dm 👍

PR #14251, opened 2023-01-27
Material Canvas: Normal input fixes, global settings to hide on node property editors, and other miscellaneous fixes

gadams3:
## What does this PR do?

-Fixed tangent space conversions to normal input values
-Added a button to the node config editor that displays the UUID and allows the user to generate a new one
-Added buttons to the node config editor that allows the user to select dynamic settings groups and add them to nodes and slots
-Added support for auto populating node and slot display names and descriptions if they are not explicitly defined in the config
-Added settings to globally toggle display and editing of slot values on node UI. Turning this off gives nodes a much cleaner look until all of the layout issues are resolved. 
-Fixed conversion from paths with aliases and relative paths to include statements.
-Changed the folder and category for all of the “object” nodes to “vertex” since they all currently wrap vertex attributes

## How was this PR tested?

Created a normal map test and confirmed that normal maps from input values or images are applied correctly if the option is enabled.
Confirmed that using the button to add node groups inserts categories for selected groups without overwriting existing ones.
Confirmed that using the reset UUID button generates a new UUID that is displayed in the inspector and saved with the node config.
Confirmed that if include path settings are active and include paths have been selected from the picker that they are inserted into the generated code correctly with relative paths.
Confirmed that settings to toggle displaying property editors on node UI does as expected.

image
image
image


cgalvan (approved):

cgalvan (approved):

Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/GraphData/Nodes/MaterialOutputs/BasePBR/MaterialGraphName_SurfaceEval.azsli
```diff
@@ -50,7 +51,7 @@ Surface EvaluateSurface_MaterialGraphName(
     Surface surface;
     surface.position = positionWS;
     surface.vertexNormal = vertexNormal;
-    surface.normal = normalize(o_normal_override_enabled ? (float3)inNormal : vertexNormal);
+    surface.normal = normalize(o_normal_override_enabled ? GetWorldSpaceNormal((float2)inNormal, vertexNormal, tangents[0], bitangents[0]) : vertexNormal);
```
jiaweig-amzn, line 54: Calling `GetWorldSpaceNormal` requires `inNormal` to be normalized (seems so in the pixel evaluation). It should be better to use `inNormal.xy` instead of casting.
I think you can directly use
`TangentSpaceToWorld(inNormal, vertexNormal, tangents[0], bitangents[0])`
So that no casting or normalization is needed.

jiaweig-amzn (approved):

Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/GraphData/Nodes/MaterialOutputs/BasePBR/MaterialGraphName_SurfaceEval.azsli
gadams3: Thank you! I have to cast instead of using .xy because the generated variable could end up being a different type, a scalar that does not have .xy.

PR #14261, opened 2023-01-27
Make it easier for users to understand static vs dynamic colliders

moraaar:
**Note**
This work has been divided into multiple PRs to make it easier to review. Merging each PR into the staging branch `PhysX_StaticRigidBody_Staging_o3de`.

**Other PRs related to this change**
- https://github.com/o3de/o3de/pull/14262
- https://github.com/o3de/o3de/pull/14264
- https://github.com/o3de/o3de/pull/14267
- https://github.com/o3de/o3de/pull/14268
- https://github.com/o3de/o3de/pull/14295

Signed-off-by: moraaar <moraaar@amazon.com>

## What does this PR do?

It's not clear to users when they can move an entity or not if there is a collider in it. There is an internal physics logic that is not explicitly said, which is "a collider with no rigid body is an static collider and the user should not try to move it either by code or script canvas". A usual mistake users do is to create an entity with just a collider and move it in script canvas, which is wrong, but there is nothing in the component that suggests that's wrong.

This PR adds the new editor component "Static Rigid Body" and makes Colliders require either an static or a dynamic rigid body. This way if the user chooses to add a static rigid body it's explicitly chosen to be static, meaning it should not move. If the customer chose to add dynamic rigid body, then it's understood it can be moved.

**List of modifications**
- Added new "PhysX Static Rigid Body" component to the editor and renamed "PhysX Rigid Body" to "PhysX Dynamic Rigid Body". "PhysicsRigidBodyService" now means any rigid body service, and in addition, each one will offer specific services "PhysicsStaticRigidBodyService" and "PhysicsDynamicRigidBodyService".
- Character Controller functions internally like a Kinematic Dynamic Rigid Body, so its services have been updated to reflect it.
- PhysX Colliders now require to add either a static or dynamic rigid body component.
- RigidBody, StaticRigidBody and CharacterController components now create the simulated bodies in `OnEntityActivated`, after all the collider components have created the shapes during activation.
- The new StaticRigidBody component (same as RigidBody) implements `RigidBodyRequests` and it sends `RigidBodyNotifications` to indicate when the rigid body is enabled/disabled. These notification will be used by other components that need to know when the simulated bodies are ready (https://github.com/o3de/o3de/pull/14262)
- Fixed bug where MeshCollider component might not load the asset in time, now blocking until it's loaded since it's required during activation to create the shapes.

**Out of Scope**
[...]

PR #14262, opened 2023-01-27
Adapt components that query simulated bodies

moraaar:
**Note**

## What does this PR do?

This PR adapts Joints, ForceRegion, NetworkRigidBody and AudioAreaEnvironment components to query for simulated bodies after the rigid body (static or dynamic) has created them by listening to `RigidBodyNotificationBus` notifications.

Joints initialization improved to only be enabled when all the rigid bodies involved in the joint are enabled.

**Components initialization order when colliders and rigid bodies are involved:**
1. Entity calls Component::Activate
    1. Rigid Body components connect to `EntityBus` to delay the creation of simulated body until `OnEntityActivated` (since shapes are not created yet)
    1. All the PhysX Collider components will create their shapes during component activation.
    1. Other components that were querying for simulated bodies during activation, they will now instead connect to `RigidBodyNotificationBus` bus (because the simulated bodies are not created yet).
1. Entity calls OnEntityActivated
    1. Rigid Body components will query all the shapes (that colliders produced during activation) and create simulated body with all the shapes
    1. Rigid Body components will enable physics on the simulate body and send a `RigidBodyNotification::OnPhysicsEnabled`.
    1. The components listening to the `RigidBodyNotificationBus` will receive the `OnPhysicsEnabled` notification and they will query the simulated bodies.

## How was this PR tested?

PhysX unit tests pass.
Multiplayer unit tests pass.

### DIFF
--- a/Gems/LmbrCentral/Code/Source/Audio/AudioAreaEnvironmentComponent.h
+++ b/Gems/LmbrCentral/Code/Source/Audio/AudioAreaEnvironmentComponent.h
@@ -9,6 +9,7 @@
 
 #include <AzCore/Component/Component.h>
 #include <AzCore/Component/TransformBus.h>
+#include <AzFramework/Physics/RigidBodyBus.h>
 #include <AzFramework/Physics/Common/PhysicsSimulatedBodyEvents.h>
 
 #include <IAudioSystem.h>
@@ -28,6 +29,7 @@ namespace LmbrCentral
      */
     class AudioAreaEnvironmentComponent
         : public AZ::Component
+        , protected Physics::RigidBodyNotificationBus::Handler
         , private AZ::TransformNotificationBus::MultiHandler
     {
         friend class EditorAudioAreaEnvironmentComponent;
@@ -69,6 +71,10 @@ namespace LmbrCentral
         }
 
         static void Reflect(AZ::ReflectContext* context);
+
+        // Physics::RigidBodyNotifications overrides...
+        void OnPhysicsEnabled(const AZ::EntityId& entityId) override;
+        void OnPhysicsDisabled(const AZ::EntityId& entityId) override;
 
     private:
         void OnTriggerEnter(const AzPhysics::TriggerEvent& triggerEvent);
--- a/Gems/LmbrCentral/Code/Source/Audio/AudioAreaEnvironmentComponent.h
+++ b/Gems/LmbrCentral/Code/Source/Audio/AudioAreaEnvironmentComponent.h
@@ -72,7 +72,7 @@ namespace LmbrCentral
 
         static void Reflect(AZ::ReflectContext* context);
 
-        // Physics::RigidBodyNotifications overrides...
+        // Physics::RigidBodyNotifications overrides ...
         void OnPhysicsEnabled(const AZ::EntityId& entityId) override;
         void OnPhysicsDisabled(const AZ::EntityId& entityId) override;
 
--- a/Gems/Multiplayer/Code/Include/Multiplayer/Components/NetworkRigidBodyComponent.h
+++ b/Gems/Multiplayer/Code/Include/Multiplayer/Components/NetworkRigidBodyComponent.h
@@ -47,7 +47,7 @@ namespace Multiplayer
         void OnDeactivate(Multiplayer::EntityIsMigrating entityIsMigrating) override;
 
     private:
-        // Physics::RigidBodyNotifications overrides...
+        // Physics::RigidBodyNotifications overrides ...
         void OnPhysicsEnabled(const AZ::EntityId& entityId) override;
 
         void OnTransformUpdate(const AZ::Transform& worldTm);
@@ -76,7 +76,7 @@ namespace Multiplayer
 #endif
 
     private:
-        // Physics::RigidBodyNotifications overrides...
+        // Physics::RigidBodyNotifications overrides ...
         void OnPhysicsEnabled(const AZ::EntityId& entityId) override;
 
         Physics::RigidBodyRequests* m_physicsRigidBodyComponent = nullptr;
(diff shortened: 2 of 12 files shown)


Gems/PhysX/Code/Source/JointComponent.cpp
```diff
@@ -90,28 +90,110 @@ namespace PhysX
...
         }
     }
 
-    void JointComponent::Deactivate()
+    void JointComponent::CreateNativeJoint()
     {
-        AZ::EntityBus::Handler::BusDisconnect();
+        if (m_jointHandle != AzPhysics::InvalidJointHandle)
+        {
+            return;
```
spham-amzn, line 176: Should this be an assertable condition? i.e. is it expected that this call is idempotent?

spham-amzn (approved):

Gems/LmbrCentral/Code/Source/Audio/AudioAreaEnvironmentComponent.h
```diff
@@ -70,6 +72,10 @@ namespace LmbrCentral
 
         static void Reflect(AZ::ReflectContext* context);
 
+        // Physics::RigidBodyNotifications overrides...
```
greerdv, line 75: nitpick - space before ...

Gems/Multiplayer/Code/Include/Multiplayer/Components/NetworkRigidBodyComponent.h
```diff
@@ -45,6 +47,9 @@ namespace Multiplayer
         void OnDeactivate(Multiplayer::EntityIsMigrating entityIsMigrating) override;
 
     private:
+        // Physics::RigidBodyNotifications overrides...
```
greerdv, line 50: nitpick - space (repeated in other places too)

Gems/Multiplayer/Code/Source/Components/NetworkRigidBodyComponent.cpp
```diff
@@ -119,11 +134,24 @@ namespace Multiplayer
...
+        m_physicsRigidBodyComponent = Physics::RigidBodyRequestBus::FindFirstHandler(entityId);
+        AZ_Assert(m_physicsRigidBodyComponent, "Physics Rigid Body is required on entity %s", GetEntity()->GetName().c_str());
+
+        m_physicsRigidBodyComponent->SetKinematic(false);
+
 #if AZ_TRAIT_SERVER
         if (IsNetEntityRoleAuthority())
         {
-            if (AzPhysics::RigidBody* rigidBody = GetParent().m_physicsRigidBodyComponent->GetRigidBody())
+            if (AzPhysics::RigidBody* rigidBody = m_physicsRigidBodyComponent->GetRigidBody())
```
greerdv, line 154: What was GetParent doing before and why is it no longer necessary?

Gems/PhysX/Code/Source/JointComponent.h
```diff
@@ -108,5 +116,9 @@ namespace PhysX
         JointMotorProperties m_motor;
         AzPhysics::JointHandle m_jointHandle = AzPhysics::InvalidJointHandle;
         AzPhysics::SceneHandle m_jointSceneOwner = AzPhysics::InvalidSceneHandle;
+
+        // Variable to keep track of the rigid bodies' entities involved in this joint
+        // and if they are enabled or disabled.
+        AZStd::unordered_map<AZ::EntityId, bool> m_rigidBodyEntityMap;
```
greerdv, line 122: unordered_map feels a little bit overkill to me for something that will contain either one or two entities.

Gems/PhysX/Code/Source/JointComponent.cpp
moraaar: In this case it's important that these `CreateNativeJoint/DestroyNativeJoint` functions validate if it's already created/destroyed the joint before doing anything because they come from physics bus calling enable/disable on the 1 or 2 entities that it relies on.

Technically speaking the `CreateNativeJoint` wouldn't need it, but the `DestroyNativeJoint` yes. But it's much better when both functions are symmetric and rely on the same logic (easier to understand the code), plus these functions are private, so they serve the purpose of internal logic, no need to answer to external users calling them.

Gems/Multiplayer/Code/Source/Components/NetworkRigidBodyComponent.cpp
moraaar: Good question. Unfortunately the controller cannot use the parent's `m_physicsRigidBodyComponent` member because when it's accessed it's not guaranteed it's been initialized yet (it's not guaranteed that `RigidBodyNotificationBus` will call the parent first and then the controller). So that's why now the controller has its own `m_physicsRigidBodyComponent` member.

Gems/PhysX/Code/Source/JointComponent.h
moraaar: I went for simplicity here since the map naturally contains the `key, element` pair needed for this case, there is no need to create additional struct or use a tuple (which could be more complicated to use), and since the code it's in a base class it won't be fixed to a number of elements, so it can scale without needing further changes (in case in the future whatever new child joint class wants/needs to depend on more entities to enable/disable the joint for whatever reason).

I could change it to a `AZStd::map` instead to be a tree and safe some memory since the look up will always be fast due to the low number of elements, would you prefer that?

Gems/Multiplayer/Code/Source/Components/NetworkRigidBodyComponent.cpp
greerdv: Cool, thanks for the explanation.

greerdv (approved):

PR #14264, opened 2023-01-27
Update Whitebox and Heightfield components' dependencies

moraaar:
**Note**

## What does this PR do?

Whitebox and Heightfield colliders handle their own collision and rigid body, so they have been explicitly made incompatible with other rigid bodies.

**Out of Scope**
- Now that there is a Static Rigid Body component, refactor Whitebox and Heightfield code so they don't have to handle their own rigid body, they user could instead add a rigid body component.

## How was this PR tested?

PhysX unit tests pass.

### DIFF
--- a/Gems/PhysX/Code/Source/HeightfieldColliderComponent.h
+++ b/Gems/PhysX/Code/Source/HeightfieldColliderComponent.h
@@ -30,12 +30,13 @@ namespace PhysX
 {
     class StaticRigidBody;
 
-    //! Component that provides a Heightfield Collider and associated Static Rigid Body.
+    //! Component that provides a Heightfield Collider.
+    //! It covers the static rigid body functionality as well, but it can be refactored out
+    //! once EditorStaticRigidBodyComponent handles the creation of the simulated body.
+    //! 
     //! The heightfield collider is a bit different from the other shape colliders in that it gets the heightfield data from a
     //! HeightfieldProvider, which can control position, rotation, size, and even change its data at runtime.
-    //! 
-    //! This component directly implements both the collider and static rigid body instead of
-    //! using BaseColliderComponent and StaticRigidBodyComponent.
+    //! Due to these differences, this component directly implements the collider instead of using BaseColliderComponent.
     class HeightfieldColliderComponent
         : public AZ::Component
         , public ColliderComponentRequestBus::Handler
--- a/Gems/PhysX/Code/Source/HeightfieldColliderComponent.h
+++ b/Gems/PhysX/Code/Source/HeightfieldColliderComponent.h
@@ -34,7 +34,7 @@ namespace PhysX
     //! The heightfield collider is a bit different from the other shape colliders in that it gets the heightfield data from a
     //! HeightfieldProvider, which can control position, rotation, size, and even change its data at runtime.
     //! 
-    //! Due to these differences, this component directly implements both the collider and static rigid body services instead of
+    //! This component directly implements both the collider and static rigid body instead of
     //! using BaseColliderComponent and StaticRigidBodyComponent.
     class HeightfieldColliderComponent
         : public AZ::Component
--- a/Gems/WhiteBox/Code/Source/Components/WhiteBoxColliderComponent.cpp
+++ b/Gems/WhiteBox/Code/Source/Components/WhiteBoxColliderComponent.cpp
@@ -33,9 +33,8 @@ namespace WhiteBox
         }
     }
 
-    void WhiteBoxColliderComponent::GetProvidedServices(AZ::ComponentDescriptor::DependencyArrayType& provided)
+    void WhiteBoxColliderComponent::GetProvidedServices([[maybe_unused]] AZ::ComponentDescriptor::DependencyArrayType& provided)
     {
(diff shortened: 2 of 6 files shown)


Gems/PhysX/Code/Source/HeightfieldColliderComponent.h
```diff
@@ -34,7 +34,7 @@ namespace PhysX
     //! The heightfield collider is a bit different from the other shape colliders in that it gets the heightfield data from a
     //! HeightfieldProvider, which can control position, rotation, size, and even change its data at runtime.
     //! 
-    //! Due to these differences, this component directly implements both the collider and static rigid body services instead of
+    //! This component directly implements both the collider and static rigid body instead of
```
mbalfour-amzn, line 37: Nit: IMO, the comment change removes the cause->effect linkage of the sentences, so it puts more burden on the user to infer *why* this component is avoiding the base classes. But if you think it's obvious anyways, I'm fine with that.

mbalfour-amzn (approved):

Gems/PhysX/Code/Source/HeightfieldColliderComponent.h
moraaar: Yeah, I removed that part because now it's not a necessity anymore to do it this way the Heightfield now is choosing to handle the static rigid body itself. 
Before it was a must because there was no static rigid body component, but now that there is one it could take that responsibility out and rely on the static rigid body component. This is the improvement that the **Out of Scope** section is referring to, but that's a refactor that we can do later.

dtamkin1 (approved):

Gems/PhysX/Code/Source/HeightfieldColliderComponent.h
mbalfour-amzn: Ah, ok, revised Nit: This comment would ideally say something like "This component currently directly implements both the collider and static rigid body instead of using BaseColliderComponent and StaticRigidBodyComponent because that was previously necessary. This is no longer necessary, so this could get refactored to use those." (Ideally something a little better and less clumsy, but you get the idea)

It just seems a little weird to me to leave the comment there as a statement of fact, without explaining any reasons.

cgalvan (approved):

spham-amzn (approved):

Gems/WhiteBox/Code/Source/Components/WhiteBoxColliderComponent.cpp
```diff
@@ -33,9 +33,8 @@ namespace WhiteBox
         }
     }
 
-    void WhiteBoxColliderComponent::GetProvidedServices(AZ::ComponentDescriptor::DependencyArrayType& provided)
+    void WhiteBoxColliderComponent::GetProvidedServices([[maybe_unused]] AZ::ComponentDescriptor::DependencyArrayType& provided)
```
greerdv, line 36: I think you can just get rid of this if it's empty

Gems/PhysX/Code/Source/HeightfieldColliderComponent.h
greerdv: Yeah, I agree that it would be good to add a little more context to the comment. Although having something that essentially amounts to a TODO in a doxygen comment may not be ideal.

Gems/WhiteBox/Code/Source/Components/WhiteBoxColliderComponent.h
```diff
@@ -18,7 +18,10 @@
 
 namespace WhiteBox
 {
-    //! Runtime White Box Collider.
+    //! Component that provides a White box Collider and associated Rigid Body.
+    //!
```
greerdv, line 22: I would suggest just getting rid of this line.

greerdv (approved):

Gems/PhysX/Code/Source/HeightfieldColliderComponent.h
moraaar: Let's see what I can put there in the comment to make it a more clear the current situation. `EditorStaticRigidBodyComponent` has being added, but it's still empty, for now it only forcing dependencies via services, the logic has to actually be moved there first, THEN these components can do the transition. So they are in a middle state at the moment.

Gems/WhiteBox/Code/Source/Components/WhiteBoxColliderComponent.cpp
moraaar: I found it actually easier to go through this code, review it, compare it and modify it when all the functions were present. Unless there are strong feelings against it, for this now I'd keep them.

Gems/WhiteBox/Code/Source/Components/WhiteBoxColliderComponent.h
moraaar: Sorry, which line would you remove? I've tried to make it so both Heightfield and Whitebox comments are similar because they are the same case.

greerdv: I meant the blank line in the comment.

Issue #8231, opened 2022-03-10
Disabling Focus Mode and Edit Mode by clicking level name in top left corner of Viewport is inconsistent


LB-JakubSkorupka:
**Describe the bug**
When **Edit mode** of any component is enabled and **Focus mode** of a prefab instance is also enabled, disabling the previously mentioned modes by clicking the **level** name in the top left corner of the **Viewport** tool, the behavior of the **Editor** is inconsistent.

The blue border present in the **Viewport** window disappears, but other tools (Entity Outliner, Asset Browser) remain greyed out. In the **Tool** dropdown menu of the Editor, tools are also greyed out as Focus/Edit mode is still enabled.

Also, when a shape component is used to reproduce this issue, the shape disappears in the **Viewport** after pressing the level name.

Please note that this issue is temporary and can be worked around by undoing the action (Ctrl + Z) or pressing the **Esc** key.

**Steps to reproduce**
1. Create a prefab in the level.
2. Enter the **Focus mode** of the created prefab.
3. Add any component, in which **Edit mode** can be enabled (e.g. Non-uniform scale, shape components).
4. Enter the **Edit mode** of the component added in step 3.
5. Press the **level** name in the top left corner of the **Viewport** window.

**Additional step (to work around the issue)**
6. Press **Esc** key or undo the action (Ctrl + Z).

**Expected behavior**
Pressing the **level** name in top left corner of the **Viewport** window disables **Focus mode** and/or **Edit mode**.

**Actual behavior**
Pressing the **level** name in top left corner of **Viewport** window does **not** disable **Focus mode** and/or **Edit mode**. Only the blue border in **Viewport** window disappears.

**Video**

https://user-images.githubusercontent.com/86951967/157667711-fa350be0-d751-477f-b3bf-d7a2c1034e30.mp4

**Found in Branch**
Development (7a75e34)

**Desktop:**
 - Device: PC
 - OS: Windows
 - Version: 10
 - CPU: Intel Core i7-8700K
 - GPU: NVIDIA GeForce RTX 2080 Ti
 - Memory: 16GB

LB-KacperMetelski:
Issue verified as fixed on commit adc2f3f.

PR #14265, opened 2023-01-27
Component Mode | Disable prefab focus breadcrumb on component mode

AMZN-daimini:
## What does this PR do?
Fixes https://github.com/o3de/o3de/issues/8231

Disables the Prefab Focus Breadcrumbs during Component Mode.
This prevents the user from being able to change prefab focus and with it the entity selection while using Component Mode and editing a nested prefab.

Also adds some code to the Breadcrumbs component so that it paints the 

## How was this PR tested?
Manual testing. Verified the behavior is now correct and Breadcrumbs links are not clickable while in Component Mode.
DisablePrefabFocusBreadcrumbsInComponentMode

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabViewportFocusPathHandler.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabViewportFocusPathHandler.cpp
@@ -22,10 +22,16 @@ namespace AzToolsFramework::Prefab
 
         // Connect to Prefab Focus Notifications
         PrefabFocusNotificationBus::Handler::BusConnect(m_editorEntityContextId);
+
+        // Connect to Viewport Editor Mode Notifications
+        ViewportEditorModeNotificationsBus::Handler::BusConnect(m_editorEntityContextId);
     }
 
     PrefabViewportFocusPathHandler::~PrefabViewportFocusPathHandler()
     {
+        // Disconnect from Viewport Editor Mode Notifications
+        ViewportEditorModeNotificationsBus::Handler::BusDisconnect();
+
         // Disconnect from Prefab Focus Notifications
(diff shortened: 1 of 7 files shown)


cgalvan (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabViewportFocusPathHandler.cpp
```diff
@@ -85,6 +91,30 @@ namespace AzToolsFramework::Prefab
...
     }
 
+    void PrefabViewportFocusPathHandler::OnEditorModeActivated(
+        [[maybe_unused]] const ViewportEditorModesInterface& editorModeState, ViewportEditorMode mode)
+    {
+        // This function triggers when the Editor changes from regular editing to a Component Mode.
+        if (mode == ViewportEditorMode::Component)
+        {
+            // Disable the widgets to prevent it from being used to change selection.
+            m_backButton->setEnabled(false);
```
tjmichaels, line 101: Nit:  You could move these two calls to a method called something like "EnableBreadcrumbWidgets(bool enable)" in case it grows in the future, etc.  Just to avoid having to update multiple locations if another widget gets added to the list.  Just noticed they were repeated in both methods with just the true/false difference

tjmichaels (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabViewportFocusPathHandler.cpp
AMZN-daimini: Good point! That widget is legacy and we will probably red-code it soon, but I'll keep that in mind if that changes or I encounter similar cases :D

PR #14266, opened 2023-01-27
Improves build time by about 9%

lawsonamzn:
## What does this PR do?
* EBUS can now be extern - common EBUS from AZCore have been extern'd
* Several heavily used templates have been extern'd
* Moved a bunch of code from heavily-trafficked headers to source

Props to @lumberyard-employee-dm  for the EBUS work.

## Testing

Just compiling, building, running the editor.  Because this doesn't represent actual change in code, just shuffling around, only light testing was done.  AR should find if this is bad or not...

## Results

The following before and after timings are done on a completely unattended 12-core machine (so no background processes, etc). Each time build is deleted entirely and built from scratch, including configure.  This is using UNITY builds, non-unity will be significantly more impacted.

These times are in wall-clock time.

* none of these optimizations: Execution time: 5106.667 s
* with all these optimizations: Execution time: 4689.982 s (417s / about 7 minutes)

The choice of what to optimize is based on using the clang build analyzer, profiler and -ftime-trace flags on windows.

## How was this PR tested?


lawsonamzn:
In case anyone wants to see the raw profiler data for the state of the engine BEFORE and AFTER these
[build_timings_unattended_no_opts.txt](https://github.com/o3de/o3de/files/10521436/build_timings_unattended_no_opts.txt)
versus with all the above:
[build_timings_with_ebus_moves.txt](https://github.com/o3de/o3de/files/10521441/build_timings_with_ebus_moves.txt)

lawsonamzn:
The validator doesn't like that I moved an AZ_PLATFORM macro from a header to a CPP file 🤡

Code/Framework/AzCore/AzCore/RTTI/BehaviorContext.h
```diff
@@ -1524,7 +1524,7 @@ namespace AZ
         virtual void OnAddEBus(const char* ebusName, BehaviorEBus* ebus)    { (void)ebusName; (void)ebus; }
         virtual void OnRemoveEBus(const char* ebusName, BehaviorEBus* ebus) { (void)ebusName; (void)ebus; }
     };
-
+
```
lumberyard-employee-dm, line 1527: nitpick: extra whitespace has been added here.
```suggestion

```

Code/Framework/AzToolsFramework/AzToolsFramework/API/EditorAssetSystemAPI.h
```diff
@@ -309,21 +259,16 @@ namespace AzToolsFramework
...
-#elif defined(AZ_PLATFORM_WINDOWS)
-            return "pc";
-#elif defined(AZ_PLATFORM_LINUX)
-            return "linux";
-#else
-            #error Unimplemented Host Asset Platform
-#endif
-        }
+        //! Returns "mac", "pc", or "linux" statically.
+        extern const char* GetHostAssetPlatform();
```
lumberyard-employee-dm, line 263: This can be moved to PlatformDefaults.h.
Also a function is external by default, so the `extern` keyword isn't needed for a function declaration.

lumberyard-employee-dm (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/API/EditorAssetSystemAPI.cpp
```diff
@@ -0,0 +1,91 @@
...
+                    ->Field("watchFolder", &JobInfo::m_watchFolder)
+                    ->Field("errorCount", &JobInfo::m_errorCount)
+                    ->Field("warningCount", &JobInfo::m_warningCount)
+                    ;
+            }
+        }
+
+        const char* GetHostAssetPlatform()
+        {
+#if defined(AZ_PLATFORM_MAC)
```
Kadino, line 79: Validator is choking on these defines, which are supposedly not intended to appear in this location.

lawsonamzn: yep, its supposed to prevent new additions of the AZ_PLATFORM macros (outdated way of doing this).

I angered it by moving the code from the header into the CPP and thus disturbed teh validator.  i'll find a better way.

cgalvan (approved):

lumberyard-employee-dm (approved):

Code/Framework/AzCore/AzCore/EBus/EBus.h
```diff
@@ -1999,3 +1999,42 @@ AZ_POP_DISABLE_WARNING
...
+   template class EBus<a, b>; \
+}
+//! Instantiates an EBus class template using only the interface argument
+//! for both the EBus Interface and BusTraits template parameters
+
+#define DECLARE_EBUS_INSTANTIATION_NO_TRAITS(a) \
+namespace AZ \
+{ \
+   template class EBus<a, a>; \
+}
```
mbalfour-amzn, line 2040: Suggestion: It seems like these would read a little better in the positive form instead of the negative form.
DECLARE_EBUS_EXTERN(a)
DECLARE_EBUS_AND_TRAITS_EXTERN(a, b)

In reading through the PR, I kept seeing "NO_TRAITS" as possibly something that was getting stripped off for some reason.

lawsonamzn: I can agree with that.  any thoughts from any other reviewers?

mbalfour-amzn (approved):

Code/Framework/AzCore/AzCore/EBus/EBus.h
mbalfour-amzn: fwiw, i actually thought about proposing the macro magic route, but backed off from it when i realized the one advantage of the current design is that it makes a mismatch between "DECLARE_EBUS_EXTERN" / "DECLARE_EBUS_INSTANTIATION" a little more obvious, because the names won't line up. a mismatch will be harder to spot if it's just the number of parameters that are different.

lawsonamzn: yeah I think I'll switch it to 
```
DECLARE_EBUS_EXTERN(a)
DECLARE_EBUS_EXTERN_WITH_TRAITS(a,b)
```
I also have to figure out whyt he compile is failing, but not locally

lumberyard-employee-dm: I am good with switch to additive terminology.

Code/Framework/AzCore/AzCore/Component/ComponentApplicationBus.h
```diff
@@ -211,3 +211,5 @@ namespace AZ
     //! Used by components to make requests of the component application.
     using ComponentApplicationBus = AZ::EBus<ComponentApplicationRequests, ComponentApplicationRequestsEBusTraits>;
 }
+
+DECLARE_EBUS_EXTERN(ComponentApplicationRequests, ComponentApplicationRequestsEBusTraits);
```
AMZN-alexpete, line 215: nit: I'm not sure if it's GitHub's viewer,  but it seems that a few of the files like this one are missing newlines at the end.

Code/Framework/AzCore/AzCore/EBus/EBus.h
AMZN-alexpete, line 2040: nit: possible missing newline at end of file

Code/Framework/AzCore/AzCore/Serialization/EditContext.cpp
```diff
@@ -414,4 +414,9 @@ namespace AZ
             return nullptr;
         }
     } // namespace Edit
+
 } // namespace AZ
+
+// pre-instantiate the extremely common ones
+template AZ::EditContext::ClassBuilder* AZ::EditContext::ClassBuilder::Attribute<AZ::Crc32>(const char *, AZ::Crc32);
+template AZ::EditContext::ClassBuilder* AZ::EditContext::ClassBuilder::Attribute<AZ::Crc32>(AZ::Crc32, AZ::Crc32);
```
AMZN-alexpete, line 422: nit: possible missing newline at end of file

Code/Framework/AzToolsFramework/Platform/Windows/AzToolsFramework/API/EditorAssetSystemAPI_Windows.cpp
```diff
@@ -0,0 +1,17 @@
...
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+namespace AzToolsFramework
+{
+    namespace AssetSystem
+    {
+        const char* GetHostAssetPlatform()
+        {
+            return "pc";
```
AMZN-alexpete, line 14: in the future might be nice to unify our platform name strings, maybe alongside the trait: `AZ_TRAIT_OS_IS_HOST_OS_PLATFORM`? And have PlatformDefaults also use them? or something like this?

```suggestion
            return AZ::PlatformDefaults::PlatformPC;
```

Though if someone were to change this to say upper case and that caused Asset lookups in the database to fail that'd be bad.

AMZN-alexpete (approved):
9% is great!!! Just some minor stuff about possible missing newlines.

Code/Framework/AzCore/AzCore/EBus/EBus.h
galibzon: I agree with @mbalfour-amzn suggestion.

hultonha (commented):
Awesome changes and a big improvement for a pretty small change, great to see! 😄

hultonha (approved):

Code/Framework/AzToolsFramework/Platform/Windows/AzToolsFramework/API/EditorAssetSystemAPI_Windows.cpp
lawsonamzn: well, in this case, the AP uses these to both auto enable platforms (in which case a enum would work) but also, to give the name of the actual folder in the cache (Cache/pc for example) to use.  Changing it to an enum would not work, we'd need another function to translate the enum to a folder name anyway (for backward compat).

But yeah, we should clean this all up in the future (no related to a build time improvement)

lumberyard-employee-dm (approved):

PR #14267, opened 2023-01-27
Adapt PhysX unit tests using the new static rigid body component

moraaar:
**Note**

## What does this PR do?

- Adapt PhysX unit tests to count for the colliders requiring a rigid body (static or dynamic).
- Fixed dyn_veg PhysXColliderSurfaceTagEmitter python test.

## How was this PR tested?

PhysX unit tests pass.

### DIFF
--- a/AutomatedTesting/Gem/PythonTests/largeworlds/dyn_veg/EditorScripts/PhysXColliderSurfaceTagEmitter_E2E_Editor.py
+++ b/AutomatedTesting/Gem/PythonTests/largeworlds/dyn_veg/EditorScripts/PhysXColliderSurfaceTagEmitter_E2E_Editor.py
@@ -115,7 +115,7 @@ def PhysXColliderSurfaceTagEmitter_E2E_Editor():
     collider_entity = hydra.Entity("Collider Surface")
     collider_entity.create_entity(
         entity_center_point,
-        ["PhysX Collider", "PhysX Collider Surface Tag Emitter"]
+        ["PhysX Collider", "PhysX Collider Surface Tag Emitter", "PhysX Static Rigid Body"]
         )
     Report.result(collider_entity_created, collider_entity.id.IsValid())
 
@@ -176,12 +176,12 @@ def PhysXColliderSurfaceTagEmitter_E2E_Editor():
     collider_entity.add_component("PhysX Collider")
     helper.wait_for_condition(lambda: editor.EditorComponentAPIBus(bus.Broadcast, 'IsComponentEnabled',
                                                                    collider_entity.components[1]), 5.0)
-    hydra.get_set_test(collider_entity, 1, "Shape Configuration|Shape", 7)
-    hydra.get_set_test(collider_entity, 1, "Shape Configuration|Asset|PhysX Mesh", test_physx_mesh_asset_id)
+    hydra.get_set_test(collider_entity, 2, "Shape Configuration|Shape", 7)
+    hydra.get_set_test(collider_entity, 2, "Shape Configuration|Asset|PhysX Mesh", test_physx_mesh_asset_id)
 
     # Set the asset scale to match the test heights of the shapes tested
     asset_scale = math.Vector3(1.0, 1.0, 9.0)
-    collider_entity.get_set_test(1, "Shape Configuration|Asset|Configuration|Asset Scale", asset_scale)
+    collider_entity.get_set_test(2, "Shape Configuration|Asset|Configuration|Asset Scale", asset_scale)
 
     # Test:  Generate a new surface on the collider.
     # There should be one instance at the very top of the collider mesh, and none on the baseline surface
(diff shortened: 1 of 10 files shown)


cgalvan (approved):

Gems/PhysX/Code/Tests/StaticRigidBodyComponentTests.cpp
```diff
@@ -51,36 +51,57 @@ namespace PhysXEditorTests
...
+        // Create editor entity
+        EntityPtr editorEntity = CreateInactiveEditorEntity("Entity");
+        editorEntity->CreateComponent<PhysX::EditorShapeColliderComponent>();
+        AddEditorBoxShapeComponent(editorEntity);
+
+        // Create game entity and verify StaticRigidBodyComponent was NOTcreated
+        EntityPtr gameEntity = CreateActiveGameEntityFromEditorEntity(editorEntity.get());
+        const auto* staticRigidBody = gameEntity->FindComponent<PhysX::StaticRigidBodyComponent>();
+
+        EXPECT_TRUE(staticRigidBody == nullptr);
```
spham-amzn, line 65: nit: It may be useful for EXPECT_ for some situations to surface the cause of failure 
```suggestion
        EXPECT_TRUE(staticRigidBody == nullptr) << "A StaticRigidBodyComponent should NOT have been created";
```

Gems/PhysX/Code/Tests/StaticRigidBodyComponentTests.cpp
```diff
@@ -51,36 +51,57 @@ namespace PhysXEditorTests
...
 
-    TEST_F(PhysXEditorFixture, StaticRigidBodyComponent_NoRigidBody_RuntimeStaticRigidBodyComponentCreated)
+    TEST_F(PhysXEditorFixture, StaticRigidBodyComponent_NoRigidBody_NoRuntimeStaticRigidBodyComponent)
+    {
+        // Create editor entity
+        EntityPtr editorEntity = CreateInactiveEditorEntity("Entity");
+        editorEntity->CreateComponent<PhysX::EditorShapeColliderComponent>();
+        AddEditorBoxShapeComponent(editorEntity);
+
+        // Create game entity and verify StaticRigidBodyComponent was NOTcreated
```
spham-amzn, line 61: nit:
```suggestion
        // Create game entity and verify StaticRigidBodyComponent was NOT created
```

spham-amzn (approved):

AutomatedTesting/Gem/PythonTests/largeworlds/dyn_veg/EditorScripts/PhysXColliderSurfaceTagEmitter_E2E_Editor.py
```diff
@@ -176,12 +176,12 @@ def validate_behavior_context():
     collider_entity.add_component("PhysX Collider")
     helper.wait_for_condition(lambda: editor.EditorComponentAPIBus(bus.Broadcast, 'IsComponentEnabled',
                                                                    collider_entity.components[1]), 5.0)
-    hydra.get_set_test(collider_entity, 1, "Shape Configuration|Shape", 7)
-    hydra.get_set_test(collider_entity, 1, "Shape Configuration|Asset|PhysX Mesh", test_physx_mesh_asset_id)
+    hydra.get_set_test(collider_entity, 2, "Shape Configuration|Shape", 7)
```
greerdv, line 179: What does that parameter do? Might be worth giving it a name. Or in python you can put the name in when you call it e.g. get_set_test(collider_entity, whatever_that_parameter_is_called=2, ...)

Gems/PhysX/Code/Tests/EditorTestUtilities.cpp
```diff
@@ -62,6 +63,10 @@ namespace PhysXEditorTests
         {
             editorEntity->CreateComponent<PhysX::EditorRigidBodyComponent>();
         }
+        else
```
greerdv, line 66: Probably out of scope for this change, but looks like we can probably refactor some duplicated code here.

greerdv (approved):

Gems/PhysX/Code/Tests/StaticRigidBodyComponentTests.cpp
moraaar: It's a good suggestion, I'll keep it in mind for future usages. It could be done later as a clean up pass (since all tests in the document do not include that extra message), but this work change is already 5 PRs, so I'd like to avoid making large clean up changes at the same time to keep it manageable. Thanks.

Gems/PhysX/Code/Tests/StaticRigidBodyComponentTests.cpp
moraaar: Thanks, will fix.

AutomatedTesting/Gem/PythonTests/largeworlds/dyn_veg/EditorScripts/PhysXColliderSurfaceTagEmitter_E2E_Editor.py
moraaar: That's the index of the component it's trying to do the `get_set` too. It got me as well why that was failing. It's not great because it has to track which index it is, initially in the test when the entity is created the collider is the first component (so it used 0) and now that it as removed it and added it it's the last one, which now the index would be 2 since there are 3 components now instead of 2 (I added the static rigid body).

Adding the parameter name is a good idea, I'll give it a go.

Gems/PhysX/Code/Tests/EditorTestUtilities.cpp
moraaar: Yeah, let's leave that for another pass.

PR #14270, opened 2023-01-27
Remove 'engines_path' and add user/project_settings.json support

AMZN-alexpete:
## What does this PR do?

1. Stop using `engines_path` in `o3de_manifest.json` because engine names are no longer unique, instead walk every engine listed in `engines` and look for the most compatible version.
2. Update o3de `edit-project-properties` to modify `user/project.json` when the `--user` arg is provided.  `user/project.json` settings are allowed to override those in the shared `project.json`
3. Add `engine_path` option that can be set in `user/project.json` only.  When looking for the engine, this path is considered first.  If this file does not exist, fall back to searching in the `o3de_manifest.json` for the most compatible engine
4. Add `engine_cmake_finder_path` option to `edit-project-properties` and update AutomatedTesting and Project templates  `CMakeList.txt` to support including this file instead of the default one in projects.  This will allow users to `undo` changes we make that may break their project's `EngineFinder.cmake` when we upgrade it.
5. Update AutomatedTesting and Project templates 'EngineFinder.cmake' to support `user/project.json` settings like `engine_path` and to look for the most compatible engine.
6. Display popup error message box when the project could not be found
image
image

7 Fix error detection when invalid engine name or missing engine found.
image
8. Moves `Dependency.*` and `Version.*` from `AzFramework` into `AzCore` so they can be used for version compatibility checks. 

Related versioning RFC: https://github.com/o3de/sig-core/issues/44

## How was this PR tested?

- configured, built and ran engine-centric project (Automated Testing)
- configured, built and ran project-centril project (MultiplayerSample)
- started up the editor in project-centric and engine-centric modes, stepping through the `FindEngineRoot` logic to verify it is considering `engine_path` and other `user/project.json` settings.
- updated unit tests for o3de cli changes


AMZN-alexpete:
A really good call out by @amzn-phist that old projects have older `EngineFinder.cmake` files that would need to be upgraded to use the `user/project_settings.json` file and not the `engines_path` setting in the manifest.  One idea is to do that as part of an "update" path via Python when the user registers their project with a new engine, or some similar command.  If a user doesn't register their project with an engine because it was already registered with one - in the case where you just pull latest dev from GitHub, we could add a CMake warning in the engine that tells you to re-register your project.

As far as updating one option is to make it part of registration like so:
1. user runs `o3de register --pp <old-project>` and Python checks either a field in `project.json` (engine or template version?) or other and determines an update is needed.  Sees this output:
```
project <old-project> needs to be updated to work with engine <new-engine> run register with the --update command to update the project.
```
2. user runs `o3de register --pp <old-project> --update`

Python runs the appropriate update script, makes any necessary backups and successfully registers the engine.

amzn-phist:
Another thing to think about is how `o3de_manifest.json` gets updated.  Any new engine registrations would not write anything to `"engines_path"` object, only add a path to the `"engines"` array.  So is there a mechanism to remove the `"engines_path"` object from the manifest?  Leaving old `"engines_path"` entries could help keep old projects working for a while, but it's likely to atrophy over time.  But it could allow for CMake configure to find the engine for a time.

If we did add a mechanism that removes the `"engines_path"` object from a user's manifest after this update, maybe we could use that to trigger an error and offer a mitigation path for fixing it.  But it is often the case that the user would be starting CMake from an old project, which would run the old version of `EngineFinder.cmake` and `CMakeFiles.txt` (in project root), and those files don't have any detection or mitigations in place.  It would error when it couldn't find the `"engines_path"` object and stop configuring without providing any details on how to fix it.

I want to make sure we think about these update/upgrade scenarios in order to offer the best user experience.

lumberyard-employee-dm:
> A really good call out by @amzn-phist that old projects have older `EngineFinder.cmake` files that would need to be upgraded to use the `user/project_settings.json` file and not the `engines_path` setting in the manifest. One idea is to do that as part of an "upgrade" path via Python when the user registers their project with a new engine, or upgrade CLI command.
> 
> If a user doesn't register their project with an engine because it was already registered with one - in the case where you just pull latest dev from GitHub, we could add a CMake warning in the engine that tells you to re-register your project (or use a CLI upgrade command).
> 
> One option for "upgrading" is to make it part of registration like so:
> 
>     1. user runs `o3de register --pp <old-project>` and Python checks either a field in `project.json` (engine or template version?) or other and determines an update is needed.  Sees this output:
> 
> 
> ```shell
> Project <old-project> needs to be updated to work with engine <new-engine>.
> Run register with the --upgrade param to update the project.
> ```
> 
>     2. user runs `o3de register --pp <old-project> --upgrade`
> 
> 
> Python runs the appropriate update script, makes any necessary backups and successfully registers the engine.
> 
> Although, we would likely still want a separate "upgrade" CLI command that didn't require registration.

For number 1, I don't see why upgrading would be part of registration and having the separate `upgrade` command is the approach I think O3DE should go with.

From a user perspective "why do I need re-register my project with the same engine that it is already registered with?".

The expectation of being telling the user to run an "upgrade" command that can update the registration, replace the existing `EngineFinder.cmake` with an upgraded version, and notify the user that if their project is part of source control, they need to submit the modification of upgraded files.
Furthermore for any upgrade paths, the hook mechanism for the scripts should require the authors to return a list of files that 1. will be upgraded and 2. were upgraded, so the user knows what to check in to source control. This is important in case a user has other modified files in their branch, they know which files that need to check in to source control or which ones they didn't modify.

AutomatedTesting/cmake/EngineFinder.cmake
```diff
@@ -36,6 +36,32 @@ if(CMAKE_MODULE_PATH)
...
+        list(APPEND CMAKE_MODULE_PATH "${engine_path}/cmake")
+        return()
+    else()
+        string(CONCAT engine_path_info "Unable to use the engine_path from "
+            "'${CMAKE_CURRENT_SOURCE_DIR}/user/project_settings.json' because "
+            "'${engine_path}/cmake' does not exist.\n"
+            "Falling back to look for an engine named '${LY_ENGINE_NAME_TO_USE}' in o3de_manifest.json")
+        message(INFO " ${engine_path_info}")
+    endif()
+endif()
```
lumberyard-employee-dm, line 63: I think we should reverse how CMake finds the engine after a project upgrade or for any new projects.

Instead of using the EngineFinder.cmake to initially to locate the engine, I think the project.json should have a "engine_finder" key that points to a path to an "engine finder cmake" script. Any relative paths would be treated relative to the `<project-root>` and absolute paths would be used as is.
Side Note: The [cmake_path](https://cmake.org/cmake/help/latest/command/cmake_path.html?highlight=cmake_path#normalization) takes care of this automatically as it works with similarly to python [os.path.join](https://docs.python.org/3/library/os.path.html#os.path.join) where an absolute path segment discards any previous path segments so far.

### project.json
```
{
    "engine_finder": "cmake/EngineFinder.cmake"
}
```

Now for the naming of the `project_settings.json` file. I think it should just be called `project.json` and it is used as an overlay for the `<project-root>/project.json` by merging its settings with it.
i.e It would work similar to the Settings Registry with how the `<project-root>/user/Registry` overlays the `<project-root>/Registry` directory.

Pretty much it can look for two files currently
```
<project-root>/project.json
<project-root>/user/project.json
```

Instead of allowing the user to specify an engine path, they can specify an "override" for the "engine" key identifier to locate an alternative registered engine.

lumberyard-employee-dm: By using an overlay of settings for the project.json this allows team and individual contributors to specify there engine to use separately from each other.

Now back to the "engine_finder" cmake key, this can be used provide upgraded Find engine scripts for the project, without needing to overwrite the existing `cmake/EngineFinder.cmake` in the project. This is useful in case the upgrade as failed and the maintainer checks in the new files `cmake/EngineFinder.cmake` files to source control, only for user to discover an issue with it.
Instead the upgrade back could be to add a `cmake/EngineFinder_v2.cmake` script to their project and update their project.json to start using that particular file.

AMZN-alexpete: I like it, so I'm assuming we'd update the project `CMakeList.txt` to do the initial loading of `project.json` and overlaying with `user/project.json` and then including the specified `engine_finder_cmake` entry?

lumberyard-employee-dm: Yes the project CMakeLists.txt would read the `project.json` and then the `user/project.json` for the "engine_finder_cmake" entry and then `include()` that path.

AMZN-alexpete:
There are enough good changes and details here I'm going to update the RFC.

AutomatedTesting/cmake/EngineFinder.cmake
```diff
@@ -47,42 +73,50 @@ Engine registration is required before configuring a project.
...
-            endif()
-
-            if(engine_path)
+                message(WARNING "Unable to read key 'engine_name' from '${engine_path}/engine.json'\nError: ${json_error}")
+            elseif(LY_ENGINE_NAME_TO_USE STREQUAL engine_name)
                 list(APPEND CMAKE_MODULE_PATH "${engine_path}/cmake")
+
+                message(INFO " Found an engine named ${LY_ENGINE_NAME_TO_USE} at ${engine_path}")
+                # save this location in user/project_settings.json for faster retrieval
+                file(WRITE ${CMAKE_CURRENT_SOURCE_DIR}/user/project_settings.json "{
```
AMZN-alexpete, line 109: Another example of writing JSON info from cmake worth considering - possible home for this functionality: https://github.com/o3de/o3de/blob/development/cmake/Projects.cmake#L123-L164

AutomatedTesting/CMakeLists.txt
```diff
@@ -8,11 +8,32 @@
 
 if(NOT PROJECT_NAME)
     cmake_minimum_required(VERSION 3.22)
-    include(cmake/EngineFinder.cmake OPTIONAL)
+
+    # Utility function to look for an optional 'engine_finder_cmake_path'setting 
+    function(get_engine_finder_cmake_path project_json_file_path path_value)
+        if(NOT $path_value AND EXISTS "${project_json_file_path}")
```
lumberyard-employee-dm, line 14: The $path_value variable needs to remove the `$`
CMake needs the name of the variable to check if has not been set or is a false value

```suggestion
        if(NOT path_value AND EXISTS "${project_json_file_path}")
```

lumberyard-employee-dm (changes requested):

AutomatedTesting/cmake/EngineFinder.cmake
```diff
@@ -7,35 +7,72 @@
...
+# the 'engine_finder_cmake_path' key in your project.json or user/project.json to specify 
+# an alternate .cmake file to use instead of this one.
 
 include_guard()
 
-# Read the engine name from the project_json file
+include(cmake/VersionUtils.cmake)
+
+set(ENGINE_FINDER_VERBOSE FALSE CACHE BOOL "Increase EngineFinder.cmake verbosity.  Useful for debugging engine selection based on compatibility.")
+
```
lumberyard-employee-dm, line 19: This isn't needed.
CMake already has [--log-level](https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-log-level) and what we should do is make the messages `VERBOSE` using that keywork.

With using a Cache variable, that value is now sticky and setting the verbosity will stick between subsequent runs until the user changes the value back to FALSE

AutomatedTesting/cmake/EngineFinder.cmake
```diff
@@ -7,35 +7,72 @@
...
-string(JSON LY_ENGINE_NAME_TO_USE ERROR_VARIABLE json_error GET ${project_json} engine)
+string(JSON O3DE_ENGINE_NAME_TO_USE ERROR_VARIABLE json_error GET ${project_json} engine)
 if(json_error)
     message(FATAL_ERROR "Unable to read key 'engine' from 'project.json'\nError: ${json_error}")
 endif()
+ 
+# Let the user override the engine to use
+if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/user/project.json)
+    file(READ ${CMAKE_CURRENT_SOURCE_DIR}/user/project.json user_project_json)
+    set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/user/project.json)
```
lumberyard-employee-dm, line 32: This only needs to be set as a CMake configure dependency if it has the "engine" key within it.

If the "engine" key is not within the `user/project.json` file, then CMake doesn't need to re-run configure when a generator project builds.

AutomatedTesting/cmake/EngineFinder.cmake
```diff
@@ -7,35 +7,72 @@
...
 if(CMAKE_MODULE_PATH)
     foreach(module_path ${CMAKE_MODULE_PATH})
         if(EXISTS ${module_path}/Findo3de.cmake)
-            file(READ ${module_path}/../engine.json engine_json)
-            string(JSON engine_name ERROR_VARIABLE json_error GET ${engine_json} engine_name)
-            if(json_error)
-                message(FATAL_ERROR "Unable to read key 'engine_name' from 'engine.json'\nError: ${json_error}")
-            endif()
-            if(LY_ENGINE_NAME_TO_USE STREQUAL engine_name)
+            o3de_engine_compatible(${module_path}/../ ${O3DE_ENGINE_NAME_TO_USE} ${O3DE_ENGINE_SPECIFIER_OP} ${O3DE_ENGINE_SPECIFIER_VERSION} is_compatible engine_version ${ENGINE_FINDER_VERBOSE})
```
lumberyard-employee-dm, line 49: The first `${module_path}/..` should be put in double quotes.

If the module_path has a space in it such as `C:\o3de\My Engine`, then it will get parsed as two arguments.

Also I advise using [cmake_path(GET ... PARENT_PATH)](https://cmake.org/cmake/help/latest/command/cmake_path.html?highlight=cmake_path#get) to retrieve the parent directory 
```suggestion
            cmake_path(GET module_path PARENT_PATH candidate_engine_path)
            o3de_engine_compatible("${candidate_engine_path}" ${O3DE_ENGINE_NAME_TO_USE} ${O3DE_ENGINE_SPECIFIER_OP} ${O3DE_ENGINE_SPECIFIER_VERSION} is_compatible engine_version ${ENGINE_FINDER_VERBOSE})
```

AutomatedTesting/cmake/EngineFinder.cmake
```diff
@@ -7,35 +7,72 @@
...
     endforeach()
 endif()
 
+# Read the user/project.json file and look for the engine_path
+if(user_engine_path)
+    if(EXISTS ${user_engine_path}/cmake)
+        # always be verbose on this compatibility check
+        o3de_engine_compatible(${user_engine_path} ${O3DE_ENGINE_NAME_TO_USE} ${O3DE_ENGINE_SPECIFIER_OP} ${O3DE_ENGINE_SPECIFIER_VERSION} is_compatible engine_version TRUE)
+        if(is_compatible)
+            message("-- Selecting engine from 'user/project.json' at '${user_engine_path}'")
```
lumberyard-employee-dm, line 63: Using the `STATUS` keyword adds a leading `-` the message

AutomatedTesting/cmake/EngineFinder.cmake
```diff
@@ -47,46 +84,71 @@ Engine registration is required before configuring a project.
...
+                        message("-- Considering engine '${O3DE_ENGINE_NAME_TO_USE}' with version '${engine_version}' from '${engine_path}' because it has a greater version number")
+                    endif()
+                    set(most_compatible_engine_path ${engine_path})
+                    set(most_compatible_engine_version ${engine_version})
+                elseif(ENGINE_FINDER_VERBOSE)
+                    message("-- Not considering engine '${O3DE_ENGINE_NAME_TO_USE}' with version '${engine_version}' from '${engine_path}' because it does not have a greater version number than the engine at '${most_compatible_engine_path}' (${most_compatible_engine_version})")
+                endif()
             endif()
+        else()
+            message(WARNING "${engine_path}/engine.json not found")
```
lumberyard-employee-dm, line 134: The foreach loop looks like a good candidate for being put into a cmake function called `get_most_compatible_engine`

AutomatedTesting/cmake/VersionUtils.cmake
```diff
@@ -0,0 +1,116 @@
...
+        set(contains_version TRUE)
+    elseif(op STREQUAL "<=" AND version VERSION_LESS_EQUAL specifier_version)
+        set(contains_version TRUE)
+    elseif(op STREQUAL ">=" AND version VERSION_GREATER_EQUAL specifier_version)
+        set(contains_version TRUE)
+    elseif(op STREQUAL "<" AND version VERSION_LESS specifier_version)
+        set(contains_version TRUE)
+    elseif(op STREQUAL ">" AND version VERSION_GREATER specifier_version)
+        set(contains_version TRUE)
+    elseif(op STREQUAL "===" AND version VERSION_EQUAL specifier_version)
```
lumberyard-employee-dm, line 30: Triple equal should use a regular STREQUAL since it is checking for an extract match
https://peps.python.org/pep-0440/#arbitrary-equality

AutomatedTesting/cmake/VersionUtils.cmake
```diff
@@ -0,0 +1,116 @@
...
+        set(contains_version TRUE)
+    elseif(op STREQUAL "~=")
+        # compatible versions have an equivalent combination of >= and == 
+        # e.g. ~=2.2 is equivalent to >=2.2,==2.*
```
lumberyard-employee-dm, line 34: So this is any value between `2.*` and `3.0` according to the [compatible release](https://peps.python.org/pep-0440/#compatible-release) clause for PEP440?

Code/Framework/AzCore/AzCore/Settings/SettingsRegistryMergeUtils.cpp
```diff
@@ -337,16 +335,17 @@ namespace AZ::SettingsRegistryMergeUtils
...
+        // as the one in project.json
         AZ::IO::FixedMaxPath projectRoot;
         settingsRegistry.Get(projectRoot.Native(), FilePathKey_ProjectPath);
         if (projectRoot.empty())
         {
             return {};
         }
 
-        // Use the project.json and engine manifest to locate the engine root.
+        // Use user/project_settings.json, project.json and engine manifest to locate the engine root.
```
lumberyard-employee-dm, line 348: This comment should be updated to be `user/project.json`
```suggestion
        // Use <project-root>/user/project.json, project.json and engine manifest to locate the engine root.
```

Code/Framework/AzCore/AzCore/Settings/SettingsRegistryMergeUtils.cpp
```diff
@@ -337,16 +335,17 @@ namespace AZ::SettingsRegistryMergeUtils
             return engineRoot;
         }
 
-        // Step 3 locate the project root and attempt to find the engine root using the registered engine
-        // for the project in the project.json file
+        // Step 3 locate the project root and attempt to find the engine using the engine_path
+        // in user/project_settings.json or find the first registered engine with the same name
```
lumberyard-employee-dm, line 339: The project_settings.json naming scheme needs to be updated.
```suggestion
        // in <project-root>/user/project.json or find the first registered engine with the same name
```

Code/Framework/AzCore/AzCore/Settings/SettingsRegistryMergeUtils.cpp
```diff
@@ -57,7 +57,7 @@ namespace AZ::Internal
 
     AZ::IO::FixedMaxPath ReconcileEngineRootFromProjectPath(SettingsRegistryInterface& settingsRegistry, const AZ::IO::FixedMaxPath& projectPath)
     {
-        // Find the engine root via the engine manifest file and project.json
+        // Find the engine root via user/project_settings.json, engine manifest and project.json
```
lumberyard-employee-dm, line 60: The name project_settings.json is no longer used.
```suggestion
        // Find the engine root via <project-root>/user/project.json, engine manifest and project.json
```

Templates/DefaultProject/Template/cmake/VersionUtils.cmake
```diff
@@ -0,0 +1,116 @@
+# {BEGIN_LICENSE}
```
lumberyard-employee-dm, line 1: The template.json hasn't been updated to copy over this file when creating a project..

The MinimalProject template also needs this update.

lumberyard-employee-dm (changes requested):

Templates/DefaultProject/Template/cmake/VersionUtils.cmake
lumberyard-employee-dm: Furthermore do not, that updating existing projects to use newer engine_finder cmake scripts is not easy to do, so we should endeavor to have as few files and as few changes as necessary to locate and validate the engine.

Once users make a project using these templates, it will much harder to update the engine finder script, so we should account for any potential issues where this file would need to be updated in the future.

Templates/DefaultProject/Template/cmake/EngineFinder.cmake
```diff
@@ -47,46 +84,71 @@ Engine registration is required before configuring a project.
 Run 'scripts/o3de register --this-engine' from the engine root.
 ]=])
 
-# Read the ~/.o3de/o3de_manifest.json file and look through the 'engines_path' object.
-# Find a key that matches LY_ENGINE_NAME_TO_USE and use that as the engine path.
+# Read the ~/.o3de/o3de_manifest.json file and look through the 'engines' array.
+# Search all known engines for one with the matching name and compatible version
```
amzn-phist, line 88: General question...
Say my engines array looks like `["C:/EngineA", "C:/EngineB"]` and I'm working on a project that uses engine B.  It will always look at and read engine A's `engine.json` first.  Does this mean that if I happen to modify engine A's `engine.json` that it will cause ZERO_CHECK on the project (that uses engine B) to force another CMake configure?

Code/Framework/AzCore/AzCore/Dependency/Dependency.h
```diff
@@ -49,7 +48,7 @@ namespace AzFramework
                 GreaterThan = 1 << 0,
                 LessThan = 1 << 1,
                 EqualTo = 1 << 2,
-                // Special operators
+                // Special operators ~> and ~=
```
lumberyard-employee-dm, line 51: We can get rid of the TwiddleWakka, that isn't needed for the semantic version support we envision using.

Code/Framework/AzCore/AzCore/Settings/SettingsRegistryMergeUtils.cpp
```diff
@@ -83,90 +237,97 @@ namespace AZ::Internal
...
-                    if (settingsRegistry.MergeSettingsFile(
-                            engineSettingsPath.Native(), AZ::SettingsRegistryInterface::Format::JsonMergePatch, EngineSettingsRootKey))
-                    {
-                        FixedValueString engineName;
-                        settingsRegistry.Get(engineName, engineMonikerKey);
-                        engineInfo.m_moniker = engineName; 
-                    }
-                }
+                        return AZ::SettingsRegistryInterface::VisitResponse::Continue;
+                    }, enginesKey);
```
lumberyard-employee-dm, line 305: We need UnitTest for this function.

Code/Framework/AzCore/AzCore/Settings/SettingsRegistryMergeUtils.cpp
```diff
@@ -83,90 +237,97 @@ namespace AZ::Internal
...
+                const auto engineVersionKey = FixedValueString::format("%s/version", EngineSettingsRootKey);
+                const auto enginesKey = FixedValueString::format("%s/engines", O3deManifestSettingsRootKey);
+
+                AZ::IO::FixedMaxPath projectUserPath;
+                settingsRegistry.Get(projectUserPath.Native(), FilePathKey_ProjectUserPath);
+
+                EngineInfo mostCompatibleEngineInfo;
+
+                // Look through the manifest engines for the most compatible engine
+                AZ::SettingsRegistryVisitorUtils::VisitArray(settingsRegistry, [&](const AZ::SettingsRegistryInterface::VisitArgs& visitArgs)
```
lumberyard-employee-dm, line 257: FYI the `visitArgs` has an `m_registry` member with a reference to the SettingsRegistry on it

Code/Framework/AzCore/AzCore/Settings/SettingsRegistryMergeUtils.cpp
```diff
@@ -83,90 +237,97 @@ namespace AZ::Internal
...
+                            // is it more compatible?
+                            else if (!engineInfo.m_version.IsZero() && engineInfo.m_version > mostCompatibleEngineInfo.m_version)
+                            {
+                                mostCompatibleEngineInfo = engineInfo;
+                            }
+                        }
 
-            const auto engineMonikerKey = FixedValueString::format("%s/engine_name", EngineSettingsRootKey);
+                        // Remove engine settings, which will be added back once we know the most compatible engine
+                        settingsRegistry.Remove(EngineSettingsRootKey);
```
lumberyard-employee-dm, line 300: If we make a local `AZ::SettingsRegistryImpl` instance, we don't have to worry about removing the  `o3de_manifest.json` and `engine.json` settings we load into.
We can let the local registry instance go out of scope and keep the `settingsRegistry` parameter passed into this function clean of any modifications until the end.

Code/Framework/AzCore/AzCore/Settings/SettingsRegistryMergeUtils.cpp
```diff
@@ -36,34 +38,186 @@ namespace AZ::Internal
...
+            {
+                if (auto engineRootAbsPath = AZ::Utils::ConvertToAbsolutePath(engineRoot.Native());
+                    engineRootAbsPath.has_value())
+                {
+                    engineRoot = AZStd::move(*engineRootAbsPath);
+                }
+            }
+
+            AZ::IO::FixedMaxPath engineJsonPath{ engineRoot / Internal::EngineJsonFilename };
+            AZ::IO::FixedMaxPath projectJsonPath{ engineRoot / projectPath / Internal::ProjectJsonFilename };
```
lumberyard-employee-dm, line 198: The projectPath is absolute at this point, so it appending it to the engineRoot is a no op
```suggestion
            AZ::IO::FixedMaxPath projectJsonPath{ projectPath / Internal::ProjectJsonFilename };
```

Code/Framework/AzCore/AzCore/Settings/SettingsRegistryMergeUtils.cpp
```diff
@@ -36,34 +38,186 @@ namespace AZ::Internal
...
+        if (!AZ::IO::SystemFile::Exists(path.c_str()))
+        {
+            return AZ::Failure(AZStd::string::format("The setreg file: '%s' does not exist and cannot be merged.", path.c_str()));
+        }
+        else if (!settingsRegistry.MergeSettingsFile(path.Native(), AZ::SettingsRegistryInterface::Format::JsonMergePatch, anchorKey))
+        {
+            return AZ::Failure(AZStd::string::format("Could not merge '%s' registry settings.", path.c_str()));
+        }
+        return AZ::Success();
+    }
```
lumberyard-employee-dm, line 53: An Outcome is convertible to bool via it's operator bool.

How about we just updated the `SettingsRegistryImpl::MergeSettingsFile` and `SettingsRegistry::MergeSettingsFileInternal` function to return an `AZ::Outcome<void, AZStd::string>` instead and update all the AZ_Warning and AZ_Error calls within those functions to instead be sent to the error string that is returned?

AutomatedTesting/CMakeLists.txt
AMZN-alexpete: good catch

AutomatedTesting/cmake/VersionUtils.cmake
AMZN-alexpete: >= 2.2 ,== 2.*  means from 2.2 (inclusive) to 3 (exclusive).  This example from the docs:
```
For example, the following groups of version clauses are equivalent:

~= 2.2
>= 2.2, == 2.*
```
... informed the code based on the [packaging implementation here](https://github.com/pypa/packaging/blob/1b9da4420e9e23046dc095a3779f2607ae6e4d62/src/packaging/specifiers.py#L376-L395) and the [Dependency implementation here](https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/AzFramework/Dependency/Dependency.inl#L166-L188).

AutomatedTesting/CMakeLists.txt
```diff
@@ -8,11 +8,32 @@
...
+            file(READ "${project_json_file_path}" project_json_data)
+            string(JSON engine_finder_cmake_value ERROR_VARIABLE json_error GET ${project_json_data} "engine_finder_cmake_path")
+            if(NOT json_error AND EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/${engine_finder_cmake_value}")
+                set(${path_value} "${engine_finder_path}" PARENT_SCOPE)
```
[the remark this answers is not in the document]
AMZN-alexpete: good catch, missed this in a refactor.

Templates/DefaultProject/Template/cmake/EngineFinder.cmake
AMZN-alexpete: That's a good question, it is my understanding that unless you use `CMAKE_CONFIGURE_DEPENDS` to add an `engine.json` as a dependency, it will not be considered when checking.  In my testing, modifications to the other engine.json files did not trigger ZERO_CHECK

Code/Framework/AzCore/AzCore/Dependency/Dependency.h
AMZN-alexpete: I left it in for backwards compatibility with the XMLDependency code - if that is no longer needed then I'm down with removing it.

Templates/DefaultProject/Template/cmake/EngineFinder.cmake
lumberyard-employee-dm: Correct, the CMAKE_CONFIGURE_DEPENDS adds a file dependency for ZERO_CHECK.

The engine.json that is being used should have that `engine.json` added as a CMAKE_CONFIGURE_DEPENDS only

Templates/DefaultProject/Template/cmake/VersionUtils.cmake
AMZN-alexpete: Good catch.  I created VersionUtils as a library that I thought might be useful to anyone making their own custom `EngineFinder.cmake`.  I'm also leaning towards adding the new `EngineFinder.cmake` changes as a new file like you mentioned - called `EngineFinderV2.cmake` and then just modify the `CMakeList.txt` to use it as the new default.  That way when we update user projects (code to be written)  it would leave the original `EngineFinder.cmake` and they could easily switch back to it if they wanted to.

AMZN-alexpete: Hmmm, I suppose what you're getting at is user's might modify the VersionUtils.cmake file and make it harder if we wanted to update it...

AutomatedTesting/cmake/VersionUtils.cmake
lumberyard-employee-dm: So a version of `2.1` would fail the check?

Templates/DefaultProject/Template/cmake/VersionUtils.cmake
lumberyard-employee-dm: Well if the user modifies the VersionUtils.cmake, then they are in advanced mode.
I more thinking of if we find a bug in VersisonUtils.cmake or EngineFinder.cmake it harders to push the changes out to existing projects.

By trying to keep it the logic as lightweight as possible, we can reduce the chances that a user needs to upgrade a project.

The point I was trying to get across is that it is a lot easier for a user to create a new project with the latest EngineFinder logic, then it is for them to upgrade an existing project to use the latest EngineFinder.cmake.

So we should try to avoid the need for the user to upgrade the EngineFinder.cmake

Code/Framework/AzCore/AzCore/Dependency/Dependency.h
lumberyard-employee-dm: @AMZN-stankowi Do you know if the XMLDependency code supports the `~>` version specifier?

It know it was used by CrySystem to check version dependencies in Lumberyard that used the old gem format, but would need clarification if the XMLDependency builder is using it.

Code/Framework/AzCore/AzCore/Settings/SettingsRegistryMergeUtils.cpp
```diff
@@ -521,8 +565,9 @@ namespace AZ::SettingsRegistryMergeUtils
             return engineRoot;
         }
 
-        // Fall back to using the project root as the engine root if the engine path could not be reconciled
-        return projectRoot;
+        // Do not use the project root as the engine root, otherwise we won't detect that the engine root
+        // is missing and notify the user.
+        return {};
```
lumberyard-employee-dm, line 571: This is going to be a problem for packaged builds.
The `engine.json` and `project.json` files would be in the bundled assets, which would be found underneath the `<project-root>/Cache/<asset-platform>` folder. We don't load the bundled asset paks until after the Settings Registry has determined all of these file paths, therefore the location of the engine would not be known at that time.

We can notifiy the user that the engine root could not be found, but we still need to set the engine to a value in `MergeSettingsToRegistry_AddRuntimeFilePaths` as a fallback.

Templates/DefaultProject/Template/cmake/VersionUtils.cmake
AMZN-alexpete: I was considering putting the compatibility check logic inside the engine to keep the engine finder in the project more simple.  e.g. for every engine path:

1. check if a `<engine root>/cmake/ProjectCompatibility.cmake` exists , and if so, include it.  Inside that file the engine would look at vars like `O3DE_ENGINE_NAME_TO_USE` and friends and would set some variable called `O3DE_COMPATIBLE_ENGINE_VERSION`  and `O3DE_COMPATIBLE_ENGINE_PATH` etc. - but only if that `O3DE_COMPATIBLE_ENGINE_PATH`variable wasn't set, or if the engine had a higher compatible version.   
2. if the `<engine root>/cmake/ProjectCompatibility.cmake` wasn't found in the engine path then assume compatible if no `O3DE_COMPATIBLE_ENGINE_PATH` has already been set by an engine that did check compatibility.
image

lumberyard-employee-dm: That could work.
Technically the only code the project needs access to is the logic to lookup the list of paths of engines in the `o3de_manifest.json`.
Once a set of candidate engine paths are available, the engine can do the rest of the lifting.

The minimum set of operations the EngineFinder.cmake needs to support each of the following three use cases are as follows
1. The user specifies the `-DCMAKE_MODULE_PATH` through the CMake command line directly, well that is the complete set of "candidate engine paths" to check in this scenario
2. The `<project-root>/user/project.json` has an `engine_path` key, that is added to the set of "candidate engine paths"
3. The `<project-root>/user/project.json` doesn't have a `engine_path` key, in that case the project only needs to query the o3de_manifest.json for the set of engines from the "engines" array and append it to it's "candidate engine paths"

Afterwards the only other operations is looping over the "candidate engine paths" and including `<engine root>/cmake/ProjectCompatibility.cmake` for each engine.

AutomatedTesting/cmake/VersionUtils.cmake
AMZN-alexpete: yup, `2.1` would fail and also `3.0`

Code/Framework/AzCore/Tests/Settings/SettingsRegistryMergeUtilsTests.cpp
```diff
@@ -751,4 +752,277 @@ tags=tools,renderer,metal)"
...
+                },
+
+                // project/project.json
+                R"({ "project_name": "TestProject", "engine":"o3de"  })",
+
+                // project/user/project.json
+                R"({})",
+
+                0 // first engine
+            },
```
lumberyard-employee-dm, line 868: Can we add a warning when an engine with the same name and version is a candidate?

Code/Framework/AzCore/Tests/Settings/SettingsRegistryMergeUtilsTests.cpp
```diff
@@ -751,4 +752,277 @@ tags=tools,renderer,metal)"
...
+            auto tempRootFolder = AZ::IO::FixedMaxPath(m_testFolder.GetDirectory());
+            AZ::IO::FixedMaxPath o3deManifestFilePath = tempRootFolder / "o3de" / "o3de_manifest.json";
+            AZ::IO::FixedMaxPath projectPath = tempRootFolder / "project" ;
+            AZ::IO::FixedMaxPath projectManifestPath = projectPath / "project.json";
+            AZ::IO::FixedMaxPath projectUserPath = tempRootFolder / "project" /  "user";
+            AZ::IO::FixedMaxPath userProjectManifestPath = projectUserPath / "project.json";
+
+            for (int i = 0; i < params.m_engineManifestsJson.size(); ++i)
+            {
+                const AZ::IO::FixedMaxPath enginePath = tempRootFolder / AZStd::string::format("engine%d",i) ;
```
lumberyard-employee-dm, line 963: nitpick: missing whitespace around comma
```suggestion
                const AZ::IO::FixedMaxPath enginePath = tempRootFolder / AZStd::string::format("engine%d" , i) ;
```

lumberyard-employee-dm (approved):

scripts/o3de/o3de/register.py
```diff
@@ -364,15 +316,8 @@ def register_engine_path(json_data: dict,
     def transform_engine_dict_to_string(engine): return engine.get('path', '') if isinstance(engine, dict) else engine
     json_data['engines'] = list(map(transform_engine_dict_to_string, engine_list))
 
-    result = register_o3de_object_path(json_data, engine_path, 'engines', 'engine.json',
+    return register_o3de_object_path(json_data, engine_path, 'engines', 'engine.json',
```
lumberyard-employee-dm, line 319: While you are in here can fix the `register_all_gems_in_folder`, `register_all_projects_in_folder`, `register_all_templates_in_folder` and `register_all_restricted_in_folder` methods to accept the force parameter: https://github.com/o3de/o3de/blob/c83090278a0901fed2b410b1de09bdef6e27da6d/scripts/o3de/o3de/register.py#L839-L845

The previous change to add version dependency checking has broken those functions, as they register a gem one at a time.

Code/Framework/AzCore/AzCore/Settings/SettingsRegistryMergeUtils.cpp
AMZN-alexpete: The way I addressed this was to set the engine path to the project path in `MergeSettingsToRegistry_AddRuntimeFilePaths` if the engine root is empty - I'm still kinda on the fence about it because imho it is a bit confusing that the engine root could ever be the project root - but that's packaged builds for you.

lumberyard-employee-dm: Well in theory in a packaged build, the "engine root" shouldn't ever be used.
There is really on the project root which is the folder containing game or simulation content.

scripts/o3de/o3de/register.py
AMZN-alexpete: Good catch - this change will be merged separately in this PR #14511

AMZN-alexpete:
Remaining known tasks
1. test moving compatibility logic to engine's `cmake/ProjectCompatibility.cmake`
2. update remaining templates
3. more testing
4. investigate `cli project upgrade` effort - might be a separate PR because this one is getting really big

Code/Framework/AzCore/AzCore/Dependency/Dependency.h
AMZN-stankowi: It's used by the XML schema system. This is a data driven solution for searching for product dependencies in XML files. This was most valuable with CryEngine code because there were many different XML file types supported by the engine, but it continues to be useful for anyone who has an XML file they don't process to binary and instead just copy to the asset cache and load at runtime.

Two good entry points for that system are here: https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/AzFramework/Asset/XmlSchemaAsset.cpp and here: https://github.com/o3de/o3de/blob/development/Gems/LmbrCentral/Code/Source/Builders/CopyDependencyBuilder/XmlBuilderWorker/XmlBuilderWorker.cpp

We've talked about expanding this to JSON, CSV, and other formats that are often just copied to the cache and not modified in the future. There's no immediate need for that, but I figured I would mention because the system was built to be grown forward to handle other data formats, too.

AutomatedTesting/cmake/EngineFinder.cmake
```diff
@@ -13,66 +13,73 @@
...
+        endif()
     endif()
-    string(JSON user_engine_path ERROR_VARIABLE json_error GET ${user_project_json} "engine_path")
 endif()
 
-# Extract optional version specifier in the engine name 
-o3de_get_version_specifier_parts("${O3DE_ENGINE_NAME_TO_USE}" O3DE_ENGINE_NAME_TO_USE O3DE_ENGINE_SPECIFIER_OP O3DE_ENGINE_SPECIFIER_VERSION)
+# Save the project json data with overrides in a global property to avoid performance hit
+# from loading the file multiple times
+set_property(GLOBAL PROPERTY O3DE_PROJECT_JSON ${o3de_project_json})
```
lumberyard-employee-dm, line 43: Seems like this can be stored in the globasl property even before checking the `<project>user/project.json` file

Templates/DefaultProject/Template/cmake/EngineFinder.cmake
```diff
@@ -13,66 +13,70 @@
...
 endif()
 
-# Extract optional version specifier in the engine name 
-o3de_get_version_specifier_parts("${O3DE_ENGINE_NAME_TO_USE}" O3DE_ENGINE_NAME_TO_USE O3DE_ENGINE_SPECIFIER_OP O3DE_ENGINE_SPECIFIER_VERSION)
+# Save the project json data with overrides in a global property to avoid performance hit
+# from loading the file multiple times
+set_property(GLOBAL PROPERTY O3DE_PROJECT_JSON ${o3de_project_json})
 
-# for backward compatibility
+# For backwards compatibility
```
lumberyard-employee-dm, line 45: The backwards compatibility isn't needed in the project templates since this is for a new project

Templates/DefaultProject/Template/cmake/EngineFinder.cmake
```diff
@@ -13,66 +13,70 @@
...
-        # always be verbose on this compatibility check
-        o3de_engine_compatible(${user_engine_path} ${O3DE_ENGINE_NAME_TO_USE} ${O3DE_ENGINE_SPECIFIER_OP} ${O3DE_ENGINE_SPECIFIER_VERSION} is_compatible engine_version TRUE)
-        if(is_compatible)
-            message("-- Selecting engine from 'user/project.json' at '${user_engine_path}'")
-            list(APPEND CMAKE_MODULE_PATH "${user_engine_path}/cmake")
-            return()
+# Option 2: Use the engine from the 'engine_path' field in <project>/user/project.json
+if(user_project_json)
+    string(JSON user_project_engine_path ERROR_VARIABLE json_error GET ${user_project_json} engine_path)
+    if(user_project_engine_path AND NOT json_error)
```
lumberyard-employee-dm, line 61: Can we add a else block as well that logs if the `<project>/user/project.json` file has a json error in it?

Previous experience has shown that users will hand edit these JSON files and they will have errors in them like missing closing quotes or a additional commas.

Templates/MinimalProject/Template/CMakeLists.txt
```diff
@@ -13,11 +13,11 @@ if(NOT PROJECT_NAME)
 
     # Utility function to look for an optional 'engine_finder_cmake_path'setting 
     function(get_engine_finder_cmake_path project_json_file_path path_value)
-        if(NOT $path_value AND EXISTS "${project_json_file_path}")
+        if(NOT path_value AND EXISTS "${project_json_file_path}")
             file(READ "${project_json_file_path}" project_json_data)
             string(JSON engine_finder_cmake_value ERROR_VARIABLE json_error GET ${project_json_data} "engine_finder_cmake_path")
             if(NOT json_error AND EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/${engine_finder_cmake_value}")
-                set(${path_value} "${engine_finder_path}" PARENT_SCOPE)
+                set(${path_value} "${CMAKE_CURRENT_SOURCE_DIR}/${engine_finder_cmake_value}" PARENT_SCOPE)
```
lumberyard-employee-dm, line 20: This needs to use `cmake_path` in order to get the correct path replacement logic when the `engine_finder_cmake_value` variable is an absolute path
https://cmake.org/cmake/help/latest/command/cmake_path.html?highlight=cmake_path#append

lumberyard-employee-dm (changes requested):

AutomatedTesting/cmake/EngineFinder.cmake
AMZN-alexpete: I removed this so that `EngineFinder.cmake` doesn't set these global properties, so now it's just the `o3deConfigVersion.cmake` files that do that.

lumberyard-employee-dm (approved):

Code/Framework/AzFramework/AzFramework/ProjectManager/ProjectManager.cpp
```diff
@@ -20,10 +20,9 @@
 
 namespace AzFramework::ProjectManager
 {
-    AZStd::tuple<AZ::IO::FixedMaxPath, AZ::IO::FixedMaxPath> FindProjectAndEngineRootPaths(const int argc, char* argv[])
+    AZ::IO::FixedMaxPath FindProjectAndEngineRootPaths(const int argc, char* argv[])
```
amzn-phist, line 23: Rename to just `FindProjectRootPath`?

cmake/Findo3de.cmake
```diff
@@ -21,24 +23,35 @@ o3de_current_file_path(current_path)
...
 # is also normalized.  They should match to avoid errors on some build systems.
 cmake_path(SET engine_json_path NORMALIZE ${current_path}/../engine.json)
-file(READ ${engine_json_path} engine_json)
 set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS ${engine_json_path})
 
-string(JSON this_engine_name ERROR_VARIABLE json_error GET ${engine_json} engine_name)
-if(json_error)
-    message(FATAL_ERROR "Unable to read key 'engine_name' from '${engine_json_path}', error: ${json_error}")
-endif()
+if(NOT LY_ENGINE_NAME_TO_USE)
```
amzn-phist, line 28: Can we change this name to `O3DE_ENGINE_NAME_TO_USE`?

cmake/Findo3de.cmake
```diff
@@ -21,24 +23,35 @@ o3de_current_file_path(current_path)
...
-    found_matching_engine
-)
+    if(this_engine_name STREQUAL LY_ENGINE_NAME_TO_USE)
+        set(found_matching_engine TRUE)
+    else()
+        message(VERBOSE "Project engine name '${LY_ENGINE_NAME_TO_USE}' does not match this engine name '${this_engine_name}' in '${engine_json_path}'")
+    endif()
+
+    find_package_handle_standard_args(o3de
+        "The engine name for this engine '${this_engine_name}' does not match the projects engine name '${O3DE_ENGINE_NAME_TO_USE}'."
```
amzn-phist, line 51: You have `O3DE_ENGINE_NAME_TO_USE` here, so I was a bit confused.

amzn-phist (approved):
Some more minor comments, but LGTM.

Code/Framework/AzCore/Tests/Settings/SettingsRegistryTests.cpp
```diff
@@ -1325,7 +1325,7 @@ namespace SettingsRegistryTests
             EXPECT_EQ(1, value);
         };
         auto testNotifier1 = m_registry->RegisterNotifier(callback);
-        bool result = m_registry->MergeSettingsFile(path.Native(), AZ::SettingsRegistryInterface::Format::JsonMergePatch, {}, nullptr);
+        bool result = m_registry->MergeSettingsFile(path.Native(), AZ::SettingsRegistryInterface::Format::JsonMergePatch, {}, nullptr).IsSuccess();
         ASSERT_TRUE(result);
```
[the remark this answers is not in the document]
AMZN-alexpete: Interestingly, AZ::Outcome does not support implicit bool conversion, just explicit, so removing the `.IsSuccess()` results in 
`Error	C2440	'initializing': cannot convert from 'AZ::Outcome<void,AZStd::string>' to 'bool`
My initial thought is this is to solve ["the safe bool problem"](https://en.cppreference.com/w/cpp/language/implicit_conversion#The_safe_bool_problem), which would allow code like this to compile also:
```
int result = AZ::Outcome<void, AZStd::string>();
auto AZ::Outcome<void, AZStd::string> someOutcome;
someOutcome << 1;
```

Re: `ASSERT_TRUE`, possibly? It could be the `ASSERT_TRUE` statements are for code required for setup, and those functions are tested in a different test, and the `EXPECT_TRUE` statements further down are testing the actual expectations of each test.

cmake/Findo3de.cmake
AMZN-alexpete: oops that's a bug, good catch.

cmake/Findo3de.cmake
AMZN-alexpete: I left it in there for backwards compatibility but the new code no longer defines `O3DE_ENGINE_NAME_TO_USE`, rather `Version.cmake` defines `O3DE_ENGINE_NAME`.

Code/Framework/AzCore/Tests/Settings/SettingsRegistryTests.cpp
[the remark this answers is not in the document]
amzn-phist: Odd.  Then why are the other tests below not producing a similar error?  E.g.
```
        auto result = m_registry->MergeSettingsFile(subPath, AZ::SettingsRegistryInterface::Format::JsonMergePatch, {}, nullptr);
        EXPECT_TRUE(result);
```
Doesn't use `.IsSuccess()` and appears to do implicit bool conversion.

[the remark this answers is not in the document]
amzn-phist: Oh duh.  There's a `bool result` in both of these lines instead of `auto result` which would be an `AZ::Outcome`.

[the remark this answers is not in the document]
AMZN-alexpete: Ya know, I may just update em to use auto and have a separate variable name for the bools

amzn-phist (approved):
Thanks!

AMZN-Phil (approved):

lumberyard-employee-dm (approved):

PR #14271, opened 2023-01-27
Improve streaming pool cvars

mbalfour-amzn:
## What does this PR do?

Changed the system buffer pool initialization to use the memory budget value from the cvar, so that way cvar values that were set from .cfg files would be taken into account.

Also added the flag to disable replication for the pool buffer cvars. Since these are rendering cvars, they seem like they should be controlled per-client instead of from the server.

## How was this PR tested?

Tried setting the cvar in a client.cfg and verified that the value "stuck". 
Used the cvar in a multiplayer session and verified that the server didn't override the local value.


rgba16f (approved):

mbalfour-amzn:
Closing the PR, after discussion with @VickyAtAZ this is a bad idea because the .setreg file should take precedence over the CVAR, not the other way around. And if the CVAR is only being used to change the value from within the Editor, the replication flag change isn't really important and might actually be the opposite of what's desired.

Issue #14169, opened 2023-01-23
Windowed Editor crashes when resized


LB-KacperMetelski:
**Describe the bug**
The Editor crashes when it's window is resized. This issue occurs only with an open level.

This issue occurs on both Windows and Linux.

**Call stack trace**
```
 	Qt5Gui.dll!00007ffb29d2c45c()	Unknown
 	Qt5Gui.dll!00007ffb29d29fc1()	Unknown
 	Qt5Gui.dll!00007ffb29d317e8()	Unknown
 	Qt5Gui.dll!00007ffb29d14c2b()	Unknown
>	AzQtComponents.dll!AzQtComponents::BreadCrumbs::sizeHint() Line 215	C++
 	AzQtComponents.dll!AzQtComponents::BreadCrumbs::resizeEvent(QResizeEvent * event) Line 361	C++
 	[External Code]	
 	AzQtComponents.dll!AzQtComponents::BreadCrumbs::resizeEvent(QResizeEvent * event) Line 363	C++
 	[External Code]	
 	The maximum number of stack frames supported by Visual Studio has been exceeded.	
```

**Steps to reproduce**
1. Launch the Editor.
2. Open or create any level.
3. Make sure the Editor's window is not maximized.
4. Resize the window (might take a couple of times).

**Expected behavior**
The Editor remains stable and is resized without issues.

**Actual behavior**
The Editor crashes.

**Screenshots/Video**
![image](https://user-images.githubusercontent.com/110163473/214057228-f7a3281e-6d98-4268-a8ca-c370369ff4d3.png)

https://user-images.githubusercontent.com/110163473/214047589-193188c7-5542-4e05-98e4-948f355d0500.mp4

**Found in Branch**
Development

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
83cf889

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Windows 
 - Version: 11 22H2
 - CPU: I7-11700f
 - GPU: Geforce RTX3070 MSI 
 - Memory: 32GB
 
**Additional context**
[Editor.zip](https://github.com/o3de/o3de/files/10480543/Editor.zip)

AMZN-daimini:
From a first investigation, it sounds like the issue may have been introduced by this PR: https://github.com/o3de/o3de/pull/13314

[...]

PR #14273, opened 2023-01-27
Fix for editor resize crash

dtamkin1:

## What does this PR do?

Fix for #14169 

## How was this PR tested?

Locally


Code/Framework/AzQtComponents/AzQtComponents/Components/Widgets/BreadCrumbs.cpp
```diff
@@ -107,6 +107,7 @@ namespace AzQtComponents
         m_label->setSizePolicy(QSizePolicy::Ignored, QSizePolicy::Preferred);
         m_lineEdit->setSizePolicy(QSizePolicy::Ignored, QSizePolicy::Preferred);
         connect(m_label, &QLabel::linkActivated, this, &BreadCrumbs::onLinkActivated);
+        setMinimumWidth(25);
```
cgalvan, line 110: I would add a comment explaining the crash and how this fixes it so someone doesn't remove it on accident :)

cgalvan (approved):

AMZN-daimini (approved):
Approving but also recommending to add a comment as cgalvan suggested.

PR #14274, opened 2023-01-27
Rework Audio Gem CMake targets, add .API public interface

amzn-phist:
## What does this PR do?

This refactors the CMake targets for the `AudioSystem` and `AudioEngineWwise` Gems.  The primary change is the addition of an `AudioSystem.API` target which provides public interface headers for other targets to depend on.  This moves audio headers out of CryCommon and into the public include path of the Gem.

There is still a dependency on `Legacy::CryCommon` for certain things, but those will be easier to remove after this change since the primary reason for the dependency (access to public audio headers) is now gone.

This also introduces an `AudioEngineWwise.Builder` target which only includes the builder source files and the module cpp.  This gives clearer separation of duties with the various output targets.  So the builder dll doesn't include any of the runtime or editor code, and similarly the Editor dll doesn't include builder code.

## How was this PR tested?

Built all targets in both affected Gems.  Ran all `.Tests` as well.  Ran Editor with a test project and was able to play sounds as before.  Will continue to test further.


Code/Legacy/CrySystem/CMakeLists.txt
```diff
@@ -33,6 +33,8 @@ ly_add_target(
             Legacy::CrySystem.XMLBinary
             Legacy::RemoteConsoleCore
             AZ::AzFramework
+        PUBLIC
+            Gem::AudioSystem.API
```
lumberyard-employee-dm, line 37: This is fine ... for now.
It is better than the Gem using CrySystem as a dependency.

Gems/AudioEngineWwise/Code/CMakeLists.txt
```diff
@@ -153,44 +157,98 @@ if (PAL_TRAIT_BUILD_HOST_TOOLS)
...
+                    Source/Engine
+                    ${pal_dir}
             BUILD_DEPENDENCIES
+                PUBLIC
+                    Gem::AudioSystem.API
                 PRIVATE
                     AZ::AzTest
-                    Gem::AudioEngineWwise.Editor.Static
+                    Gem::${gem_name}.Editor.Private.Object
+                    Gem::${gem_name}.Builder.Private.Object
```
lumberyard-employee-dm, line 243: This test needs the private Builder STATIC targets as well?

lumberyard-employee-dm (approved):

Code/Legacy/CrySystem/CMakeLists.txt
amzn-phist: Yes, the idea is to further remove calls to audio system in CrySystem later.

Gems/AudioEngineWwise/Code/CMakeLists.txt
amzn-phist: This test module combines Editor and Builder tests together, rather than having separate Editor.Tests and Builder.Tests.  So it effectively contains the same code as before.

lumberyard-employee-dm: Thanks for the clarification.

byrcolin (approved):

PR #14275, opened 2023-01-27
Update TAA to make a copy of its output for next frame's history.

invertednormal:
## What does this PR do?

Previously, the TAA pass would swap between two images to manage its history buffer. The current frame's output would become the next frame's history, and the current frame's history would be overwritten on the next frame. This avoids needing to copy the image to the history buffer every frame. However, this method had a problem due to the way the pass system currently works.

The pass system has no way to mark textures as read-only, which means when a pass's output is used as the input for another pass, that pass may overwrite the same image with its own modifications. This was causing TAA's output buffer to be overwritten by all the postfx passes that came after it - bloom, dof, etc. So unfortunately, a copy needs to be added for now to avoid this problem.

I implemented this change in a way that keeps the old method around, so we should be able to easily revert back when the pass system is improved.

## How was this PR tested?

Tested primary on the loft scene which has a lot of post fx to ensure there was no longer a feedback loop on the post fx output being used as the history buffer.

### DIFF
--- a/Gems/Atom/Feature/Common/Code/Source/PostProcessing/TaaPass.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/PostProcessing/TaaPass.cpp
@@ -69,10 +69,29 @@ namespace AZ::Render
 
         m_shaderResourceGroup->SetConstant(m_constantDataIndex, cb);
 
-
         Base::CompileResources(context);
     }
-    
+
+    void TaaPass::BuildCommandListInternal(const RHI::FrameGraphExecuteContext& context)
+    {
+        Base::BuildCommandListInternal(context);
+        if (ShouldCopyHistoryBuffer)
+        {
+            context.GetCommandList()->Submit(m_copyItem);
+        }
+    }
+
+    void TaaPass::SetupFrameGraphDependencies(RHI::FrameGraphInterface frameGraph)
+    {
+        if (ShouldCopyHistoryBuffer)
+        {
+            // Override the estimated item count to include the copy item.
+            frameGraph.SetEstimatedItemCount(2);
+        }
+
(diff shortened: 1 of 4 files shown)


rgba16f (approved):

amzn-tommy (approved):

Gems/Atom/Feature/Common/Code/Source/PostProcessing/TaaPass.cpp
```diff
@@ -69,10 +69,29 @@ namespace AZ::Render
...
+
+    void TaaPass::SetupFrameGraphDependencies(RHI::FrameGraphInterface frameGraph)
+    {
+        if (ShouldCopyHistoryBuffer)
+        {
+            // Override the estimated item count to include the copy item.
+            frameGraph.SetEstimatedItemCount(2);
+        }
+
+        Base::SetupFrameGraphDependencies(frameGraph);
```
VickyAtAZ, line 92: I think you need to change the order of these two calls. 
The Base::SetupFrameGraphDependencies will call frameGraph.SetEstimatedItemCount(1); which override your frameGraph.SetEstimatedItemCount(2);

VickyAtAZ (changes requested):
One minor issue.

Gems/Atom/Feature/Common/Code/Source/PostProcessing/TaaPass.cpp
invertednormal: While technically not required since `SetEstimatedItemCount()` isn't called by any of the base classes, I'll go ahead and change this just in case they are updated some day to do that.

VickyAtAZ (approved):

PR #14282, opened 2023-01-28
Improve detach prefab workflow with selective undo nodes

forkercat:

## What does this PR do?

This PR improves detach prefab workflow using the selective undo nodes to update prefab template data. Previously, we used PrefabUndoInstance node to update everything at once.

Similar PR (create-prefab): https://github.com/o3de/o3de/pull/12506

### What issues are fixed and improved?
1. After the operation, the template DOM is now always in sync with object data with the correct entity & instance DOM info. So the prefab DOM won't be corrupt after the operation.
    1. **Before this PR:** `no registered owning instance` occurs when undoing detach OR detach a prefab with nested prefabs. It is because sort order array contains `""`.
1. Removes link id before instance DOM comparison to prevent generating a link id override patch
    1. **Before this PR:** An expected add link id patch generated when detach a prefab with nested prefabs
1. It should improve the performance if the parent instance contains many entities and instances, since we no longer compare large DOMs to generate patches.

### What issues are not fixed?
1. Parent entity and detached container entity are still reloaded in next template propagation. It is needed to maintain sort order after detaching. More info is in comment. Will create a GHI for this.
    - So, if we do not reload the entity sort array in parent entity, the sort order is corrupt in Outliner and internally the sort array contains invalid data, that would fail the new unit test:
    - > PrefabDetachPrefabTests.cpp(258): error: Expected equality of these values:
  entityOrderArrayAfterDetach.size()
    Which is: 2
  3

## How was this PR tested?

- [x] Manually test detach prefab in several test cases
- [x] Pass all existing unit and automated tests
- [x] Add missing unit tests to test the workflow. This includes testing child entity order for parent entity and detached container entity. **UPDATE: Added 5 tests**


hultonha (commented):
Overall change looks very solid and awesome to see the thorough integration test coverage. It would be cool to see the output from the tests (and to see any failing tests with the original logic) but otherwise looks good to me! 👍

Code/Framework/AzToolsFramework/Tests/Prefab/PrefabDetachPrefabTests.cpp
```diff
@@ -0,0 +1,256 @@
...
+        AZ::IO::Path carPrefabFilepath = engineRootPath / carPrefabName;
+        AZ::IO::Path wheelPrefabFilepath = engineRootPath / wheelPrefabName;
+
+        AZ::EntityId engineEntityId = CreateEditorEntityUnderRoot(engineEntityName);
+        AZ::EntityId tireEntityId = CreateEditorEntityUnderRoot(tireEntityName);
+        AZ::EntityId wheelContainerId = CreateEditorPrefab(wheelPrefabFilepath, { tireEntityId });
+        AZ::EntityId batteryEntityId = CreateEditorEntityUnderRoot(batteryEntityName);
+        AZ::EntityId carContainerId = CreateEditorPrefab(carPrefabFilepath, { engineEntityId, wheelContainerId, batteryEntityId });
+
+        // Validate child entity order under car before detaching the car prefab.
```
forkercat, line 213: This part is failing because entity order issue in create-prefab. It should pass once this PR https://github.com/o3de/o3de/pull/14184 gets merged.

Code/Framework/AzToolsFramework/Tests/Prefab/PrefabDetachPrefabTests.cpp
```diff
@@ -0,0 +1,256 @@
...
+        AZ::TransformBus::EventResult(parentEntityIdToCheck, tireEntityIdAfterDetach, &AZ::TransformInterface::GetParentId);
+        EXPECT_EQ(carEntityIdAfterDetach, parentEntityIdToCheck);
+    }
+
+    TEST_F(PrefabDetachPrefabTests, DetachPrefabWithNestedPrefabSucceeds)
+    {
+        // Level
+        // | Car       (prefab)  <-- detach prefab
+        //   | Wheel   (prefab)
+        //     | Tire
```
srikappa-amzn, line 63: Another important use case to test would be to detach a prefab that is parented to another entity within the owning prefab. Example:

- Level
  - entity1
    - car.prefab <-- detach
      - wheel entity

srikappa-amzn (approved):
Nice job adding the tests!

Looks good to me. Gave a few suggestions about improving test coverage for detach

forkercat:
> It would be cool to see the output from the tests (and to see any failing tests with the original logic)

@hultonha I updated the description to include the information. The original workflow does not have this issue since it already reloads parent entity. This is a new issue I discovered in the new workflow not doing the reloading. I could verify the test would fail. So, `updateCache=false` is introduced to make it reload to pass the test.

forkercat:
I was planning to add a helper function to validate child entity sort order to encapsulate EXPECT_EQs in a for-loop and can print out array values for expected and actual data for better readability, but I held back doing it until we see more usage to validate sort order in prefab tests. @michabr

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabUndoHelpers.cpp
```diff
@@ -97,11 +108,12 @@ namespace AzToolsFramework
                 const PrefabDomValue& entityDomBeforeUpdatingEntity,
                 const PrefabDomValue& entityDomAfterUpdatingEntity,
                 AZ::EntityId entityId,
-                UndoSystem::URSequencePoint* undoBatch)
+                UndoSystem::URSequencePoint* undoBatch,
+                bool updateCache)
```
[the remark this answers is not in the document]
forkercat: > an `enum` is an improvement or a separate function potentially

I think it is a good suggestion. But I anticipate that this `UpdateEntity` function will be changed and improved very soon when the sort order rework is in. I would prefer doing that later.

I'm curious how the `enum` can be used here. As this parameter only has two states of values, then would it be an enum with just two values?

hultonha:
> > It would be cool to see the output from the tests (and to see any failing tests with the original logic)
> 
> @hultonha I updated the description to include the information. The original workflow does not have this issue since it already reloads parent entity. This is a new issue I discovered in the new workflow not doing the reloading. I could verify the test would fail. So, `updateCache=false` is introduced to make it reload to pass the test.

Awesome stuff! 😄 Thanks for doing this @forkercat, much appreciated! 👍

hultonha (approved):

Issue #13667, opened 2022-12-13
Material Type Hot Reload Is Broken in Material Editor


santorac:
**Describe the bug**
Material Editor does not hot reload changes to the .materialtype and .shader files. This was working before.

(Note some of these fail in Editor.exe too).

**Assets required**
Use the AtomSampleViewer or AutomatedTesting project, both have access to the TestData materials referenced here.
You'll also need this material file (remove the ".txt" extension): [TestMaterialReload.material.txt](https://github.com/o3de/o3de/files/10214480/TestMaterialReload.material.txt)

**Steps to reproduce**
1. Open Material Editor
2. Use the lighting preset called "Test", and the "Shaderball" model preset.
3. Open TestMaterialReload.material in Material Editor. Notice the "glowie" parts around the shadow edges, this is subsurface scattering.
4. Option 1
    1. Open TestMaterialReload.material file in a text editor.
    2. Change subsurfaceScattering.enableSubsurfaceScattering to false in notepad, and save.
    3. Wait for the Asset Processor to finish
    4. Nothing changes in Material Editor's preview window (you might see the asset browser thumbnail update though)
    5. In Material Editor, close and open TestMaterialReload.material again.
    6. Now subsurface scattering is gone.
    7. (Set enableSubsurfaceScattering back to true, save, and reload if you want to continue with the other options)
5. Option 2
    1. Open SubsurfaceAndTransmissionPropertyGroup.json in a text editor.
    2. Change scatterColor defaultValue to "[0.0, 1.0, 0.0]" and save
    3. Wait for the Asset Processor to finish
    3. Nothing changes in Material Editor's preview window (you might see the asset browser thumbnail update though)
    4. In Material Editor, close and open TestMaterialReload.material again.
    5. Now the surface is green
    6. (Set scatterColor defaultValue back to the prior value, save, and reload if you want to continue with the other options)
6. Option 3 (This one also fails in Editor.exe)
    1. Open EnhancedPBR_ForwardPass_EDS.shader in a text editor
    2. Change "DrawList" from "forwardWithSubsurfaceOutput" to "forward" and save
    3. Wait for the Asset Processor to finish
    4. Nothing changes in Material Editor's preview window
    5. In Material Editor, close and open TestMaterialReload.material again.
    6. Now subsurface scattering is gone.
    7. (Set enableSubsurfaceScattering back to true, save, and reload if you want to continue with the other options)

**Expected behavior**
Material Editor should always reload changes to the material type or shaders.

[...]

PR #14288, opened 2023-01-30
Materials: Fixing problems with material hot reloading after shader and material type changes (WIP)

gadams3:
## What does this PR do?

This PR will address multiple problems with material asset dependencies, hot reloading, and stability. Without these changes, materials, material types, and shaders will not be synchronized when reloading, leading to performance loss, spam related to mismatching material property layouts, spam and rendering anomalies because of missing or mismatching material SRGs, materials not refreshing after changes in material canvas or manual edits to shaders and material types.

Changes include:

Code and options have been removed that allowed opting out of fully baking materials at build time. This prevented asset dependencies and notifications from being propagated correctly and the correct version of assets from being referenced consistently. The material finalization settings were intended to prevent changes made to low level, common shader code utilized by the standard PBR material type from triggering model assets dependent on standard PBR from rebuilding. Instead, the intention is to remove the dependency on standard PBR from the model builder.

Material initialization resets member variables pointing to property layouts in shader resource groups. Without this, reinitialized materials would retain pointers to the previous version of those objects. In the case of material canvas, a generated material may not initially have a material SRG. Once material inputs are added, the material SRG is generated but the reinitialized material ignored it. The same situation would arise if material inputs were moved or the layout changed in any way. The result was log spam, objects rendering incorrectly, if at all, and Instability.

Exploring removing shader reload notification support from the material and asset classes. Handling these notifications attempts to reinitialize the material by passing in its own outdated version of the asset. By doing this, it’s also passing in an outdated version of the material type, shaders, and other assets unless those embedded assets are also handling the shader reload notifications and updating themselves internally. Either way, it’s not the fully baked version of the material being passed in. Instead, high level systems like the material component and other material providers can create a new material instance or reinitialize after they receive material ready and reload notifications from the asset bus.

Resolves https://github.com/o3de/o3de/issues/13667

## How was this PR tested?

Updated and ran unit tests. All tests passed.
Testing and verified hot reloading of iterative changes in material canvas and material editor.
Preparing to test ASV.

Continuing to test these changes but so far material component and material canvas reloading is consistent and no longer reporting any errors or asserts. Materials refresh and reflect changes as expected.

### DIFF
--- a/Gems/Atom/RPI/Code/Source/RPI.Builders/Material/MaterialBuilder.cpp
+++ b/Gems/Atom/RPI/Code/Source/RPI.Builders/Material/MaterialBuilder.cpp
@@ -29,7 +29,7 @@ namespace AZ
 
         AZStd::string MaterialBuilder::GetBuilderSettingsFingerprint() const
         {
-            return AZStd::string::format("[BuildersShouldFinalizeMaterialAssets=%d]", MaterialBuilderUtils::BuildersShouldFinalizeMaterialAssets());
+            return "";
         }
 
         void MaterialBuilder::RegisterBuilder()
@@ -113,9 +113,10 @@ namespace AZ
 
             if (!parentMaterialPath.empty())
             {
-                // Register dependency on the parent material source file so we can load it and use it's data to build this variant material.
-                // Note, we don't need a direct dependency on the material type because the parent material will depend on it.
-                MaterialBuilderUtils::AddPossibleDependencies(request.m_sourceFile,
+                // Register dependency on the parent material source file so we can load it and use it's data to build this variant
+                // material. Note, we don't need a direct dependency on the material type because the parent material will depend on it.
+                MaterialBuilderUtils::AddPossibleDependencies(
+                    request.m_sourceFile,
                     parentMaterialPath,
                     JobKey,
                     outputJobDescriptor.m_jobDependencyList,
@@ -136,7 +137,8 @@ namespace AZ
                 // At this point the builder does not know which is the case, without loading the .materialtype file and inspecting its data. The builder
                 // avoids that because it could slow things down, and instead just registers both dependencies.
 
-                MaterialBuilderUtils::AddPossibleDependencies(request.m_sourceFile,
+                MaterialBuilderUtils::AddPossibleDependencies(
+                    request.m_sourceFile,
                     materialTypePath,
                     MaterialTypeBuilder::FinalStageJobKey,
                     outputJobDescriptor.m_jobDependencyList,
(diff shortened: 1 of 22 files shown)


galibzon (approved):

Gems/Atom/RPI/Code/Source/RPI.Builders/Material/MaterialBuilder.cpp
```diff
@@ -113,9 +113,10 @@ namespace AZ
 
             if (!parentMaterialPath.empty())
             {
-                // Register dependency on the parent material source file so we can load it and use it's data to build this variant material.
-                // Note, we don't need a direct dependency on the material type because the parent material will depend on it.
-                MaterialBuilderUtils::AddPossibleDependencies(request.m_sourceFile,
+                // Register dependency on the parent material source file so we can load it and use it's data to build this variant
+                // material. Note, we don't need a direct dependency on the material type because the parent material will depend on it.
+                MaterialBuilderUtils::AddPossibleDependencies(
```
rgba16f, line 118: This comment seems to conflict with the comment below the if, One says a dependecy on the material type is not needed and the other says it is.

Gems/Atom/RPI/Code/Source/RPI.Builders/Material/MaterialBuilderUtils.cpp
```diff
@@ -24,12 +24,9 @@ namespace AZ::RPI::MaterialBuilderUtils
...
     {
         bool dependencyFileFound = false;
-            
-        const bool currentFileIsMaterial = AzFramework::StringFunc::Path::IsExtension(currentFilePath.c_str(), MaterialSourceData::Extension);
-        const bool referencedFileIsMaterialType = AzFramework::StringFunc::Path::IsExtension(referencedParentPath.c_str(), MaterialTypeSourceData::Extension);
-        const bool shouldFinalizeMaterialAssets = MaterialBuilderUtils::BuildersShouldFinalizeMaterialAssets();
 
-        AZStd::vector<AZStd::string> possibleDependencies = RPI::AssetUtils::GetPossibleDepenencyPaths(currentFilePath, referencedParentPath);
+        AZStd::vector<AZStd::string> possibleDependencies =
+            RPI::AssetUtils::GetPossibleDepenencyPaths(currentFilePath, referencedParentPath);
```
rgba16f, line 29: performance nit: function returns a whole vector for 2 strings. Would be great to update to a fixed_vector of fixed_string's so no allocations are needed.

rgba16f (approved):

PR #14297, opened 2023-01-30
Action Manager | Fix issues that resulted in prefab focus and entity selection being lost when leaving component mode via Esc

AMZN-daimini:
## What does this PR do?
Introduces a few fixes to the Action Manager:

- Restored an old fix that prevents Shortcuts from being auto-repeated when the key sequence is held. This is a default in Qt, but given the peculiarities of O3DE it usually results in the Editor getting overloaded and is rarely intentional on the users' part. UX consensus was agreed when this fix was applied to the legacy Action Manager years ago, and is just being ported over;
- Fixed issue in the Action Context widget event filter where actions bound to a shortcut were identified and triggered in a single loop. That could result in undefined behavior since triggering an action could change the enabled state of other actions while the loop was still happening, making it so that which actions are executed depended on the order they were registered;
- Migrated the `Esc` actions for the entity selection and prefab focus mode to use Actions in the new Action Manager.
- Removed a workaround to deliver the `Esc` shortcut event via the Viewport that was necessary with the legacy ShortcutDispatcher but was now causing the shortcut to be triggered twice when sent via the viewport.

The current setup of the `Esc` keypress in the Editor partially still bypasses the Action Manager. This is fine since the Action Manager actions are unreachable in Game Mode, and the custom code enforces that `Esc` can be pressed to leave it. The other code path is now only used in Entity Pick mode, and even after migrating fully to the new Action Manager I would be ok with leaving it as is since Ctrl interactions and shortcuts are not (and probably should not) be migrated to the Action Manager.

## How was this PR tested?

Manual testing.

Before
RetainSelectionOnComponentModeEsc_before

After
RetainSelectionOnComponentModeEsc

Test steps:

(Need to have a level loaded with a nested prefab containing at least one entity with a component that has a component mode).

- Focus on the nested prefab
- Select the entity inside the nested prefab
- Switch to a Component Mode
- Press `Esc`

Expected result: Component Mode is closed, but the nested prefab is still in focus and the the entity selection persists.

Actual result (before this change): Component Mode, prefab focus and selection are all cleared.


cgalvan (approved):

hultonha (approved):
Changes look good but it would be great to see more tests to verify this change in behavior. While we're rewriting this new system it's definitely good to continue to focus on tests to make things more stable over the long term. Enumerating the manual testing steps would be useful in the PR description too.

AMZN-daimini:
> Changes look good but it would be great to see more tests to verify this change in behavior. While we're rewriting this new system it's definitely good to continue to focus on tests to make things more stable over the long term. Enumerating the manual testing steps would be useful in the PR description too.

Hum the specific test steps are a bit complex (need to open component mode while focusing on a nested prefab, then press `Esc`). I will add manual test steps but I'm not sure an automated test for such a compounded set of operations makes sense.

hultonha:
> > Changes look good but it would be great to see more tests to verify this change in behavior. While we're rewriting this new system it's definitely good to continue to focus on tests to make things more stable over the long term. Enumerating the manual testing steps would be useful in the PR description too.
> 
> Hum the specific test steps are a bit complex (need to open component mode while focusing on a nested prefab, then press `Esc`). I will add manual test steps but I'm not sure an automated test for such a compounded set of operations makes sense.

I just mean you could write a test to verify the functionality of the code you changed/fixed even in an integration test or unit test to verify the functionality. There's different levels we can test at and ensuring the new escape handling is working as expected would be super useful for example.

AMZN-daimini:
The new Escape handling is just treating `Esc` like any other shortcut, so that part is covered :)
Changes to the event filter are also still covered in the tests Chris added in the previous PR.

I'll give this another look to see if there's any blind spot that I can't think of right now, as soon as I have completed the other PR I'm working on :D

AMZN-daimini:
> The new Escape handling is just treating `Esc` like any other shortcut, so that part is covered :) Changes to the event filter are also still covered in the tests Chris added in the previous PR.
> 
> I'll give this another look to see if there's any blind spot that I can't think of right now, as soon as I have completed the other PR I'm working on :D

Double checked and I think existing tests cover this new use case. We will do a test run before switching this on by default to verify if any other issue comes up that could use regression tests.

hultonha:
> > The new Escape handling is just treating `Esc` like any other shortcut, so that part is covered :) Changes to the event filter are also still covered in the tests Chris added in the previous PR.
> > I'll give this another look to see if there's any blind spot that I can't think of right now, as soon as I have completed the other PR I'm working on :D
> 
> Double checked and I think existing tests cover this new use case. We will do a test run before switching this on by default to verify if any other issue comes up that could use regression tests.

Great to hear, thank you for double checking 🙂

PR #14298, opened 2023-01-30
Resolving intermittent test failures in Lua/SC Spawnables tests

jckand-amzn:
## What does this PR do?

- Adds new unique prefabs for each test as Game Mode collisions were causing unexpected entity transforms to be validated
- Moves tests to Main suite

## How was this PR tested?

- Ran updated tests 100x locally with 0 failures


cgalvan (approved):

Kadino (approved):
Excellent fix! Slightly increases the data footprint, but the stability tradeoff is worth it.

srikappa-amzn:
Maybe we should wait until the underlying gameentitycontext bug gets resolved? Much of this PR needs to be reverted and the new files deleted once we fix that. If we confirm that fixing that will take a lot of time, then we can do this temporary fix but it feels like we should at least attempt fixing the root cause now that we have found it.

i feel that this sets out a bad precedent to other test writers who may feel that they need a new prefab for writing a test around spawnables when that's not the case. We can't monitor all PRs and eventually the number of prefabs will grow in size. I think it's good to create files in memory if possible and check in files for testing only as a last resort. In this particular case, I feel that we don't need to add new files and make so many changes when the fix for the bug may end up being a few lines of code in a class.

jckand-amzn:
Moving this PR to draft as fixing the root cause of the intermittent failures may be an easier solution.

jckand-amzn:
Closing this PR out as the root cause was able to be fixed in https://github.com/o3de/o3de/pull/14889.

PR #14300, opened 2023-01-30
Add deprecated var aliases back to Version.cmake

AMZN-alexpete:

## What does this PR do?

Adds back deprecated CMake version vars as aliases till after the next release when they will be removed.

These version vars are still in use by the 3rd party packaging scripts 
https://github.com/o3de/3p-package-source/search?q=LY_VERSION_ENGINE_NAME

## How was this PR tested?

Ran CMake Configure


amzn-phist (approved):

lawsonamzn (approved):

cmake/Version.cmake
```diff
@@ -92,3 +92,9 @@ o3de_set_from_env_with_default(O3DE_INSTALL_VERSION_STRING O3DE_INSTALL_VERSION
 o3de_set_from_env_with_default(O3DE_INSTALL_DISPLAY_VERSION_STRING O3DE_INSTALL_DISPLAY_VERSION "@O3DE_DISPLAY_VERSION_STRING@" CACHE STRING "Open 3D Engine's display version for the INSTALL target")
 o3de_set_from_env_with_default(O3DE_INSTALL_BUILD_VERSION O3DE_INSTALL_BUILD_VERSION "@O3DE_BUILD_VERSION@" CACHE STRING "Open 3D Engine's build number for the INSTALL target")
 o3de_set_from_env_with_default(O3DE_INSTALL_ENGINE_NAME O3DE_INSTALL_ENGINE_NAME "@O3DE_ENGINE_NAME@" CACHE STRING "Open 3D Engine's engine name for the INSTALL target")
+
+# Deprecated version aliases
+set(LY_VERSION_COPYRIGHT_YEAR ${O3DE_COPYRIGHT_YEAR})
+set(LY_VERSION_STRING ${O3DE_VERSION_STRING})
+set(LY_VERSION_BUILD_NUMBER ${O3DE_BUILD_VERSION})
+set(LY_VERSION_ENGINE_NAME ${O3DE_ENGINE_NAME})
```
lumberyard-employee-dm, line 100: So there is a bit of a problem here.
Even if these `LY_VERSION_*` variables are added back, because they are not CMAKE Cache variables, specifying them via the `-DLY_VERSION_*` during CMake configure will not work and that means the documentation is still incorrect

AMZN-alexpete: Good call - I've created a GHI to update the docs https://github.com/o3de/o3de.org/issues/2214 and will start with the smaller change that we discusses which just sets the LY_VERSION_ENGINE_NAME alias for 3p packages.

AMZN-alexpete: Related PR https://github.com/o3de/o3de.org/pull/2215

lawsonamzn (approved):

PR #14303, opened 2023-01-30
Fixed default seeds: Removed missing files, added files that were required but not in the default list.

AMZN-stankowi:

## What does this PR do?

Fixes it so, creating a bundle using the default seeds command, will give you enough files for the engine to load.

## How was this PR tested?

I built a monolithic release build of the Automated Testing project, and got it to load with a bundle generated from the asset bundler that just had the default seeds.


cgalvan (approved):
Thanks!

Assets/Engine/SeedAssetList.seed
```diff
@@ -128,14 +128,6 @@
       <Class name="unsigned int" field="platformFlags" value="127" type="{43DA906B-7DEF-4CA8-9790-854106D3F983}"/>
       <Class name="AZStd::string" field="pathHint" value="fonts/default.font" type="{03AAAB3F-5C47-5A66-9EBC-D5FA4DB353C9}"/>
     </Class>
-    <Class name="SeedInfo" field="element" version="2" type="{FACC3682-2ACA-4AA4-B85A-07AD276D18A0}">
```
tkothadev, line 131: what assets did these seed hashes correspond to, if we can know?

tkothadev:
Was it only the seed list in the Atom gem that was causing issues for the bundler, or are there other seed lists we need to be concerned about (like in a game project)?

tkothadev (approved):

Assets/Engine/SeedAssetList.seed
```diff
@@ -232,14 +224,6 @@
       <Class name="unsigned int" field="platformFlags" value="255" type="{43DA906B-7DEF-4CA8-9790-854106D3F983}"/>
       <Class name="AZStd::string" field="pathHint" value="scripts/enginecommon.luac" type="{03AAAB3F-5C47-5A66-9EBC-D5FA4DB353C9}"/>
```
AMZN-alexpete, line 225: Can we remove this one too? I believe I've deleted this in the redcode change https://github.com/aws-lumberyard-dev/o3de/commit/adc2f3f14d123db215da57fd212b32278175342f#diff-4e5ea674abf1a32fc45dbe021f8052fb955314bd3858023400730febf38b1878

AMZN-alexpete (approved):
Thanks for cleaning these up - sorry I didn't remove enginecommon.lua/luac from the seedlist when I removed it with the redcode change!

Assets/Engine/SeedAssetList.seed
AMZN-alexpete: the .luac file is the compiled version of the .lua file and should no longer get generated or copied over

AMZN-stankowi:
> Was it only the seed list in the Atom gem that was causing issues for the bundler, or are there other seed lists we need to be concerned about (like in a game project)?

Just those, I don't get any errors, and the launcher loads with these changes.

Assets/Engine/SeedAssetList.seed
AMZN-stankowi: I just merged latest, I think you did already remove it there.

rgba16f (approved):

lumberyard-employee-dm (approved):
Should we convert these seed files over to use Json Serialization?

tkothadev:
that would be much more straightforward imo if we could do that

AMZN-stankowi:
> Should we convert these seed files over to use Json Serialization?

I'll create a backlog task for that on our internal tool

PR #11016, opened 2022-08-02
Finished PBR shader refactor

antonmic:
Final refactor: BasePBR, StandardPBR, EnhancedPBR, Skin and Eye have all been updated to use a new material shader authoring paradigm. This paradigm is split into 4 stages:
- Vertex (object to world space calculation, TBN calculation)
- PixelGeometry (things like tangents space calculation and parallax occlusion mapping)
- Surface (Albedo, roughness, specular, etc.)
- Lighting (executes the full light loop, also contains decals for the moment (could be separated out into it's own stage) 
The only tricky thing about this new paradigm is a bit of #define macro magic, though this is kept to a minimum.
None of the structures used by the code (VsInput, VsOutput, PixelGeometryData, Surface, LightingData) are defined directly by any file. Rather the files in question have a macro that, if not already defined, defines the structure in question to the definition in the file. Same thing for the functions. What this allows for is code reuse without name conflicts. EvaluateSurface_StandardPBR can call EvaluateSurface_BasePBR, and can #define EvaluateSurface to EvaluateSurface_StandardPBR.

In terms of functionality, all functionality for materials is exactly as it was, nothing has changed and all material and parallax screenshot tests in ASV pass.

Signed-off-by: antonmic <56370189+antonmic@users.noreply.github.com>

PR #14320, opened 2023-01-31
Fix ASV screenshot

jiaweig-amzn:

## What does this PR do?

1. Fix screenshot capture for secondary windows.
2. Fix issue that was exposed by multiscene sample

The multiscene sample started to fail from this commit:
https://github.com/o3de/o3de/pull/11016/commits/b0d199ad67aa50b798f23504dd77d9e4d6838341
It didn't introduce bug, but in fact exposed an older bug where there were different behavior after calculating specularAA, which is whether to set back linear roughness after calculating specular AA.

Check
[Gems/Atom/Feature/Common/Assets/Shaders/Materials/BasePBR/BasePBR_SurfaceData.azsli](https://github.com/o3de/o3de/pull/11016/files#diff-d7a09bf799d2de8986dda7232c78614abe97385120b888c75c0535df8780f78e) line 49 - 82
[Gems/Atom/Feature/Common/Assets/ShaderLib/Atom/Features/PBR/Surfaces/StandardSurface.azsli](https://github.com/o3de/o3de/pull/11016/files#diff-717ad64bab26699482e10183d3983f833f87771397d1c9c2c0fc8d66e6bae81b) line 70 - 100

The screenshot of multiscene used to not apply roughness back after calculating specular AA. Now it does after the PBR shader refactor. So the fix is updating the screenshot and fix other places that don't apply roughness back.
Screenshot is updated here: https://github.com/o3de/o3de-atom-sampleviewer/pull/577

## How was this PR tested?
2 errors are:
Multiscene screenshot, which is updated as mentioned above.
The other is in the backlog to clear. [Error] (PassSystem) - Do not remove child passes outside of the removal, reset, or build phases.
Automation: Running script 'scripts/RenderTargetTexture.bv.luac'...

ASV 1 31

### DIFF
--- a/Gems/Atom/Feature/Common/Assets/ShaderLib/Atom/Features/PBR/Surfaces/StandardSurface.azsli
+++ b/Gems/Atom/Feature/Common/Assets/ShaderLib/Atom/Features/PBR/Surfaces/StandardSurface.azsli
@@ -89,6 +89,9 @@ void Surface::ApplySpecularAA()
     float kernelRoughnessA2 = min(2.0 * variance , varianceThresh );
     float filteredRoughnessA2 = saturate ( roughnessA2 + kernelRoughnessA2 );
     roughnessA2 = filteredRoughnessA2;
+
+    roughnessA = sqrt(roughnessA2);
+    roughnessLinear = sqrt(roughnessA);
 }
 
 void Surface::CalculateRoughnessA()
--- a/Gems/Atom/Feature/Common/Code/Source/FrameCaptureSystemComponent.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/FrameCaptureSystemComponent.cpp
@@ -582,7 +582,7 @@ namespace AZ
             const AZStd::string& outputFilePath,
             RPI::PassAttachmentReadbackOption option)
         {
-            CaptureHandle captureHandle = InternalCapturePassAttachment(passHierarchy, slot, outputFilePath, option, AZStd::bind(&FrameCaptureSystemComponent::CaptureAttachmentCallback, this, AZStd::placeholders::_1));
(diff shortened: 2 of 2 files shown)


gadams3 (approved):

Gems/Atom/Feature/Common/Assets/ShaderLib/Atom/Features/PBR/Surfaces/StandardSurface.azsli
```diff
@@ -89,6 +89,9 @@ void Surface::ApplySpecularAA()
     float kernelRoughnessA2 = min(2.0 * variance , varianceThresh );
     float filteredRoughnessA2 = saturate ( roughnessA2 + kernelRoughnessA2 );
     roughnessA2 = filteredRoughnessA2;
+
+    roughnessA = sqrt(roughnessA2);
```
moudgils, line 93: Why is there CalculateSpecularAA and ApplySpecularAA both doing the same things but now we are modifying one? It all seems very confusing. Can we just remove one entirely and refactor it out a bit?

jiaweig-amzn: ApplySpecularAA exists in various surface classes before CalculateSpecularAA is created. But some have linear roughness set back, some don't, like this one. Now all the shaders except multilayer PBR in Feature\Common\Assets\Shaders\Materials have been refactored to use CalculateRoughnessValues, it calls CalculateSpecularAA and sets back roughness value.
I am not sure why multilayer PBR is left untouched. If we are doing another refactoring, I think it should go into another PR.

moudgils (approved):

PR #14340, opened 2023-02-01
Change the order of module initialization to stay the same in monolithic builds

mbalfour-amzn:
## What does this PR do?

In monolithic builds, the static engine modules (AzCore, AzFramework, AzNetworking, AzGameFramework) were getting initialized *after* the "dynamic" modules that become static, where in non-monolithic builds, they get initialized first.

This was causing a crash in the Multiplayer Gem in monolithic builds because it relied on AzNetworking being initialized first.

The reason for the order change is because the modules were getting added to the list in this order:
* ComponentApplication::CreateStaticModules()
** startupParameters.m_createStaticModulesCallback()
** AzCoreModule
* Application::CreateStaticModules()
** AzNetworkingModule
** AzFrameworkModule
* GameApplication::CreateStaticModules()
** AzGameFrameworkModule
* CreateDynamicModules()

In monolithic builds, the createStaticModulesCallback() contains all the modules that appear in CreateDynamicModules() in non-monolithic builds.

The fix is to change the various CreateStaticModules() calls to push their modules to the front instead of the end, so that they appear before both the createStaticModulesCallback() and the CreateDynamicModules() list.

## How was this PR tested?

Ran a monolithic and non-monolithic build and verified the module order appears correct in both, and Multiplayer no longer crashes in either scenario.

### DIFF
--- a/Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp
+++ b/Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp
@@ -1214,12 +1214,15 @@ namespace AZ
     //=========================================================================
     void ComponentApplication::CreateStaticModules(AZStd::vector<AZ::Module*>& outModules)
     {
+        // Add this to the front of the list. In monolithic builds, dynamic modules become static modules,
+        // so we want this to load before the dynamic modules load regardless of the build type.
+        outModules.insert(outModules.begin(), aznew AzCoreModule());
+
         if (m_startupParameters.m_createStaticModulesCallback)
         {
             m_startupParameters.m_createStaticModulesCallback(outModules);
         }
 
-        outModules.emplace_back(aznew AzCoreModule());
     }
 
     void ComponentApplication::LoadModules()
--- a/Code/Framework/AzFramework/AzFramework/Application/Application.cpp
+++ b/Code/Framework/AzFramework/AzFramework/Application/Application.cpp
@@ -404,10 +404,14 @@ namespace AzFramework
 
     void Application::CreateStaticModules(AZStd::vector<AZ::Module*>& outModules)
     {
+        // Add this to the front of the list. In monolithic builds, dynamic modules become static modules,
+        // so we want this to load before the dynamic modules load regardless of the build type.
+        // Note that because of the order we add these in, AzNetworkingModule will appear before AzFrameworkModule in the list.
+        outModules.insert(outModules.begin(), aznew AzFrameworkModule());
+        outModules.insert(outModules.begin(), aznew AzNetworking::AzNetworkingModule());
+
         AZ::ComponentApplication::CreateStaticModules(outModules);
 
-        outModules.emplace_back(aznew AzNetworking::AzNetworkingModule());
-        outModules.emplace_back(aznew AzFrameworkModule());
     }
 
     const char* Application::GetCurrentConfigurationName() const
--- a/Code/Framework/AzGameFramework/AzGameFramework/Application/GameApplication.cpp
+++ b/Code/Framework/AzGameFramework/AzGameFramework/Application/GameApplication.cpp
@@ -111,9 +111,11 @@ namespace AzGameFramework
 
     void GameApplication::CreateStaticModules(AZStd::vector<AZ::Module*>& outModules)
     {
-        AzFramework::Application::CreateStaticModules(outModules);
+        // Add this to the front of the list. In monolithic builds, dynamic modules become static modules,
+        // so we want this to load before the dynamic modules load regardless of the build type.
+        outModules.insert(outModules.begin(), aznew AzGameFrameworkModule());
 
-        outModules.emplace_back(aznew AzGameFrameworkModule());
+        AzFramework::Application::CreateStaticModules(outModules);
     }
 
     void GameApplication::QueryApplicationType(AZ::ApplicationTypeQuery& appType) const


AMZN-stankowi (approved):

lawsonamzn (approved):

cgalvan (approved):

lumberyard-employee-dm (approved):

lumberyard-employee-dm:
In theory, the order of module initialization shouldn't break other modules, but it appears the Multplayer Gem is causing an issue.

The way we initialize modules [DynamicModules](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp#L1357-L1375) uses a phase based approach where the all the Modules loaded before moving on to the next phase, followed by all the modules reflecting there component descriptors, followed by all the modules adding their required system components.

The [StaticModule](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp#L1243-L1247) loading workflow does not initialize ALL modules for a specific phase at a time and I believe that is the cause of this issue.

Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp
```diff
@@ -1214,12 +1214,15 @@ namespace AZ
     //=========================================================================
     void ComponentApplication::CreateStaticModules(AZStd::vector<AZ::Module*>& outModules)
     {
+        // Add this to the front of the list. In monolithic builds, dynamic modules become static modules,
+        // so we want this to load before the dynamic modules load regardless of the build type.
+        outModules.insert(outModules.begin(), aznew AzCoreModule());
```
lumberyard-employee-dm, line 1219: In addition to this change I recommend changing the static module initialization code in `ComponentApplication::LoadStaticModules`, to perform initialization of ALL modules in particular phase before moving on to next phase.

```c++
    void ComponentApplication::LoadStaticModules()
    {
        // Load static modules
        auto LoadStaticModules = [this](ModuleManagerRequests* moduleManager)
        {
            auto PopulateStaticModules = [this](AZStd::vector<AZ::Module*>& outModules) { this->CreateStaticModules(outModules); };

            // 1. Load - the first call is to ensure all dynamic modules are loaded
            // 2. CreateClass - the second call is to create specific AZ::Module class instances for each dynamic module after its loaded
            // 3. RegisterComponentDescriptors - the third call is to perform a horizontal register step for each module component descriptors after
            //    module has been loaded and created
            for (ModuleInitializationSteps lastStepToPerform : { ModuleInitializationSteps::Load,
                ModuleInitializationSteps::CreateClass, ModuleInitializationSteps::RegisterComponentDescriptors })
            {
                moduleManager->LoadStaticModules(PopulateStaticModules, lastStepToPerform);
            };
        };

        ModuleManagerRequestBus::Broadcast(LoadStaticModules);
    }

```

mbalfour-amzn:
> In theory, the order of module initialization shouldn't break other modules, but it appears the Multplayer Gem is causing an issue.
> 
> The way we initialize modules [DynamicModules](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp#L1357-L1375) uses a phase based approach where the all the Modules loaded before moving on to the next phase, followed by all the modules reflecting there component descriptors, followed by all the modules adding their required system components.
> 
> The [StaticModule](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp#L1243-L1247) loading workflow does not initialize ALL modules for a specific phase at a time and I believe that is the cause of this issue.

The reason that module initialization order matters is that there's no code anywhere that sorts the system entity / component creation based on dependencies. They simply get created based on the order they're declared, for both static and dynamic modules. The issue with the Multiplayer System Component is the following:

```
    void MultiplayerSystemComponent::GetRequiredServices(AZ::ComponentDescriptor::DependencyArrayType& required)
    {
        required.push_back(AZ_CRC_CE("NetworkingService"));
        required.push_back(AZ_CRC_CE("MultiplayerStatSystemComponent"));
    }
```

It requires a component that meets "NetworkingService" to exist before it gets started. And sure enough, it goes on to call `AZ::Interface<INetworking>::Get()->CreateNetworkInterface` as a part of initialization, which fails if nothing is registered to the INetworking interface. Normally, this "just works" because AzNetworking gets started as a static module before any dynamic modules are started. However, if all the modules are static, AzNetworking was getting started last, not first, so this requirement wasn't met and things crashed.

The larger flaw here is that it appears that the required services for system components don't cause any reordering of initialization to ensure that components are started in the correct order across modules, or at the very least the dependencies don't generate any visible errors when they're unmet.

lumberyard-employee-dm:
> > In theory, the order of module initialization shouldn't break other modules, but it appears the Multplayer Gem is causing an issue.
> > The way we initialize modules [DynamicModules](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp#L1357-L1375) uses a phase based approach where the all the Modules loaded before moving on to the next phase, followed by all the modules reflecting there component descriptors, followed by all the modules adding their required system components.
> > The [StaticModule](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp#L1243-L1247) loading workflow does not initialize ALL modules for a specific phase at a time and I believe that is the cause of this issue.
> 
> The reason that module initialization order matters is that there's no code anywhere that sorts the system entity / component creation based on dependencies. They simply get created based on the order they're declared, for both static and dynamic modules. The issue with the Multiplayer System Component is the following:
> 
> ```
>     void MultiplayerSystemComponent::GetRequiredServices(AZ::ComponentDescriptor::DependencyArrayType& required)
>     {
>         required.push_back(AZ_CRC_CE("NetworkingService"));
>         required.push_back(AZ_CRC_CE("MultiplayerStatSystemComponent"));
>     }
> ```
> 
> It requires a component that meets "NetworkingService" to exist before it gets started. And sure enough, it goes on to call `AZ::Interface<INetworking>::Get()->CreateNetworkInterface` as a part of initialization, which fails if nothing is registered to the INetworking interface. Normally, this "just works" because AzNetworking gets started as a static module before any dynamic modules are started. However, if all the modules are static, AzNetworking was getting started last, not first, so this requirement wasn't met and things crashed.
> 
> The larger flaw here is that it appears that the required services for system components don't cause any reordering of initialization to ensure that components are started in the correct order across modules, or at the very least the dependencies don't generate any visible errors when they're unmet.

We DO sort the System Component, when the [System Entity Activates](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Module/ModuleManager.cpp#L710-L751)

mbalfour-amzn:
> We DO sort the System Components, when the [System Entity Activates](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Module/ModuleManager.cpp#L710-L751)

Ohhh. I'll close out this PR and not submit it. The real problem here is that the MultiplayerSystemComponent is relying on the NetworkingSystemComponent from its constructor, not from Activate.

mbalfour-amzn:
I'll open a new PR that defers the Multiplayer initialization that uses the Networking system till the Activate().

lumberyard-employee-dm:
> > We DO sort the System Components, when the [System Entity Activates](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Module/ModuleManager.cpp#L710-L751)
> 
> Ohhh. I'll close out this PR and not submit it. The real problem here is that the MultiplayerSystemComponent is relying on the NetworkingSystemComponent from its constructor, not from Activate.

Ahh that is problem.
The Components can be **created** in any order, but their **activation** follows the service dependency order

PR #14348, opened 2023-02-02
Adding internal pipeline metrics uploading

evanchia-ly-sdets:

## What does this PR do?
Adds metrics uploading to the internal pipeline

## How was this PR tested?
Tested scripts locally, but will test in nightly pipeline.

### DIFF
--- a/scripts/build/Platform/Windows/build_config.json
+++ b/scripts/build/Platform/Windows/build_config.json
@@ -1,4 +1,32 @@
 {
+  "public_metrics_pipe": {
+    "TAGS": [
+      "periodic-incremental-daily-internal"
+    ],
+    "steps": [
+      "generate_metrics",
+      "upload_metrics"
+    ]
+  },
+  "generate_metrics": {
+    "TAGS": [
+      "periodic-incremental-daily-internal"
+    ],
+    "COMMAND": "python_windows.cmd",
+    "PARAMETERS": {
+      "SCRIPT_PATH": "scripts\\metrics\\ctest_metrics_xml_to_csv.py",
+      "SCRIPT_PARAMETERS": "build\\windows --output-directory build\\windows\\Testing\\csv --branch %BRANCH_NAME% -w"
+    }
+  },
+  "upload_metrics": {
+    "TAGS": [
+      "periodic-incremental-daily-internal"
+    ],
+    "COMMAND": "upload_metrics_windows.cmd",
+    "PARAMETERS": {
+      "CSV_DIRECTORY": "build\\windows\\Testing\\csv"
+    }
+  },
   "clean": {
     "TAGS": [],
     "COMMAND": "clean_windows.cmd",
--- a/scripts/build/Platform/Windows/upload_metrics_windows.cmd
+++ b/scripts/build/Platform/Windows/upload_metrics_windows.cmd
@@ -1,0 +1,14 @@
+@ECHO OFF
+REM
+REM Copyright (c) Contributors to the Open 3D Engine Project.
+REM For complete copyright and license terms please see the LICENSE at the root of this distribution.
+REM
+REM SPDX-License-Identifier: Apache-2.0 OR MIT
+REM
+REM
+
+ECHO [ci_build] aws s3 cp --recursive %CSV_DIRECTORY% s3://o3de-metrics/csv/ --acl bucket-owner-full-control
+
+aws s3 cp --recursive %CSV_DIRECTORY% s3://o3de-metrics/csv/ --acl bucket-owner-full-control
+
+EXIT /b %ERRORLEVEL%
--- a/scripts/metrics/ctest_metrics_xml_to_csv.py
+++ b/scripts/metrics/ctest_metrics_xml_to_csv.py
@@ -43,17 +43,26 @@ def main():
     args = parse_args()
 
     # Construct the full path to the xml file
-    file_path = _get_test_xml_path(args.build_folder, args.ctest_log)
+    xml_file_path = _get_test_xml_path(args.build_folder, args.ctest_log)
 
     # Create csv file
-    if os.path.exists(args.csv_file):
-        logger.warning(f"The file {args.csv_file} already exists. It will be overriden.")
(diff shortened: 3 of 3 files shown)


scripts/build/Platform/Windows/build_config.json
```diff
@@ -1,4 +1,32 @@
...
+      "periodic-incremental-daily-internal"
+    ],
+    "steps": [
+      "generate_metrics",
+      "upload_metrics"
+    ]
+  },
+  "generate_metrics": {
+    "TAGS": [
+      "periodic-incremental-daily-internal"
```
brianherrera, line 13: You only need to add the tag to the pipe config if they are meant to run together, otherwise they will run separately.

scripts/build/Platform/Windows/upload_metrics_windows.cmd
```diff
@@ -0,0 +1,14 @@
...
+REM Copyright (c) Contributors to the Open 3D Engine Project.
+REM For complete copyright and license terms please see the LICENSE at the root of this distribution.
+REM
+REM SPDX-License-Identifier: Apache-2.0 OR MIT
+REM
+REM
+
+ECHO [ci_build] aws s3 cp --recursive %CSV_DIRECTORY% s3://o3de-metrics/csv/ --acl bucket-owner-full-control
+
+aws s3 cp --recursive %CSV_DIRECTORY% s3://o3de-metrics/csv/ --acl bucket-owner-full-control
```
brianherrera, line 12: I recommend having the bucket name provided to the script as an argument instead of hardcoding it here. In Jenkins we can then set it as an env var to supply to the script.

scripts/build/Platform/Windows/build_config.json
```diff
@@ -1,4 +1,43 @@
...
+      "SCRIPT_PATH": "scripts\\metrics\\ctest_metrics_xml_to_csv.py",
+      "SCRIPT_PARAMETERS": "build\\windows --output-directory build\\windows\\Testing\\csv --branch %BRANCH_NAME% -w"
+    }
+  },
+  "upload_metrics": {
+    "TAGS": [],
+    "COMMAND": "upload_metrics_windows.cmd",
+    "PARAMETERS": {
+      "CSV_DIRECTORY": "build\\windows\\Testing\\csv",
+      "BUCKET": "s3://o3de-metrics/csv/"
```
amzn-changml, line 25: Probably might be better to make this a Jenkins expandable variable like `!O3DE_METRICS_S3_BUCKET!`, as this will not be accessible in other Jenkins environments

scripts/build/Platform/Windows/upload_metrics_windows.cmd
```diff
@@ -0,0 +1,14 @@
+@ECHO OFF
+REM
+REM Copyright (c) Contributors to the Open 3D Engine Project.
+REM For complete copyright and license terms please see the LICENSE at the root of this distribution.
+REM
+REM SPDX-License-Identifier: Apache-2.0 OR MIT
+REM
+REM
+
+ECHO [ci_build] aws s3 cp --recursive %CSV_DIRECTORY% %BUCKET% --acl bucket-owner-full-control
```
Kadino, line 10: @o3de/sig-build is [ci_build] an appropriate tag to log here?

scripts/build/Platform/Windows/build_config.json
```diff
@@ -1,4 +1,46 @@
 {
+  "public_metrics_pipe": {
```
Kadino, line 2: I'm a bit confused on how this is defining when these metrics are collected. Is this running its own forked periodic pipeline? Shouldn't metrics get uploaded as part of existing pipelines?

amzn-changml: The `periodic-incremental-daily-internal` tag will mean that this will run in any internal Jenkins that is not the O3DE one. I did question whether this is necessary or just a sandboxing measure while this is in active development

scripts/build/Platform/Windows/upload_metrics_windows.cmd
amzn-changml: This is probably not something we need to emit, as the command below will be visible as well. Is this possibly a debug message?

scripts/build/Platform/Windows/build_config.json
evanchia-ly-sdets: These metrics are intended to be ran only on the internal Jenkins because these metrics are not yet public.

They are running in a separate "pipe" in the current Periodic Incremental Daily Jenkins job because all current build steps are "blocking" which means if a build/test step fails, it will not trigger the metrics step.

scripts/build/Platform/Windows/upload_metrics_windows.cmd
evanchia-ly-sdets: Removed debug line

scripts/metrics/ctest_metrics_xml_to_csv.py
```diff
@@ -70,8 +79,21 @@ def parse_args():
     )
     parser.add_argument(
         "--csv-file", action="store", default=_get_default_csv_filename(),
-        help=f"The directory and file name for the csv to be saved (defaults to YYYY_MM_DD)."
+        help=f"The directory and file name for the csv to be saved (defaults to YYYY_MM_DD_HH_mm)."
```
Kadino, line 82: Would be nice to have a strict contract of how folders are structured in S3, to share between upload and download/querying. If it doesn't get added in code, be sure to document it after iteration is done.

Kadino (approved):
Looks like this should upload metrics without interfering with the existing pipeline. I won't hold this up, as I expect to see plenty more PRs that modify this.

scripts/metrics/ctest_metrics_xml_to_csv.py
evanchia-ly-sdets: Folder structure contract should be held by unit tests. Unsure if the download/querying has tests, but will cut a task if not.

brianherrera (approved):

PR #14349, opened 2023-02-02
Fix crash in dropping fbx as new-prefab override

forkercat:

## What does this PR do?

This PR fixes a crash when we drag an FBX file from Asset Browser to a non-focused prefab in Outliner. Since we already disabled instantiating prefab as override earlier in https://github.com/o3de/o3de/pull/13578/, we should not allow this to happen as well.

Before the fix, it would show up `Prefab Instantiation Error` dialog and then crash in DetachPrefab workflow.

The fix is just not go into the DetachPrefab workflow if the InstantiatePrefab work won't succeed.

## How was this PR tested?

I manually tested dragging one or couple of FBX to Outliner under a non-focused prefab. The dialog showed up, but it did not crash.


AMZN-stankowi (approved):
Would it be possible to add an automated test that fails without this change, and succeeds with it?

lawsonamzn (approved):
I'm hoping once we get overrides on as default we can switc this to always instantiate with overrides.

dtamkin1 (approved):

forkercat:
Looking into adding a unit test in PrefabSaveHandler or an automated test that test 'DragAndDrop' for this issue.

forkercat:
> Looking into adding a unit test in PrefabSaveHandler or an automated test that test 'DragAndDrop' for this issue.

One possible place to add the test is in a new file `PrefabSaveLoadHandlerTests` with new test fixture. I need to create a new test fixture similar to `AssetBrowserTests` in order to create prefab assets to be dropped. I prefer skipping it for now and consider adding the tests later after refactoring the `PrefabSaveLoadHandler` class.

Issue #12339, opened 2022-09-30
Frequent use of GetFeatureProcessor function causes performance issues


amzn-tommy:
**Describe the bug**
GetFeatureProcessor is not a free lookup function due to its use of RTTI, and generally speaking pointers to feature processors should be cached and re-used. This has a noticeable performance impact when the transform of many mesh components are updated every frame, since UpdateObjectSrg is calling GetFeatureProcessor many times per frame.

Several Pass::IsEnabled() functions also call GetFeatureProcessor every frame.

**Steps to reproduce**
1. Create a scene with a large number of mesh components that move every frame.

**Screenshots/Video**
![GetFeatureProcessor](https://user-images.githubusercontent.com/82672795/193346331-120e7e1a-dcd1-423b-b198-19ffc0967bf7.PNG)

burelc-amzn:
Does #14421 alleviate any of this issue?

amzn-tommy:
This issue was actually fixed at a higher level by https://github.com/o3de/o3de/pull/13686 which updates it to not call GetFeatureProcessor so frequently, but instead call it once and cache the result.

So I'm not sure if #14421 helped, or if this was a case where not enough information is known at compile time for it to be evaluated at compile time.

PR #14350, opened 2023-02-02
Optimized scene queries by replacing RTTI with request type

SergeyAMZN:

## What does this PR do?

Improves performance of scene queries in PhysX gem by replacing expensive AZ_RTTI mechanism with storing request type in the base class. 

## How was this PR tested?

Unit tests

image

### DIFF
--- a/Code/Framework/AzFramework/AzFramework/Physics/Common/PhysicsSceneQueries.h
+++ b/Code/Framework/AzFramework/AzFramework/Physics/Common/PhysicsSceneQueries.h
@@ -151,7 +151,7 @@ namespace AzPhysics
         static void Reflect(AZ::ReflectContext* context);
         virtual ~SceneQueryRequest() = default;
 
-        RequestType m_requestType = Undefined;
+        RequestType m_requestType = RequestType::Undefined;
         AZ::u32 m_maxResults = 32; //!< The Maximum results for this request to return, this is limited by the value set in the SceneConfiguration
         CollisionGroup m_collisionGroup = CollisionGroup::All; //!< Collision filter for the query.
         SceneQuery::QueryType m_queryType = SceneQuery::QueryType::StaticAndDynamic; //!< Object types to include in the query
--- a/Code/Framework/AzFramework/AzFramework/Physics/Common/PhysicsSceneQueries.h
+++ b/Code/Framework/AzFramework/AzFramework/Physics/Common/PhysicsSceneQueries.h
@@ -132,12 +132,27 @@ namespace AzPhysics
     //! Not valid to be used with Scene::QueryScene functions
     struct SceneQueryRequest
     {
+        enum RequestType : AZ::u8
+        {
+            Undefined = 0,
+            Raycast,
+            Shapecast,
+            Overlap
+        };
+
         AZ_CLASS_ALLOCATOR_DECL;
         AZ_RTTI(SceneQueryRequest, "{76ECAB7D-42BA-461F-82E6-DCED8E1BDCB9}");
+
+        explicit SceneQueryRequest(RequestType requestType)
+            : m_requestType(requestType)
+        {
+        }
+        SceneQueryRequest() = default;
         static void Reflect(AZ::ReflectContext* context);
         virtual ~SceneQueryRequest() = default;
 
-        AZ::u64 m_maxResults = 32; //!< The Maximum results for this request to return, this is limited by the value set in the SceneConfiguration
+        RequestType m_requestType = Undefined;
+        AZ::u32 m_maxResults = 32; //!< The Maximum results for this request to return, this is limited by the value set in the SceneConfiguration
         CollisionGroup m_collisionGroup = CollisionGroup::All; //!< Collision filter for the query.
         SceneQuery::QueryType m_queryType = SceneQuery::QueryType::StaticAndDynamic; //!< Object types to include in the query
     };
@@ -149,6 +164,12 @@ namespace AzPhysics
     {
         AZ_CLASS_ALLOCATOR_DECL;
         AZ_RTTI(RayCastRequest, "{53EAD088-A391-48F1-8370-2A1DBA31512F}", SceneQueryRequest);
+
+        RayCastRequest()
+            : SceneQueryRequest(RequestType::Raycast)
+        {
+        }
+
         static void Reflect(AZ::ReflectContext* context);
 
         float m_distance = 500.0f; //!< The distance to cast along the direction.
@@ -165,6 +186,11 @@ namespace AzPhysics
     {
         AZ_CLASS_ALLOCATOR_DECL;
         AZ_RTTI(ShapeCastRequest, "{52F6C536-92F6-4C05-983D-0A74800AE56D}", SceneQueryRequest);
+
+        ShapeCastRequest()
+            : SceneQueryRequest(RequestType::Shapecast)
+        {
+        }
         static void Reflect(AZ::ReflectContext* context);
 
         float m_distance = 500.0f; //! The distance to cast along the direction.
@@ -207,6 +233,12 @@ namespace AzPhysics
     {
         AZ_CLASS_ALLOCATOR_DECL;
         AZ_RTTI(OverlapRequest, "{3DC986C2-316B-4C54-A0A6-8ABBB8ABCC4A}", SceneQueryRequest);
+
+        OverlapRequest()
+            : SceneQueryRequest(RequestType::Overlap)
+        {
+        }
+
         static void Reflect(AZ::ReflectContext* context);
 
         AZ::Transform m_pose = AZ::Transform::CreateIdentity(); //!< Initial shape pose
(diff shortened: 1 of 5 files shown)


hultonha:
FYI @adamdbrw, @pawelbudziszewski, @michalpelka - you might be interested in this 🙂

Code/Framework/AzFramework/AzFramework/Physics/Common/PhysicsSceneQueries.h
```diff
@@ -132,12 +132,27 @@ namespace AzPhysics
     //! Not valid to be used with Scene::QueryScene functions
     struct SceneQueryRequest
     {
+        enum RequestType : AZ::u8
```
hultonha, line 135: suggestion: Could we make this `enum class`?

Gems/PhysX/Code/Source/Scene/PhysXScene.cpp
```diff
@@ -974,27 +974,29 @@ namespace PhysX
...
+            }
+        case AzPhysics::SceneQueryRequest::Shapecast:
+            {
+                return Internal::ShapeCast(
+                    static_cast<const AzPhysics::ShapeCastRequest*>(request), s_sweepBuffer, m_pxScene, queryData, m_shapecastBufferSize);
+            }
+        case AzPhysics::SceneQueryRequest::Overlap:
+            {
+                return Internal::OverlapQuery(
+                    static_cast<const AzPhysics::OverlapRequest*>(request), s_overlapBuffer, m_pxScene, queryData, m_overlapBufferSize);
```
hultonha, line 992: question: Do we definitely need these casts? Will things not compile without them?

hultonha (approved):
The changes look good to me I just have a couple of minor suggestions/questions. Thanks!

moraaar (approved):

Gems/PhysX/Code/Source/Scene/PhysXScene.cpp
```diff
@@ -974,27 +974,29 @@ namespace PhysX
...
         {
-            return Internal::OverlapQuery(azdynamic_cast<const AzPhysics::OverlapRequest*>(request),
-                s_overlapBuffer, m_pxScene, queryData, m_overlapBufferSize);
-        }
-        else
-        {
-            AZ_Warning("Physx", false, "Unknown Scene Query request type.");
-        }
-        return AzPhysics::SceneQueryHits();
+        case AzPhysics::SceneQueryRequest::Raycast:
```
burelc-amzn, line 979: I'm surprised that this change is significantly faster. The code
```cpp
azrtti_istypeof<AzPhysics::RayCastRequest>(request)
```
is a macro that expands to
```cpp
AZ::RttiIsTypeOf<AzPhysics::RayCastRequest>(request))
```

Substituting in our types in this case, that template function is calling:
```cpp
AZ::Internal::RttiIsTypeOfHelper<AzPhysics::RayCastRequest, AzPhysics::SceneQueryRequest*>::Check(request, typename AZ::HasAZRtti<AzPhysics::SceneQueryRequest>::kind_type())
```

_That_ function is the same as
```cpp
request && request->RTTI_IsTypeOf(AZ::AzTypeInfo<AzPhysics::RayCastRequest>::Uuid()))
```

Now, that `RTTI_IsTypeOf` function is virtual. For the two classes in question, `RayCastRequest` and `SceneQueryRequest`, they have a `IsTypeOf` overload as:
```cpp
virtual inline bool RTTI_IsTypeOf(const AZ ::TypeId& typeId) const
{
    return RTTI_IsContainType(typeId);
}
```

Continuing down the rabbit hole, we get to the actual test:
```cpp
static constexpr bool SceneQueryRequest::RTTI_IsContainType(const AZ ::TypeId& id)
{
    return id == RTTI_Type();
}

static bool RayCastRequest::RTTI_IsContainType(const AZ ::TypeId& id)
{
    if (id == RTTI_Type())
    {
        return true;
    }
    return AZ ::Internal ::RttiCaller<SceneQueryRequest>::RTTI_IsContainType(id);
}
```

Those `id == RTTI_Type()` calls is doing the real comparison:
```cpp
constexpr bool Uuid::operator==(const Uuid& rhs) const
{
    return AZStd::ranges::equal(m_data, rhs.m_data);
}
```
which compares the `AZStd::byte[16]`'s of each Uuid.

From what I can tell, there's 1 virtual function call, and 2 `AZStd::byte[16]` comparisons for each call to `azrtti_istypeof`. So where's the speed going?

lumberyard-employee-dm: Instead of performing an `azrtti_istypeof` followed by an `azdynamic_cast`, it is quicker to just perform the `azdynamic_cast`
The `azrtti_istypeof` check is unnecessary when a cast is going to be performed as it is checking the same information twice.

SergeyAMZN: @burelc-amzn it's because of `AZ::Uuid::CreateStringSkipWarning` 
Similar issue to this: https://github.com/o3de/o3de/issues/12339
Up to 20% of the entire raycast call was RTTI.. 

I'll try test with `azdynamic_cast`, thanks @lumberyard-employee-dm

amzn-tommy: Avoiding a redundant azrtti_istypeof is better and can be done anyways, but I expect you'll still see that azdynamic_cast is still expensive due to the RTTI functions. Ultimately, getting to the bottom of why we end up creating a Uuid from the string from the RTTI macro at runtime instead of having the Uuid itself created at compile time would be the biggest win across all systems.

lumberyard-employee-dm: Actually it has to be created at runtime because RTTI_GetType() is a virtual function and therefore relies on a vtable created at runtime.

Gems/PhysX/Code/Source/Scene/PhysXScene.cpp
SergeyAMZN: Since we're casting from base class pointer to a derived one, have to have the casts..

Gems/PhysX/Code/Source/Scene/PhysXScene.cpp
SergeyAMZN: Yes, still getting `CreateStringSkipWarnings` with `azdynamic_cast` too. 

```
AZ::Uuid::CreateStringSkipWarnings(AZStd::basic_string_view<char,AZStd::char_traits<char>> uuidString, bool skipWarnings) Line 94	C++
AZ::Uuid::CreateString(AZStd::basic_string_view<char,AZStd::char_traits<char>> uuidString) Line 84	C++
AZ::Uuid::Uuid(AZStd::basic_string_view<char,AZStd::char_traits<char>> uuidString) Line 60	C++
AzPhysics::RayCastRequest::TYPEINFO_Uuid() Line 166	C++
AzPhysics::RayCastRequest::RTTI_Type() Line 166	C++
AzPhysics::RayCastRequest::RTTI_AddressOf(const AZ::Uuid & id) Line 166	C++
AZ::Internal::RttiCastHelper<AzPhysics::RayCastRequest const *,AzPhysics::SceneQueryRequest const *>(const AzPhysics::SceneQueryRequest * ptr, const std::integral_constant<enum AZ::RttiKind,1> & __formal) Line 614	C++
AZ::RttiCast<AzPhysics::RayCastRequest const *,AzPhysics::SceneQueryRequest const *>(const AzPhysics::SceneQueryRequest * ptr) Line 913	C++
PhysX::PhysXScene::QueryScene(const AzPhysics::SceneQueryRequest * request) Line 1003	C++

```

I tried to see if it's possible to cache the UUID in a static local variable inside TYPEINFO_Uuid() since that function will be unique per class type but it seems there are some issues with constexpr in this case. 

Here's my attempt:
```
    static constexpr AZ::TypeId TYPEINFO_Uuid() \
    { \
        static AZ::TypeId cachedUuid = AZ::TypeId::CreateNull(); \
        if (cachedUuid.IsNull()) \
            cachedUuid = AZ::TypeId(_ClassUuid) + AZ::Internal::AggregateTypes<__VA_ARGS__>::Uuid(); \
        return cachedUuid; \
    }

error C3615: constexpr function 'AZ::AttributeData<AZ::Script::Attributes::ExcludeFlags>::TYPEINFO_Uuid' cannot result in a constant expression
```

lumberyard-employee-dm: I working on the AZ TypeInfo code right now.
This should be addressed soon.

FYI, a static local cannot be used inside of a constexpr function.
It makes the function unable to be evaluated at compile time.

burelc-amzn: It should avoid the runtime calls to parse Uuid strings if you use a static constexpr variable instead, like in d59c241ec8945f0420bab6ccca48ce27529776e6. But it sounds like there’s a larger change coming.

Gems/PhysX/Code/Source/Scene/PhysXScene.cpp
hultonha: Ah I see now, got you thanks

hultonha (approved):

PR #5771, opened 2021-11-19
CMake Xcode Generator fix

lumberyard-employee-dm:
Removed the generator expression from the `LY_ARCHIVE_FILE_SEARCH_MODE` define as Xcode doesn't support per-config per-file definitions.

Updated the #else block in ArchiveVars to default to PakOnly mode if the `LY_ARCHIVE_FILE_SEARCH_MODE` define isn't set.

Moved the the `LY_ARCHIVE_FILE_SEARCH_MODE` Cache Variable from cmake/Deployment.cmake to feature_options.cmake within the AzFramework folder.

Signed-off-by: lumberyard-employee-dm <56135373+lumberyard-employee-dm@users.noreply.github.com>

PR #14353, opened 2023-02-02
Mbalfour/networking init cleanups

mbalfour-amzn:
## What does this PR do?

Fixes a couple of different init/shutdown errors in the networking code:
1. The MultiplayerEditorConnection was getting constructed as a part of the MultiplayerSystemComponent constructor. However, it relied on AzNetworking being active at that point or else it would crash. AzNetworking is only guaranteed to be active for the Activate() call, not during construction, so this crashed in monolithic builds. The solution is to defer the MEC construction until the Activate() call.
2. Both AWSCore and HttpRequestor were trying to init and shutdown the AwsNativeSDK. This caused a crash on shutdown because some AwsNativeSDK-allocated pointers could no longer get freed after the first shutdown. The solution is to add ownership guards to both gems so that only the first gem to activate owns the init/shutdown sequence, and the other one doesn't. Since neither gem appears to depend on the other currently, I didn't want to introduce any cross-dependencies between the two gems just to solve this issue. However, the implication is that if any future gems get added that also try to own AwsNativeSDK startup/shutdown, they will need to add the same ownership guard too.

## How was this PR tested?

Ran a monolithic build and verified that it no longer crashes on shutdown due to bad AwsNativeSDK shutdown order. (It still crashes on shutdown in ScriptCanvasRegistry::ReleaseDescriptors(), but that's well after the AWS module shutdowns and doesn't appear to be related)

### DIFF
--- a/Gems/AWSCore/Code/Source/AWSCoreSystemComponent.cpp
+++ b/Gems/AWSCore/Code/Source/AWSCoreSystemComponent.cpp
@@ -137,12 +137,21 @@ namespace AWSCore
 
     void AWSCoreSystemComponent::InitAWSApi()
     {
-        AWSNativeSDKInit::InitializationManager::InitAwsApi();
+        // Multiple different Gems might try to use the AWSNativeSDK, so make sure it only gets initialized / shutdown once
+        // by the first Gem to try using it.
+        m_ownsAwsNativeInitialization = !AWSNativeSDKInit::InitializationManager::IsInitialized();
+        if (m_ownsAwsNativeInitialization)
+        {
+            AWSNativeSDKInit::InitializationManager::InitAwsApi();
+        }
     }
 
     void AWSCoreSystemComponent::ShutdownAWSApi()
     {
-        AWSNativeSDKInit::InitializationManager::Shutdown();
+        if (m_ownsAwsNativeInitialization)
+        {
+            AWSNativeSDKInit::InitializationManager::Shutdown();
+        }
     }
 
     AZ::JobContext* AWSCoreSystemComponent::GetDefaultJobContext()
--- a/Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
+++ b/Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
@@ -263,7 +263,7 @@ namespace Multiplayer
 
     void MultiplayerSystemComponent::Activate()
     {
-#if !defined(AZ_RELEASE_BUILD)
+#if (O3DE_EDITOR_CONNECTION_LISTENER_ENABLE)
         m_editorConnectionListener = AZStd::make_unique<MultiplayerEditorConnection>();
 #endif
 
@@ -351,7 +351,7 @@ namespace Multiplayer
 
         m_networkEntityManager.Reset();
 
-#if !defined(AZ_RELEASE_BUILD)
+#if (O3DE_EDITOR_CONNECTION_LISTENER_ENABLE)
         m_editorConnectionListener.reset();
 #endif
     }
--- a/Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
+++ b/Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
@@ -263,7 +263,7 @@ namespace Multiplayer
 
     void MultiplayerSystemComponent::Activate()
     {
-#if (O3DE_EDITOR_CONNECTION_LISTENER_ENABLE)
+#if !defined(AZ_RELEASE_BUILD)
         m_editorConnectionListener = AZStd::make_unique<MultiplayerEditorConnection>();
 #endif
 
@@ -351,7 +351,7 @@ namespace Multiplayer
 
         m_networkEntityManager.Reset();
 
-#if (O3DE_EDITOR_CONNECTION_LISTENER_ENABLE)
+#if !defined(AZ_RELEASE_BUILD)
         m_editorConnectionListener.reset();
 #endif
     }
--- a/Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
+++ b/Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
@@ -263,6 +263,10 @@ namespace Multiplayer
 
     void MultiplayerSystemComponent::Activate()
     {
+#if !defined(AZ_RELEASE_BUILD)
+        m_editorConnectionListener = AZStd::make_unique<MultiplayerEditorConnection>();
+#endif
+
         RegisterMetrics();
 
         AzFramework::RootSpawnableNotificationBus::Handler::BusConnect();
@@ -346,6 +350,10 @@ namespace Multiplayer
         AzFramework::RootSpawnableNotificationBus::Handler::BusDisconnect();
 
         m_networkEntityManager.Reset();
+
+#if !defined(AZ_RELEASE_BUILD)
+        m_editorConnectionListener.reset();
+#endif
     }
 
     bool MultiplayerSystemComponent::StartHosting(uint16_t port, bool isDedicated)
--- a/Gems/AWSCore/Code/Source/AWSCoreSystemComponent.h
+++ b/Gems/AWSCore/Code/Source/AWSCoreSystemComponent.h
@@ -77,6 +77,8 @@ namespace AWSCore
         AZStd::unique_ptr<AWSCoreConfiguration> m_awsCoreConfiguration;
         AZStd::unique_ptr<AWSCredentialManager> m_awsCredentialManager;
         AZStd::unique_ptr<AWSResourceMappingManager> m_awsResourceMappingManager;
+
+        bool m_ownsAwsNativeInitialization; // Track whether or not this module initialized the native layer
     };
 
 } // namespace AWSCore
--- a/Gems/AWSCore/Code/Source/AWSCoreSystemComponent.h
+++ b/Gems/AWSCore/Code/Source/AWSCoreSystemComponent.h
@@ -78,7 +78,7 @@ namespace AWSCore
         AZStd::unique_ptr<AWSCredentialManager> m_awsCredentialManager;
         AZStd::unique_ptr<AWSResourceMappingManager> m_awsResourceMappingManager;
 
-        bool m_ownsAwsNativeInitialization; // Track whether or not this module initialized the native layer
+        bool m_ownsAwsNativeInitialization = false; // Track whether or not this module initialized the native layer
     };
 
 } // namespace AWSCore
--- a/Gems/HttpRequestor/Code/Source/HttpRequestManager.cpp
+++ b/Gems/HttpRequestor/Code/Source/HttpRequestManager.cpp
@@ -35,7 +35,14 @@ namespace HttpRequestor
         desc.m_name = s_loggingName;
         desc.m_cpuId = AFFINITY_MASK_USERTHREADS;
         m_runThread = true;
-        AWSNativeSDKInit::InitializationManager::InitAwsApi();
+
+        // Multiple different Gems might try to use the AWSNativeSDK, so make sure it only gets initialized / shutdown once
+        // by the first Gem to try using it.
+        m_ownsAwsNativeInitialization = !AWSNativeSDKInit::InitializationManager::IsInitialized();
+        if (m_ownsAwsNativeInitialization)
+        {
+            AWSNativeSDKInit::InitializationManager::InitAwsApi();
+        }
         auto function = [this]
         {
             ThreadFunction();
@@ -53,7 +60,10 @@ namespace HttpRequestor
         }
 
         // Shutdown after background thread has closed.
-        AWSNativeSDKInit::InitializationManager::Shutdown();
+        if (m_ownsAwsNativeInitialization)
+        {
+            AWSNativeSDKInit::InitializationManager::Shutdown();
+        }
     }
 
     void Manager::AddRequest(Parameters&& httpRequestParameters)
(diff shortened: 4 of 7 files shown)


Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
```diff
@@ -263,6 +263,10 @@ namespace Multiplayer
 
     void MultiplayerSystemComponent::Activate()
     {
+#if !defined(AZ_RELEASE_BUILD)
+        m_editorConnectionListener = AZStd::make_unique<MultiplayerEditorConnection>();
```
AMZN-Olex, line 267: FYI I found that this ended up breaking in non-release monolithic, such as profile monolithic as I found Editor features weren't available but this class was expected it. That said I'm not sure if we have a monolithic macro on or not. We might have to add that in cmake somewhere.

AMZN-Olex (commented):
Overall this looks good but I'm unsure about AWSNativeSDKInit::InitializationManager::IsInitialized and would rather Allie take a look at this.

allisaurus (commented):
One question about initialization, but otherwise LGTM! 
The other AWS gems should depend on `AWSCore` to initialize the SDK (and `HttpRequestor` should honestly be migrated away from using it), so I believe the future scope of needed ownership guards will be small.

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
mbalfour-amzn: I was able to run profile monolithic with this change, so I don't think there's a need to guard this any further.

allisaurus (approved):

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
lumberyard-employee-dm: Instead of using not defined `AZ_RELEASE_BUILD` monolithic check, it would be more configurable to add an O3DE_EDITOR_CONNECTION_LISTENER_ENABLE define that default enables if it is not defined and the configuration is release,
```c++
#if !defined (O3DE_EDITOR_CONNECTION_LISTENER_ENABLE)
    #if defined(AZ_RELEASE_BUILD)
        #define O3DE_EDITOR_CONNECTION_LISTENER_ENABLE 1
    #endif
#endif
```

That way users who want to debug using profile without the Editor Connection Manager could override the value.

In the `Gems/Multiplayer/Code/CMakeLists.txt` we can add a define that is set based on a CMake Cache Variable.
Such as `FORCE_ENABLE_O3DE_EDITOR_CONNECTION_LISTENER`
```cmake
set(FORCE_ENABLE_O3DE_EDITOR_CONNECTION_LISTENER "" CACHE BOOL "Overrides the creation of the Editor Connection Listener\n\
    If unset, the build configuration is determines if the Editor Connection Listener is created")

set(O3DE_EDITOR_CONNECTION_LISTENER_ENABLE $<$<BOOL:"${FORCE_ENABLE_O3DE_EDITOR_CONNECTION_LISTENER}">:O3DE_EDITOR_CONNECTION_LISTENER_ENABLE=1>)

ly_add_source_properties(
    SOURCES
        Source/MultiplayerSystemComponent.cpp
    PROPERTY COMPILE_DEFINITIONS
    VALUES ${O3DE_EDITOR_CONNECTION_LISTENER_ENABLE})
```

Gems/AWSCore/Code/Source/AWSCoreSystemComponent.cpp
```diff
@@ -137,12 +137,21 @@ namespace AWSCore
...
     void AWSCoreSystemComponent::InitAWSApi()
     {
-        AWSNativeSDKInit::InitializationManager::InitAwsApi();
+        // Multiple different Gems might try to use the AWSNativeSDK, so make sure it only gets initialized / shutdown once
+        // by the first Gem to try using it.
+        m_ownsAwsNativeInitialization = !AWSNativeSDKInit::InitializationManager::IsInitialized();
+        if (m_ownsAwsNativeInitialization)
+        {
+            AWSNativeSDKInit::InitializationManager::InitAwsApi();
+        }
```
lumberyard-employee-dm, line 146: This seems like a good candidate to store the InitializationManager inside of an `AZ::Interface`

The [AZ::Interface::Register](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Interface/Interface.h#L71-L74) function will not register an interface that is already registered
```c++
// In AWSCoreSystemComponent.h
AZStd::unique_ptr<AWSNativeSDKInit::InitializationManager> m_awsInitializationManager = AZStd::make_unique<AWSNativeSDKInit::InitializationManager>();
// ...
// In AWSCoreSystemComponent.cpp InitAWSApi
if(auto registerOutcome = AZ::Interface<AWSNativeSDKInit::InitializationManager>::Register(m_awsInitializationManager.get());
    !registerOutcome)
{
    AZ_Warning("AWS Core", "AWS Native SDK initialization has failed with error %s", registerOutcome.GetError().c_str());
}

// In AWSCoreSystemComponent.cpp ShutdownAWSApi
// Unregister will only unregister an the instance if the memory address matches
// otherwise a failure outcome will be returned.
// For this use case the failure result is needed
AZ::Interface<AWSNativeSDKInit::InitializationManager>::Unregister(m_awsInitializationManager.get());

```

The [AWSNativeSDKInit.h](https://github.com/o3de/o3de/blob/development/Code/Tools/AWSNativeSDKInit/include/AWSNativeSDKInit/AWSNativeSDKInit.h#L28-L30) InitializationManager class could just add a use of the `AZ_RTTI` macro and then it would be able to be stored in an AZ::Interface

mbalfour-amzn: In this case, I'd be inclined to leave it as-is. The InitializationManager is literally only used to call InitAwsApi() / Shutdown(), so having it available through an Interface doesn't provide any additional value. Using the Interface to prevent multiple initializations just seems to me like a slightly more obfuscated way to accomplish the same goal. I'm not _opposed_ to changing it if you feel strongly though, I only have a small preference for this one.

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
mbalfour-amzn: Followed the suggestion, but implemented it slightly differently from above:
- Changed it so the override can override in either direction (disable or enable), so it can get disabled in profile builds, enabled in release builds, etc.
- Added the compile definition to all of Multiplayer.Unified.Static, not just the .cpp file, since it's affecting the definition of the class in the header file too.

lumberyard-employee-dm: This will not actually work on MacOS when using Xcode.
The problem is that build configurations can't have different values for a C preprocessor define.
So all 3 configurations will get the same value.

lumberyard-employee-dm: We found the issue with Xcode previously in this PR https://github.com/o3de/o3de/pull/5771 when adding support for the `LY_ARCHIVE_FILE_SEARCH_MODE` define

mbalfour-amzn: Given that, I'm inclined to move it back to just using AZ_RELEASE_BUILD the way it originally was then. There isn't a specific use case right now for disabling the Editor listener in non-release builds *or* for enabling the Editor listener in release builds I don't think, so I'm not sure there's value in adding two separate build flags for force-enabling and force-disabling across all configurations.

lumberyard-employee-dm: There is only a single define being used and a single CMake Cache variable being added.

This line is `set(O3DE_EDITOR_CONNECTION_LISTENER_ENABLE $<$<BOOL:"${FORCE_ENABLE_O3DE_EDITOR_CONNECTION_LISTENER}">:O3DE_EDITOR_CONNECTION_LISTENER_ENABLE=1>)` is just setting a local variable that uses the Cache Variable if it is set.

mbalfour-amzn: Sorry, let me back up. In the original code, the goal is to only create a MultiplayerEditorConnection instance in non-release builds, because in release builds there is no Editor that's expected to exist and connect to the server.

The proposal above is adding a variable to make it possible to enable the MultiplayerEditorConnection in release builds as well. What I was trying to say is that there's a second hypothetical use case here too, which is to disable the MultiplayerEditorConnection in non-release builds. IMO, they're both equally hypothetical because neither situation has been needed yet.

So if we catered to _both_ situations, we would need to add two defines and two cache variables. One would force enabling in all builds, and one would force disabling in all builds. And in both situations, as you've shown, the code itself still needs to have a dependency on AZ_RELEASE_BUILD for the default case, since we apparently can't handle the default case of varying between release/non-release in the cmake file.

I can add the single define and cache variable if you believe it's truly necessary. It just seems like an arbitrary choice to me that wouldn't come up in practice, since the common use case is for the Editor to connect to the server that it auto-launches, not for the Editor to connect to a separate release-build server.

lumberyard-employee-dm: Let's back up even more.
The goal of only creating the MultiplayerEditorConnection in non-release builds is incorrect.
It should be not made in **non-monolithic** builds.
The Editor does exist in release builds, it does NOT exist in monolithic builds.
If anything the check should be for `AZ_MONOLITHIC_BUILD`, not `AZ_RELEASE_BUILD`.

lumberyard-employee-dm: Now to speak on the point as to why two variables Cache variables aand two defienes aren't needed, it is because the BOOL Cache variable would represent more than **two** states.

If the cache variable is not set, no `define` is added injected through the compiler definition.
If the cache variable is set, then the value of the variable is injected into the compiler definition.

If a user doesn't specify `O3DE_EDITOR_CONNECTION_LISTENER_ENABLE`, then no `#define` is set in code and then it would delegate to checking the `AZ_MONOLITHC_BUILD` define.
If the user does specify `O3DE_EDITOR_CONNECTION_LISTENER_ENABLE`, that value is used directly.

mbalfour-amzn: Sorry, please help me understand how this works.

First, I'm assuming the code side should look like this:
```
#if !defined (O3DE_EDITOR_CONNECTION_LISTENER_ENABLE)
    #if !defined(AZ_MONOLITHIC_BUILD)
        #define O3DE_EDITOR_CONNECTION_LISTENER_ENABLE 1
    #endif
#endif
```
Specifically, switching to use "!defined" for the monolithic build instead of "defined". Otherwise, won't this _only_ enable in monolithic builds if the user hasn't defined the override, which seems like the exact opposite of what we need?

Provided that's correct, for this line:
```
set(O3DE_EDITOR_CONNECTION_LISTENER_ENABLE $<$<BOOL:"${FORCE_ENABLE_O3DE_EDITOR_CONNECTION_LISTENER}">:O3DE_EDITOR_CONNECTION_LISTENER_ENABLE=1>)
```
if the FORCE value is set to true, I see how it sets ENABLE to 1, so the user can force enabling. However, if the FORCE value is false, won't this leave ENABLE undefined, causing us to fall back into the code-side conditional where it checks for monolithic build? I don't understand how this works in the case where the user specifically wants to force ENABLE to 0 (i.e. force disabling the connection).

lumberyard-employee-dm: Sorry for not explaining this well: 
The check here is to detect if the string stored in the `FORCE_ENABLE_O3DE_EDITOR_CONNECTION_LISTENER` variable is empty, not the whether the `FORCE_ENABLE_O3DE_EDITOR_CONNECTION_LISTENER` variable has a value of TRUE or FALSE

lumberyard-employee-dm: I updated the CMake Cache string and snippet to be specify the value of the cache string instead of just 1
```cmake
set(FORCE_ENABLE_O3DE_EDITOR_CONNECTION_LISTENER "" CACHE STRING "Overrides the creation of the Editor Connection Listener\n\
    If unset, the build configuration is determines if the Editor Connection Listener is created.
    If set to 0 then the Editor Connection Listener is never created
    If set to 1 then the Connection Listener is always created")

set(O3DE_EDITOR_CONNECTION_LISTENER_ENABLE $<$<BOOL:"${FORCE_ENABLE_O3DE_EDITOR_CONNECTION_LISTENER}">:O3DE_EDITOR_CONNECTION_LISTENER_ENABLE=${FORCE_ENABLE_O3DE_EDITOR_CONNECTION_LISTENER}>)
```

Gems/AWSCore/Code/Source/AWSCoreSystemComponent.cpp
lumberyard-employee-dm: I am cool with leaving this as is.

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
lumberyard-employee-dm: For reference I made this mistake in the past: https://github.com/o3de/o3de/pull/5771#discussion_r752805795

mbalfour-amzn: The bool expressions were always returning true for me, even when the string was empty. I've changed the expression to $<NOT:<$STREQUAL:... which seems to work.

I still disagree with the value of this change, but I've pushed it to the PR to hopefully resolve this concern.

lumberyard-employee-dm (approved):

PR #14381, opened 2023-02-02
Fix linker error in Android builds on Linux 

hjanetzek:
## What does this PR do?

Fixing linker error  by not linking libraries from ${LY_NDK_ABI_ROOT}/usr/lib since LY_NDK_ABI_ROOT is never set.
`'ld: error: --fix-cortex-a53-843419 is only supported on AArch64 targets' ` 

## How was this PR tested?

Android build on linux ran succesfully.

### DIFF
--- a/cmake/Platform/Android/Configurations_android.cmake
+++ b/cmake/Platform/Android/Configurations_android.cmake
@@ -56,7 +56,6 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
 
             -u ANativeActivity_onCreate
 
-            -L${LY_NDK_ABI_ROOT}/usr/lib
             -L${LY_NDK_SRC_ROOT}/cxx-stl/llvm-libc++/libs/${ANDROID_ABI}
 
         LINK_NON_STATIC_DEBUG
--- a/cmake/Platform/Android/Configurations_android.cmake
+++ b/cmake/Platform/Android/Configurations_android.cmake
@@ -56,8 +56,6 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
 
             -u ANativeActivity_onCreate
 
-            -L${LY_NDK_SRC_ROOT}/cxx-stl/llvm-libc++/libs/${ANDROID_ABI}
-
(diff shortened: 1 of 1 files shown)


cmake/Platform/Android/Configurations_android.cmake
```diff
@@ -56,7 +56,6 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
 
             -u ANativeActivity_onCreate
 
-            -L${LY_NDK_ABI_ROOT}/usr/lib
```
lumberyard-employee-dm, line 59: So the NDK on Windows has this set, but not NDK on Linux?

hjanetzek: I guess a regular Windows system does not have /usr/lib containing conflicting files

hjanetzek: On Windows this line should have no effect at all as it likely sets a non-existent library path, but it breaks the build on linux and probably macos as well

hjanetzek: https://github.com/o3de/o3de/search?q=LY_NDK_ABI_ROOT this is the single occurrence of LY_NDK_ABI_ROOT so the removed line should expand to `-L/usr/lib` on Windows as well and not to a path within NDK

lumberyard-employee-dm: It also appears that `LY_NDK_SRC_ROOT` is also never set

lumberyard-employee-dm (approved):

cmake/Platform/Android/Configurations_android.cmake
hjanetzek: sure!

gadams3 (approved):

burelc-amzn (approved):
These 2 variables used to be set by the `cmake/Platform/Android/Toolchain_android.cmake` file, but were removed quite some time ago.

cgalvan (approved):

spham-amzn (approved):

PR #14385, opened 2023-02-03
Add manual test for rga

etherealchain:

## What does this PR do?

Add a manual test for radeon gpu analyzer, the test is using test_suite periodic, so it won't be run by automated test. 
The test will copy the test data into the Assets folder, and check if the analysis data had been generated.

## How was this PR tested?

Test locally with command `>python\python.cmd -m pytest --build-directory .\build\windows\bin\profile .\AutomatedTesting\Gem\PythonTests\Atom\TestSuite_Manual.py`

Note the test uses many functions from `hydra_ShaderAssetBuilder_RecompilesShaderAsChainOfDependenciesChanges.py`, so I modulize the functions and put them in `atom_tools_utils.py`.

The chain dependency test still worked correctly when I run command `>python\python.cmd -m pytest --build-directory .\build\windows\bin\profile .\AutomatedTesting\Gem\PythonTests\Atom\TestSuite_Periodic_Null_Render_01.py`


AutomatedTesting/Gem/PythonTests/Atom/CMakeLists.txt
```diff
@@ -173,4 +173,16 @@ if(PAL_TRAIT_BUILD_HOST_TOOLS AND PAL_TRAIT_BUILD_TESTS_SUPPORTED)
         COMPONENT
             Atom
     )
+    ly_add_pytest(
+        NAME AutomatedTesting::RGATest 
+        TEST_SUITE periodic
```
galibzon, line 178: This test should be in TEST_SUITE sandbox. `periodic` tests are ran periodically in Jenkins and this test would cause it to fails because it needs the special CMAKE variable to fetch the RGA executable.

etherealchain: Thanks!

galibzon (approved):

moudgils (approved):

PR #14393, opened 2023-02-03
Action Manager | Disable "End Component Mode" action when Editor is in Entity Pick mode

AMZN-daimini:
## What does this PR do?
Adds an Enabled State CallBack to the "End Component Mode" action so that it is disabled when the Editor is in Entity Pick mode.
This allows the usage of the `Esc` shortcut in a nested fashion: when entering Entity Pick mode inside a Component Mode, pressing `Esc` a first time will just end the Entity Pick mode, while pressing it a second time will close the Component Mode.

Took the chance to verify that multiple Enabled State Callbacks on a single action work as intended in a real environment (unit tests already exist).

## How was this PR tested?
Manual testing.

EntityPickingEscBefore
Before this PR

EntityPickingEscAfter
After this PR

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeActionHandler.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeActionHandler.cpp
@@ -11,6 +11,7 @@
 #include <AzToolsFramework/ActionManager/Action/ActionManagerInterface.h>
 #include <AzToolsFramework/ActionManager/Menu/MenuManagerInterface.h>
 #include <AzToolsFramework/ActionManager/HotKey/HotKeyManagerInterface.h>
+#include <AzToolsFramework/API/ViewportEditorModeTrackerInterface.h>
 #include <AzToolsFramework/ComponentMode/EditorBaseComponentMode.h>
 #include <AzToolsFramework/ComponentMode/EditorComponentModeBus.h>
 #include <AzToolsFramework/Editor/ActionManagerIdentifiers/EditorActionUpdaterIdentifiers.h>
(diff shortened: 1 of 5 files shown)


cgalvan (approved):

amzn-ahmadkrm (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeActionHandler.cpp
```diff
@@ -100,6 +101,26 @@ namespace AzToolsFramework
                 }
             );
 
+            // Disable this action if the Entity Picker is enabled while in Component Mode.
+            m_actionManagerInterface->InstallEnabledStateCallback(
+                actionIdentifier,
```
hultonha, line 106: suggestion: It would be good to run this change by @mbalfour-amzn just to check this doesn't break anything with pick mode in the painting tool with terrain as it works slightly differently to the default entity pick mode. Should hopefully fine but would be good to verify with them. Thanks!

Code/Framework/AzToolsFramework/AzToolsFramework/Manipulators/EditorVertexSelection.cpp
```diff
@@ -83,7 +83,7 @@ namespace AzToolsFramework
         componentModeCollectionInterface->EnumerateActiveComponents(
             [&emptySelection](const AZ::EntityComponentIdPair& entityComponentIdPair, const AZ::Uuid&)
             {
-                int count;
+                int count = 0;
```
hultonha, line 86: Thanks for fixing this 🙂

Code/Framework/AzToolsFramework/AzToolsFramework/ComponentMode/ComponentModeActionHandler.cpp
AMZN-daimini: This fix was added for that exact case. Pressing `Esc` to end Entity Picking mode would also end the Paintbrush Component Mode, while this change prevents that.

hultonha: Okay perfect! So this works for Component Mode use cases and normal pick mode when in the regular viewport?

AMZN-daimini: At the moment, this only impacts the "Done" action for Component Mode. The Behavior of the Entity Picker mode `Esc` interaction is still using the old "OnEscape" Notification bus call since that system has not been migrated yet and I didn't want to expand the scope of the changes and testing.

hultonha: I see got you, thanks for the explanation 👍

PR #14408, opened 2023-02-03
Support for Calling 'host' From a Setreg File

### LINKED ISSUE #14405: Multiplayer Crash: SetReg File "Host"
Calling "host" from inside a .setreg JSON file will crash the game.

**Assets required**
[settings.multiplayersample_serverlauncher.zip](https://github.com/o3de/o3de/files/10583280/settings.multiplayersample_serverlauncher.zip)

**Steps to reproduce**
1. Create setreg or use the one provided and call "host"

_Registry/settings.multiplayersample_serverlauncher.setreg_
```
{
    "Amazon": {
        "AzCore": {
            "Runtime": {
                "ConsoleCommands": {
                    "host": "",
                }
            }
        }
    }
}
```

3. Run ServerLauncher.exe

**Expected behavior**
Host is called and the server begins hosting

**Actual behavior**
Runtime crash due to null reference. The AZ::Interface<IMultiplayer>is null.

**Found in Branch**
development

AMZN-Gene:
- Dedicated servers will automatically begin hosting on app startup unless told otherwise (sv_dedicated_host_onstartup=0)
- Updating 'host' command to register itself once it's safe to run (after application systems are all up and running)
Crash fix #14405 

## How was this PR tested?
1. Running dedicated server and seeing that it begins hosting on app start

2. Created a serverlauncher setreg with host and see that it no longer crashes and also automatically calls host once it's safe

**Registry/settings.multiplayersample_serverlauncher.setreg**
```
{
    "Amazon": {
        "AzCore": {
            "Runtime": {
                "ConsoleCommands": {
                    "host": "",
                }
            }
        }
    }
}
```


Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
```diff
@@ -98,6 +98,7 @@ namespace Multiplayer
         "The base used for blending between network updates, 0.1 will be quite linear, 0.2 or 0.3 will "
         "slow down quicker and may be better suited to connections with highly variable latency");
     AZ_CVAR(bool, bg_multiplayerDebugDraw, false, nullptr, AZ::ConsoleFunctorFlags::Null, "Enables debug draw for the multiplayer gem");
+    AZ_CVAR(bool, sv_host_onstartup, false, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Whether to host as soon as the Multiplayer system becomes available.");
```
AMZN-alexpete, line 101: I'd be down for naming this `sv_host` but I see the cl version also has a '_onstartup' prefix 😕

AMZN-alexpete (approved):

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
kberg-amzn: Wouldn't we want dedicated server targets that have the multiplayer gem enabled to just always call host? Putting this particular command in a setreg file seems redundant. I could totally see the default map being in a cfg or setreg file though.

AMZN-Olex: That seems reasonable to make it true by default for dedicated servers.

AMZN-Gene: Fixed! Dedicated servers will automatically host unless told otherwise (sv_dedicated_host_onstartup=false)

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
```diff
@@ -280,22 +281,64 @@ namespace Multiplayer
...
+                        "Opens a multiplayer connection as a host for other clients to connect to",
+                        AZ::ConsoleFunctorFlags::DontReplicate | AZ::ConsoleFunctorFlags::DontDuplicate,
+                        AZ::TypeId{},
+                        *this,
+                        &MultiplayerSystemComponent::HostConsoleCommand);
+
+                    AZ::Interface<AZ::IConsole>::Get()->ExecuteDeferredConsoleCommands();
+
+                    // Begin hosting or connecting if desired
+                    if (const auto console = AZ::Interface<AZ::IConsole>::Get())
```
lumberyard-employee-dm, line 305: The AZ::IConsole is being retrieved twice here.
The `ExecuteDeferredConsoleCommands` call can be inside of the if block

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
```diff
@@ -280,22 +281,64 @@ namespace Multiplayer
...
+                    // It's now safe to register and execute the "host" command
+                    m_hostConsoleCommand = AZStd::make_unique<AZ::ConsoleFunctor<MultiplayerSystemComponent, false>>(
+                        "host",
+                        "Opens a multiplayer connection as a host for other clients to connect to",
+                        AZ::ConsoleFunctorFlags::DontReplicate | AZ::ConsoleFunctorFlags::DontDuplicate,
+                        AZ::TypeId{},
+                        *this,
+                        &MultiplayerSystemComponent::HostConsoleCommand);
+
+                    AZ::Interface<AZ::IConsole>::Get()->ExecuteDeferredConsoleCommands();
```
lumberyard-employee-dm, line 302: You may want to add a comment, here that this call to `ExecuteDeferredConsoleCommands` would execute any previous deferred "host" commands now that it has been registered with the AZ Console.

lumberyard-employee-dm (approved):

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
AMZN-Gene: Thank you, fixed!

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
AMZN-Gene: Good call

Issue #13804, opened 2022-12-21
Crash when reparenting an entity outside its owning instance


forkercat:
**Describe the bug**

Reparenting an entity outside its owning prefab can crash.

Note: This only happens when the override flag is on, because users are not able to reparent entities outside owning prefabs when the flag is off.

**Steps to reproduce**
1. Enable the override flag `EnableOverridesUx` in 'AutomatedTesting/user/Registry/editorpreferences.setreg'
1. Create an empty level
2. Create a prefab with one entity inside
3. (Do not focus on it) Drag the entity inside out from its owning prefab

**Expected behavior**

No crash

**Actual behavior**

Crash

**Screenshots/Video**

https://user-images.githubusercontent.com/11157226/208986774-6e6d0b48-39d7-4061-8cd6-ae2214f5242d.mp4

**Found in Branch**

Development https://github.com/o3de/o3de/commit/d8470ef0ac7c1ff9f3b2106e34d866ee77ed072b

**Desktop/Device:**
* Device: PC
* OS: Windows 10
* CPU: AMD Ryzen 5950x
* GPU: NVIDIA RTX 3070
* Memory: 64 GB

**Additional context**

Reparenting on a container entity outside also crashes. We should be testing that in the fix as well.

<details>
  <summary>CLICK ME to see the call stack</summary>

```
> EditorLib.dll!AzToolsFramework::Prefab::Link::ConstructLinkDomFromPatches(rapidjson::GenericValue<rapidjson::UTF8<char>,rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator>> & linkDom, rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator> & allocator) Line 263	C++
[Inline Frame] EditorLib.dll!AZ::Debug::ProfileScope::BeginRegion(AZ::Debug::Budget *) Line 37	C++
[Inline Frame] EditorLib.dll!AZ::Debug::ProfileScope::{ctor}(AZ::Debug::Budget * budget, const char *) Line 64	C++
EditorLib.dll!AzToolsFramework::Prefab::Link::GetLinkDom(rapidjson::GenericValue<rapidjson::UTF8<char>,rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator>> & linkDom, rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator> & allocator) Line 119	C++
[...]

Issue #13823, opened 2022-12-22
Reparenting a prefab entity to the level prefab causes a crash


raktalen:
**Describe the bug**
When an entity that is a part of a prefab is re-parented to the level prefab, the editor crashes.

**Steps to reproduce**
1. Create a prefab with multiple entities
2. Drag one of the entities in this prefab to the level

**Expected behavior**
The Entity can be reparented to the level

**Actual behavior**
Editor crashes

**Screenshots/Video**
![image](https://user-images.githubusercontent.com/118857486/209209685-544083bb-1e4b-4759-8c41-482a0bf9dceb.png)

**Found in Branch**
development - with overrides enabled

AMZN-daimini:
I think this operation requires some UX thought, as it is going to be executed as an override but it may be hard to visualize that.
@rainbj

forkercat:
This is a special use case in "Reparent As Override". Users should not be allowed to add entities/prefabs directly under level prefab as overrides. It should not be a valid use case in the first place.

(Internally, the level prefab instance does not have a link object and the instance has invalid link id in it.)

A workaround is blocking this operation from happening when Override UX flag is on.

forkercat:
Reparenting to Level is disabled. Closing this issue.

PR #14411, opened 2023-02-04
Disable parenting an entity to another instance as override

forkercat:

## What does this PR do?

This PR disables moving entity/prefab to another prefab instance, including to prefab that is not focused. It just removes the focus check that was introduced in the previous PR https://github.com/o3de/o3de/pull/11659. This would make prefab override more stable and not corrupt prefab data when override feature is on.

Fixes: https://github.com/o3de/o3de/issues/13804, https://github.com/o3de/o3de/issues/13823

As discussed, the next step for this operation would be add support for copying/pasting entity/prefab into another prefab instance.

## How was this PR tested?

- [x] Manually tested in editor.
- [ ] Add new unit tests to validate the use cases

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/UI/Outliner/EntityOutlinerListModel.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/UI/Outliner/EntityOutlinerListModel.cpp
@@ -853,43 +853,36 @@ namespace AzToolsFramework
         {
             return false;
         }
-
-        // Check if we're in focus mode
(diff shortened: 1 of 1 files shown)


amzn-ahmadkrm (approved):

srikappa-amzn (approved):
The tests are great. But it'd be good to also add a few 'success' tests so that we know that reparenting works in certain scenarios. otherwise, we may run into a scenario where it always returns false and we don't catch it.

forkercat:
**UPDATE:** Added two successful reparent-entity and reparent-prefab tests with some variants.

PR #14421, opened 2023-02-06
Use a `constexpr` static variable to force evaluation of typeid strings at compile time

burelc-amzn:
This yields a 13.5x speed increase to the `TYPEINFO_Uuid()` method for classes, without adding any new symbols. It does this by forcing the calculation of a class's uuid at compile time, avoiding the expensive call to `AZ::Uuid::CreateStringSkipWarnings` at runtime.

Because the class's `TYPEINFO_Uuid_v` static variable is never ODR-used, no symbol is emitted for it into the generated binary.

The speed was measured with this benchmark:

```cpp
class Base
{
public:
    virtual ~Base() = default;
    AZ_RTTI(Base, "{4699b4b3-2a44-4a02-9f70-cd9889b4b36d}")
};

class Derived
    : public Base
{
public:
    AZ_RTTI(Derived, "{c89b7eac-26a0-4b37-a14e-cec20fcad7a8}", Base)
};

static void BenchmarkIsTypeof(benchmark::State& state)
{
    auto derived = std::make_unique<Derived>();
    Base* base = derived.get();
    for (auto _ : state)
    {
        azrtti_istypeof<Derived>(base);
    }
}
BENCHMARK(BenchmarkIsTypeof);
```

```
Benchmark                  Time             CPU   Iterations
------------------------------------------------------------
BenchmarkIsTypeof.old   64.0 ns         64.0 ns     10901595
BenchmarkIsTypeof.new   4.60 ns         4.60 ns    152030119
```

This should fix the issue that #14350 was designed to work around.

### DIFF
--- a/Code/Framework/AzCore/AzCore/RTTI/TypeInfo.h
+++ b/Code/Framework/AzCore/AzCore/RTTI/TypeInfo.h
@@ -1390,7 +1390,10 @@ namespace AZ
     } \
     static constexpr AZ::TypeId TYPEINFO_Uuid() \
     { \
-        return AZ::TypeId(_ClassUuid) + AZ::Internal::AggregateTypes<__VA_ARGS__>::Uuid(); \
+        /* using a local constexpr variable ensures that the TypeId's string is
+         * parsed at compile time */ \
+        constexpr auto typeId = AZ::TypeId(_ClassUuid) + AZ::Internal::AggregateTypes<__VA_ARGS__>::Uuid(); \
+        return typeId; \
     }
 
 // Template class type info
--- a/Code/Framework/AzCore/AzCore/RTTI/TypeInfoSimple.h
+++ b/Code/Framework/AzCore/AzCore/RTTI/TypeInfoSimple.h
@@ -20,7 +20,12 @@ namespace AZ
     struct TYPEINFO_Enable{}; \
     struct TypeNameInternal { constexpr const char* operator()() const { return #_ClassName; } }; \
     static constexpr const char* TYPEINFO_Name() { return TypeNameInternal{}(); } \
-    static constexpr AZ::TypeId TYPEINFO_Uuid() { return AZ::TypeId(_ClassUuid); }
+    static constexpr AZ::TypeId TYPEINFO_Uuid() { \
+        /* using a local constexpr variable ensures that the TypeId's string is
+         * parsed at compile time */ \
+        constexpr AZ::TypeId typeId(_ClassUuid); \
+        return typeId; \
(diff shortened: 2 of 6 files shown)


cgalvan (approved):

lumberyard-employee-dm (approved):

lsemp3d (approved):

hultonha (approved):
Nice find! 😄 Any chance we can include the benchmark too? 🙂

burelc-amzn:
> Just curious, what is the compile time impact?

@kberg-amzn I only ran one test, just using the `time` command on Linux. Both were done after running `ninja clean; ninja -t cleandead; rm -rf bin/profile;`:

| | `development` @ 139a0ac304 | `SpeedUpTypeinfoMethod` branch |
| --- | --- | --- |
| real | 64m38.631s | 53m28.877s |
| user | 912m33.662s | 929m50.223s |
| sys | 46m48.307s | 47m37.449s |

I'm not exactly sure what would account for 11 minute decrease in real time, I made an effort to run the builds while not much else was running on my machine, but maybe the `development` build had to use more swap memory. I think the 1m11s `sys` time increase is more indicative of the actual impact.

> And I see the original issue on scene queries that was linked, but how much do we actually see in runtime performance improvements?

I'd have to get @SergeyAMZN to rerun their test, I'm not sure what he did to generate the graph in #14350.

SergeyAMZN:
> 
> I'd have to get @SergeyAMZN to rerun their test, I'm not sure what he did to generate the graph in #14350.
>

Nice, I can give a try with these changes!
The graphs come from Excel, no automated process yet unfortunately.

Although in our particular case of scene queries we might get rid of that inheritance RTTI hierarchy entirely, it was a bit too much for a simple data container. 
AZ_RTTI also adds a need for a virtual destructor and vtable pointer and we never serialize those structs anyway..

SergeyAMZN (approved):

rgba16f (approved):

burelc-amzn:
I changed the approach taken here slightly. Using a `static constexpr` class member has certain restrictions:
1. Function-local classes cannot have `static constexpr` members, so local classes that use the `AZ_RTTI`/`AZ_TYPE_INFO` macro would break.
2. Class templates that use the CRTP pattern are not able to instantiate, because the definition of the static class member from the derived class is not defined when the base class is instantiated.

This change now uses a local `constexpr` variable in the `TYPEINFO_Uuid` method. The effect is the same: the UUID string is parsed at compile time, but still allows for the above two scenarios to also work.

PR #14428, opened 2023-02-06
Intermediate assets that have had their source removed are no longer processed.

AMZN-stankowi:

## What does this PR do?

Before this change:
If the source file for an intermediate asset was removed, and changes were made outside that intermediate asset in a way that would cause it to fail to process, the intermediate asset would fail to process instead of just being removed/ignored like it should be.

To test this bug, follow these steps:
1) Sync to 955f2b2028914fa38ded66e7dcf11cf81a76fd48
2) Build & process assets
3) Sync to 4f917ff7561db3f454be1141185c11f6d493a59a (the material update)
4) Build & process assets
5) Wait for all assets to finish processing - some assets, like FBX files, will fail at first, but may succeed later.
6) Notice that some assets are still marked as failed to process, even when the Asset Processor is idle.
7) Look at those specific assets: Notice they are intermediate assets.
8) Look for the source for those intermediate assets: Notice they are missing.
9) Close Asset Processor
10) Open it again
11) Notice all assets process correctly now

This change adds an extra step in job evaluation: If the file is an intermediate asset, the file operation is not a delete operation, it's not a folder, then this looks up the source for the intermediate asset, and verifies the source still exists on disk. If not, it skips processing this intermediate asset. Other cleanup happens later, this removes that initial failure that causes Jenkins to get stuck and would confuse people on first run of Asset Processor.

## How was this PR tested?

Manual testing with the above steps in several variations.

Ran existing and new automated tests.


Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
```diff
@@ -869,6 +869,42 @@ namespace AssetProcessor
         return m_lfsPointerFileValidator->IsLfsPointerFile(filePath);
     }
 
+    AZ::Outcome<QString, void> AssetProcessorManager::GetFirstTopmostSourceAssetForProductPath(const AZStd::string& productPath)
```
amzn-mike, line 872: I think this is providing the same functionality as AssetUtilities::GetTopLevelSourceForIntermediateAsset

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
```diff
@@ -3184,6 +3220,27 @@ namespace AssetProcessor
                 // Do not process this file yet.  When the job is done it will retrigger processing for this file.
                 return;
             }
+
+            AZ::Outcome<QString, void> topmostSourceAsset = GetFirstTopmostSourceAssetForProductPath(productName);
+
+            if (topmostSourceAsset.IsSuccess())
+            {
+                if (!QFile::exists(topmostSourceAsset.GetValue()))
```
amzn-mike, line 3228: You may want to make use of the file state cache to do this check, hitting the disk tends to be more expensive.
```
auto* diskCache = AZ::Interface<IFileStateRequests>()::Get()
if(!diskCache->Exists(topmostSourceAsset.GetValue()))
```

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
```diff
@@ -3184,6 +3220,27 @@ namespace AssetProcessor
...
+                {
+                    AZ_Printf("AP", "Topmost source '%s' for intermediate asset '%.*s' no longer exists, "
+                        "so this intermediate asset will not be reprocessed.",
+                        topmostSourceAsset.GetValue().toUtf8().data(),
+                        AZ_STRING_ARG(productName));
+                    return;
```
amzn-mike, line 3234: I think this should be calling `CheckDeletedSourceFile` to clean up the asset.  Or does that happen later during CheckMissingFiles?

Code/Framework/AzCore/AzCore/Memory/AllocationRecords.cpp
```diff
@@ -467,11 +467,8 @@ namespace AZ::Debug
             m_requestedAllocs = 0;
         }
 
-        AZ_Warning(
-            "Memory", m_mode != RECORD_NO_RECORDS || mode == RECORD_NO_RECORDS,
-            "Records recording was disabled and now it's enabled! You might get assert when you free memory, if a you have allocations "
-            "which were not recorded!");
-
+        // Don't print anything out - if an issue occurs, an assert will fire off later.
```
lmbr-pip, line 470: nit: Maybe move this to a comment at start of function, also worth checking with SIG/core maintainers

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
lmbr-pip: I would also ensure we define 'topmost' in the comment. FirstTopMost is a confusing mash of naming to me and I feel either needs a better name or a comment explaining the why of this method

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
```diff
@@ -3184,6 +3220,27 @@ namespace AssetProcessor
...
+                }
+            }
+            else
+            {
+                // Failed to find topmost source - don't process this asset, there's a gap or issue in the asset database.
+                // This will hopefully be resolved once all of AP finishes processing.
```
lmbr-pip, line 3240: nit: "This hopefully..." seems oddly vague and wishful. Is there future work here or a gap that should be closed with another feature or issue.

Code/Framework/AzCore/AzCore/Memory/AllocationRecords.cpp
AMZN-stankowi: I've updated the comment and moved it. Sig/core's been tagged on this, I'll wait to push this until I get a review.

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
AMZN-stankowi: I've altered the language to describe the situation more in detail. Now the "hopefully" can just be read between the lines - if there's a gap there then this will get missed.

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
AMZN-stankowi: Done. Something I found that was slightly concerning. I used three different exist functions in building this: QFile, AZ::IO::FileIOBase::GetInstance()->Exists(fileName), and the IFileStateRequests.

QFile and IFileStateRequests handled it correctly with the path I gave it. AZ::IO::FileIOBase::GetInstance()->Exists(fileName) did not work, it was reporting the file as not existing when calling QFile and IFileStateRequests both said the file did exist. I don't have enough info here to open a separate ticket, but figured I would point this out for now in case anyone else has encountered something similar.

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
AMZN-stankowi: I've switched to using GetTopLevelSourceForIntermediateAsset. This also means I've removed the phrase "topmost" so we're now consistently using TopLevel instead.

lumberyard-employee-dm (approved):

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
AMZN-stankowi: image

Here's an example from one of the tests, at least, that shows the new message followed by the delete. This is a contrived situation, the test specifically calls these, though, but I think it shows the general idea and that cleanup does happen when AssessDeletedFile is eventually called on the source.

amzn-mike (approved):

amzn-mike (approved):

PR #14432, opened 2023-02-07
Disable Editor mode feedback when enter game mode in Editor and re-enable it when exit game mode.

VickyAtAZ:
## What does this PR do?
Disable Editor mode feedback when enter game mode in Editor and re-enable it when exit game mode.
Also fixed a gpu timestamp profiling issue when there are passes are disabled (the frame name in timestamp profiling wasn't correct since it uses time from disabled passes)

The Editor mode feedback related passes cost around 0.5 ms gpu time. This time will be saved when running level in Editor game mode.

## How was this PR tested?
Run Editor. Enable gpu profiler's timestamp profiling. Check the gpu cost for EditorModeFeedback passes. Enter game mode and check it again.


rgba16f (approved):

galibzon (approved):

Gems/Atom/Utils/Code/Source/ImGuiGpuProfiler.cpp
```diff
@@ -2014,14 +2014,17 @@ namespace AZ
...
 
                 // Add new entry to the timestamp map.
-                PassEntry* entry = addPassEntry(pass, parent);
-
-                // Recur if it's a parent.
-                if (passAsParent)
+                if (pass->IsEnabled())
                 {
-                    for (const auto& childPass : passAsParent->GetChildren())
+                    PassEntry* entry = addPassEntry(pass, parent);
```
hultonha, line 2019: nit: We could move `entry` inside the scope of `passAsParent` as it's not needed outside that scope and is closer to first use

Gems/AtomLyIntegration/EditorModeFeedback/Code/Source/EditorModeFeedbackFeatureProcessor.cpp
```diff
@@ -143,5 +143,24 @@ namespace AZ
...
+            
+            if (!m_editorStatePassSystem)
+            {
+                return;
+            }
+
+            const auto templateName = Name(m_editorStatePassSystem->GetParentPassTemplateName());
+
+            auto passFilter = AZ::RPI::PassFilter::CreateWithTemplateName(templateName, GetParentScene());
+            AZ::RPI::PassSystemInterface::Get()->ForEachPass(passFilter,  [this](RPI::Pass* pass) -> RPI::PassFilterExecutionFlow
```
hultonha, line 159: suggestion: Rather than capturing this we could do

```
[enableRender = m_enableRender]
```

in the capture clause and use that as that is all we need from the surrounding code

Gems/AtomLyIntegration/EditorModeFeedback/Code/Source/EditorModeFeedbackFeatureProcessor.h
```diff
@@ -44,6 +44,9 @@ namespace AZ
             // AZ::TickBus overrides ...
             void OnTick(float deltaTime, AZ::ScriptTimePoint time) override;
 
+            //! Enable or disable the rendering editor mode feedback
```
hultonha, line 47: uber nit: Missing period at EOL

Gems/AtomLyIntegration/EditorModeFeedback/Code/Source/EditorModeFeedbackSystemComponent.cpp
```diff
@@ -75,5 +81,35 @@ namespace AZ
...
+
+        void EditorModeFeedbackSystemComponent::OnStopPlayInEditor()
+        {
+            SetEnableRender(true);
+        }
+
+        void EditorModeFeedbackSystemComponent::SetEnableRender(bool enableRender)
+        {
+            auto sceneSystem = AzFramework::SceneSystemInterface::Get();
+            if (sceneSystem)
```
hultonha, line 98: nit: Could move this assignment inside the if to limit scope

```
if (auto* sceneSystem = AzFramework::SceneSystemInterface::Get())
// or
if (auto* sceneSystem = AzFramework::SceneSystemInterface::Get();
    sceneSystem)
```

Depending on your preference

Gems/AtomLyIntegration/EditorModeFeedback/Code/Source/EditorModeFeedbackSystemComponent.cpp
```diff
@@ -75,5 +81,35 @@ namespace AZ
...
+                    if (rpiScene)
+                    {
+                        if (AZ::Render::EditorModeFeatureProcessor* fp = (*rpiScene)->GetFeatureProcessor<AZ::Render::EditorModeFeatureProcessor>())
+                        {
+                            fp->SetEnableRender(enableRender);
+                        }
+                    }
+                }
+            }
+        }
```
hultonha, line 113: suggestion: Using early returns with flipped conditions might be nicer here to limit the amount of indentation required

Gems/AtomLyIntegration/EditorModeFeedback/Code/Source/EditorModeFeedbackSystemComponent.h
```diff
@@ -40,8 +42,16 @@ namespace AZ
             // EditorModeFeedbackInterface overrides ...
             bool IsEnabled() const override;
 
-        private:
-            //! Settings registery override for enabling/disabling editor mode feedback.
+        private:            
+            ///////////////////////////////////////////////////////////////////////
```
hultonha, line 46: nit: Can we not add the //////////// lines, they just add noise and visual clutter in my mind, we don't have them in Editor code at all anymore for the most part

Gems/AtomLyIntegration/EditorModeFeedback/Code/Source/EditorModeFeedbackSystemComponent.h
```diff
@@ -40,8 +42,16 @@ namespace AZ
...
+            // AzToolsFramework::EditorEntityContextNotificationBus overrides
```
hultonha, line 47: nit: `EditorEntityContextNotificationBus overrides ...` (omit namespace in this context for brevity and include ellipsis)

Gems/AtomLyIntegration/EditorModeFeedback/Code/Source/EditorModeFeedbackSystemComponent.h
```diff
@@ -40,8 +42,16 @@ namespace AZ
...
+            void OnStartPlayInEditorBegin() override;
+            void OnStopPlayInEditor() override;
+
+            // Enable/disable editor mode feedback rendering for the level viewport (main scene)
+            void SetEnableRender(bool enableRender);
+
+            //! Settings registry override for enabling/disabling editor mode feedback.
```
hultonha, line 54: Thanks for adding this comment! 👍

hultonha (approved):
Changes look great, thanks @VickyAtAZ 👍 My feedback is mostly pretty minor, nothing blocking. One thing that would be nice though is some numbers of before/after in the description of how much time this saved. Some extra data/stats would be really useful to see to gauge the impact of this change. Thanks!

jonawals (approved):
LGTM! Thanks for fixing this!

Gems/Atom/Utils/Code/Source/ImGuiGpuProfiler.cpp
VickyAtAZ: I don't think we can. 
The entry is return value of addPassEntry(). And addPassEntry needs to be called outside passAsParent.

Gems/AtomLyIntegration/EditorModeFeedback/Code/Source/EditorModeFeedbackFeatureProcessor.h
VickyAtAZ: I think we usually don't add period in the end if the comment is just one line. Although I will add it since it's not consistent in this file.

PR #14445, opened 2023-02-07
Add missing diffuse probe grid assets to seed list.

mbalfour-amzn:
## What does this PR do?

Adds some diffuse probe grid assets that were missing from the Atom seed list, which caused errors when trying to run release pak builds.

## How was this PR tested?

Ran a MultiplayerSample build in profile with the AP disconnected and verified that the missing asset errors I was seeing from this no longer happen. There are still some asset load errors related to shader variants, but a GHI has been filed for that.

### DIFF
--- a/Gems/Atom/Feature/Common/Assets/seedList.seed
+++ b/Gems/Atom/Feature/Common/Assets/seedList.seed
@@ -336,6 +336,46 @@
 			<Class name="unsigned int" field="platformFlags" value="1" type="{43DA906B-7DEF-4CA8-9790-854106D3F983}"/>
 			<Class name="AZStd::string" field="pathHint" value="passes/diffuseprobegridvisualizationpassrequest.azasset" type="{03AAAB3F-5C47-5A66-9EBC-D5FA4DB353C9}"/>
 		</Class>
+		<Class name="SeedInfo" field="element" version="2" type="{FACC3682-2ACA-4AA4-B85A-07AD276D18A0}">
+			<Class name="AssetId" field="assetId" version="1" type="{652ED536-3402-439B-AEBE-4A5DBC554085}">
+				<Class name="AZ::Uuid" field="guid" value="{8CC9B575-39D7-5914-AC42-92A77833E63A}" type="{E152C105-A133-4D03-BBF8-3D4B2FBA3E2A}"/>
+				<Class name="unsigned int" field="subId" value="268692035" type="{43DA906B-7DEF-4CA8-9790-854106D3F983}"/>
+			</Class>
+			<Class name="unsigned int" field="platformFlags" value="1" type="{43DA906B-7DEF-4CA8-9790-854106D3F983}"/>
+			<Class name="AZStd::string" field="pathHint" value="models/diffuseprobesphere.azmodel" type="{03AAAB3F-5C47-5A66-9EBC-D5FA4DB353C9}"/>
+		</Class>
+		<Class name="SeedInfo" field="element" version="2" type="{FACC3682-2ACA-4AA4-B85A-07AD276D18A0}">
+			<Class name="AssetId" field="assetId" version="1" type="{652ED536-3402-439B-AEBE-4A5DBC554085}">
+				<Class name="AZ::Uuid" field="guid" value="{A6F9D127-D963-5D78-B19F-A31E73C5C2B5}" type="{E152C105-A133-4D03-BBF8-3D4B2FBA3E2A}"/>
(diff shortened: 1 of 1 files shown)


galibzon (approved):

Gems/Atom/Feature/Common/Assets/seedList.seed
```diff
@@ -336,6 +336,46 @@
...
+			<Class name="unsigned int" field="platformFlags" value="1" type="{43DA906B-7DEF-4CA8-9790-854106D3F983}"/>
+			<Class name="AZStd::string" field="pathHint" value="shaders/diffuseglobalillumination/diffuseprobegridquery.azshader" type="{03AAAB3F-5C47-5A66-9EBC-D5FA4DB353C9}"/>
+		</Class>
+		<Class name="SeedInfo" field="element" version="2" type="{FACC3682-2ACA-4AA4-B85A-07AD276D18A0}">
+			<Class name="AssetId" field="assetId" version="1" type="{652ED536-3402-439B-AEBE-4A5DBC554085}">
+				<Class name="AZ::Uuid" field="guid" value="{B3B664B6-4340-5A35-9A67-DC00A5361D62}" type="{E152C105-A133-4D03-BBF8-3D4B2FBA3E2A}"/>
+				<Class name="unsigned int" field="subId" value="0" type="{43DA906B-7DEF-4CA8-9790-854106D3F983}"/>
+			</Class>
+			<Class name="unsigned int" field="platformFlags" value="1" type="{43DA906B-7DEF-4CA8-9790-854106D3F983}"/>
+			<Class name="AZStd::string" field="pathHint" value="shaders/diffuseglobalillumination/diffuseprobegridvisualizationprepare.azshader" type="{03AAAB3F-5C47-5A66-9EBC-D5FA4DB353C9}"/>
```
AMZN-stankowi, line 377: the title of this shader and the model added make it sound like this is a debug visualization tool that we might not want in release builds. Maybe the code should be instead updated to not need these to exist, to work?

Or am I mis-understanding what these assets are, and they do have a functional use to end-players of distributed games?

AMZN-stankowi (changes requested):

Gems/Atom/Feature/Common/Assets/seedList.seed
dmcdiarmid-ly: The visualization passes, shaders, and sphere model are used for development purposes to debug lighting.

mbalfour-amzn: Is there anything that currently prevents them from trying to load in release builds?

AMZN-stankowi: I think for now it might make more sense to temporarily add those seeds to the Multiplayer Sample seed list, to encourage the proper fix for these systems here. If we put in this work around, it will mean that we ship a debug feature in release builds longer than necessary.

However, if people disagree with me, I can drop the requested change and instead change to a ship it. I'm just worried that if this goes in, there's no incentive to address the actual underlying problem.

mbalfour-amzn: IMO, it's better to leave things in a working but non-optimal state vs a broken state that requires anyone creating release builds to go through the same sort of debugging I just did to discover the same problem and solution. Having me institute a workaround in my project _also_ removes the incentive to address the underlying problem, so why not create functioning release builds for _all_ projects by default in the meantime?

That said, if @dmcdiarmid-ly or someone has concrete plans to fix this in the very near-term future, I'm happy to drop this PR.

AMZN-stankowi: I guess part of it is how harmful is it to ship the debug content in release builds. For teams really trying to optimize build sizes, this is extra content, including an azmodel, and an entire feature they probably don't want in that build.

Most of the cases I've been working on with release builds have been to get release builds as small as possible (mobile games, or games with frequent updates).

I'll drop my "Requested Changes" and change to a ship it for now so this doesn't block other work.

AMZN-stankowi (approved):

Gems/Atom/Feature/Common/Assets/seedList.seed
mbalfour-amzn: I submitted a GHI for cleaning up the release build behavior of Diffuse Probes: https://github.com/o3de/o3de/issues/14463 .

AMZN-stankowi: Perfect, thanks.

Issue #14263, opened 2023-01-27
Bug Report: Material Canvas sorting below output node options


ParasiteSt:
**Describe the bug**
One time a Material Input parameter was put below the output nodes parameters in the material instance/editor.

**Steps to reproduce**
n/a - not sure why it happened, possibly because there is 47 other material inputs and none are sorted into groups so they all default into the general group.

**Expected behavior**
All parameters should be above the output node's parameters.

**Actual behavior**
It is possible for an input parameter to be below the output nodes parameters.

**Screenshots/Video**
![image](https://user-images.githubusercontent.com/100494538/215145990-5ce272b5-7ac7-4896-85f3-754e27d00ace.png)

**Found in Branch**
Development

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
0a666643fd39795fbb33383764017a8ec8e1a4d7

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Windows
 - Version 11

PR #14448, opened 2023-02-07
Material Canvas: Sorting generated material type groups and properties

gadams3:
## What does this PR do?

This change implements automatic sorting of material type groups and properties emitted from material canvas.

Material input nodes are used to expose material properties on the generated material type. These nodes have fields that allow the user to configure the material type property names, groups, and descriptions. The material type groups and properties will be registered in the order nodes are traversed. That order is not deterministic. There is currently no variable manager or other mechanism to explicitly dictate the organization of groups and properties. In lieu of that, the groups and properties will be sorted alphabetically.

Resolves https://github.com/o3de/o3de/issues/14263

## How was this PR tested?

Created included test graph with multiple texture material inputs, each with a different name and group.
Confirmed that generated material type recursively sorted all of the groups and properties by name.
Confirmed that properties appeared in the correct order in material editor and instance inspector.

### DIFF
--- a/Gems/Atom/RPI/Code/Source/RPI.Edit/Material/MaterialTypeSourceData.cpp
+++ b/Gems/Atom/RPI/Code/Source/RPI.Edit/Material/MaterialTypeSourceData.cpp
@@ -36,25 +36,23 @@ namespace AZ
             void SortMaterialTypeSourceDataPropertyGroups(
                 AZStd::vector<AZStd::unique_ptr<MaterialTypeSourceData::PropertyGroup>>& propertyGroups)
             {
-                // Sort child groups alphabetically
+                // Sort child groups alphabetically with general groups to the back of the list
                 AZStd::sort(
                     propertyGroups.begin(),
                     propertyGroups.end(),
                     [](auto& group1, auto& group2)
                     {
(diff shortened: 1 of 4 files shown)


cgalvan (approved):

Gems/Atom/RPI/Code/Source/RPI.Edit/Material/MaterialTypeSourceData.cpp
```diff
@@ -32,6 +32,40 @@ namespace AZ
...
+                AZStd::stable_sort(
+                    propertyGroups.begin(),
+                    propertyGroups.end(),
+                    [](auto& group1, auto& group2)
+                    {
+                        static constexpr AZStd::string_view generalGroupName = "general";
+                        const auto& groupName1 = group1->GetName();
+                        const auto& groupName2 = group2->GetName();
+                        return AZ::StringFunc::Equal(groupName1, generalGroupName) < AZ::StringFunc::Equal(groupName2, generalGroupName);
+                    });
```
lumberyard-employee-dm, line 58: It seems like these two sorts can be combined where the any general group name check occurs first.

galibzon (approved):

PR #14450, opened 2023-02-07
Ensure All AZ::Console Commands Eventually Get Called 

### LINKED ISSUE #14449: AZ::Console PerformCommand Never Executed If It's Called Too Early
It's possible to call a console command that is never dispatched. It is expected that console commands will be deferred for a later time if they are called upon too early before the command has been registered with the AZ::Console

**Steps to reproduce**
Call AZ::Interface<AZ::IConsole>::Get()->PerformCommand on any CVAR before the Gem owning that module is loaded.
1. Open \Code\LauncherUnified\Launcher.cpp
3. Call AZ::Interface<AZ::IConsole>::Get()->PerformCommand("sv_isDedicatedServer false") before gameApplication.Start
4. Query sv_isDedicatedServer some time after (for example inside a system component OnActivate())

**Expected behavior**
sv_isDedicatedServer should return false

**Actual behavior**
sv_isDedicatedServer remains it's default value ("true") even though we set it to false.

**Found in Branch**
development

AMZN-Gene:
Ensure AZ::Console::PerformCommand() will defer commands to be executed at a later time if the command isn't yet registered
Fixes #14449 

## How was this PR tested?
Call PerformCommand on a cvar before the gem module owning that cvar was registered. Later checked the cvar and noticed the cvar is now properly set.

### DIFF
--- a/Code/Framework/AzCore/AzCore/Console/Console.cpp
+++ b/Code/Framework/AzCore/AzCore/Console/Console.cpp
@@ -110,7 +110,21 @@ namespace AZ
         ConsoleFunctorFlags requiredClear
     )
     {
-        return DispatchCommand(command, commandArgs, silentMode, invokedFrom, requiredSet, requiredClear);
+        if (!DispatchCommand(command, commandArgs, silentMode, invokedFrom, requiredSet, requiredClear))
+        {
+            // If the command could not be dispatched at this time add it to the deferred commands queue
+            DeferredCommand deferredCommand{ AZStd::string_view{ command },
+                                             DeferredCommand::DeferredArguments{ commandArgs.begin(), commandArgs.end() },
+                                             silentMode,
+                                             invokedFrom,
+                                             requiredSet,
+                                             requiredClear };
+
+            m_deferredCommands.emplace_back(AZStd::move(deferredCommand));
+            return false;
+        }
+
+        return true;
     }
 
     void Console::ExecuteConfigFile(AZStd::string_view configFileName)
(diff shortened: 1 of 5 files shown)


lumberyard-employee-dm (approved):

AMZN-Olex (approved):

Code/Framework/AzCore/AzCore/Console/Console.cpp
```diff
@@ -110,7 +110,21 @@ namespace AZ
...
+            // If the command could not be dispatched at this time add it to the deferred commands queue
+            DeferredCommand deferredCommand{ AZStd::string_view{ command },
+                                             DeferredCommand::DeferredArguments{ commandArgs.begin(), commandArgs.end() },
+                                             silentMode,
+                                             invokedFrom,
+                                             requiredSet,
+                                             requiredClear };
+
+            m_deferredCommands.emplace_back(AZStd::move(deferredCommand));
+            return false;
```
AMZN-stankowi, line 124: Nitpick: It might be better to change this to either return an enum (return PerformCommandResult::Defferred;) or an AZ::Outcome, to make it clear to callers that getting a non-success result doesn't mean their command won't eventually run.

Especially because right now, the comment on the interface says it returns true on success, and false otherwise. In this case, you don't know if it's successful or not on running the command yet because it's been deferred, so it would be useful for callers to know it's been deferred and not just failed to invoke.

https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Console/IConsole.h#L46

AMZN-stankowi (approved):

Code/Framework/AzCore/AzCore/Console/Console.cpp
AMZN-Gene: True, I was thinking the same. This is a widely used method so it'll definitely require proper deprecation..

lumberyard-employee-dm: If an `AZ::Outcome<void, AZStd::string>` is returned, it should be pretty transparent to the user.

lumberyard-employee-dm: Doing a search, there is only one location in the codebase where we check the result of PerformCommand and that is an the SettingsRegistryConsoleUtilsTest.

https://sourcegraph.com/search?q=context:global+repo:github%5C.com/o3de/+PerformCommand&patternType=standard&sm=1&groupBy=repo

AMZN-Gene: Ah I see, AZ::Outcome has a bool operator. Wow, I'm surprised by the amount of times we use AZ::Outcome<void,.,.>Enum seems more natural... I think Outcome should be when there's a return value the user actually wants to retrieve. I'm cool using enum even if it requires the deprecating step if it makes more sense. Thoughts?

AMZN-Gene: Fixed!

lumberyard-employee-dm (approved):

PR #14452, opened 2023-02-07
Fixed error while loading PhysX materials.

moraaar:

## What does this PR do?

Fixes a bug that causes a crash when while waiting for a physics material asset to load the asset manager dispatches events for other assets. These events can cause `OnAssetReady` of other assets to be called and if those end up trying to create the same physics material then it leads to the crash.

Solved by creating the material first, then block until loaded.

Also renamed the utility function `GetShapesFromAsset` to `CreateShapesFromAsset` since that reflects better what it does.

## How was this PR tested?

Run PhysX unit tests.
Tested using a local level that was crashing while loading due to the issue described above. After this fix the level loads correctly. Also debugged the loading of colliders and rigid bodies to verify it works correctly.


Code/Framework/AzFramework/AzFramework/Physics/Material/PhysicsMaterialManager.cpp
```diff
@@ -62,24 +62,27 @@ namespace Physics
...
-            m_materials.emplace(newMaterial->GetId(), newMaterial);
+            assetLocal.QueueLoad();
+
+            // The call to BlockUntilLoadComplete must happen after the new material
+            // has been added to m_materials container. This is because BlockUntilLoadComplete
+            // will dispatch events for other assets while waiting, and if other OnAssetRead
+            // code ends up trying find/create the same material it needs to find it in the
+            // container and return it.
+            assetLocal.BlockUntilLoadComplete();
         }
```
moraaar, line 86: An alternative approach to this would be to make all the physx components that use physx materials to delay their initialization until all the physics materials it uses are loaded. But we cannot go for this solution at the moment because all the colliders need to create the physx shapes during activation. The current implementation cannot wait a few frames later to create the shapes, the the rigid body components also need the shaped to create the simulated body on entity activated.

If we change how of all physx components are initialized with a more dynamic approach based on events to allow the creation of shapes and simulated bodies at any time, then it would be possible to delay the creation until all the assets are ready.

Not sure if it's a good idea overall to delay the creation of physical entities though, that might lead to objects falling though floors, or trespassing walls.

Code/Framework/AzFramework/AzFramework/Physics/Material/PhysicsMaterialManager.cpp
```diff
@@ -62,24 +62,27 @@ namespace Physics
...
         {
-            assetLocal.QueueLoad();
-            assetLocal.BlockUntilLoadComplete();
-
-            // Failed to load the asset
-            if (!assetLocal.IsReady())
-            {
-                return nullptr;
-            }
+            m_materials.emplace(newMaterial->GetId(), newMaterial);
```
hultonha, line 68: suggestion: As `newMaterial` isn't used after this it might be worth moving to an if initializer - we might also then want to `AZStd::move` the value into the `m_materials` container

```

if (auto newMaterial = CreateMaterialInternal(id, materialAsset);
    newMaterial)
{
    m_materials.emplace(newMaterial->GetId(), AZStd::move(newMaterial));
```

Code/Framework/AzFramework/AzFramework/Physics/Material/PhysicsMaterialManager.cpp
```diff
@@ -62,24 +62,27 @@ namespace Physics
...
         }
 
-        auto newMaterial = CreateMaterialInternal(id, assetLocal);
-        if (newMaterial)
+        // Block until the material asset is loaded.
+        // It's important that the loading of physics material assets is not delayed,
+        // otherwise objects would react differently during simulation during a few
+        // frames until the material is loaded, causing unexpected behaviours.
```
hultonha, line 74: uber nit: We're recommended to use American English in code and comments so this should be `behavior` I'm afraid 😅

hultonha (approved):
Change looks good to me! I just have a few minor suggestions 👍 Be good to get the blessing of @mbalfour-amzn and @AMZN-stankowi/@amzn-mike. Thanks

Code/Framework/AzFramework/AzFramework/Physics/Material/PhysicsMaterialManager.cpp
moraaar: `newMaterial` is returned at the end, so it can't be moved.

Code/Framework/AzFramework/AzFramework/Physics/Material/PhysicsMaterialManager.cpp
moraaar: Thanks for letting me know, I confuse them :)

moraaar:
I messed up this PR by pushing in the middle of a merge and it hasn't fix doing a subsequent merge, so I'll close this and open a new PR.

moraaar:
New PR for this: https://github.com/o3de/o3de/pull/14462

Issue #5541, opened 2021-11-11
[Light component] The parameter Height for the Polygon light is set to 0 improperly when operating any other parameter


instoneks2018:
**Describe the bug** 
the parameter Height for the Polygon light is set to `0.0` improperly when operating any other parameter on the `Light` panel,
and then, I try to fix this（delete the line of code https://github.com/o3de/o3de/blob/23faf4d36aafdb9f2fc38186d090d9143946e29d/Gems/AtomLyIntegration/CommonFeatures/Code/Source/CoreLights/PolygonLightDelegate.cpp#L22 ), the `Height` value is unchanged, PTAL.      
**To Reproduce**
1.Open Editor.
2.Create a new Entity.
3.Add Light Component.
4.Select `Polygon` as Light type for light component.
5.Change the `Height` value in `Polygon Prism Shape`.
6.Change any other parameter on the `Light` panel.

**Expected behavior**
 The `Height` value set previously  is unchanged
**Screenshots**
1.Before changing  parameter, the `Height` value is `10.0` .
![before](https://user-images.githubusercontent.com/45937683/141291544-395512d2-11d2-4bea-813d-6e8e5a58f735.png)

2.After changing the change `Mode` from `Automatic` to `Explicit`，  the `Height` value is changed from `10.0` to `0.0`. If you change any other parameter on `Light` panel, the result is also the same.
![after](https://user-images.githubusercontent.com/45937683/141291638-2edab4f6-51f8-44b3-ab58-b3768caac096.png)
**Desktop/Device (please complete the following information):**
- Device: [PC]
- OS: [Windows]
- Version [10]
- CPU [Intel I7-10700 2.90GHz (16 CPUs)]
- GPU [Intel(R) UHD Graphics 630]
- Memory [32GB]

**Additional context**

invertednormal:
This isn't a bug, the behavior is intentional. Extruding the polygon doesn't change the lighting - polygon lights can only emit from a polygon plane, not a 3d surface. Perhaps we should create some way for the polygon light to disable or hide the height field to avoid this confusion.

invertednormal:
Created a new issue for the UX problem #14472. Closing this one .

Issue #6127, opened 2021-12-03
Polygon light type height value is reverted to 0 after leaving Game Mode


LB-BartoszDudzinski:
**Describe the bug**
Polygon Prism Shape in an entity with a Light component has it's height value reset after leaving the Game Mode. This issue only occurs if a Polygon Prism Shape component is used in conjunction with a Light component.

**Steps to reproduce**
1. Create an entity.
2. Add a Light component to it.
3. Select Polygon as the light type.
4. Change the height value.
5. Enter and leave the Game Mode.

**Expected behavior**
The height value does not change.

**Actual behavior**
The height value is reverted to 0.

**Screenshots/Video**

https://user-images.githubusercontent.com/86952009/144581439-9bc4ce4c-bea3-4149-ad65-e42ccc4d06da.mp4

**Found in Branch**
Development ([cfb426b](https://github.com/o3de/o3de/commit/cfb426b86c34be6474095433c10844610fcbd070))
The issue occurs in O3DE Stable 21.11 Release.

**Desktop/Device:**

- Device: PC
- OS: Windows
- Version 10
- CPU Intel I5-9600k
- GPU RTX 2060
- Memory 16GB

monroegm:
Sending to Graphics-Audio, the owners of the component

invertednormal:
The polygon light only emits light from the 2d surface of a flat polygon plane. The fact that height is adjustable is just an artifact of using the polygon prism component to define the polygon shape. This is why the height is always set back to 0 whenever you make some change to the light settings, or enter/exit game mode. I've created https://github.com/o3de/o3de/issues/14472 to help clear up the UX issue.

LB-BartoszDudzinski:
Confirming the issue has been closed and will be tracked in [#14472](https://github.com/o3de/o3de/issues/14472).

Issue #5931, opened 2021-11-25
[Polygon light]the Polygon light source has illumination at  height of 0.0, but may not have  illumination at other location of other height


instoneks2018:
**Describe the bug**
The `Polygon light` source has illumination at  height of 0.0, but may not have  illumination at other location of height.

**Steps to reproduce**
1.Open Editor.
2.Create a new Entity.
3.Add Light Component.
4.Select `Polygon` as Light type for light component.
5. Change the `Height` value in `Polygon Prism Shape`.
6.Move the height of the light source and observe the illumination effect.

**Expected behavior**
The `Polygon light` source illumination is not affected by height position,each height point should be illuminated.

**Screenshots**
```
1. The height position is 0.0,  the objects (cube and rabbit) in the scene have bright illumination.
```
![before2](https://user-images.githubusercontent.com/45937683/143442955-24f21165-eed0-44ed-86a8-50df3333fe4a.png)
]
```
2. Translate the `Polygon light` from right to left，the height position is 10.0,  the object cube and rabbit of the scene are dimmed and the dividing line in both directions is also at height of 0.0.
```
![after2](https://user-images.githubusercontent.com/45937683/143443296-31a75ae8-8e87-4cd9-b860-08790392e4b3.png)

**Desktop/Device (please complete the following information):**
- Device: [PC]
- OS: [Windows]
- Version [10

- CPU [Intel I7-10700 2.90GHz (16 CPUs)]
- GPU [Intel(R) UHD Graphics 630]
- Memory [32GB]

**Additional context**

invertednormal:
The polygon light only emits light from the 2d surface of a flat polygon plane. The fact that height is adjustable is just an artifact of using the polygon prism component to define the polygon shape. This is why the height is always set back to 0 whenever you make some change to the light settings. I've created #14472 to help clear up the UX issue.

PR #14455, opened 2023-02-08
Fix the height of the polygon light will be reset to 0 if other parameters are modified

Zhang-Baochong:
Signed-off-by: Zhang-Baochong <zhangbaochong@huawei.com>

## What does this PR do?
 The parameter Height for the Polygon light is set to 0 improperly when operating any other parameter.
Fixes : #5541

## How was this PR tested?
Manually tested in editor.
1.Open Editor.
2.Create a new Entity.
3.Add `Light` Component.
4.Select `Polygon` as Light type.
5.Change the Height value.
6.Change any other parameter on the Light panel, the Height value set previously is unchanged.

### DIFF
--- a/Gems/AtomLyIntegration/CommonFeatures/Code/Source/CoreLights/PolygonLightDelegate.cpp
+++ b/Gems/AtomLyIntegration/CommonFeatures/Code/Source/CoreLights/PolygonLightDelegate.cpp
@@ -17,7 +17,6 @@ namespace AZ::Render
         , m_shapeBus(shapeBus)
     {
         InitBase(entityId);
-        shapeBus->SetHeight(0.0f);
     }
 
     void PolygonLightDelegate::SetLightEmitsBothDirections(bool lightEmitsBothDirections)


Jackie9527 (approved):
lgtm

cgalvan (approved):

invertednormal:
This behavior is intentional. The polygon prism shape allows you to create 3d extrusions of polygons, but for the polygon light, extruding the polygon doesn't make sense since the light is only 2d. What problem does this actually solve - is there a bug dealing with the height? is it just confusing that you can set the height to a non-zero value only to have it changed back to zero? Maybe we need to add the ability for the polygon light to disable the height field?

invertednormal:
Sorry, just noticed the bug link. I think the real issue is UX - it shouldn't be possible to extrude the polygon when its used for polygon lights because the polygon light is 2d. I commented on the bug.

invertednormal (changes requested):
The polygon height should stay at 0 when used in a polygon light. For a better UX fix, we should open up the ability for the polygon shape component to exist in a 2d-only mode with height disabled or completely hidden.

invertednormal:
I created a new bug for the UX issue here - #14472

Zhang-Baochong:
> The polygon height should stay at 0 when used in a polygon light. For a better UX fix, we should open up the ability for the polygon shape component to exist in a 2d-only mode with height disabled or completely hidden.

So #5931 #6127 may be not a bug too?

invertednormal:
> > The polygon height should stay at 0 when used in a polygon light. For a better UX fix, we should open up the ability for the polygon shape component to exist in a 2d-only mode with height disabled or completely hidden.
> 
> So #5931 #6127 may be not a bug too?

Correct - I'll close both of those and refer them to the UX issue. Thanks for bringing those to my attention.

Issue #3007, opened 2021-08-10
[PhysX] Color preview looks different when color picker is open when selecting color for a Box Shape.


Jackie9527:
**Describe the bug**
Set color of Box Shape to 255,0,0 with color picker, when color picker is open, then the color preview is brighter than color picker  is closed.

**To Reproduce**
1. Open Editor.
2. Create a new Entity.
3. Add PhysX Collider and Box Shape.
4. Set Shape Color with color picker, set RGB value as 255,0,0.
5. Open/Close the color picker several times, check the color preview.
6. See difference.

Also, when going into game mode, the color of box shape looks different with the color preview.
1. Open Editor.
2. Create a new Entity.
3. Add PhysX Collider and Box Shape.
4. Set Shape Color with color picker, set RGB value as 255,0,0.
5. Enable Game View.
6. Make a screen clip with the color preview in it.
7. Go to game mode, check the color of box with the color in clip.
8. See difference.

**Expected behavior**
No difference whether the color picker is open or closed.

**Screenshots**

**Desktop/Device (please complete the following information):**
 - Device: [e.g. PC, Mac, iPhone, Samsung] 
 - OS: [e.g. Windows, macOS, iOS, Android]
 - Version [e.g. 10, Bug Sur, Oreo]
 - CPU [e.g. Intel I9-9900k , Ryzen 5900x, ]
 - GPU [AMD 6800 XT, NVidia RTX 3090]
 - Memory [e.g. 16GB]

**Additional context**

HogJonny-AMZN:
Just reviewing this and commenting in my own words.

1. When I create a Box Shape, the default color is yellow-ish
2. It renders this color in the viewport
3. The color swatch for the Box Shape color is brown-ish
4. If I click-on the color swatch it open the color picker
5. The color picker displays a yollow-ish color (the correct sRGB color.)
6. The color swatch of the Box Shape suddenly pops to display yellow (same as in the picker.)
7. When I hit OK and close the Color Picker, the Box Shape color swatch has the same numerical value but now remains yellow.

What this implies to me ... the brown you initially see, likely is the Linear RGB value of this yellow (many colors, such as those in materials, are converted from the color picker working color space to the renderer Linear RGB color space).  And that the Color Picker is somehow putting this swatch into a state where both the working (picking) and the display/storage state are all in the same color space of sRGB.

[...]

PR #14458, opened 2023-02-08
bugfix for 3007

Jackie9527:

## What does this PR do?

fix for #3007

## How was this PR tested?
before

after
3007-box
3007-capsule
3007-cylinder
3007-disk
3007-polygon
3007-quad
3007-sphere
3007-tube

### DIFF
--- a/Gems/LmbrCentral/Code/Source/Shape/EditorBaseShapeComponent.cpp
+++ b/Gems/LmbrCentral/Code/Source/Shape/EditorBaseShapeComponent.cpp
@@ -200,11 +200,12 @@ namespace LmbrCentral
 
     void EditorBaseShapeComponent::OnShapeColorChanged()
     {
-        m_shapeColor.SetA(AzFramework::ViewportColors::DeselectedColor.GetA());
+        AZ::Color drawColor(m_shapeColor);
+        drawColor.SetA(AzFramework::ViewportColors::DeselectedColor.GetA());
 
         if (m_shapeConfig)
         {
-            m_shapeConfig->SetDrawColor(m_shapeColor);
+            m_shapeConfig->SetDrawColor(drawColor);
         }
     }
 
--- a/Gems/LmbrCentral/Code/Source/Shape/EditorCapsuleShapeComponent.cpp
+++ b/Gems/LmbrCentral/Code/Source/Shape/EditorCapsuleShapeComponent.cpp
@@ -119,7 +119,7 @@ namespace LmbrCentral
             {
                 DrawShape(
                     debugDisplay,
-                    ShapeDrawParams{ m_shapeColor, m_shapeWireColor, m_displayFilled },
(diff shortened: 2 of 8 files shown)


moraaar (approved):
LTGM

Gems/LmbrCentral/Code/Source/Shape/EditorBaseShapeComponent.cpp
```diff
@@ -200,11 +200,12 @@ namespace LmbrCentral
 
     void EditorBaseShapeComponent::OnShapeColorChanged()
     {
-        m_shapeColor.SetA(AzFramework::ViewportColors::DeselectedColor.GetA());
+        AZ::Color drawColor(m_shapeColor);
+        drawColor.SetA(AzFramework::ViewportColors::DeselectedColor.GetA());
```
hultonha, line 204: question: Is this change definitely needed? I think I understand the other ones but do things work as expected with this change reverted? I don't quite follow how this change is different, though maybe it is something subtle I'm missing

gadams3 (approved):

cgalvan:
I'd like to get @HogJonny-AMZN input on this before it goes in. We've had some discussions in the past on color picker inconsistencies.

HogJonny-AMZN:
Based on my validation of the bug: https://github.com/o3de/o3de/issues/3007
- I left some screenshots, something is wonky and I could validate it
- In the bug, the issue seems to be isolated to the color swatch display of the shape components.

From what I see of the code changes here, they are isolated to the Shape Components (and don't try to cascade out into changing the color picker, or changes that would spill over into color elsewhere like for instance in a material.)

I approve of these changes.

HogJonny-AMZN (approved):

Gems/LmbrCentral/Code/Source/Shape/EditorBaseShapeComponent.cpp
cgalvan: @Jackie9527 can you speak to this feedback from @hultonha ?

Jackie9527: The color of label would be modified, because its alpha value is changed to `0.4`, which is weird.
image
image

cgalvan (approved):

Gems/LmbrCentral/Code/Source/Shape/EditorBaseShapeComponent.cpp
cgalvan: @hultonha Can you take a look at Jackie's response and see if it answers your question? Thanks!

hultonha: Sorry I missed this before, thanks for pinging me @cgalvan. Sorry @Jackie9527 I think I understand now, apologies I misunderstood the original code and see why the temporary is needed now to not modify the member variable. Looks good to me thanks for getting back to me 👍

hultonha (approved):

PR #14459, opened 2023-02-08
bugfix for color and fill mode of shapes.

Jackie9527:

## What does this PR do?

When I was working on #3007, I found that there is something wrong with the shape color of quad and disk in scene.
It's fix for the problem.
before

## How was this PR tested?
after


Code/Framework/AzFramework/AzFramework/UnitTest/TestDebugDisplayRequests.cpp
```diff
@@ -58,8 +58,9 @@ namespace UnitTest
         m_points.push_back(tm.TransformPoint(AZ::Vector3(0.5f * width, 0.0f, 0.5f * height)));
     }
 
-    void TestDebugDisplayRequests::DrawQuad(float width, float height)
+    void TestDebugDisplayRequests::DrawQuad(float width, float height, bool drawShaded)
```
hultonha, line 61: nit: For new code please prefer `[[maybe_unused]]` in the parameter list as opposed to cast to void in the function body

hultonha (approved):
Code change looks good to me (thanks for looking at this!) but would be useful to get approval from @o3de/sig-ui-ux to confirm the behavior change is inline with what we want.

One quick question is why the cube looks the same as before and different to all the others now, should we try and make that consistent too?

Tagging @rgba16f and @invertednormal for visibility as they know about this code.

gadams3 (approved):

rgba16f (approved):

Code/Framework/AzFramework/AzFramework/UnitTest/TestDebugDisplayRequests.cpp
Jackie9527: `void` is removed.

invertednormal (approved):
Looks good, but like @hultonha mentioned, I wonder if the same treatment should be given to `DrawSolidBox()` and `DrawSolidOBB()`

cgalvan (approved):

Code/Framework/AzFramework/AzFramework/UnitTest/TestDebugDisplayRequests.cpp
hultonha: Awesome thanks!

PR #14461, opened 2023-02-08
EMFX Raycasts optimization

SergeyAMZN:

## What does this PR do?

Optimizes EMotionFX raycasts by replacing EBus calls with AZ::Interface 

## How was this PR tested?

MP Sample with 100 characters using these raycasts for the foot IK, unit tests

EMFXRaycasts

The average cost of a single EMFX raycast call in a test (MP SampleBase with 58 characters) is:

* Via EBus: 198.4321 us (peaking 4231 us)
* Via AZ::Interface: 25.3555 us (peaking 1107 us)


Gems/EMotionFX/Code/EMotionFX/Source/BlendTreeRaycastNode.cpp
```diff
@@ -88,18 +88,20 @@ namespace EMotionFX
...
         }
 
-        Integration::RaycastRequests::RaycastRequest rayRequest;
+        Integration::IRaycastRequests::RaycastRequest rayRequest;
         rayRequest.m_start      = rayStart;
         rayRequest.m_direction  = rayDirection;
         rayRequest.m_distance   = maxDistance;
         rayRequest.m_queryType  = AzPhysics::SceneQuery::QueryType::Static;
-        rayRequest.m_hint       = Integration::RaycastRequests::UsecaseHint::Generic;
+        rayRequest.m_hint = Integration::IRaycastRequests::UsecaseHint::Generic;
```
amzn-ahmadkrm, line 96: nit: keep the extra spacing here? just to make it consistent with the rest

amzn-ahmadkrm (approved):
LGTM!

Gems/EMotionFX/Code/Include/Integration/AnimationBus.h
```diff
@@ -169,17 +169,13 @@ namespace EMotionFX
...
-         * The raycast request bus, which EMotion FX calls in order to perform ray cast tests.
+         * The raycast request interface, which EMotion FX calls in order to perform ray cast tests.
          * This allows you to perform custom filtering.
          */
-        class RaycastRequests : public AZ::EBusTraits
+        class IRaycastRequests
         {
         public:
-            // Enable multi-threaded access by locking primitive using a mutex when connecting handlers to the EBus or executing events.
-            using MutexType = AZStd::recursive_mutex;
```
moraaar, line 179: Change looks good. Quick question before approving. Normally when the mutex is used in a bus is because is needed to be called from multiple threads and this protects the bus so the calls don't happen at the same time. Will the interface work appropriately if it's called from multiple threads?

gadams3 (approved):
Looks good to me other than the same question about multithreading. With interfaces you lose the automatic support for wrapping function calls and bus connections and disconnections with a mutex. But if that is an issue here then the mutex or other protection can be moved directly into the class implementing the interface.

Gems/EMotionFX/Code/Include/Integration/AnimationBus.h
SergeyAMZN: Calling from multiple threads is exactly when the performance issues from EBus locks arise. 
Yes, the raycast itself is thread safe.

moraaar (approved):

Gems/EMotionFX/Code/EMotionFX/Source/BlendTreeFootIKNode.cpp
```diff
@@ -311,18 +311,20 @@ namespace EMotionFX
             heightOffset += uniqueData->m_legs[legId].m_toeHeight * actorInstanceScale;
         }
 
-        Integration::RaycastRequests::RaycastRequest rayRequest;
+        Integration::IRaycastRequests::RaycastRequest rayRequest;
```
hultonha, line 314: question: Is it absolutely necessary to add the `I` prefix to this type? If this is the convention for interfaces so be it but this is one of those Hungarian notation style things that may be redundant/not advised these days

Gems/EMotionFX/Code/Source/Integration/System/SystemComponent.cpp
```diff
@@ -564,13 +564,12 @@ namespace EMotionFX
         //////////////////////////////////////////////////////////////////////////
         void SystemComponent::EnableRayRequests()
         {
-            RaycastRequestBus::Handler::BusDisconnect();
-            RaycastRequestBus::Handler::BusConnect();
+            AZ::Interface<IRaycastRequests>::Register(this);
```
hultonha, line 567: question: Does `Register` gracefully handle being called twice without a disconnect? (as there was a disconnect then connect before it just made me wonder)

Gems/EMotionFX/Code/Source/Integration/System/SystemComponent.h
```diff
@@ -98,10 +98,10 @@ namespace EMotionFX
             ////////////////////////////////////////////////////////////////////////
 
             ////////////////////////////////////////////////////////////////////////
-            // RaycastRequestBus
+            // IRaycastRequests
```
hultonha, line 101: nit: `IRaycastRequests overrides ...`

hultonha (approved):
LGTM I just have a few minor bits of feedback/questions

Gems/EMotionFX/Code/EMotionFX/Source/BlendTreeFootIKNode.cpp
spham-amzn: Especially given that `rayRequest` looks like an instantiation of a concrete class in this declaration, not really an interface

spham-amzn (approved):

PR #14462, opened 2023-02-08
Fixed error while loading PhysX materials.

moraaar:

## What does this PR do?

Fixes a bug that causes a crash when while waiting for a physics material asset to load the asset manager dispatches events for other assets. These events can cause `OnAssetReady` of other assets to be called and if those end up trying to create the same physics material then it leads to the crash.

Solved by creating the material first, then block until loaded.

Also renamed the utility function `GetShapesFromAsset` to `CreateShapesFromAsset` since that reflects better what it does.

## How was this PR tested?

Run PhysX unit tests.
Tested using a local level that was crashing while loading due to the issue described above. After this fix the level loads correctly. Also debugged the loading of colliders and rigid bodies to verify it works correctly.

### DIFF
--- a/Code/Framework/AzFramework/AzFramework/Physics/Material/PhysicsMaterialManager.cpp
+++ b/Code/Framework/AzFramework/AzFramework/Physics/Material/PhysicsMaterialManager.cpp
@@ -62,24 +62,27 @@ namespace Physics
             return materialFound;
         }
 
-        // Take a reference so we can mutate it.
-        AZ::Data::Asset<AZ::Data::AssetData> assetLocal = materialAsset;
-        if (!assetLocal.IsReady())
-        {
-            assetLocal.QueueLoad();
-            assetLocal.BlockUntilLoadComplete();
-
-            // Failed to load the asset
-            if (!assetLocal.IsReady())
-            {
-                return nullptr;
-            }
-        }
-
-        auto newMaterial = CreateMaterialInternal(id, assetLocal);
+        auto newMaterial = CreateMaterialInternal(id, materialAsset);
         if (newMaterial)
         {
             m_materials.emplace(newMaterial->GetId(), newMaterial);
+        }
+
+        // Block until the material asset is loaded.
+        // It's important that the loading of physics material assets is not delayed,
+        // otherwise objects would react differently during simulation for a few
+        // frames until the material is loaded, causing unexpected behaviors.
+        AZ::Data::Asset<AZ::Data::AssetData> assetLocal = materialAsset; // Take a reference so we can mutate it for loading.
+        if (!assetLocal.IsReady())
+        {
+            assetLocal.QueueLoad();
+
+            // The call to BlockUntilLoadComplete must happen after the new material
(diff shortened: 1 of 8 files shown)


Code/Framework/AzFramework/AzFramework/Physics/Material/PhysicsMaterialManager.cpp
```diff
@@ -62,24 +62,27 @@ namespace Physics
...
         {
-            m_materials.emplace(newMaterial->GetId(), newMaterial);
+            assetLocal.QueueLoad();
+
+            // The call to BlockUntilLoadComplete must happen after the new material
+            // has been added to m_materials container. This is because BlockUntilLoadComplete
+            // will dispatch events for other assets while waiting, and if other OnAssetRead
+            // code ends up trying find/create the same material it needs to find it in the
+            // container and return it.
+            assetLocal.BlockUntilLoadComplete();
```
moraaar, line 85: An alternative approach to this would be to make all the physx components that use physx materials to delay their initialization until all the physics materials it uses are loaded. But we cannot go for that solution at the moment because all the colliders need to create the physx shapes during activation. The current implementation cannot wait a few frames later to create the shapes, the rigid body components also need the shapes to create the simulated body on "entity activated" event.

If we change how of all physx components are initialized with a more dynamic approach based on events to allow the creation of shapes and simulated bodies at any time, then it would be possible to delay the creation until all the assets are ready.

Not sure if it's a good idea overall to delay the creation of physical entities though, that might lead to objects falling though floors, or trespassing walls.

Code/Framework/AzFramework/AzFramework/Physics/Material/PhysicsMaterialManager.cpp
```diff
@@ -62,24 +62,27 @@ namespace Physics
...
+        // frames until the material is loaded, causing unexpected behaviors.
+        AZ::Data::Asset<AZ::Data::AssetData> assetLocal = materialAsset; // Take a reference so we can mutate it for loading.
+        if (!assetLocal.IsReady())
         {
-            m_materials.emplace(newMaterial->GetId(), newMaterial);
+            assetLocal.QueueLoad();
+
+            // The call to BlockUntilLoadComplete must happen after the new material
+            // has been added to m_materials container. This is because BlockUntilLoadComplete
+            // will dispatch events for other assets while waiting, and if other OnAssetRead
```
AMZN-stankowi, line 82: nitpick typo: OnAssetRead -> OnAssetReady

AMZN-stankowi (approved):
Would it be possible to add an automated test that fails without this change, and passes with it?

moraaar:
> Would it be possible to add an automated test that fails without this change, and passes with it?

I guess I could add an automated test that opens the local level I had, but it's such a rare case that an asset just finished to load at the same time `BlockUntilLoadComplete` is called that it might not repro 100% of the time, so I'm a bit reluctant to add it, I think I was lucky to find this bug.

hultonha (approved):

amzn-mike (approved):

PR #14469, opened 2023-02-08
Update templates when prefab source file was renamed in Asset Browser

### LINKED ISSUE #14015: Asset Browser: Renaming a prefab in the AB does not update instances of that prefab
**Describe the bug**
Renaming a prefab in the AB does not update instances of that prefab.

**Steps to reproduce**
1. Launch the Editor.
3. Instantiate any prefab into the level.
4. Via the Asset Browser, find the prefab that was instantiated, then right-click and rename it.

**Expected behavior**
Instantiated prefab is updated with the new filename.

**Actual behavior**
Instantiated prefabs are not updated with the new filename. They become sort of like an unsaved prefab at that point. If any changes are made to the renamed prefab, those changes will not propagate to the already instantiated prefab. If any changes are made to the instantiated prefab and saved, a new prefab is created with the original name.

**Screenshots/Video**
https://user-images.githubusercontent.com/82226555/211922746-a9d9331c-b40a-45cf-a60d-2d6222fb9488.mp4

**Found in Branch**
development

cbbf5d6f3970f405c7128c212da31c8b2b43920f

michabr:

## What does this PR do?

Handle updating in-memory templates and instances when a user renames or moves a prefab source asset or folder in Asset Browser.

Fixes #14015

## How was this PR tested?

Tested renaming and moving a prefab file and folder from Asset Browser with override UX mode enabled and disabled.


Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/AssetBrowserBus.h
```diff
@@ -377,5 +377,23 @@ namespace AzToolsFramework
         };
         using AssetBrowserFileCreationNotificationBus = AZ::EBus<AssetBrowserFileCreationNotifications>;
 
+        //////////////////////////////////////////////////////////////////////////
+        // File action notifications
+        //////////////////////////////////////////////////////////////////////////
+
+        //! Used for sending and/or recieving notifications regarding source file manipulation through the Asset Browser.
+        class AssetBrowserFileActionNotifications
```
AMZN-stankowi, line 385: I think this is OK for now, but I'd like to have a discussion on what this should look like long term. We'll probably have Asset Processor sending file manipulation events in the future, so I'm thinking we'll end up wanting to move this to a more central location that isn't as specific to the AssetBrowser?

Maybe a new "AssetStateMessages.h" added to:
https://github.com/o3de/o3de/tree/development/Code/Framework/AzToolsFramework/AzToolsFramework/Asset

Something that does not imply the originator of the message, I think listeners to this won't care if the relocation was done via the Asset Browser, Asset Processor, or externally on disk directly (which would probably route through the Asset Processor, which would detect and send the message)

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabSystemComponent.cpp
```diff
@@ -144,6 +146,20 @@ namespace AzToolsFramework
             GarbageCollectTemplates(false);
         }
 
+        void PrefabSystemComponent::OnSourceFilePathNameChanged(const AZStd::string_view fromPathName, const AZStd::string_view toPathName)
+        {
+            // Convert to prefab relative path
+            AZ::IO::Path fromRelativePath = m_prefabLoader.GenerateRelativePath(fromPathName);
```
AMZN-stankowi, line 152: Relative paths with assets make me nervous. What happens if I rename this file to move it between scan folders?

Example 1: I rename from MyProject\MyPrefab.prefab to Gems\MyGem\Assets\MyPrefab.prefab. The relative path will be the same, but the scan folder will change.
Example 2: I rename from MyProject\MyPrefab.prefab to Gems\MyGem\Assets\DifferentPrefabName.prefab. The relative path and scan folder have changed.

AMZN-stankowi (approved):
Would it be possible to get automated tests added to this? I think with the Python in-Editor you could do a lot of this, right? Create a prefab, instantiate it in the current level, rename the initial prefab via the Asset Browser, and verify things have updated properly?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabSystemComponent.cpp
```diff
@@ -144,6 +146,20 @@ namespace AzToolsFramework
...
+
+            // Update loaded template with new pathname
+            if (Prefab::TemplateId templateId = GetTemplateIdFromFilePath(fromRelativePath); templateId != InvalidTemplateId)
```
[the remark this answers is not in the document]
michabr: We are just supporting file renames and moves within Asset Browser.

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/AssetBrowserBus.h
michabr: Yeah, this isn't necessarily an Asset Browser specific event. If/when Asset Processor supports file moves, it seems like it would make sense to add the event to the AssetSystemNotificationBus and refine SourceFileChanged.

For now, the AssetBrowserBus.h file already contained an existing AssetBrowserFileCreationNotificationBus which is similar, so I just followed that pattern.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabSystemComponent.cpp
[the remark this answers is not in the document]
amzn-mike: I know, but the API sending this ebus event has a flag `isFolder`, so while I haven't tested it (I know APB supports it at least) I assume a whole folder containing many prefabs can be moved/renamed.  If that's the case, it seems like the event would broadcast that the folder was renamed, then `GetTemplateIdFromFilePath` would likely return nothing (since the folder isn't a prefab) and end up exiting out.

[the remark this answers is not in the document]
michabr: Oh I actually didn't realize users can rename/move folders in AB. I'll update to handle this case. Thanks!

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabSystemComponent.cpp
forkercat: Looks like prefab system does not identify prefabs by scan folders. If a prefab file is referencing `MyPrefab.prefab` in Source field, I suppose it currently depends on this call to determine which one to look for.

```cpp
// Inside PrefabLoader::GenerateRelativePath()...
// The asset system allows for paths to be relative to multiple root folders, using a priority system.
// This request will make the input path relative to the most appropriate, highest-priority root folder.
AzToolsFramework::AssetSystemRequestBus::BroadcastResult(
    pathFound, &AzToolsFramework::AssetSystemRequestBus::Events::GenerateRelativeSourcePath, path.Native(), 
    relativePath, rootFolder);
```

forkercat: I also noticed that the editor would crash when I tried:
- first creating a `MyPrefab` prefab at `MyProject\MyPrefab.prefab` and
- then creating another prefab at `Gems\Camera\Assets\MyPrefab.prefab`

It immediately crashes during the second creation. Inside `CreatePrefabInMemory()`, new template id and DOM seem to be invalid at these lines ([PrefabPublicHandler.cpp#L243-L244](https://github.com/o3de/o3de/blob/development/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp#L243-L244)).

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabDomUtils.cpp
```diff
@@ -505,13 +505,19 @@ namespace AzToolsFramework
                     {
                         continue;
                     }
+                    // Source
+                    else if (patchPath == PathMatchingSource)
+                    {
+                        // Ignore nested instance file path changes resulting from source asset renaming
+                        continue;
```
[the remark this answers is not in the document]
michabr: This just prevents the warning that catches unwanted patch types.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabSystemComponent.cpp
michabr: Yeah, we'll need to address how we store prefab asset paths in a separate task.

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Views/AssetBrowserTreeView.cpp
```diff
@@ -979,7 +1003,8 @@ namespace AzToolsFramework
                     }
                     if (isFolder)
                     {
-                        AZ::IO::SystemFile::DeleteDir(fromPath.substr(0, fromPath.size() - 2).data());
+                        AZStd::string fromPathNoWildcard(fromPath.substr(0, fromPath.size() - 2));
+                        AZ::IO::SystemFile::DeleteDir(fromPathNoWildcard.c_str());
                     }
```
michabr, line 1008: @AMZN-stankowi FYI I added this fix because the folder wasn't getting deleted.

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Views/AssetBrowserTreeView.cpp
```diff
@@ -954,14 +966,26 @@ namespace AzToolsFramework
                     AssetChangeReportResponse moveResponse;
                     if (SendRequest(moveRequest, moveResponse))
                     {
-                        if (!response.m_lines.empty())
+                        // Send notification for listeners to update any in-memory states
+                        if (isFolder)
+                        {
+                            AZStd::string fromPathNoWildcard(fromPath.substr(0, fromPath.size() - 2));
+                            AZStd::string toPathNoWildcard(toPath.substr(0, toPath.size() - 2));
```
amzn-mike, line 973: Can you add a comment on what the -2 is for?

amzn-mike (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Views/AssetBrowserTreeView.cpp
AMZN-stankowi: Looking at SystemFile_WinAPI.cpp, it doesn't print out an error or anything if this call fails. Would it make sense to print out a message?

```
if(!AZ::IO::SystemFile::DeleteDir(fromPathNoWildcard.c_str()))
{
    AZ_Printf("AssetBrowser", false, "Failed to delete empty directory %s.", fromPathNoWildcard.c_str());
}
```

No problem if not, it doesn't look like many other places that call AZ::IO::SystemFile::DeleteDir are checking, either.

forkercat (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Views/AssetBrowserTreeView.cpp
michabr: Makes sense. Added as a warning.

PR #14483, opened 2023-02-08
Fix nested prefab bug in detach prefab operation

forkercat:
## What does this PR do?

This PR fixes a bug that was introduced by the detach-prefab improvement PR https://github.com/o3de/o3de/pull/14282.

**Bug:** In detached prefab, nested prefab that is not under top level entity would be reparented beyond level. Nested entity is fine.

**Cause:** The issue happens in `CreateLink` for nested prefab. It creates a reparent patch pointing to a parent entity that is in a state before detach. Moving `DetachEntities` call before `DetachNestedInstances` call can resolve the bug.

https://user-images.githubusercontent.com/11157226/217672566-e32bf27c-096c-454a-8d4d-fb6397ecb9fc.mp4

## How was this PR tested?

- [x] Verified the issue is gone after the fix and existing tests are not broken
- [x] Updated one simple test to check entity hierarchy after detach
- [x] Added 2 unit tests to validate hierarchy and entity order under a top level entity after detach. Verified they would fail before the fix
```
[==========] 7 tests from 1 test case ran. (1508 ms total)
[  PASSED  ] 5 tests.
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] PrefabDetachPrefabTests.DetachPrefabWithNestedPrefabUnderTopLevelEntitySucceeds
[  FAILED  ] PrefabDetachPrefabTests.DetachPrefabValidatesTopLevelChildEntityOrder
```
- [x] Manually tested undo/redo/save DetachPrefab operation in a complex hierarchy:

### DIFF
--- a/Code/Framework/AzToolsFramework/Tests/Prefab/PrefabDetachPrefabTests.cpp
+++ b/Code/Framework/AzToolsFramework/Tests/Prefab/PrefabDetachPrefabTests.cpp
@@ -44,7 +44,7 @@ namespace UnitTest
         ValidateNestedInstanceNotUnderInstance(GetRootContainerEntityId(), carInstanceAlias);
 
         InstanceOptionalReference levelInstance = m_instanceEntityMapperInterface->FindOwningInstance(GetRootContainerEntityId());
-        EXPECT_TRUE(levelInstance.has_value());
+        ASSERT_TRUE(levelInstance.has_value());
 
         // Validate there are three entities in the level prefab instance.
         EXPECT_EQ(levelInstance->get().GetEntityAliasCount(), 3);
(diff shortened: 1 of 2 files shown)


Code/Framework/AzToolsFramework/Tests/Prefab/PrefabDetachPrefabTests.cpp
```diff
@@ -204,6 +216,95 @@ namespace UnitTest
...
+        // Validate there is wheels entity in the level prefab instance.
+        EntityAlias wheelsEntityAlias = FindEntityAliasInInstance(GetRootContainerEntityId(), wheelsEntityName);
+        ValidateEntityUnderInstance(GetRootContainerEntityId(), wheelsEntityAlias, wheelsEntityName);
+
+        // Validate there is wheel instance in the level prefab instance.
+        InstanceAlias wheelInstanceAlias = FindNestedInstanceAliasInInstance(GetRootContainerEntityId(), wheelPrefabName);
+        ValidateNestedInstanceUnderInstance(GetRootContainerEntityId(), wheelInstanceAlias);
+
+        InstanceOptionalReference levelInstance = m_instanceEntityMapperInterface->FindOwningInstance(GetRootContainerEntityId());
+        EXPECT_TRUE(levelInstance.has_value());
```
michabr, line 269: ASSERT_TRUE?

michabr (approved):

srikappa-amzn (approved):

PR #14484, opened 2023-02-08
Provide a CharacterNotifications bus to drive association of active PhysX character and Network Character Controllers

mbalfour-amzn:
## What does this PR do?

The recent change to Static Rigid Bodies changed the timing of the Character Controller initialization. It used to occur during the component's Activate(), which meant that other components on the entity could be guaranteed that they could fetch a pointer to the component during their Activate() calls, as long as they had the service dependencies set up correctly.

However, the Character Controller now activates during OnEntityActivated(), which means that other components can no longer rely on the service dependencies to know when the Character Controller is available.

To fix this, the Character Controller now has a new notification bus that it uses to signal when the character is activated and deactivated. Dependent components can listen for these notifications to know when it is safe to communicate with the Character Controller and when it isn't.

## How was this PR tested?

Ran the Multiplayer Sample and verified that the character can once again move around in networked conditions.
Run physx gem unit tests.
Run multiplayer gem unit tests.

### DIFF
--- a/Code/Framework/AzFramework/AzFramework/Physics/CharacterBus.h
+++ b/Code/Framework/AzFramework/AzFramework/Physics/CharacterBus.h
@@ -103,4 +103,22 @@ namespace Physics
     };
 
     using CharacterRequestBus = AZ::EBus<CharacterRequests>;
-} // namespace Physics
+
+    /// Messages sent by character controllers.
+    class CharacterNotifications : public AZ::ComponentBus
+    {
+    public:
+        virtual ~CharacterNotifications() = default;
+
+        virtual void OnCharacterActivated()
+        {
+        }
+
+        virtual void OnCharacterDeactivated()
+        {
+        }
+    };
+
+    using CharacterNotificationBus = AZ::EBus<CharacterNotifications>;
+
+    } // namespace Physics
--- a/Code/Framework/AzFramework/AzFramework/Physics/CharacterBus.h
+++ b/Code/Framework/AzFramework/AzFramework/Physics/CharacterBus.h
@@ -9,6 +9,7 @@
 #pragma once
 
 #include <AzCore/Component/ComponentBus.h>
+#include <AzCore/Component/Entity.h>
 
 namespace AZ
 {
@@ -107,20 +108,67 @@ namespace Physics
     /// Messages sent by character controllers.
     class CharacterNotifications : public AZ::ComponentBus
     {
+    private:
+        template<class Bus>
+        struct CharacterNotificationsConnectionPolicy : public AZ::EBusConnectionPolicy<Bus>
+        {
+            static void Connect(
+                typename Bus::BusPtr& busPtr,
+                typename Bus::Context& context,
+                typename Bus::HandlerNode& handler,
+                typename Bus::Context::ConnectLockGuard& connectLock,
+                const typename Bus::BusIdType& id = 0)
+            {
+                AZ::EBusConnectionPolicy<Bus>::Connect(busPtr, context, handler, connectLock, id);
+
+                AZ::Entity* entity = nullptr;
+                AZ::ComponentApplicationBus::BroadcastResult(entity, &AZ::ComponentApplicationBus::Events::FindEntity, id);
+                if (entity)
+                {
+                    // Only immediately dispatch if the entity is already active, otherwise when
+                    // entity will get activated it will make the notifications itself.
+                    const AZ::Entity::State entityState = entity->GetState();
+                    if (entityState == AZ::Entity::State::Active)
+                    {
+                        // Only immediately dispatch if the entity is a CharacterRequests' handler.
+                        CharacterRequestBus::EnumerateHandlersId(
+                            id,
+                            [&handler, id](const CharacterRequests* characterHandler)
+                            {
+                                if (characterHandler->IsPresent())
+                                {
+                                    handler->OnCharacterActivated(id);
+                                }
+                                else
+                                {
+                                    handler->OnCharacterDeactivated(id);
+                                }
+                                return true;
+                            });
+                    }
+                }
+            }
+        };
+
     public:
+        /// With this connection policy, CharacterNotifications::OnCharacterActivated and
+        /// CharacterNotifications::OnCharacterDeactivated events will be immediately
+        /// dispatched when a handler connects to the bus.
+        template<class Bus>
+        using ConnectionPolicy = CharacterNotificationsConnectionPolicy<Bus>;
+
         virtual ~CharacterNotifications() = default;
 
-        /// Notifies that the character controller has activated on this entity.
-        /// Other components on the entity that depend on the character controller can't simply rely on the component service
-        /// dependencies because the character controller isn't necessarily activated at component activation.
-        virtual void OnCharacterActivated()
+        /// Notifies that the character controller has activated on an entity.
+        /// Other components that depend on the character controller can't simply rely on the component service
+        /// dependencies because the character controller is activated at OnEntityActivated, after the
+        /// components' activation.
+        virtual void OnCharacterActivated([[maybe_unused]] const AZ::EntityId& entityId)
         {
         }
 
-        // Notifies that the character controller has deactivated on this entity.
-        /// Other components on the entity that depend on the character controller can't simply rely on the component service
-        /// dependencies because the character controller isn't necessarily deactivated at component deactivation.
-        virtual void OnCharacterDeactivated()
+        /// Notifies that the character controller has deactivated on an entity.
+        virtual void OnCharacterDeactivated([[maybe_unused]] const AZ::EntityId& entityId)
         {
         }
     };
--- a/Code/Framework/AzFramework/AzFramework/Physics/CharacterBus.h
+++ b/Code/Framework/AzFramework/AzFramework/Physics/CharacterBus.h
@@ -110,10 +110,16 @@ namespace Physics
     public:
         virtual ~CharacterNotifications() = default;
 
+        /// Notifies that the character controller has activated on this entity.
+        /// Other components on the entity that depend on the character controller can't simply rely on the component service
+        /// dependencies because the character controller isn't necessarily activated at component activation.
         virtual void OnCharacterActivated()
         {
         }
 
+        // Notifies that the character controller has deactivated on this entity.
+        /// Other components on the entity that depend on the character controller can't simply rely on the component service
+        /// dependencies because the character controller isn't necessarily deactivated at component deactivation.
         virtual void OnCharacterDeactivated()
         {
         }
--- a/Gems/Multiplayer/Code/Include/Multiplayer/Components/NetworkCharacterComponent.h
+++ b/Gems/Multiplayer/Code/Include/Multiplayer/Components/NetworkCharacterComponent.h
@@ -10,6 +10,7 @@
 
 #include <Source/AutoGen/NetworkCharacterComponent.AutoComponent.h>
 #include <Multiplayer/Components/NetBindComponent.h>
+#include <AzFramework/Physics/CharacterBus.h>
 
 namespace Physics
 {
@@ -37,6 +38,7 @@ namespace Multiplayer
     //! Provides multiplayer support for game-play player characters.
     class NetworkCharacterComponent
         : public NetworkCharacterComponentBase
+        , protected Physics::CharacterNotificationBus::Handler
     {
         friend class NetworkCharacterComponentController;
 
@@ -52,6 +54,9 @@ namespace Multiplayer
         void OnInit() override {}
         void OnActivate(Multiplayer::EntityIsMigrating entityIsMigrating) override;
         void OnDeactivate(Multiplayer::EntityIsMigrating entityIsMigrating) override;
+
+    protected:
+        void OnCharacterActivated() override;
 
     private:
         void OnTranslationChangedEvent(const AZ::Vector3& translation);
--- a/Gems/Multiplayer/Code/Include/Multiplayer/Components/NetworkCharacterComponent.h
+++ b/Gems/Multiplayer/Code/Include/Multiplayer/Components/NetworkCharacterComponent.h
@@ -56,7 +56,7 @@ namespace Multiplayer
         void OnDeactivate(Multiplayer::EntityIsMigrating entityIsMigrating) override;
 
     protected:
-        void OnCharacterActivated() override;
+        void OnCharacterActivated(const AZ::EntityId& entityId) override;
 
     private:
         void OnTranslationChangedEvent(const AZ::Vector3& translation);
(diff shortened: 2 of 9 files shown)


Gems/Multiplayer/Code/Include/Multiplayer/Components/NetworkCharacterComponent.h
```diff
@@ -53,6 +55,9 @@ namespace Multiplayer
         void OnActivate(Multiplayer::EntityIsMigrating entityIsMigrating) override;
         void OnDeactivate(Multiplayer::EntityIsMigrating entityIsMigrating) override;
 
+    protected:
+        void OnCharacterActivated() override;
```
lmbr-pip, line 59: Should we have an OnCharacterDeactiveated() call?

lmbr-pip, line 59: Maybe on `OnPhysicsCharacterActivated()` to make it intent clearer?

Gems/Multiplayer/Code/Source/Components/NetworkCharacterComponent.cpp
```diff
@@ -107,7 +107,13 @@ namespace Multiplayer
     {
     }
 
+
     void NetworkCharacterComponent::OnActivate([[maybe_unused]] Multiplayer::EntityIsMigrating entityIsMigrating)
+    {
+        Physics::CharacterNotificationBus::Handler::BusConnect(GetEntityId());
+    }
+
+    void NetworkCharacterComponent::OnCharacterActivated()
```
lmbr-pip, line 116:
```
void NetworkCharacterComponent::OnPhysicsCharacterDeactivated()
{
    m_physicsCharacter = nullptr;
}
```

lmbr-pip (commented):
LG.

Thanks for adding this.

Gems/Multiplayer/Code/Source/Components/NetworkCharacterComponent.cpp
```diff
@@ -107,10 +107,22 @@ namespace Multiplayer
...
-        Physics::CharacterRequests* characterRequests = Physics::CharacterRequestBus::FindFirstHandler(GetEntityId());
-        m_physicsCharacter = (characterRequests != nullptr) ? characterRequests->GetCharacter() : nullptr;
+        // During activation the character controller is not created yet.
+        // Connect to CharacterNotificationBus to listen when it's activated after creation.
+        Physics::CharacterNotificationBus::Handler::BusConnect(GetEntityId());
+    }
+
+    void NetworkCharacterComponent::OnCharacterActivated(const AZ::EntityId& entityId)
+    {
+        Physics::CharacterNotificationBus::Handler::BusDisconnect();
```
moraaar, line 120: Notice how `NetworkCharacterComponent` is only using `CharacterNotificationBus` for "creation" only since it disconnects right away.

I did this to keep the exact same logic as before since it wasn't handling the character being enabled/disabled. But having into consideration that enabling/disabling a character recreates the body entirely internally my suggestion would be to keep listening to `OnCharacterActivated` and `OnCharacterDeactivated` to refresh everything inside `NetworkCharacterComponent`. The only reason I didn't doit is because I didn't know the consequences of such change and so I kept the same logic as before.

This is now exactly the same as in NetworkRigidBodyComponent and NetworkRigidBodyControllerComponent, so they are aligned now, but if you guys prefer to go ahead and change it please change it in all of them.

Gems/Multiplayer/Code/Source/Components/NetworkCharacterComponent.cpp
moraaar: I answered this in a separate code comment.

moraaar (commented):
I'll leave other folks to approve since I contributed, but it's got my **virtual approval** :D

moraaar:
> Not a fan of the spaghetti of messages being thrown around when something activates just to get all the timing down right. Any chance we could have the physics folks look at their change again and make sure the ordering change was actually necessary? It's a breaking change and we're not sure what else may have broken..

In this case the change was necessary, I talked today again with the other physics folks and we still think this is the appropriate change.

It is indeed an impactful change and we spent 1 week testing it with 4 people and PRs went sent to multiple sigs. It's difficult to cover everything, we tried to cover as many repos as possible, but unfortunately network character component fell through the cracks and I am very sorry that affected other members of the team. I'll be paying much more attention to it from now on.

Please do not hesitate to contact me with any questions about colliders or rigid bodies.

Code/Framework/AzFramework/AzFramework/Physics/CharacterBus.h
```diff
@@ -103,4 +104,75 @@ namespace Physics
     };
 
     using CharacterRequestBus = AZ::EBus<CharacterRequests>;
-} // namespace Physics
+
+    /// Messages sent by character controllers.
```
hultonha, line 108: uber nit: I believe `//!` style is generally preferred now if possible

Code/Framework/AzFramework/AzFramework/Physics/CharacterBus.h
```diff
@@ -103,4 +104,75 @@ namespace Physics
...
+    class CharacterNotifications : public AZ::ComponentBus
+    {
+    private:
+        template<class Bus>
+        struct CharacterNotificationsConnectionPolicy : public AZ::EBusConnectionPolicy<Bus>
```
hultonha, line 113: question: Based on the recent EBus changes by @lawsonamzn I wonder if it's preemptively worth moving this to the .cp and using the new `extern` approach? It might be worth waiting to profile the compile time I just wanted to sanity check in case.

Code/Framework/AzFramework/AzFramework/Physics/CharacterBus.h
```diff
@@ -103,4 +104,75 @@ namespace Physics
...
+                const typename Bus::BusIdType& id = 0)
+            {
+                AZ::EBusConnectionPolicy<Bus>::Connect(busPtr, context, handler, connectLock, id);
+
+                AZ::Entity* entity = nullptr;
+                AZ::ComponentApplicationBus::BroadcastResult(entity, &AZ::ComponentApplicationBus::Events::FindEntity, id);
+                if (entity)
+                {
+                    // Only immediately dispatch if the entity is already active, otherwise when
+                    // entity will get activated it will make the notifications itself.
```
hultonha, line 129: uber nit: make the notifications -> send the notifications

Code/Framework/AzFramework/AzFramework/Physics/CharacterBus.h
```diff
@@ -103,4 +104,75 @@ namespace Physics
...
+                            });
+                    }
+                }
+            }
+        };
+
+    public:
+        /// With this connection policy, CharacterNotifications::OnCharacterActivated and
+        /// CharacterNotifications::OnCharacterDeactivated events will be immediately
+        /// dispatched when a handler connects to the bus.
```
hultonha, line 156: nit (repeated): Prefer `//!` (the reason for this was for accessibility because by default in Visual Studio in the dark them the color selected for these 'XML' comments is really hard to read, there is no such problem with `//!` however so all new code should prefer it).

Gems/PhysX/Code/Source/PhysXCharacters/Components/CharacterControllerComponent.cpp
```diff
@@ -586,30 +589,39 @@ namespace PhysX
                 sceneInterface->RegisterSceneSimulationStartHandler(m_attachedSceneHandle, m_sceneSimulationStartHandler);
             }
         }
+
+        AZ::TransformNotificationBus::Handler::BusConnect(GetEntityId());
+        Physics::CharacterRequestBus::Handler::BusConnect(GetEntityId());
+        Physics::CollisionFilteringRequestBus::Handler::BusConnect(GetEntityId());
+        AzPhysics::SimulatedBodyComponentRequestsBus::Handler::BusConnect(GetEntityId());
```
hultonha, line 596: nit: Could extract `EntityId` to a local to stop having to call this function repeatedly

```
const AZ::EntityId entityId = GetEntityId();
AZ::TransformNotificationBus::Handler::BusConnect(entityId);
... // etc
```

hultonha (approved):
Change looks good to me - I realize this change is needed ASAP but I think it would be great to take a breath after this change and think about how we could create a test to encapsulate this behavior (this could serve as an easy to find example as well to see how to handle this particular case). We just need some way to now 'lock' in this behavior so something breaks if we change something in RigidBody in future (this can be in a separate PR but it would be great to do and see if we can have a simplified example of how to handle the `OnEntityActivated` behavior)

Gems/Multiplayer/Code/Source/Components/NetworkCharacterComponent.cpp
moraaar: @lmbr-pip @kberg-amzn would be doing `m_physicsCharacter = nullptr;` in a `OnCharacterDeactivated` function enough to revert all that `OnCharacterActivated` is doing or would it need to do something else. Is it harmless that if the character is activate again all the code of `OnCharacterActivated` is executed again?

mbalfour-amzn: @lmbr-pip @kberg-amzn I'm not very familiar with this code, but I'm from looking at the OnCharacterActivated(), I'm thinking we need the following?
- Add the OnCharacterDeactivated() to set m_physicsCharacter to null.
- Move these lines to Activate() instead of OnCharacterActivated(), so that way in case OnCharacterActivated/Deactivated ever gets called more than once, things continue to work ok. These seem like they shouldn't get called more than once, and that they're ok to call even before the character exists?
```
        GetNetBindComponent()->AddEntitySyncRewindEventHandler(m_syncRewindHandler);
```
```
        if (!HasController())
        {
            GetNetworkTransformComponent()->TranslationAddEvent(m_translationEventHandler);
        }
```

Code/Framework/AzFramework/AzFramework/Physics/CharacterBus.h
```diff
@@ -103,4 +104,75 @@ namespace Physics
...
+        template<class Bus>
+        using ConnectionPolicy = CharacterNotificationsConnectionPolicy<Bus>;
+
+        virtual ~CharacterNotifications() = default;
+
+        /// Notifies that the character controller has activated on an entity.
+        /// Other components that depend on the character controller can't simply rely on the component service
+        /// dependencies because the character controller is activated at OnEntityActivated, after the
+        /// components' activation.
+        virtual void OnCharacterActivated([[maybe_unused]] const AZ::EntityId& entityId)
```
greerdv, line 166: I'm wondering if there's any value in having the entity Id as a parameter here. Wouldn't you already need to know the entity Id in order to have connected to the bus?

Code/Framework/AzFramework/AzFramework/Physics/CharacterBus.h
moraaar: This entire file is using /// so I'd rather keep it consistent and not do unnecessary changes in other places now.

Code/Framework/AzFramework/AzFramework/Physics/CharacterBus.h
moraaar: I'll change that. Thanks.

Gems/Multiplayer/Code/Source/Components/NetworkCharacterComponent.cpp
moraaar: I would not move the code to `Activate()` because those register to call the function `NetworkCharacterComponent::OnSyncRewind` which needs `m_physicsCharacter` anyway, so I think it's more correct to do it all `OnCharacterActivated`.

Probably `OnCharacterDeactivated` would need to register those events too. Right?

greerdv (approved):

Code/Framework/AzFramework/AzFramework/Physics/CharacterBus.h
moraaar: I'll wait if there is another change... I already kicked of AR and probably not worth it to do it again for this. But I'll keep it in mind if I have to push again.

Code/Framework/AzFramework/AzFramework/Physics/CharacterBus.h
mbalfour-amzn: I believe @lawsonamzn specifically said _not_ to preemptively use the extern approach, that it should only be used for highly-used EBuses? This one seems like it would be more sparsely used IMO...

Code/Framework/AzFramework/AzFramework/Physics/CharacterBus.h
moraaar: Good question. This is useful when it's required to know if more that 1 character is ready, so it can be easily used with a Multihandler and therefore it's very useful to know which one is calling.

Gems/Multiplayer/Code/Source/Components/NetworkCharacterComponent.cpp
moraaar: By the way, I'd be inclined to leave this for a subsequent PR.

Code/Framework/AzFramework/AzFramework/Physics/CharacterBus.h
mbalfour-amzn: Yeah, it should be "gettable" inside the OnCharacterActivated() via "CharacterNotificationBus::GetCurrentBusId()". That's also admittedly a little bit less obvious and readable. I'm personally pretty ambivalent on this one, but that's why I didn't initially include the entityId in the first version of this event. :)

Gems/Multiplayer/Code/Include/Multiplayer/Components/NetworkCharacterComponent.h
moraaar: I answered this in a separate code comment below.

Code/Framework/AzFramework/AzFramework/Physics/CharacterBus.h
greerdv: Yeah, I don't think it's a big deal either way, I was just curious.

Gems/Multiplayer/Code/Include/Multiplayer/Components/NetworkCharacterComponent.h
mbalfour-amzn: I personally like the suggestion, but at least for now this naming is in line with the other physics bus names, so I'd prefer not to deviate. I'll leave it to the physics folks to decide if they want to make it more verbose in a follow-up PR.

Gems/Multiplayer/Code/Source/Components/NetworkCharacterComponent.cpp
mbalfour-amzn: i discussed with @AMZN-Olex and he's in favor of moving those chunks to Activate(), and adding the Disconnect() calls to Deactivate(), so that the network handling is kept more obviously separated from the physics character handling. i'll put that up for review in a follow-up PR.

mbalfour-amzn:
This passed AR and development is green, so I'll submit as-is. I'll put up a follow-up PR to clean up the Activate() / Deactivate() logic in NetworkCharacterComponent.

PR #14486, opened 2023-02-09
Change DX12 staging memory allocator settings so it won't need allocate some memory pages every frame.

VickyAtAZ:
## What does this PR do?
Change DX12 staging memory allocator settings so it won't need allocate some memory pages every frame.
It saves around 1ms with https://github.com/michalpelka/test-atom-multicam when running game mode in Editor

## How was this PR tested?
Editor with  https://github.com/michalpelka/test-atom-multicam project.

### DIFF
--- a/Gems/Atom/RHI/DX12/Code/Source/RHI/StagingMemoryAllocator.cpp
+++ b/Gems/Atom/RHI/DX12/Code/Source/RHI/StagingMemoryAllocator.cpp
@@ -27,6 +27,7 @@ namespace AZ
                 poolDesc.m_pageSizeInBytes = descriptor.m_mediumPageSizeInBytes;
                 poolDesc.m_collectLatency = descriptor.m_collectLatency;
                 poolDesc.m_getHeapMemoryUsageFunction = [this]() { return &m_memoryUsage; };
+                poolDesc.m_recycleOnCollect = true;
                 m_mediumPageAllocator.Init(poolDesc);
             }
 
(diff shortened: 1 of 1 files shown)


Gems/Atom/RHI/DX12/Code/Source/RHI/StagingMemoryAllocator.cpp
```diff
@@ -27,6 +27,7 @@ namespace AZ
                 poolDesc.m_pageSizeInBytes = descriptor.m_mediumPageSizeInBytes;
                 poolDesc.m_collectLatency = descriptor.m_collectLatency;
                 poolDesc.m_getHeapMemoryUsageFunction = [this]() { return &m_memoryUsage; };
+                poolDesc.m_recycleOnCollect = true;
```
moudgils, line 30: Dont we need this for Vulkan and Metal too?

moudgils (approved):

Gems/Atom/RHI/DX12/Code/Source/RHI/StagingMemoryAllocator.cpp
VickyAtAZ: Vulkan and Metal is using BufferPool which is hard-coded for all bufferPool instances. 
Although I didn't find similar allocations when running with Vulkan.

moudgils: For Vk/Metal we use m_stagingBufferPool and all BufferPools are getting hardcoded to 
`allocatorDescriptor.m_recycleOnCollect = false;`  within Vk/Metal::BufferPool::InitInternal. 

We could modify RHI::BufferPoolDescriptor to add m_recycleOnCollect which can be used to set that variable in BufferPool::InitInternal instead of always hardcoding it to false.

PR #14493, opened 2023-02-09
fix collider debug draw and dynamic rigid body editor world body not …

greerdv:
…updating when collider is modified

## What does this PR do?
Fixes a couple of issues:
- collider debug draw was not getting updated when colliders were modified
- if the entity had a dynamic rigid body, the editor world body was not getting updated when colliders were modified

## How was this PR tested?
Manual testing in the editor.

### DIFF
--- a/Gems/PhysX/Code/Source/EditorColliderComponent.cpp
+++ b/Gems/PhysX/Code/Source/EditorColliderComponent.cpp
@@ -639,9 +639,11 @@ namespace PhysX
     void EditorColliderComponent::CreateStaticEditorCollider()
     {
         m_cachedAabbDirty = true;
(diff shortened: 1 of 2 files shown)


greerdv:
Had a quick chat with @moraaar and I'm going to make some updates here, so maybe hold off on reviewing till I push those. The gist is that it's a bit weird for the collider changed notification to be a side effect of creating the editor rigid body, so the plan is to add a new function UpdateCollider() which will call CreateStaticEditorCollider and send the OnColliderChanged event. Then all the places which call CreateStaticEditorCollider will be changed to call UpdateCollider instead.

moraaar (approved):
LGTM

hultonha (approved):

PR #14494, opened 2023-02-09
Enable TIAF gating in AR.

jonawals:
Signed-off-by: John <jonawals@amazon.co.uk>

## What does this PR do?

This PR enables gating for failed tests run by TIAF in AR.

## How was this PR tested?

Tested locally. Will be tested in AR.


amzn-ahmadkrm (approved):

hultonha (approved):

hultonha (commented):
Change looks good but I just want to check is this tested @jonawals? Can you do a quick sanity check like @moraaar did last week to prove a failing test in TIAF will report the right error in Jenkins? Would be great to see before/after to verify. Once we have that I can approve and we can get this merged early tomorrow so we can monitor it more closely. Thanks!

moraaar (approved):
Change looks good. I agree with Tom's comment of testing a failing test case just to be sure it'll be catch.

PR #14495, opened 2023-02-09
Using default editor to open log files.

Jackie9527:

## What does this PR do?

1. Set LuaIDE as Scripts Editor.
image
2. Open to see log files of Editor.

3. LuaIDE starts and gives a misleading prompt.
image

## How was this PR tested?

Follow the same repo steps, and see.
image

### DIFF
--- a/Code/Editor/CryEdit.cpp
+++ b/Code/Editor/CryEdit.cpp
@@ -2652,7 +2652,9 @@ void CCryEditApp::OnEditLevelData()
 //////////////////////////////////////////////////////////////////////////
 void CCryEditApp::OnFileEditLogFile()
 {
-    CFileUtil::EditTextFile(CLogFile::GetLogFileName(), 0, IFileUtil::FILE_TYPE_SCRIPT);
+    QString file = CLogFile::GetLogFileName();
+    QString fullPathName = Path::GamePathToFullPath(file);
+    QDesktopServices::openUrl(QUrl::fromLocalFile(fullPathName));
 }
 
 #ifdef ENABLE_SLICE_EDITOR


Code/Editor/Util/FileUtil.cpp
```diff
@@ -305,6 +305,12 @@ void CFileUtil::EditTextFile(const char* txtFile, int line, IFileUtil::ETextFile
...
 
+    if (fileType == IFileUtil::FILE_TYPE_SYSTEM)
+    {
+        QDesktopServices::openUrl(QUrl::fromLocalFile(fullPathName));
+        return;
+    }
+
     Common::EditFileType editFileType = Common::EditFileType::FILE_TYPE_SCRIPT;
 
     switch (fileType)
```
gadams3, line 316: nit -this if could be merged with the switch statement below but it is doing slightly different behavior so I see why it's separated

Code/Editor/Include/IFileUtil.h
```diff
@@ -80,6 +80,7 @@ struct IFileUtil
         FILE_TYPE_SCRIPT,
         FILE_TYPE_SHADER,
         FILE_TYPE_BSPACE,   // added back in 3.8 integration, may not end up needing this.
+        FILE_TYPE_SYSTEM,
```
gadams3, line 83: nit - from the code it's clear that this is going to use the OS file type associations to open the file but having a comment here to explain it would be nice

gadams3 (approved):

Code/Editor/Util/FileUtil.cpp
```diff
@@ -305,6 +305,12 @@ void CFileUtil::EditTextFile(const char* txtFile, int line, IFileUtil::ETextFile
     ExtractFile(fullPathName);
     QString cmd = Platform::MakePlatformFileEditString(fullPathName, line);
 
+    if (fileType == IFileUtil::FILE_TYPE_SYSTEM)
+    {
+        QDesktopServices::openUrl(QUrl::fromLocalFile(fullPathName));
+        return;
+    }
+
```
cgalvan, line 313: This API is for using the pre-defined editors that are specified by the user. The `FILE_TYPE_SCRIPT` type is typically meant to be some kind of text editor so users can modify scripts (e.g. python, lua). If the user specified the LUAIde as their text editor, they wouldn't be able to open python files in there either, so I see that more of a misconfiguration.

I could potentially see this change being reduced to just calling `QDesktopServices::openUrl(QUrl::fromLocalFile(CLogFile::GetLogFileName()));` in `CCryEditApp::OnFileEditLogFile()`, but I don't think the file edit API should be modified for this.

cgalvan (changes requested):

Code/Editor/Util/FileUtil.cpp
Jackie9527: done.
Please review the code.

Code/Editor/Include/IFileUtil.h
Jackie9527: Got a simple way as @cgalvan said.

Code/Editor/Util/FileUtil.cpp
Jackie9527: Got a simple way as @cgalvan said.

Code/Editor/Util/FileUtil.cpp
cgalvan: Thank you!

cgalvan (approved):

PR #14498, opened 2023-02-09
Fix memory leak from TrajectoryTracePath object.

Jackie9527:

## What does this PR do?

Fix memory leak from `TrajectoryTracePath` object.
Objects of `TrajectoryTracePath` are created but no code to delete them.

## How was this PR tested?

Using `AZStd::unique_ptr` to track the memory allocated with `TrajectoryTracePath`.

emotionfx-memory-leak-1
emotionfx-memory-leak-2

### DIFF
--- a/Gems/AtomLyIntegration/EMotionFXAtom/Code/Source/AtomActorDebugDraw.cpp
+++ b/Gems/AtomLyIntegration/EMotionFXAtom/Code/Source/AtomActorDebugDraw.cpp
@@ -961,7 +961,7 @@ namespace AZ::Render
     void AtomActorDebugDraw::UpdateActorInstance(EMotionFX::ActorInstance* actorInstance, float deltaTime)
     {
         // Find the corresponding trajectory trace path for the given actor instance
-        TrajectoryTracePath* trajectoryPath = FindTrajectoryPath(actorInstance);
+        auto trajectoryPath = FindTrajectoryPath(actorInstance);
         if (!trajectoryPath)
         {
             return;
@@ -1084,22 +1084,21 @@ namespace AZ::Render
     // Find the trajectory path for a given actor instance
     AtomActorDebugDraw::TrajectoryTracePath* AtomActorDebugDraw::FindTrajectoryPath(const EMotionFX::ActorInstance* actorInstance)
     {
-        for (TrajectoryTracePath* trajectoryPath : m_trajectoryTracePaths)
+        for (const auto& trajectoryPath : m_trajectoryTracePaths)
         {
             if (trajectoryPath->m_actorInstance == actorInstance)
             {
-                return trajectoryPath;
+                return trajectoryPath.get();
             }
         }
 
         // We haven't created a path for the given actor instance yet, do so
-        TrajectoryTracePath* tracePath = new TrajectoryTracePath();
+        auto tracePath = AZStd::make_unique<TrajectoryTracePath>();
 
         tracePath->m_actorInstance = actorInstance;
         tracePath->m_traceParticles.reserve(512);
 
-        m_trajectoryTracePaths.emplace_back(tracePath);
-        return tracePath;
+        return m_trajectoryTracePaths.emplace_back(AZStd::move(tracePath)).get();
     }
 
     void AtomActorDebugDraw::RenderTrajectoryPath(AzFramework::DebugDisplayRequests* debugDisplay,
@@ -1107,7 +1106,7 @@ namespace AZ::Render
         const AZ::Color& headColor,
         const AZ::Color& pathColor)
     {
-        TrajectoryTracePath* trajectoryPath = FindTrajectoryPath(actorInstance);
+        auto trajectoryPath = FindTrajectoryPath(actorInstance);
         if (!trajectoryPath)
         {
             return;
--- a/Gems/AtomLyIntegration/EMotionFXAtom/Code/Source/AtomActorDebugDraw.h
+++ b/Gems/AtomLyIntegration/EMotionFXAtom/Code/Source/AtomActorDebugDraw.h
@@ -137,7 +137,7 @@ namespace AZ::Render
 
         Physics::CharacterPhysicsDebugDraw m_characterPhysicsDebugDraw;
         // Motion extraction paths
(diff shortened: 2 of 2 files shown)


Gems/AtomLyIntegration/EMotionFXAtom/Code/Source/AtomActorDebugDraw.cpp
```diff
@@ -961,7 +961,7 @@ namespace AZ::Render
     void AtomActorDebugDraw::UpdateActorInstance(EMotionFX::ActorInstance* actorInstance, float deltaTime)
     {
         // Find the corresponding trajectory trace path for the given actor instance
-        TrajectoryTracePath* trajectoryPath = FindTrajectoryPath(actorInstance);
+        auto trajectoryPath = FindTrajectoryPath(actorInstance);
```
burelc-amzn, line 964: There's not really any shared ownership of this object going on. In both the `UpdateActorInstance` and the `RenderTrajectoryPath` methods, the trajectory path is only viewed, they don't need to be involved in the ownership. That's because the `AtomActorDebugDraw` instance itself owns the `TrajectoryPath`, and the `AtomActorDebugDraw` object won't be destroyed in the middle of one of its methods running.

I think this is a great find, but we should use `unique_ptr` instead, because we don't need shared ownership. Make the `m_trajectoryTracePaths` member a `AZStd::vector<AZStd::unique_ptr<TrajectoryTracePath>>`, and have `FindTrajectoryPath` return a raw pointer (`TrajectoryTracePath*`).

Gems/AtomLyIntegration/EMotionFXAtom/Code/Source/AtomActorDebugDraw.cpp
```diff
@@ -1082,9 +1082,10 @@ namespace AZ::Render
     }
 
     // Find the trajectory path for a given actor instance
-    AtomActorDebugDraw::TrajectoryTracePath* AtomActorDebugDraw::FindTrajectoryPath(const EMotionFX::ActorInstance* actorInstance)
+    AZStd::shared_ptr<AtomActorDebugDraw::TrajectoryTracePath> AtomActorDebugDraw::FindTrajectoryPath(
+        const EMotionFX::ActorInstance* actorInstance)
     {
-        for (TrajectoryTracePath* trajectoryPath : m_trajectoryTracePaths)
+        for (auto trajectoryPath : m_trajectoryTracePaths)
```
burelc-amzn, line 1088: This is another reason to use a `unique_ptr`: this loop has overhead, because using `auto` here will deduce the value type `AZStd::shared_ptr<TrajectoryTracePath>`, and thus will _copy_ each contained shared_ptr in the vector. Copying a `shared_ptr` incurs a cost of incrementing the internal atomic reference counter, which is paired with an atomic decrement when the copy goes out of scope at the end of the loop iteration.

When you change this to use `unique_ptr` instead, you'll have to change this to `auto&`, since `unique_ptr` is not copyable. (`const auto&` would be even better). Of course, with the shared_ptr solution, using `auto&` would avoid the atomic increment/decrement also, but preferring `unique_ptr` _forces_ you to avoid that pitfall.

Jackie9527: done

Gems/AtomLyIntegration/EMotionFXAtom/Code/Source/AtomActorDebugDraw.cpp
Jackie9527: done

Gems/AtomLyIntegration/EMotionFXAtom/Code/Source/AtomActorDebugDraw.cpp
```diff
@@ -1084,30 +1084,30 @@ namespace AZ::Render
...
             if (trajectoryPath->m_actorInstance == actorInstance)
             {
-                return trajectoryPath;
+                return trajectoryPath.get();
             }
         }
 
         // We haven't created a path for the given actor instance yet, do so
-        TrajectoryTracePath* tracePath = new TrajectoryTracePath();
+        AZStd::unique_ptr<TrajectoryTracePath> tracePath = AZStd::make_unique<TrajectoryTracePath>();
```
burelc-amzn, line 1096: You could choose to use `auto` here instead of spelling out the type twice. But it's not a big deal.

Gems/AtomLyIntegration/EMotionFXAtom/Code/Source/AtomActorDebugDraw.cpp
```diff
@@ -1084,30 +1084,30 @@ namespace AZ::Render
...
 
         tracePath->m_actorInstance = actorInstance;
         tracePath->m_traceParticles.reserve(512);
 
-        m_trajectoryTracePaths.emplace_back(tracePath);
-        return tracePath;
+        m_trajectoryTracePaths.emplace_back(AZStd::move(tracePath));
+        return tracePath.get();
```
burelc-amzn, line 1102: This uses the `tracePath` variable after it has been moved from, so you don't know exactly what the state of that variable will be. Because `emplace_back` returns a reference to the inserted object, it would be better to write this instead:
```suggestion
        return m_trajectoryTracePaths.emplace_back(AZStd::move(tracePath)).get();
```

burelc-amzn (commented):
Just a few things left to tidy up.

Gems/AtomLyIntegration/EMotionFXAtom/Code/Source/AtomActorDebugDraw.cpp
Jackie9527: done.

Gems/AtomLyIntegration/EMotionFXAtom/Code/Source/AtomActorDebugDraw.cpp
Jackie9527: done.

amzn-rhhong (approved):

cgalvan (approved):

PR #14507, opened 2023-02-09
Add support to show override icon in DPE based entity inspector

srikappa-amzn:
## What does this PR do?

This PR adds the initial support to show override icons next to overridden properties in a DPE based entity inspector. For example, this is how the comment component with override would look like when loaded
iconAtBeginning

The code is wrapped behind a new feature flag.

RFC: https://github.com/o3de/sig-content/issues/112

What works:
If an override is present in the prefab file, and the prefab is loaded, the icon shows up

What doesn't work:
Icons won't show up unless it resets the entire adapter, which is why the icon doesn't show up for every component when it gets modified (because its not resetting entire adapter for a single property change). Probably why it doesn't work for certain properties like uniform scale and isstatic in transformcomponent. These issues will be addressed in subsequent PRs.

Next steps:

- More tighter integration for prefabs with DPE so that prefab system can use DPE patches to generate prefab patches.
- More testing

## How was this PR tested?

Manually. Will add tests as the feature gets fully developed in coming PRs


Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/PropertyEditorNodes.h
```diff
@@ -204,6 +204,11 @@ namespace AZ::DocumentPropertyEditor::Nodes
         static constexpr auto OnActivate = CallbackAttributeDefinition<void()>("OnActivate");
     };
 
+    struct PrefabOverrideProperty : PropertyEditorDefinition
+    {
+        static constexpr AZStd::string_view Name = "PrefabOverrideProperty";
+    };
```
amzn-tmryan, line 210: Similar to comment above. My understanding is the intention for PropertyEditorNodes and DocumentSchema to contain the general stuff used by many editors, and for more specific impl to go in files focused around that particular area.

That being said, I think this is one of the first implementations specific to a particular system, like prefabs. If it isn't possible yet then it definitely won't break anything to leave it here.

@alexmontAmazon Does that sound right?

Code/Framework/AzCore/AzCore/Component/Component.h
```diff
@@ -118,6 +118,13 @@ namespace AZ
          */
         void SetId(const ComponentId& id)   { m_id = id; }
 
+        virtual void SetSerializedIdentifier(const AZStd::string&){};
+
+        virtual AZStd::string GetSerializedIdentifier()
+        {
+            return AZStd::string();
+        }
```
lumberyard-employee-dm, line 126: To help with build times we should move implementations to cpp files whenever possible.

Code/Framework/AzCore/AzCore/Prefab/PrefabEditorPreferences.cpp
```diff
@@ -37,4 +39,24 @@ namespace AzToolsFramework::Prefab
...
+
+        bool dpeEnabled = false;
+        if (auto* console = AZ::Interface<AZ::IConsole>::Get(); console != nullptr)
+        {
+            console->GetCvarValue("ed_enableDPE", dpeEnabled);
+        }
+
+        return (isInspectorOverrideManagementEnabled && prefabOverridesUxEnabled && dpeEnabled);
+    }
+
```
lumberyard-employee-dm, line 61: Why was this moved to AzCore?

This is purely The PrefabEditorPreferences is purely related to a Tools workflow.

lumberyard-employee-dm (dismissed):

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/AdapterBuilder.cpp
```diff
@@ -165,6 +165,8 @@ namespace AZ::DocumentPropertyEditor
     {
         BeginLabel();
         Attribute(Nodes::Label::Value, text);
+        Attribute(Nodes::PropertyEditor::SharePriorColumn, true);
+        Attribute(Nodes::PropertyEditor::Alignment, Nodes::PropertyEditor::Align::AlignLeft);
```
lumberyard-employee-dm, line 169: Should thees attributes be hardcoded?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/PropertyEditorToolsSystem.cpp
```diff
@@ -27,6 +28,7 @@ namespace AzToolsFramework
     void PropertyEditorToolsSystem::RegisterDefaultHandlers()
     {
         PropertyEditorToolsSystemInterface::RegisterHandler<ContainerActionButtonHandler>();
+        PropertyEditorToolsSystemInterface::RegisterHandler<Prefab::OverrideIconHandler>();
```
amzn-tmryan, line 31: Similar to comments above. This might be more specific than this scope was intended for. I'm not aware of an existing way to register handlers from elsewhere though. That's something we might need, especially for handlers implemented in Gems.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
```diff
@@ -480,40 +522,51 @@ namespace AZ::DocumentPropertyEditor
...
-                        // string-based Filter adapter. Without this, things like Vector3 will not have searchable values by text. These
-                        // advantages would have to be measured against the size changes in the DOM and the time taken to populate and parse them.
-                        return newValue;
-                    },
-                    false, hashValue);
+
+                if (AZ::Prefab::IsInspectorOverrideManagementEnabled() && !serializedPathAttribute.GetString().empty())
+                {
+                    VisitValueWithSerializedPath(access, attributes);
+                }
```
lumberyard-employee-dm, line 529: Why is Prefab specific functionality being added to the ReflectionAdapter?

Code/Framework/AzCore/AzCore/Prefab/PrefabEditorPreferences.cpp
srikappa-amzn: I moved it here because there are some changes in AzCore/Component/* and AzFramework/DocumentPropertyEditor classes that are safer to wrap behind the feature flag until the overrides feature is fully ready. So I put just the helper function in AzCore to make that possible. I don't plan on adding any other prefab related code to AzCore, just the preferences.

Alternatively, I can move this to AzFramework too since nothing in AzCore needs this right now at least. Let me know your thoughts

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
srikappa-amzn: This is only temporary. The idea is that the code wrapped in AZ::Prefab::IsInspectorOverrideManagementEnabled() would become the new normal after the feature is fully developed. If you look at VisitValueWithSerializedPath() , you wouldn't find any prefab specific code in there. The flag is just a safety guardrail for now so that it can be removed once we gain more confidence in the new approach and can be removed.

lumberyard-employee-dm: Since AZ Attributes values can be read from functors?

It seems like a solution is to move the `AZ::Prefab::IsInspectorOverrideManagerEnabled()` check to the reflected type that adds the SerializedPath attribute

amzn-tmryan: An alternative might be to override `VisitObjectBegin` in a more content specific adapter. That might require copying the existing code into the overridden `VisitObjectBegin` though, depending on what you're trying to do.

Code/Framework/AzCore/AzCore/Prefab/PrefabEditorPreferences.cpp
lumberyard-employee-dm: This function can be made as a lambda inside of the ReflectionAdapater.cpp `VisitObjectBegin` function if it is "temporary".

There is no reason to add the entire file Prefab file to AzCore.
Architectural wise, this shouldn't be in AzCore or AzFramework.

If a particular setting needs to be read temporary, then just use the string value directly `"/O3DE/Preferences/Prefabs/EnableInspectorOverrideManagement"`

Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorComponentBase.h
```diff
@@ -117,6 +117,16 @@ namespace AzToolsFramework
...
 
+            void SetSerializedIdentifier(const AZStd::string& alias) override
+            {
+                m_alias = alias;
+            }
+
+            AZStd::string GetSerializedIdentifier() override
+            {
+                return m_alias;
+            }
```
lumberyard-employee-dm, line 128: This should be moved to the cpp file

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.h
```diff
@@ -60,6 +62,10 @@ namespace AZ::DocumentPropertyEditor
         //! property editor instances has altered its value.
         void NotifyPropertyChanged(const PropertyChangeInfo& changeInfo);
 
+        virtual void OnBeginRow(AdapterBuilder*, const AZStd::string_view&)
+        {
+        }
```
lumberyard-employee-dm, line 67: Repeat the implementation should be moved to the cpp file.
Also the `const AZStd::string_view&` should be `AZStd::string_view`
Finally since this is an interface function, it needs a doxygen comment explaining what it is for.

Code/Framework/AzCore/AzCore/Prefab/PrefabEditorPreferences.cpp
srikappa-amzn: Yeah, I agree with the reasoning. I put the file in there since we can't reach AzToolsFramework from AzCore. But its probably an overkill and sets a bad precedent. I'll change it to query the setreg key directly

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/Reflection/LegacyReflectionBridge.cpp
```diff
@@ -228,7 +228,7 @@ namespace AZ::Reflection
                     nullptr);
 
                 const StackEntry& nodeData = m_stack.back();
-                m_serializeContext->EnumerateInstance(&context, nodeData.m_instance, nodeData.m_typeId, nullptr, nullptr);
+                m_serializeContext->EnumerateInstance(&context, nodeData.m_instance, nodeData.m_typeId, nodeData.m_classData, nodeData.m_classElement);
```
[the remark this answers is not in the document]
lumberyard-employee-dm: EnumerateInstance uses the `ClassData` parameter as the class data for the pointer instance supplied.
When `ClassData` is nullptr, it uses the class TypeId to lookup the ClassData from the SerializeContext.

Now the `ClassElement` is used to determine whether the void pointer instance refers to a type T or to a pointer to type T
For example if the type being passed references `AZ::Entity*`, then the void pointer supplied to EnumerateInstance is treated as a `void**` and deferenced in order to access the `AZ::Entity` instance.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/DocumentPropertyEditor/PrefabAdapter.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include <AzFramework/DocumentPropertyEditor/AdapterBuilder.h>
+#include <AzFramework/DocumentPropertyEditor/PropertyEditorNodes.h>
+#include <AzToolsFramework/Prefab/DocumentPropertyEditor/PrefabAdapter.h>
+#include <AzToolsFramework/Prefab/Overrides/PrefabOverridePublicInterface.h>
+
+namespace AzToolsFramework::Prefab
+{
+    PrefabAdapter::PrefabAdapter()
+    {
+        m_prefabOverridePublicInterface = AZ::Interface<AzToolsFramework::Prefab::PrefabOverridePublicInterface>::Get();
```
lumberyard-employee-dm, line 18: Unless the PrefabOverridPublicPublicInterface is being held by reference, I don't see a good reason to hold onto it as a member.

It is dangerous if the registered PrefabOverridPublicPublicInterface needs to get replace on the AZ::Interface during the run.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/AdapterBuilder.cpp
alexmontAmazon: labels should definitely not share prior column by default -- that's a specialty flag that combines columns. Please contact me offline so that we can look into what issue you're trying to solve, and hopefully we can fix it on the DPE side. Ditto that alignLeft shouldn't be necessary to everything. If you can show me what it looks like with and without that, I'm sure we can make the desired effect the default in the DPE, and not something everything specifies in the DOM. Keep in mind that any attribute the builder is setting will have to be written, stored, and interpreted for every one of those nodes. Ideally, the default behavior would be the most common behavior, and specifying a particular attribute will be relatively rare.

Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorComponentBase.cpp
```diff
@@ -94,5 +94,15 @@ namespace AzToolsFramework
         {
             return AzToolsFramework::IsSelected(GetEntityId());
         }
+
+        void EditorComponentBase::SetSerializedIdentifier(const AZStd::string& alias)
+        {
+            m_alias = alias;
+        }
```
lumberyard-employee-dm, line 101: This should be accepted by value and moved into the m_alias member, to avoid an unnecessary copy when this function is invoked with a string literal.
i.e
```
MyEditorComponent editorComponent;
editorComponent->SetSerializedIdentifier("foo"); // Has to construct a new AzStd::string and bind it to a lvalue reference
// Then inside the function it must be make a copy because the parameter is const lvalue.
```
```suggestion
        void EditorComponentBase::SetSerializedIdentifier(AZStd::string alias)
        {
            m_alias = AZStd::move(alias);
        }
```

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/PropertyEditorNodes.cpp
```diff
@@ -96,6 +96,8 @@ namespace AZ::DocumentPropertyEditor::Nodes
         system->RegisterNodeAttribute<ContainerActionButton>(ContainerActionButton::Action);
         system->RegisterNodeAttribute<ContainerActionButton>(ContainerActionButton::OnActivate);
 
+        system->RegisterPropertyEditor<PrefabOverrideProperty>();
```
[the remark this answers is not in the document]
alexmontAmazon: +1 to this, let's keep the main generic code modules clean and... well, generic.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/PropertyEditorNodes.h
alexmontAmazon: +1. Confine prefab-related things to a prefab file. We hope to eventually have *tons* of editors based on the DPE, so we can't afford to jam everything into the base.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DPEComponentAdapter.cpp
```diff
@@ -71,11 +75,18 @@ namespace AZ::DocumentPropertyEditor
...
         AzToolsFramework::ToolsApplicationEvents::Bus::Handler::BusConnect();
         AzToolsFramework::PropertyEditorGUIMessages::Bus::Handler::BusConnect();
         AZ::Uuid instanceTypeId = azrtti_typeid(m_componentInstance);
         SetValue(m_componentInstance, instanceTypeId);
+        m_componentAlias = componentInstance->GetSerializedIdentifier();
+        auto owningInstance = AZ::Interface<AzToolsFramework::Prefab::InstanceEntityMapperInterface>::Get()->FindOwningInstance(m_entityId);
+        AZ_Assert(owningInstance.has_value(), "Entity owning the component doesn't have an owning prefab instance.");
+        auto entityAlias = owningInstance->get().GetEntityAlias(m_entityId);
+        AZ_Assert(entityAlias.has_value(), "Owning entity of component doesn't have a valid entity alias in the owning prefab.");
+        m_entityAlias = entityAlias->get();
```
lumberyard-employee-dm, line 89: What is the entityAlias used for?
It is not passed into `OnBeginRow`, so it is currently unused.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
```diff
@@ -246,6 +249,55 @@ namespace AZ::DocumentPropertyEditor
...
+                    auto outputWriter = newInstancePointerValue.GetWriteHandler();
+                    auto convertToAzDomResult =
+                        AZ::Dom::Json::VisitRapidJsonValue(serializedValue, *outputWriter, AZ::Dom::Lifetime::Temporary);
+                    return newInstancePointerValue;
+                },
+                false,
+                false);
+        }
+
+        bool IsInspectorOverrideManagementEnabled()
```
alexmontAmazon, line 289: I don't think the ReflectionAdapter should need to know about Inspector/override specific concepts. Can you encapsulate this in a prefab or Insepctor specific class?

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
```diff
@@ -480,40 +536,51 @@ namespace AZ::DocumentPropertyEditor
...
-                        // In future, to properly support multiple DPEs on one adapter, we will need to solve this. One way would be to store
-                        // the json serialized value (which is mostly human-readable text) as an attribute, so any change to the Json would
-                        // trigger an update. This would have the advantage of allowing opaque and pointer types to be searchable by the
-                        // string-based Filter adapter. Without this, things like Vector3 will not have searchable values by text. These
-                        // advantages would have to be measured against the size changes in the DOM and the time taken to populate and parse them.
-                        return newValue;
-                    },
-                    false, hashValue);
+
+                if (IsInspectorOverrideManagementEnabled() && !serializedPathAttribute.GetString().empty())
```
alexmontAmazon, line 540: I'd rather not have IsInspector-type stuff here. Let's find a way to hook into this that isn't Inspector specific, something like your OnBeginRow mechanism above. You can call a virtual function here that you override in your derived adapter, or something like that. There are editors (like the AssetEditor) that use the ReflectionAdapter but have nothing to do with prefabs or overrides. Let's keep this clean.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
```diff
@@ -480,40 +536,51 @@ namespace AZ::DocumentPropertyEditor
...
+
+                            // serialize the new value to Json, using the original valuePointer as a reference object to generate a minimal
+                            // diff
+                            JsonSerialization::Store(
+                                buffer, buffer.GetAllocator(), marshalledPointer, valuePointer, valueType, serializeSettings);
+
+                            // now deserialize that value into the original location
+                            JsonSerialization::Load(valuePointer, valueType, buffer, deserializeSettings);
+
+                            // NB: the returned value for serialized pointer values is instancePointerValue, but since this is passed by
```
alexmontAmazon, line 570: this is a good comment, but we should be tracking this as a GHI. Could you write one up? I think this is actually already testable, right? If you create an inspector looking at a prefab, then pin it, then create another Inspector looking at the same object, changes made in one of the Inspectors won't show up in the other one when there are pointers like this involved.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.h
```diff
@@ -60,6 +62,8 @@ namespace AZ::DocumentPropertyEditor
         //! property editor instances has altered its value.
         void NotifyPropertyChanged(const PropertyChangeInfo& changeInfo);
 
+        virtual void OnBeginRow(AdapterBuilder*, AZStd::string_view);
```
alexmontAmazon, line 65: I like this, it's pleasantly generic / multi-purpose

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/PropertyEditorToolsSystem.cpp
alexmontAmazon: I think you can register them wherever you want, so long as they're not somehow used first. I'd put this somewhere prefab-specific, like when your adapter is initialized.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyEditorAPI_Internals.h
```diff
@@ -265,7 +268,19 @@ namespace AzToolsFramework
             auto value = AZ::DocumentPropertyEditor::Nodes::PropertyEditor::Value.ExtractFromDomNode(node);
             if (value.has_value())
             {
-                m_proxyValue = AZ::Dom::Utils::ValueToType<WrappedType>(value.value()).value_or(m_proxyValue);
+                if (!AzToolsFramework::Prefab::IsInspectorOverrideManagementEnabled())
```
alexmontAmazon, line 271: this layer is just for RPE-specific stuff. This is very much a bridge to the old system. Please find an Inspector-specific place to put this so it doesn't bleed into generic PropertyEditor stuff

alexmontAmazon (changes requested):

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabEditorPreferences.cpp
```diff
@@ -6,12 +6,13 @@
  *
  */
 
-#include <AzToolsFramework/Prefab/PrefabEditorPreferences.h>
 #include <AzCore/Settings/SettingsRegistry.h>
+#include <AzToolsFramework/Prefab/PrefabEditorPreferences.h>
 
 namespace AzToolsFramework::Prefab
 {
     static constexpr AZStd::string_view EnablePrefabOverridesUxKey = "/O3DE/Preferences/Prefabs/EnableOverridesUx";
```
forkercat, line 14: I would suggest renaming it to `EnableOutlinerOverrideManagement` in a task to keep consistency with the new flag. `EnableOverridesUx`'s purpose is not very clear to me compared to the new flag.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabFocusHandler.cpp
```diff
@@ -276,7 +276,7 @@ namespace AzToolsFramework::Prefab
         // Close all container entities in the old path.
         SetInstanceContainersOpenState(m_rootAliasFocusPath, false);
 
-        if (IsPrefabOverridesUxEnabled())
+        if (AzToolsFramework::Prefab::IsPrefabOverridesUxEnabled())
         {
```
forkercat, line 280: Is this needed? Same comment for below checks as well.

Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorComponentBase.h
```diff
@@ -117,6 +117,10 @@ namespace AzToolsFramework
             virtual void Deactivate() override;
             //////////////////////////////////////////////////////////////////////////
 
+            void SetSerializedIdentifier(const AZStd::string& alias) override;
+
+            AZStd::string GetSerializedIdentifier() const override;
+
```
forkercat, line 123: Would be good to add the missing comments

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverridePublicInterface.h
```diff
@@ -25,7 +26,7 @@ namespace AzToolsFramework
             //! by the class implmenting this interface based on certain selections in the editor. eg: the prefab currently being edited.
             //! @param entityId The id of the entity to check for overrides.
```
forkercat, line 27: nit: Missing comment for `prefix`.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverridePublicHandler.cpp
```diff
@@ -51,11 +51,15 @@ namespace AzToolsFramework
             }
         }
 
-        bool PrefabOverridePublicHandler::AreOverridesPresent(AZ::EntityId entityId)
+        bool PrefabOverridePublicHandler::AreOverridesPresent(AZ::EntityId entityId, const AZ::Dom::Path& prefix)
         {
```
forkercat, line 55: I feel like we can have a separate function to check if overrides are present for components and properties. I'm okay with both solutions.

Another thought is we can use `relativePathToValue` or something similar rather than `prefix`? I used to find it hard to follow (in some functions in our interfaces) what prefix it is. From my first glance, I thought it was a prefix to an entity.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Outliner/EntityOutlinerListModel.cpp
```diff
@@ -1262,7 +1262,7 @@ namespace AzToolsFramework
 
     void EntityOutlinerListModel::OnContainerEntityStatusChanged(AZ::EntityId entityId, [[maybe_unused]] bool open)
     {
-        if (!Prefab::IsPrefabOverridesUxEnabled())
+        if (!AzToolsFramework::Prefab::IsPrefabOverridesUxEnabled())
```
forkercat, line 1265: same comment: Is it still needed in latest revision?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabIntegrationManager.cpp
```diff
@@ -446,7 +446,7 @@ namespace AzToolsFramework
                                 );
                             }
 
-                            if (IsPrefabOverridesUxEnabled())
+                            if (AzToolsFramework::Prefab::IsPrefabOverridesUxEnabled())
```
forkercat, line 449: same comment: Is it still needed in latest revision?

forkercat (commented):
Thanks for putting up the changes. Left some comments about changes in prefab code. Will continue reviewing other parts.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/PropertyEditorNodes.cpp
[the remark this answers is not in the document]
NicholasVanSickle: Yeah I think this is hitting a growing pain of "how do non-RPE systems register to the DPE?" - there should probably be a bus or something that single-shot notifies consumers when `PropertyEditorSystemInterface` is available so that other systems (inc gems) can plug in. @alexmontAmazon / @srikappa-amzn feel free to ping me on that if you've got design questions

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
NicholasVanSickle: I think in general we should figure out a delineation between a prefab adapter and a general reflection adapter - the most obvious to my mind is making `PrefabAdapter` a subclass of `ReflectionAdapter` and exposing these touch-points via virtual methods that are a no-op for reflection, but I don't personally care too much as long as there's some kind of divide. (See @alexmontAmazon , sometimes I DO lean towards inheritance over composition!)

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.h
NicholasVanSickle: Hmm, yeah, this gave me pause but I think I'm peaceful with this as a general mechanism in the reflection adapter - `AdapterBuilder` isn't meant to be generally required for building DPEs (it's just a bag of utility functions to make building the tree easier), but it should be fine to expose it as an implementation detail for plugging into `ReflectionAdapter` specifically.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/Reflection/LegacyReflectionBridge.cpp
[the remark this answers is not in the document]
NicholasVanSickle: I'll happily defer to @lumberyard-employee-dm on this one - last I remember we were using a forked copy of this in the edit context, so I'm way out of date on how this bit works.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/AdapterBuilder.cpp
NicholasVanSickle: Honestly I'm wondering if there shouldn't just be a Prefab-specific `OverridablePropertyLabel` that handles the icon, text styling, etc rather than trying to do it with multiple handlers and dealing with layout shenanigans - should be straightforward to wrap where the ReflectionAdapter is making property labels with a virtual method you can override.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverridePublicHandler.cpp
NicholasVanSickle: +1, interface seems fine but I'd call out that it's specifically a dom subpath of the entity

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyEditorAPI_Internals.h
NicholasVanSickle: This looks like a gradual switch-over to loading JSON-serialized values here which is very cool! Maybe we can just start doing that everywhere instead of gating it on overrides? There aren't that many other DPE adapters so I'm expected a light lift, but I also could be super wrong - maybe there are some reflected types that will really struggle with this - long term I'd love to see these all just marshal serialized values in and out though

NicholasVanSickle (commented):
Did a quick lookthrough - really exciting progress! I think there's some things to figure out in terms of finding the right abstractions for building on top of the DPE, especially on top of the ReflectionAdapter, but I'm heartened to see this shaping up. I'll try to keep an eye on activity around this for a bit in case I can be of help (also feel free to tell me to go away if I'm *not* being of help, I'm definitely out of date on several bits)

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/AdapterBuilder.cpp
srikappa-amzn: Okay. Removed it now. We have follow up tasks to fix up ux issues with the override icon. This can be done then. I like the idea of making a virtual method and letting the DPEComponentAdapter & PrefabAdapter handle the aligning. We can even do the aligning only when overrides are present too if we want to be really specific.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/Reflection/LegacyReflectionBridge.cpp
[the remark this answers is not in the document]
srikappa-amzn: Okay. Based on the discussion above, I think I'll revert it to what it was before since it's not a required change for this functionality to work. We can revisit this later if this becomes an issue.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DPEComponentAdapter.cpp
srikappa-amzn: Thanks for catching that! Removed this.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/DocumentPropertyEditor/PrefabAdapter.cpp
srikappa-amzn: PrefabOverridePublicHandler is a singleton that is initialized once when the editor starts. It registers with the PrefabOverridPublicPublicInterface when it is constructed and doesn't unregister from it until the application is terminated. So we will not run into the issue you mentioned above.

I agree that holding onto raw pointers is always risky. But it has been common practice in the prefab code to hold on to interface pointers backed by singletons to avoid fetching them repeatedly and making the code bloated (because interfaces are used heavily in prefab code). So I am inclined to keep it this way for now.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
srikappa-amzn: This helper function simply checks whether a particular feature flag is enabled or not. It doesn't have any prefab specific logic in it. This is only temporary and will be removed once the work in https://github.com/o3de/sig-content/issues/112 is finished. 

I agree that it feels odd to have it in this class but it'd be odd to have it anywhere else in AzFramework as well since everything about prefabs is in AzToolsFramework. So it's better to keep it here locally than make it a common utils method in AzFramework elsewhere. See https://github.com/o3de/o3de/pull/14507#discussion_r1101914970 for related discussion on this.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
srikappa-amzn: Similar to my other response, the IsInspectorOverrideManagementEnabled() function is simply a feature flag check and is only temporary and will go away.

The serialized path attribute mechanism is a new way of storing opaque types in DPE dom, where we will be storing fully serialized values( if there is a serialized path associated with the value). It's not prefab or inspector specific code. So it belongs in this class imo and will become the new way of storing opaque values in DPE DOM after the work in https://github.com/o3de/sig-content/issues/112 finishes. 

I just put this behind a flag so that we can test it properly and not impact other DPE editor already released like the cvar editor.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/PropertyEditorNodes.cpp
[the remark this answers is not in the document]
srikappa-amzn: While refactoring this code, I realized that even if I don't call RegisterPropertyEditor<PrefabOverrideProperty>() , it still works and I can see the override icon. So I removed this registering part altogether.

@NicholasVanSickle , what exactly does RegisterPropertyEditor<>() do and how does it work even without it? Can we skip this for now?

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/PropertyEditorNodes.h
srikappa-amzn: Moved this to AzToolsFramework/Prefab/DocumentPropertyEditor/PrefabPropertyEditorNodes.h

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/PropertyEditorToolsSystem.cpp
srikappa-amzn: Moved this registration to the constructor of PrefabAdapter

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyEditorAPI_Internals.h
srikappa-amzn: Yes, this will eventually be the new way of loading opaque values from DPE DOM. This isn't inspector only logic. The check is only temporary and is put in place so that we can do enough testing and gain confidence instead of changing everything at once.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
srikappa-amzn: Sure. Here it is https://github.com/o3de/o3de/issues/14583

This is a pre-existing comment. So I'm not 100% sure what it is suggesting. But if the solution is to store serialized values in DPE dom, this PR already does it behind a flag. So it might just be a case of testing out with an actual use case and seeing where things are at.

Few questions about your comment:

1. By 'pin it' , do you mean 'create an override' or something else?
2. What does create another inspector mean? There will only be 1 entity inspector in the editor right? I'm trying to understand what an actual use case for this would be. Please update it in the ghi created above and feel free to edit the issue if its missing any information.

alexmontAmazon: No, I mean pin it using the pin button:
image
This leaves a "pinned" copy of the Inspector looking at the current Entity. There can be many inspectors open at the same time.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
alexmontAmazon: ok, but you have it copy-pasted in two places, and this one is a file called ReflectionAdapter that doesn't need to know anything about prefabs! Can you at least merge these two identical functions into one place and have them both call the same method?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabEditorPreferences.cpp
```diff
@@ -37,4 +38,16 @@ namespace AzToolsFramework::Prefab
         return prefabOverridesUxEnabled;
     }
 
+    bool IsInspectorOverrideManagementEnabled()
```
alexmontAmazon, line 41: this is the duplicate.

forkercat (approved):
Overall looks good to me. I'd like to follow up to learn more about why PrefabOverrideProperty nodes and attributes can be used without registration.

Additionally, I would recommend updating the screenshot in description since the spacing is not fixed by the latest revision of the PR. And it would be good to point out in 'What doesn't work' that for some property labels like 'Uniform Scale' and 'IsStatic' the override icons still do not show up even if they are overridden.

alexmontAmazon (approved):
I'm okay with the temporary changes for now, but please add some kind of "todo" comments in places (like the feature flag query) that we hope to replace soon.

PR #14516, opened 2023-02-09
Fix thumbnail view names

### LINKED ISSUE #14399: Asset Browser: Filenames in Thumbnail View are often truncated to the point they are indistinguishable from each other
**Describe the bug**
Filenames in Thumbnail View are often truncated to the point they are indistinguishable from each other.

**Steps to reproduce**
```
```
2. Open the Asset Browser, and switch to List View if necessary.
3. Create a new folder. 
4. In the new folder, create several new assets of the same type.
5. Switch to Thumbnail View.

**Expected behavior**
Filenames can be clearly read or, if truncated, hovering over the assets displays the full filename.

**Actual behavior**
Assets are indistinguishable from each other by filename.

**Screenshots/Video**
image

**Found in Branch**
development

c8d69fb61c196601651391f4e22d1408971052d8

dtamkin1:

## What does this PR do?
Adds a second line to each thumbnail to make it easier to distinguish them from each other
Fixes #14399.

xHqfBZwVZM

## How was this PR tested?
Locally 

- [x] Tooltip when hovering over assets
- [x] Elide in the middle, show name and file extension
- [x] default height spacing, normal font weight, no extra spacing
- [x] Left align if two rows
- [ ] 12 point font at its smallest


cgalvan (approved):
You might want to take a quick look and see if you can pass the `Qt::TextWordWrap` flag as well in the flags you're already sending to `drawText` to see if it will do the wrapping for you. Since we probably don't want to give it unlimited height, your approach might still be better since you can explicitly restrict to two lines and then elide the rest.

https://doc.qt.io/qt-6/qt.html#TextFlag-enum

gadams3 (approved):

dtamkin1:
@rainbj @bhanuja-s Is a maximum of two rows a good number, and does the spacing between each thumbnail look good?

rainbj:
Hi @dtamkin1. I feel that if we go more than two lines, it will end up making the UI look very strange (odd gaps under some rows). Some great ideas from Luis plus some of my own combined.



- If an asset name takes up more than one row. Make sure the font size is set to the lowest number we use. Which should be 12 pt font. From an accessibility perspective, we should never go smaller than 12 pt font, as legibility becomes almost impossible for some readers. 
- make sure the font weight is set to normal (not bold)
- Make sure the line height spacing is the default for the font type. Make sure you clear out any extra spacing.
- Left align the text instead of center aligning if the text is two rows. Center aligning two rows could make the text staggered, and make the text harder to read. 
- Use an ellipse at the front of two lines of text instead of at the end. This is because knowing the file extension type is more important than seeing the full name at a glance. 
- Tooltips should work for large file names such as this. Hiding part of a file name without any method of seeing the full name will be problematic.

@bhanuja-s to include any details I might have missed

jckand-amzn:
>     * Tooltips should work for large file names such as this. Hiding part of a file name without any method of seeing the full name will be problematic.

+1, I think tooltips are a necessity.

bhanuja-s:
> Use an ellipse at the front of two lines of text instead of at the end. This is because knowing the file extension type is more important than seeing the full name at a glance.

Rather than an ellipsis at the front of the text, maybe we could have an ellipsis in the middle? This way users have a way of distinguishing by file name, and extension type?

I'm attaching an example from macOS' Finder: 

image

rainbj:
> In addition to the font size, I believe we should have a line height in the range of 1.2–1.5x the font size (12 in this case). Currently, the second line is merging with the first—this should increase readability.
> 

- Small point of clarification @bhanuja-s . "The font size at it's smallest should be 12 (pt)"- not **(in)** (see above) 12 inches would basically be an entire page :)
- Second point. Line height. 1 vs 1.2 vs 1.5 would just need to be viewed in real time, in the editor. My stance is that the text string shouldn't be as big as an entire row of the thumbnail images. So I would say it should be about 1/2 or at most 3/4 of the icon row size. So this final value is a hard to nail down because processed fonts sometimes do odd things on screens (antialiasing, graphics interpretation of what a pt is measured out at on any given graphical display, etc) So basically if 1.5 looks like it fits within those parameters then I'm ok with it.

> > Use an ellipse at the front of two lines of text instead of at the end. This is because knowing the file extension type is more important than seeing the full name at a glance.
> 
> Rather than an ellipsis at the front of the text, maybe we could have an ellipsis in the middle? This way users have a way of distinguishing by file name, and extension type? @rainbj what are your thoughts?

- I like this idea! Made my idea even better. Sold :) Question to @jckand-amzn if this is this possible in a reasonable amount of time OR will this take a substantial amount of programming to make this work? Thanks B

dtamkin1:
@rainbj @bhanuja-s I've implemented the following changes to the PR:

- Added tooltips when you mouse over assets
- Left align if text is two lines
- Move file extension to always appear on the second line
- Modified ellipsis to show the beginning and end of a file, prioritizing file types if present to always show them
- Made sure that height and weight were default, with no extra spacing.
 
The font size used for each asset is already 12 point, so I did not modify the font size. Here are some images:

image
image

lsemp3d (approved):

PR #14518, opened 2023-02-09
Added support for a generic QLabel as UIElement

cgalvan:
## What does this PR do?

This feature adds a generic label handler that can be used to display text in a property editor. This has been requested several times over the years, most recently for the PhysX Static Rigid Body component that no longer has actual properties to be edited, so it would be helpful to still give the user some information.

A brief list of the changes are:
- Added a label handler that derives from `GenericPropertyHandler` so it isn't tied to specified data type
- Fixed logic error that would prevent anything other than a button being used as the handler for a `UIElement`
- Added an example of using the label as a UIElement in the Qt control gallery
- Updated the PhysX Static Rigid Body component to use the new `Label` handler instead of `MultiLineEdit`
- Fixed the `MultiLineEdit` to handle the `ValueText` attribute (using this handler was the original requested approach)
- Fixed up some attributes I came across to use `AZ_CRC_CE` and to use the constants across the Editor instead of replicating the CRC

UIElementLabel_GalleryExample

UIElementLabel_RigidBody

## How was this PR tested?

Tested all components modified to verify they still work as expected. Tested other components as well to ensure modifying the logic in the handler resolver didn't break other components getting the proper handler.


amzn-tmryan (approved):

gadams3 (approved):

phuchau1989 (approved):

Gems/PhysX/Code/Source/EditorStaticRigidBodyComponent.cpp
```diff
@@ -33,12 +33,10 @@ namespace PhysX
...
-                    ->UIElement(AZ::Edit::UIHandlers::MultiLineEdit, "", ToolTip)
-                    ->Attribute(AZ::Edit::Attributes::ValueText,"<i>Component properties not required</i>")
-                    ->Attribute(AZ::Edit::Attributes::ReadOnly, true)
-                    ->UIElement(AZ::Edit::UIHandlers::MultiLineEdit, "", ToolTip)
-                    ->Attribute(AZ::Edit::Attributes::ValueText, "Non-movable rigid body in PhysX")
-                    ->Attribute(AZ::Edit::Attributes::ReadOnly, true);
+                    ->UIElement(AZ::Edit::UIHandlers::Label, "", ToolTip)
+                        ->Attribute(AZ::Edit::Attributes::NameLabelOverride, "")
+                        ->Attribute(AZ::Edit::Attributes::ValueText, "<i>Component properties not required</i><br>Non-movable rigid body in PhysX")
+                    ;
```
moraaar, line 39: very cool! Thanks!

moraaar (approved):
This is a great addition! Thank you very much @cgalvan

dtamkin1 (approved):

PR #14521, opened 2023-02-09
NetworkCharacterComponent cleanup from previous PR feedback

mbalfour-amzn:
## What does this PR do?

This addresses the feedback that appeared in #14484 .

1. Added OnCharacterDeactivated() which sets m_physicsCharacter back to null.
2. Moved the network connect / disconnect code into Activate() / Deactivate() instead of OnCharacterActivated() / OnCharacterDeactivated().

## How was this PR tested?

Ran MultiplayerSample and verified the character still moved around successfully.

### DIFF
--- a/Gems/Multiplayer/Code/Source/Components/NetworkCharacterComponent.cpp
+++ b/Gems/Multiplayer/Code/Source/Components/NetworkCharacterComponent.cpp
@@ -113,17 +113,24 @@ namespace Multiplayer
         // During activation the character controller is not created yet.
         // Connect to CharacterNotificationBus to listen when it's activated after creation.
         Physics::CharacterNotificationBus::Handler::BusConnect(GetEntityId());
+
+        // Set up the network event handlers. These can be bound before the character controller is created
+        // because they check for the character controller to exist inside of the handlers.
+
+        GetNetBindComponent()->AddEntitySyncRewindEventHandler(m_syncRewindHandler);
+
+        if (!HasController())
+        {
+            GetNetworkTransformComponent()->TranslationAddEvent(m_translationEventHandler);
+        }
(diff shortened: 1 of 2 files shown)


AMZN-Olex (approved):

lmbr-pip (approved):

Gems/Multiplayer/Code/Source/Components/NetworkCharacterComponent.cpp
```diff
@@ -113,17 +113,24 @@ namespace Multiplayer
         // During activation the character controller is not created yet.
         // Connect to CharacterNotificationBus to listen when it's activated after creation.
         Physics::CharacterNotificationBus::Handler::BusConnect(GetEntityId());
+
+        // Set up the network event handlers. These can be bound before the character controller is created
+        // because they check for the character controller to exist inside of the handlers.
+
```
AMZN-Gene, line 119: remove newline between the comment and code?

AMZN-Gene (approved):

moraaar (approved):
nit: Connecting to events unnecessarily and spend the time sending those events to do no ops is less optimal. My preferred option would be to remove the `if (m_physicsCharacter)` checks in the code and connect to the events at the right moment.

But the current code works and it's perfectly valid too so approve.

Gems/Multiplayer/Code/Source/Components/NetworkCharacterComponent.cpp
mbalfour-amzn: It's a style choice. I tend to use a newline separator if the comment refers to a series of lines that follow it, and no newline if it only refers to a single line.

PR #14525, opened 2023-02-10
Adding GravitySets & EntityFallsUnderGravity tests

FuzzyCarterAWS:
Signed-off-by: FuzzyCarterAWS <jscar@amazon.com>

## What does this PR do?

Adds two new GTest unit tests to the CharacterGameplayControllerTests

## How was this PR tested?

Tests have passed locally in isolation and as part of entire gtests suite.

```
[----------] 2 tests from PhysXDefaultWorldTest
[ RUN      ] PhysXDefaultWorldTest.CharacterGameplayController_GravitySets
[       OK ] PhysXDefaultWorldTest.CharacterGameplayController_GravitySets (2 ms)
[ RUN      ] PhysXDefaultWorldTest.CharacterGameplayController_GravitySetsWhileMoving
[       OK ] PhysXDefaultWorldTest.CharacterGameplayController_GravitySetsWhileMoving (6 ms)
[----------] 2 tests from PhysXDefaultWorldTest (13 ms total)

[----------] 7 tests from PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture
[ RUN      ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/0
[       OK ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/0 (4 ms)
[ RUN      ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/1
[       OK ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/1 (10 ms)
[ RUN      ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/2
[       OK ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/2 (18 ms)
[ RUN      ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/3
[       OK ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/3 (26 ms)
[ RUN      ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/4
[       OK ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/4 (33 ms)
[ RUN      ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/5
[       OK ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/5 (38 ms)
[ RUN      ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/6
[       OK ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/6 (51 ms)
[----------] 7 tests from PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture (201 ms total)

[==========] 9 tests from 2 test cases ran. (672 ms total)
[  PASSED  ] 9 tests.
```


Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
```diff
@@ -124,4 +133,33 @@ namespace PhysX
...
+        for (int i = 0; i < timeStepCount; i++)
+        {
+            basis.Update();
+            totalTime += timeStep;
+        }
+
+        auto endPosition = basis.m_controller->GetPosition();
+        
+        // calculate distance fallen (d = 0.5 * g * t^2)
+        distanceFell = 0.5f * (-10) * (totalTime * totalTime);
```
hultonha, line 160: nit: Prefer `-10.0f` (stop `int`/`float` conversion) - also you can omit the parens

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
```diff
@@ -124,4 +133,33 @@ namespace PhysX
...
+        
+        EXPECT_THAT(endPosition.GetZ(), testing::FloatNear(startPosition.GetZ() + distanceFell, 0.1f));
```
hultonha, line 162: nit: `0.1f` tolerance seems high, might be worth decreasing this to a much smaller value for this sort of test

hultonha (commented):
Overall looks good I just have a few minor suggestions, thanks!

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
```diff
@@ -107,6 +107,15 @@ namespace PhysX
         float m_timeStep = AzPhysics::SystemConfiguration::DefaultFixedTimestep;
     };
 
+    TEST_F(PhysXDefaultWorldTest, CharacterGameplayController_GravitySets)
+    {
+        float expectedGravityMultiplier = 2.5f;
```
moraaar, line 112: `expectedGravityMultiplier` can be const.

````
const float expectedGravityMultiplier = 2.5f;
````

File: Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp [hunk 4]
```diff
@@ -124,4 +133,33 @@ namespace PhysX
             EXPECT_THAT(expectedGravityMultiplier + i, testing::FloatEq(basis.m_gameplayController->GetGravityMultiplier()));
         }
     }
+
+    TEST_F(PhysXDefaultWorldTest, CharacterGameplayController_EntityFallsUnderGravity)
+    {
+        GameplayTestBasis basis(m_testSceneHandle, DefaultGravityMultiplier, DefaultGroundDetectionBoxHeight, DefaultFloorTransform);
+        auto* sceneInterface = AZ::Interface<AzPhysics::SceneInterface>::Get();
+        sceneInterface->RemoveSimulatedBody(basis.m_sceneHandle, basis.m_floor->m_bodyHandle);
```
moraaar, line 141: A comment indicating why the floor rigid body is being deleted would help to understand more the test.

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
```diff
@@ -124,4 +133,33 @@ namespace PhysX
...
+        
+        // Let scene run for a few moments so the entity can be manipulated by gravity from the gameplay component
+        auto startPosition = basis.m_controller->GetPosition();
+        
+        int timeStepCount = 10;
+        float totalTime = 0.0f;
+        float timeStep = AzPhysics::SystemConfiguration::DefaultFixedTimestep;
+        float distanceFell = 0.0f;
```
moraaar, line 149: - `startPosition` can be const.
- `timeStepCount` can be const
- `timeStep` can be const.
- `distanceFell` is not used until later so it can be moved where it's needed and be const.
````
const float distanceFell = 0.5f * (-10.0f) * (totalTime * totalTime);
`````

File: Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp [hunk 6]
```diff
@@ -124,4 +133,33 @@ namespace PhysX
...
+
+        for (int i = 0; i < timeStepCount; i++)
+        {
+            basis.Update();
+            totalTime += timeStep;
+        }
+
+        auto endPosition = basis.m_controller->GetPosition();
```
moraaar, line 157: `endPosition` can be const.

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
moraaar, line 160: `(-10.0f)` is a magic number. Please assign to a const variable.
````
const float gravity = -10.0f;
````
Question: Is -10 for gravity used on purpose for this test for some reason? If yes then please add a comment. If not maybe just use `AzPhysics::DefaultGravity.GetZ()` instead.

File: Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp (hunk 2 above)
Comment (moraaar) [reply Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp:162]: +1

File: Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp (hunk 1 above)
Comment (FuzzyCarterAWS) [reply Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp:160]: This was actually a debugging artifact. Reverted the code back to getting the gravity from the scene.

File: Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp (hunk 3 above)
Comment (FuzzyCarterAWS) [reply Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp:112]: Added

File: Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp (hunk 4 above)
Comment (FuzzyCarterAWS) [reply Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp:141]: Refactored to test basis and added a docstring to the function as well as a comment to the test.

File: Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp (hunk 5 above)
Comment (FuzzyCarterAWS) [reply Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp:149]: Udpated

File: Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp (hunk 1 above)
Comment (FuzzyCarterAWS) [reply Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp:160]: I was using -10 for debugging and mistakenly removed the code to get the gravity from the scene using `basis.m_testScene->GetGravity();` and using that object's z value.

File: Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp (hunk 6 above)
Comment (FuzzyCarterAWS) [reply Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp:157]: Updated

File: Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp (hunk 2 above)
Comment (FuzzyCarterAWS) [reply Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp]: @hultonha @moraaar @greerdv 

If I change the tolerance to `0.01f` or lower these test will fail (I've pushed up the code using this value atm).

```
E:/gws/o3de/Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp(170): error: Value of: endPosition.GetZ()
Expected: is approximately -0.1362505 (absolute error <= 0.0099999998)
  Actual: -0.122625 (of type float), which is 0.013625 from -0.13625
```

Is there something I can do to get the calculations to be closer to use a tighter tolerance, or is it ok in this instance to use a tolerance of `0.1f`?

Comment (moraaar) [reply Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp]: Unless the math done to calculate `distanceFell` matches the math used internally by PhysX I'm not sure you'll ever get a reliable comparison. Maybe a more reliable test would be to check the end position is lower than the start position. Tagging @greerdv

Comment (hultonha) [reply Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp]: Yeah I agree, I'm not sure how helpful this is if the tolerance is large and the values don't actually really match

File: Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp (hunk 1 above)
Comment (hultonha) [reply Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp]: Nice one thanks 👍

File: Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp (hunk 2 above)
Comment (greerdv) [reply Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp]: The expected value calculation here is what you would get if the simulation was perfect. The discrepancy is coming about because the simulation is stepping time in discrete intervals. When I implemented the character gameplay component I used a pretty naive integration scheme, which takes a snapshot of the velocity at a timestep and applies that over a whole timestep. It looks kind of like this

image

The upshot of that is that there is a discrepancy in the distance travelled which is roughly equal to half a time step's worth of motion. Because the distance fallen is quadratic, and the total time here is pretty small, that difference is large compared to the distance fallen. It works out that the relative error is about equal to the timestep divided by the total time, so for this example, where the total time is 1/6s and the timestep is 1/60s, you get a relative error of (1/60)/(1/6) = 1/10 so about 10% error. If you simulated for longer e.g. 2s, you would get a relative error (1/60)/2 = 1/120. In that case the distance fallen would be a bit less than 20m, and the tolerance would need to be about 0.2m (about 1%).

I'd still much rather have some idea of how far it should have fallen feed into the expectation. If you just expect it to be lower than it started, you could pass the test whether you'd fallen 1mm or 1km.

Comment (FuzzyCarterAWS) [reply Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp]: Based on @greerdv's feedback, I've adjusted the calculations to compensate for the discrete intervals vs continuous and was able to set a much tighter tolerance. 

This also inspired me to use use a bit of variability in the number of time steps taken to test some minor bounds of time that are reasonable for a unit scale test.

File: Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp [hunk 7]
```diff
@@ -124,4 +141,35 @@ namespace PhysX
...
+    TEST_F(PhysXDefaultWorldTest, CharacterGameplayController_EntityFallsUnderGravity)
+    {
+        GameplayTestBasis basis(m_testSceneHandle, DefaultGravityMultiplier, DefaultGroundDetectionBoxHeight, DefaultFloorTransform);
+
+        // Remove the Ground so we can can test falling without worrying about accidental collisions with objects.
+        basis.DeleteGround();        
+        
+        // Let scene run for a few moments so the entity can be manipulated by gravity from the gameplay component
+        const auto startPosition = basis.m_controller->GetPosition();
+        const int timeStepCount = rand() % 180 + 10;
```
moraaar, line 154: Please avoid using random values in the test. The unit tests need to be deterministic and use the same values to determine if when the code changes will still pass the test or not. Adding randomness eliminates that certainty since different executions will use different values, which potentially can pass on some values and fail in another.

If testing with different amount of time steps is relevant then I would recommend choosing a deliberate amount of values and always test them all. You can use the following approach to run the same test with different values:

````
// Use the PhysXDefaultWorldTestWithParam class for the fixture to indicate the parameter type is going to be int
using PhysXDefaultWorldTestWithParamFixture= PhysXDefaultWorldTestWithParam<int>;

// Use the new PhysXDefaultWorldTestWithParamFixture fixture.
TEST_F(PhysXDefaultWorldTestWithParamFixture, CharacterGameplayController_EntityFallsUnderGravity)
    {
        GameplayTestBasis basis(m_testSceneHandle, DefaultGravityMultiplier, DefaultGroundDetectionBoxHeight, DefaultFloorTransform);

        // Remove the Ground so we can can test falling without worrying about accidental collisions with objects.
        basis.DeleteGround();        

        // Let scene run for a few moments so the entity can be manipulated by gravity from the gameplay component
        const auto startPosition = basis.m_controller->GetPosition();
        const int timeStepCount = GetParam(); // <-- Gets the current value to use.

       ...
}

// Indicate the values to call all the tests that use PhysXDefaultWorldTestWithParamFixture.
INSTANTIATE_TEST_CASE_P(PhysXDefaultWorldTest, PhysXDefaultWorldTestWithParamFixture, ::testing::Values(10, 30, 60, 90, 120, 136, 190));
````

Comment (hultonha) [reply Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp:154]: +1 to @moraaar's comment, though one consideration is it might be okay to use random so long as you use a fixed seed (so the same random numbers are produced each run). If you do want to do this it's also best to avoid `rand` and use `uniform_int_distribution` and a `random_device` (see https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution for more info)

Comment (FuzzyCarterAWS) [reply Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp:154]: Thanks for this. After previous discussions about the relative error in the distance calculations, it became clear that a few different times should be used.

I'm going to use the approach that @moraaar has provided to strengthen this test.

File: Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp [hunk 8]
```diff
@@ -151,7 +154,7 @@ namespace PhysX
         
         // Let scene run for a few moments so the entity can be manipulated by gravity from the gameplay component
         const auto startPosition = basis.m_controller->GetPosition();
-        const int timeStepCount = rand() % 180 + 10;
+        const int timeStepCount = GetParam(); // <-- Gets the current value to use
```
moraaar, line 157: nit: Please remove the comment, I put them in the feedback to remark the changes for you :)

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
```diff
@@ -164,12 +167,15 @@ namespace PhysX
...
+        const auto relativeError = timeStep / totalTime;
         // calculate distance fallen (d = 0.5 * g * t^2)
         const float relativeDistanceFell = 0.5f * gravity.GetZ() * (totalTime * totalTime);
         const float calculatedDistanceFell = relativeDistanceFell - (relativeDistanceFell * relativeError);
 
-        EXPECT_THAT(endPosition.GetZ(), testing::FloatNear((startPosition.GetZ() + calculatedDistanceFell), 0.0001f));
+        EXPECT_THAT(endPosition.GetZ(), testing::FloatNear((startPosition.GetZ() + calculatedDistanceFell), 0.001f));
     }
 
+    // Indicate the values to call all the tests that use PhysXDefaultWorldTestWithParamFixture.
```
moraaar, line 178: nit: Please remove the comment, I put them in the feedback to remark the changes for you :)

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
```diff
@@ -98,6 +98,14 @@ namespace PhysX
             }
         }
 
+        /**
+        * Function to delete the ground for tests where the ground can interfere with the intent of the test. 
+        */
```
hultonha, line 103: nit: Please prefer `//!` style comments now (multi-line ones are harder to maintain and more fiddly to write)

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
```diff
@@ -98,6 +98,14 @@ namespace PhysX
...
+        void DeleteGround()
```
hultonha, line 104: nit: Potentially prefer 'Destroy' or 'Remove' ground (as 'delete' has different connotations in C++)

hultonha (approved):
LGTM (and +1 to @moraaar's comments before merging). Nice work @FuzzyCarterAWS 👍

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
FuzzyCarterAWS: Removed

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
FuzzyCarterAWS: changed

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
FuzzyCarterAWS: Updated to RemoveGround

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
FuzzyCarterAWS: Removed

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
hultonha: Thank you! 🙂

PR #14544, opened 2023-02-11
Enable the ability to create profile builds that use pak files

mbalfour-amzn:
## What does this PR do?

Non-release builds weren't able to run correctly with pak files due to some of the different ways they loaded and handled registry settings. The existence and usage of user registry settings changed which directories were searched for pak files, causing the workflow to fail.

This PR fixes the problem by turning the deviating behavior into a compile flag so that a user can optionally compile a profile build to behave the same way as a release build (and vice versa). 

The primary change is the addition of the optional "ALLOW_SETTINGS_REGISTRY_DEVELOPMENT_OVERRIDES" compile flag that can be set to 0 (allow no overrides) to 3 (allow all overrides). If the flag isn't set, then all overrides are used in non-release builds, and no overrides are used in release builds. This was created as a compile setting instead of a runtime flag because it shouldn't exist to be hacked in releasable builds and because the flag itself literally controls whether or not runtime flags can get used. :) 

A few other changes were also necessary to support this:
* The Settings Registry builder that produces the bootstrap.game.*.setreg file used by the launcher was merging in user settings for non-release setreg files, which is undesirable, since user settings should only affect the local user and not end up in baked data. The user settings are still applied for non-release builds of the Launcher by the GameApplication code, so they don't need to get baked in.
* ComponentApplication and GameApplication had unintentionally deviated slightly in their logic for merging registry settings. This change aligns the two to use exactly the same logic except for GameApplication also merging in the bootstrap.game.*.setreg file.
* The archive code would assert if a pak file tried to open before the serialization code was initialized. This removes the assert, since it could validly happen from the GameApplication constructor. The pak file just gets deferred to open later when that happens.

## How was this PR tested?

* Ran a regular profile build without the setting and verified it worked the same as before.
* Ran a regular release build without the setting and verified it worked the same as before.
* Ran a profile build with the setting set to 0 and verified that it could work with pak files, where before it couldn't.


Code/Framework/AzCore/AzCore/Component/ComponentApplication.h
```diff
@@ -392,5 +402,19 @@ namespace AZ
...
+            None,                       // 0 = no overrides are allowed
+            CommandLineOnly,            // 1 = registry overrides are allowed from the command line
+            CommandLineAndProject,      // 2 = registry overrides are allowed from the command line, engine, gem, and project files
+            CommandLineProjectAndUser   // 3 = registry overrides are allowed from the command line, engine, gem, project, and user files
+        };
+
+        //! Determines which development settings (user registry files, project registry files, etc) should
+        //! be allowed to override the default settings.
+        //! @return the level of development settings overrides that are allowed
+        static constexpr DevelopmentSettingsOverrides GetDevelopmentSettingsOverrides();
```
lumberyard-employee-dm, line 418: We can move the enum to be defined in the `ComponentApplication.cpp` as well as remove the declaration of `GetDevelopmentSettingsOverrides` in the class declaration.

Since we want `GetDevelopmentSettingsOverrides` to be constexpr it would make sense to be only inline in the cpp file

Code/Framework/AzCore/CMakeLists.txt
```diff
@@ -20,6 +20,25 @@ if(PAL_TRAIT_PROF_PIX_SUPPORTED)
...
+  If unset, development overrides are used in development builds and ignored in release builds.
+  0 disables the development overrides in all builds.
+  1 = registry overrides are allowed from the command line.
+  2 = registry overrides are allowed from the command line, engine, gem, and project files.
+  3 = registry overrides are allowed from the command line, engine, gem, project, and user files.")
+
+# If there's a value in ALLOW_SETTINGS_REGISTRY_DEVELOPMENT_OVERRIDES, the compiler flag will get set to it.
+# Otherwise, the variable will be empty and no compiler flag will be set, leaving it to the code to decide
+# whether or not to allow the Settings Registry development overrides.
+set(ALLOW_SETTINGS_REGISTRY_DEVELOPMENT_OVERRIDES_FLAG $<$<NOT:$<STREQUAL:"${ALLOW_SETTINGS_REGISTRY_DEVELOPMENT_OVERRIDES}","">>:ALLOW_SETTINGS_REGISTRY_DEVELOPMENT_OVERRIDES=$<BOOL:${ALLOW_SETTINGS_REGISTRY_DEVELOPMENT_OVERRIDES}>>)
```
lumberyard-employee-dm, line 38: In this case we don't want to convert the Cache variable value into a boolean here, so we should remove the `$<BOOL: ` generator expression
```suggestion
set(ALLOW_SETTINGS_REGISTRY_DEVELOPMENT_OVERRIDES_FLAG $<$<NOT:$<STREQUAL:"${ALLOW_SETTINGS_REGISTRY_DEVELOPMENT_OVERRIDES}","">>:ALLOW_SETTINGS_REGISTRY_DEVELOPMENT_OVERRIDES=${ALLOW_SETTINGS_REGISTRY_DEVELOPMENT_OVERRIDES}>)
```

Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp
```diff
@@ -923,24 +939,27 @@ namespace AZ
...
+            SettingsRegistryMergeUtils::MergeSettingsToRegistry_O3deUserRegistry(
+                registry, AZ_TRAIT_OS_PLATFORM_CODENAME, specializations, &scratchBuffer);
+            SettingsRegistryMergeUtils::MergeSettingsToRegistry_CommandLine(registry, m_commandLine, false);
+            // Project User Registry is merged after the command line here to allow make sure the any command line override of the project
+            // path is used for merging the project's user registry
+            SettingsRegistryMergeUtils::MergeSettingsToRegistry_ProjectUserRegistry(
+                registry, AZ_TRAIT_OS_PLATFORM_CODENAME, specializations, &scratchBuffer);
+            SettingsRegistryMergeUtils::MergeSettingsToRegistry_CommandLine(registry, m_commandLine, false);
+            SettingsRegistryMergeUtils::MergeSettingsToRegistry_AddRuntimeFilePaths(registry);
+        }
```
lumberyard-employee-dm, line 961: We should provide similar logic for making sure the command line is merged as we do in MergeUserSettings, when we the override settings are `CommandLineOnly` or `CommandLineAndProject`

We can check if the `GetDevelopmentSettingOverrides()` enum is greater than or equal to the `CommandLineOnly` option to enter the block below and then add an inner check to merge the o3de user registry and project user registry if the value is `CommandLineProjectAndUser`
```suggestion
        if constexpr (GetDevelopmentSettingsOverrides() >= DevelopmentSettingsOverrides::CommandLineOnly)
        {
            // In development builds apply the o3de registry and the command line to allow early overrides. This will
            // allow developers to override things like default paths or Asset Processor connection settings. Any additional
            // values will be replaced by later loads, so this step will happen again at the end of loading.
            if constexpr (GetDevelopmentSettingsOverrides() == DevelopmentSettingsOverrides::CommandLineProjectAndUser)
            {
                SettingsRegistryMergeUtils::MergeSettingsToRegistry_O3deUserRegistry(
                    registry, AZ_TRAIT_OS_PLATFORM_CODENAME, specializations, &scratchBuffer);
                SettingsRegistryMergeUtils::MergeSettingsToRegistry_CommandLine(registry, m_commandLine, false);
            }
                // Project User Registry is merged after the command line here to allow make sure the any command line override of the project
                // path is used for merging the project's user registry
                if constexpr (GetDevelopmentSettingsOverrides() == DevelopmentSettingsOverrides::CommandLineProjectAndUser)
                {
                    SettingsRegistryMergeUtils::MergeSettingsToRegistry_ProjectUserRegistry(
                        registry, AZ_TRAIT_OS_PLATFORM_CODENAME, specializations, &scratchBuffer);
                }
                
            // Make sure the command line is merged at least once, before updating the runtime filepaths
            SettingsRegistryMergeUtils::MergeSettingsToRegistry_CommandLine(registry, m_commandLine, false);
            SettingsRegistryMergeUtils::MergeSettingsToRegistry_AddRuntimeFilePaths(registry);
        }

```

cgalvan (approved):
Thanks for doing this! Will be super helpful for debugging builds that use pak files.

lumberyard-employee-dm (approved):
Looks good!

PR #14547, opened 2023-02-11
For Android, create a public, common bufferpool object for Passes/SRG…

Jackie9527:

## What does this PR do?

When using vulkan as backend, for each pass RHI creates a bufferpool object to allocate constant memory blocks.
If RPI creates N pass, such as 70+, then it could consume up to 70 * 16 MB = 1120 MB, which is unacceptable for Android apps.

srg-bufferpool-11
srg-bufferpool-11-1

## How was this PR tested?
srg-bufferpool-04

For windows, set name for each bufferpool object created for Shader Resource Group.
srg-bufferpool-02
srg-bufferpool-02-1

For Android, set name for the common bufferpool object, and share a single bufferpool object for all the Shader Resource Group objects.

srg-bufferpool-03-1

### DIFF
--- a/Gems/Atom/RHI/Vulkan/Code/Source/RHI/ShaderResourceGroupPool.cpp
+++ b/Gems/Atom/RHI/Vulkan/Code/Source/RHI/ShaderResourceGroupPool.cpp
@@ -34,16 +34,7 @@ namespace AZ
 
             m_descriptorSetCount = RHI::Limits::Device::FrameCountMax;
 
-            const uint32_t constantSize = layout.GetConstantDataSize();
-            if (constantSize > 0)
-            {
-                BufferPoolDescriptor bufferPoolDescriptor;
-                bufferPoolDescriptor.m_bindFlags = RHI::BufferBindFlags::Constant;
-                bufferPoolDescriptor.m_heapMemoryLevel = RHI::HeapMemoryLevel::Host;
-                m_constantBufferPool = BufferPool::Create();
-                m_constantBufferPool->SetName(AZ::Name(AZStd::string::format("%s_ConstantBufferPool", GetName().GetCStr())));
-                m_constantBufferPool->Init(device, bufferPoolDescriptor);
-            }
+            m_constantBufferPool = device.GetConstantBufferPool();
 
             DescriptorSetLayout::Descriptor layoutDescriptor;
             layoutDescriptor.m_device = &device;
--- a/Gems/Atom/RHI/Vulkan/Code/Source/RHI.Reflect/BufferPoolDescriptor.cpp
+++ b/Gems/Atom/RHI/Vulkan/Code/Source/RHI.Reflect/BufferPoolDescriptor.cpp
@@ -27,7 +27,16 @@ namespace AZ
 
         BufferPoolDescriptor::BufferPoolDescriptor()
         {
-            m_bufferPoolPageSizeInBytes = RHI::RHISystemInterface::Get()->GetPlatformLimitsDescriptor()->m_platformDefaultValues.m_bufferPoolPageSizeInBytes;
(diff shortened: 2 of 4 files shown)


Gems/Atom/RHI/Vulkan/Code/Source/RHI/ShaderResourceGroupPool.cpp
```diff
@@ -34,15 +34,21 @@ namespace AZ
 
             m_descriptorSetCount = RHI::Limits::Device::FrameCountMax;
 
+#if !defined(AZ_PLATFORM_ANDROID)
```
rgba16f, line 37: Why not do this for all vulkan platforms?

rgba16f (approved):
Change looks good to me, only question is why limit it to Android?

Gems/Atom/RHI/Vulkan/Code/Source/RHI/ShaderResourceGroupPool.cpp
```diff
@@ -34,15 +34,21 @@ namespace AZ
...
             {
                 BufferPoolDescriptor bufferPoolDescriptor;
                 bufferPoolDescriptor.m_bindFlags = RHI::BufferBindFlags::Constant;
                 bufferPoolDescriptor.m_heapMemoryLevel = RHI::HeapMemoryLevel::Host;
                 m_constantBufferPool = BufferPool::Create();
+                static int index = 0;
+                m_constantBufferPool->SetName(Name(AZStd::string::format("SRG_%s_%d", layout.GetName().GetCStr(), ++index)));
                 m_constantBufferPool->Init(device, bufferPoolDescriptor);
             }
+#else
```
moudgils, line 49: Platform specific code can not go in like this. You have to sue the PAL layer. Take a look at C:\GitHub\o3de\Gems\Atom\RHI\Vulkan\Code\Include\Platform to get better idea. Also I think we can just adopt a single allocator for all platforms. Dont need any sort of platform specific code for SRG constants.

moudgils (changes requested):
See my comments below

Gems/Atom/RHI/Vulkan/Code/Source/RHI/ShaderResourceGroupPool.cpp
rgba16f: This use of the AZ_PLATFORM_* macro is also causing the change to fail vaildation, I believe we want platform specific code to be moved over to the platform abstraction layer. 
Platform specific code for the vulkan RHI is placed here - Gems/Atom/RHI/Vulkan/Code/Include/Platform & Gems/Atom/RHI/Vulkan/Code/Source/Platform.

Gems/Atom/RHI/Vulkan/Code/Source/RHI/ShaderResourceGroupPool.cpp
Jackie9527: done.
`AZ_PLATFORM_ANDROID` has been removed.

Gems/Atom/RHI/Vulkan/Code/Source/RHI/ShaderResourceGroupPool.cpp
Jackie9527: done.
`AZ_PLATFORM_ANDROID` has been removed.

rgba16f (approved):

rgba16f:
Automated Review run started: https://jenkins.build.o3de.org/job/O3DE/view/change-requests/job/PR-14547/1/

VickyAtAZ (approved):

moudgils (approved):

PR #14549, opened 2023-02-13
[ShaderManagementConsole] List all material which use the selected option

etherealchain:

Add a function in SMC, by double-clicking on the option name, a window will pop up to show a list of all the material that is using this option.

image

## How was this PR tested?

Test locally on my pc.


gadams3 (approved):
The implementation looks good but I think invoking this could be moved somewhere more obvious or discoverable than double clicking the option header. Clicking or double clicking the header would usually be reserved for something like sorting. Users might accidentally trigger this behavior and not understand why it happened.

I recommend changing this to allow right clicking anywhere in the column, opening a context menu, and giving the specific options for operations that can be performed. Then invoke this from that context menu. 

It's not required but I also recommend moving this to a Python script since the majority of the shader variant enumeration is done from Python and there will soon be other factors besides materials. Then in the context menu good potentially make other scripts available. Scripts could also be registered and run from the main menu if functions are added to retrieve the currently selected option.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleTableView.cpp
```diff
@@ -26,13 +29,15 @@ namespace ShaderManagementConsole
         setSelectionBehavior(QAbstractItemView::SelectItems);
         setSelectionMode(QAbstractItemView::SingleSelection);
         horizontalHeader()->setSectionResizeMode(QHeaderView::ResizeToContents);
+        connect(horizontalHeader(), &QHeaderView::sectionDoubleClicked, this, &ShaderManagementConsoleTableView::OnOptionClicked);
```
galibzon, line 32: I'd agree here with the comment from @gadams3 , this should be a right-click with a popup menu. Also the dialog box should show the material count total.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleTableView.cpp
```diff
@@ -241,6 +246,79 @@ namespace ShaderManagementConsole
...
+            materialAssetIdList,
+            &ShaderManagementConsoleRequestBus::Events::GetAllMaterialAssetIds);
+
+        QString materialList = "";
+
+        QProgressDialog progressDialog("Gather shader variant information...", "Cancel", 0, materialAssetIdList.size());
+        progressDialog.setWindowModality(Qt::WindowModal);
+        progressDialog.setMaximumWidth(400);
+        progressDialog.setMaximumHeight(100);
+        progressDialog.setWindowTitle("Gather information from material assets");
```
cgalvan, line 265: Couple of suggestions here:
- Any user facing strings should be wrapped in `tr()` so they can be translated in the future
- The default cancel button text is "Cancel", it doesn't need to be specified
- Any modal dialogs should get parented to the currently active window instead of `nullptr` so they get positioned properly, which we have a nice helper method for
- The default minimum is already 0 so we don't need to specify it

```suggestion
        QProgressDialog progressDialog(AzToolsFramework::GetActiveWindow());
        progressDialog.setWindowModality(Qt::WindowModal);
        progressDialog.setMaximum(materialAssetIdList.size());
        progressDialog.setMaximumWidth(400);
        progressDialog.setMaximumHeight(100);
        progressDialog.setWindowTitle(tr("Gather information from material assets"));
        progressDialog.setLabelText(tr("Gather shader variant information..."));
```

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleTableView.cpp
```diff
@@ -241,6 +246,79 @@ namespace ShaderManagementConsole
...
+            {
+                return;
+            }
+        }
+
+        progressDialog.close();
+
+        if (materialList != "")
+        {
+            QMessageBox msgBox;
```
cgalvan, line 309: Same parenting advice as above
```suggestion
            QMessageBox msgBox(AzToolsFramework::GetActiveWindow());
```

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleTableView.cpp
```diff
@@ -241,6 +246,79 @@ namespace ShaderManagementConsole
...
+            QString message = QString("The materials which used %1 are listed here.").arg(shaderOptionDescriptor.GetName().GetCStr());
```
cgalvan, line 310:
```suggestion
            QString message = QString(tr("The materials which used %1 are listed here.")).arg(shaderOptionDescriptor.GetName().GetCStr());
```

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleTableView.cpp
```diff
@@ -241,6 +246,79 @@ namespace ShaderManagementConsole
...
+            msgBox.setText(message);
+            msgBox.setDetailedText(materialList);
+            msgBox.exec();
+        }
+        else
+        {
+            QMessageBox msgBox;
```
cgalvan, line 317:
```suggestion
            QMessageBox msgBox(AzToolsFramework::GetActiveWindow());
```

cgalvan (changes requested):
Some minor feedback + the suggestion from @gadams3 , but overall this looks great

etherealchain:
I want to confirm the requirement here, when you generate a shader variant list from script:
image

Looks about right? Did I miss anything?

etherealchain:
Actually, maybe we shouldn't play with column this way, since there are containers indexed with column.
How about a tooltip like this?
image

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleTableView.cpp
gadams3: You might have to use the atom tools framework utilities version of this function

etherealchain:
Wouldn't that result in a lot of repetition in UI? Since most of the option is boolean. Something like this
image
And what if you change the option value in SMC? The number should be updated right away?

etherealchain:
It would be simpler to subclass `QTableWidgetItem` and display the data as we want it, instead of adding columns for each option.

etherealchain:
Another question, when you run the script `GenerateShaderVariantListForMaterials.py`, it only collects the material that is using the .shader you selected, but it's possible a material that is not using the shader still using one of the options in the list.
Should the counting only operate on the material collected (i.e. the materials that are related to the shader you selected), or it should operate on all materials in the code base?

galibzon:
> Wouldn't that result in a lot of repetition in UI? Since most of the option is boolean. Something like this image And what if you change the option value in SMC? The number should be updated right away?

It's ok to see all those values. They are not always the same and that's the point... to be able to spot the differences.

galibzon:
> Another question, when you right-click on .shader to run the script `GenerateShaderVariantListForMaterials.py`, it only collects the material that is using the .shader you selected, but it's totally possible a material that is not using the shader still using one of the options in the list. Should the counting only operate on the material collected (i.e. the materials that are related to the shader you selected), or it should operate on all materials in the code base?

We only care for Materials that refer to the shader, and for which the shader references the option. In the end if an option is not referenced by a shader, then that option should not influence how many variants we are going to create for that shader.

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Document/AtomToolsDocumentMainWindow.cpp
```diff
@@ -277,7 +277,9 @@ namespace AtomToolsFramework
 
             if (!result)
             {
-                return false;
+                // Currently, result will only be false when a view doesn't have document, e.g. ShaderManagementConsoleStatisticView
+                // We can just close the tab here
+                RemoveDocumentTab(documentId);
             }
```
gadams3, line 283: These changes are not needed. See comments about a global view and assumptions about how documents are handled.

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Document/AtomToolsDocumentSystem.cpp
```diff
@@ -297,7 +297,13 @@ namespace AtomToolsFramework
...
         }
 
-        DestroyDocument(documentId);
+        if(!DestroyDocument(documentId))
+        {
+            // documentId doesn't exist in m_documentMap
+            // This can happen when a view doesn't have document, e.g. ShaderManagementConsoleStatisticView
+            return false;
+        }
+
```
gadams3, line 306: These changes are not needed. See comments about a global view and assumptions about how documents are handled.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleWindow.cpp
```diff
@@ -92,6 +92,100 @@ namespace ShaderManagementConsole
...
+            progressDialog.setValue(i);
+
+            if (progressDialog.wasCanceled())
+            {
+                return;
+            }
+        }
+        progressDialog.close();
+
+        AddDocumentTab(AZ::Uuid::CreateRandom(), new ShaderManagementConsoleStatisticView(statisticData, this));
```
gadams3, line 187: Is this supposed to create a view of statistics for all materials in the entire project or just statistics for materials using options in the currently loaded shader variant list?

If this is supposed to be a global view, not directly connected with a document, then I would recommend not adding a dummy document to register the view with. It does not need to be tabbed with the other documents. Instead, you can just create a regular, free floating or docked window.

If this is supposed to focus on statistics relative to the open shader variant list document, then I would instead create and register a tab or stack widget with a table view and statistics view as child pages, and register that as the document view. That way everything related to the document is packaged together.

gadams3 (changes requested):
Main change requested is to revert the changes to the base document window and system to not affect assumptions made in other places about document existence or validity. if it is a global view then create a floating or docked window since that's not really a document tab on its own.

gadams3 (approved):

galibzon (approved):

cgalvan (approved):

PR #14555, opened 2023-02-13
add prompt when overwriting existing files.

Jackie9527:

## What does this PR do?

When saving the UI canvas file with an existed filename, UI editor will directly overwrite the existing file without prompting.

## How was this PR tested?

1. Create a1.uicanvas, and save to disk.
2. Create another uicanvas file, using filename as `a1.uicanvas`, and Click Save.
3. UI Editor will prompt if user want to overwrite the existing file.

### DIFF
--- a/Gems/LyShine/Code/Editor/EditorWindow.cpp
+++ b/Gems/LyShine/Code/Editor/EditorWindow.cpp
@@ -733,8 +733,7 @@ bool EditorWindow::SaveCanvasToXml(UiCanvasMetadata& canvasMetadata, bool forceA
             QString(),
             dir,
             "*." UICANVASEDITOR_CANVAS_EXTENSION,
-            nullptr,
-            QFileDialog::DontConfirmOverwrite);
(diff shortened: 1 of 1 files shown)


AMZN-daimini (approved):

Gems/LyShine/Code/Editor/EditorWindow.cpp
```diff
@@ -733,8 +733,7 @@ bool EditorWindow::SaveCanvasToXml(UiCanvasMetadata& canvasMetadata, bool forceA
             QString(),
             dir,
             "*." UICANVASEDITOR_CANVAS_EXTENSION,
-            nullptr,
-            QFileDialog::DontConfirmOverwrite);
+            nullptr);
```
cgalvan, line 736: Nit: Can omit the last `nullptr` since the `selectedFilter` param has a default value of `nullptr`.

cgalvan (approved):

PR #14560, opened 2023-02-13
Model import support transparent material.

Zhang-Baochong:
Signed-off-by: Zhang-Baochong <zhangbaochong@huawei.com>

## What does this PR do?

Import model with transparent material, the material is opaque in o3de editor.

## How was this PR tested?
Import model with transparent material, check if it is transparent in editor.

### DIFF
--- a/Gems/Atom/Feature/Common/Code/Source/Material/MaterialConverterSystemComponent.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/Material/MaterialConverterSystemComponent.cpp
@@ -159,6 +159,10 @@ namespace AZ
             }
 
             sourceData.SetPropertyValue(Name{"opacity.factor"}, materialData.GetOpacity());
+            if (1.0f - materialData.GetOpacity() > AZ::Constants::FloatEpsilon)
+            {
+                sourceData.SetPropertyValue(Name{ "opacity.mode" }, AZStd::string("Blended"));
+            }
 
             auto applyOptionalPropertiesFunc = [&sourceData, &anyPBRInUse](const auto& propertyGroup, const auto& propertyName, const auto& propertyOptional)
             {
@@ -182,6 +186,14 @@ namespace AZ
             if (roughness.has_value())
             {
                 sourceData.SetPropertyValue(Name{"roughness.factor"}, roughness.value());
+            }
+            else if (materialData.GetShininess() > AZ::Constants::FloatEpsilon)
+            {
+                // When the MaterialData provides Shininess instead of Roughness, it is necessary to convert Shininess to Roughness.
+                // Normalized Blinn-Phong: D_p(m)=((alpha_p+2)/(2*PI))dot(n, m)^alpha_p, Usually use formula: alpha_p=2*alpha^-2 - 2,
+                // alpha = roughness^2
+                sourceData.SetPropertyValue(
+                    Name{ "roughness.factor" }, aznumeric_cast<float>(pow(2.0f / (materialData.GetShininess() + 2.0f), 0.25)));
             }
 
             handleTexture("emissive", "textureMap", SceneAPI::DataTypes::IMaterialData::TextureMapType::Emissive);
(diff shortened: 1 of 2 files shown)


AMZN-daimini (approved):

Gems/Atom/Feature/Common/Code/Source/Material/MaterialConverterSystemComponent.cpp
```diff
@@ -183,6 +187,13 @@ namespace AZ
                 sourceData.SetPropertyValue(Name{"roughness.factor"}, roughness.value());
             }
 
+            if (!materialData.GetRoughnessFactor().has_value() && abs(materialData.GetShininess()) > AZ::Constants::FloatEpsilon)
+            {
+                // When the MaterialData provides Shininess instead of Roughness, it is necessary to convert Shininess to Roughness.
```
gadams3, line 192:
```suggestion
            }
            else if (abs(materialData.GetShininess()) > AZ::Constants::FloatEpsilon)
            {
                // When the MaterialData provides Shininess instead of Roughness, it is necessary to convert Shininess to Roughness.
```

Gems/Atom/Feature/Common/Code/Source/Material/MaterialConverterSystemComponent.cpp
```diff
@@ -183,6 +187,13 @@ namespace AZ
...
+                sourceData.SetPropertyValue(
+                    Name{ "roughness.factor" }, aznumeric_cast<float>(pow(2 / (materialData.GetShininess() + 2), 0.25)));
```
amzn-tommy, line 194: Can you add a comment explaining why this is the correct conversion from shininess to roughness?

nitpick: should probably use 2.0 instead of 2 since they are not integers here.

Gems/Atom/Feature/Common/Code/Source/Material/MaterialConverterSystemComponent.cpp
```diff
@@ -183,6 +187,13 @@ namespace AZ
                 sourceData.SetPropertyValue(Name{"roughness.factor"}, roughness.value());
             }
 
+            if (!materialData.GetRoughnessFactor().has_value() && abs(materialData.GetShininess()) > AZ::Constants::FloatEpsilon)
```
amzn-tommy, line 190: Whats the abs needed for? Is a negative value valid/expected for shininess?

Gems/Atom/Feature/Common/Code/Source/Material/MaterialConverterSystemComponent.cpp
```diff
@@ -158,6 +158,10 @@ namespace AZ
             }
 
             sourceData.SetPropertyValue(Name{"opacity.factor"}, materialData.GetOpacity());
+            if (abs(materialData.GetOpacity() - 1.0f) > AZ::Constants::FloatEpsilon)
```
amzn-tommy, line 161: Why not just use 1.0 - opacity?

Zhang-Baochong: the value of opacity between 0 to 1.0f, its ok to use `1.0 - opacity`.

Gems/Atom/Feature/Common/Code/Source/Material/MaterialConverterSystemComponent.cpp
Zhang-Baochong: there is no clear cut interpretation/conversion for this. 
Normalized Blinn-Phong: D_p(m)=((alpha_p+2)/(2*PI))dot(n,m)^alpha_p
Usually use formula (UE4 use this): alpha_p=2*alpha^-2 - 2, alpha=roughness^2

Gems/Atom/Feature/Common/Code/Source/Material/MaterialConverterSystemComponent.cpp
Zhang-Baochong: changed.

PR #14564, opened 2023-02-13
Adding pytest metrics script

evanchia-ly-sdets:

## What does this PR do?
Adds a script to convert pytest xml metrics to csv. This will be added to the internal pipeline for testing.

## How was this PR tested?
Tested by running script locally.


scripts/metrics/tests/test_pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,43 @@
...
+
+Unit tests for metrics_xml_to_csv.py
+"""
+import unittest
+import unittest.mock as mock
+
+import pytest_metrics_xml_to_csv
+
+
+class TestMetricsXMLtoCSV(unittest.TestCase):
```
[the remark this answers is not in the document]
evanchia-ly-sdets: Added more unit tests

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,159 @@
...
+
+# Setup logging.
+logger = logging.get_logger("test_metrics")
+logging.setup_logger(logger)
+
+# Create the csv field header
+PYTEST_FIELDS_HEADER = [
+    'test_name',
+    'status',
+    'duration_seconds',
```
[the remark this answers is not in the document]
evanchia-ly-sdets: In my opinion, seconds are more readable than ms on average. If there are tests in the minutes, it becomes difficult to read if in ms.

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,159 @@
...
+    'sig_owner'
```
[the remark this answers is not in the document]
evanchia-ly-sdets: That's a possible solution. I'm pretty sure we can input custom properties via marks or something that will appear in the pytest xml. That would require manual marking of tests, or we would need to add an automated marking feature.

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,159 @@
...
+    """
+    Inspects the test case node and determines the test result based on the presence of known element
+    names such as 'error', 'failure' and 'skipped'. If the elements are not found, the test case is assumed
+    to have passed. This is how the JUnit XML schema is generated for failed tests.
+
+    :param test_case_node: The element node for the test case.
+    :return: The test result
+    """
+    # Mapping from JUnit elements to test result to keep it consistent with CTest result reporting.
+    failure_elements = [('error', 'failed'), ('failure', 'failed'), ('skipped', 'skipped')]
```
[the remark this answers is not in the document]
evanchia-ly-sdets: Skip is not counted as "failed", it is counted as "skipped"

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,159 @@
...
+        for test in xml_root.findall('./testsuite/testcase'):
+            # There are some testcase fields that are not tests. They do not have 'name' attributes are skipped
+            if 'name' not in test.attrib:
+                logger.debug(f'We found a testcase where it did not have a test field. Printing attribs:\n'
+                             f'{test.attrib}')
+                continue
+            test_data_dict['test_name'] = test.attrib['name']
+            test_data_dict['duration_seconds'] = float(test.attrib['time'])
+            # using 'status' to keep it consistent with CTest xml schema
+            test_data_dict['status'] = _determine_test_result(test)
```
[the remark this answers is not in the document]
evanchia-ly-sdets: I'm collected all entries for the moment for testing purposes. I'm intending to remove the pass data in the future.

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,159 @@
...
+            # Index 1 is the sig owner, replace slashes to match codeowners file
+            sig_owner = codeowners_hint.get_codeowners(test.attrib['file'].replace("\\", "/"))[1]
```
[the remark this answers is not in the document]
evanchia-ly-sdets: I could cache some values to improve performance when obtaining sig owners. Might be cleaner than trying to relate CTest to their respective files.

scripts/metrics/pytest_metrics_xml_to_csv.py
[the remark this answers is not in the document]
Kadino: Yeah, if we want to enable that metrics feature, we will need to update to the way CTest and its sub-runners record their artifacts.

scripts/metrics/pytest_metrics_xml_to_csv.py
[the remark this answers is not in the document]
Kadino: Shame we don't yet have relation of individual test to CTest module or the path to the CMake file that registers the module. IMO it's ideal for tests to rely on the module owner, and then redundant data doesn't need to be calculated or stored per test.

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,166 @@
...
+            # using 'status' to keep it consistent with CTest xml schema
+            test_data_dict['status'] = _determine_test_result(test)
+            # replace slashes to match codeowners file
+            test_file_path = test.attrib['file'].replace("\\", "/")
+            try:
+                sig_owner = SIG_OWNER_CACHE[test_file_path]
+            except KeyError:
+                # Index 1 is the sig owner
+                sig_owner = codeowners_hint.get_codeowners(test_file_path)[1]
+                SIG_OWNER_CACHE[test_file_path] = sig_owner
```
Kadino, line 157: minor: python typically recommends ignoring tuple return values with an underscore variable `_, sig_owner, _ = codeowners_hint.get_codeowners(test_file_path)`

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,166 @@
...
+                continue
+            test_data_dict['test_name'] = test.attrib['name']
+            test_data_dict['duration_seconds'] = float(test.attrib['time'])
+            # using 'status' to keep it consistent with CTest xml schema
+            test_data_dict['status'] = _determine_test_result(test)
+            # replace slashes to match codeowners file
+            test_file_path = test.attrib['file'].replace("\\", "/")
+            try:
+                sig_owner = SIG_OWNER_CACHE[test_file_path]
+            except KeyError:
```
Kadino, line 154: Instead of try/except, prefer `if test_file_path in SIG_OWNER_CACHE:`

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,166 @@
...
+                # Index 1 is the sig owner
+                sig_owner = codeowners_hint.get_codeowners(test_file_path)[1]
+                SIG_OWNER_CACHE[test_file_path] = sig_owner
+            test_data_dict['sig_owner'] = sig_owner if sig_owner else "N/A"
+
+            writer.writerow(test_data_dict)
+    except KeyError as exc:
+        logger.exception(f"KeyError when parsing xml file: {full_xml_path}. Check xml keys for changes.", exc)
```
Kadino, line 162: This exception will currently break out of the inner loop, causing data loss. Since it is not reraised, the script will proceed as if the transformation succeeded.

scripts/metrics/pytest_metrics_xml_to_csv.py
evanchia-ly-sdets: Fixed

scripts/metrics/pytest_metrics_xml_to_csv.py
evanchia-ly-sdets: Fixed

scripts/metrics/pytest_metrics_xml_to_csv.py
evanchia-ly-sdets: Fixed exception to inside loop to prevent more data loss.

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,163 @@
...
+    parser.add_argument(
+        "-o", "--output-directory", action="store", default="",
+        help=f"The directory where the csv to be saved. Prepends the --csv-file arg."
+    )
+    parser.add_argument(
+        "-b", "--branch", action="store", default="",
+        help="The branch the metrics were generated on. Used for directory saving."
+    )
+    parser.add_argument(
+        "-w", "--weeks", action="store_true",
```
Kadino, line 96: Why should the locally-saved CSV need this information, isn't it only important for the S3 upload location?

That aside, week being optional is surprising. I think this should automatically store in a specific way, not provide an option. Either stored in YYYY/WW or YYYY/MM/DD (which would have smaller buckets to filter through). Seems extreme to store everything together from an entire month, let alone year.

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,163 @@
...
+        "--csv-file", action="store", default=_get_default_csv_filename(),
+        help=f"The directory and file name for the csv to be saved (defaults to YYYY_MM_DD_hh_mm_pytest)."
+    )
+    parser.add_argument(
+        "-o", "--output-directory", action="store", default="",
+        help=f"The directory where the csv to be saved. Prepends the --csv-file arg."
+    )
+    parser.add_argument(
+        "-b", "--branch", action="store", default="",
+        help="The branch the metrics were generated on. Used for directory saving."
```
Kadino, line 93: Seems like CSV transformation doesn't require this, only the S3 upload should.

scripts/build/Platform/Windows/build_config.json
```diff
@@ -5,18 +5,27 @@
...
       "SCRIPT_PATH": "scripts\\metrics\\ctest_metrics_xml_to_csv.py",
       "SCRIPT_PARAMETERS": "build\\windows --output-directory build\\windows\\Testing\\csv --branch %BRANCH_NAME% -w"
     }
   },
+  "generate_pytest_metrics": {
+    "TAGS": [],
+    "COMMAND": "python_windows.cmd",
+    "PARAMETERS": {
+      "SCRIPT_PATH": "scripts\\metrics\\pytest_metrics_xml_to_csv.py",
+      "SCRIPT_PARAMETERS": "build\\windows\\Testing\\Pytest --output-directory build\\windows\\Testing\\csv --branch %BRANCH_NAME% -w"
```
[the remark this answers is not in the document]
Kadino: More concerned about old XMLs than about old CSVs, but both could be an issue.

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,163 @@
...
+PYTEST_FIELDS_HEADER = [
+    'test_name',
+    'status',
+    'duration_seconds',
+    'sig_owner'
+]
+SIG_OWNER_CACHE = {}
+
+
+def main():
```
LesaelR, line 43: Nitpick: Could add a docstring to explain what is happening here. 
There's also a few other functions that could use docstrings.

LesaelR (approved):
Some minor nitpicks but other than that, LGTM.

scripts/build/Platform/Windows/build_config.json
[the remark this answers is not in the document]
evanchia-ly-sdets: Yes there are already cleanup steps in the pipeline that the other metrics jobs utilize.

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,163 @@
...
+
+SPDX-License-Identifier: Apache-2.0 OR MIT
+
+Scrapes metrics from Pytest xml files and creates csv formatted files.
+"""
+import os.path
+import sys
+import csv
+import argparse
+import xml.etree.ElementTree as xmlElementTree
```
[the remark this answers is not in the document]
evanchia-ly-sdets: I think adding a lambda would overcomplicate the pipeline because the xml files are cleaned up after every pipeline run. We would have to persist the files, trigger a lambda to execute the script, then clean up the old workspace.

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,163 @@
...
+import ly_test_tools.cli.codeowners_hint as codeowners_hint
+from common import logging, exception
+
+
+TESTING_DIR = 'Testing'
+
+
+def _get_default_csv_filename():
+    # Format default file name based off of date
+    now = datetime.datetime.now()
```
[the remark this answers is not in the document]
evanchia-ly-sdets: Fixed

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,163 @@
...
+    full_path = os.path.join(args.output_directory, args.branch)
+    if args.weeks:
+        week = datetime.datetime.now().isocalendar().week
+        full_path = os.path.join(full_path, f"Week{week:02d}")
+    full_path = os.path.join(full_path, args.csv_file)
+    if os.path.exists(full_path):
+        logger.warning(f"The file {full_path} already exists. It will be overridden.")
+    if not os.path.exists(os.path.split(full_path)[0]):
+        # Create directory if it doesn't exist
+        os.makedirs(os.path.split(full_path)[0])
```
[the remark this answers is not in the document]
evanchia-ly-sdets: Fixed

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,163 @@
...
+                parse_pytest_xmls_to_csv(os.path.join(args.xml_folder, filename), writer)
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description='This script assumes that Pytest xml files have been produced. The xml files will be parsed and '
+                    'written to a csv file.')
+    parser.add_argument(
+        'xml_folder',
+        help="Path to where the Pytest xml files live. O3DE CTest defaults this to {build_folder}/Testing/Pytest"
```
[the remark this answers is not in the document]
evanchia-ly-sdets: Positional args are already implicitly required

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,163 @@
...
+    )
+    parser.add_argument(
+        "--csv-file", action="store", default=_get_default_csv_filename(),
+        help=f"The directory and file name for the csv to be saved (defaults to YYYY_MM_DD_hh_mm_pytest)."
+    )
+    parser.add_argument(
+        "-o", "--output-directory", action="store", default="",
```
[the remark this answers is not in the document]
evanchia-ly-sdets: Fixed

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,163 @@
...
+    parser = argparse.ArgumentParser(
+        description='This script assumes that Pytest xml files have been produced. The xml files will be parsed and '
+                    'written to a csv file.')
+    parser.add_argument(
+        'xml_folder',
+        help="Path to where the Pytest xml files live. O3DE CTest defaults this to {build_folder}/Testing/Pytest"
+    )
+    parser.add_argument(
+        "--csv-file", action="store", default=_get_default_csv_filename(),
+        help=f"The directory and file name for the csv to be saved (defaults to YYYY_MM_DD_hh_mm_pytest)."
```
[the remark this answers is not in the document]
evanchia-ly-sdets: Fixed

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,163 @@
...
+from common import logging, exception
+
+
+TESTING_DIR = 'Testing'
+
+
+def _get_default_csv_filename():
+    # Format default file name based off of date
+    now = datetime.datetime.now()
+    return f"{now.year}_{now.month:02d}_{now.day:02d}_{now.hour:02d}_{now.minute:02d}_pytest.csv"
```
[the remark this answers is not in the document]
evanchia-ly-sdets: I will clarify the proposed folder structure saving system which will answer some questions below. It is proposed that csv metrics will be stored in S3 by the following format: csv/<Branch>/<Week>/<metrics>.csv
It is also proposed to not use an uploader script, and instead directly upload the metrics via AWS CLI. The uploading does not handle any folder structure logic and simply only uploads files/folders as is. The metrics script contains this folder structure logic which is why there are branch/weeks options in the script args. These are optional as users can still use the script locally (and for testing locally).

scripts/metrics/pytest_metrics_xml_to_csv.py
evanchia-ly-sdets: See answer above

scripts/metrics/pytest_metrics_xml_to_csv.py
evanchia-ly-sdets: See answer above

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,163 @@
...
+    # Create csv file
+    if os.path.exists(args.csv_file):
+        logger.warning(f"The file {args.csv_file} already exists. It will be overriden.")
+    with open(full_path, 'w', encoding='UTF8', newline='') as csv_file:
+        writer = csv.DictWriter(csv_file, fieldnames=PYTEST_FIELDS_HEADER, restval='N/A')
+        writer.writeheader()
+
+        # Parse Pytest xml's and write to csv file
+        for filename in os.listdir(args.xml_folder):
+            if os.path.splitext(filename)[-1] == '.xml':
```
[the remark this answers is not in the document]
evanchia-ly-sdets: Fixed

scripts/metrics/pytest_metrics_xml_to_csv.py
evanchia-ly-sdets: Fixed

scripts/metrics/pytest_metrics_xml_to_csv.py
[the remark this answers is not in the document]
Kadino: Mostly brought this up as https://github.com/o3de/sig-build/issues/77 mentions moving operations to GitHub actions, which are quite similar to AWS Lambda. But I think we can defer that discussion outside of this PR. Becomes a question of whether persisting the XML to another stage is cheaper/easier.

scripts/metrics/pytest_metrics_xml_to_csv.py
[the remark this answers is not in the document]
Kadino: In that case, I think the script should not have options on how the data gets generated. It should be a one-shot that creates it in the "current" correct location.

My sentiments about ISO timestamp stands.

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,173 @@
...
+    """
+    Returns a default filename to be saved in the format of YYYY_MM_DD_HH_mm_pytest.csv
+    :return: The filename as a string
+    """
+    # Format default file name based off of date
+    now = datetime.datetime.now(tz=datetime.timezone.utc)
+    return f"{now.year}_{now.month:02d}_{now.day:02d}_{now.hour:02d}_{now.minute:02d}_pytest.csv"
+
+
+# Setup logging.
```
Kadino, line 33: minor: good to set this up before defining any functions

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,173 @@
...
+logger = logging.get_logger("test_metrics")
+logging.setup_logger(logger)
+
+# Create the csv field header
+PYTEST_FIELDS_HEADER = [
```
Kadino, line 38: minor: recommend setting up logging and defining constants immediate after imports, before defining functions

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,173 @@
...
+        "-o", "--output-directory", action="store", default=os.getcwd(),
+        help=f"The directory where the csv to be saved. Prepends the --csv-file arg."
+    )
+    parser.add_argument(
+        "-b", "--branch", action="store", default="",
+        help="The branch the metrics were generated on. Used for directory saving."
+    )
+    parser.add_argument(
+        "-w", "--weeks", action="store_true",
+        help="Boolean whether to include the week number in the created csv path."
```
Kadino, line 107: Weeks should not be an optional param, it should be automatically included (unless gets removed from the data schema, in which case it should never be included)

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,173 @@
...
+    parser.add_argument(
+        "--csv-file", action="store", default=_get_default_csv_filename(),
+        help=f"The file name for the csv to be saved."
+    )
+    parser.add_argument(
+        "-o", "--output-directory", action="store", default=os.getcwd(),
+        help=f"The directory where the csv to be saved. Prepends the --csv-file arg."
+    )
+    parser.add_argument(
+        "-b", "--branch", action="store", default="",
```
Kadino, line 102: default should probably explicitly be "unknown_branch" ...or this should be marked required, such that the script signals that the user is operating under an error.

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,173 @@
...
+    parser = argparse.ArgumentParser(
+        description='This script assumes that Pytest xml files have been produced. The xml files will be parsed and '
+                    'written to a csv file.')
+    parser.add_argument(
+        'xml_folder',
+        help="Path to where the Pytest xml files live. O3DE CTest defaults this to {build_folder}/Testing/Pytest"
+    )
+    parser.add_argument(
+        "--csv-file", action="store", default=_get_default_csv_filename(),
+        help=f"The file name for the csv to be saved."
```
Kadino, line 95: Why is it important to allow the user to override the filename? Is the timestamp not critical to the metrics schema? If it is important to provide the option, this should probably suggest why yet recommend relying on the default.

scripts/metrics/pytest_metrics_xml_to_csv.py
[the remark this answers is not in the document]
evanchia-ly-sdets: Here is an example of a path from a file that was previously uploaded
s3://o3de-metrics/csv/development/Week08/2023_02_20_08_02.csv

Fixed filename to use isoformat and will look like the following:
'2023_02_21T15_35_48_158328Z_pytest.csv'

scripts/metrics/pytest_metrics_xml_to_csv.py
evanchia-ly-sdets: Fixed

scripts/metrics/pytest_metrics_xml_to_csv.py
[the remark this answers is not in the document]
evanchia-ly-sdets: actually I got rid of the microseconds

scripts/metrics/pytest_metrics_xml_to_csv.py
evanchia-ly-sdets: Removed as an optional param

scripts/metrics/pytest_metrics_xml_to_csv.py
evanchia-ly-sdets: Set a default

scripts/metrics/pytest_metrics_xml_to_csv.py
evanchia-ly-sdets: Having an override is convenient for local testing, I can add a recommendation for default in the comments

scripts/metrics/ctest_metrics_xml_to_csv.py
```diff
@@ -47,9 +46,8 @@ def main():
 
     # Create csv file
     full_path = os.path.join(args.output_directory, args.branch)
-    if args.weeks:
-        week = datetime.datetime.now().isocalendar().week
-        full_path = os.path.join(full_path, f"Week{week:02d}")
+    week = datetime.datetime.now().isocalendar().week
```
Kadino, line 49: Today and one year from now appear to map into the same folder, which can create a reporting issue. I think that week needs to be preceded by year, though it may be more appropriate to instead have folders for year/month/day and let any reporting scripts ingest whichever days they require.

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,168 @@
...
+    """
+    # Parse args
+    args = parse_args()
+
+    if not os.path.exists(args.xml_folder):
+        raise exception.MetricsExn(f"Cannot find directory: {args.xml_folder}")
+
+    # Create csv file
+    full_path = os.path.join(args.output_directory, args.branch)
+    week = datetime.datetime.now().isocalendar().week
```
Kadino, line 61: Same year/week or year/month/day issue. May want to deduplicate the path-defining code if they both write a file with the same pattern.

scripts/metrics/pytest_metrics_xml_to_csv.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+def _get_default_csv_filename():
+    """
+    Returns a default filename to be saved in the format of YYYY_MM_DDTHH_mm_SS_pytest.csv
+    :return: The filename as a string
+    """
+    # Format default file name based off of date
+    now = datetime.datetime.isoformat(datetime.datetime.now(tz=datetime.timezone.utc), timespec='seconds')
+    return f"{now.replace('+00:00', 'Z').replace('-', '_').replace('.', '_').replace(':', '_')}_pytest.csv"
```
Kadino, line 43: With the possibility of pipelines running in parallel and uploading to the same location, these file names won't be guaranteed to be unique. While including seconds helps reduce collision, I a collision handling mechanism still needs to be considered. I assume the S3 uploader doesn't provide this, so perhaps the file names should use a UUID? (possibly instead of a timestamp? If we really need it, the timestamp info could instead be a metric added into the CSV file)

scripts/metrics/ctest_metrics_xml_to_csv.py
evanchia-ly-sdets: I was planning on having an archive folder for other years, but perhaps that may be overcomplicating it and years can just be included in the folder path. Preference?

scripts/metrics/pytest_metrics_xml_to_csv.py
evanchia-ly-sdets: Hm, yeah the uploader won't help here. Unfortunately I think we may need to use uuid to avoid collisions. I was trying to prevent that as they are less readable.

evanchia-ly-sdets: I could include microseconds in the timestamped filename, but that also reduces readability and does not guarantee collision handling

lumberyard-employee-dm: How about making the folder a Uuid?

evanchia-ly-sdets: With that suggestion path will look like:
`o3de-metrics\csv\development\Week08\1a9207c0-7a31-4e5b-a20d-775639de8c77\2023_02_23T21_27_35Z_pytest.csv`

I'm not sure if it would be better to just have:
`o3de-metrics\csv\development\Week08\1a9207c0-7a31-4e5b-a20d-775639de8c77.csv`

lumberyard-employee-dm: How about separating the Uuids using <dot>, then?
(Also Week08 seems redundant if the timestamp is part of the filename, as that can be calculated from the timestamp)
`o3de-metrics\csv\development\2023_02_23T21_27_35Z.1a9207c0-7a31-4e5b-a20d-775639de8c77.pytest.csv

scripts/metrics/ctest_metrics_xml_to_csv.py
Kadino: That archive pattern creates an odd boundary condition when one year transitions to the next. I think archiving should be set by a data retention policy, not by the folder structure.

I think `branch/year/month/day` existing as folders makes sense, with timestamp and for UUID the filename.

scripts/metrics/pytest_metrics_xml_to_csv.py
Kadino: > Perhaps the first 8 digits of the Uuid would be sufficient. `o3de-metrics\csv\development\2023_02_23T21_27_35Z.1a9207c0.pytest.csv`

I proposed separating YYYY/MM/DD into folders above, but I also like this file naming proposal. Salting with 8 UUID digits seems sufficient since we're unlikely to get more than a few collisions on the same second.

There are likely to be hundreds of files generated per day, and saving everything in the same `development` folder would result in needing to regex filter a large set of files. Pre-partitioning by day provides report generators simpler collection of every relevant file in known folders.

evanchia-ly-sdets: New format will be:
`o3de-metrics\csv\development\2023\02\23\f56aa22b.2023_02_23T22_00_58Z_pytest.csv`

Kadino (approved):
I'm okay with this new format duplicating date info between folder and filename, but suggest it would be more readable as:
`o3de-metrics\csv\BRANCHNAME\2023\02\23\22_00_58Z_f56aa22b_RUNNERNAME.csv`

PR #14566, opened 2023-02-13
Multiplayer Component to Debug Player Id

AMZN-Gene:
A component (**NetworkDebugPlayerIdComponent**) for network player to display player ids and how many players are connected. 
Helps when testing client apps to tell players apart and see how many players are currently connected.

- Use _cl_debugPlayerIdRender_ to enable/disable rendering at runtime (or simply remove the component from the player at edit time to disable the feature)

Multiplayer system doesn't directly track the number of players, this instead just assign this player an id by checking how many machines are already connected to this host.
Note 1: This doesn't support reassigning player-ids to rejoining players
Note 2: This doesn't consider multi-server connections; it's possible that connection count includes other servers (not just player machines)
Note 3: Client-server player count will be wrong by -1; client-server has its own player without any connections.

image


AMZN-Olex:
Very useful! It would awesome to add a cvar to toggle this debug info.

Gems/Multiplayer/Code/Source/Components/NetworkDebugPlayerIdComponent.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+    void NetworkDebugPlayerIdComponentController::OnDeactivate([[maybe_unused]] Multiplayer::EntityIsMigrating entityIsMigrating)
+    {
+        if (IsNetEntityRoleAuthority())
+        {
+            AZ::TickBus::Handler::BusDisconnect();
+        }
+    }
+
+    void NetworkDebugPlayerIdComponentController::OnTick([[maybe_unused]]float deltaTime, [[maybe_unused]]AZ::ScriptTimePoint time)
```
kberg-amzn, line 161: Updating this every tick seems a bit brutal.. do we have an event for connect/disconnect on the server?

AMZN-Gene: Unfortunately, there's no event for connect/disconnect. If I added functionality to Multiplayer System, I would bypass connection count altogether and actually track OnPlayerJoin and OnPlayerLeave events.

AMZN-Gene: Oops, nope, you're right... there is a connection/disconnect event. So the on startup I'll grab the connection count directly, and then begin listening for new connections. Stay tuned!

kberg-amzn: And don't bother trying to increment/decrement based on those events.. every time the event pops just grab the source of truth connection count from the network interface.

lmbr-pip:
I would argue we have two use cases for the displayed txt. IMHO:

* Player identity near the player
* Connected player count should go to the bottom right hand corner with the other networking "stats" ie server, ip etc

Gems/Multiplayer/Code/Source/Components/NetworkDebugPlayerIdComponent.cpp
AMZN-Gene: fixed! thank you

AMZN-Gene:
> I would argue we have two use cases for the displayed txt. IMHO:
> 
> * Player identity near the player, its relevant to the actor so can stay where is or move slightly above player.
> * Connected player count should go to the bottom right hand corner with the other networking "information" ie server, ip etc

Fixed! Screenshot updated... thanks :)

AMZN-Gene:
> Very useful! It would awesome to add a cvar to toggle this debug info.

"Ask and you shall receive" - Gene 2023
_cl_debugPlayerIdRender_

Gems/Multiplayer/Code/Source/Components/NetworkDebugPlayerIdComponent.cpp
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    void NetworkDebugPlayerIdComponent::OnTick([[maybe_unused]] float deltaTime, [[maybe_unused]] AZ::ScriptTimePoint time)
+    {
+        if (!cl_debugPlayerIdRender)
+        {
+            return;
+        }
+
+        AZ::Vector3 renderWorldSpace = GetEntity()->GetTransform()->GetWorldTranslation() + m_translationOffset;
```
AMZN-Olex, line 93: You could move this after the if checks that bail out.

Gems/Multiplayer/Code/Source/Components/NetworkDebugPlayerIdComponent.cpp
```diff
@@ -0,0 +1,185 @@
...
+            renderWorldSpace, m_viewport->GetCameraViewMatrixAsMatrix3x4(), m_viewport->GetCameraProjectionMatrix(), AzFramework::ScreenSize(windowSize.m_width, windowSize.m_height));
+        m_drawParams.m_hAlign = AzFramework::TextHorizontalAlignment::Center;
+        m_drawParams.m_position = AZ::Vector3(aznumeric_cast<float>(renderScreenpoint.m_x), aznumeric_cast<float>(renderScreenpoint.m_y), 0.f);
+
+        AZStd::string playerIdText = AZStd::string::format("Player %i", GetPlayerId());
+        m_fontDrawInterface->DrawScreenAlignedText2d(m_drawParams, playerIdText);
+
+        if (IsNetEntityRoleAutonomous())
+        {
+            const auto controller = static_cast<NetworkDebugPlayerIdComponentController*>(GetController());
```
AMZN-Olex, line 117: Isn't this garatuneed if the role is autonomous?

AMZN-Olex (approved):
Approved assuming my comments are addressed.

Gems/Multiplayer/Code/Source/Components/NetworkDebugPlayerIdComponent.cpp
```diff
@@ -0,0 +1,185 @@
...
+        }
+
+        NetworkDebugPlayerIdComponentBase::Reflect(context);
+    }
+
+    void NetworkDebugPlayerIdComponent::OnActivate([[maybe_unused]] EntityIsMigrating entityIsMigrating)
+    {
+        #if (!AZ_TRAIT_CLIENT)
+            return;
+        #endif
```
AMZN-stankowi, line 51: it might be worth wrapping the rest of this function in an #else. I think I've seen compilers give warnings on unreachable code with early outs like this. I don't know if we have that warning enabled in any build configuration, but I wouldn't be surprised if Mac builds had that compile warning on and failed with this.

AMZN-stankowi (commented):
How possible would it be to get some automated tests written for this component?

Gems/Multiplayer/Code/Include/Multiplayer/Components/NetworkDebugPlayerIdComponent.h
```diff
@@ -0,0 +1,84 @@
...
+#include <AzCore/Component/TickBus.h>
+#include <AzFramework/Font/FontInterface.h>
+
+namespace AZ
+{
+    namespace RPI
+    {
+        class ViewportContext;
+    }
+}
```
AMZN-alexpete, line 21:
```suggestion
namespace AZ::RPI
{
    class ViewportContext;
}
```

Gems/Multiplayer/Code/Source/Components/NetworkDebugPlayerIdComponent.cpp
```diff
@@ -0,0 +1,185 @@
...
+#include <Multiplayer/Components/NetworkDebugPlayerIdComponent.h>
+
+#include <Atom/RPI.Public/ViewportContextBus.h>
+#include <AzCore/Serialization/SerializeContext.h>
+#include <AzFramework/Viewport/ViewportScreen.h>
+#include <AzNetworking/Framework/INetworking.h>
+
+namespace Multiplayer
+{
+    AZ_CVAR(bool, cl_debugPlayerIdRender, true, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Whether NetworkDebugPlayerIdComponent should render the player id and connection count on-screen.");
```
AMZN-alexpete, line 18: Are you sure you want to default this to true?  I'd go with setting this to false in the gem so it isn't on by default for all projects, and setting it to true in our specific MPS Registry/game.setreg

Gems/Multiplayer/Code/Source/Components/NetworkDebugPlayerIdComponent.cpp
AMZN-Gene: It well ought to be guaranteed! Hence the Assert; there must be a problem with the code itself should it be null. I guess that's too cautious? Better to leave out the null check entirely?

Gems/Multiplayer/Code/Source/Components/NetworkDebugPlayerIdComponent.cpp
```diff
@@ -0,0 +1,185 @@
...
+            serializeContext->Class<NetworkDebugPlayerIdComponent, NetworkDebugPlayerIdComponentBase>()
+                ->Version(1)
+                ->Field("translationOffset", &NetworkDebugPlayerIdComponent::m_translationOffset)
+                ->Field("scale", &NetworkDebugPlayerIdComponent::m_fontScale)
+                ->Field("color", &NetworkDebugPlayerIdComponent::m_fontColor)
+            ;
+
+            if (AZ::EditContext* editContext = serializeContext->GetEditContext())
+            {
+                editContext->Class<NetworkDebugPlayerIdComponent>("Network Debug Connection Counter", "Renders the player id as debug text over network players.")
```
AMZN-alexpete, line 33: Long names don't fit super well in the entity outliner, recommend something shorter
```suggestion
                editContext->Class<NetworkDebugPlayerIdComponent>("Network Debug Player ID", "Renders the player id as debug text over network players.")
```

Gems/Multiplayer/Code/Source/Components/NetworkDebugPlayerIdComponent.cpp
AMZN-stankowi: If it's off by default, it could be confusing to use this component. You find it in the component list, add it to an entity hoping to get extra info, and nothing happens. I don't know how discoverable the cvars are from the entity inspector that shows this component.

Would it be possible to add a boolean checkbox toggle that actually flips the cvar, on the inspector for this component?

AMZN-Gene: _"Would it be possible to add a boolean checkbox toggle that actually flips the cvar, on the inspector for this component?"_
Now that would be awesome. Like a component editor property that very clearly allows users to change a global setting. Is that possible? Is there an existing UX example of a component doing that?

AMZN-alexpete: Good point, I wonder if this should be a single system component instead that monitors the cvar, instead of a component you need to remember to add to your players.

AMZN-stankowi: No idea on how possible that would be. My guess is yes? Can't you, with the component property editor, mark some properties as not being variables on a component but instead the result of a function?

I haven't spent a lot of time working with the edit context, though, so I'm not totally positive.

AMZN-Gene:
> How possible would it be to get some automated tests written for this component?

Yes, in fact, I would likely use this component for the existing multiplayer AR tests that are currently sandboxed! Multiplayer Component tests still need hardening. We have 1, but still see it flake out about once a week.

Gems/Multiplayer/Code/Source/Components/NetworkDebugPlayerIdComponent.cpp
AMZN-Gene: " I wonder if this should be a single system component instead that monitors the cvar, instead of a component you need to remember to add to your players."
I actually really like this idea. Sort of nice multiplayer-components handles the traffic, but this could instead be sending network packets to and from a system component. That system component could have a cvar controlled on the server side... server cvars can be marked for replication so clients know when it's toggled on and off; clients could listen in and receive the packets and render player id's when the cvar is enabled. The packet would include each player's id and currently world-space location. Would need to be careful to only update the players that haven't moved much. It also wouldn't be predictive, so you'd be seeing your own player id location in your past (jarring?). Sending packets directly would also be missing out on the implicit net-entity update filtration for hacker avoidance (ie: not sending other player position properties which the player cant currently see). However, this is for debugging anyways and devs would disable it for release.

AMZN-Gene: > No idea on how possible that would be. My guess is yes? Can't you, with the component property editor, mark some properties as not being variables on a component but instead the result of a function?
> 
> I haven't spent a lot of time working with the edit context, though, so I'm not totally positive.

I could link the component's edit-context to a cvar, but I'm more worried about the ux. Maybe just label it like... "GLOBAL SETTING: Enable Rendering". I that that would be a first... the cleanest way would be probably be @AMZN-alexpete suggestion and just remove this component completely and have a system component render all this.

AMZN-Gene: > It also wouldn't be predictive

Actually, that might not be true. I could send the network entity id instead of the position to the clients. The system component could use grab the entities by hand. That would well reduce net traffic too since you could send the player id once and forget it. Net Entity Id is persistent forever (I THINK)

Gems/Multiplayer/Code/Source/Components/NetworkDebugPlayerIdComponent.cpp
```diff
@@ -0,0 +1,185 @@
...
+            ;
+
+            if (AZ::EditContext* editContext = serializeContext->GetEditContext())
+            {
+                editContext->Class<NetworkDebugPlayerIdComponent>("Network Debug Player ID", "Renders the player id as debug text over network players.")
+                    ->ClassElement(AZ::Edit::ClassElements::EditorData, "")
+                        ->Attribute(AZ::Edit::Attributes::Category, "Multiplayer")
+                        ->Attribute(AZ::Edit::Attributes::AppearsInAddComponentMenu, AZ_CRC_CE("Game"))
+                    ->DataElement(AZ::Edit::UIHandlers::Default, &NetworkDebugPlayerIdComponent::m_translationOffset, "Translation Offset", "The world-space offset from the player position to render the debug text.")
+                    ->DataElement(AZ::Edit::UIHandlers::Default, &NetworkDebugPlayerIdComponent::m_fontScale, "Font Scale", "Apply a scale to the default debug font rendering size.")
```
allisaurus, line 38: what _is_ the default debug font rendering size?

Gems/Multiplayer/Code/Source/Components/NetworkDebugPlayerIdComponent.cpp
AMZN-Gene: The if-check uses the world space position (making sure we dont render players behind the player)

Gems/Multiplayer/Code/Source/Components/NetworkDebugPlayerIdComponent.cpp
```diff
@@ -0,0 +1,185 @@
...
+        # if AZ_TRAIT_SERVER
+            m_networkInterface = AZ::Interface<AzNetworking::INetworking>::Get()->RetrieveNetworkInterface(AZ::Name(MpNetworkInterfaceName));
+            if (IsNetEntityRoleAuthority())
+            {
+                uint32_t currentConnectionCount = m_networkInterface->GetConnectionSet().GetConnectionCount();
+                //SetConnectionCount(currentConnectionCount);
+
+                // Multiplayer system doesn't directly track the number of players.
+                // Instead just assign this player an id by checking how many machines are already connected to this host.
+                // Note 1: Players can have duplicate ids based on how many players were connected at the time they activated.
```
allisaurus, line 155: (maybe a future change) if giving IDs based on connection count is unreliable (i join when there's 3 players so i'm player 4, one leaves, the next person to join is _also_ player 4) what do you think about random assignment of an adjective + noun? like "VibrantEagle", "TerrificSnake", etc. - that way uniqueness is better guaranteed

allisaurus (approved):

Gems/Multiplayer/Code/Source/Components/NetworkDebugPlayerIdComponent.cpp
AMZN-Gene: "This big" _*closes one eye and looks through pinched fingers_

Oh geez, I wish I knew. It looks like 12pt font, but I'd be surprised if it's using a standard [font size unit](https://m2.material.io/design/typography/the-type-system.html#type-scale)

Gems/Multiplayer/Code/Source/Components/NetworkDebugPlayerIdComponent.cpp
AMZN-Gene: Rendering a player id above the player's head does seem like the sort of thing each player would be in charge of though, right? Rather than a system component I mean? I do think sending the total connection count would be nicer handled via system component though (for a later improvement).

Gems/Multiplayer/Code/Source/Components/NetworkDebugPlayerIdComponent.cpp
AMZN-Gene: Haha, that would be cool. Definitely more game specific though. I was thinking at the very least broadcast event the MultiplayerAgentDatum (which does actually contain a unique player id) to the player entity.

Gems/Multiplayer/Code/Source/Components/NetworkDebugPlayerIdComponent.cpp
AMZN-Gene: https://github.com/o3de/o3de.org/issues/2249

Gems/Multiplayer/Code/Source/Components/NetworkDebugPlayerIdComponent.cpp
allisaurus: Nice, that sounds like a much more fully-baked idea 😄

Gems/Multiplayer/Code/Source/Components/NetworkDebugPlayerIdComponent.cpp
AMZN-Gene: removed the check

Gems/Multiplayer/Code/Source/Components/NetworkDebugPlayerIdComponent.cpp
AMZN-Gene: Thank you! Fixed.

Gems/Multiplayer/Code/Include/Multiplayer/Components/NetworkDebugPlayerIdComponent.h
```diff
@@ -0,0 +1,84 @@
...
+
+        #if AZ_TRAIT_SERVER
+            void SetOwningConnectionId(AzNetworking::ConnectionId connectionId) override;
+        #endif
+
+        // AZ::TickBus::Handler overrides...
+        void OnTick(float deltaTime, AZ::ScriptTimePoint time) override;
+
+    private:
+        // Properties exposed to editor
```
lmbr-pip, line 45: nit: Consider comments that discuss what these control/how set

Gems/Multiplayer/Code/Include/Multiplayer/Components/NetworkDebugPlayerIdComponent.h
```diff
@@ -0,0 +1,84 @@
...
+#include <AzFramework/Font/FontInterface.h>
+
+namespace AZ::RPI
+{
+    class ViewportContext;
+}
+
+namespace Multiplayer
+{
+    class NetworkDebugPlayerIdComponent
```
lmbr-pip, line 22: Can we have a public class description that talks to intent/responsibilities?

Gems/Multiplayer/Code/Include/Multiplayer/Components/NetworkDebugPlayerIdComponent.h
AMZN-Gene: Fixed

Gems/Multiplayer/Code/Include/Multiplayer/Components/NetworkDebugPlayerIdComponent.h
AMZN-Gene: Good call

PR #14572, opened 2023-02-14
fix for non-unique name being used as handler pool ID

alexmontAmazon:
Signed-off-by: Alex Montgomery <alexmont@amazon.com>

## What does this PR do?
before handlerId->m_name was assumed to be unique for each handler type, but this has proven not to be the case. This fix switches to using a stringified AZ::Name of the HandlerID instead of using the member m_name.

## How was this PR tested?
Locally with the DPE standalone app


forkercat (approved):

srikappa-amzn (approved):
It'd be good to have some tests in place to ensure that the components in inspector show up without issues when DPE is enabled.

@alexmontAmazon , can you add some tests to this PR or create a follow up github issue for them?

dtamkin1 (approved):

PR #14581, opened 2023-02-14
Action Manager | Enable the new Action Manager by default

AMZN-daimini:
Note: this PR has been created early to streamline the submission process, but is pending confirmation before being submitted.

## What does this PR do?

Enables the new Action Manager by default.

Note: if this change ends up causing issues to any workflow, please file a bug directed to sig/content or myself.
It is then possible to temporarily revert this change and use the legacy systems by setting the `EnableNewActionManager` value to false manually in any `.setreg` file:

```
{
    "O3DE": {
        "ActionManager": {
            "EnableNewActionManager": false
        }
}
```

## How was this PR tested?

Functionality has been live behind a flag in development for a long time and has gone through multiple test passes.


hultonha:
@AMZN-daimini Is it worth reaching out to LB to do a test pass on this branch before merging or has enough testing been done already with this change enabled? As it's so simple to turn off it might not even be worth it if people do run into problems but just thought I'd check.

AMZN-daimini:
> @AMZN-daimini Is it worth reaching out to LB to do a test pass on this branch before merging or has enough testing been done already with this change enabled? As it's so simple to turn off it might not even be worth it if people do run into problems but just thought I'd check.

We've done a couple test passes on both Windows and Linux (LB handled the Linux ones), which I think uncovered all of the issues we can gather from our point of view. It is possible there's still a few bugs in some areas, but given how simple it is to revert locally or for the whole branch I think it's worth it to just switch the functionality on by default and handle any issue as it happens. Hopefully it will go smooth, but we have a few alternative plans if anything comes up :)

hultonha (approved):

cgalvan (approved):
The unit tests should pass once https://github.com/o3de/o3de/pull/14604 gets merged

PR #14587, opened 2023-02-14
Fixed application focus being lost with EntityPropertyEditor refresh

### LINKED ISSUE #14446: Action Manager: Exiting component modes requires a click before ESC works as a hotkey to exit
**Describe the bug**
Exiting component modes requires a mouse click somewhere in the Editor window before ESC works as a hotkey to exit.

**Steps to reproduce**
1. Launch the Editor with the new Action Manager created Menus active:
```
Editor.exe --regset=O3DE/ActionManager/EnableNewActionManager=true
```
3. Add a new entity to the level with a Box Shape component.
4. Click the Edit button on the Box Shape component to enter component mode.
5. Hit ESC.

**Expected behavior**
Component mode is immediately exited.

**Actual behavior**
Nothing happens.

**Found in Branch**
development

57a7df88f77a5e741e0a70dd06831314e3cf427c

**Additional context**
1. After the user enters component mode, if they click somewhere in the Editor window, the ESC hotkey will then work. This is a regression from the pre-Action Manager setup.

cgalvan:
## What does this PR do?

Fixes #14446

When the `EntityPropertyEditor` does a queued full refresh, part of the logic disables certain widgets to prevent user input during the refresh, but if one of those widgets (or children/grand-children of those widgets) was the previously focused widget, this results in the focus being cleared, which removes focus from the entire application. This prevents the application from receiving key events/shortcuts.

This fix detects this loss of focus and ensures the `EntityPropertyEditor` retains focus after the refresh, which could fix various different issues, most specifically in #14446 where the user can't use the `Escape` key to exit component mode until they click somewhere in the Editor to regain focus.

## How was this PR tested?

Tested reproduction steps and other ad-hoc testing to ensure Editor shortcuts still work after entering/exiting component modes. Also ran all unit/automated tests.


dtamkin1 (approved):

AMZN-daimini (approved):

burelc-amzn (approved):
It seems it would be nicer if we could restore focus directly to the new widget for the same property. Do we have any way to identify some kind of path to a property, so that we could identify that widget after the refresh to restore the focus?

cgalvan:
> It seems it would be nicer if we could restore focus directly to the new widget for the same property. Do we have any way to identify some kind of path to a property, so that we could identify that widget after the refresh to restore the focus?

I thought about that first. In the case of the actual use-case for the bug that was filed, it wouldn't help since the button you click to go into component mode no longer exists after the refresh, but for other cases I think it would be possible. I'd like to revisit it once the DPE is enabled by default, so that we don't spend too much time on RPE fixes that will be (hopefully) soon a waste when we're able to deprecate it.

hultonha (approved):
Nice! Thanks for the detailed PR description and comment too 👍

PR #14589, opened 2023-02-15
Add verification before saving material document.

Zhang-Baochong:
## What does this PR do?
If rename opening material source, then `Ctrl S` save the material, there will be two materials in asset browser.
For this situation, need to add verification before saving material document.
material-save-2

## How was this PR tested?
1、create a new material document
2、Open it in explorer， and rename
3、Change any param for origin material, and Ctrl+S to save, Check if the issue has been solved
add-check-save-material

### DIFF
--- a/Gems/Atom/Tools/AtomToolsFramework/Code/Source/Document/AtomToolsDocumentSystem.cpp
+++ b/Gems/Atom/Tools/AtomToolsFramework/Code/Source/Document/AtomToolsDocumentSystem.cpp
@@ -350,6 +350,12 @@ namespace AtomToolsFramework
         }
 
         const QFileInfo saveInfo(savePath.c_str());
+        if (!saveInfo.exists())
+        {
+            QMessageBox::critical(QApplication::activeWindow(), "Error", QObject::tr("Document does not exist:\n%1").arg(savePath.c_str()));
+            return false;
+        }
+
         if (saveInfo.exists() && !saveInfo.isWritable())
         {
             DisplayErrorMessage(


tjmgd (approved):
Looks good

gadams3 (changes requested):
Hello! Can you please explain what the bug is here? This seems like a breaking change. The user should be able to save files whether they exist or not. 

The document system allows creating documents in memory that can be saved after the fact, which we do in material canvas and shader management console. Material editor currently asks to save up front because of a requirement that no longer exists, where the file had to be saved so that it could be picked up by the asset processor to create an asset and instance for editor functors and displaying in the viewport. Material editor now bypasses the AP and creates primary material assets in memory.

The attached video shows the user renaming the file via explorer or the OS file browser. Like many other tools, material editor is operating on what's in memory and not monitoring external file renames or deletions. It is valid to be able to rename or delete a file then save a new one with the old name.

We have ignored most other file operations except if the currently open document or one of its dependencies gets updated, which asks the user if they want to reload.

Zhang-Baochong:
> Hello! Can you please explain what the bug is here? This seems like a breaking change. The user should be able to save files whether they exist or not.
> 
> The document system allows creating documents in memory that can be saved after the fact, which we do in material canvas and shader management console. Material editor currently asks to save up front because of a requirement that no longer exists, where the file had to be saved so that it could be picked up by the asset processor to create an asset and instance for editor functors and displaying in the viewport. Material editor now bypasses the AP and creates primary material assets in memory.
> 
> The attached video shows the user renaming the file via explorer or the OS file browser. Like many other tools, material editor is operating on what's in memory and not monitoring external file renames or deletions. It is valid to be able to rename or delete a file then save a new one with the old name.
> 
> We have ignored most other file operations except if the currently open document or one of its dependencies gets updated, which asks the user if they want to reload.

Ok, I agree with you after thinking，the user should be able to save files whether they exist or not. so its no need to add verification before saving material document. I will close this PR, thx.

PR #14269, opened 2023-01-27
'Be this camera' fixes and improvements

hultonha:
## What does this PR do?

Fixes #13525
Fixes #14285
Fixes #13763
Fixes #13744
Fixes #13476
Fixes #10428
Fixes #10427

_These are just the ones I could find... there could be more..._

This PR fixes a _ton_ of issues related to 'Be this camera'... it's been hard to keep track, this really was a rabbit whole.

I'm pretty confident things are much better now than they were before and ~~I'm in the progress of adding tests to verify the corrected functionality~~ (__5 total test have been added__).

The problems included but were not limited to...

1. Going to game mode and back when being Editor camera would not return to the previous position (instead the camera would go to the saved view bookmark 'last known location', and was only updated when saved).
2. Changes to `EditorCameraComponent` would not persist when going to `GameMode` after using 'Be this camera'.
3. View bookmark warning would fire because `SetDefaultCamera` in `EditorViewportWidget` would be called before view bookmarks are available.
4. An error message was printed when changing properties on the camera without it selected when it Edit Mode (`[Error] (EditorViewportWidget) - GetCurrentViewEntityId called while the current view is being changed. You may get inconsistent results if you make use of the returned entity ID. This is an internal error, please report it as a bug.` and `[Error] (EditorViewportWidget) - Internal logic error - active view changed to an entity which is not an editor camera. Please report this as a bug.`).
5. Changing transform component when in 'Be this camera' would not change the camera transform (instead it would stutter).
6. Going to game mode with multiple cameras in the scene would produce `[Warning] (EditorViewportWidget) - Tried to change the editor camera during play game in editor; this is currently unsupported`.
7. Opening a level and just switching between EditorCamera and Camera (CameraComponent) gets the Editor into a bad state.
8. Going to game mode when in 'Be this camera' would print `[Warning] (EditorViewportWidget) - Tried to change the editor camera during play game in editor; this is currently unsupported`.
9. Opening a new level would go to the origin when in normal EditorCamera.
10. Always would return to EditorCamera when starting Game Mode as 'Be in this camera' and then returning to Edit Mode.
11. When in 'Be this camera' and then changing level, the camera would get stuck in a weird state (still assigned to a camera from the previous scene when in EditorCamera mode).

## How was this PR tested?

Extensive manual testing in the Editor (changing cameras, changing level, going in and out of game mode, modifying camera properties, etc...)

New automation tests to verify a significant proportion of the error lists above.

[...]

PR #14594, opened 2023-02-15
Update notifications for Editor Entity Context

hultonha:
> Note: This PR is intentionally going to `development` and not `stabilization/2305` as this isn't an important fix needed for the release but is an important long-term improvement.

## What does this PR do?

The incorrect ordering of a couple of entity notification events was discovered as part of this change (https://github.com/o3de/o3de/pull/14269). As the fixes weren't required by the previous PR, these changes were extracted and are being presented separately.

After discussing with @AMZN-daimini, we took the opportunity to update the notification bus name `PrepareForContextReset` to `OnPrepareForContextReset` to make it clearer it is a notification event (it was not widely used so should not be a concern for changes outside the core engine/editor code). Types previously implementing `OnContextReset` have also been updated to use `OnPrepareForContextReset` to ensure behavior remains the same as before.

## How was this PR tested?

Tested loading/changing levels in the editor to ensure the notifications are sent at the right time.

A new integration test has been added to confirm the events are received and appear in the correct order.

### Before

```bash
[==========] Running 1 test from 1 test case.
[----------] 1 test from EditorEntityContextNotificationFixture
[ RUN      ] EditorEntityContextNotificationFixture.EditorEntityContextNotificationsReceivedInCorrectOrder
C:/dev/o3de/Code/Framework/AzToolsFramework/Tests/Prefab/PrefabEditorEntityNotificationTests.cpp(55): error: Value of: m_prepareForContextReset
Expected: is true
  Actual: false (of type bool)
C:/dev/o3de/Code/Framework/AzToolsFramework/Tests/Prefab/PrefabEditorEntityNotificationTests.cpp(84): error: Value of: m_editorEntityContextNotificationOrderingDetector.m_prepareForContextReset
Expected: is true
  Actual: false (of type bool)
[  FAILED  ] EditorEntityContextNotificationFixture.EditorEntityContextNotificationsReceivedInCorrectOrder (359 ms)
[----------] 1 test from EditorEntityContextNotificationFixture (360 ms total)

[==========] 1 test from 1 test case ran. (363 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] EditorEntityContextNotificationFixture.EditorEntityContextNotificationsReceivedInCorrectOrder
```

### After

```bash
[==========] Running 1 test from 1 test case.
[----------] 1 test from EditorEntityContextNotificationFixture
[ RUN      ] EditorEntityContextNotificationFixture.EditorEntityContextNotificationsReceivedInCorrectOrder
[       OK ] EditorEntityContextNotificationFixture.EditorEntityContextNotificationsReceivedInCorrectOrder (368 ms)
[----------] 1 test from EditorEntityContextNotificationFixture (369 ms total)

[==========] 1 test from 1 test case ran. (370 ms total)
[  PASSED  ] 1 test.
```


srikappa-amzn (approved):

AMZN-daimini (approved):

cgalvan (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Entity/EditorEntityContextBus.h
```diff
@@ -162,9 +162,9 @@ namespace AzToolsFramework
         virtual ~EditorEntityContextNotification() = default;
 
         /// Called before the context is reset.
-        virtual void PrepareForContextReset() {}
+        virtual void OnPrepareForContextReset() {}
```
srikappa-amzn, line 165: A deprecation notice might be needed if we are renaming this since there can be other usages outside the o3de repo. Maybe force 'PrefapareForContextReset()' to call 'OnPrepareForContextReset()?' and add a deprecation notice?

Personally, I think 'PrepareForContextReset()' is a good enough name, but if the reasoning behind the renaming is to start notification events with 'On...' , I understand.

srikappa-amzn (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Entity/EditorEntityContextBus.h
hultonha: That's likely not a bad call, I was on the fence about if it was worth it or not as this existing API I don't think was even really used by our code, so it did seem very unlikely it would be used by any external code, but it's likely better to be safe than sorry so I'll create one to track.

I found renaming to `OnPrepareForContextReset` was worth it as it helped disambiguate the separate virtual call in `EntityContext` called the same thing (which isn't easily visible in the PR).

hultonha: Created https://github.com/o3de/o3de/issues/15310

PR #14604, opened 2023-02-15
Fixed EditorTransformComponentSelection tests with new action manager enabled

cgalvan:
## What does this PR do?

Fixed several `EditorTransformComponentSelection` unit tests that fail when the new action manager is enabled by default, which was uncovered in this PR (https://github.com/o3de/o3de/pull/14581). These tests failed because they trigger operations by injecting key presses into a dummy widget that gets constructed as part of the test fixture, but these actions are registered in a different workflow when the new action manager is enabled. This required the test fixture to be updated to construct a similar dummy widget for the new action manager to register its default action context with.

## How was this PR tested?

Ran all unit tests with and without the new action manager enabled and verified they all pass.


AMZN-daimini (approved):

gadams3 (approved):

dtamkin1 (approved):

hultonha (commented):
Thanks for fixing this @cgalvan - it'll be nice to clean-up the test code (remove the widget that's being used in the other branches) when we do eventually remove the old action manager code, but that can happen down the line. Thanks!

PR #14607, opened 2023-02-15
Fix array/vector network serialization to have unique names.

mbalfour-amzn:
## What does this PR do?

The array and vector network serialization helpers were using "Element" as the base name for every entry in the array/vector. This could lead to inconsistent serialized state like the following:
image

Note that the ActiveShots.Size is 0, but there is an entry for an element in ActiveShots. This occurs when multiple weapons states in an array are serialized, where the first one has a size of 1, and the second has a size of 0. The second size overwrites the first, but the rest of the previous entry is left as-is.

The fix changes the helpers to generate unique names like "00", "01", etc for each array entry instead of "Element" for every entry. The assert in StringifySerializer is there to provide more visibility in case something else ends up creating non-unique names in the future.

## How was this PR tested?

Ran MultiplayerSample. First, added the assert and verified that it triggered from state overwrites. Then, added the name fix. Finally, verified that the assert no longer triggers, and manually examined the state to verify that it got multiple adjacent element entries instead of a single merged one.

### DIFF
--- a/Code/Framework/AzNetworking/AzNetworking/Serialization/StringifySerializer.cpp
+++ b/Code/Framework/AzNetworking/AzNetworking/Serialization/StringifySerializer.cpp
@@ -136,6 +136,15 @@ namespace AzNetworking
     {
         const AZStd::string keyString = m_prefix + name;
         AZ::CVarFixedString valueString = AZ::ConsoleTypeHelpers::ValueToString(value);
+
+        // Verify that the serialization isn't silently accidentally replacing an existing key with a different value.
+        AZ_Assert(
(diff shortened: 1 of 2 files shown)


kberg-amzn (approved):
Nice catch! Thanks Mike!

lmbr-pip (approved):
LG. Did we run unit tests? Should we add one?

mbalfour-amzn:
> LG. Did we run unit tests? Should we add one?

The unit tests for StringifySerializer are pretty slim, and there aren't any unit tests for the SerializeNetworkPropertyHelper methods. I'll create a follow-up PR to add unit tests for the condition fixed here.

PR #14609, opened 2023-02-16
bugfix: correct change status of "Actor name".

Jackie9527:
## What does this PR do?

1. Click Tools -> Animation Editor.
2. Click File -> Open Actor.
3. Check `Actor name`, click the input text and select the value, but do not make any change.
image
4. Click else where.
5. Check `Actors`.
image

## How was this PR tested?
Follow the same steps.

### DIFF
--- a/Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/SceneManager/ActorPropertiesWindow.cpp
+++ b/Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/SceneManager/ActorPropertiesWindow.cpp
@@ -334,6 +334,12 @@ namespace EMStudio
             return;
         }
 
+        // If the names are the same, do not change them.
+        if (m_nameEdit->text().compare(m_actor->GetName()) == 0)
+        {
+            return;
+        }
+ 
         // execute the command
         const AZStd::string command = AZStd::string::format("AdjustActor -actorID %i -name \"%s\"", m_actor->GetID(), m_nameEdit->text().toUtf8().data());
 


hultonha (commented):
Change looks good @Jackie9527 but could you please use `AZStd` instead of `std`. Also could you explain what was incorrect before please, it's not immediately clear to see from the screen shots. Thanks!

Jackie9527:
> Change looks good @Jackie9527 but could you please use `AZStd` instead of `std`. Also could you explain what was incorrect before please, it's not immediately clear to see from the screen shots. Thanks!

image
The font of actor name is changed to italic, which is wrong when I just click actor name, not modify it.

Gems/EMotionFX/Code/EMotionFX/Tools/EMotionStudio/Plugins/StandardPlugins/Source/SceneManager/ActorPropertiesWindow.cpp
```diff
@@ -334,6 +334,12 @@ namespace EMStudio
             return;
         }
 
+        // If the names are the same, do not change them.
+        if (std::strcmp(m_actor->GetName(), m_nameEdit->text().toUtf8().data()) == 0)
```
spham-amzn, line 338: `m_nameEdit->text()` is a QString already right? Why not just use its built-in `compare` function?
```suggestion
        if (m_nameEdit->text().compare(QStringView(m_actor->GetName())==0)
```

Jackie9527: done.

hultonha (approved):
LGTM! Nice work 🥳

spham-amzn (approved):

PR #14629, opened 2023-02-16
Added unit tests for the StringifySerializer and the serializer helpers.

mbalfour-amzn:
## What does this PR do?

Added unit tests to test the previously-submitted fixes to StringifySerializer and the serializer helpers to verify that they create unique entries and don't overwrite on serialization.

Also discovered from the unit test that the assert had a condition wrong, so I fixed that too.

## How was this PR tested?

Wrote and ran the unit tests.

### DIFF
--- a/Gems/Multiplayer/Code/Tests/MultiplayerComponentTests.cpp
+++ b/Gems/Multiplayer/Code/Tests/MultiplayerComponentTests.cpp
@@ -1,0 +1,77 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project. For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#include <CommonNetworkEntitySetup.h>
+#include <MockInterfaces.h>
+#include <AzCore/UnitTest/TestTypes.h>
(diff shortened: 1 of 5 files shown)


Gems/Multiplayer/Code/Tests/MultiplayerComponentTests.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include <AzCore/UnitTest/TestTypes.h>
+#include <AzCore/UnitTest/UnitTest.h>
+#include <AzNetworking/Serialization/StringifySerializer.h>
+#include <AzTest/AzTest.h>
+#include <Multiplayer/Components/MultiplayerComponent.h>
+
+namespace Multiplayer
+{
+    using namespace testing;
+    using namespace ::UnitTest;
```
lumberyard-employee-dm, line 19: We should avoid using namespace at namespace scope as well.

mbalfour-amzn: Whoops. Accidentally cut-and-pasted this from a different test file. Removed it from both mine and the other file. Turns out they weren't needed in either one, making them even *more* egregious. :(

lumberyard-employee-dm (approved):
Thanks!

AMZN-stankowi (approved):

PR #14633, opened 2023-02-16
Reduced recompile time when changing O3DE_STACK_CAPTURE_DEPTH

lumberyard-employee-dm:
CMake contains a cache variable called `O3DE_STACK_CAPTURE_DEPTH` which determines the amount of stack frames to capture when a memory leak occurs in the C++ UnitTest.

When that cache variable is changed, it modified the public COMPILE_DEFINITION of `O3DE_STACK_CAPTURE_DEPTH` on the `O3DEKernel` library target.

As the `O3DEKernel` is a dependency of AzCore and AzCore is a dependency on all the other libraries within the code base, this required the entire code base to recompile when the option is changed.

To reduce the time needed to recompile, the `O3DE_STACK_CAPTURE_DEPTH` CMake cache variable will only change the COMPILE_DEFINITION for the `OSAllocator.cpp`, `PoolAllocator.cpp` and `SystemAllocator.cpp`.

Therefore only those 3 files need to recompile.
There is still a cost to relinking AzCore to other libraries, but is a lot less than recompiling every source file.

## How was this PR tested?

Validated that the O3DE_STACK_CAPTURE_DEPTH define were only set for the source files of `OSAllocator.cpp`, `PoolAllocator.cpp` and `SystemAllocator.cpp`
image

image

image

As can be seen it is not set for other files such as Memory.cpp
image

### DIFF
--- a/Code/Framework/AzCore/CMakeLists.txt
+++ b/Code/Framework/AzCore/CMakeLists.txt
@@ -51,9 +51,6 @@ ly_add_target(
             .
             ${pal_dir}
             ${common_dir}
-    COMPILE_DEFINITIONS
-        PUBLIC
-            O3DE_STACK_CAPTURE_DEPTH=${O3DE_STACK_CAPTURE_DEPTH}
 )
 
 ly_add_target(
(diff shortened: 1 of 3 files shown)


Code/Framework/AzCore/CMakeLists.txt
```diff
@@ -96,6 +93,18 @@ ly_add_source_properties(
...
+# This reduces re-compile time when the cache variable value changes.
+set(O3DE_STACK_CAPTURE_DEPTH_DEFINE $<$<NOT:$<STREQUAL:"${O3DE_STACK_CAPTURE_DEPTH}","">>:O3DE_STACK_CAPTURE_DEPTH=${O3DE_STACK_CAPTURE_DEPTH}>)
+ly_add_source_properties(
+    SOURCES
+        AzCore/Memory/OSAllocator.cpp
+        AzCore/Memory/PoolAllocator.cpp
+        AzCore/Memory/SystemAllocator.cpp
+    PROPERTY COMPILE_DEFINITIONS
+    VALUES ${O3DE_STACK_CAPTURE_DEPTH_DEFINE}
+)
```
mbalfour-amzn, line 106: Instead of adding O3DE_STACK_CAPTURE_DEPTH_DEFINE here, since O3DE_STACK_CAPTURE_DEPTH is already in Configurations.cmake, it seems like this should just add "O3DE_STACK_CAPTURE_DEPTH=${O3DE_STACK_CAPTURE_DEPTH}" to those three files instead of introducing a level of indirection.

srikappa-amzn (approved):
Thanks for adding this!

Code/Framework/AzCore/CMakeLists.txt
lumberyard-employee-dm: Actually there is a problem if user if the user overrides the `O3DE_STACK_CAPTURE_DEPTH` to be empty via `-DO3DE_STACK_CAPTURE_DEPTH=`, then it results in a compiler define of `O3DE_STACK_CAPTURE_DEPTH=` which I believe doesn't work everywhere
image

mbalfour-amzn (approved):

lawsonamzn (approved):

PR #14636, opened 2023-02-17
Fix an issue RenderPass::BindSrg function

VickyAtAZ:
## What does this PR do?
Sometime the bindSrg is called for a pass which won't be rendered. This leads to issue that the m_shaderResourceGroupsToBind vector grow without proper clear.

This fix is to solve the issue by setting srg for based on its binding slot (which is limited)

## How was this PR tested?
Editor with AutomatedTesting project
ASV
robotec's test level with rbgd sensors

### DIFF
--- a/Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Pass/RenderPass.h
+++ b/Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Pass/RenderPass.h
@@ -153,7 +153,7 @@ namespace AZ
 
             // List of all ShaderResourceGroups to be bound during rendering or computing
             // Derived classed may call BindSrg function to add other srgs the list
-            ShaderResourceGroupList m_shaderResourceGroupsToBind;
+            AZStd::unordered_map<uint8_t, const RHI::ShaderResourceGroup*> m_shaderResourceGroupsToBind;
             
             // View tag used to associate a pipeline view for this pass.
             PipelineViewTag m_viewTag;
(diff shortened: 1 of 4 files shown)


Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Pass/RenderPass.h
```diff
@@ -153,7 +153,7 @@ namespace AZ
 
             // List of all ShaderResourceGroups to be bound during rendering or computing
             // Derived classed may call BindSrg function to add other srgs the list
-            ShaderResourceGroupList m_shaderResourceGroupsToBind;
+            AZStd::unordered_map<uint8_t, const RHI::ShaderResourceGroup*> m_shaderResourceGroupsToBind;
```
rgba16f, line 156: This changes from no memory allocations to num entries + 1, How bad of an error does it fix?

rgba16f: would fixed_unordered_map work?

dmcdiarmid-ly (approved):

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Pass/RenderPass.h
VickyAtAZ: Basically the BindSrg became a public function due a special need from terrain FP. 
We can't control how often the function is called so it might be easily broken if it's called many time before the pass is rendered. 
it's not the best fix since I don't like that it won't be cleared when a pass is not rendered because any reasons. But it's a ok solution for the current problem. 
I am collecting similar problems for the pass system so I may come back for a better solution for these kind of problems.

moudgils (approved):

rgba16f (approved):

PR #14638, opened 2023-02-17
Adding proper BehaviorContext reflection of AZ Path

lumberyard-employee-dm:
The AZ::IO::Path reflected functions model the python Pathlib API and use the same method names.

Added AsUri function and (StringAsUri/FixedAstringAsUri) functions to `AZ::IO::Path*` classes to allow a path to be encoded as a uri using the `file://` protocol

Other Changes:
* Moved the reflection for AZStd::string and AZStd::string_view to the AzStdOnDemandReflectionSpecializations.cpp file to reduce compilation time
  * Added fixed_string<1024> BehaviorContext reflection also inside of the AzStdOnDemandReflectionSpecializations.cpp file
* Added a utility function for retriving a project's product folder for a platform and the project's log folder

## How was this PR tested?

Added UnitTest that validates the reflected AZ IO Path Scripting functions as well as the `AsUri` function

### DIFF
--- a/Code/Framework/AzCore/AzCore/IO/Path/Path.cpp
+++ b/Code/Framework/AzCore/AzCore/IO/Path/Path.cpp
@@ -132,7 +132,7 @@ namespace AZ::IO::Internal
     }
 
     template <class StringType>
-    constexpr StringType AsUri(const PathView& pathView) noexcept
+    StringType AsUri(const PathView& pathView) noexcept
     {
         StringType resultUri;
         // The syntax for the "file:" URI scheme is documented at https://www.rfc-editor.org/rfc/rfc8089#page-3
--- a/Code/Framework/AzCore/AzCore/IO/Path/Path.cpp
+++ b/Code/Framework/AzCore/AzCore/IO/Path/Path.cpp
@@ -50,4 +50,122 @@ namespace AZ::IO
         const PathIterator<const Path>& rhs);
     template bool operator!=<const FixedMaxPath>(const PathIterator<const FixedMaxPath>& lhs,
         const PathIterator<const FixedMaxPath>& rhs);
+
+    // string conversion
+    AZStd::string PathView::String() const
+    {
+        return AZStd::string(m_path);
+    }
+
+    // as_posix
+    AZStd::string PathView::StringAsPosix() const
+    {
+        AZStd::string resultPath(m_path);
+        AZStd::replace(resultPath.begin(), resultPath.end(), AZ::IO::WindowsPathSeparator, AZ::IO::PosixPathSeparator);
+        return resultPath;
+    }
+
+    // as_uri
+    AZStd::string PathView::StringAsUri() const
+    {
+        return Internal::AsUri<AZStd::string>(*this);
+    }
+}
+
+namespace AZ::IO::Internal
+{
+    //! Quotes a URL like path string
+    //! Reference RFC 3986 for quoting: https://www.rfc-editor.org/rfc/rfc3986
+    //! The unreserved characters are left as is https://www.rfc-editor.org/rfc/rfc3986#section-2.3
+    //! and all reserved characters are URL encoded except for the '/' character as that is needed
+    //! in a URI path
+    static constexpr auto URIQuoteTable = []()
+    {
+        // Hex to Character macro for converting nibbles to characters
+        constexpr AZStd::string_view HexToChar = "0123456789ABCDEF";
+        static_assert(HexToChar.size() == 16);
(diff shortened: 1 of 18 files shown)


Code/Framework/AzCore/AzCore/IO/Path/Path.inl
```diff
@@ -263,6 +283,11 @@ namespace AZ::IO
         return m_path.empty();
     }
 
+    constexpr const char PathView::PreferredSeparator() const noexcept
+    {
+        return m_preferred_separator;
```
AMZN-alexpete, line 288: I'm guessing we went with `m_preferred_separator` instead of `m_preferredSeparator` to more closely model `std::filesystem::path::preferred_separator`?

Code/Framework/AzCore/AzCore/Utils/Utils.cpp
```diff
@@ -241,6 +240,42 @@ namespace AZ::Utils
...
+        if (settingsRegistry != nullptr)
+        {
+            if (AZ::IO::FixedMaxPathString settingsValue;
+                settingsRegistry->Get(settingsValue, AZ::SettingsRegistryMergeUtils::FilePathKey_CacheRootFolder))
+            {
+                return settingsValue;
+            }
+        }
+        return {};
+    }
```
AMZN-alexpete, line 277: we're getting enough of these similar `GetXYZPath` util functions where the only difference seems to be the key - I wonder if we should have a 
`AZ::IO::FixedMaxPathString GetPathFromKey(settingsRegistry, key)` function to DRY things up a bit.

Code/Framework/AzCore/AzCore/IO/Path/Path.cpp
```diff
@@ -50,4 +50,121 @@ namespace AZ::IO
...
+            }
+        }
+
+        return percentQuoteTable;
+    }();
+
+    // Quick validation of some of quote conversions
+    static_assert(URIQuoteTable['a'] == "a");
+    static_assert(URIQuoteTable['-'] == "-");
+    static_assert(URIQuoteTable['@'] == "%40");
```
AMZN-alexpete, line 117: ' ' == "%20" is also a good one

Code/Framework/AzCore/AzCore/RTTI/AzStdOnDemandReflectionSpecializations.cpp
```diff
@@ -9,18 +9,19 @@
...
+#include <AzCore/StringFunc/StringFunc.h>
 
 #include <AzCore/RTTI/AzStdOnDemandReflection.inl>
 #include <AzCore/RTTI/AzStdOnDemandReflectionLuaFunctions.inl>
+
 namespace AZ::CommonOnDemandReflections
 {
-    void ReflectCommonString(ReflectContext* context)
+    template<class T>
+    void ReflectCommonStringAPI(ReflectContext* context)
```
AMZN-alexpete, line 22: I only die a little inside when I see these classes reflected in lua like basic_string_char_traits_char_ref, I'm guessing if we provided some friendlier class names like "String" it would break existing code?

Code/Framework/AzCore/AzCore/IO/Path/Path.inl
lumberyard-employee-dm: The original implementation was based on `std::filesystem::path` before it was transitioned over to be based on python Pathlib.

Code/Framework/AzCore/AzCore/IO/Path/Path.cpp
lumberyard-employee-dm: I'll add that as well

Code/Framework/AzCore/AzCore/RTTI/AzStdOnDemandReflectionSpecializations.cpp
lumberyard-employee-dm: We do provide friendlier names for [ScriptCanvas](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/RTTI/AzStdOnDemandPrettyName.inl#L53-L60) at least

Code/Framework/AzCore/AzCore/Utils/Utils.cpp
lumberyard-employee-dm: That would decrease the number of lines of code a bit

AMZN-alexpete (approved):
Nice!

cgalvan (approved):
Thanks for adding this!

PR #14643, opened 2023-02-17
Fix the bug there are deleted material infos in .prefab file

Zhang-Baochong:
## What does this PR do?

1、Create a new entity, add Mesh Componet, selete a mesh (such as the chicken fbx which has three material slots) 
2、Add Material of the Mesh, and select material assets, and save scene.

3、Select a different Model Asset of the same Entity,  such as the shaderball fbx, add Material and save the scene.

4、Open the .prefab file of the scene, there is previous residual info of the chicken materials.

## How was this PR tested?
Follow the same steps, and the residual material infos has been deleted.

### DIFF
--- a/Gems/AtomLyIntegration/CommonFeatures/Code/Source/Mesh/EditorMeshComponent.cpp
+++ b/Gems/AtomLyIntegration/CommonFeatures/Code/Source/Mesh/EditorMeshComponent.cpp
@@ -254,6 +254,7 @@ namespace AZ
                 &AzToolsFramework::ToolsApplicationEvents::InvalidatePropertyDisplay,
                 AzToolsFramework::Refresh_EntireTree);
 
+            MaterialComponentRequestBus::Event(GetEntityId(), &MaterialComponentRequestBus::Events::ClearMaterialsOnInvalidSlots);
         }
 
         AZ::u32 EditorMeshComponent::OnConfigurationChanged()


gadams3 (changes requested):
This behavior is intentional but worth discussing because it’s come up in other conversation recently.

The material assignments are supposed to persist so that they can be reapplied to a model with compatible material slots or if the component is copied and pasted to another entity.

I recently added better handling and messaging when these residual materials are carried around. If materials are missing, the remaining materials will still be applied.

The residual materials can be cleaned up from the material component context menu, which has several options for managing materials.

We can also add a label or button onto the material component UI to inform users that these residual materials are present and can be cleared from the button or context menu.

In any event, this type of handling should be done in the material component itself since it can be used in conjunction with other components like actors. It should go in the editor material component in a way that does not break undo redo, which is handled correctly in the context menu actions.

Zhang-Baochong:
> This behavior is intentional but worth discussing because it’s come up in other conversation recently.
> 
> The material assignments are supposed to persist so that they can be reapplied to a model with compatible material slots or if the component is copied and pasted to another entity.
> 
> I recently added better handling and messaging when these residual materials are carried around. If materials are missing, the remaining materials will still be applied. #14345
> 
> The residual materials can be cleaned up from the material component context menu, which has several options for managing materials.
> 
> We can also add a label or button onto the material component UI to inform users that these residual materials are present and can be cleared from the button or context menu.
> 
> In any event, this type of handling should be done in the material component itself since it can be used in conjunction with other components like actors. It should go in the editor material component in a way that does not break undo redo, which is handled correctly in the context menu actions.

This problem attracts my attention because using the scriptCanvas GetMaterialOverrides will get the previous material information, which may cause problems in setting material parameters. For users, it should not get materials that do not belong to this mesh, but the implementation really needs to be carefully analyzed.

gadams3:
> This problem attracts my attention because using the scriptCanvas GetMaterialOverrides will get the previous material information, which may cause problems in setting material parameters. For users, it should not get materials that do not belong to this mesh, but the implementation really needs to be carefully analyzed.

Between the handling and the warnings that I added in the other PR and the option to add a label or button to the component, that should help clarify the state of things and the available options without breaking portability between models and materials or the undo redo stack in the editor. 

If I understand correctly, you expect to be able to request a list of material overrides that are guaranteed to be applicable to the current model. While it's not ideal, your script could call the function to clear all of the invalid materials when it starts. We could add another method that only returns the material overrides applicable to the current model. 

A user can change the current model or any of the material overrides at runtime from script canvas or lua. That is another reason we don't want to clear the overrides whenever the component activates or model changes. The script could just be switching to a model or material that reflects a different state, like a damaged state of an object.

Prior to prefabs, there was a special build game entity function where the editor version of a component had the opportunity to make any kind of modifications to the game component configuration before saving it. That would be the perfect opportunity to clean up materials specifically for game mode or the launcher if something like that still exists. I will investigate while we consider other options.

gadams3:
I see that you closed this but I was going to mention then I confirmed that the editor component BuildGameEntity function still exists and allows you to customize the component and configuration that gets exported for spawnables. We can override the function on the editor material component to generate a new material component and controller with these invalid materials removed. Then everything will stay in place for the editor component but they will be removed from the game component used in game mode and the launcher.

Issue #13568, opened 2022-12-06
Atom RHI should avoid selecting llvmpipe as it doesn't support the required features


rgba16f:
**Describe the bug**
If the Atom RHI selects llvmpipe as it's vulkan gpu the executable will immediately crash with a missing api error

**Steps to reproduce**
1. Launch editor or game launcher on a machine that only has llvmpipe

**Expected behavior**
The user should see a graceful fail explaining that llvmpipe is not a supported gpu

**Actual behavior**
App crashes on startup

**Screenshots/Video**
If applicable, add screenshots and/or a video to help explain your problem.

**Found in Branch**
22.10, development

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Fedora
 - Version 35
 - CPU Intel i7
 - GPU llvmpipe & intel
 - Memory 

**Additional context**
log output
> RHISystem: Initializing RHI...
RHISystem: 	Enumerated physical device: llvmpipe (LLVM 13.0.0, 256 bits)
RHISystem: 	Enumerated physical device: Intel(R) UHD Graphics 620 (WHL GT2)
RHISystem: 	Using physical device: llvmpipe (LLVM 13.0.0, 256 bits)
WARNING: lavapipe is not a conformant vulkan implementation, testing use only.
System: 
==================================================================
System: Trace::Assert
 /data/workspace/o3de/Gems/Atom/RHI/Vulkan/Code/Source/RHI/Vulkan.h(96): (140080659850560) 'void AZ::Vulkan::AssertSuccess(VkResult)'
System: ASSERT: Vulkan API method failed: Feature not present
System: ------------------------------------------------
System: std::__throw_bad_optional_access() (+0x26ca8) [0x7f66913db538]
System: std::__throw_bad_optional_access() (+0x958b23) [0x7f66807d5f63]
System: std::__throw_bad_optional_access() (+0x8e0be0) [0x7f668075e020]
System: std::__throw_bad_optional_access() (+0x8e03a6) [0x7f668075d7e6]
System: std::__throw_bad_optional_access() (+0x51e599) [0x7f668039b9d9]
System: std::__throw_bad_optional_access() (+0xf6f1d) [0x7f667ff7435d]
System: OutputDebugString(char const*) (+0x37bbf5) [0x7f670c5f3d25]
System: OutputDebugString(char const*) (+0x37c808) [0x7f670c5f4938]
System: OutputDebugString(char const*) (+0x37c897) [0x7f670c5f49c7]
[...]

PR #14644, opened 2023-02-17
Possibility to initialize multiple RHI devices.

jhmueller-huawei:
## What does this PR do?

This PR makes it possible to initialize multiple devices in the RHISystem, but does not make them usable yet. It also removes some unused/unnecessary functions.

It is a preparation for the bigger changes coming for multi-device support, where a much bigger pull request is planned to add multi-device resources first. See: https://github.com/o3de/sig-graphics-audio/issues/120

## How was this PR tested?

Functionality actually does not change at all. Tests were carried out running the atom-sampleviewer and checking out some examples.


Gems/Atom/RHI/Code/Source/RHI/RHISystem.cpp
```diff
@@ -270,19 +288,14 @@ namespace AZ
 
         ConstPtr<PlatformLimitsDescriptor> RHISystem::GetPlatformLimitsDescriptor() const
```
moudgils, line 289: lets add a default device index (defaulted to 0) for this method as well.

Gems/Atom/RHI/Code/Source/RHI/Device.cpp
```diff
@@ -75,6 +75,7 @@ namespace AZ
             m_nearestSupportedFormats.fill(Format::Unknown);
             
             m_physicalDevice = &physicalDevice;
+            m_index = index;
```
moudgils, line 78: Lets add an assert here.
AZ_Assert(index > 0, "Multi gpu support is not implemented yet.")

Gems/Atom/RHI/Code/Source/RHI/RHISystem.cpp
jhmueller-huawei: I wonder how device specific `PlatformLimitsDescriptor` is/should be? Maybe it would make sense to split it up? E.g., separate out the `PlatformDefaultValues`. Some parts appear to be generic/global, while other parts might depend on the backend or on the device?

rgba16f (approved):
Overall the change looks good, I'd like to see unit tests of the multi-gpu init as a follow on change.

Gems/Atom/RHI/Code/Source/RHI/RHISystem.cpp
```diff
@@ -96,92 +97,113 @@ namespace AZ
...
-            RHI::PhysicalDevice* preferredUserDevice{};
-            RHI::PhysicalDevice* preferredVendorDevice{};
+            if (AzFramework::StringFunc::Equal(multipleDevicesValue.c_str(), "enable"))
+            {
+                AZ_Printf("RHISystem", "\tUsing multiple devices\n");
 
-            for (RHI::Ptr<RHI::PhysicalDevice>& physicalDevice : physicalDevices)
+                usePhysicalDevices = AZStd::move(physicalDevices);
+            }
+            else
```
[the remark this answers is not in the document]
jhmueller-huawei: How would you go about deciding, which devices to include?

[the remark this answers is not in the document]
rgba16f: The physical device descriptor already classifies GPUs as either Discrete, Integrated, Virtual, CPU, or Fake (Mock for unit test, etc). I would guess that we probably want options to limit multiGPU to either Only Discrete, All local HW (Discrete + integrated), or All hw (Discrete, Integrated, virtual). We may also need an option to allow multiple Fake GPU's for unit tests.

I'm unaware of a reason we would want to run multiple Software GPU instances, or include a software GPU in the group of GPUS.

I don't think we need this added as a requirement for this change.

We've previously identified that selecting LLVMPipe as the gpu is something we want to avoid (unless explicitly called for) - https://github.com/o3de/o3de/issues/13568

Issue #13398, opened 2022-11-25
Changing Translation Offset in entity with Noise Gradient is not reflected on Terrain Layer Spawner entity


LB-KacperKapusta:
**Describe the bug**
Modifying Translation Offset in entity with Noise Gradient component targeting a Terrain Layer Spawner entity does not change the terrain being generated. Although, the change is reflected in the Preview window of the Noise Gradient component.

**Prerequisites**

- Translation Offset mode enabled - Add "EnableShapeComponentTranslationOffset": true under "Amazon" -> "Preferences" in editorpreferences.setreg.
- A project with enabled Terrain gem is built.

**Steps to reproduce**
1. Add Terrain World and Terrain World Renderer components to Level entity.
2. Create Entity1 with Terrain Layer Spawner, Axis Aligned Box Shape and Terrain Height Gradient List components.
3. Change Axis Aligned Box Shape's Dimensions to e.g. X:10, Y:10 and move the Entity1 above Z:0 for the terrain to be rendered.
4. Create Entity2 with Perlin or Random Noise Gradient component and add the required components to it (as shape use Axis Aligned Box Shape, Box Shape, Capsule Shape or Sphere Shape component).
5. Select Entity1 and choose Entity2 as target in Terrain Height Gradient List component.
6. Select Entity2 and modify Translation Offset in its shape component.

**Expected behavior**
The generated terrain updates as the Noise Gradient entity's Preview does.

**Actual behavior**
The Noise Gradient entity's Preview updates but the change is not reflected on the generated terrain.

**Screenshots/Video**

https://user-images.githubusercontent.com/86953659/204005712-d006eff2-6d85-4c04-8550-32be06efb94b.mp4

**Found in Branch**
Development (da550ef)

**Desktop/Device**
- Device: PC
- OS: Windows
- Version: 10
- CPU: AMD Ryzen 5 3600
- GPU: NVIDIA GeForce RTX 2060 SUPER
- Memory: 16GB

**Additional context**

greerdv:
Talking to @mbalfour-amzn it seems like this is the expected behaviour, although it did lead to discovering a couple of other issues (https://github.com/o3de/o3de/pull/14648 https://github.com/o3de/o3de/issues/14650)

greerdv:
Closing as by design

LB-KacperKapusta:
Confirming the issue has been closed as by design as per comment above.

PR #14648, opened 2023-02-17
allow gradient transform to handle asymmetrical shapes

greerdv:
## What does this PR do?
Allows gradient transforms to handle shapes which are not symmetrical about their centre e.g. shapes with translation offsets, polygon prisms. Related to #13398.

## How was this PR tested?
Manual testing in editor and running unit tests.

### DIFF
--- a/Gems/GradientSignal/Code/Include/GradientSignal/Components/GradientTransformComponent.h
+++ b/Gems/GradientSignal/Code/Include/GradientSignal/Components/GradientTransformComponent.h
@@ -55,6 +55,7 @@ namespace GradientSignal
 
         bool m_overrideBounds = false;
         AZ::Vector3 m_bounds = AZ::Vector3::CreateOne(); //1m sq default value chosen by design, to start small and expand as needed
+        AZ::Vector3 m_center = AZ::Vector3::CreateZero(); // to handle asymmetrical shapes such as polygon prism
 
         TransformType m_transformType = TransformType::World_ThisEntity;
         bool m_overrideTranslate = false;
@@ -129,6 +130,9 @@ namespace GradientSignal
 
         AZ::Vector3 GetBounds() const override;
         void SetBounds(AZ::Vector3 bounds) override;
+
+        AZ::Vector3 GetCenter() const override;
+        void SetCenter(AZ::Vector3 center) override;
 
         TransformType GetTransformType() const override;
         void SetTransformType(TransformType type) override;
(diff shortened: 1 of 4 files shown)


Gems/GradientSignal/Code/Source/Components/GradientTransformComponent.cpp
```diff
@@ -83,7 +84,9 @@ namespace GradientSignal
                     ->Attribute(AZ::Edit::Attributes::ReadOnly, &GradientTransformConfig::IsAdvancedModeReadOnly)
                     ->DataElement(0, &GradientTransformConfig::m_bounds, "Bounds", "Local (untransformed) bounds of a box used to remap, clamp, wrap, scale incoming coordinates")
                     ->Attribute(AZ::Edit::Attributes::ReadOnly, &GradientTransformConfig::IsBoundsReadOnly)
-                    
+                    ->DataElement(0, &GradientTransformConfig::m_bounds, "Center", "Local (untransformed) center of a box used to remap, clamp, wrap, scale incoming coordinates")
```
mbalfour-amzn, line 87: If this stays here, it should use m_center, not m_bounds.

greerdv: good catch, thanks

mbalfour-amzn (approved):

Gems/GradientSignal/Code/Include/GradientSignal/Components/GradientTransformComponent.h
```diff
@@ -130,6 +131,9 @@ namespace GradientSignal
         AZ::Vector3 GetBounds() const override;
         void SetBounds(AZ::Vector3 bounds) override;
 
+        AZ::Vector3 GetCenter() const override;
+        void SetCenter(AZ::Vector3 center) override;
```
moraaar, line 135: The parameter in `SetCenter` is passed by value and therefore making copies. Any particular reason for not using const ref?

Same for `SetBounds`, `SetTranslate`, `SetRotate` and `SetScale`. It'd be nice to use const ref in all of them.

moraaar (approved):

Gems/GradientSignal/Code/Include/GradientSignal/Components/GradientTransformComponent.h
greerdv: I wondered about that too, but I just copied what all the other functions here were doing.

greerdv: ok, can do.

greerdv: updated

moraaar: Thanks!

PR #14657, opened 2023-02-18
Speed up GetProjects() and set local "engine_path" in user/project.json

AMZN-alexpete:
## What does this PR do?

- makes GetProjects() more efficient by only loading `engine.json` files once.  Made Project page refresh go from ~500ms to ~150ms.
- set local "engine_path" when selecting an engine locally 
- have Project Properties `engine` combo show when an engine is not specified or an engine is unregistered:
image
- have Project Properties show engine version in `engine` combo box
image

## How was this PR tested?

- verified project properties page showed correct engine assignment and path
- verified project could be assigned to an engine path when multiple engines with same name and version were registered
- verified project showed correct path after assigning and returning to the project properties page
- ran o3de unit tests and verified they passed


Code/Tools/ProjectManager/Source/ProjectsScreen.cpp
```diff
@@ -301,6 +301,11 @@ namespace O3DE::ProjectManager
                 }
             });
 
+
+            // It's more efficient to update the project engine by loading engine infos once
+            // instead of loading them all each time we want to know what project an engine uses
+            auto engineInfoResult = PythonBindingsInterface::Get()->GetAllEngineInfos();
```
lumberyard-employee-dm, line 307: How long does this take if have 10 engines registered?

We don't want it to be unresponsive for the user opening the project dialog the first time.

I use 10 engines registered, because I don't think a user would any more than 10 engines that are actually found on the machine and not stale entries.
So that should be good cap on long it would take.

scripts/o3de/o3de/project_manager_interface.py
```diff
@@ -138,7 +139,39 @@ def get_all_project_infos() -> list:
...
+    engines = manifest.get_manifest_engines()
+    for engine in engines:
+        if isinstance(engine, dict):
+            engine_path = pathlib.Path(engine['path']).resolve()
+        else:
+            engine_path = pathlib.Path(engine).resolve()
+        engine_json_data = manifest.get_engine_json_data(engine_path=engine_path)
+        if not engine_json_data:
+            continue
+        engines_json_data[engine_path] = engine_json_data
```
lumberyard-employee-dm, line 155: This logic seems like it should be in a method on its own.
i.e `refresh_engine_info_cache`

Code/Tools/ProjectManager/Source/ProjectsScreen.cpp
AMZN-alexpete: Let me pull the numbers - I made this change because without it the Project Manager was much slower because it was inefficient in how it was asking for project data.  I really want us to move all our `PythonBindings` calls to the `project-manager-interface.py` where we can safely implement more efficient use of o3de python code and even limited caching.  I think we may need to address this so the UI doesn't freeze up if it tries to display 30+ projects and 1k+ gems that it knows about because of remote repos.  Some of the pages that will pull in remote content will need to use async updates, pagination at some point.

AMZN-alexpete: With 12 projects and 9 engines this whole function (ResetProjectContents) takes ~500ms before these changes and ~150ms after based on 2-3 runs in each implementations.

This specific line of code `GetAllEngineInfos` takes ~65ms to load and parse 9 `engine.json`s found in the `o3de_manifest.json` so not really noticeable unless every time you get the project info you are loading every `engine.json` which is what was happening before.  Still lots of room for improvement in startup time - this function is run 3 times alone on startup!

scripts/o3de/o3de/project_manager_interface.py
AMZN-alexpete: I like it, something similar to `get_gems_json_data_by_name`

AMZN-Phil (approved):

Code/Tools/ProjectManager/Source/ProjectsScreen.cpp
lumberyard-employee-dm: That is good enough and won't be noticeable by the user, if we do it on startup.

tkothadev (approved):

Code/Tools/ProjectManager/Resources/ProjectManager.qss
```diff
@@ -710,7 +710,7 @@ QTabBar::tab:focus {
 }
 
 #projectButton QLabel#otherEngineLabel {
-    color: #dd0;
+    color: #ddd;
```
lumberyard-employee-dm, line 713: So the color here is now #dddddd.
So a very light grey?

scripts/o3de/o3de/manifest.py
```diff
@@ -404,6 +405,20 @@ def get_gem_templates(gem_path: pathlib.Path) -> list:
                         gem_object['templates'])) if 'templates' in gem_object else []
     return []
 
+def get_engines_json_data_by_path():
+    # use an OrderedDict to preserve the order found in the o3de manifest
+    engines_json_data = OrderedDict()
```
lumberyard-employee-dm, line 410: Python dictionaries maintain order as of 3.7, so the OrderedDict collection is no longer needed.
https://mail.python.org/pipermail/python-dev/2017-December/151283.html

lumberyard-employee-dm (approved):

Code/Tools/ProjectManager/Resources/ProjectManager.qss
AMZN-alexpete: yes, very close to white, but not pure white

scripts/o3de/o3de/manifest.py
AMZN-alexpete: oh cool! Not sure how I missed this.

PR #14661, opened 2023-02-20
add parent widget to StringDlg object.

Jackie9527:
## What does this PR do?
I have two screens.
1. Open Editor, at 1st screen.
2. Click Tools -> Track View.
3. Move `Track View` to 2nd screen.
4. Create a sequence, and create a folder, as follows:
image

## How was this PR tested?
image


AMZN-daimini (approved):

Jackie9527:
> Change looks good but any chance of adding a higher resolution image next time? The screen shot is super helpful but it's a little hard to see what's going on as it's quite small. Thanks!

Sorry about that.
I will provide higher resolution images next time.

Issue #6094, opened 2021-12-02
Track View tracks become disabled when playing a sequence in play mode with Track View open


windbagjacket:
**Describe the bug**
Tracks in Track View become invalid/disabled when playing a sequence via play mode or rendering to file, while it is open in Track View. 

**Steps to reproduce**
1. Create a sequence in TrackView and add a track to translate an entity. Set the sequence to auto-play.
2. With the sequence open in TrackView, enter play mode in the main editor window. Or, render the sequence to file.
3. Exit play mode, and check TrackView.

**Expected behavior**
The sequence should appear unchanged in Track View.

**Actual behavior**
The track turns red and the track names (e.g. "translation" and "rotation") become "Unknown". 
Workaround: restart Track View, or disable and enable the parent object of each track.

**Found in Branch**
stabilization/2110

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Windows
 - Version: 10
 - CPU: Intel i7-10750H
 - GPU: NVidia RTX 2080 Super
 - Memory: 32GB

PR #14662, opened 2023-02-20
bugfix: light color on track is inconsistent.

Jackie9527:
## What does this PR do?
fix color change rule from editor to track.

1. Open Editor, create a level, create an entity with Light Component, named as `light1`.
2. Open Track View.
3. Create a sequence.
4. Add `light1` to sequence, add Color track.
5. Set color value as `255, 0, 0`.
image
6. Create a key frame at color track, set color value as `0, 255, 0`.
image
7. Click the key frame, open the color picker dialog, don't make any change, just click `Cancel`, then color value is changed.
image

## How was this PR tested?
Follow the same steps.

### DIFF
--- a/Code/Editor/TrackView/TrackViewDopeSheetBase.cpp
+++ b/Code/Editor/TrackView/TrackViewDopeSheetBase.cpp
@@ -2051,7 +2051,8 @@ void CTrackViewDopeSheetBase::OnCurrentColorChange(const AZ::Color& color)
 
 void CTrackViewDopeSheetBase::UpdateColorKey(const QColor& color, bool addToUndo)
 {
-    ColorF colArray(static_cast<f32>(color.redF()), static_cast<f32>(color.greenF()), static_cast<f32>(color.blueF()), static_cast<f32>(color.alphaF()));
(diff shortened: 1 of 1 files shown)


hultonha (approved):
Thanks for fixing all these TrackView issues @Jackie9527 👍 If you're in this area of the code and want to take a look at https://github.com/o3de/o3de/issues/6094 that would be amazing! 😅 Thanks again!

AMZN-daimini (approved):

Jackie9527:
> Thanks for fixing all these TrackView issues @Jackie9527 👍 If you're in this area of the code and want to take a look at #6094 that would be amazing! 😅 Thanks again!

ok, I will follow #6094.

hultonha:
> > Thanks for fixing all these TrackView issues @Jackie9527 👍 If you're in this area of the code and want to take a look at #6094 that would be amazing! 😅 Thanks again!
> 
> ok, I will follow #6094.

That's very kind! 😄 If it it looks like a rabbit hole/nightmare let me know, it's just something I spotted recently that I'd love to fix but have been struggling to find time for I'm afraid. Feel free to ping me if you have any questions. Cheers!

PR #14665, opened 2023-02-20
fix uint64 value serialize/deserialize by CXmlNode.

Jackie9527:
## What does this PR do?
Currently, due to serialize/deserialize problems, we cannot copy a node between sequences.

1. Open Editor.
2. Click Tools -> Track View.
3. Create two sequence, and add some nodes, as follows:
image
image
4. Right click a node of sequence 1, copy it.
image
5. Change to sequence 2.
6. Right click, and paste.
image
7. The node of sequence 1 fails to copy to sequence 2.
image

## How was this PR tested?
image

image

### DIFF
--- a/Code/Legacy/CrySystem/XML/xml.cpp
+++ b/Code/Legacy/CrySystem/XML/xml.cpp
@@ -320,7 +320,7 @@ void CXmlNode::setAttr(const char* key, int64 value)
 //////////////////////////////////////////////////////////////////////////
 void CXmlNode::setAttr(const char* key, uint64 value, bool useHexFormat)
 {
-    char str[32];
+    char str[32] = { 0 };
     if (useHexFormat)
     {
         sprintf_s(str, "%" PRIX64, value);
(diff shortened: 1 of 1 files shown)


moraaar (approved):

Code/Legacy/CrySystem/XML/xml.cpp
```diff
@@ -320,7 +320,7 @@ void CXmlNode::setAttr(const char* key, int64 value)
 //////////////////////////////////////////////////////////////////////////
 void CXmlNode::setAttr(const char* key, uint64 value, bool useHexFormat)
 {
-    char str[32];
+    char str[32] = { 0 };
```
hultonha, line 323: nit: Rather than zero initializing the whole buffer, it can be preferable to write `'\0'` to the first element...

```
char str[32];
str[0] = '\0';
```

In this case it's not a huge deal but for large buffers you might not want to pay the cost of zeroing them before writing some other values to them.

hultonha (approved):

PR #14666, opened 2023-02-20
Scene queries optimization: Exposed QueryScene function doing no allocations

SergeyAMZN:
## What does this PR do?

Scene queries optimization: Exposed `QueryScene` function doing no allocations + exposed Shape::GetMaterialId to prevent using `shared_ptr` in a performance critical code. 

Benchmark creating 32 threads running 1 million raycasts against 128 boxes, measuring overall execution time:
Before: 48583 ms 
After: 456 ms
CPU: AMD Ryzen Threadripper 3970X (32 cores)

image

## How was this PR tested?

Unit tests, benchmarks.

### DIFF
--- a/Code/Framework/AzFramework/AzFramework/Physics/PhysicsScene.cpp
+++ b/Code/Framework/AzFramework/AzFramework/Physics/PhysicsScene.cpp
@@ -51,7 +51,10 @@ namespace AzPhysics
                 ->Attribute(AZ::Script::Attributes::Category, "Physics")
                 ->Method("GetOnGravityChangeEvent", getOnGravityChange)
                     ->Attribute(AZ::Script::Attributes::AzEventDescription, gravityChangedEventDescription)
-                ->Method("QueryScene", &Scene::QueryScene)
+                ->Method("QueryScene", [](Scene* selfPtr, const SceneQueryRequest* request)
+                {
+                    return selfPtr->QueryScene(request);
+                })
                 ;
         }
     }
--- a/Code/Framework/AzFramework/AzFramework/Physics/PhysicsScene.cpp
+++ b/Code/Framework/AzFramework/AzFramework/Physics/PhysicsScene.cpp
@@ -51,9 +51,9 @@ namespace AzPhysics
                 ->Attribute(AZ::Script::Attributes::Category, "Physics")
                 ->Method("GetOnGravityChangeEvent", getOnGravityChange)
                     ->Attribute(AZ::Script::Attributes::AzEventDescription, gravityChangedEventDescription)
-                ->Method("QueryScene", [](Scene* selfPtr, const SceneQueryRequest* request)
+                ->Method("QueryScene", [](Scene* self, const SceneQueryRequest* request)
                 {
-                    return selfPtr->QueryScene(request);
+                    return self->QueryScene(request);
                 })
                 ;
         }
--- a/Code/Framework/AzFramework/AzFramework/Physics/PhysicsScene.h
+++ b/Code/Framework/AzFramework/AzFramework/Physics/PhysicsScene.h
@@ -132,6 +132,13 @@ namespace AzPhysics
         //! @return Returns a structure that contains a list of Hits. Depending on flags set in the request, this may only contain 1 result.
         virtual SceneQueryHits QueryScene(SceneHandle sceneHandle, const SceneQueryRequest* request) = 0;
 
+        //! Make a blocking query into the scene.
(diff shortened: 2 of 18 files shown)


Code/Framework/AzFramework/AzFramework/Physics/PhysicsScene.cpp
```diff
@@ -51,7 +51,10 @@ namespace AzPhysics
                 ->Attribute(AZ::Script::Attributes::Category, "Physics")
                 ->Method("GetOnGravityChangeEvent", getOnGravityChange)
                     ->Attribute(AZ::Script::Attributes::AzEventDescription, gravityChangedEventDescription)
-                ->Method("QueryScene", &Scene::QueryScene)
+                ->Method("QueryScene", [](Scene* selfPtr, const SceneQueryRequest* request)
```
hultonha, line 54: uber nit: Could just use the variable `self` and omit `Ptr` from the name (less Hungarian style notation)

Code/Framework/AzFramework/AzFramework/Physics/PhysicsScene.h
```diff
@@ -132,6 +132,13 @@ namespace AzPhysics
         //! @return Returns a structure that contains a list of Hits. Depending on flags set in the request, this may only contain 1 result.
         virtual SceneQueryHits QueryScene(SceneHandle sceneHandle, const SceneQueryRequest* request) = 0;
 
+        //! Make a blocking query into the scene.
+        //! @param sceneHandle A handle to the scene to make the scene query with.
+        //! @param request The request to make. Should be one of RayCastRequest || ShapeCastRequest || OverlapRequest
+        //! @param result A structure that contains a list of Hits. Depending on flags set in the request, this may only contain 1 result.
+        //! @return Returns true if there is at least one hit.
+        virtual bool QueryScene(SceneHandle sceneHandle, const SceneQueryRequest* request, SceneQueryHits& result) = 0;
```
hultonha, line 140: suggestion: Consider using `AZStd::optional` here as opposed to the `bool` return with out param, it's essentially the same thing and allows RVO to kick in.

Code/Framework/AzFramework/AzFramework/Physics/PhysicsScene.h
```diff
@@ -344,6 +351,12 @@ namespace AzPhysics
         //! @return Returns a structure that contains a list of Hits. Depending on flags set in the request, this may only contain 1 result.
         virtual SceneQueryHits QueryScene(const SceneQueryRequest* request) = 0;
 
+        //! Make a blocking query into the scene.
+        //! @param request The request to make. Should be one of RayCastRequest || ShapeCastRequest || OverlapRequest
+        //! @param result A structure that contains a list of Hits. Depending on flags set in the request, this may only contain 1 result.
+        //! @return Returns true if there is at least one hit.
+        virtual bool QueryScene(const SceneQueryRequest* request, SceneQueryHits& result) = 0;
```
hultonha, line 358: suggestion (repeated): See earlier comment

Gems/PhysX/Code/Source/Scene/PhysXScene.cpp
```diff
@@ -326,15 +326,16 @@ namespace PhysX
                     }
                 }
             }
-            return hits;
+            return status;
         }
 
         //helper to preform a shape cast
```
hultonha, line 332: uber nit: We could add a space after // while we're here 🙂

hultonha (approved):
Absolutely incredible improvement @SergeyAMZN! 😄 Amazing to see, great to see the new benchmark added too (we just need to get these running regularly now to make sure nothing causes the performance to regress).

Code/Framework/AzFramework/AzFramework/Physics/PhysicsScene.h
SergeyAMZN: Do you mean `AZStd::optional<bool>` instead of `bool`? 
Would RVO optimize unnecessary copies of a bool? looks more readable without optional tbh :)

hultonha: No sorry I mean `AZStd::optional<SceneQueryHits>` instead (and no out param)

hultonha: Discussed offline - this API choice was intentional to reduce allocations, sorry for missing this 😅 🙈

PR #14668, opened 2023-02-20
Disabling TIAF for python tests

moraaar:

## What does this PR do?

Python tests signed up for TIAF are executed in the "test_impact_analysis_profile_python" node, which doesn't gate errors and never goes red. This means that if errors were added in code AR will not report them and let the PR be merged. Disabling TIAF temporarily until a safer approach can be taken.

## How was this PR tested?

Before
1. Created branch and added an error to whitebox python tests
2. Run AR on that branch and noticed how it doesn't go red. (test was executed in "test_impact_analysis_profile_python" node and failed, but the node finished green since it doesn't gate)

After
1. Created branch and added an error to whitebox python tests
2. Run AR on that branch and it goes red.


hultonha (approved):
This looks good to me until @jonawals is back to investigate. Definitely keep a close eye on the build machines to see if any failures to crop up! 😅 Thanks @moraaar 👍

SergeyAMZN (approved):

PR #14673, opened 2023-02-20
fix sequence select/edit.

Jackie9527:
## What does this PR do?

1. Open Editor.
2. Click Tools -> Track View.
3. Create a sequence, named as `1`.
4. Add some nodes.
image
5. Change sequence selection to none.
image
6. Click `Edit Sequence Properties` in the toolbar, the dialog shows up with properties of sequence `1`.
image

## How was this PR tested?
image

### DIFF
--- a/Code/Editor/TrackView/TrackViewDialog.cpp
+++ b/Code/Editor/TrackView/TrackViewDialog.cpp
@@ -1279,6 +1279,11 @@ void CTrackViewDialog::OnSequenceComboBox()
         GetIEditor()->GetAnimation()->SetSequence(nullptr, false, false);
         return;
     }
+    if (sel == 0)
+    {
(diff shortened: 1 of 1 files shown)


Code/Editor/TrackView/TrackViewDialog.cpp
```diff
@@ -1279,6 +1279,11 @@ void CTrackViewDialog::OnSequenceComboBox()
         GetIEditor()->GetAnimation()->SetSequence(nullptr, false, false);
         return;
     }
+    if (sel == 0)
```
AMZN-daimini, line 1282: nit: could be turned into a `switch`, or at least an `else if`?

AMZN-daimini (approved):

cgalvan (approved):

PR #14680, opened 2023-02-20
Intermediate asset jobs are now cleared if the source asset no longer emits them

AMZN-stankowi:
## What does this PR do?

Asset processor now checks for leftover common platform jobs from previous runs. If the source asset reprocesses and a common platform job was emitted in a previous run that does not get emitted in this run (either no common platform job, or some job parameter doesn't match like the key), now it properly removes that job + products of that job.

This is a fix for a bug where material types would sometimes leave behind stray intermediate assets. Material types had a check in CreateJobs, and would emit either a job with an intermediate asset product, or a regular product asset, based on data in the material type. If this data changed from emitting an intermediate asset product to emitting the regular product asset, it was not removing the old intermediate asset. This would then cause farther issues, having an orphaned intermediate asset.

## How was this PR tested?

Tested locally with the bug, and then the automated test.

Verified the automated test fails if I remove the bug fix, passes with the bug fix.


Code/Tools/AssetProcessor/native/tests/assetmanager/AssetProcessorManagerTest.cpp
```diff
@@ -706,6 +707,149 @@ TEST_F(AssetProcessorIntermediateAssetTests, NestedIntermediateAsset_SourceDelet
...
+    QCoreApplication::processEvents();
+
+    // There is one active file, because it has been modified.
+    m_assetProcessorManager->CheckFilesToExamine(0);
+    m_assetProcessorManager->CheckActiveFiles(1);
+    m_assetProcessorManager->CheckJobEntries(0);
+
+    // Assess the modified file
+    m_assetProcessorManager->AssessModifiedFile(m_testFilePath.c_str());
+    QCoreApplication::processEvents();
```
amzn-mike, line 810: How come this is calling AssessModifiedFile twice.  Also the rest these are using invokeMethod to call AssessModifiedFile

Code/Tools/AssetProcessor/native/tests/assetmanager/AssetProcessorManagerTest.cpp
```diff
@@ -706,6 +707,149 @@ TEST_F(AssetProcessorIntermediateAssetTests, NestedIntermediateAsset_SourceDelet
...
+    // Process the file, which triggers CheckMissingJobs to be called, which actually deletes the no longer emitted file.
+    // This doesn't call ProcessSingleStep because the files to examine and active files won't match what ProcessSingleStep expects.
+
+    // m_jobDetailsList lets this test verify the job ran that was expected to run.
+    m_jobDetailsList.clear();
+    m_fileCompiled = false;
+    m_fileFailed = false;
+    QCoreApplication::processEvents();
+    m_assetProcessorManager->ProcessFilesToExamineQueue();
+    QCoreApplication::processEvents(); // execute ProcessFilesToExamineQueue
```
amzn-mike, line 826: This is also weird, usually ProcessFilesToExamineQueue isn't call directly, it would be called eventually from the processEvents calls.

Code/Tools/AssetProcessor/native/tests/assetmanager/AssetProcessorManagerTest.cpp
```diff
@@ -706,6 +707,149 @@ TEST_F(AssetProcessorIntermediateAssetTests, NestedIntermediateAsset_SourceDelet
...
+    m_assetProcessorManager->CheckFilesToExamine(0);
+    m_assetProcessorManager->CheckActiveFiles(0);
+    m_assetProcessorManager->CheckJobEntries(0);
+
+    // Nothing should have failed to process.
+    EXPECT_FALSE(m_fileFailed);
+
+    // The intermediate asset should be deleted and gone, because CheckMissingJobs removed it for no longer being a product
+    // of any source asset.
+    auto expectedIntermediatePath = GetIntermediateAssetsDir() / AZStd::string("test.stage2");
```
amzn-mike, line 847: Nit: I think there's a MakePath function to do this

Code/Tools/AssetProcessor/native/tests/assetmanager/AssetProcessorManagerTest.cpp
```diff
@@ -706,6 +707,149 @@ TEST_F(AssetProcessorIntermediateAssetTests, NestedIntermediateAsset_SourceDelet
...
+    QFile sourceAsset(expectedIntermediatePath.c_str());
+    EXPECT_FALSE(sourceAsset.exists());
```
amzn-mike, line 849: Nit: Usually prefer to use AZ::IO::FileIOBase now since it can be mocked (there is a drop-in mock class for this now) to allow us to make these tests entirely virtual so they don't write to disk.

amzn-mike (approved):

Code/Tools/AssetProcessor/native/tests/assetmanager/AssetProcessorManagerTest.cpp
AMZN-stankowi: Switched to using it.

Code/Tools/AssetProcessor/native/tests/assetmanager/AssetProcessorManagerTest.cpp
AMZN-stankowi: Switched to FileIOBase, and added an extra check that the intermediate asset was created earlier.

Code/Tools/AssetProcessor/native/tests/assetmanager/AssetProcessorManagerTest.cpp
AMZN-stankowi: Good point, I've cleaned up the asset processing logic.

AMZN-alexpete (approved):
Nice!

PR #14683, opened 2023-02-20
Polygon light bounds fix

invertednormal:
## What does this PR do?

This PR fixes a few small issues with the polygon light.
- Depending on the order of initialization, it was possible for AZ::Hemisphere to complain about a non-normalized direction. This happened when the polygon tried updating its bounds before having any points.
- The polygon light would still create a "valid" default initialized sphere when it was invalid
- When the polygon's points are removed and lowered to below 3, it wouldn't update to not render, instead it would retain a brighter version of the last valid point configuration.

## How was this PR tested?

Tested the Polygon light in various configurations.

### DIFF
--- a/Gems/Atom/Feature/Common/Code/Source/CoreLights/PolygonLightFeatureProcessor.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/CoreLights/PolygonLightFeatureProcessor.cpp
@@ -227,12 +227,12 @@ namespace AZ::Render
     void PolygonLightFeatureProcessor::SetPolygonPoints(LightHandle handle, const Vector3* vertices, const uint32_t vertexCount, const Vector3& direction)
     {
         AZ_Warning("PolygonLightFeatureProcessor", vertexCount <= MaxPolygonPoints, "Too many polygon points on polygon light. Only using the first %lu vertices.", MaxPolygonPoints);
-        AZ_Warning("PolygonLightFeatureProcessor", vertexCount >= 2, "Polygon light must have at least three points - ignoring points.");
+        AZ_Warning("PolygonLightFeatureProcessor", vertexCount > 2, "Polygon light must have at least three points - ignoring points.");
 
         PolygonLightData& data = m_lightData.GetData<0>(handle.GetIndex());
         if (vertexCount < 3)
         {
-            data.SetEndIndex(data.GetStartIndex());
+            data.SetEndIndex(data.GetStartIndex()); // Set the number of points to 0.
         }
         else
         {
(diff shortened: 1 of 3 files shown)


Gems/Atom/Feature/Common/Code/Source/CoreLights/PolygonLightFeatureProcessor.cpp
```diff
@@ -228,26 +228,27 @@ namespace AZ::Render
...
-            pointArray.at(i).x = vertices[i].GetX();
-            pointArray.at(i).y = vertices[i].GetY();
-            pointArray.at(i).z = vertices[i].GetZ();
+            PolygonPoints& pointArray = m_lightData.GetData<1>(handle.GetIndex());
+            uint32_t clippedCount = AZ::GetMin<uint32_t>(vertexCount, MaxPolygonPoints);
+            for (uint32_t i = 0; i < clippedCount; ++i)
+            {
+                pointArray.at(i).x = vertices[i].GetX();
+                pointArray.at(i).y = vertices[i].GetY();
+                pointArray.at(i).z = vertices[i].GetZ();
```
rgba16f, line 245: Is there a way to avoid calling `at(i)` 3 times? Each one tests `i` to see if it's within the pointArray.

rgba16f (approved):

Gems/Atom/Feature/Common/Code/Source/CoreLights/PolygonLightFeatureProcessor.cpp
invertednormal: When optimized the compiler doesn't do those checks because it has enough information to know that clippedCount can't go out of bounds. I can add an extra line that creates a ref to pointArray.a(i) though.

invertednormal: But I can eliminate some code with StoreToFloat4 😄

rgba16f (approved):

Gems/Atom/Feature/Common/Code/Source/CoreLights/PolygonLightFeatureProcessor.cpp
```diff
@@ -228,26 +228,26 @@ namespace AZ::Render
     {
         AZ_Warning("PolygonLightFeatureProcessor", vertexCount <= MaxPolygonPoints, "Too many polygon points on polygon light. Only using the first %lu vertices.", MaxPolygonPoints);
         AZ_Warning("PolygonLightFeatureProcessor", vertexCount >= 2, "Polygon light must have at least three points - ignoring points.");
```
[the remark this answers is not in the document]
invertednormal: Condition should be >, will fix

VickyAtAZ (approved):

PR #14685, opened 2023-02-20
Adds "EditorAndServerLauncher" as a target.

mbalfour-amzn:
## What does this PR do?

This is a small Quality of Life improvement to make it easier to build both the Editor and the Server when working in the Editor. 

When using the Editor with a multiplayer project, it will automatically launch the ServerLauncher for you in game mode. However, building the Editor doesn't automatically build the ServerLauncher, so it's very easy for the two to get out of sync.

This change adds an "EditorAndServerLauncher" target into the project next to the launchers that will simply build both the Editor _and_ the ServerLauncher for that project, and it will run the Editor when the project is run. This makes it much easier to avoid the mistake of forgetting to build the ServerLauncher.

## How was this PR tested?

Built the MultiplayerSample project and verified that the new target worked as expected.

image

### DIFF
--- a/Code/LauncherUnified/launcher_generator.cmake
+++ b/Code/LauncherUnified/launcher_generator.cmake
@@ -201,6 +201,12 @@ foreach(project_name project_path IN ZIP_LISTS O3DE_PROJECTS_NAME LY_PROJECTS)
 
         # Associate the Servers Gem Variant with each projects ServerLauncher
         ly_set_gem_variant_to_load(TARGETS ${project_name}.ServerLauncher VARIANTS Servers)
+
+        # If the Editor is getting built, it should rebuild the ServerLauncher as well. The Editor's game mode will attempt
+        # to launch it, and things can break if the two are out of sync.
+        if(PAL_TRAIT_BUILD_HOST_TOOLS)
+            ly_add_dependencies(Editor ${project_name}.ServerLauncher)
+        endif()
     endif()
 
     ################################################################################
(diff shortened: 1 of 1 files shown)


AMZN-stankowi (approved):

lawsonamzn (approved):

Code/LauncherUnified/launcher_generator.cmake
```diff
@@ -257,6 +257,29 @@ foreach(project_name project_path IN ZIP_LISTS O3DE_PROJECTS_NAME LY_PROJECTS)
...
+
+    ################################################################################
+    # EditorWithServer
+    ################################################################################
+    if(PAL_TRAIT_BUILD_HOST_TOOLS)
+        if(PAL_TRAIT_BUILD_SERVER_SUPPORTED)
+            add_custom_target(${project_name}.EditorWithServer
+            )
+            ly_add_dependencies(${project_name}.EditorWithServer Editor)
+            ly_add_dependencies(${project_name}.EditorWithServer ${project_name}.ServerLauncher)
```
lumberyard-employee-dm, line 269: Would you want the custom target to build the assets (TARGET `${project_name}.Assets`) as well or is that not needed?

lumberyard-employee-dm (approved):

Code/LauncherUnified/launcher_generator.cmake
mbalfour-amzn: I don't think that's needed, since we're still running the Editor, which runs the AP, which builds the assets.

mbalfour-amzn:
Per some on-the-side feedback, I renamed "EditorWithServer" to "EditorAndServerLauncher" for improved clarity. I'm also looking into whether or not we can build it from Periodic builds to make sure it's safe.

Some other things that have been discussed but are _not_ being done here:
* The ServerLauncher target _could_ automatically add itself as an Editor dependency so that the Editor will always build it. The downside to that approach is that if multiple projects are active, the Editor will build _all_ the ServerLauncher targets on code changes, which seems excessive when trying to iterate on a single Editor/server combination.
* We _could_ change the Editor and EditorAndServerLauncher targets so that the Editor target passes in `-editorsv_clientserver=true` when running and EditorAndServerLauncher passes in `editorsv_clientserver=false`. This would make it so that the Editor runs a built-in server and doesn't launch ServerLauncher by default, and the EditorAndServerLauncher target would launch ServerLauncher by default. The concern with this approach is that it adds some invisible configuration switching that might confuse users.

Code/LauncherUnified/launcher_generator.cmake
```diff
@@ -257,6 +257,29 @@ foreach(project_name project_path IN ZIP_LISTS O3DE_PROJECTS_NAME LY_PROJECTS)
...
         ly_set_gem_variant_to_load(TARGETS ${project_name}.UnifiedLauncher VARIANTS Unified)
     endif()
+
+    ################################################################################
+    # EditorWithServer
+    ################################################################################
+    if(PAL_TRAIT_BUILD_HOST_TOOLS)
+        if(PAL_TRAIT_BUILD_SERVER_SUPPORTED)
+            add_custom_target(${project_name}.EditorWithServer
+            )
```
AMZN-alexpete, line 267:
```suggestion
            add_custom_target(${project_name}.EditorWithServer)
```

AMZN-alexpete (approved):

Code/LauncherUnified/launcher_generator.cmake
lumberyard-employee-dm: Ok, I was wondering if there was a need to streamline the AssetProcessing, for the first time launch, but it looks like their isn't a need.

mbalfour-amzn:
I'm solving this a different way, so I'll open up a new PR to avoid any confusion.

PR #14686, opened 2023-02-20
Allow editable breadcrumbs to use and display full file paths

### LINKED ISSUE #14431: Asset Browser: Windows Explorer URLs/paths do not work in the Asset Browser
**Describe the bug**
Windows Explorer URLs/paths do not work in the Asset Browser. Per the feature request outlined in https://github.com/o3de/o3de/issues/12854, users should be able to copy/paste an Explorer URL into the AB to jump directly to a specific directory.

**Steps to reproduce**
```
```
2. Open the Asset Browser.
3. In Windows Explorer, copy/paste the path to AutomatedTesting/Assets, (e.g. D:\GitHub\o3de\AutomatedTesting\Assets)
4. Paste the copied path into the Asset Browser URL/breadcrumbs bar.

**Expected behavior**
The pasted directory is jumped to in the AB.

**Actual behavior**
The pasted directory is cleared out and the previously opened directory replaces the pasted content.

**Found in Branch**
development

1fdc3d4c1c2bbc38b729733c1df53bb8e15cb231

**Additional context**
1. This also occurs in reverse. Copying from the Asset Browser URL bar doesn't append letter drives or anything to the path, so copying out something like AutomatedTesting/Assets/Prefabs won't be able to pull anything up in Windows Explorer.

dtamkin1:
## What does this PR do?

- Allows users to paste full file paths from file explorer and have the asset browser navigate to the correct directory
- Display the full file path when editing the URL, allowing users to copy the full file path
- Fixes #14431.

Editor_GZSoptphT7

## How was this PR tested?

Locally

### DIFF
--- a/Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
+++ b/Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
@@ -278,6 +278,7 @@ AzAssetBrowserWindow::AzAssetBrowserWindow(QWidget* parent)
         {
             // No need to select the folder ourselves, callback from Breadcrumbs will take care of that
             m_ui->m_pathBreadCrumbs->pushPath(QString::fromUtf8(folderEntry->GetVisiblePath().c_str()));
+            m_ui->m_pathBreadCrumbs->setFullPath(QString::fromUtf8(folderEntry->GetFullPath().c_str()));
         }
     });
 
@@ -624,15 +625,18 @@ void AzAssetBrowserWindow::UpdateBreadcrumbs(const AzToolsFramework::AssetBrowse
     using namespace AzToolsFramework::AssetBrowser;
 
     QString entryPath;
+    QString fullPath;
     if (selectedEntry)
     {
         const AssetBrowserEntry* folderEntry = Utils::FolderForEntry(selectedEntry);
         if (folderEntry)
         {
             entryPath = QString::fromUtf8(folderEntry->GetVisiblePath().c_str());
+            fullPath = QString::fromUtf8(folderEntry->GetFullPath().c_str());
         }
     }
     m_ui->m_pathBreadCrumbs->pushPath(entryPath);
+    m_ui->m_pathBreadCrumbs->setFullPath(fullPath);
 }
 
 void AzAssetBrowserWindow::SetTwoColumnMode(QWidget* viewToShow)
--- a/Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
+++ b/Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
@@ -638,8 +638,8 @@ void AzAssetBrowserWindow::UpdateBreadcrumbs(const AzToolsFramework::AssetBrowse
         const AssetBrowserEntry* folderEntry = Utils::FolderForEntry(selectedEntry);
         if (folderEntry)
         {
-            entryPath = FromStdString(folderEntry->GetVisiblePath().c_str());
-            fullPath = FromStdString(folderEntry->GetFullPath().c_str());
+            entryPath = FromStdString(folderEntry->GetVisiblePath());
+            fullPath = FromStdString(folderEntry->GetFullPath());
         }
     }
     m_ui->m_pathBreadCrumbs->pushFullPath(fullPath, entryPath);
--- a/Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
+++ b/Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
@@ -45,6 +45,14 @@ static constexpr int s_narrowModeThreshold = 700;
 
 using namespace AzToolsFramework::AssetBrowser;
 
+namespace
+{
+    inline QString FromStdString(AZStd::string_view string)
+    {
+        return QString::fromUtf8(string.data(), static_cast<int>(string.size()));
+    }
+}
+
 namespace AzToolsFramework
 {
     namespace AssetBrowser
@@ -277,8 +285,7 @@ AzAssetBrowserWindow::AzAssetBrowserWindow(QWidget* parent)
         if (folderEntry)
         {
             // No need to select the folder ourselves, callback from Breadcrumbs will take care of that
-            m_ui->m_pathBreadCrumbs->pushPath(QString::fromUtf8(folderEntry->GetVisiblePath().c_str()));
-            m_ui->m_pathBreadCrumbs->setFullPath(QString::fromUtf8(folderEntry->GetFullPath().c_str()));
+            m_ui->m_pathBreadCrumbs->pushFullPath(FromStdString(folderEntry->GetFullPath()), FromStdString(folderEntry->GetVisiblePath()));
         }
     });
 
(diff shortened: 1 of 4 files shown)


Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
```diff
@@ -278,6 +278,7 @@ AzAssetBrowserWindow::AzAssetBrowserWindow(QWidget* parent)
         {
             // No need to select the folder ourselves, callback from Breadcrumbs will take care of that
             m_ui->m_pathBreadCrumbs->pushPath(QString::fromUtf8(folderEntry->GetVisiblePath().c_str()));
+            m_ui->m_pathBreadCrumbs->setFullPath(QString::fromUtf8(folderEntry->GetFullPath().c_str()));
```
alexmontAmazon, line 281: do we need both of these calls? Is it possible to either make it so that pushPath knows how to build the fullPath and sets it itself, or maybe have setFullPath know to push the end string? I'm just suspicious that everywhere that pushPath is called, setFullPath is called too, so it seems like these might be able to be combined.

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Views/AssetBrowserTreeView.cpp
```diff
@@ -352,9 +352,11 @@ namespace AzToolsFramework
         const AssetBrowserEntry* AssetBrowserTreeView::GetEntryByPath(QStringView path)
         {
             QModelIndex current;
-            for (auto token : AZStd::ranges::split_view(path, '/'))
+            bool currentlyFindingRows = false;
+            AZ::IO::Path p = path.toUtf8().constData();
```
alexmontAmazon, line 356: I think single letter variables names are against our coding standard. Could you change this to something like "pathString"?

alexmontAmazon (approved):
a couple of small suggested changes. Otherwise looks good :)

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Views/AssetBrowserTreeView.cpp
```diff
@@ -352,9 +352,11 @@ namespace AzToolsFramework
...
+            for (auto it = p.begin(); it != p.end(); ++it)
             {
-                QStringView pathPart{ token.begin(), token.end() };
+                QString pathPart = QString::fromUtf8((*it).c_str());
                 const int rows = model()->rowCount(current);
                 bool rowFound = false;
                 for (int row=0; row < rows; ++row)
```
burelc-amzn, line 362: If I'm reading this code correctly, the objective here is to find an item in the tree view's model based on the path parts in `path`, while skipping the leading parts of `path` that are not in the model. I think that using `QAbstractItemModel::match()` would be clearer than the loop made here. (I couldn't select the whole function, but I'd replace the whole body of the function with this version). This version also doesn't need the `currentlyFindingRows` bool, as that same information is encoded in the test `current.isValid()`.

Separately, this change results in two additional copies of the input `path`. The old version is using `split_view` along with `QStringView` to avoid any copies. With this version, the use of `split_view` is maintained, but each path part is copied into a `QString`, because the `match()` method works on a `QVariant` value, and `QVariant` doesn't work with `QStringView`.

```suggestion
            QModelIndex current;
            for (const auto token : AZStd::ranges::split_view(path, '/'))
            {
                QString pathPart(token.begin(), AZStd::distance(token.begin(), token.end()));

                QModelIndexList next = model()->match(
                    /*start =*/model()->index(0, 0, current),
                    /*role =*/Qt::DisplayRole,
                    /*value =*/pathPart,
                    /*hits =*/1,
                    /*flags =*/Qt::MatchExactly
                );
                if (next.size() == 1)
                {
                    current = next[0];
                }
                else if (current.isValid())
                {
                    return nullptr;
                }
            }
            return GetEntryFromIndex<AssetBrowserEntry>(current);
```

Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
dtamkin1: Done! I've combined the methods, so pushFullPath calls pushPath and sets itself.

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Views/AssetBrowserTreeView.cpp
dtamkin1: Yup will do, thanks!

cgalvan (approved):
LGTM! Agree w/ other Chris' feedback as well.

Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
```diff
@@ -624,15 +632,17 @@ void AzAssetBrowserWindow::UpdateBreadcrumbs(const AzToolsFramework::AssetBrowse
...
     QString entryPath;
+    QString fullPath;
     if (selectedEntry)
     {
         const AssetBrowserEntry* folderEntry = Utils::FolderForEntry(selectedEntry);
         if (folderEntry)
         {
-            entryPath = QString::fromUtf8(folderEntry->GetVisiblePath().c_str());
+            entryPath = FromStdString(folderEntry->GetVisiblePath().c_str());
+            fullPath = FromStdString(folderEntry->GetFullPath().c_str());
```
burelc-amzn, line 642: Don't pass the result of `.c_str()` to a `string_view` parameter, it degrades the type down to a `const char*` and loses the length information.
```suggestion
            entryPath = FromStdString(folderEntry->GetVisiblePath());
            fullPath = FromStdString(folderEntry->GetFullPath());
```

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Views/AssetBrowserTreeView.cpp
```diff
@@ -352,30 +352,21 @@ namespace AzToolsFramework
         const AssetBrowserEntry* AssetBrowserTreeView::GetEntryByPath(QStringView path)
         {
             QModelIndex current;
-            for (auto token : AZStd::ranges::split_view(path, '/'))
+            const QByteArray byteArray = path.toUtf8();
+            const AZ::IO::Path azpath{ AZStd::string_view{ byteArray.constData(), static_cast<size_t>(byteArray.size()) } };
```
burelc-amzn, line 356: You can avoid the internal copy of the utf8 converted data that's present in the `QByteArray` by using `AZ::IO::PathView` here
```suggestion
            const AZ::IO::PathView azpath{ AZStd::string_view{ byteArray.constData(), static_cast<size_t>(byteArray.size()) } };
```

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Views/AssetBrowserTreeView.cpp
```diff
@@ -352,30 +352,21 @@ namespace AzToolsFramework
...
+            for (const auto& pathPart : azpath)
             {
-                QStringView pathPart{ token.begin(), token.end() };
-                const int rows = model()->rowCount(current);
-                bool rowFound = false;
-                for (int row=0; row < rows; ++row)
+                QModelIndexList next = model()->match(
```
burelc-amzn, line 359: Nit: this can be `const`

dtamkin1: Thanks, fixed!

Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
dtamkin1: Fixed!

PR #14699, opened 2023-02-21
Project exporter cli external script prototype

tkothadev:
this is just a prototype demonstrating the possibilities of project export via loading an external python file.

I've included the export script that resides in the Newspaper Delivery Game Project folder here for reference.

I'm able to import the O3DE CLI (and the packaging script can even pre-inject variable names for the export script to use).

I made comments about pros and cons of this approach, but that is still currently being evaluated, so that list is not final.

### DIFF
--- a/scripts/o3de.py
+++ b/scripts/o3de.py
@@ -35,7 +35,7 @@ def add_args(parser: argparse.ArgumentParser) -> None:
     sys.path.insert(0, str(o3de_package_dir))
     from o3de import engine_properties, engine_template, gem_properties, \
         global_project, register, print_registration, get_registration, \
-        enable_gem, disable_gem, project_properties, sha256, download
+        enable_gem, disable_gem, project_properties, sha256, download, packaging
     # Remove the temporarily added path
     sys.path = sys.path[1:]
 
@@ -75,6 +75,9 @@ def add_args(parser: argparse.ArgumentParser) -> None:
     # download
     download.add_args(subparsers)
 
+    #packaging
+    packaging.add_args(subparsers)
+
 
 if __name__ == "__main__":
     # parse the command line args
@@ -85,6 +88,7 @@ if __name__ == "__main__":
 
     # parse args
     the_args = the_parser.parse_args()
+
 
     # if empty print help
     if len(sys.argv) == 1:
--- a/scripts/o3de/example_project_export_scripts/newspaper_delivery_project_export.py
+++ b/scripts/o3de/example_project_export_scripts/newspaper_delivery_project_export.py
@@ -1,0 +1,44 @@
+import sys
+# import o3de.packaging as pkg
+from o3de.packaging import *
+import o3de.manifest as man
+import o3de.project_manager_interface as pm
+
+#from other_local_script import *  #our script can import from both o3de CLI and from the local folder it resides in
+
+#this is a simple linear script
+
+#pros for basic python scripting approach
+# very simple to setup
+# data oriented (one workflow per python script)
+# if API is minimal, it is readily accessible, and function arguments can be provided by user
+# both the script's local directory and O3DE CLI are accessible for import logic
+
+#drawbacks
+# requires user to code in python (we have customers that would rather not)
+# it can be daunting as a blank slate. coding patterns are not standardized/enforceable, and is up to user to code out correctly
+# can't do pre-validation of workflow
+# lack of step annotation
+# because the entire workflow is python code, it is non-trivial to parse (and with custom team-code, nigh impossible), and we won't be able to enforce data patterns
+    # from CLI perspective this is not a big deal, but for ProjectManager UI, this can impede our ability to provide meaningful UX feedback or enable automated auditing
+# too much more than what is actually needed for project export use case?
+
+
+package_hi()
+print("Now running the packaging code")
+
+
+#apparently our module script can be pre-injected with known variable values!!!
+#this comes from o3de/packaging.py
+print(o3de_project_path)
+
+
+
+package_windows_standalone_monolithic(project_path= o3de_project_path,
+                                          zipped = False)
+
+
+
+#before doing full compare contrast, get the json version of this working
+
+
--- a/scripts/o3de/o3de/packaging.py
+++ b/scripts/o3de/o3de/packaging.py
@@ -1,0 +1,298 @@
+#
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+#
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+#
+import os
+import platform
+import argparse
+import json
+import logging
+import pathlib
+import sys
+import subprocess
+import shutil
+import pathlib
+
+from o3de import manifest, validation, utils, repo
+
+logger = logging.getLogger('o3de.print_registration')
+logging.basicConfig(format=utils.LOG_FORMAT)
+
+SUCCESS = 0
+
+#goal: simple CLI to package and build/automate what we need. This is proven with the prototype, so we're good to start RFC.
+# we will need sufficient time to flesh out the RFC, then the engineering plan. Otherwise it fees.
+# thankfully, I'm not blocking anyone since this is an automation task
+# can I ship this before RFC is accepted (feature flags? or separate branch) talk with Joe if this is acceptable for smallest MVP
+# all this being said, it is better that I do not overengineer at  this point. Better to write the RFC/design, get feedback, and adjust accordingly. Make sure to supply as many relevant considerations as possible
+
+#design review - ask joe about this, or if he still wants me to write RFC first. (do the minimum important things. Let the design sheperd guide you)
+
+#review automated test code for running executables
+#issue with timeouts: this can be extremely difficult
+#if someone clones a mature project...
+# SPIKE: figure out good system for timeouts. It would suck if it stalls out, but it also sucks if it prematurely times out
+# how do we show progress?
+# how do we handle logging?
+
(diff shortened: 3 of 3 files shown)


scripts/o3de/example_project_export_scripts/newspaper_delivery_project_export.py
```diff
@@ -0,0 +1,44 @@
...
+# both the script's local directory and O3DE CLI are accessible for import logic
+
+#drawbacks
+# requires user to code in python (we have customers that would rather not)
+# it can be daunting as a blank slate. coding patterns are not standardized/enforceable, and is up to user to code out correctly
+# can't do pre-validation of workflow
+# lack of step annotation
+# because the entire workflow is python code, it is non-trivial to parse (and with custom team-code, nigh impossible), and we won't be able to enforce data patterns
+    # from CLI perspective this is not a big deal, but for ProjectManager UI, this can impede our ability to provide meaningful UX feedback or enable automated auditing
+# too much more than what is actually needed for project export use case?
```
tkothadev, line 24: The last one is most likely wrong, b/c we could have all kinds of use cases. Python would give us flexibility to handle what is needed. It's encumbant on script writer to not go too far.

scripts/o3de/o3de/packaging.py
```diff
@@ -0,0 +1,318 @@
...
+    module_name = os.path.basename(external_script_path)
+
+    #works for python 3.5+
+    sys.path.insert(0, os.path.split(pathlib.Path(external_script_path))[0])
+    sys.path.insert(0, os.getcwd())
+    spec = importlib.util.spec_from_file_location(module_name, external_script_path)
+    foo = importlib.util.module_from_spec(spec)
+    sys.modules[module_name] = foo
+    foo.o3de_project_path =  args.project_path #customer facing variable name for accessible scripting
+    spec.loader.exec_module(foo)
```
tkothadev, line 240: this segment basically demonstrates how simple our export project system can be. We can completely offload the responsibility of exporting out to an external script. From there it's up to teams working on individual use cases to make sure their export scripts are correct, and we can just focus on providing base API to make that job easier.

scripts/o3de/o3de/packaging.py
```diff
@@ -0,0 +1,318 @@
...
+    parser.add_argument('-sf', '--script-file', type=pathlib.Path, required=True, help='external script file to run')
+    # parser.add_argument('-z', '--zipped', action='store_true',
+    #                     default=False,
+    #                     help='takes the packaged build and zips it')
+    parser.add_argument('-pp', '--project-path', type=pathlib.Path, required=False,
+                        help='Project to package')
+    # parser.add_argument('-ep', '--engine-path', type=pathlib.Path, required=False,
+    #                     help='Engine for building tooling')
+    parser.set_defaults(func=_run_packaging)
+
```
tkothadev, line 281: in this prototype, we only take 2 arguments: project path, and script file path. These are separate because we have no guarantees as to where the user put the script file.

scripts/o3de/o3de/packaging.py
```diff
@@ -0,0 +1,318 @@
...
+# step 8: putting everything into container for distribution (zip file, apk file, ipa file)
+#enable code re-use. This could also inform _how_ we setup our keys
+
+# we can rely on making this code file expandable as a fallback
+# we should provide legacy warnings as code updates progress
+
+# how do we handle with engine customer/contributor
+# RFC will help inform my logistical questions on what needs to be split and when
+# test driven development would help
+def package_windows_standalone_monolithic(project_path: str,
```
AMZN-stankowi, line 103: this should be broken into smaller functions.

scripts/o3de.py
```diff
@@ -35,7 +35,7 @@ def add_args(parser: argparse.ArgumentParser) -> None:
     sys.path.insert(0, str(o3de_package_dir))
     from o3de import engine_properties, engine_template, gem_properties, \
         global_project, register, print_registration, get_registration, \
-        enable_gem, disable_gem, project_properties, sha256, download
+        enable_gem, disable_gem, project_properties, sha256, download, packaging
```
AMZN-stankowi, line 38: we should use the term project export here instead of packaging, because packaging means different things to different people

scripts/o3de.py
```diff
@@ -35,7 +35,7 @@ def add_args(parser: argparse.ArgumentParser) -> None:
     sys.path.insert(0, str(o3de_package_dir))
```
AMZN-stankowi, line 35: with the python based approach, what is the benefit to using o3de.py as the entry point to calling this? Why not just skip invoking o3de.py and call what that calls, directly?

scripts/o3de/example_project_export_scripts/newspaper_delivery_project_export.py
```diff
@@ -0,0 +1,44 @@
...
+print("Now running the packaging code")
+
+
+#apparently our module script can be pre-injected with known variable values!!!
+#this comes from o3de/packaging.py
+print(o3de_project_path)
+
+
+
+package_windows_standalone_monolithic(project_path= o3de_project_path,
```
AMZN-stankowi, line 37: another reason to break package_windows_standalone_monolithic into smaller functions is flexibility. There's no way for this project to alter the order or how the steps are run, or interject a new step between other steps.

scripts/o3de/example_project_export_scripts/newspaper_delivery_project_export.py
```diff
@@ -0,0 +1,44 @@
+import sys
```
AMZN-stankowi, line 1: This file, newspaper_delivery_project_export.py, should be in the newspaper delivery project folder and not in the global engine folder

scripts/o3de/o3de/packaging.py
```diff
@@ -0,0 +1,318 @@
+#
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+#
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+#
+import os
+import platform
+import argparse
```
AMZN-stankowi, line 10: I suggest putting your imports in alphabetical order, it's easier to manage them and avoid duplicates that way

scripts/o3de/o3de/packaging.py
```diff
@@ -0,0 +1,318 @@
...
+
+#review automated test code for running executables
+#issue with timeouts: this can be extremely difficult
+#if someone clones a mature project...
+# SPIKE: figure out good system for timeouts. It would suck if it stalls out, but it also sucks if it prematurely times out
+# how do we show progress?
+# how do we handle logging?
+
+#add parameters for if we want to expose logging or not? otherwise render progress?
+def process_command(cmd_args, project_path):
```
AMZN-stankowi, line 42: there are a lot of existing wrappers for running commands on the command line in the lytesttools folders. We should re-use an existing one if possible. I think at least one wraps subprocess.popen.

scripts/o3de.py
AMZN-stankowi: We had a discussion on this: The grouping kind of matches other uses of o3de.py, it can help with discoverability.

It also helps with parameter management: o3de.py is setup to be aware of things like engine folder. We may be able to automatically populate properties for the project exporting scripts from o3de.py easier than elsewhere.

scripts/o3de/o3de/packaging.py
```diff
@@ -0,0 +1,318 @@
+#
```
AMZN-stankowi, line 1: Feedback from the discussion: Let's make sure to include a task to write scoped automated tests (ideally unit tests) for anything we write here

AMZN-stankowi: https://github.com/o3de/o3de/tree/development/scripts/o3de/tests

scripts/o3de/example_project_export_scripts/newspaper_delivery_project_export.py
AMZN-stankowi: The author is aware of this issue - but it's here to make it easier to review the whole concept, it let this file be in this draft pull request for reference.

AMZN-stankowi: Other feedback: For the deliverable, we can link to examples in the docs or elsewhere

scripts/o3de/example_project_export_scripts/newspaper_delivery_project_export.py
AMZN-stankowi: We discussed this: The author did this for the prototype to get this running. In a future iteration, this will be broken up.

There is even an option to make use of the templating system to be able to quickly get started by creating a new workflow python script from a template file.

Other feedback: It can be useful to have this single, monolithic function as a helper function that calls the discrete, individual step functions. Then you could call that if you want, or look into it and call those steps yourself.

scripts/o3de/o3de/packaging.py
```diff
@@ -0,0 +1,318 @@
...
+
+    packaging_rules = None
+    with open(PACKAGING_RULES_PATH, "r") as prules_file:
+        packaging_rules = json.load(prules_file)
+        print(packaging_rules)
+
+    #note: we need ability to supply a default option!
+    ENGINE_PATH = packaging_rules["engine-path"]
+    THIRD_PARTY_PATH  = packaging_rules["third-party-path"]
+    CMAKE_GENERATOR = packaging_rules["cmake-generator"]
```
AMZN-stankowi, line 121: Feedback from our review: Is there an existing cmake python interface we use elsewhere, that we can re-use here?

AMZN-stankowi: What about https://github.com/o3de/o3de/blob/development/scripts/o3de/o3de/cmake.py ?

PR #14701, opened 2023-02-21
Fix compatibility not considering scan-up engine

AMZN-alexpete:
## What does this PR do?

1. Allows version compatibility check to succeed when no `engine` is specified in `project.json` but the engine would be found using scan-up logic because the project is inside the engine, e.g. AutomatedTesting
1. Fixes incorrect cmake error message when a `project.json` doesn't have an `engine` key.  In this case, the error message should state that the project needs to be registered, not that the engine needs to be registered.

## How was this PR tested?

- configured AutomatedTesting successfully from the AutomatedTesting folder
- created a new project outside an engine folder and verified that the warning message is displayed when the engine is not registered.


amzn-phist (approved):

cmake/o3deConfigVersion.cmake
```diff
@@ -63,6 +63,25 @@ set(PACKAGE_VERSION ${engine_version})
...
+            if(is_same_path)
+                message(VERBOSE "The project folder is a subdirectory of this engine.")
+                set(PACKAGE_VERSION_COMPATIBLE TRUE)
+                return()
+            else()
+                cmake_path(GET candidate_engine_path PARENT_PATH parent_path)
+                cmake_path(COMPARE "${candidate_engine_path}" EQUAL "${parent_path}" reached_root_dir)
+                set(candidate_engine_path ${parent_path})
+            endif()
+        endwhile()
```
lumberyard-employee-dm, line 82: Instead of looping, the [cmake_path(IS_PREFIX](https://cmake.org/cmake/help/latest/command/cmake_path.html?highlight=cmake_path#query) function should be used here.

AMZN-alexpete: nice! great suggestion!

lumberyard-employee-dm: Nevermind, it looks like you are trying to locate the engine, not check if the project is a subdirectory

AMZN-alexpete: I was trying to see if the project is a subdirectory of the engine, IS_PREFIX is much simpler

AMZN-Phil (approved):

lumberyard-employee-dm (approved):

cmake/o3deConfigVersion.cmake
lumberyard-employee-dm: Sounds good.

moraaar:
Thanks for fixing this @AMZN-alexpete!

PR #14703, opened 2023-02-21
Disable Revert Override option on entities added as overrides

raktalen:
## What does this PR do?

Entities added as overrides cannot be reverted as revert is the same as delete. This change Greys out the RevertOverride option on entities added as overrides and adds a tooltip explaining why.

## How was this PR tested?

Tested in editor

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabIntegrationManager.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabIntegrationManager.cpp
@@ -720,6 +720,12 @@ namespace AzToolsFramework
                         m_prefabOverridePublicInterface->AreOverridesPresent(selectedEntity))
                     {
                         QAction* revertAction = menu->addAction(QObject::tr("Revert Overrides"));
+                        revertAction->setToolTip(QObject::tr("Revert all overrides on this entity to whatever is in the prefab file."));
+                        if (m_prefabOverridePublicInterface->GetOverrideType(selectedEntity) == OverrideType::AddEntity)
+                        {
+                            revertAction->setToolTip(QObject::tr("Cannot revert overrides on entities that are added as overrides."));
+                            revertAction->setEnabled(false);
+                        }
                         QObject::connect(
                             revertAction,
                             &QAction::triggered,


AMZN-stankowi (approved):
Can you add an automated test for this? This seems like it would work well with the Python based automation.

forkercat (approved):

raktalen:
> Can you add an automated test for this? This seems like it would work well with the Python based automation.

Sure, I'll add it as a followup CR :)

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabIntegrationManager.cpp
```diff
@@ -720,6 +720,12 @@ namespace AzToolsFramework
                         m_prefabOverridePublicInterface->AreOverridesPresent(selectedEntity))
                     {
                         QAction* revertAction = menu->addAction(QObject::tr("Revert Overrides"));
+                        revertAction->setToolTip(QObject::tr("Revert all overrides on this entity to whatever is in the prefab file."));
+                        if (m_prefabOverridePublicInterface->GetOverrideType(selectedEntity) == OverrideType::AddEntity)
+                        {
+                            revertAction->setToolTip(QObject::tr("Cannot revert overrides on entities that are added as overrides."));
+                            revertAction->setEnabled(false);
```
srikappa-amzn, line 727: I think it's better to not show this at all if this was an entity added as an override. This would be consistent with what we have been doing with other actions that user should not be able to perform. We simply hide them instead of showing them in a disabled state.

raktalen: A user not seeing a revert option on one entity and seeing it on others would rightly be confused as to why it exists on one and not on the other. @monroegm any thoughts ?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabIntegrationManager.cpp
```diff
@@ -720,6 +720,12 @@ namespace AzToolsFramework
                         m_prefabOverridePublicInterface->AreOverridesPresent(selectedEntity))
                     {
                         QAction* revertAction = menu->addAction(QObject::tr("Revert Overrides"));
+                        revertAction->setToolTip(QObject::tr("Revert all overrides on this entity to whatever is in the prefab file."));
```
michabr, line 723: Does it revert to values in the prefab files, or values in the loaded in-memory template (which could be unsaved)?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabIntegrationManager.cpp
```diff
@@ -720,6 +720,12 @@ namespace AzToolsFramework
...
+                        if (m_prefabOverridePublicInterface->GetOverrideType(selectedEntity) == OverrideType::AddEntity)
+                        {
+                            revertAction->setToolTip(QObject::tr("Cannot revert overrides on entities that are added as overrides."));
```
michabr, line 726: Suggestion: indicate that user can delete the entity instead.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabIntegrationManager.cpp
raktalen: That's a good question. From a user perspective, I would imagine that this goes to the last saved state. @forkercat  / @srikappa-amzn do you know what actually happens ?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabIntegrationManager.cpp
raktalen: Sure, I was considering adding (Use delete instead) at the end of the tooltip, but it felt a little long. I'll add it though

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabIntegrationManager.cpp
forkercat: It reverts the values in memory which also supports undo/redo. Yeah I believe the moment we do the reversion the focused prefab becomes dirty and unsaved.

raktalen: So it reverts to the unsaved state that is in memory as opposed to the file on disk ?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabIntegrationManager.cpp
forkercat: I prefer making it disabled rather than hidden. As @raktalen points out, I think the user would be confused bc unlike other workflows there is a "+" icon on the entity. Additionally, it gives us a chance to let users know in tooltip that the entity can be deleted as @michabr mentions.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabIntegrationManager.cpp
raktalen: Is this by design ?

forkercat: I think I might not understand the question correctly.

> Does it revert to values in the prefab files, or values in the loaded in-memory template (which could be unsaved)?

For 'or values in the loaded in-memory template', @michabr do you refer to new and unsaved override edits or template edits?

What revert does is just drop all override patches stored in the current focused prefab on that entity. Values would be reverted to the state what its owning prefab see.

michabr: Yeah, I wanted to see if we need to reword the tooltip to make it more accurate because reverting the overrides does not restore any values from disk.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabIntegrationManager.cpp
lawsonamzn: Is there a way to just make it show up and instead just _be_ a delete?  Least confusing?  Or still bad?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabIntegrationManager.cpp
forkercat: I think we can keep it short as "Revert all overrides on this entity" and include the "Use delete instead" info.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Prefab/PrefabIntegrationManager.cpp
raktalen: The concern I have is that most users are unlikely to understand the fact that the "Revert" will actually just make that entity go away. Asking them to delete instead makes the operation clear

michabr (approved):

PR #14706, opened 2023-02-21
Add basic splash screen rendering to Atom

jiaweig-amzn:
## What does this PR do?

1. Add basic splash screen rendering to Atom.
   - It is achieved by adding a separate pass to the main pipeline. Ideally, we could have the main pipeline holding 2 paths (or more), making the original pipeline into a parent pass, and disable the other path if one is in use.
   - Only a fading effect is added to demonstrate the usage of the shader and the pass.
   - Settings are retrieved from setreg in Atom Feature. This may be moved to somewhere else.
2. Move add pass helper function from terrain gem to utilities.
   - Will do a follow up to update other places that add passes to the main pipeline
3. Cooperation with other systems
   - We need further support for handling cooperation with other systems. Now it sets the tick scale to 0 so that everything else will be paused until splash screen passes. But this is not ideal and may have conflicts when multiple systems are scaling the tick.

## How was this PR tested?

Tested with a default level in game launcher.


Gems/Atom/Feature/Common/Code/Include/Atom/Feature/SplashScreen/SplashScreenSettings.h
```diff
@@ -0,0 +1,30 @@
...
+ *
+ */
+#pragma once
+
+#include <AzCore/Memory/SystemAllocator.h>
+#include <AzCore/Serialization/SerializeContext.h>
+
+namespace AZ::Render
+{
+    class SplashScreenSettings
```
AMZN-Gene, line 15: Add comment for public classes regarding their functionality/intended use

Gems/Atom/Feature/Common/Assets/Shaders/SplashScreen/SplashScreenPass.azsl
```diff
@@ -0,0 +1,67 @@
...
+        AddressU = Clamp;
+        AddressV = Clamp;
+        AddressW = Clamp;
+    };
+
+    Texture2D<float4> m_splashScreenImage;
+
+    struct SplashScreenParams
+    {
+        float m_fadingFactor;
```
AMZN-Gene, line 27: Bonus points for fading :D

Gems/Atom/Feature/Common/Code/Include/Atom/Feature/SplashScreen/SplashScreenSettings.h
```diff
@@ -0,0 +1,30 @@
...
+        AZ_RTTI(AZ::Render::SplashScreenSettings, "{4B441F73-8F72-4DCA-9451-416C40693487}");
+        AZ_CLASS_ALLOCATOR(AZ::Render::SplashScreenSettings, AZ::SystemAllocator, 0);
+
+        SplashScreenSettings() = default;
+        virtual ~SplashScreenSettings() = default;
+
+        static void Reflect(AZ::ReflectContext* context);
+
+        AZStd::string m_imagePath;
+        float m_lastingTime = 10.0f;
```
AMZN-Gene, line 27: Assuming seconds. Does this behave as a minimum time? Some usefully of splash screen is something to look at while the game is loading. In that case, a splash screen may have a minimum time (so it doesn't flash for hal;f a second if loading is super fast), but generally it's useful to fade it out as soon as the loading has finished. For example, in LY1.28 you could set a global splashscreen that would automatically appear during app startup and level loads. Each level could also override the splashscreen using a level.cfg.

Gems/Atom/Feature/Common/Code/Source/SplashScreen/SplashScreenPass.h
```diff
@@ -0,0 +1,69 @@
...
+        AZ_CLASS_ALLOCATOR(AZ::Render::SplashScreenPass, AZ::SystemAllocator, 0);
+        virtual ~SplashScreenPass();
+
+        static AZ::RPI::Ptr<SplashScreenPass> Create(const AZ::RPI::PassDescriptor& descriptor);
+
+        // TickBus overrides...
+        void OnTick(float deltaTime, AZ::ScriptTimePoint time) override;
+
+        // Scope producer functions...
+        void SetupFrameGraphDependencies(AZ::RHI::FrameGraphInterface frameGraph) override;
```
AMZN-Gene, line 36: Comments for the indented use of public methods. Helpfully to call out if game developers or engine developers would need to use this API.

AMZN-Gene (approved):

AMZN-Gene:
Thanks @jiaweig-amzn. Splashscreens have been a long time coming, this is great!

AMZN-alexpete:
@jiaweig-amzn Great new feature! I think this is an important change to recognize by incrementing the minor version of the following affected gems from `0.0.0` to `0.1.0`
```
Gems/Atom/gem.json
Gems/Atom/Bootstrap/gem.json
Gems/Atom/Feature/Common/gem.json
```

Gems/Atom/Feature/Common/Code/Source/SplashScreen/SplashScreenPass.h
```diff
@@ -0,0 +1,69 @@
...
+
+        bool m_beginTimer = false;  // Flag to begin timing.
+        float m_screenTime = 10.0f; // How long the slash screen lasts. Will be initialized by setreg.
+        float m_lastRealTimeStamp;  // Time stamp in seconds, used to calculate unscaled delta time.
+
+        // Data struct passed to the shader.
+        struct SplashScreenParams
+        {
+            float m_fadingFactor;
+            uint32_t m_flags;
```
AMZN-alexpete, line 56: these flags appear to be unused
```suggestion
```

Gems/Atom/Feature/Common/Assets/Shaders/SplashScreen/SplashScreenPass.azsl
```diff
@@ -0,0 +1,67 @@
...
+        uint m_flags;
```
AMZN-alexpete, line 28:
```suggestion
```
These flags appear to be unused

Gems/Atom/Feature/Common/Code/Include/Atom/Feature/SplashScreen/SplashScreenSettings.h
AMZN-alexpete: +1 
```suggestion
        float m_secondsTillFadeOut= 10.0f;
```

Gems/Atom/Feature/Common/Code/Source/SplashScreen/SplashScreenFeatureProcessor.cpp
```diff
@@ -0,0 +1,47 @@
...
+        {
+            return;
+        }
+
+        bool enable = false;
+
+        auto settingsRegistry = AZ::SettingsRegistry::Get();
+        static const AZStd::string setregPath = "/O3DE/Atom/Feature/SplashScreenEnable";
+
+        if (settingsRegistry && settingsRegistry->GetObject(&enable, azrtti_typeid(enable), setregPath.c_str()))
```
AMZN-alexpete, line 39:
```suggestion
        static constexpr char* setregPath = "/O3DE/Atom/Feature/SplashScreenEnable";

        if (settingsRegistry && settingsRegistry->GetObject(&enable, azrtti_typeid(enable), setregPath))
```

Gems/Atom/Feature/Common/Code/Source/SplashScreen/SplashScreenFeatureProcessor.cpp
```diff
@@ -0,0 +1,47 @@
...
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#include <SplashScreen/SplashScreenFeatureProcessor.h>
+#include <Atom/RPI.Public/RPIUtils.h>
+#include <AzCore/Settings/SettingsRegistryImpl.h>
+
+namespace AZ::Render
+{
```
AMZN-alexpete, line 14:
```suggestion
{
    namespace Internal {
        static constexpr AZStd::string_view SetregSplashScreenEnableKey{ "/O3DE/Atom/Feature/SplashScreenEnable" };
    }
```

Gems/Atom/Feature/Common/Code/Source/SplashScreen/SplashScreenFeatureProcessor.cpp
```diff
@@ -0,0 +1,47 @@
...
+        AZ::ComponentApplicationBus::Broadcast(&AZ::ComponentApplicationBus::Events::QueryApplicationType, appType);
+        if (!appType.IsValid() || appType.IsEditor())
+        {
+            return;
+        }
+
+        bool enable = false;
+
+        auto settingsRegistry = AZ::SettingsRegistry::Get();
+        static const AZStd::string setregPath = "/O3DE/Atom/Feature/SplashScreenEnable";
```
AMZN-alexpete, line 37: I'm tempted to recommend you put this in `SplashScreenFeatureProcessor.h` like
```
inline constexpr const char* SplashScreenEnableKey= "/O3DE/Atom/Feature/SplashScreenEnable";
```
but it is only used in 2 places so...

Gems/Atom/Feature/Common/Registry/splash_screen.setreg
```diff
@@ -0,0 +1,16 @@
+{
+    "O3DE": {
```
AMZN-alexpete, line 2: I think we should consider adding this `.setreg` in the project templates e.g. https://github.com/o3de/o3de/tree/development/Templates/DefaultProject/Template/Registry

Gems/Atom/Feature/Common/Code/Source/SplashScreen/SplashScreenPass.cpp
```diff
@@ -0,0 +1,147 @@
...
+        m_splashScreenImage = AZ::RPI::LoadStreamingTexture(m_settings.m_imagePath);
+        m_screenTime = m_settings.m_lastingTime;
+
+        if (!m_splashScreenImage)
+        {
+            return;
+        }
+
+        m_splashScreenImageIndex.Reset();
+        m_splashScreenImageIndex.Reset();
```
AMZN-alexpete, line 60: why reset this twice?
```suggestion
        m_splashScreenImageIndex.Reset();
```

Gems/Atom/Feature/Common/Code/Source/SplashScreen/SplashScreenPass.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+        AZ::TickBus::Handler::BusConnect();
+    }
+
+    void SplashScreenPass::Clear()
+    {
+        m_splashScreenImage = nullptr;
+
+        m_splashScreenImageIndex.Reset();
+        m_splashScreenImageIndex.Reset();
```
AMZN-alexpete, line 73: ditto - why twice?

Gems/Atom/Feature/Common/Code/Source/SplashScreen/SplashScreenPass.cpp
```diff
@@ -0,0 +1,147 @@
...
+        m_lastRealTimeStamp = currentRealTimeStamp;
+
+        if (m_beginTimer)
+        {
+            m_screenTime -= realDeltaTime;
+        }
+
+        if (m_settings.m_fading)
+        {
+            m_splashScreenParams.m_fadingFactor = m_screenTime < 0.0f ? 0.0f : m_screenTime / 10.0f;
```
AMZN-alexpete, line 132: It looks like the duration is hard-coded to 10 seconds here
```suggestion
            m_splashScreenParams.m_fadingFactor = m_screenTime < 0.0f ? 0.0f : m_screenTime / m_settings.m_lastingTime;
```
maybe a `1-x^3` curve would be nice?
```suggestion
            m_splashScreenParams.m_fadingFactor = m_screenTime < 0.0f ? 0.0f : (1.f - powf(m_screenTime / m_settings.m_lastingTime, 3.f));
```

AMZN-alexpete (approved):
just some minor things.  Also, we may want to init the window background to black (on windows) so it isn't white before Atom has a chance to draw anything: 
```
HBRUSH brush = CreateSolidBrush(RGB(0, 0, 0));
SetClassLongPtr(hwnd, GCLP_HBRBACKGROUND, (LONG_PTR)brush);
```

Gems/Atom/Feature/Common/Code/Source/SplashScreen/SplashScreenPass.h
jiaweig-amzn: I'll remove it. I intended to add some control to render an image at the center with a fixed size.
Currently, the image will be stretched to fit the window size.

Gems/Atom/Feature/Common/Code/Source/SplashScreen/SplashScreenPass.cpp
jiaweig-amzn: My bad. One of them should be the other index.

Gems/Atom/Feature/Common/Code/Include/Atom/Feature/SplashScreen/SplashScreenSettings.h
jiaweig-amzn: We need a minimal set of resources that can be loaded to render the splash screen. But now the engine reaches the MainLoop only after everything are initialized. We should need some further work to allow Atom to take some priority, something like preInit/postInit.

Gems/Atom/Feature/Common/Code/Source/SplashScreen/SplashScreenPass.cpp
jiaweig-amzn: I was originally trying to fade out the image.
It actually doesn't matter that much. I mainly used the fading factor as a case of how to manipulate the shader behavior for this pass. But lacking of a use case made it hard to imagine what we need. For example, we can have multiple images display in a row, accepting input to skip the splash screen. Or we may need animation for scrolling text.

Gems/Atom/Feature/Common/Code/Source/SplashScreen/SplashScreenPass.h
AMZN-alexpete: oh, yeah a fixed image at center would be a nice option, although - being able to render a UICanvas would probably be the nicest option because then you could handle all the of the positioning, scaling even animating in the UICanvas - but I'm guessing this is before all of that is loaded?

Gems/Atom/Feature/Common/Registry/splash_screen.setreg
```diff
@@ -0,0 +1,16 @@
...
+        "Atom": {
+            "Feature": {
+                "SplashScreenEnable": false,
+                "SplashScreen": {
+                    "$type": "AZ::Render::SplashScreenSettings",
```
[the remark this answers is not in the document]
jiaweig-amzn: I think the type is still necessary to deserialize to the data class.
There are a few settings and can be more in the future. Grouping them in a class can save some lines of code.

PR #14708, opened 2023-02-21
Fixes FileExistsError in LTT artifact manager

evanchia-ly-sdets:
## What does this PR do?
Adds exist_ok params to LTT artifact manager mkdir calls and also adds extra debug text when errors are caught.

## How was this PR tested?
Ran unit tests locally.

### DIFF
--- a/Tools/LyTestTools/ly_test_tools/_internal/managers/artifact_manager.py
+++ b/Tools/LyTestTools/ly_test_tools/_internal/managers/artifact_manager.py
@@ -115,14 +115,22 @@ class ArtifactManager(object):
                                  artifact_name if artifact_name is not None else os.path.basename(artifact_path))
         if os.path.exists(dest_path):
             dest_path = self._get_collision_handled_filename(dest_path, amount)
-
         logger.debug("Copying artifact from '{}' to '{}'".format(artifact_path, dest_path))
 
-        if os.path.isdir(artifact_path):
-            shutil.copytree(artifact_path, dest_path)
-        else:
-            shutil.copy(artifact_path, dest_path)
-            os.chmod(dest_path, stat.S_IWRITE | stat.S_IREAD | stat.S_IEXEC)
+        dest_folder, _ = os.path.split(dest_path)
+        try:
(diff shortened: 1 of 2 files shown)


Tools/LyTestTools/ly_test_tools/_internal/managers/artifact_manager.py
```diff
@@ -115,14 +115,22 @@ def save_artifact(self, artifact_path, artifact_name=None, amount=1):
...
-        else:
-            shutil.copy(artifact_path, dest_path)
-            os.chmod(dest_path, stat.S_IWRITE | stat.S_IREAD | stat.S_IEXEC)
+        dest_folder, _ = os.path.split(dest_path)
+        try:
+            if not os.path.exists(dest_folder):
+                os.makedirs(dest_folder, exist_ok=True)
+
+            if os.path.isdir(artifact_path):
+                shutil.copytree(artifact_path, dest_path, dirs_exist_ok=True)
```
Kadino, line 126: This will overwrite duplicate files that already exist in that directory. Is this appropriate?

The parent folder `dest_path` should already be collision-handled, such that the only reason a duplicate filename would exist is that the user somehow sent a folder with multiple artifacts with an identical sub-path. This seems improbable, and could probably only exist when thrashing the directory during iteration.

Kadino (approved):

Tools/LyTestTools/ly_test_tools/_internal/managers/artifact_manager.py
evanchia-ly-sdets: Overriding duplicate files should already be the default behavior and appropriate in my opinion.

Yeah I wasn't sure if I was being overly cautious, but I figured it's better to have the dirs_exist_ok=True than not have it.

Kadino: Guess I should have clarified my comments by saying "[...] Is this appropriate? I think so." before exploring why it appears to be.

jckand-amzn (approved):

PR #14712, opened 2023-02-22
Project exporter cli json workflow prototype

tkothadev:
This is a prototype of the JSON based workflow for exporting projects.

Instead of relying on a python script directly, we use a json workflow from the project directory to coordinate various steps of the building process. The data oriented nature, and ability to annotate steps does make it easier for the user to keep track of what steps are occuring, and can enable us more introspection from the Project Manager UI.

We also do not lose any of the features of just running the external python scripts directly (and can even have those scripts import the o3de cli, along with passing along arguments by pre-injecting variables into the arguments list before the script is executed). The code is more complex to run through a JSON workflow than just loading a single external script, but we gain the ability to validate various steps.

basic replacements also work as a simple variable mechanism for injecting values as desired.

### DIFF
--- a/scripts/o3de/example_project_export_scripts/newspaper_delivery_project_export_workflow.json
+++ b/scripts/o3de/example_project_export_scripts/newspaper_delivery_project_export_workflow.json
@@ -1,0 +1,23 @@
+{
+    "version":"0.0.0",
+    "workflow":[
+        {"step":"build editor tools", "executable":"cmake",
+            "args":["--build","build/windows_vs2022","--target","Editor","--target","AssetProcessorBatch","--config","profile", "--", "/m   "],
+            "cwd":"${EnginePath}"},
+
+        {"print":"Editor tools built! Now starting Asset Processor"},
+        {"comment":"This was used to successfully test loading python scripts", "step":"python-test-script", "using-python":true, "executable":"${ProjectPath}\\export_rules\\test_export.py",
+            "args":["${EnginePath}", "${ProjectPath}"]},
+
+        {"step":"asset processor", "executable":"${EnginePath}\\build\\windows_vs2022\\bin\\profile\\AssetProcessorBatch.exe",
+            "args":["--platform=pc", "--project-path=${ProjectPath}"]},
+        
+        {"print":"Assets processed! Now to build the project!"},
+
+        {"step":"prepare project", "executable":"cmake",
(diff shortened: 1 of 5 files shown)


scripts/o3de/example_project_export_scripts/newspaper_delivery_project_export_workflow.json
```diff
@@ -0,0 +1,23 @@
+{
```
AMZN-stankowi, line 1: In favor of the python based solution for the short term: This is less human readable than Python.

I think a json based workflow system works best with a UI of some kind to manage the file

AMZN-stankowi:
A few other things to investigate that came up in a discussion on this prototype:
1. Can you make use of existing systems for interacting with cmake from python? Maybe https://github.com/o3de/o3de/blob/development/scripts/o3de/o3de/cmake.py
2. For cmake configuration, we should defer to existing cmake functionality as much as possible, avoid having this exporter set flags itself unless required. Can we use cmake presets for generating monolithic and non-monolithic build artifacts, instead of using the DLY_MONOLITHIC_GAME flag? https://github.com/o3de/o3de/blob/development/cmake/Platform/Windows/CMakePresets.json
3. Do we have configuration flags we change in configuration for server builds? If so, we might need another cmake folder generation step for servers. Right now we don't think that's true, but it's something to think about when looking at how MultiplayerSample would use this functionality.
4. For multiplayer projects, like multiplayer sample, what should the export directory structure look like? It's likely we don't want to mix and match server and client exported projects, especially assets. This is also incentive for us to take a deeper look at the concept of platform and how we are processing assets for server and clients.
5. The above notes really show how MultiplayerSample really stress tests this system, so we'll want to involve people familiar with that project in later discussions.
6. Open question - can we, should re re-use artifacts generated outside this system to accelerate steps? For example, if I've already built non-monolithic profile locally, should we try to re-use those build artifacts, instead of creating a subdirectory for cmake non-monolithic artifacts? Same with processed assets. Should we make use of the existing asset cache and save time if assets have been processed, or specifically process assets to a clean cache? Should we sandbox artifacts from the project exporter tool.
7. Need to farther investigation versioning needs for these builds. Current versioning is mostly around API compatibility. Project versioning helps with things like "My project's dedicated server allows connections from clients with this version range", tracking bugs reported from players who have downloaded the game, etc. Do we want to support multiplayer layers of versioning? For example, a game may put out a binary update once a month on the application stores its released on, but it might do weekly gameplay balance patches that are handled through asset bundles downloaded weekly.
8. We should explore branching steps more. For example, if I provide a seed list to use on the command line, then the step that creates a seed list should be skipped, and the passed in seed list used instead for the following step, asset list generation.

PR #14713, opened 2023-02-22
bugfix for inconsistent rotation calculation order.

Jackie9527:
## What does this PR do?
1. Open Editor.
2. Click Tools -> Track View.
3. Create a sequence.
4. Add some nodes.
5. Record a key frame, modify the rotation of the entity.
image

## How was this PR tested?
image

### DIFF
--- a/Code/Legacy/CryCommon/Cry_Quat.h
+++ b/Code/Legacy/CryCommon/Cry_Quat.h
@@ -395,6 +395,30 @@ struct Quat_tpl
         return q;
     }
 
+    // creatr rotation quatation base on ZYX axis
+    ILINE void SetRotationZYX(const Ang3_tpl<F>& a)
+    {
+        assert(a.IsValid());
+        F sx, cx;
+        sincos_tpl(F(a.x * F(0.5)), &sx, &cx);
+        F sy, cy;
+        sincos_tpl(F(a.y * F(0.5)), &sy, &cy);
+        F sz, cz;
+        sincos_tpl(F(a.z * F(0.5)), &sz, &cz);
+        w = cx * cy * cz - sx * sy * sz;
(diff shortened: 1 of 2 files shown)


gadams3 (approved):
Even though these are “legacy”, are there any tests for the functions? At some point, everything should be migrated to the AZ math types.

Code/Legacy/CryCommon/Cry_Quat.h
```diff
@@ -395,6 +395,30 @@ struct Quat_tpl
         return q;
     }
 
+    // creatr rotation quatation base on ZYX axis
```
AMZN-daimini, line 398: nit: "create rotation quaternion based on ZYX axis", I guess? :)

AMZN-daimini (approved):
Same as what Guthrie said, it would make sense to look into moving this to use AZ math types over time. No need to block this tho.

Code/Legacy/CryCommon/Cry_Quat.h
```diff
@@ -395,6 +395,30 @@ struct Quat_tpl
...
+    ILINE void SetRotationZYX(const Ang3_tpl<F>& a)
```
hultonha, line 399: update: If we could look at AZ::Quaternion and see if this functionality is already supported could we use that instead (we are long-term planning to deprecate and remove CryMath so adding more functionality should ideally be avoided if possible

hultonha: We could do this as a follow-up PR if needed, but might be nice to do a pass over Track View (with tests) to update the use of the math library (this is of course a large change though, so might need some thought/planning).

cgalvan (approved):

Issue #14597, opened 2023-02-15
Editor: Toggle Viewport Helpers selection should be saved as a preference.


galibzon:
**Is your feature request related to a problem? Please describe.**
Every time I open the Editor, I have to go to the  "Toggle Viewport Helpers" menu and uncheck "Helpers" and "Icons".

**Describe the solution you'd like**
Whatever selection is made for  "Toggle Viewport Helpers", it should be stored in disk as any other Editor Preference, so the next time I open the Editor those helpers are shown or hidden depending on the last selection.

**Describe alternatives you've considered**
N/A

**Additional context**
This can save a lot of time in scenarios where the Editor is shutdown and restarted and each time the user doesn't have to change the "Toggle Viewport Helpers" selections.

PR #14724, opened 2023-02-22
Helper visibility settings changes persist

amzn-ahmadkrm:
## What does this PR do?

Fixes https://github.com/o3de/o3de/issues/14597

This PR addresses https://github.com/o3de/o3de/issues/14597 where the helpers visibility settings do not persist on restarting the Editor and have to be changed every time. These are now saved to the registry when they are changed.

This PR also changes the name of the "Show Helpers" setting to "Show Helpers for all entities", and similarly "Show Helpers for selected entities" to clearly differentiate the two settings. Previously the setting to only show helpers for selected entities was dependent on "Show Helpers" also being selected. They are now mutually exclusive settings. 

image

Video in action:

https://user-images.githubusercontent.com/105638312/221668731-6416fd91-d030-439d-ac2b-cce9c31bf23b.mp4

## How was this PR tested?

This PR was manually tested and also passes the tests in EditorViewportHelperTests.cpp


cgalvan (approved):
This looks good to me, but I'd like to get @rainbj 's eyes on the verbiage.

AMZN-daimini:
If they're mutually exclusive, should we also provide a "No Helpers" option?

I envision that menu to either be like this:

- Icons
- SEPARATOR
- Helpers for all entities
- Helpers for selected entities
- No Helpers

amzn-ahmadkrm:
> If they're mutually exclusive, should we also provide a "No Helpers" option?
> 
> This could be the final menu:
> 
>     * Icons
> 
>     * SEPARATOR
> 
>     * Helpers for all entities
> 
>     * Helpers for selected entities
> 
>     * No Helpers

This could work but would we then need an option for 'No Icons'? Is it un-intuitive to uncheck the helpers options?

AMZN-daimini:
> > If they're mutually exclusive, should we also provide a "No Helpers" option?
> > This could be the final menu:
> > ```
> > * Icons
> > 
> > * SEPARATOR
> > 
> > * Helpers for all entities
> > 
> > * Helpers for selected entities
> > 
> > * No Helpers
> > ```
> 
> This could work but would we then need an option for 'No Icons'? Is it un-intuitive to uncheck the helpers options?

It is if they're mutually exclusive. We have a similar pattern in the menu right next to it that shows the Atom debug info
image

hultonha:
> This looks good to me, but I'd like to get @rainbj 's eyes on the verbiage.

I believe @bhanuja-s already gave input on this so I think we have UX's blessing but more eyes can't hurt 🙂

hultonha:
> > > If they're mutually exclusive, should we also provide a "No Helpers" option?
> > > This could be the final menu:
> > > ```
> > > * Icons
> > > 
> > > * SEPARATOR
> > > 
> > > * Helpers for all entities
> > > 
> > > * Helpers for selected entities
> > > 
> > > * No Helpers
> > > ```
> > 
> > 
> > This could work but would we then need an option for 'No Icons'? Is it un-intuitive to uncheck the helpers options?
> 
> It is if they're mutually exclusive. We have a similar pattern in the menu right next to it that shows the Atom debug info image

Yeah it seems with the separator that would help distinguish them so we wouldn't need the 'No Helpers' option in that case.

hultonha:
Changes look good but it might be helpful to include a video to show the change in action to see how changing the options work and what things look like. Also the divider suggested by @AMZN-daimini would be nice to see too. It's super close though, great work @amzn-ahmadkrm 🥳

AMZN-daimini:
> Yeah it seems with the separator that would help distinguish them so we wouldn't need the 'No Helpers' option in that case.

I still feel like a "No Helpers" option is necessary, with this kind of setup you don't expect to need to click on the checkmark to disable.

hultonha:
> > Yeah it seems with the separator that would help distinguish them so we wouldn't need the 'No Helpers' option in that case.
> 
> I still feel like a "No Helpers" option is necessary, with this kind of setup you don't expect to need to click on the checkmark to disable.

Cool I'm good with that, is that okay to add @amzn-ahmadkrm?

hultonha:
Be good to check with @bhanuja-s and @rainbj too

amzn-ahmadkrm:
Yes working on adding 'No Helpers' now :)

hultonha:
> Yes working on adding 'No Helpers' now :)

Awesome thanks, when you have a moment could you upload the updated video with that change? Thanks!

AMZN-daimini (approved):

Issue #14717, opened 2023-02-22
Deprecation of PhysX::EditorColliderComponentRequests (partially) and PhysX::EditorColliderValidationRequests


moraaar:
## Deprecated APIs
````
virtual void PhysX::EditorColliderComponentRequests::SetSphereRadius(float);
virtual float PhysX::EditorColliderComponentRequests::GetSphereRadius() const;
virtual void PhysX::EditorColliderComponentRequests::SetCapsuleRadius(float);
virtual float PhysX::EditorColliderComponentRequests::GetCapsuleRadius() const;
virtual void PhysX::EditorColliderComponentRequests::SetCapsuleHeight(float);
virtual float PhysX::EditorColliderComponentRequests::GetCapsuleHeight() const;
virtual void PhysX::EditorColliderComponentRequests::SetCylinderRadius(float);
virtual float PhysX::EditorColliderComponentRequests::GetCylinderRadius() const;
virtual void PhysX::EditorColliderComponentRequests::SetCylinderHeight(float);
virtual float PhysX::EditorColliderComponentRequests::GetCylinderHeight() const;
virtual void PhysX::EditorColliderComponentRequests::SetCylinderSubdivisionCount(AZ::u8);
virtual AZ::u8 PhysX::EditorColliderComponentRequests::GetCylinderSubdivisionCount() const;
virtual void PhysX::EditorColliderComponentRequests::SetShapeType(Physics::ShapeType);

virtual void PhysX::EditorColliderComponentRequests::SetAssetScale(const AZ::Vector3&);
virtual AZ::Vector3 PhysX::EditorColliderComponentRequests::GetAssetScale() const;

class EditorColliderValidationRequests;
using EditorColliderValidationRequestBus = AZ::EBus<EditorColliderValidationRequests>;
````

## Alternatives APIs
```
// Name modification EditorColliderComponentRequests -> EditorPrimitiveColliderComponentRequests
virtual void PhysX::EditorPrimitiveColliderComponentRequests::SetSphereRadius(float);
virtual float PhysX::EditorPrimitiveColliderComponentRequests::GetSphereRadius() const;
virtual void PhysX::EditorPrimitiveColliderComponentRequests::SetCapsuleRadius(float);
virtual float PhysX::EditorPrimitiveColliderComponentRequests::GetCapsuleRadius() const;
virtual void PhysX::EditorPrimitiveColliderComponentRequests::SetCapsuleHeight(float);
[...]

PR #14725, opened 2023-02-22
Separate editor PhysX Collider in two components: PhysX Primitive Collider and PhysX Mesh Collider

moraaar:
**Note**
This work has been divided into multiple PRs to make it easier to review. Merging each PR into the staging branch `MeshColliderComponent_Staging`.

**Other PRs related to this change**
- https://github.com/o3de/o3de/pull/14726
- https://github.com/o3de/o3de/pull/14729
- https://github.com/o3de/o3de/pull/14732
- https://github.com/o3de/o3de/pull/14730

## What does this PR do?

In order to improve the UX using colliders and make it simpler to use at an entry level we've split **PhysX Collide**r in 2 components:
- **PhysX Primitive Collider**: This collider will take care of Box, Sphere, Capsule and Cylinder shape colliders, which are easy to use don't require any additional asset.
- **PhysX Mesh Collider**: This collider will handle the case of using a PhysX Mesh Asset (.pxmesh product), which requires to be created in FBX Settings.

Changes done:
- Added Editor Mesh Collider component class to handle the physics asset shape type.
- Removed code related to physics asset in editor collider component and renamed it in UI from 'PhysX Collider' to 'PhysX Primitive Collider'.
- Split the `EditorColliderComponentRequests` interface to separate better the functionality of primitive vs mesh colliders, following deprecation policy (https://github.com/o3de/o3de/issues/14717).

NOTE: Collider runtime components have not changed, only the editor one (`EditorColliderComponent`).

## How was this PR tested?

Manual testing of PhysX Primitive Collider and PhysX Mesh Collider.
Created unit tests for PhysX Mesh Collider.
AR passes, all unit tests and python tests pass.


Gems/PhysX/Code/Source/EditorColliderComponent.cpp
```diff
@@ -379,9 +330,28 @@ namespace PhysX
...
+        // O3DE_DEPRECATION_NOTICE(GHI-14718)
+        // If initial value is PhysicsAsset, default to Box and update the UI.
+        // To be removed when m_shapeType intial value is changed to Box with GHI-14718.
+        if (m_shapeConfiguration.m_shapeType == Physics::ShapeType::PhysicsAsset)
+        {
+            m_shapeConfiguration.m_shapeType = Physics::ShapeType::Box;
+            AzToolsFramework::PropertyEditorGUIMessages::Bus::Broadcast(
+                &AzToolsFramework::PropertyEditorGUIMessages::RequestRefresh,
+                AzToolsFramework::PropertyModificationRefreshLevel::Refresh_AttributesAndValues);
+        }
```
moraaar, line 354: This entire cpp file is the same as before but removing anything to do with PhysX Asset (which will appear in the `EditorMeshColliderComponent.cpp` new file later).

This section is the part more worth highlighting.

Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
```diff
@@ -0,0 +1,1101 @@
...
+    }
+
+    void EditorMeshColliderComponent::GetDependentServices(AZ::ComponentDescriptor::DependencyArrayType& dependent)
+    {
+        dependent.push_back(AZ_CRC_CE("NonUniformScaleService"));
+    }
+
+    void EditorMeshColliderComponent::Reflect(AZ::ReflectContext* context)
+    {
+        EditorProxyAssetShapeConfig::Reflect(context);
```
moraaar, line 122: This file is basically a copy of `EditorColliderComponent` keeping everything related to PhysX Asset and removing anything to do with primitive shapes. This class now uses `EditorProxyAssetShapeConfig` which only configures a physx asset.

cgalvan (approved):

Gems/PhysX/Code/Source/EditorColliderComponent.h
```diff
@@ -215,22 +205,20 @@ namespace PhysX
...
-            const AZ::Data::Instance<AZ::RPI::Model>& model) override;
-
-        // PhysX::ColliderShapeBus
+        // PhysX::ColliderShapeBus overrides ...
         AZ::Aabb GetColliderShapeAabb() override;
         bool IsTrigger() override;
 
-        // PhysX::EditorColliderComponentRequestBus
+        // PhysX::EditorColliderComponentRequestBus overrides ...
         void SetColliderOffset(const AZ::Vector3& offset) override;
```
greerdv, line 213: Does this stuff end up getting duplicated between the two collider components? How much stuff is duplicated?

Gems/PhysX/Code/Include/PhysX/EditorColliderComponentRequestBus.h
```diff
@@ -39,85 +39,191 @@ namespace PhysX
...
+    AZ_PUSH_DISABLE_WARNING(4996, "-Wdeprecated-declarations")
     using EditorColliderValidationRequestBus = AZ::EBus<EditorColliderValidationRequests>;
+    AZ_POP_DISABLE_WARNING
+    
+    //! This is a Bus in order to communicate the status of the meshes of the collider and avoid dependencies with the rigidbody
+    class EditorMeshColliderValidationRequests : public AZ::ComponentBus
+    {
+    public:
+        //! Checks if the the mesh in the collider is correct with the current state of the Rigidbody!
+        virtual void ValidateRigidBodyMeshGeometryType() = 0;
```
hultonha, line 223: suggestion: This might be outside the scope of this change but it seems like it might be useful if this function returned a `bool` to indicate success or failure of the validation?

Gems/PhysX/Code/Source/EditorColliderComponent.cpp
```diff
@@ -71,36 +67,22 @@ namespace PhysX
         }
     }
 
-    void EditorProxyAssetShapeConfig::Reflect(AZ::ReflectContext* context)
+    // O3DE_DEPRECATION_NOTICE(GHI-14718)
```
hultonha, line 70: nit: You can leave this but it's generally fine to leave API deprecation comments in the .h files that are public (rather than repeating in multiple places in the implementation)

Gems/PhysX/Code/Source/EditorColliderComponent.cpp
```diff
@@ -204,7 +173,7 @@ namespace PhysX
         if (auto serializeContext = azrtti_cast<AZ::SerializeContext*>(context))
         {
             serializeContext->Class<EditorColliderComponent, EditorComponentBase>()
-                ->Version(1 + (1<<PX_PHYSICS_VERSION_MAJOR)) // Use PhysX version to trigger prefabs recompilation when switching between PhysX 4 and 5.
+                ->Version(2)
```
hultonha, line 176: @SergeyAMZN do we need to keep the previous logic? IIRC correct that was quite important if customers would switch between PhysX 4 or 5

Gems/PhysX/Code/Source/EditorColliderComponent.cpp
```diff
@@ -379,9 +330,28 @@ namespace PhysX
         }
     }
 
+    bool EditorProxyShapeConfig::IsNonUniformlyScaledPrimitive() const
+    {
+        return m_hasNonUniformScale && (Utils::IsPrimitiveShape(GetCurrent()) || IsCylinderConfig());
```
hultonha, line 335: nit: Second parens around `Utils::IsPrimitiveShape` are redundant and I feel make this a little harder to read as it looks like the brackets might be around the final condition

Gems/PhysX/Code/Source/EditorMeshColliderComponent.h
```diff
@@ -0,0 +1,266 @@
...
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#pragma once
+
+#include <AzCore/Component/TickBus.h>
```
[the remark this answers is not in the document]
hultonha: Good call!

Gems/PhysX/Code/Source/EditorMeshColliderComponent.h
```diff
@@ -0,0 +1,266 @@
...
+            const AZ::Data::Asset<AZ::Data::AssetData>& renderMeshAsset, const AZStd::vector<AZ::Data::AssetId>& physicsAssets);
+
+        void ValidateAssetMaterials();
+
+        EditorProxyAssetShapeConfig m_shapeConfiguration;
+        Physics::ColliderConfiguration m_configuration;
+
+        using ComponentModeDelegate = AzToolsFramework::ComponentModeFramework::ComponentModeDelegate;
+        ComponentModeDelegate m_componentModeDelegate; //!< Responsible for detecting ComponentMode activation
+                                                       //!< and creating a concrete ComponentMode.
```
hultonha, line 229: It looks like this can be removed too as I don't think this Editor Component should have a component mode right?

Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
hultonha: I think as myself and @greerdv mentioned the Component Mode logic should now be removed from here too

Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
```diff
@@ -0,0 +1,1101 @@
...
+
+        if (auto serializeContext = azrtti_cast<AZ::SerializeContext*>(context))
+        {
+            serializeContext->Class<EditorMeshColliderComponent, EditorComponentBase>()
+                ->Version(1 + (1<<PX_PHYSICS_VERSION_MAJOR)) // Use PhysX version to trigger prefabs recompilation when switching between PhysX 4 and 5.
```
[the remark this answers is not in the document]
hultonha: @SergeyAMZN can speak to this, I believe the intention was just to have a big change to ensure the assets would get reprocessed and it intentionally didn't want just a single increment when switching back and forward

Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
```diff
@@ -0,0 +1,1101 @@
...
+        AzToolsFramework::ShapeManipulatorRequestBus::Handler::BusConnect(
+            AZ::EntityComponentIdPair(GetEntityId(), GetId()));
+        ColliderShapeRequestBus::Handler::BusConnect(GetEntityId());
+        AZ::Render::MeshComponentNotificationBus::Handler::BusConnect(GetEntityId());
+        EditorColliderComponentRequestBus::Handler::BusConnect(AZ::EntityComponentIdPair(GetEntityId(), GetId()));
+        EditorMeshColliderComponentRequestBus::Handler::BusConnect(AZ::EntityComponentIdPair(GetEntityId(), GetId()));
+        EditorMeshColliderValidationRequestBus::Handler::BusConnect(GetEntityId());
+        AzFramework::BoundsRequestBus::Handler::BusConnect(GetEntityId());
+        AzToolsFramework::EditorComponentSelectionRequestsBus::Handler::BusConnect(GetEntityId());
+        m_nonUniformScaleChangedHandler = AZ::NonUniformScaleChangedEvent::Handler(
```
hultonha, line 249: @greerdv Maybe non-uniform scale stuff can now be removed for this type? Or is it still needed?

Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
```diff
@@ -0,0 +1,1101 @@
...
+    void EditorMeshColliderComponent::Display(
+        [[maybe_unused]] const AzFramework::ViewportInfo& viewportInfo,
+        AzFramework::DebugDisplayRequests& debugDisplay) const
+    {
+        if (!m_colliderDebugDraw.HasCachedGeometry())
+        {
+            BuildDebugDrawMesh();
+        }
+
+        if (m_colliderDebugDraw.HasCachedGeometry())
```
hultonha, line 710: question: This check seems redundant unless `BuildDebugDrawMesh` could fail? Can that happen do we know?

Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
```diff
@@ -0,0 +1,1101 @@
...
+        return GetAabb();
+    }
+
+    AZ::Aabb EditorMeshColliderComponent::GetLocalBounds()
+    {
+        AZ::Aabb worldBounds = GetWorldBounds();
+        if (worldBounds.IsValid())
+        {
+            return worldBounds.GetTransformedAabb(m_cachedWorldTransform.GetInverse());
+        }
```
hultonha, line 1034: comment: I realize this was probably like this before but it does seem odd we get the local bounds from the world bounds instead of the other way around 😅 not a huge deal though

hultonha (commented):
Looks solid to me! Nice work 🙂 I agree with @greerdv's comments, once those and a couple of mine are addressed I can approve 👍

Gems/PhysX/Code/Include/PhysX/EditorColliderComponentRequestBus.h
```diff
@@ -39,85 +39,191 @@ namespace PhysX
...
+    };
+
+    using EditorPrimitiveColliderComponentRequestBus = AZ::EBus<EditorPrimitiveColliderComponentRequests>;
+
+    //! Request bus for colliders using PhysX mesh assets.
+    class EditorMeshColliderComponentRequests
+        : public AZ::EntityComponentBus
+    {
+    public:
+
```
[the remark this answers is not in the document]
moraaar: will remove

Gems/PhysX/Code/Include/PhysX/EditorColliderComponentRequestBus.h
```diff
@@ -39,85 +39,191 @@ namespace PhysX
...
 
+    AZ_PUSH_DISABLE_WARNING(4996, "-Wdeprecated-declarations")
     using EditorColliderValidationRequestBus = AZ::EBus<EditorColliderValidationRequests>;
+    AZ_POP_DISABLE_WARNING
+    
+    //! This is a Bus in order to communicate the status of the meshes of the collider and avoid dependencies with the rigidbody
+    class EditorMeshColliderValidationRequests : public AZ::ComponentBus
+    {
+    public:
+        //! Checks if the the mesh in the collider is correct with the current state of the Rigidbody!
```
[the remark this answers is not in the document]
moraaar: will remove

Gems/PhysX/Code/Include/PhysX/EditorColliderComponentRequestBus.h
```diff
@@ -39,85 +39,191 @@ namespace PhysX
...
     public:
         /// Checks if the the mesh in the collider is correct with the current state of the Rigidbody!
         virtual void ValidateRigidBodyMeshGeometryType() = 0;
     };
 
+    AZ_PUSH_DISABLE_WARNING(4996, "-Wdeprecated-declarations")
     using EditorColliderValidationRequestBus = AZ::EBus<EditorColliderValidationRequests>;
+    AZ_POP_DISABLE_WARNING
+    
+    //! This is a Bus in order to communicate the status of the meshes of the collider and avoid dependencies with the rigidbody
```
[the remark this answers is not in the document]
moraaar: I just copied it, but your version is better, I'll change it in both buses.

Gems/PhysX/Code/Source/EditorColliderComponent.h
```diff
@@ -83,12 +84,18 @@ namespace PhysX
         EditorProxyShapeConfig() = default;
         EditorProxyShapeConfig(const Physics::ShapeConfiguration& shapeConfiguration);
 
+        // O3DE_DEPRECATION_NOTICE(GHI-14718)
+        // Initial value for m_shapeType needs to remain PhysicsAsset
+        // to support command to convert EditorColliderComponent to
+        // EditorMeshColliderComponent. This is because the lack of 'ShapeType'
+        // property in prefabs still need to mean it's using PhysicsAsset type.
```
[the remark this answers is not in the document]
moraaar: >I think this would read better if you change "need" to "needs".
Thanks, I'll fix that.

>Is it because nothing gets serialized to json if the property is equal to its default value
Exactly, so if you change the default value now you've lost knowing what value it had before.

>its default previously was PhysicsAsset so it has to stay like that otherwise it will deserialize old data which had the default value incorrectly?
No, its default previously was PhysicsAsset so it has to stay like that otherwise when the converter opens the prefab and reads the entity it'll think it's a box (if box is used as default) and therefore we would lose the information of knowing it was a PhysicsAsset (and therefore it needs to be converted).

Gems/PhysX/Code/Source/EditorColliderComponent.h
```diff
@@ -97,52 +104,44 @@ namespace PhysX
...
         AZ::u32 OnConfigurationChanged();
 
-        Physics::ShapeType m_lastShapeType = Physics::ShapeType::PhysicsAsset;
+        Physics::ShapeType m_lastShapeType = Physics::ShapeType::Box;
     };
 
-    class EditorColliderComponentDescriptor;
-    //! Editor PhysX Collider Component.
+    //! Editor PhysX Primitive Collider Component.
     //!
```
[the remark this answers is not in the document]
moraaar: will remove

Gems/PhysX/Code/Source/EditorColliderComponent.h
```diff
@@ -168,15 +167,16 @@ namespace PhysX
...
+        void Init() override;
         void Activate() override;
+        void Deactivate() override;
 
         void UpdateShapeConfiguration();
 
-        void Deactivate() override;
-
-        //! AzToolsFramework::EntitySelectionEvents
+        //! AzToolsFramework::EntitySelectionEvents overrides ...
```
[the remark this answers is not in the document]
moraaar: agreed, will remove.

Gems/PhysX/Code/Source/EditorColliderComponent.h
```diff
@@ -278,7 +254,8 @@ namespace PhysX
 
         void BuildDebugDrawMesh() const;
 
-        AZ::ComponentDescriptor::StringWarningArray GetComponentWarnings() const { return m_componentWarnings; };
+        EditorProxyShapeConfig m_shapeConfiguration;
```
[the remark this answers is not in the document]
moraaar: Yeah, I can change that.

Gems/PhysX/Code/Source/EditorForceRegionComponent.cpp
```diff
@@ -291,21 +292,31 @@ namespace PhysX
...
-            , &ColliderShapeRequestBus::Events::GetColliderShapeAabb);
+        const AZStd::vector<AZ::Vector3> shapeAabbPoints = [this]()
+        {
+            AZ::Aabb aabb;
+            ColliderShapeRequestBus::EventResult(aabb, GetEntityId(), &ColliderShapeRequestBus::Events::GetColliderShapeAabb);
+
+            const AZ::Vector3 halfExtents = aabb.GetExtents() * 0.5f;
+            AZStd::vector<AZ::Vector3> points = Utils::Geometry::GenerateBoxPoints(-halfExtents, halfExtents);
+
+            const AZ::Vector3 aabbCenter = aabb.GetCenter();
```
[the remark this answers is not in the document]
moraaar: Very good suggestion. Will change.

Gems/PhysX/Code/Source/EditorForceRegionComponent.cpp
```diff
@@ -327,31 +338,22 @@ namespace PhysX
...
+                    , GetWorldTM()
+                    , editorColliderComponent->GetColliderConfiguration().m_position
+                    , editorColliderComponent->GetColliderConfiguration().m_rotation
+                    , m_cachedNonUniformScale);
 
                 DrawForceArrows(randomPoints, debugDisplayRequests);
             }
 
-            else if (editorShapeColliderComponent)
+            if ([[maybe_unused]] auto* editorMeshColliderComponent = azrtti_cast<EditorMeshColliderComponent*>(component))
```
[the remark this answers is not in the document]
moraaar: Well spotted both suggestions. Will change.

Gems/PhysX/Code/Source/EditorMeshColliderComponent.h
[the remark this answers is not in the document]
moraaar: I'll do that in another pass as I'd need to build with non-unity builds locally to make sure I don't break anything and I'm not in a moment to be able to switch. But I'll put a reminder to do a clean up pass later. Thanks.

Gems/PhysX/Code/Source/EditorMeshColliderComponent.h
```diff
@@ -0,0 +1,266 @@
...
+
+    private:
+        bool ShowingSubdivisionLevel() const;
+        AZ::u32 OnConfigurationChanged();
+    };
+
+    class EditorMeshColliderComponentDescriptor;
+
+    //! Editor PhysX Mesh Collider Component.
+    //!
```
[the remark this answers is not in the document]
moraaar: will remove

Gems/PhysX/Code/Source/EditorMeshColliderComponent.h
```diff
@@ -0,0 +1,266 @@
...
+        static void GetRequiredServices(AZ::ComponentDescriptor::DependencyArrayType& required);
+        static void GetDependentServices(AZ::ComponentDescriptor::DependencyArrayType& dependent);
+
+        EditorMeshColliderComponent() = default;
+        EditorMeshColliderComponent(
+            const Physics::ColliderConfiguration& colliderConfiguration, const EditorProxyAssetShapeConfig& proxyAssetShapeConfig);
+        EditorMeshColliderComponent(
+            const Physics::ColliderConfiguration& colliderConfiguration, const Physics::PhysicsAssetShapeConfiguration& assetShapeConfig);
+
+        // these functions are made virtual because we call them from other modules
```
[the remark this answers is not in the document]
moraaar: I see. I've done a search it doesn't look like they are called from outside PhysX gem. I'll remove the comment and make them non-virtual. Thanks.

Gems/PhysX/Code/Source/EditorMeshColliderComponent.h
```diff
@@ -0,0 +1,266 @@
...
+    private:
+        AZ_DISABLE_COPY_MOVE(EditorMeshColliderComponent);
+
+        // AZ::Component overrides ...
+        void Activate() override;
+        void Deactivate() override;
+
+        void UpdateShapeConfiguration();
+
+        //! AzToolsFramework::EntitySelectionEvents overrides ...
```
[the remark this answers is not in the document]
moraaar: will change to normal comment.

Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
```diff
@@ -0,0 +1,1101 @@
...
+                    ->Attribute(AZ::Edit::Attributes::AutoExpand, true)
+                    ->DataElement(AZ::Edit::UIHandlers::Default, &EditorMeshColliderComponent::m_configuration, "Collider Configuration", "Configuration of the collider.")
+                    ->Attribute(AZ::Edit::Attributes::Visibility, AZ::Edit::PropertyVisibility::ShowChildrenOnly)
+                    ->Attribute(AZ::Edit::Attributes::ChangeNotify, &EditorMeshColliderComponent::OnConfigurationChanged)
+                    ->DataElement(
+                        AZ::Edit::UIHandlers::Default,
+                        &EditorMeshColliderComponent::m_shapeConfiguration,
+                        "Physics Asset",
+                        "Configuration of physics asset shape.")
+                    //->Attribute(AZ::Edit::Attributes::Visibility, AZ::Edit::PropertyVisibility::ShowChildrenOnly)
```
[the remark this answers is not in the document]
moraaar: Ups, forgot to remove, not needed. Thanks!

Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
```diff
@@ -0,0 +1,1101 @@
...
+                AzToolsFramework::PropertyEditorGUIMessages::Bus::Broadcast(&AzToolsFramework::PropertyEditorGUIMessages::RequestRefresh,
+                    AzToolsFramework::PropertyModificationRefreshLevel::Refresh_AttributesAndValues);
+            });
+
+        AzToolsFramework::Components::EditorComponentBase::Activate();
+        AzToolsFramework::EntitySelectionEvents::Bus::Handler::BusConnect(GetEntityId());
+        PhysX::MeshColliderComponentRequestsBus::Handler::BusConnect(GetEntityId());
+        AZ::TransformNotificationBus::Handler::BusConnect(GetEntityId());
+        AzToolsFramework::ShapeManipulatorRequestBus::Handler::BusConnect(
+            AZ::EntityComponentIdPair(GetEntityId(), GetId()));
```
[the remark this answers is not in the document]
moraaar: Yeah, I can change that.

Gems/PhysX/Code/Source/EditorColliderComponent.h
moraaar: The `EditorColliderComponentRequestBus` implementation is very similar in both and the skeleton of both classes look the same, but eventually they arrive to do different things in private functions. As a first pass of splitting the component in 2 I think it makes sense as it is now, but in the future we can take another look and see if it's worth refactoring something out. 

I'm glad we're doing this split because I noticed `EditorMeshColliderComponent` has to do a lot than `EditorColliderComponent` (primitive one) just for having to deal with an asset and custom error checking, and it's great all that is out of the primitive collider.

Gems/PhysX/Code/Include/PhysX/EditorColliderComponentRequestBus.h
moraaar: I understand why you'd make that suggestion, but in this case the caller doesn't need to know. This is just a notification from certain points so the collider runs a validation check on itself to show a message to the user if it's wrongly configured.

Gems/PhysX/Code/Source/EditorColliderComponent.cpp
moraaar: There is no second parenthesis around `Utils::IsPrimitiveShape`.

Gems/PhysX/Code/Source/EditorMeshColliderComponent.h
moraaar: Editor Mesh Collider still needs to support component mode to help modify its `Offset`, `Rotation` parameters.

Gems/PhysX/Code/Source/EditorMeshColliderComponent.h
```diff
@@ -0,0 +1,266 @@
...
+        void SetMeshAsset(const AZ::Data::AssetId& id) override;
+        void UpdateMaterialSlotsFromMeshAsset();
+
+        // TransformBus overrides ...
+        void OnTransformChanged(const AZ::Transform& local, const AZ::Transform& world) override;
+
+        // non-uniform scale handling
+        void OnNonUniformScaleChanged(const AZ::Vector3& nonUniformScale);
+
+        // AzToolsFramework::ShapeManipulatorRequestBus overrides ...
```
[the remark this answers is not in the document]
moraaar: Is this needed to control the `Asset Scale`, `Offset` and `Rotation` properties of the mesh collider?

Gems/PhysX/Code/Source/EditorColliderComponent.h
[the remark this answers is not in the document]
greerdv: Yeah, that's what I meant. But like I say it took me a while to figure that out, so maybe the comment can be made clearer?

Gems/PhysX/Code/Source/EditorMeshColliderComponent.h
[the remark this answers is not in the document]
greerdv: Cool, sounds good. Maybe we should make a task to do a pass over the whole gem to see if there are opportunities to tidy up includes.

Gems/PhysX/Code/Source/EditorMeshColliderComponent.h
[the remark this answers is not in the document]
greerdv: For asset component mode, I don't think that happens using this bus, but I'm not 100% sure. I thought it was only used by the primitive modes though.

Gems/PhysX/Code/Source/EditorColliderComponent.h
greerdv: Cool, sounds good.

Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
greerdv: No, you can have non-uniform scale with asset colliders. In fact, that was the reason we made collider component compatible with NUS in the first place. Since prior to this change, the collider component supported both assets and primitives, it actually meant a bunch of work had to be done to replace primitives with convex meshes if they had NUS. If this change had happened first, we could actually have just made primitive colliders incompatible with NUS component.

greerdv: As an aside, I would really like to either get rid of asset scale or make it uniform. I don't think having NUS in there adds value, since if you want it you can use the NUS component. But it does break primitives inside asset colliders when there is no NUS component.

Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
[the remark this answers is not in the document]
moraaar: Correct, this is to make it so the class has different versions when switching PhysX versions because cooked data can be added to the serialized data of the component, changing the version forces asset processor to reprocess the prefab again. But I thought I only needed it in the Mesh Collider, which is incorrect, the primitives colliders can also cooked data, for example for cylinders. I'll add it back to the `EditorColliderComponent` class too. I verified this with @SergeyAMZN.

Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
greerdv: PhysX only provides the bounds in world space, you have to do it this way round unless you want some very complex (and probably fairly expensive) logic for computing the local bounds.

greerdv (approved):

Gems/PhysX/Code/Source/EditorColliderComponent.cpp
moraaar: I talked to @SergeyAMZN and yes, we need to keep this, I thought it was only for Mesh Collider, but that's not true, it's needed for whenever it cooks data, which it's done when using cylinder (handled by this class) or non-uniform scale, I added it back.

Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
moraaar: > As an aside, I would really like to either get rid of asset scale or make it uniform. I don't think having NUS in there adds value, since if you want it you can use the NUS component. But it does break primitives inside asset colliders when there is no NUS component.

+1 on this as a follow up task

Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
moraaar: You might be right. These editor classes have evolved having multiple checks along the way, probably because they where needed at some point. So I'd rather keep it to be safe and have the same code as before.

Gems/PhysX/Code/Source/EditorColliderComponent.h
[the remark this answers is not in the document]
moraaar: Yeah, I wasn't inspired, was I. This the troubling bit 

"This is because the lack of 'ShapeType' property in prefabs still needs to mean it's using PhysicsAsset type."

What about this instead?

"This is because prefabs do not store in JSON the default values and therefore the converter would lose the ability to know if the type was PhysX Asset before to convert the component to a Editor Mesh Collider."

Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
moraaar: Editor Mesh Collider uses component mode to handle the offset, rotation and asset scale of the component.

Gems/PhysX/Code/Source/EditorMeshColliderComponent.h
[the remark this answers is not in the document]
moraaar: I dived into this and you are right, `ShapeManipulatorRequestBus` is only used by primitives. I'll remove it from Editor Mesh Component! Thanks!

AMZN-alexpete:
@moraaar if these changes are breaking any APIs or adding new functionality, please update the appropriate version fields - looks like it'd just be the one in `Gems/PhysX/gem.json` if any.

Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
SergeyAMZN: Asset scale is just for the collider mesh/convex while NUS component applies for the entire entity. 
There may be use cases when you want to adjust the scale just for the collision geom without changing the entity scale.

moraaar:
> @moraaar if these changes are breaking any APIs or adding new functionality, please update the appropriate version fields - looks like it'd just be the one in `Gems/PhysX/gem.json` if any.

Done

Gems/PhysX/Code/Include/PhysX/EditorColliderComponentRequestBus.h
hultonha: Okay understood, thanks for the explanation 👍

Gems/PhysX/Code/Source/EditorColliderComponent.cpp
hultonha: Okay perfect thanks 👍

Gems/PhysX/Code/Source/EditorColliderComponent.cpp
hultonha: Apologies I misread this, ignore me 🤦

Gems/PhysX/Code/Source/EditorMeshColliderComponent.h
hultonha: I see good to know 👍

Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
hultonha: I see got you thanks! 👍

hultonha (approved):

Gems/PhysX/Code/Source/EditorMeshColliderComponent.h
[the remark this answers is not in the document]
moraaar: Follow up PR doing this clean up: https://github.com/o3de/o3de/pull/14955

Issue #9660, opened 2022-05-19
Multiplayer Components Missing Tooltips 


AMZN-Gene:
**Describe the bug**
Multiplayer components are missing tooltips

**Steps to reproduce**
1. Open editor  
3. Add a Multiplayer Component to an entity (for example Network Transform)

**Expected behavior**
Components have tooltips 

**Actual behavior**
Tooltip just says the name of the component
![image](https://user-images.githubusercontent.com/32776221/169362535-f43d0416-e574-409b-8065-efbba8edfa86.png)

**Found in Branch**
development latest

lmbr-pip:
Relates to [8662](https://github.com/o3de/o3de/issues/8662)

AMZN-Gene:
This came up again during latest WF6 session.

PR #14727, opened 2023-02-22
Fix descriptions on generated components.

mbalfour-amzn:
## What does this PR do?

According to the documentation, AutoComponent.xml files should use "Description" as the name of the field to use for descriptions:
https://www.o3de.org/docs/user-guide/networking/multiplayer/autocomponents/

However, the jinja file was looking for "Desc" instead, causing correctly-formatted AutoComponent.xml files to fail to generate descriptions in the generated components.

This changes the jinja file to look for "Description" so the descriptions show up.

## How was this PR tested?

Regenerated the files in MultiplayerSample and verified that the Description fields are now populated for the components. Spot-checked both the code and some of the tooltips in the Editor.

image


AMZN-Olex (approved):

cgalvan (approved):

lmbr-pip:
Believe this will address https://github.com/o3de/o3de/issues/9660

PR #14569, opened 2023-02-13
Disabling/moving consistently failing tests from the Periodic Test Suites

jckand-amzn:
## What does this PR do?

Moves several consistently and/or intermittently failing tests to the various feature area's Sandbox suite, or skips/conditionally skips the affected tests.

Tickets have been cut to fix the failing tests for each feature area:

https://github.com/o3de/o3de/issues/14540
https://github.com/o3de/o3de/issues/14520
https://github.com/o3de/o3de/issues/14519
https://github.com/o3de/o3de/issues/14514

## How was this PR tested?

Ran updated Periodic test suite on internal Jenkins with 0 test failures.

PR #14728, opened 2023-02-22
Updating conditional skips for several Asset Pipeline tests

jckand-amzn:
## What does this PR do?

- Updates conditional skips for several inaccurately skipped tests in the Asset Pipeline team's test suite
- Addresses feedback from https://github.com/o3de/o3de/pull/14569

## How was this PR tested?

- Running Periodic test suite on internal Jenkins

### DIFF
--- a/AutomatedTesting/Gem/PythonTests/assetpipeline/asset_processor_tests/asset_relocator_tests.py
+++ b/AutomatedTesting/Gem/PythonTests/assetpipeline/asset_processor_tests/asset_relocator_tests.py
@@ -260,13 +260,13 @@ class TestsAssetRelocator_WindowsAndMac(object):
         # Look for expected message inside the log and verify that no move or delete occurs in the log
         utils.validate_log_output(ap_batch_output, [expected_message], [unexpected_message])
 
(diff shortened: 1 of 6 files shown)


AutomatedTesting/Gem/PythonTests/assetpipeline/asset_processor_tests/asset_relocator_tests.py
```diff
@@ -260,13 +260,13 @@ def test_WindowsMacPlatforms_AllowBrokenDependenciesWithoutMoveDelete_RelocatorW
         # Look for expected message inside the log and verify that no move or delete occurs in the log
         utils.validate_log_output(ap_batch_output, [expected_message], [unexpected_message])
 
-    @pytest.mark.skip(reason="https://github.com/o3de/o3de/issues/14514")
+    @pytest.mark.skipif(ly_test_tools.WINDOWS, reason="https://github.com/o3de/o3de/issues/14514")
```
AMZN-stankowi, line 263: I like this, instead of marking skip because both platforms fail, you're leaving the skip for both so the separate reasons its failing are tracked.

AMZN-stankowi (approved):

cgalvan (approved):

Issue #14718, opened 2023-02-22
Deprecation of legacy reflection data in EditorColliderComponent


moraaar:
The functionality of using a Physics Asset as a collider has been moved from `EditorColliderComponent` to `EditorMeshColliderComponent`.

The serialization data of physics asset data has been kept in `EditorColliderComponent` (unexposed to UI) to allow the console command `ed_physxUpdatePrefabsWithColliderComponents` to automatically update prefabs for users, converting `EditorColliderComponent` to `EditorMeshColliderComponent` when needed.

This GHI is for the removal eventually of both the serialization data and `ed_physxUpdatePrefabsWithColliderComponents` command.

## Last Release
The addition of `EditorMeshColliderComponent` and the console command `ed_physxUpdatePrefabsWithColliderComponents` will appear in the release notes in the release after 22.10.0. The old serialization data and the command will be removed by the **third** release after 22.10.0.

## Additional Context
The deprecation policy indicates that deprecated code can be removed after 1 release, but since **`EditorColliderComponent` is a very fundamental component I'd recommend to keep the old data and conversion command for longer and remove it after 2 releases instead of 1.**

iamjbq:
I plan to execute the removal of this depreciated functionality following the upgrade of PhysX to version 5.6.1. #19595. Any code using this deprecated functionality would need to migrate their collider explicitly to the `EditorMeshColliderComponent`.

PR #14729, opened 2023-02-22
Command 'ed_physxUpdatePrefabsWithColliderComponents' update

moraaar:
**Note**
This is part of the work to separate PhysX Collider into 2 components: https://github.com/o3de/o3de/pull/14725. Merging into the staging branch `MeshColliderComponent_Staging`.

## What does this PR do?

Command 'ed_physxUpdatePrefabsWithColliderComponents' now also updates all prefabs in a project by detecting PhysX Colliders using physics assets and converting them into PhysX Mesh Colliders.

Deprecation GHI (https://github.com/o3de/o3de/issues/14718)

## How was this PR tested?

Used the command to update all the prefabs in AutomatedTesting project and it worked as expected: https://github.com/o3de/o3de/pull/14730

### DIFF
--- a/Gems/PhysX/Code/Editor/Source/Components/Conversion/CollidersPrefabConversion.cpp
+++ b/Gems/PhysX/Code/Editor/Source/Components/Conversion/CollidersPrefabConversion.cpp
@@ -131,7 +131,7 @@ namespace PhysX
                 return false;
             }
 
-            // Once the component is removed from the entity we are responsable for its destruction.
+            // Once the component is removed from the entity we are responsible for its destruction.
             delete editorColliderComponent;
 
             entityModified = true;
--- a/Gems/PhysX/Code/Editor/Source/Components/Conversion/CollidersPrefabConversion.cpp
+++ b/Gems/PhysX/Code/Editor/Source/Components/Conversion/CollidersPrefabConversion.cpp
@@ -13,6 +13,7 @@
 #include <AzToolsFramework/API/ToolsApplicationAPI.h>
 
 #include <EditorColliderComponent.h>
+#include <EditorMeshColliderComponent.h>
 #include <EditorShapeColliderComponent.h>
 #include <EditorRigidBodyComponent.h>
 #include <EditorStaticRigidBodyComponent.h>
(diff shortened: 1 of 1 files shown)


Gems/PhysX/Code/Editor/Source/Components/Conversion/CollidersPrefabConversion.cpp
```diff
@@ -23,11 +24,13 @@ namespace PhysX
...
     void UpdatePrefabsWithColliderComponents(const AZ::ConsoleCommandContainer& commandArgs);
 
+    // O3DE_DEPRECATION_NOTICE(GHI-14718)
     AZ_CONSOLEFREEFUNC(
         "ed_physxUpdatePrefabsWithColliderComponents",
         UpdatePrefabsWithColliderComponents,
         AZ::ConsoleFunctorFlags::Null,
-        "Finds entities with collider components and no rigid bodies and updates them to the new pattern which requires a static rigid body component.");
+        "Finds entities with collider components and no rigid bodies and updates them to the new pattern which requires a static rigid body component. "
+        "Finds entities with collider components using physx asset and replace it with a mesh collider component.");
```
hultonha, line 33: Quick question: Does this command now just do both operations in one?

hultonha (approved):
Change looks good, I just wanted to check are there some test cases for this (I guess you have run it a bunch already and will be opening the levels you test it on but just wanted to sanity check we'd tried quite a few different prefabs with this). Thanks!

Gems/PhysX/Code/Editor/Source/Components/Conversion/CollidersPrefabConversion.cpp
moraaar: Yes, this is intentional so in the next o3de release it's only 1 command we ask users to run in the release notes.

hultonha: Understood perfect, and is there a warning/pop-up that we can present users with when opening the level to let them know they need to run this command?

moraaar: At the moment we don't have a way to detect if a prefab needs to be updated or not I'm afraid, but that feature is becoming more and more important and it'd be ideal if the editor will just tell customers when they have to do some process to update their content to work with newer versions of the engine, but that support doesn't exist at the moment. 

It's worth mentioning I'm updating the most important repos in o3de, so all the content in our repos should be updated, it's only for user's own prefabs they will have to run it.

hultonha: I see 🙁 That is a real shame as just having a way to detect and then prompt the user would be incredibly helpful and guard against people not understanding why their level won't open.

Understood, I just think as time goes by that number is going to increase significantly and we should really think about addressing it before then.

Tagging @srikappa-amzn, @monroegm and @forkercat for visibility and I wonder if/when @o3de/sig-content might be able to add support for this.

srikappa-amzn: question: What happens if a user tries to load a prefab with the old data? Will they still be able to load the rest of the prefab except for the entities with the old collider components?  If that is not the case, we should not be pushing this change in its current state imo.

suggestion: Looking at the code in Gems\PhysX\Code\Editor\Source\Components\Conversion\PrefabConversionUtils.cpp , it looks like we are directly making changes to the Prefab DOM data. If that's the case, it shouldn't be hard to provide a notification when a prefab is loaded in the editor (whether it is through loading the level or instantiating a prefab). And systems like PhysX can listen to that notification and execute the same code in UpdatePrefabPhysXColliders(). Then you'll have the ability to either force update user's data or ask them for their permission or just show them a warning.

AMZN-daimini (approved):

PR #14753, opened 2023-02-23
GamePlayController tests for Falling Velocity and Ground Detection

FuzzyCarterAWS:
## What does this PR do?

Adding tests:
- FallingVelocitySets
- FallingVelocitySetsWhileMoving
- SetGroundDetectionHeight
- GroundDetectedWhileMoving

Updates:
- Inverts the logic for creating the ground on a test basis method call instead of on creation of the test basis. This was done to prevent accidental ground collisions delaying test development. Provides updates to existing tests to support this logic.
- Fixed the matcher order on existing tests that were flipping the expected vs actual with actual vs expected so that way if a test fails the logic reported reads correctly. EG: `EXPECT_THAT(actual, expected)` becomes `EXPECT_THAT(expected, actual)`.
- Cleaned up & added comments to existing and new testsj.

## How was this PR tested?

Tests have passed locally in Isolation & within the entire test suite

```
[----------] 6 tests from PhysXDefaultWorldTest
[ RUN      ] PhysXDefaultWorldTest.CharacterGameplayController_GravitySets
[       OK ] PhysXDefaultWorldTest.CharacterGameplayController_GravitySets (3 ms)
[ RUN      ] PhysXDefaultWorldTest.CharacterGameplayController_GravitySetsWhileMoving
[       OK ] PhysXDefaultWorldTest.CharacterGameplayController_GravitySetsWhileMoving (9 ms)
[ RUN      ] PhysXDefaultWorldTest.CharacterGameplayController_FallingVelocitySets
[       OK ] PhysXDefaultWorldTest.CharacterGameplayController_FallingVelocitySets (2 ms)
[ RUN      ] PhysXDefaultWorldTest.CharacterGameplayController_FallingVelocitySetsWhileMoving
[       OK ] PhysXDefaultWorldTest.CharacterGameplayController_FallingVelocitySetsWhileMoving (8 ms)
[ RUN      ] PhysXDefaultWorldTest.CharacterGameplayController_SetGroundDetectionHeight
[       OK ] PhysXDefaultWorldTest.CharacterGameplayController_SetGroundDetectionHeight (1 ms)
[ RUN      ] PhysXDefaultWorldTest.CharacterGameplayController_GroundDetectedWhileMoving
[       OK ] PhysXDefaultWorldTest.CharacterGameplayController_GroundDetectedWhileMoving (19 ms)
[----------] 6 tests from PhysXDefaultWorldTest (70 ms total)

[----------] 7 tests from PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture
[ RUN      ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/0
[       OK ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/0 (8 ms)
[ RUN      ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/1
[       OK ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/1 (21 ms)
[ RUN      ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/2
[       OK ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/2 (38 ms)
[ RUN      ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/3
[       OK ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/3 (55 ms)
[ RUN      ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/4
[       OK ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/4 (70 ms)
[ RUN      ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/5
[       OK ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/5 (80 ms)
[ RUN      ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/6
[       OK ] PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture.CharacterGameplayController_EntityFallsUnderGravity/6 (104 ms)
[----------] 7 tests from PhysXDefaultWorldTest/PhysXDefaultWorldTestWithParamFixture (391 ms total)

[==========] 13 tests from 2 test cases ran. (983 ms total)
[  PASSED  ] 13 tests.
```


Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
```diff
@@ -98,7 +95,12 @@ namespace PhysX
             }
         }
 
-        
+        //! Function to add the ground for tests where the ground is needed
+        void AddGround(const AZ::Transform& floorTransform = DefaultFloorTransform)
```
hultonha, line 99: nit: It might be better to call this `SetGround` as calling this multiple times will override the same value (and not append multiple floors)

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
```diff
@@ -137,18 +139,90 @@ namespace PhysX
...
+            basis.Update();
+
+            if (basis.m_gameplayController->IsOnGround())
+            {
+                groundDetected = true;
+                break;
+            }
+        }
+
+        EXPECT_THAT(groundDetected, testing::AllOf(testing::IsTrue(), testing::Ne(startingDetected)));
```
hultonha, line 218: suggestion: Could we add some comments for this test, it's a little tricky to see exactly what's going on. Do we need the `testing::IsTrue` here anymore? (maybe that was from some testing?)

hultonha (commented):
Overall looks good but I do think a few more comments/context in the new tests would be super helpful to understand exactly what they're verifying. Great start!

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
FuzzyCarterAWS: Renamed to SetupGround and added better comments.

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
FuzzyCarterAWS: I'll update the comments for all the tests with this PR.

This is needed. The condition is making sure groundDetected is actually true while also checking that the starting state was not detected in the first place.  If for some reason the test got into a state where the starting detection was True and the end detection is False we would throw a false passing situation. 

This is the logical breakdown:

```
if (groundDetected) && (groundDetected != startingDetected):
     Pass
else:
    Fail
```

hultonha: I see understood, thank you for clarifying! I see `groundDetected` is being checked against all conditions in the `testing::AllOf` matcher (apologies I think I initially read it as equivalent to `if (true)`, but I see what it's doing now). Thanks!

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
```diff
@@ -98,7 +95,13 @@ namespace PhysX
             }
         }
 
-        
+        //! Add the ground tot he scene via the test basis for tests where the ground is needed.
```
hultonha, line 98: nit: tot he -> to the

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
```diff
@@ -114,41 +117,136 @@ namespace PhysX
...
+
+            // The end velocity should be decelerated by gt when g is gravity acceleration and t is the length of time
+            const float expectedDeceleration = gravity.GetZ() * basis.m_timeStep;
+
+            // Verify that the end velocity is the set velocity minus the deceleration for a single time step
+            EXPECT_THAT(endVelocity.GetZ(), testing::FloatNear(setVelocity.GetZ() + expectedDeceleration, 0.001f));
+        }
+    }
+
+    //! Test that verifies that the Ground Detec Set & Get methods function as expected.
```
hultonha, line 196: nit: Detec -> Detection?

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
```diff
@@ -114,41 +117,136 @@ namespace PhysX
...
+    TEST_F(PhysXDefaultWorldTest, CharacterGameplayController_SetGroundDetectionHeight)
+    {
+        GameplayTestBasis basis(m_testSceneHandle, DefaultGravityMultiplier, DefaultGroundDetectionBoxHeight);
+
+        const float expectedHeight = 10.0f;
+        const auto originalHeight = basis.m_gameplayController->GetGroundDetectionBoxHeight();
+
+        // Set the ground detection to the excpected heigh + the original height to ensure that it changes.
```
hultonha, line 204: nit: heigh -> height

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
```diff
@@ -114,41 +117,136 @@ namespace PhysX
...
+        basis.m_gameplayController->SetGroundDetectionBoxHeight(expectedHeight + originalHeight);
+
+        // Get the end height and verify that it is set to the expected value set.
+        auto endHeight = basis.m_gameplayController->GetGroundDetectionBoxHeight();
+        EXPECT_THAT(endHeight, testing::FloatNear(expectedHeight + originalHeight, 0.001f));
+    }
+
+    //! Test to determine if the ground will detect while moving.
```
hultonha, line 212: nit: (rephrase slightly) _Test to determine if the ground is detected while moving_.

hultonha (approved):
Thanks for all the extra details/comments @FuzzyCarterAWS. Change looks good! I just have a few minor nits on typos, otherwise all good 👍

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
```diff
@@ -114,41 +117,136 @@ namespace PhysX
...
+
+        // Get the original velocity so we can determine that set actually changes things.
+        auto originalVelocity = basis.m_gameplayController->GetFallingVelocity();
+
+        // Set the falling velocity to the expected + original velocity to verify that the velociy will change.
+        basis.m_gameplayController->SetFallingVelocity(originalVelocity + expectedVelocity);
+
+        // Get the end velocity and validate that it is the expected value.
+        auto endVelocity = basis.m_gameplayController->GetFallingVelocity();
+        EXPECT_THAT(endVelocity.GetZ(), testing::FloatNear(expectedVelocity.GetZ(), 0.001f));
```
greerdv, line 163: The falling velocity should be whatever you set it to (immediately after you set it - it will change when you update physics). I think this test can only be passing because the original falling velocity was 0 or very close to it. If you've never ticked physics, then the original velocity would be 0.

I think a better test would be setting the falling velocity, ticking physics once and verifying that the new position was consistent with the velocity value you set.

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
```diff
@@ -98,7 +95,13 @@ namespace PhysX
...
-        
+        //! Add the ground tot he scene via the test basis for tests where the ground is needed.
+        //! This function will only ever add one ground element to the scene.
+        void SetupGround(const AZ::Transform& floorTransform = DefaultFloorTransform)
+        {
+            m_floor = PhysX::TestUtils::AddStaticFloorToScene(m_sceneHandle, floorTransform);
+        }
+
         //! Function to remove the ground for tests where the ground can interfere with the intent of the test. 
         void RemoveGround()
```
greerdv, line 106: Are you using this any more? I think you can probably just not add the ground rather than add it and then remove it.

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
```diff
@@ -114,41 +117,136 @@ namespace PhysX
...
+    }
+
+    //! Test the verifies that the Falling Velocity sets and uses the set value to interact with the tick incrementing.
+    TEST_F(PhysXDefaultWorldTest, CharacterGameplayController_FallingVelocitySetsWhileMoving)
+    {
+        const AZ::Vector3 expectedVelocity(0.0f, 0.0f, 22.0f);
+        GameplayTestBasis basis(m_testSceneHandle, DefaultGravityMultiplier, DefaultGroundDetectionBoxHeight);
+
+        // Get the original velocity so we can determine that set actually changes things.
+        auto originalVelocity = basis.m_gameplayController->GetFallingVelocity();
```
greerdv, line 173: Again the original velocity is going to be zero because you haven't ticked physics yet. So adding it to things is not going to do anything, and doesn't really make any sense.

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
```diff
@@ -114,41 +117,136 @@ namespace PhysX
...
+            basis.m_gameplayController->SetFallingVelocity(setVelocity);
+
+            basis.Update();
+
+            // Get the endVelocity & current gravity to perform calculations on.
+            auto endVelocity = basis.m_gameplayController->GetFallingVelocity();
+            const auto gravity = basis.m_testScene->GetGravity();
+
+            // The end velocity should be decelerated by gt when g is gravity acceleration and t is the length of time
+            const float expectedDeceleration = gravity.GetZ() * basis.m_timeStep;
```
greerdv, line 189: Nitpick - I would call this something like expectedVelocityChange. The name expectedDeceleration implies that it is a rate of change of velocity, which is slightly different.

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
```diff
@@ -114,41 +117,136 @@ namespace PhysX
...
+            EXPECT_THAT(endVelocity.GetZ(), testing::FloatNear(setVelocity.GetZ() + expectedDeceleration, 0.001f));
+        }
+    }
+
+    //! Test that verifies that the Ground Detec Set & Get methods function as expected.
+    TEST_F(PhysXDefaultWorldTest, CharacterGameplayController_SetGroundDetectionHeight)
+    {
+        GameplayTestBasis basis(m_testSceneHandle, DefaultGravityMultiplier, DefaultGroundDetectionBoxHeight);
+
+        const float expectedHeight = 10.0f;
```
greerdv, line 201: Nitpick - doesn't really have any impact on this test, but this is an unrealistically large value for the ground detection box height (it's much taller than the character).

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
greerdv: You could just write

EXPECT_FALSE(startingDetected);
EXPECT_TRUE(groundDetected);

which I would say is much easier to read.

And you could just move the first expectation to line 224 and say

EXPECT_FALSE(basis.m_gameplayController->IsOnGround());

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
FuzzyCarterAWS: fixed

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
FuzzyCarterAWS: Fixed

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
FuzzyCarterAWS: Removed

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
FuzzyCarterAWS: updated

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
FuzzyCarterAWS: I was intentionally using a large number to make it obvious that the value was set.

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
FuzzyCarterAWS: After discussion with @greerdv, I've adjusted the logic here to follow his advice.

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
FuzzyCarterAWS: Reconfigured the logic here.

I was intending to use the starting velocity to shortcut setting the velocity to remove the starting velocity's default value as a failsafe for a false positive.

I've been more explicit about the testing of values at the end of the test.

As this test is trying to isolate whether or not the get & sets work outside of the physics tick modifying the velocity, I have not updated the physics tick and have another test that does this.

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
FuzzyCarterAWS: Removed

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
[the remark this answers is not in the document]
FuzzyCarterAWS: Fixed.

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
greerdv: OK, that makes sense with regard to not ticking the physics.

If you're worried about a false positive, in this case the original velocity should be 0, because physics has never ticked, so you could also just check that the original velocity was 0. The way you were doing things before just complicates the test, and before you explained why you were doing it, as someone reading the code I had no idea what the reason was for that complication. In general, I would say it's always better to aim for more explicitness and simplicity.

I would also argue that the scenario you're trying to guard against (that SetFallingVelocity is broken and GetVelocity just happens to return 22 by accident) is so unlikely that it's not worth testing for.

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
```diff
@@ -186,42 +179,44 @@ namespace PhysX
...
     {
         GameplayTestBasis basis(m_testSceneHandle, DefaultGravityMultiplier, DefaultGroundDetectionBoxHeight);
 
-        const float expectedHeight = 10.0f;
+        const float expectedHeight = 1.2f;
         const auto originalHeight = basis.m_gameplayController->GetGroundDetectionBoxHeight();
 
-        // Set the ground detection to the excpected heigh + the original height to ensure that it changes.
-        basis.m_gameplayController->SetGroundDetectionBoxHeight(expectedHeight + originalHeight);
+        // Set the ground detection to the expected height + the original height to ensure that it changes.
```
greerdv, line 197: Think this comment is out of date now.

FuzzyCarterAWS: fixed

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
FuzzyCarterAWS: removed that check & unused originalVelocity variable.

Gems/PhysX/Code/Tests/CharacterGameplayControllerTests.cpp
FuzzyCarterAWS: Updated

greerdv (approved):

PR #14756, opened 2023-02-23
Multiplayer Simple Player Spawner: Expose Spawn Points

AMZN-Gene:
- Exposing Simple Player Spawner methods for [games to utilize](https://github.com/o3de/o3de-multiplayersample/issues/210)
- Added unit tests

## How was this PR tested?
Run Multiplayer Sample using the new API and see that players still spawn and respawn in the correct positions
Ran Multiplayer gem unit tests


Gems/Multiplayer/Code/Source/Components/SimplePlayerSpawnerComponent.cpp
```diff
@@ -72,36 +74,47 @@ namespace Multiplayer
...
+        m_spawnIndex = ++m_spawnIndex % m_spawnPoints.size();
+        return nextSpawnPoint;
+    }
 
+    AZ::Transform SimplePlayerSpawnerComponent::GetNextSpawnPoint() const
+    {
         if (m_spawnPoints.empty())
         {
-            AZLOG_WARN("SimplePlayerSpawnerComponent is missing spawn points. Spawning new player at the origin.")
+            AZLOG_WARN("SimplePlayerSpawnerComponent is missing spawn points. Returning spawn point at the world's origin.")
```
mbalfour-amzn, line 88:
```suggestion
            AZLOG_WARN("SimplePlayerSpawnerComponent is missing spawn points. Returning spawn point at the world's origin.");
```

Gems/Multiplayer/Code/Source/Components/SimplePlayerSpawnerComponent.cpp
```diff
@@ -72,36 +74,47 @@ namespace Multiplayer
...
 
+        AZ::Entity* spawnPointEntity = nullptr;
+        AZ::ComponentApplicationBus::BroadcastResult(
+            spawnPointEntity, &AZ::ComponentApplicationBus::Events::FindEntity, spawnPointEntityId);
+        if (spawnPointEntity != nullptr)
+        {
+            return spawnPointEntity->GetTransform()->GetWorldTM();
+        }
+
+        AZLOG_WARN("The spawn point entity id for index %i is invalid. Returning spawn point at the world's origin.", m_spawnIndex)
```
mbalfour-amzn, line 108:
```suggestion
        AZLOG_WARN("The spawn point entity id for index %i is invalid. Returning spawn point at the world's origin.", m_spawnIndex);
```

Gems/Multiplayer/Code/Include/Multiplayer/Components/SimplePlayerSpawnerComponentBus.h
```diff
@@ -0,0 +1,36 @@
...
+{
+    //! @class SimplePlayerSpawnerRequests
+    //! @brief The SimplePlayerSpawnerRequest event-bus exposes helper methods regarding network player spawners.
+    //! For example, it's common for game servers want information regarding valid spawn locations in order to reset a player's position after the player dies or the round restarts (respawning).
+    class SimplePlayerSpawnerRequests : public AZ::EBusTraits
+    {
+    public:
+        static const AZ::EBusHandlerPolicy HandlerPolicy = AZ::EBusHandlerPolicy::Single;
+        static const AZ::EBusAddressPolicy AddressPolicy = AZ::EBusAddressPolicy::Single;
+
```
mbalfour-amzn, line 24: This is going on a component, but doesn't use Entity Id as the address? What happens when the component is added multiple times?

From reading the description of the component, maybe it should be turned into a level component so that only one can ever be added?

Gems/Multiplayer/Code/Include/Multiplayer/Components/SimplePlayerSpawnerComponentBus.h
```diff
@@ -0,0 +1,36 @@
...
+        
+        //! Visits and returns the next player spawn point, causing the current spawn point index to increment.
+        //! Method is only valid if called from the multiplayer host/authority; clients are not given information regarding spawn points.
+        //! @return AZ::Transform The location of the next spawn point
+        virtual AZ::Transform VisitNextSpawnPoint() = 0;
+
+        //! Returns the next player spawn point. Unlike VisitNextSpawnPoint, this will not cause the current spawn point index to increment.
+        //! Method is only valid if called from the multiplayer host/authority; clients are not given information regarding spawn points.
+        //! @return AZ::Transform The location of the next spawn point
+        virtual AZ::Transform GetNextSpawnPoint() const = 0;
```
mbalfour-amzn, line 33: Just curious about the API here. Is there any particular reason for exposing it this way, vs something like:

int GetNumSpawnPoints()
Transform GetSpawnPoint(int index)
int GetCurrentSpawnPointIndex();
void SetCurrentSpawnPointIndex(int index);
// And maybe some APIs for adding / removing / changing spawn point locations too?

The API just surprised me on first read, since there's nothing that exposes the set of spawn points as a whole, and the naming difference between "VisitNextSpawnPoint" vs "GetNextSpawnPoint" is a little bit subtle.

I'm not entirely sure what the intended uses are for the APIs as written, it seems like they're intended to maybe enable spawning something other than players? If so, maybe the API should be SpawnAsset(AssetId) for spawning something other than players, or SpawnPlayer(xxx), or something?

Gems/Multiplayer/Code/Include/Multiplayer/Components/SimplePlayerSpawnerComponentBus.h
AMZN-Gene: Yep! This is a level component

Gems/Multiplayer/Code/Include/Multiplayer/Components/SimplePlayerSpawnerComponentBus.h
AMZN-Gene: The API and component is inspired by the Multiplayer Sample game for respawning. The MPS custom spawner calls a "VisitNextSpawnPoint" method (actually just confusingly called GetNextSpawnPoint but it updates the spawn index internally).

AMZN-Olex: Visit* is a bit of a curious name. It doesn't make it clear enough that something is modified behind the scenes. Maybe something along the lines of Get and Increment? Or something. Hmm.

Gems/Multiplayer/Code/Include/Multiplayer/Components/SimplePlayerSpawnerComponentBus.h
mbalfour-amzn: Oh whoops, missed that. ty! :)

Gems/Multiplayer/Code/Include/Multiplayer/Components/SimplePlayerSpawnerComponentBus.h
mbalfour-amzn: Hm, what about:
```
AZ::Transform AdvanceSpawnPoint(); // advances the spawn point index then returns the new spawn point
AZ::Transform GetCurrentSpawnPoint() const; // gets the current spawn point that pointed at
```

AMZN-Gene: Oops, just saw your naming suggestions. I changed the name to GetNextSpawnPoint vs RoundRobinNextSpawnPoint. Does that make it more distinguished? 

I like using "next" vs "current" since spawning refers to a future action, as opposed to an ongoing action.

Gems/Multiplayer/Code/Include/Multiplayer/Components/SimplePlayerSpawnerComponentBus.h
```diff
@@ -0,0 +1,56 @@
...
+    {
+    public:
+        static const AZ::EBusHandlerPolicy HandlerPolicy = AZ::EBusHandlerPolicy::Single;
+        static const AZ::EBusAddressPolicy AddressPolicy = AZ::EBusAddressPolicy::Single;
+        
+        //! Returns the location where the next joining player will be spawned, and then advances the spawn point index.
+        //! The index will wrap back to 0 in a round-robin fashion so that a valid spawn point is always available.
+        //! Method is only valid if called from the multiplayer host/authority; clients are not given information regarding the spawn point index.
+        //! @return AZ::Transform The location of the next spawn point
+        virtual AZ::Transform RoundRobinNextSpawnPoint() = 0;
```
mbalfour-amzn, line 30: Suggestion: I'd still suggest "Advance" over "RoundRobin" but it depends on your goals for the API. 

My thinking is that the actual selection algorithm would ideally be data-driven, and not "baked" into the different API names. For example, you could imagine having a dropdown on the component that says "RoundRobin" / "Random" / "Farthest" / "Nearest". In that case, you'd want the API to use whatever was on the dropdown.

However, I could see an alternate argument where the dropdown could contain the _default_ and you want the API to override it. Though even in that case, I'd think the API would take in an enum of algorithms instead of having a different call per algorithm.

A third point to consider - even if you _don't_ want multiple algorithms, by exposing the choice of algorithm in the name of the API, it means that if you ever want to change which algorithm this component uses, the API name will need to change.

Gems/Multiplayer/Code/Include/Multiplayer/Components/SimplePlayerSpawnerComponentBus.h
```diff
@@ -0,0 +1,56 @@
...
+        //! Returns the location where the next joining player will be spawned. Unlike RoundRobinNextSpawnPoint(), this will not cause the current spawn point index to increment.
+        //! Method is only valid if called from the multiplayer host/authority; clients are not given information regarding the spawn point index.
+        //! @return AZ::Transform The location of the next spawn point
+        virtual AZ::Transform GetNextSpawnPoint() const = 0;
+
+
+        //! Returns a mutable list of all the spawn points.
+        //! Only access/edit this list on the multiplayer host; spawn points are not synced across the network and the host is responsible for spawning players.
+        //! @return AZStd::vector<AZ::EntityId> List of spawn points.
+        virtual AZStd::vector<AZ::EntityId>& GetSpawnPoints() = 0;
```
mbalfour-amzn, line 41: Suggestion: This should probably be an _immutable_ list. It's a bit dangerous to expose the internal storage of the component directly outwards through the API. For example, suppose the component needed to keep a second data structure in sync with this list. Like maybe a "num times used" counter per spawn point, to support a "semi-random" selection that ensures that every point is used once in a rotation before using it the next time. If the list itself is exposed outwards, there's no way to build a second data structure like that and keep it in sync.

Gems/Multiplayer/Code/Include/Multiplayer/Components/SimplePlayerSpawnerComponentBus.h
```diff
@@ -0,0 +1,56 @@
...
+        //! @return Number of spawn points.
+        virtual uint32_t GetSpawnPointCount() const = 0;
+
+        //! Returns the spawn point index where the next joining player will be spawned.
+        //! @return AZ::Outcome<uint32_t, AZStd::string> If success, the value present a valid index that can be used to lookup into the spawn points array.
+        virtual AZ::Outcome<uint32_t, AZStd::string> GetNextSpawnPointIndex() const = 0;
+
+        //! Overwrites the next joining player's spawn index. The spawn index provided must be a valid (in-bounds) index into the array of available spawn points.
+        //! @return AZ::Outcome<void, AZStd::string> Success means a valid index was provided and the next spawn point was updated; failure signifies a bad (out-of-bounds) index.
+        virtual AZ::Outcome<void, AZStd::string> SetNextSpawnPointIndex(uint32_t index) = 0;
```
mbalfour-amzn, line 53: Suggestion: IMO, "Outcome" is a pretty heavy-handed choice for a runtime API, especially since it's constructing strings for failure conditions in this particular case. If success/failure needs to be conveyed, maybe consider an optional or just a bool, but I'd question the overall need of conveying a result on operations as simple as these.

Alternatively, if you're committed to providing information on success/failure, then I'd question why only these two APIs get that, and not _all_ of them...

Gems/Multiplayer/Code/Source/Components/SimplePlayerSpawnerComponent.cpp
```diff
@@ -72,36 +74,74 @@ namespace Multiplayer
...
+        {
+            return AZ::Transform::Identity();
+        }
+
+        const AZ::EntityId spawnPointEntityId = m_spawnPoints[m_spawnIndex];
+        const AZ::Entity* spawnPointEntity = AZ::Interface<AZ::ComponentApplicationRequests>::Get()->FindEntity(spawnPointEntityId);
+        if (spawnPointEntity != nullptr)
+        {
+            return spawnPointEntity->GetTransform()->GetWorldTM();
+        }
```
mbalfour-amzn, line 96: Just curious, is this better in some way than using the Transform bus?

Gems/Multiplayer/Code/Tests/SimplePlayerSpawnerTests.cpp
```diff
@@ -0,0 +1,129 @@
...
+        EXPECT_EQ(nextSpawnPoint.GetTranslation().GetX(), 3);
+
+        // Ran through each spawn point; make sure the spawner goes back to spawn point #1
+        SimplePlayerSpawnerRequestBus::BroadcastResult(nextSpawnPoint, &SimplePlayerSpawnerRequests::RoundRobinNextSpawnPoint);
+        EXPECT_EQ(nextSpawnPoint.GetTranslation().GetX(), 1);
+
+        AZ::Outcome<uint32_t, AZStd::string> spawnPointIndexOutcome;
+        SimplePlayerSpawnerRequestBus::BroadcastResult(spawnPointIndexOutcome, &SimplePlayerSpawnerRequests::GetNextSpawnPointIndex);
+        EXPECT_TRUE(spawnPointIndexOutcome.IsSuccess());
+        EXPECT_EQ(spawnPointIndexOutcome.GetValue(), 1); // Spawn point #2 (aka index 1)
```
mbalfour-amzn, line 107: Nit: It's _really_ not important, but consider making the constants floats (1.0f, 2.0f, etc) since that's what you're actually comparing here.

Gems/Multiplayer/Code/Include/Multiplayer/Components/SimplePlayerSpawnerComponentBus.h
AMZN-Gene: Only Get/Set index warrant an outcome since they are the only ones that can fail. For example, GetIndex might return an out of bounds index if the array is empty (using the index in a lookup would crash), where as even if the array is empty GetNextSpawnPoint will return the actual position the next player will spawn (the worlds origin).
In either case, if the gut is that outcome is too heavy, I’ll remove it.

mbalfour-amzn: Totally up to you if you want to leave or remove it, it's just a suggestion. It's not like these will get called a ton, it just struck me as odd when looking at it. From my point of view, if the API is being called from code only, I'd expect failures here to be asserts because it would mean I'm using the API wrong. If the API is being called from script, or we just want to make it friendlier to begin with, I can see making it handle the errors gracefully (which it already does) and just print AZ_ERROR. The "heavy-handed" question for me is that it's forcing the API user into some level of runtime error handling by needing to check for success/failure. To put it another way, I'm not sure what the API user _should_ do in failure conditions beyond what the API internally itself is already doing other than rewrite their code to use the API correctly.

Gems/Multiplayer/Code/Include/Multiplayer/Components/SimplePlayerSpawnerComponentBus.h
AMZN-Gene: Changed to "Advance" thanks!

Gems/Multiplayer/Code/Include/Multiplayer/Components/SimplePlayerSpawnerComponentBus.h
AMZN-Gene: Good call! Fixed!

Gems/Multiplayer/Code/Include/Multiplayer/Components/SimplePlayerSpawnerComponentBus.h
AMZN-Gene: Fixed! I removed the outcomes. Yeah, ultimately I want this to be user friendly. I appreciate the suggestions.

Gems/Multiplayer/Code/Tests/SimplePlayerSpawnerTests.cpp
AMZN-Gene: good call! especially easier on the eyes to differentiate the other numbers like when checking indices. Much clearer to read. Thanks!

Gems/Multiplayer/Code/Source/Components/SimplePlayerSpawnerComponent.cpp
AMZN-Gene: Nope, I just derped >.<
Updated to just use the ebus. Thank you, good catch.

Gems/Multiplayer/Code/Include/Multiplayer/Components/SimplePlayerSpawnerComponentBus.h
AMZN-Gene: Very true, now with this wording this API can be generic and used for other spawner class. I wonder if I should rename this to just _PlayerSpawnerBus.h_?

Gems/Multiplayer/Code/Include/Multiplayer/Components/SimplePlayerSpawnerComponentBus.h
```diff
@@ -0,0 +1,55 @@
...
+
+namespace Multiplayer
+{
+    //! @class SimplePlayerSpawnerRequests
+    //! @brief The SimplePlayerSpawnerRequest event-bus exposes helper methods regarding network player spawners.
+    //! Although the Multiplayer System automatically spawns in players, it's common for game specific server logic to retrieve valid spawn locations when respawning a player.
+    class SimplePlayerSpawnerRequests : public AZ::EBusTraits
+    {
+    public:
+        static const AZ::EBusHandlerPolicy HandlerPolicy = AZ::EBusHandlerPolicy::Single;
```
kberg-amzn, line 23: Single/Single kind of cries out for an Interface.. This is also a little extra weird since it's already inheriting from the IMultiplayerSpawner interface, and all your use-cases appear to be native code.

Gems/Multiplayer/Code/Include/Multiplayer/Components/SimplePlayerSpawnerComponentBus.h
```diff
@@ -0,0 +1,55 @@
...
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#pragma once
+
+#include <AzCore/EBus/EBus.h>
+#include <AzCore/Outcome/Outcome.h>
```
kberg-amzn, line 12: Outcome is not used, remove it to keep compile times down. Remember there's a bit push on getting our compile-times under control right now.

Gems/Multiplayer/Code/Include/Multiplayer/Components/SimplePlayerSpawnerComponentBus.h
AMZN-Gene: I was thinking that we didn't want to lump in any assumptions about iMultiplayerSpawner besides the responsibility of spawning a player OnPlayerJoin. The original custom code for MPS used a separate IMultiplayerSpawner and IPlayerSpawner (for having spawn points and round-robin indexing); this carries the tradition of separation.

Gems/Multiplayer/Code/Include/Multiplayer/Components/SimplePlayerSpawnerComponentBus.h
AMZN-Gene: Thank you!

Gems/Multiplayer/Code/Include/Multiplayer/Components/SimplePlayerSpawnerComponentBus.h
AMZN-Gene: Fixed! Is now an AZ::Interface.

Gems/Multiplayer/Code/Include/Multiplayer/Components/SimplePlayerSpawnerComponentBus.h
AMZN-Gene: To follow up, I actually just removed the "advance" option. Avoids added complexity, and a game can simply Get/Set the index to their liking.

Gems/Multiplayer/Code/Include/Multiplayer/Components/ISimplePlayerSpawner.h
```diff
@@ -0,0 +1,46 @@
...
+    //! @brief The SimplePlayerSpawnerRequest event-bus exposes helper methods regarding network player spawners.
+    //! Although the Multiplayer System automatically spawns in players, it's common for game specific server logic to retrieve valid spawn locations when respawning a player.
+    class ISimplePlayerSpawner
+    {
+    public:
+        AZ_RTTI(ISimplePlayerSpawner, "{6DAC5CDE-5D63-4C0B-9F7B-9F5B09079DBB}");
+
+        virtual ~ISimplePlayerSpawner() = default;
+
+        //! Returns the location where the next joining player will be spawned. Unlike RoundRobinNextSpawnPoint(), this will not cause the current spawn point index to increment.
```
mbalfour-amzn, line 25: Nit: There is no RoundRobinNextSpawnPoint() any longer.

Gems/Multiplayer/Code/Source/Components/SimplePlayerSpawnerComponent.cpp
```diff
@@ -72,35 +74,69 @@ namespace Multiplayer
...
-            if (spawnPointEntity != nullptr)
-            {
-                transform = spawnPointEntity->GetTransform()->GetWorldTM();
-            }
-            else
-            {
-                AZLOG_WARN("The spawn point entity id for index %i is invalid. Spawning new player at the origin.", m_spawnIndex)
-            }
+            AZ_Assert(false, "SimplePlayerSpawnerComponent has an out-of-bounds m_spawnIndex %i. Please ensure spawn index is always valid.", m_spawnIndex);
+            return static_cast<uint32_t>(-1);
```
mbalfour-amzn, line 116: Suggestion: I think for an empty list it would be more reasonable to return an index of 0 instead of 4 billion. 0 would still be easily detectable as invalid because it would be >= GetSpawnPointCount().

mbalfour-amzn (approved):

Gems/Multiplayer/Code/Source/Components/SimplePlayerSpawnerComponent.cpp
```diff
@@ -72,35 +74,69 @@ namespace Multiplayer
...
+        m_spawnIndex = index;
+    }
+
+    NetworkEntityHandle SimplePlayerSpawnerComponent::OnPlayerJoin(
+        [[maybe_unused]] uint64_t userId, [[maybe_unused]] const MultiplayerAgentDatum& agentDatum)
+    {
+        const PrefabEntityId prefabEntityId(AZ::Name(m_playerSpawnable.m_spawnableAsset.GetHint().c_str()));
+
+        const AZ::Transform transform = GetNextSpawnPoint();
+        m_spawnIndex = ++m_spawnIndex % m_spawnPoints.size();
```
lmbr-pip, line 139: nit: Do we think this is thread safe? I assume the worst that can happen is multiple players spawning on same point?

AMZN-Gene: Yeah, I guess players would spawn in on the same point. We're spawning on the main thread, but yeah, maybe it's a problem if we move handling network traffic to a new thread.

Gems/Multiplayer/Code/Source/Components/SimplePlayerSpawnerComponent.cpp
AMZN-Gene: Updated!

Gems/Multiplayer/Code/Include/Multiplayer/Components/ISimplePlayerSpawner.h
AMZN-Gene: Thank you!

Gems/Multiplayer/Code/Source/Components/SimplePlayerSpawnerComponent.cpp
```diff
@@ -72,35 +74,69 @@ namespace Multiplayer
...
-            if (spawnPointEntity != nullptr)
-            {
-                transform = spawnPointEntity->GetTransform()->GetWorldTM();
-            }
-            else
-            {
-                AZLOG_WARN("The spawn point entity id for index %i is invalid. Spawning new player at the origin.", m_spawnIndex)
-            }
+            AZ_Assert(false, "SimplePlayerSpawnerComponent has an out-of-bounds m_spawnIndex %i. Please ensure spawn index is always valid.", m_spawnIndex);
+            return static_cast<uint32_t>(-1);
```
lmbr-pip, line 116: Why not return zero?

lmbr-pip (approved):

Issue #13764, opened 2022-12-19
StandardPBR Emissive Not Compatible with Non-Opaque surfaces (Blended, TintedTransperancy)


AMZN-Gene:
**Describe the bug**
Adding an emissive texture to a material appears to have no affect.

**Steps to reproduce**
1. Create a new _standard pbr_ material
2. Enable the _emissive_ property
3. Make sure the base color isn't contributing to the color. Turn down the base color factor to 0.0, or change the base color to black <0,0,0>, or remove the base texture.
4. Assign an emissive texture

**Expected behavior**
The emissive texture will light the material

**Actual behavior**
The object is black

**Screenshots/Video**
![image](https://user-images.githubusercontent.com/32776221/208370326-9ffcedd2-4b7b-4f6d-a799-05ee2d479cef.png)

**Found in Branch**
Development 

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
47eea0c820a3639fcbcb04aff991235e9132adb1 
Monday Dec 12 2022

gadams3:
Do you have transparency enabled for this material? That is my assumption because depth is not being respected between the model and the grid. I think there was a known issue with emissive not working on transparent materials. Can you confirm if the emissive settings work if your material is set to opaque?

HogJonny-AMZN:
I have verified ...
1. emissive works fine on a opaque material, including cutout (clip)
2. emissive is non-functional with Blended and TintedTransperancy

invertednormal:
This was fixed in #14758

PR #14758, opened 2023-02-23
Fixing emissive to work with alpha. 

invertednormal:
## What does this PR do?

Moved the emissive contribution out of LightingData and into the individual VertexAndPixel shaders (its usage was inconsistent before). Emissive is now added onto diffuse instead of specular. 

image

Added a new slider to the emissive section of materials to control how much the alpha should affect the emissive contribution.

image

## How was this PR tested?

Tested with atom sample viewer (in progess) and custom materials in the material editor.

### DIFF
--- a/Gems/Atom/Feature/Common/Assets/Materials/Pipelines/Common/TransparentPassVertexAndPixel.azsli
+++ b/Gems/Atom/Feature/Common/Assets/Materials/Pipelines/Common/TransparentPassVertexAndPixel.azsli
@@ -45,6 +45,8 @@ ForwardPassOutput PixelShader(VsOutput IN, bool isFrontFace : SV_IsFrontFace)
     float3 diffuse = lightingData.diffuseLighting.rgb;
     diffuse *= surface.alpha;
     
+    diffuse += lightingData.emissiveLighting;
+
     // m_opacityAffectsSpecularFactor controls how much the alpha masks out specular contribution.
     float3 specular = lightingData.specularLighting.rgb;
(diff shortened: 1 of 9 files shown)


HogJonny-AMZN (approved):

gadams3 (approved):
Awesome! Wondering if “emissive affected by alpha” should be an option instead. Should it be premultiplied by intensity instead of being passed into the emissive input function? We need to expose emissive intensity on the material canvas standard PBR node as well. Would either want to expose the same flag there or force the user to manually multiply alpha by whatever they're using for intensity.

invertednormal:
> Awesome! Wondering if “emissive affected by alpha” should be an option instead. Should it be premultiplied by intensity instead of being passed into the emissive input function? We need to expose emissive intensity on the material canvas standard PBR node as well. Would either want to expose the same flag there or force the user to manually multiply alpha by whatever they're using for intensity.

I did it this way for flexibility. You could make it a boolean, or you could just make it a 0-1 float and allow people any combination. Performance wise, it's a lerp, so almost nothing.

As far as premultiply, it makes no difference to the output. I lean towards making it function arguments to avoid everyone having to do it themselves. The number of args is getting kinda crazy though and wonder if a descriptor would be better. The equation is just emissive * intensity * lerp(1, alpha, affectedByAlpha). You could do that in any order, but multiplying by intensity first might be a good idea to help the compiler realize it doesn't need to multiply against every channel individually.

dmcdiarmid-ly (approved):

invertednormal:
Fixes #13764

Issue #14722, opened 2023-02-22
No pop-up message when taking a screenshot


LB-KacperMetelski:
**Describe the bug**
When taking a screenshot, there is no output in the console or a pop-up message related to this action, despite the screenshot being taken.

This issue occurs on both Linux and Windows. However, on Windows there is a console window displayed for a split second.

**Steps to reproduce**
1. Open the Material/Script Canvas.
2. Capture a screenshot.

**Expected behavior**
A pop-up message informing about the screenshot is displayed.

**Actual behavior**
No indication related to the screenshot being captured.

**Screenshots/Video**
https://user-images.githubusercontent.com/110163473/220658575-1f49b231-9814-4b86-ab6a-774c1851f14c.mp4

**Found in Branch**
Development

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
9c57ede

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Windows
 - Version: 11 22H2
 - CPU: I7-11700f
 - GPU: Geforce RTX3070 MSI
 - Memory: 32GB

monroegm:
Note from SIG-Content Triage: It doesn't necessarily need to be a pop up, but we need some sort of visual feedback. @o3de/sig-ui-ux please advise

bhanuja-s:
I agree, I don't want to disrupt the user's task – we can simply show them a Toast. We have guidelines on O3DE Docs: https://www.o3de.org/docs/tools-ui/ux-patterns/error/components/toasts/ on how to use them; they are a more passive and non-disruptive manner of notifying user. If absolutely necessary, you can use a Toast with Link which is a persistent notification, and the user has to manually dismiss it – it also gives the user a chance to view their screenshot.

> Screenshot saved. View your screenshot here.

PR #14764, opened 2023-02-23
Add a toast notification for Script and Material canvas

amzn-ahmadkrm:
## What does this PR do? 

This PR adds a toast notification when a user screenshots using the screenshot button. This addresses https://github.com/o3de/o3de/issues/14722. The issue also mentioned a window that would display for a tick, but this change seems to stop that aswell.

https://user-images.githubusercontent.com/105638312/220892649-639ba050-09bb-475f-9ef7-b6ee8a89ddd0.mp4

## How was this PR tested?
The toast notifications were tested in script and material canvas.


gadams3 (approved):

gadams3:
Is creating these kind of notifications accessible from outside graphics view using a bus?

amzn-ahmadkrm:
> Is creating these notifications accessible from outside graphics view using a bus?

Yes these are creatable with the `ToastRequestBus` but GraphCanvas here is overriding the bus calls so I'm just using the local functions.

PR #14784, opened 2023-02-23
Add a new test fixture and a couple of tests for prefab inspector overrides work

srikappa-amzn:
## What does this PR do?

Add a new test fixture and a couple of tests for prefab inspector overrides work.

RFC: https://github.com/o3de/sig-content/issues/112

## How was this PR tested?

Ran the new tests successfully.


Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DPEComponentAdapter.cpp
```diff
@@ -78,9 +78,9 @@ namespace AZ::DocumentPropertyEditor
         AzToolsFramework::PropertyEditorEntityChangeNotificationBus::MultiHandler::BusConnect(m_entityId);
         AzToolsFramework::ToolsApplicationEvents::Bus::Handler::BusConnect();
         AzToolsFramework::PropertyEditorGUIMessages::Bus::Handler::BusConnect();
+        m_componentAlias = componentInstance->GetSerializedIdentifier();
```
srikappa-amzn, line 81: moving this before 'SetValue' because 'SetValue' eventually calls DPEComponentAdapter::OnBeginRow() and it looks for the alias in that function. If it isn't set by then, the paths forwarded to the PrefabAdapter won't have the component alias in them.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/DocumentPropertyEditor/PrefabAdapter.cpp
```diff
@@ -44,7 +44,6 @@ namespace AzToolsFramework::Prefab
         if (m_prefabOverridePublicInterface->AreOverridesPresent(entityId, relativePathFromEntity.ToString()))
         {
             adapterBuilder->BeginPropertyEditor<PrefabPropertyEditorNodes::PrefabOverrideProperty>();
-            adapterBuilder->Attribute(AZ::DocumentPropertyEditor::Nodes::PropertyEditor::SharePriorColumn, true);
```
srikappa-amzn, line 47: Removed this as it was throwing errors if the icon is the first thing put in a row. This can be added at a later point if we want this to come after a certain property/label

Code/Framework/AzToolsFramework/AzToolsFramework/API/EntityPropertyEditorRequestsBus.h
```diff
@@ -31,6 +33,10 @@ namespace AzToolsFramework
         //! Explicitly sets a component as having been the most recently added.
         //! This means that the next time the UI refreshes, that component will be ensured to be visible.
         virtual void SetNewComponentId(AZ::ComponentId componentId) = 0;
+
+        //! Provides access to the component editors in an EntityPropertyEditor via a callback.
+        //! @param callback The callback that iterates over all the component editors within an EntityPropertyEditor.
+        virtual void GetComponentEditors(const GetComponentEditorsCallback& callback) const = 0;
```
hultonha, line 39: suggestion: Should this potentially be called something like `Process` or `Update` rather than `Get` (as no value is being returned and we're iterating and potentially modifying the ComponentEditors?)

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/ComponentEditor.hxx
```diff
@@ -97,6 +99,10 @@ namespace AzToolsFramework
         AZStd::vector<AZ::Component*>& GetComponents();
         const AZStd::vector<AZ::Component*>& GetComponents() const;
 
+        //! Gets the contents of the DPEComponentAdapter used by this ComponentEditor.
+        //! @param callback The callback to use to retrieve the DPEComponentAdapter contents.
+        void GetComponentAdapterContents(const GetComponentAdapterContentsCallback& callback) const;
```
hultonha, line 104: nit: Same suggestion as before about the naming

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/EntityPropertyEditor.cpp
```diff
@@ -742,6 +742,22 @@ namespace AzToolsFramework
...
+        for (const ComponentEditor* componentEditor : m_componentEditors)
+        {
+            if (!componentEditor)
+            {
+                continue;
+            }
+
+            if (!callback(componentEditor))
+            {
+                return;
```
hultonha, line 756: question: Just was curious what this early out was for, is this so all further processing of component editors can stop? What might this be useful? Also is it worth returning this value to the outer function? Just curious if the result might be useful to know/detect.

Code/Framework/AzToolsFramework/Tests/Prefab/Overrides/PrefabInspectorOverrideTests.cpp
```diff
@@ -0,0 +1,83 @@
...
+            EXPECT_EQ(
+                translateRow[0][AZ::DocumentPropertyEditor::Nodes::PropertyEditor::Type.GetName()].GetString(),
+                AzToolsFramework::Prefab::PrefabPropertyEditorNodes::PrefabOverrideProperty::Name);
+
+            // Then comes the label of the row.
+            EXPECT_EQ(translateRow[1][AZ::DocumentPropertyEditor::Nodes::Label::Value.GetName()].GetString(), "Translate");
+
+            // Then comes the actual property editor of the translate vector.
+            AZ::Dom::Value translateVectorInDom = translateRow[2][AZ::DocumentPropertyEditor::Nodes::PropertyEditor::Value.GetName()];
+            EXPECT_EQ(translateVectorInDom[0].GetDouble(), 10.0);
```
hultonha, line 51: suggestion: There is definitely a balance with this but sometimes having fewer expectations/assertions per test can make failures a lot clearer (ideally we want there to be one reason and one reason only for a test to fail). This is a guideline and not a rule and doing what you've done here might make sense but now there are a few different ways the test could fail and it may make diagnosing failures more difficult. I just wanted to mention it for reference as sometimes having a few more tests and refactoring to avoid some duplication can be really useful long-term.

hultonha (approved):
Really great to see more of these tests 🙂 Awesome work @srikappa-amzn, I only have a few pretty minor comments/suggestions, nothing blocking.

Code/Framework/AzToolsFramework/Tests/Prefab/Overrides/PrefabInspectorOverrideTestFixture.h
```diff
@@ -0,0 +1,38 @@
...
+ *
+ */
+
+#pragma once
+
+#include <AzCore/std/smart_ptr/unique_ptr.h>
+#include <AzToolsFramework/Prefab/Overrides/PrefabOverridePublicInterface.h>
+#include <AzToolsFramework/Prefab/PrefabFocusPublicInterface.h>
+#include <Prefab/PrefabTestFixture.h>
+#include <Prefab/Overrides/PrefabOverrideTestFixture.h>
```
forkercat, line 15: nit: include order

Code/Framework/AzToolsFramework/Tests/Prefab/Overrides/PrefabInspectorOverrideTestFixture.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include <Prefab/Overrides/PrefabInspectorOverrideTestFixture.h>
+
+namespace UnitTest
+{
+    void PrefabInspectorOverrideTestFixture::SetUpEditorFixtureImpl()
+    {
+        // Enable feature flags for DPE and InspectorOverrideManagement
+        AZ::SettingsRegistryInterface* registry = AZ::SettingsRegistry::Get();
+        registry->Set("/O3DE/Preferences/Prefabs/EnableInspectorOverrideManagement", true);
+        registry->Set("/O3DE/Autoexec/ConsoleCommands/ed_enableDPE", true);
```
forkercat, line 20: Can `EnableInspectorOverrideManagement` be imported from PrefabEditorPreferences if we make it public?

Code/Framework/AzToolsFramework/Tests/Prefab/Overrides/PrefabInspectorOverrideTestFixture.cpp
```diff
@@ -0,0 +1,57 @@
...
+    {
+        m_testEntityPropertyEditor.reset();
+        PrefabOverrideTestFixture::TearDownEditorFixtureImpl();
+    }
+
+    void PrefabInspectorOverrideTestFixture::GenerateComponentAdapterDoms(AZ::EntityId entityId)
+    {
+        // Calling UpdateContents will trigger the ReflectionAdapters of the components to build the DPE DOMs.
+        AzToolsFramework::ToolsApplicationRequestBus::Broadcast(
+            &AzToolsFramework::ToolsApplicationRequests::SetSelectedEntities, AzToolsFramework::EntityIdList{ entityId });
```
forkercat, line 37: Is `AzToolsFramework::EntityIdList { entityId }` needed? I think this can be just `{ entityId }`.

Code/Framework/AzToolsFramework/Tests/Prefab/Overrides/PrefabInspectorOverrideTests.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+namespace UnitTest
+{
+    using PrefabInspectorOverrideTest = PrefabInspectorOverrideTestFixture;
+
+    // These paths depends on multiple factors like the data in the component, how its reflected to serialize and edit contexts,
+    // how different DPE adapters likes ReflectionAdapter and PrefabAdapter construct the DPE DOM etc. Therefore, these may
+    // change in the future if the data stored in the DPE DOM itself changes and need to be modified accordingly to prevent test failures.
+    constexpr AZStd::string_view domPathToTranslateProperty = "/0/3/2";
+    constexpr AZStd::string_view domPathToOverriddenTranslateProperty = "/0/3/3";
```
forkercat, line 21: I think we can add a comment about the what the path looks like for each DOM path, e.g. `/Transform Data/Translate/0`.

Code/Framework/AzToolsFramework/Tests/Prefab/Overrides/PrefabInspectorOverrideTests.cpp
```diff
@@ -0,0 +1,83 @@
...
+        CreateAndValidateEditEntityOverride(newEntityId, grandparentContainerId);
+
+        GenerateComponentAdapterDoms(newEntityId);
+
+        AzToolsFramework::ComponentEditor::GetComponentAdapterContentsCallback callback = [](const AZ::Dom::Value& adapterContents)
+        {
+            EXPECT_FALSE(adapterContents.IsArrayEmpty());
+
+            AZ::Dom::Path pathToTranslateProperty(domPathToOverriddenTranslateProperty);
+            [[maybe_unused]] AZ::Dom::Value translateRow = adapterContents[pathToTranslateProperty];
```
forkercat, line 37: curious why it is marked `[[maybe_unused]]`.

Code/Framework/AzToolsFramework/Tests/Prefab/Overrides/PrefabInspectorOverrideTestFixture.cpp
srikappa-amzn: Yes, I wanted to avoid putting the static string in the header file as that'll make a copy in every cpp that includes it. To avoid this, I made it an extern and put it in the header now.

Code/Framework/AzToolsFramework/AzToolsFramework/API/EntityPropertyEditorRequestsBus.h
srikappa-amzn: Yes, I like visit better than 'Process' or 'Update' because the callback merely provides access to the component editors by iterating over the vector, it doesn't modify or update them. I'll change it to VisitComponentEditors() since that'll be more aligned to the visitor pattern being implemented here.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/EntityPropertyEditor.cpp
srikappa-amzn: 1. Yes, it's basically an early-out mechanism. If the caller wants to know if atleast one component editor of a type exists or if a component editor matching a particular id/name exists, they can return false to stop iterating over the rest of the components. 
2. I don't think this function should be returning the value because the idea of implementing this with a function is so that api doesn't directly copy or return the component editors. Besides, the callback will have access to the editors. So callers will still be able to know when the stoppage point happens.

Code/Framework/AzToolsFramework/Tests/Prefab/Overrides/PrefabInspectorOverrideTestFixture.h
```diff
@@ -0,0 +1,38 @@
...
+
+namespace AzToolsFramework
+{
+    class EntityPropertyEditor;
+}
+
+namespace UnitTest
+{
+    using namespace AzToolsFramework::Prefab;
```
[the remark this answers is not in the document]
srikappa-amzn: I generally avoid this in code but I think an exception can be granted for tests and test fixtures. This fixture particularly tests prefab overrides by enabling prefab feature flags and talking to the prefab systems apis. It has very less chances of running into namespace collisions due to dependencies on other systems. Using fully qualified names will be an overkill and make the code really verbose and annoying to write imo. If anything, I'd argue that we change the namespace of this class itself to be 'AzToolsFramework::Prefab' instead of 'UnitTest' if we want to stop doing 'using namespace' . I picked 'UnitTest' because that's what most test fixtures seem to be using but I think it's okay to not use it for every fixture?

Code/Framework/AzToolsFramework/Tests/Prefab/Overrides/PrefabInspectorOverrideTests.cpp
srikappa-amzn: I agree with the sentiment of having a test responsible for a single validation instead of a bunch of different once. However, what's happening here is that we are validating data in a DOM. So, if we have a structure like

- TopNode
  - Child1
     - GrandChild1
      - GrandChild2
        - property1 : true
        
We are basically going down this dom data and validating it at each stage. It makes sense to keep these expectations in one test in this case. Each dom is different and requires different validations. Breaking these expectations into multiple tests would make these much harder to write an manage because of the nature of these tests.

Code/Framework/AzToolsFramework/AzToolsFramework/API/EntityPropertyEditorRequestsBus.h
hultonha: Nice good call! 👍

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/EntityPropertyEditor.cpp
hultonha: Cool makes sense, thanks for the explanation 👍

Code/Framework/AzToolsFramework/Tests/Prefab/Overrides/PrefabInspectorOverrideTests.cpp
hultonha: Okay understood, it might be nice to look at some of the test matcher options to see if any of this could be streamlines but it's not a huge deal. You're right if the checks are all consistent/coherent and related that's definitely not as bad and makes sense 👍

Issue #14782, opened 2023-02-23
DPE: It is possible for the new key dialog ComboBox to be empty when they key type is enum


amzn-tmryan:
**Description**
The default constructed instance value for a reflected enum is 0. This means that when we try to select a value in a ComboBox populated from the default value, the ComboBox will appear empty (see example gif).

![dpeEnumKeyPropertyHandlerDefaultSelection](https://user-images.githubusercontent.com/104796591/221013510-dcc4f2f8-0cd3-4d85-a4b4-4e281098e13c.gif)

**Reproduction Steps**
1. Open a DPE enabled editor (such as in the DPEDebugViewStandalone app).
2. Click the + button next to an associative container that has some enum as its key type, where the enum doesn't contain 0 as a value.
3. ---> The resulting property editor will have no value selected (an invalid selection).

PR #14786, opened 2023-02-23
DPE: Add support for EnumValueKey attribute

amzn-tmryan:
**Description**
Prior to this change, ComboBox property editors which were populated from the SerializeContext's data as opposed to the EditContext would have an empty dropdown in the editor.

This change adds support for `AZ::Serialize::Attributes::EnumValueKey` to the DPE so that enum data coming from the SerializeContext can be used to populate ComboBoxes in the editor.

**Testing**
- Verified the ComboBox property editor in the new key dialog for enum keyed associative contianers are populated
- Verified existing ComboBox related attributes, property handlers, and property editors didn't break

**Introduced Issue**
- The now populated ComboBox is created without a value selected by default in cases where the enum does not contain a 0 value; for example, an `enum class : int { "Thing" = 1, "Stuff" }`
  - A GHI was entered to track this issue: https://github.com/o3de/o3de/issues/14782


Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/DocumentSchema.h
```diff
@@ -276,6 +276,18 @@ namespace AZ::DocumentPropertyEditor
...
 
+            // This case is for handling legacy EnumValueKey attributes
+            if constexpr (AZStd::is_same_v<GenericValueType, uint64_t>)
+            {
+                using EnumConstantBaseType = AZ::SerializeContextEnumInternal::EnumConstantBase;
+                if (auto data = azdynamic_cast<AttributeData<AZStd::unique_ptr<EnumConstantBaseType>>*>(attribute); data != nullptr)
+                {
+                    EnumConstantBaseType* value = static_cast<EnumConstantBaseType*>(data->Get(instance).get());
+                    return ValueToDom(
+                        AZStd::make_pair<GenericValueType, AZStd::string>(value->GetEnumValueAsUInt(), value->GetEnumValueName()));
```
lumberyard-employee-dm, line 287: Do not pass in argument to `AZStd::make_pair`, it's entire purpose is to deduce the parameters.
I have a change in flight where this will be a compile error and it would require the user to either use `AZStd::pair` with the template arguments specified or AZStd::make_pair without the arguments specified.
```suggestion
                        AZStd::pair<GenericValueType, AZStd::string>(value->GetEnumValueAsUInt(), value->GetEnumValueName()));
```

lumberyard-employee-dm (changes requested):

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/Reflection/LegacyReflectionBridge.cpp
```diff
@@ -625,8 +622,26 @@ namespace AZ::Reflection
...
+                            if (name == enumValueKeyName &&
+                                !visitedAttributes.contains(enumValuesCrcName) &&
+                                !visitedAttributes.contains(genericValueListName))
+                            {
+                                genericValueCache.ArrayPushBack(attributeValue);
+                                // Forcing the node's typeId to uint64_t so the correct property handler will be chosen
+                                // in the PropertyEditorSystem.
+                                // This is reasonable since the attribute's value is an enum with an underlying integral
+                                // type which is safely convertible to unint64_t.
+                                nodeData.m_typeId = azrtti_typeid(0Ui64);
```
lumberyard-employee-dm, line 644: This should use `AZ::TypeInfo<AZ::u664>::Uuid`

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/DocumentSchema.h
```diff
@@ -276,6 +276,18 @@ namespace AZ::DocumentPropertyEditor
                 return {};
             }
 
+            // This case is for handling legacy EnumValueKey attributes
+            if constexpr (AZStd::is_same_v<GenericValueType, uint64_t>)
```
lumberyard-employee-dm, line 280: It is better to use an `AZ::u64` as that is always an `unsigned long long` across all of O3DE support platforms, where as `uint64_t` has different underlying types on Linux and Android vs Windows and Mac(`unsigned long` vs `unsigned long long`)

This matters because the Uuid is different for those types: https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/RTTI/TypeInfo.h#L1305-L1306

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/Reflection/LegacyReflectionBridge.cpp
```diff
@@ -571,13 +574,13 @@ namespace AZ::Reflection
...
                     if (!name.IsEmpty())
                     {
-                        // If an attribute of the same name is already loaded then ignore the new value
-                        // unless it is a GenericValue attribute since each represents an individual
-                        // pair destined for a combobox and thus multiple are expected
-                        if (visitedAttributes.find(name) != visitedAttributes.end() && name != genericValueName)
+                        // If an attribute of the same name is already loaded then ignore the new value unless
+                        // it is a GenericValue or EnumValueKey attribute since each represents an individual
+                        // (value, description) pair destined for a combobox and thus multiple are expected
+                        if (visitedAttributes.find(name) != visitedAttributes.end() && name != genericValueName && name != enumValueKeyName)
```
lumberyard-employee-dm, line 583: `visitedAttributes.contains(name)` can be used here.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/DocumentSchema.h
amzn-tmryan: Understood. I thought it was the other way around. Will get that changed here and throughout.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/Reflection/LegacyReflectionBridge.cpp
amzn-tmryan: Embarrassed that I knew that and didn't notice it. Fixing.

amzn-tmryan:
Updated with changes recommended by @lumberyard-employee-dm

I amended out of habit instead of adding a new commit with the updates. Sorry about that.

lumberyard-employee-dm (approved):

lumberyard-employee-dm:
> Updated with changes recommended by @lumberyard-employee-dm
> 
> I amended out of habit instead of adding a new commit with the updates. Sorry about that.

It doesn't matter to me if you amend the commit.
It keeps the commit history simpler.

alexmontAmazon (approved):
looks good to me :)

PR #14790, opened 2023-02-23
Add missing shader to seed list

mbalfour-amzn:
## What does this PR do?

Another diffuse raytracing shader was discovered to be missing from the seed list. Added.

## How was this PR tested?

Built MPS with pak files and verified that it ran without asserting and crashing from missing shaders.


dmcdiarmid-ly (approved):

lmbr-pip (approved):
Thanks for fixing this.

PR #14795, opened 2023-02-24
Introduce an interface to get the polygon point buffer of polygon light.

hsli-in-hw:
## What does this PR do?

Adds an interface to get the polygon point buffer of a polygon light. This is useful for some custom Gems that needs the polygon light point information.

## How was this PR tested?

Using a custom gem calling this function to get the buffer, and set to another shader resource group, and access it in the shader.


rgba16f (approved):

invertednormal (approved):
I'm ok with this for now to unblock work. I think longer term, if we go bindless for all resources, a gem's shader would be able to access this data directly without needing to upload it to another srg. It could just get some offset into the bindless buffer for the data.

moudgils (approved):

rgba16f:
AR run started: https://jenkins.build.o3de.org/job/O3DE/view/change-requests/job/PR-14795/1/

PR #14806, opened 2023-02-24
avoid a crash if the input-image for the TAA pass has no attachment

kh-huawei:
## What does this PR do?

Changes the checks if the Attachments of the TAA - pass are valid. Currently wrong attachments lead to a crash, which makes debugging the actual cause somewhat difficult.

## How was this PR tested?

Used with a custom render-pipeline, where the SizeSource - Slot for the TAA - images was connected, but had no actual attachment.

### DIFF
--- a/Gems/Atom/Feature/Common/Code/Source/PostProcessing/TaaPass.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/PostProcessing/TaaPass.cpp
@@ -134,13 +134,18 @@ namespace AZ::Render
         // Make sure the attachments have images when the pass first loads.
         for (auto i : { 0, 1 })
         {
-            if (m_accumulationAttachments[0] && !m_accumulationAttachments[i]->m_importedResource)
+            if (m_accumulationAttachments[i])
             {
-                attachmentsValid &= UpdateAttachmentImage(i);
+                attachmentsValid = UpdateAttachmentImage(i);
+                if (!attachmentsValid)
+                {
+                    break;
+                }
             }
             else
             {
                 attachmentsValid = false;
+                break;
             }
         }
         if (!attachmentsValid)
@@ -199,7 +204,7 @@ namespace AZ::Render
         if (attachment->m_importedResource && imageDesc.m_size == currentImage->GetDescriptor().m_size)
         {
             // If there's a resource already and the size didn't change, just keep using the old AttachmentImage.
-            return false;
+            return true;
         }
         
         Data::Instance<RPI::AttachmentImagePool> pool = RPI::ImageSystemInterface::Get()->GetSystemAttachmentPool();
--- a/Gems/Atom/Feature/Common/Code/Source/PostProcessing/TaaPass.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/PostProcessing/TaaPass.cpp
@@ -6,6 +6,7 @@
  *
  */
 
+#include <Atom/RHI.Reflect/Format.h>
 #include <PostProcessing/TaaPass.h>
 
 #include <AzCore/Math/Random.h>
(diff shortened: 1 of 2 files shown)


rgba16f (dismissed):

Gems/Atom/Feature/Common/Code/Source/PostProcessing/TaaPass.cpp
```diff
@@ -169,23 +176,30 @@ namespace AZ::Render
...
+            return false;
+        }
+
         RPI::AttachmentImage* currentImage = azrtti_cast<RPI::AttachmentImage*>(attachment->m_importedResource.get());
 
         if (attachment->m_importedResource && imageDesc.m_size == currentImage->GetDescriptor().m_size)
         {
             // If there's a resource already and the size didn't change, just keep using the old AttachmentImage.
-            return;
+            return false;
```
invertednormal, line 202: While this case shouldn't happen when `BuildInternal()` is called (only on `FrameBeginInternal()`), it seems odd to return false for this case since everything's really OK, it just doesn't need to update.

invertednormal (changes requested):
Looks good, just one minor issue.

Gems/Atom/Feature/Common/Code/Source/PostProcessing/TaaPass.cpp
```diff
@@ -129,21 +130,27 @@ namespace AZ::Render
...
+            {
+                attachmentsValid &= UpdateAttachmentImage(i);
+            }
+            else
             {
-                if (!m_accumulationAttachments[i]->m_importedResource)
-                {
-                    UpdateAttachmentImage(i);
-                }
+                attachmentsValid = false;
```
jiaweig-amzn, line 143: It can break here to skip checking the second attachment.

Gems/Atom/Feature/Common/Code/Source/PostProcessing/TaaPass.cpp
```diff
@@ -129,21 +130,27 @@ namespace AZ::Render
...
-        if (hasAttachments)
+        bool attachmentsValid = true;
+        // Make sure the attachments have images when the pass first loads.
+        for (auto i : { 0, 1 })
         {
-            // Make sure the attachments have images when the pass first loads.
-            for (auto i : { 0, 1 })
+            if (m_accumulationAttachments[0] && !m_accumulationAttachments[i]->m_importedResource)
+            {
+                attachmentsValid &= UpdateAttachmentImage(i);
```
jiaweig-amzn, line 139: Bitwise operation is not recommended for bool. It may cause different behavior from logical operation in some case like 1&2=0. And it doesn't have the ability to do short circuit, in case when `attachmentsValid` is false, but you can break in the else branch.

Gems/Atom/Feature/Common/Code/Source/PostProcessing/TaaPass.cpp
```diff
@@ -129,21 +130,27 @@ namespace AZ::Render
...
-        AZ_Error("TaaPass", hasAttachments, "TaaPass must have Accumulation1 and Accumulation2 ImageAttachments defined.");
-
-        if (hasAttachments)
+        bool attachmentsValid = true;
+        // Make sure the attachments have images when the pass first loads.
+        for (auto i : { 0, 1 })
         {
-            // Make sure the attachments have images when the pass first loads.
-            for (auto i : { 0, 1 })
+            if (m_accumulationAttachments[0] && !m_accumulationAttachments[i]->m_importedResource)
```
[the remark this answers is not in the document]
kh-huawei: fixed

Gems/Atom/Feature/Common/Code/Source/PostProcessing/TaaPass.cpp
kh-huawei: changed to return true.

Gems/Atom/Feature/Common/Code/Source/PostProcessing/TaaPass.cpp
kh-huawei: removed the bitwise operations, and break early if one attachment is invalid

rgba16f (approved):

rgba16f:
AR run started: https://jenkins.build.o3de.org/job/O3DE/view/change-requests/job/PR-14806/1/

jiaweig-amzn (approved):

invertednormal (approved):

PR #14808, opened 2023-02-24
Fix an assert with the normalized direction for polygon lights

kh-huawei:
## What does this PR do?

Placing a polygonal light in the scene would continuously assert, because the direction wasn't normalized.

## How was this PR tested?

Place a Polygonal light in a scene.

### DIFF
--- a/Gems/Atom/Feature/Common/Code/Source/CoreLights/PolygonLightFeatureProcessor.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/CoreLights/PolygonLightFeatureProcessor.cpp
@@ -289,7 +289,7 @@ namespace AZ::Render
         else
         {
             AZ::Vector3 normal = AZ::Vector3::CreateFromFloat3(data.m_direction.data());
-            bounds.emplace<Hemisphere>(AZ::Hemisphere(position, radius, normal));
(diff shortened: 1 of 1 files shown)


rgba16f (approved):

gadams3 (approved):

Gems/Atom/Feature/Common/Code/Source/CoreLights/PolygonLightFeatureProcessor.cpp
```diff
@@ -289,7 +289,7 @@ namespace AZ::Render
         else
         {
             AZ::Vector3 normal = AZ::Vector3::CreateFromFloat3(data.m_direction.data());
-            bounds.emplace<Hemisphere>(AZ::Hemisphere(position, radius, normal));
+            bounds.emplace<Hemisphere>(AZ::Hemisphere(position, radius, normal.GetNormalized()));
```
invertednormal, line 292: This really should be done in `SetPolygonPoints()`, not here. That way the normalize call only happen when it needs to.

invertednormal (changes requested):
Thanks for the fix, please just move the normalize to where the direction is set instead of where it's used.

rgba16f:
AR run started: https://jenkins.build.o3de.org/job/O3DE/view/change-requests/job/PR-14808/1/

invertednormal (approved):

PR #14809, opened 2023-02-24
make hand-written pass-templates more consistent with json serialization

kh-huawei:
## What does this PR do?

Changes a few values in the json templates so they are somewhat more future proof, and stick to the case convention as in the Reflect - functions. Mostly this makes parsing the templates by an external script a bit easier.

## How was this PR tested?

Display a basic scene. The PR didn't introduce any functional difference.

### DIFF
--- a/Gems/Atom/Feature/Common/Assets/Passes/BloomDownsample.pass
+++ b/Gems/Atom/Feature/Common/Assets/Passes/BloomDownsample.pass
@@ -54,7 +54,7 @@
                     },
                     "ImageDescriptor": {
                         "MipLevels": "5",
-                        "SharedQueueMask": 3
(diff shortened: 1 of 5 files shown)


gadams3 (approved):
I noticed the same thing with shader source data files. Most of them were originally hand edited. But if you load them using the JSON serializer and resave they will eliminate unnecessary fields and format everything consistently, arrange the data according to how it's defined in code, with the correct casing and other substitutions.

moudgils (approved):

invertednormal (approved):

rgba16f:
AR run started: https://jenkins.build.o3de.org/job/O3DE/view/change-requests/job/PR-14809/1/

PR #14781, opened 2023-02-23
Add property icon and label handler for override in DPE Entity Inspector

forkercat:
## What does this PR do?

_Feature is behind two flags: `ed_enableDPE` and `EnableInspectorOverrideManagement`_

This PR improves property override visualization in DPE Entity Inspector. It fixes spacing and indentation issues that were observed in previous PR: https://github.com/o3de/o3de/pull/14507.

### Change List
- Replace `OnBeginRow` with `CreateLabel` function and subclass adapter can handle label node creation differently
- Replace `OverridePropertyHandler` with a new handler that encapsulates icon and label widgets
- Replace `PrefabOverrideProperty` with a new node and attributes
- Disable override visualization and interaction for container entity
- Context menu can be right clicked for both icon and label including the empty area between text and property editor
- Context menu can show up only when the property is overridden

Inspector Override RFC: https://github.com/o3de/sig-content/issues/112

## How was this PR tested?

Manually tested in Editor. Validate the icon can show up for different property types. There are some known and documented issues in the inspector where the icon status might not be updated accordingly.

PR #14810, opened 2023-02-24
Add component override visualization in DPE based inspector

michabr:
## What does this PR do?

Inspector changes to represent prefab overrides on the component level:
- Overlay an icon on top of the component icon in the component header to represent a component added as an override or a component containing property overrides
- Display a "revert overrides" context menu when clicking on the component icon

## How was this PR tested?

Tested in Editor with associated registry flags enabled and disabled.

Screenshots (note that individual property icons will be represented in a separate PR: https://github.com/o3de/o3de/pull/14781)
ComponentOverrides
ComponentOverridesMenu


srikappa-amzn (commented):
The new icons are great and they fit well on top of existing icons! 

I have a few thoughts on how they look currently look though. I'm not sure whether its the size of them or color of them but they look like they are blending with the original icons rather than indicating 'hey, this is a different icon'. Here are a few suggestion to make them stand apart:

1. Increase the size of them
2. Maybe lose the black border around the icon and around the plus icon as they appear to make the icon even darker and less distinguishable.

If UX okayed the current look, then I'm okay with keeping them as is. This is just what I felt on first look.
@rainbj , fyi

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverrideHandler.cpp
```diff
@@ -62,10 +62,62 @@
...
+                    break;
+                default:
+                    break;
+                }
+            }
+
+            return overrideType;
+        }
+
+        AZStd::optional<OverrideType> PrefabOverrideHandler::GetComponentOverrideType(AZ::Dom::Path path, LinkId linkId) const
```
srikappa-amzn, line 91: I don't think this class should have 2 methods for fetching entity and component override types. Irrespective of it being an entity or a component, the logic to fetch the type is the same. So, we can just have a single method called GetOverrideType. And the PrefabPublicHandler can have 2 different methods that convert these types to component/entity types as needed.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverridePublicInterface.h
```diff
@@ -34,12 +34,23 @@ namespace AzToolsFramework
...
+
             //! Revert overrides on the entity matching the entity id. Returns false if no overrides are present on the entity.
             //! @param entityId The id of the entity on which overrides should be reverted.
             //! @return Whether overrides are successfully reverted on the entity.
             virtual bool RevertOverrides(AZ::EntityId entityId) = 0;
+
+            //! Revert overrides on the given component. Returns false if no overrides are present on the component.
+            //! @param component The component on which overrides should be reverted.
+            //! @return Whether overrides are successfully reverted on the component.
+            virtual bool RevertComponentOverrides(const AZ::Component& component) = 0;
```
srikappa-amzn, line 53: The idea is to keep this interface scriptable so that we can write python automated tests by calling this. As such, the parameters of the functions here should be usable in tests. I don't think AZ::Component is a type that python tests can interpret. Since we only care about GetSerializedIdentifier() within the component, I suggest you pass that as a string to these new functions.

Another alternative could be to pass the component id and fetch the component from the entity using the id. However, you'd have to jump multiple hoops to get to the serialized identifier in that case. So the above approach would be better imo

srikappa-amzn (changes requested):

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverrideHandler.cpp
michabr: The reason for the two methods is to distinguish between the user wanting the override type on entity scope or on a component scope.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverridePublicInterface.h
michabr: I thought about using component Id and using Entity::FindComponent() to find the component. The "find" method isn't that efficient though, but it might be OK enough for this use case. I'll look into using the serialized identifier as well.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverrideHandler.cpp
srikappa-amzn: Yes. I'm basically suggesting that we move this calculation logic under PrefabOverridePublicHandler::GetComponentOverrideType itself because that function still accepts a component ref as a variable. 

So, for a given api (PrefabOverridePublicHandler), saying 'I'll give you the component, you give me the component override type' makes more sense than, doing something for PrefabOverrideHandler like 'I'll give you the path, you give me the component override type'.  Because what if a path corresponding to a 'remove-entity' patch is provided to this function? In that case it would return 'remove-component' as the type. So the api is ambiguous.

michabr: Makes sense. Changed.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverridePublicInterface.h
michabr: Changed to use EntityComponentIdPair.

michabr:
> 

OK, I'll also wait to get feedback from @rainbj as he provided the icons.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/ComponentEditor.cpp
```diff
@@ -201,7 +201,12 @@ namespace AzToolsFramework
 
         if (DocumentPropertyEditor::ShouldReplaceRPE())
         {
-            m_adapter->SetComponent(componentInstance);
+            if (!aggregateInstance)
```
srikappa-amzn, line 204: Nice!

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DPEComponentAdapter.h
```diff
@@ -44,6 +44,9 @@ namespace AZ::DocumentPropertyEditor
         //! Sets the component, connects the appropriate Bus Handlers and sets the reflect data for this instance
         void SetComponent(AZ::Component* componentInstance);
 
+        const AZ::Component* GetConstComponent();
+        AZ::Component* GetComponent();
+
```
srikappa-amzn, line 49: These are not needed and should be removed

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/ComponentEditorHeader.hxx
```diff
@@ -45,6 +45,13 @@ namespace AzToolsFramework
         /// Set an icon. Passing a null icon will hide the widget.
         void SetIcon(const QIcon& icon);
 
+        /// Set a secondary icon to be drawn on top of the main icon.
+        void SetIconOverlay(const QIcon& icon);
+
+        // Set whether the component icon is clickable
+        void SetComponentIconClickable(bool clickable);
+        bool IsComponentIconClickable() const;
```
srikappa-amzn, line 53: Looks like this is not being used in the PR. Can be deleted?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/EntityPropertyEditor.cpp
```diff
@@ -1815,6 +1884,27 @@ namespace AzToolsFramework
...
+                // Subscribe to DPE property changes to keep the component icon updated based on override state
+                auto propertyChangeHandler = AZ::DocumentPropertyEditor::ReflectionAdapter::PropertyChangeEvent::Handler(
+                    [this, componentEditor](const AZ::DocumentPropertyEditor::ReflectionAdapter::PropertyChangeInfo& changeInfo)
+                    {
+                        if (changeInfo.changeType == AZ::DocumentPropertyEditor::Nodes::ValueChangeType::FinishedEdit)
+                        {
+                            UpdateOverrideVisualization(*componentEditor);
+                        }
+                    });
+                componentEditor->ConnectPropertyChangeHandler(propertyChangeHandler);
```
srikappa-amzn, line 1905: This will create a copy of the the propertyChangeHandler in the ComponentEditor class. Instead of creating it here and passing it as a const ref, maybe you can create it in the ComponenEditor class itself? 

If for whatever reason you can't create it in the ComponentEditor class itself, you can at least move it from here instead of copying.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/ComponentEditor.cpp
```diff
@@ -952,6 +963,13 @@ namespace AzToolsFramework
         // refresh which Component Editor/Card looks selected in the Entity Outliner
         SetSelected(componentType == m_componentType);
     }
+
+    void ComponentEditor::ConnectPropertyChangeHandler(
```
srikappa-amzn, line 967: Is this class needed at all? Can we not create and connect a handler at the end of the ComponentEditor constructor?

Code/Framework/AzQtComponents/AzQtComponents/Components/Widgets/CardHeader.cpp
```diff
@@ -186,20 +218,45 @@ namespace AzQtComponents
...
+    {
+        m_iconLabel->SetClickable(clickable);
+    }
+
+    bool CardHeader::isIconClickable() const
+    {
+        return m_iconLabel->GetClickable();
+    }
+
+    void CardHeader::updateIconLabel()
```
srikappa-amzn, line 240: It'd be good to get someone who worked on AzQtComponents to review these changes since this'll affect all card headers.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/ComponentEditorHeader.hxx
michabr: SetComponentIconClickable is being used in EntityPropertyEditor.cpp. Are you suggesting to remove the getter method?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/ComponentEditor.cpp
michabr: I'm keeping the prefab override code outside of ComponentEditor because ComponentEditor is also used in non-prefab related tools.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/ComponentEditorHeader.hxx
srikappa-amzn: I meant the 'IsComponentIconClickable()' method

michabr: OK, I typically like to have an accompanying getter method but I can remove it.

Code/Framework/AzQtComponents/AzQtComponents/Components/Widgets/CardHeader.cpp
michabr: @cgalvan, @AMZN-daimini or @alexmontAmazon could you take a look at these changes if you get a chance?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverridePublicHandler.cpp
```diff
@@ -111,5 +171,36 @@ namespace AzToolsFramework
...
+
+        AZStd::pair<AZ::Dom::Path, LinkId> PrefabOverridePublicHandler::GetComponentPathAndLinkIdFromFocusedPrefab(
+            const AZ::EntityComponentIdPair& entityComponentIdPair)
+        {
+            AZStd::pair<AZ::Dom::Path, LinkId> pathAndLinkIdPair = GetPathAndLinkIdFromFocusedPrefab(entityComponentIdPair.GetEntityId());
+            if (!pathAndLinkIdPair.first.IsEmpty() && pathAndLinkIdPair.second != InvalidLinkId)
+            {
+                AZ::Entity* entity = nullptr;
+                AZ::ComponentApplicationBus::BroadcastResult(
+                    entity, &AZ::ComponentApplicationRequests::FindEntity, entityComponentIdPair.GetEntityId());
```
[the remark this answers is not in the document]
michabr: I updated the PR with 1. to get the entity from the instance.

michabr:
> @michabr , since there's quite a bit of new functionality being added in this PR, I assume there'll be follow up tests to cover this work?
> 
> I'll leave it up to you to come up with exact tests but looking at this PR, I think these are the cases that'd be good to be covered:
> 
> * Test that 'add-component-as-override' icon appears
> * Test that 'edit-component-as-override' icon appears
> * Test that the above icons are clickable
> * Test that regular (non-override) icons are not clickable
> * Test that Revert option shows up and clicking it reverts
>   
>   * Test undo/redo on this too

@srikappa-amzn I will make a follow up task to add automated tests for this!

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/EntityPropertyEditor.cpp
```diff
@@ -1790,6 +1827,38 @@ namespace AzToolsFramework
...
+                }
+            }
+
+            AZStd::string prevIconOverlayPath(componentEditor.property("IconOverlayPath").toString().toUtf8());
+            if (prevIconOverlayPath != iconOverlayPath)
+            {
+                componentEditor.setProperty("IconOverlayPath", iconOverlayPath.c_str());
+                componentEditor.GetHeader()->SetIconOverlay(QIcon(iconOverlayPath.c_str()));
+                componentEditor.GetHeader()->SetComponentIconClickable(!iconOverlayPath.empty());
+            }
```
forkercat, line 1858: Is it for updating visualization only when the icon path changes?

srikappa-amzn (approved):

forkercat (approved):

forkercat:
> I'll leave it up to you to come up with exact tests but looking at this PR, I think these are the cases that'd be good to be covered:
> 
> * Test that 'add-component-as-override' icon appears
> * Test that 'edit-component-as-override' icon appears
> * Test that the above icons are clickable
> * Test that regular (non-override) icons are not clickable
> * Test that Revert option shows up and clicking it reverts
>   
>   * Test undo/redo on this too

+1. Before adding the tests, I think we need to add tests that first verify `GetComponentOverrideType()` and `GetEntityOverrideType()` work as expected, which also checks `GetPatchType()`.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/EntityPropertyEditor.cpp
```diff
@@ -902,6 +905,33 @@ namespace AzToolsFramework
...
+        const AZStd::vector<AZ::Component*>& components = componentEditor->GetComponents();
+        AZ_Assert(!components.empty() && components[0], "ComponentEditor should have at least one component.");
+        QMenu menu;
+
+        QAction* revertAction = menu.addAction(QObject::tr("Revert Overrides"));
+        QObject::connect(
+            revertAction,
+            &QAction::triggered,
+            this,
+            [components]
```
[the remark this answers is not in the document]
michabr: I think I passed in the vector for when we support multiple selection. I'll add a comment.

rainbj:
> I'm not sure whether its the size of them or color of them but they look like they are blending with the original icons rather than indicating 'hey, this is a different icon'. Here are a few suggestion to make them stand apart
>
> 1. Increase the size of them
> 2. Maybe lose the black border around the icon and around the plus icon as they appear to make the icon even darker and less distinguishable.
> 
> If UX okayed the current look, then I'm okay with keeping them as is. This is just what I felt on first look.

1 - I'm not sure increasing the icon here is a good call. The icon itself is 16 x 16 and the dot is 9x9 making it over 50% of the original icon size. My suggestion is we move away from the blue as an override color and go back to the orange as it stands out more and is easier to see from further away. 



2 I don't recommend losing the border color, as it makes complex shape icons in the background hard to distinguish from the foreground dot at scale.

rainbj:
> Visualization looks good! One minor UX comment is do we want 'Revert overrides' option to show up only in header (via right click)? Do we need to add the option in the below context menu as well? @rainbj

I think it makes sense to have it in both places. However, the question here is about maintenance. Can we do it in both places, but the code references the same function so that if we ever update the functionality it updates it everywhere? If yes, then yes please.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/EntityPropertyEditor.cpp
michabr: Yes, it's so the component label pixmap only gets updated when the icon path changes.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/DocumentPropertyEditor/Resources/OverrideResources.qrc
```diff
@@ -0,0 +1,6 @@
+<!DOCTYPE RCC><RCC version="1.0">
+    <qresource prefix="/OverrideResources">
+		<file>component_modified_as_override.svg</file>
+		<file>component_added_as_override.svg</file>
+    </qresource>
+</RCC>
```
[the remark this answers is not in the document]
rainbj: I was considering using the color inside the property row (unselected state) to be the color of the border. Do you have an example of what it looks like with the darker color? Do you like it? FYI the icon from above is not centered, so please don't use the exact same image as it will look off centered. You can get the centered on in the same location as the original icons provided from above.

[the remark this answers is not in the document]
forkercat: I had this thought because I noticed in your prototype it has the black border. But yeah I need to prototype it first so we can see if it can fit well.

forkercat:
> > Visualization looks good! One minor UX comment is do we want 'Revert overrides' option to show up only in header (via right click)? Do we need to add the option in the below context menu as well? @rainbj
> 
> I think it makes sense to have it in both places. However, the question here is about maintenance. Can we do it in both places, but the code references the same function so that if we ever update the functionality it updates it everywhere? If yes, then yes please.

Okay. I think we can come back to this later since the current version (right click on header) is quite adequate and straightforward. It doesn't seem to be hard to extend the functionality to that context menu.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverrideHandler.cpp
```diff
@@ -78,22 +78,22 @@
...
                         if (opPath == "remove")
                         {
-                            overrideType = OverrideType::RemoveEntity;
+                            patchType = PatchType::Remove;
                         }
                         else if (opPath == "add")
                         {
-                            overrideType = OverrideType::AddEntity;
+                            patchType = PatchType::Add;
                         }
```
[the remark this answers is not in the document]
michabr: You're right. I ported the method over which was originally meant to handle paths to the entity only. I updated this new function to handle PatchType::Edit if exact path to a property is passed in.

Code/Framework/AzQtComponents/AzQtComponents/Components/Widgets/CardHeader.cpp
```diff
@@ -21,12 +21,43 @@
...
+            {
+                emit clicked(event->globalPos());
+            }
+            else
+            {
+                event->ignore();
+            }
+        }
+    }
+
```
cgalvan, line 60: Instead of making a clickable label, you could just change `m_iconLabel` to be a `QToolButton`, which would give you both an icon (pixmap) and be clickable.

Code/Framework/AzQtComponents/AzQtComponents/Components/Widgets/CardHeader.cpp
```diff
@@ -186,20 +218,45 @@ namespace AzQtComponents
...
+    {
+        if (!m_icon.isNull())
         {
-            m_iconLabel->setPixmap(icon.pixmap(s_iconSize, s_iconSize));
+            QPixmap pixmap = m_icon.pixmap(s_iconSize, s_iconSize);
+            if (!m_iconOverlay.isNull())
+            {
+                QPainter paint(&pixmap);
+                paint.drawPixmap(0, 0, m_iconOverlay.pixmap(s_iconSize, s_iconSize));
+            }
```
cgalvan, line 249: Nice! This would still work with the above comment about switching to a `QToolButton` instead of a `QLabel` just need to convert the pixmap back to an icon.

cgalvan (approved):

Code/Framework/AzQtComponents/AzQtComponents/Components/Widgets/CardHeader.cpp
cgalvan: Reviewed!

Code/Framework/AzQtComponents/AzQtComponents/Components/Widgets/CardHeader.cpp
michabr: Good to know! I remembered that I also added the custom clickable QLabel to be able to set a clickable flag on it rather than enabling/disabling the QWidget because the enable/disable property was already being set for visual reasons.

PR #14815, opened 2023-02-24
Automated test for current (not desired) source dependency behavior

AMZN-stankowi:
## What does this PR do?

There is an issue where asset jobs that emit job dependencies on intermediate assets will run before the intermediate assets are created, and then run a second time after the intermediate asset is created.

Currently, this surfaces as an issue where FBX files initially fail to process, and eventually succeed by the time asset processor goes idle.

This ticket is an automated test for the current, not desired, behavior. This is to get this out and in front of people, I'm going to next work on adjusting this behavior. I thought it would be much easier to iterate on a fix with an automated test that demonstrates the undesired workflow for FBX files, instead of constantly reprocessing a lot of the asset cache, and we will want an automated test for the fix when it goes in.

## How was this PR tested?

Ran test locally, it passes.


Code/Tools/AssetProcessor/native/tests/assetmanager/AssetProcessorManagerTest.cpp
```diff
@@ -846,6 +847,330 @@ TEST_F(AssetProcessorIntermediateAssetTests, IntermediateAsset_SourceNoLongerEmi
...
+                auto product =
+                    JobProduct{ outputFile.c_str(), AZ::Data::AssetType::CreateName(m_intermediateExtension.c_str()), AssetSubId };
+
+                product.m_outputFlags = ProductOutputFlags::IntermediateAsset;
+
+                product.m_dependenciesHandled = true;
+                response.m_outputProducts.push_back(product);
+                response.m_resultCode = ProcessJobResult_Success;
+            },
+            "fingerprint");
```
amzn-mike, line 911: This looks functionally pretty similar to the default builder setup provided by the base class.  Is there a reason for a custom setup here?

Code/Tools/AssetProcessor/native/tests/assetmanager/AssetProcessorManagerTest.cpp
```diff
@@ -846,6 +847,330 @@ TEST_F(AssetProcessorIntermediateAssetTests, IntermediateAsset_SourceNoLongerEmi
...
+        AssetProcessorIntermediateAssetTests::SetUp();
+
+        GenerateAssetPaths();
+        CreateTestAssets();
+
+        // Jobs with dependencies need those dependencies to have updated the asset catalog before the job with the dependency runs.
+        SetCatalogToUpdateOnJobCompletion();
+
+        CreateBuilders();
+
```
amzn-mike, line 1020: Nit: extra blank line

Code/Tools/AssetProcessor/native/tests/assetmanager/AssetProcessorManagerTest.cpp
```diff
@@ -846,6 +847,330 @@ TEST_F(AssetProcessorIntermediateAssetTests, IntermediateAsset_SourceNoLongerEmi
...
+    UnitTests::JobSignalReceiver receiver;
+
+    // Processing asset B will produce several warnings right now (listed at the end of this block, where they are checked).
+    m_errorChecker.Begin();
+
+    // Process the first asset in the queue, which will be the job for B, because it's a PC job, which takes priority over the Common platform job in the queue.
+    QCoreApplication::processEvents();
+    receiver.WaitForFinish();
+    QCoreApplication::processEvents();
+    QCoreApplication::processEvents();
```
amzn-mike, line 1119: I know its a bit of a pain but if you can put comments on what function should be running during each processEvents, it helps keep track of what's happening when reading the test and also with knowing where to look to debug.

amzn-mike (approved):

Code/Tools/AssetProcessor/native/tests/assetmanager/AssetProcessorManagerTest.cpp
AMZN-stankowi: Removed the custom builders for the two cases that didn't need them right now. I originally had these to manage the job priority, before I discovered Common platform jobs always come after Host platform jobs.

Code/Tools/AssetProcessor/native/tests/assetmanager/AssetProcessorManagerTest.cpp
AMZN-stankowi: I went through and commented them, as well as collapsed one set into a function.

amzn-mike (approved):

AMZN-stankowi:
The bug this was written to validate was fixed in this pull request, with an update to the automated test https://github.com/o3de/o3de/pull/14862

PR #14816, opened 2023-02-25
Atom/galibzon/shader variant list builder

galibzon:
## What does this PR do?

This is the new ShaderVariantListBuilder.
It fixes the problem we have today that any change
to a `.shadervariantlist` file would cause recompilation
of all the shader variants listed in the `.shadervariantlist`.
  
This builder works based on the concept of
Intermediate Assets.

Current State of Affairs:  
`*.shadervariantlist` => ShaderVariantAssetBuilder => `*.azshadervarianttree` + several `*.azshadervariant` assets 
The End Game:  
`*.shadervariantlist` => ShaderVariantListBuilder ==> one `*.hashedvariantlist` + several `*.hashedvarianinfo` (one per variant) ==> ShaderVariantAssetBuilder => `*.azshadervarianttree` + several `*.azshadervariant` assets 

---
=== The first time around ===  
The ShaderVariantListBuilder will find
that a new `<shader_name>.shadervariantlist` file has been added to the filesystem
and the ShaderVariantListBuilder produces two kinds of Intermediate Assets from a single
.shadervariantlist file.
The first intermediate asset is the ".hashedvariantlist" file:
    Will contain a consolidated list of all variants that will be necessary to create a ShaderVariantTreeAsset.
The second intermediate asset is the ".hashedvariantinfo" files:
    There will be one of these for each variant specified in the .shadervariantlist file.
    These files will be named as <Shader Name>_<StableId>.hashedvariantinfo.

The ShaderVariantAssetBuilder will wake up and find that there's a ".hashedvariantlist" file
and will spawn a job to create the ShaderVariantTreeAsset (*.azshadervarianttree).
Additionally, for each `<Shader Name>_<StableId>.hashedvariantinfo` it will spawn a job
to create a set of ShaderVariantAsset (*.azshadervariant), one for each combination of Supervariant,
RHI and StableId.
  
---

=== The second time around ===  
After the user modifies `<shader_name>.shadervariantlist`,
either by adding a new variant or modifying an existing one, the ShaderVariantListBuilder
will wake up and produce all the intermediate assets mentioned above, +/- the new additions
or removals.

The AssetProcessor will find that most *.hashedvariantinfo files have not changed, and will wake up
the ShaderVariantAssetBuilder to recreate the ShaderVariantTreeAsset, and create only the new variants.  

---
REMARK1: This PR introduces the ShaderVariantListBuilder,
but for the sake of simplifying the Code Review, this builder
will be inactive because it listens for file with extension `*.shadervariantlist2`
which no such file exist yet in the engine.

REMARK2: During debugging the ShaderVariantListBuilder the new algorithms
found that `o3de\Gems\Atom\Feature\Common\Assets\Shaders\SkyAtmosphere\SkyRayMarching.shadervariantlist`
had a redundant variant with StableId=4, which was the same variant as StableId=2.
This has been fixed by removing the variant with StableId=4.

## How was this PR tested?

Played with several scenarios of deleting and updating `SkyRayMarching.shadervariantlist2`, and use it as a project customization in AutomatedTesting and as part of the `Atom/Features/Common` gem

### DIFF
--- a/Gems/Atom/Asset/Shader/Code/Source/Editor/HashedVariantListSourceData.h
+++ b/Gems/Atom/Asset/Shader/Code/Source/Editor/HashedVariantListSourceData.h
@@ -1,0 +1,71 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+#pragma once
+
+#include <AzCore/std/string/string.h>
+#include <AzCore/std/containers/unordered_map.h>
+#include <AzCore/Memory/SystemAllocator.h>
+#include <Atom/RPI.Edit/Shader/ShaderVariantListSourceData.h>
+
+namespace AZ
+{
+    class ReflectContext;
+
+    namespace ShaderBuilder
+    {
+        //! This is a simple data structure that represents a .hashedvariantlist file.
+        //! Users are not expected to create .hashedvariantlist files. These files are produced by the ShaderVariantListBuilder
+        //! as intermediate assets.
+        //! Provides configuration data about which shader variants should be generated for a given shader.
+        struct HashedVariantListSourceData
+        {
+            AZ_TYPE_INFO_WITH_NAME_DECL(HashedVariantListSourceData);
+            AZ_CLASS_ALLOCATOR(HashedVariantListSourceData, AZ::SystemAllocator);
+
+            static constexpr const char* Extension = "hashedvariantlist";
+            static constexpr uint32_t SubId = 0;
+
+            static void Reflect(ReflectContext* context);
+
+            //! A struct that describes each hashed shader variant data.
+            struct HashedVariantInfo
+            {
+                AZ_TYPE_INFO_WITH_NAME_DECL(HashedVariantInfo);
+
+                AZ::RPI::ShaderVariantListSourceData::VariantInfo m_variantInfo;
+                size_t m_hash = 0; // Hash of all the data in @m_variantInfo
+                bool m_isNew = false; // If true, the corresponding ShaderVariantAsset should be built/rebuilt.
+
+                //! Hash combines all the data in @optionValues.
+                static size_t HashCombineShaderOptionValues(size_t startingHash, const AZ::RPI::ShaderOptionValuesSourceData& optionValues);
+
+                //! Hash combines the result of a previous call to HashCombineShaderOptionValues, which is passed to this function
+                //! in @optionValuesHash argument, with the rest of the data in @variantInfo.
+                static size_t CalculateHash(size_t optionValuesHash, const AZ::RPI::ShaderVariantListSourceData::VariantInfo& variantInfo);
+
+                //! Same as above, but uses this.m_variantInfo and stores the result in this.m_hash.
+                void CalculateHash(size_t optionValuesHash);
+            };
+
+            // A time stamp is necessary, because building shader variants takes time.
+            // We calculate hashes to figure out the variants that changed and based on the match
+            // we set HashedVariantInfo::m_isNew to false or true.
+            // Imagine a user makes changes to a .shadervariantlist, and some HashedVariantInfos are marked as new.
+            // then within a few seconds later they make another change to the .shadervariantlist file. Because
+            // it happened so quickly those HashedVariantInfos that were set as new would become "old" and those
+            // shader variants won't be compiled.
+            // This timestamp comes to the rescue and we can measure if the change happened too quick
+            // and it such case we preserve the state of each HashedVariantInfo::m_isNew.
+            long long m_timeStamp;
+
+            AZStd::string m_shaderFilePath; // .shader file.
+            AZStd::vector<HashedVariantInfo> m_hashedVariants;
+        };
+
+    } // namespace ShaderBuilder
+} // namespace AZ
--- a/Gems/Atom/Asset/Shader/Code/Source/Editor/HashedVariantListSourceData.h
+++ b/Gems/Atom/Asset/Shader/Code/Source/Editor/HashedVariantListSourceData.h
@@ -18,10 +18,52 @@ namespace AZ
 
     namespace ShaderBuilder
     {
+        //! This structure represents the content of an intermediate asset
+        //! with file extension "hashedvariantinfo".
+        //! The ShaderVariantAssetBuilder will react to this file pattern and create
+        //! a single product with extension "azshadervariant" (ShaderVariantAsset) per "hashedvariantinfo".
+        //!
+        //! This struct is also leveraged by HashedVariantListSourceData (see below)
+        //! to create a single list of all variants, which will be used to create
+        //! another intermediate asset called the "hashedvariantlist" which will be used
+        //! by the ShaderVariantAssetBuilder to build the "azshadervarianttree" output asset
+        //! that the runtime will load a ShaderVariantTreeAsset.
+        //!
+        //! REMARK1: Users are not expected to create .hashedvariantinfo files. These files are produced by the ShaderVariantListBuilder
+        //! as intermediate assets.
+        //!
+        //! REMARK2: These files will be named <Shader Name>_<StableId>.hashedvariantinfo, Where the StableId is a 1-based index.
+        //!
+        //! REMARK3: As an Intermediate Asset the Product SubID will be the StableId, because SubId 0 is reserved for the
+        //! ".hashedvariantlist".
+        struct HashedVariantInfoSourceData
+        {
+            AZ_TYPE_INFO_WITH_NAME_DECL(HashedVariantInfoSourceData);
+
+            static constexpr const char* Extension = "hashedvariantinfo";
+
+            AZ::RPI::ShaderVariantListSourceData::VariantInfo m_variantInfo;
+            size_t m_hash = 0; // Hash of all the data in @m_variantInfo
+
+            //! Hash combines all the data in @optionValues.
+            static size_t HashCombineShaderOptionValues(size_t startingHash, const AZ::RPI::ShaderOptionValuesSourceData& optionValues);
+
+            //! Hash combines the result of a previous call to HashCombineShaderOptionValues, which is passed to this function
+            //! in @optionValuesHash argument, with the rest of the data in @variantInfo.
+            static size_t CalculateHash(size_t optionValuesHash, const AZ::RPI::ShaderVariantListSourceData::VariantInfo& variantInfo);
+
+            //! Same as above, but uses this.m_variantInfo and stores the result in this.m_hash.
+            void CalculateHash(size_t optionValuesHash);
+        };
+
         //! This is a simple data structure that represents a .hashedvariantlist file.
         //! Users are not expected to create .hashedvariantlist files. These files are produced by the ShaderVariantListBuilder
         //! as intermediate assets.
-        //! Provides configuration data about which shader variants should be generated for a given shader.
+        //! Provides configuration data about which shader variants should be used to create a ShaderVariantTreeAsset.
+        //!
+        //! REMARK:  These files will be named <Shader Name>.hashedvariantlist.
+        //! Using the name and the subpath of this file We can figure out the name of the
+        //! *.shader file.
         struct HashedVariantListSourceData
         {
             AZ_TYPE_INFO_WITH_NAME_DECL(HashedVariantListSourceData);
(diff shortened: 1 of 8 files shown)


Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantListBuilder.cpp
```diff
@@ -0,0 +1,586 @@
...
+                serializeContext->Class<HashedVariantInfo>()
+                    ->Version(1) // Added Radeon GPU Analyzer
+                    ->Field("VariantInfo", &HashedVariantInfo::m_variantInfo)
+                    ->Field("Hash", &HashedVariantInfo::m_hash)
+                    ->Field("IsNew", &HashedVariantInfo::m_isNew)
+                    ;
+
+                serializeContext->Class<HashedVariantListSourceData>()
+                    ->Version(1)
+                    ->Field("timeStamp", &HashedVariantListSourceData::m_timeStamp)
```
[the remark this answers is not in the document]
galibzon: Fixed.

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantListBuilder.cpp
```diff
@@ -0,0 +1,586 @@
...
+                hashToStableIdMap.emplace(optionValuesHash, variantInfo.m_stableId);
+            
+                HashedVariantListSourceData::HashedVariantInfo hashedVariantInfo;
+                hashedVariantInfo.m_isNew = true;
+                hashedVariantInfo.m_variantInfo = variantInfo;
+                hashedVariantInfo.CalculateHash(optionValuesHash);
+                hashedVariantList.m_hashedVariants.emplace_back(AZStd::move(hashedVariantInfo));
+            }
+
+            AZStd::chrono::steady_clock::time_point now = AZStd::chrono::steady_clock::now();
```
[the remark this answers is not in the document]
galibzon: You are right. Changed it to system_clock.

Gems/Atom/Asset/Shader/Code/Source/Editor/HashedVariantListSourceData.h
```diff
@@ -0,0 +1,71 @@
...
+            // A time stamp is necessary, because building shader variants takes time.
+            // We calculate hashes to figure out the variants that changed and based on the match
+            // we set HashedVariantInfo::m_isNew to false or true.
+            // Imagine a user makes changes to a .shadervariantlist, and some HashedVariantInfos are marked as new.
+            // then within a few seconds later they make another change to the .shadervariantlist file. Because
+            // it happened so quickly those HashedVariantInfos that were set as new would become "old" and those
+            // shader variants won't be compiled.
+            // This timestamp comes to the rescue and we can measure if the change happened too quick
+            // and it such case we preserve the state of each HashedVariantInfo::m_isNew.
+            long long m_timeStamp;
```
santorac, line 64: I'm a little concerned this timestamp concept will have weird edge cases. I wonder if something like this would work, instead of chaining ShaderVariantListBuilder with ShaderVariantAssetBuilder....
Just have a single ShaderVariantListBuilder stand on its own. The first time a list is processed, it will output HashVariantListSourceData to the cache and will also compile all the shader variants. On subsequent updates, it will load the HashVariantListSourceData that it output last time, and check the hashes for changes. It will then compile shaders for only the new/updated variants. You won't need any IsNew flag because the same builder that caches the hashes will also check the hashes and compile the shaders.

galibzon: At face value your idea sounds good. But need to make sure there are no quirks. thinking...

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantListBuilder.cpp
```diff
@@ -0,0 +1,586 @@
...
+            hashedVariantList.m_timeStamp = now.time_since_epoch().count();
+
+            if (enableHashCompare)
+            {
+                // Time to load the previously generated *.hashedvariantlist file and compare hashes for each variantInfo.
+                // hashes that match, we'll flag as `hashedVariantInfo.m_isNew = false;` so the ShaderVariantAsset does not rebuild those.
+                AZStd::string previousHashedVariantListPath = GetAbsolutePathOfIntermediateAsset(variantListFullPath);
+                AZ_Trace(ShaderVariantListBuilderName, "Previous path found at %s\n", previousHashedVariantListPath.c_str());
+                HashedVariantListSourceData prevHashedVariantList;
+                if (AZ::RPI::JsonUtils::LoadObjectFromFile(previousHashedVariantListPath, prevHashedVariantList, 10 * 1024 * 1024))
```
gadams3, line 555: There should be some constant with an explanation for 10 * 1024 * 1024

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantListBuilder.cpp
```diff
@@ -0,0 +1,586 @@
...
+            RPI::ShaderVariantListSourceData shaderVariantList;
+            if (!RPI::JsonUtils::LoadObjectFromFile(variantListFullPath, shaderVariantList, AZStd::numeric_limits<size_t>::max()))
+            {
+                AZ_Error(ShaderVariantListBuilderName, false, "Failed to parse Shader Variant List Descriptor JSON from [%s]", variantListFullPath.c_str());
+                response.m_resultCode = AssetBuilderSDK::ProcessJobResult_Failed;
+                return;
+            }
+
+            auto resultOutcome = RPI::ShaderVariantTreeAssetCreator::ValidateStableIdsAreUnique(shaderVariantList.m_shaderVariants);
+            if (!resultOutcome.IsSuccess())
```
gadams3, line 507: nit - calling this resultOutcome is a little redundant. Maybe uniqueStableIdOutcome? If you are only checking for success, you can just move the function call into the if().

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantListBuilder.cpp
```diff
@@ -0,0 +1,586 @@
...
+                response.m_createJobOutputs.push_back(jobDescriptor);
+            }
+
+            response.m_result = AssetBuilderSDK::CreateJobsResultCode::Success;
+        }  // End of CreateJobs
+
+
+        // This function is temporarily needed until the AssetSystem APIs provides
+        // an API to locate intermediate assets.
+        static AZStd::string GetAbsolutePathOfIntermediateAsset(const AZStd::string sourceShaderVariantListAbsolutePath)
```
gadams3, line 327:
```suggestion
        static AZStd::string GetAbsolutePathOfIntermediateAsset(const AZStd::string& sourceShaderVariantListAbsolutePath)
```

Gems/Atom/Asset/Shader/Code/Source/Editor/HashedVariantListSourceData.h
galibzon: See my final conclusion below, as I answer a question from @gadams3

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantListBuilder.h
```diff
@@ -0,0 +1,71 @@
...
+
+            // Setting this to false will always rebuild all variants when a .shadervariantlist changes.
+            // Enabled by default.
+            static constexpr char EnableHashCompareRegistryKey[] = "/O3DE/Atom/Shaders/ShaderVariantListBuilder/EnableHashCompare";
+            static constexpr bool EnableHashCompareRegistryDefaultValue = true;
+
+            // If a .shadervariantlist changes several times within this time period,
+            // the IsNew status of each previous variant in .shadervariantlist will be preserved.
+            static constexpr char SuddenChangeInMinutesRegistryKey[] = "/O3DE/Atom/Shaders/ShaderVariantListBuilder/SuddenChangeInMinutes";
+            static constexpr unsigned SuddenChangeInMinutesRegistryDefaultValue = 5;
```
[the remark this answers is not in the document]
galibzon: My initial analysis, looking forward into the future was a case of a `shadervariantlist` with, let's say 10K variants.
In a **conservative scenario** the ShaderVariantListBuilder would split the original list into 10K individual intermediate assets.
Subsequently each one of those assets would become an individual asset, and if it didn't change in content then the AssetProcessor wouldn't start that job anymore. All good and solid, except for the 10K File Open/Save/Close operations.

To overcome the problem There's the idea of hashing variants, doing a single file I/O operation, BUT now we have to implement a little bit of fuzzy logic (hence the sudden change logic, etc).

Long story short based from @santorac's and your feedback, it seems the clean way moving forward for now is to simply split the assets per the  **conservative scenario** mentioned above and if there are bottle-necks we can always revisit this idea of hashes and having a single `hashedvariantlist` with hashes per variant to avoid redundant asset compilation.

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantListBuilder.h
```diff
@@ -0,0 +1,71 @@
...
+
+            ShaderVariantListBuilder() = default;
+            ~ShaderVariantListBuilder() = default;
+
+            // Asset Builder Callback Functions ...
+            void CreateJobs(const AssetBuilderSDK::CreateJobsRequest& request, AssetBuilderSDK::CreateJobsResponse& response) const;
+            void ProcessJob(const AssetBuilderSDK::ProcessJobRequest& request, AssetBuilderSDK::ProcessJobResponse& response) const;
+
+            // AssetBuilderSDK::AssetBuilderCommandBus interface overrides ...
+            void ShutDown() override { };
```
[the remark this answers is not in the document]
galibzon: will do

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantListBuilder.cpp
```diff
@@ -0,0 +1,586 @@
...
+            {
+                AZ_Error(ShaderVariantListBuilderName, false, "Failed to parse Shader Variant List Descriptor JSON from [%s]", variantListFullPath.c_str());
+                response.m_result = AssetBuilderSDK::CreateJobsResultCode::Failed;
+                return;
+            }
+
+            // There can be several <ShaderName>[_<*>].shadervariantlist files, if they share the same <ShaderName> then
+            // we will submit a job only on behalf of the file which is named as <ShaderName>.shadervariantlist, later
+            // in ProcessJob we'll merge all <ShaderName>[_<*>].shadervariantlist and eventually
+            // generate a single Intermediate Source Asset called <ShaderName>.hashedvariantlist.
```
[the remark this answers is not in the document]
galibzon: Will add explanation. the idea is to have "layers" of variants that can be produced by different means, and the later merged by the ShaderVariantListBuilder. The main "layer" would be <ShaderName>.shadervariantlist, the "sublayers" would be the files named <ShaderName>_<something>.shadervariantlist.

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantListBuilder.h
```diff
@@ -0,0 +1,71 @@
...
+
+#include "ShaderBuilderUtility.h"
+
+namespace AZ
+{
+    namespace ShaderBuilder
+    {
+        struct AzslData; 
+
+        class ShaderVariantListBuilder
```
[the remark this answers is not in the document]
galibzon: Will do

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantListBuilder.cpp
galibzon: This code has been removed in the latest commit.

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantListBuilder.cpp
galibzon: Done. Changed it to uniqueStableIdOutcome.

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantListBuilder.cpp
galibzon: Function was removed in the last commit.

gadams3 (approved):
When this time, if there's any overlap with the other builders using intermediate assets, we should consolidate some of the common or utility functions.

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantListBuilder.cpp
```diff
@@ -0,0 +1,514 @@
...
+        //! The "Higher Precedence" expected valid location for the .shadervariantlist would be:
+        //!     - <GameProject>/ShaderVariants/Materials/Types/StandardPBR_ForwardPass.shadervariantlist.
+        //! The "Lower Precedence" valid location would be:
+        //!     - @gemroot:Atom_Feature_Common@/Assets/Materials/Types/StandardPBR_ForwardPass.shadervariantlist.
+        //! @shouldExitEarlyFromProcessJob [out] Set to true if ProcessJob should do no work but return successfully.
+        //!     Set to false if ProcessJob should do work and create assets.
+        //!     When @shaderVariantListAbsolutePath is provided by a Gem/Feature instead of the Game Project
+        //!     We check if the game project already defined the shader variant list, and if it did it means
+        //!     ProcessJob should do no work, but return successfully nonetheless.
+        static bool ValidateShaderVariantListFolder(const AZStd::string& shaderVariantListAbsolutePath,
```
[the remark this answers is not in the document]
galibzon: This one is only relevant for `shadervariantlist` files. In particular because it cares about the subdirectory `ShaderVariants/`.

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantListBuilder.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+        static constexpr char ShaderVariantListBuilderName[] = "ShaderVariantListBuilder";
+
+        //! Adds source file dependencies for every place a referenced file may appear, and detects if one of
+        //! those possible paths resolves to the expected file.
+        //! @param variantListFullpath - The full path to the shader variant list file being processed.
+        //! @param originalShaderPath - The path to a *.shader file as described inside the shader variant list file.
+        //! @param sourceFileDependencies - new source file dependencies will be added to this list
+        //! @return absolute path of the shader file, if it exists, otherwise, an empty string.
+        static AZStd::string GetSourceShaderAbsolutePath(
```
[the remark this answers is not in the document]
galibzon: Yes, potentially the Material Type Builder would benefit.

Gems/Atom/Asset/Shader/Code/Source/Editor/ShaderVariantListBuilder.cpp
```diff
@@ -0,0 +1,514 @@
...
+            bool success = false;
+            AzToolsFramework::AssetSystemRequestBus::BroadcastResult(success
+                , &AzToolsFramework::AssetSystem::AssetSystemRequest::GenerateRelativeSourcePath
+                , shaderAbsolutePath, shaderProductFileRelativePath, scanFolderFullPath);
+            if (!success)
+            {
+                AZ_Error(ShaderVariantListBuilderName, false, "Couldn't get the scan folder for shader [%s]", shaderAbsolutePath.c_str());
+                return false;
+            }
+            AZ_Trace(ShaderVariantListBuilderName, "For shader [%s], Scan folder full path [%s], relative file path [%s]", shaderAbsolutePath.c_str(), scanFolderFullPath.c_str(), shaderProductFileRelativePath.c_str());
```
amzn-mike, line 146: I think you still need \n at the end for AZ_Trace but I haven't used it yet to confirm

amzn-mike (approved):

PR #14817, opened 2023-02-25
Multiplayer Editor: Tether ServerLauncher Lifetime to Editor

### LINKED ISSUE #14586: As a Dev I want Editor to Close/Warn of Existing Rogue ServerLauncher
Multiplayer Editor simulation will fail if a server is already running. Unless _editorsv_launch_ is false, all existing server launchers should be shut down before starting up a new server when pressing CTRL+G

If the editor is entering into a multiplayer simulation, search for other applications matching the server name and display a popup in Editor is an existing server exists. Provide a prompt to forcefully shutdown the rogue server processes
This would work both when editorsv_clientserver is enabled or not.

AMZN-Gene:
- Adding an option to process launch info to tether the lifetime of the process to the parent (only implemented on Windows and Linux)
- Tethering the editor-server’s lifetime to the Editor
- Find and warns user about existing ServerLaunchers.exe processes before launching a new one for the editor (only implemented on Windows)
Fixes #14586 

## How was this PR tested?
Open a few server launchers by hand, then run CTRL+G in Editor and see a warning log before Launching a fresh editor-server.
Shut down the editor in task manager or ALT+F4 and see the server also terminates (Tested on both WIndows/Linux)


AMZN-Olex (commented):
This needs to be PAL-fied.

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
```diff
@@ -38,11 +38,14 @@
 #include <AzToolsFramework/ViewportSelection/EditorSelectionUtil.h>
 #include <Atom/RPI.Public/RPISystemInterface.h>
 #include <AzCore/Settings/SettingsRegistryMergeUtils.h>
-#include <AzFramework/Entity/EntityDebugDisplayBus.h>
 
 #include <QMenu>
 #include <QAction>
 
+#ifdef _WIN32
+#include <tlhelp32.h>
```
AMZN-Olex, line 46: This needs to be PAL-fied.

kberg-amzn: Agreed, it does, but this is a decent temp stop-gap for Windows..

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
```diff
@@ -576,7 +579,35 @@ namespace Multiplayer
                     remoteAddress.c_str()) return;
             }
 
-            AZ_Printf("MultiplayerEditor", "Editor is listening for the editor-server...")
+            #ifdef _WIN32
+                // Find and terminate all ServerLauncher processes before launching a new one for the editor
```
kberg-amzn, line 583: Do people manually launch servers much? Will this interact weirdly in that case?

AMZN-Gene: This helps in 2 situations. 1) I manually launch servers often and then forget to tear it down before running editor CTRL+G. 2) The Editor crashes often and leaves its rogue server running and hidden as a background process.

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
```diff
@@ -576,7 +576,20 @@ namespace Multiplayer
...
-            AZ_Printf("MultiplayerEditor", "Editor is listening for the editor-server...")
+            // Terminate any existing server launchers before launching a new one.
+            // It's possible for a rogue server launcher to exist if the Editor shutdown unexpectedly while running a previous multiplayer session.
+            // It's also common to open ServerLaunchers by hand for testing, but then to forget to shut it down before starting the editor play mode.
+            const AZStd::string serverExeFilename(AZ::Utils::GetProjectName() + ".ServerLauncher" + AZ_TRAIT_OS_EXECUTABLE_EXTENSION);
+            if (AzFramework::ProcessUtils::TerminateProcess(serverExeFilename))
+            {
+                AZ_Warning("MultiplayerEditorSystemComponent", false,
+                    "Terminating an existing server launcher (%s) before opening a new server. "
+                    "If your intention was to connect to this server instead of automatically launching one from the Editor set editorsv_launch = false.",
```
mbalfour-amzn, line 587: Nice! This is a helpful explanation!

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
```diff
@@ -576,7 +576,20 @@ namespace Multiplayer
                     remoteAddress.c_str()) return;
             }
 
-            AZ_Printf("MultiplayerEditor", "Editor is listening for the editor-server...")
+            // Terminate any existing server launchers before launching a new one.
+            // It's possible for a rogue server launcher to exist if the Editor shutdown unexpectedly while running a previous multiplayer session.
+            // It's also common to open ServerLaunchers by hand for testing, but then to forget to shut it down before starting the editor play mode.
+            const AZStd::string serverExeFilename(AZ::Utils::GetProjectName() + ".ServerLauncher" + AZ_TRAIT_OS_EXECUTABLE_EXTENSION);
+            if (AzFramework::ProcessUtils::TerminateProcess(serverExeFilename))
```
lumberyard-employee-dm, line 583: This is not going to link on other platforms.

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
```diff
@@ -576,7 +576,20 @@ namespace Multiplayer
                     remoteAddress.c_str()) return;
             }
 
-            AZ_Printf("MultiplayerEditor", "Editor is listening for the editor-server...")
+            // Terminate any existing server launchers before launching a new one.
+            // It's possible for a rogue server launcher to exist if the Editor shutdown unexpectedly while running a previous multiplayer session.
+            // It's also common to open ServerLaunchers by hand for testing, but then to forget to shut it down before starting the editor play mode.
+            const AZStd::string serverExeFilename(AZ::Utils::GetProjectName() + ".ServerLauncher" + AZ_TRAIT_OS_EXECUTABLE_EXTENSION);
```
lumberyard-employee-dm, line 582: This is making a big assumption that the filename of the ServerLauncher is renamed on disk.

lumberyard-employee-dm (changes requested):

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
```diff
@@ -576,7 +576,20 @@ namespace Multiplayer
...
+            if (AzFramework::ProcessUtils::TerminateProcess(serverExeFilename))
+            {
+                AZ_Warning("MultiplayerEditorSystemComponent", false,
+                    "Terminating an existing server launcher (%s) before opening a new server. "
+                    "If your intention was to connect to this server instead of automatically launching one from the Editor set editorsv_launch = false.",
+                    serverExeFilename.c_str());
+            }
+
+            AZ_Printf("MultiplayerEditor", "Editor is listening for the editor-server...");
```
lumberyard-employee-dm, line 591: `AZ_Printf` does not automatically add a newline to the end of its output.
So this call needs to add a newline.
```suggestion
            AZ_Printf("MultiplayerEditor", "Editor is listening for the editor-server...\n");
```

lumberyard-employee-dm:
I suggest investigating the ap_tether_lifetime option that works correctly on our host platforms of [Windows](https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/Platform/Windows/AzFramework/Asset/AssetSystemComponentHelper_Windows.cpp#L146-L150) and [Linux](https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/Platform/Linux/AzFramework/Asset/AssetSystemComponentHelper_Linux.cpp#L118-L126) to terminate the process when the parent process terminates.

That logic can be exposed to the ProcessWatcher API in AzFramework.

Furthermore the [ProcessWatcher](https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/Platform/Windows/AzFramework/Process/ProcessWatcher_Win.cpp#L337-L358) already has a [TerminateProcess](https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/AzFramework/Process/ProcessWatcher.h#L84-L85) function that works for all platforms

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
AMZN-Gene: Yep! I wanted to see if this was a good place to put it before implementing the other platforms. I was worried there may have been a rejection of AzFramework/Process/ProcessUtil.cpp.

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
AMZN-Gene: Not sure I follow. The ServerLauncher is typically <game>.ServerLauncher.exe.

AMZN-Gene:
> I suggest investigating the ap_tether_lifetime option that works correctly on our host platforms of [Windows](https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/Platform/Windows/AzFramework/Asset/AssetSystemComponentHelper_Windows.cpp#L146-L150) and [Linux](https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/Platform/Linux/AzFramework/Asset/AssetSystemComponentHelper_Linux.cpp#L118-L126) to terminate the process when the parent process terminates.
> 
> That logic can be exposed to the ProcessWatcher API in AzFramework.
> 
> Furthermore the [ProcessWatcher](https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/Platform/Windows/AzFramework/Process/ProcessWatcher_Win.cpp#L337-L358) already has a [TerminateProcess](https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/AzFramework/Process/ProcessWatcher.h#L84-L85) function that works for all platforms

Users can open a server by hand outside of the Editor. We want to shut that down.
We do call ProcessWatcher::TerminateProcess, but can't if the Editor crashes. I think the ap_tether_lifetime would work even Edior crashes; this should be implemented at least to have proper house-keeping, but we're still left with the first problem.

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
AMZN-Gene: But why does it newline in the log?
image

lumberyard-employee-dm: Because the log file goes through a different path code path where it adds a [missing newline](https://github.com/o3de/o3de/blob/development/Code/Legacy/CrySystem/Log.cpp#L1323-L1347), than when logged to the [CXConsole](https://github.com/o3de/o3de/blob/development/Code/Legacy/CrySystem/Log.cpp#L1358-L1373)

lumberyard-employee-dm:
> > I suggest investigating the ap_tether_lifetime option that works correctly on our host platforms of [Windows](https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/Platform/Windows/AzFramework/Asset/AssetSystemComponentHelper_Windows.cpp#L146-L150) and [Linux](https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/Platform/Linux/AzFramework/Asset/AssetSystemComponentHelper_Linux.cpp#L118-L126) to terminate the process when the parent process terminates.
> > That logic can be exposed to the ProcessWatcher API in AzFramework.
> > Furthermore the [ProcessWatcher](https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/Platform/Windows/AzFramework/Process/ProcessWatcher_Win.cpp#L337-L358) already has a [TerminateProcess](https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/AzFramework/Process/ProcessWatcher.h#L84-L85) function that works for all platforms
> 
> Users can open a server by hand outside of the Editor... we want to shut that down. Would ap_tether_lifetime fix that too? We do call ProcessWatcher::TerminateProcess, but can't if the Editor crashes. If ap_tether_lifetime works even when Editor crashes the Multiplayer gem should also use that to have proper house-keeping, but we're still left with the first problem.

That seems quite intrusive and kind of a like malware, if another application can close process, that it didn't open.
I would understand the Editor terminating the ServerLauncher if it launched it, but my opinion is that it should be terminating the ServerLauncher process if it wasn't responsible for launching it.

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
AMZN-Gene: Not sure what we could do to prevent that. Is there a way to tag processes?

AMZN-Gene:
> That seems quite intrusive and kind of a like malware, if another application can close process, that it didn't open.
> I would understand the Editor terminating the ServerLauncher if it launched it, but my opinion is that it should be terminating the ServerLauncher process if it wasn't responsible for launching it.

Fixed! I'll settle for a warning. The warning might be lost in a sea of logs, but it's something and I can use it as a jumping off point for better UX. Changed _AzFramework::ProcessUtils::TerminateProcess()_ to _ProcessCount()_

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
AMZN-Gene: No longer terminating based on name. I’ll warn users based on name and also tether the server lifetime to the editor.

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
```diff
@@ -576,7 +577,21 @@ namespace Multiplayer
                     remoteAddress.c_str()) return;
             }
 
-            AZ_Printf("MultiplayerEditor", "Editor is listening for the editor-server...")
+            // Find any existing server launchers before launching a new one.
+            // It's possible for a rogue server launcher to exist if the Editor shutdown unexpectedly while running a previous multiplayer session.
+            // It's also common to open ServerLaunchers by hand for testing, but then to forget to shut it down before starting the editor play mode.
+            const AZStd::string serverExeFilename(AZ::Utils::GetProjectName() + ".ServerLauncher" + AZ_TRAIT_OS_EXECUTABLE_EXTENSION);
+            int existingServers = AzFramework::ProcessUtils::ProcessCount(serverExeFilename);
```
mbalfour-amzn, line 584: Won't this cause link errors on Mac/Linux since there aren't any stub or real implementations of ProcessCount() in this PR?

lumberyard-employee-dm: @AMZN-Gene  You need to implement PAL files for for Linux and Mac that returns 0.

AMZN-Gene: Added stubs to return 0. Also added implemented tethering to work on Linux.

lumberyard-employee-dm (approved):

Issue #10825, opened 2022-07-22
Physics::MaterialAsset being registered with the AssetCatalog multiple times in the AssetBuilder 


lumberyard-employee-dm:
**Describe the bug**
The Physics::MaterialAsset is being registered with the AssetCatalog multiple times leading to an error message being output in the Asset builder

```
[2022-07-22T06:17:55.311Z]   AssetBuilder: E: AssetDatabase: Trace::Error

[2022-07-22T06:17:55.311Z]   AssetBuilder: E: >	D:/workspace/o3de/Code/Framework/AzCore/AzCore/Asset/AssetManager.cpp(717): 'void __cdecl AZ::Data::AssetManager::RegisterCatalog(class AZ::Data::AssetCatalog *,const struct AZ::Uuid &)'

[2022-07-22T06:17:55.311Z]   AssetBuilder: E: >	Asset type {E4EF58EE-B1D1-46C8-BE48-BB62B8247386} already has a catalog registered! New registration ignored!

[2022-07-22T06:17:55.311Z]   AssetBuilder: E: AssetDatabase: Trace::Error

[2022-07-22T06:17:55.311Z]   AssetBuilder: E: >	D:/workspace/o3de/Code/Framework/AzCore/AzCore/Asset/AssetManager.cpp(648): 'void __cdecl AZ::Data::AssetManager::RegisterHandler(class AZ::Data::AssetHandler *,const struct AZ::Uuid &)'

[2022-07-22T06:17:55.311Z]   AssetBuilder: E: >	Asset type {E4EF58EE-B1D1-46C8-BE48-BB62B8247386} already has a handler registered! New registration ignored!
```

Each time the MaterialSystemComponent activates it registers the PhysicsMaterial with the AssetCatalog through the GenericAssetHandler::Register function, but during deactivation it does NOT invoke the Unregister function.
That can be seen [here](https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/AzFramework/Physics/Material/PhysicsMaterialSystemComponent.cpp#L58-L68)

Also the [Physics::Material](https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/AzFramework/Physics/Material/PhysicsMaterial.h#L18-L21) class incorrectly specifies the Physics::MaterialAsset in RTTI and that would need to be fixed as well

**Steps to reproduce**
1. Run the AssetProcessorBatch with the project-path set to the location of the AutomatedTesting project
[...]

PR #14824, opened 2023-02-27
Typeinfo simplification

lumberyard-employee-dm:
## What does this PR do?

Simplified AzTypeInfo to reduce Template instantiations and improve build time.
Frontend compiler build time has improved by about 7.5% and overall build time has improved about 2%(CodeGen is slower)
The CodeGen can be time can be improved by iterative using the new AZ_TYPE_INFO and AZ_RTTI Declaration and Implementation macros to define the AzTypeInfo and Rtti virtuals functions in a cpp file.

Instead of specializing AzTypeInfo<T>, classes now opt in to type info support by overloading a `GetO3deTypeName, and `GetO3deTypeId` in their class namespace or in the AZ namespace.

The `AZ_TYPE_INFO*` macros have been updated to declare and define free functions those functions.
In additional if the `AZ_TYPE_INFO*` macros supply additional parameters, the `GetO3deClassTemplateId` function will also be overloaded which associates a Uuid with any specialization class template.
For example `AZStd::vector<int>` and `AZStd::vector<AZ::Entity*>` will have the same class template id

The following is an example of how to manually add AzTypeInfo for a type
```c++
namespace Particles
{
    class Effect;
    // ...
    // Function which returns the name to use for the type
    AZ::TypeNameString GetO3deTypeName(AZ::Adl, AZStd::type_identity<Effect>);
    // Function which returns the Uuid to use for the type
    AZ::TypeId GetO3deTypeId(AZ::Adl, AZStd::type_identity<Effect>);
}
```

NOTE: the functions can be defined in a cpp file as well and don't have to be inline within a header

Moved all the AZ_RTTI macros into their own RTTI_Macros header.

Added new user facing AZ_RTTI macros
* `AZ_RTTI_NO_TYPE_INFO_DECL` -> Declares the virtual functions(RTTI_Type, RTTI_TypeName, etc...). Pairs with `AZ_RTTI_NO_TYPE_INFO_IMPL` to implement RTTI within a cpp file
* `AZ_RTTI_NO_TYPE_INFO_IMPL` -> Defines the RTTI virtual functions(RTTI_Type, RTTI_TypeName, etc...)
* `AZ_RTTI_NO_TYPE_INFO_IMPL_INLINE` -> Same as `AZ_RTTI_NO_TYPE_INFO_IMPL`, with the addition of the `inline` keyword in front of each definition. This is suitable for adding out of line definitions for a class template in an inline or header file.
* `AZ_RTTI_WITH_NAME` -> Same as the AZ_RTTI macro, accept a custom name can be given for the type

Added new user facing AZ_TYPE_INFO macros for declaring and defining in separate locations
* `AZ_TYPE_INFO_WITH_NAME` - Same as AZ_TYPE_INFO, but allows a custom type name to be associated with the Uuid
* `AZ_TYPE_INFO_WITH_NAME_DECL` - Adds non-member friend functions declarations of the GetO3deTypeName and GetO3deTypeId functions for the specified class.
* `AZ_TYPE_INFO_WITH_NAME_IMPL` - Adds non-member friend functions definitions of the GetO3deTypeName and GetO3deTypeId functions for the specified class. The type name and Uuid is specified in this macro. This is useful for defining RTTI in a cpp file.
* `AZ_TYPE_INFO_WITH_NAME_IMPL_INLINE` - Same as `AZ_TYPE_INFO_WITH_NAME_IMPL` with the addition of marking the function definitions with the `inline` keyword. Useful for defining TypeInfo in an inline or header file .
* `AZ_TYPE_INFO_SPECIALIZE_WITH_NAME_DECL` - Can be used to add free functions declarations of the GetO3deTypeName and GetO3deTypeId for any type be that a class, enum or fundamental type(`int`, `bool`, etc...)
* `AZ_TYPE_INFO_SPECIALIZE_WITH_NAME_IMPL` - Adds free function definitions of the GetO3deTypeName and GetO3deTypeId functions. This pairs well with the `AZ_TYPE_INFO_SPECIALIZE_WITH_NAME_DECL` macro
* `AZ_TYPE_INFO_SPECIALIZE_WITH_NAME_IMPL_INLINE` - Same as the `AZ_TYPE_INFO_SPECIALIZE_WITH_NAME_IMPL` and adds `inline` keyword in front of the functions
* `AZ_TYPE_INFO_TEMPLATE_WITH_NAME_DECL` - Adds free function declarations of GetO3deTypeName, GetO3deTypeId and GetO3deClassTemplateId for a class template. Also adds a declaration of GetO3deTemplateId for use with the class template directly(no-specialization i.e `AZStd::vector`, not `AZStd::vector<int>`)
* `AZ_TYPE_INFO_TEMPLATE_WITH_NAME_IMPL` - Adds free function declarations of GetO3deTypeName, GetO3deTypeId and GetO3deClassTemplateId for a class template. Also adds a declaration of GetO3deTemplateId for use with the class template.
* `AZ_TYPE_INFO_TEMPLATE_WITH_NAME_IMPL_INLINE` - Same as `AZ_TYPE_INFO_TEMPLATE_WITH_NAME_IMPL` and adds the `inline` keyword in front of the `GetO3de*` functions
* `AZ_TYPE_INFO_TEMPLATE_WITH_NAME_INSTANTIATE` - Allows explicit instantiation the `GetO3deTypeName`, `GetO3deTypeId` and `GetO3deClassTemplateId` functions for a class template specialization. Useful for restricting the TypeInfo support for a template type. For example the `AZStd::basic_string` class template only provides a few instantiations of type info such as `AZStd::string` and `AZ::OSString`. This should be used in a cpp file to guarantee only a single instantiation is crated
* `AZ_TEMPLATE_INFO_WITH_NAME` - Provides a definition of the `GetO3deTemplateId` function for a class template. This allows the Uuid associated with the template to be queried using `AZ::AzGenericTypeInfo::Uuid<ClassTemplate>`

AZ TypeInfo definitions of fundamental types, AZStd and string based types to the TypeInfo.cpp file to reduce their inline type info function count when compiling

Moved all SerializeContext, ReflectContext and BehaviorContext AZ_TYPE_INFO and AZ_RTTI definitions to the their cpp files if to reduce type info function compilation.

Moved PoolAllocatorHelper template TypeInfo to the PoolAllocator cpp file to reduce compile time.

Moved SystemAllocator TypeInfo to SystemAllocator.cpp

Removed `ranges_algorithm.h` and `ranges_to.h` dependency from Uuid.inl to reduce compile time building the AZ::Uuid code.

Fixed Incorrect usages of AZ_TYPE_INFO and AZ_RTTI through the codebase (invalid type names, duplicate uuids, **correct type names but in the wrong namespace**, etc...)

Fixed the AZ_FOR_EACH/AZ_FOR_EACH_BIND1ST macro to call joined macro names in the second pass over a macro. Previously if the AZ_MACRO_CALL macro was used anywhere on the way to the AZ_FOR_EACH macro, the expansion of the second AZ_MACRO_CALL would be suppressed.

### Limitations

Using the AZ_TYPE_INFO macro within a local class is no longer supported(i.e declaring a class within a function). 
The only uses of this was in the ObjectStream Serialization Unit Test
```
TEST_F(ObjectStreamSerialization, LoadNonDeprecatedElement_FollowedByZeroSizeDeprecatedElement_DoesNotAssert)
    {
        struct EmptyDeprecatedClass
        {
            //! Not supported
            AZ_TYPE_INFO(EmptyDeprecatedClass, "{73890A64-9ADB-4639-B0E0-93294CE81B19}");
        };
```

The solution is just move the local class outside of the function.

## How was this PR tested?

Used [ClangBuildAnalyzer](https://github.com/aras-p/ClangBuildAnalyzer) to measure compilation time using the Windows clang build

Current measurements
> **** Time summary:
Compilation (2628 times):
  Parsing (frontend):        23253.0 s
  Codegen & opts (backend):  24228.4 s

Measurement with changes**** Time summary:
Compilation (2627 times):
  Parsing (frontend):        21526.7 s
  Codegen & opts (backend):  25107.9 s

Used a couple of powershell scripts to automate running the ClangBuildAnalyzer and kicking off O3DE builds
[analyzer-no-changes.ps1.txt](https://github.com/o3de/o3de/files/10836120/analyzer-no-changes.ps1.txt)
[analyzer-with-changes.ps1.txt](https://github.com/o3de/o3de/files/10836119/analyzer-with-changes.ps1.txt)


Gems/PhysX/Code/Include/PhysX/MeshAsset.h
```diff
@@ -28,7 +28,7 @@ namespace PhysX
         {
         public:
             AZ_CLASS_ALLOCATOR(AssetColliderConfiguration, AZ::SystemAllocator);
-            AZ_TYPE_INFO(MeshAssetData, "{463AA6A7-8A1A-42B6-B103-F6939CC7A8A5}");
```
moraaar, line 31: Good catch!

Code/Framework/AzFramework/AzFramework/Physics/Material/PhysicsMaterialSystemComponent.cpp
```diff
@@ -61,6 +61,14 @@ namespace Physics
...
     void MaterialSystemComponent::Deactivate()
     {
+        for (auto& assetHandler : m_assetHandlers)
+        {
+            if (auto materialAssetHandler = azrtti_cast<AzFramework::GenericAssetHandler<MaterialAsset>*>(assetHandler.get());
+                materialAssetHandler != nullptr)
+            {
+                materialAssetHandler->Unregister();
+            }
+        }
```
moraaar, line 71: Good catch! This might be related to this issue https://github.com/o3de/o3de/issues/10825 

This means there is also a `Unregister` missing for [this one too](https://github.com/o3de/o3de/blob/76074304573950182e2d48283ff4b01b798bab03/Gems/PhysX/Code/Editor/Source/Components/EditorSystemComponent.cpp#L81), can you add it as well please? Thanks!

moraaar (commented):
Reviewed the changes to physics parts of this PR, which look good. The rest I'm going to lean on other people from sig-core to review.

Code/Framework/AzFramework/AzFramework/Physics/Material/PhysicsMaterialSystemComponent.cpp
moraaar: Thanks!

lumberyard-employee-dm: Done: https://github.com/o3de/o3de/pull/14824/commits/d5962dd52fddb8e36c668e715dc26610799f19e1

lawsonamzn (approved):
Any time we've touched RTTI or type info in the past we've had to fight serialize/deserialize or old data issues.  IDeally, AR would surface this, but some manual testing of existing data (load some script canvas files, emotionfx, etc) would go a long way.

lumberyard-employee-dm:
> Any time we've touched RTTI or type info in the past we've had to fight serialize/deserialize or old data issues. IDeally, AR would surface this, but some manual testing of existing data (load some script canvas files, emotionfx, etc) would go a long way.

Actually I hit some issues while testing the changes with ScriptCanvas JSON Serialization of NodeFunctionGeneric nodes.
Older ScriptCanvas files that were using JSON serialization with a truncated type name, were failing to have their type names matched in the SerializeContext
Found out that there was some bugs in how the Deprecated Name logic, which I have since corrected in the DeprecatedName logic for function pointers (https://github.com/o3de/o3de/blame/0f0d20603645feac8263ec041dcbb38e6295c6df/Code/Framework/AzCore/AzCore/RTTI/TypeInfo.h#L589-L647)

For example the [ScriptCanvas_ShapeCast.scriptcanvas](https://github.com/aws-lumberyard-dev/o3de/blob/behavior-context-refactor/AutomatedTesting/ScriptCanvas/ScriptCanvas_ShapeCast.scriptcanvas#L166) file stores the type name of the NodeFunctionGeneric node associated with the [SphereCastWithGroup](https://github.com/aws-lumberyard-dev/o3de/blob/behavior-context-refactor/Gems/ScriptCanvas/Code/Include/ScriptCanvas/Libraries/Deprecated/PhysX/WorldNodes.h#L288-L301) C++ function.

The deprecated type name were calculated differently and so the logic to calculate those names in the old manner has been replicated. The old logic involves making sure that `const` keyword is place before the type name, etc...

gadams3 (approved):
Surprised at how many types were not even set up with the right name. Will changing or removing namespaces affect any of the types that were serialized with their name?

lumberyard-employee-dm:
> Surprised at how many types were not even set up with the right name. Will changing or removing namespaces affect any of the types that were serialized with their name?

For ObjectStream serialization changing the name will not have an effect on the content.
For JSON serialization, if the typename is serialized using the StoreType::Yes option through the [JsonSerializer::Store*](https://github.com/aws-lumberyard-dev/o3de/blob/behavior-context-refactor/Code/Framework/AzCore/AzCore/Serialization/Json/JsonSerializer.cpp#L95) parameter to write `$type` variables, then this affects those types.

However the mitigation for that was the addition of the `AZ_TYPE_INFO_WITH_NAME` macro that allows specifying the name that should be written when writing out a type id.

Any place in the Codebase that serialized one of the corrected typenames to JSON were replaced with the AZ_TYPE_INFO_WITH_NAME macro where the old name is being specified.

PR #14830, opened 2023-02-27
Action Manager | Refactor EditorActionsHandler::UpdateRecentFileActions to prevent crash.

AMZN-daimini:
## What does this PR do?

Refactors EditorActionsHandler::UpdateRecentFileActions to only check a file's validity when the index is in bounds, preventing a crash.

## How was this PR tested?

Verified the crash no longer occurs locally.

### DIFF
--- a/Code/Editor/Core/EditorActionsHandler.cpp
+++ b/Code/Editor/Core/EditorActionsHandler.cpp
@@ -2091,13 +2091,13 @@ bool EditorActionsHandler::IsRecentFileActionActive(int index)
 
 bool EditorActionsHandler::IsRecentFileEntryValid(const QString& entry, const QString& gameFolderPath)
 {
-    QFileInfo info(entry);
-    if (!info.exists())
+    if (entry.isEmpty())
     {
         return false;
     }
 
-    if (entry.isEmpty())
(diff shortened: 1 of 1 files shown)


galibzon (approved):

Code/Editor/Core/EditorActionsHandler.cpp
```diff
@@ -2169,34 +2169,44 @@ void EditorActionsHandler::UpdateRecentFileActions()
     const QString gameDirPath = gameDir.absolutePath();
 
     m_recentFileActionsCount = 0;
-    int counter = 0;
+
+    int index = 0;
 
     // Update all names
-    for (int index = 0; (index < maxRecentFiles) || (index < recentFilesSize); ++index)
+    for (int counter = 0; counter < maxRecentFiles; ++counter)
```
hultonha, line 2176: This is super minor but you could do...

```
for (int counter = 0, index = 0; counter < maxRecentFiles; ++counter)
```

to keep the counter local to the loop, but it's not a huge deal

hultonha (approved):
LGTM 👍

dtamkin1 (approved):
It looks like a community member is also trying to fix this

PR #14832, opened 2023-02-27
Fix a crash if the number of recently opened files is < 10

kh-huawei:
## What does this PR do?

Fixes a crash in the File Menu when the number of recently opened files is less that 10.

## How was this PR tested?

Clicking on the file-menu with 3 recently opened files.

### DIFF
--- a/Code/Editor/Core/EditorActionsHandler.cpp
+++ b/Code/Editor/Core/EditorActionsHandler.cpp
@@ -2172,7 +2172,7 @@ void EditorActionsHandler::UpdateRecentFileActions()
     int counter = 0;
 
     // Update all names
-    for (int index = 0; (index < maxRecentFiles) || (index < recentFilesSize); ++index)
+    for (int index = 0; (index < maxRecentFiles) && (index < recentFilesSize); ++index)
     {
         if (!IsRecentFileEntryValid((*recentFiles)[index], gameDirPath))
         {


sphrose (commented):
It looks to me as if this function is supposed to overrun the recentFiles list, and display placeholders for the remaining entries. Wouldn't a more sympathetic fix be to move the `IsRecentFileEntryValid` call inside the `if (index < recentFilesSize` conditional?

AMZN-daimini:
I have merged a fix for the issue less than 50 minutes ago: https://github.com/o3de/o3de/pull/14830
Could you pull latest and verify if it you can still reproduce? Thank you :)

kh-huawei:
> I have merged a fix for the issue less than 50 minutes ago: #14830
> Could you pull latest and verify if it you can still reproduce? Thank you :)

Can confirm that the issue is fixed. Thanks!

> It looks to me as if this function is supposed to overrun the recentFiles list, and display placeholders for the remaining entries. Wouldn't a more sympathetic fix be to move the `IsRecentFileEntryValid` call inside the `if (index < recentFilesSize` conditional?

I thought so too, but the "Recent File #xx" placeholder is also added in the OnActionRegistrationHook(), so i'm not entirely sure why it's here as well.

However, the fix from #14830 works and looks better, so i'm closing this MR.

hultonha:
Thanks for jumping on this and getting a proposed fix out so quickly @kh-huawei! I know @AMZN-daimini just beat you to it but it's much appreciated! 🙂

Issue #14800, opened 2023-02-24
Material Editor: Errors when opening material graph template materials


LB-KacperMetelski:
**Describe the bug**
Opening a Material Document type named '_Material Graph Name_' causes a pop-up with multiple errors to be displayed.

This issue occurs on both Linux and Windows. Moreover, there is a chance of the Material Editor crashing after the pop-up is dismissed.

**Steps to reproduce**
1. Open the Material Editor.
2. Create a new document by navigating to File -> New Material Document -> Material Graph Name.

**Expected behavior**
A Material Document is created without issues.

**Actual behavior**
Errors occur.

**Screenshots/Video**
![image (1)](https://user-images.githubusercontent.com/110163473/221185393-2c839a35-2fc3-4fac-9e8e-60c638288cbe.png)
![image](https://user-images.githubusercontent.com/110163473/221185399-273c2f27-af83-49c8-b94c-f855856d8cc1.png)

**Found in Branch**
Development

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
37b6a48

**Desktop/Device:**
- Device: PC
- OS: Ubuntu
- Version: 22.04.1 LTS
- CPU: I7-11700f
- GPU: Geforce RTX3070 MSI
- Memory: 32GB

LB-KacperMetelski:
Issue verified as fixed in commit 15fce3e.

PR #14833, opened 2023-02-27
Material Canvas: Updated lua scripts to check for missing options before reading them

gadams3:
## What does this PR do?

Updated base PBR and standard PBR lua scripts that come with material canvas to handle missing shader options. This should only occur when opening the material graph template materials and material types directly, since the options are only added to materials generated from them. The lua functor API should be updated to have default values passed in as a second parameter when reading options and properties.

Resolves https://github.com/o3de/o3de/issues/14800

## How was this PR tested?

Opened base and standard PBR material graph template materials in the material editor without errors.
Created new materials from material canvas and opened those without errors.
Confirmed that options still behave correctly when toggling them on generated materials.

### DIFF
--- a/Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/GraphData/Nodes/MaterialOutputs/BasePBR/HandleCastShadows.lua
+++ b/Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/GraphData/Nodes/MaterialOutputs/BasePBR/HandleCastShadows.lua
@@ -14,11 +14,13 @@ function GetMaterialPropertyDependencies()
 end
 
 function Process(context)
-    -- The only reason we need this script is when the material type uses GeneralCommonPropertyGroup.json
-    -- which does not connect general.cast_shadows, and the only way to add a connection after importing
-    -- a common .json file is to use a lua functor. If we had another way to add a connection to a property
-    -- after defining it via $import, that would probably be simpler.
-    local castShadows = context:GetMaterialPropertyValue_bool("general.cast_shadows")
+    local castShadows = true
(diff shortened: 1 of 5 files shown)


Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/GraphData/Nodes/MaterialOutputs/BasePBR/HandleCastShadows.lua
```diff
@@ -14,11 +14,13 @@ function GetMaterialPropertyDependencies()
...
-    -- after defining it via $import, that would probably be simpler.
-    local castShadows = context:GetMaterialPropertyValue_bool("general.cast_shadows")
+    local castShadows = true
+    if context:HasMaterialProperty("general.cast_shadows") then
+        castShadows = context:GetMaterialPropertyValue_bool("general.cast_shadows")
+    end
+
     local shadowMap = context:GetShaderByTag("Shadowmap")
-    shadowMap:SetEnabled(castShadows)
+    if shadowMap then
```
rgba16f, line 23: should this if check castShadows as well before enabling the shadowmap?

rgba16f (approved):

Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/GraphData/Nodes/MaterialOutputs/BasePBR/HandleCastShadows.lua
gadams3: No because it's enabling or disabling it based on the flag

invertednormal (approved):

PR #14834, opened 2023-02-27
remove caching for IsOnGround result

greerdv:
## What does this PR do?
Removes the caching behaviour for IsOnGround queries on the character gameplay controller component. The caching was intended to avoid the expense of scene queries, but can lead to incorrect results if the controller or other physics bodies move outside the physics update.

## How was this PR tested?
Unit tests and manual testing in the editor.

### DIFF
--- a/Gems/PhysX/Code/Source/PhysXCharacters/Components/CharacterGameplayComponent.cpp
+++ b/Gems/PhysX/Code/Source/PhysXCharacters/Components/CharacterGameplayComponent.cpp
@@ -110,29 +110,17 @@ namespace PhysX
     // CharacterGameplayRequestBus
     bool CharacterGameplayComponent::IsOnGround() const
     {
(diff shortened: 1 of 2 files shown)


Gems/PhysX/Code/Source/PhysXCharacters/Components/CharacterGameplayComponent.cpp
```diff
@@ -145,8 +133,7 @@ namespace PhysX
 
         if (state.touchedActor != nullptr || (state.collisionFlags & physx::PxControllerCollisionFlag::eCOLLISION_DOWN) != 0)
```
greerdv, line 134: I'm wondering if we should get rid of this check as well and go straight to the overlap query.

mbalfour-amzn (approved):

hultonha (approved):
LGTM! 👍

SergeyAMZN (approved):

PR #14843, opened 2023-02-27
Register sets engines_path for older projects

AMZN-alexpete:
## What does this PR do?

Reverts some code changed in https://github.com/o3de/o3de/pull/14270 that required each engine to be registered in `o3de_manifest.json` in an `engines_path` dictionary.

Older projects rely on this key to find the engine, so users were confused when they installed o3de from dev but their project's failed to find the engine.

This change:
1. registers the engine in the `engines_path` dictionary in the `o3de_manifest.json` if no entry exists
2. prints a warning instead of an error if a different engine is already registered with the engine name in `engines_path` 

Example:
```
$ scripts/o3de.bat register --this-engine
[INFO] o3de.register: Engine successfully registered.
[WARNING] o3de.register: Important notice regarding older projects created with engine version 22.10 or earlier:
An engine named "o3de" is already registered at "D:/git/o3de-dev"
If you have projects created with engine version 22.10 or earlier, and you want them to use this engine you will need to:
  1. backup the project's existing "cmake/EngineFinder.cmake" file
  2. copy the "Templates/DefaultProject/Template/cmake/EngineFinder.cmake" from this engine into the project.
 Alternately, to force all older projects that use "o3de" to use this engine path, specify the -f/--force option.
```

## How was this PR tested?

- ran `register --this-engine` from a new engine with an existing engine and without and verified the warning prints if an existing engine is registered with that name, otherwise no warning is printed.
- in any case, the engine is registered in the `engines` list in `o3de_manifest.json` so new projects can find it

### DIFF
--- a/scripts/o3de/o3de/register.py
+++ b/scripts/o3de/o3de/register.py
@@ -225,6 +225,60 @@ def register_all_repos_in_folder(repos_path: pathlib.Path,
     return register_all_o3de_objects_of_type_in_folder(repos_path, 'repo', remove, force, None, engine_path=engine_path)
 
 
+def remove_engine_name_to_path(json_data: dict,
+                               engine_path: pathlib.Path) -> int:
+    """
+    Remove the engine at the specified path if it exist in the o3de manifest
+    :param json_data in-memory json view of the o3de_manifest.json data
+    :param engine_path path to engine to remove from the manifest data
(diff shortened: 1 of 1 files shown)


AMZN-Phil (approved):

kberg-amzn (approved):
Awesome! This will help remove a lot of friction from the next release. 👍

PR #14846, opened 2023-02-27
Custom page/min allocation/max allocation sizes to AZ::PoolAllocator

amzn-tommy:
Moved the creation of PoolSchemaImpl to PoolSchema::Create, because otherwise the PoolSchemaImpl is constructed with the default values before the PoolAllocator's constructor runs.

## What does this PR do?

Adds support to optionally use custom page size, min allocation size, and max allocation size for a PoolAllocator.

## How was this PR tested?

Added unit tests, and used this change locally for changes in another branch.


Code/Framework/AzCore/Tests/Memory/PoolAllocatorTests.cpp
```diff
@@ -0,0 +1,125 @@
...
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+#include <AzCore/UnitTest/TestTypes.h>
+#include <AzCore/Memory/PoolAllocator.h>
+#include <AzCore/std/containers/vector.h>
+
+constexpr size_t s_testPoolPageSize = 1024;
```
lumberyard-employee-dm, line 12: These constants and classes should be placed within the UnitTest namespace.

The global namespace is not allowed for declaring types and variables

Code/Framework/AzCore/Tests/Memory/PoolAllocatorTests.cpp
```diff
@@ -0,0 +1,125 @@
...
+    void DoTestAllocations(const PoolAllocatorTestParameters& testParameters)
+    {
+        AZStd::vector<void*, AZ::AZStdAlloc<AZ::OSAllocator>> allocations;
+        const size_t totalNumberOfAllocations = testParameters.m_allocationSizes.size() * testParameters.m_numberOfAllocationsPerSize;
+        for (size_t i = 0; i < totalNumberOfAllocations; ++i)
+        {
+            const size_t allocationIndex = allocations.size();
+            const size_t allocationSize = testParameters.m_allocationSizes[allocationIndex % testParameters.m_allocationSizes.size()];
+            void* allocation = AZ::AllocatorInstance<TestPoolAllocator>::Get().Allocate(allocationSize, s_testPoolMinAllocationSize);
+            EXPECT_NE(nullptr, allocation);
```
lumberyard-employee-dm, line 67: This is a scenario where `ASSERT_NE` is more applicable as if the allocation is nullptr, the test should end.

Code/Framework/AzCore/Tests/Memory/PoolAllocatorTests.cpp
```diff
@@ -0,0 +1,125 @@
...
+        AZ::AllocatorInstance<TestPoolAllocator>::Get().DeAllocate(result);
+    }
+
+    TEST_F(PoolAllocatorDeathTestFixture, PoolAllocator_AllocateMoreThanMaxAllocationSize_FailsToAllocate)
+    {
+        // Validate that the maximum allocation size of a custom PoolAllocator is respected
+        AZ_TEST_START_ASSERTTEST;
+        AZ::IAllocator::pointer result =
+            AZ::AllocatorInstance<TestPoolAllocator>::Get().Allocate(s_testPoolMaxAllocationSize + 1, 1);
+        AZ_TEST_STOP_ASSERTTEST(1);
```
lumberyard-employee-dm, line 122: `AZ_TEST_START_ASSERTTEST` and `AZ_TEST_STOP_ASSERTTEST` macros are [deprecated](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/UnitTest/UnitTest.h#L290-L301)

The `AZ_TEST_START_TRACE_SUPPRESSION` and `AZ_TEST_STOP_TRACE_SUPPRESSION` macros should be used

Code/Framework/AzCore/Tests/Memory/PoolAllocatorTests.cpp
```diff
@@ -0,0 +1,125 @@
...
+        , public ::testing::WithParamInterface<PoolAllocatorTestParameters>
+    {
+    };
+
+    class PoolAllocatorDeathTestFixture : public PoolAllocatorTestFixture
+        {};
+
+    void DoTestAllocations(const PoolAllocatorTestParameters& testParameters)
+    {
+        AZStd::vector<void*, AZ::AZStdAlloc<AZ::OSAllocator>> allocations;
```
lumberyard-employee-dm, line 60: This is an alis for AZ::AZStdAlloc<AZ::OSAllocator> in osallocator.h called [OSStdAllocator](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Memory/OSAllocator.h#L59)
```suggestion
        AZStd::vector<void*, OSStdAllocator> allocations;
```

Code/Framework/AzCore/Tests/Memory/PoolAllocatorTests.cpp
```diff
@@ -0,0 +1,125 @@
...
+#include <AzCore/UnitTest/TestTypes.h>
+#include <AzCore/Memory/PoolAllocator.h>
+#include <AzCore/std/containers/vector.h>
+
+constexpr size_t s_testPoolPageSize = 1024;
+constexpr size_t s_testPoolMinAllocationSize = 64;
+constexpr size_t s_testPoolMaxAllocationSize = 256;
+
+// Define a custom pool allocator
+class TestPoolAllocator final : public AZ::Internal::PoolAllocatorHelper<AZ::PoolSchema>
```
lumberyard-employee-dm, line 17: This class name needs to be different to avoid issue with Unity files combining both `VectorAndArray.cpp` and `PoolAllocatorTests.cpp` into a single unit

Code/Framework/AzCore/Tests/Memory/PoolAllocatorTests.cpp
```diff
@@ -0,0 +1,125 @@
...
+{
+public:
+    AZ_CLASS_ALLOCATOR(TestPoolAllocator, AZ::SystemAllocator, 0)
+    AZ_TYPE_INFO(TestPoolAllocator, "{5AB88680-581F-4F0B-B67F-BAEAA78D3122}")
```
lumberyard-employee-dm, line 21: minor nitpick: missing semicolon after macros

lumberyard-employee-dm:
FYI I have a change up for PR that modifies PoolAllocator that shouldn't affect these changes, but is something to keep an eye out for. 
https://github.com/o3de/o3de/commit/9db0894b904cdd84f00e35ccc2a23f2c8b108523#diff-86999ab38d628aae10484dd9ea088b32cdf3039b72058670410013fd140bec22

Code/Framework/AzCore/Tests/Memory/PoolAllocatorTests.cpp
```diff
@@ -0,0 +1,125 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+#include <AzCore/UnitTest/TestTypes.h>
```
lumberyard-employee-dm, line 8: We actually have PoolAllocatorTest in Memory.cpp https://github.com/o3de/o3de/pull/14824/files#diff-66fc5113ad002d6ac22403c3f7aaa33d8b2d6050e10b61dc386f72eb0d8bffa7R669-R676

Code/Framework/AzCore/Tests/Memory/HphaAllocator.cpp
```diff
@@ -10,35 +10,38 @@
...
-static const AllocationSizeArray s_mixedAllocationSizes = { 2, s_kiloByte, 59, 4 * s_kiloByte, 128, 200 * s_kiloByte, 250, s_megaByte, 512, 2 * s_megaByte };
+    static const size_t s_kiloByte = 1024;
+    static const size_t s_megaByte = s_kiloByte * s_kiloByte;
+    using AllocationSizeArray = AZStd::array<size_t, 10>;
+    static const AllocationSizeArray s_smallAllocationSizes = { 2, 16, 20, 59, 100, 128, 160, 250, 300, 512 };
+    static const AllocationSizeArray s_bigAllocationSizes = {
+        513,        s_kiloByte,    2 * s_kiloByte, 4 * s_kiloByte, 10 * s_kiloByte, 64 * s_kiloByte, 128 * s_kiloByte, 200 * s_kiloByte,
+        s_megaByte, 2 * s_megaByte
+    };
+    static const AllocationSizeArray s_mixedAllocationSizes = { 2,   s_kiloByte, 59,  4 * s_kiloByte, 128, 200 * s_kiloByte,
```
lumberyard-employee-dm, line 42: Thanks for putting this inside of a namespace!

lumberyard-employee-dm (approved):

lumberyard-employee-dm (approved):

Code/Framework/AzCore/AzCore/Memory/PoolAllocator.cpp
```diff
@@ -665,8 +674,7 @@ namespace AZ
     // [9/15/2009]
     //=========================================================================
     PoolSchema::PoolSchema()
-        : m_impl(new (AZStd::stateless_allocator().allocate(sizeof(PoolSchemaImpl), AZStd::alignment_of<PoolSchemaImpl>::value))
-                 PoolSchemaImpl())
+        : m_impl(nullptr)
```
burelc-amzn, line 677: The allocators need to fully initialize themselves inside their constructors. The reason for this is to ensure that the destruction order is correct. For instance, say that the `PoolSchemaImpl` object internally used some other dependent allocator, like the `SystemAllocator`. In order to ensure that the `SystemAllocator` is still valid when the `PoolSchema`'s destructor is run, the `SystemAllocator`'s constructor has to complete before the `PoolSchema` constructor completes. The way that is guaranteed is by having each allocator construct all of their dependent allocators. This is why the `SystemAllocator` calls `AllocatorInstance<OSAllocator>::Get()` in its constructor:
https://github.com/o3de/o3de/blob/69dbcd08a56539315bfb0472984daf0f46e7a966/Code/Framework/AzCore/AzCore/Memory/SystemAllocator.cpp#L33

amzn-tommy: The SystemAllocator is fully constructed before the PoolAllocatorHelper's constructor completes, but not before the PoolSchema's constructor completes. Is that sufficient? The order with my change is
```
TestPoolAllocator()->PoolAllocatorHelper()->SimpleSchemaAllocator()->AllocatorBase()
                                |                                   |
                                |                                   -->SimpleSchemaAllocator::Create()->PoolSchema()
                                -->PoolSchema::Create()->stateless_allocator::allocate()->AZ_OS_MALLOC
                                                                     ->PoolSchemaImpl()->AllocatorInstance<SystemAllocator>::Get()
                                                                                                    ->PoolAllocation<PoolSchemaImpl>()->SystemAllocator::allocate()->HphaSchema::allocate()
```
The reason I moved the PoolsSchemaImpl construction is because the SimpleSchemaAllocator's constructor always creates a PoolSchema inside SimpleSchemaAllocator::Create(). There are no parameters for pageSize, minAllocationSize, or maxAllocationSize there, so there is no way for me to forward that info to PoolSchemaImpl's constructor. By creating PoolSchemImple inside PoolSchema::Create(), I have the option to have the PoolAllocatorHelper constructor take those arguments and use them to create the appropriate Impl.

I can think of a couple of alternatives to my implementation, I'd like to know which you think is best:
1. Keep what I have here, if it is sufficient to have SystemAllocator constructed after PoolSchema() completes, but before PoolAllocatorHelper() completes.
2. Would it be sufficient to simply call AllocatorInstance<SystemAllocator>::Get() in PoolSchema(), that way it is constructed before PoolSchema() completes?
3. I could add an alternative constructor and Create function to SimpleSchemaAllocator with pageSize, minAllocation, and maxAllocation parameters. Then have the PoolAllocatorHelper(pageSize, minAllocation, maxAllocation) constructor use that one instead. The only reason I didn't do this initially was that modifying the SimpleSchemaAllocator seemed like it was getting outside the purview of modifying the PoolAllocator. But now that I understand things better it seems like a reasonable solution to me, although it's bleeding implementation details of the underlying schema into the SimpleSchemaAllocator.

burelc-amzn: Thanks for the diagram. It looks like this will work as written. I was thrown off by the fact that this is a `Schema` instead of an `Allocator`. Because `Schema::Create` is called in the constructor of the `Allocator` type, we still have the guarantee that the dependent allocators' constructors complete before the `PoolAllocator`'s does.

burelc-amzn (approved):

PR #14854, opened 2023-02-28
Action Manager | Restore Lua Editor as an item in the Tools Menu (+ minor refactoring)

AMZN-daimini:
## What does this PR do?

Restore the `Lua Editor` menu item in the Tools menu.
Took the chance to apply some minor refactorings:
- Split action and menu binding definitions for the Material Editor into separate hooks;
- Added an API call to generate a sort key for an action given its identifier, which can be used to sort alphabetically.

I preferred introducing an API call for this purpose as it makes the call opaque, giving us the chance to change it down the line if we need more granularity, and it makes it easier to expose to Python or other purposes.

## How was this PR tested?

Tested locally, menu item now appears in the menu.

Before
image

After
image

### DIFF
--- a/Code/Editor/Core/EditorActionsHandler.cpp
+++ b/Code/Editor/Core/EditorActionsHandler.cpp
@@ -1053,6 +1053,30 @@ void EditorActionsHandler::OnActionRegistrationHook()
         m_actionManagerInterface->AssignModeToAction(AzToolsFramework::DefaultActionContextModeIdentifier, actionIdentifier);
     }
 
+    // -- Tools Actions
+
+    // Lua Editor
+    {
+        constexpr AZStd::string_view actionIdentifier = "o3de.action.tools.luaEditor";
+        AzToolsFramework::ActionProperties actionProperties;
+        actionProperties.m_name = "Lua Editor";
+        actionProperties.m_category = "Tools";
+        actionProperties.m_menuVisibility = AzToolsFramework::ActionVisibility::AlwaysShow;
+
+        m_actionManagerInterface->RegisterAction(
+            EditorIdentifiers::MainWindowActionContextIdentifier,
+            actionIdentifier,
+            actionProperties,
+            []
+            {
+                AzToolsFramework::EditorRequestBus::Broadcast(&AzToolsFramework::EditorRequests::LaunchLuaEditor, nullptr);
+            }
+        );
+
+        // This action is only accessible outside of Component Modes
+        m_actionManagerInterface->AssignModeToAction(AzToolsFramework::DefaultActionContextModeIdentifier, actionIdentifier);
+    }
+
     // --- View Actions
 
     // Component Entity Layout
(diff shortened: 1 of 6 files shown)


Code/Editor/Core/EditorActionsHandler.cpp
```diff
@@ -1785,6 +1809,15 @@ void EditorActionsHandler::OnMenuBindingHook()
         }
     }
 
+    // Tools
+    {
+        m_menuManagerInterface->AddActionToMenu(
+            EditorIdentifiers::ToolsMenuIdentifier,
+            "o3de.action.tools.luaEditor",
```
hultonha, line 1816: question: Could we use the constant `actionIdentifier` here from above instead of duplicating from the above?

Code/Framework/AzToolsFramework/AzToolsFramework/ActionManager/Action/ActionManager.cpp
```diff
@@ -420,6 +420,20 @@ namespace AzToolsFramework
         return AZ::Success();
     }
 
+    int ActionManager::GenerateActionAlphabeticalSortKey(const AZStd::string& actionIdentifier)
+    {
+        auto actionIterator = m_actions.find(actionIdentifier);
+        if (actionIterator == m_actions.end())
+        {
+            return AZStd::numeric_limits<int>::max();
```
hultonha, line 428: suggestion: Returning `AZStd::numeric_limits<int>::max();` seems like a sentinel value to me 😝 Might this be better to use an optional instead so users clearly know if there's a valid value or not?

hultonha (commented):
Overall looks good but I did have one minor suggestion 🙂 Let me know what you think, thanks!

gadams3 (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/ActionManager/Action/ActionManager.cpp
AMZN-daimini: Returning an optional would mean that each individual caller would have to handle the case where no value is returned differently, causing fragmentation in behavior. I would prefer to provide a default somehow.
I returned max so that items would default to last, but it's more of a way to allow for more complex behavior in the future than anything.

Code/Editor/Core/EditorActionsHandler.cpp
AMZN-daimini: Oh, it's a different function. So we could if I moved it outside the function's scope...
Not worth holding this PR for since it passed AR, but I'll think about it if it makes sense for a future refactoring of this class.

cgalvan (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/ActionManager/Action/ActionManager.cpp
hultonha: Okay understood, if this is just last and the correct behavior will happen if not checked for that's fine, I just wanted to double check, thanks for the additional context

Issue #14514, opened 2023-02-09
Periodic Test Failure: Several Asset Pipeline tests are consistently failing in Periodic test suites


jckand-amzn:
**Describe the bug**
Several Asset Pipeline tests are consistently failing in Periodic test suites

```
AssetPipelineTests.Periodic.periodic
	- SubID_NoChange_MeshChanged[linux-windows_editor-AutomatedTesting]
	- SubID_WarningReported_AssetRemoved[linux-windows_editor-AutomatedTesting]
	
APTests.Batch_Periodic.periodic
	- test_TwoAssetsWithSameProductName_ShouldProcessAfterRename[linux-AutomatedTesting]
	- test_InvalidServerAddress_Warning_Logs[linux-AutomatedTesting]
	- test_validateDirectPreloadDependency_Found[windows-AutomatedTesting]
	
APTests.Gui_Periodic.periodic
	- test_ProcessAssets_ReprocessDeletedCache[linux-AutomatedTesting]
	- test_APStop_TimesOut[windows-AutomatedTesting]
	
APTests.Gui_2_Periodic.periodic
	- test_AllSupportedPlatforms_DeleteCachedAssets_AssetsReprocessed[linux-AutomatedTesting]
	- test_AllSupportedPlatforms_ModifyAssetInfo_AssetsReprocessed[linux-AutomatedTesting]
	- test_WindowsMacPlatforms_GUIFastScanEnabled_GameLauncherWorksWithAP[windows-AutomatedTesting]
	
APTests.AssetRelocator_Periodic.periodic
	ERRORS
	- test_WindowsMacPlatforms_RelocatorLeaveEmptyFolders_WithAndWithoutConfirm[linux-C21968381_a-True-False-True-True-True-relocate-AutomatedTesting] + ALL
	- test_WindowsMacPlatforms_DeleteWithParameters_DeleteSuccess[windows-C21968355-False-True-True-expected_queries0-unexpected_queries0-AutomatedTesting] + ALL
	FAILURES
	- test_WindowsMacPlatforms_MoveCommand_CommandResult[linux-C19462747-AutomatedTesting] + ALL (PASSES ON WINDOWS)
	
	
AssetProcessorCacheServerTests.Periodic.periodic
	- test_AssetCacheServer_LocalWorkUnaffected[linux-AutomatedTesting]
```

**Failed Jenkins Job Information:**
[...]

PR #14861, opened 2023-02-28
Updating assetpipline/assetprocessor tests to fix test issues.

LesaelR:
- Addressing test issues and flaky test in asset_processor_tests/asset_processor_batch_tests.py
 - Updating test to run on Linux asset_processor_tests/asset_processor_cache_server.py
 - Updating asset_processor_tests/asset_processor_cache_server.py name to match other tests.
Signed-off-by: Rosario Cox <lesaelr@amazon.com>

## What does this PR do?
This PR tackles some of the test issues that had caused a few of the [assetpipeline tests to be disabled.](https://github.com/o3de/o3de/issues/14514)
It also updates the Asset Processor Cache Server tests to run on Linux

## How was this PR tested?
Locally tested on Windows and Linux.


AutomatedTesting/Gem/PythonTests/assetpipeline/asset_processor_tests/asset_processor_batch_tests.py
```diff
@@ -18,7 +18,6 @@
 import shutil
 
 # Import LyTestTools
```
AMZN-stankowi, line 20: you should remove the comment, too, now that the import's gone.

AutomatedTesting/Gem/PythonTests/assetpipeline/asset_processor_tests/asset_processor_batch_tests.py
```diff
@@ -352,6 +350,9 @@ def run_ap_debug_skip_atom_output(asset_processor):
 
         shutil.copyfile(data_for_rename, file_to_rename)
 
+        # Clear the cache folder
+        asset_processor.delete_temp_cache()
```
AMZN-stankowi, line 354: won't this alter the behavior? I think the test is specifically trying to make sure the naming collision is resolved cleanly, but if the cache is modified between runs, won't that impact how the test runs?

AutomatedTesting/Gem/PythonTests/assetpipeline/asset_processor_tests/asset_processor_batch_tests.py
```diff
@@ -650,7 +653,6 @@ def test_AllSupportedPlatforms_CorruptPrefab_BatchReportsProcessingFail(self, as
 
         assert error_line_found, "The error could not be found in the newest run of the AP Batch log."
 
-    @pytest.mark.skipif(ly_test_tools.WINDOWS, reason="https://github.com/o3de/o3de/issues/14514")
     @pytest.mark.BAT
```
AMZN-stankowi, line 656: what change was made so this starts passing on Windows now?

AutomatedTesting/Gem/PythonTests/assetpipeline/asset_processor_tests/asset_processor_batch_tests.py
LesaelR: Yeah that's a good point, I figured changing this could help as the processing order could make the test flaky. (If B.fbx processed before A.fbx) - sometimes one of them wouldn't reprocess because it wasn't needed - I could change the verification at the end to ensure all the expected product assets are present in cache instead of ensuring 2 assets processed.

AutomatedTesting/Gem/PythonTests/assetpipeline/asset_processor_tests/asset_processor_batch_tests.py
LesaelR: None - After pulling from lastest, this hasn't failed for me locally, on either Windows or Linux.

AutomatedTesting/Gem/PythonTests/assetpipeline/asset_processor_tests/asset_processor_batch_tests.py
AMZN-stankowi: We have a command line parameter to run asset jobs in a very specific order, --reprocessFileList
That might help control the order this runs the second time.

AutomatedTesting/Gem/PythonTests/assetpipeline/asset_processor_tests/asset_processor_batch_tests.py
AMZN-stankowi: Hmm, I'm OK with turning it back on because otherwise we'll not be able to track these flaky tests down.

I have the feeling that this failing was Jenkins only, so we'll need to catch it failing again there to figure out next steps.

amzn-mike (approved):

AutomatedTesting/Gem/PythonTests/assetpipeline/asset_processor_tests/asset_processor_batch_tests.py
```diff
@@ -330,41 +326,42 @@ def test_TwoAssetsWithSameProductName_ShouldProcessAfterRename(self, asset_proce
         """
 
         def run_ap_debug_skip_atom_output(asset_processor):
-            result, output = asset_processor.batch_process(capture_output=True, extra_params=["--debugOutput",
-                                                                                              "--regset=\"/O3DE/SceneAPI/AssetImporter/SkipAtomOutput=true\""])
+            result, output = asset_processor.batch_process(capture_output=True, extra_params=["--regset=\"/O3DE/SceneAPI/AssetImporter/SkipAtomOutput=true\""])
```
Kadino, line 329: This nested function's name suggests that `--debugOutput` was an important param. Appears like an unintentional change, but the AP maintainers should be able to clarify whether the change is appropriate. Should probably remove debug from the name if it's no longer used.

AutomatedTesting/Gem/PythonTests/assetpipeline/asset_processor_tests/asset_processor_batch_tests.py
```diff
@@ -330,41 +326,42 @@ def test_TwoAssetsWithSameProductName_ShouldProcessAfterRename(self, asset_proce
...
             'AssetProcessorBatch should have failed because the generated output products should share the same name.'
 
         # Renaming output files so they won't collide in cache after second processing
-        file_to_rename = os.path.join(asset_processor.temp_asset_root(), "AutomatedTesting", "test_TwoAssetsWithSameProductName_ShouldProcessAfterRename", "a.fbx.assetinfo")
+        asset_info_file = os.path.join(asset_processor.temp_asset_root(), "AutomatedTesting", "test_TwoAssetsWithSameProductName_ShouldProcessAfterRename", "a.fbx.assetinfo")
         data_for_rename = os.path.join(asset_processor.temp_asset_root(), "AutomatedTesting", "test_TwoAssetsWithSameProductName_ShouldProcessAfterRename", "rename.txt")
 
-        assert os.path.exists(file_to_rename), "An assetinfo file is missing."
-        assert os.path.exists(data_for_rename), "Text file to copy into the assetinfo file is missing."
+        os.remove(asset_info_file)
```
Kadino, line 346: `os.remove` will raise an exception if the file does not currently exist on disk, is this intentional? The previous code was asserting that the path existed with a helpful message before attempting to modify it.

Tools/LyTestTools/ly_test_tools/o3de/asset_processor.py
```diff
@@ -1089,5 +1089,5 @@ def get_num_failed_processed_assets(output):
     return -1 if not result else int(result)
 
 
-def has_invalid_server_address(output):
-    return parse_output_value(output, 'Invalid server address') is not None
+def has_invalid_server_address(output, serverAddress):
+    return parse_output_value(output, f'Server address ({serverAddress}) is invalid! Reverting back to ()') is not None
```
Kadino, line 1093: This is an extremely brittle check on the entire specific failure message, where most any change will require updating this file. It also reuses a fairly inefficient parsing function. I recommend instead using:
```python
address_invalid_warning = f"({serverAddress}) is invalid"
for line in output:
    if address_invalid_warning in line:
        return True
return False
```

Kadino (commented):
Some recommended changes

AutomatedTesting/Gem/PythonTests/assetpipeline/asset_processor_tests/asset_processor_batch_tests.py
LesaelR: Yeah debug wasn't needed in this case. I'll remove it.
I think this was actually duplicate code from another module, which needed both, it would likely be best to add the skip atom output either to one of the assetpipeline fixtures or to Tools/LyTestTools/ly_test_tools/o3de/AssetProcessor.py

AutomatedTesting/Gem/PythonTests/assetpipeline/asset_processor_tests/asset_processor_batch_tests.py
LesaelR: I was assuming the file always existed as it's one of the test assets copied in, we wouldn't actually get this far in the test if the file was missing actually as AP would pass when it should have failed -  the assetinfo file it's referring to is what causes AP to intentionally fail in the step before. 
I could still add it for clarity though, thanks for the call out!

AutomatedTesting/Gem/PythonTests/assetpipeline/asset_processor_tests/asset_processor_batch_tests.py
FuzzyCarterAWS, line 329: Nice to have: Can we add this param to an AP test util as a class of consts? Manually typing in strings in a test file like this is fragile and can be difficult to maintain when repeated across multiple test files. 

It would be nice to see all params like this moved to such a data structure to make it easier to read and maintain the code.

AutomatedTesting/Gem/PythonTests/assetpipeline/asset_processor_tests/asset_processor_batch_tests.py
```diff
@@ -330,41 +326,42 @@ def test_TwoAssetsWithSameProductName_ShouldProcessAfterRename(self, asset_proce
...
+        waiter.wait_for(lambda: not os.path.exists(asset_info_file), timeout=5)
+        assert not os.path.exists(asset_info_file), "The assetinfo file was not successfully removed missing."
```
FuzzyCarterAWS, line 348: Provide the info in the assert.

Also, the assert message "was not successfully removed missing" is confusing.

```suggestion
        os.remove(asset_info_file)
        waiter.wait_for(lambda: not os.path.exists(asset_info_file), timeout=5)
        
        assert not os.path.exists(asset_info_file), f"The assetinfo file \"{asset_info_file}\" was not successfully removed and is still present."
```

However, I'm not sure how valid this assert is. [`os.remove`](https://docs.python.org/3/library/os.html#os.remove) is a python command and should be thoroughly tested by the Python team. If the command fails it will throw an exception and can be caught using a try/catch and asserting on that with a meaningful error (EG: File is in use and cannot be removed).

You can also use `ly_test_tools.environment.file_system.delete(..)` to remove with a wait to allow a timeout to continue or to try removing until the timeout before allowing the error to be thrown and handled in the test.

Alternatively you can use python to check if the file is in use and wait for the file to become available via waiter before attempting to delete and throwing an error if the file doesn't become available in time for deletion. 

EG:

```def file_has_handle(file_to_delete: os.path):
    for proc in psutil.process_iter():
        try:
            for file in proc.open_files():
                if file_to_delete == file.path:
                    return True
        except Exception:
            pass

    return False```

File: AutomatedTesting/Gem/PythonTests/assetpipeline/asset_processor_tests/asset_processor_batch_tests.py [hunk 8]
```diff
@@ -330,41 +326,42 @@ def test_TwoAssetsWithSameProductName_ShouldProcessAfterRename(self, asset_proce
...
 
-        shutil.copyfile(data_for_rename, file_to_rename)
+        os.rename(data_for_rename, asset_info_file)
+        waiter.wait_for(lambda : os.path.exists(asset_info_file), timeout=5)
+        assert os.path.exists(asset_info_file), "The assetinfo file is missing."
```
FuzzyCarterAWS, line 352: This again seems like we're trying to test python and not the system.

AutomatedTesting/Gem/PythonTests/assetpipeline/asset_processor_tests/asset_processor_batch_tests.py
```diff
@@ -380,10 +377,13 @@ def test_InvalidServerAddress_Warning_Logs(self, asset_processor):
 
         asset_processor.create_temp_asset_root()
         # Launching AP and making sure that the warning exists
-        result, output = asset_processor.batch_process(capture_output=True, extra_params=["/serverAddress=InvalidAddress", "/server"])
+        result, output = asset_processor.batch_process(capture_output=True, extra_params=[
+            '--regset="/O3DE/AssetProcessor/Settings/Server/cacheServerAddress=InvalidAddress"',
+            '--regset="/O3DE/AssetProcessor/Settings/Server/assetCacheServerMode=Server"'
```
FuzzyCarterAWS, line 382: Nice to have: More cases for an AP test util as a class of consts.

AutomatedTesting/Gem/PythonTests/assetpipeline/asset_processor_tests/asset_processor_batch_tests.py
LesaelR: I've added both of these to the Asset Processor build command in asset_processor.py
Will make a task for the class of constants request.

AutomatedTesting/Gem/PythonTests/assetpipeline/assetpipeline_utils/assetpipeline_constants.py
```diff
@@ -0,0 +1,21 @@
...
+TIMEOUT_5 = 5
+
+"""
+Asset Cache Server Constants
+"""
+
+ASSET_CACHE_INVALID_SERVER_ADDRESS = '--regset="/O3DE/AssetProcessor/Settings/Server/cacheServerAddress=InvalidAddress"'
+INVALID_SERVER_ADDRESS = "InvalidAddress"
+ASSET_CACHE_SERVER_MODE = '--regset="/O3DE/AssetProcessor/Settings/Server/assetCacheServerMode=Server"'
+ASSET_CACHE_CLIENT_MODE = '--regset="/O3DE/AssetProcessor/Settings/Server/assetCacheServerMode=Client"'
```
FuzzyCarterAWS, line 21: NIT: PEP 8 New Line at end of file expected.

AutomatedTesting/Gem/PythonTests/assetpipeline/assetpipeline_utils/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""
+Copyright (c) Contributors to the Open 3D Engine Project.
+For complete copyright and license terms please see the LICENSE at the root of this distribution.
+
+SPDX-License-Identifier: Apache-2.0 OR MIT
+"""
```
FuzzyCarterAWS, line 6: NIT: PEP 8 New Line at end of file expected.

FuzzyCarterAWS (approved):

Tools/LyTestTools/ly_test_tools/environment/file_system.py
```diff
@@ -311,6 +311,42 @@ def delete(file_list, del_files, del_dirs):
...
+    :param dst: Full path to renamed file
+
+    Returns a boolean for success or failure of the operation.
+    """
+
+    logger.info(f"Renaming {src} to {dst}.")
+    def _rename_helper(src: str | bytes, dst: str | bytes) -> bool:
+        """ Helper: Change file permissions and renames the file."""
+        try:
+            os.chmod(src, 0o777)
```
lmbr-pip, line 328: octal for the win!

lmbr-pip (approved):

Tools/LyTestTools/ly_test_tools/environment/file_system.py
```diff
@@ -311,6 +311,42 @@ def delete(file_list, del_files, del_dirs):
...
+        logger.error(f"No file located at: {src}")
+        return False
+    if os.path.exists(dst):
+        logger.error(f"File already exists at: {dst}")
+        return False
+
+    if os.path.isfile(src) and not os.path.isdir(dst):
+        return _rename_helper(src, dst)
+
+    if os.path.isdir(src) and os.path.isdir(dst):
```
amzn-mike, line 345: I don't think these if statements are working as you expect.  isdir and isfile only return true if the path exists already, so likely the first if statement is just always true (since isdir(dst) will always be false).  There's really no way to tell if a path refers to a file or a folder unless it already exists, so you might want to just remove these checks.

amzn-mike (approved):

Tools/LyTestTools/ly_test_tools/environment/file_system.py
```diff
@@ -311,6 +311,42 @@ def delete(file_list, del_files, del_dirs):
...
+        return _rename_helper(src, dst)
+
+    logger.error(f"Rename failed. Please ensure source and destination are both either file or directory paths.")
```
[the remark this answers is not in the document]
LesaelR: Removed this error, I was mistaken as to how to identify if a path was leading to a file or a directory.

Tools/LyTestTools/ly_test_tools/environment/file_system.py
LesaelR: Removed the statements.

PR #14862, opened 2023-02-28
Jobs with missing source dependencies are now run later, in case the dependencies resolve

AMZN-stankowi:
## What does this PR do?

Before this change: If a job declared a source dependency that wasn't available, the job would be run at regular priority. If the source dependency was on an intermediate asset, then Common platform jobs took lower priority than Host platform jobs, so the job would fail the first run, before the intermediate asset was created. The Asset Processor would re-run the failed job when the dependency was made available. The result of this was FBX jobs failing on running from a clean cache, but eventually all passing. This caused confusion, and resulted in running jobs when they were known to fail.

This change make it so, if a source dependency isn't available, the job is queued at a lower priority. Then, if the dependency ever shows up (the Common platform job completes, an intermediate asset is created that matches the dependency), the job ends up re-created and the old job is dropped, with the new job being put into priority where it would have been before.

If the source dependency never becomes available, then these jobs still end up running, but at the back of the queue.

## How was this PR tested?

Ran automated test.

This will be in draft until I do some manual testing.

I'm also going to leave this in draft while I think about if it's worth also fixing the cancelled job bug while I'm here, or doing that as a separate fix.


Code/Tools/AssetProcessor/native/resourcecompiler/RCQueueSortModel.cpp
```diff
@@ -55,12 +56,61 @@ namespace AssetProcessor
...
+                        actualJob->GetJobEntry().GetAbsoluteSourcePath().toUtf8().constData());
+                    for (const JobDependencyInternal& jobDependencyInternal : actualJob->GetJobDependencies())
+                    {
+                        // Include the source name so if only this log line is seen, it's still useful.
+                        AZ_Printf(AssetProcessor::ConsoleChannel, "\tSource %s has dependency - File: %s, JobKey: %s, Platform: %s",
+                            actualJob->GetJobEntry().GetAbsoluteSourcePath().toUtf8().constData(),
+                            jobDependencyInternal.m_jobDependency.m_sourceFile.m_sourceFileDependencyPath.c_str(),
+                            jobDependencyInternal.m_jobDependency.m_jobKey.c_str(),
+                            jobDependencyInternal.m_jobDependency.m_platformIdentifier.c_str());
+                    }
```
amzn-mike, line 106: Is this intended to be committed?  If so, should this be using AZ_Trace?

amzn-mike (approved):

Code/Tools/AssetProcessor/native/resourcecompiler/RCQueueSortModel.cpp
AMZN-stankowi: Yeah, that's intentional. It's additional context for the wanring a few lines up. I'll switch it to Trace. I thought we weren't using TracePrintf anymore and used Printf now, though? Or do I have that backward?

gadams3 (approved):

AMZN-stankowi:
This also fixes a Linux only issue effecting automated review, where DefaultPBR.material fails to process causing the automated review to fail.

Code/Tools/AssetProcessor/native/resourcecompiler/RCQueueSortModel.cpp
AMZN-stankowi: I removed this printf, it was causing an unused variable issue on Jenkins, and the printf isn't important enough to be worth wrapping it with unused or #ifdef.

PR #14871, opened 2023-03-01
Change Physics Asset serialization in Editor Mesh Collider to match the same property paths than previously.

moraaar:

## What does this PR do?

This is to simplify the transition from old PhysX Colliders and not having to update the property paths on prefabs' patches or python scripts.

Reverted the property paths to PhysX Mesh Collider's properties on python scripts since now it matches the same as before (one less change to do to adapt to PhysX Mesh Colliders).

Exposed collider debug draw display flag so it can be queried when converting from PhysX Collider to PhysX Mesh Collider.

## How was this PR tested?

Tested converting prefabs on complex levels (like MultiplayerSample and loft scene) with `ed_physxUpdatePrefabsWithColliderComponents` and no prefab patching warnings appear anymore after the conversion.

### DIFF
--- a/AutomatedTesting/Editor/Scripts/scene_mesh_to_prefab.py
+++ b/AutomatedTesting/Editor/Scripts/scene_mesh_to_prefab.py
@@ -169,9 +169,11 @@ def update_manifest(scene):
                                       entity_id, "{20382794-0E74-4860-9C35-A19F22DC80D4} EditorMeshColliderComponent")
 
             json_update = json.dumps({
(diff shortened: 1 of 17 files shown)


hultonha (approved):
LGTM :+1:

greerdv:
I don't think I can approve any more, but LGTM.

moraaar:
> I don't think I can approve any more, but LGTM.

Thanks for reviewing it!

PR #14883, opened 2023-03-01
Standalone tool framework

l-sempe:
## What does this PR do?

Abstracts and moves the AtomToolsDocument and associated utilities into AzToolsFramework.

As part of AzToolsFramework the ToolsDocument allows the creating of standalone tools that do not require the Atom renderer.

image

Known issue:

It is currently spawning a separate window:

image


gadams3:
These files were copied from atom tools framework instead of moved so they won’t maintain the connection or history, which is important for any other incoming merges or changes.

This merges the base application with the document application, which eliminates the possibility of having non document applications or simple tools that don’t need all of that support.

This does not update the existing tools to use the relocated files.

If the goal is to support non atom centric tools and keep everything running on the same code base, it might be easier for me/us to quickly move/rename the entire gem to an “extended/external/standalone tools framework” gem and ensure that all of the tools using this framework are updated and tested at the same time.

After that, we can extract everything that is specific to atom. There are other utility functions in other atom modules for working with assets and JSON files that need to be moved beforehand. Those can go straight to AZ core, AZ framework, AZ tools framework.

There has been some discussion, as mentioned in the material canvas RFC, that the tools currently operate as standalone only but may be updated to also support running natively in the main editor. Keeping some of the code separate or distinguishing it from AZ tools framework will help ensure that some of the unique requirements are maintained.

l-sempe:
> These files were copied from atom tools framework instead of moved so they won’t maintain the connection or history, which is important for any other incoming merges or changes.
> 
> This merges the base application with the document application, which eliminates the possibility of having non document applications or simple tools that don’t need all of that support.
> 
> This does not update the existing tools to use the relocated files.
> 
> If the goal is to support non atom centric tools and keep everything running on the same code base, it might be easier for me/us to quickly move/rename the entire gem to an “extended/external/standalone tools framework” gem and ensure that all of the tools using this framework are updated and tested at the same time.
> 
> After that, we can extract everything that is specific to atom. There are other utility functions in other atom modules for working with assets and JSON files that need to be moved beforehand. Those can go straight to AZ core, AZ framework, AZ tools framework.
> 
> There has been some discussion, as mentioned in the material canvas RFC, that the tools currently operate as standalone only but may be updated to also support running natively in the main editor. Keeping some of the code separate or distinguishing it from AZ tools framework will help ensure that some of the unique requirements are maintained.

Agreed, this was meant as a proof of concept. I will open a new draft PR in which I will move the actual files out of Atom, into their own gem and renamed accordingly. Thanks!

PR #14885, opened 2023-03-01
o3de CLI enable-gem dependency resolution

AMZN-alexpete:
## What does this PR do?

1. adds dependency resolution to the `enable-gem` CLI command
```
o3de enable-gem -pp c:/myproject -gn popcornfx>=1.2.3
```
2. CMake runs `cmake.py` for gem dependency resolution (see `Subdirectores.cmake`) and supports gem names with version specifiers e.g. `atom>=1.2.3`.  Dependency resolution can be turned off behind the variable `O3DE_DISABLE_GEM_DEPENDENCY_RESOLUTION`
3. add support for gem names with version specifiers in `project.json` `gem_names` field: Example
```
"gem_names": [
    "atom",
    "popcornfx>=1.2.3"
]
```
3. deprecates `enabled_gems.cmake` in Templates  and moves gem names into 'gem_names' field in 'project.json'
3. adds [`resolvelib`](https://github.com/sarugaku/resolvelib/ )Python dependency, [the dependency resolver used by PyPi](https://github.com/pypa/pip/issues/7406)
4. updates `manifest.get_registered()` to support returning the most compatible object (gem/engine/project) and accept object names with optional version specifiers.  e.g. o3de==1.0.0
```
o3de get-registered -en o3de>=2.3.4
```
6. removes Project Manager popup if another engine exists with the same name

Related RFC: https://github.com/o3de/sig-core/issues/44

## How was this PR tested?

- updated and ran existing unit tests
- added new unit tests
- ran configuration before and after test and compared the output list of paths generated by CMake, all non-versioned engine gem paths remained stable and the version gem paths I was testing with were correctly selected after the changes.
- verified the correct gem appeared in visual studio solution
- verified editor built and launched AutomatedTesting - though got into odd asset type not registered error that am currently looking into - possibly related to some recent type changes
- MultiplayerSample config, build and run 
- Created new project, configured and built


scripts/o3de/o3de/enable_gem.py
```diff
@@ -115,7 +115,7 @@ def enable_gem_in_project(gem_name: str = None,
         if manifest.get_project_engine_path(project_path):
             # Note: we don't remove gems that are not active or dependencies
             # because they will be implicitely found and activated via cmake 
-            incompatible_objects = compatibility.get_gem_project_incompatible_objects(gem_path, gem_json_data, project_path)
+            incompatible_objects = compatibility.get_gem_project_incompatible_objects(gem_path, gem_json_data, project_path, gem_name=gem_name)
```
lumberyard-employee-dm, line 118: typo: "implicitely" -> "implicitly"

Code/Tools/ProjectManager/Source/Application.cpp
```diff
@@ -203,40 +203,6 @@ namespace O3DE::ProjectManager
...
-                engineInfo.m_name,
-                &okPressed);
-
-            if (!okPressed)
-            {
-                // user elected not to change the name or force registration
-                return false;
-            }
-        }
-
```
lumberyard-employee-dm, line 239: It seems strange that we would still force the registration with the new logic that the engine with the same name identifier and different version can be registered.
Is that still needed?

cmake/FileUtil.cmake
```diff
@@ -141,6 +141,18 @@ function(ly_file_read path content)
...
+function(ly_file_read_cache path content)
+    unset(file_content)
+    set(file_cache_var_name "O3DE_FILE_CACHE_${path}")
+    get_property(file_content GLOBAL PROPERTY ${file_cache_var_name})
+    if(NOT file_content)
+        ly_file_read(${path} file_content)
+        set_property(GLOBAL PROPERTY ${file_cache_var_name} ${file_content})
+    endif()
+    set(${content} ${file_content} PARENT_SCOPE)
+endfunction()
```
lumberyard-employee-dm, line 155: I suggest the path is normalized in case it is slightly different(case differences on Windows or backslash/forward slash issues)

cmake/O3DEJson.cmake
```diff
@@ -49,11 +49,35 @@ function(o3de_read_json_array read_output_array input_json_path array_key)
...
+        else()
+            string(JSON value ERROR_VARIABLE manifest_json_error GET ${manifest_json_data} ${key})
+            if(manifest_json_error)
+                message(WARNING "Error reading field at key ${key} in file \"${input_json_path}\" : ${manifest_json_error}")
+            else()
+                set(${arg} ${value} PARENT_SCOPE)
+            endif()
+            unset(key)
+        endif()
+    endforeach()
```
lumberyard-employee-dm, line 82: Uh, how are the keys being split from the value output parameter.

The function doc example indicates that the `key1` parameter is separate from the output parameter of `out_key1_value`, but this code is assuming the value parameter is empty.
If it is not, then it gets used as a key.

If anything we can use a separator to encode a single argument passed to this function as a key value.
Such as embeding a comma or newline in the parameter

Code/Tools/ProjectManager/Source/Application.cpp
AMZN-alexpete: Good call, we shouldn't force registration anymore because that would break projects using older engines.

cmake/O3DEJson.cmake
AMZN-alexpete: OK, I'll take a look at that change, though I may just remove this function for now as it is currently not used anywhere, it's from a previous iteration of the code and I thought it might be useful.

AMZN-alexpete: Hmm, I tried to detect the issue when a value was not empty and would be used as a key, maybe I don't understand what the potential issues is yet, but in my test I didn't see any issue:
```
        unset(test_gem_name)
        unset(test_gem_display_name)

        # Grab the values when variables are empty
        o3de_read_json_keys("${gem_path}/gem.json" "gem_name" test_gem_name "display_name" test_gem_display_name)
        MESSAGE(INFO "gem_name: ${test_gem_name} display_name: ${test_gem_display_name}")

        # Grab the values when variables are not empty
        o3de_read_json_keys("${gem_path}/gem.json" "gem_name" test_gem_name "display_name" test_gem_display_name)
        MESSAGE(INFO "gem_name: ${test_gem_name} display_name: ${test_gem_display_name}")
   ```
   
   Prints out:
   ```
INFOgem_name: ArchVis display_name: Samples: ArchVis Assets
INFOgem_name: ArchVis display_name: Samples: ArchVis Assets
```

cmake/Subdirectories.cmake
```diff
@@ -153,24 +160,112 @@ endfunction()
...
+                " could not be found in any gem.json from the following list of registered external subdirectories:"
+                "\n  ${external_subdirs_formatted}")
                 break()
             endif()
         endforeach()
     endif()
     set(${output_gem_dirs} ${gem_dirs} PARENT_SCOPE)
 endfunction()
 
+#! Use cmake.py to get a resolved list of gem names and paths for this object (engine or project)
```
tkothadev, line 204: so how is the cmake system used in conjunction with cmake.py?
Does cmake.py run cmake commands while injecting necessary values into it?

AMZN-alexpete: 1. `Subdirectores.cmake` runs `cmake.py`, giving it the path to the engine or project.  `cmake.py` does not run any cmake commands, only Python
2. `cmake.py` determines all the gems an engine or project uses and runs dependency resolution on them
3. `cmake.py`  writes the output as a CMake list to a file 
4. `Subdirectories.cmake` loads the file and writes the resolved gem name to gem path results into CMake global property variables, so that when any logic wants to know the path to some gem, it get's the resolved path to the gem from that global property.

This way, if you have say 100 versions of PopcornFX gems on your computer, CMake can determine what single gem path should be used for PopcornFX in your project.

cmake/Subdirectories.cmake
```diff
@@ -153,24 +160,112 @@ endfunction()
...
+        OUTPUT_VARIABLE O3DE_CLI_OUT 
+        ERROR_VARIABLE O3DE_CLI_OUT
+        )
+
+    if(O3DE_CLI_RESULT)
+        message(WARNING "Dependecy resolution failed\n  Error: ${O3DE_CLI_OUT}")
+        return()
+    endif()
+
+    file(READ "${output_path}" resolved_gem_dependency_output)
```
lumberyard-employee-dm, line 232: Whoa, writing to a file during configuration this is a point of failure we should avoid.
This should write the dependencies to stdout and the OUTPUT_VARIABLE should be used.
Having to write and read a file on disk opens up issues with file permissions and sync issues.

lumberyard-employee-dm (changes requested):

cmake/Subdirectories.cmake
lumberyard-employee-dm: We can default to writing to [stdout](https://stackoverflow.com/a/7864158) if no output path is giving

cmake/Version.cmake
```diff
@@ -14,6 +16,81 @@ ly_file_read("${LY_ROOT_FOLDER}/engine.json" tmp_json_data)
...
+    if("${input}" MATCHES "^(.*)(~=|==|!=|<=|>=|<|>|===)(.*)$")
+        if(${CMAKE_MATCH_COUNT} GREATER_EQUAL 1)
+            set(${dependency_name} ${CMAKE_MATCH_1} PARENT_SCOPE)
+        endif()
+        if(${CMAKE_MATCH_COUNT} GREATER_EQUAL 2)
+            set(${operator} ${CMAKE_MATCH_2} PARENT_SCOPE)
+        endif()
+        if(${CMAKE_MATCH_COUNT} GREATER_EQUAL 3)
+            set(${version} ${CMAKE_MATCH_3} PARENT_SCOPE)
+        endif()
```
lumberyard-employee-dm, line 35: The match count here can never be less than 4.
The regular expression cannot succeed in any other way.
A capture 0 is implicit and capture 1-3 aren't optional, that must be their.
They capture group may be an empty string but they must be there.
Also there is a flaw with the regular expression
It can no characters on either size of the relational operator
i.e "o3de==" will match, "==1.0.0" will match as well.
Second problem if for some reason the user specifies the version specifier as `o3de == 1.0.0` that will also fail because the `dependency_name` and `version` values don't strip surrounding whitespace.
```suggestion
    if("${input}" MATCHES "^(.+)(~=|==|!=|<=|>=|<|>|===)(.+)$")
        string(STRIP ${CMAKE_MATCH_1} _dependency_name)
        set(${dependency_name} ${_dependency_name} PARENT_SCOPE)
        set(${operator} ${CMAKE_MATCH_2} PARENT_SCOPE)
        string(STRIP ${CMAKE_MATCH_3} _version)
        set(${version} ${_version} PARENT_SCOPE)
```

cmake/Version.cmake
```diff
@@ -14,6 +16,81 @@ ly_file_read("${LY_ROOT_FOLDER}/engine.json" tmp_json_data)
...
+    elseif(op STREQUAL ">" AND version VERSION_GREATER specifier_version)
+        set(contains_version TRUE)
+    elseif(op STREQUAL "===" AND version STREQUAL specifier_version)
+        set(contains_version TRUE)
+    elseif(op STREQUAL "~=")
+        # compatible versions have an equivalent combination of >= and == 
+        # e.g. ~=2.2 is equivalent to >=2.2,==2.*
+        if(version VERSION_GREATER_EQUAL specifier_version)
+            string(REPLACE "." ";" specifer_version_part_list ${specifier_version})
+            list(LENGTH specifer_version_part_list list_length)
```
lumberyard-employee-dm, line 70: typo: specifer -> specifier
```suggestion
            list(LENGTH specifier_version_part_list list_length)
```

scripts/o3de/o3de/cmake.py
```diff
@@ -243,3 +242,120 @@ def get_enabled_gem_cmake_file(project_name: str = None,
...
+                             )
+
+def add_parser_args(parser):
+    group = parser.add_argument_group("resolve gem dependencies")
+    group.add_argument('-pp', '--project-path', type=pathlib.Path, required=False,
+                       help='The path to the project.')
+    group.add_argument('-ep', '--engine-path', type=pathlib.Path, required=False,
+                       help='The path to the engine.')
+    group.add_argument('-gpof', '--gem-paths-output-file', type=pathlib.Path, required=True,
+                       help='The path to the resolved gem paths output file.')
```
lumberyard-employee-dm, line 350: This shouldn't be required.
If it is not set, then the file to open should be set to `sys.stdout`

scripts/o3de/o3de/cmake.py
```diff
@@ -243,3 +242,120 @@ def get_enabled_gem_cmake_file(project_name: str = None,
...
+        # Try without optional gems
+        gem_names_without_optional = utils.get_gem_names_set(active_gem_names, include_optional=False)
+        results, errors = compatibility.resolve_gem_dependencies(gem_names_without_optional, 
+                                                                 all_gems_json_data, 
+                                                                 engine_json_data,
+                                                                 include_optional=False)
+
+    if errors:
+        logger.error(f'Failed to resolve dependencies:\n '+ 
+                    "\n  ".join(errors))
```
lumberyard-employee-dm, line 322: Why is there a single space after the first newline.
But too spaces after every other newlines.
The output would like something as follows
```
Failed to resolve dependencies
 error 1
  error 2
  error 3

File: scripts/o3de/o3de/cmake.py [hunk 11]
```diff
@@ -27,75 +29,55 @@ def add_gem_dependency(cmake_file: pathlib.Path,
...
-                if gem_name in gem_name_list:
-                    logger.info(f'{gem_name} is already enabled in file {str(cmake_file)}.')
-                    return 0
+            # regex to replace every version of the gem name 
+            regex_replace_gem_name = re.compile(f'(?:^|\s+){gem_name}[=><~]*[\S]*')
+            all_gem_names = regex_replace_gem_name.sub('', all_gem_names_matches.group('gem_names'))
 
-            t_data.append(line)
-            line_index += 1
+            all_gem_names += f'\n{" "  * indent}{gem_name_with_version_specifier}'
```
lumberyard-employee-dm, line 67: The reason regular expressions weren't used is because it is not powerful enough to parse all the obscure cases like having multiple gems on the same with version specifiers in them or the fact that the ENABLED_GEMS might not be terminated correctly with a trailing `)` causing further corruption of the enable_gems.cmake when writing back to it.

scripts/o3de/o3de/cmake.py
```diff
@@ -27,75 +29,55 @@ def add_gem_dependency(cmake_file: pathlib.Path,
...
-            if parsed_line.startswith(enable_gem_start_marker):
-                # Skip pass the 'set(ENABLED_GEMS' marker just in case their are gems declared on the same line
-                parsed_line = parsed_line[len(enable_gem_start_marker):]
-                # Set the flag to indicate that we are in the ENABLED_GEMS variable
-                in_gem_list = True
-                start_marker_line_index = line_index
+        file_contents = s.read()
+        if file_contents:
+            # regex to isolate pre, gem names and post 
+            regex_all_gem_names = re.compile('(?P<pre>[\s\S]*set\s*\(\s*ENABLED_GEMS\s*)?(?P<gem_names>[^\)]+)(?P<post>\)[\s\S]*)?')
```
lumberyard-employee-dm, line 51: This sequence being matched in the `pre` and `post` named capture group checks doesn't make any sense.
This is matching zero or more characters that **is whitespace** character or **is not a whitespace**.
In that case `.` can be used.

lumberyard-employee-dm (commented):
The gem names in the `enabled_gems.cmake` file was kept as simple as possible, because we are not in the business of trying to parse and write valid cmake.

My opinion is that we should eschew trying to write or even read version information from the enable_gems.cmake files and instead just write it to the `"gem_names"` or `"gem_dependencies"` if you want to call it that array in the project.json.

cmake/PAL.cmake
```diff
@@ -71,6 +71,91 @@ function(o3de_find_gem_with_registered_external_subdirs gem_name output_gem_path
...
+            if(EXISTS ${candidate_gem_path})
+                unset(candidate_paths)
+                list(APPEND candidate_paths ${external_subdir})
+                o3de_read_json_key(candidate_gem_name ${candidate_gem_path} "gem_name")
+
+                set(cached_candidate_paths_property "${candidate_gem_name}:CandidatePaths")
+                get_property(cached_candidate_paths GLOBAL PROPERTY ${cached_candidate_paths_property})
+                if(cached_candidate_paths)
+                    list(APPEND candidate_paths ${cached_candidate_paths})
+                endif()
```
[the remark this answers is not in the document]
AMZN-alexpete: Ah thanks, that makes it simpler! I was under the impression if the property didn't exist then the `list(APPEND` call would error.

cmake/O3DEJson.cmake
AMZN-alexpete: Thanks! I'll make it use `if(NOT DEFINED` instead to avoid the "falsey" problem - which is also my new band name btw.

cmake/Subdirectories.cmake
AMZN-alexpete: Sounds good, and I agree, it'd be safer. I went the file route to make it easier to troubleshoot failures and assumed if you can't write to the build folder then you've got bigger issues to solve 😄

cmake/FileUtil.cmake
```diff
@@ -141,6 +141,20 @@ function(ly_file_read path content)
     set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS ${path})
 endfunction()
 
+#! ly_file_read: wrap ly_file_read but store the file in a cache to avoid 
+#  extra reads.
+function(ly_file_read_cache path content)
```
spham-amzn, line 146: maybe we should start naming functions starting with `o3de` now?
```suggestion
function(o3de_file_read_cache path content)
```

scripts/o3de/o3de/cmake.py
spham-amzn: or... you can also add a `--dry-run` option to output to `sys.stdout`

scripts/o3de/o3de/disable_gem.py
```diff
@@ -88,12 +88,13 @@ def disable_gem_in_project(gem_name: str = None,
         return 1
 
     # remove the gem
-    error_code = cmake.remove_gem_dependency(enabled_gem_file, gem_json_data['gem_name'])
+    gem_name = gem_name if gem_name else gem_json_data.get('gem_name','')
```
spham-amzn, line 91: you can also do this
```suggestion
    gem_name = gem_name or gem_json_data.get('gem_name','')
```

scripts/o3de/o3de/cmake.py
AMZN-alexpete: looks like a bug to me!

spham-amzn: In general, if you are using a logger, you wouldn't want to use anything multi-line. logger can be registered with formatters that may append things like log level, timestamp, etc, which will throw off the identing anyways.

scripts/o3de/o3de/cmake.py
AMZN-alexpete: Do the unit tests cover these edge cases? If so, then they are currently passing with these regex.

AMZN-alexpete:
> The gem names in the enabled_gems.cmake file was kept as simple as possible, because we are not in the business of trying to parse and write valid cmake.
> 
> My opinion is that we should eschew trying to write or even read version information from the enable_gems.cmake files and instead just write it to the "gem_names" or "gem_dependencies" if you want to call it that array in the project.json.

I'd be fine with not including  versions in `enabled_gems.cmake`, and would like to move that content into `project.json` and `engine.json`, but am concerned about needing extra code to handle backwards compatibility.

I can take some time trying the route of only handling version specifiers in `enabled_gems.cmake` and potentially moving those gems in to "project.json"

scripts/o3de/o3de/compatibility.py
```diff
@@ -326,3 +369,173 @@ def has_compatible_version(name_and_version_specifier_list:list, object_name:str
             pass
 
     return False 
+
+class GemRequirement(namedtuple("GemRequirement", ["name", "specifier"])):  # noqa
```
spham-amzn, line 373: what is `#noqa` ?

AMZN-alexpete: oh... that's a comment from resolvelib's code I used initially - will remove

spham-amzn: Nice use of `namedtuple`!

scripts/o3de/o3de/cmake.py
lumberyard-employee-dm: The dot character matches newlines as well as long it doesn't used the `$` anchor at the end: https://regex101.com/r/pGU4Bn/1

scripts/o3de/o3de/cmake.py
lumberyard-employee-dm: The Unit test does cover the complexity for regular gem names.
Test would have to be added to make sure specifying gem dependencies on the same line would work.
i.e
```cmake
set(ENABLED_GEMS
    gem1=1.0.0 gem2=2.0.0 gem3 gem4=1.0.1)
```

spham-amzn (approved):

lumberyard-employee-dm:
> > The gem names in the enabled_gems.cmake file was kept as simple as possible, because we are not in the business of trying to parse and write valid cmake.
> > My opinion is that we should eschew trying to write or even read version information from the enable_gems.cmake files and instead just write it to the "gem_names" or "gem_dependencies" if you want to call it that array in the project.json.
> 
> I'd be fine with not including versions in `enabled_gems.cmake`, and would like to move that content into `project.json` and `engine.json`, but am concerned about needing extra code to handle backwards compatibility.
> 
> I can take some time trying the route of only handling version specifiers in `enabled_gems.cmake` and potentially moving those gems in to "project.json"

Well we don't have any backwards compatibility concerns with the `enabled_gems.cmake` format.

It currently only supports referencing a gem only by name identifier.
We don't have to add support for it to reference a gem by `<gem_name>=<version>`.

It would keep parsing of existing `enabled_gems.cmake` files simpler and we can redirect the `enable_gems.py` command to always write to the "gem_names" field of the project.json instead of writing to `enabled_gems.cmake`.

That would provide a builtin incentive for users to prefer setting gem dependencies in the `project.json` instead of `enabled_gems.cmake` file as the `enabled_gems.cmake` wouldn't support versioning.

scripts/o3de/o3de/cmake.py
AMZN-alexpete: I do have this test:
```
            # when multiple versions of the same gem are found expect all are replaced with the new entry
            pytest.param("""
                        # Comment
                        set(ENABLED_GEMS
                            foo bar
                            TestGem==1.0.0 TestGem TestGem>=2.0.0
                            baz)
                    """, 'TestGem', set(['foo', 'bar', 'baz', 'TestGem']), 0),
```

scripts/o3de/o3de/cmake.py
AMZN-alexpete: It does not seem to capture multiple lines, i.e. everything up until `ENABLED_GEMS` https://regex101.com/r/nDfF0a/1

scripts/o3de/o3de/cmake.py
```diff
@@ -243,3 +242,126 @@ def get_enabled_gem_cmake_file(project_name: str = None,
...
+    else:
+        active_gem_names = engine_json_data.get('gem_names',[])
+
+    # some gem name entries will be dictionaries - convert to a set of strings 
+    gem_names_with_optional_gems = utils.get_gem_names_set(active_gem_names, include_optional=True)
+    if not gem_names_with_optional_gems:
+        logger.info(f'No gem names were found to use as input to resolve gem dependencies.')
+        if resolved_gem_dependencies_output_path:
+            with resolved_gem_dependencies_output_path.open('w') as output:
+                output.write('')
```
lumberyard-employee-dm, line 299: We could use a function that wraps a [context manager](https://docs.python.org/3/library/contextlib.html?highlight=contextlib#contextlib.contextmanager) to encapsulate the output file
```python
@contextlib.contextmanager
def open_output(output_path:pathlib.Path):   
    output = resolved_gem_dependencies_output_path.open('w') if resolved_gem_dependencies_output_path else sys.stdout
    yield output
    if resolved_gem_dependencies_output_path:
        output.close()

# ...
with open_output(resolved_gem_dependencies_output_path) as output:
    output.write('')
```

scripts/o3de/o3de/cmake.py
lumberyard-employee-dm: That should work.
Can you add a test make sure the `enable_gems.cmake` file isn't modified if `set(ENABLED_GEMS)` variable is invalid such as it is missing an ending `)`

cmake/Subdirectories.cmake
```diff
@@ -153,24 +160,108 @@ endfunction()
...
+                if(NOT O3DE_DISABLE_GEM_DEPENDENCY_RESOLUTION)
+                    message(WARNING "Failed to find resolved gem path for gem '${gem_name_with_version_specifier}'")
+                    # Make a best effort attempt to find the most compatible gem
+                    o3de_find_most_compatible_gem_with_registered_external_subdirs(${gem_name_with_version_specifier} gem_path "${registered_external_subdirs}")
+                else()
+                    o3de_find_gem_with_registered_external_subdirs(${gem_name_with_version_specifier} gem_path "${registered_external_subdirs}")
+                endif()
+
+                if(gem_path)
+                    set_property(GLOBAL PROPERTY "@GEMROOT:${gem_name}@" "${gem_path}")
```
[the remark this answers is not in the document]
AMZN-alexpete: Yes, before these changes, `add_o3de_object_json_external_subdirectories` would re-open all the `gem.json` files to determine the gem paths, so I left the logic to resolve those gem paths in that location, but if there is a better place to resolve these dependencies and set those gem paths, please let me know.  It's very likely I do not fully understand the gem dependency cmake logic.

scripts/o3de/o3de/cmake.py
lumberyard-employee-dm: It looks like the dot for python matches zero or more characters that are not line terminators
> matches any character (except for line terminators)
image

So I guess the regular expression you have is a funky way to match multiple lines, though I don't think that was the intent of the regular expression engine.

scripts/o3de/o3de/compatibility.py
AMZN-alexpete: I can't take credit - this is based on resolvelib's examples

scripts/o3de/o3de/cmake.py
AMZN-alexpete: well, I'm currently looking into remove the need for `enable_gems.cmake` if that goes well I'll revert these changes

cmake/Subdirectories.cmake
```diff
@@ -255,27 +346,42 @@ function(get_all_external_subdirectories_for_o3de_object output_subdirs object_t
...
 
+        # Remove any version specifier from the gem name
+        o3de_get_name_and_version_specifier(${gem_name_with_version_specifier} gem_name spec_op spec_version)
+
         # Set a global "optional" property on the gem name
         set_property(GLOBAL PROPERTY "${gem_name}_OPTIONAL" ${gem_optional})
+
         # Build the gem_names list with extracted names
-        list(APPEND gem_names ${gem_name})
+        list(APPEND gem_names ${gem_name_with_version_specifier})
```
lumberyard-employee-dm, line 374: After this loop a call to `ly_enable_gems(GEM ${gem_name} PROJECT_NAME ${object_name}` can be added to make sure the gems referenced in the "gem_name" field of a project.json or engine.json is added to the generated `cmake_dependencies.*.setreg` for loading

engine.json
```diff
@@ -101,6 +101,15 @@
...
         "Gems/WhiteBox"
     ],
+    "gem_names": [
+        "AtomShader",
+        "Atom_Bootstrap",
+        "CommonFeaturesAtom",
+        "Camera",
+        "Maestro",
+        "PrefabBuilder",
+        "SceneProcessing"
```
lumberyard-employee-dm, line 111: The `ly_enable_gems` call can be removed from those CMakeLists.txt

lawsonamzn:
Q: What does this mean for compatibility with existing projects?  We're trying to not make changes that would require all existing gems and projects out there to update their stuff (unless we can update it automatically for them)

Also, on a separate topic...

I'm all for the removal of enabled_gems.cmake, tbh.  

However, there is a distinction between 
* enabled_gems.cmake, which lists the names of cmake **targets** to affix ".client", ".server", ".tools", ".builder" to, and attempts to establish a dependency on it, but also operates at the cmake target / module level...
* and "gem_names" field in project.json, which is the name of individual gems (not, inside the gem.json, which have been registered and may have more than one module in it.

_However_

It's probably still ok.  The reason being, by _convention_, people have named the 'main module' of their gem called ZZZZZ as 'ZZZZZ' also.   This means you can currently assume that for any gem called ZZZZZ you can automatically add the CMake Target called ZZZZZ(.clients|.servers|.builders|.tools) to the dependency tree.

There were at one point plans to allow the gems to override this behavior, or offer additional optional modules in addition to their main one, declaritively, inside Gem.json, but so far there's just been no need whatsoever for such optional modules, as they themselves could be additional gems in the same package that depend on the main gem anyway...

cmake/Subdirectories.cmake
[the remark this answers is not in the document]
lumberyard-employee-dm: We can set each gems `@GEMROOT:<gem-name>@` property in `get_external_subdirectories_in_use` after all the external_subdirectories in use has been gathered.

If we have a mapping of "candidate gem_path to gem_name" that we populate when we `gem.json` file for the first time, then that would make sure we capture gems of the same name but different versions.
Then at the bottom of `get_external_subdirectories_in_use` e can loop over the "external subdirectories that will be used by the project" and use them as a key into the "candiadate_gem_path to gem_name" map to create mapping of `@GEMROOT@:<gem-name>` -> `<gem-path>`

lumberyard-employee-dm:
> Q: What does this mean for compatibility with existing projects? We're trying to not make changes that would require all existing gems and projects out there to update their stuff (unless we can update it automatically for them)
> 
> Also, on a separate topic...
> 
> I'm all for the removal of enabled_gems.cmake, tbh.
> 
> However, there is a distinction between
> 
>     * enabled_gems.cmake, which lists the names of cmake **targets** to affix ".client", ".server", ".tools", ".builder" to, and attempts to establish a dependency on it, but also operates at the cmake target / module level...
> 
>     * and "gem_names" field in project.json, which is the name of individual gems (inside the gem.json, which have been registered with the engine) which may have more than one actual module in it with different names.
> 
> 
> _However_
> 
> It's probably still ok. The reason being, by _convention_, people have named the 'main module' of their gem called ZZZZZ as 'ZZZZZ' also. This means you can currently assume that for any gem called ZZZZZ you can automatically add the CMake Target called ZZZZZ(.clients|.servers|.builders|.tools) to the dependency tree. The Project Manager UI kind of depends on it anyway, that is, it automatically adds "ZZZZZ" to the list of cmake TARGETS to depend on (in enabled_gems.cmake), when you add ZZZZZ to the list of _gems_ you use even though those are technically unrelated. By enforcing this workflow, it has resulted in every gem having its main module be named the same as its gem name.
> 
> There were at one point plans to allow the gems to override this behavior, or offer additional optional modules in addition to their main one, declaritively, inside Gem.json, but so far there's just been no need whatsoever for such optional modules, as they themselves could be additional gems in the same package that depend on the main gem anyway...

By using the convention it is the simplest approach as we have even updated the gem templates to use the ${gem_name} as a [prefix for the CMake targets](https://github.com/o3de/o3de/blob/development/Templates/DefaultGem/Template/Code/CMakeLists.txt#L29) in this [PR](https://github.com/o3de/o3de/commit/7b3434a216d74310834f608a9c2106ab2edb967f).

Also following the convention makes it easier for users to still have their gems work in Project Manager when renaming them, as only the "gem_name" in the gem.json file needs to be updated and not the CMake Targets..

That being said the `enabled_gems.cmake` file is still being read and the gem module "prefix" value for lack of a better term is still supported. It is just being moved to more of an advanced workflow.

Pretty much if a gem author wants to have a gem with a different name than the prefixes it uses it's gem module CMake Targets, than it can, but it won't work with ProjectManager and would require users of the gem to manually add an `ly_enable_gems` call that specifies that prefix.

But 99%+ of the cases the prefix used for the CMake targets for a Gem such as `ZZZZZ.Clients`, `ZZZZZ.Servers`, etc... is going to be the same as gem name `ZZZZZ`.

Since the Gem templates and Project templates have been updated to generate the CMakeLists.txt Gem Targets using the `${gem_name}` variable as a prefix for the target, it makes it even more likely that gem name is going to match the "gem target prefix".
The gem author would have to go out their way to change the name of the Gem without changing the name of CMake targets can vice-versa

AMZN-alexpete:
> Q: What does this mean for compatibility with existing projects? We're trying to not make changes that would require all existing gems and projects out there to update their stuff (unless we can update it automatically for them)
I initially avoided making `enabled_gems.cmake` deprecated because of my recent track record for breaking things 😄 but it makes the code simpler and has the other benefits mentioned already and the functionality I left in still reads the file if it exists.  Anyone with a project that has an `enabled_gems.cmake` and removes a gem from their project, will have that gem removed from `enabled_gems.cmake` and the `gem_names` fields in `project.json`.  When a gem is added to a project it will be added to the `gem_names` field only.

There are probably edge cases that I am attempting to capture in the updates to the unit tests I'm making

scripts/o3de/o3de/cmake.py
AMZN-alexpete: I've removed this function and related tests as it is no longer necessary to add gems to `enabled_gems.cmake`

scripts/o3de/o3de/cmake.py
AMZN-alexpete: As discussed, I've deprecated `enabled_gems.cmake` and the functions that add gems to it

cmake/Subdirectories.cmake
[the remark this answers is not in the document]
AMZN-alexpete: I spent some time trying this change and discovered that `query_gem_paths_from_external_subdirs` gets called for gems in the `external_subdirectories` in `engine.json` that are not used by the project or engine.  If we don't store a gem_name -> gem_path mapping like we are with `@GEMROOT:,gem-name>@` then we run logic that searches every external_subdirectory looking in all the gem.json files for a gem with that name.  None of those gem_name -> gem_path mappings will exist if we rely on the resolved gem_name -> gem_path mappings because the logic is looking for gems that the project doesn't use.  I think the reasoning here is that, even if your project doesn't use all the gems in the engine, it should still show them in the Visual Studio solution?

For example: The SaveData and GameStateSamples gems appear in Visual Studio even though they are not enabled in AutomatedTesting.

scripts/o3de/o3de/cmake.py
```diff
@@ -144,7 +71,7 @@ def remove_gem_dependency(cmake_file: pathlib.Path,
                 while gem_name in gem_name_list:
                     gem_name_list.remove(gem_name)
                     removed = True
-
+
```
lumberyard-employee-dm, line 74: nitpick: extra whitespace has been added here.

cmake/Version.cmake
AMZN-alexpete: In my tests the old regex looks like it works OK and maybe handles one case better (where version is missing) - except for the space stripping, thoughts? https://godbolt.org/z/3dnqqod9a

```
OLD 'o3de == 1.0.0' object_name:'o3de' op:'==' ver: '1.0.0'
NEW 'o3de == 1.0.0' object_name:'o3de' op:'==' ver: '1.0.0'

OLD 'o3de>=1.2.3' object_name:'o3de' op:'>=' ver: '1.2.3'
NEW 'o3de>=1.2.3' object_name:'o3de' op:'>=' ver: '1.2.3'

OLD 'o3de' object_name:'o3de' op:'' ver: ''
NEW 'o3de' object_name:'o3de' op:'' ver: ''

OLD 'o3de>=' object_name:'o3de' op:'>=' ver: ''
NEW 'o3de>=' object_name:'o3de' op:'>' ver: '='
```

lawsonamzn (approved):
Looks good - I also fetched this branch and tried it out from a scratch situation (no python downloaded, no cache) to ensure that it didn't cause a weird issue like needing python to get cmake going... worked well.

scripts/o3de/o3de/manifest.py
```diff
@@ -16,7 +16,7 @@
 from packaging.version import Version
 from collections import deque
 
-from o3de import validation, utils, repo, compatibility
+from o3de import validation, utils, repo, compatibility, cmake
```
lumberyard-employee-dm, line 19: This is a circular import:
The cmake script imports manifest.py

scripts/o3de/o3de/manifest.py
```diff
@@ -301,6 +301,83 @@ def get_engine_templates() -> list:
...
+    Example: {"gemA>=1.2.3":"c:/gemA", "gemB":"c:/gemB"}
+    :param project_path The path to the project
+    :param include_gem_dependencies True to include all gem dependencies, otherwise just return
+    gems listed in project.json and the deprecated enabled_gems.json
+    """
+    project_json_data = get_project_json_data(project_path=project_path)
+    active_gem_names = project_json_data.get('gem_names',[])
+    enabled_gems_file = cmake.get_enabled_gem_cmake_file(project_path=project_path)
+    if enabled_gems_file and enabled_gems_file.is_file():
+        active_gem_names.extend(cmake.get_enabled_gems(enabled_gems_file))
```
lumberyard-employee-dm, line 316: Moving the `get_enabled_gem_cmake_file` and `get_enable_gems` methods to this file will resolve the need to import the cmake module

scripts/o3de/o3de/manifest.py
```diff
@@ -301,6 +301,83 @@ def get_engine_templates() -> list:
...
+    engine_path = get_project_engine_path(project_path=project_path)
+    if not engine_path:
+        engine_path = get_this_engine_path()
+        if not engine_path:
+            logger.error('Failed to find an engine path for the project at '
+                            f'"{project_path}" which is required to resolve gem dependencies.')
+            return result
+
+        logger.warning('Failed to determine the correct engine for the project at '
+                        f'"{project_path}", falling back to this engine at {engine_path}.')
```
lumberyard-employee-dm, line 334: Does `get_proejct_engine_path` take into the actual path the project is using?
Such as if their `user/project.json` is set to a different engine?

scripts/o3de/o3de/manifest.py
AMZN-alexpete: good catch

scripts/o3de/o3de/manifest.py
AMZN-alexpete: yes, it calls `get_most_compatible_project_engine_path` which takes `user/project.json` into account.

scripts/o3de/o3de/manifest.py
```diff
@@ -301,6 +301,83 @@ def get_engine_templates() -> list:
...
+        resolved_gems, errors = compatibility.resolve_gem_dependencies(gem_names_without_optional, 
+                                                                 all_gems_json_data, 
+                                                                 engine_json_data,
+                                                                 include_optional=False)
+    if not errors:
+        for _, gem in resolved_gems.items():
+            gem_name = gem.gem_json_data['gem_name']
+            gem_name_with_specifier = gem_names_with_version_specifiers.get(gem_name,gem_name)
+            if gem_name_with_specifier in result or include_dependencies:
+                result[gem_name_with_specifier] = gem.gem_json_data['path'].resolve().as_posix()
```
lumberyard-employee-dm, line 452: Uh, isn't this path relative to the gem directory containing the "external_subdirectory" entry for this sub gem?
Wouldn't this result in the path being resolved against the current working directory therefore resulting in the wrong path?

scripts/o3de/o3de/manifest.py
```diff
@@ -760,11 +837,45 @@ def get_repo_path(repo_uri: str, cache_folder: str or pathlib.Path = None) -> pa
...
+    :param gem_json_data_by_name Gem data from get_gems_json_data_by_name()
+    """
+    gem_name_with_version_specifier = gem_name
+    gem_name, version_specifier = utils.get_object_name_and_optional_version_specifier(gem_name)
+    if not gem_name in gem_json_data_by_name:
+        return None
+
+    matching_paths = deque()
+    most_compatible_version = Version('0.0.0')
+    for gem_json_data in gem_json_data_by_name[gem_name]:
```
lumberyard-employee-dm, line 936: We should probably use `gem_json_data_by_name.get(gem_name, {})` to be on the safe side

scripts/o3de/o3de/project_properties.py
```diff
@@ -43,6 +43,19 @@ def _edit_gem_names(proj_json: dict,
...
+            candidate_gem_name = candidate if isinstance(candidate, str) else candidate.get('name')
+            candidate_gem_name_only, _ = utils.get_object_name_and_optional_version_specifier(candidate_gem_name)
+            return candidate_gem_name_only == gem_name
+
+        # remove any versions of the existing gem first
+        if proj_json.get('gem_names',[]):
+            for new_gem_name in add_list:
+                new_gem_name = new_gem_name if isinstance(new_gem_name,str) else new_gem_name['name']
+                gem_name_only, _ = utils.get_object_name_and_optional_version_specifier(new_gem_name)
+                proj_json['gem_names'] = [gem for gem in proj_json['gem_names'] if not is_version_of_gem(gem, gem_name_only)]
```
lumberyard-employee-dm, line 57: Doesn't the delete_gem_names code need this logic as well?

If the project.json contained the following entry
```json
{
    "gem_names": [ "gemA=1.0.0" ]
}
```

And the user requested to remove gem "gemA", then "gemA=1.0.0" would still exist.

scripts/o3de/o3de/manifest.py
AMZN-alexpete: I probably don't need to call resolve here,  the paths should already be absolute at this point. 
The 'path' comes from `get_gems_json_data_by_name` which pulls the paths from 
`get_manifest_external_subdirectories`, `get_project_external_subdirectories`, and then uses `get_gems_from_external_subdirectories`.
It appears that all the functions return absolute paths by appending the 'external_subdirectory' except `get_manifest_external_subdirectories`

It seems to be working fine for all the Atom sub gems so far, but couldn't hurt to try the MPS gems too.

scripts/o3de/o3de/project_properties.py
AMZN-alexpete: That's probably a good idea. Currently you have to specify the exact gem name with specifier or it won't delete it, but I bet if a user tries to remove a gem and doesn't provide the specifier, we can assume they want to remove any version of the gem.

scripts/o3de/o3de/manifest.py
AMZN-alexpete: I verified with MPS that all paths are absolute for gems in the engine and manifest including subdir gems like those in `Atom` and `MPSGems`

AMZN-Phil (approved):

PR #14888, opened 2023-03-02
Fix parsing of 'TracerPid' ID to account for spaces AND tabs before the number

spham-amzn:
## What does this PR do?

This fixes an issue on linux where the detection of a debugger being attached is always reporting 'true' whether or not one is attached at all. The original logic relied on only checking for the space character, but on ubuntu, it was a tab character before the pid.

## How was this PR tested?

Added temporary trace logs to show results for `IsDebuggerPresent` with and without a debugger (lldb) attached.

### DIFF
--- a/Code/Framework/AzCore/Platform/Common/UnixLike/AzCore/Debug/Trace_UnixLike.cpp
+++ b/Code/Framework/AzCore/Platform/Common/UnixLike/AzCore/Debug/Trace_UnixLike.cpp
@@ -76,6 +76,8 @@ namespace AZ::Debug
         {
             if (isEnabled)
             {
+                // Signals to handle based on glibc may be defined at 
+                // [bits/signum-generic.h](https://sourceware.org/git/?p=glibc.git;a=blob;f=bits/signum-generic.h;h=67131853c220b942aa9bc599eca7db5e5e797f98;hb=e4e11b1dba261cb650e631978622bf3b4a4d8c37#l28)
                 signal(SIGSEGV, ExceptionHandler);
                 signal(SIGTRAP, ExceptionHandler);
                 signal(SIGILL, ExceptionHandler);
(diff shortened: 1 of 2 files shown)


lumberyard-employee-dm (approved):

cgalvan (approved):

cgalvan:
@spham-amzn Awesome thanks for catching this!

mbalfour-amzn (approved):
Lol nice fix!

lumberyard-employee-dm (approved):
Is Jenkins now trying to actually attach a debugger to the process when a death test fails?

spham-amzn:
> Is Jenkins now trying to actually attach a debugger to the process when a death test fails?

Actually found out that the googletest package that was built has death tests disabled, and the failing test was a death test. It needed to be wrapped with an #if/#endif guard like the other DEATH tests

lumberyard-employee-dm:
> > Is Jenkins now trying to actually attach a debugger to the process when a death test fails?
> 
> Actually found out that the googletest package that was built has death tests disabled, and the failing test was a death test. It needed to be wrapped with an #if/#endif guard like the other DEATH tests

That is not right.
We need to fix the googletest package for Linux to enable the DEATH tests.

dtamkin1 (approved):

spham-amzn:
> > > Is Jenkins now trying to actually attach a debugger to the process when a death test fails?
> > 
> > 
> > Actually found out that the googletest package that was built has death tests disabled, and the failing test was a death test. It needed to be wrapped with an #if/#endif guard like the other DEATH tests
> 
> That is not right. We need to fix the googletest package for Linux to enable the DEATH tests.
That is an issue with the 3P package, not O3DE. Create GHI : https://github.com/o3de/3p-package-source/issues/184

lumberyard-employee-dm:
> > > > Is Jenkins now trying to actually attach a debugger to the process when a death test fails?
> > > 
> > > 
> > > Actually found out that the googletest package that was built has death tests disabled, and the failing test was a death test. It needed to be wrapped with an #if/#endif guard like the other DEATH tests
> > 
> > 
> > That is not right. We need to fix the googletest package for Linux to enable the DEATH tests.
> 
> That is an issue with the 3P package, not O3DE. Create GHI : [o3de/3p-package-source#184](https://github.com/o3de/3p-package-source/issues/184)

Thanks for making the ticket!

cgalvan (approved):

lumberyard-employee-dm (approved):

Code/Framework/AzCore/Platform/Common/UnixLike/AzCore/Debug/Trace_UnixLike.cpp
```diff
@@ -121,21 +123,16 @@ namespace AZ::Debug
...
+        // Continue to exit the process when a signal is caught by this signal handler. 
+        // The exit code will conform to https://tldp.org/LDP/abs/html/exitcodes.html.
+        // - Signal based exit codes will be calculated based on 128+signal
+        // - Exit codes will not be greater than 255
+        //
+        // Any exit code that results in anything outside of this range will result in 255: Exit status out of range
+        constexpr int exitCodeSignalBase = 128;
+        constexpr int maxExitCode = 255;
+        int signalExitCode = exitCodeSignalBase + signal;
+        _exit(signalExitCode<maxExitCode?signalExitCode:maxExitCode);
```
lumberyard-employee-dm, line 135: whitespace would be nice.
```suggestion
        _exit(signalExitCode < maxExitCode ? signalExitCode : maxExitCode);
```

lumberyard-employee-dm (approved):

cgalvan (approved):

PR #14892, opened 2023-03-02
Bugfix for device access during initialization.

jhmueller-huawei:
## What does this PR do?

The issue is that during device initialization, at least for Vulkan, a BufferPoolDescriptor is already created, which accesses the device in the RHISystem, before it finished adding.

 Also, there is a minor change creating a new variable instead of accessing a vector thrice.

### DIFF
--- a/Gems/Atom/RHI/Code/Source/RHI/RHISystem.cpp
+++ b/Gems/Atom/RHI/Code/Source/RHI/RHISystem.cpp
@@ -40,12 +40,13 @@ namespace AZ
     
         void RHISystem::Init()
         {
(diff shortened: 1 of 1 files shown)


rgba16f:
There is a parallel change to fix the same crash - https://github.com/o3de/o3de/pull/14891

I prefer the fix here though as it also guards against out of range accesses.

moudgils:
> There is a parallel change to fix the same crash - #14891
> 
> I prefer the fix here though as it also guards against out of range accesses.

I did contemplate using this but went with .empty() check. It may be better to do the range check as an assert.

moudgils (approved):

PR #14917, opened 2023-03-03
Add support for reverting property override in DPE Entity Inspector

forkercat:
## What does this PR do?

_Feature is behind two flags: `ed_enableDPE` and `EnableInspectorOverrideManagement`_

This PR adds support for reverting property override in DPE-based Entity Inspector.

Change Summary:
- Change to have `PrefabComponentAdapter` that subclasses `DPE::ComponentAdapter`
    - `ComponentEditor` now has a dependency on the prefab adapter, but it can be removed later once we change to handle component adapter creation in `EntityPropertyEditor`.
- Add new callback attribute `RevertOverride` to handle revert operation

Inspector Override RFC: https://github.com/o3de/sig-content/issues/112

Recording Demo:

https://user-images.githubusercontent.com/11157226/222603981-437610bb-5ebf-498d-8913-f6b97b2e1715.mp4

(I sometimes hit ctrl+s to save the level so the * on the level prefab would disappear)

## How was this PR tested?

- [x] Update existing unit tests to validate new attributes are present if the property is overridden
- [x] Manually tested in Editor with undo/redo support

Issues (will update the list if I notice more issues):
1. For light component, changing light type from null to spot light will create a sphere shape component as an add-component override. Reverting the light type property won't remove that component. But I think this is okay and users can delete that add-component override manually.


srikappa-amzn (commented):
I love how smooth the iteration with the new dpe based inspector is! There are no visible refreshes of the component editors, which is neat.

srikappa-amzn (commented):
Do the 'revert' changes propagate to other prefab instances like they should? For example, if the level has 2 'supercars' that have the car prefabs under them and if you focus on the supercar prefab and make/revert an override , the other car should also get the override

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/ComponentEditor.cpp
```diff
@@ -150,7 +151,15 @@ namespace AzToolsFramework
...
         {
             // Instantiate the DPE without the RPE
-            m_adapter = AZStd::make_shared<AZ::DocumentPropertyEditor::ComponentAdapter>();
+            if (Prefab::IsInspectorOverrideManagementEnabled())
+            {
+                m_adapter = AZStd::make_shared<Prefab::PrefabComponentAdapter>();
+            }
+            else
+            {
+                m_adapter = AZStd::make_shared<AZ::DocumentPropertyEditor::ComponentAdapter>();
```
srikappa-amzn, line 160: is this all we need to ensure that opening UI editor with dpe enabling doesn't crash?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/DocumentPropertyEditor/PrefabComponentAdapter.cpp
```diff
@@ -0,0 +1,81 @@
...
+        using PrefabPropertyEditorNodes::PrefabOverrideLabel;
+
+        adapterBuilder->BeginPropertyEditor<PrefabOverrideLabel>();
+        adapterBuilder->Attribute(PrefabOverrideLabel::Text, labelText);
+
+        AZ::Dom::Path relativePathFromEntity;
+        if (!serializedPath.empty())
+        {
+            relativePathFromEntity /= PrefabDomUtils::ComponentsName;
+            relativePathFromEntity /= m_componentAlias;
```
srikappa-amzn, line 45: might be good to throw an assert here if the alias is empty as that should never happen.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/DocumentPropertyEditor/PrefabComponentAdapter.cpp
```diff
@@ -0,0 +1,81 @@
...
+            adapterBuilder->Attribute(PrefabOverrideLabel::IsOverridden, false);
+        }
+        else
+        {
+            bool isOverridden = m_prefabOverridePublicInterface->AreOverridesPresent(m_entityId, relativePathFromEntity.ToString());
+            adapterBuilder->Attribute(PrefabOverrideLabel::IsOverridden, isOverridden);
+
+            if (isOverridden)
+            {
+                adapterBuilder->Attribute(PrefabOverrideLabel::RelativePath, relativePathFromEntity.ToString());
```
srikappa-amzn, line 61: why is this only added when the property is overridden? Regardless of it being an override or not, the relative path from entity would stay the same and might be useful to add it all the time so that non-override use cases can also take advantage of it.

srikappa-amzn (commented):
I believe this code was added so that propagation still happens when DPE is enabled : https://github.com/o3de/o3de/blob/development/Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DPEComponentAdapter.cpp#L101-L144


If you override 'HandleMessage', then the above code might never get called and entities never get marked dirty and propagation never happens.

I think propagation wouldn't work with these changes because the above code never gets called?

If that's the case, I believe that code should either 
1. Get moved to PrefabComponentAdapter::HandleMessage function
2. Or it stays there and at the end of PrefabComponentAdapter::HandleMessage, we call DPEComponentAdapter::HandleMessage()

Whether we do 1 or 2 depends on whether DPEComponentEditor on its own without the prefab system would want to mark the entities dirty or not. My gut says it wouldn't and we can pick option 1 but probably good to check with @amzn-tmryan and @alexmontAmazon

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/DocumentPropertyEditor/PrefabComponentAdapter.cpp
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    AZ::Dom::Value PrefabComponentAdapter::HandleMessage(const AZ::DocumentPropertyEditor::AdapterMessage& message)
+    {
+        auto handleRevertOverride = [this](AZStd::string_view relativePathFromEntity)
+        {
+            m_prefabOverridePublicInterface->RevertOverrides(m_entityId, relativePathFromEntity);
+        };
+
+        message.Match(PrefabPropertyEditorNodes::PrefabOverrideLabel::RevertOverride, handleRevertOverride);
```
michabr, line 76: Should the return value for this be checked, or does it not return anything?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/DocumentPropertyEditor/PrefabOverrideLabelHandler.cpp
```diff
@@ -64,7 +67,9 @@ namespace AzToolsFramework::Prefab
 
         if (selectedItem == revertAction)
         {
-            // This is the place where revert override code will be added later.
+            AZStd::string_view relativePathFromEntity =
+                PrefabPropertyEditorNodes::PrefabOverrideLabel::RelativePath.ExtractFromDomNode(m_node).value_or("");
+            PrefabPropertyEditorNodes::PrefabOverrideLabel::RevertOverride.InvokeOnDomNode(m_node, relativePathFromEntity);
```
michabr, line 72: It would be nice to have a way to send a message to the node or getting the node associated with the handler instead of caching it.

michabr (commented):
Nice! Left some questions in the comments.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/ComponentEditor.cpp
michabr: @srikappa-amzn I am working on that part in a separate PR.

michabr:
> Do the 'revert' changes propagate to other prefab instances like they should? For example, if the level has 2 'supercars' that have the car prefabs under them and if you focus on the supercar prefab and make/revert an override , the other car should also get the override

Looks like they should because m_prefabOverridePublicInterface->RevertOverrides() handles the undo/redo nodes etc.

forkercat:
> Do the 'revert' changes propagate to other prefab instances like they should? For example, if the level has 2 'supercars' that have the car prefabs under them and if you focus on the supercar prefab and make/revert an override , the other car should also get the override

Thanks for mentioning this. Yes I verified that and it worked as expected. So when making and reverting that override, I need to focus on the supercar. I saw that the other supercar gets updated accordingly.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/DocumentPropertyEditor/PrefabComponentAdapter.cpp
forkercat: Yeah I can add it. I also think it would be good for debugging purpose.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/DocumentPropertyEditor/PrefabComponentAdapter.cpp
forkercat: The type is `AZ::Dom::Value`. It currently always returns null since the callback does not return anything.

We can take advantage of it if we want to do something more based on the result.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DPEComponentAdapter.cpp
```diff
@@ -146,17 +146,6 @@ namespace AZ::DocumentPropertyEditor
...
     void ComponentAdapter::CreateLabel(AdapterBuilder* adapterBuilder, AZStd::string_view labelText, AZStd::string_view serializedPath)
     {
-        auto* prefabAdapterInterface = AZ::Interface<AzToolsFramework::Prefab::PrefabAdapterInterface>::Get();
-        if (prefabAdapterInterface)
-        {
-            AZ::Dom::Path relativePathFromEntity;
-            if (!serializedPath.empty())
-            {
-                relativePathFromEntity /= AzToolsFramework::Prefab::PrefabDomUtils::ComponentsName;
-                relativePathFromEntity /= m_componentAlias;
```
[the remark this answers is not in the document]
srikappa-amzn: Yes, m_componentAlias  should definitely be part of the PrefabComponentAdapter because it is nothing but the identifier of a component in a serialized prefab json.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/DocumentPropertyEditor/PrefabOverrideLabelHandler.cpp
forkercat: I notice that the cached node is also needed for doing invoking 'RevertOverride'. It does check if the attribute exists in the node in `InvokeOnDomNode(m_node)`. Afaik there is no other place to bind the node to the handler, unless we cache the node in `SetValueFromDom()`.

fyi I followed the pattern that we used in [ContainerActionButtonHandler](https://github.com/o3de/o3de/blob/development/Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/ContainerActionButtonHandler.cpp#L36).

michabr: Oh, I see, so we're already using this pattern elsewhere. Thanks for the info!

srikappa-amzn:
> I believe this code was added so that propagation still happens when DPE is enabled : https://github.com/o3de/o3de/blob/development/Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DPEComponentAdapter.cpp#L101-L144
> 
> If you override 'HandleMessage', then the above code might never get called and entities never get marked dirty and propagation never happens.
> 
> I think propagation wouldn't work with these changes because the above code never gets called?
> 
> If that's the case, I believe that code should either
> 
> 1. Get moved to PrefabComponentAdapter::HandleMessage function
> 2. Or it stays there and at the end of PrefabComponentAdapter::HandleMessage, we call DPEComponentAdapter::HandleMessage()
> 
> Whether we do 1 or 2 depends on whether DPEComponentEditor on its own without the prefab system would want to mark the entities dirty or not. My gut says it wouldn't and we can pick option 1 but probably good to check with @amzn-tmryan and @alexmontAmazon

Never mind. I see that you are already calling DPEComponentAdapter::HandleMessage(). So, propagation should continue to work.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/DocumentPropertyEditor/PrefabComponentAdapter.cpp
michabr: OK. I guess there's no way to know whether the base class ComponentAdapter::HandleMessage still needs to be called? I'm still not super familiar with the DPE message handling.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/DocumentPropertyEditor/PrefabComponentAdapter.cpp
```diff
@@ -30,6 +30,15 @@ namespace AzToolsFramework::Prefab
     {
     }
 
+    void PrefabComponentAdapter::SetComponent(AZ::Component* componentInstance)
+    {
+        // Set the component alias before calling SetValue() in base SetComponent().
+        // Otherwise, an empty component alias will be used in DOM data.
+        m_componentAlias = componentInstance->GetSerializedIdentifier();
```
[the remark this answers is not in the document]
forkercat: oh yeah. Forgot to add assert here as well

srikappa-amzn (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/DocumentPropertyEditor/PrefabComponentAdapter.cpp
forkercat: I guess maybe return value of Match() can tell whether the message is matched or not. I am still looking into why it always return null. Maybe @amzn-tmryan knows. But I think it's okay to not check this since base class HandleMessage()s do Match() as well.

Even if we change to handle property edits in PrefabComponentAdapter we still need to call ComponentAdapter::HandleMessage() to "handle" the message. [ReflectionAdapter::HandleMessage()](https://github.com/o3de/o3de/blob/development/Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DPEComponentAdapter.cpp#L142) might accept the message.

michabr (approved):

forkercat:
Update after rebase to choose prefab component adapter when both DPE and InspectorOverride flags are on.

Testing in editor...

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/DocumentPropertyEditor/PrefabComponentAdapter.cpp
amzn-tmryan: The reason you always see null returned by Match is because the handlers being executed don't return anything:
https://github.com/o3de/o3de/blob/0a666643fd39795fbb33383764017a8ec8e1a4d7/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/DocumentSchema.h#L579-L590

If a handler returns a value, then it will be returned by Match.

alexmontAmazon (approved):
looks good :)

amzn-tmryan (approved):

forkercat:
I noticed a Lua spawnable test is failing in Prefab Periodic Suite. 

> FAILED - AutomatedTesting\Gem\PythonTests\Prefab\TestSuite_Periodic.py::TestAutomationNoOverrides::test_Lua_Spawnables_NestedSpawn[windows-AutomatedTesting-windows_editor] - Failed: Test test_Lua_Spawnables_NestedSpawn.

After some investigation and discussion, I think it is an intermittent issue that could not be consistently reproduced. Since the AR is green, I am merging this PR to unblock other work.

PR #14918, opened 2023-03-03
SerializeContext header include dependency reduction 

lumberyard-employee-dm:
## What does this PR do?

The goal of the changes is to remove unnecessary inclusion of SerializeContext.h from header files .

Whenever possible if a Reflect function was defined in header, it's
definition was moved to a cpp file and the header was replaced with a
forward declaration to ReflectContext.

The SerializeContext VersionConverter functions that were members of a class declaration were
instead moved to cpp files as static functions.
Declaring the SerializeContext VersionConverter functions in header
inhibits forward declaration of the SerializeContext due to the need to
access the local class of SerializeContext::DataElementNode

The AnimSplineTrack specializations for Vec2 for both the LyShine and
Maestro Gems were moved to a cpp file, with forward declarations for
the `TAnimSplineTrack<Vec2>` and `TUiAnimSplineTrack<Vec2>` member
functions added to remove the include of the
AnimSplineTrack_Vec2Specialization.cpp file

Finally the `AZ_TYPE_INFO_SPECIALIZE*` macros were moved from TypeInfo.h
to TypeInfoSimple.h which allows code to avoid including `TypeInfo.h`
when needing to use that macro.

The TypeInfo.h include specializes AZ TypeInfo for AZStd container
templates such as `AZStd::vector`, `AZStd::unordered_map`, etc...
So this should save a little bit of include time.

---
Added non-template helper functions for the SerializeContext to reduce compiling code code in frequently used templates.

The `SerializeContext::Class`, `SerializeContext::ClassBuilder::Class`
and `SerializeContext::ClassData::Create` functions have had
non-template helper functions added that encapsulates the non-template
logic in the SerializeContext.cpp file.

## How was this PR tested?

Built all of O3DE using Windows clang with unity builds turned off.

Measures show a 6% reduction in non-compilation times when building in with unity builds turned.
i.e `LY_UNITY_BUILD=OFF`

### Before Changes
```
Analyzing build trace from 'D:/Playground/output.dat'...
**** Time summary:
Compilation (14018 times):
  Parsing (frontend):        77580.2 s
  Codegen & opts (backend):  35841.6 s
```

### After Changes
```
Analyzing build trace from 'D:/Playground/output.dat'...
**** Time summary:
Compilation (14029 times):
  Parsing (frontend):        74311.3 s
  Codegen & opts (backend):  32472.1 s
```

Measurement statistics for gathered on a 32-core/64-hyperthread  machine with 64-GiB of Ram are attached.
[development-no-unity.1b552ded3f5830947068029a333820fe1735b76d.txt](https://github.com/o3de/o3de/files/10877166/development-no-unity.1b552ded3f5830947068029a333820fe1735b76d.txt)
[serialize-context-no-unity.1b552ded3f5830947068029a333820fe1735b76d.txt](https://github.com/o3de/o3de/files/10877165/serialize-context-no-unity.1b552ded3f5830947068029a333820fe1735b76d.txt)


gadams3 (approved):
Are you using something to do static analysis, remove unused headers, and autoformat?

lumberyard-employee-dm:
> Are you using something to do static analysis, remove unused headers, and autoformat?

Yes.
For the compile timings a combination of the `-ftime-trace` compiler option and the ClangBuildAnalyzer is used to record compile timings using clang
1. O3DE needs to be configured to build using clang the the `-ftime-trace` compile flag, which records compilation timing for each cpp file that is compiled in a JSON file. 
2. The [ClangBuildAnalyzer](https://github.com/aras-p/ClangBuildAnalyzer) tool captures the recorded artifacts output to the build folder and can analyze that data to output a text filke

This is done on Windows by building both LLVM and the ClangBuildAnalyzer from source

For unused headers, the [include-what-you-use](https://github.com/include-what-you-use/include-what-you-use) tool is used to determine which includes are used by a cpp file.

moraaar (approved):
Thanks for doing this! I've reviewed the Physics related files. I have feedback about one of the changes, but nothing blocking.

Gems/Atom/RHI/Code/Include/Atom/RHI.Reflect/PipelineLibraryData.h
```diff
@@ -53,7 +65,10 @@ namespace AZ
             PipelineLibraryData(AZStd::vector<uint8_t>&& data);
             PipelineLibraryData() = default;
 
-            AZ_SERIALIZE_FRIEND();
+            template <typename, typename>
+            friend struct AnyTypeInfoConcept;
+            template <typename, bool, bool>
+            friend struct Serialize::InstanceFactory;
```
mbalfour-amzn, line 71: I'm not particularly a fan of getting rid of the macro and replacing it with the expanded version. By doing this we lose the comments explaining what it is and why it's here, and it's not as easy to understand how to add it to new classes. I won't block the PR over it, but please at least consider bringing back the macro in a much smaller SerializeContextFriend.h or something.

Gems/EMotionFX/Code/EMotionFX/Source/BlendTreeBlendNNode.cpp
```diff
@@ -645,7 +645,7 @@ namespace EMotionFX
         data->SetTrajectoryDeltaMirrored(delta);
     }
 
-    bool BlendTreeBlendNNode::VersionConverter(AZ::SerializeContext& context, AZ::SerializeContext::DataElementNode& classElement)
+    static bool BlendTreeBlendNNodeVersionConverter(AZ::SerializeContext& context, AZ::SerializeContext::DataElementNode& classElement)
```
mbalfour-amzn, line 648: I wonder if we should just be removing these version converters entirely, since they aren't compatible with prefabs anyways...

Gems/EMotionFX/Code/MCore/Source/StringIdPool.cpp
```diff
@@ -10,8 +10,12 @@
 #include "StringIdPool.h"
 #include "LogManager.h"
 
+#include <AzCore/Serialization/SerializeContext.h>
+
 namespace MCore
 {
+    AZ_TYPE_INFO_SPECIALIZE_WITH_NAME_IMPL(StringIdPoolIndex, "StringIdPoolIndex", "{C374F051-8323-49DB-A1BD-C6B6CF0333C0}");
```
mbalfour-amzn, line 17: Does this need to have the name "MCore::StringIdPoolIndex" or is it never serialized out to storage?

Gems/GraphModel/Code/Source/Model/Module/ModuleNode.cpp
```diff
@@ -32,16 +33,14 @@ namespace GraphModel
...
         : Node(ownerGraph)
         , m_moduleGraphFileId(moduleGraphFileId)
     {
         // The module file name (without extension) is the node title
-        if (!AzFramework::StringFunc::Path::GetFileName(moduleGraphFileName.data(), m_nodeTitle))
-        {
-            AZ_Error(ownerGraph->GetSystemName(), false, "Could not get node name from file string [%s]", moduleGraphFileName.data());
-        }
+        m_nodeTitle = AZ::IO::PathView(moduleGraphFileName).Filename().String();
+        AZ_Error(ownerGraph->GetSystemName(), !m_nodeTitle.empty(), "Could not get node name from file string [%.*s]", AZ_STRING_ARG(moduleGraphFileName));
```
mbalfour-amzn, line 43: Nice change!

mbalfour-amzn (approved):

Gems/EMotionFX/Code/MCore/Source/StringIdPool.cpp
lumberyard-employee-dm: This is never serialized to storage, so the including the name of "MCore::" isn't needed.

cgalvan (approved):

Gems/PhysX/Code/Source/Pipeline/MeshGroup.h
```diff
@@ -211,7 +211,10 @@ namespace PhysX
             const ConvexDecompositionParams& GetConvexDecompositionParams() const;
 
         protected:
-            static bool VersionConverter(AZ::SerializeContext& context, AZ::SerializeContext::DataElementNode& classElement);
+            // void* parameter is used for the ClassElement to avoid the need to reference
+            // the SerializeContext::ClassElement type. That would require the include
+            // of SerializeContext
+            static bool VersionConverter(AZ::SerializeContext& context, void* classElement);
```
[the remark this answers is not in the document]
lawsonamzn: I'm not saying that its wrong to remove it (it may be correct) but bear in mind that the files that O3DE has in its projects is a tiny window into all the files that exist somewhere out there in user's machines.  Searching for a particular thing in the files that are part of O3DE doesn't provide full confidence that it doesn't still exist somewhere in say, a user project.  However, it can provide evidence of how old it is / ie, before it became O3DE / multiple years ago / etc.

Gems/EMotionFX/Code/EMotionFX/Source/BlendTreeBlendNNode.cpp
lawsonamzn: Yes, as long as the objectstream format is not being used anymore (ie, its json now) or its a product (which gets remade whenever a version is bumped).  The objectsream VersionConverters will not acutally port to json format, even if someone made the effort, the actual structure is different, so they are not "save-able".

However, some files still use objectstream, as the source (right?) and if these objects appear in there, the version converter must be maintained.

Gems/PhysX/Code/Source/Pipeline/MeshGroup.h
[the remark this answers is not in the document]
lumberyard-employee-dm: Doing a check on the MeshGroup VersionConveter reflection code itself, it is only for `MeshGroup` instances of [version 1 and below](https://github.com/o3de/o3de/blame/ec52f77dde0d3e7e334810a7241adf1085b8156d/Gems/PhysX/Code/Source/Pipeline/MeshGroup.cpp#L856-L914)

At initial commit to O3DE to reflect the `MeshGroup` class, it was [Version 2](https://github.com/o3de/o3de/blame/9a63716b39d6e83ac76559a5eaedce00227fa0c9/Gems/PhysX/Code/Source/Pipeline/MeshGroup.cpp#L596-L608).

Therefore the only content that could have a lower version of the `MeshGroup` type instance serialized would be within the O3DE repo, at the initial commit time of the O3DE code.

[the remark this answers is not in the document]
lawsonamzn: kill it with fire then :)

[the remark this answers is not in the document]
lumberyard-employee-dm: A couple of the AutomatedTetring Physics assetinfo files that came with O3DE contains version 1 data for the `MeshGroup` class: 
https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/o3de/o3de%24+version%3D%221%22+type%3D%22%7B5B03C8E6&patternType=standard&sm=1&groupBy=path

There are 2 files with Version 1 data in them currently
* [Material_PerFaceMatgerialGetsMaterial/test.fbx.assetinfo](https://github.com/o3de/o3de/blob/development/AutomatedTesting/Levels/Physics/Material_PerFaceMaterialGetsCorrectMaterial/test.fbx.assetinfo#L5)
* [Collider_MultipleSurfaceSlots/test.fbx.assetinfo](https://github.com/o3de/o3de/blob/development/AutomatedTesting/Assets/Physics/Collider_MultipleSurfaceSlots/test.fbx.assetinfo#L5)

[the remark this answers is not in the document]
lumberyard-employee-dm: I used the `SerializeContextTools` convert command to load those two files using ObjectStream with the Version Converter for Version 1 and saved backout using json with the command of
```shell
<build-dir>/bin/debug/SerializeContextTools.exe convert --project-path=D:/o3de/o3de/AutomatedTesting --specializations=editor --files=D:/o3de/o3de/AutomatedTesting/Levels/Physics/Material_PerFaceMaterialGetsCorrectMaterial/test.fbx.assetinfo;D:/o3de/o3de/AutomatedTesting/Assets/Physics/Collider_MultipleSurfaceSlots/test.fbx.assetinfo  -ext=json
```

Gems/Atom/RHI/Code/Include/Atom/RHI.Reflect/PipelineLibraryData.h
lumberyard-employee-dm: Micro include added as [SerializeContext_fwd.h](https://github.com/aws-lumberyard-dev/o3de/blob/serialize-context-class-builder-updates/Code/Framework/AzCore/AzCore/Serialization/SerializeContext_fwd.h)

lawsonamzn (approved):
How likely are these changes to affect 3rd Party Gems?  if its _just_ modifying our own headers using these things that'd be one thing, but I notice changes to things like SerializeContext.h itself.

Will 3p gems have to be updated to still compile?  It might be worth trying popcornfx or cesium or something with it...

basically, conditional approval here.  If we have to go tell everyone who's ever made a gem to change their code, that'd be a no-go.

lumberyard-employee-dm:
> How likely are these changes to affect 3rd Party Gems? if its _just_ modifying our own headers using these things that'd be one thing, but I notice changes to things like SerializeContext.h itself.
> 
> Will 3p gems have to be updated to still compile? It might be worth trying popcornfx or cesium or something with it...
> 
> basically, conditional approval here. If we have to go tell everyone who's ever made a gem to change their code, that'd be a no-go.

The changes to the SerializeContext are backwards compatible and no additional includes are needed by 3rdParty gem authors.

Type aliases have been added for all the old inner classes and will continue to funciton.

Gems/PhysX/Code/Source/Pipeline/MeshGroup.h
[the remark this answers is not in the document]
lumberyard-employee-dm: The only way for users to have left over assets that would need the version converter, is if they are trying to use fbx.assetinfo files created with Lumberyard in O3DE.
Otherwise at minimum the user would have a version 2 MeshGroup serialized ObjectStream file, where the Version Converter wouldn't do anything as it checks specify for two cases
1. That the version is less than 1 to remove the Material Rule: https://github.com/o3de/o3de/blob/development/Gems/PhysX/Code/Source/Pipeline/MeshGroup.cpp#L832-L835
2. That the version is less than 2 to move all the fields to subclasses: https://github.com/o3de/o3de/blob/development/Gems/PhysX/Code/Source/Pipeline/MeshGroup.cpp#L832-L855

Code/Framework/AzCore/AzCore/Memory/Memory_fwd.h
```diff
@@ -0,0 +1,369 @@
...
+    #define aznewex(_Name)                                          new
+
+/// azmalloc(size)
+    #define azmalloc_1(_1)                                          AZ::AllocatorInstance< AZ::SystemAllocator >::Get().allocate(_1)
+/// azmalloc(size,alignment)
+    #define azmalloc_2(_1, _2)                                      AZ::AllocatorInstance< AZ::SystemAllocator >::Get().allocate(_1, _2)
+/// azmalloc(size,alignment,Allocator)
+    #define azmalloc_3(_1, _2, _3)                                  AZ::AllocatorInstance< _3 >::Get().allocate(_1, _2)
+
+/// azcreate(class,params)
```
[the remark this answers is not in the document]
lumberyard-employee-dm: I'll examine the logic for the macros.

The only change was to move the Macros to this file, but I'll give it a thorough examination.

Code/Framework/AzCore/AzCore/Serialization/SerializeContext.cpp
```diff
@@ -888,2398 +878,2612 @@ namespace AZ
...
-    int SerializeContext::DataElementNode::AddElement(SerializeContext& sc, const char* name, const ClassData& classData)
+    SerializeContext::ClassBuilder* SerializeContext::ClassBuilder::DataContainer(IDataContainer* dataContainer)
     {
-        (void)sc;
-        AZ_Assert(name != nullptr && strlen(name) > 0, "Empty name is an INVALID element name!");
-        u32 nameCrc = Crc32(name);
-
-    #if defined(AZ_ENABLE_TRACING)
-        if (!m_classData->m_container && FindElement(nameCrc) != -1)
+        if (m_context->IsRemovingReflection())
```
spham-amzn, line 1153: nit: This logic can be inverted to remove a `return this`

```
if (!m_context->m_context->IsRemovingReflection())
{
    m_classData->second.m_container = dataContainer; // Only set if we have not removed the class data
}
return this;

File: Code/Framework/AzCore/AzCore/Serialization/SerializeContext.cpp [hunk 8]
```diff
@@ -888,2398 +878,2612 @@ namespace AZ
...
-                                    if (actualClassData)
-                                    {
-                                        resolvedObject = elementData->m_azRtti->Cast(resolvedObject, actualClassData->m_azRtti->GetTypeId());
-                                    }
-                                }
+                                #if defined (AZ_ENABLE_SERIALIZER_DEBUG)
+                                // Additional error information: Provide Type IDs and the serialization stack from our enumeration
+                                AZStd::string sourceTypeID = dataClassInfo->m_typeId.ToString<AZStd::string>();
+                                AZStd::string targetTypeID = classElement->m_typeId.ToString<AZStd::string>();
+                                AZStd::string error = AZStd::string::format("EnumarateElements RTTI Cast Error: %s -> %s", sourceTypeID.c_str(), targetTypeID.c_str());
```
spham-amzn, line 1238:
```suggestion
                                AZStd::string error = AZStd::string::format("EnumerateElements RTTI Cast Error: %s -> %s", sourceTypeID.c_str(), targetTypeID.c_str());
```

Code/Framework/AzCore/AzCore/Memory/Memory_fwd.h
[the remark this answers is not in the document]
lumberyard-employee-dm: The `#if defined(_RELEASE)` block has been removed

Code/Framework/AzCore/AzCore/Serialization/SerializeContext.cpp
```diff
@@ -888,2398 +878,2612 @@ namespace AZ
...
         }
+        m_classData->second.m_doSave = doSave;
+        return this;
     }
 
     //=========================================================================
-    // SetName
-    // [1/16/2013]
+    // EnumerateInstanceConst
+    // [10/31/2012]
```
spham-amzn, line 1189: That is so long ago

Code/Framework/AzCore/AzCore/Serialization/SerializeContext.cpp
```diff
@@ -888,2398 +878,2612 @@ namespace AZ
...
-
-                    if (errorHandler)
-                    {
-                        errorHandler->ReportError(error.c_str());
-                    }
-                    else
-                    {
-                        AZ_Error("Serialize", false, "%s", error.c_str());
-                    }
+            }
```
spham-amzn, line 1246: Looks like this is not returning any boolean (fall through)

Code/Framework/AzCore/AzCore/Serialization/SerializeContext.cpp
lumberyard-employee-dm: Those comments are also useless.
The data provides the reader with no meaningful information.

The name of the function is extra point failure for mismatched documentation as it would need to be kept in sync as well, for no meaningful information.
This makes the comments have an overall net negative value to the code.

Code/Framework/AzCore/AzCore/Serialization/SerializeContext.cpp
```diff
@@ -888,2398 +878,2612 @@ namespace AZ
...
                         }
                         else
                         {
-                            if (m_element.m_id == childElement->m_typeId)
-                            {
-                                classElement = *childElement;
-                                elementFound = true;
-                            }
+                            AZ_Error("Serialization", false, "Failed to find class data for 'Dynamic Serializable Field' with type=%s address=%p. Make sure this type is reflected, \
+                                otherwise you will lose data during serialization!\n", dynamicFieldDesc->m_typeId.ToString<AZStd::string>().c_str(), dynamicFieldDesc->m_data);
```
spham-amzn, line 1355: I think this will include the entire string so a huge gap in between `reflected,` and `otherwise` will appear in the message.
```suggestion
                            AZ_Error("Serialization", false, "Failed to find class data for 'Dynamic Serializable Field' with type=%s address=%p. Make sure this type is reflected, "\
                                "otherwise you will lose data during serialization!\n", dynamicFieldDesc->m_typeId.ToString<AZStd::string>().c_str(), dynamicFieldDesc->m_data);
```

lumberyard-employee-dm: In that case the backslash isn't needed then.

Code/Tools/SceneAPI/SceneCore/DataTypes/IManifestObject.h
```diff
@@ -22,11 +26,21 @@ namespace AZ
         }
         namespace DataTypes
         {
-            class IManifestObject
+            class SCENE_CORE_CLASS IManifestObject
             {
             public:
-                AZ_RTTI(IManifestObject, "{3B839407-1884-4FF4-ABEA-CA9D347E83F7}");
-                static void Reflect(AZ::ReflectContext* context);
+                SCENE_CORE_API virtual AZ::TypeId RTTI_GetType() const;
```
spham-amzn, line 32: I thought if SCENE_CORE_CLASS defines the declspec import/export, you don't need to do it for the members

lumberyard-employee-dm: Well SCENE_CORE_CLASS is not defined on [Windows platforms](https://github.com/o3de/o3de/blob/a10351f38ddf50af2b01ed071c04cda52d6069ed/Code/Tools/SceneAPI/SceneCore/SceneCoreConfiguration.h#L18).
It is only set for non-Windows.

lawsonamzn (approved):

PR #14929, opened 2023-03-03
Asset Processor - Metadata Create Delay

amzn-mike:
## What does this PR do?

Add configurable delay before processing a new file which would create a metadata file.

Also made it so AP doesn't try to create metadata files in the middle of a move/rename command from editor.

This avoids some cases where renaming causing a new metadata to be created immediately.  It is mainly focused on the use case of a person trying to manually rename a file.  It may help some with source control updates but isn't intended to directly address that case.

Delay is currently defaulted to 0 (disabled) for now.  This can be enabled with `--regset="/O3DE/AssetProcessor/Settings/Metadata/MetaCreationDelayMs=5000"`

## How was this PR tested?

Manually + added/ran unit tests


AMZN-stankowi (approved):

Code/Tools/AssetProcessor/native/AssetManager/AssetRequestHandler.cpp
```diff
@@ -466,9 +471,9 @@ void AssetRequestHandler::OnRequestAssetExistsResponse(NetworkRequestID groupID,
         AZ_TracePrintf(AssetProcessor::DebugChannel, "OnRequestAssetExistsResponse: No such compile group found, ignoring.\n");
         return;
     }
-    
-    AZ_TracePrintf(AssetProcessor::DebugChannel, "GetAssetStatus / CompileAssetSync: Asset %s is %s.\n", 
-        located.value().GetDisplayString().toUtf8().constData(), 
+
+    AZ_TracePrintf(AssetProcessor::DebugChannel, "GetAssetStatus / CompileAssetSync: Asset %s is %s.\n",
```
lmbr-pip, line 475: nit: Please use `AZ_Trace`

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
```diff
@@ -5826,4 +5847,208 @@ namespace AssetProcessor
...
+                {
+                    return true;
+                }
+            }
+        }
+
+        return false;
+    }
+
+    bool AssetProcessorManager::IsDelayProcessTimerElapsed(qint64 elapsedTime)
```
lmbr-pip, line 5922: nit: HasDelayProcessTimerElapsed reads slightly better IMHO, though may not be worth effort to rename if only change in PR

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.h
```diff
@@ -495,6 +520,18 @@ namespace AssetProcessor
         typedef QHash<QString, FileEntry> FileExamineContainer;
         FileExamineContainer m_filesToExamine; // order does not actually matter in this (yet)
 
+        // Set of files which are metadata-enabled but don't have a metadata file.
+        // These files will be delayed for processing for a short time to wait for a metadata file to show up.
+        AZStd::unordered_map<AZ::IO::Path, QDateTime> m_delayProcessMetadataFiles;
```
lmbr-pip, line 525: Do we have any concerns about size at edge cases here? Do we have a sense of memory footprint per tracked file? Imagine this set should always be small(ish)

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.h
```diff
@@ -495,6 +520,18 @@ namespace AssetProcessor
...
+        // Indicates if DelayedMetadataFileCheck has already been queued to run or not.
+        bool m_delayProcessMetadataQueued = false;
+        // Max delay time before creating a metadata file.
+        // Avoid setting this too high as it will delay processing of new files.
+        AZ::u32 m_metaCreationDelayMs = 0;
+        // Set of files/folders that have been reported as pending for move.  bool: false = old file path, true = new file path
+        AZStd::unordered_map<AZ::IO::Path, bool> m_pendingMoves;
```
lmbr-pip, line 532: Same is true here.

lmbr-pip (commented):
LG overall. Thank you for adding unit tests.

Would be curious about what stressing we've done/memory size tracking? Obviously want to get this in and optimize where appropriate. But I think it would be good to talk to edge cases and testing performed.

Ie 100,000 files being renamed etc because the top level folder of my game assets needs to change or similar.

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.h
amzn-mike: Shouldn't be any real concerns, there are lots of other maps APM is managing and this one will be cleared out every few seconds

Code/Tools/AssetProcessor/native/AssetManager/AssetRequestHandler.cpp
amzn-mike: Updated

Code/Tools/AssetProcessor/native/AssetManager/assetProcessorManager.cpp
amzn-mike: Updated

lawsonamzn:
FWIW we've tried a delay like this before in a different system that also used metafiles (many years ago) and found that it _can_ improve things but isn't a silver bullet.  Specifically, P4 and GIT and other source control systems (and including windows file moves) can have random patterns when dealing with files, and with filters like LFS in place you often see a pattern where, for example, it gets files in layers by _type_ rather than by folder, and it can be minutes between getting a file and getting its meta file.

So this might help, but any time based solution is going to be problematic compared to a fence based solution or some other eventing.  However, eventing and fencing is not always available.

lawsonamzn:
Also, in situations _under our control_ such as when files are moved in the editor using the editor's gui, could we not move the metafiles first?

lawsonamzn (commented):
The code looks okay but shouldn't we respond to file change notifications for the files by _resetting_ their timers, to catch the case of a big file being modified (written to) over and over again?  once it settles we can process it.

Also, why start with 0 timeout?  We only have to apply this timeout to files that have no meta (completely new files) so we can give it some room here of a few seconds as it won't affect the actual iteration loop of modify, modify, modify...

amzn-mike:
> Also, in situations _under our control_ such as when files are moved in the editor using the editor's gui, could we not move the metafiles first? I also assume that if you move a normal file using the editor GUI, it should move the metafile first even if the metafile was not selected... and of course, any tools inside our toolbox or editor that _create_ metafiles should do so before they create the corresponding asset - or use a temp file and smart move...

Moving the metadata file first would cause the original source file to trigger, which then no longer has a metadata file and it would create one.  There is a smaller part of this change which adds a system for indefinitely pausing processing for a move that's about to occur, and its hooked into the API which the editor uses for renaming files.  So when a file is renamed via the asset browser, AP will use that system instead of the X second delay system.

amzn-mike:
> The code looks okay but shouldn't we respond to file change notifications for the files by _resetting_ their timers, to catch the case of a big file being modified (written to) over and over again? once it settles we can process it. FWIW the "true worst case" we saw was adobe photoshop and a autodesk product whihc would open a file for writing, write the header to it, then _close the file_ and then several seconds later, reopen it to write the file data (ugh!)
> 
> Also, why start with 0 timeout? We only have to apply this timeout to files that have no meta (completely new files) so we can give it some room here of a few seconds as it won't affect the actual iteration loop of modify, modify, modify...

Hmm, I suppose we could reset the timer in that case.  I was mainly focused on the use-case of a human renaming a file, but to your earlier point, in the case of a slow source control system, resetting while it writes some large file could help avoid creating metadata files when they already exist.

As for defaulting to 0 for now, I wanted to make this opt-in for the time being as I'm concerned about the impact on newly created files taking longer to start processing.

amzn-mike:
> FWIW we've tried a delay like this before in a different system that also used metafiles (many years ago) and found that it _can_ improve things but isn't a silver bullet. Specifically, P4 and GIT and other source control systems (and including windows file moves) can have random patterns when dealing with files, and with filters like LFS in place you often see a pattern where, for example, it gets files in layers by _type_ rather than by folder, and it can be minutes between getting a file and getting its meta file.
> 
> So this might help, but any time based solution is going to be problematic compared to a fence based solution or some other eventing. However, eventing and fencing is not always available.

Good point to bring up.  This is definitely not a silver bullet; this was mainly meant as a tool for people to use so they can keep AP running while they move/rename files, as the alternative solution is having to shutdown AP every time you want to rename files.  For updating from source control, the best workaround right now is probably to still shutdown AP.  Perhaps we could add a button in AP to pause it, but that's been contentious in the past.

lawsonamzn (approved):

Issue #12834, opened 2022-10-27
Feature Request: Multiple Camera Workflow through Bitflags


ParasiteSt:
**Is your feature request related to a problem? Please describe.**
Right now O3DE does not have any way to render two cameras and put them together into a final picture for the player. This is a vital piece to most FPS/FPP games.

**Describe the solution you'd like**
A solution for this would to add on the camera component a Bitflag category which would specify what that camera sees and renders. The Bitflags would be set on a per entity basis, all starting with a default bitflag which is what a camera would render by default as well. Setting Bitflags on meshes/visual assets would be done either by having a category per component, for example the Mesh component would have a category to set what Bitflag it is, or the better approach is to have a new component called Bitflag which would need to be added to entities to change the default bitflag to a custom one you have created. 

**Describe alternatives you've considered**
An alternative is to set up all materials to be able to be drawn in any order that overrides the depth pass, but this would be a less attractive way of doing this.

**Additional context**
This Feature is a large one, and would more than likely have to be added in steps. Beginning with allowing the camera component to render based off of a bitflag/flag. Then after the camera can support it, I would suggest one of two methods, the first would be the best end result and that is creating the new component that holds all bitflags, and can be assigned to any entity that holds a visual component. The other way would be to introduce bitflags to one visual component like the mesh component, and after it is working reuse the code for all other components like Actor, etc. But this method has a lot of reusing the same code, that should be stored within its own component.

ParasiteSt:
![CameraBitflagAddition](https://user-images.githubusercontent.com/100494538/198187589-2fdfe877-956b-42d9-8858-ab2ee640c926.png)
![BitflagComponentIdea](https://user-images.githubusercontent.com/100494538/198187598-6f6daf37-9643-4846-a705-02846a3942b1.png)
This is just a couple quick drafts in photoshop of what it _could_ look like. I also think the bitflag addition may not be added directly from the component, and instead it has the same dropdown as the camera, and bitflags are set up in a separate file that is one per project, possibly utilizing the Asset Editor window for creating new bitflags.

rgba16f:
It seems like Atom can already do what is described. We can split what is rendered by render pipeline pass, so as an example ui can be rendered separately from the main 3d scene. In the described example a 3d ui pass could be added to render first person character hands, etc.

Other cases can be handled as a camera rendering to a texture that is then used in the main viewport.

If I've misunderstood please give a more detailed example of what cannot be achieved with Atom at the moment.

[...]

PR #14934, opened 2023-03-03
Add RenderToTexture function to camera component

VickyAtAZ:
## What does this PR do?
The camera component can specify an AttachmentImageAsset which it renders to.
The camera component can specify the root pass template it may use for its render pipeline.

To test this feature, we may create a UICanvas which uses this AttachmentImageAsset for its image component. Then add this UICanvas to an entity inside the level via UICanvas Ref component. When enter the game mode in Editor, it should show the render target image on screen in UI. 

## How was this PR tested?

Tested in Editor for AutomatedTesting project. 
Tested with AutomatedTesting.GameLauncher. 

Test steps in Editor:
1. Create a new level
2. Add an entity with camera component and set it as the activate camera for the level
3. Add another entity with camera, set the camera1.attimage (from the following package) to its Render Texture. 
4. Create a new UICanvas file. Add an image element to the canvas, select image type Render Target, then choose the camera1.attimage. 
5. Add an entity in the level with UICanvas AssetRef component. Choose the UICanvas file for this component.
6. Enter game mode in Editor. You should see the captured image showing in the UI

Test resources:
[RenderTargets.zip](https://github.com/o3de/o3de/files/10886646/RenderTargets.zip)

ezgif-5-eeff6e738f

## extra info
Related RHI:
https://github.com/o3de/o3de/issues/12834
https://github.com/o3de/o3de/issues/14362

### DIFF
--- a/Gems/Atom/RPI/Code/Include/Atom/RPI.Public/RenderPipeline.h
+++ b/Gems/Atom/RPI/Code/Include/Atom/RPI.Public/RenderPipeline.h
@@ -98,6 +98,9 @@ namespace AZ
                                                                    const ViewType viewType = ViewType::Default);
             static RenderPipelinePtr CreateRenderPipelineForWindow(Data::Asset<AnyAsset> pipelineAsset, const WindowContext& windowContext);
 
+            // Create a render pipeline which renders to the specified attachment image
+            static RenderPipelinePtr CreateRenderPipelineForImage(const RenderPipelineDescriptor& desc, Data::Asset<AttachmentImageAsset> imageAsset);
+
             // Data type for render pipeline's views' information
             using PipelineViewMap = AZStd::map<PipelineViewTag, PipelineViews, AZNameSortAscending>;
             using ViewToViewTagMap = AZStd::map<const View*, PipelineViewTag>;
@@ -283,6 +286,9 @@ namespace AZ
             // End of functions accessed by Scene class
             //////////////////////////////////////////////////
 
+            // update viewport and scissor based on pass tree's output
+            void UpdateViewportScissor();
+
             RenderMode m_renderMode = RenderMode::RenderEveryTick;
 
             // The Scene this pipeline was added to.
@@ -335,6 +341,10 @@ namespace AZ
 
             // View type associated with the Render Pipeline.
             ViewType m_viewType = ViewType::Default;
+
+            // viewport and scissor for frame update
+            RHI::Viewport m_viewport;
+            RHI::Scissor m_scissor;
         };
 
     } // namespace RPI
--- a/Gems/Atom/RPI/Code/Include/Atom/RPI.Public/RenderPipeline.h
+++ b/Gems/Atom/RPI/Code/Include/Atom/RPI.Public/RenderPipeline.h
@@ -98,7 +98,10 @@ namespace AZ
                                                                    const ViewType viewType = ViewType::Default);
             static RenderPipelinePtr CreateRenderPipelineForWindow(Data::Asset<AnyAsset> pipelineAsset, const WindowContext& windowContext);
 
-            // Create a render pipeline which renders to the specified attachment image
+            //! Create a render pipeline which renders to the specified attachment image
+            //! The render pipeline's root pass is created from the pass template specified from RenderPipelineDescriptor::m_rootPassTemplate
+            //! The input AttachmentImageAsset is used to connect to first output attachment of the root pass template
+            //! Note: the AttachmentImageAsset doesn't need to be loaded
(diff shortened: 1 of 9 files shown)


Gems/Camera/Code/Source/CameraComponentController.cpp
```diff
@@ -26,7 +28,7 @@ namespace Camera
         if (auto serializeContext = azrtti_cast<AZ::SerializeContext*>(context))
         {
             serializeContext->Class<CameraComponentConfig, AZ::ComponentConfig>()
-                ->Version(4)
+                ->Version(5)
```
AMZN-alexpete, line 31: It may be good to increase the minor version in `gems/Camera/gems.json` from 0.0.0 to 0.1.0 to account for this new capability

Gems/Camera/Code/Source/CameraComponentController.cpp
```diff
@@ -280,6 +299,28 @@ namespace Camera
...
+    void CameraComponentController::CreateRenderPipelineForTexture()
+    {
+        auto scene = AZ::RPI::RPISystemInterface::Get()->GetSceneByName(AZ::Name("Main"));
+
+        const AZStd::string pipelineName = "Camera_" + m_entityId.ToString() + "_Pipeline";
+
+        auto defaultPipelineDesc = scene->GetDefaultRenderPipeline()->GetDescriptor();
+
+        // Use default render pipeline for the camera render pipeline
+        // We can add function to use customized render pipeline for camera in the future
```
AMZN-alexpete, line 311: That would be cool

Gems/Camera/Code/Source/CameraComponentController.cpp
```diff
@@ -280,6 +299,28 @@ namespace Camera
         DeactivateAtomView();
     }
 
+    void CameraComponentController::CreateRenderPipelineForTexture()
+    {
+        auto scene = AZ::RPI::RPISystemInterface::Get()->GetSceneByName(AZ::Name("Main"));
+
+        const AZStd::string pipelineName = "Camera_" + m_entityId.ToString() + "_Pipeline";
+
+        auto defaultPipelineDesc = scene->GetDefaultRenderPipeline()->GetDescriptor();
```
AMZN-alexpete, line 308: This `defaultPipelineDesc` variable does not appear to be used

AMZN-alexpete (approved):
Nice!  Just one possibly unused variable. 
In the future it would be nice to be able to avoid the PostProcessing effects in the main scene from affecting the post processing in the render to texture image - or is it already possible to prevent the main pipeline post processing settings affecting the render to texture pipeline?

VickyAtAZ:
> 

The post effect layer is supposed to associate to camera. So if a camera entity doesn't have postfx component, then it shouldn't have any post effect. I think the implementation was setup that way. But it's possible if it's broken. And we should fix it.

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/RenderPipeline.h
```diff
@@ -98,6 +98,9 @@ namespace AZ
                                                                    const ViewType viewType = ViewType::Default);
             static RenderPipelinePtr CreateRenderPipelineForWindow(Data::Asset<AnyAsset> pipelineAsset, const WindowContext& windowContext);
 
+            // Create a render pipeline which renders to the specified attachment image
```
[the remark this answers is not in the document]
VickyAtAZ: Sure. I will update the comment. 

Basically for the current root pass templates, such MainPipeline.pass or the new CameraPipeline.pass, we can't just create a pass from it and make it the root pass of the render pipeline. This is because they don't have output attachment specified. 
The CreateRenderPipelineForWindow function create a SwapchainPass as root pass which provide input for the MainPipeline, in other case such as preview renderer, it creates RenderToTexture pass which provides a transient attachment as the output for the previewRenderer pass template. Other than providing output attachment, they were also responding to setup viewport and scissor.

Gems/Camera/Code/Source/CameraComponentController.cpp
VickyAtAZ: It's actually using a root pass template which can be specified in `m_config.m_pipelineTemplate`. So user can specify other template other than the one in Atom common feature gem.

VickyAtAZ: I removed the comment.

Gems/Camera/gem.json
```diff
@@ -1,6 +1,6 @@
 {
     "gem_name": "Camera",
-    "version": "0.0.0",
+    "version": "0.1.0",
```
AMZN-alexpete, line 3: thank you!

AMZN-alexpete (approved):

moudgils (approved):

PR #14939, opened 2023-03-04
Projected shadows calculate once across multiple pipelines in a scene

invertednormal:
## What does this PR do?

This PR changes the way projected shadows work so if there are multiple pipelines in the same scene, only one pipeline will process the projected and projected esm shadows. This can lead to considerable performance improvements in scene with many cameras like robotics simulation. 

The projected feature processor has taken over many responsibilities that used to be owned by the projected shadow pass. It finds the relevant passes and tells them what to do, like which resource to use for their output, whether they should be enabled, etc. It also creates and destroys the passes that do individual shadow renders. It has to take great care any time there are any pipeline changes to keep the primary render passes up to date.

The way this was accomplished wasn't particularly straightforward and the pass system really needs a more generic way for passes to accomplish this. Shadows aren't the only case of render passes that only need to be run once per scene.

## How was this PR tested?

This was primarily tested with a custom test scene and two viewports. Some additional testing is still in progress to make sure nothing is broken by this change.


Gems/Atom/Feature/Common/Code/Source/Shadows/ProjectedShadowFeatureProcessor.cpp
```diff
@@ -52,8 +59,15 @@ namespace AZ::Render
...
-        
-        CachePasses();
+
+        auto pipelines = GetParentScene()->GetRenderPipelines();
+        AZStd::for_each(pipelines.begin(), pipelines.end(),
+            [this](RPI::RenderPipelinePtr pipelinePtr)
+            {
+                CachePasses(pipelinePtr.get());
+            }
+        );
```
AMZN-alexpete, line 69: nit, would something simpler be sufficient here?
```suggestion
        for(auto& renderPipeline: GetParentScene()->GetRenderPipelines())
        {
            CachePasses(renderPipeline->get());
        );
```

Gems/Atom/RPI/Code/Source/RPI.Public/RPISystem.cpp
```diff
@@ -119,10 +119,6 @@ namespace AZ
             Debug::TraceMessageBus::Handler::BusConnect();
 #endif
             m_descriptor = rpiSystemDescriptor;
-
-            // set the default multisample state to MSAA 4x
-            // the default render pipeline may override this when it is loaded
-            m_multisampleState.m_samples = 4;
```
AMZN-alexpete, line 125: did you mean to remove this?

Gems/Atom/Feature/Common/Code/Source/Shadows/ProjectedShadowFeatureProcessor.cpp
invertednormal: yeah much better, I even wrote the loop that way in another place further down, not sure what I was thinking here.

Gems/Atom/RPI/Code/Source/RPI.Public/RPISystem.cpp
invertednormal: It is kinda unrelated, but it also seems quite wrong. I think it was set to 4 back when we used to use msaa 4x as the norm, but we use 2x more regularly now, and the RPI shouldn't assume msaa usage at all.

Gems/Atom/RPI/Code/Source/RPI.Public/Image/ImageSystem.cpp
```diff
@@ -209,6 +209,68 @@ namespace AZ
...
+            createImageRequest.m_imageDescriptor = imageDescriptor;
+            createImageRequest.m_imageName = "SystemAttachmentImage";
+            createImageRequest.m_isUniqueName = false;
+
+            RHI::ImageViewDescriptor viewDesc = RHI::ImageViewDescriptor::Create(imageDescriptor.m_format, 0, 0);
+            viewDesc.m_aspectFlags = formatAspectFlag;
+
+            auto systemAttachmentImage = RPI::AttachmentImage::Create(createImageRequest);
+            m_systemAttachmentImages[format] = systemAttachmentImage;
+            return systemAttachmentImage;
```
michalpelka, line 271: Compiling with clang it complained that the value from the stack is returned as reference here `-Wreturn-stack-address`

Gems/Atom/Feature/Common/Code/Source/Shadows/ProjectedShadowFeatureProcessor.h
```diff
@@ -14,6 +14,7 @@
 #include <Atom/Feature/Utils/MultiSparseVector.h>
 #include <CoreLights/EsmShadowmapsPass.h>
 #include <CoreLights/ProjectedShadowmapsPass.h>
+#include <CoreLights/ShadowmapPass.h>
```
michalpelka, line 17: With clang, the header was missing:
```
#include <Atom/RPI.Public/Shader/Shader.h>
```

michalpelka (commented):
I found two minor issues testing this PR against the robotic application.
To experience a successful build on clang and ubuntu I needed to address two issues that are mentioned in the review. Thanks!

Gems/Atom/Feature/Common/Code/Source/CoreLights/ProjectedShadowmapsPass.cpp
```diff
@@ -60,302 +46,64 @@ namespace AZ
...
+                AttachImageToSlot(Name("Shadowmap"), depthAttachmentImage);
                 this->SetEnabled(false);
+                return;
             }
 
-            RPI::PassAttachmentBinding& binding = GetOutputBinding(0);
-            binding.SetAttachment(attachment);
-
+            AttachImageToSlot(Name("Shadowmap"), m_atlasAttachmentImage);
+            this->SetEnabled(true);
```
VickyAtAZ, line 92: nit: remove "this->"

Gems/Atom/Feature/Common/Code/Source/CoreLights/ProjectedShadowmapsPass.cpp
```diff
@@ -60,302 +46,64 @@ namespace AZ
...
-            drawPacketBuilder.SetDrawArguments(AZ::RHI::DrawLinear(1, 0, 3, 0));
-
-            AZ::RHI::DrawPacketBuilder::DrawRequest drawRequest;
-            drawRequest.m_listTag = m_clearShadowShader->GetDrawListTag();
-            drawRequest.m_pipelineState = pipelineState;
-            drawRequest.m_sortKey = AZStd::numeric_limits<RHI::DrawItemSortKey>::min();
-
-            drawPacketBuilder.AddDrawItem(drawRequest);
-            m_clearShadowDrawPacket = drawPacketBuilder.End();
+            QueueForBuildAndInitialization();
```
VickyAtAZ, line 98: Can you add comment (in the code) about why this QueueForBuildAndInitialization is called here?

Gems/Atom/RPI/Code/Source/RPI.Public/Image/ImageSystem.cpp
VickyAtAZ: Yes. I had similar issue. And I changed it to `return m_systemAttachmentImage[format]`

Gems/Atom/Feature/Common/Code/Source/Shadows/ProjectedShadowFeatureProcessor.cpp
```diff
@@ -299,50 +315,222 @@ namespace AZ::Render
...
+        }
+        if (changeType == RPI::SceneNotification::RenderPipelineChangeType::Added || changeType == RPI::SceneNotification::RenderPipelineChangeType::PassChanged)
+        {
+            CachePasses(renderPipeline);
+        }
+
+        // Check to see if the primary passes have changed, and if so removes the children from the old primary pass and creates them
+        // on the new primary pass. This is necessary if an earlier render pipeline adds or removes references to the shadow map
+        // passes, forcing this feature processor to change which pipeline it uses to render shadows for all pipelines in the scene.
+        UpdatePrimaryPasses();
```
VickyAtAZ, line 352: It seems UpdatePrimaryPasses() would set new primary passes. But the CachePasses() function is using primary passes to disable non-primary passes. With current order, CachePasses() is called before the new primary passes are set. Would it be a problem?

Gems/Atom/RPI/Code/Source/RPI.Public/Pass/Pass.cpp
```diff
@@ -1320,7 +1320,12 @@ namespace AZ
                 return;
             }
 
-            AZ_Error("PassSystem", m_state == PassState::Idle, "Pass::FrameBegin - Pass [%s] is attempting to render, but is not in the Idle state.", m_path.GetCStr());
+            if (m_state != PassState::Idle)
+            {
+                AZ_Error("PassSystem", m_state == PassState::Idle,
+                    "Pass::FrameBegin - Pass [%s] is attempting to render, and should be in the 'Idle' or 'Queued' state, but is in the '%s' state.",
+                    m_path.GetCStr(), ToString(m_state).data());
+            }
```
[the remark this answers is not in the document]
invertednormal: Ah yes, I had it this way so I could break-point on it and need to change it back. Thanks.

Gems/Atom/Feature/Common/Code/Source/Shadows/ProjectedShadowFeatureProcessor.cpp
invertednormal: Should be ok. The check is just comparing pointers in the esm case, the primary pass is never actually accessed. when there's some initialization going on and the primary esm pass hasn't yet been set, it'll disable it in `CachePasses()` then re-enable it later in `UpdateFilterParameters()`

Gems/Atom/RPI/Code/Source/RPI.Public/RenderPipeline.cpp
```diff
@@ -443,6 +443,9 @@ namespace AZ
 
                 // Reset change flags
                 m_pipelinePassChanges = PipelinePassChanges::NoPassChanges;
+
+                // Process any changes that may have happened due to SceneNotification Events.
+                m_passTree.ProcessQueuedChanges();
```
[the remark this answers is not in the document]
invertednormal: I think it needs to stay where it is actually. `ProcessQueuedChanges()` can flip `m_pipelinePassChanges` away from the `NoPassChanges` state. That still needs to be handled even if it's not till next frame.

Gems/Atom/RPI/Code/Source/RPI.Public/Image/ImageSystem.cpp
```diff
@@ -209,6 +209,68 @@ namespace AZ
...
+                formatBindFlag = RHI::ImageBindFlags::DepthStencil;
+                formatAspectFlag = RHI::ImageAspectFlags::DepthStencil;
+                break;
+            }
+
+            RHI::ImageDescriptor imageDescriptor;
+            imageDescriptor.m_size = RHI::Size(1, 1, 1);
+            imageDescriptor.m_format = format;
+            imageDescriptor.m_arraySize = 1;
+            imageDescriptor.m_bindFlags = formatBindFlag | RHI::ImageBindFlags::ShaderRead;
```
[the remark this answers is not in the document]
invertednormal: I left it this way because I couldn't think of  a reason you'd want write access to these. Actually if any shader did write to one of these temp images it could change the value in the pixel, which would be bad if you were expecting black for instance. This is more in your purview though so if you really think we should make it writable I'll change it.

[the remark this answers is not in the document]
VickyAtAZ: Yeah. I prefer we add ShaderWrite which is just to have safe fallback when it was written to by mistake

Gems/Atom/RPI/Code/Source/RPI.Public/RenderPipeline.cpp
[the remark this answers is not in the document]
VickyAtAZ: Maybe update the comment? since the intention is not aligned with the comment

Gems/Atom/RPI/Code/Source/RPI.Public/Image/ImageSystem.cpp
```diff
@@ -209,6 +209,68 @@ namespace AZ
...
+            imageDescriptor.m_bindFlags = formatBindFlag | RHI::ImageBindFlags::ShaderReadWrite;
+            imageDescriptor.m_sharedQueueMask = RHI::HardwareQueueClassMask::All;
+
+            RPI::CreateAttachmentImageRequest createImageRequest;
+            createImageRequest.m_imagePool = m_systemAttachmentPool.get();
+            createImageRequest.m_imageDescriptor = imageDescriptor;
+            createImageRequest.m_imageName = "SystemAttachmentImage";
+            createImageRequest.m_isUniqueName = false;
+
+            RHI::ImageViewDescriptor viewDesc = RHI::ImageViewDescriptor::Create(imageDescriptor.m_format, 0, 0);
```
AMZN-alexpete, line 266: Should the view descriptor be passed to the createImageRequest?
```suggestion
            createImageRequest.m_imageViewDescriptor = RHI::ImageViewDescriptor::Create(imageDescriptor.m_format, 0, 0);
```

AMZN-alexpete (approved):
Looks good to me aside from the view descriptor that doesn't appear to be used in `GetSystemAttachmentImage`,  seems like it should be assigned or removed.

VickyAtAZ (approved):

Gems/Atom/RPI/Code/Source/RPI.Public/Image/ImageSystem.cpp
invertednormal: After chatting with @VickyAtAZ, looks like we can just remove. Will have a fix up for this shortly.

AMZN-alexpete (approved):

PR #14942, opened 2023-03-04
Remove unused function SetPropertyVisibility.

Jackie9527:
## What does this PR do?

Fix the following warning when building with clang-cl shipped with VS2022.
```
error : parameter 'flags' set but not used [-Werror,-Wunused-but-set-parameter]
```

The I found that the whole function is not used.

## How was this PR tested?

Rebuild o3de using VS2022 cl and clang-cl.


moraaar (approved):

gadams3 (approved):

lawsonamzn (approved):
I kinda assumed the bug here was that flags was passed in without a ref.

If you remove it and nobody uses it then I guess no harm done until someone needs to use it.

cgalvan (approved):

PR #14943, opened 2023-03-04
Fix linking error duplicate symbols when building with clang-cl.

Jackie9527:
## What does this PR do?

Fix linking error duplicate symbols when building with clang-cl.

## How was this PR tested?

Rebuild o3de using VS2022 cl and clang-cl.

### DIFF
--- a/Gems/DebugDraw/Code/CMakeLists.txt
+++ b/Gems/DebugDraw/Code/CMakeLists.txt
@@ -59,7 +59,10 @@ if(PAL_TRAIT_BUILD_HOST_TOOLS)
                 DEBUGDRAW_GEM_EDITOR=1
         BUILD_DEPENDENCIES
             PRIVATE
-                Gem::DebugDraw.Static
+                AZ::AtomCore
+                Gem::Atom_RPI.Public
+                Gem::Atom_Bootstrap.Headers
+                Legacy::CryCommon
                 AZ::AzToolsFramework
     )
 


gadams3 (approved):

Gems/DebugDraw/Code/CMakeLists.txt
```diff
@@ -59,7 +59,10 @@ if(PAL_TRAIT_BUILD_HOST_TOOLS)
                 DEBUGDRAW_GEM_EDITOR=1
         BUILD_DEPENDENCIES
             PRIVATE
-                Gem::DebugDraw.Static
+                AZ::AtomCore
+                Gem::Atom_RPI.Public
+                Gem::Atom_Bootstrap.Headers
+                Legacy::CryCommon
```
lumberyard-employee-dm, line 65: It looks like here that the [debugdraw_editor_files.cmake](https://github.com/o3de/o3de/blob/development/Gems/DebugDraw/Code/debugdraw_editor_files.cmake) contains a superset of the [debugdraw_files.cmake](https://github.com/o3de/o3de/blob/development/Gems/DebugDraw/Code/debugdraw_files.cmake) from the Static library target.

We could remove the same files contained in the `debugdraw_editor_files.cmake` and still link the `Gem::DebugDraw.Static`

lumberyard-employee-dm (approved):

lawsonamzn (approved):

cgalvan (approved):

Gems/DebugDraw/Code/CMakeLists.txt
Jackie9527: https://github.com/o3de/o3de/blob/602b668c35203c7c3db590c0a8ae126de8cac306/Gems/DebugDraw/Code/CMakeLists.txt#L59

When building `DebugDraw.Editor`, it needs a macro `DEBUGDRAW_GEM_EDITOR` to enable some functions that are not used in runtime.
That's why it doesn't work to remove the same files contained in the `debugdraw_editor_files.cmake` and still link the `Gem::DebugDraw.Static`.

lumberyard-employee-dm: Thanks for the explanation. It is not ideal that we reuse the same files for both Editor and non-Editor using `#if`, but it is fine.

PR #14949, opened 2023-03-05
Material Canvas: Support for nodes accessing local and world space vertex attributes

gadams3:
## What does this PR do?

- Updating shaders to rename normal, tangent, bitangent variables to make it clear they’re in world space
- Cleaning up eval function macro parameters, formatting, and spacing
- Updating material pipeline surface eval functions to pass in untransformed vertex data
- Replacing defines with constants for material canvas integration
- Adding constants for local and world space vertex attributes
- Updating and adding nodes for normal, tangent, bitangent in local and world space

## How was this PR tested?

Confirmed that all automated testing shader, material, and model assets compiled successfully in the AP.
Confirmed that new material canvas local and world normal nodes operate as expected.
Confirmed that atom sample viewer assets compiled successfully in the AP.
Attempting to confirm that all atom sample viewer screenshot tests pass but the full test suite stops progressing after The dynamic material test. Investigating...

Corresponding ASV PR https://github.com/o3de/o3de-atom-sampleviewer/pull/592

### DIFF
--- a/Gems/Atom/Feature/Common/Assets/Materials/Types/StandardMultilayerPBR.azsli
+++ b/Gems/Atom/Feature/Common/Assets/Materials/Types/StandardMultilayerPBR.azsli
@@ -558,7 +558,7 @@ option enum class OpacityMode {Opaque, Cutout, Blended, TintedTransparent} o_opa
         return surface;
     }
 
-    Surface EvaluateSurface_StandardMultilayerPBR(PixelGeometryData geoData)
+    Surface EvaluateSurface_StandardMultilayerPBR(VsOutput IN, PixelGeometryData geoData)
     {
         return EvaluateSurface_StandardMultilayerPBR(
             geoData.positionWS,
@@ -585,7 +585,7 @@ option enum class OpacityMode {Opaque, Cutout, Blended, TintedTransparent} o_opa
         return ShouldHandleParallax();
     }
     
-    void EvaluateSurface(PixelGeometryData geoData)
+    void EvaluateSurface(VsOutput IN, PixelGeometryData geoData)
     {
         // do nothing, this is where alpha clip can be done if it's supported
     }
--- a/Gems/Atom/Feature/Common/Assets/Materials/Types/StandardMultilayerPBR.azsli
+++ b/Gems/Atom/Feature/Common/Assets/Materials/Types/StandardMultilayerPBR.azsli
@@ -119,14 +119,14 @@ option enum class OpacityMode {Opaque, Cutout, Blended, TintedTransparent} o_opa
     PixelGeometryData EvaluatePixelGeometry_StandardMultilayerPBR(
         inout float4 positionSV,
         float3 positionWS,
-        float3 normal,
-        float3 tangent,
-        float3 bitangent,
+        float3 normalWS,
+        float3 tangentWS,
+        float3 bitangentWS,
         float2 uvs[UvSetCount],
         bool isFrontFace,
         float3 vertexBlendMask)
     {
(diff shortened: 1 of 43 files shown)


galibzon (approved):

Gems/Atom/Feature/Common/Assets/Materials/Types/StandardMultilayerPBR.azsli
```diff
@@ -145,15 +145,15 @@ option enum class OpacityMode {Opaque, Cutout, Blended, TintedTransparent} o_opa
         float4x4 objectToWorld = ObjectSrg::GetWorldMatrix();
         float3x3 objectToWorldIT = ObjectSrg::GetWorldMatrixInverseTranspose();
 
-        float3 vertexNormal, vertexTangent, vertexBitangent;
```
amzn-tommy, line 148: nit: I would consider renaming this to vertexNormalWS to help delineate a vertex normal in world space vs. the surface normal.

Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/GraphData/Nodes/MaterialOutputs/StandardPBR/MaterialGraphName_SurfaceEval.azsli
```diff
@@ -91,7 +91,7 @@ Surface EvaluateSurface_MaterialGraphName(
         surface.clearCoat.roughness = max(inClearCoatRoughness, 1e-3);
 
         // Flip normal if back face is rendered   
-        surface.clearCoat.normal = normalize(o_clearCoat_normal_override_enabled ? GetWorldSpaceNormal((float2)inClearCoatNormal, vertexNormal, tangents[0], bitangents[0]) : vertexNormal);
```
amzn-tommy, line 94: Here is an example where I think using vertexNormalWS would be beneficial. It's a bit confusing to see That GetWorldSpaceNormal takes a normalWS as input.

cgalvan (approved):

amzn-tommy (approved):
Is there a use-case where EvaluateSurface needs to work with the vertex data in object space, rather than the geoData that is already in world space? I would expect all the object space work to be done in EvaluatePixelGeometry.

Gems/Atom/Tools/MaterialCanvas/Assets/MaterialCanvas/GraphData/Nodes/MaterialOutputs/StandardPBR/MaterialGraphName_SurfaceEval.azsli
gadams3: I see your point. What about renaming the function as well?

gadams3:
> Is there a use-case where EvaluateSurface needs to work with the vertex data in object space, rather than the geoData that is already in world space? I would expect all the object space work to be done in EvaluatePixelGeometry.

Current material canvas templates use evaluate vertex data and evaluate surface data as “entry points” to inject all of the generated code. Right before each code injection site, there is a block of specially named constants and defines for all the vertex attributes and other data that can be referenced in material graph node code snippets. Before the material pipeline refactor, they were all defined at the top of the vertex and pixel shader functions, which contained all of the logic in one function instead of spread across multiple evaluate functions. With everything broken up into stages, the material canvas nodes still need access to all of the values they were expecting and any new ones we add later. Instead of passing the vertex data into the function, we could use standard or base pixel and surface data structures that contain all the required variables and default values.

PR #14963, opened 2023-03-06
Create MeshInstanceManager for managing draw packets

amzn-tommy:
## What does this PR do?

When r_enableMeshInstancing is true, the draw packets are managed by the MeshInstanceManager, and when it is false, they continue to be managed by the ModelDataInstance. Either way, at this stage the draw calls are still not-instanced and they are still submitted by the culling system.

Adding a sync point between Init and UpdateCulling jobs has ~.3ms overhead on the 10k levels on my machine, which is why I didn't have the non-instanced path also go through the MeshInstanceManager. That sync point is unnecessary when everything is owned by the ModelDataInstance, so this overhead doesn't exist when r_enableMeshInstancing is false. This should be fixable with the StableDynamicStructOfArrays work, but we don't want to introduce any overhead in the meantime when instancing is disabled.

- Created an r_enableMeshInstancing cvar. There is no actual instancing at this stage, but it hides any of the extra work and keeps the existing draw submission path working.
- Created a ReInit function that makes it easier to destroy and re-create the resources that are managed by the MeshFeatureProcessor. This is also used to make it so the instancing cvar can be changed at runtime.
- Changed ModelDataInstance functions to take MeshFeatureProcessor, to prevent constantly changing them every time need to access something is added or removed (e.g. mesh instance manager, ray tracing feature processor, etc.)
- Created a ReInit function that makes it easier to destroy and re-create the resources that are managed by the MeshFeatureProcessor.
- Created an r_enableMeshInstancing cvar. There is no actual instancing at this stage, but it hides any of the extra work and keeps the existing draw submission path working.
- Created a MeshInstanceManager, which maps meshes to instance groups, and uses ref-counting to track how many total meshes belong to a group. The MeshFeatureProcessor is responsible for mirroring calls to Add and Remove to ensure the refcount is up to date.
- Made pageIterator type public on StableDynamicArray so that the MeshInstanceManager can return page iterators for parallel jobs
- Added == and != operators to StableDynamicArrayWeakHandle, which is useful for unit testing.

## How was this PR tested?

 Tested in the Editor's 10kEntity and 10kVegInstances test levels with Vulkan + DX12 and device validation enabled. Tested the cvar with both true and false, and toggling between the two, and tested with and without the TaskGraph enabled (which determines whether or not Simulate uses a parent job).
 Ran AtomSampleViewer tests with and without r_enableMeshInstancing
 Unit tests
 Tested entering and exiting game mode in MultiplayerSample's Alpha level


jiaweig-amzn (approved):

invertednormal (commented):
Updating review with what I have so far, just nits. Still working on MeshFeatureProcessor.cpp

invertednormal (approved):

Issue #14848, opened 2023-02-27
Play in editor (ctrl + g) doesn't work when AWSGameLift gem is enabled


allisaurus:
**Describe the bug**
After enabling the AWSGameLift gem on the project by adding it to the list of enabled gems and adding the `Gem::AWSGameLift.Client.Static` and `Gem::AWSGameLift.Server.Static` targets to the game's client & server targets, the Editor and launchers build, but fail to run with the following errors:

```
...
(MultiplayerEditor) - Editor TCP connection attempt #2.
[Error] (EditorServer) - [2023-02-27 13:49:46] [info] asio async_connect error: system:10061 (No connection could be made because the target machine actively refused it)
[Error] (EditorServer) - [2023-02-27 13:49:46] [info] Error getting remote endpoint: system:10057 (A request to send or receive data was disallowed because the socket is not connected and (when sending on a datagram socket using a sendto call) no address was supplied)
[Error] (EditorServer) - [2023-02-27 13:49:46] [error] handle_connect error: Underlying Transport Error
(MultiplayerEditor) - Editor TCP connection attempt #3.
[Error] (EditorServer) - [2023-02-27 13:49:53] [info] asio async_connect error: system:10061 (No connection could be made because the target machine actively refused it)
[Error] (EditorServer) - [2023-02-27 13:49:53] [info] Error getting remote endpoint: system:10057 (A request to send or receive data was disallowed because the socket is not connected and (when sending on a datagram socket using a sendto call) no address was supplied)
[Error] (EditorServer) - [2023-02-27 13:49:53] [error] handle_connect error: Underlying Transport Error

... (eventually stops trying after failed attempt #5)
```

**Assets required**
A working O3DE multiplayer game, such as [o3de-multiplayersample](https://github.com/o3de/o3de-multiplayersample)

**Steps to reproduce**
1. Follow [the instructions to enable the AWS GameLift gem](https://www.o3de.org/docs/user-guide/gems/reference/aws/aws-gamelift/gem-setup/#2-enable-the-aws-gamelift-gem)
2. Rebuild the project editor and server launcher
3. Start the editor
4. Hit `ctrl` + `g` to start playing
5. See error logs described above

**Expected behavior**
In-editor play of the multiplayer game would work.

**Actual behavior**
Game play did not work, appears server failed to start.

**Found in Branch**
development, `40682b7ef30`

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Windows 10
 - CPU:  Intel
[...]

PR #14964, opened 2023-03-06
Add sv_useGameLiftServer CVAR to allow local play without GameLift

allisaurus:
## What does this PR do?

Addresses https://github.com/o3de/o3de/issues/14848

Some questions I'd like feedback on:
1. what's the best way to _enable_ GameLift for local play, add `sv_useGameLiftServer=true` in the `server.cfg`? Will this work for playing in the Editor?
2. should we have a corresponding CVAR for the client, i.e., `cl_useGameLift`? This would mostly prevent the client from connecting to the various GameLift request buses, which may help performance, but not sure its worth complicating things with an additional setting to flip.

## How was this PR tested?

* Using [o3de-multiplayersample](https://github.com/o3de/o3de-multiplayersample) with AWSGameLift gem enabled, can run `ctrl` + `g` in Editor w/o running GameLiftLocal.
* setting `"sv_useGameLiftServer": "true"` in _commands.multiplayersample_serverlauncher.setreg_ and running server launcher shows connection to GL local

### DIFF
--- a/Gems/AWSGameLift/Code/AWSGameLiftServer/Source/AWSGameLiftServerSystemComponent.cpp
+++ b/Gems/AWSGameLift/Code/AWSGameLiftServer/Source/AWSGameLiftServerSystemComponent.cpp
@@ -19,6 +19,8 @@
 
 namespace AWSGameLift
 {
+    AZ_CVAR(bool, sv_useGameLiftServer, false, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Activate GameLift server manager and SDK");
+
     AWSGameLiftServerSystemComponent::AWSGameLiftServerSystemComponent()
         : m_gameLiftServerManager(AZStd::make_unique<AWSGameLiftServerManager>())
     {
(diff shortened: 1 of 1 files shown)


Gems/AWSGameLift/Code/AWSGameLiftServer/Source/AWSGameLiftServerSystemComponent.cpp
```diff
@@ -74,14 +76,21 @@ namespace AWSGameLift
...
 
     void AWSGameLiftServerSystemComponent::Deactivate()
     {
-        m_gameLiftServerManager->DeactivateManager();
-        m_gameLiftServerManager->HandleDestroySession();
+        if (sv_useGameLiftServer)
+        {
+            m_gameLiftServerManager->DeactivateManager();
+            m_gameLiftServerManager->HandleDestroySession();
+        }
```
mbalfour-amzn, line 93: Because there's nothing preventing a user from trying to change this cvar after the system component has activated, the deactivate should probably be based on whether or not the m_gameLiftServerManager is actually active, not on the value of the cvar.

AMZN-Olex: How about just checking if m_gameLiftServerManager is valid and is active?

allisaurus: `m_gameLiftServerManager` gets instantiated as a member var, so AFAICT it'll always exist ...unless that should change, and we should only instantiate it if `sv_useGameLiftServer`? Then after activation, we could check whether it exists to determine shut down behavior instead of adding a dedicate "active" status variable

AMZN-Olex (approved):

mbalfour-amzn (approved):

PR #14965, opened 2023-03-06
Makes the visuals of physics joint scale as you zoom

lawsonamzn:
## What does this PR do?

Previously, if you zoom in far enough the physics joint visuals would cover the entire screen unlike the rest of the gizmos and visualizations.

This change causes the visualizers to remain the same size on screen regardless of your zoom.

This also adds a different visualizer for the revolving joint (the hinge joint with no limits).  I made sure that adding rotation in the positive direction caused the body to rotate in the direction the arrows indicate.

## How was this PR tested?

Manually, by viewing the gizmos at various scales.

Hinge Joint with no limit

fixed_revolute

Hinge joint with limit
limited

The rest are unchanged, except that they remain a constant size on screen regardless of whether you zoom in or out.

### DIFF
--- a/Gems/PhysX/Code/Source/EditorBallJointComponent.cpp
+++ b/Gems/PhysX/Code/Source/EditorBallJointComponent.cpp
@@ -11,6 +11,7 @@
 #include <AzCore/Serialization/EditContext.h>
 #include <AzCore/Serialization/SerializeContext.h>
 #include <AzToolsFramework/ViewportSelection/EditorSelectionUtil.h>
+#include <AzToolsFramework/Viewport/ViewportSettings.h>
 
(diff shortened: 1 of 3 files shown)


lawsonamzn:
now that I look at it closely, I gonna need to flip at least 2 of those arrows.  Standby

lawsonamzn:
That's better!
And I made sure that if I give it some positive initial rotation, it does spin in that direction, so it is the correct "positive" direction.
fixed_revolute

AMZN-daimini (approved):

hultonha (approved):

lawsonamzn:
failed due to a different issue already present in dev.  I'll rebase, and try again.

PR #14967, opened 2023-03-06
Fixed assert in the ActionManagerSystemComponent::Init

lumberyard-employee-dm:
When the ActionManagerSystemComponent is loaded in an Application that hasn't create a QGUIApplication, it asserts when a QWidget is created with the message
> QWidget: Must construct a QApplication before a QWidget

The fix is to avoid creating the QWidget if there is no QApplication.

Side Note: We should investigate why the ActionManagerSystemComponent is creating a QWidget. It is the only SystemComponent that creates a QT GUI Object.

## How was this PR tested?

Successfully loaded activated all the active AutomatedTesting SystemComponents when using loading the Editor gem modules in `SerializeContextTools` application.

The command used were as follows
```
<build-dir>/bin/debug/SerializeContextTools.exe convert --project-path=<engine-root>/AutomatedTesting --specializations=editor --files=<engine-root>/AutomatedTesting/Levels/Physics/Material_PerFaceMaterialGetsCorrectMaterial/test.fbx.assetinfo;<engine-root>/AutomatedTesting/Assets/Physics/Collider_MultipleSurfaceSlots/test.fbx.assetinfo  -ext=json
```

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/ActionManager/ActionManagerSystemComponent.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/ActionManager/ActionManagerSystemComponent.cpp
@@ -12,6 +12,7 @@
 
 #include <AzToolsFramework/ActionManager/ActionManagerRegistrationNotificationBus.h>
 #include <AzToolsFramework/Editor/ActionManagerUtils.h>
+#include <QApplication>
 
 namespace AzToolsFramework
 {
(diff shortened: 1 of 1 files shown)


gadams3 (approved):

cgalvan (approved):
The `QWidget` is used as a default parent for menu action widgets and toolbars.

lumberyard-employee-dm:
> The `QWidget` is used as a default parent for menu action widgets and toolbars.

I wonder if there is a better way however to avoid creating the Widget in the Action Manager System Component.
Perhaps we can wrap  the `IsNewActionManagerEnabled()` check in another function that checks if the ActionManager widget is enabled and can be created.

AMZN-daimini (approved):

PR #14969, opened 2023-03-06
DrawPacket clone

jiaweig-amzn:
## What does this PR do?

DrawPacketBuilder is able to clone a draw packet and set instance count and root constant.
Based on changes from:
https://github.com/o3de/o3de/pull/14152/files#diff-650885225375dc0ef0f7f6b2cc4d92395f6e1d752ab1264bfe1889257a62b44d

Rearranged unit tests so that they can access DrawPacket private members and test them. Added unit tests to validate the operations.

## How was this PR tested?

Ran the new unit tests.


Gems/Atom/RHI/Code/Include/Atom/RHI/DrawPacket.h
```diff
@@ -11,6 +11,12 @@
 #include <Atom/RHI/DrawList.h>
 #include <AzCore/std/smart_ptr/intrusive_base.h>
 
+// Predefination for unit test friend class
```
moudgils, line 14: Typo -> Predefinition

Gems/Atom/RHI/Code/Include/Atom/RHI/DrawPacket.h
```diff
@@ -61,13 +69,22 @@ namespace AZ
             //! Overloaded operator delete for freeing a draw packet.
             void operator delete(void* p, size_t size);
 
+            //! Update the root constant at the specefied interval. The same root constants are shared by all draw items in the draw packet
```
moudgils, line 72: specified

Gems/Atom/RHI/Code/Source/RHI/DrawPacket.cpp
```diff
@@ -47,5 +50,29 @@ namespace AZ
...
+                AZ_Assert(sizeValid, "New root constants exceed the original size.");
+            }
+        }
+
+        void DrawPacket::SetInstanceCount(uint32_t instanceCount)
+        {
+            for (size_t drawItemIndex = 0; drawItemIndex < m_drawItemCount; ++drawItemIndex)
+            {
+                const DrawItem* drawItemConst = m_drawItems + drawItemIndex;
+                DrawItem* drawItem = const_cast<DrawItem*>(drawItemConst);
```
moudgils, line 72: Add a comment as to why we need to drop constness. I.e the need to mutate instance count for mesh instancing.

moudgils (approved):

Gems/Atom/RHI/Code/Tests/DrawPacketTests.cpp
```diff
@@ -177,246 +185,503 @@ namespace UnitTest
...
+                EXPECT_EQ(rootConstantOld[i], drawPacket->m_rootConstants[i]);
             }
 
-            tagIndex++;
+            // Compare the part being set
+            for (uint8_t i = drawPacket->m_rootConstantSize - 8; i < drawPacket->m_rootConstantSize; ++i)
+            {
+                EXPECT_EQ(rootConstantNew[i - (drawPacket->m_rootConstantSize -8)], drawPacket->m_rootConstants[i]);
+            }
+
```
amzn-tommy, line 617: For both this test and the SetInstanceCount test, can you add something at the end that clones the draw packet, sets the root constant/instance count on the cloned draw packet, and then verifies that the original did not change? Since that's the main use-case for these functions. It will validate that our cloned draw packets aren't referring to some shared data in memory.

Gems/Atom/RHI/Code/Include/Atom/RHI/DrawPacketBuilder.h
```diff
@@ -74,6 +74,8 @@ namespace AZ
 
             const DrawPacket* End();
 
+            const DrawPacket* Clone(const DrawPacket* original);
```
amzn-tommy, line 77: Let's add a comment here to be more explicit about the DrawSrg, since I think that might surprise someone at some point.
//! Make a copy of an existing DrawPacket.
//! Note: the copy will reference the same DrawSrg as the original, so it is not possible to vary the DrawSrg values between the original draw packet and the cloned one.
//! Only settings that can be modified via the DrawPacket interface can be changed after cloning, such as SetRootConstant and SetInstanceCount

amzn-tommy (approved):

Issue #11446, opened 2022-08-22
[RHI] The engine uses quite a lot of VRAM


CyaNinix:
Hello there, sorry if I step a bit outside of the bug format since this isn't really a bug, it's more about raising awareness of an issue with the way O3DE handle video memory.

# The issue

O3DE D3D12 and Vulkan buffers allocators use a tremendous amount of video memory (VRAM), which prevents us to launch our game of most of our computers.

This is because of multiple issues, first O3DE will allocate a lot of buffers in their own memory block (unique allocations), I've seen a lot of calls to vkAllocateMemory and ID3D12Device::CreateCommittedResource with very little size (~200 bytes), which is problematic with D3D12 because every unique allocation is aligned to `D3D12_DEFAULT_RESOURCE_PLACEMENT_ALIGNMENT` (65335):
```cpp
        BufferMemoryView BufferMemoryAllocator::AllocateUnique(const RHI::BufferDescriptor& bufferDescriptor)
        {
            AZ_PROFILE_FUNCTION(RHI);

            const size_t alignedSize = RHI::AlignUp(bufferDescriptor.m_byteCount, Alignment::CommittedBuffer);
```

When you have several differents meshes and actors, you can easily gets out of memory errors.

On Vulkan you have the very same issue because "Allocations returned by vkAllocateMemory are guaranteed to meet any alignment requirement of the implementation.", but it's even more problematic because the number of unique allocations supported by the API is quite low:
> The maximum number of valid memory allocations that can exist simultaneously within a [VkDevice](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkDevice.html) may be restricted by implementation- or platform-dependent limits. The [maxMemoryAllocationCount](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#limits-maxMemoryAllocationCount) feature describes the number of allocations that can exist simultaneously before encountering these internal limits.
(https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkAllocateMemory.html)

[...]

PR #14971, opened 2023-03-07
Change Vulkan impl of BufferPools to use AMD's VulkanMemoryAllocator library

rgba16f:
Implementation was contributed by Cyanide.


Gems/Atom/RHI/Vulkan/Code/Source/RHI/Buffer.h
```diff
@@ -68,7 +68,7 @@ namespace AZ
 
         private:
             Buffer() = default;
-            RHI::ResultCode Init(Device& device, const RHI::BufferDescriptor& bufferDescriptor, BufferMemoryView& memmoryView);
+            RHI::ResultCode Init(Device& device, const RHI::BufferDescriptor& bufferDescriptor, const BufferMemoryView& memmoryView);
```
lumberyard-employee-dm, line 71: repeat: The memmoryView variable has a typo in it.
```suggestion
            RHI::ResultCode Init(Device& device, const RHI::BufferDescriptor& bufferDescriptor, const BufferMemoryView& memoryView);
```

Gems/Atom/RHI/Vulkan/Code/Source/RHI/MemoryPageAllocator.cpp
```diff
@@ -49,9 +49,11 @@ namespace AZ
 
         void MemoryPageFactory::ShutdownObject(Memory& memory, bool isPoolShutdown)
         {
+/*          // Allocation statistics are broken with VMA change
             RHI::HeapMemoryUsage& heapMemoryUsage = *m_descriptor.m_getHeapMemoryUsageFunction();
             heapMemoryUsage.m_totalResidentInBytes -= memory.GetDescriptor().m_sizeInBytes;
             heapMemoryUsage.Validate();
+*/
```
lumberyard-employee-dm, line 56: This code should be removed or a GHI should be made to address fixing htis.

Gems/Atom/RHI/Vulkan/Code/Source/RHI/Device.h
```diff
@@ -214,6 +224,11 @@ namespace AZ
 
             BindlessDescriptorPool m_bindlessDescriptorPool;
             ShadingRateImageMode m_imageShadingRateMode = ShadingRateImageMode::None;
+
+            RHI::ResultCode InitVulkanAllocator(RHI::PhysicalDevice& physicalDevice);
+            void ShutdownVulkanAllocator();
+
+            VmaAllocator m_vmaMemoryAllocator;
```
lumberyard-employee-dm, line 231: `VmaAllocator` can be made a pointer and forward declare here.
That allow `vk_mem_alloc.h` to only be included in the cpp file

Gems/Atom/RHI/Vulkan/Code/Source/RHI/Device.cpp
```diff
@@ -31,6 +31,13 @@
 #include <Vulkan_Traits_Platform.h>
 #include <Atom/RHI.Reflect/VkAllocator.h>
 
+#define VMA_IMPLEMENTATION
+#define VMA_STATIC_VULKAN_FUNCTIONS 0
+
+#pragma warning(push, 3)
+#include <vma/vk_mem_alloc.h>
+#pragma warning(pop)
```
lumberyard-employee-dm, line 39: This should use the [AZ_PUSH_DISABLE_WARNING/AZ_POP_DISABLE_WARNING](https://github.com/o3de/o3de/blob/a10351f38ddf50af2b01ed071c04cda52d6069ed/Code/Framework/AzCore/AzCore/PlatformDef.h#L73-L80) macros.

This will not work on non-Windows.

lumberyard-employee-dm (changes requested):

Gems/Atom/RHI/Vulkan/Code/CMakeLists.txt
```diff
@@ -40,6 +40,7 @@ if(NOT PAL_TRAIT_ATOM_RHI_VULKAN_SUPPORTED)
                 AZ::AzCore
                 Gem::Atom_RHI.Reflect
                 3rdParty::glad_vulkan
+                3rdParty::amd_vma
```
moudgils, line 43: Why do we need vma for stub gem?

Gems/Atom/RHI/Vulkan/Code/CMakeLists.txt
```diff
@@ -68,7 +69,7 @@ if(NOT PAL_TRAIT_ATOM_RHI_VULKAN_SUPPORTED)
                     Gem::Atom_RHI.Edit
                     Gem::Atom_RHI.Reflect
                     3rdParty::glad_vulkan
-
+                    3rdParty::amd_vma
```
moudgils, line 72: Dont think we will need to add vma for Mac.

Gems/Atom/RHI/Vulkan/Code/CMakeLists.txt
```diff
@@ -160,7 +161,8 @@ ly_add_target(
             AZ::AzCore
         PUBLIC
             3rdParty::glad_vulkan
-)
+            3rdParty::amd_vma
```
moudgils, line 164: Why is this added here instead of Atom_RHI_Vulkan.Private.Static

Gems/Atom/RHI/Vulkan/Code/Source/RHI/BufferPool.h
```diff
@@ -11,11 +11,20 @@
 #include <Atom/RHI.Reflect/Vulkan/BufferPoolDescriptor.h>
 #include <RHI/BufferMemoryAllocator.h>
 
+
+//#define DEBUG_TRACK_VMA_ALLOCATIONS
+#ifdef DEBUG_TRACK_VMA_ALLOCATIONS
+#include <unordered_map>
```
moudgils, line 17: use az version

Gems/Atom/RHI/Vulkan/Code/Source/RHI/MemoryPageAllocator.cpp
rgba16f: I haven't removed this yet. Basic statistics work in straightforward cases, but seem to be broken in others (e.g. going into game mode in the editor). I'm guessing it was removed because the tracking is being done elsewhere.

Gems/Atom/RHI/Vulkan/Code/Source/RHI/Device.h
rgba16f: VmaAllocator is already a typedef of a ptr. I just recreated the typedef at the top of the file and removed the include of vk_mem_alloc.h.

Gems/Atom/RHI/Vulkan/Code/Source/RHI/BufferPool.h
```diff
@@ -63,6 +72,9 @@ namespace AZ
             //////////////////////////////////////////////////////////////////////////
 
             BufferMemoryAllocator m_memoryAllocator;
+#ifdef DEBUG_TRACK_VMA_ALLOCATIONS
+            std::unordered_map<BufferMemory*, VkDeviceSize> m_allocations;
+#endif
```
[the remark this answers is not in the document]
rgba16f: Updated to use AZStd.

Gems/Atom/RHI/Vulkan/Code/Source/RHI/Device.cpp
rgba16f: removed the pragma's

rgba16f:
> 

BufferPool covers most of our buffer allocations. And is done first because on Vulkan we were forcing InputAssembly buffers to be unique allocations. Now all BufferPool allocations go through VMA. This can save GB's of mem, and help prevent us hitting the Vulkan allocation cap.

Gems/Atom/RHI/Vulkan/Code/Source/RHI/BufferMemory.cpp
```diff
@@ -18,6 +18,43 @@ namespace AZ
...
+            createInfo.queueFamilyIndexCount = static_cast<uint32_t>(queueFamilies.size());
+            createInfo.pQueueFamilyIndices = queueFamilies.empty() ? nullptr : queueFamilies.data();
+
+            VmaAllocationCreateInfo allocInfo = {};
+            allocInfo.flags = VMA_ALLOCATION_CREATE_HOST_ACCESS_RANDOM_BIT;
+            allocInfo.usage = VMA_MEMORY_USAGE_AUTO;
+
+            VmaAllocator allocator = device.GetMemoryAllocator();
+            if (vmaCreateBuffer(allocator, &createInfo, &allocInfo, &m_vkBuffer, &m_vmaAllocation, &m_vmaAllocationInfo) != VK_SUCCESS)
+                return RHI::ResultCode::OutOfMemory;
```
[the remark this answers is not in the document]
rgba16f: updated.

Gems/Atom/RHI/Vulkan/Code/Source/RHI/BufferMemory.cpp
```diff
@@ -39,14 +76,45 @@ namespace AZ
...
+
         CpuVirtualAddress BufferMemory::Map(size_t offset, size_t size, RHI::HostMemoryAccess hostAccess)
         {
-            return m_memoryView.GetMemory()->Map(m_memoryView.GetOffset() + offset, size, hostAccess);
+            Device& device = static_cast<Device&>(GetDevice());
+
+            void* mappedPtr;
+            VkResult vkResult = vmaMapMemory(device.GetMemoryAllocator(), m_vmaAllocation, &mappedPtr);
+            if (vkResult != VK_SUCCESS)
+                return nullptr;
```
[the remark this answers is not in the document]
rgba16f: updated

Gems/Atom/RHI/Vulkan/Code/Source/RHI/BufferMemory.cpp
```diff
@@ -39,14 +76,45 @@ namespace AZ
...
+
+            if (hostAccess == RHI::HostMemoryAccess::Read)
+            {
+                // VMA will check if this is necessary
+                vkResult = vmaInvalidateAllocation(device.GetMemoryAllocator(), m_vmaAllocation, offset, size);
+                if (vkResult != VK_SUCCESS)
+                    return nullptr;
```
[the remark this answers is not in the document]
rgba16f: updated

Gems/Atom/RHI/Vulkan/Code/Source/RHI/BufferPool.h
rgba16f: updated

Gems/Atom/RHI/Vulkan/Code/CMakeLists.txt
rgba16f: Fixed too many modules having amd_vma added.

Gems/Atom/RHI/Vulkan/Code/CMakeLists.txt
rgba16f: Removed vma from the stubbed gems.

rgba16f:
> > > > > It feels like there's still some work missing because the BufferMemory class is using VMA and also direct memory allocations is different cases. Is there any plans to replace all buffer allocations with VMA? Depending on the "Init" function used, the BufferMemory class may or may not have a valid MemoryView object. Also the need to expose if the BufferMemory was allocated using VMA is a sign that something is weird is going on. That information should not be needed outside the class.
> > > > 
> > > > 
> > > > BufferPool covers most of our buffer allocations. And is done first because on Vulkan we were forcing InputAssembly buffers to be unique allocations. Now all BufferPool allocations go through VMA. This can save GB's of mem, and help prevent us hitting the Vulkan allocation cap.
> > > 
> > > 
> > > I get that this handles the problem with multiple "unique" allocations (e.g. InputAssembly), but now the BufferMemory class is this weird hybrid between VMA and direct memory. I would prefer to do a proper replacement for all memory allocations to use VMA. Also, forcing all allocations to be unique in the BufferPool is removing our VkBuffer sub-allocation strategy.
> > 
> > 
> > Removing our sub-allocation strategy is intentional! The intention is to move all the allocation strategy over to VMA, This is the first part of that work. I will also be implementing the same change for DX12. Metal doesn't have an equivalent library so more work will be needed there.
> > There is an RFC for this work here - [o3de/sig-graphics-audio#121](https://github.com/o3de/sig-graphics-audio/issues/121) And the original discussion by Cyanide here - #11446
> 
> I'm talking about the buffer sub-allocation, not the memory sub-allocation. As far as I know VMA doesn't do buffer sub-allocations. You need to do it yourself ([GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator#88](https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/issues/88)) which is what we are doing.
> 
> I do think we should replace all memory allocations with VMA, but I think I more global solution should be implemented. IMHO this PR it's a patch to workaround the problem that Cyanide had. I think we should consider a more general integration of VMA into the RHI Vulkan.

Yes, we're removing the buffer sub-allocation strategy. We're going over to Heaps that are placement allocated, which is how VMA works.

rgba16f:
> > > > > > > It feels like there's still some work missing because the BufferMemory class is using VMA and also direct memory allocations is different cases. Is there any plans to replace all buffer allocations with VMA? Depending on the "Init" function used, the BufferMemory class may or may not have a valid MemoryView object. Also the need to expose if the BufferMemory was allocated using VMA is a sign that something is weird is going on. That information should not be needed outside the class.
> > > > > > 
> > > > > > 
> > > > > > BufferPool covers most of our buffer allocations. And is done first because on Vulkan we were forcing InputAssembly buffers to be unique allocations. Now all BufferPool allocations go through VMA. This can save GB's of mem, and help prevent us hitting the Vulkan allocation cap.
> > > > > 
> > > > > 
> > > > > I get that this handles the problem with multiple "unique" allocations (e.g. InputAssembly), but now the BufferMemory class is this weird hybrid between VMA and direct memory. I would prefer to do a proper replacement for all memory allocations to use VMA. Also, forcing all allocations to be unique in the BufferPool is removing our VkBuffer sub-allocation strategy.
> > > > 
> > > > 
> > > > Removing our sub-allocation strategy is intentional! The intention is to move all the allocation strategy over to VMA, This is the first part of that work. I will also be implementing the same change for DX12. Metal doesn't have an equivalent library so more work will be needed there.
> > > > There is an RFC for this work here - [o3de/sig-graphics-audio#121](https://github.com/o3de/sig-graphics-audio/issues/121) And the original discussion by Cyanide here - #11446
> > > 
> > > 
> > > I'm talking about the buffer sub-allocation, not the memory sub-allocation. As far as I know VMA doesn't do buffer sub-allocations. You need to do it yourself ([GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator#88](https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/issues/88)) which is what we are doing.
> > > I do think we should replace all memory allocations with VMA, but I think I more global solution should be implemented. IMHO this PR it's a patch to workaround the problem that Cyanide had. I think we should consider a more general integration of VMA into the RHI Vulkan.
> > 
> > 
> > Yes, we're removing the buffer sub-allocation strategy. We're going over to Heaps that are placement allocated, which is how VMA works.
> 
> Could you point me to the "placement allocated" documentation on VMA? I couldn't find it. The only thing I could find is that they recommend to do buffer sub-allocation in the vmaCreateBuffer function.
> 
> `This function creates a new VkBuffer. Sub-allocation of parts of one large buffer, although recommended as a good practice, is out of scope of this library and could be implemented by the user as a higher-level logic on top of VMA.`

I see it here - https://gpuopen-librariesandsdks.github.io/VulkanMemoryAllocator/html/choosing_memory_type.html

> Dedicated allocations
> Memory for allocations is reserved out of larger block of VkDeviceMemory allocated from Vulkan internally. That is the main feature of this whole library.

Gems/Atom/RHI/Vulkan/Code/Source/RHI/BufferPool.cpp
```diff
@@ -145,7 +172,7 @@ namespace AZ
             buffer.m_memoryView = BufferMemoryView();
             buffer.Invalidate();
 
-            // Allocate a new BufferMemmory
+            // Allocate a new BufferMemory
             BufferMemoryView memoryView = m_memoryAllocator.Allocate(buffer.GetDescriptor().m_byteCount, 1);
```
moudgils, line 176: Doesnt this method need to be converted over as well?

Gems/Atom/RHI/Vulkan/Code/Source/RHI/BufferMemory.cpp
```diff
@@ -39,14 +74,54 @@ namespace AZ
...
+        const VmaAllocationInfo& BufferMemory::GetVmaAllocationInfo() const
+        {
+            AZ_Assert(IsVmaAllocated(), "Buffer memory was not allocated with VMA");
+            return m_vmaAllocationInfo;
+        }
+
         CpuVirtualAddress BufferMemory::Map(size_t offset, size_t size, RHI::HostMemoryAccess hostAccess)
         {
-            return m_memoryView.GetMemory()->Map(m_memoryView.GetOffset() + offset, size, hostAccess);
+            Device& device = static_cast<Device&>(GetDevice());
```
moudgils, line 90: What if the memory came from m_memoryView (i.e no vma)? Can we take out m_memoryView completely?

Gems/Atom/RHI/Vulkan/Code/Source/RHI/Device.cpp
```diff
@@ -1199,6 +1216,94 @@ namespace AZ
...
+                context.BindBufferMemory,
+                context.BindImageMemory,
+                context.GetBufferMemoryRequirements,
+                context.GetImageMemoryRequirements,
+                context.CreateBuffer,
+                context.DestroyBuffer,
+                context.CreateImage,
+                context.DestroyImage,
+                context.CmdCopyBuffer,
+#if VMA_DEDICATED_ALLOCATION || VMA_VULKAN_VERSION >= 1001000
```
moudgils, line 1247: Where will these macros be defined?

moudgils:
I do think that it will be a good idea to atleast convert over Vk::BufferPool **fully** to use VMA. At the moment it is using both vma and direct which is confusing. The only other remaining use case is aliased heap related buffers and that can be done as a separate PR.

Gems/Atom/RHI/Vulkan/Code/Source/RHI/Device.cpp
rgba16f: They are defined by the VMA header.
```c++
#if !defined(VMA_DEDICATED_ALLOCATION)
    #if VK_KHR_get_memory_requirements2 && VK_KHR_dedicated_allocation
        #define VMA_DEDICATED_ALLOCATION 1
    #else
        #define VMA_DEDICATED_ALLOCATION 0
    #endif
#endif

#if !defined(VMA_BIND_MEMORY2)
    #if VK_KHR_bind_memory2
        #define VMA_BIND_MEMORY2 1
    #else
        #define VMA_BIND_MEMORY2 0
    #endif
#endif

```

rgba16f:
@moudgils I think this PR can be closed now that @akioCL has his change up for review.

PR #14985, opened 2023-03-07
NetBind component now removes its entity from network tracking on Deactivate

mbalfour-amzn:
## What does this PR do?

A previous recent change to the NetBind component made it so that networked prefabs could be despawned just by calling DespawnAllEntities. However, by going through that API instead of the NetworkEntityManager API, this caused a tracking leak where the networked entities still had entries in a network entity table.

This PR fixes the leak by having the NetBind component remove its entity from the tracking table.

## How was this PR tested?

Ran MPS and printed out the tracked network entity counts to verify they no longer kept growing on every round. No other obvious errors or warnings from the change showed up as a result either.


kberg-amzn (approved):
Good catch!

cgalvan (approved):

PR #14997, opened 2023-03-08
Show the 'Is Simulated' option as disabled when the collider is acting as a trigger.

moraaar:
## What does this PR do?

Show the 'Is Simulated' option as disabled when the collider is acting as a trigger,  this way it'll be clear that when the collider is a trigger it won't be simulated.

image

## How was this PR tested?

Checked 'Is Simulated' appears disabled when 'Is Trigger' is enabled in Primitive, Mesh and Shape collider components.

### DIFF
--- a/Code/Framework/AzFramework/AzFramework/Physics/Shape.h
+++ b/Code/Framework/AzFramework/AzFramework/Physics/Shape.h
@@ -72,6 +72,13 @@ namespace Physics
     private:
         void OnRestOffsetChanged();
         void OnContactOffsetChanged();
+
+        // m_stubIsSimulated is used for EditContext only, it will always be false and read-only.
+        // It will be shown instead of the real m_isSimulated property when m_isTrigger is enabled,
+        // this way it'll be clear that when the collider is a trigger it won't be simulated.
+        bool m_stubIsSimulated = false;
+        AZ::Crc32 GetSimulatedPropertyVisibility() const;
+        AZ::Crc32 GetStubSimulatedPropertyVisibility() const;
     };
 
     struct RayCastRequest;
--- a/Code/Framework/AzFramework/AzFramework/Physics/Shape.h
+++ b/Code/Framework/AzFramework/AzFramework/Physics/Shape.h
@@ -73,12 +73,12 @@ namespace Physics
         void OnRestOffsetChanged();
         void OnContactOffsetChanged();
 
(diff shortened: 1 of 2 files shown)


Code/Framework/AzFramework/AzFramework/Physics/Shape.h
```diff
@@ -72,6 +72,13 @@ namespace Physics
     private:
         void OnRestOffsetChanged();
         void OnContactOffsetChanged();
+
+        // m_stubIsSimulated is used for EditContext only, it will always be false and read-only.
+        // It will be shown instead of the real m_isSimulated property when m_isTrigger is enabled,
+        // this way it'll be clear that when the collider is a trigger it won't be simulated.
+        bool m_stubIsSimulated = false;
```
hultonha, line 79: suggestion: It might be clearer to use 'dummy' instead of 'stub' in this context to indicate what this is used for

hultonha (approved):
Looks good, just a small suggestion on the naming but otherwise LGTM :+1:

Code/Framework/AzFramework/AzFramework/Physics/Shape.h
moraaar: will change to dummy.

hultonha:
Any chance you could include a screenshot in the description please @moraaar just for @o3de/sig-ui-ux  to take a look at too.

cgalvan (approved):

Code/Framework/AzFramework/AzFramework/Physics/Shape.h
```diff
@@ -72,6 +72,13 @@ namespace Physics
     private:
         void OnRestOffsetChanged();
         void OnContactOffsetChanged();
+
+        // m_dummyIsSimulated is used for EditContext only, it will always be false and read-only.
+        // It will be shown instead of the real m_isSimulated property when m_isTrigger is enabled,
+        // this way it'll be clear that when the collider is a trigger it won't be simulated.
+        bool m_dummyIsSimulated = false;
```
mbalfour-amzn, line 79: Can this be made const, or does that fail to work with the EditContext?

mbalfour-amzn (approved):

Code/Framework/AzFramework/AzFramework/Physics/Shape.h
moraaar: I tried that first, it does fail to compile.

Issue #14907, opened 2023-03-02
PhysX Mesh Component: Asset Scale (with non-uniform value) does not work when the asset contains primitives


moraaar:
**Describe the bug**
The PhysX Mesh Component has the property "Asset Scale" which allows the PhysX asset to be scaled independently from the transform scale.

**The PhysX asset can be exported as 3 shape types from the FBX Settings' PhysX tab: primitive, convex or triangle mesh.**

The Asset Scale property works correctly with convex and triangle mesh asset types, they scale as expected with or without the entity having a non-uniform scale component.

**When using primitive asset type, the Asset Scale only works correctly if the entity has the non-uniform scale component.**

**Steps to reproduce**
1. Open Editor using AutomatedTesting project
2. Create an entity in a level
3. Add a PhysX Static Rigid Body component.
4. Add a PhysX Mesh Collider component.
5. In the PhysX Mesh property assign the asset `_sphere_1x1` (this asset contains a .pxmesh exported as primitive)
6. Set the Asset Scale value to (1,2,1)

**Expected behavior**
The sphere scales in the Y axis.

**Actual behavior**
The sphere scales uniformly in all axis.

Workaround: Click on the "Add non-uniform scale" button at the Transform component, notice how the sphere is correctly scaled when there is a Non-Uniform Scale component present.

**Unit Test**
There is a unit tests for this in file `o3de\Gems\PhysX\Code\Tests\EditorMeshColliderComponentTests.cpp` which is currently disabled:

````
    // [GHI-14907]
    // Asset Scale (with non-uniform value) does not work when the asset contains primitives and there
    // is no non-uniform scale component present.
    TEST_F(PhysXEditorFixture, DISABLED_EditorMeshColliderComponent_AssetWithPrimitive_AssetScale_CorrectRuntimeEntity)
````

Once this issue is fixed that test should be enabled and pass.

PR #15001, opened 2023-03-08
Fixed asset scale when Mesh Collider has primitive shapes and improved unit tests.

moraaar:
## What does this PR do?

Fixed asset scale when Mesh Collider has primitive shapes. (https://github.com/o3de/o3de/issues/14907)
Subdivision level property now shows when there are primitive shapes.

The tests were checking only the runtime components created from the editor component (which is good to test), but it was missing to check the editor component itself. Refactored the tests to check results in both editor and runtime components.

Fixed bug when creating an `EditorMeshColliderComponent` via its constructor passing physics asset configuration, it was missing to set the proxy's `m_pxAsset` variable from the configuration, which caused the editor on activation to assume there was no asset.

## How was this PR tested?

Run PhysX gem unit tests and passed.
Checked the Asset Scale property inside Mesh Collider Component works properly with all types of shapes (primitive, convex and triangle mesh), also tested with and without a non-uniform scale component.

### DIFF
--- a/Gems/PhysX/Code/Source/BaseColliderComponent.cpp
+++ b/Gems/PhysX/Code/Source/BaseColliderComponent.cpp
@@ -328,7 +328,10 @@ namespace PhysX
             return false;
         }
 
-        const bool hasNonUniformScale = (AZ::NonUniformScaleRequestBus::FindFirstHandler(GetEntityId()) != nullptr);
+        const bool isAssetScaleUniform =
+            AZ::IsClose(physicsAssetConfiguration.m_assetScale.GetX(), physicsAssetConfiguration.m_assetScale.GetY()) &&
+            AZ::IsClose(physicsAssetConfiguration.m_assetScale.GetX(), physicsAssetConfiguration.m_assetScale.GetZ());
+        const bool hasNonUniformScale = !isAssetScaleUniform || (AZ::NonUniformScaleRequestBus::FindFirstHandler(GetEntityId()) != nullptr);
         Utils::CreateShapesFromAsset(physicsAssetConfiguration, componentColliderConfiguration, hasNonUniformScale,
             physicsAssetConfiguration.m_subdivisionLevel, m_shapes);
 
--- a/Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
+++ b/Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
@@ -93,6 +93,7 @@ namespace PhysX
     EditorProxyAssetShapeConfig::EditorProxyAssetShapeConfig(
         const Physics::PhysicsAssetShapeConfiguration& assetShapeConfiguration)
     {
+        m_physicsAsset.m_pxAsset = assetShapeConfiguration.m_asset;
         m_physicsAsset.m_configuration = assetShapeConfiguration;
     }
 
--- a/Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
+++ b/Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
@@ -130,30 +130,25 @@ namespace PhysX
                     switch (cookedMeshShapeConfiguration->GetMeshType())
                     {
                     case Physics::CookedMeshShapeConfiguration::MeshType::Convex:
-                        {
-                            shapeTypes.push_back(ShapeType::Convex);
-                        }
+                        shapeTypes.push_back(ShapeType::Convex);
+                        break;
                     case Physics::CookedMeshShapeConfiguration::MeshType::TriangleMesh:
-                        {
-                            shapeTypes.push_back(ShapeType::TriangleMesh);
-                        }
+                        shapeTypes.push_back(ShapeType::TriangleMesh);
+                        break;
                     default:
-                        {
-                            shapeTypes.push_back(ShapeType::Invalid);
-                        }
+                        shapeTypes.push_back(ShapeType::Invalid);
+                        break;
                     }
-                    break;
                 }
+                break;
             case Physics::ShapeType::Sphere:
             case Physics::ShapeType::Box:
             case Physics::ShapeType::Capsule:
-                {
-                    shapeTypes.push_back(ShapeType::Primitive);
-                }
+                shapeTypes.push_back(ShapeType::Primitive);
+                break;
             default:
-                {
-                    shapeTypes.push_back(ShapeType::Invalid);
-                }
+                shapeTypes.push_back(ShapeType::Invalid);
+                break;
             }
         }
 
@@ -174,21 +169,13 @@ namespace PhysX
         switch (shapeTypes[0])
         {
         case ShapeType::Primitive:
-            {
-                return assetName + " (Primitive)";
-            }
+            return assetName + " (Primitive)";
         case ShapeType::Convex:
-            {
-                return assetName + " (Convex)";
-            }
+            return assetName + " (Convex)";
         case ShapeType::TriangleMesh:
-            {
-                return assetName + " (Triangle Mesh)";
-            }
+            return assetName + " (Triangle Mesh)";
         default:
-            {
-                return assetName;
-            }
+            return assetName;
         }
     }
 
--- a/Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
+++ b/Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
@@ -97,77 +97,96 @@ namespace PhysX
         m_physicsAsset.m_configuration = assetShapeConfiguration;
     }
 
-    AZStd::string EditorProxyAssetShapeConfig::PhysXMeshAssetShapeTypeName() const
+    AZStd::vector<EditorProxyAssetShapeConfig::ShapeType> EditorProxyAssetShapeConfig::GetShapeTypesInsideAsset() const
     {
-        const AZStd::string assetName = "Asset";
-
         if (!m_physicsAsset.m_pxAsset.IsReady())
         {
-            return assetName;
+            return {};
         }
 
         Physics::ColliderConfiguration defaultColliderConfiguration;
         Physics::PhysicsAssetShapeConfiguration physicsAssetConfiguration = m_physicsAsset.m_configuration;
         physicsAssetConfiguration.m_asset = m_physicsAsset.m_pxAsset;
+        physicsAssetConfiguration.m_assetScale = AZ::Vector3::CreateOne(); // Remove the scale so it doesn't affect the query for the asset mesh type
+        const bool hasNonUniformScale = false;
 
         AzPhysics::ShapeColliderPairList shapeConfigList;
         Utils::GetColliderShapeConfigsFromAsset(
-            physicsAssetConfiguration,
-            defaultColliderConfiguration,
-            m_hasNonUniformScale,
-            m_subdivisionLevel,
-            shapeConfigList);
+            physicsAssetConfiguration, defaultColliderConfiguration, hasNonUniformScale, m_subdivisionLevel, shapeConfigList);
 
-        if (shapeConfigList.empty())
+        AZStd::vector<ShapeType> shapeTypes;
+        shapeTypes.reserve(shapeConfigList.size());
+        for (const auto& shapeConfig : shapeConfigList)
+        {
+            const Physics::ShapeConfiguration* shapeConfiguration = shapeConfig.second.get();
+            AZ_Assert(shapeConfiguration, "GetShapeTypesInsideAsset: Invalid shape-collider configuration pair");
+
+            switch (shapeConfiguration->GetShapeType())
+            {
+            case Physics::ShapeType::CookedMesh:
+                {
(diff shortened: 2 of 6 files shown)


Gems/PhysX/Code/Tests/EditorMeshColliderComponentTests.cpp
```diff
@@ -147,37 +194,24 @@ namespace PhysXEditorTests
...
-
-        AzPhysics::ShapeColliderPairList shapeConfigList = meshColliderComponent->GetShapeConfigurations();
-        EXPECT_EQ(shapeConfigList.size(), 1);
-
-        for (const auto& shapeConfigPair : shapeConfigList)
+        auto checkSimulatedBody = [](const AZ::EntityId& entityId)
         {
-            EXPECT_EQ(shapeConfigPair.second->GetShapeType(), Physics::ShapeType::PhysicsAsset);
-        }
+            AzPhysics::SimulatedBody* simulatedBody = nullptr;
```
greerdv, line 199: There are a whole bunch of blocks of code which look pretty much like this one. I would be tempted to make more helper functions, so the lambda can just look something like

{
    ExpectGeometryType(entityId, physx::PxGeometryType::eSPHERE);
    ExpectAabb(entityId, min, max, tolerance);
}

Or to be even more succinct you could even do

for (auto entityId : {editorEntity->GetId(), gameEntity->GetId()})
{
    ExpectGeometryType(entityId, physx::PxGeometryType::eSPHERE);
    ExpectAabb(entityId, min, max, tolerance);
}

and not need the lambda at all.

greerdv: github mangled the indentation there :(

moraaar: Good feedback, thanks, I was looking into that myself :)

Gems/PhysX/Code/Source/BaseColliderComponent.cpp
```diff
@@ -328,7 +328,10 @@ namespace PhysX
             return false;
         }
 
-        const bool hasNonUniformScale = (AZ::NonUniformScaleRequestBus::FindFirstHandler(GetEntityId()) != nullptr);
+        const bool isAssetScaleUniform =
+            AZ::IsClose(physicsAssetConfiguration.m_assetScale.GetX(), physicsAssetConfiguration.m_assetScale.GetY()) &&
+            AZ::IsClose(physicsAssetConfiguration.m_assetScale.GetX(), physicsAssetConfiguration.m_assetScale.GetZ());
```
hultonha, line 333: thought: It's a shame we don't have a nice util for this thinking about it (I had a look on the `Vector3` API and couldn't see anything). I realize it's likely not worth it for this change but I do wonder how often this 'IsUniform' check comes up (maybe @greerdv wrote a helper for this?) or even if we had something like 'ElementsEqual/Close' for vector types... just thinking out loud really 😅

Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
```diff
@@ -93,88 +93,117 @@ namespace PhysX
...
+                        }
                     }
+                    break;
+                }
+            case Physics::ShapeType::Sphere:
+            case Physics::ShapeType::Box:
+            case Physics::ShapeType::Capsule:
+                {
+                    shapeTypes.push_back(ShapeType::Primitive);
+                }
```
hultonha, line 152: update: It looks like we're missing a `break` here, otherwise I believe `Invalid` will also be added 😬

Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
```diff
@@ -93,88 +93,117 @@ namespace PhysX
...
         }
     }
 
     bool EditorProxyAssetShapeConfig::ShowingSubdivisionLevel() const
     {
-        return m_hasNonUniformScale;
+        const AZStd::vector<ShapeType> shapeTypes = GetShapeTypesInsideAsset();
+
+        return m_hasNonUniformScale &&
+            AZStd::any_of(
```
hultonha, line 200: +1 to the algorithm use here 🙂

Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
```diff
@@ -845,6 +869,11 @@ namespace PhysX
 
     void EditorMeshColliderComponent::UpdateShapeConfigurationScale()
     {
+        const AZ::Vector3& assetScale = m_proxyShapeConfiguration.m_physicsAsset.m_configuration.m_assetScale;
+        const bool isAssetScaleUniform =
+            AZ::IsClose(assetScale.GetX(), assetScale.GetY()) && AZ::IsClose(assetScale.GetX(), assetScale.GetZ());
```
hultonha, line 874: nit: As this is the same 'IsUniform' check mentioned above, maybe it is worth moving to a helper (or using an existing one if one exists - tagging @greerdv who may know this).

Gems/PhysX/Code/Source/EditorRigidBodyComponent.cpp
```diff
@@ -78,11 +78,18 @@ namespace PhysX
                     continue;
                 }
 
+                const AZ::Vector3& assetScale = shapeConfigurationProxy.m_physicsAsset.m_configuration.m_assetScale;
+                const bool isAssetScaleUniform =
+                    AZ::IsClose(assetScale.GetX(), assetScale.GetY()) && AZ::IsClose(assetScale.GetX(), assetScale.GetZ());
```
hultonha, line 83: nit: Another IsUniform check (mentioned above)

Gems/PhysX/Code/Tests/EditorTestUtilities.cpp
```diff
@@ -85,6 +85,15 @@ namespace PhysXEditorTests
...
         }
 
+        // If using a dynamic rigid body then reactivate the entity to recreate the editor world body.
+        // This is because the dynamic rigid body only applies the changes in the next simulation tick,
+        // which happens automatically in the editor but not in test environment.
+        if (rigidBodyType == RigidBodyType::Dynamic)
+        {
+            editorEntity->Deactivate();
+            editorEntity->Activate();
+        }
```
hultonha, line 95: suggestion: Should we move this and the comment to a function to reduce the duplication below? Can this just be a member function?

hultonha (commented):
Overall looks good and the test changes/tidy-up in `EditorMeshColliderComponentTests.cpp` are great but I have a few small suggestions before approving (+ one update on a case statement that could be a bug)

Gems/PhysX/Code/Source/BaseColliderComponent.cpp
greerdv: moraaar and I already discussed this earlier. I'm not aware of a utility for this off the top of my head. I don't think it actually comes up that often to be honest.

Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
moraaar: great catch! It used to be a return, so I missed to put the break when I changed the code. Thanks!

Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
moraaar: Covered in the previous comment.

Gems/PhysX/Code/Source/EditorRigidBodyComponent.cpp
moraaar: Covered in the previous comment.

Gems/PhysX/Code/Tests/EditorTestUtilities.cpp
moraaar: > Yeah, I was just thinking the same thing. There's quite a lot of duplication in general between these functions. It could be nice to have them do something like
>
> SetupGenericStuff();
// shape specific logic
...
DoGenericFinalization();

I'm happy to continue refactoring, but I think the granularity at the moment of these `CreateXXXColliderEditorEntity` functions is quite nice because it's very easy to read the steps of how to setup a collider. It is the same pattern, but what they do is different when creating the entity and configuring the components, so refactoring beyond this point I think it's when it'll start to obfuscate the coded rather than helping. Yes, they'll be less lines of code, but for people looking for code snippets or examples of how to do create colliders it'll just make it more complicated to follow.

If you feel strongly I should give it another pass to refactor all these helpers let me know.

Cheers

moraaar: >I'm also wondering if there's a slightly less clunky was to trigger the editor body recreation. I was wondering if it would be OK to make either CreateEditorWorldRigidBody or OnConfigurationChanged public? I guess you'd still need a comment to explain why that was getting called though.

The `OnColliderChanged` is public and it's called by the collider, but the issue is the rigid body just caches a boolean indicating it'll recreate itself in the next simulation tick. I guess this was done because there are many notifications that could trigger recreate the rigid body and many can happen in 1 frame, so it'd be very slow to recreate the body for each notification.

This could be done instead to make the rigid body recreate the body:

````
        // Update editor world to let updates to be applied
        auto* physicsSystem = AZ::Interface<AzPhysics::SystemInterface>::Get();
        physicsSystem->Simulate(0.1f);
````

But I didn't like this approach because these are helpers to create the entity, so trigger simulation in the middle of the "setup" step of an unit tests seems odd.

It might be better to add a function to a bus to force the recreation of the rigid body and call it. But this is now out of scope of this PR I'd say.

Comment (greerdv) [reply Gems/PhysX/Code/Tests/EditorTestUtilities.cpp:95]: I was thinking about OnConfigurationChanged rather than OnColliderChanged, but it's not important.

These are all just suggestions, the PR is perfectly fine with or without them.

File: Gems/PhysX/Code/Source/BaseColliderComponent.cpp (hunk 2 above)
Comment (FuzzyCarterAWS) [reply Gems/PhysX/Code/Source/BaseColliderComponent.cpp:333]: As a general abstraction, adding some helpers to AZ Math's Transform to check uniform for various data types/patterns might be a good place to consider for a later support.

I had also noticed that @shawstar had done some uniform checks from vectors for an export tool they working on, so there is definitely some potential need for this.

File: Gems/PhysX/Code/Tests/EditorMeshColliderComponentTests.cpp [hunk 8]
```diff
@@ -34,6 +34,56 @@
 
 namespace PhysXEditorTests
 {
+    static bool MeshColliderHasOnePhysicsAssetShapeType(PhysX::MeshColliderComponent* meshColliderComponent)
+    {
+        if (!meshColliderComponent)
+        {
+            return false;
+        }
```
FuzzyCarterAWS, line 42: If the tests depend on these, is it wise to throw a return here instead an assert since the test is going to fail anyways? 

If you're planning on handling these possibilities in tests, using an enumerated return code pattern and failing tests that way would provide greater value.

Gems/PhysX/Code/Tests/EditorTestUtilities.cpp
hultonha: Maybe only have the massive comment once then?

hultonha: It would be trivial to make this a free function you call from all the places you've added it

Gems/PhysX/Code/Tests/EditorMeshColliderComponentTests.cpp
FuzzyCarterAWS: NVM, I misread what was going on here after going back and forth between Python & C++ all day.

Gems/PhysX/Code/Tests/EditorTestUtilities.cpp
moraaar: I'll move it to a function.

Gems/PhysX/Code/Source/BaseColliderComponent.cpp
hultonha: Understood, my thought was a few function somewhere (not a member) but agree it's not always easy to know where it should go (though I'm sure we probably do have a MathsHelpers file somewhere). It's fine for now but this is the sort of thing that is definitely useful to add (and test! 🙂)

Gems/PhysX/Code/Source/EditorMeshColliderComponent.cpp
hultonha: Good to hear! 🙂

Gems/PhysX/Code/Tests/EditorTestUtilities.cpp
hultonha: Thank you! ❤️

hultonha (approved):
LGTM 👍 Thanks for making the updates

Issue #14903, opened 2023-03-02
Material Canvas: Asset processing fails when PBR Node Groups are being continuously collapsed and expanded


LB-KacperMetelski:
**Describe the bug**
Collapsing and expanding Node Groups containing multiple Material Output Nodes may cause the Asset Processor to fail processing the untitled_DepthPass.shader; moreover, there is a slight chance that the processing may fail indefinitely (the Viewport is stuck with the base material) until the user decides to save the Material Graph.

The reproduction rate of this issue may vary greatly due to timing in which the Group is collapsed/expanded.

**Steps to reproduce**
1. Open the Material Canvas.
2. Add both Material Outputs(PBRs).
3. Group them.
4. Collapse and expand the group at least 4 times.

**Expected behavior**
Assets are processed successfully.

**Actual behavior**
Asset processor fails to process the asset. (rarely)
![image](https://user-images.githubusercontent.com/110163473/222438427-9107fba0-c71c-4a3c-aff3-0a709d247d80.png)

**Screenshots/Video**
https://user-images.githubusercontent.com/110163473/222448519-874e43d1-c8e4-4909-9adb-dffd10013895.mp4

**Found in Branch**
Development

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
ba791dc

**Desktop/Device (please complete the following information):**
- Device: PC
- OS: Windows
- Version: 11 21H2
- CPU: I7-11700f
- GPU: Geforce RTX3070 MSI
- Memory: 32GB

galibzon:
@gadams3

LB-KacperMetelski:
Issue verified as fixed on commit 3a3fce2.

PR #15012, opened 2023-03-08
Material Canvas: Use a unique suffix for substitution names if a graph has multiple output nodes

gadams3:
## What does this PR do?

Material canvas currently supports multiple output nodes on the same graph. While it’s not typical, this allows generating multiple materials, strategies, or other types of content from the same graph. To keep the generated data of one element node from stopping over the other, a unique suffix will be added onto graph name substitutions for each output node after the first one.

If other issues are encountered, this can be changed to return after the first output is processed and ignore the others.

Resolves https://github.com/o3de/o3de/issues/14903

## How was this PR tested?

Confirmed that having multiple output nodes on the same graph will generate data for both nodes with the new suffix.

### DIFF
--- a/Gems/Atom/Tools/MaterialCanvas/Code/Source/Document/MaterialGraphCompiler.h
+++ b/Gems/Atom/Tools/MaterialCanvas/Code/Source/Document/MaterialGraphCompiler.h
@@ -126,6 +126,9 @@ namespace MaterialCanvas
             const AZStd::string& templateInputPath,
             const AZStd::string& templateOutputPath) const;
 
+        // Returns the name that will be used to replace material graph name during any substitutions 
+        AZStd::string GetUniqueGraphName() const;
+
         // All slots and nodes will be visited to collect all of the unique include paths.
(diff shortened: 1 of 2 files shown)


cgalvan (approved):

Gems/Atom/Tools/MaterialCanvas/Code/Source/Document/MaterialGraphCompiler.h
```diff
@@ -141,6 +144,9 @@ namespace MaterialCanvas
         // match types and sizes of values on related slots.
         AZStd::map<GraphModel::ConstSlotPtr, AZStd::any> m_slotValueTable;
 
+        // This counter will be used as a suffix for graph name substitutions in case multiple template nodes are included in the same graph
+        int m_templateNodeCount = 0;
```
jiaweig-amzn, line 148: Should we use int32_t?

jiaweig-amzn (approved):

dtamkin1 (approved):

invertednormal (approved):

PR #13959, opened 2023-01-06
Added a new error message for Erroneous Assets 

raktalen:
Signed-off-by: Nikhil Raktale <raktalen@amazon.com>

## What does this PR do?

Fixes : https://github.com/o3de/o3de/issues/4934

- Adds better error messaging for assets that were deleted on disk after being referenced
- Adds better iconography and messaging for assets that are supposed to have valid defaults, but the defaults are missing

## How was this PR tested?

Tested in editor

PR #15019, opened 2023-03-09
Fix Physics Material Asset property from showing invalid asset in components

moraaar:
## What does this PR do?

Physics material reflection to edit context does not need an asset assigned, the components handle that if there is no asset it uses the default material. In order to show the text "(default)" when there is no asset assigned Physics materials were passing an null asset to the DefaultAsset property. This was ok before, but after [this change](https://github.com/o3de/o3de/pull/13959) this is reported as an error.

- Fixed by passing the default physics material from the material manager, which is an in-memory asset. 
- This was still reported as error because the asset property handler expected an asset job for the default asset, but since it's an in-memory asset there is no job for it and it was reported it as missing asset. Fixed by only reporting as missing asset if it is not found in asset manager.

## How was this PR tested?

Checked PhysX and terrain components that use physics material property and verified that they don't show as invalid anymore when it doesn't have a material assigned.

Before:
image
image

After:
image
image

### DIFF
--- a/Gems/Terrain/Code/Source/Components/TerrainPhysicsColliderComponent.h
+++ b/Gems/Terrain/Code/Source/Components/TerrainPhysicsColliderComponent.h
@@ -41,12 +41,14 @@ namespace Terrain
 
         AZStd::vector<AZStd::pair<AZ::u32, AZStd::string>> BuildSelectableTagList() const;
         void SetTagListProvider(const EditorSurfaceTagListProvider* tagListProvider);
-        AZ::Data::AssetId GetDefaultPhysicsAssetId() const;
 
         SurfaceData::SurfaceTag m_surfaceTag;
         AZ::Data::Asset<Physics::MaterialAsset> m_materialAsset;
 
     private:
+        friend class EditorTerrainPhysicsColliderComponent;
+        AZ::Data::AssetId GetDefaultPhysicsMaterialAssetIdEditContext() const;
+
         const EditorSurfaceTagListProvider* m_tagListProvider = nullptr;
     };
 
@@ -57,10 +59,12 @@ namespace Terrain
         AZ_TYPE_INFO(TerrainPhysicsColliderConfig, "{E9EADB8F-C3A5-4B9C-A62D-2DBC86B4CE59}");
         static void Reflect(AZ::ReflectContext* context);
 
-        AZ::Data::AssetId GetDefaultPhysicsAssetId() const;
-
         AZ::Data::Asset<Physics::MaterialAsset> m_defaultMaterialAsset;
         AZStd::vector<TerrainPhysicsSurfaceMaterialMapping> m_surfaceMaterialMappings;
+
+    private:
+        friend class EditorTerrainPhysicsColliderComponent;
+        AZ::Data::AssetId GetDefaultPhysicsMaterialAssetIdEditContext() const;
     };
 
 
--- a/Gems/Terrain/Code/Source/Components/TerrainPhysicsColliderComponent.h
+++ b/Gems/Terrain/Code/Source/Components/TerrainPhysicsColliderComponent.h
@@ -46,9 +46,6 @@ namespace Terrain
         AZ::Data::Asset<Physics::MaterialAsset> m_materialAsset;
 
     private:
-        friend class EditorTerrainPhysicsColliderComponent;
-        AZ::Data::AssetId GetDefaultPhysicsMaterialAssetIdEditContext() const;
-
         const EditorSurfaceTagListProvider* m_tagListProvider = nullptr;
     };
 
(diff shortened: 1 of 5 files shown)


Gems/Terrain/Code/Source/Components/TerrainPhysicsColliderComponent.h
```diff
@@ -57,10 +59,12 @@ namespace Terrain
...
         static void Reflect(AZ::ReflectContext* context);
 
-        AZ::Data::AssetId GetDefaultPhysicsAssetId() const;
-
         AZ::Data::Asset<Physics::MaterialAsset> m_defaultMaterialAsset;
         AZStd::vector<TerrainPhysicsSurfaceMaterialMapping> m_surfaceMaterialMappings;
+
+    private:
+        friend class EditorTerrainPhysicsColliderComponent;
+        AZ::Data::AssetId GetDefaultPhysicsMaterialAssetIdEditContext() const;
```
hultonha, line 67: suggestion: It might not be worth the hassle of making these types friends and just instead make `GetDefaultPhysicsMaterialAssetIdEditContext` public (especially as it's `const` anyway). The problem with friendship is now everything else in the class (not just this function) is accessible. I realize this is a POD type anyway and things are public by default, but it's not always the case. You can leave as-is but I generally try and avoid `friend` wherever possible, it's usually more trouble than it's worth.

Gems/Terrain/Code/Source/EditorComponents/EditorTerrainPhysicsColliderComponent.cpp
```diff
@@ -51,7 +52,7 @@ namespace Terrain
                     ->Attribute(AZ::Edit::Attributes::AutoExpand, true)
                     ->DataElement(AZ::Edit::UIHandlers::Default, &TerrainPhysicsColliderConfig::m_defaultMaterialAsset,
                         "Default Surface Physics Material", "Select a material to be used by unmapped surfaces by default")
-                        ->Attribute(AZ::Edit::Attributes::DefaultAsset, &TerrainPhysicsColliderConfig::GetDefaultPhysicsAssetId)
+                        ->Attribute(AZ::Edit::Attributes::DefaultAsset, &TerrainPhysicsColliderConfig::GetDefaultPhysicsMaterialAssetIdEditContext)
```
hultonha, line 55: update: Actually looking at this you shouldn't need to add the member function at all, you can just move `static AZ::Data::AssetId GetDefaultPhysicsMaterialAssetId()` up and then pass it here....

```
->Attribute(AZ::Edit::Attributes::DefaultAsset, &GetDefaultPhysicsMaterialAssetId)
```

I'm pretty sure that should work no?

hultonha (commented):
I think the AssetId usage in the EditContext could be simplified (see comment) as no member data is being used. Would you be able to double check and see if that's possible? Thanks!

hultonha:
Thanks for including the before/after screenshots too, super helpful to see 👍

Gems/Terrain/Code/Source/Components/TerrainPhysicsColliderComponent.h
moraaar: `GetDefaultPhysicsMaterialAssetIdEditContext` is only meant to be used for edit context for the asset attribute, if this is made public it looks like users can call it to obtain the default material and that's not something these classes has to offer.

In this particular case the friend class `EditorTerrainPhysicsColliderComponent` is actually doing the edit context reflection for it, so I think it makes sense it can access its properties. Because of this I think it's justified to make it friend.

A different solution would have been to move the edit context reflection to itself, but that's a larger change than required.

Gems/Terrain/Code/Source/EditorComponents/EditorTerrainPhysicsColliderComponent.cpp
hultonha: I'm pretty sure free functions do work, so you can remove the member functions and friends entirely and keep the function in the .cpp

moraaar: Nice! It works! I'll remove the function members. I thought this line was checking the passed type belonged to the class https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Serialization/EditContext.h#L903

Thanks!

moraaar: Changed. Thanks!

Gems/Terrain/Code/Source/Components/TerrainPhysicsColliderComponent.h
moraaar: Using free function to remove the member functions. See comment below. Thanks @hultonha for the suggestion!

hultonha (approved):
I do think it would be fine to have this be in a `.h` file somewhere so we could avoid the small amount of duplication (especially if this pops up in other places) but it's not a huge deal, otherwise looks good 👍

moraaar:
> I do think it would be fine to have this be in a `.h` file somewhere so we could avoid the small amount of duplication (especially if this pops up in other places) but it's not a huge deal, otherwise looks good 👍

Thanks. It's only in these 2 places because terrain deals with Physics Materials directly, but all the other components deal with Physics Material Slots. So I don't expect more components dealing with materials directly any time soon.

gadams3 (approved):

cgalvan (approved):

FuzzyCarterAWS (approved):

PR #15033, opened 2023-03-09
Add `--prune` option to AzAutoGen

lumberyard-employee-dm:
The prune option will remove any files from the generated output directories that were NOT produced by the current run of AzAutoGen.

This can be used to clean up state generated files that are still being tracked by CMake.

## How was this PR tested?

Commented out AutoGen generation for the [RpcTesterComponent.AutoComponent.xml](https://github.com/o3de/o3de-multiplayersample/blob/development/Gem/Code/multiplayersample_files.cmake#L11) and performed two iterative builds of the the MultiplayerSample.Client.Static target

Verified that the **first** iterative build re-ran CMake and removed the `RpcTesterComponent.AutoComponent.h` and `RpcTesterComponent.AutoComponent.cpp` files

Verified that the **second** iterative did not re-run CMake

### DIFF
--- a/cmake/AzAutoGen.py
+++ b/cmake/AzAutoGen.py
@@ -11,14 +11,17 @@ import re
 import sys
 import time
 import errno
-import shutil
 import fnmatch
-import filecmp
 import fileinput
-import importlib
+import logging
 import argparse
 import hashlib
+import pathlib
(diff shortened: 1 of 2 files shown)


amzn-ahmadkrm (approved):

cgalvan (approved):
Nice! Thanks for adding this

PR #15036, opened 2023-03-10
DPE: Support nested containers and specific patching in ReflectionAdapter

amzn-tmryan:
DRAFT

Contains:
- Fix for broken nested container action buttons which weren't working due to BoundContainer's combining state for root containers, nested containers, and non-container elements
- Created BoundContainer factory method to simplify their creation and maintenance
- Separated GetContainerNode's visiting into two cases; one for container elements (including nested containers) and one for containers
- Implemented Add/Remove/ClearNotify handlers in ReflectionAdapter::HandleMessage instead of performing DOM fixup in the BoundContainer
- Handler for container Notify messages above passes the root container node (the first container in any container subtree) back through the visitor in order to fixup BoundAdapterMessage paths, labels with container size information, etc

Problems/WIP:
- Re-visiting nodes does not perform as hoped; SerializeContext::EnumerateInstance does not acquire classElement/editData for the instance provided, so parent node data is lost and things like editData, parent container data, and visibility are lost
- Make constexpr strings constexpr, etc


amzn-tmryan:
This is being merged as https://github.com/o3de/o3de/pull/15587.

PR #15038, opened 2023-03-10
DPE: Update expansion state saving for chained adapters

amzn-tmryan:
DRAFT

Uploading current changes for chained adapter expansion state saving in case someone picks up this work.

For sure the fixup algorithms are missing. The rest just suffered a pretty nasty merge, but should be mostly intact. Mostly.


Code/Framework/AzCore/AzCore/DOM/DomPath.h
```diff
@@ -68,7 +68,7 @@ namespace AZ::Dom
     class Path final
     {
     public:
-        AZ_TYPE_INFO(AZ::Dom::Path, "{C0081C45-F15D-4F46-9680-19535D33C312}")
+        AZ_RTTI(AZ::Dom::Path, "{C0081C45-F15D-4F46-9680-19535D33C312}");
```
lumberyard-employee-dm, line 71: AZ_RTTI is not needed for reflection.
It is only needed if the Path class is base class of some other class that needs TypeInfo.
Also for AZ_RTTI, the fully qualified name should not be used, but instead just the name of "Path"
```suggestion
        AZ_TYPE_INFO(Path, "{C0081C45-F15D-4F46-9680-19535D33C312}");
```

PR #15039, opened 2023-03-10
Moving material assignment data types to LY integration and optimizing/batching material related notifications

gadams3:
## What does this PR do?

The material assignment data types were originally created for the material component to manage, look up, and serialize material and property overrides mapped to meshes and other objects. The material assignment class contains data needed to assign and serialize property overrides and other data that is really only used by the material component. This data is not needed for systems that work directly with material instances.

This PR relocates the material assignment data types to live with the material component, which is the only component using it as intended. The white box component was also using it, presumably as a convenience, but it only required a single material instance and was reapplying it to the mesh handle unnecessarily.

The mesh feature processor interface was updated with its own custom material map types that use the minimum amount of data needed by the feature processor. This includes the material instances and UV name remapping. With the removal of the material assignment notification bus, all of the material change tracking was also removed from the feature processor. Instead of doing this work, a single function was added to allow the owner of the mesh handle to set a ray tracing dirty flag.

The material component notification bus functions were all changed to send notifications affecting batches of materials. There is now a notification sent once per tick if needed when materials are initially created, when material properties change, and when materials are otherwise updated. 

The mesh and actor components were updated with all of these changes and now handle the material property change notification to set the ray tracing dirty flag on the handle. This should provide equivalent functionality without monitoring every material asset or sending notifications every time a material instance is rebuilt.

See related PR https://github.com/o3de/o3de-atom-sampleviewer/pull/594

## How was this PR tested?

Manual testing and visual confirmation with the material component.
Manual testing and visual confirmation with ASV, which was also updated with these changes.
ASV full test suite script stops progressing after the dynamic material test.
Continuing to test while PR is up.


Gems/Atom/Feature/Common/Code/Include/Atom/Feature/Mesh/MeshFeatureProcessor.h
```diff
@@ -190,9 +170,9 @@ namespace AZ
             const RPI::MeshDrawPacketLods& GetDrawPackets(const MeshHandle& meshHandle) const override;
             const AZStd::vector<Data::Instance<RPI::ShaderResourceGroup>>& GetObjectSrgs(const MeshHandle& meshHandle) const override;
             void QueueObjectSrgForCompile(const MeshHandle& meshHandle) const override;
-            void SetMaterialAssignmentMap(const MeshHandle& meshHandle, const Data::Instance<RPI::Material>& material) override;
-            void SetMaterialAssignmentMap(const MeshHandle& meshHandle, const MaterialAssignmentMap& materials) override;
-            const MaterialAssignmentMap& GetMaterialAssignmentMap(const MeshHandle& meshHandle) const override;
+            void SetCustomMaterials(const MeshHandle& meshHandle, const Data::Instance<RPI::Material>& material) override;
```
invertednormal, line 173: nit: seems weird for the function to be named `SetCurstomMaterials` when it's only for setting a singular material.

Gems/AtomLyIntegration/CommonFeatures/Code/Source/Material/MaterialComponentController.cpp
```diff
@@ -788,9 +806,18 @@ namespace AZ
...
         }
 
-        void MaterialComponentController::QueueMaterialUpdateNotification()
+        void MaterialComponentController::QueueMaterialsCreatedNotification()
+        {
+            m_queuedMaterialsCreatedNotification = true;
+            if (!SystemTickBus::Handler::BusIsConnected())
+            {
+                SystemTickBus::Handler::BusConnect();
+            }
```
invertednormal, line 815: nit: Doesn't `BusConnect()` already check internally for `BusIsConnected()`?

Gems/AtomLyIntegration/CommonFeatures/Code/Source/Mesh/MeshComponentController.cpp
```diff
@@ -409,6 +424,14 @@ namespace AZ
                 MaterialAssignmentMap materials;
                 MaterialComponentRequestBus::EventResult(materials, entityId, &MaterialComponentRequests::GetMaterialMap);
 
+                AZ::Render::CustomMaterialMap customMaterials;
+                for (const auto& materialAssignment : materials)
+                {
+                    customMaterials.emplace(
+                        AZ::Render::CustomMaterialId{ materialAssignment.first.m_lodIndex, materialAssignment.first.m_materialSlotStableId },
+                        AZ::Render::CustomMaterialInfo{ materialAssignment.second.m_materialInstance, materialAssignment.second.m_matModUvOverrides });
+                }
```
invertednormal, line 433: nit: Could make a function for this since it's identical to above.

Gems/AtomLyIntegration/EMotionFXAtom/Code/Source/AtomActorInstance.cpp
```diff
@@ -670,8 +685,15 @@ namespace AZ::Render
             meshDescriptor.m_isAlwaysDynamic = true;
             meshDescriptor.m_excludeFromReflectionCubeMaps = true;
 
+            AZ::Render::CustomMaterialMap customMaterials;
+            for (const auto& materialAssignment : materials)
+            {
+                customMaterials.emplace(
+                    AZ::Render::CustomMaterialId{ materialAssignment.first.m_lodIndex, materialAssignment.first.m_materialSlotStableId },
+                    AZ::Render::CustomMaterialInfo{ materialAssignment.second.m_materialInstance, materialAssignment.second.m_matModUvOverrides });
+            }
```
invertednormal, line 694: nit: More duplication... maybe this material assignment map to custom materials code could be in one shared location.

Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
```diff
@@ -1736,56 +1716,25 @@ namespace AZ
...
-            {
-                if (CheckForMaterialChanges())
-                {
-                    m_needsSetRayTracingData = true;
-
-                    // update the material changeId list with the latest materials
-                    UpdateMaterialChangeIds();
-                }
-            }
+            return CustomMaterialInfo{};
```
invertednormal, line 1737: nit: this could be simplified a bit to something like:
```
CustomMaterialInfo ModelDataInstance::GetCustomMaterialWithFallback(const CustomMaterialId& customId) const
{
    const CustomMaterialId ignoreLodId(DefaultCustomMaterialLodIndex, id.second);
    for (auto id : {customId, ignoreLodId, DefaultCustomMaterialId})
    {
        if (auto itr = m_customMaterials.find(id); itr != m_customMaterials.end() && itr->second.m_material)
        {
            return itr->second;
        }
    }
    return CustomMaterialInfo{};
}
```

invertednormal (approved):
Looks good, just a few nits. Only thing I'm not thrilled about is the need to explicitly set the ray tracing to dirty instead of it being able to pick up something has changed automatically.

amzn-tommy (approved):
I'd still prefer to see an RPI event for notifying when a material instance changes. But overall this looks like an improvement over what we have, since the existing notification is still relying on something higher level sending the notification anyways. So I don't consider that blocking.

Gems/AtomLyIntegration/CommonFeatures/Code/Source/Material/MaterialComponentController.cpp
gadams3: Yes it does indeed!

gadams3:
> I'd still prefer to see an RPI event for notifying when a material instance changes. But overall this looks like an improvement over what we have, since the existing notification is still relying on something higher level sending the notification anyways. So I don't consider that blocking.

Adding the events to the material class should be a pretty easy follow up to this.
I also want to change the feature processor acquire mesh function and descriptor so that custom materials are assigned to the descriptor. That's probably also a good place to move the model change notification event. Then event handlers can be connected prior to calling acquire so we don't need to check if the model is loaded every time we call acquire mesh.

PR #15052, opened 2023-03-10
StableDynamicStructOfArrays

amzn-tommy:
## What does this PR do?

This PR introduces a modified version of StableDynamicArray, which is a container that stores data in pages. StableDynamicStructOfArrays has multiple 'rows', such that each row exists in its own chunk of memory, and systems that only need to iterate over part of the data can do so with better cache efficiency. The motivation is that the MeshFeatureProcessor, which uses StableDynamicArray, has object data that has grown over time, and it is now slow to iterate over due to its size. StableDynamicStructOfArrays will allow the MeshFeatureProcessor to split that data up, and when only a subset of the data needs to be iterated over, it can do so more efficiently. This is especially useful for mesh instancing, which has a post-culling process that is only interested in a small subset of the data from a ModelDataInstance.

To review this change, I recommend looking at a diff between the first commit of the change and the final commit. The first commit is an exact copy-paste of StableDynamicArray, only renamed to StableDynamicStructOfArrays. Starting from there, you can see the difference between StableDynamicArray and StableDynamicStructOfArrays.

One feature of StableDynamicArray that I did not support was the ability to move handles between types. This was originally implemented to support a pattern that we no longer use in Atom, which was to have a RenderProxy type associated with a feature processor, and have a RenderProxyInterface that was exposed to outside gems such as AtomLyIntegration. That pattern is no longer used (everything goes through a FeatureProcessorInterface instead), so I recommend we deprecate that functionality from the StableDynamicArray as well.

There are two pieces of syntactic sugar that I left out of this initial commit, which I think it would benefit from long term.
1. Rows must be accessed by index, not by type. AZStd::tuple::get<> supports both type and index based access. The type based access is friendlier to use, but it's less generic since it will not work if there are multiple rows of the same type.
2. Iterators can only access data via a GetItem<> function, instead of via operator* or operator->. This is because they need to know which row to access, and GetItem<> is templated to specify which row. I think it would be beneficial to have typed iterators that know which specific row they are iterating over, such that they can only access data from that row, and can then use operator* or operator-> to access the data like the original StableDynamicArray.

With these two changes, I think we could deprecate the original StableDynamicArray completely, and use StableDynamicStructOfArrays with a single row for cases where StableDynamicArray is used now, since StableDynamicStructOfArrays is a generalization of the single row use-case. But they are out of scope for the initial commit.

## How was this PR tested?

Unit tests. An earlier iteration of this container was tested in the MeshInstancing prototype (and confirmed the performance improvements gained by splitting up the data), but it has been through several rounds of updates since that have only been tested with the unit tests.

### DIFF
--- a/Gems/Atom/Utils/Code/Include/Atom/Utils/StableDynamicStructOfArrays.h
+++ b/Gems/Atom/Utils/Code/Include/Atom/Utils/StableDynamicStructOfArrays.h
@@ -157,14 +157,14 @@ namespace AZ
         /// True if this page is completely empty
         bool IsEmpty() const;
 
-        AZStd::tuple<value_types *...> GetItems(size_t index);
+        ItemTupleType GetItems(size_t index);
 
         /**
         * Gets a specific item from the Page
         * Note: may return empty slots.
         */
         template<size_t RowIndex>
-        AZStd::tuple_element_t<RowIndex, AZStd::tuple<value_types...>>* GetItem(size_t index);
+        auto* GetItem(size_t index);
 
         /// Gets the number of items allocated on this page.
         size_t GetItemCount() const;
@@ -197,7 +197,7 @@ namespace AZ
         // TODO: is this operator needed?
         const this_type& operator*() const;
         template <size_t RowIndex>
-        auto* GetItem() const;
+        auto& GetItem() const;
 
         bool operator==(const this_type& rhs) const;
         bool operator!=(const this_type& rhs) const;
@@ -231,7 +231,7 @@ namespace AZ
         explicit const_iterator(Page* firstPage);
 
         template<size_t RowIndex>
-        auto* GetItem() const;
+        auto& GetItem() const;
 
         this_type& operator++();
         this_type operator++(int);
@@ -252,7 +252,7 @@ namespace AZ
 
         const this_type& operator*() const;
         template<size_t RowIndex>
-        auto* GetItem() const;
+        auto& GetItem() const;
 
         bool operator==(const this_type& rhs) const;
         bool operator!=(const this_type& rhs) const;
@@ -282,6 +282,9 @@ namespace AZ
     class StableDynamicStructOfArraysWeakHandle
     {
     public:
+        using ItemTupleType = AZStd::tuple<value_types*...>;
+
+        /// Default constructor creates an invalid handle.
         StableDynamicStructOfArraysWeakHandle() = default;
 
         /// Returns true if this Handle currently holds a valid value.
@@ -290,15 +293,15 @@ namespace AZ
         /// Returns true if this Handle doesn't contain a value (same as !IsValid()).
         bool IsNull() const;
 
-        AZStd::tuple<value_types*...>& operator*();
+        ItemTupleType& operator*();
 
         template <size_t RowIndex>
-        auto* GetItem() const;
+        auto& GetItem() const;
         
     private:
-        StableDynamicStructOfArraysWeakHandle(const AZStd::tuple<value_types*...>& data);
+        StableDynamicStructOfArraysWeakHandle(const ItemTupleType& data);
 
-        AZStd::tuple<value_types*...> m_data;
+        ItemTupleType m_data;
 
         template<typename... value_types>
         friend class StableDynamicStructOfArraysHandle;
@@ -315,6 +318,7 @@ namespace AZ
         template<size_t ElementsPerPage, class Allocator, typename... value_types>
         friend class StableDynamicStructOfArrays;
     public:
+        using ItemTupleType = AZStd::tuple<value_types*...>;
 
         /// Default constructor creates an invalid handle.
         StableDynamicStructOfArraysHandle() = default;
@@ -337,17 +341,17 @@ namespace AZ
         /// Returns true if this Handle doesn't contain a value (same as !IsValid()).
         bool IsNull() const;
 
-        AZStd::tuple<value_types*...>& operator*();
+        ItemTupleType& operator*();
 
         /// Returns a pointer to the item in the row specified by RowIndex
         template <size_t RowIndex>
-        auto* GetItem() const;
+        auto& GetItem() const;
 
         /// Returns a non-owning weak handle to the data
         StableDynamicStructOfArraysWeakHandle<value_types...> GetWeakHandle() const;
     private:
         template<typename PageType>
-        StableDynamicStructOfArraysHandle(AZStd::tuple<value_types *...> data, PageType* page);
+        StableDynamicStructOfArraysHandle(ItemTupleType data, PageType* page);
 
         StableDynamicStructOfArraysHandle(const StableDynamicStructOfArraysHandle&) = delete;
 
@@ -357,7 +361,7 @@ namespace AZ
         /// Called for valid handles on delete so the underlying data can be removed from the StableDynamicArray
         HandleDestructor m_destructorCallback = nullptr; 
         void* m_page = nullptr; ///< The page the data this Handle points to was allocated on.
-        AZStd::tuple<value_types *...> m_data;
+        ItemTupleType m_data;
     };
     
     /// Used for returning information about the internal state of the StableDynamicStructOfArrays.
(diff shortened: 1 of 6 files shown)


Gems/Atom/Utils/Code/Tests/StableDynamicStructOfArraysTests.cpp
```diff
@@ -0,0 +1,887 @@
...
+        static int s_testItemsModified;
+    };
+
+    int StableDynamicStructOfArraysHandleTests::s_testItemsConstructed = 0;
+    int StableDynamicStructOfArraysHandleTests::s_testItemsDestructed = 0;
+    int StableDynamicStructOfArraysHandleTests::s_testItemsModified = 0;
+
+    // Class used to test that the right number of items are created, modified, and destroyed
+    // Follows similar pattern to what a FeatureProcessor might do
+    class StableDynamicStructOfArraysOwner
```
lumberyard-employee-dm, line 574: Since the `StableDynamicStructOfArrays` was written to be used in scenarios where the performance `StableDynamicArray` was not sufficient.

This should add AZ Benchmark test that validates that this container performances better in those use cases.

Gems/Atom/Utils/Code/Include/Atom/Utils/StableDynamicStructOfArrays.inl
```diff
@@ -314,46 +354,77 @@ namespace AZ
...
-    void StableDynamicStructOfArrays<T, ElementsPerPage, Allocator>::Page::Free(T* item)
+    template<size_t ElementsPerPage, class Allocator, typename... value_types>
+    void StableDynamicStructOfArrays<ElementsPerPage, Allocator, value_types...>::Page::Free(ItemTupleType item)
     {
-        // use the difference between the item pointer and the page data to find the index
-        size_t index = item - reinterpret_cast<T*>(&m_data);
+        // Use the first row of data to get the index within the page
+        AZStd::tuple_element_t<0, ItemTupleType> pageStart = reinterpret_cast<AZStd::tuple_element_t<0, ItemTupleType>>(&AZStd::get<0>(m_data));
+        AZStd::tuple_element_t<0, ItemTupleType> itemStart = AZStd::get<0>(item);
+        size_t index =  itemStart - pageStart;
```
invertednormal, line 369: nit: extra space

Gems/Atom/Utils/Code/Include/Atom/Utils/StableDynamicStructOfArrays.inl
```diff
@@ -314,46 +354,77 @@ namespace AZ
...
+
         // This item's flag will be in the uint64_t at index >> 6 (index / 64). Mark the appropriate bit as 0 (Free).
         AZ_Assert(m_bits[index >> 6] & (1ull << (index & 0x3F)), "Freeing item that is already marked as free!"); // The IsValid() check on handles should prevent this ever happening.
         m_bits[index >> 6] &= ~(1ull << (index & 0x3F));
         // Set the bit start index so the next Reserve() starts on a uint_64 that likely has space.
-        m_bitStartIndex = AZStd::min(index >> 6, m_bitStartIndex);
+        m_bitStartIndex = AZStd::min(static_cast<size_t>(index) >> 6, m_bitStartIndex);
```
invertednormal, line 375: nit: since index is a size_t, it doesn't seem like this should be necessary.

Gems/Atom/Utils/Code/Include/Atom/Utils/StableDynamicStructOfArrays.inl
```diff
@@ -314,46 +354,77 @@ namespace AZ
...
+        PageType page,
+        size_t pageElementIndex,
+        AZStd::index_sequence<Ints...>)
+    {
+        ((AZStd::get<Ints>(items) = page->GetItem<Ints>(pageElementIndex)), ...);
+    }
+
+    template<size_t ElementsPerPage, class Allocator, typename... value_types>
+    typename StableDynamicStructOfArrays<ElementsPerPage, Allocator, value_types...>::Page::ItemTupleType
+        StableDynamicStructOfArrays<ElementsPerPage, Allocator, value_types...>::Page::GetItems(size_t elementIndex)
```
invertednormal, line 405: nit: good candidate for putting the type after the function (`->`)

Gems/Atom/Utils/Code/Include/Atom/Utils/StableDynamicStructOfArrays.inl
```diff
@@ -696,82 +762,62 @@ namespace AZ
...
-        m_data = nullptr;
+        AZStd::apply(
+            // We take a reference to the dataItem pointers here, since we're modifying the pointers
+            // themselves not the data they're pointing to
+            [](value_types *&... dataItems)
+            {
+                ((dataItems = nullptr), ...);
+            },
+            m_data
+        );
```
invertednormal: nit: technically you only need to set the first pointer to null. Setting them all to null feels a bit cleaner, but it's not as fast 😄  up to you.

invertednormal (approved):

Gems/Atom/Utils/Code/Include/Atom/Utils/StableDynamicStructOfArrays.h
```diff
@@ -0,0 +1,389 @@
...
+        size_t m_bitStartIndex = 0; ///< Index of the first uint64_t that might have space.
+        Page* m_nextPage = nullptr; ///< pointer to the next page.
+        StableDynamicStructOfArrays<ElementsPerPage, Allocator, value_types...>*
+            m_container; ///< pointer to the container this page was allocated from.
+        size_t m_pageIndex = 0; ///< used for comparing pages when items are freed so the earlier page in the list can be cached.
+        size_t m_itemCount = 0; ///< the number of items in the page.
+        AZStd::array<uint64_t, NumUint64_t> m_bits; ///< Bits representing free slots in the array. Free slots are 1, occupied slots are 0.
+
+        using PageTupleType = AZStd::tuple<AZStd::aligned_storage_t<ElementsPerPage * sizeof(value_types), alignof(value_types)>...>;
+        PageTupleType m_data; ///< aligned storage for all the actual data.
```
lumberyard-employee-dm, line 188: [aligned_storage_t](https://stackoverflow.com/a/71828512) has been deprecated and instead we should something like `alignas(value_types) AZStd::byte[sizeof(value_types) * ElementsPerPage]`
```suggestion
        using PageTupleType = AZStd::tuple<AZStd::byte[ElementsPerPage * sizeof(value_types)] ...>;
        alignas(value_types...) PageTupleType m_data; ///< aligned storage for all the actual data.
```

Gems/Atom/Utils/Code/Tests/StableDynamicStructOfArraysTests.cpp
amzn-tommy: Great suggestion. I'm not getting the same improvements in the benchmark that I saw when using this in the MeshFeatureProcessor. I don't know why it worked so well on real data, possibly I had another change mixed in with the switch to the struct-of-arrays that was the true cause of the speedup.

But I have a pretty good idea why I'm not seeing an improvement in the benchmark. When passing in two structs, the data from the two structs is still lumped together in a single chunk of memory. I had mistakenly thought that the aligned_storage_t in the original StableDynamicArray was a reference to the actual data in a separate allocation from the Page itself, but now I can see that the aligned storage is part of the page itself. 

So in this case, a tuple of aligned_storage for two structs is no better than a single aligned storage for one. I think what I really need is pointers to the data from each row, plus the ability to pass in a different pool allocator for each row so that the page data for each row is closer together in memory. I will have to experiment more with the benchmark.

Gems/Atom/Utils/Code/Include/Atom/Utils/StableDynamicStructOfArrays.h
```diff
@@ -19,60 +19,65 @@ namespace AZ
...
     * arrays.
     *   It will always place new items at the front-most slot of the first array with available space.
     * DefragmentHandle() can be called to reorganize data to reduce the amount of empty slots.
     **/
-    template<typename T, size_t ElementsPerPage = 512, class Allocator = AZStd::allocator>
+
+    // The template paramaterization is a little different than StableDynamicArray. The ElementsPerPage and Allocator
+    // come first, and do not have default arguments, so that it is explicit which arguments are part of the parameter pack (...)
+    // see https://en.cppreference.com/w/cpp/language/parameter_pack
+    // "In a primary class template, the template parameter pack must be the final parameter in the template parameter list."
```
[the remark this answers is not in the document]
siliconvoodoo: still nice educatory comment

PR #15072, opened 2023-03-13
Make inheritance of parent entity's transform optional at runtime

greerdv:
## What does this PR do?
Implements https://github.com/o3de/sig-simulation/issues/57. Adds an option to the Transform interface which allows the behaviour of updating a child entity's transform when its parent's transform changes to be disabled at runtime.

## How was this PR tested?
Manual testing in the editor.

https://user-images.githubusercontent.com/82226198/224714366-26ff36a4-fa3e-4cf2-b052-59bc2879d014.mp4

### DIFF
--- a/Code/Framework/AzCore/AzCore/Component/TransformBus.h
+++ b/Code/Framework/AzCore/AzCore/Component/TransformBus.h
@@ -29,6 +29,13 @@ namespace AZ
     enum class ChildChangeType { Added, Removed };
     using ChildChangedEvent = Event<ChildChangeType, EntityId>;
 
+    //! Used to control the behavior of an entity's transform when its parent's transform changes at runtime.
+    enum class OnParentChangedBehavior : AZ::u8
+    {
+        Update, //!< Update this entity's transform based on the parent's new world transform and this entity's local transform.
+        DoNotUpdate //!< Do not update this entity's world transform when the parent's transform changes.
+    };
+
     //! Interface for AZ::TransformBus, which is an EBus that receives requests
     //! to translate (position), rotate, and scale an entity in 3D space. It
     //! also receives requests to get and set the parent of an entity and get
@@ -286,6 +293,9 @@ namespace AZ
         //! A static transform is unmovable and does not respond to requests that would move it.
         virtual void SetIsStaticTransform([[maybe_unused]] bool isStatic) {}
         //! @}
+
+        //! Set the behavior at runtime when this entity's parent's transform changes.
+        virtual void SetOnParentChangedBehavior([[maybe_unused]] OnParentChangedBehavior onParentChangedBehavior) {}
     };
 
     //! The EBus for requests to position and parent an entity.
--- a/Code/Framework/AzFramework/AzFramework/Components/TransformComponent.cpp
+++ b/Code/Framework/AzFramework/AzFramework/Components/TransformComponent.cpp
@@ -642,7 +642,7 @@ namespace AzFramework
                 // but keeps our world transform unchanged. Do not send any notifications here, because our world
                 // transform has not changed, and with this OnParentChangedBehavior setting the expectation is that
                 // another system will update our transform, and the notification will be triggered then.
-                m_localTM = m_parentTM->GetWorldTM().GetInverse() * m_worldTM;
+                m_localTM = parentWorldTM.GetInverse() * m_worldTM;
             }
         }
     }
--- a/Code/Framework/AzFramework/AzFramework/Components/TransformComponent.cpp
+++ b/Code/Framework/AzFramework/AzFramework/Components/TransformComponent.cpp
@@ -466,6 +466,11 @@ namespace AzFramework
         return m_isStatic;
     }
 
+    void TransformComponent::SetOnParentChangedBehavior(AZ::OnParentChangedBehavior onParentChangedBehavior)
+    {
+        m_onParentChangedBehavior = onParentChangedBehavior;
+    }
+
     void TransformComponent::OnTransformChanged(const AZ::Transform& parentLocalTM, const AZ::Transform& parentWorldTM)
     {
         OnTransformChangedImpl(parentLocalTM, parentWorldTM);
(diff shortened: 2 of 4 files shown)


Code/Framework/AzCore/AzCore/Component/TransformBus.h
```diff
@@ -286,6 +293,9 @@ namespace AZ
         //! A static transform is unmovable and does not respond to requests that would move it.
         virtual void SetIsStaticTransform([[maybe_unused]] bool isStatic) {}
         //! @}
+
+        //! Set the behavior at runtime when this entity's parent's transform changes.
+        virtual void SetOnParentChangedBehavior([[maybe_unused]] OnParentChangedBehavior onParentChangedBehavior) {}
```
moraaar, line 298: Worth adding a getter to the API as well?

Code/Framework/AzFramework/AzFramework/Components/TransformComponent.cpp
```diff
@@ -624,10 +629,21 @@ namespace AzFramework
...
+                    m_notificationBus, &AZ::TransformNotificationBus::Events::OnTransformChanged, m_localTM, m_worldTM);
+                m_transformChangedEvent.Signal(m_localTM, m_worldTM);
+            }
+            else
+            {
+                // Update the local transform to make sure it is consistent with the parent's new transform
+                // but keeps our world transform unchanged. Do not send any notifications here, because our world
+                // transform has not changed, and with this OnParentChangedBehavior setting the expectation is that
+                // another system will update our transform, and the notification will be triggered then.
+                m_localTM = parentWorldTM.GetInverse() * m_worldTM;
```
moraaar, line 645: When another system (rigid body for now) will update the transform, is it updating both the world and the local transforms? or is the local transform getting "out of sync" with the parent?

Code/Framework/AzCore/AzCore/Component/TransformBus.h
```diff
@@ -29,6 +29,13 @@ namespace AZ
     enum class ChildChangeType { Added, Removed };
     using ChildChangedEvent = Event<ChildChangeType, EntityId>;
 
+    //! Used to control the behavior of an entity's transform when its parent's transform changes at runtime.
+    enum class OnParentChangedBehavior : AZ::u8
+    {
+        Update, //!< Update this entity's transform based on the parent's new world transform and this entity's local transform.
```
hultonha, line 35: suggestion: Might be worth adding a note that this is the 'default' behavior

Code/Framework/AzFramework/AzFramework/Components/TransformComponent.h
```diff
@@ -186,5 +187,7 @@ namespace AzFramework
         bool m_parentActive = false; ///< Keeps track of the state of the parent entity.
         bool m_onNewParentKeepWorldTM = true; ///< If set, recompute localTM instead of worldTM when parent becomes active.
         bool m_isStatic = false; ///< If true, the transform is static and doesn't move while entity is active.
+        /// Behavior for this entity's transform when its parent's transform changes.
```
hultonha, line 190: suggestion: Might be worth mentioning something like 'see OnParentChangedBehavior definition for definitions of behavior types' (maybe overkill but could be useful to give readers a hint)

Gems/PhysX/Code/Source/RigidBodyComponent.cpp
```diff
@@ -404,6 +404,8 @@ namespace PhysX
         m_interpolator = std::make_unique<TransformForwardTimeInterpolator>();
         m_interpolator->Reset(transform.GetTranslation(), rotation);
 
+        GetEntity()->GetTransform()->SetOnParentChangedBehavior(AZ::OnParentChangedBehavior::DoNotUpdate);
```
hultonha, line 407: suggestion: A comment here to indicate _why_ we're doing this would be helpful for context

hultonha (commented):
Changes look good but it really would be good to have a couple tests to sanity check this low level change in behavior. This could be done as a follow-up PR but I do think it's important we have some kind of test to validate this new behavior.

Code/Framework/AzFramework/AzFramework/Components/TransformComponent.cpp
greerdv: The point is that the world transform will not change, but the local would (because the parent has moved relative to the child).

Code/Framework/AzCore/AzCore/Component/TransformBus.h
greerdv: Good call, will update.

Code/Framework/AzCore/AzCore/Component/TransformBus.h
greerdv: Yeah, I guess it at least won't do any harm.

Code/Framework/AzFramework/AzFramework/Components/TransformComponent.h
greerdv: That does feel a bit overkill to me tbh. The first thing I would do reading the code is F12 on the type, which takes you straight to the definitions. Also I think they're fairly clear from the names.

Gems/PhysX/Code/Source/RigidBodyComponent.cpp
greerdv: Yeah, good idea, will update.

Code/Framework/AzFramework/AzFramework/Components/TransformComponent.cpp
moraaar: >When physics (or whatever) calls SetWorldTransform, that will update both the world and child transform, so they won't get out of sync.

Perfect, thanks for the clarification!

moraaar (approved):

greerdv:
> Changes look good but it really would be good to have a couple tests to sanity check this low level change in behavior. This could be done as a follow-up PR but I do think it's important we have some kind of test to validate this new behavior.

I think I'll need to do that a follow up. I looked at the transform component tests and they aren't in a fantastic state.

hultonha:
> > Changes look good but it really would be good to have a couple tests to sanity check this low level change in behavior. This could be done as a follow-up PR but I do think it's important we have some kind of test to validate this new behavior.
> 
> I think I'll need to do that a follow up. I looked at the transform component tests and they aren't in a fantastic state.

Okay understood, thanks for checking. Would be nice to have something small for a follow-up, hopefully more can then be added in future as there'll be less inertia to add new tests. Thanks!

hultonha (approved):

PR #15077, opened 2023-03-13
Fix for component alias not being initialized

michabr:
## What does this PR do?

Add a helper to get the component alias and initialize it if it hasn't been initialized yet.
This fix will need some tweaks when we remove the component Id dependency on the alias.

## How was this PR tested?

Manual testing in editor.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/DocumentPropertyEditor/PrefabComponentAdapter.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/DocumentPropertyEditor/PrefabComponentAdapter.cpp
@@ -34,7 +34,8 @@ namespace AzToolsFramework::Prefab
     {
         // Set the component alias before calling SetValue() in base SetComponent().
         // Otherwise, an empty component alias will be used in DOM data.
-        m_componentAlias = componentInstance->GetSerializedIdentifier();
+        AZ_Assert(componentInstance, "PrefabComponentAdapter::SetComponent - component is null.")
+        m_componentAlias = PrefabDomUtils::GetComponentAliasWithInitialization(*componentInstance);
         AZ_Assert(!m_componentAlias.empty(), "PrefabComponentAdapter::SetComponent - Component alias should not be empty.");
 
         ComponentAdapter::SetComponent(componentInstance);
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabDomUtils.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabDomUtils.cpp
@@ -545,6 +545,16 @@ namespace AzToolsFramework
                 return prefabBuffer.GetString();
             }
 
+            AZStd::string GetComponentAliasWithInitialization(AZ::Component& component)
+            {
+                AZStd::string componentAlias = component.GetSerializedIdentifier();
+                if (componentAlias.empty())
+                {
+                    componentAlias = AZStd::string::format("Component_[%llu]", component.GetId());
+                    component.SetSerializedIdentifier(componentAlias);
+                }
+                return componentAlias;
+            }
         } // namespace PrefabDomUtils
     } // namespace Prefab
 } // namespace AzToolsFramework
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabDomUtils.h
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabDomUtils.h
@@ -214,6 +214,8 @@ namespace AzToolsFramework
 
             AZStd::string PrefabDomValueToString(const PrefabDomValue& prefabDomValue);
 
+            AZStd::string GetComponentAliasWithInitialization(AZ::Component& component);
+
             //! An empty struct for passing to JsonSerializerSettings.m_metadata that is consumed by InstanceSerializer::Store.
             //! If present in metadata, linkIds will be stored to instance dom.
             struct LinkIdMetadata


Code/Framework/AzCore/AzCore/Component/Component.cpp
```diff
@@ -163,7 +163,7 @@ namespace AZ
         }
     }
 
-    AZStd::string Component::GetSerializedIdentifier() const
+    AZStd::string Component::GetSerializedIdentifier()
```
michabr, line 166: I'm considering renaming this method since it's now non-const...

Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorComponentBase.cpp
```diff
@@ -102,9 +106,21 @@ namespace AzToolsFramework
             m_alias = alias;
         }
 
-        AZStd::string EditorComponentBase::GetSerializedIdentifier() const
+        AZStd::string EditorComponentBase::GetSerializedIdentifier()
         {
+            // Initialize if necessary. This is needed because not all derived components are calling their base class Init()
+            if (m_alias.empty())
+            {
+                InitSerializedIdentifier();
```
srikappa-amzn, line 114: Instead of leaving letting alias be created in any Get() call, I think it would be easier to maintain this if we explicitly create the alias in EditorComponentBase() constructor. If the id is empty by the time the constructor is called, you can create the id there too doing this : https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Component/Component.cpp#L156

The id gets set later at some point when SetEntity() is called on the component. So, I think its fine to call this during the component creation itself, (particularly because the creation of id seems to have nothing to do with entity data itself).

lumberyard-employee-dm: The Id is always going to be empty when the constructor.
It is the constructor that initializes the data.

The steps of initialization that are relevant to determining the component ID are as follows
1. Component construction - Component ID is not set yet.
1. (If Component is loaded from serialized data) Entity de-serialization - Component ID can be loaded from prefab data
1. Entity::Init - Calls `Component::SetEntity`  to set Component ID

michabr: I'm not sure if it makes sense to initialize the alias in the constructor because a component that is loaded from serialized data will set that value to something else.

srikappa-amzn: I think the fact that deserialization may alter the alias is fine imo. The problem this PR is trying to solve is , "I created a new component in the editor, I don't know what its alias should be". For cases where we load the component from prefab data, there is currently no issue. We simply read it from json.

However, if we create a new component, we don't want to deserialize again immediately just to give it a unique alias. So, starting with a placeholder alias in the constructor sounds like a reasonable approach to me. If the id is empty in the constructor, we can create that too. I can't think of that creating any issues.

lumberyard-employee-dm: That breaks an invariant that the Component Id is set to `InvalidComponentId` if it is not attached to an entity.

There is code that uses that constraint to determine how to lookup a component on an entity (https://github.com/o3de/o3de/blob/9ef05ce7fa90d6299e384af775c27ca680cd5f7a/Code/Framework/AzToolsFramework/AzToolsFramework/Entity/EntityUtilityComponent.cpp#L124-L131).

michabr: We don't have to incorporate componentId to construct the alias. We can generate our own Id without referencing componentId.

forkercat (approved):
Looks good. Does it also fix the assertions from "Create Entity"?

Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorComponentBase.cpp
lumberyard-employee-dm: Because the ComponentId is set, it wouldn't go into the else block that uses the `AZ::TypeId`, instead it would use the created component Id which should work in all circumstances unless their is a random number collision which should be on the order of 2^64.

lumberyard-employee-dm (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorComponentBase.cpp
```diff
@@ -49,6 +49,17 @@ namespace AzToolsFramework
             m_transform = nullptr;
         }
 
+        void EditorComponentBase::SetEntity(AZ::Entity* entity)
```
srikappa-amzn, line 52: I've noticed that SetEntity can get called multiple times with the same entity. So, it'd be good to have a 
```
if (m_entity != entity)
```
check wrapping the code in this function too similar to Component::SetEntity()

Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorComponentBase.cpp
```diff
@@ -49,6 +49,17 @@ namespace AzToolsFramework
...
+        {
+            Component::SetEntity(entity);
+
+            if (m_alias.empty())
+            {
+                AZ_Assert(GetId() != AZ::InvalidComponentId, "ComponentId is invalid.");
```
srikappa-amzn, line 58: Component::SetEntity() and Entity::AddComponent() are not guaranteed to have a valid m_id by the time code reaches this point. So, instead of asserting here, just check that m_id is valid and only create alias if it is.

Code/Framework/AzCore/AzCore/Component/EntitySerializer.cpp
```diff
@@ -206,7 +206,9 @@ namespace AZ
         {
             if (component)
             {
-                componentMapOut.emplace(AZStd::string::format("Component_[%llu]", component->GetId()), component);
+                AZStd::string componentAlias = component->GetSerializedIdentifier();
+                AZ_Assert(!componentAlias.empty(), "Component alias is empty.");
```
srikappa-amzn, line 210: It'd be good to have a fallback solution of ("Component_[%llu]")in case the alias is empty so that serialization doesn't stop/break.

srikappa-amzn (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorComponentBase.cpp
michabr: Oh yeah it looks like m_id can be invalid if SetEntity is called with null. I'll keep the assert but add a check if entity is valid before setting the alias. m_id is expected to be valid if the component is on a valid entity.

Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorComponentBase.cpp
michabr: I think it's safer to always call the base class Component::SetEntity() so we don't make assumptions on the base class behavior.

Code/Framework/AzCore/AzCore/Component/EntitySerializer.cpp
michabr: OK, I'll add a fallback.

PR #15080, opened 2023-03-13
Fix server entities not ticking when outside of a camera frustum view

AMZN-Olex:
## What does this PR do?

On a server with a camera, even with -rhi=null, only the entities within the camera frustum would tick. This results in an incorrect root transform for the players and their animation actors. In turns, this breaks bones' positions, and the firing ends up broken as well, since the server disagrees with the client on the origin of the shot based on the shooting bone.

## How was this PR tested?

in o3de-multiplayer with a dedicated server with camera looking away from players


AMZN-stankowi (approved):

mbalfour-amzn (approved):

AMZN-Gene (approved):
Nice find!

Issue #14865, opened 2023-02-28
Reflection: Component should not display the component name row in DPE Inspector


forkercat:
**Describe the bug**

Prerequisite: `ed_enableDPE` is on

Seeing an extra row for the component name, which is redundant and it takes up space of one level of indentation.

**Steps to reproduce**
1. Enable DPE flag
1. Create an empty level
1. Create an entity and see it in Entity Inspector

**Expected behavior**

Like we're seeing in RPE Entity Inspector:

<img width=200px src="https://user-images.githubusercontent.com/11157226/222004297-742ded88-0723-49a7-bcb8-87966397f600.png">

**Actual behavior**

<img width=200px src="https://user-images.githubusercontent.com/11157226/222004354-e9a00747-3921-4807-a46d-fe4332123b7a.png">

**Screenshots/Video**

See screenshot above

**Found in Branch**

Development

**Desktop/Device**
* Device: PC
* OS: Windows 10
* CPU: AMD Ryzen 5950x
* GPU: NVIDIA RTX 3070
* Memory: 64 GB

**Additional context**

Looks like it is related to `Visibility` attribute for Row.

forkercat:
Looks like it can be fixed by changing a condition in LegacyReflectionBridge. Will verify and put up PR soon.

PR #15082, opened 2023-03-13
Change the visibility of root node for DPE editors

forkercat:
## What does this PR do?

This PR changes the visibility of class name for DPE editors. Specifically, it changes the visibility for the root node in stack in LegacyReflectionBridge. For DPE, node visibility is defaulted to `Show` and we pick what to hide. So for the root node, we need to set to `ShowChildrenOnly`.

Fixes https://github.com/o3de/o3de/issues/14865

Before & After:

<div>

</div>

In debug view, now you won't see the root row. So for DPE Entity Inspector, we then need to infer the component type based on the actual row values.

## How was this PR tested?

Tested in editor and standalone app. TestContainer row is now not present.

### DIFF
--- a/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/Reflection/LegacyReflectionBridge.cpp
+++ b/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/Reflection/LegacyReflectionBridge.cpp
@@ -173,13 +173,24 @@ namespace AZ::Reflection
             using HandlerCallback = AZStd::function<bool()>;
             AZStd::unordered_map<AZ::TypeId, HandlerCallback> m_handlers;
 
+            // Specify whether the visit starts from the root of the instance.
+            bool m_visitFromRoot = true;
+
             virtual ~InstanceVisitor() = default;
 
-            InstanceVisitor(IReadWrite* visitor, void* instance, const AZ::TypeId& typeId, SerializeContext* serializeContext)
+            InstanceVisitor(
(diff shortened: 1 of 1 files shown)


Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/Reflection/LegacyReflectionBridge.cpp
```diff
@@ -173,13 +173,24 @@ namespace AZ::Reflection
...
             virtual ~InstanceVisitor() = default;
 
-            InstanceVisitor(IReadWrite* visitor, void* instance, const AZ::TypeId& typeId, SerializeContext* serializeContext)
+            InstanceVisitor(
+                IReadWrite* visitor, void* instance,
+                const AZ::TypeId& typeId,
+                SerializeContext* serializeContext,
+                bool visitFromRoot = true)
                 : m_visitor(visitor)
                 , m_serializeContext(serializeContext)
```
forkercat, line 187: Exposing this flag such that caller can change it. For example, we may want to visit data like this, where it is not visiting from root.

```cpp
Reflection::VisitLegacyInMemoryInstance(
    m_impl.get(), boundContainer->m_containerInstance, boundContainer->m_containerTypeId);
```

alexmontAmazon (approved):

dtamkin1 (approved):

lumberyard-employee-dm (approved):

PR #15084, opened 2023-03-13
Auto-pass the NullRenderer flag to server when -rhi=null

mbalfour-amzn:
## What does this PR do?

When the Editor auto-launches a server, the rhi command-line value is set by a cvar. If that cvar is set to "-rhi=null", we should probably also pass "-NullRenderer" to avoid running the higher-level rendering code as well.

There's an argument for making that another cvar instead, if anyone feels strongly about it on this PR i'll change it. It _seems_ like overkill to me, but if there's a good use case we can go for it.

## How was this PR tested?

Ran the Editor with MPS and verified that the additional flag was getting passed when the cvar was set to "null".


lmbr-pip (approved):
Awesome

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
```diff
@@ -304,11 +304,15 @@ namespace Multiplayer
             server_rhi = static_cast<AZ::CVarFixedString>(editorsv_rhi_override);
         }
 
+        // If the null rhi is requested, then also automatically set the -NullRenderer flag.
+        const char* nullRendererFlag = (server_rhi == AZ::Name("null")) ? "-NullRenderer" : "";
+
         processLaunchInfo.m_commandlineParameters = AZStd::string::format(
-            R"("%s" --project-path "%s" --editorsv_isDedicated true --bg_ConnectToAssetProcessor false --rhi "%s" --editorsv_port %i --bg_enableNetworkingMetrics %i --sv_dedicated_host_onstartup false)",
+            R"("%s" --project-path "%s" --editorsv_isDedicated true --bg_ConnectToAssetProcessor false --rhi "%s" %s --editorsv_port %i --bg_enableNetworkingMetrics %i --sv_dedicated_host_onstartup false)",
```
lumberyard-employee-dm, line 311: Honestly this would be simpler if the vector arguments are used instead of string.
the `m_commandlineParameters` container is a [variant](https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/AzFramework/Process/ProcessWatcher.h#L33-L41) that contains a vector where the arguments can be added individually

lumberyard-employee-dm (approved):

lumberyard-employee-dm (approved):

mbalfour-amzn:
Closing on development, will reopen on stabilization

PR #15085, opened 2023-03-13
Prefab Override | Enable override in Entity Outliner by default

forkercat:
## What does this PR do?

This PR enables the new prefab override feature in Entity Outliner by default.

Once enabled, prefab containers can be manually expanded or collapsed. Any override edits would be stored in the _prefab that is being edited_ (highlighted with blue capsule). The level prefab is in edit mode by default.

(Blue "o" icon indicates there is an override edit; Blue "+" icon indicates the entity is added as an override)

See more in [prefab override user guide](https://development--o3deorg.netlify.app/docs/learning-guide/tutorials/entities-and-prefabs/override-a-prefab/).

See RFC: https://github.com/o3de/sig-content/issues/75

See the [known issue list](https://github.com/o3de/o3de/issues?q=is%3Aopen+is%3Aissue+label%3Afeature%2Fprefabs+%22prefab+overrides%22) for prefab override.

### How to turn off this feature?

If it causes any issue to any workflow, you can manually turn this feature off by setting this flag in `<project_root>/user/Registry/editorpreferences.setreg`:

```json
{
    "O3DE": {
        "Preferences": {
            "Prefabs": {
                "EnableOutlinerOverrideManagement": false
            }
        }
    }
}
```

## How was this PR tested?

- [x] Functionality has been manually tested in editor.
- [x] We have unit tests and automated tests that cover the supported operations.
- [x] Validated the flag is on by default in AutomatedTesting and any new project if users create.
- [x] Verified other project repos do not have existing dependencies on the flag, so we don't need to update those flags manually.

### DIFF
--- a/Registry/prefab.editor.setreg
+++ b/Registry/prefab.editor.setreg
@@ -17,5 +17,12 @@
                 }
             }
         }
+    },
+    "O3DE": {
+        "Preferences": {
+            "Prefabs": {
+                "EnableOutlinerOverrideManagement": true
+            }
+        }
     }
 }
--- a/AutomatedTesting/Registry/editorpreferences.setreg
+++ b/AutomatedTesting/Registry/editorpreferences.setreg
@@ -7,7 +7,7 @@
     "O3DE": {
         "Preferences": {
             "Prefabs": {
-                "EnableOutlinerOverrideManagement": false,
+                "EnableOutlinerOverrideManagement": true,
                 "EnableInspectorOverrideManagement": false
             }
         }


Registry/prefab.editor.setreg
```diff
@@ -17,5 +17,12 @@
                 }
             }
         }
+    },
+    "O3DE": {
+        "Preferences": {
+            "Prefabs": {
+                "EnableOutlinerOverrideManagement": true
+            }
```
forkercat, line 25: By adding the default value here, the key/value would be present in `<project_root>/user/Registry/editorpreferences.setreg`, so users can change `true` to `false` if they want to disable the feature.

lumberyard-employee-dm: Instead of plaching this setting the engine `Registry/prefab.editor.setreg` it could be placed in the `*.setreg` that is part of the [DefaultProject](https://github.com/o3de/o3de/tree/development/Templates/DefaultProject/Template/Registry) Template and [MinimalProject](https://github.com/o3de/o3de/tree/development/Templates/MinimalProject/Template/Registry)

lumberyard-employee-dm (approved):
Either way the changes are good to go even without being in the Project Templates.

Registry/prefab.editor.setreg
forkercat: GP. That is also a place to consider. Another alternative is default the flag to `true` in code like we did for Action Manager and Prefab System.

https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/AzFramework/Application/Application.cpp#L759

Will see what others think about these options.

srikappa-amzn: Regarding only storing it in DefaultProject and MinimalProject, 

1. Would it be possible to create a new project without using these 2 templates? If so, the value will be false by default
2. If we create a new template in the future like 'DefaultStarterGameProject' , then the creator of the template would need to remember and know that they need to enable the flag. Relying on project creators to know and remember this would be asking a lot imo.
3. Also, existing projects will still see the old 'non-override' mode. Ideally, we would want the 'override' mode to be the default for all projects and want users to switch to non-override mode if they run into issues (this can't be forever though) 

Considering the above reasons, I think it's better to either have a global setreg that flips it to true like shown here or change it in the code.

lumberyard-employee-dm: For number 1 Creating a project without using either of those templates is an unsupported workflow. Someone could copy and paste an existing project and rename all the content within it to make a "new" project, but is off the beaten path

For number 2, we create project templates from existing projects. So if the user created a `DefaultStarterGameProject` template they would do so from a project created from the Default Project template.

For number 3, the default value in code should be `true`, so the need to have a engine setreg file or even a project setreg is unneeded.

forkercat: > For number 1 Creating a project without using either of those templates is an unsupported workflow. Someone could copy and paste an existing project and rename all the content within it to make a "new" project, but is off the beaten path.

I think we don't need to change the flag in the project template. If we do that, if I understand right we need to manually do the same every time a new project template is published (e.g. MultiplayerProject, FPSProject, etc), unless all other templates can inherit data from those two basic templates.

> For number 3, the default value in code should be true, so the need to have a engine setreg file or even a project setreg is unneeded.

Can we do both (code + global setreg file)?

forkercat: Due to time constraint, I need to get the PR in before stabilization cut on 3/15. I would go with the current version (global setreg file). Let me know if you still have any concern @srikappa-amzn, or I'll try to merge it before the time on 3/15.

srikappa-amzn (approved):

Registry/prefab.editor.setreg
srikappa-amzn: I'm fine with this going as is. The code change won't be needed if we are doing this.

PR #15095, opened 2023-03-14
[Spawnable] Simplify inheritance chain of `EntityAliasVisitor`

burelc-amzn:
The classes `EntityAliasVisitor` and `EntityAliasConstVisitor` both inherited from `EntityAliasVisitorBase`, which defines a number of protected methods, each taking a `const EntityAliasList*`. Then, the individual derived classes `EntityAliasVisitor` and `EntityAliasConstVisitor` both store a `EntityAliasList*`, the only difference being that the one stored in `EntityAliasConstVisitor` is `const`. This design was chosen to have single implementations of the methods declared in the base class, that work for both the `const` and non-`const` member of the derived classes.

However, this design can be simplified. All of the base class's methods are `protected`, only called by public methods of the derived classes. If the same code works for both `const` and non-`const` members, then it can be defined once in the `EntityAliasConstVisitor` class, and inherited by the non-const `Visitor`. That reduces 3 separate functions down to 1.

This change moves the `EntityAliasConstVisitor` to serve the role that the `EntityAliasVisitorBase` did previously. It defines `const Spawnable&` and `const EntityAliasList*` members. All the methods in the const visitor class are `const` already, so they do not need to be adjusted.

The non-`const` Visitor gains a private `GetEntityAliasList()` method, that uses a `const_cast`, which always requires some further explanation. The `EntityAliasVisitor()` constructor accepts a non-`const` `EntityAliasList*`, so the caller has to already have a mutable type. It is merely stored in the base type's `const` member. Because a non-const variable must be used to construct the non-const Visitor, it is always allowed to read that member in a non-const way.

### DIFF
--- a/Code/Framework/AzFramework/AzFramework/Spawnable/Spawnable.cpp
+++ b/Code/Framework/AzFramework/AzFramework/Spawnable/Spawnable.cpp
@@ -27,113 +27,12 @@ namespace AzFramework
             m_sourceIndex < other.m_sourceIndex;
     }
 
-
-    //
-    // EntityAliasVisitorBase
-    //
-
-    bool Spawnable::EntityAliasVisitorBase::IsValid(const EntityAliasList* aliases) const
-    {
-        return aliases != nullptr;
-    }
-
-    bool Spawnable::EntityAliasVisitorBase::HasAliases(const EntityAliasList* aliases) const
-    {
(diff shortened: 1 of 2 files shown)


AMZN-Gene:
+1 red code

AMZN-Gene (approved):

dtamkin1 (approved):

PR #15110, opened 2023-03-14
Fixing cached shadows not clearing properly for their initial render.

invertednormal:
## What does this PR do?

Cached shadows weren't correctly handling the case of multithreaded draw submits with regards to their first render when cached shadows were initially turned on. Depending on the order things were called, it was possible for `m_forceRenderNextFrame` to be set to false before the clear draw packet was actually submitted since `SubmitDrawItems()` can be called multiple times per frame on different threads. This fixes it so `m_forceRenderNextFrame` isn't set to false until `FrameEndInternal()` which will be called only once per frame.

## How was this PR tested?

Tested in warehouse scene with 108 light views and a fair number of draws.

### DIFF
--- a/Gems/Atom/Feature/Common/Code/Source/CoreLights/ShadowmapPass.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/CoreLights/ShadowmapPass.cpp
@@ -190,7 +190,6 @@ namespace AZ
                     }
                 }
             }
-            m_forceRenderNextFrame = false;
 
             if (m_clearShadowDrawPacket)
             {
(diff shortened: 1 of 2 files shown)


rgba16f (approved):

Gems/Atom/Feature/Common/Code/Source/CoreLights/ShadowmapPass.cpp
```diff
@@ -212,5 +211,16 @@ namespace AZ
...
 
+        void ShadowmapPass::FrameEndInternal()
+        {
+            if (m_clearShadowDrawPacket)
+            {
+                // If m_clearShadowDrawPacket is valid, then this pass would have rendered this frame if
+                // m_forceRenderNextFrame was set to true, so set it to false now. This can't be done safely
+                // in SubmitDrawItems because it may be called multiple times from different threads.
+                m_forceRenderNextFrame = false;
+            }
```
VickyAtAZ, line 222: You may want to add `Base::FrameEndInternal();`

amzn-tommy (approved):

VickyAtAZ (approved):

pawelbudziszewski (approved):
It works as it should, cached shadows are back :)
Tested using Warehouse Assets Gem

PR #15113, opened 2023-03-15
Added TypeIds header file which contains string literals of Uuids to the Project and Gem Templates

lumberyard-employee-dm:
Added an AZ_COMPONENT_BASE_DECL and AZ_COMPONENT_BASE_IMPL macros to allow AZ::Component derived classes to define their `CreateDescriptor` functions in a cpp file.

Updated the Project and Gem templates to move definitions of TypeInfo, RTTI, Component Descriptors and Allocator methods to cpp files.

## How was this PR tested?

Instantiated the Gem and Project Templates and built their specific targets using `cmake --build <build-dir> --target`


Templates/CppToolGem/Template/Code/Source/Clients/${Name}SystemComponent.h
```diff
@@ -22,7 +22,11 @@ namespace ${SanitizedCppName}
         , public AZ::TickBus::Handler
     {
     public:
-        AZ_COMPONENT(${SanitizedCppName}SystemComponent, "${SysCompClassId}");
+        AZ_TYPE_INFO_WITH_NAME_DECL(${SanitizedCppName}SystemComponent)
+        AZ_RTTI_NO_TYPE_INFO_DECL();
+        AZ_COMPONENT_INTRUSIVE_DESCRIPTOR_TYPE(${SanitizedCppName}SystemComponent);
+        AZ_COMPONENT_BASE_DECL();
+        AZ_CLASS_ALLOCATOR_DECL
```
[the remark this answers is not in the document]
lumberyard-employee-dm: That sounds good to add as well.
Now that we have all the individual macros decomposed into DECL and IMPL variants, we can have a combined version for the DECL variants and IMPL variants.

Templates/CppToolGem/Template/Code/Source/Clients/${Name}SystemComponent.cpp
```diff
@@ -10,27 +10,25 @@
...
-#include <AzCore/Serialization/EditContext.h>
-#include <AzCore/Serialization/EditContextConstants.inl>
 
 namespace ${SanitizedCppName}
 {
+    AZ_TYPE_INFO_WITH_NAME_IMPL(${SanitizedCppName}SystemComponent, "${SanitizedCppName}SystemComponent",
+        ${SanitizedCppName}SystemComponentTypeId);
+    AZ_RTTI_NO_TYPE_INFO_IMPL(${SanitizedCppName}SystemComponent, AZ::Component);
+    AZ_COMPONENT_BASE_IMPL(${SanitizedCppName}SystemComponent);
+    AZ_CLASS_ALLOCATOR_IMPL(${SanitizedCppName}SystemComponent, AZ::SystemAllocator);
```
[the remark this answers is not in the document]
lumberyard-employee-dm: I'll add that macro using the combined IMPL macros for TypeInfo, RTTI, Component Base and Allocator

Templates/CppToolGem/Template/Code/Include/${Name}/${Name}TypeIds.h
```diff
@@ -0,0 +1,28 @@
...
+
+    // Module derived classes TypeIds
+    inline constexpr const char* ${SanitizedCppName}ModuleInterfaceTypeId = "{${Random_Uuid}}";
+    inline constexpr const char* ${SanitizedCppName}ModuleTypeId = "${ModuleClassId}";
+    // The Editor Module by default is mutually exclusive with the Client Module
+    // so they use the Same TypeId
+    inline constexpr const char* ${SanitizedCppName}EditorModuleTypeId = ${SanitizedCppName}ModuleTypeId;
+
+    // Interface TypeIds
+    inline constexpr const char* ${SanitizedCppName}RequestsTypeId = "{${Random_Uuid}}";
```
AMZN-alexpete, line 27: When the user adds another system component or anything with a TypeId, do they need to add it to this file if they want to use the new _DECL/_IMPL macros?  If so, it might be good to add a comment somewhere in here `Add the typeIds for your components and classes here`  Though I do miss the simplicity of having the type Id defined in the header with the DECL.

Templates/PythonToolGem/Template/Code/Source/${Name}ModuleInterface.cpp
```diff
@@ -0,0 +1,32 @@
...
+    AZ_RTTI_NO_TYPE_INFO_IMPL(${SanitizedCppName}ModuleInterface, AZ::Module);
+    AZ_CLASS_ALLOCATOR_IMPL(${SanitizedCppName}ModuleInterface, AZ::SystemAllocator);
+
+    ${SanitizedCppName}ModuleInterface::${SanitizedCppName}ModuleInterface()
+    {
+    }
+
+    AZ::ComponentTypeList ${SanitizedCppName}ModuleInterface::GetRequiredSystemComponents() const
+    {
+        return AZ::ComponentTypeList{
```
AMZN-alexpete, line 29: Might be nice to have a comment here stating `This template has no c++ system component, but if you create one, add the type here so it is automatically activated`

lumberyard-employee-dm: This does have a System Component.
It is added in the derived [EditorModule](https://github.com/o3de/o3de/blob/36b98f01825e0bffdbf343d2d54eb36f64c2ff92/Templates/PythonToolGem/Template/Code/Source/Tools/%24%7BName%7DEditorModule.cpp#L47-L56).

Templates/UnifiedMultiplayerGem/Template/gem.json
```diff
@@ -21,5 +22,9 @@
     "documentation_url": "${DocumentationURL}",
     "dependencies": [
```
AMZN-alexpete, line 23: I think this gem should declare a dependency on Multiplayer because it is used in the CMakeList.txt
```suggestion
    "dependencies": [
        "Multiplayer"
```

Templates/CppToolGem/Template/Code/Include/${Name}/${Name}TypeIds.h
lumberyard-employee-dm: Well the primary purposes of the TypeId header is to avoid the need to bring in header files with declarations of classes especially component classes.

The ComponentApplication uses a registration system that allows modules to register their components with it as a centralized location for component creation.
Any other module or system can then create these components without needing to know include any headers containing the definitions for the components.

This not only helps with compile times in other libraries, but also allows the module to better separate it's public API and private implementation.
Gems such as PhysX wouldn't need to expose it's PhysX Rigid Body Component as part of it's public API, but instead only needs to expose a TypeId. This allows greater flexibility for gem to perform actions such as change the C++ identifier name of component, refactor it's implementation, add any amount of includes within headers while only affecting the PhysX gem compile time and not downstream users of the gem, etc....

Also what should be noticed is that the `<GemName>TypeIds.h` contains no include files itself.
Making it cheap to include in other Gems as far as compile time is concerned.

Templates/CppToolGem/Template/Code/Source/Clients/${Name}SystemComponent.h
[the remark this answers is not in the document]
lumberyard-employee-dm: The `AZ_COMPONENT_DECL` macro has been added

Templates/CppToolGem/Template/Code/Source/Clients/${Name}SystemComponent.cpp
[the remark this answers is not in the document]
lumberyard-employee-dm: The `AZ_COMPONENT_IMPL` has been added and is being used within the templates.

Code/Framework/AzCore/AzCore/Component/Component.h
```diff
@@ -317,6 +369,127 @@ namespace AZ
...
+    //! ```
+    //! template<class T, class U, size_t N>
+    //! class TestTemplateGemSystemComponent : public BaseSystemComponent {};
+    //!  //... Later in a cpp file
+    //! AZ_COMPONENT_IMPL_INLINE((TestTemplateGemSystemComponent, AZ_CLASS, AZ_CLASS, AZ_AUTO),
+    //!     "TestTemplateGemSystemComponent", TestTemplateGemSystemComponentTypeId, BaseSystemComponent)
+    //! ```
+    //! The `AZ_CLASS` is a placeholder macro that is used to substitute `class` template parameters
+    //! For non-type template parameters such as `size_t` or `int` the `AZ_AUTO` placeholder can be used
+    #define AZ_COMPONENT_IMPL(_ComponentClassOrTemplate, _DisplayName, _Uuid, ...) \
```
AMZN-alexpete, line 459: I like this, I wonder if for this version most people will use the same display name as the class name?
```suggestion
    #define AZ_COMPONENT_IMPL(_ComponentClassOrTemplate, #_ComponentClassOrTemplate, _Uuid, ...) \
```

Code/Framework/AzCore/AzCore/Component/Component.h
```diff
@@ -317,6 +369,127 @@ namespace AZ
...
+    AZ_COMPONENT_MACRO_CALL(AZ_TYPE_INFO_WITH_NAME_IMPL, \
+        AZ_USE_FIRST_ARG(AZ_UNWRAP(_ComponentClassOrTemplate)), \
+        _DisplayName, \
+        _Uuid \
+        AZ_VA_OPT(AZ_COMMA_SEPARATOR, AZ_SKIP_FIRST_ARG(AZ_UNWRAP(_ComponentClassOrTemplate))) \
+        AZ_SKIP_FIRST_ARG(AZ_UNWRAP(_ComponentClassOrTemplate)) \
+    ) \
```
AMZN-alexpete, line 466: I like this, I wonder if for this version most people will use the same display name as the class name?
```suggestion
    #define AZ_COMPONENT_IMPL(_ComponentClassOrTemplate, _Uuid, ...) \
    AZ_COMPONENT_MACRO_CALL(AZ_TYPE_INFO_WITH_NAME_IMPL, \
        AZ_USE_FIRST_ARG(AZ_UNWRAP(_ComponentClassOrTemplate)), \
        #_ComponentClassOrTemplate, \
        _Uuid \
        AZ_VA_OPT(AZ_COMMA_SEPARATOR, AZ_SKIP_FIRST_ARG(AZ_UNWRAP(_ComponentClassOrTemplate))) \
        AZ_SKIP_FIRST_ARG(AZ_UNWRAP(_ComponentClassOrTemplate)) \
    ) \
```

Templates/CppToolGem/Template/Code/Source/Clients/${Name}SystemComponent.h
```diff
@@ -22,7 +22,7 @@ namespace ${SanitizedCppName}
         , public AZ::TickBus::Handler
     {
     public:
-        AZ_COMPONENT(${SanitizedCppName}SystemComponent, "${SysCompClassId}");
+        AZ_COMPONENT_DECL(${SanitizedCppName}SystemComponent);
```
AMZN-alexpete, line 25: nice!

AMZN-alexpete (approved):
Just some minor comments.

Code/Framework/AzCore/AzCore/RTTI/TypeInfoSimple.h
```diff
@@ -602,14 +602,20 @@ namespace AZ
 // This pairs with the AZ_TYPE_INFO_WITH_NAME_IMPL/AZ_TYPE_INFO_WITH_NAME_IMPL_INLINE where an implemenation can be provided
 // in a translation unit(.cpp) or a an inline(.inl) file in order to help reduce compile times
 // https://godbolt.org/z/EGPvKr7xM
-#define AZ_TYPE_INFO_WITH_NAME_DECL(_ClassName) \
+#define AZ_TYPE_INFO_WITH_NAME_DECL_HELPER(_ClassName, _TemplatePlaceholders) \
+    AZ_TYPE_INFO_SIMPLE_TEMPLATE_ID _TemplatePlaceholders \
     friend AZ::TypeNameString GetO3deTypeName(AZ::Adl, AZStd::type_identity<_ClassName>); \
+    AZ_TYPE_INFO_SIMPLE_TEMPLATE_ID _TemplatePlaceholders \
```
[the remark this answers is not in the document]
lumberyard-employee-dm: Yes, The second `AZ_TYPE_INFO_SIMPLE_TEMPLATE_ID _TemplatePlaceholder` is for the friend function on the line below

Code/Framework/AzCore/AzCore/Component/Component.h
lumberyard-employee-dm: For backwards compatibility going forward I think it is better to be explicit with the display name.

For example we have had [several cases](https://github.com/o3de/o3de/commit/cade456db3d683c8245870d35c4abec4da9d2ad2) where a developer changes the identifier of C++ class which effects the name that gets serialized out when using JSON Serialization.

Attempting to load that content fails as the type name used for reflection fails.

By making the display name explicit, a developer can change the C++ class name or even prefix it or remove a namespace from the call to `AZ_COMPONENT_IMPL` and type name registered with the TypeInfo system remains the same allowing Serialization to succeed.

Something that is seemingly innocuous as removing the namespace from in front of fully qualified class name, could break serialization.
For example removing the `PhysX::` from in front of the RagdollComponent identifier used in the RagdollComponent.h `AZ_COMPONENT` macro call would break loading the Ragdoll component in any prefabs that it serialized to it

https://github.com/o3de/o3de/blob/development/Gems/PhysX/Code/Source/PhysXCharacters/Components/RagdollComponent.h#L21-L31
```c++
        AZ_COMPONENT(PhysX::RagdollComponent, "{B89498F8-4718-42FE-A457-A377DD0D61A0}");
```

to
```c++
        AZ_COMPONENT(RagdollComponent, "{B89498F8-4718-42FE-A457-A377DD0D61A0}");
```

PR #15122, opened 2023-03-15
Thumbnailview icon and scrollbar fixes

### LINKED ISSUE #14422: Asset Browser: Majority of Thumbnail View icons are small/pixelated
**Describe the bug**
The vast majority of Thumbnail View icons are default file type icons (e.g. Lua Script/Folders/Script Canvas, etc.) that are small and pixelated when displayed in their thumbnail box in the Asset Browser. The only asset types that display anything other than a default icon are textures which display a preview of the texture itself.

**Steps to reproduce**
```
```
2. Open the Asset Browser, and switch to Thumbnail View if necessary.
3. Click through a few different directories, noting the appearance of the thumbnail icons for various asset types.

**Expected behavior**
Asset icons are clear and attractive.

**Actual behavior**
The majority of asset icons appear they are too small to be displayed in the thumbnail view, and appear pixelated in the Asset Browser.

**Screenshots/Video**
image

**Found in Branch**
development

1fdc3d4c1c2bbc38b729733c1df53bb8e15cb231

dtamkin1:
## What does this PR do?

- Updates scrollbar speed to once per item
- Enlarges icons and icon resolution
- Adds a couple missing icons

image

Fixes #14422 

## How was this PR tested?

Locally


Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/AssetBrowserThumbnailViewProxyModel.cpp
```diff
@@ -38,18 +46,93 @@ namespace AzToolsFramework
...
+
+                            ThumbnailerRequestBus::BroadcastResult(
+                                thumbnail, &ThumbnailerRequests::GetThumbnail, assetBrowserEntry->GetThumbnailKey());
+                            AZ_Assert(thumbnail, "The shared thumbnail was not available from the ThumbnailerRequestBus.");
+                            if (thumbnail && thumbnail->GetState() != Thumbnail::State::Failed)
+                            {
+                                return thumbnail->GetPixmap();
+                            }
+                        }
                     }
```
gadams3, line 66: Why does this need to enumerate the factories if it doesn't directly do anything with them? Couldn't it just be this?'

                            SharedThumbnail thumbnail;

                            ThumbnailerRequestBus::BroadcastResult(
                                thumbnail, &ThumbnailerRequests::GetThumbnail, assetBrowserEntry->GetThumbnailKey());
                            AZ_Assert(thumbnail, "The shared thumbnail was not available from the ThumbnailerRequestBus.");
                            if (thumbnail && thumbnail->GetState() != Thumbnail::State::Failed)
                            {
                                return thumbnail->GetPixmap();
                            }

lsemp3d (approved):

Code/Framework/AzQtComponents/AzQtComponents/Components/Widgets/AssetFolderThumbnailView.cpp
```diff
@@ -334,6 +344,7 @@ namespace AzQtComponents
         config.topItemsHorizontalSpacing = 18;
         config.topItemsVerticalSpacing = 18;
         config.childrenItemsHorizontalSpacing = 7;
+        config.scrollSpeed = 45;
```
spham-amzn, line 347: `45` is an arbitrary magic number, maybe make a constant here with a name describing what `45` means

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/AssetBrowserThumbnailViewProxyModel.cpp
dtamkin1: So I ran into an issue where if I simply tried to get all thumbnails, the resulting icons were tiny. Scaling them up resulted in extremely blurry icons, and if I smoothed them then they would be incredibly blurry. However, this is only an issue for entries that do not have a previewer. Therefore, I'm enumerating the factories as a way to check for entries with a previewer, because we can just return the thumbnail for that set of entries.

The code below this part digs through the directories and returns a path to each icon file so that we can create an appropriately sized icon for that entry. I'm not sure how else to differentiate between these two sets of entries, so I'm open to any suggestions :)

Code/Framework/AzQtComponents/AzQtComponents/Components/Widgets/AssetFolderThumbnailView.cpp
```diff
@@ -156,11 +156,20 @@ namespace AzQtComponents
...
-        if (!iconVariant.isNull())
+        const auto& qVariant = index.data(Qt::DecorationRole);
+        if (!qVariant.isNull())
         {
-            const auto& icon = iconVariant.value<QIcon>();
-            icon.paint(painter, imageRect);
+            QIcon icon;
+            if (const auto& path = qVariant.value<QString>(); !path.isEmpty())
+            {
+                icon.addFile(path, rect.size(), QIcon::Normal, QIcon::Off);
```
spham-amzn, line 165: Will the file path be guaranteed to exist? If not, we may want to check file existence as well

spham-amzn (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/AssetBrowserThumbnailViewProxyModel.cpp
gadams3: OK. Since you're not using the factories, could you just check if the container is empty instead of looping over them? Very supportive of any optimization, simplification, improvement to the thumbnail system. Work still needs to be done to fix all of the loading animations from getting stuck or occupying all of the available threads. Separate issue though.

gadams3 (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/AssetBrowserThumbnailViewProxyModel.cpp
dtamkin1: Unfortunately the container is never empty and always contains two factories, which are nullptrs most of the time. I tried seeing if there was a way I could query the factory for the thumbnail, but I'm not seeing that info stored anywhere 😞

Code/Framework/AzQtComponents/AzQtComponents/Components/Widgets/AssetFolderThumbnailView.cpp
dtamkin1: If the path doesn't exist then no icon will be painted on the thumbnail, maybe we could print a warning if there is no file?

Code/Framework/AzQtComponents/AzQtComponents/Components/Widgets/AssetFolderThumbnailView.cpp
dtamkin1: Sure I'll add a comment, there are a ton of magic numbers as part of this struct, but I think that's on purpose since it's a custom abstract item layout

Issue #10799, opened 2022-07-21
Atom: Editor can crash when running Python automated tests using the null renderer


jckand-amzn:
**Describe the bug**
Editor can crash when running Python automated tests using the null renderer. The repro rate of this is very low, but can seemingly occur in any test that uses the null renderer, of which the majority do in AutomatedTesting. This may only be reproducible in tests which don't run in -BatchMode but use the null renderer, though this is difficult to reproduce, so hasn't been proven.

```
E   ---------------
E   |  Stacktrace |
E   ---------------
E   Logged at Thu Jul 21 11:26:52 2022
E   FileVersion: 0.0.0.0
E   ProductVersion: 0.0.0.0
E   LogFile: @log@/editor_test.log
E   ProjectDir: D:\GitHub\o3de\AutomatedTesting
E   Executable: Editor.exe (debug: no
E   Exception Code: 0xC0000005
E   Exception Addr: 0x0033:0x00007FFA8AA3991C
E   Exception Module:
E   Exception Description: EXCEPTION_ACCESS_VIOLATION, Attempt to write data to address 0x00000015
E
E   The memory could not be "written"
E
E   Call Stack Trace:
E   23) 00007FFB445D0057 (KERNELBASE) : UnhandledExceptionFilter
E   22) 00007FFB468F53B0 (ntdll) : memset
E   21) 00007FFB468DC766 (ntdll) : _C_specific_handler
E   20) 00007FFB468F229F (ntdll) : _chkstk
E   19) 00007FFB468A1454 (ntdll) : RtlRaiseException
E   18) 00007FFB468F0DCE (ntdll) : KiUserExceptionDispatcher
E   17) D:\GitHub\o3de\Code\Framework\AzCore\AzCore\std\containers\intrusive_set.h (606) : AZStd::intrusive_multiset<AZ::HpAllocator::free_node,AZStd::intrusive_multiset_base_hook<AZ::HpAllocator::free_node>,AZStd::less<void> >::erase
E   16) D:\GitHub\o3de\Code\Framework\AzCore\AzCore\Memory\HphaSchema.cpp (1474) : AZ::HpAllocator::coalesce_block
E   15) D:\GitHub\o3de\Code\Framework\AzCore\AzCore\Memory\HphaSchema.cpp (1978) : AZ::HpAllocator::tree_free
[...]

PR #15128, opened 2023-03-15
Enable output callstack for main thread for error.log

VickyAtAZ:
## What does this PR do?

The current error log only contains the call stack for the thread where the exception happened. Sometime this is not enough and including the main thread's call stack would help to diagnose the crash. 
This PR enables output call stack for main thread if error thread is a job thread. 
And this may help find the root cause for this issue which only happens on jenkins AR runs: https://github.com/o3de/o3de/issues/10799

## How was this PR tested?

Run editor with an intentional nullptr crash which happens in a job thread. Check the error.log file under %project%/user/log/ folder. Noticed that the call stack for main thread is added after the call stack of the crashing job thread.

example error.log file

[error.log](https://github.com/o3de/o3de/files/10984541/error.log)

### DIFF
--- a/Code/Legacy/CrySystem/DebugCallStack.cpp
+++ b/Code/Legacy/CrySystem/DebugCallStack.cpp
@@ -52,6 +52,10 @@ static bool IsFloatingPointException(EXCEPTION_POINTERS* pex);
 extern LONG WINAPI CryEngineExceptionFilterWER(struct _EXCEPTION_POINTERS* pExceptionPointers);
 extern LONG WINAPI CryEngineExceptionFilterMiniDump(struct _EXCEPTION_POINTERS* pExceptionPointers, const char* szDumpPath, MINIDUMP_TYPE mdumpValue);
 
+void MarkThisThreadForDebugging(const char* name);
+void UnmarkThisThreadFromDebugging();
+void UpdateFPExceptionsMaskForThreads();
+
 //=============================================================================
 CONTEXT CaptureCurrentContext()
 {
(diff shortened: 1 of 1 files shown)


Code/Legacy/CrySystem/DebugCallStack.cpp
```diff
@@ -52,6 +52,10 @@ static bool IsFloatingPointException(EXCEPTION_POINTERS* pex);
 extern LONG WINAPI CryEngineExceptionFilterWER(struct _EXCEPTION_POINTERS* pExceptionPointers);
 extern LONG WINAPI CryEngineExceptionFilterMiniDump(struct _EXCEPTION_POINTERS* pExceptionPointers, const char* szDumpPath, MINIDUMP_TYPE mdumpValue);
 
+void MarkThisThreadForDebugging(const char* name);
+void UnmarkThisThreadFromDebugging();
+void UpdateFPExceptionsMaskForThreads();
```
spham-amzn, line 57: You shouldn't need to prototype these 2 functions since theyre not called in this file

moudgils (approved):

Code/Legacy/CrySystem/DebugCallStack.cpp
VickyAtAZ: I intent to group them together since they are related.

invertednormal (approved):

Code/Legacy/CrySystem/DebugCallStack.cpp
lumberyard-employee-dm: Unless they are called they are going to be dead-stripped.

PR #15129, opened 2023-03-15
Fix unused variable error by deleting unnecessary code

spham-amzn:
## What does this PR do?

Fixes a Linux/Clang compile error:
```
In file included from /home/github/o3de/build/linux_aarch64/External/RemoteTools-bd846853/Code/CMakeFiles/RemoteTools.Private.Static.dir/Unity/unity_0_cxx.cxx:3:
/home/github/o3de/Gems/RemoteTools/Code/Source/RemoteToolsSystemComponent.cpp:568:14: error: variable 'hasActiveConnection' set but not used [-Werror,-Wunused-but-set-variable]
        bool hasActiveConnection = false;
             ^
```

## How was this PR tested?

Built locally on Linux

### DIFF
--- a/Gems/RemoteTools/Code/Source/RemoteToolsSystemComponent.cpp
+++ b/Gems/RemoteTools/Code/Source/RemoteToolsSystemComponent.cpp
@@ -564,14 +564,6 @@ namespace RemoteTools
             m_joinThread->Start();
         }
 
-        // Check if there are any active connections, if not stop the outbox thread
-        [[maybe_unused]] bool hasActiveConnection = false;
-        for (auto registryIt = m_entryRegistry.begin(); registryIt != m_entryRegistry.end(); ++registryIt)
-        {
-            hasActiveConnection =
-                AZ::Interface<AzNetworking::INetworking>::Get()->RetrieveNetworkInterface(registryIt->second.m_name) != nullptr;
-        }
-
     }
 
 } // namespace RemoteTools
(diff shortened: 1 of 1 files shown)


gadams3 (approved):

Gems/RemoteTools/Code/Source/RemoteToolsSystemComponent.cpp
```diff
@@ -565,7 +565,7 @@ namespace RemoteTools
         }
 
         // Check if there are any active connections, if not stop the outbox thread
-        bool hasActiveConnection = false;
+        [[maybe_unused]] bool hasActiveConnection = false;
```
spham-amzn, line 568: This looks like temporary testing code to me, nothing is done with this boolean. Can it be removed completely?

spham-amzn: Looks like the following block was removed after this check for active connection:
```
        if (!hasActiveConnection)
        {
            m_outboxThread->Stop();
            m_outboxThread->Join();
        }
```
It looks like we don't need this check at all

lumberyard-employee-dm: This code can be removed.

hasActiveConnection is written to below on line 571, but nothing actually reads it at all.

lumberyard-employee-dm (approved):

Gems/RemoteTools/Code/Source/RemoteToolsSystemComponent.cpp
spham-amzn: Removed

PR #15146, opened 2023-03-16
Add operation name in undo/redo console log

forkercat:
## What does this PR do?

This PR adds the operation names in undo/redo logs in console.

Before:

After:

## How was this PR tested?

Tested in editor

### DIFF
--- a/Code/Editor/Undo/Undo.cpp
+++ b/Code/Editor/Undo/Undo.cpp
@@ -324,6 +324,13 @@ void CUndoManager::Redo(int numSteps)
             m_bRedoing = true;
             CUndoStep* redo = m_redoStack.back();
             redo->Redo();
+
+            GetIEditor()->GetLogFile()->FormatLine(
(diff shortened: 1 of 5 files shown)


Code/Editor/Undo/Undo.cpp
```diff
@@ -335,7 +342,6 @@ void CUndoManager::Redo(int numSteps)
     {
         GetIEditor()->UpdateViews(eUpdateObjects);
     }
-    GetIEditor()->GetLogFile()->FormatLine("Redo (Undo:%d,Redo:%d)", m_undoStack.size(), m_redoStack.size());
     GetIEditor()->GetObjectManager()->InvalidateVisibleList();
```
forkercat, line 345: Move the log to the above while loop. If `numSteps` is N (>1), it should output for N times.
Also, by putting it here, no assertion is needed for to check the stack size before using node in it.

srikappa-amzn (approved):
Good addition! This should make debugging undo/redo operations easier.

gadams3 (approved):

PR #15148, opened 2023-03-16
Asset Processor - Process metadata types with same rel path

amzn-mike:
## What does this PR do?

Update asset processor to handle multiple metadata enabled assets to have the same relative path.  AP will now process all metadata enabled files instead of just the "winner".  When rcjob is copying files from the working directory into the cache, it will append the scanfolder ID to all products.  When APM is finishing up with an asset and writing things to the database, it will do a final check to remove the prepended scanfolder ID from the products of the asset with the highest priority for backwards compatibility with systems expecting the original filename.  All other, lower priority products from assets with the same relative path will keep the prepended scanfolder ID.

## How was this PR tested?

Tested manually with multiple files with the same name in different scanfolders, opened editor and verified all the products existed and could be referenced/renamed.

Added and ran unit tests.

### DIFF
--- a/Code/Tools/AssetProcessor/native/utilities/ProductOutputUtil.cpp
+++ b/Code/Tools/AssetProcessor/native/utilities/ProductOutputUtil.cpp
@@ -1,0 +1,147 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#include <native/utilities/ProductOutputUtil.h>
+#include <QString>
+#include <AzCore/std/containers/vector.h>
+#include <AzFramework/IO/FileOperations.h>
+#include <native/utilities/UuidManager.h>
+
+namespace AssetProcessor
+{
+    AZStd::string ProductOutputUtil::GetPrefix(AZ::s64 scanfolderId)
+    {
+        return AZStd::string::format("(%" PRIu64 ")", scanfolderId);
+    }
+
+    void ProductOutputUtil::ModifyProductPath(QString& outputFilename, AZ::s64 sourceScanfolderId)
+    {
+        auto prefix = GetPrefix(sourceScanfolderId);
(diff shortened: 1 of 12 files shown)


AMZN-alexpete (approved):

Code/Tools/AssetProcessor/native/utilities/ProductOutputUtil.cpp
```diff
@@ -0,0 +1,147 @@
...
+#include <QString>
+#include <AzCore/std/containers/vector.h>
+#include <AzFramework/IO/FileOperations.h>
+#include <native/utilities/UuidManager.h>
+
+namespace AssetProcessor
+{
+    AZStd::string ProductOutputUtil::GetPrefix(AZ::s64 scanfolderId)
+    {
+        return AZStd::string::format("(%" PRIu64 ")", scanfolderId);
```
AMZN-stankowi, line 19: Nitpick: This might cause compile warnings (and then errors) on Linux, you're implicitly converting from a signed value to an unsigned string format value.

Code/Tools/AssetProcessor/native/utilities/ProductOutputUtil.cpp
```diff
@@ -0,0 +1,147 @@
...
+                    bool result = AssetUtilities::MoveFileWithTimeout(oldAbsolutePath.c_str(), newAbsolutePath.c_str(), 1);
+
+                    if (!result)
+                    {
+                        AZ_Error(
+                            "ProductOutputUtil",
+                            false,
+                            "Failed to move product from " AZ_STRING_FORMAT " to " AZ_STRING_FORMAT,
+                            AZ_STRING_ARG(oldAbsolutePath),
+                            AZ_STRING_ARG(newAbsolutePath));
```
AMZN-stankowi, line 108: This call doesn't store data on the specific failure anywhere you can look up, but it does print out log messages that can help track why. It would probably be worth updating the error message below to have at the end "See previous log messages for details on why the move failed.", so people skimming or just looking at error output will know to look at the non-error output.

Same with farther down in RenameProduct.

AMZN-stankowi (approved):

Issue #15121, opened 2023-03-15
[GraphModel gem] Asserts reported when running Asset Processor


moraaar:
**Describe the bug**
There are a few asserts being reported when running Asset Processor (see attached Asserts.txt file below)

[Asserts.txt](https://github.com/o3de/o3de/files/10983281/Asserts.txt)

**Steps to reproduce**
1. Build Editor using AutomatedTesting project
2. Run Asset Processor

**Expected behavior**
No asserts reported in the log.

**Actual behavior**
Asserts are reported in the log

**Screenshots/Video**
If applicable, add screenshots and/or a video to help explain your problem.

**Found in Branch**
development

**Commit ID**
7993300824

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Windows
 - Version 10

moraaar:
The issue was introduced in https://github.com/o3de/o3de/pull/14886 (tagging @gadams3) in `GraphModel` gem. More specifically the reflection of `GraphContext::Reflect`, it's reflecting methods that return `AZStd::vector<DataTypePtr>` and `DataTypePtr`, but it's complaining that the serializer doesn't know about those. I saw the following code in `GraphModelSystemComponent::Reflect`

````
        // Reflect core graph classes
        GraphModel::Graph::Reflect(context);
        GraphModel::DataType::Reflect(context);
````

And I thought it was a matter of reflecting `DataType` first (because the assert comes from within `Graph::Reflect`), but that didn't solve the issue because it's reflecting `DataType` and not `DataTypePtr` (which is the one used by `GraphContext::Reflect`).

@gadams3 do you mind taking a further look and think of the appropriate solution for this case? Thanks

PR #15155, opened 2023-03-16
Updating graph model reflection and unbinding function causing AP asserts

gadams3:
## What does this PR do?

This change reorders reflection calls and removes a behavior context binding for a graph model graph controller function that was causing asserts in the AP.

Resolves https://github.com/o3de/o3de/issues/15121

## How was this PR tested?

Confirmed that asserts no longer appear in the AP log after changes.

### DIFF
--- a/Gems/GraphModel/Code/Source/GraphModelSystemComponent.cpp
+++ b/Gems/GraphModel/Code/Source/GraphModelSystemComponent.cpp
@@ -33,22 +33,32 @@ namespace GraphModel
     void GraphModelSystemComponent::Reflect(AZ::ReflectContext* context)
     {
         // Reflect core graph classes
-        GraphModel::Graph::Reflect(context);
-        GraphModel::DataType::Reflect(context);
+        DataType::Reflect(context);
+        GraphContext::Reflect(context);
+        GraphElement::Reflect(context);
+        SlotId::Reflect(context);
+        Slot::Reflect(context);
+        Node::Reflect(context);
+        Connection::Reflect(context);
+        Graph::Reflect(context);
+
+        // Reflect module node data types
+        BaseInputOutputNode::Reflect(context);
+        GraphInputNode::Reflect(context);
+        GraphOutputNode::Reflect(context);
+        ModuleNode::Reflect(context);
+
+        // Reflect data types for integrating graph model with graph canvas
         GraphModelIntegration::GraphCanvasMetadata::Reflect(context);
 
-        // Mime Events for Graph Canvas nodes
+        // Reflect MIME events for graph canvas nodes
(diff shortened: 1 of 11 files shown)


Gems/GraphModel/Code/Source/GraphModelSystemComponent.cpp
```diff
@@ -57,8 +67,8 @@ namespace GraphModel
                 ->Version(0)
                 ;
 
-            serialize->RegisterGenericType<GraphModel::NodePtrList>();
-            serialize->RegisterGenericType<GraphModel::SlotPtrList>();
+            serialize->RegisterGenericType<NodePtrList>();
+            serialize->RegisterGenericType<SlotPtrList>();
```
moraaar, line 71: `SlotPtrList` is already registered in Slot.cpp, I suppose this can be removed.

Gems/GraphModel/Code/Source/GraphModelSystemComponent.cpp
```diff
@@ -57,8 +67,8 @@ namespace GraphModel
                 ->Version(0)
                 ;
 
-            serialize->RegisterGenericType<GraphModel::NodePtrList>();
-            serialize->RegisterGenericType<GraphModel::SlotPtrList>();
+            serialize->RegisterGenericType<NodePtrList>();
```
moraaar, line 70: `NodePtrList` is already registered in Node.cpp, I suppose this can be removed.

Gems/GraphModel/Code/Source/Model/GraphContext.cpp
```diff
@@ -36,7 +38,7 @@ namespace GraphModel
                 ->Attribute(AZ::Script::Attributes::Module, "editor.graph")
                 ->Method("GetSystemName", &GraphContext::GetSystemName)
                 ->Method("GetModuleFileExtension", &GraphContext::GetModuleFileExtension)
-                ->Method("GetAllDataTypes", &GraphContext::GetAllDataTypes)
+                //->Method("GetAllDataTypes", &GraphContext::GetAllDataTypes) // GHI-15121 Unbinding until asserts with AP behavior context reflection are resolved
```
moraaar, line 41: This should be fixed by this PR changes, forgot to uncomment it and remove the comment?

gadams3: This is the only function that is problematic. No other errors or asserts are being reported. I'll comment this out for now, since it's not being used in script or tests yet, and uncomment once there's more time to investigate.

moraaar (approved):

galibzon (approved):

PR #15161, opened 2023-03-17
Reduce ComponentApplication startup time during tests

Kadino:
## What does this PR do?

Adds a StartupParameters option to ComponentApplication to disable writing to the settings registry, and disables it in test fixtures which have no test asserts depending on it. This reduces execution time by around 100ms per test which launches a ComponentApplciation, of which there are thousands of such tests. This reduces parallel test exeuction in AR by around two minutes (721-609 = 112sec, 15%), and reduces serial-only test execution by around five minutes (1993-1679 = 314sec, 15%). If serial-only tests are excluded, the parallelizeable tests duration is reduced by 35% (323-211 = 112sec). The parallel execution impact is relatively large as AZ::AzToolsFramework.Tests.main was the long-pole parallel test module, and still remains the longest by around 15sec.

The primary change is proposed in ComponentApplication.cpp and .h, with all other test files consuming this change.

aside: It does not seem important for all of these tests to integrate with the full application context, but it would also take significant effort to decouple the dependency and more directly verify each tests' targeted logic.

## How was this PR tested?

Executed tests locally multiple times, using a similar command as Jenkins:
`ctest -j6 -C profile --output-on-failure -L "(SUITE_smoke|SUITE_main)" -LE "(REQUIRES_gpu|REQUIRES_tiaf)" -T Test --no-tests=error`

CTest summary from db92736cd9 before this change:
```
100% tests passed, 0 tests failed out of 121

Label Time Summary:
COMPONENT_Editor                = 263.30 sec*proc (1 test)
COMPONENT_MetadataRelocation    =  31.20 sec*proc (1 test)
COMPONENT_Physics               =  34.54 sec*proc (1 test)
COMPONENT_ScriptCanvas          =  33.34 sec*proc (1 test)
COMPONENT_Smoke                 =   2.19 sec*proc (1 test)
COMPONENT_TestTools             =  93.45 sec*proc (6 tests)
FRAMEWORK_googletest            = 1502.71 sec*proc (92 tests)
FRAMEWORK_pytest                = 491.74 sec*proc (28 tests)
SUITE_main                      = 1893.79 sec*proc (100 tests)
SUITE_smoke                     = 110.40 sec*proc (21 tests)

Total Test time (real) = 721.33 sec
```

After the proposed change:
```
100% tests passed, 0 tests failed out of 121

Label Time Summary:
COMPONENT_Editor                = 260.59 sec*proc (1 test)
COMPONENT_MetadataRelocation    =  30.28 sec*proc (1 test)
COMPONENT_Physics               =  32.71 sec*proc (1 test)
COMPONENT_ScriptCanvas          =  33.65 sec*proc (1 test)
COMPONENT_Smoke                 =   2.21 sec*proc (1 test)
COMPONENT_TestTools             =  78.98 sec*proc (6 tests)
FRAMEWORK_googletest            = 1206.29 sec*proc (92 tests)
FRAMEWORK_pytest                = 473.66 sec*proc (28 tests)
SUITE_main                      = 1586.55 sec*proc (100 tests)
SUITE_smoke                     = 103.12 sec*proc (21 tests)

Total Test time (real) = 608.83 sec
```


Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp
```diff
@@ -1589,6 +1598,11 @@ namespace AZ
         return AZ::TimeUsToSeconds(gameTickTime);
     }
 
+    void ComponentApplication::SetSettingsRegistryEnabled(bool enabled = true)
```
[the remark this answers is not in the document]
lumberyard-employee-dm: Default parameters in the function definition would be useless, outside of the ComponentApplication.cpp file, as other files would only have access to the signature in ComponentApplication.h

lumberyard-employee-dm:
I have a suggestion for further flexibility.

How about we add a [constructor overload](https://github.com/o3de/o3de/blob/e3f53517d17c10d8e93fd2c955a49e284a45bee7/Code/Framework/AzCore/AzCore/Component/ComponentApplication.h#L150-L151) for the ComponentApplication that accepts a struct that can configure creation of systems such as the SettingsRegistry(and potentially other systems in the future).

That would shave off a bit more startup time, as the SettingsRegistry initialize merging in the constructor of command line parameters and any `~/.o3de/Regsitry/*.setreg(patch)` files can be skipped.

```c++
struct AppConstructParams
{
    constexpr size_t MaxCommandArgsCount = 128;
    using CommandArgsArray = AZStd::fixed_vector<char*, MaxCommandArgsCount>;
    CommandArgsArray m_commandArgs;
    bool m_settingsRegistryEnabled{ true };
};
ComponentApplication(const AppConstructParams& appConstructParams);
// ...
ComponentApplication::ComponentApplication(const AppConstructParams& appConstructParams)
    : m_timeSystem(AZStd::make_unique<TimeSystem>())
{
   // Use the existing ComponentApplication(int argC, char** argV); overload
   // ...
}

// Delegate to the AppConstructParams overload
ComponentApplication::ComponentApplication(int argC, char** argV)
    : ComponentApplication(AppConstructParams{{argV, argv + argC} })
{
}

File: Code/Framework/AzCore/AzCore/Component/ComponentApplication.h [hunk 2]
```diff
@@ -355,6 +358,7 @@ namespace AZ
         bool                                        m_isStarted{ false };
         IAllocator*                                 m_osAllocator{ nullptr };
         EntitySetType                               m_entities;
+        bool                                        m_isSettingsRegistryEnabled{ true };
```
lumberyard-employee-dm, line 361: The name for this variable containing the verb is seems a off here.

Using a term like "isSettingsRegistryEnabled" is for checking the state that the SettingsRegistry is _already_ enabled.
Seems like the variable should better indicate that this is actually affecting the settings registry enabled state.
i.e `m_settingsRegistryEnabled` or `m_SettingsRegistryActive`

Better yet, the SettingsRegistry isn't a singleton, it is regular class that can be registered with a global `AZ::Interface`.
A better term might be "is this settings registered globally".
Such as `m_settingsRegistryRegistered` or `m_settingsRegistryRegisteredGlobally`

A third the approach is to in the constructor of the ComponentApplication, not create the SettingsRegistry if a constructor parameter is supplied to indicate creation should be skipped.

Kadino: Good catch, plan to use `m_settingsRegistryRegistered` along with trying the AppConstructParams approach above, to see which tests the initialization matters for. It's possible that init is currently depended on, and some tests/fixtures may want to only disable the later calls to merge.

Kadino: AppConstructParams actually seems quite similar to the existing [StartupParameters](https://github.com/o3de/o3de/pull/15161/files#diff-ae0fc2ebd65b47a83c2de4ea98436904a271494c6ff458b1ac6cf2230cdb6e0aR128) so I'll add it there.

Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp
```diff
@@ -552,52 +552,54 @@ namespace AZ
 
     void ComponentApplication::InitializeSettingsRegistry()
     {
-        SettingsRegistryMergeUtils::ParseCommandLine(m_commandLine);
+        if (m_startupParameters.m_loadSettingsRegistry)
```
lumberyard-employee-dm, line 555: Uh... InitializeSettingsRegistry is called in the `constructor` and user provided StartupParameters aren't supplied until `ComponentApplication::Create` is called.

Kadino: Changing to your constructor overload suggestion above.

Kadino: Constructor overloading is proving fairly messy due to it also prompting adding constructors across the broader class hierarchy of Applications, such as EditorToolsApplication and EditorQtApplication. Debating cutting out this attempt to include initialization.

Kadino:
> One question I have is it looks like the default is to have the settings registry enabled, and then in lots of places we're turning it off. Are there many more tests that need it enabled? If not, then perhaps we should flip the default, so it's off normally, and then tests that need it explicitly opt into it? Then potentially fewer files need to be modified. I just wanted to confirm before approving.

The actual production application code wants to use the SettingsRegistry, so I want to avoid modify that default behavior to suit test requirements. The latest diff also minimizes modifying the public interface. It's possible to go further into Application's constructors an initialization, but also even more sprawling of a change.

I didn't see an obvious path to make this change yet modify fewer files. There's thousands of individual tests which depend on the 64 modified test files, which contain fixtures shared by the tests. Most of these fixtures are separate from one another, and it didn't appear appropriate to deduplicate their similarities by imposing a new hierarchy. Plus that would still modify at least as many files. I also want to avoid promoting new variant application specializations "configured for testing" because that pattern encourages avoiding testing what ships, and bloats the codebase.

hultonha (approved):

lumberyard-employee-dm (approved):

PR #15112, opened 2023-03-15
Fixes race condition when debugging Lua files

galibzon:
Closes #14823

## What does this PR do?

The TcpConnection class has a m_sendRingBuffer
that is NOT thread safe.

NetworkSystemComponent uses the Main thread
to send messages, and hence use the m_sendRingBuffer.

On the other hand, before this fix, the RemoteToolsSystemComponent.cpp was sending packets with the help of a second thread controlled by the class RemoteToolsOutboxThread.cpp. This second thread is totally unnecessary and, even worst, was causing race conditions. This race condition was corrupting the messages being sent from Editor.exe to LuaIDE.exe.

## How was this PR tested?

Did long sessions of Lua files testing. For small and large Lua files. No problems at all with this fix.
Without this fix the LuaIDE is mostly unusable.

## Other Notes.
In particular the Race Condition exacerbates when this PR was merged: https://github.com/o3de/o3de/pull/10132
That PR was necessary, but it forces the TcpConnection to flush pending messages in the Main thread, while at the same time the RemoteToolsSystemComponent was also sending messages using the RemoteToolsOutboxThread.

Here is an example of the Race Condition:

**Main Thread**  
```
 	EditorLib.dll!AzNetworking::TcpRingBufferImpl::AdvanceReadBuffer(unsigned int numBytes) Line 54	C++
 	EditorLib.dll!AzNetworking::TcpRingBuffer<1048576>::AdvanceReadBuffer(unsigned int numBytes) Line 48	C++
 	EditorLib.dll!AzNetworking::TcpConnection::UpdateSend() Line 112	C++
 	EditorLib.dll!AzNetworking::TcpNetworkInterface::HandleConnectionSend(AzNetworking::SocketFd socketFd) Line 219	C++
 	EditorLib.dll!AzNetworking::TcpNetworkInterface::Update::__l2::<lambda_2>::operator()(AzNetworking::SocketFd socketFd) Line 115	C++
 	EditorLib.dll!AZStd::Internal::INVOKE<`AzNetworking::TcpNetworkInterface::Update'::`2'::<lambda_2> &,enum AzNetworking::SocketFd>(AzNetworking::TcpNetworkInterface::Update::__l2::<lambda_2> & f, AzNetworking::SocketFd && <args_0>) Line 209	C++
[...]

PR #15163, opened 2023-03-17
Fix build linux/android in release

moraaar:
## What does this PR do?

Fixes Linux and android on release configurations. `TaskGraphEvent::m_label` is only used inside `AZ_Assert` statements, so it was warning as not used in release, causing build to fail.

Marking as well `test_BasicEditorWorkflows_LevelEntityComponentCRUD` as xfail since it's regularly failing affecting the speed to submit bug fixes to stabilization branch. (#15172)

Fixed also the following error introduced in #15112 (tagging @galibzon)
````
/data/workspace/o3de/Gems/RemoteTools/Code/Source/RemoteToolsSystemComponent.cpp:568:14: error: variable 'hasActiveConnection' set but not used [-Werror,-Wunused-but-set-variable]
        bool hasActiveConnection = false;
              ^
````

## How was this PR tested?

Built Linux and android on release successfully.

### DIFF
--- a/Code/Framework/AzCore/AzCore/Task/TaskGraph.h
+++ b/Code/Framework/AzCore/AzCore/Task/TaskGraph.h
@@ -90,7 +90,7 @@ namespace AZ
         AZStd::binary_semaphore m_semaphore;
         AZStd::atomic_int       m_waitCount = 0;
         TaskExecutor*           m_executor = nullptr;
-        const char* m_label;
+        [[maybe_unused]] const char* m_label;
     };
 
     // The TaskGraph encapsulates a set of tasks and their interdependencies. After adding
--- a/Code/Framework/AzCore/AzCore/Task/TaskGraph.h
+++ b/Code/Framework/AzCore/AzCore/Task/TaskGraph.h
@@ -90,7 +90,7 @@ namespace AZ
         AZStd::binary_semaphore m_semaphore;
         AZStd::atomic_int       m_waitCount = 0;
         TaskExecutor*           m_executor = nullptr;
-        [[maybe_unused]] const char* m_label;
+        [[maybe_unused]] const char* m_label = nullptr;
     };
 
     // The TaskGraph encapsulates a set of tasks and their interdependencies. After adding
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by AMZN-daimini — APPROVED** [review]

**Review by hultonha — APPROVED** [review]

File: Code/Framework/AzCore/AzCore/Task/TaskGraph.h [hunk 1]
```diff
@@ -90,7 +90,7 @@ namespace AZ
         AZStd::binary_semaphore m_semaphore;
         AZStd::atomic_int       m_waitCount = 0;
         TaskExecutor*           m_executor = nullptr;
-        const char* m_label;
+        [[maybe_unused]] const char* m_label;
```
hultonha, line 93: suggestion: If AR has started it's not a huge deal but while we're here it might be nice to initialize this to `nullptr` too 🙂

moraaar: I had to run AR again, so I added it :)

Gems/RemoteTools/Code/Source/RemoteToolsSystemComponent.cpp
```diff
@@ -563,15 +563,6 @@ namespace RemoteTools
...
             m_joinThread->Start();
         }
-
-        // Check if there are any active connections, if not stop the outbox thread
-        bool hasActiveConnection = false;
-        for (auto registryIt = m_entryRegistry.begin(); registryIt != m_entryRegistry.end(); ++registryIt)
-        {
-            hasActiveConnection =
-                AZ::Interface<AzNetworking::INetworking>::Get()->RetrieveNetworkInterface(registryIt->second.m_name) != nullptr;
-        }
```
moraaar, line 573: This query is not necessary anymore because before it used to do

````
        if (!hasActiveConnection)
        {
            m_outboxThread->Stop();
            m_outboxThread->Join();
        }
````

But not anymore since this change: https://github.com/o3de/o3de/pull/15112

galibzon (approved):

Issue #14998, opened 2023-03-08
Deleting a Transform component when the "Pick an object in the viewport" mode is active crashes the Editor


LB-TomaszSerwanski:
**Describe the bug**
It is possible to delete an entity's _Transform_ component when the _Pick an object in the viewport_ mode is active. This results in a crash.

Please note that the crash reproduces even if the _Pick an object in the viewport_ was initiated from another component or tool (i.e. _Entity Reference_ component or _Script Canvas_ tool). 

Please refer to the attached videos for the two ways of reproducing the issue and log files for more details.

**Call stack trace**
```
25) 00007FFCFA260207 (KERNELBASE) : UnhandledExceptionFilter
24) 00007FFCF9732486 (InProcessClient64) : (function-name not available)
23) 00007FFCF9732518 (InProcessClient64) : (function-name not available)
22) 00007FFCF97323E1 (InProcessClient64) : (function-name not available)
21) 00007FFCFC515530 (ntdll) : memset
20) 00007FFCFC4FC876 (ntdll) : _C_specific_handler
19) 00007FFCFC51241F (ntdll) : _chkstk
18) 00007FFCFC4C14A4 (ntdll) : RtlRaiseException
17) 00007FFCFC510F4E (ntdll) : KiUserExceptionDispatcher
16) D:\Git\o3de\Code\Framework\AzCore\AzCore\std\containers\vector.h (1095) : AZStd::vector<AZ::EntityId,AZStd::allocator>::insert_iter<AZStd::basic_const_iterator<AZStd::list_iterator<AZ::EntityId> > >
15) D:\Git\o3de\Code\Framework\AzToolsFramework\AzToolsFramework\ViewportSelection\EditorTransformComponentSelection.cpp (321) : AzToolsFramework::EntityIdVectorFromContainer<AZStd::unordered_set<AZ::EntityId,AZStd::hash<AZ::EntityId>,AZStd::equal_to<AZ::Ent
14) D:\Git\o3de\Code\Framework\AzToolsFramework\AzToolsFramework\ViewportSelection\EditorTransformComponentSelection.cpp (2602) : `AzToolsFramework::EditorTransformComponentSelection::OnActionRegistrationHook'::`7'::<lambda_2>::operator()
13) D:\Git\o3de\Code\Framework\AzToolsFramework\AzToolsFramework\ActionManager\Action\EditorAction.cpp (280) : AzToolsFramework::EditorAction::Trigger
[...]

PR #15173, opened 2023-03-17
Action Manager | Prevent entity deletion while in Entity Picker mode

AMZN-daimini:
## What does this PR do?

Fixes https://github.com/o3de/o3de/issues/14998

Disables the `Delete` action when the Entity Picker mode is enabled. This prevents edge cases where users can try and delete the currently selected entity while picking an entity for a property of one of its components, causing a crash.

## How was this PR tested?

Manual testing.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/ViewportSelection/EditorTransformComponentSelection.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/ViewportSelection/EditorTransformComponentSelection.cpp
@@ -21,6 +21,7 @@
 #include <AzToolsFramework/ActionManager/Action/ActionManagerInterface.h>
 #include <AzToolsFramework/ActionManager/Menu/MenuManagerInterface.h>
 #include <AzToolsFramework/ActionManager/HotKey/HotKeyManagerInterface.h>
+#include <AzToolsFramework/API/ViewportEditorModeTrackerInterface.h>
 #include <AzToolsFramework/Commands/EntityManipulatorCommand.h>
 #include <AzToolsFramework/Commands/SelectionCommand.h>
 #include <AzToolsFramework/ComponentMode/ComponentModeSwitcher.h>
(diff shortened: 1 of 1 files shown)


Code/Framework/AzToolsFramework/AzToolsFramework/ViewportSelection/EditorTransformComponentSelection.cpp
```diff
@@ -2651,6 +2652,17 @@ namespace AzToolsFramework
...
+                    // Don't allow users to delete entities while in Entity Picker mode.
+                    if (auto viewportEditorModeTracker = AZ::Interface<AzToolsFramework::ViewportEditorModeTrackerInterface>::Get())
+                    {
+                        auto viewportEditorModes =
+                            viewportEditorModeTracker->GetViewportEditorModes({ AzToolsFramework::GetEntityContextId() });
+                        if (viewportEditorModes->IsModeActive(AzToolsFramework::ViewportEditorMode::Pick))
+                        {
+                            return;
+                        }
+                    }
```
hultonha, line 2664: suggestion: It might be worth pulling this out into a simple local function/lambda with`auto` and then capturing it in the callbacks, e.g.

```
auto isInEditorPickMode = []
{
    // Don't allow users to delete entities while in Entity Picker mode.
    if (auto viewportEditorModeTracker = AZ::Interface<AzToolsFramework::ViewportEditorModeTrackerInterface>::Get())
    {
        auto viewportEditorModes =
            viewportEditorModeTracker->GetViewportEditorModes({ AzToolsFramework::GetEntityContextId() });
        if (viewportEditorModes->IsModeActive(AzToolsFramework::ViewportEditorMode::Pick))
        {
            return true;
        }
    }
    return false;
}
...

[this, isInEditorPickMode]()
{
    AZ_PROFILE_FUNCTION(AzToolsFramework);
    if (isInEditorPickMode())
    {
        return;
    }
    ...
```

hultonha (approved):
LGTM just had one minor suggestion

PR #15175, opened 2023-03-17
Add OnObjectSrgCreated event

CyaNinix:
## What does this PR do?

This PR adds an event to know when a new ObjectSrg is created for a mesh or an actor.

This allows to easily sets custom per-object data using only a few lines of code, usage example at the end.

The main issue is that when you acquire a mesh handle, ObjectSrg have not yet been created, and except waiting for the next tick there's not way to get them.

This PR adds a new MeshComponent event which allows us to be able to know when a new ObjectSrg has been created by the mesh feature processor. This way we can configure it using our already configured data, or query all object SRGs when data changes.

## How was this PR tested?

This PR was tested using a similar simple component as present at the end of this PR, adding the component, instancing a prefab containing a mesh/material and colored components and changing the color at will.

## Usage example

```hlsl
// full shader not included
partial ShaderResourceGroup ObjectSrg : SRG_PerObject
{
    float4 m_objectColor;
}
```

```cpp
#pragma once

#include <AzCore/Component/Component.h>
#include <AtomLyIntegration/CommonFeatures/Mesh/MeshComponentBus.h>
#include <AtomLyIntegration/CommonFeatures/Mesh/MeshHandleStateBus.h>
#include <MyProject/Types.h>
#include <MyProject/Rendering/ColoredMaterialBus.h>

namespace MyProject
{
	class ColoredMaterialComponentBase
		: private ColoredMaterialBus::Handler
		, private AZ::Render::MeshComponentNotificationBus::Handler
		, private AZ::Render::MeshHandleStateNotificationBus::Handler
	{
	public:
		AZ_TYPE_INFO(ColoredMaterialComponentBase, ColoredMaterialComponentBaseGuid);

		static void GetRequiredServices(AZ::ComponentDescriptor::DependencyArrayType& required);

		const AZ::Color& GetColor() const;

		void SetColorIndex(const AZ::Color& color) override;

	protected:
		void Activate(AZ::EntityId entityId);
		void Deactivate();

		/////////////////////////////////////////////////////////////////////////////////
		// MeshComponentNotificationBus overrides
		void OnModelReady(const AZ::Data::Asset<AZ::RPI::ModelAsset>& modelAsset, const AZ::Data::Instance<AZ::RPI::Model>& model) override;
		void OnObjectSrgCreated(const AZ::Data::Instance<AZ::RPI::ShaderResourceGroup>& objectSrg) override;
		/////////////////////////////////////////////////////////////////////////////////

		/////////////////////////////////////////////////////////////////////////////////
		// MeshComponentNotificationBus overrides
		void OnMeshHandleSet(const AZ::Render::MeshFeatureProcessorInterface::MeshHandle* meshHandle) override;
		/////////////////////////////////////////////////////////////////////////////////

		void UpdateObjectSrgs();
		void UpdateObjectSrgs(const AZ::Data::Instance<AZ::RPI::ShaderResourceGroup>* objectSrgs, size_t srgCount);

		AZ::EntityId m_entityId;
		AZ::Color m_color;
		const AZ::Render::MeshFeatureProcessorInterface::MeshHandle* m_meshHandle = nullptr;
	};
}
```

```cpp
#include "ColoredMaterialComponentBase.h"

#include <AzCore/Asset/AssetSerializer.h>
#include <AzCore/RTTI/ReflectContext.h>
#include <AzCore/Serialization/EditContext.h>
#include <AzCore/Serialization/SerializeContext.h>
#include <Atom/RPI.Public/Scene.h>
#include <cmath>

namespace MyProject
{
	void ColoredMaterialComponentBase::GetRequiredServices(AZ::ComponentDescriptor::DependencyArrayType& required)
	{
		required.push_back(AZ_CRC("MaterialProviderService"));
	}

	const AZ::Color& ColoredMaterialComponentBase::GetColorIndex() const
	{
		return m_color;
	}

	void ColoredMaterialComponentBase::SetColorIndex(const AZ::Color& color)
	{
		m_color = color;

		UpdateObjectSrgs();
	}

	void ColoredMaterialComponentBase::Activate(AZ::EntityId entityId)
	{
		m_entityId = entityId;

		ColoredMaterialBus::Handler::BusConnect(entityId);
		AZ::Render::MeshComponentNotificationBus::Handler::BusConnect(entityId);
		AZ::Render::MeshHandleStateNotificationBus::Handler::BusConnect(entityId);
	}

	void ColoredMaterialComponentBase::Deactivate()
	{
		ColoredMaterialBus::Handler::BusDisconnect();
		AZ::Render::MeshComponentNotificationBus::Handler::BusDisconnect();
		AZ::Render::MeshHandleStateNotificationBus::Handler::BusDisconnect();

		m_entityId.SetInvalid();
		m_meshHandle = nullptr;
	}

	void ColoredMaterialComponentBase::OnModelReady(const AZ::Data::Asset<AZ::RPI::ModelAsset>& /*modelAsset*/, const AZ::Data::Instance<AZ::RPI::Model>& /*model*/)
	{
		UpdateObjectSrgs();
	}

	void ColoredMaterialComponentBase::OnObjectSrgCreated(const AZ::Data::Instance<AZ::RPI::ShaderResourceGroup>& objectSrg)
	{
		UpdateObjectSrgs(&objectSrg, 1);
	}
	
	void ColoredMaterialComponentBase::OnMeshHandleSet(const AZ::Render::MeshFeatureProcessorInterface::MeshHandle* meshHandle)
	{
		m_meshHandle = meshHandle;
		UpdateObjectSrgs();
	}

	void ColoredMaterialComponentBase::UpdateObjectSrgs()
	{
		if (!m_entityId.IsValid())
			return;

		if (!m_meshHandle || !m_meshHandle->IsValid())
			return;

		auto* meshFeatureProcessor = AZ::RPI::Scene::GetFeatureProcessorForEntity<AZ::Render::MeshFeatureProcessorInterface>(m_entityId);
		if (!meshFeatureProcessor)
			return;

		const AZStd::vector<AZ::Data::Instance<AZ::RPI::ShaderResourceGroup>>& objectSrgs = meshFeatureProcessor->GetObjectSrgs(*m_meshHandle);
		if (objectSrgs.empty())
			return;

		UpdateObjectSrgs(objectSrgs.data(), objectSrgs.size());
		meshFeatureProcessor->QueueObjectSrgForCompile(*m_meshHandle);
	}
	
	void ColoredMaterialComponentBase::UpdateObjectSrgs(const AZ::Data::Instance<AZ::RPI::ShaderResourceGroup>* objectSrgs, size_t srgCount)
	{
		for (size_t i = 0; i < srgCount; ++i)
		{
			AZ::RHI::ShaderInputConstantIndex colorIndex = objectSrgs[i]->FindShaderInputConstantIndex(AZ::Name("m_objectColor"));
			if (colorIndex.IsValid())
			{
				objectSrgs[i]->SetConstant(colorIndex, m_color);
			}
		}
	}
}
```

### DIFF
--- a/Gems/Atom/Feature/Common/Code/Include/Atom/Feature/Mesh/MeshFeatureProcessor.h
+++ b/Gems/Atom/Feature/Common/Code/Include/Atom/Feature/Mesh/MeshFeatureProcessor.h
@@ -36,8 +36,12 @@ namespace AZ
             friend class MeshLoader;
 
         public:
+            using ObjectSrgCreatedEvent = MeshFeatureProcessorInterface::ObjectSrgCreatedEvent;
+
             const Data::Instance<RPI::Model>& GetModel() { return m_model; }
             const RPI::Cullable& GetCullable() { return m_cullable; }
+
+            ObjectSrgCreatedEvent& GetObjectSrgCreatedEvent() { return m_objectSrgCreatedEvent; }
 
         private:
             class MeshLoader
@@ -113,6 +117,7 @@ namespace AZ
 
             //! List of object SRGs used by meshes in this model 
             AZStd::vector<Data::Instance<RPI::ShaderResourceGroup>> m_objectSrgList;
+            MeshFeatureProcessorInterface::ObjectSrgCreatedEvent m_objectSrgCreatedEvent;
             AZStd::unique_ptr<MeshLoader> m_meshLoader;
             RPI::Scene* m_scene = nullptr;
             RHI::DrawItemSortKey m_sortKey;
@@ -179,6 +184,7 @@ namespace AZ
             void SetCustomMaterials(const MeshHandle& meshHandle, const CustomMaterialMap& materials) override;
             const CustomMaterialMap& GetCustomMaterials(const MeshHandle& meshHandle) const override;
             void ConnectModelChangeEventHandler(const MeshHandle& meshHandle, ModelChangedEvent::Handler& handler) override;
+            void ConnectObjectSrgCreatedEventHandler(const MeshHandle& meshHandle, ObjectSrgCreatedEvent::Handler& handler) override;
 
             void SetTransform(const MeshHandle& meshHandle, const AZ::Transform& transform,
                 const AZ::Vector3& nonUniformScale = AZ::Vector3::CreateOne()) override;
--- a/Gems/Atom/Feature/Common/Code/Include/Atom/Feature/Mesh/MeshFeatureProcessorInterface.h
+++ b/Gems/Atom/Feature/Common/Code/Include/Atom/Feature/Mesh/MeshFeatureProcessorInterface.h
@@ -74,6 +74,7 @@ namespace AZ
 
             using MeshHandle = StableDynamicArrayHandle<ModelDataInstance>;
             using ModelChangedEvent = Event<const Data::Instance<RPI::Model>>;
+            using ObjectSrgCreatedEvent = Event<const Data::Instance<RPI::ShaderResourceGroup>&>;
 
             //! Returns the object id for a mesh handle.
             virtual TransformServiceFeatureProcessorInterface::ObjectId GetObjectId(const MeshHandle& meshHandle) const = 0;
@@ -115,6 +116,9 @@ namespace AZ
             virtual const CustomMaterialMap& GetCustomMaterials(const MeshHandle& meshHandle) const = 0;
             //! Connects a handler to any changes to an RPI::Model. Changes include loading and reloading.
             virtual void ConnectModelChangeEventHandler(const MeshHandle& meshHandle, ModelChangedEvent::Handler& handler) = 0;
+
+            //! Connects a handler to ObjectSrg creation
+            virtual void ConnectObjectSrgCreatedEventHandler(const MeshHandle& meshHandle, ObjectSrgCreatedEvent::Handler& handler) = 0;
 
             //! Sets the transform for a given mesh handle.
             virtual void SetTransform(const MeshHandle& meshHandle, const Transform& transform,
--- a/Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
@@ -443,6 +443,14 @@ namespace AZ
             }
         }
 
+        void MeshFeatureProcessor::ConnectObjectSrgCreatedEventHandler(const MeshHandle& meshHandle, ObjectSrgCreatedEvent::Handler& handler)
+        {
+            if (meshHandle.IsValid())
+            {
+                handler.Connect(meshHandle->GetObjectSrgCreatedEvent());
+            }
+        }
+
         void MeshFeatureProcessor::SetTransform(const MeshHandle& meshHandle, const AZ::Transform& transform, const AZ::Vector3& nonUniformScale)
         {
             if (meshHandle.IsValid())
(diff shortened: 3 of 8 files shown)


gadams3 (approved):
LGTM. In this case, I'm not sure what using events is adding over just registering a callback function with the FP or handle. Maybe an unnecessary layer of indirection since it's resulting in a notification bus call anyway.

invertednormal (approved):
While I'm not a huge fan of having an event trigger another event (az::event to ebus) that's a precedent that's already been set in this system with the model change event.

My one recommendation is to make this event less specific, like instead of announcing the event for objectsrg creation in the middle of `BuildDrawPacketList()`, what if you wait until the end of `BuildDrawPacketList()` and then announce that the draw packet has been created?

CyaNinix:
> LGTM. In this case, I'm not sure what using events is adding over just registering a callback function with the FP or handle. Maybe an unnecessary layer of indirection since it's resulting in a notification bus call anyway.

In this case it allows for multiple components to be aware of ObjectSrgs, a callback would limit it to only one.

> My one recommendation is to make this event less specific, like instead of announcing the event for objectsrg creation in the middle of BuildDrawPacketList(), what if you wait until the end of BuildDrawPacketList() and then announce that the draw packet has been created?

The idea here was to notify only when an ObjectSrg was created, not just when the draw packets were rebuilt (which occurs a lot more often), because to my understanding an ObjectSrg keeps its current data no matter what.

So by allowing a component to know when an ObjectSrg was created + querying and settings them when the mesh handles is retrieves allows one to set custom per-object data only once (or once per data update).

invertednormal:
> > LGTM. In this case, I'm not sure what using events is adding over just registering a callback function with the FP or handle. Maybe an unnecessary layer of indirection since it's resulting in a notification bus call anyway.
> 
> In this case it allows for multiple components to be aware of ObjectSrgs, a callback would limit it to only one.
> 
> > My one recommendation is to make this event less specific, like instead of announcing the event for objectsrg creation in the middle of BuildDrawPacketList(), what if you wait until the end of BuildDrawPacketList() and then announce that the draw packet has been created?
> 
> The idea here was to notify only when an ObjectSrg was created, not just when the draw packets were rebuilt (which occurs a lot more often), because to my understanding an ObjectSrg keeps its current data no matter what.
> 
> So by allowing a component to know when an ObjectSrg was created + querying and settings them when the mesh handles is retrieves allows one to set custom per-object data only once (or once per data update).

Does redrawing the draw packets actually get called much more often? It looks like only on `Init()` and when `SetUseForwardPassIblSpecular()` is called. `Init()` can happen for a few reasons, but usually for significant changes that could affect the objectSrg, like materials changing. Actually it could potentially cut down on events if the model has multiple object srgs for different meshes, then you could send just one event out for the model draw packet being done then all object srgs could be updated. Having a more generic event also could help in the future if there's some other aspect of building draw packets that someone wants to know has completed. Ultimately this isn't a blocking issue though, just a recommendation.

CyaNinix:
You may be right, I don't know very well the internal of this class. I set up an ObjectSrg event because I saw code that would reuse ObjectSrg when possible. Do you want me to change this or is it something that could be done after merging?

invertednormal:
> You may be right, I don't know very well the internal of this class. I set up an ObjectSrg event because I saw code that would reuse ObjectSrg when possible. Do you want me to change this or is it something that could be done after merging?

It'd be fine to do later - you have my 👍, so any remaining suggestions are just suggestions, nothing blocking. I'll kick off an AR run.

invertednormal:
> Since this new event can be triggered from multiple threads, I think it needs to be protected by a mutex (with lockless dispatch). Maybe this belongs to a new bus?

Hmm... maaaaybe. Each MeshInstance has its own event instance, and I don't think there's a way for the same exact event to signal at the same time from multiple threads, just different MeshInstances signaling simultaneously. That seems like it'd be ok though as long as the handler doesn't do anything that would affect MeshInstances other than itself. Do you see any concrete causes for concern?

rgba16f:
@CyaNinix This change appears to be causing a crash in the Editor loading this level - AutomatedTesting/Levels/Performance/10KEntityCpuPerfTest

PR #15179, opened 2023-03-17
Fix sfmt value repetition bugs

### LINKED ISSUE #13791: Test_Suite_main: test_AltitudeFilter_ComponentAndOverrides_InstancesPlantAtSpecifiedAltitude can fail with an Atom assert
**Describe the bug**
Test failed in main development AR run. Cannot reproduce locally. An unexpected number of instances were found during the test. It appears that instances failed to cleanup. There is an Atom assert regarding instances:

```
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  ==================================================================
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  <15:53:11> (System) - Trace::Assert
[2022-12-20T15:55:49.154Z] E   [editor_test.log]   D:\workspace\o3de\Code\Framework\AtomCore\AtomCore/Instance/InstanceDatabase.h(409): (1716) 'class AZStd::intrusive_ptr<class AZ::RPI::ShaderResourceGroup> __cdecl AZ::Data::InstanceDatabase<class AZ::RPI::ShaderResourceGroup>::EmplaceInstance(const struct AZ::Data::InstanceId &,const class AZ::Data::Asset<class AZ::Data::AssetData> &,const class AZStd::any *)'
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  <15:53:11> (System) - Instance creation for asset id {388A9D03-6DA9-4ED6-A55C-082C80F6E833}:0:0000000000000000 resulted in a recursive creation of that asset, which was unexpected. This asset might be erroneously referencing itself as a dependent asset.
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  <15:53:11> (System) - ------------------------------------------------
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  <15:53:11> (System) - D:\workspace\o3de\Code\Framework\AtomCore\AtomCore\Instance\InstanceDatabase.h (405) : AZ::Data::InstanceDatabase<AZ::RPI::ShaderResourceGroup>::EmplaceInstance
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  <15:53:11> (System) - D:\workspace\o3de\Code\Framework\AtomCore\AtomCore\Instance\InstanceDatabase.h (352) : AZ::Data::InstanceDatabase<AZ::RPI::ShaderResourceGroup>::Create
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  <15:53:11> (System) - D:\workspace\o3de\Gems\Atom\RPI\Code\Source\RPI.Public\Shader\Shader.cpp (541) : AZ::RPI::Shader::CreateDrawSrgForShaderVariant
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  <15:53:11> (System) - D:\workspace\o3de\Gems\Atom\RPI\Code\Source\RPI.Public\MeshDrawPacket.cpp (289) : `AZ::RPI::MeshDrawPacket::DoUpdate'::`2'::<lambda_1>::operator()
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  <15:53:11> (System) - D:\workspace\o3de\Code\Framework\AzCore\AzCore\std\function\function_template.h (173) : AZStd::Internal::function_util::get_invoker<bool __cdecl(AZ::Name const &,AZ::RPI::ShaderCollection::Item const &),AZStd::Internal::function_util::function_obj_tag,AZ
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  <15:53:11> (System) - D:\workspace\o3de\Gems\Atom\RPI\Code\Source\RPI.Public\Material\Material.cpp (171) : AZ::RPI::Material::ForAllShaderItems
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  <15:53:11> (System) - D:\workspace\o3de\Gems\Atom\RPI\Code\Source\RPI.Public\MeshDrawPacket.cpp (366) : AZ::RPI::MeshDrawPacket::DoUpdate
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  <15:53:11> (System) - D:\workspace\o3de\Gems\Atom\RPI\Code\Source\RPI.Public\MeshDrawPacket.cpp (151) : AZ::RPI::MeshDrawPacket::Update
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  <15:53:11> (System) - D:\workspace\o3de\Gems\Atom\Feature\Common\Code\Source\Mesh\MeshFeatureProcessor.cpp (1043) : AZ::Render::ModelDataInstance::BuildDrawPacketList
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  <15:53:11> (System) - D:\workspace\o3de\Gems\Atom\Feature\Common\Code\Source\Mesh\MeshFeatureProcessor.cpp (928) : AZ::Render::ModelDataInstance::Init
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  <15:53:11> (System) - D:\workspace\o3de\Gems\Atom\Feature\Common\Code\Source\Mesh\MeshFeatureProcessor.cpp (180) : `AZ::Render::MeshFeatureProcessor::Simulate'::`6'::<lambda_1>::operator()
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  <15:53:11> (System) - D:\workspace\o3de\Code\Framework\AzCore\AzCore\Jobs\Internal\JobManagerBase.cpp (29) : AZ::Internal::JobManagerBase::Process
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  <15:53:11> (System) - D:\workspace\o3de\Code\Framework\AzCore\AzCore\Jobs\Internal\JobManagerWorkStealing.cpp (421) : AZ::Internal::JobManagerWorkStealing::ProcessJobsInternal
[2022-12-20T15:55:49.154Z] E   [editor_test.log]  
[2022-12-20T15:55:49.155Z] E   [editor_test.log]  <15:53:11> (System) - D:\workspace\o3de\Code\Framework\AzCore\AzCore\std\parallel\thread.h (193) : AZStd::Internal::thread_info_impl<`AZStd::thread::thread<`AZ::Internal::JobManagerWorkStealing::CreateWorkerThreads'::`4'::<lambda_1> >'::`2'::<lambda_1> >::execute
[2022-12-20T15:55:49.155Z] E   [editor_test.log]  
[2022-12-20T15:55:49.155Z] E   [editor_test.log]  <15:53:11> (System) - D:\workspace\o3de\Code\Framework\AzCore\Platform\Common\WinAPI\AzCore\std\parallel\internal\thread_WinAPI.cpp (38) : AZStd::Internal::thread_run_function
[2022-12-20T15:55:49.155Z] E   [editor_test.log]  
[2022-12-20T15:55:49.155Z] E   [editor_test.log]  <15:53:11> (System) - 00007FFA9D2A268A (ucrtbase) : o_exp
[2022-12-20T15:55:49.155Z] E   [editor_test.log]  
[2022-12-20T15:55:49.155Z] E   [editor_test.log]  <15:53:11> (System) - 00007FFA9D607974 (KERNEL32) : BaseThreadInitThunk
[2022-12-20T15:55:49.155Z] E   [editor_test.log]  
[2022-12-20T15:55:49.155Z] E   [editor_test.log]  <15:53:11> (System) - 00007FFAA041A371 (ntdll) : RtlUserThreadStart
[2022-12-20T15:55:49.155Z] E   [editor_test.log]  
[2022-12-20T15:55:49.155Z] E   [editor_test.log]  <15:53:11> (System) - ==================================================================
[2022-12-20T15:55:49.155Z] E   [editor_test.log]  
[2022-12-20T15:55:49.155Z] E   [editor_test.log]  <15:53:11> (System) - ====Assert added to ignore list by spec and verbosity setting.====
```

**Failed Jenkins Job Information:**
https://jenkins.build.o3de.org/blue/organizations/jenkins/O3DE/detail/development/3666/pipeline/892

**Attachments**
[log.txt](https://github.com/o3de/o3de/files/10270709/log.txt)

**Additional context**
Initial checks  using the same asset were successful, as well as several other tests using the same azmodel for creating a temporary prefab for use in the test. The error message below does not seem valid for the asset in use:

```
(System) - Instance creation for asset id {388A9D03-6DA9-4ED6-A55C-082C80F6E833}:0:0000000000000000 resulted in a recursive creation of that asset, which was unexpected. This asset might be erroneously referencing itself as a dependent asset.
```

mbalfour-amzn:
## What does this PR do?

The Sfmt code had some subtle (and some not-so-subtle) bugs that led to race conditions which led to O3DE producing the same "random" UUIDs far more often than the odds would predict. This manifested as an intermittent assert ("Instance creation for asset id X resulted in a recursive creation of that asset, which was unexpected. This asset might be erroneously referencing itself as a dependent asset.") from Atom in the Instance database, since it is the most frequent consumer of parallel random UUID generation.

Thanks to @amzn-tommy for providing the unit test that made this bug nearly 100% reproducible!

The heart of the problem was some bad overly-clever threading logic that attempted to remain lockless as much as possible, but introduced bugs that were thread-unsafe and logically incorrect. For reference, here is the incorrect code:
```c++
        int index = m_index.fetch_add(1);
        if (index >= SfmtInternal::N32)
        {
            AZStd::lock_guard<decltype(m_generationMutex)> lock(m_generationMutex);
            // if this thread is the one that sets m_index to 0, then this thread
            // does the generation
            index += 1; // compare against the result of fetch_add(1) above
            if (m_index.compare_exchange_strong(index, 0))
            {
                SfmtInternal::gen_rand_all(*this);
            }
            // try again, with the new table
            return Rand32();
        }
        return m_psfmt32[index];
```
The problems are as follows:
1. m_index was getting set to 0 _before_ calling gen_rand_all(). If another thread then calls Rand32() before gen_rand_all() is called, it won't enter the locking path and will return a sequence of m_psfmt32 values that have already been used. This led to entire identical 128-bit UUIDs getting recreated. It might seem like this could be solved by delaying the m_index reset until after the gen_rand_all() call, but...
2. If a thread grabbed an index value and then was swapped out before using it, it's possible for the index to roll all the way back around and get used twice on the same m_psfmt32 table generation, meaning that the same 32-bit (or 64-bit) value will get returned multiple times. This is _less_ problematic than 1, because at least it doesn't cause an entire series of numbers to repeat, but it's still a problem for code that expects 32-bit or 64-bit numbers to collide very infrequently.
3. Other functions such as Seed() or FillArray32() modified the m_psfmt32 table without locking a mutex, so the table could get modified on one thread while in the midst of using it on a different thread. This isn't _necessarily_ a problem, but the algorithm will certainly be affected in unpredictable ways if two threads start modifying the table at the same time.

At the end of the day, the easiest and most verifiably correct solution is to guard all use of the m_psfmt32 table with a mutex. It might be possible to invent an alternate solution that can remain mostly-lockless, but it would likely be complicated by tracking a separate value to account for the number of table regenerations, and then continuously retrying until the table regeneration count remains stable during the entire lookup. Or it could be done with a boolean array to ensure each value is only used once per table generation. etc. They're all more complicated and not obviously-better solutions than just a good old-fashioned simple mutex.

The fixes consist of the following:
* Simplified the threading logic in Sfmt to simply be a mutex in each function that needs to run exclusively from the other functions. Also removed the sketchy "recursive retry" logic in Rand32/Rand64 since it's no longer necessary.
* Added a second assert to InstanceDatabase to help distinguish between duplicate IDs and recursive asset creation.
* Changed UUID creation to use 2 Rand64() calls instead of 4 Rand32() calls for improved performance.
* Added unit tests from @amzn-tommy to verify (as best as we can) that the race conditions no longer exist.

Fixes #13791 .

## How was this PR tested?

Ran the unit tests for Sfmt and UUID.
Ran the Editor / MultiplayerSample to spot-check that everything still behaved normally.

### DIFF
--- a/Code/Framework/AtomCore/AtomCore/Instance/InstanceDatabase.h
+++ b/Code/Framework/AtomCore/AtomCore/Instance/InstanceDatabase.h
@@ -391,8 +391,8 @@ namespace AZ
             // there might be a bug / race condition in the id generator. The same assert also occurs *after*
             // instance creation to help differentiate between a non-random id vs recursive creation of the same id.
             AZ_Assert(
-                m_database.contains(id),
-                "Database already contains an instance for this id, possibly a random id generation collision?",
+                !m_database.contains(id),
+                "Database already contains an instance for this id (%s), possibly a random id generation collision?",
                 id.ToFixedString().c_str());
 
             // Emplace a new instance and return it.
@@ -411,10 +411,10 @@ namespace AZ
             if (instance)
             {
                 AZ_Assert(
-                    m_database.find(id) == m_database.end(),
+                    !m_database.contains(id),
                     "Instance creation for asset id %s resulted in a recursive creation of that asset, which was unexpected. "
                     "This asset might be erroneously referencing itself as a dependent asset.",
-                    id.ToString<AZStd::string>().c_str());
+                    id.ToFixedString().c_str());
 
                 instance->m_id = id;
                 instance->m_parentDatabase = this;
--- a/Code/Framework/AtomCore/AtomCore/Instance/InstanceDatabase.h
+++ b/Code/Framework/AtomCore/AtomCore/Instance/InstanceDatabase.h
@@ -387,6 +387,14 @@ namespace AZ
         Data::Instance<Type> InstanceDatabase<Type>::EmplaceInstance(
             const InstanceId& id, const Data::Asset<AssetData>& asset, const AZStd::any* param)
         {
+            // This assert is here to catch any potential non-randomness in our id generation. If it triggers,
+            // there might be a bug / race condition in the id generator. The same assert also occurs *after*
+            // instance creation to help differentiate between a non-random id vs recursive creation of the same id.
+            AZ_Assert(
+                m_database.find(id) == m_database.end(),
+                "Database already contains an instance for this id, possibly a random id generation collision?",
+                id.ToString<AZStd::string>().c_str());
+
             // Emplace a new instance and return it.
             // It's possible for the m_createFunction call to recursively trigger another FindOrCreate call, so be aware that
             // the contents of m_database may change within this call.
--- a/Code/Framework/AtomCore/AtomCore/Instance/InstanceDatabase.h
+++ b/Code/Framework/AtomCore/AtomCore/Instance/InstanceDatabase.h
@@ -391,9 +391,9 @@ namespace AZ
             // there might be a bug / race condition in the id generator. The same assert also occurs *after*
             // instance creation to help differentiate between a non-random id vs recursive creation of the same id.
             AZ_Assert(
-                m_database.find(id) == m_database.end(),
+                m_database.contains(id),
                 "Database already contains an instance for this id, possibly a random id generation collision?",
-                id.ToString<AZStd::string>().c_str());
+                id.ToFixedString().c_str());
 
             // Emplace a new instance and return it.
             // It's possible for the m_createFunction call to recursively trigger another FindOrCreate call, so be aware that
--- a/Code/Framework/AtomCore/AtomCore/Instance/InstanceDatabase.h
+++ b/Code/Framework/AtomCore/AtomCore/Instance/InstanceDatabase.h
@@ -393,7 +393,7 @@ namespace AZ
             AZ_Assert(
                 !m_database.contains(id),
                 "Database already contains an instance for this id (%s), possibly a random id generation collision?",
-                id.ToFixedString().c_str());
+                id.ToString<AZStd::string>().c_str());
 
             // Emplace a new instance and return it.
             // It's possible for the m_createFunction call to recursively trigger another FindOrCreate call, so be aware that
@@ -414,7 +414,7 @@ namespace AZ
                     !m_database.contains(id),
                     "Instance creation for asset id %s resulted in a recursive creation of that asset, which was unexpected. "
                     "This asset might be erroneously referencing itself as a dependent asset.",
-                    id.ToFixedString().c_str());
+                    id.ToString<AZStd::string>().c_str());
 
                 instance->m_id = id;
                 instance->m_parentDatabase = this;
--- a/Code/Framework/AtomCore/AtomCore/Instance/InstanceDatabase.h
+++ b/Code/Framework/AtomCore/AtomCore/Instance/InstanceDatabase.h
@@ -393,7 +393,7 @@ namespace AZ
             AZ_Assert(
                 !m_database.contains(id),
                 "Database already contains an instance for this id (%s), possibly a random id generation collision?",
-                id.ToString<AZStd::string>().c_str());
+                id.ToString<AZStd::fixed_string<64>>().c_str());
 
             // Emplace a new instance and return it.
             // It's possible for the m_createFunction call to recursively trigger another FindOrCreate call, so be aware that
@@ -414,7 +414,7 @@ namespace AZ
                     !m_database.contains(id),
                     "Instance creation for asset id %s resulted in a recursive creation of that asset, which was unexpected. "
                     "This asset might be erroneously referencing itself as a dependent asset.",
-                    id.ToString<AZStd::string>().c_str());
+                    id.ToString<AZStd::fixed_string<64>>().c_str());
 
                 instance->m_id = id;
                 instance->m_parentDatabase = this;
--- a/Code/Framework/AzCore/AzCore/Math/Sfmt.cpp
+++ b/Code/Framework/AzCore/AzCore/Math/Sfmt.cpp
@@ -392,6 +392,8 @@ namespace AZ
     //=========================================================================
     void Sfmt::Seed(AZ::u32* keys, int numKeys)
     {
+        AZStd::lock_guard<decltype(m_generationMutex)> lock(m_generationMutex);
+
         using SfmtInternal::N;
         using SfmtInternal::N32;
         int i, j, count;
@@ -551,6 +553,8 @@ namespace AZ
     //=========================================================================
     void Sfmt::FillArray32(AZ::u32* array, int size)
     {
+        AZStd::lock_guard<decltype(m_generationMutex)> lock(m_generationMutex);
+
         AZ_MATH_ASSERT(m_index == SfmtInternal::N32, "Invalid m_index! Reinitialize!");
         AZ_MATH_ASSERT(size % 4 == 0, "Size must be multiple of 4!");
         AZ_MATH_ASSERT(size >= SfmtInternal::N32, "Size must be bigger than %d GetMinArray32Size()!", SfmtInternal::N32);
@@ -565,6 +569,8 @@ namespace AZ
     //=========================================================================
     void Sfmt::FillArray64(AZ::u64* array, int size)
     {
+        AZStd::lock_guard<decltype(m_generationMutex)> lock(m_generationMutex);
+
         AZ_MATH_ASSERT(m_index == SfmtInternal::N32, "Invalid m_index! Reinitialize!");
         AZ_MATH_ASSERT(size % 4 == 0, "Size must be multiple of 4!");
         AZ_MATH_ASSERT(size >= SfmtInternal::N64, "Size must be bigger than %d GetMinArray64Size()!", SfmtInternal::N64);
--- a/Code/Framework/AzCore/AzCore/Math/Sfmt.cpp
+++ b/Code/Framework/AzCore/AzCore/Math/Sfmt.cpp
@@ -553,7 +553,7 @@ namespace AZ
     //=========================================================================
     void Sfmt::FillArray32(AZ::u32* array, int size)
     {
-        AZStd::lock_guard<decltype(m_generationMutex)> lock(m_generationMutex);
+        AZStd::scoped_lock lock(m_generationMutex);
 
         AZ_MATH_ASSERT(m_index == SfmtInternal::N32, "Invalid m_index! Reinitialize!");
         AZ_MATH_ASSERT(size % 4 == 0, "Size must be multiple of 4!");
--- a/Code/Framework/AzCore/AzCore/Math/Sfmt.cpp
+++ b/Code/Framework/AzCore/AzCore/Math/Sfmt.cpp
@@ -515,21 +515,16 @@ namespace AZ
     //=========================================================================
     AZ::u32 Sfmt::Rand32()
     {
-        int index = m_index.fetch_add(1);
-        if (index >= SfmtInternal::N32)
+        AZStd::lock_guard<decltype(m_generationMutex)> lock(m_generationMutex);
+
+        m_index++;
+        if (m_index >= SfmtInternal::N32)
         {
-            AZStd::lock_guard<decltype(m_generationMutex)> lock(m_generationMutex);
-            // if this thread is the one that sets m_index to 0, then this thread
-            // does the generation
-            index += 1; // compare against the result of fetch_add(1) above
-            if (m_index.compare_exchange_strong(index, 0))
-            {
-                SfmtInternal::gen_rand_all(*this);
-            }
-            // try again, with the new table
-            return Rand32();
+            SfmtInternal::gen_rand_all(*this);
+            m_index = 0;
         }
-        return m_psfmt32[index];
+
+        return m_psfmt32[m_index];
     }
 
     //=========================================================================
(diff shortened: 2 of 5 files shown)


AMZN-Olex (approved):
Good find!

amzn-tommy (approved):

Code/Framework/AtomCore/AtomCore/Instance/InstanceDatabase.h
```diff
@@ -387,6 +387,14 @@ namespace AZ
         Data::Instance<Type> InstanceDatabase<Type>::EmplaceInstance(
             const InstanceId& id, const Data::Asset<AssetData>& asset, const AZStd::any* param)
         {
+            // This assert is here to catch any potential non-randomness in our id generation. If it triggers,
+            // there might be a bug / race condition in the id generator. The same assert also occurs *after*
+            // instance creation to help differentiate between a non-random id vs recursive creation of the same id.
+            AZ_Assert(
+                m_database.find(id) == m_database.end(),
+                "Database already contains an instance for this id, possibly a random id generation collision?",
+                id.ToString<AZStd::string>().c_str());
```
lumberyard-employee-dm, line 396: It is better to use ToFixedString to avoid a heap allocation for the Uuid when the assert occurs.
Also the `.contains` function can be used to determine if the id is in the map container.
```suggestion
            AZ_Assert(
                m_database.contains(id),
                "Database already contains an instance for this id, possibly a random id generation collision?",
                id.ToFixedString().c_str());
```

Code/Framework/AzCore/AzCore/Math/Sfmt.cpp
```diff
@@ -564,6 +553,8 @@ namespace AZ
     //=========================================================================
     void Sfmt::FillArray32(AZ::u32* array, int size)
     {
+        AZStd::lock_guard<decltype(m_generationMutex)> lock(m_generationMutex);
```
lumberyard-employee-dm, line 556: Class template argument deduction can be used here to deduce the argument type.
```suggestion
        AZStd::scoped_lock lock(m_generationMutex);
```

Code/Framework/AzCore/AzCore/Math/Sfmt.cpp
```diff
@@ -578,6 +569,8 @@ namespace AZ
     //=========================================================================
     void Sfmt::FillArray64(AZ::u64* array, int size)
     {
+        AZStd::lock_guard<decltype(m_generationMutex)> lock(m_generationMutex);
```
lumberyard-employee-dm, line 572: repeat: Class template argument deduction can be used here.
```suggestion
        AZStd::scoped_lock lock(m_generationMutex);
```

Code/Framework/AzCore/AzCore/Math/Sfmt.h
```diff
@@ -106,10 +106,10 @@ namespace AZ
         void    PeriodCertification();
 
         SfmtInternal::w128_t        m_sfmt[SfmtInternal::N];
-        AZStd::atomic_int           m_index;   ///  Index into the pre-generated tables
+        size_t                      m_index;   ///  Index into the pre-generated tables
```
lumberyard-employee-dm, line 109: This index isn't initialized at all.
I recommend initializing the `m_index` and `m_psfmt32` and `m_psfmt64` variables to 0 and nullptr

lumberyard-employee-dm (changes requested):

Code/Framework/AzCore/AzCore/Math/Sfmt.cpp
```diff
@@ -392,6 +392,8 @@ namespace AZ
     //=========================================================================
     void Sfmt::Seed(AZ::u32* keys, int numKeys)
     {
+        AZStd::lock_guard<decltype(m_generationMutex)> lock(m_generationMutex);
```
lumberyard-employee-dm: Can you update this function to accept a `const AZ::u32*`?
The keys aren't being modified by this function.
```suggestion
    void Sfmt::Seed(const AZ::u32* keys, int numKeys)
    {
        AZStd::scoped_lock lock(m_generationMutex);
```

Code/Framework/AzCore/AzCore/Math/Sfmt.cpp
```diff
@@ -515,21 +517,16 @@ namespace AZ
     //=========================================================================
     AZ::u32 Sfmt::Rand32()
     {
-        int index = m_index.fetch_add(1);
-        if (index >= SfmtInternal::N32)
+        AZStd::lock_guard<decltype(m_generationMutex)> lock(m_generationMutex);
```
lumberyard-employee-dm: Class Template Argument deduction can be used here.
```suggestion
        AZStd::scoped_lock lock(m_generationMutex);
```

Code/Framework/AzCore/AzCore/Math/Sfmt.cpp
```diff
@@ -538,24 +535,16 @@ namespace AZ
     //=========================================================================
     AZ::u64 Sfmt::Rand64()
     {
-        int index = m_index.fetch_add(2);
-        if (index >= (SfmtInternal::N32 - 1))
+        AZStd::lock_guard<decltype(m_generationMutex)> lock(m_generationMutex);
```
lumberyard-employee-dm: repeat: Class Template argument deduction can be used here.
```suggestion
        AZStd::scoped_lock lock(m_generationMutex);
```

Code/Framework/AzCore/Tests/Math/SfmtTests.cpp
```diff
@@ -107,6 +109,82 @@ namespace UnitTest
...
+        template<size_t NumThreads, size_t NumResultsPerThread, typename ResultType>
+        bool ParallelTest()
+        {
+            Sfmt sfmt;
+
+            // Arbitrary but deterministic seed values to guarantee that we get the same random numbers produced every time.
+            AZ::u32 buffer[32] = { 0x8236ed73, 0x854aa369, 0xc7b68864, 0x5f1e49da, 0x58ea5a08, 0xa33fc74b, 0x0336bd81, 0x8d3c11ac,
+                                   0x7312bc57, 0x92fee3d1, 0x4b852f9f, 0xa7ac02ad, 0x4d72fb7a, 0x630641c6, 0x1edbeebf, 0xc18fdabe,
+                                   0xc6588dba, 0x6a821cf5, 0xec7e24b3, 0x44246d7e, 0x24af2f32, 0x4f64d44c, 0x44b24116, 0x65585572,
+                                   0x0f95038d, 0xd3e4c521, 0x6eea6bc1, 0x6d651ab5, 0x25dfc39e, 0x7502d183, 0xd32fc9bf, 0x9854093f };
```
lumberyard-employee-dm, line 122: This buffer can be `constexpr`
```suggestion
            // Arbitrary but deterministic seed values to guarantee that we get the same random numbers produced every time.
            constexpr AZ::u32 buffer[32] = { 0x8236ed73, 0x854aa369, 0xc7b68864, 0x5f1e49da, 0x58ea5a08, 0xa33fc74b, 0x0336bd81, 0x8d3c11ac,
                                   0x7312bc57, 0x92fee3d1, 0x4b852f9f, 0xa7ac02ad, 0x4d72fb7a, 0x630641c6, 0x1edbeebf, 0xc18fdabe,
                                   0xc6588dba, 0x6a821cf5, 0xec7e24b3, 0x44246d7e, 0x24af2f32, 0x4f64d44c, 0x44b24116, 0x65585572,
                                   0x0f95038d, 0xd3e4c521, 0x6eea6bc1, 0x6d651ab5, 0x25dfc39e, 0x7502d183, 0xd32fc9bf, 0x9854093f };
```

Code/Framework/AzCore/Tests/Math/SfmtTests.cpp
```diff
@@ -107,6 +109,82 @@ namespace UnitTest
...
+
+                auto threadFunc = [&startSync, startOffset, &results, &sfmt]()
+                {
+                    // Block so that we can start all the threads at the same time after creation.
+                    startSync.acquire();
+
+                    for (int i = 0; i < NumResultsPerThread; ++i)
+                    {
+                        // Call Rand32 when uint32_t is the passed-in type, and Rand64 if uint64_t is the passed-in type.
+                        if constexpr (AZStd::is_same<ResultType, uint32_t>::value)
```
lumberyard-employee-dm, line 150: The `_v` shortcut can be used for [AZStd::is_same](https://en.cppreference.com/w/cpp/types/is_same) to save a little bit of typing.
```suggestion
                        if constexpr (AZStd::is_same_v<ResultType, uint32_t>)
```

Code/Framework/AtomCore/AtomCore/Instance/InstanceDatabase.h
mbalfour-amzn: InstanceId (which isn't a UUID but does contain one) doesn't have a ToFixedString() method.

lumberyard-employee-dm: Ok, alternative solution is to use the template argument to pass in `Uuid::FixedString`

lumberyard-employee-dm (approved):

Code/Framework/AzCore/AzCore/Math/Sfmt.cpp
mbalfour-amzn: Also fixed the constructor.

lumberyard-employee-dm: Thanks!

PR #15198, opened 2023-03-20
Asset Processor - Handle edge case of product prefix conflict

amzn-mike:
## What does this PR do?

Fixes an edge case where asset products could be overwritten if they used the same naming scheme as is used for override prefixing.

When files are copied to the cache they now are given a temporary but longer prefix that shouldn't accidentally conflict with an asset naming scheme.

## How was this PR tested?

Added and ran unit test

### DIFF
--- a/Code/Tools/AssetProcessor/native/tests/assetmanager/AssetProcessorManagerTest.cpp
+++ b/Code/Tools/AssetProcessor/native/tests/assetmanager/AssetProcessorManagerTest.cpp
@@ -380,6 +380,9 @@ public:
 
         using namespace AssetBuilderSDK;
 
+        // Set up a custom builder with a ProcessJob stage that will output 2 files, one of which is intentionally
+        // designed to output a product with a name that conflicts with the prefixing scheme (the (2) prefix).
+        // .stage1 is the input and .stage2 is the output.  This unit test framework currently requires those extensions.
         m_builderInfoHandler.CreateBuilderDesc(
             "stage1",
             AZ::Uuid::CreateRandom().ToFixedString().c_str(),
--- a/Code/Tools/AssetProcessor/native/tests/assetmanager/AssetProcessorManagerTest.cpp
+++ b/Code/Tools/AssetProcessor/native/tests/assetmanager/AssetProcessorManagerTest.cpp
@@ -380,7 +380,44 @@ public:
 
         using namespace AssetBuilderSDK;
 
-        CreateBuilder("stage1", "*.stage1", "stage2", false, ProductOutputFlags::ProductAsset);
+        m_builderInfoHandler.CreateBuilderDesc(
+            "stage1",
+            AZ::Uuid::CreateRandom().ToFixedString().c_str(),
+            { AssetBuilderPattern{ "*.stage1", AssetBuilderPattern::Wildcard } },
+            CreateJobStage("stage1", false),
+            [](const ProcessJobRequest& request, ProcessJobResponse& response)
+            {
+                AZ::IO::FixedMaxPath outputFile = AZ::IO::FixedMaxPath(request.m_sourceFile);
+                outputFile.ReplaceExtension("stage2");
+                outputFile = outputFile.Filename();
+
+                AZ::IO::Result result = AZ::IO::FileIOBase::GetInstance()->Copy(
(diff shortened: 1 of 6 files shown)


Code/Tools/AssetProcessor/native/utilities/ProductOutputUtil.h
```diff
@@ -17,15 +17,22 @@ namespace AssetProcessor
 {
     struct ProductOutputUtil
     {
-        static AZStd::string GetPrefix(AZ::s64 scanfolderId);
+        //! Gets the prefix to apply to products when copying into the cache before being finalized.
+        //! This is different from the final prefix version to avoid conflicts unprefixed assets that may use the same prefix as the final prefix.
+        //! The temp prefix shouldn't show up in any UI so its longer and uglier.
```
lmbr-pip, line 22: nit: I understand what you are trying to say with uglier, but it looks odd in a comment. Maybe "Its formatted as "TMP<number>" to make it clear that its an intermediate value.

Code/Tools/AssetProcessor/native/utilities/ProductOutputUtil.h
```diff
@@ -17,15 +17,22 @@ namespace AssetProcessor
...
+        static AZStd::string GetTempPrefix(AZ::s64 scanfolderId);
+        //! Gets the prefix to apply to products when finalizing.
+        static AZStd::string GetFinalPrefix(AZ::s64 scanfolderId);
 
-        // Modifies product paths to have the scanfolder prepended to the filename
-        // This needs to be done before the files are copied into the cache to avoid overwriting the legacy non-prepended version
-        static void ModifyProductPath(QString& outputFilename, AZ::s64 sourceScanfolderId);
+        //! Modifies product paths to have the temporary prefixed version while waiting on final processing.
+        //! This needs to be done before the files are copied into the cache to avoid overwriting the legacy non-prepended version.
+        static void GetTempProductPath(QString& outputFilename, AZ::s64 sourceScanfolderId);
```
lmbr-pip, line 29: See comment about avoid "temp" + "path" names

Code/Tools/AssetProcessor/native/utilities/ProductOutputUtil.cpp
```diff
@@ -96,23 +119,73 @@ namespace AssetProcessor
...
+        QString filename = AZ::IO::PathView(product.m_productFileName).Filename().FixedMaxPathString().c_str();
+
+        int prefixPos = filename.indexOf(currentPrefix.c_str());
+
+        if (prefixPos < 0)
+        {
+            AZ_Error(
+                "ProductOutputUtil",
+                false,
+                "Product " AZ_STRING_FORMAT " is expected to be prefixed but was not",
```
AMZN-stankowi, line 162: This error won't mean a lot to most people. Maybe prefix it with Programmer Error like you've done with other errors, or "Asset Processor Internal Error"? Then non-engineers will know to grab an engineer to help, the engineer will find this code and if they're unfamiliar with the prefix system, there's enough context in this function to work backward toward solving the problem.

AMZN-stankowi (approved):

Code/Tools/AssetProcessor/native/tests/assetmanager/AssetProcessorManagerTest.cpp
```diff
@@ -380,7 +380,44 @@ class MetadataOverrides : public UnitTests::AssetManagerTestingBase
 
         using namespace AssetBuilderSDK;
 
-        CreateBuilder("stage1", "*.stage1", "stage2", false, ProductOutputFlags::ProductAsset);
+        m_builderInfoHandler.CreateBuilderDesc(
+            "stage1",
+            AZ::Uuid::CreateRandom().ToFixedString().c_str(),
+            { AssetBuilderPattern{ "*.stage1", AssetBuilderPattern::Wildcard } },
+            CreateJobStage("stage1", false),
```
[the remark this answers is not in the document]
amzn-mike: Added a bit of a comment but this test follows the pattern of all the others using this framework.  The framework was originally designed alongside the intermediate asset testing, so it makes some assumptions about naming conventions which has forced other tests using this to also use .stageX format.

lmbr-pip (approved):

Issue #9789, opened 2022-05-26
Re-enable Spawnable tests.


AMZN-mnaumov:
Lua spawnable integration tests are currently broken on Linux because of an unknown intermittent issue.
https://jenkins.build.o3de.org/blue/organizations/jenkins/O3DE/detail/development/2409/pipeline

jckand-amzn:
Re-opening as tests still intermittently fail on both Windows and Linux.

jckand-amzn:
Ran the Prefab test suite 100x on both Windows and Linux with 0 local failures. @chiyenteng has encountered failures when running locally, and on AR.

chiyenteng:
On AR: https://jenkins.build.o3de.org/blue/organizations/jenkins/O3DE/detail/PR-9933/3/pipeline/665

chiyenteng:
E   [editor_test.log]  <10:31:11> (python_test) -
E   [editor_test.log]  <10:31:11> (python_test) - Info: Spawned Entity Scale: Found 8.889683573327787e-18, expected 5.0
E   [editor_test.log]  <10:31:11> (python_test) -
E   [editor_test.log]  <10:31:11> (python_test) - Info: Spawned Entity Scale: Found 8.889683573327787e-18, expected 5.0
E   [editor_test.log]  <10:31:11> (python_test) -
E   [editor_test.log]  <10:31:11> (python_test) - Test Lua_Spawnables_NestedSpawn finished.
E   [editor_test.log]  Report:
E   [editor_test.log]  EXCEPTION raised:
E   [editor_test.log]    Traceback (most recent call last):
E   [editor_test.log]      File "d:\ly\o3de\o3de\automatedtesting\gem\pythontests\editorpythontesttools\editor_python_test_tools\utils.py", line 305, in start_test
E   [editor_test.log]        test_function()
E   [editor_test.log]      File "D:/ly/o3de/o3de/AutomatedTesting/Gem/PythonTests/Prefab/tests/spawnables/Lua_Spawnables_NestedSpawn.py", line 57, in Lua_Spawnables_NestedSpawn
E   [editor_test.log]        expected_spawned_entity_scale)
E   [editor_test.log]      File "D:\ly\o3de\o3de\AutomatedTesting\Gem\Editor\Scripts/../../PythonTests\Prefab\tests\PrefabTestUtils.py", line 301, in validate_spawned_entity_transform
[...]

PR #15202, opened 2023-03-20
Moving spawnables tests into main suite

jckand-amzn:
## What does this PR do?

- Moves spawnables tests back into the main suite
- Resolves https://github.com/o3de/o3de/issues/9789

## How was this PR tested?

Ran spawnables tests 100x locally without failure.
Ran updated main test suite without failure.
Ran updated periodic suite to ensure dummy test does not cause a failure.

### DIFF
--- a/AutomatedTesting/Gem/PythonTests/Prefab/TestSuite_Periodic.py
+++ b/AutomatedTesting/Gem/PythonTests/Prefab/TestSuite_Periodic.py
@@ -7,7 +7,7 @@ SPDX-License-Identifier: Apache-2.0 OR MIT
 
 import pytest
 
-from ly_test_tools.o3de.editor_test import EditorBatchedTest, EditorTestSuite
+from ly_test_tools.o3de.editor_test import EditorBatchedTest, EditorSingleTest, EditorTestSuite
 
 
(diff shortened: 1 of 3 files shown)


AutomatedTesting/Gem/PythonTests/Prefab/TestSuite_Periodic.py
```diff
@@ -7,7 +7,7 @@
 
 import pytest
 
-from ly_test_tools.o3de.editor_test import EditorBatchedTest, EditorTestSuite
+from ly_test_tools.o3de.editor_test import EditorBatchedTest, EditorSingleTest, EditorTestSuite
```
forkercat, line 11: Looks like "EditorBatchedTest" can be removed

forkercat (approved):

AutomatedTesting/Gem/PythonTests/Prefab/TestSuite_Periodic.py
jckand-amzn: It'll be re-used as new tests are added. Don't think it hurts to leave it as-is for now.

cgalvan (approved):

PR #15211, opened 2023-03-21
Fix streaming issue when changing resolution on Linux

RogerShaw:
## What does this PR do?
Run GameLauncher process on Euler or Ubuntu System. This process can encode video and streaming. If I change resolution during the process is running, the image of encoding and streaming will stop. This PR can fix this issue.

## How was this PR tested?
1. Strat a virtual window using "Startx";
2. Export virtual window and run GameLauncher process on Euler or Ubuntu System;
3. Change resolution using code below
EBUS_EVENT(AzFramework::WindowRequestBus, ResizeClientArea, AzFramework::WindowSize(width, height));
4. Check whether streaming of GameLauncher process if normal;

### DIFF
--- a/Code/Framework/AzFramework/Platform/Common/Xcb/AzFramework/XcbNativeWindow.cpp
+++ b/Code/Framework/AzFramework/Platform/Common/Xcb/AzFramework/XcbNativeWindow.cpp
@@ -275,9 +275,18 @@ namespace AzFramework
     void XcbNativeWindow::ResizeClientArea(WindowSize clientAreaSize, const WindowPosOptions& options)
     {
         const uint32_t values[] = { clientAreaSize.m_width, clientAreaSize.m_height };
-
+        
+        if (m_activated)
+        {
+            xcb_unmap_window(m_xcbConnection, m_xcbWindow);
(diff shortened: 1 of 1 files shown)


hultonha (approved):
Thank you very much for the contribution @RogerShaw! Change looks good to me. One minor suggestion would be it might be useful to add a comment around the new added statements to indicate _why_ they're needed (just to help future maintainers) but otherwise looks good. Let me know your thoughts and I can kick off a build shortly 👍

RogerShaw:
My thought is simple. When resolution of window is changed, do "Deactivate" and "Activate" again. @hultonha

hultonha:
> My thought is simple. When resolution of window is changed, do "Deactivate" and "Activate" again. @hultonha

Makes total sense, fix is good, it just can be helpful to include what `xcb_unmap_window` and `xcb_map_window` are doing and why `xcb_flush` is needed. It's not a huge deal but very useful for context. Let me know if you think it's worth adding any extra info in the form of a comment and then I can start a build one way or the other :+1:

rgba16f (approved):

PR #15231, opened 2023-03-21
Action Manager | Introduce "Create" context sub-menu, restore "Create multiplayer entity" menu item.

AMZN-daimini:
## What does this PR do?
- Introduces a "Create" sub-menu to the Outliner and Viewport context menus. This can be used to provide entity presets to streamline entity creation.
- Restores the "Create multiplayer entity" menu item that was present before the migration to the new Action Manager system. This is now included in the "Create" sub-menu described above.

## How was this PR tested?
Verified a multiplayer entity is created correctly.

In verifying this change, I realized that in both the new Action Manager and the legacy system the behavior for creating entities by right-clicking in the viewport is inconsistent. I'm working on a follow-up PR to standardize the behavior and possibly provide helper functions to simplify adding new menu items to the "Create" menu down the line with consistent behavior.

image

### DIFF
--- a/Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
+++ b/Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
@@ -29,8 +29,14 @@
 #include <AzFramework/API/ApplicationAPI.h>
 #include <AzFramework/Process/ProcessUtils.h>
 #include <AzNetworking/Framework/INetworking.h>
+#include <AzToolsFramework/ActionManager/Action/ActionManagerInterface.h>
+#include <AzToolsFramework/ActionManager/HotKey/HotKeyManagerInterface.h>
+#include <AzToolsFramework/ActionManager/Menu/MenuManagerInterface.h>
 #include <AzToolsFramework/API/EntityCompositionRequestBus.h>
 #include <AzToolsFramework/ContainerEntity/ContainerEntityInterface.h>
+#include <AzToolsFramework/Editor/ActionManagerIdentifiers/EditorActionUpdaterIdentifiers.h>
+#include <AzToolsFramework/Editor/ActionManagerIdentifiers/EditorContextIdentifiers.h>
+#include <AzToolsFramework/Editor/ActionManagerIdentifiers/EditorMenuIdentifiers.h>
 #include <AzToolsFramework/Entity/PrefabEditorEntityOwnershipInterface.h>
 #include <AzToolsFramework/Entity/ReadOnly/ReadOnlyEntityInterface.h>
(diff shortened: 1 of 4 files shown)


cgalvan (approved):

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
```diff
@@ -664,6 +672,107 @@ namespace Multiplayer
...
+
+            actionManagerInterface->InstallEnabledStateCallback(
+                actionIdentifier,
+                [readOnlyEntityPublicInterface]()
+                {
+                    AzToolsFramework::EntityIdList selectedEntities;
+                    AzToolsFramework::ToolsApplicationRequests::Bus::BroadcastResult(
+                        selectedEntities, &AzToolsFramework::ToolsApplicationRequests::Bus::Events::GetSelectedEntities);
+
+                    if (selectedEntities.size() == 0)
```
hultonha, line 736: nit: For consistency with above it would be good to use `empty()` instead

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.h
```diff
@@ -95,15 +97,19 @@ namespace Multiplayer
         //! @{
         void OnStartPlayInEditorBegin() override;
         void OnStartPlayInEditor() override;
+        void OnStopPlayInEditorBegin() override;
```
hultonha, line 100: question: How come this was added? I don't see the implementation 🤔

hultonha (commented):
Overall changes look good but would be super useful to see a screen shot or video of what the updated menu looks like in the PR description if possible. Also maybe in the follow-up PR would be nice to see if this functionality can be tested too 🙂

gadams3 (approved):

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.h
hultonha: Haha I see thanks! 😅

hultonha (approved):

AMZN-daimini:
> Overall changes look good but would be super useful to see a screen shot or video of what the updated menu looks like in the PR description if possible. Also maybe in the follow-up PR would be nice to see if this functionality can be tested too 🙂

Added a screenshot for reference :)

hultonha:
> > Overall changes look good but would be super useful to see a screen shot or video of what the updated menu looks like in the PR description if possible. Also maybe in the follow-up PR would be nice to see if this functionality can be tested too 🙂
> 
> Added a screenshot for reference :)

Thank you 🧡

PR #15249, opened 2023-03-21
Turn on new Asset Browser

dtamkin1:
## What does this PR do?

Turns on the new asset browser and moves it to be tabbed with the console instead of under the Entity Outliner. Also changes the Asset Browser Inspector to be tabbed next to the asset browser.

## How was this PR tested?

Locally


AMZN-daimini (approved):

cgalvan:
> I would also recommend changing the default editor layout so that the Asset Browser is tabbed with the Console instead of under the Entity Outliner. So if the table view is default, the widget can spread horizontally and be much more readable :)

I agree! @dtamkin1 check out `QtViewPaneManager::RestoreDefaultLayout` to change the default layout of the panes.

cgalvan (approved):

jjjoness (approved):

Code/Editor/QtViewPaneManager.cpp
```diff
@@ -1144,6 +1144,25 @@ void QtViewPaneManager::RestoreDefaultLayout(bool resetSettings)
         m_mainWindow->addDockWidget(Qt::BottomDockWidgetArea, consoleViewPane->m_dockWidget);
         consoleViewPane->m_dockWidget->setFloating(false);
 
+        if (assetBrowserViewPane)
+        {
+
```
cgalvan, line 1149: Nit: Remove extra whitespace

Code/Editor/QtViewPaneManager.cpp
```diff
@@ -1144,6 +1144,25 @@ void QtViewPaneManager::RestoreDefaultLayout(bool resetSettings)
...
+            m_mainWindow->addDockWidget(Qt::BottomDockWidgetArea, assetBrowserViewPane->m_dockWidget);
+            assetBrowserViewPane->m_dockWidget->setFloating(false);
+
+            AzQtComponents::DockTabWidget* bottomTabWidget = m_advancedDockManager->tabifyDockWidget(assetBrowserViewPane->m_dockWidget, consoleViewPane->m_dockWidget, m_mainWindow);
+            if (bottomTabWidget)
```
cgalvan, line 1154: Nit: We shouldn't fail to create a tab dock widget here, so could change to assert here instead

Code/Editor/QtViewPaneManager.cpp
```diff
@@ -1144,6 +1144,25 @@ void QtViewPaneManager::RestoreDefaultLayout(bool resetSettings)
         m_mainWindow->addDockWidget(Qt::BottomDockWidgetArea, consoleViewPane->m_dockWidget);
         consoleViewPane->m_dockWidget->setFloating(false);
 
+        if (assetBrowserViewPane)
```
cgalvan, line 1147: General question for this part and the below logic that was removed: Do we want to keep the old default layout if `ed_useWIPAssetBrowserDesign` gets disabled by the user? If it looks OK in both scenarios, then we can leave the new layout for both cases.

cgalvan (approved):

Code/Editor/QtViewPaneManager.cpp
dtamkin1: It looks okay in both scenarios, the old AB doesn't offer much functionality so it still looks and works fine when it is tabbed at the bottom with the console.

PR #15256, opened 2023-03-21
Add component type name based aliases for components in prefabs

srikappa-amzn:
## What does this PR do?

This PR uses typenames of components as the serialized identifier (alias) for components in prefabs. 

Previously, an editor component in a prefab file used to look like 
```
"Component_[1438000288113878947]": {
    "$type": "EditorEntityIconComponent",
    "Id": 1438000288113878947
}
```

Now the same component would look like
```
"EditorEntityIconComponent": {
    "$type": "EditorEntityIconComponent",
    "Id": 546527165471463920
}
```

If there are multiple components of the same type on the entity, the alias will follow a numbering pattern like below
```
"EditorCommentComponent": {
    "$type": "EditorCommentComponent",
    "Id": 1335213497284657070
},
"EditorCommentComponent_2": {
    "$type": "EditorCommentComponent",
    "Id": 16928103960535337869
},
```

## How was this PR tested?

Added unit tests to test new behavior.

Also tested manually the following things:
1. Data in existing prefabs won't change. This ensures that any overrides on components will continue to work
2. Underlying component type names are correctly fetched for components that are wrapped in a GenericComponentWrapper
```
"FlyCameraInputComponent": {
    "$type": "GenericComponentWrapper",
    "Id": 17838737984030005354,
    "m_template": {
        "$type": "FlyCameraInputComponent"
    }
}
```


Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorComponentBase.cpp
```diff
@@ -49,14 +49,43 @@ namespace AzToolsFramework
...
-                AZ_Assert(GetId() != AZ::InvalidComponentId, "ComponentId is invalid.");
-                m_alias = AZStd::string::format("Component_[%llu]", GetId());
+                AZStd::unordered_set<AZStd::string> serializedIdentifiersMatchingType;
+
+                for (const AZ::Component* componentInEntity : entity->GetComponents())
+                {
+                    if (componentInEntity == this)
+                    {
+                        return;
+                    }
```
forkercat, line 63: Should it be 'continue'?

forkercat (commented):
Very nice to change to have type name as the component alias.

Since this change is going into dev and the outliner override management is already in stab/2305, do we need to think about adding migration support for the old override patches? In other words, in this release, users will start making override patches in prefab file:

```json
{
    "op": "replace",
    "path": "/Entities/Entity_[16218641747988]/Components/Component_[3813021942463140055]/IsStatic",
    "value": true
}
```

Is it still recognized by prefab override handler after this PR? I assume the new path is like this:

```
/Entities/Entity_[16218641747988]/Components/EditorTagComponent/IsStatic
```

Code/Framework/AzToolsFramework/AzToolsFramework/Entity/EditorEntityHelpers.cpp
```diff
@@ -213,6 +213,18 @@ namespace AzToolsFramework
...
     }
 
+    AZStd::string GetNameFromComponentClassData(const AZ::Component* component)
+    {
+        const char* className = component->RTTI_GetTypeName();
+        const AZ::SerializeContext::ClassData* classData = GetComponentClassData(component);
+        if (!classData)
+        {
+            return className;
+        }
```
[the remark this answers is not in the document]
srikappa-amzn: This is a safety check. If you look at GetComponentClassData(), you'll see that it's fetching the class data for the underlying component type and it can be a nullptr. So, in the off chance that it happens, we are using fallback logic to give it the typename of component. 

An example where this can can happen if a component forgets to reflect its data to serializecontext or don't want to reflect at all. In that case, this'll fetch 'GenericComponentWrapper' as the typename. Will add a comment here.

Code/Framework/AzCore/AzCore/Component/Component.cpp
```diff
@@ -163,6 +163,10 @@ namespace AZ
         }
     }
 
+    void Component::PrepareComponentForEntityAddition(Entity*)
+    {
+    }
```
lumberyard-employee-dm, line 168: What was the issue with using SetEntity?

Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorComponentBase.cpp
srikappa-amzn: This is stopping the function execution if the component is already added to the entity. We already checked that m_alias shouldn't be empty on line 54. So the idea is that if a component already has an alias and its already added to the entity, there's nothing to do here; return. 

I'll add a comment here to make this clearer.

Code/Framework/AzCore/AzCore/Component/Component.cpp
srikappa-amzn: 2 issues primarily:
1. It's being used both when adding a component and removing a component. And even for adding, it's being called twice from EditorEntityActionComponent. So it was getting harder to test that the function only sets the alias once correctly
2. Since it's a public function, there are many calls to this already in the code base https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/o3de/o3de%24+SetEntity&patternType=standard&sm=1&groupBy=path . We don't want components to call this manually to set an alias. So it felt like we are trying to piggyback our required behavior (to set a valid alias) on to an existing call that seems to have many use cases for many components.

srikappa-amzn:
> Very nice to change to have type name as the component alias.
> 
> question: Since this change is going into dev and the outliner override management is already in stab/2305, do we need to think about adding migration support for the old override patches? In other words, in this release, users will start making override patches in prefab file:
> 
> ```json
> {
>     "op": "replace",
>     "path": "/Entities/Entity_[16218641747988]/Components/Component_[3813021942463140055]/IsStatic",
>     "value": true
> }
> ```
> 
> Is it still recognized by prefab override handler after this PR? I assume the new path is like this:
> 
> ```
> /Entities/Entity_[16218641747988]/Components/EditorTagComponent/IsStatic
> ```

Yes, overrides will continue to work even with old component aliases since we are keeping aliases read from prefabs intact. Regarding merging this PR to stabilization, I think it's a good idea as we will have more user data created with typename aliases, which will be helpful down the road if we decide to enforce the rule that an alias can only be the typename. But we can revisit that discussion once this is merged to development and things look stable for a while.

Code/Framework/AzCore/AzCore/Component/Component.cpp
lumberyard-employee-dm: Thanks for the explanation. I understand the purpose of this function much better now.

How about given it a name that is more inline how we name similar functions in the code base.
Such as `OnBeforeEntityAdded` or `PreAddEntity`?

lumberyard-employee-dm: That is fine.

Code/Framework/AzToolsFramework/Tests/CustomSerializeContextTestFixture.h
```diff
@@ -0,0 +1,50 @@
...
+        AZ::BehaviorContext* GetBehaviorContext() override;
+        AZ::JsonRegistrationContext* GetJsonRegistrationContext() override;
+        const char* GetEngineRoot() const override;
+        const char* GetExecutableFolder() const override;
+        void EnumerateEntities(const EntityCallback& /*callback*/) override;
+        void QueryApplicationType(AZ::ApplicationTypeQuery& /*appType*/) const override;
+        ////
+    protected:
+        AZStd::unique_ptr<AZ::SerializeContext> m_serializeContext;
+    };
```
lumberyard-employee-dm, line 49: Instead of making a fixture for the SerializeContext.
The MockComponentApplication can be used and the GetSerializeContext function for it can just return the

srikappa-amzn: Yes, I like that. Makes the code concise. I've left the new test fixture in place though to prevent code duplication, as this is needed in 3 test fixtures so far and that number may grow in the future.

lumberyard-employee-dm (approved):

forkercat (approved):

Issue #15246, opened 2023-03-21
DPE/Prefab Overrides: Override edit on container/list property does not work


forkercat:
**Describe the bug**

Feature Flags: Enable DPE and inspector override management

Making an override edit on container does not work. The data is actually not updated. Undo node is skipping the operation.

<img height=120 src="https://user-images.githubusercontent.com/11157226/226711609-8b7d7476-2d08-4598-aae9-8fb65e3cb66d.png">

Note: The visualization is showing up because the code sets it to true.

**Steps to reproduce**
1. Create a new prefab with an entity inside
2. Focus on the prefab and create a Tag component, then exit the focus
3. Make sure level prefab is being focused
4. Click "+" button to add a new element and change it value

**Expected behavior**

The element is being added as an override.

**Actual behavior**

No override edit is performed.

**Screenshots/Video**

https://user-images.githubusercontent.com/11157226/226712442-3f0d5f11-266c-4fe6-9e06-a5fbdc6ddc99.mp4

**Found in Branch**

Development https://github.com/o3de/o3de/commit/9c369f68def99c45a8b79c0e2f98a2438bf9d15b

**Desktop/Device**
* Device: PC
* OS: Windows 10
* CPU: AMD Ryzen 5950x
* GPU: NVIDIA RTX 3070
* Memory: 64 GB

**Additional context**

I remembered it worked before the changes of using DPE patches, i.e. using dirty entity mechanism.

forkercat:
So, the root cause is the current value DOM from the focused template in this case is null, the current workflow skips further processing.

Issue #15242, opened 2023-03-21
DPE/Prefab Overrides: Changing Y would clear X value in vector-type property


forkercat:
**Describe the bug**

Feature Flags: Enable DPE and inspector override management

For a container-type property, if we change one individual value as an override edit, it would clear override edits on other individual values.

**Steps to reproduce**
1. Create a new prefab with an entity inside
2. Make sure level prefab is being focused
3. Change translate X to 3.0, and then change Y to 4.0.

**Expected behavior**

The X value should not change.

**Actual behavior**

The X value is cleared.

**Screenshots/Video**

https://user-images.githubusercontent.com/11157226/226682319-feb3932e-ee4c-4540-89db-b45bb7e6b9fc.mp4

**Found in Branch**

Development https://github.com/o3de/o3de/commit/9c369f68def99c45a8b79c0e2f98a2438bf9d15b

**Desktop/Device**
* Device: PC
* OS: Windows 10
* CPU: AMD Ryzen 5950x
* GPU: NVIDIA RTX 3070
* Memory: 64 GB

**Additional context**

No

Issue #15241, opened 2023-03-21
DPE/Prefab Overrides: Visualization does not update properly if property is unchanged during editing


forkercat:
**Describe the bug**

Feature Flags: Enable DPE and inspector override management

If the property value is 3.0, I change it to 3.0 during editing. The visualization is not updated correctly.

**Steps to reproduce**
1. Create a new prefab with an entity inside
2. Make sure level prefab is being focused
3. Change translate X value to 3.0, then click an empty space in inspector to complete the editing
4. Repeat this operation again and again

**Expected behavior**

The visualization should not change. The override icon should stay.

**Actual behavior**

The icon would appear and disappear after this operation.

**Screenshots/Video**

https://user-images.githubusercontent.com/11157226/226679314-1337de61-659e-4161-9549-4e24ea44d06b.mp4

**Found in Branch**

Development https://github.com/o3de/o3de/commit/9c369f68def99c45a8b79c0e2f98a2438bf9d15b

**Desktop/Device**
* Device: PC
* OS: Windows 10
* CPU: AMD Ryzen 5950x
* GPU: NVIDIA RTX 3070
* Memory: 64 GB

**Additional context**

No

Issue #15243, opened 2023-03-21
DPE/Prefab Overrides: Undoing/Reverting property override does not work properly


forkercat:
**Describe the bug**

Feature Flags: Enable DPE and inspector override management

Undoing and reverting property override does not work properly. The data is changed, but the visualization is not updated correctly for the edit field.

**Steps to reproduce**
1. Create a new prefab with an entity inside
2. Make sure level prefab is being focused
3. Change translate X value to 3.0
4. Undo or revert the value

**Expected behavior**

The data is changed and the visualization should be updated correctly.

**Actual behavior**

The data is changed and the visualization is not updated correctly.

**Screenshots/Video**

https://user-images.githubusercontent.com/11157226/226685360-7b717756-e4de-45fa-8944-9deb0d24201a.mp4

**Found in Branch**

Development https://github.com/o3de/o3de/commit/9c369f68def99c45a8b79c0e2f98a2438bf9d15b

**Desktop/Device**
* Device: PC
* OS: Windows 10
* CPU: AMD Ryzen 5950x
* GPU: NVIDIA RTX 3070
* Memory: 64 GB

**Additional context**

- Interestingly, if the override edits are already present after loading a level, the override can be reverted and the visualization is updated correctly.
- Looks like in the case that the visualization is not updated correctly. The SetComponent or DoRefresh in DPE::ComponentAdapter is not invoked.

PR #15264, opened 2023-03-22
Prefab Overrides: Fix property-edit issues in DPE Entity Inspector

forkercat:
## What does this PR do?

_Feature is behind two flags: `ed_enableDPE` and `EnableInspectorOverrideManagement`_

This PR fixes couple of issues after we changed to adopt DPE patches in https://github.com/o3de/o3de/pull/14988.

Inspector Override RFC: https://github.com/o3de/sig-content/issues/112

**Change List:**
1. Fixes https://github.com/o3de/o3de/issues/15241
2. Fixes https://github.com/o3de/o3de/issues/15242
3. Fixes https://github.com/o3de/o3de/issues/15246 (this PR adds add-element support)
4. Adds logic of skipping update in undo node if property is not changed.
5. Improves handling of default override values:
    1. Override patch is created if property is changed to non-default value and won't go away even users change it back to default value.
    2. Override patch won't be generated if property is not changed to other value from default value.
6. Vector-type property is now stored as a vector rather than individual values as shown below.

```json
{
    "op": "replace",
    "path": "/Entities/Entity_[123]/Components/Component_[8896558300816993180]/Transform Data/Translate",
    "value": [0, 3, 0]
}
```

**Issues Not Fixed:**
1. Undo/redo and revert is not tested. It should be addressed in a different PR. See GHI: https://github.com/o3de/o3de/issues/15243
1. For dynamic container list, 'delete' does not work. From DPE, we are receiving index paths pointing at individual values. Still investigating how to make DPE to generate path to the whole value.

https://user-images.githubusercontent.com/11157226/226967528-043ea7c5-7136-435c-b5fe-2b0d6549fc93.mp4

## How was this PR tested?

Manually tested in editor. Will have a follow-up PR to add unit tests, but it should be done after undo/redo issues are fixed.

Here are some workflows I tested and verified patches stored in file.

**Translate vector in TransformComponent:**
1. Change value from [0, 0, 0], to [0, 0, 0] (no icon shows up)
1. Change value from [0, 0, 0], to [3, 0, 0] (icon shows up), to [3, 5, 0] (icon shows up), to [0, 5, 0] (icon shows up), to [0, 0, 0] (icon still shows up)
1. Change value from [0, 0, 0] to [3, 0, 0], then to [3, 0, 0]. The last edit is not valid. Undoing once will change value back to [0, 0, 0].
1. Focus on the owning prefab and change the value to [3, 0, 0]. Exit focus. Change to [5, 0, 0] (icon shows up), to [3, 0, 0] (icon still shows up).
1. Tested 'Tab' to complete editing. It works as expected.

**Container list in TagComponent:**
1. Tested adding new elements.
1. Tested updating values in new elements.

Also tested switch, slider, color picker, text, enum type properties.


Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoComponentPropertyOverride.cpp
```diff
@@ -73,62 +74,77 @@ namespace AzToolsFramework
...
-
-            m_overriddenPropertyPathFromFocusedPrefab /= relativePathFromOwningPrefab;
-
-            const TemplateId topTemplateId = link->get().GetTargetTemplateId();
-            const PrefabDom& topTemplateDom = m_prefabSystemComponentInterface->FindTemplateDom(topTemplateId);
+            // Configure paths from focused and top templates. The top template is the source template of the link.
+            AZ::Dom::Path pathToPropertyFromTop =
+                AZ::Dom::Path(PrefabInstanceUtils::GetRelativePathFromClimbedInstances(climbUpResult.m_climbedInstances, true)) /
+                relativePathFromOwningPrefab;
+            AZ::Dom::Path pathToPropertyFromFocused(PrefabDomUtils::InstancesName);
```
srikappa-amzn, line 81: nit: pathToPropertyFromFocusedPrefab to be clearer

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoComponentPropertyOverride.cpp
```diff
@@ -73,62 +74,77 @@ namespace AzToolsFramework
...
+            pathToPropertyFromFocused /= climbUpResult.m_climbedInstances.back()->GetInstanceAlias(); // top instance alias
+            pathToPropertyFromFocused /= pathToPropertyFromTop;
+
+            // Get the current state of property value from focused template.
+            const TemplateId focusedTemplateId = link->get().GetTargetTemplateId();
+            const PrefabDom& focusedTemplateDom = m_prefabSystemComponentInterface->FindTemplateDom(focusedTemplateId);
+            PrefabDomPath domPathToPropertyFromFocused(pathToPropertyFromFocused.ToString().c_str());
```
srikappa-amzn, line 88: same: domPathToPropertyFromFocusedPrefab

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoComponentPropertyOverride.cpp
```diff
@@ -73,62 +74,77 @@ namespace AzToolsFramework
...
-                // DOM value pointers can't be relied upon if the original DOM gets modified after pointer creation.
-                PrefabDomPath overriddenPropertyDomPath(pathFromFocusedPrefab.ToString().c_str());
-                const PrefabDomValue* overriddenPropertyDomInTopTemplate = overriddenPropertyDomPath.Get(topTemplateDom);
-
-                
-                PrefabDom overridePatches;
-                if (overriddenPropertyDomInTopTemplate)
+                // Get the default state of property value from top template (direct child prefab of focused template).
+                // Note: We can't get the default state from the owning template because top template might have stored overrides
+                // on the owning instance. Those overrides should be treated as default/base values from the focused template.
```
srikappa-amzn, line 107: Do you mean to say , 'we can't get the default state from the "focused template"' ? The comment is not clear imo and can be removed altogether as I don't see it clarifying things.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoUtils.h
```diff
@@ -17,6 +17,12 @@ namespace AzToolsFramework
     {
         namespace PrefabUndoUtils
         {
+            enum class PatchOperationType
+            {
+                Add,
+                Replace
+            };
```
srikappa-amzn, line 24: why not reuse https://github.com/o3de/o3de/blob/development/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Overrides/PrefabOverrideTypes.h#L15-L20 ?

srikappa-amzn (commented):
Looking great! Quite a few issues fixed, which is nice!

Unit tests in the future for this will be crucial to prevent regressions as clearly there are several intricate edge cases here.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoComponentPropertyOverride.cpp
forkercat: I think I can update it. I want to put a note here to address reader's question like "why can't we just use the source template from owning instance as the default state?". The reason is there could be the case that `topInstance != owningInstance` and topInstance has an override edit to change the default state from source.

forkercat:
Might need to merge from dev to resolve conflicts. Will run AR later.

srikappa-amzn (approved):

AMZN-daimini (approved):

PR #15267, opened 2023-03-22
Merge script manager from ASV to ScriptAutomation gem

jiaweig-amzn:
## What does this PR do?

Merge script manager from ASV to ScriptAutomation gem, including:
- Various additional script binding calls
- Imgui control
- Camera conrtol


Gems/ScriptAutomation/Code/Source/ImGui/ImGuiAssetBrowser.cpp
```diff
@@ -0,0 +1,468 @@
...
+
+    void ImGuiAssetBrowser::OnCatalogChanged(const AZ::Data::AssetId& assetId)
+    {
+        AZ::Data::AssetInfo assetInfo;
+        AZ::Data::AssetCatalogRequestBus::BroadcastResult(assetInfo, &AZ::Data::AssetCatalogRequestBus::Events::GetAssetInfoById, assetId);
+
+        if (m_includedAssetFilter && m_includedAssetFilter(assetInfo))
+        {
+            m_needsRefresh = true;
+        }
```
rgba16f, line 84: This doesn't set needs refresh if no asset filter is supplied, is that intentional?

Gems/ScriptAutomation/Code/Source/ImGui/ImGuiAssetBrowser.cpp
```diff
@@ -0,0 +1,468 @@
...
+                Utils::AssetEntry entry;
+                entry.m_path = assetInfo.m_relativePath;
+                entry.m_assetId = id;
+                entry.m_name = assetInfo.m_relativePath;
+
+                m_assets.push_back(entry);
+            }
+        };
+
+        auto endCB = []() {};
```
rgba16f, line 116: Does this function need to save the supplied shouldInclude function as the includedAssetFilter?

Gems/ScriptAutomation/Code/Source/Utils.cpp
```diff
@@ -0,0 +1,315 @@
...
+            static int32_t s_memCaptureEnabled = 0;
+
+            ProfileTelemetryRequestBus::Broadcast(&ProfileTelemetryRequestBus::Events::SetAddress, s_telemetryAddress, s_telemetryPort);
+
+            const MaskType fullCaptureMask = s_telemetryCaptureMask | (s_memCaptureEnabled ? AZ_PROFILE_CAT_TO_RAD_CAPFLAGS(MemoryReserved) : 0);
+            ProfileTelemetryRequestBus::Broadcast(&ProfileTelemetryRequestBus::Events::SetCaptureMask, fullCaptureMask);
+
+            ProfileTelemetryRequestBus::Broadcast(&ProfileTelemetryRequestBus::Events::ToggleEnabled);
+#endif // AZ_PROFILE_TELEMETRY
+        }
```
rgba16f, line 82: I think this can be removed, we don't support it anymore

Gems/ScriptAutomation/Code/Source/ScriptAutomationScriptBindings.cpp
```diff
@@ -6,307 +6,264 @@
...
-#ifndef _RELEASE //AZ_TracePrintf does nothing in release, ignore this call
-            auto func = [message]()
+            auto operation = [message]()
             {
                 AZ_TracePrintf("ScriptAutomation", "Script: %s\n", message.c_str());
             };
 
-            ScriptAutomationInterface::Get()->QueueScriptOperation(AZStd::move(func));
-#endif
+            ScriptAutomationInterface::Get()->QueueScriptOperation(AZStd::move(operation));
```
rgba16f, line 62: AZ_TracePrintF, AZ_Warning, & AZ_Error are all nop's in release, please re-add the #ifndef's

rgba16f (changes requested):

rgba16f:
Why are we doing this? It seems unnecessary.

jiaweig-amzn:
> Why are we doing this? It seems unnecessary.

The task is to have a sample level in AutomatedTesting to perform the same tests in ASV.
But ScriptAutomation doesn't have many necessary functionality, including: camera control, screenshot comparison reporting, friendly interface to execute a script.

rgba16f:
> > Why are we doing this? It seems unnecessary.
> 
> The task is to have a sample level in AutomatedTesting to perform the same tests in ASV. But ScriptAutomation doesn't have many necessary functionality, including: camera control, screenshot comparison reporting, friendly interface to execute a script.

I believe the aim is to have less code to maintain rather than bring all our possible code over from ASV. I'd like to lose all the IMGUI related classes as o3de already has a UI, and revert most of the Scripting Automation changes. The aim is not to make scripting automation a clone of ASV, it's to give it the minimum necessary features to enable writing tests.

PR #15327, opened 2023-03-23
Fix movement issues when client input id wraps around.

mbalfour-amzn:
## What does this PR do?

The ClientInputId is defined as a uint16 that is incremented every frame. At 60 fps, it takes ~18.2 minutes for this value to wrap around back to 0. However, the LocalPredictionPlayerInputComponent wasn't accounting for wraparound, so it had logic like this:
```
if (clientInputId <= m_lastClientInputId) { return; }
```
Once m_lastClientInputId reached 65535, the clientInputId would always be <= m_lastClientInputId, and no further inputs could get processed. This manifested as a bug in MultiplayerSample where the player could still move around locally and could still see other players moving around, but the server version of their player didn't move and nobody else saw them move. They also couldn't interact with the environment since the server version didn't move.

This changes the logic so that it correctly handles value wraparounds.

While testing, I also noticed that the logic for handling the first input array sent was a little bit fragile. "m_lastClientId" is always initialized to 0, so if the first client id received is 50, it would try to process 50 inputs, which means processing the 8th input 42 times, and then the other 7 inputs once. Now it initializes the m_lastClientId to the oldest ClientInputId in the array.

Note that the existing logic always assumed that the entire array is filled out with valid entries, so I've preserved that assumption. It seems like it _should_ look for invalid entries based on invalid host frame id values or something, but since the SendClientInput is always filling out the whole array, I guess it isn't necessary. It could always be done as a separate cleanup if desired though.

The changes to the test code are just some preliminary changes that will be needed when writing a unit test to test out this new logic. The unit test itself will be submitted in a separate PR once it's finished being written.

## How was this PR tested?

Ran the MultiplayerSample client and server with the m_clientInputId and m_lastCorrectionInputId locally initialized to 65000 instead of 0. Verified that the loss of movement occurred after about 30 seconds when the ID rolled over. Ran the same situation again with this fix and verified that movement continued to work through the rollover.

Ran the Multiplayer unit tests and verified they continued to pass.

### DIFF
--- a/Gems/Multiplayer/Code/Source/Components/LocalPredictionPlayerInputComponent.cpp
+++ b/Gems/Multiplayer/Code/Source/Components/LocalPredictionPlayerInputComponent.cpp
@@ -10,6 +10,7 @@
 #include <AzCore/Serialization/SerializeContext.h>
 #include <AzCore/Serialization/EditContext.h>
 #include <AzCore/std/smart_ptr/make_shared.h>
+#include <AzNetworking/ConnectionLayer/SequenceGenerator.h>
 #include <AzNetworking/Serialization/HashSerializer.h>
 #include <AzNetworking/Serialization/StringifySerializer.h>
 #include <Multiplayer/Components/NetworkHierarchyRootComponent.h>
@@ -188,18 +189,23 @@ namespace Multiplayer
             return;
         }
 
+        // After receiving the first input from the client, start the update event to check for slow hacking.
+        // Also initialize the lastClientInputId to one before the oldest available one in the inputArray so that
+        // we process everything available to us on the first call.
+        if (!m_updateBankedTimeEvent.IsScheduled())
+        {
+            // This subtraction intentionally wraps around.
+            m_lastClientInputId = inputArray[NetworkInputArray::MaxElements - 1].GetClientInputId() - ClientInputId(1);
+
+            m_updateBankedTimeEvent.Enqueue(sv_InputUpdateTimeMs, true);
+        }
+
         const ClientInputId clientInputId = inputArray[0].GetClientInputId();
-        if (clientInputId <= m_lastClientInputId)
+        if (!AzNetworking::SequenceMoreRecent(clientInputId, m_lastClientInputId))
         {
             AZLOG(NET_Prediction, "Discarding old or out of order move input (current: %u, received %u)",
                 aznumeric_cast<uint32_t>(m_lastClientInputId), aznumeric_cast<uint32_t>(clientInputId));
             return;
-        }
-
-        // After receiving the first input from the client, start the update event to check for slow hacking
-        if (!m_updateBankedTimeEvent.IsScheduled())
-        {
-            m_updateBankedTimeEvent.Enqueue(sv_InputUpdateTimeMs, true);
         }
 
         const AZ::TimeMs currentTimeMs = AZ::GetElapsedTimeMs();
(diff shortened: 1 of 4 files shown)


Gems/Multiplayer/Code/Source/Components/LocalPredictionPlayerInputComponent.cpp
```diff
@@ -188,20 +189,25 @@ namespace Multiplayer
             return;
         }
 
+        // After receiving the first input from the client, start the update event to check for slow hacking.
+        // Also initialize the lastClientInputId to one before the oldest available one in the inputArray so that
+        // we process everything available to us on the first call.
+        if (!m_updateBankedTimeEvent.IsScheduled())
+        {
+            // This subtraction intentionally wraps around.
+            m_lastClientInputId = inputArray[NetworkInputArray::MaxElements - 1].GetClientInputId() - ClientInputId(1);
```
kberg-amzn, line 198: I forget, is this input array always flushed out with events..? What do we do on the very first few inputs?

kberg-amzn: Double checked, we flush with the oldest element, so what you have works great 👍

Gems/Multiplayer/Code/Source/Components/LocalPredictionPlayerInputComponent.cpp
```diff
@@ -210,15 +216,16 @@ namespace Multiplayer
         m_lastInputReceived = inputArray;
         SetLastInputId(m_lastInputReceived[0].GetClientInputId()); // Set this variable in case of migration
 
-        while (m_lastClientInputId < clientInputId)
+        // Since id values can wrap around, we intentionally compare with a "!=" instead of a "<".
+        while (m_lastClientInputId != clientInputId)
```
kberg-amzn, line 220: This always makes me a little nervous, since if we have a double-increment during iteration for some reason this could loop forever.. if we wanted to be more defensive then we could possibly use a sequence check here as well to control loop termination.

kberg-amzn (approved):
Wow! What a great fix! 😄

Gems/Multiplayer/Code/Source/Components/LocalPredictionPlayerInputComponent.cpp
mbalfour-amzn: Yeah, it works fine the way it's written, I'm just not personally enamored of it because it's continuing to make an assumption that the API doesn't really enforce, that all entries are valid. For example, we've got unit tests that do _not_ fill out all the elements but also don't fail. I'd rather see this code stop making the assumption, either by using a fixed_vector instead of an array and checking .size(), or adding flags to say if a given entry is valid, or worst-case relying on the host frame id being initialized to invalid to detect it.

But none of this is required for this bugfix, so I'm leaving it as-is for now.

Gems/Multiplayer/Code/Source/Components/LocalPredictionPlayerInputComponent.cpp
mbalfour-amzn: I'd personally be _more_ worried about letting lastClientInputId grow larger than clientInputId with defensive code than about a bug causing an infinite loop. The infinite loop will produce an obvious symptom that needs to be fixed. Letting the lastClientInputId get too large will cause some number of inputs to get invisibly dropped and/or desync corrected.

PR #15331, opened 2023-03-23
[Terrain] Ray tracing terrain sectors

invertednormal:
## What does this PR do?

This PR removes the old rather prototypy ray tracing support in terrain and replaces it with ray traced geometry that should exactly match the geometry in the forward pass. This is done by copying the geometry data for each sector into formats that are digestible by ray tracing. Color is still not supported, so all of the terrain is assigned a base color of medium gray (0.18).

image
_(debug output from nsight graphics)_

## How was this PR tested?

Tested using several ray tracing scenes and observing the TLAS in nsight graphics.

### DIFF
--- a/Gems/Terrain/Code/Source/TerrainRenderer/TerrainMeshManager.cpp
+++ b/Gems/Terrain/Code/Source/TerrainRenderer/TerrainMeshManager.cpp
@@ -78,7 +78,6 @@ namespace Terrain
         AZ::RPI::ShaderSystemInterface::Get()->Connect(m_handleGlobalShaderOptionUpdate);
 
         AZ::RHI::Ptr<AZ::RHI::Device> rhiDevice = AZ::RHI::RHISystemInterface::Get()->GetDevice();
-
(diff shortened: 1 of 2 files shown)


galibzon (approved):

mbalfour-amzn (approved):

Gems/Terrain/Code/Source/TerrainRenderer/TerrainMeshManager.cpp
```diff
@@ -662,18 +748,27 @@ namespace Terrain
...
+            aznumeric_cast<uint32_t>(indices.size()),
+            indices.data());
 
-        struct VertexPosition
+        if (m_rayTracingEnabled)
         {
-            uint8_t m_posx;
-            uint8_t m_posy;
-        };
+            // Generate a 32 bit index buffer for ray tracing by copying and transforming the 16 bit index buffer.
```
dmcdiarmid-ly, line 756: 16-bit RayTracing index buffers are supported, so it should be possible to reuse the index buffer.

dmcdiarmid-ly (approved):

Issue #15330, opened 2023-03-23
Windows version of ProcessCommunicator does not actually block and wait for events in its WaitForReadyOutputs function


lawsonamzn:
**Describe the bug**
the WaitForReadyOutputs function in the ProcessCommunicator is implemented differently for each host platform (Linux, mac, windows, etc).  On other platforms, it uses a platform specific ability to wait for data to arrive in STDOUT/STDERR, for example, linux epoll.

However, on windows, it creates anonymous pipes, and then attempts to call WaitForMultipleObjects to wait for the pipes to have data.  But it is invalid to call WaitForMultipleObjects on pipes.  They are not a supported type.  This causes it to return immediately without waiting or yielding.   This means that any subprocess calls on windows will spin instead of sleep.

Note that this doesn't actually break anything in terms of functioning, but it means whenever an app is waiting for another app via this method, the host application will be using up a CPU core per sub child waiting in a busy loop.

**Assets required**
No sample assets are required.  This is a bug that can only be seen in the code.

**Steps to reproduce**
1. Build the project (in debug or in profile) on windows.
2. Ensure that the AzToolsFramework tests are your startup project.  Consider adding the filter `--gtest_filter=*ProcessLaunch*` to debug options to narrow the scope 
3. Run (add traceprints or just use breakpoints to examine things) and notice how `StdInOutProcessCommunicator::WaitForReadyOutputs` does not wait at all if they're pipes. 

**Expected behavior**
Just like on the other platforms some sort of wake event based or io completion port approach should be used to not waste cpu waiting for a child process.  Adding a sleep would not be okay becuase that may slow down child process execution.

**Actual behavior**
The calls to StdInOutProcessCommunicator::WaitForReadyOutputs should actually wait for output to be available if there is no output instead of instantly returning.  It should do so using event or async approaches (not by calling sleep or busy looping).

**Found in Branch**
Stabilization 2023

**Desktop/Device (please complete the following information):**
 - OS: Windows ONLY!
 
**Additional context**
This is not a critical problem, although, I did find it while investigating a deadlock.  I'll be posting a fix that fixes the deadlock, but not a fix that fixes this problem, as it would require a rewrite to use io completion ports or some other approach (threading, for example).  The deadlock is currently more important to fix than the wasted CPU. (sorry)

lawsonamzn:
Note that the above fix won't fix this problem, but will fix a deadlock that was discovered while investigating this.

PR #15333, opened 2023-03-23
Fix file handle starvation linux

### LINKED ISSUE #15332: Linux engine crashes at random due to having more than 1024 file handles open and calling FD_SET to buffer overflow
**Describe the bug**

Background:  If you need to wait for files to have data and you don't want to busy loop while waiting but would rather sleep and wake up when data is ready, one common approach is to use the system call select().

The select() system call on linux will sleep until any one of the file handles you feed it in a fd_set structure has data ready.  So to sleep until data is ready and wake up when it is,  you use the macro FD_SET() to populate a fd_set struct with the file handles you'd like to wait on, and then invoke select() on that fd_set, telling it to block until one of those file handles in the set has data.  It will stop using CPU on your thread (letting other stuff on the system have that compute power) until those handles are readable, at which point it will wake your thread up and you can read the data.

The FD_SET function, however, is a bitset, and in most linux systems, is hardcoded to have at most room for a file handle with an integer handle of up to 1023.  Example in openbsd:  But linux is similar.
https://github.com/openbsd/src/blob/fba4865f1dbe0cc6c4725437366d812456e9331d/sys/sys/select.h#L58

Using FD_SET(int) and giving an integer > 1024 will cause a buffer overflow, and apparmor/buffer overflow detection will kill your application.

O3DE's streamer settings by default cause it to keep 1024 file handles open.  file handles are globally shared amonst all applications in linux.  This means that not just in O3DE, but any other application that you run after starting O3DE, if that application is using FD_SET and select, that application will buffer overrun and crash!  Every file handle any application has open increases the file handle integer, it is a global resource.

We should fix this in 2 ways
1. futureproof it by using epoll instead of select() so that it wont crash even if a lot of handles are in use
2. reduce the number of handles o3de keeps open by default, perhaps to a reasonable 128 or 256, since this will destabilize other applications on linux if abused.

I have a fix for this, pending PR.  I will link it here. 

**Steps to reproduce**
1. Start O3DE and load a fairly complicated level.  I was using the ROS2 demo.
2. Cause O3DE to launch a child process (such as XACRO, or any other child process, for example, kill AP and let it restart AP for you).  Shader compiler, anything
**Expected behavior**
Everything works

**Actual behavior**
Your entire linux system is destabilized.  random applications that still use the old API will die.  Any part of O3DE that launches a subapplication (like the import robot dialog xacro invoker and waits for output) will crash and take the editor with it.

**Found in Branch**
Stabilization 2023

 - Device: Ubuntu linux

lawsonamzn:
## What does this PR do?
fixes #15332 and also fixes a deadlock discovered by fixing that fix, in windows.
The original bug was discovered here: https://github.com/o3de/o3de-extras/issues/118 and applies to o3de-extras, but the actual problem is in the o3de source code.

see the bug report above for a description of the problem.
For the windows deadlock, the same PR (this one) also fixes the windows deadlock which was always there, but the new test reveals.

Order that this happened in
1. discover the bug on linux and develop a fix using epoll, yay!
2. write a test to ensure its working correctly and that all the data is present and read regardless of flush/buffered
3. run the test all green on linux, push to branch!
4. test branch on windows and find a deadlock even though no windows code was modified at all.  turns out we had a deadlock all along!
5. Find a workaround for deadlock on windows.  File https://github.com/o3de/o3de/issues/15330 to cover a more long term fix for it.

## How was this PR tested?
Auto test, and manual test, on linux and windows.

### DIFF
--- a/Code/Framework/AzFramework/AzFramework/Process/ProcessCommunicator.cpp
+++ b/Code/Framework/AzFramework/AzFramework/Process/ProcessCommunicator.cpp
@@ -45,13 +45,15 @@ namespace AzFramework
     void ProcessCommunicator::ReadIntoProcessOutput(ProcessOutput& processOutput)
     {
         OutputStatus status;
-        char readBuffer[s_readBufferSize];
 
         // read from the process until the handle is no longer valid
         while (true)
         {
             WaitForReadyOutputs(status);
-            ReadFromOutputs(processOutput, status, readBuffer, s_readBufferSize);
+            if (status.shouldReadErrors || status.shouldReadOutput)
+            {
+                ReadFromOutputs(processOutput, status);
+            }
 
             if (!status.outputDeviceReady && !status.errorsDeviceReady)
             {
@@ -60,27 +62,26 @@ namespace AzFramework
         }
     }
 
-    void ProcessCommunicator::ReadFromOutputs(ProcessOutput& processOutput, OutputStatus& status, char* buffer, AZ::u32 bufferSize)
+    void ProcessCommunicator::ReadFromOutputs(ProcessOutput& processOutput, OutputStatus& status)
     {
         AZ::u32 bytesRead = 0;
+        // Send in the size + 1 to leave room for us to write out the 0 in
+        char readBuffer[s_readBufferSize+1];
 
         if (status.shouldReadOutput)
         {
-            // Send in the size - 1 to leave room for us to write out the 0 in
-            // bytesRead position on the next line
-            bytesRead = ReadOutput(buffer, bufferSize - 1);
-            buffer[bytesRead] = 0;
-            processOutput.outputResult.append(buffer, bytesRead);
+            bytesRead = ReadOutput(readBuffer, s_readBufferSize);
+            readBuffer[bytesRead] = 0;
+            processOutput.outputResult.append(readBuffer, bytesRead);
         }
 
         if (status.shouldReadErrors)
         {
-            // Send in the size - 1 to leave room for us to write out the 0 in
-            // bytesRead position on the next line
-            bytesRead = ReadError(buffer, bufferSize - 1);
-            buffer[bytesRead] = 0;
-            processOutput.errorResult.append(buffer, bytesRead);
+            bytesRead = ReadError(readBuffer, s_readBufferSize);
+            readBuffer[bytesRead] = 0;
+            processOutput.errorResult.append(readBuffer, bytesRead);
         }
+
     }
 
     AZ::u32 ProcessCommunicatorForChildProcess::BlockUntilInputAvailable(AZStd::string& readBuffer)
--- a/Code/Framework/AzFramework/AzFramework/Process/ProcessCommunicator.cpp
+++ b/Code/Framework/AzFramework/AzFramework/Process/ProcessCommunicator.cpp
@@ -141,6 +141,7 @@ namespace AzFramework
         m_stdInWrite->Close();
         m_stdOutRead->Close();
         m_stdErrRead->Close();
+        m_communicatorData.reset();
         m_initialized = false;
     }
 
--- a/Code/Framework/AzFramework/Platform/Linux/AzFramework/Process/ProcessCommunicator_Linux.cpp
+++ b/Code/Framework/AzFramework/Platform/Linux/AzFramework/Process/ProcessCommunicator_Linux.cpp
@@ -287,7 +287,7 @@ namespace AzFramework
         return true;
     }
 
-    void StdInOutProcessCommunicator::WaitForReadyOutputs(OutputStatus& status) const
+    void StdInOutProcessCommunicator::WaitForReadyOutputs(OutputStatus& status)
     {
         using namespace ProcessCommunicatorLinuxInternal;
 
--- a/Code/Framework/AzFramework/Platform/Linux/AzFramework/Process/ProcessCommunicator_Linux.cpp
+++ b/Code/Framework/AzFramework/Platform/Linux/AzFramework/Process/ProcessCommunicator_Linux.cpp
@@ -21,7 +21,7 @@ namespace ProcessCommunicatorLinuxInternal
 
     class CommunicatorDataImpl : public AzFramework::StdInOutProcessCommunicatorData
     {
-        public:
+    public:
         AZ_CLASS_ALLOCATOR(CommunicatorDataImpl, AZ::SystemAllocator);
         CommunicatorDataImpl() = default;
         virtual ~CommunicatorDataImpl()
@@ -80,7 +80,7 @@ namespace AzFramework
     {
         m_handle = handle;
         // also create an epoll handle
-        if ((handle != -1)&&(createEPollHandle))
+        if ( (handle != -1) && (createEPollHandle) )
         {
             int m_epollHandle = epoll_create1(0);
 
--- a/Code/Framework/AzFramework/Platform/Linux/AzFramework/Process/ProcessCommunicator_Linux.cpp
+++ b/Code/Framework/AzFramework/Platform/Linux/AzFramework/Process/ProcessCommunicator_Linux.cpp
@@ -67,13 +67,17 @@ namespace AzFramework
 
     void CommunicatorHandleImpl::Close()
     {
-        close(m_handle);
+        if (m_handle != -1)
+        {
+            close(m_handle);
+            m_handle = -1;
+        }
+
         if (m_epollHandle != -1)
         {
             close(m_epollHandle);
             m_epollHandle = -1;
         }
-        m_handle = -1;
     }
 
     bool CommunicatorHandleImpl::SetHandle(int handle, bool createEPollHandle)
@@ -84,7 +88,7 @@ namespace AzFramework
         {
             int m_epollHandle = epoll_create1(0);
 
-            AZ_Assert(m_epollHandle -1, "Failed to create epoll handle");
+            AZ_Assert(m_epollHandle -1, "Failed to create epoll handle. errno: %d", errno);
             if (m_epollHandle == -1) 
             {
                 return false;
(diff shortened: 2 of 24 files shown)


lawsonamzn:
Implementation notes
This PR consists of 2 fixes for the same system, one on linux, and one on windows.

On linux, the fix was done by creating an epoll handle for each of the stdout, stderr handles and storing it in the already available handle structure.  However, it also needed to create an epoll handle that represents BOTH of those, to use to wait on when you wanted to wake-on-first-available.  There was nowhere to store these, so a new platform-specific data structure to store this was created.  The default is empty/not installed on platforms that don't need it.

A test was written which spams 16k (twice the buffer size) to both stdout and stderr.

Once the system was converted to epoll the problem no longer presented itself.  As an extra precaution, the total number of file handles we force to keep open at all times is reduced to 128 in the settings registry.  feel free to increase or decrease if you're shipping a game.  Actual shipping games already had it trimmed down to 64.

However, the test started to deadlock on windows.

It turns out windows always had a deadlock capability here.  If any process wrote to both stdout and stderror enough data to fill the os buffers, it would deadlock on windows.   This is because you cannot actually call WaitForxxxxx on pipes on windows, it always immediately returns with 0 as the result.  We interpreted 0 as meaning 'stdout is ready for reading' and attempt to read stdout, which is empty, which causes a deadlock.  This would essentially cause a deadlock in windows of any child process that wrote only to stderr.

Code/Framework/AzFramework/Platform/Linux/AzFramework/Process/ProcessCommunicator_Linux.cpp
```diff
@@ -11,7 +11,32 @@
...
+
 #include <AzFramework/Process/ProcessCommunicator.h>
+#include <AzCore/Memory/SystemAllocator.h>
+
+namespace ProcessCommunicatorLinuxInternal
+{
+
+    class CommunicatorDataImpl : public AzFramework::StdInOutProcessCommunicatorData
+    {
+        public:
```
lumberyard-employee-dm, line 24: nitpick: public is normally unindented inside the class
```suggestion
    class CommunicatorDataImpl : public AzFramework::StdInOutProcessCommunicatorData
    {
    public:
```

Code/Framework/AzFramework/Platform/Linux/AzFramework/Process/ProcessCommunicator_Linux.cpp
```diff
@@ -38,12 +68,40 @@ namespace AzFramework
...
+        }
         m_handle = -1;
     }
 
-    void CommunicatorHandleImpl::SetHandle(int handle)
+    bool CommunicatorHandleImpl::SetHandle(int handle, bool createEPollHandle)
     {
         m_handle = handle;
+        // also create an epoll handle
+        if ((handle != -1)&&(createEPollHandle))
```
lumberyard-employee-dm, line 83: nitpick: whitespace would help with readability here.
```suggestion
        if ((handle != -1) && (createEPollHandle))
```

Code/Framework/AzFramework/Platform/Linux/AzFramework/Process/ProcessCommunicator_Linux.cpp
```diff
@@ -178,61 +238,104 @@ namespace AzFramework
...
-            if (status.errorsDeviceReady)
-            {
-                int currentHandle = m_stdErrRead->GetHandle();
-                FD_SET(currentHandle, &readSet);
-                maxHandle = currentHandle > maxHandle ? currentHandle : maxHandle;
-            }
-
-            if (select(maxHandle + 1, &readSet, nullptr, nullptr, nullptr) != -1)
+            struct epoll_event events[2];
+            if (int numberSignalled = epoll_wait(platformImpl->m_stdInAndOutPollHandle, events, 1, -1) > -1)
```
lumberyard-employee-dm, line 302: I don't think this function actually needed to be changed from `const`

A `const CommunicatorDataImpl*` could still access the `m_stdInAndOutPollHandle` int and work just fine.

lawsonamzn: it was changed due to windows, actually.  Windows uses Poll(Handle) which is non const cuz it could do handle->Break().

lumberyard-employee-dm: Ok, thanks for the clarification.

Code/Framework/AzFramework/Tests/ProcessLaunchTestTokens.h
```diff
@@ -0,0 +1,22 @@
...
+
+namespace ProcessLaunchTestInternal
+{
+    // the default buffer size for stdout/stdin on some platforms is 4k.  This number
+    // should just be higher than a couple multiples of buffer sizes to ensure that it overwhelms
+    // any buffer and reveals any deadlock caused by not reading from stdout/stderr when its full.
+    const size_t s_numOutputBytesInPlentyMode = 16 * 1024 ; 
+    const char* s_beginToken = "BEGIN";
+    const char* s_endToken = "END";
+    const char* s_midToken = "x";
```
lumberyard-employee-dm, line 21: extreme nitpick: these can be made `constexpr`
```suggestion
    constexpr size_t s_numOutputBytesInPlentyMode = 16 * 1024; 
    constexpr const char* s_beginToken = "BEGIN";
    constexpr const char* s_endToken = "END";
    constexpr const char* s_midToken = "x";
```

Code/Framework/AzFramework/Tests/ProcessLaunchMain.cpp
```diff
@@ -29,12 +31,54 @@ void OutputArgs(const AzFramework::CommandLine& commandLine)
...
+    int exitCode = 0;
     {
         AzFramework::CommandLine commandLine;
 
         commandLine.Parse(argc, argv);
-        OutputArgs(commandLine);
+
+        if (commandLine.GetNumSwitchValues("exitCode") > 0)
+        {
+            exitCode = atoi(commandLine.GetSwitchValue("exitCode").c_str());
```
lumberyard-employee-dm, line 45: [AZStd::from_chars](https://en.cppreference.com/w/cpp/utility/from_chars) can be used here from `#include<AzCore/std/utility/charconv.h>`
That at the very least returns an error code if something goes wrong.
```suggestion
            const AZStd::string& exitCodeStr = commandLine.GetSwitchValue("exitCode");
            AZStd::from_chars(exitCodeStr.begin(), exitCodeStr.end(), exitCode);
```

lumberyard-employee-dm (approved):

Code/Framework/AzFramework/AzFramework/Process/ProcessCommunicator.cpp
```diff
@@ -60,27 +62,26 @@ namespace AzFramework
...
         if (status.shouldReadOutput)
         {
-            // Send in the size - 1 to leave room for us to write out the 0 in
-            // bytesRead position on the next line
-            bytesRead = ReadOutput(buffer, bufferSize - 1);
-            buffer[bytesRead] = 0;
-            processOutput.outputResult.append(buffer, bytesRead);
+            bytesRead = ReadOutput(readBuffer, s_readBufferSize);
+            readBuffer[bytesRead] = 0;
+            processOutput.outputResult.append(readBuffer, bytesRead);
```
spham-amzn, line 75: Shouldn't we be reading from the buffer until `bytesRead` reaches zero?

Code/Framework/AzFramework/Platform/Linux/AzFramework/Process/ProcessCommunicator_Linux.cpp
```diff
@@ -38,12 +68,40 @@ namespace AzFramework
     void CommunicatorHandleImpl::Close()
     {
         close(m_handle);
+        if (m_epollHandle != -1)
```
spham-amzn, line 71: Should we check if `m_handle` is closed as well and do the similar logic?

Code/Framework/AzFramework/Platform/Linux/AzFramework/Process/ProcessCommunicator_Linux.cpp
```diff
@@ -38,12 +68,40 @@ namespace AzFramework
...
-    void CommunicatorHandleImpl::SetHandle(int handle)
+    bool CommunicatorHandleImpl::SetHandle(int handle, bool createEPollHandle)
     {
         m_handle = handle;
+        // also create an epoll handle
+        if ( (handle != -1) && (createEPollHandle) )
+        {
+            int m_epollHandle = epoll_create1(0);
+
+            AZ_Assert(m_epollHandle -1, "Failed to create epoll handle");
```
spham-amzn, line 87: may be worth it to add `errno` to the assert message here for troubleshooting
```suggestion
            AZ_Assert(m_epollHandle -1, "Failed to create epoll handle: error %d", errno);
```

Code/Framework/AzFramework/Platform/Linux/AzFramework/Process/ProcessCommunicator_Linux.cpp
lawsonamzn: We weren't before, and I've tried to touch as little as possible.  its not checking the result value of close, and you can't close handle -1.  So worst case this would just result in errno being set.

Code/Framework/AzFramework/Platform/Linux/AzFramework/Process/ProcessCommunicator_Linux.cpp
lawsonamzn: I'll add that, actually.

Code/Framework/AzFramework/AzFramework/Process/ProcessCommunicator.cpp
spham-amzn: Good point, this is called within a loop anyways

spham-amzn (approved):

PR #15344, opened 2023-03-23
Adding utility command of `createuuid` to SerializeContextTools

lumberyard-employee-dm:
The `createuuid` command outputs a uuid created from a string using the SHA1 hash algorithm that O3DE uses.

This is suitable for outputting Uuids used as part of AssetIds.

The command supports a `--quiet` option for suppressing output of the string used to generate the UUID.
The `--values=<string...>` option can be specified 1 or more times to specify strings to convert.
The `--with-curly-braces=<true|false>` option formats the Uuid with `{}` surrounding it
The `--with-dashes=<true|false>` option formats the Uuid with `-` after the 8th, 12th, 16th, 20th nibble in the Uuid
The `--output-file=<file>` option allows outputting the Uuid to a file instead of stdout.

## How was this PR tested?

Ran the `createuuid` command with the following arguments

`SerializeContextTools.exe createuuid --values="materials/types/basepbr.materialtype" --values="materials/types/enhancedpbr.materialtype" --values=engine.json`
image

`SerializeContextTools.exe createuuid --values="materials/types/basepbr.materialtype" --values="materials/types/enhancedpbr.materialtype" --values=engine.json --with-curly-braces=false --with-dashes=0`
image

`SerializeContextTools.exe createuuid --values="materials/types/basepbr.materialtype" --values="materials/types/enhancedpbr.materialtype" --values=engine.json --quiet`
image

### DIFF
--- a/Code/Tools/SerializeContextTools/Dumper.cpp
+++ b/Code/Tools/SerializeContextTools/Dumper.cpp
@@ -465,7 +465,7 @@ namespace AZ::SerializeContextTools
             {
                 AZ_Printf(
                     "createtype",
-                    R"(Unable to specified output-file "%s". Object will not be dumped)"
+                    R"(Unable to open specified output-file "%s". Object will not be dumped)"
                     "\n",
                     outputPath.c_str());
                 return false;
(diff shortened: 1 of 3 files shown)


forkercat (approved):

lumberyard-employee-dm:
> Is this compatible with the sidecar file based asset relocation work? Looking at the code here, the UUID you would generate by calling this won't necessarily match the actual source asset's UUID for a path given.

Well the Uuid generated here can be used to create GUIDs that match the Asset Id paths, but can be used for more than just AssetID GUID generation.

That is why it doesn't modify the string the user inputs and uses as is.

Code/Tools/SerializeContextTools/Dumper.cpp
```diff
@@ -596,6 +596,94 @@ namespace AZ::SerializeContextTools
...
+            {
+                outputPath = outputPathView.LexicallyNormal();
+            }
+
+            constexpr AZ::IO::OpenMode openMode = AZ::IO::OpenMode::ModeWrite | AZ::IO::OpenMode::ModeCreatePath;
+
+            if (auto& fileStream = outputStream.emplace<AZ::IO::SystemFileStream>(outputPath.c_str(), openMode);
+                !fileStream.IsOpen())
+            {
+                AZ_Printf(
```
cgalvan, line 625: Nit: Should we make this an error or warning instead?

Code/Tools/SerializeContextTools/Runner.cpp
```diff
@@ -93,6 +93,26 @@ namespace SerializeContextTools
         AZ_Printf("Help", R"(    example: 'createtype --type-name="AZ::Entity" --json-prefix="/My/Anchor"')" "\n");
         AZ_Printf("Help", R"(    example: 'createtype --type-name="AZ::Entity" --output-file=object.json)" "\n");
         AZ_Printf("Help", "\n");
+        AZ_Printf("Help", "  'createuuid': Create a UUID using a SHA1 hash from a string and output the contents to stdout or a file.\n");
+        AZ_Printf("Help", "    [arg] --values=<string...>: One or more strings to convert to UUID.\n");
```
cgalvan, line 97: Nit: Could also mention the delimiter in the help string

cgalvan (approved):

Code/Tools/SerializeContextTools/Dumper.cpp
lumberyard-employee-dm: Actually not in the SerializeContext tools.
When a non-zero return code is received from a command it redirect [output to stderr](https://github.com/o3de/o3de/blob/development/Code/Tools/SerializeContextTools/Runner.cpp#L185-L199), which is normal for a CLI application

Code/Tools/SerializeContextTools/Runner.cpp
lumberyard-employee-dm: Actually this is part of the CommandLine API.
Specifying multiple `--values` arguments or using a comma is the way to separate multiple of an [arugment](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Settings/CommandLine.h#L20-L30)

cgalvan (approved):

PR #15346, opened 2023-03-24
Fixes code generation for script-bind on network properties

kberg-amzn:
Fixes code generation for script-bind on network properties that should not be accesible from clients

## What does this PR do?
Fixes code generation when scriptbind is set to true on network properties that can only be written from a server target

### DIFF
--- a/Gems/Multiplayer/Code/Include/Multiplayer/AutoGen/AutoComponent_Source.jinja
+++ b/Gems/Multiplayer/Code/Include/Multiplayer/AutoGen/AutoComponent_Source.jinja
@@ -888,9 +888,9 @@ enum class NetworkProperties
 {%    endif %}
     })
 {%  if Property.attrib['Container'] == 'Vector' or Property.attrib['Container'] == 'Array' %}
-    ->Method("Set{{ UpperFirst(Property.attrib['Name']) }}", [](AZ::EntityId id, int32_t index, const {{ Property.attrib['Type'] }}& value) -> void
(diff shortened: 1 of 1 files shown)


galibzon (approved):

AMZN-Olex (approved):

Gems/Multiplayer/Code/Include/Multiplayer/AutoGen/AutoComponent_Source.jinja
```diff
@@ -914,10 +914,16 @@ enum class NetworkProperties
             return;
         }
 
+{%     if (ReplicateFrom == 'Authority' and ReplicateTo == 'Server') or (ReplicateTo == 'Client' and Property.attrib['IsPredictable'] | booleanTrue == false) %}
```
mbalfour-amzn, line 917: What does the `| booleanTrue == false` mean in this context? Just seeing it in the PR, it looks either meaningless (always false?) or horrifying (sometimes true?) ...

kberg-amzn: It pipes the IsPredictable property into a helper function that deduces whether it's a true value or not.. This just abstracts away all the numerous ways people might type a true value into a helper.. 'T' 't' 'True' 'true' 'TRUE' '1', etc..

mbalfour-amzn: Thanks for the explanation! I'm not familiar with our Jinja setup, so I didn't realize "booleanTrue" is a helper method (defined in AzAutoGen.py from the looks of it). The helper method name would've been easier for me to understand as a casual reader if it was called StringToBoolean or something, but that ship has long sailed. Anyways, I'll understand it going forward, thanks!

PR #15364, opened 2023-03-24
Re-factor the material_canvas_utils.py to use Graph and Node objects instead of module-level functions.

jromnoa:
## What does this PR do?
Re-factors MaterialCanvas bus calls for tests to be more object oriented by using the new `Node` and `Graph` class objects.

## How was this PR tested?
Passing test run using command `C:\git\o3de\python\runtime\python-3.10.5-rev1-windows\python\python.exe -m pytest -vv -s --build-directory="C:\git\o3de\build\bin\profile" C:\git\o3de\AutomatedTesting\Gem\PythonTests\Atom\TestSuite_Main_Null_Render_MaterialCanvas_01.py -k MaterialCanvas_Atom_BasicTests`:
```py
Starting test MaterialCanvas_BasicFunctionalityChecks_AllChecksPass...
Test MaterialCanvas_BasicFunctionalityChecks_AllChecksPass finished.
Report:
[SUCCESS] Success: Existing material graph opened successfully.
[SUCCESS] Success: Closed existing material graph.
[SUCCESS] Success: Closed all open material graphs.
[SUCCESS] Success: Verified all expected material graphs are opened.
[SUCCESS] Success: Verified all expected material graphs are opened.
[SUCCESS] Success: Verified all expected material graphs are opened.
[SUCCESS] Success: Verified all expected material graphs are opened.
[SUCCESS] Success: Verified all expected material graphs are opened.
[SUCCESS] Success: Closed all open material graphs except for one.
[SUCCESS] Success: Node Palette pane is visible.
[SUCCESS] Success: Verified material graph name is 'test1'
[SUCCESS] Success: New AZStd::shared_ptr<Node> object created in memory as world_position_node
[SUCCESS] Success: New AZStd::shared_ptr<Node> object created in memory as standard_pbr_node.
[SUCCESS] Success: Connected world_position_node to standard_pbr_node successfully.
Test result:  SUCCESS
```


AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
```diff
@@ -10,101 +10,101 @@
...
+        """
+        Creates a node in memory using _create_node_by_name then adds it to this Graph object.
+        :param node_name: String representing the type of node to add to the graph.
+        :param position: math.Vector2(x,y) value that determines where to place the node on the Graph's grid.
+        :return: None, but appends the new node to self.nodes list.
+        """
+        node_object = self._create_node_by_name(node_name)
+        node_id = graph.GraphControllerRequestBus(
+            bus.Event, GraphControllerRequestBusEvents.ADD_NODE, self.graph_id, node_object, position)
+        self.nodes.append(Node(node_object, node_id, node_name, self.graph_id, position))
```
gadams3, line 55: This should verify that the node was created. You should also be able to get the nodes from the graph after all of the functions I exposed. I'm not sure if you can get position but that should be added if you can't.

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
```diff
@@ -10,101 +10,101 @@
...
+
+    def __init__(self, node_object: object, node_id: int, node_name: str, graph_id: math.Uuid, position: math.Vector2):
+        self.node_object = node_object
+        self.node_id = node_id
+        self.node_name = node_name
+        self.graph_id = graph_id
+        self.position = position
+        self.outbound_slot = graph.GraphModelSlotId('outPosition')
+        self.inbound_slot = graph.GraphModelSlotId('inPositionOffset')
+
```
gadams3, line 71: These slots are specific to the nodes used in the test. Every node type has its own slots. You should also be able to get the slots from each node now that I exposed all of those graph model functions.

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
```diff
@@ -10,101 +10,101 @@
...
+    def connect_to_node_inbound_slot(self, target_node: Node) -> object:
+        """
+        Adds a new connection between this node's outbound slot to target_node's inbound slot.
+        :param target_node: A Node() object for the target node to connect to.
+        :return: azlmbr.object.PythonProxyObject representing a C++ AZStd::shared_ptr<Connection> object.
+        """
+        return graph.GraphControllerRequestBus(
+            bus.Event, GraphControllerRequestBusEvents.ADD_CONNECTION_BY_SLOT_ID, self.graph_id,
+            self.node_object, self.outbound_slot, target_node, target_node.inbound_slot)
```
gadams3, line 90: You should be able to create connections between any slots on different nodes but the system enforces that you're connecting between an output slot and an input slot. I would move this function to the graph class and pass in the source node and source slot and the target node and target slot. Basically what the add connection function is doing already.

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
```diff
@@ -10,101 +10,101 @@
...
+
+    def is_connected_to_node_inbound_slot(self, target_node: Node) -> bool:
+        """
+        Determines if self.node_object's outbound slot is connected to the target_node's inbound slot.
+        :param target_node: A Node() object for the target node to connect to.
+        :return: bool representing success (True) or failure (False)
+        """
+        return graph.GraphControllerRequestBus(
+            bus.Event, GraphControllerRequestBusEvents.ARE_SLOTS_CONNECTED, self.graph_id,
+            self.node_object, self.outbound_slot, target_node, target_node.inbound_slot)
```
gadams3, line 110: I would also move this to the graph class and use the signature that's already there on the bus. No reason to differentiate between inbound and outbound when they're doing the same thing.

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
```diff
@@ -10,101 +10,101 @@
...
+    """
+
+    def __init__(self, node_object: object, node_id: int, node_name: str, graph_id: math.Uuid, position: math.Vector2):
+        self.node_object = node_object
+        self.node_id = node_id
+        self.node_name = node_name
+        self.graph_id = graph_id
+        self.position = position
+        self.outbound_slot = graph.GraphModelSlotId('outPosition')
+        self.inbound_slot = graph.GraphModelSlotId('inPositionOffset')
```
smurly, line 70: should these be initialized to empty list of GraphModelSlotId. I think you'd need to call one of the new functions to populate these with actual slots on the node.

also I think you might need a class Slot which has attributes id, name, inbound: bool (or some way of indicating inbount/outbound, etc
slot.id can then be passed around and these could be list of Slot objects

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
```diff
@@ -10,101 +10,101 @@
...
+
+    def connect_to_node_outbound_slot(self, target_node: Node) -> object:
```
smurly, line 72: could be slightly clearer function name
def connect_inbound_to_target_outbound_slot(self, target_node: Node, target_slot: SlotId
or
def connect_to_target_node(self, self.Slot, target_node: Node, target_slot: SlotId

this might do well with a helper for GraphModelSlotId common name. Since each node has one or more inbound and outbound slots it may be unwieldy to just simply connect two nodes.

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
```diff
@@ -10,101 +10,101 @@
...
+        """
+        Adds a new connection between this node's inbound slot to target_node's outbound slot.
+        :param target_node: A Node() object for the target node to connect to.
+        :return: azlmbr.object.PythonProxyObject representing a C++ AZStd::shared_ptr<Connection> object.
+        """
+        return graph.GraphControllerRequestBus(
+            bus.Event, GraphControllerRequestBusEvents.ADD_CONNECTION_BY_SLOT_ID, self.graph_id,
+            self.node_object, self.inbound_slot, target_node, target_node.outbound_slot)
+
+    def connect_to_node_inbound_slot(self, target_node: Node) -> object:
```
smurly, line 82: again the names could be more suggestive like
def connect_to_target_node
def connect_from_target_node

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
```diff
@@ -10,101 +10,101 @@
...
+        """
+        Adds a new connection between this node's outbound slot to target_node's inbound slot.
+        :param target_node: A Node() object for the target node to connect to.
+        :return: azlmbr.object.PythonProxyObject representing a C++ AZStd::shared_ptr<Connection> object.
+        """
+        return graph.GraphControllerRequestBus(
+            bus.Event, GraphControllerRequestBusEvents.ADD_CONNECTION_BY_SLOT_ID, self.graph_id,
+            self.node_object, self.outbound_slot, target_node, target_node.inbound_slot)
+
+    def is_connected_to_node_outbound_slot(self, target_node: Node) -> bool:
```
smurly, line 92: def is_connected_from_node_slot(self, target_node: Node, target_slot: SlotId)
def has_connection_in_from_node_slot(
this probably needs to also specify which inbound slot since there may be more than one.

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
```diff
@@ -10,101 +10,101 @@
...
-        source_node, source_slot, target_node, target_slot)
+class Graph(object):
+    """
+    Opens the specified materialgraph file when inside MaterialCanvas which appears as a Graph.
+    Opening the Graph is required since it also sets the document_id for use with other functions.
+    """
+
+    def __init__(self, material_graph_file_path: str):
+        self.material_graph_file_path = material_graph_file_path
+        self.document_id = atom_tools_utils.open_document(self.material_graph_file_path)
```
smurly, line 26: do we need to expose document model CRUD on the Graph object or should the object model be centered on the material_graph_file_path as a docuemnt which contains a graph.
what if I want to create a new graph from scratch (new document)
can I Graph.save_as(document_name:str)
should I be able to call Graph.close() or Graph.delete()

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
```diff
@@ -10,101 +10,101 @@
...
+    def _create_node_by_name(self, node_name: str) -> object:
+        """
+        Creates a new node in memory matching the node_name string on the specified graph object.
+        i.e. "World Position" for node_name would create a World Position node.
+        :param node_name: String representing the type of node to add to the graph.
+        :return: azlmbr.object.PythonProxyObject representing a C++ AZStd::shared_ptr<Node> object.
+        """
+        return atomtools.DynamicNodeManagerRequestBus(
+            bus.Broadcast, DynamicNodeManagerRequestBusEvents.CREATE_NODE_BY_NAME, self.graph_object, node_name)
+
```
smurly, line 44: we could use
remove_node(Node)
remove_nodes([Node])
maybe a remove_all_nodes()

jromnoa: These are out of scope of the test and should be added when tests that test this functionality are added. I'm not trying to map everything in the re-factor, just have a starting point of what there is to work with for MaterialCanvas.

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
jromnoa: Out of scope, that would be a periodic test case and this re-factor isn't trying to add any new functionality, just re-factor the existing functionality to use the new classes.
Reason being is the extra testing and debugging time may impact completion.

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
jromnoa: I will try to map them using the info in MaterialCanvas

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
jromnoa: We like to split verification functions from the utils and only verify stuff in the test scripts themselves. If its incorrect it should fail on the test steps.

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
gadams3: To be more specific, I mean that these slot names only apply to the node created in your test and should not be referenced in the utilities.

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
gadams3: That's fine. I'm just saying that should not be added to the list if it is not a valid node unless you plan to crawl the list and check manually

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
```diff
@@ -8,103 +8,109 @@
...
+        self.position = position
+        self.slots = []
+
+    def add_slot(self, slot_name: str) -> None:
+        """
+        Adds a new Slot object to this Node to identify one of its existing slots.
+        :param slot_name: Name of the slot to lookup using graph.GraphModelSlotId().
+        :return: None, but adds the new slot to the self.slots list.
+        """
+        self.slots.append(Slot(slot_name))
```
gadams3, line 92: Exposed API does not support adding slots so I'm not sure what this is doing. Should be able to get all of the slot data from the node

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
```diff
@@ -8,103 +8,109 @@
...
+
+    def connect_slots(self, source_slot: Slot, target_node: (), target_slot: Slot) -> azlmbr.object.PythonProxyObject:
```
gadams3, line 94: This would be cleaner if it was moved to the graph class and mirrored the same parameters as the graph controller request bus

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
```diff
@@ -8,103 +8,109 @@
...
+        :param source_slot: A Slot() object for the source_node in a given Node() object.
+        :param target_node: A Node() object for the target node that holds the target_slot we are connecting to.
+        :param target_slot: A Slot() object for the slot we are connecting to from the source_node.
+        :return: azlmbr.object.PythonProxyObject representing a C++ AZStd::shared_ptr<Connection> object.
+        """
+        return graph.GraphControllerRequestBus(
+            bus.Event, GraphControllerRequestBusEvents.ADD_CONNECTION_BY_SLOT_ID, self.graph_id,
+            self.node_object, source_slot.id, target_node.node_object, target_slot.id)
+
+    def are_slots_connected(self, source_slot: Slot, target_node: (), target_slot: Slot) -> bool:
```
gadams3, line 106: This would be cleaner if it was moved to the graph class and mirrored the same parameters as the graph controller request bus

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
```diff
@@ -8,103 +8,142 @@
...
-    """
-    return graph.GraphControllerRequestBus(
-        bus.Event, GraphControllerRequestBusEvents.ARE_SLOTS_CONNECTED, graph_id,
-        source_node, source_slot, target_node, target_slot)
+class Node(object):
+    """
+    Represents a node inside of a graph and contains node related functions.
+    """
+
+    def __init__(self, node_python_proxy: azlmbr.object.PythonProxyObject, node_id: azlmbr.object.PythonProxyObject):
```
gadams3, line 23: you don't need to pass in the node id because you can get it from the node
it might be safest to only pass the document id and node id instead then always lookup the graph and node
either way

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
```diff
@@ -8,103 +8,142 @@
...
+
+class Graph(object):
+    """
+    Binds this Graph object to the opened graph matching the document_id passed to construct this object.
+    """
+
+    def __init__(self, document_id: str):
+        self.document_id = document_id
+
+    def get_graph_id(self) -> azlmbr.object.PythonProxyObject:
```
gadams3, line 53: graoh id is a uuid

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
```diff
@@ -8,103 +8,142 @@
...
+            bus.Event, GraphDocumentRequestBusEvents.GET_GRAPH_NAME, self.document_id)
+
+    def get_graph(self) -> azlmbr.object.PythonProxyObject:
+        """
+        Returns the AZStd::shared_ptr<Graph> object for this Graph object.
+        :return: azlmbr.object.PythonProxyObject containing a AZStd::shared_ptr<Graph> object.
+        """
+        return atomtools.GraphDocumentRequestBus(bus.Event, GraphDocumentRequestBusEvents.GET_GRAPH, self.document_id)
+
+    def get_nodes(self, graph_id: azlmbr.object.PythonProxyObject) -> list[azlmbr.object.PythonProxyObject]:
```
gadams3, line 76: don't need to pass graph id because class has get_graph_id

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
```diff
@@ -8,103 +8,142 @@
...
+        """
+        Gets the current Nodes in this Graph and returns them as a list of azlmbr.object.PythonProxyObject objects.
+        :param graph_id: azlmbr.object.PythonProxyObject containing an int value for the graph.
+        :return: list of azlmbr.object.PythonProxyObject objects each representing a Node in this Graph.
+        """
+        nodes = graph.GraphControllerRequestBus(bus.Event, GraphControllerRequestBusEvents.GET_NODES, graph_id)
+        return nodes
```
gadams3, line 83:
```suggestion
        returm graph.GraphControllerRequestBus(bus.Event, GraphControllerRequestBusEvents.GET_NODES, get_graph_id())
```

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
```diff
@@ -8,103 +8,142 @@
...
+
+    def create_node_by_name(
+            self, graph: azlmbr.object.PythonProxyObject, node_name: str) -> azlmbr.object.PythonProxyObject:
```
gadams3, line 86: don't need to pass graph because class has get_graph

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
```diff
@@ -8,103 +8,142 @@
...
+        :param node_name: String representing the type of node to add to the graph.
+        :return: azlmbr.object.PythonProxyObject representing a C++ AZStd::shared_ptr<Node> object.
+        """
+        return atomtools.DynamicNodeManagerRequestBus(
+            bus.Broadcast, DynamicNodeManagerRequestBusEvents.CREATE_NODE_BY_NAME, graph, node_name)
+
+    def add_node(self,
+                 graph_id: azlmbr.object.PythonProxyObject,
+                 node: azlmbr.object.PythonProxyObject,
+                 position: math.Vector2) -> Node:
```
gadams3, line 101: don't need to pass graph id because class has get_graph_id

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
```diff
@@ -8,103 +8,142 @@
...
+        node_id = graph.GraphControllerRequestBus(
+            bus.Event, GraphControllerRequestBusEvents.ADD_NODE, graph_id, node, position)
+        return Node(node, node_id)
+
+    def add_connection_by_slot_id(self,
+                                  graph_id: azlmbr.object.PythonProxyObject,
+                                  source_node: Node,
+                                  source_slot: graph.GraphModelSlotId,
+                                  target_node: Node,
+                                  target_slot: graph.GraphModelSlotId) -> azlmbr.object.PythonProxyObject:
```
gadams3, line 118: don't need to pass graph id because class has get_graph_id

AutomatedTesting/Gem/PythonTests/Atom/atom_utils/material_canvas_utils.py
```diff
@@ -8,103 +8,142 @@
...
+        return graph.GraphControllerRequestBus(
+            bus.Event, GraphControllerRequestBusEvents.ADD_CONNECTION_BY_SLOT_ID, graph_id,
+            source_node.node_python_proxy, source_slot, target_node.node_python_proxy, target_slot)
+
+    def are_slots_connected(self,
+                            graph_id: azlmbr.object.PythonProxyObject,
+                            source_node: Node,
+                            source_slot: graph.GraphModelSlotId,
+                            target_node: Node,
+                            target_slot: graph.GraphModelSlotId) -> bool:
```
gadams3, line 137: don't need to pass graph id because class has get_graph_id

gadams3 (approved):

PR #15369, opened 2023-03-25
All decals no longer break if a decal material is missing

### LINKED ISSUE #15193: Bug Report - Decals stop working when one of the decal materials is removed
**Describe the bug**
If the material for one of the decals on the scene is removed, all decals disappear (after reloading the scene). This applies to existing, as well as new/modified decals. The only way to solve this is to re-assign the missing material. However, it may be not easy to find the decal with incorrect material, especially if there is a large number of decals on the scene. Typically, if the material for any entity is incorrect, it's marked red icon in the "Entity Inspector". But not in this case - the material is displayed normally. There are also no clear messages informing about the problem. Only the following warning message in the output when the level is loaded:
```
[Warning] (AssetManager) - GetAsset called for asset which does not exist in asset catalog and cannot be loaded. Asset may be missing, not processed or moved. AssetId: {D34B50C3-D2FD-519E-8F52-6D2FE4761477}:0
```

**Assets required**
Tested on default O3DE project

**Steps to reproduce**
1. Create a default project
2. Create 2 materials with bitmaps as a base color. Name them "decal1_mat" and "decal2_mat"
3. On the scene create 2 entities "decal1" and "decal2". Create a decal component for each and assign decal materials respectively
4. Place both decals in the way, the bitmaps are visible (eg. on the floor)
5. Delete "decal1_mat" using "Asset Browser" or file manager
6. Restart O3DE and load the level

Result:
- "decal2" (with the correct material) is invisible. Changing material doesn't help.
- All new-created decals are invisible
- In the entity inspector of the "decal1" nothing indicates, that the material is missing

8fbad22e9545137ea45b9019941fe3fea63c0af6

Linux, Ubuntu 22.04

invertednormal:
## What does this PR do?

Previously if a single decal material was missing, all decals would break. In addition to being a bug, it also made it difficult to track down which was the offending decal. With this change, it simply warns for the missing material's asset id, and all other decals continue to work.

Fixes #15193

## How was this PR tested?

Tested using repro steps from original bug.

### DIFF
--- a/Gems/Atom/Feature/Common/Code/Source/Decals/DecalTextureArrayFeatureProcessor.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/Decals/DecalTextureArrayFeatureProcessor.cpp
@@ -549,15 +549,18 @@ namespace AZ
         {
             const auto materialAsset = QueueMaterialAssetLoad(materialId);
 
-            m_materialLoadTracker.TrackAssetLoad(handle, materialAsset);
-
             if (materialAsset.IsLoading())
             {
+                m_materialLoadTracker.TrackAssetLoad(handle, materialAsset);
                 AZ::Data::AssetBus::MultiHandler::BusConnect(materialId);
             }
             else if (materialAsset.IsReady())
             {
                 OnAssetReady(materialAsset);
+            }
+            else if (materialAsset.IsError())
+            {
+                AZ_Warning("DecalTextureArrayFeatureProcessor", false, "Unable to load material for decal. Asset ID: %s", materialId.ToString<AZStd::string>().c_str());
             }
             else
             {


pawelbudziszewski:
It is working as it should now. The only thing that could still be improved is how the missing/incorrect material is displayed in the component view. It could be grayed out or have an additional red icon (like in the case of mesh materials). Currently, incorrect material can't be distinguished from the correct one.

Example of incorrect material:
Screenshot from 2023-03-25 18-31-43

Gems/Atom/Feature/Common/Code/Source/Decals/DecalTextureArrayFeatureProcessor.cpp
```diff
@@ -549,16 +549,19 @@ namespace AZ
         {
             const auto materialAsset = QueueMaterialAssetLoad(materialId);
 
-            m_materialLoadTracker.TrackAssetLoad(handle, materialAsset);
-
             if (materialAsset.IsLoading())
             {
+                m_materialLoadTracker.TrackAssetLoad(handle, materialAsset);
```
moudgils, line 554: Maybe add a quick comment why this was moved here.

moudgils (approved):

cgalvan (approved):

Gems/Atom/Feature/Common/Code/Source/Decals/DecalTextureArrayFeatureProcessor.cpp
invertednormal: I don't think a comment in the code is necessary since anyone reading it won't know about the history. But basically, before it was tracking loads for assets that were already loaded, which isn't necessary.

pawelbudziszewski (approved):

Issue #15361, opened 2023-03-24
Scene Settings: Updating fbx with Actor group and Prefab group fails


LB-MichalRodzos:
**Describe the bug**
This issue occurs on Windows platform and Linux platform.

If fbx has Actor group and Prefab group created, any update of the fbx fails in the Scene Settings and Asset Processor. Removing one of the groups allows User to update changes without fail.

Errors in Scene Settings and Asset Processor are the same. Please refer to the attached screenshot for more information.

**Steps to reproduce**
1. In the Asset Browser, find any fbx that supports Actor groups (e.g. sphere.fbx).
2. Make a change in the Actor tab (e.g. add a new Actor group).
3. Make sure that in the Procedural Prefab tab, Prefab group exist (if necessary add one).
4. Save changes.

**Expected behavior**
Fbx is updated without any errors in Scene Settings and Asset Processor.

**Actual behavior**
Updating fbx results in errors and fail in Asset Processor.

**Screenshots/Video**
![update_fail](https://user-images.githubusercontent.com/86952318/227556537-e985e248-443e-4db0-9751-6f601c9ae158.png)

**Found in Branch**
Stabilization/2305 Installer build ( https://o3debinaries.org/stabilization-2305/Latest/Windows/o3de_installer.exe; https://o3debinaries.org/stabilization-2305/Latest/Linux/O3DE_latest.deb)
Development (90289a5)

**Desktop/Device:**

- Device: PC
- OS: Windows
- Version: 10 Pro
- CPU: AMD Ryzen 5 3600 6-Core Processor
- GPU: NVIDIA GeForce RTX 2060 SUPER
- Memory: 16 GB

amzn-mike:
Closing as fixed in stabilization

LB-JakubSkorupka:
Issue verified as fixed on build Stabilization/2305 installer.

PR #15379, opened 2023-03-27
Fix error when prefab generator skips generation

amzn-mike:
## What does this PR do?

When a prefab group exists in the manifest but generation is skipped due to a registry setting, it leaves behind an empty prefab group which causes exporting to fail later on since the group still exists.  This fix causes the group to be removed before skipping the entry and adds an info message to notify users that a registry setting caused generation to be skipped.

Fixes https://github.com/o3de/o3de/issues/15361

## How was this PR tested?

Followed repro steps in linked bug and verified asset processes correctly after fix.


cgalvan (approved):

amzn-mike:
I added some unit tests to verify this fix: https://github.com/o3de/o3de/pull/15385

I'll merge them in after this fix is merged

PR #15381, opened 2023-03-27
Local prediction unit tests for client input id wraparound

mbalfour-amzn:
## What does this PR do?

This adds unit tests for HandleSendClientInput and HandleSendClientInputCorrection to verify that they correctly handle the case where the ClientInputId wraps around to 0. 

In the process, also found and fixed another wraparound bug with HandleSendClientInputCorrection where the logic to ignore old corrections would fail if the corrections were sent more than 10 minutes apart because there's no way to reliably tell just from the input id which correction is older if one or more wraparounds occur. The fix is to also send the client input's host frame id with the correction and ensure that we're processing a later frame or an equal frame but a later input id.

## How was this PR tested?

Verified that the new unit tests failed when the fixes to the Handle* methods were reverted and succeeded when the fixes were reinstated.

### DIFF
--- a/Gems/Multiplayer/Code/Source/Components/LocalPredictionPlayerInputComponent.cpp
+++ b/Gems/Multiplayer/Code/Source/Components/LocalPredictionPlayerInputComponent.cpp
@@ -334,8 +334,11 @@ namespace Multiplayer
  #endif
 
                 // Send correction. Include both the latest client input host frame id and the latest client input id processed so that
-                // the client can ensure that it doesn't try to process out-of-order corrections. The client input id by itself
-                // can roll over its value too quickly to be useful for detecting out-of-order conditions.
+                // the client can ensure that it doesn't try to process out-of-order corrections. The client input id is a uint16, which
+                // can roll over in (65536 / 60 fps) which is < 20 minutes. If half that time or more passes between corrections,
+                // so if we only tried to rely on the client input id to detect out-of-order corrections, we wouldn't be able to tell if
+                // the difference is telling us that it's out of order or if a long time had passed. By sending the host frame id too,
+                // we can distinguish between the two cases.
                 SendClientInputCorrection(m_lastInputReceived[0].GetHostFrameId(), m_lastClientInputId, correction);
             }
         }
--- a/Gems/Multiplayer/Code/Source/Components/LocalPredictionPlayerInputComponent.cpp
+++ b/Gems/Multiplayer/Code/Source/Components/LocalPredictionPlayerInputComponent.cpp
@@ -333,8 +333,11 @@ namespace Multiplayer
                 }
  #endif
 
-                // Send correction
-                SendClientInputCorrection(m_lastClientInputId, correction);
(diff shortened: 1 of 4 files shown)


Gems/Multiplayer/Code/Source/Components/LocalPredictionPlayerInputComponent.cpp
```diff
@@ -333,8 +333,10 @@ namespace Multiplayer
                 }
  #endif
 
-                // Send correction
-                SendClientInputCorrection(m_lastClientInputId, correction);
+                // Send correction. Include both the latest client input host frame id and the latest client input id processed so that
+                // the client can ensure that it doesn't try to process out-of-order corrections. The client input id by itself
+                // can roll over its value too quickly to be useful for detecting out-of-order conditions.
```
AMZN-Gene, line 338: Maybe explain a bit further about ClientInputId is uint16_t, and so this value rolls over to 0 again if it reaches 2^16, which at 30 inputs a second, can happen after ~2000 seconds. etc etc

Gems/Multiplayer/Code/Source/Components/LocalPredictionPlayerInputComponent.cpp
```diff
@@ -418,25 +420,58 @@ namespace Multiplayer
...
+                    NET_Prediction,
+                    "Discarding old correction for client host frame %u input id %u, host frame is older than last processed correction.",
+                    aznumeric_cast<uint32_t>(inputHostFrameId),
+                    aznumeric_cast<uint32_t>(inputId));
+                return;
+            }
+            else
+            {
+                // It's possible to receive corrections where the host frame is identical but the client input ids are out of sequence
+                // if we sent multiple inputs in the same frame, the server received and processed them across multiple frames, and
```
AMZN-Gene, line 460: +1 for code comments

AMZN-Gene (approved):

Gems/Multiplayer/Code/Source/Components/LocalPredictionPlayerInputComponent.cpp
mbalfour-amzn: Sure thing! I've put those comments in other places already, but no harm in having it here too.

mbalfour-amzn: Changing it to this:
```
                // Send correction. Include both the latest client input host frame id and the latest client input id processed so that
                // the client can ensure that it doesn't try to process out-of-order corrections. The client input id is a uint16, which
                // can roll over in (65536 / 60 fps) which is < 20 minutes. If half that time or more passes between corrections,
                // so if we only tried to rely on the client input id to detect out-of-order corrections, we wouldn't be able to tell if
                // the difference is telling us that it's out of order or if a long time had passed. By sending the host frame id too,
                // we can distinguish between the two cases.
```

AMZN-Olex (approved):

PR #15409, opened 2023-03-28
Atom/qingtao/atom 18375

VickyAtAZ:
## What does this PR do?

MeshMotionVector pass optimization: only render moved or always dyanmic meshes to MeshMotionVector pass.
Note, to avoid update MeshDrawPacket frequent, if a model was moved once after it's initialized, then we always treat it as dynamic object and draw it to MeshMotionVector pass

If mesh instancing is enabled (cvar r_meshInstancingEnabled), if any model is dynamic then the draw packet it reference to would enable draw motion vector.

## How was this PR tested?

Editor. Using ImGui GPU profiler to observe gpu cost for MeshMotionVectorPass. Using ImGui pass tree tool to observe MeshMotionVectorPass's output image.

Example capture of before/after optimization
Before
after


rgba16f (approved):

Gems/Atom/RPI/Code/Source/RPI.Public/MeshDrawPacket.cpp
```diff
@@ -126,9 +150,36 @@ namespace AZ
...
             m_shaderOptions.clear();
         }
 
+        void MeshDrawPacket::SetEnableDraw(RHI::DrawListTag drawListTag, bool enableDraw)
+        {
+            if (drawListTag.IsNull())
+            {
+                return;
+            }
+            // set the bit to true (1) if it's enabled for drawing
```
amzn-tommy, line 163: nit: The comment states that the following code is going to set it to true if it's enabled, but that kind of implies that it will only set it to true, not false. But the code is also going to set it to false if it's disabled.

Maybe something like: // set the bit to true (1) if it's enabled for drawing, false otherwise

Or just remove the comment. I think the code here is self explanatory, and the comment in the header clearly describes what the function does. You could also copy that header comment here, since it's more clearly written.

amzn-tommy (approved):

Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
```diff
@@ -647,10 +651,35 @@ namespace AZ
...
-                modelData.m_cullBoundsNeedsUpdate = true;
-                modelData.m_objectSrgNeedsUpdate = true;
+                modelData.m_flags.m_cullBoundsNeedsUpdate = true;
+                modelData.m_flags.m_objectSrgNeedsUpdate = true;
                 modelData.m_cullable.m_flags = modelData.m_cullable.m_flags | m_meshMovedFlag.GetIndex();
 
+                // Only set m_dynamic flag if the model instance is initialized.
+                if (!modelData.m_flags.m_dynamic)
+                {
+                    modelData.m_flags.m_dynamic = (modelData.m_model && !modelData.m_flags.m_needsInit)?true:false;
```
invertednormal, line 661: nit: usually there's more spacing `) ? true : false;`

invertednormal (approved):
I think this is an improvement on what we have but would really like to see the ability for meshes to turn off their dynamic flag once they're no longer moving in a future PR. Especially in the editor, this will mean over time as the user moves things around more and more things will get flagged even though they're not moving frame to frame. Perhaps that work can come after there's a visibility scene representation of static vs dynamic objects. If a mesh was marked as static then it wouldn't ever draw motion vectors even if it moved.

VickyAtAZ:
> I think this is an improvement on what we have but would really like to see the ability for meshes to turn off their dynamic flag once they're no longer moving in a future PR. Especially in the editor, this will mean over time as the user moves things around more and more things will get flagged even though they're not moving frame to frame. Perhaps that work can come after there's a visibility scene representation of static vs dynamic objects. If a mesh was marked as static then it wouldn't ever draw motion vectors even if it moved.

I think during editing, the entities are often get re-activated when any of its properties changed which always triggers MeshDrawPacket got re-created. Although it might not be true just for transformation change. I will need to check it.

invertednormal (approved):

PR #15425, opened 2023-03-28
Simplify and improve prefab undo utility functions

forkercat:
## What does this PR do?

This PR simplifies and improves utility functions in 'PrefabUndoUtils'. They are used to encapsulate repeated code to generate undo/redo patches and update value in cached instance DOM.

The purpose of this PR is to remove duplicate code and avoid confusion in usage.

**Change List:**
- Keep functions for entity-level or property-level updates. Entity-level would call property-level version.
- Replaced `GenerateUpdateEntityPatch` by `GenerateAndAppendPatch`.
- Changed to initialize undo/redo patch arrays in member initialization list, rather than calling `SetArray` before usage.
- For path helper function, use 'Append' to 'Prepend' in naming, and use assertions to check if the provided patches is array type.

## How was this PR tested?

Does not alter workflow logic. Passed existing unit tests.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoBase.h
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoBase.h
@@ -36,15 +36,16 @@ namespace AzToolsFramework
             void virtual Redo(InstanceOptionalConstReference instanceToExclude);
 
         protected:
-            bool m_changed;
-            TemplateId m_templateId;
-
             PrefabDom m_redoPatch;
             PrefabDom m_undoPatch;
+
+            TemplateId m_templateId;
 
             InstanceToTemplateInterface* m_instanceToTemplateInterface = nullptr;
             InstanceDomGeneratorInterface* m_instanceDomGeneratorInterface = nullptr;
             PrefabSystemComponentInterface* m_prefabSystemComponentInterface = nullptr;
+
+            bool m_changed;
         };
     }
 }
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoBase.h
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoBase.h
@@ -36,6 +36,7 @@ namespace AzToolsFramework
             void virtual Redo(InstanceOptionalConstReference instanceToExclude);
 
         protected:
+            bool m_changed;
             TemplateId m_templateId;
 
             PrefabDom m_redoPatch;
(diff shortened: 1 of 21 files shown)


Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoBase.h
```diff
@@ -36,6 +36,7 @@ namespace AzToolsFramework
             void virtual Redo(InstanceOptionalConstReference instanceToExclude);
 
         protected:
+            bool m_changed;
```
srikappa-amzn, line 39: It's a good practice to keep datatypes of the small size like bools at the end of all member variables. This ensures that these small types can take advantage of padding and reduce the memory allocation for the objects of the class.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoUtils.cpp
```diff
@@ -16,16 +16,6 @@ namespace AzToolsFramework
...
         {
-            void AppendAddEntityPatch(
-                PrefabDom& patches,
-                const PrefabDomValue& newEntityDom,
-                const AZStd::string& newEntityAliasPath)
-            {
-                AZ_Assert(patches.IsArray(), "AppendAddEntityPatch - Provided patches should be an array object DOM value.");
-
-                AppendUpdateValuePatch(patches, newEntityDom, newEntityAliasPath, PatchType::Add);
-            }
```
srikappa-amzn, line 27: Having clear named functions like this can actually help readability. In fact,  I'd recommend having functions like these call the underlying 'AppendUpdateValuePatch' and 'AppendUpdateValuePatchByComparison' functions rather than trying to route all calls to those functions.

srikappa-amzn (approved):

michabr (commented):
Nice to see API simplification!

michabr (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Undo/PrefabUndoUtils.h
```diff
@@ -46,44 +37,33 @@ namespace AzToolsFramework
...
-
-            //! Update the entity in instance DOM with the provided entity DOM.
+            //! Update the value in instance DOM with the provided DOM value.
             //! @param instanceDom The given instance DOM.
-            //! @param entityDom The entity DOM that will be put in the instance DOM.
-            //! @param entityAliasPath The given alias path to the entity.
-            void UpdateEntityInInstanceDom(
+            //! @param domValue The DOM value that will be put in the instance DOM.
+            //! @param pathToValue The given path to the value.
+            void UpdateValueInInstanceDom(
```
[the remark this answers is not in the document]
forkercat: yeah. It is good to use 'prefab' over 'instance'.

PR #15432, opened 2023-03-28
Make ModelReloaderSystem support multiple MeshReload requests at once

srikappa-amzn:
## What does this PR do?

This PR fixes a bug where the ModelReloaderSystem is attaching a singleton event hander to every 'ModelReloader' object it creates. Since a handler can only be attached to a single event, this causes an assert and a crash.

## How was this PR tested?

Manually. Noticed this happening when a procprefab is modified and it reloads the meshes of the procprefab in the editor. Was hitting an assert and crashes before these changes. It's fixed now.


gadams3 (approved):

gadams3:
This should go to stabilization

Issue #13919, opened 2023-01-05
Material Canvas: Undo/redo actions cause selection of all nodes in graph


LB-KacperMetelski:
**Describe the bug**
Undoing or redoing an action selects all of the nodes added to the graph.

**Steps to reproduce**
1. Open Material Canvas.
2. Create three or more random nodes.
3. Press Ctrl-Z or CTRL+Y.

**Expected behavior**
Only the undo/redo action is executed.

**Actual behavior**
Every node in the graph is selected.

**Screenshots/Video**
https://user-images.githubusercontent.com/110163473/210789650-c274f2c3-e43d-4ac3-9204-75684ba882ef.mp4

**Found in Branch**
Development

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
af94a59

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Windows
 - Version: Windows 11 21H1
 - CPU: i7-17000F
 - GPU: RTX-3070 MSI
 - Memory: 32GB

LB-MichalRodzos:
Issue verified as fixed on build Stabilization/2305 (169ab99). On Development build (73e54c6) the issue still occurs on Windows platform.

PR #15434, opened 2023-03-29
Material Canvas: Save and restore graph element selection on load, undo, and redo

gadams3:
## What does this PR do?

This change implements graph model selection state serialization so that it can be saved and restored when graphs are loaded and after undo and redo operations. Prior to these changes, loading, undue, and reduce selected all nodes in the graph.

In order to accomplish this, rather than creating new components and injecting them into all of the graph canvas native types, graph controller creates a router for entity state save data to inject and retrieve selection and position data with the rest of the metadata.

Additionally, this change adds batching for drag and drop operations so that tracking multiple nodes from the node palette can be undone and redone as an atomic operation.

Resolves https://github.com/o3de/o3de/issues/13919

## How was this PR tested?

Tested by creating multiple combinations of nodes, dragging from the docked node palette as well as using the context menu node palette. Performed undo, redo, save, and open operations to validate that selection state and position are restored consistently in material canvas.

Doing additional testing in landscape canvas.

### DIFF
--- a/Gems/GraphCanvas/Code/Source/Components/SceneComponent.cpp
+++ b/Gems/GraphCanvas/Code/Source/Components/SceneComponent.cpp
@@ -341,15 +341,15 @@ namespace GraphCanvas
     {
         GRAPH_CANVAS_PROFILE_FUNCTION();
 
+        // Once we finalize the node, we want to release the undo state, and push a new undo.
+        ScopedGraphUndoBatch undoBatch(GetSceneId());
+
         m_spliceTimer.stop();
 
         m_displayStateStateSetter.ResetStateSetter();
 
         if (m_splicingNode.IsValid())
         {
-            // Once we finalize the node, we want to release the undo state, and push a new undo.
-            GraphModelRequestBus::Event(GetSceneId(), &GraphModelRequests::RequestUndoPoint);
-
             SceneRequestBus::Event(GetSceneId(), &SceneRequests::ClearSelection);
             SceneMemberUIRequestBus::Event(m_splicingNode, &SceneMemberUIRequests::SetSelected, true);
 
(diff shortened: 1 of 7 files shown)


cgalvan (approved):

Gems/GraphCanvas/Code/Source/Components/SceneComponent.cpp
```diff
@@ -1246,23 +1245,22 @@ namespace GraphCanvas
...
-            if (constructSaveData)
+            if (GraphUtils::IsNodeGroup(currentEntity->GetId()))
             {
+                GraphCanvasConstructSaveData* constructSaveData = aznew GraphCanvasConstructSaveData();
+                constructSaveData->m_constructType = ConstructType::NodeGroup;
                 EntitySaveDataRequestBus::Event(currentEntity->GetId(), &EntitySaveDataRequests::WriteSaveData, constructSaveData->m_saveDataContainer);
                 saveData->m_constructs.push_back(constructSaveData);
+                continue;
             }
         }
```
rgba16f, line 1265: Are there future planned changes here? Just wondering why the refactor to duplicate the call to the ebus?

rgba16f (approved):

Gems/GraphCanvas/Code/Source/Components/SceneComponent.cpp
gadams3: This particular change happened organically while trying to understand how selection was handled in the first place and how all of the save data was serialized. I was originally going to try to inject the selection data here but it was not going to catch all of the cases and would impact script canvas. Thankfully, since all of the items read and write data via a bus, I was able to use the bus router to connect to all of the graph items noninvasively. I could revert this block but I think it's cleaner. It would be even better to use a lambda and pass in the container, entity ID, and construct type for all three blocks.

PR #15440, opened 2023-03-29
Fixed Shader hot reload for FullscreenTrianglePass

msat-huawei:
## What does this PR do?

Shader hot reload did not work for the `FullscreenTrianglePass`. The `ShaderReloadNotificationBus` overrides did not actually reload the shader, but only called `UpdateSrg` for the old shader. The new code actually reloads the shader.

I also made the `ShaderReloadNotificationBus` overrides protected (formerly they were private), so derived passes can react to reloaded shaders.

## How was this PR tested?

Windows and Vulkan. Tested hot reload on a custom compute pass.


galibzon (approved):

rgba16f (approved):

gadams3 (approved):
we should prioritize removal or simplification of the shader reload notification bus

should this go to the stabilization branch?

jhmueller-huawei (approved):

PR #15448, opened 2023-03-29
Fix networked character physics hit volumes

mbalfour-amzn:
## What does this PR do?

This fixes the hit volumes on networked characters. Previously, the character physics didn't support rotation of the character controller, so the attached hit volumes always stayed unrotated, regardless of the visual character's rotation. With this fix, the hit volumes rotate with the networked characters.

Changes:
* Changed the NetworkCharacterComponent to set the rotation on the physics character.
* Changed the CharacterController to store a copy of the rotation and correctly apply it to the shadow body, and added a unit test to trivially validate that this works.
* Changed the NetworkHitVolumes to destroy themselves when the physics character is destroyed, so that they'll recreate correctly when a new physics character is created. (This fixes a teleportation bug with the hit volumes where they weren't associated correctly after the physics character was disabled / enabled)
* Added an assert to MultiplayerSystemComponent. There's an occasional crash related to corrupt net binding components ending up in the gathered entities list; this assert will hopefully catch the problem a little bit sooner than the crash to help with debugging if/when it happens again.

## How was this PR tested?

Ran the unit test before and after the changes to verify that it failed before and succeeded after.
Ran MPS before and after the changes and verified that the hit volumes now rotate correctly.

Before:
https://user-images.githubusercontent.com/82224783/228595750-6dc48ead-b1c5-40c5-95a0-a784397b979b.mp4

After:
https://user-images.githubusercontent.com/82224783/228595620-c32efc14-f14d-4f3b-9668-d4eb7dd9b3c1.mp4

### DIFF
--- a/Gems/Multiplayer/Code/Source/Components/NetworkCharacterComponent.cpp
+++ b/Gems/Multiplayer/Code/Source/Components/NetworkCharacterComponent.cpp
@@ -220,20 +220,23 @@ namespace Multiplayer
             Multiplayer::GetNetworkTime()->SyncEntitiesToRewindState(entitySweptBounds);
         }
 
-        if ((GetParent().m_physicsCharacter == nullptr) || (velocity.GetLengthSq() <= 0.0f))
+        if (GetParent().m_physicsCharacter != nullptr)
         {
-            return GetEntity()->GetTransform()->GetWorldTranslation();
+            GetParent().m_physicsCharacter->SetRotation(GetEntity()->GetTransform()->GetWorldRotationQuaternion());
+
+            if (velocity.GetLengthSq() > 0.0f)
+            {
+                GetParent().m_physicsCharacter->Move(velocity * deltaTime, deltaTime);
+                GetEntity()->GetTransform()->SetWorldTM(GetParent().m_physicsCharacter->GetTransform());
+                AZLOG(
+                    NET_Movement,
+                    "Moved to position %f x %f x %f",
+                    GetParent().m_physicsCharacter->GetBasePosition().GetX(),
+                    GetParent().m_physicsCharacter->GetBasePosition().GetY(),
+                    GetParent().m_physicsCharacter->GetBasePosition().GetZ());
+            }
         }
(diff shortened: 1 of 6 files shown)


greerdv (approved):

AMZN-Olex (approved):

Gems/Multiplayer/Code/Source/Components/NetworkCharacterComponent.cpp
```diff
@@ -220,20 +220,23 @@ namespace Multiplayer
...
+        if (GetParent().m_physicsCharacter != nullptr)
         {
-            return GetEntity()->GetTransform()->GetWorldTranslation();
+            GetParent().m_physicsCharacter->SetRotation(GetEntity()->GetTransform()->GetWorldRotationQuaternion());
+
+            if (velocity.GetLengthSq() > 0.0f)
+            {
+                GetParent().m_physicsCharacter->Move(velocity * deltaTime, deltaTime);
+                GetEntity()->GetTransform()->SetWorldTM(GetParent().m_physicsCharacter->GetTransform());
+                AZLOG(
```
AMZN-Gene, line 231: Log looks spammy, but surprised it’s always been there. I don’t recall seeing logs every time I TryMove. I guess leave it if it not a problem?

AMZN-Gene (approved):

Gems/Multiplayer/Code/Source/Components/NetworkCharacterComponent.cpp
kberg-amzn: It's a tagged log, you have to enable it. We have a whole bloom filter sitting outfront of the new log system that allows you to turn on and off logs at runtime through the console. Try running:
Logger.EnableLog NET_Movement
Works way better than putting in debug logs, commenting them out, deleting them, needing to add them back in next time you run into a problem, etc..

kberg-amzn (approved):
Excellent

hultonha (approved):
Nice!

PR #15457, opened 2023-03-29
Contains crash fixes + fixes related to Renderdoc capture

moudgils:
## What does this PR do?

- Fixes app crash when launched from Renderdoc with openxr enabled
- Fix a crash when calling VK::SetNameInternal with null object
- Add support to unregister XR if rhi is not vulkan essentially preventing a crash. 

## How was this PR tested?

Tested on openxr and Automated testing on PC and quest2

### DIFF
--- a/Gems/Atom/RPI/Code/Source/RPI.Public/RPISystem.cpp
+++ b/Gems/Atom/RPI/Code/Source/RPI.Public/RPISystem.cpp
@@ -30,6 +30,7 @@
 #include <Atom/RHI/Factory.h>
 #include <Atom/RHI/Device.h>
 #include <Atom/RHI.Reflect/PlatformLimitsDescriptor.h>
+#include <Atom/RHI/RHIUtils.h>
 #include <Atom/RHI/XRRenderingInterface.h>
 
 #include <AzCore/Interface/Interface.h>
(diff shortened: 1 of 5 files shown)


Gems/Atom/RPI/Code/Source/RPI.Public/RPISystem.cpp
```diff
@@ -83,12 +84,17 @@ namespace AZ
             if (m_xrSystem)
             {
                 AZ::RHI::ResultCode resultCode = m_xrSystem->InitInstance();
-                // Fail result code can happen if no compatible device is attached. UnRegister xr system if that happens
-                if (resultCode == AZ::RHI::ResultCode::Fail)
+                // Check if the result code is a Success. Failure can happen if no compatible device is attached.
+                // Check if the active rhi backend is not vulkan
+                // UnRegister xr system if any of the criteria is not met
+                if (resultCode == AZ::RHI::ResultCode::Fail || !AZ::RHI::IsVulkanRHI())
```
moudgils, line 90: if resultCode is success, but the RHI is not vulkan, it will throw out an error and simply unregister XR so that it doesnt crash.

rgba16f (approved):

Gems/Atom/RHI/Code/Include/Atom/RHI/Factory.h
```diff
@@ -130,6 +130,9 @@ namespace AZ
             //! Returns the default priority of the factory in case there's no priorities set in the FactoryManager.
             virtual APIPriority GetDefaultPriority() = 0;
 
+            //! Returns true if the factory supports XR api
+            virtual bool SupportXR() const = 0;
+
```
rgba16f, line 135: grammer nit: SupportsXR

rgba16f (approved):

PR #15466, opened 2023-03-29
Revert "Fix file handle starvation linux (#15333)"

lawsonamzn:
This reverts commit 0906af50592ba00be123f29632933f0c3a506d36 to see if it fixes stabilization

I was unable to reproduce the crash encountered by CI on either windows or linux, even using the same commit hash as the CI machine is using.  So I'm going to try reverting it to see if it resolves it.


lawsonamzn:
never mind.  The thing was dying due to a bad asset cache

PR #15469, opened 2023-03-29
Setting registry support for r_streamingImageMipBias

VickyAtAZ:
## What does this PR do?

Add streaming image mip bias setting to setting registry: "/O3DE/Atom/RPI/Initialization/ImageSystemDescriptor/SystemStreamingImagePoolMipBias"
Also add value syncing for image streaming cvars and settings. 

Now user can use following way to setup texture mip bias when start app:
1. setting registry  "/O3DE/Atom/RPI/Initialization/ImageSystemDescriptor/SystemStreamingImagePoolMipBias" 
2. use --r_streamingImageMipBias=x in application's command line
3. add the cvar to <project-root>/Cache/<platform>/user.cfg

## How was this PR tested?

Editor

### DIFF
--- a/Gems/Atom/RPI/Code/Source/RPI.Public/Image/ImageSystem.cpp
+++ b/Gems/Atom/RPI/Code/Source/RPI.Public/Image/ImageSystem.cpp
@@ -392,11 +392,15 @@ namespace AZ
             // Sync values from ImageSystemDescriptor back to the cvars
             // Note 1: we need the sync here because one instance of the cvars might be initialized early than setting registry,
             // so it can't be initialized properly. See cvar_r_streamingImagePoolBudgetMb_Init and cvar_r_streamingImageMipBias_Init
-            // Note 2: we need to use PerformCommand instead of assign value directly because of this issue https://github.com/o3de/o3de/issues/5537
-            AZ::CVarFixedString commandString = AZ::CVarFixedString::format("r_streamingImagePoolBudgetMb %" PRIu64, desc.m_systemStreamingImagePoolSize);
-            AZ::Interface<AZ::IConsole>::Get()->PerformCommand(commandString.c_str());
-            commandString = AZ::CVarFixedString::format("r_streamingImageMipBias %" PRIu16, desc.m_systemStreamingImagePoolMipBias);
-            AZ::Interface<AZ::IConsole>::Get()->PerformCommand(commandString.c_str());
+            // Note 2: we need to use PerformCommand instead of assign value directly because of this issue https://github.com/o3de/o3de/issues/5537            
+            AZ::IConsole* console = AZ::Interface<AZ::IConsole>::Get();
+            if (console)
+            {
+                AZ::CVarFixedString commandString = AZ::CVarFixedString::format("r_streamingImagePoolBudgetMb %" PRIu64, desc.m_systemStreamingImagePoolSize);
+                console->PerformCommand(commandString.c_str());
+                commandString = AZ::CVarFixedString::format("r_streamingImageMipBias %" PRId16, desc.m_systemStreamingImagePoolMipBias);
+                console->PerformCommand(commandString.c_str());
+            }
 
             const SystemImagePoolDescriptor systemStreamingPoolDescriptor{ desc.m_systemStreamingImagePoolSize, "ImageSystem::SystemStreamingImagePool" };
             const SystemImagePoolDescriptor systemAttachmentPoolDescriptor{desc.m_systemAttachmentImagePoolSize, "ImageSystem::AttachmentImagePool" };
--- a/Gems/Atom/RPI/Code/Source/RPI.Public/Image/ImageSystem.cpp
+++ b/Gems/Atom/RPI/Code/Source/RPI.Public/Image/ImageSystem.cpp
@@ -29,6 +29,7 @@
 #include <AzCore/Console/IConsole.h>
 #include <AzCore/Interface/Interface.h>
 #include <AzCore/Math/Color.h>
+#include <AzCore/Settings/SettingsRegistry.h>
 
 AZ_DECLARE_BUDGET(RPI);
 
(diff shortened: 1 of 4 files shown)


Gems/Atom/RPI/Registry/atom_rpi.setreg
```diff
@@ -6,7 +6,8 @@
                     "CommonSrgsShaderAssetPath": "shaders/sceneandviewsrgs.azshader",
                     "ImageSystemDescriptor": {
                         "SystemStreamingImagePoolSize": 0,
-                        "SystemAttachmentImagePoolSize": 0 
+                        "SystemStreamingImagePoolMipBias": 1,
```
mbalfour-amzn, line 9: Should this be 0 instead of 1?

mbalfour-amzn (approved):

Gems/Atom/RPI/Registry/atom_rpi.setreg
VickyAtAZ: you are right. I forgot to change it back from testing

AMZN-Olex (approved):

AMZN-alexpete (approved):

Gems/Atom/RPI/Code/Source/RPI.Public/Image/ImageSystem.cpp
```diff
@@ -347,6 +389,14 @@ namespace AZ
                 Data::AssetId m_assetId;
             };
 
+            // Sync values from desc back to the cvars
+            // Note: we need this code because one of the instance of the cvar might be initialized early than setting registry
+            // due to this issue: https://github.com/o3de/o3de/issues/5537
```
lumberyard-employee-dm, line 394: This issue has nothing to do with the Settings Registry and more to do with the CVar value being set in the data space of the loaded Gem modules at the time.
For example if the CVar is set when the `Atom_RPI.Editor` module is loaded, but before the `EMotionFX_Atom.Editor` module is loaded, then only the static variable copy of `r_streamingImagePoolBudgetMb` inside of `Atom_RPI.Editor` is set and not the static variable of `r_streamingImagePoolBudgetMb` within the `EMotionFX_Atom.Editor` module.

So performing a direct access of the `r_streamingImagePoolBudgetMb` could result in two different values being returned depending on which gem module the code is executing within.

Gems/Atom/RPI/Code/Source/RPI.Public/Image/ImageSystem.cpp
```diff
@@ -347,6 +389,14 @@ namespace AZ
...
+            AZ::CVarFixedString commandString = AZ::CVarFixedString::format("r_streamingImagePoolBudgetMb %d", desc.m_systemStreamingImagePoolSize);
```
lumberyard-employee-dm, line 395: The percent specifier here needs to be "%" PRiu64 since `m_systemStreamingImagePoolSize` is a [uint64_t](https://en.cppreference.com/w/c/types/integer#Format_macro_constants)
```suggestion
            AZ::CVarFixedString commandString = AZ::CVarFixedString::format("r_streamingImagePoolBudgetMb %" PRIu64, desc.m_systemStreamingImagePoolSize);
```

Gems/Atom/RPI/Code/Source/RPI.Public/Image/ImageSystem.cpp
```diff
@@ -347,6 +389,14 @@ namespace AZ
...
+            AZ::Interface<AZ::IConsole>::Get()->PerformCommand(commandString.c_str());
+            commandString = AZ::CVarFixedString::format("r_streamingImageMipBias %d", desc.m_systemStreamingImagePoolMipBias);
```
lumberyard-employee-dm, line 397: The format specifier here needs to be `"%" PRIu16`: https://en.cppreference.com/w/c/types/integer#Format_macro_constants
```suggestion
            commandString = AZ::CVarFixedString::format("r_streamingImageMipBias %" PRIu16, desc.m_systemStreamingImagePoolMipBias);
```

lumberyard-employee-dm (changes requested):

Gems/Atom/RPI/Code/Source/RPI.Public/Image/ImageSystem.cpp
```diff
@@ -37,6 +38,31 @@ namespace AZ
 
     namespace
     {
+        const char* MemoryBudgetSettingPath = "/O3DE/Atom/RPI/Initialization/ImageSystemDescriptor/SystemStreamingImagePoolSize";
+        const char* MipBiasSettingPath = "/O3DE/Atom/RPI/Initialization/ImageSystemDescriptor/SystemStreamingImagePoolMipBias";
+        
+        size_t cvar_r_streamingImagePoolBudgetMb_Init()
+        {
+            size_t value = 0;
```
lumberyard-employee-dm, line 46: This type needs to be a `AZ::u64` in order to be read on Linux.
The SettingsRegistry only supports reading an integer into an `AZ::u64` or `AZ::s64`

Gems/Atom/RPI/Code/Source/RPI.Public/Image/ImageSystem.cpp
VickyAtAZ: I added initialization function to set the default value from the loaded setting registry. So it shouldn’t need to set values here again and I intended to remove these lines here. But somehow when the first time the init function get called, the setting registry wasn’t created. So it won’t get the value.
My comment is for this use case but not for using the performcommand function vs directly assign

lumberyard-employee-dm: Thank you for the clarification here.

VickyAtAZ: I wanted to use the GHI to explain why there might be multiple instances of the same cvar.  I may explain a bit more about the situation and remove the GHI since it may confuse people

lumberyard-employee-dm (approved):

HogJonny-AMZN (approved):

PR #15470, opened 2023-03-29
Organizes SingleTests by test case name and saves game/server logs

evanchia-ly-sdets:
## What does this PR do?
Creates sub artifact folders for SingleTests. Also saves Game.log and Server.log if they exist.

## How was this PR tested?
Tested by running SingleTests locally and checking logs.

### DIFF
--- a/Tools/LyTestTools/ly_test_tools/o3de/multi_test_framework.py
+++ b/Tools/LyTestTools/ly_test_tools/o3de/multi_test_framework.py
@@ -898,8 +898,15 @@ class MultiTestSuite(object):
             return_code = executable.get_returncode()
             executable_log_content = log_content_function(run_id, log_name, workspace)
             # Save the executable log.
+            workspace.artifact_manager.set_dest_path(request.node.originalname)
             workspace.artifact_manager.save_artifact(
                 os.path.join(log_path_function(run_id, workspace), log_name), f'({run_id}){log_name}')
+            server_log_name = os.path.join(workspace.paths.project_log(), 'Server.log')
+            game_log_name = os.path.join(workspace.paths.project_log(), 'Game.log')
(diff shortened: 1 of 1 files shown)


Tools/LyTestTools/ly_test_tools/o3de/multi_test_framework.py
```diff
@@ -898,8 +898,15 @@ def _exec_single_test(self,
...
             # Save the executable log.
+            workspace.artifact_manager.set_dest_path(request.node.originalname)
             workspace.artifact_manager.save_artifact(
                 os.path.join(log_path_function(run_id, workspace), log_name), f'({run_id}){log_name}')
+            server_log_name = os.path.join(workspace.paths.project_log(), 'Server.log')
+            game_log_name = os.path.join(workspace.paths.project_log(), 'Game.log')
+            if os.path.exists(server_log_name) and os.path.getsize(server_log_name):
+                workspace.artifact_manager.save_artifact(server_log_name)
+            if os.path.exists(game_log_name) and os.path.getsize(game_log_name):
+                workspace.artifact_manager.save_artifact(game_log_name)
```
cgalvan, line 909: Might be worth moving the `os.path.getsize` check into the `save_artifact` method itself so that any usage of that API would ignore an empty file. Unless it might be valuable to see if a file was there but empty?

cgalvan (approved):

Kadino (approved):

Tools/LyTestTools/ly_test_tools/o3de/multi_test_framework.py
evanchia-ly-sdets: Yeah I don't mind leaving the check outside in case of some log stomping issues. The reason why I'm checking here is because there was already something intentionally creating blank Server & Game logs.

Issue #15360, opened 2023-03-24
Project Manager: Change of project's engine is not visible under the project tile in the projects list


LB-KacperKapusta:
**Describe the bug**
Changing engine version in the Project Manager for a project does not update the project tile in the projects list. 
Note that the issue cannot be verified on Linux due to the reported [#15357](https://github.com/o3de/o3de/issues/15357).

Please note that upon launching the 2nd instance of the engine (2.0.0) installed, all projects created using the first one (1.2.0) change to 2.0.0 in project tile. The projects fail to launch until rebuilt (there is no information about the rebuild being required).

Please refer to the attached screenshot and video for more information.

**Steps to reproduce**
1. Download and install the O3DE Stabilization/2305 Installer (https://o3debinaries.org/stabilization-2305/Latest/Windows/O3DE_latest.deb).
2. Launch the O3DE from the _/opt/O3DE/1.2.0/bin/Windows/profile/Default_ directory.
3. Create the project, build it and launch the Editor verifying it works.
4. Close the Editor and the Project Manager.
5. Download and install the O3DE Development Installer (https://o3debinaries.org/development/Latest/Windows/O3DE_latest.deb).
6. Launch the O3DE from the _/opt/O3DE/2.0.0/bin/Windows/profile/Default_ directory.
7. Observe the project tile.
8. Press the hamburger icon next to the project created in Step 3. and select Edit Project Settings.
9. Change Engine entry and press save.
10. Observe the project tile.

**Expected behavior**
Project tile displays the version of the O3DE it has been built in. Changing the Engine entry for the project updates the engine version in the project tile and it becomes yellow if it is different than Project Manager version.

**Actual behavior**
Project tile displays the version of the O3DE that has been last built. Changing the Engine entry for the project does not update the engine version in the project tile and it never changes to yellow.

**Screenshots/Video**
![EngineVersionPM](https://user-images.githubusercontent.com/86953659/227544881-2b4b0022-6bb4-4479-bc4d-8d3ecb670a92.PNG)

https://user-images.githubusercontent.com/86953659/227544929-90745dbb-f129-4b2f-bf30-05f50b0b751c.mp4

**Found in Branch**
Stabilization/2305 Installer build (https://o3debinaries.org/stabilization-2305/Latest/Linux/O3DE_latest.deb)
Development Installer build (https://o3debinaries.org/development/Latest/Linux/O3DE_latest.deb)

**Desktop/Device:**
 - Device: PC
 - OS: Windows
[...]

Issue #12901, opened 2022-10-31
Visual changes in the gem catalog persist on a new project if it shares the same name


LB-JaroslawGaweda:
**Describe the bug**
Creating a new project with the same name after deleting the previous one will have the same configuration of active gems. It should be noted that this is a UI visual persistence, as deactivated gems will work on a new project that should have them activated by default (e.g. deleting a project with deactivated white box gem and creating a new one from the same default template and the same name will have it deactivated, but the gem will work inside the new project). 
It should be noted that on the new project when entering the configure gems tab there will be a notification by the gem bag icon of a gem to be activated. 

**Steps to reproduce**
1. Create a new project from default template, build it.
4. From the project's hamburger menu, choose Configure Gems.
5. Switch the White Box gem off.
6. Delete the project.
7. Create a new project from default template, with the same name, build it.
8. From the project's hamburger menu, choose Configure Gems.

**Expected behavior**
The gems catalog should display the actual correct gem configuration.

**Actual behavior**
The gem catalog does not display the actual correct gem configuration.

**Found in Branch**
Development (e0e4557)

Desktop :
• Device: PC
• OS: Windows
• Version 10
• CPU Intel Core i7-9700K
• GPU Nvidia GeForce RTX 2070
• Memory 16GB

LB-JaroslawGaweda:
Issue verified as fixed in the latest Development (eb6620e) and Stabilization/2305 (1832bb1) branches.

PR #15474, opened 2023-03-30
fix misc project bugs in Project Manager

AMZN-alexpete:
## What does this PR do?

- fixes https://github.com/o3de/o3de/issues/15360
- fixes https://github.com/o3de/o3de/issues/12901
- fixes issue where users could not add projects with compatibility warnings
- fixes issue where adding a project didn't register it to the specific engine in use.  If the user had multiple engines with the same name, the project just used the most compatible engine name, which may not be the engine in use.
- fixes missing error and success messages when adding a project
image

image

- fixes https://github.com/o3de/o3de/issues/15557
- shows the "download template" when the user tries to modify gems before downloading
- fixes https://github.com/o3de/o3de/issues/15578

## How was this PR tested?

- ran o3de pytests
- tested adding and removing projects that had invalid project.json files, missing project.json files, and specified incompatible engines
- tested adding multiple projects with the same name and different versions
- verified updates to the local user's project.json `engine_path` existed in the project folder after adding it
- tested adding the o3de-extras remote repo as a .git repo, and downloading the Multiplayer Sample project template


lumberyard-employee-dm:
> I'm debating if we should add the specifier that registers the project to a specific engine version e.g. o3de~=2.0.0, because I don't want users working with a development engine to have to keep registering their project (or edit the project.json to remove the specifier), but it could be really useful for anyone using a release engine. So, I might make this change in a later PR that involves showing engine compatibility in the project properties UI

This should be determined by the user. There is no way to handle the users "wants" here by making a decision on their behalf.

lumberyard-employee-dm (changes requested):

Code/Tools/ProjectManager/Source/UpdateProjectCtrl.cpp
```diff
@@ -47,6 +47,7 @@ namespace O3DE::ProjectManager
 
         connect(m_projectGemCatalogScreen, &ScreenWidget::ChangeScreenRequest, this, &UpdateProjectCtrl::OnChangeScreenRequest);
         connect(m_gemRepoScreen, &GemRepoScreen::OnRefresh, m_projectGemCatalogScreen, &ProjectGemCatalogScreen::Refresh);
+        connect(reinterpret_cast<ScreensCtrl*>(parent), &ScreensCtrl::NotifyProjectRemoved, m_projectGemCatalogScreen, &GemCatalogScreen::NotifyProjectRemoved);
```
lumberyard-employee-dm, line 50: repeat: This needs to be a `static_cast`
```suggestion
        connect(static_cast<ScreensCtrl*>(parent), &ScreensCtrl::NotifyProjectRemoved, m_projectGemCatalogScreen, &GemCatalogScreen::NotifyProjectRemoved);
```

lumberyard-employee-dm (approved):

tkothadev (approved):

PR #15483, opened 2023-03-30
Fix open in other asset browser

### LINKED ISSUE #14841: Asset Browser: "Open in Another Asset Browser" context menu option opens a new Asset Browser, but without the asset selected
**Describe the bug**
"Open in Another Asset Browser" context menu option opens a new Asset Browser, but without the asset selected

**Steps to reproduce**
```
```
2. Open the Asset Browser. 
3. Switch to List view (necessary currently due to https://github.com/o3de/o3de/issues/14397)
4. Drill down to any asset.
5. Right-click the asset and "Open in Another Asset Browser".

**Expected behavior**
A new Asset Browser is opened with the specified asset selected in it.

**Actual behavior**
A new Asset Browser with default settings is opened. No asset is selected.

**Found in Branch**
development

945c0ee96955487f0fabc35deaff1ea6629a4fd3

sphrose:
## What does this PR do?

This PR resolves #14841. It now selects the correct folder in the treeview and asset in the thumbnail view.

## How was this PR tested?

Tested manually in the new thumbnail browser and verified the old browser still works.

### DIFF
--- a/Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
+++ b/Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
@@ -20,6 +20,7 @@
 #include <AzToolsFramework/AssetBrowser/AssetBrowserModel.h>
 #include <AzToolsFramework/AssetBrowser/AssetBrowserTableModel.h>
 #include <AzToolsFramework/AssetBrowser/AssetBrowserFilterModel.h>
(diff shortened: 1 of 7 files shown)


dtamkin1:
Does this fix take into account the current view we're in? For example if a new AB is opened from table view, then the second AB should open to the table view with the correct asset selected. Same thing if we're in list view

sphrose:
It doesn't, I'll look at getting that working, thanks.

PR #15485, opened 2023-03-30
New fixture and unit tests for property override undo node

forkercat:
## What does this PR do?

This PR adds a new test fixture and unit tests for `PrefabUndoComponentPropertyOverride` undo node.

It aims to capture the issues we just fixed in this PR: https://github.com/o3de/o3de/pull/15264

Change List:
- New test fixture that does DOM value conversion, patch generation, and testing scene creation
- New unit tests for property-edit overrides: Default value behavior, override on nested prefab, etc

Next step: Add unit tests for the other undo node: `PrefabUndoComponentPropertyEdit`.

## How was this PR tested?

Tests passed.


Code/Framework/AzToolsFramework/Tests/Prefab/PrefabUndoComponentPropertyOverrideTests.cpp
```diff
@@ -0,0 +1,287 @@
...
+        InstanceOptionalReference carInstance = m_instanceEntityMapperInterface->FindOwningInstance(carInstanceInfo.m_containerEntityId);
+        EXPECT_TRUE(carInstance.has_value());
+
+        // Redo
+        undoNode.CaptureAndRedo(carInstance->get(), pathToPropertyFromOwningPrefab, propertyDomValue);
+        EXPECT_TRUE(undoNode.Changed());
+        PropagateAllTemplateChanges();
+
+        bool currentStaticValue = defaultStaticValue; // default to opposite value
+        AZ::TransformBus::EventResult(currentStaticValue, wheelEntityInfo.m_entityId, &AZ::TransformBus::Events::IsStaticTransform);
```
michabr, line 50: We could add a check here to validate that wheelEntityInfo.m_entityId is still a valid entity and wasn't recreated from the PropagateAllTemplateChanges().

Code/Framework/AzToolsFramework/Tests/Prefab/PrefabUndoComponentPropertyTestFixture.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+        m_prefabOverridePublicInterface = AZ::Interface<PrefabOverridePublicInterface>::Get();
+        ASSERT_TRUE(m_prefabOverridePublicInterface);
+    }
+
+    bool PrefabUndoComponentPropertyTestFixture::CreateWheelEntityHierarchy(EntityInfo& wheelEntityInfo)
+    {
+        AZ::EntityId wheelEntityId = CreateEditorEntityUnderRoot(WheelEntityName);
+        EntityAlias wheelEntityAlias = FindEntityAliasInInstance(GetRootContainerEntityId(), WheelEntityName);
+        EXPECT_FALSE(wheelEntityAlias.empty());
```
michabr, line 30: Perhaps ASSERT_FALSE/ASSERT_TRUE would be more appropriate for setup related calls?

forkercat: Yeah. I should not assert on the return type (bool); instead, I should use all assertions in the setup calls.

Code/Framework/AzToolsFramework/Tests/Prefab/PrefabUndoComponentPropertyOverrideTests.cpp
forkercat: I think we're certain that (like what we're currently doing in other tests) the entity id is valid and unchanged after PropagateAllTemplateChanges(). But the entity object needs to be retrieved again. In these tests, the following calls use the entity id rather than the entity reference, so I think it is okay.

michabr: I guess if the entity gets recreated and entityId changed, then the test would fail anyway in the next line because the expected transform value would be wrong.

Code/Framework/AzToolsFramework/Tests/Prefab/PrefabUndoComponentPropertyOverrideTests.cpp
```diff
@@ -0,0 +1,287 @@
...
+        PropagateAllTemplateChanges();
+
+        currentStaticValue = defaultStaticValue; // reset to opposite value
+        AZ::TransformBus::EventResult(currentStaticValue, wheelEntityInfo.m_entityId, &AZ::TransformBus::Events::IsStaticTransform);
+        EXPECT_EQ(currentStaticValue, overriddenStaticValue);
+        EXPECT_TRUE(
+            m_prefabOverridePublicInterface->AreOverridesPresent(wheelEntityInfo.m_entityId, changePatch.m_pathToProperty.ToString()));
+    }
+
+    TEST_F(PrefabUndoComponentPropertyOverrideTests, EditTransformStaticNoOverrideForUnchangedDefaultValue)
```
srikappa-amzn, line 76: I recommend adding a comment here to make it clearer. Something like changing 'false (default)' to 'false'

Code/Framework/AzToolsFramework/Tests/Prefab/PrefabUndoComponentPropertyOverrideTests.cpp
```diff
@@ -0,0 +1,287 @@
...
+        PropagateAllTemplateChanges();
+
+        currentWheelTranslate = AZ::Vector3(-1.0f); // reset
+        AZ::TransformBus::EventResult(currentWheelTranslate, wheelEntityInfo.m_entityId, &AZ::TransformBus::Events::GetLocalTranslation);
+        EXPECT_EQ(currentWheelTranslate, overriddenTranslation);
+        EXPECT_TRUE(
+            m_prefabOverridePublicInterface->AreOverridesPresent(wheelEntityInfo.m_entityId, changePatch.m_pathToProperty.ToString()));
+    }
+
+    TEST_F(PrefabUndoComponentPropertyOverrideTests, EditTranslationOnEntityInNestedPrefabSucceeds)
```
srikappa-amzn, line 173: This can be added later but one more good test case can be creating the super car prefab hierarchy and focusing on the supercar and making it override the car prefab. It'll verify that overrides work not just from the topmost(level) prefab but will work from nested prefabs too.

srikappa-amzn (approved):

Code/Framework/AzToolsFramework/Tests/Prefab/PrefabUndoComponentPropertyOverrideTests.cpp
forkercat: I will keep this test in this PR and leave this thread open. Can add it later.

PR #15486, opened 2023-03-30
Standardized path resolution for wildcard product dependencies.

AMZN-stankowi:
## What does this PR do?

There are two code paths when resolving product path dependencies with wildcards.
1. The product with a dependency checks all existing files already resolved in the database.
2. Later jobs finish, and the existing previously emitted wildcard path dependencies are checked against the newly completed files.

Without this change, there was a difference in those code paths:
1. This did a database search with database style wildcard. This meant a wildcard matched against the whole string, and didn't stop at directory markers.
2. This did a path search with path style wildcards. This meant a wildcard stopped matching at directory change markers.

This meant that, if you re-processed a file with a wildcard dependency after the asset processor went idle, you could get new dependencies, due to the difference in how it was resolved.

This change standardizes the behavior to 2, which matches how other systems have worked with paths and wildcards.

## How was this PR tested?

Tested locally with multiplayer sample.

This is meant to be a quick fix for stabilization - I'm going to separately evaluate automation test options here, and re-writing some of this logic to share code instead of duplicate code, for development.


mbalfour-amzn (approved):

amzn-mike (approved):

amzn-mike (approved):

AMZN-stankowi:
I accidentally merged the wrong branch - I need to create a new PR

PR #15491, opened 2023-03-30
Support instanced object transforms in shaders

amzn-tommy:
## What does this PR do?

Added support for instanced object transforms in shaders. These can be retrieved by the DefaultDrawSrg, which has new GetWorldMatrix* functions. The instanced transforms are off by default and controlled via a shader option. They fall back to using the ObjectSrg when the option is false. This opt-in behavior allows existing systems to render objects without needing to prepare instance buffers. 

Because of the fallback behavior, when including the DrawSrg, the  ObjectSrg must be included first. Any custom ObjectSrg can still be used with the DefaultDrawSrg, as long as it implements GetWorldMatrix, GetWorldMatrixInverseTranspose, and GetWorldMatrixPrev. For example, the SkinObjectSrg is used by the skin shaders.

Some shaders were including the DefaultDrawSrg even without using any DrawSrg data because a DrawSrg must exist for shader options to be used. That coupled them to changes in the DefaultDrawSrg, such as now needing an ObjectSrg to be provided. To de-couple them, I created an UnusedFallbackDrawSrg that can be included instead.

Added a new SystemValues struct named SV that is used to pass the instanceId around. I chose a struct instead of a uint so that future system value semantics can be passed in, like triangleId and vertexId, which will be useful whenever we start doing programmable vertex pulling. This struct can be customized by shaders in the same manner that VsInput and VsOutput are.

There were a few places such as MultilayerSetPixelDepth that called ObjectSrg::GetWorldTransform() after it had already been called higher up in the callstack. Rather than pass the instanceId down through the callstacks, I just pass the world transform down so it doesn't need to be fetched twice.

## This is an impactful change.
Here is a draft of the impactful change message:

**What is changing?**
We are adding support to fetch object to world transforms from an instance buffer to O3DE's shaders.

**What do I need to do?**
This change will only impact you if you are working with shaders.

--To keep your shaders working as-is:
•	If you are not building shaders based on O3DE's PBR shaders (that is, you are not using the EvaluateVetexGeometry/EvaluatePixelGeometry/etc. macros), you can continue using the ObjectSrg to get matrix transforms.
•	If you use the EvaluateVertexGeometry macro in your custom shader, you will need to update it to ignore the new SystemValues struct that is a parameter in that function
            Instead of using something like
            #define EvaluateVertexGeometry EvaluateVertexGeometry_DepthPass
            use
            #define EvaluateVertexGeometry(IN, SV) EvaluateVertexGeometry_DepthPass(IN)
            to ignore the SystemValues struct.

-- To support instancing in your shaders:
1.	Change the parameters for your EvaluateVertexGeometry_ function from (VsInput IN) to (VsInput IN, VsSystemValues SV).
2.	Define VsSystemValues wherever you define VsInput.
-------- Add a uint m_instanceId to VsSystemValues.
3.	#include <Atom/Features/InstancedTransforms.azsli> in any shader where you want to get the world transform
4.	If you have a custom ObjectSrg, update it to include GetWorldMatrix, GetWorldMatrixPrev, and GetWorldMatrixInverseTranspose. The Skin shader does this, as an example.
-------- This is used as a fallback for when instancing is disabled.
5.	Use GetObjectToWorldMatrix(instanceId) instead of ObjectSrg::GetWorldMatrix. Same for GetWorldMatrixInverseTranspose and GetWorldMatrixPrev
6.	Add uint m_instanceId to your VsOutput struct.
7.	Set output.m_instanceId somewhere in your EvaluateVertexGeometry function.
8.	See the PR for examples of how the existing shaders were updated. https://github.com/o3de/o3de/pull/15491
 
**When is this happening?**
PR related to this work is here https://github.com/o3de/o3de/pull/15491. Aim is to merge it into development tomorrow.

**Why are we doing this?**
This change will pave the way for mesh instancing, which improves CPU performance for scenes with a lot of repeated meshes & materials.

## How was this PR tested?

Tested using AtomSampleViewer on both DX12 and Vulkan.
Tested the Editor with the AutomatedTesting project
Tested MaterialCanvas


Gems/Atom/RPI/Assets/ShaderLib/Atom/RPI/ShaderResourceGroups/DefaultDrawSrg.azsli
gadams3, line 1: Would it be cleaner for these functions to live in their own file instead of the SRG to avoid the dependencies on the other SRGs?

gadams3 (approved):

Gems/Atom/Feature/Common/Assets/Shaders/Reflections/ReflectionScreenSpaceBlurVertical.azsl
```diff
@@ -20,7 +20,8 @@ ShaderResourceGroup PassSrg : SRG_PerPass
     float m_invOutputScale;
 }
 
-#include <Atom/RPI/ShaderResourceGroups/DefaultDrawSrg.azsli>
+// The shader variant system needs a fallback, using UnusedFallbackDrawSrg for this.
```
moudgils, line 23: What error do you get if you simply remove DefaultDrawSrg.azsli?

amzn-tommy: Generally speaking, you'll see something like

4/4/2023 9:27 AM | ERROR | AZSLc: D:\GitHub\tinyprbranch\o3de\AutomatedTesting\user\AssetProcessorTemp\JobTemp-VmDfMW\LightCullingTilePrepare_dx12.azslin(,) : IR error #129: If you have non-static options, one SRG must be designated as the default ShaderVariantFallback from AssetBuilder

Although in this particular case, the original shader did not need to include the DefaultDrawSrg in the first place, because it does not use any shader variants, so this one can just be removed.

Gems/Atom/RPI/Assets/ShaderLib/Atom/RPI/ShaderResourceGroups/DefaultDrawSrg.azsli
amzn-tommy: It's a good point. The original reason this was in the DrawSrg was just because I tried both using a root constant and the DrawSrg to hold the instance offset, to see which was more performant. Updating the DrawSrg every frame on the cpu side turned out to be prohibitively expensive relative to updating a root constant. So that original reason of putting it in this file is no longer valid, as it's no longer coupled to anything in the DrawSrg.

A separate file would make the DefaultDrawSrg not dependent on having the ObjectSrg with the appropriate functions defined first. However, that means yet another #include that must be sprinkled everywhere. And _that_ file is still going to depend on the ObjectSrg for a fallback. So I don't think there's really anything to be gained. Also, m_rootConstantInstanceDataOffset is per-draw data (every instanced draw call uses it to fetch the per-instance data), so I think it's natural to keep it in the same file as the rest of the per-draw data.

amzn-tommy: Thinking about this a little more, and there is potentially another advantage to using a separate file, regarding how this feature is enabled/rolled out.

First I'll describe how I imagined it going. It should be opt-in. If someone writes their own EvaluateVertexGeometry function, and they use ObjectSrg::GetWorldTransform, the MeshFeatureProcessor can't lump those together into a single instanced draw call. It needs to know somehow that this shader/material type doesn't support instanced draw calls. We can't just check to see if o_meshInstancingEnabled exists in the shader, because anyone who is using the DefaultDrawSrg but has not updated their shader to call DrawSrg::GetWorldMatrix instead of ObjectSrg::GetWorldMatrix, would end up with a broken shader. So I thought we could add a property to a .materialtype for 'supports instancing'. The MeshFP checks for that property, and if it exists/is true, it sets the o_meshInstancingEnabled shader option to true. If not, that object gets its own separate draw call. Then over time, we could deprecate the need for the property, and just have an implicit assumption that if you use the DefaultDrawSrg, you have updated your shaders to support instancing.

Now I'm thinking that we could remove this from the DrawSrg, and the way of opting-in is to check for the existence o_meshInstancingEnabled option. Existing shaders would not be including this new InstancedTransforms.azsil (or some such named) file, even if they currently use the DefaultDrawSrg. So they would be opting-out by default, and anything that includes the new file would be explicitly opting in that way, and we'd have no need for a material property. It also squashes the list of "What do I need to do" down to just updating EvaluateVertexGeometry overrides to have a second parameter, unless they want to support instancing, which would just be #include the new file and use a new GetInstancedWorldTransform(instanceId) function. So I like this idea better.

What do you think?

Gems/Atom/Feature/Common/Assets/Shaders/Reflections/ReflectionScreenSpaceBlurVertical.azsl
moudgils: yeah. My original confusion is why do we need to include an azsli file  whose SRG is not even used in the shader.

amzn-tommy: Yeah, even if the Srg is not used, we still need a fallback for the shader variant system (some shaders use SRG_PerPass_WithFallback for this purpose, others use SRG_PerDraw). But in this case, it also does not use shader options, so it is not necessary.

invertednormal (approved):
Nice work.

moudgils (approved):

PR #15505, opened 2023-03-31
Fix lag when switching views in the Asset Browser

### LINKED ISSUE #15451: Asset Browser: Returning to List/Tree view from Table/Thumbnail views takes several seconds
**Describe the bug**
Returning to List/Tree view from Table/Thumbnail views takes several seconds. This behavior is worsened by not toggling on the options for Hide Engine Folders/Unusable Product Assets.

**Steps to reproduce**
1. Launch the Editor.
2. Enable the new Asset Browser if necessary:
```
ed_useWIPAssetBrowserDesign=true
```
3. Toggle off the Hide Engine Folders/Unusable Product Assets options in the Asset Browser.
4. Switch from Table or Thumbnail view back to List/Tree view.

**Expected behavior**
View is switched quickly.

**Actual behavior**
View takes several seconds to update when going to List/Tree view from Thumbnail or Table views.

**Screenshots/Video**

https://user-images.githubusercontent.com/82226555/228606065-978d92c1-3043-4fdf-a569-dc7afe93728b.mp4

**Found in Branch**
development

9f4341ff5f5eb0ad4ddf997469488ba8642cc762

**Desktop/Device:**
 - Device: PC
 - OS: Windows 10 21H1
 - CPU: Ryzen 5950x
 - GPU: Nvidia 2070
 - Memory: 64GB

dtamkin1:
## What does this PR do?

When switching views in the asset browser, we were calling a recursive method to check whether each entry should be expanded or not. Since there is no change in expansion state when switching views, this method can be circumvented. Makes view switching nearly instantaneous.

Switching to list view now expands the currently selected folder, and scrolls to it while centering it. Also does nothing if user click a view switch button and they're already in that view

scrolling to folders when clicked on in the thumbnail view and table view centers them as well

Fixes #15451 

## How was this PR tested?

Locally


Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
```diff
@@ -665,9 +678,18 @@ void AzAssetBrowserWindow::SetTwoColumnMode(QWidget* viewToShow)
...
-    m_ui->m_thumbnailView->SetThumbnailActiveView(false);
+    if (m_ui->m_thumbnailView->GetThumbnailActiveView() || m_ui->m_expandedTableView->GetExpandedTableViewActive())
+    {
+        m_ui->m_middleStackWidget->hide();
+        m_ui->m_assetBrowserTreeViewWidget->SetApplySnapshot(false);
+        m_ui->m_searchWidget->RemoveFolderFilter();
+        if (!m_ui->m_assetBrowserTreeViewWidget->selectionModel()->selectedRows().isEmpty())
+        {
+            m_ui->m_assetBrowserTreeViewWidget->expand(m_ui->m_assetBrowserTreeViewWidget->selectionModel()->selectedRows()[0]);
+        }
```
cgalvan, line 689: Can there be multiple selected rows? If so, should we expand each of those indexes?

cgalvan (approved):

Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
dtamkin1: There can be, but it seems to me that having a selected row in the tree view doesn't necessarily mean the user wants all those folders expanded if they switch views? 
I was thinking that it makes sense to expand the current folder that the user is viewing, but not all of them since the multi-select might be for folder operations

lsemp3d (approved):

PR #15510, opened 2023-03-31
Created a template that produces a Gem with Atom support 

l-sempe:
## What does this PR do?

This template generates a gem that is configured to support Atom rendering components and basic dependencies.

## How was this PR tested?

Code was generated, compiled and used breakpoints to verify that components were registered, created and activated

Created a gem using 

`scripts\o3de create-gem -tp Templates/GraphicsGem -gp Gems\GraphicsGem_AR_Test -r ${GemName} GraphicsGem_AR_Test`

Pushed this gem into this PR in order to have AR run over the generated code for all platforms.

Once AR passed, I reverted the commits related to `Gems\GraphicsGem_AR_Test`


Templates/GraphicsGem/Template/Code/Source/${Name}Module.h
```diff
@@ -0,0 +1,30 @@
...
+#include <AzCore/Module/Module.h>
+
+namespace AZ::Render
+{
+    class ${Name}Module
+        : public AZ::Module
+    {
+    public:
+        AZ_RTTI(${Name}Module, "{${Random_Uuid}}", AZ::Module);
+        AZ_CLASS_ALLOCATOR(${Name}Module, AZ::SystemAllocator);
```
lumberyard-employee-dm, line 21: I have made updates to the existing gem templates to move the RTTI, TypeInfo, Component Descriptor and Allocator definitions to cpp files.

This template should follow that same approach.
The reason why is to reduce compile time cost, as o3de compile time is suffering death by thousand cuts with unnecessary compiling of functions in headers and includes.
Here is a link to the those Template updates: https://github.com/o3de/o3de/commit/b16fb0aa4fd56df9aa214cbfeb43ca93e99ce8ff

The changes would invoke using the `AZ_TYPE_INFO_WITH_NAME_DECL` macro, `AZ_RTTI_NO_TYPE_INFO_DECL` and `AZ_CLASS_ALLOCATOR_DECL` macro in headers.

The `AZ_COMPONENT` macro can be replaced with the `AZ_COMPONENT_DECL` macro.

lsemp3d: Ok, I had seen it, though when I first used it I got some errors, likely they were due to missing include files. I'll do it that way and report back any issues

Templates/GraphicsGem/Template/Code/CMakeLists.txt
```diff
@@ -0,0 +1,84 @@
...
+    BUILD_DEPENDENCIES
+        PUBLIC
+            AZ::AzCore
+            AZ::AzFramework
+            Gem::Atom_RPI.Public
+            Gem::Atom_Utils.Static
+            Gem::Atom_Feature_Common
+            Gem::AtomLyIntegration_CommonFeatures.Static
+            Gem::LmbrCentral
+)
```
lumberyard-employee-dm, line 31: This is missing the .API targets which is used to avoid exposing the internal API of the gem.
The .Static target is meant as a private `STATIC` library that is only linked with the gem.
The .API exposes the public API of the gem and should only provide an include directory INTERFACE for the "Include" directory.
Furthermore the .API target has minimal build dependencies to make it as lean as possible to use as it relates to compile time.

If anything I recommend making this Gem based off of a gem created from the [DefaultGem](https://github.com/o3de/o3de/blob/development/Templates/DefaultGem/Template/Code/CMakeLists.txt) template.

You can 
1. Use the `o3de.py create-gem` command to a create gem scratch gem using the default template
2. Update the gem with the Graphic files and make sure it all builds.
3. Use the `o3de.py create-from-template` command to create this Gem template again.
4. Use the `o3de.py create-gem -tn Graphics` to create a gem using this template
5. **Most importantly:** Make sure the gem instantiated from the Graphics template actually builds.

Step 5 is a requirement, as AR doesn't validate that the gem instantiated from the template builds.

A good test you can do is to add a temporary commit to the branch that instantiates the gem template and registers it with the `engine.json`, which will happen automatically if the destination folder is within engine-root.
Next, you can then run AR to make sure the Gem builds for all the AR supported platforms.

If that succeeds, can you add that information to the github description and then rebase or revert out the commit that added the temporary gem

lumberyard-employee-dm (changes requested):

Templates/GraphicsGem/Template/Code/CMakeLists.txt
l-sempe: I did this, it compiles and the code is correct, however, there is a problem. The component does not show up in the component list, so far I haven't found the cause.

l-sempe: This is done, AR passed with created gem and gem has been reverted

lumberyard-employee-dm: Thanks for validating the create gem workflow with the new gem.

Templates/GraphicsGem/Template/Code/Source/Render/${Name}FeatureProcessor.h
```diff
@@ -0,0 +1,33 @@
...
+        ${Name}FeatureProcessor() = default;
+        virtual ~${Name}FeatureProcessor() = default;
+
+        // FeatureProcessor overrides
+        void Activate() override;
+        void Deactivate() override;
+        void Simulate(const FeatureProcessor::SimulatePacket& packet) override;
+
+    };
+}
```
lumberyard-employee-dm, line 33: nitpick missing newline at the end of this file.

Templates/GraphicsGem/Template/Code/Source/Render/${Name}FeatureProcessor.cpp
```diff
@@ -0,0 +1,37 @@
...
+    void ${Name}FeatureProcessor::Deactivate()
+    {
+        
+    }
+
+    void ${Name}FeatureProcessor::Simulate([[maybe_unused]] const FeatureProcessor::SimulatePacket& packet)
+    {
+        
+    }    
+}
```
lumberyard-employee-dm, line 37: repeated nitpick: missing newline at the end of this file.

Templates/GraphicsGem/Template/Code/Include/${Name}/${Name}TypeIds.h
```diff
@@ -0,0 +1,26 @@
...
+#pragma once
+
+namespace ${Name}
+{
+    // System Component TypeIds
+    inline constexpr const char* ${Name}SystemComponentTypeId = "${SysCompClassId}";
+    inline constexpr const char* ${Name}EditorSystemComponentTypeId = "${EditorSysCompClassId}";
+
+    // Module derived classes TypeIds
+    inline constexpr const char* ${Name}ModuleInterfaceTypeId = "{E22C9D32-526E-481F-8440-29E489EEB349}";
```
lumberyard-employee-dm, line 18: This needs to be replaced with `"{${Random_Uuid}}"`

The Uuid logic template logic in the [engine_template.py](https://github.com/o3de/o3de/blob/development/scripts/o3de/o3de/engine_template.py#L709-L734) will need to be updated in the future to be updated replace actual Uuids in this file with the `"{${Random_Uuid}}"` placeholder

Templates/GraphicsGem/Template/Code/Include/${Name}/${Name}TypeIds.h
```diff
@@ -0,0 +1,26 @@
...
+    inline constexpr const char* ${Name}ModuleTypeId = "${ModuleClassId}";
+    // The Editor Module by default is mutually exclusive with the Client Module
+    // so they use the Same TypeId
+    inline constexpr const char* ${Name}EditorModuleTypeId = ${Name}ModuleTypeId;
+
+    // Interface TypeIds
+    inline constexpr const char* ${Name}RequestsTypeId = "{D6D4B90B-2392-40D1-8177-9C71383B57AD}";
```
lumberyard-employee-dm, line 25: repeat: This needs to be replaced with `"{${Random_Uuid}}"`

Templates/GraphicsGem/Template/Code/Source/Tools/Components/Editor${Name}Component.cpp
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    void Editor${Name}Component::OnTick([[maybe_unused]] float deltaTime, [[maybe_unused]] AZ::ScriptTimePoint time)
+    {
+        if (!m_controller.m_featureProcessor)
+        {
+            return;
+        }
+    }
+}
```
lumberyard-employee-dm, line 88: repeated nitpick: Missing newline at the end of this file

Templates/GraphicsGem/Template/Code/Source/Components/${Name}ComponentController.h
```diff
@@ -0,0 +1,73 @@
...
+    {
+    public:
+        AZ_RTTI(${Name}ComponentConfig, "{${Random_Uuid}}", ComponentConfig);
+        AZ_CLASS_ALLOCATOR(${Name}ComponentConfig, AZ::SystemAllocator);
+        static void Reflect(AZ::ReflectContext* context);
+
+        ${Name}ComponentConfig() = default;
+
+        AZ::u64 m_entityId{ AZ::EntityId::InvalidEntityId };
+    };
```
gadams3, line 30: nit I would consider moving the config to its own h/cpp files but it's not critical.

gadams3 (approved):
Should the template provide the system component and editor system component? I suppose it's better to be verbose and provide both because someone can always delete whatever they don't need.

Can the components be created from a separate template in case we want to create more than one components in the same gem?

lsemp3d:
> Should the template provide the system component and editor system component? I suppose it's better to be verbose and provide both because someone can always delete whatever they don't need.
> 
> Can the components be created from a separate template in case we want to create more than one components in the same gem?

it does provide them because it's the starting point towards registering and reflecting all associated data types, it creates a stubbed out component with a feature processor so that developers can go straight into writing graphics code vs. spending time writing all the boilerplate.

Templates/GraphicsGem/Template/Code/Include/${Name}/${Name}TypeIds.h
```diff
@@ -11,16 +11,16 @@
 namespace ${Name}
 {
     // System Component TypeIds
-    inline constexpr const char* ${Name}SystemComponentTypeId = "${SysCompClassId}";
-    inline constexpr const char* ${Name}EditorSystemComponentTypeId = "${EditorSysCompClassId}";
+    inline constexpr const char* ${Name}SystemComponentTypeId = "{${Random_Uuid}}";
+    inline constexpr const char* ${Name}EditorSystemComponentTypeId = "{${Random_Uuid}}";
```
lumberyard-employee-dm, line 15: These placeholders were fine, it was just the actual Uuids that needed to be replaced with the "{${Random_Uuid}}" placeholder.

Templates/GraphicsGem/Template/Code/Include/${Name}/${Name}TypeIds.h
```diff
@@ -11,16 +11,16 @@
...
 
     // Module derived classes TypeIds
-    inline constexpr const char* ${Name}ModuleInterfaceTypeId = "{E22C9D32-526E-481F-8440-29E489EEB349}";
-    inline constexpr const char* ${Name}ModuleTypeId = "${ModuleClassId}";
+    inline constexpr const char* ${Name}ModuleInterfaceTypeId = "{${Random_Uuid}}";
+    inline constexpr const char* ${Name}ModuleTypeId = "{${Random_Uuid}}";
```
lumberyard-employee-dm, line 19: repeat: "${ModuleClassId}" was fine here.

lumberyard-employee-dm (approved):

Templates/GraphicsGem/Template/Code/Include/${Name}/${Name}TypeIds.h
lsemp3d: ok, I'll bring them back

Templates/GraphicsGem/Template/Code/Include/${Name}/${Name}TypeIds.h
lsemp3d: ok, I'll bring them back

PR #15520, opened 2023-04-03
UI theme refactoring | Foundational changes

AMZN-daimini:
Kickstarts a short effort to refactor the way Editor UI properties are loaded to allow for easier re-skinning of the Editor, and support multiple themes in the process.

### What this does

- Introduces a system to load `themeProperties.json` files into a properties map that can be referenced throughout the UI. ;
- Hardcodes the loading of the `O3DE_original` theme at Editor startup, which is exactly identical to the current Editor theme;
- Repurposes the Stylesheet Preprocessor (unused in the current UI, remnant of the old system) to allow for SCSS-like variable functionality. Writing `$PropertyName` in the `qss` files will allow stylesheets to reference the value of a theme property (the `$PropertyName` will be replaced with the value of the property during this preprocessing phase);
- Starts the process of removing the `Config.ini` files in the UI Component Library with the theme property file mentioned above;
- Introduces an interface that allows C++ code to retrieve the values of properties programmatically, and use them in painting code that is not reachable via qss;
- Removes some of the old "UI1.0" classes and functions that have been unused since the first O3DE release and would just be confusing. (more red-coding to come as this progresses, most likely).

### What's still missing/what is going to change

- The actual structure of the theme properties file may change over time in this phase, as we determine what needs to be consolidated and what needs to be separated;
- Theme selection is currently hardcoded; as a follow-up, we will find a way to control which theme is selected via a settings registry property first, and an Editor preferences setting later;
- The Stylesheet Preprocessing code is currently in a legacy class, we will proceed to move it to a better location in a follow-up PR;
- Only about 5/6 UI Components have been updated to use the new theme properties so far, the rest of the work will come separately.


Code/Framework/AzQtComponents/AzQtComponents/Components/StyleManager.cpp
```diff
@@ -181,6 +238,14 @@ namespace AzQtComponents
         m_autoCustomWindowDecorations = new AutoCustomWindowDecorations(this);
         m_autoCustomWindowDecorations->setMode(AutoCustomWindowDecorations::Mode_AnyWindow);
 
+        AZStd::string themeName = "O3DE_Original";
+
+        AZStd::string themePropertiesPath = AZStd::string::format(
+            "D:\\GitRepos\\O3DE\\o3de\\Code\\Framework\\AzQtComponents\\AzQtComponents\\Themes\\%s\\themeProperties.json",
+             themeName.c_str());
```
cgalvan, line 245: Find this without hard-coding the path. Might need to go under `Assets/Editor` so it can be found by the installer as well?

cgalvan (approved):

Code/Editor/Plugins/ComponentEntityEditorPlugin/SandboxIntegration.cpp
```diff
@@ -1567,12 +1567,8 @@ AZ::EntityId SandboxIntegrationManager::ContextMenu_NewLayer()
...
 
     // Make sure the color is created fully opaque.
-    static QColor newLayerDefaultColor = GetIEditor()->GetColorByName("NewLayerDefaultColor");
-    AZ::Color newLayerColor(
-        aznumeric_cast<float>(newLayerDefaultColor.redF()),
-        aznumeric_cast<float>(newLayerDefaultColor.greenF()),
-        aznumeric_cast<float>(newLayerDefaultColor.blueF()),
-        aznumeric_cast<float>(newLayerDefaultColor.alphaF()));
+    AZ::Color newLayerColor;
+    newLayerColor.SetA(1.0f);
```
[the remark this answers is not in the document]
pollend: would this be easier?
```
AZ::Color newLayerColor{0.0f,0.0f,0.0f,1.0f};
```

Code/Framework/AzQtComponents/AzQtComponents/Components/Widgets/ToolButton.h
```diff
@@ -24,39 +24,35 @@ namespace AzQtComponents
...
+        //! Margin around ToolButton controls, in pixels. All directions get the same margin.
+        inline static int s_defaultButtonMargin = 1;
+        //! Width of the menu indicator arrow in pixels.
+        inline static int s_menuIndicatorWidth = 10;
+        //! Background color for checkable ToolButtons set to the checked state.
+        inline static QColor s_checkedStateBackgroundColor = QStringLiteral("#1E70EB");
+        //! Path to the indicator icon. Svg images recommended.
+        inline static QString s_menuIndicatorIcon = QStringLiteral(":/stylesheet/img/UI20/menu-indicator.svg");
+        //! Size of the indicator icon. Size must be proportional to the icon's ratio.
+        inline static QSize s_menuIndicatorIconSize = QSize(6, 3);
```
pollend, line 56: nit: should these be const?

```suggestion
        //! The default size of button icons in pixels.
        static constexpr int ButtonIconSize = 16;
        //! Margin around ToolButton controls, in pixels. All directions get the same margin.
        static constexpr int DefaultButtonMargin = 1;
        //! Width of the menu indicator arrow in pixels.
        static constexpr int MenuIndicatorWidth = 10;
        //! Background color for checkable ToolButtons set to the checked state.
        static const QColor CheckedStateBackgroundColor = QStringLiteral("#1E70EB");
        //! Path to the indicator icon. Svg images recommended.
        static const QString MenuIndicatorIcon = QStringLiteral(":/stylesheet/img/UI20/menu-indicator.svg");
        //! Size of the indicator icon. Size must be proportional to the icon's ratio.
        static const QSize MenuIndicatorIconSize = QSize(6, 3);
```

Code/Editor/Plugins/ComponentEntityEditorPlugin/SandboxIntegration.cpp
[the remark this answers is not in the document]
gadams3: +1

Code/Framework/AzQtComponents/AzQtComponents/Components/StyleManager.cpp
```diff
@@ -47,9 +51,43 @@ namespace AzQtComponents
...
+    QColor StyleManager::GetStylePropertyAsColor(const char* propertyKey) const
+    {
+        if (m_themeProperties.contains(propertyKey))
+        {
+            return QColor(m_themeProperties.value(propertyKey));
+        }
+
+        return QColor();
+    };
+
```
gadams3, line 90: Instead of doing the look up twice by calling contains and then value, you should be able to use find to get the iterator and return the value if it's a valid iterator. 

QHash has an overloaded version of the value function that allows you to pass in a default value to return if the value is not in the container. That will collapse all of these functions down to one line.

gadams3 (approved):

lsemp3d:
I tested this PR, unfortunately it may be missing some data files causing the UI to be completely dark, it needs additional work to complete it
image

JT-SCB-GameDesign:
NOTE: this was linked in Discord because the topic is being picked back up.
https://discord.com/channels/805939474655346758/1457148819396362362

Issue #15271, opened 2023-03-22
AssetBundler Row is Empty if Asset Not Found in Cache


AMZN-Gene:
The asset bundle will show an empty row if the asset it references doesn't exist in the cache folder.

**Steps to reproduce**
1. Add an asset to a seedlist
2. Delete an asset and let AP process the deletion
3. Reopen AssetBundler.exe

**Expected behavior**
AssetBundler seedlist should display a warning message on the row. And provide the pathHint to show which asset went missing.

**Actual behavior**
The row is empty and the UI looks broken.

**Screenshots/Video**
![image](https://user-images.githubusercontent.com/32776221/227030678-e510d652-ebbe-42d6-b14a-2daaa4554481.png)

**Found in Branch**
development

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
e460ef653b82605ee9696c18c21187400c9535cc
Mar 21 18:47:37 2023

PR #15525, opened 2023-04-03
When an asset in a seed file is no longer available, an error icon and error message is now displayed instead of a blank line.

AMZN-stankowi:
## What does this PR do?

Without this change: When you view a seed list with a missing asset in the Asset Bundler GUI, it displays a blank line. There is an error message in the log output at the bottom of the screen, but it's still confusing having a blank line.

With this change: Instead a blank line, the asset hint and ID are displayed (if available), and an error icon is displayed to call attention to this row.

This change addresses this issue: https://github.com/o3de/o3de/issues/15271

This is what this change looks like:

image

## How was this PR tested?

Manual testing - we don't have automation for asset bundler GUI available yet.


lmbr-pip (approved):
Thanks for fixing this

mbalfour-amzn (approved):
Yay!!!

cgalvan (approved):

PR #15528, opened 2023-04-03
Fixed the intermittent timeouts when connecting to editor server

mbalfour-amzn:
## What does this PR do?

When connecting from the Editor to a multiplayer server, we would see intermittent timeouts during the level send causing the Editor/server connection to fail, even though the level send would retry multiple times over 20 seconds. Upon further investigation, the root cause was that the output of the server was being piped back to the Editor process. If these pipes fill up, the entire server process locks up on outputting to the pipe until the pipe is free to send again. So if the server attempted to print too much to stdout during the connection process, and the client was in a tight loop attempting to retry level sends without processing the pipes, the connection would fail.

The solution is to add "m_serverProcessTracePrinter->Pump();" in the retry loop to ensure that the pipes continue to get processed.

While adding and testing this, I also added more display information to show how many bytes were sent to the server. I've left this in the PR because it seems like a helpful addition. There aren't any screen redraws during the level send so the numbers don't refresh, but it's now more clear where any delays are happening at.

## How was this PR tested?

Ran the MPS level from the Editor several times with a 100% success rate.

image

### DIFF
--- a/Gems/Multiplayer/Code/Source/Debug/MultiplayerConnectionViewportMessageSystemComponent.h
+++ b/Gems/Multiplayer/Code/Source/Debug/MultiplayerConnectionViewportMessageSystemComponent.h
@@ -37,7 +37,7 @@ namespace Multiplayer
         static constexpr char OnServerLaunchFailMessage[] = "(1/3) Could not launch editor server.\nSee console for more info.";
         static constexpr char OnEditorConnectionAttemptMessage[] = "(2/3) Attempting to connect to server in order to send level data.\nAttempt %i of %i";
         static constexpr char OnEditorConnectionAttemptsFailedMessage[] = "(2/3) Failed to connect to server after %i attempts!\nPlease exit play mode and try again.";  
-        static constexpr char OnEditorSendingLevelDataMessage[] = "(3/3) Editor is sending the editor-server the level data packet.";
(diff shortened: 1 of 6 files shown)


Gems/Multiplayer/Code/Source/Debug/MultiplayerConnectionViewportMessageSystemComponent.h
```diff
@@ -37,7 +37,7 @@ namespace Multiplayer
         static constexpr char OnServerLaunchFailMessage[] = "(1/3) Could not launch editor server.\nSee console for more info.";
         static constexpr char OnEditorConnectionAttemptMessage[] = "(2/3) Attempting to connect to server in order to send level data.\nAttempt %i of %i";
         static constexpr char OnEditorConnectionAttemptsFailedMessage[] = "(2/3) Failed to connect to server after %i attempts!\nPlease exit play mode and try again.";  
-        static constexpr char OnEditorSendingLevelDataMessage[] = "(3/3) Editor is sending the editor-server the level data packet.";
+        static constexpr char OnEditorSendingLevelDataMessage[] = "(3/3) Editor is sending the editor-server the level data packet.\n Bytes %u / %u sent.";
```
AMZN-Olex, line 40: Nice!

AMZN-Olex (approved):

AMZN-Gene (approved):

kberg-amzn (approved):
Very very nice work 👍

PR #15529, opened 2023-04-03
Fix deadlock in FileDescriptorCapturer

lumberyard-employee-dm:
The FileDescriptorCapturer class would deadlock when the file descriptor data that was being captured filled up the pipe buffer used to redirect the output.

This would result in the application being blocked attempting to write captured output to one end of the pipe on the same thread where a read of the other end of the pipe would occur later.

The changes here involve making a dedicated "Flush" thread, whose purposes to perform reads on one of the pipe when it receives data and then forward that data to a user provided callback.

On Windows a Completion Routine is used to perform asynchronous IO the flush thread to have it yield when it is waiting for input on the read side of the pipe

On Linux and Android, `epoll` is used to wake up the flush thread when data is available to read from the pipe

On MacOS and iOS, `kqueue` is used to receive a signal to read data from the pipe.

## How was this PR tested?

Verified running the `SerializeContextTools create-slice` command on does not deadlock on Windows, Linux and MacOS when running in a condition where data is sent to stdout. This can occur when a warning or error is output such as project not being completely built or the AssetProcessor isn't running in the background.

Validated that the TraceTests file descriptor redirection test succeed on Windows when running with a --gtest_repeat value of 1000 and on Linux and Mac when running with a repeat count of 10


mbalfour-amzn:
Hmmm, i just hit the same type of problem today with the ProcessCommunicatorTracePrinter with Editor/Server connections that has a more localized fix over on https://github.com/o3de/o3de/pull/15528 . Should that one be changed to use a flush thread as well?

lumberyard-employee-dm:
> Hmmm, i just hit the same type of problem today with the ProcessCommunicatorTracePrinter with Editor/Server connections that has a more localized fix over on #15528 . Should that one be changed to use a flush thread as well?

Well having the ability to off load processing of the stdout and stderr for a child processing in a asynchronous manner could be used to prevent the current thread from blocking on reading the output.

Invoking the [ProcessCommunicator::ReadIntoProcessOutput](https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/AzFramework/Process/ProcessCommunicator.cpp#L45-L63) function in a dedicated thread would alleviate the pipe pressure, by pumping the output as it is received.

mbalfour-amzn (approved):
Nice find and fix!

lawsonamzn (approved):
LGTM.  Did one of the tests that already exist identify the deadlock?

lumberyard-employee-dm:
> LGTM. Did one of the tests that already exist identify the deadlock?

The two test in Trace.cpp of [TraceTests.RedirectToRawOutputToStderr_DoesNotOutputToStdout](https://github.com/o3de/o3de/blob/e8ce34098effb38eef372b8d2c9d6b0041d3c8d6/Code/Framework/AzCore/Tests/Debug/Trace.cpp#L142-L185) and [TraceTests.RedirectToRawOutputNone_DoesNotOutputToStdout_NorStderr](https://github.com/o3de/o3de/blob/e8ce34098effb38eef372b8d2c9d6b0041d3c8d6/Code/Framework/AzCore/Tests/Debug/Trace.cpp#L187-L245) is able to reproduce the deadlock when the [pipe size is smaller](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/IO/SystemFile.h#L204-L205) than the output sent to the descriptor.
For example when 65k of data is sent into the 64k pipe buffer without the pipe being pumped it was causing a deadlock.

I'll add an explicit test to make sure that we output more than 64k of data to the pipe buffer to catch any regressions.

lumberyard-employee-dm:
> > LGTM. Did one of the tests that already exist identify the deadlock?
> 
> The two test in Trace.cpp of [TraceTests.RedirectToRawOutputToStderr_DoesNotOutputToStdout](https://github.com/o3de/o3de/blob/e8ce34098effb38eef372b8d2c9d6b0041d3c8d6/Code/Framework/AzCore/Tests/Debug/Trace.cpp#L142-L185) and [TraceTests.RedirectToRawOutputNone_DoesNotOutputToStdout_NorStderr](https://github.com/o3de/o3de/blob/e8ce34098effb38eef372b8d2c9d6b0041d3c8d6/Code/Framework/AzCore/Tests/Debug/Trace.cpp#L187-L245) is able to reproduce the deadlock when the [pipe size is smaller](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/IO/SystemFile.h#L204-L205) than the output sent to the descriptor. For example when 65k of data is sent into the 64k pipe buffer without the pipe being pumped it was causing a deadlock.
> 
> I'll add an explicit test to make sure that we output more than 64KiB of data to the pipe buffer to catch any regressions.

I added a test to the SystemFile Unit Test that writes 128KiB + 1 bytes of data to validate that a deadlock does not occur
image

As part of the testing however, I discovered that not all the data to stdout was being captured, as the pipe is buffered on Windows and was not being flushed before being disconnected.

I put in a fix for that and now all the data in the captured file descriptor will be sent over the pipe to the callback

Issue #12795, opened 2022-10-25
Import graphics files without UV maps


michalpelka:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I work with poorly exported meshes - it is rendered without any shading.
Such meshes are commonly provided in ROS2 description packages. Some examples you can be found here
https://github.com/husky/husky/blob/foxy-devel/husky_description/meshes/wheel.dae
The problem is missing UV mapping. 
In the image left monkey was created in Blender and exported without UV, right, on another hand was created with UV.
![Screenshot from 2022-10-25 16-16-00](https://user-images.githubusercontent.com/3209244/197799861-1341e5e6-f4a1-4505-93d0-9b759df02ab6.png)

**Describe the solution you'd like**
Have some simple shading when mesh without UV maps is imported

**Describe alternatives you've considered**
Generate placeholder by AssetProcessor

**Additional context**
Two DAE files I used for screenshots:
https://gist.github.com/michalpelka/b82d5898cda935c6cb84d99d2430b438

lmbr-pip:
Most likely involves a default shader when then are no UVs, assigning to Graphics and removing asset pipeline labels.

AMZN-stankowi:
STL files are a good way to test this, too. The STL file format does not support UVs.

pawelbudziszewski:
And it seems that models of some robots are distributed in STL instead of Collada (eg. https://github.com/husarion/rosbot_ros), which causes more problems when importing such robots.

lawsonamzn:
For the robotics usecase the prevalance of things that have no uv coords is actually very high, this needs to be fixed somehow.

lawsonamzn:
I see at least 2 possible ways to fix this...

One way to do it may be to give normals and uvs the same same treatment as we do tangents - that is, there is an option to "generate tangents" (based on normals and uvs).   In the same way, it might be possible to generate normals using average or face normals, and then use xatlas or even just basic spherical coordinate mapping to generate uvs.  Then generate tangents from that.  So there's already kind of a trail blazed here - however the 'generate tangents' feature works (and shows up in the GUI) could be duplicated for 'generate uvs' and 'generate normals'.

lawsonamzn:
I've got a working prototype of this (in case anyone was planning to work on it, trying to avoid duplicate work)
stay tuned

lawsonamzn:
[...]

PR #15530, opened 2023-04-03
Adds the ability to generate UVs automatically during mesh import

lawsonamzn:
## What does this PR do?

This adds a new modifier to the available list of modifiers you can add to a mesh during its import in its import settings.

Pictures:
Before:
no_uvs

Add a modifier in the import settings
add_uvs

And done
after

The new modifier lets you select how to generate UVs for a mesh. Currently, only 1 generator is supported (Spherical Positional projection) but all the hard work here is done to support additional UV-unwrappers in the future by following the pattern (ie, adding to the case statement and adding to the enum).

https://github.com/o3de/o3de/issues/12795 is addressed partially by this.

The old behavior was that meshes that had no UV data had to have UVs added to them in order to render with the current set of PBR shaders.  The only recourse was to go into something like blender and add UVs and re-export.  However, in many situations like for example robotics or simulation, much of the input data do not have UVs and are meant for simple color or shape representation, or procedural texturing.

Spherical unwrapping isn't going to be as good as one of the more advanced unwrappers, but for getting some shading on meshes that really don't care about their material, its a start.

This adds a new modifier to the available modifiers for when you ingest meshes, just like the already existing "generate tangents" rule, you can now add a "generate UVs" rule.   It follows that pattern, so doesn't really invent anything new.

If no modifier is present (all existing data), no changes are made.
If the modifier is present, the default is to use spherical UVs.  The default can be changed in the regset file.

The change to the scenesettings file is as simple as adding
```
 {
                        "$type": "UVsRule"
}
```
to the list of rules applicable to the model.  With optional "replaceExisting" 

## How was this PR tested?

Testing it manually using meshes that have no UV data and meshes that previously existed.   The defaults are configured in such a way to have no effect whatsoever to existing data.  Someone has to go and actually add these rules to the import settings for individual meshes to get the benefit.  (Basically following in the footsteps of the tangent generation system).


lawsonamzn:
Consider turning off whitespace compare, as one of the files was ancient tab format and needed converting.

Code/Tools/SceneAPI/SceneData/Rules/UVsRule.cpp
```diff
@@ -0,0 +1,129 @@
...
+                        else
+                        {
+                            AZ_Warning(
+                                AZ::SceneAPI::Utilities::WarningWindow,
+                                false,
+                                "'" AZ_STRING_FORMAT "' is not a valid default UV generation method. Check the value of %s in your "
+                                "settings registry, and change "
+                                "it to 'LeaveSceneDataAsIs' or 'SphericalProjection'",
+                                stringFromRegistry.c_str(),
+                                AZ_STRING_ARG(regKey));
```
lumberyard-employee-dm, line 62: These format strings need to be swapped.

`"%s"` should be used for the default UV generation method string, while AZ_STRING_FORMAT(or `"%.*s"`) should be used for the reg key path.
```suggestion
                                "'%s' is not a valid default UV generation method. Check the value of " AZ_STRING_FORMAT " in your "
                                "settings registry, and change "
                                "it to 'LeaveSceneDataAsIs' or 'SphericalProjection'",
                                stringFromRegistry.c_str(),
                                AZ_STRING_ARG(regKey));
```

Code/Tools/SceneAPI/SceneData/Rules/UVsRule.cpp
```diff
@@ -0,0 +1,129 @@
...
+            void UVsRule::Reflect(AZ::ReflectContext* context)
+            {
+                AZ::SerializeContext* serializeContext = azrtti_cast<AZ::SerializeContext*>(context);
+                if (!serializeContext)
+                {
+                    return;
+                }
+
+                serializeContext->Class<UVsRule, DataTypes::IRule>()
+                    ->Version(2)
```
lumberyard-employee-dm, line 103: How is this version 2 already?

lawsonamzn: True!  I can fix this, I was iterating on it internally.

Gems/SceneProcessing/Code/Source/Generation/Components/UVsGenerator/UVsGenerateComponent.cpp
```diff
@@ -0,0 +1,254 @@
...
+        {
+        case AZ::SceneAPI::DataTypes::UVsGenerationMethod::SphericalProjection:
+            {
+                allSuccess &= AZ::UVsGeneration::Mesh::SphericalMapping::GenerateUVsSphericalMapping(meshData, dataToFill);
+                break;
+            }
+            // for future expansion - add new methods here if you want to support additional methods of UV auto generation.
+        default:
+            {
+                AZ_Assert(false, "Unknown UVs generation method selected (%d) cannot generate UVs.\n", static_cast<AZ::u32>(generationMethod));
```
lumberyard-employee-dm, line 164: You may get a warning here when building with clang due to %d being used instead of %u.
```suggestion
                AZ_Assert(false, "Unknown UVs generation method selected (%u) cannot generate UVs.\n", static_cast<AZ::u32>(generationMethod));
```

Gems/SceneProcessing/Code/Source/Generation/Components/UVsGenerator/UVsPreExportComponent.h
```diff
@@ -0,0 +1,34 @@
...
+
+namespace AZ::SceneGenerationComponents
+{
+    //! This is the component responsible for actually hooking into the scene API's processing flow
+    //! during the generation step.
+    class UVsPreExportComponent
+        : public AZ::SceneAPI::SceneCore::GenerationComponent
+    {
+    public:
+        AZ_COMPONENT(UVsPreExportComponent, "{64F79C1E-CED6-42A9-8229-6607F788C731}", AZ::SceneAPI::SceneCore::GenerationComponent)
```
lumberyard-employee-dm, line 26: FYI we have a new AZ_COMPONENT_DECL and AZ_COMPONENT_IMPL macros that allows the move the inline definitions of Component Descriptor, TypeInfo, RTTI and Allocator functions to the cpp file.

That could save a tiny bit of compile time.

Gems/SceneProcessing/Code/Source/Generation/Components/UVsGenerator/UVsGenerators/SphereMappingUVsGenerator.cpp
```diff
@@ -0,0 +1,54 @@
...
+        {
+            AZ_Trace(AZ::SceneAPI::Utilities::LogWindow, "Mesh has 0 vertex count, skipping UV generation.")
+            return true;
+        }
+
+        uvData->ReserveContainerSpace(meshData->GetVertexCount());
+
+        // calculate mesh center by bounding box center.
+        AZ::Aabb meshAabb = AZ::Aabb::CreateNull();
+        for (unsigned int vertexIndex = 0; vertexIndex < vertexCount ; ++vertexIndex)
```
lumberyard-employee-dm, line 36: extreme nitpick: extra space before semicolon
```suggestion
        for (unsigned int vertexIndex = 0; vertexIndex < vertexCount; ++vertexIndex)
```

Gems/SceneProcessing/Code/Source/Generation/Components/UVsGenerator/UVsPreExportComponent.h
lawsonamzn: Looking at the code, everything in scene api works by trawling the serialize context for classes, there's no reason for any of these components to even show up in the headers at all.  I'll move the entire components to the impl files, that way they'll be in just 1 compile unit.

lawsonamzn:
Force pushed to get it to latest dev and also refactored it so that the components are hidden in CPP files.  The headers now only declare enough forwards to spawn the component in the module and the classes they use to communicate with each other, such as the Context classes.

gadams3 (approved):
Very cool. Like you said, this should also make it easy to add other texture coordinate generators. Not sure how the spherical wrapping would look on something flat but it is clear how to add other methods that use position or normals to project onto the nearest plane.

Gems/SceneProcessing/Code/Source/Generation/Components/UVsGenerator/UVsGenerateComponent.cpp
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    bool UVsGenerateComponent::GenerateUVsForMesh(
+        AZ::SceneAPI::Containers::Scene& scene,
+        const AZ::SceneAPI::Containers::SceneGraph::NodeIndex& nodeIndex,
+        AZ::SceneAPI::DataTypes::IMeshData* meshData,
+        const AZ::SceneAPI::DataTypes::UVsGenerationMethod generationMethod,
+        const bool replaceExisting)
+    {
+        if (generationMethod == SceneAPI::DataTypes::UVsGenerationMethod::LeaveSceneDataAsIs)
```
AMZN-stankowi, line 159: Why not shortcut and early out earlier if the setting is LeaveSceneDataAsIs, in GenerateUVsData before the loop that calls GenerateUVsForMesh is run?

Gems/SceneProcessing/Code/Source/Generation/Components/UVsGenerator/UVsGenerateComponent.cpp
```diff
@@ -0,0 +1,300 @@
...
+    {
+        *outUVsData = nullptr;
+
+        AZStd::shared_ptr<SceneData::GraphData::MeshVertexUVData> uvData =
+            AZStd::make_shared<AZ::SceneData::GraphData::MeshVertexUVData>();
+
+        AZ_Assert(uvData, "Failed to allocate UV data for scene graph.");
+        if (!uvData)
+        {
+            AZ_Error(AZ::SceneAPI::Utilities::ErrorWindow, false, "Failed to allocate UV data.\n");
```
AMZN-stankowi, line 278: It's useful to provide additional context in these failure causes, to help content creators track down the issue. A suggested change to fix the error, and information about the node that generated the error.

AMZN-stankowi (approved):
It would be really helpful to have some automated tests around this new rule.

Gems/SceneProcessing/Code/Source/Generation/Components/UVsGenerator/UVsGenerateComponent.cpp
lawsonamzn: I was largely following the pattern for tangent gen, but this is a good idea, I'll move it.

lawsonamzn:
> It would be really helpful to have some automated tests around this new rule.

I was following the pattern for tangent gen.  Are there patterns to follow for tests of these generator components?  I'll do a search and see if I can find them.  If none exist, I'll see if its easy to add.  OTOH, I don't have a huge amount of bandwidth to spend on this at the moment, else I'd have done something a lot more elaborate than just sphere map :)

Gems/SceneProcessing/Code/Source/Generation/Components/UVsGenerator/UVsGenerateComponent.cpp
lawsonamzn: In this particular case you are out of RAM.  I'm actually not sure what to recommend, or whether you will actually see this message... maybe I shouldn't check... not sure! :)

AMZN-stankowi: Hmm, displaying the current node could lead to confusion there if out of ram was caused by something else. However, some messaging around that could be useful, otherwise if a content creator encounters this, they might not know the steps to take to resolve this, especially if it's consistent because it's a very large scene file.

AMZN-stankowi:
> I was following the pattern for tangent gen. Are there patterns to follow for tests of these generator components? I'll do a search and see if I can find them. If none exist, I'll see if its easy to add. OTOH, I don't have a huge amount of bandwidth to spend on this at the moment, else I'd have done something a lot more elaborate than just sphere map :)

I think the Python tests that process FBX files and compare the output to a last known good are probably enough for here: https://github.com/o3de/o3de/tree/development/AutomatedTesting/Gem/PythonTests/assetpipeline/scene_tests

Functionally these tests end up comparing the output to a last known good. I'll add some additional details on the tests and the strategy around them in the other pull request.

amzn-mike (approved):

lawsonamzn:
I'm actually not sure how useful these fbx_tests are.   They don't run automatically, and largely seem like they cause a debug export of the scene data and cause it to be compared with expected scene data?  That seems like I'd be just setting maintainers up for a huge headache anytime anyone changed anything about scenes.

I'm going to see if there's some other way to test that this basic functionality works.

lawsonamzn:
FWIW, I tried to fix and update those tests, (they were failing), only to find out that they're essentially useless. 
* Brittle
* Respond to any change by failing, even benign changes in the data. 
* Respond to any additional product asset by failing
* considers mutations of floating point numbers or hashes as warnings, not errors or asserts **so the tests dont even fail when this happns**
* currently full of failures
* does not run automatically (sandboxed)

lumberyard-employee-dm (approved):

PR #15535, opened 2023-04-04
Use O3DE_EXTERNAL_SUBDIRS cmake variable correctly.

jhmueller-huawei:
## What does this PR do?

The  O3DE_EXTERNAL_SUBDIRS` variable did not work properly for me, when I tried using it to specify a subdirectory that was not specified in the o3de manifest. This pull request fixes the issues I encountered.

## How was this PR tested?

This PR now allows me to build an engine, together with a project and an external gem that the project depends on  without having the engine and gem directories registered in the o3de manifest using the cmake variables  `O3DE_EXTERNAL_SUBDIRS` and `LY_PROJECTS`. This is useful, for example, when you want to build a project together with the engine in a CI setup since you can do it without setting up the o3de manifest. It also allows you to build multiple versions of a gem on one machine without running into issues, since registering both versions in the manifest leads to naming conflicts.

### DIFF
--- a/cmake/Subdirectories.cmake
+++ b/cmake/Subdirectories.cmake
@@ -134,7 +134,8 @@ endfunction()
 #! of the engine.json, project.json, o3de_manifest.json and any gem.json files found visiting
 function(get_all_external_subdirectories output_subdirs)
     # Gather user supplied external subdirectories via the Cache Variable
-    get_property(all_external_subdirs GLOBAL PROPERTY O3DE_EXTERNAL_SUBDIRS)
+    get_property(o3de_external_subdirs CACHE O3DE_EXTERNAL_SUBDIRS PROPERTY VALUE)
+    list(APPEND all_external_subdirs ${o3de_external_subdirs})
     get_property(manifest_external_subdirs GLOBAL PROPERTY O3DE_EXTERNAL_SUBDIRS_O3DE_MANIFEST)
     list(APPEND all_external_subdirs ${manifest_external_subdirs})
     get_property(engine_external_subdirs GLOBAL PROPERTY O3DE_EXTERNAL_SUBDIRS_ENGINE)
@@ -201,8 +202,12 @@ function(resolve_gem_dependencies object_type object_path)
 
     set(ENV{PYTHONNOUSERSITE} 1)
     string(TOLOWER ${object_type} object_type_lower)
+    get_property(user_external_subdirs CACHE O3DE_EXTERNAL_SUBDIRS PROPERTY VALUE)
+    if(user_external_subdirs)
+        set(user_external_subdir_option -ed "${user_external_subdirs}")
+    endif()
     execute_process(COMMAND 
-        ${LY_PYTHON_CMD} "${LY_ROOT_FOLDER}/scripts/o3de/o3de/cmake.py" --${object_type_lower}-path "${object_path}"
+        ${LY_PYTHON_CMD} "${LY_ROOT_FOLDER}/scripts/o3de/o3de/cmake.py" --${object_type_lower}-path "${object_path}" ${user_external_subdir_option}
         WORKING_DIRECTORY ${LY_ROOT_FOLDER}
         RESULT_VARIABLE O3DE_CLI_RESULT
         OUTPUT_VARIABLE resolved_gem_dependency_output 
--- a/scripts/o3de/o3de/cmake.py
+++ b/scripts/o3de/o3de/cmake.py
@@ -98,6 +98,7 @@ def remove_gem_dependency(cmake_file: pathlib.Path,
 def resolve_gem_dependency_paths(
         engine_path:pathlib.Path,
         project_path:pathlib.Path,
+        external_subdirectories:str or list or None,
         resolved_gem_dependencies_output_path:pathlib.Path or None):
     """
     Resolves gem dependencies for the given engine and project and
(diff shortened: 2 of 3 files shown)


cmake/Subdirectories.cmake
```diff
@@ -134,7 +134,8 @@ endfunction()
 #! of the engine.json, project.json, o3de_manifest.json and any gem.json files found visiting
 function(get_all_external_subdirectories output_subdirs)
     # Gather user supplied external subdirectories via the Cache Variable
-    get_property(all_external_subdirs GLOBAL PROPERTY O3DE_EXTERNAL_SUBDIRS)
+    get_property(o3de_external_subdirs GLOBAL PROPERTY O3DE_EXTERNAL_SUBDIRS)
+    list(APPEND all_external_subdirs ${o3de_external_subdirs})
```
lumberyard-employee-dm, line 138: Actually can you fix this to grab the O3DE_EXTERNAL_SUBDIRS from the CACHE variable using CACHE PROPERTY
```suggestion
    get_property(o3de_external_subdirs CACHE PROPERTY O3DE_EXTERNAL_SUBDIRS)
    list(APPEND all_external_subdirs ${o3de_external_subdirs})
```

cmake/Subdirectories.cmake
```diff
@@ -202,7 +203,7 @@ function(resolve_gem_dependencies object_type object_path)
     set(ENV{PYTHONNOUSERSITE} 1)
     string(TOLOWER ${object_type} object_type_lower)
     execute_process(COMMAND 
-        ${LY_PYTHON_CMD} "${LY_ROOT_FOLDER}/scripts/o3de/o3de/cmake.py" --${object_type_lower}-path "${object_path}"
+        ${LY_PYTHON_CMD} "${LY_ROOT_FOLDER}/scripts/o3de/o3de/cmake.py" --${object_type_lower}-path "${object_path}" -ed "${O3DE_EXTERNAL_SUBDIRS}"
```
lumberyard-employee-dm, line 206: To make sure we are retrieving O3DE_EXTERNAL_SUBDIRS from the cache and not some override local variable we can use `get_property(<var name> CACHE PROPERTY O3DE_EXTERNAL_SUBDIRS)`

We also should only specify the `-ed` option if the list of user provided subdirectories list is not empty

```suggestion
        get_property(user_external_subdirs CACHE PROPERTY O3DE_EXTERNAL_SUBDIRS)
        if(user_external_subdirs)
            set(user_external_subdir_option -ed "${user_external_subdirs}")
        endif()
        ${LY_PYTHON_CMD} "${LY_ROOT_FOLDER}/scripts/o3de/o3de/cmake.py" --${object_type_lower}-path "${object_path}" ${user_external_subdir_option}
```

scripts/o3de/o3de/cmake.py
```diff
@@ -208,6 +211,8 @@ def add_parser_args(parser):
                        help='The path to the project.')
     group.add_argument('-ep', '--engine-path', type=pathlib.Path, required=False,
                        help='The path to the engine.')
+    group.add_argument('-ed', '--external-subdirs', type=str, required=False,
+                       help='Additional list of subdirectories.')
```
lumberyard-employee-dm, line 215: Can you also add a the `nargs='*'` option so this option can be used 0 or more times?

scripts/o3de/o3de/cmake.py
```diff
@@ -158,7 +159,8 @@ def resolve_gem_dependency_paths(
     all_gems_json_data = manifest.get_gems_json_data_by_name(engine_path=engine_path, 
                                                              project_path=project_path, 
                                                              include_manifest_gems=True, 
-                                                             include_engine_gems=True)
+                                                             include_engine_gems=True,
+                                                             external_subdirectories=external_subdirs.split(';') if external_subdirs else None)
```
lumberyard-employee-dm, line 163: A safer approach here is to split if the external_subdirs is an instance of str, otherwise assume it is a list already
```suggestion
                                                             external_subdirectories=external_subdirs.split(';') if isinstance(external_subdirs, str) else external_subdirs.)
```

lumberyard-employee-dm:
> I agree with all of @lumberyard-employee-dm suggestions, but otherwise, looks like a good addition. Is this technically a fix for something that is broken? If so we may want to add it to `stabilization/2305` so it is included in the next release.

Well the [O3DE_EXTERNAL_SUBDIRS](https://github.com/o3de/o3de/blob/19619553dd0f6600d6b7e2851710cd59f11f477b/cmake/GeneralSettings.cmake#L15-L16) variable is used to allow users to allow CMake to perform`add_subdirectory` calls on gems that aren't registered in any o3de manifest file.

It is useful for building additional gems as part of a CI/CD pipeline.

How ever it should be noted that because the external subdirectories aren't registered with a manifest, the metadata information such as the gem source root and the gem name can't be queried at runtime. 

So when running a C++ application, while the gem can be active and visited using [VisitActiveGems](https://github.com/o3de/o3de/blob/19619553dd0f6600d6b7e2851710cd59f11f477b/Code/Framework/AzCore/AzCore/Settings/SettingsRegistryMergeUtils.h#L360-L365) assuming it is a dependency of another gem, it's gem.json will not be able to found when visiting the [manifest files](https://github.com/o3de/o3de/blob/19619553dd0f6600d6b7e2851710cd59f11f477b/Code/Framework/AzCore/AzCore/Settings/SettingsRegistryMergeUtils.h#L383-L410)

jhmueller-huawei:
I fixed all issues and the CI finished (https://jenkins.build.o3de.org/blue/organizations/jenkins/O3DE/detail/PR-15535/4/tests/), I don't know, why it still states that it is waiting for the status to be reported?!

lumberyard-employee-dm:
It looks like the build wasn't attached to the latest commit for some reason.

lumberyard-employee-dm (approved):

jhmueller-huawei:
Thanks, works now :)

PR #15539, opened 2023-04-04
Fix open in other asset browser

sphrose:
Add open in other browser to other view modes.

## What does this PR do?

This PR resolves https://github.com/o3de/o3de/issues/14841. It now selects the correct folder in the treeview and asset in the the thumbnail, expanded table views and simple treeview.

## How was this PR tested?

Tested manually with and without the `ed_useWIPAssetBrowserDesign` flag set.


Code/Editor/AzAssetBrowser/AzAssetBrowserRequestHandler.cpp
```diff
@@ -539,16 +539,43 @@ void AzAssetBrowserRequestHandler::AddContextMenuActions(QWidget* caller, QMenu*
                     });
             }
 
-            menu->addAction(QObject::tr("Open in another Asset Browser"), [fullFilePath, caller](){
+            menu->addAction(QObject::tr("Open in another Asset Browser"), [fullFilePath, thumbnailView, expandedTableView, caller](){
                 auto* browser1 = qobject_cast<AzAssetBrowserWindow*>(QtViewPaneManager::instance()->OpenPane(LyViewPane::AssetBrowser)->Widget());
```
pollend, line 543: while your here you can use a better variable name?

Code/Editor/AzAssetBrowser/AzAssetBrowserRequestHandler.cpp
```diff
@@ -539,16 +539,43 @@ void AzAssetBrowserRequestHandler::AddContextMenuActions(QWidget* caller, QMenu*
...
                 const QString name2 = QString("%1 (2)").arg(LyViewPane::AssetBrowser);
                 auto* browser2 = qobject_cast<AzAssetBrowserWindow*>(QtViewPaneManager::instance()->OpenPane(name2)->Widget());
```
pollend, line 545: not sure what name2 is suppose to be

Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
```diff
@@ -786,32 +790,48 @@ static void ExpandTreeToIndex(QTreeView* treeView, const QModelIndex& index)
...
-                treeView->scrollTo(index);
-                treeView->setCurrentIndex(index);
+                    QTreeView* treeView = m_ui->m_assetBrowserTreeViewWidget;
+                    ExpandTreeToIndex(treeView, modelIndex);
 
-                treeView->selectionModel()->select(index, QItemSelectionModel::ClearAndSelect);
-            });
+                    treeView->scrollTo(modelIndex);
+                    treeView->setCurrentIndex(modelIndex);
+                    treeView->selectionModel()->select(modelIndex, QItemSelectionModel::ClearAndSelect);
```
dtamkin1, line 832: I know this is in the old AB, but it seems like this would break multi-select in the tree view

dtamkin1 (approved):

Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
sphrose: It's in the old AB and the existing code.

pollend (approved):

Issue #15549, opened 2023-04-04
AssetJobsFloodTest.ContainerCoreTest_BasicDependencyManagement_Success test failed in latest development/stabilization builds


jckand-amzn:
**Describe the bug**
AssetJobsFloodTest.ContainerCoreTest_BasicDependencyManagement_Success test failed in latest development/stabilization builds

**Failed Jenkins Job Information:**
https://jenkins.build.o3de.org/job/O3DE/job/stabilization%252F2305/68/testReport/junit/(root)/AssetJobsFloodTest/ContainerCoreTest_BasicDependencyManagement_Success/

https://jenkins.build.o3de.org/job/O3DE/job/development/4204/testReport/junit/(root)/AssetJobsFloodTest/ContainerCoreTest_BasicDependencyManagement_Success/

**Attachments**
```
Error Message

/data/workspace/o3de/Code/Framework/AzCore/Tests/Asset/AssetManagerLoadingTests.cpp:1082
Expected equality of these values:
  m_testAssetManager->GetRemainingJobs()
    Which is: 1
  0

Stacktrace

/data/workspace/o3de/Code/Framework/AzCore/Tests/Asset/AssetManagerLoadingTests.cpp:1082
Expected equality of these values:
  m_testAssetManager->GetRemainingJobs()
    Which is: 1
  0
```

AMZN-stankowi:
I've adjusted the timeout, I'll leave this ticket open another day or two before closing it, in case it wasn't long enough or it was failing for other reasons https://github.com/o3de/o3de/pull/15556

lumberyard-employee-dm:
> I've adjusted the timeout, I'll leave this ticket open another day or two before closing it, in case it wasn't long enough or it was failing for other reasons #15556
> 
> I went with a change to adjust the timeout because it's the lowest risk change for stabilization. Having these tests that rely on timeouts isn't necessarily ideal, but refactoring it will take a lot of time.

Should this test be disabled then since it can't 100% deterministically pass

lumberyard-employee-dm:
> > I've adjusted the timeout, I'll leave this ticket open another day or two before closing it, in case it wasn't long enough or it was failing for other reasons #15556
> > I went with a change to adjust the timeout because it's the lowest risk change for stabilization. Having these tests that rely on timeouts isn't necessarily ideal, but refactoring it will take a lot of time.
> 
> Should this test be disabled then since it can't 100% deterministically pass?

[...]

PR #15556, opened 2023-04-04
Adjusted a timeout on Linux for a flaky test.

AMZN-stankowi:
## What does this PR do?

A test that uses this timeout has been infrequently failing on Linux, likely due to this timeout. https://github.com/o3de/o3de/issues/15549

This is the lowest risk change to correct it. The test using this timeout doesn't wait until the timeout all the time, only if the assets aren't processed in time.

## How was this PR tested?

This is a test intended for Jenkins, so we'll watch what this does for intermittent test failures.


Kadino (approved):
This remains a reasonably short timeout

gadams3 (approved):

PR #15566, opened 2023-04-05
Removed unused legacy image histogram classes

cgalvan:
## What does this PR do?

Removed unused legacy image histogram classes.

## How was this PR tested?

Ran/built the Editor and verified unit and automated tests still pass.


AMZN-daimini (approved):

amzn-tmryan (approved):

cgalvan:
> Totally reasonable to delete it, but also kind of a shame, since we definitely could _use_ histogram controls in some places. :(

True. This class as-is was restricted to just a subset of image formats that Qt supports, but we could always resurrect it in the future or use it as a springboard if we want to add a histogram control in the future :)

PR #15571, opened 2023-04-05
Switch trace processing to use threads for editor/server connection without impacting AP or asset builders

mbalfour-amzn:
## What does this PR do?

This improves upon a previous PR by introducing threaded trace printing for the editor/server connection instead of relying on well-placed Pump() calls. Also fixed a tiny bug where the tick bus wouldn't get connected to if server logging wasn't enabled, so the detection of unexpected server disconnects also wouldn't happen in that case.

In adding this support, I also left the previous pump-based trace printing because the Asset Builders rely on it pretty deeply, since they expect the trace printing to occur on specific threads. The change was spiraling out too far to get the Asset Builders to use the threaded trace printing, so I left them working the same as before with a comment explaining the situation.

## How was this PR tested?

Rebuilt the entire cache in the AP to verify that builder logging still works.
Ran Multiplayersample and verified that there are still no timeouts when starting up the Editor/server connection.

### DIFF
--- a/Code/Framework/AzFramework/AzFramework/Process/ProcessCommunicatorTracePrinter.h
+++ b/Code/Framework/AzFramework/AzFramework/Process/ProcessCommunicatorTracePrinter.h
@@ -8,6 +8,8 @@
 
 #pragma once
 
+#include <AzCore/std/parallel/thread.h>
+#include <AzCore/std/parallel/atomic.h>
 #include <AzFramework/Process/ProcessCommunicator.h>
 
 //! ProcessCommunicatorTracePrinter listens to stderr and stdout of a running process and writes its output to the AZ_Trace system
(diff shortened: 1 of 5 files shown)


Code/Framework/AzFramework/AzFramework/Process/ProcessCommunicatorTracePrinter.h
```diff
@@ -8,33 +8,59 @@
...
+    //! End the pump thread and flush the current data.
+    void FlushInternal();
+
     AZStd::string m_window;
     AzFramework::ProcessCommunicator* m_communicator;
     char m_streamBuffer[128];
     AZStd::string m_stringBeingConcatenated;
     AZStd::string m_errorStringBeingConcatenated;
+    AZStd::thread m_pumpThread;
+    AZStd::atomic_bool m_runThread = true;
```
lumberyard-employee-dm, line 64: I think m_runThread should default to `false`, since the thread doesn't get started until the first call to `StartPumpThread`

AMZN-Gene (approved):

AMZN-Gene:
> Can we have a slightly more descriptive title as this does impact AssetProcessor as well? Want to make it easy for future maintainer to find this.

Noice! Improvement for anyone using ProcessCommunicator

mbalfour-amzn:
> Can we have a slightly more descriptive title as this does impact AssetProcessor as well? Want to make it easy for future maintainer to find this.

Well, assuming I did my job correctly, it does _not_ impact AssetProcessor. :P That's the whole point of leaving the polling pathway in the code. i'll change the title though if that helps

lumberyard-employee-dm (approved):

PR #15572, opened 2023-04-05
Added a simple welcome screen to the Asset Processor, to make it easier to onboard to this tool.

AMZN-stankowi:
## What does this PR do?

Adds a simple welcome screen to the Asset Processor. The goal here is to provide a softer landing point in this tool, especially for new users, who find the previous default tab, the Jobs tab, confusing and overwhelming.

A screenshot of what this looks like:

image

This is in draft until I hear back from UX.

## How was this PR tested?

Ran locally, saw the new screen.

### DIFF
--- a/Code/Tools/AssetProcessor/native/ui/MainWindow.ui
+++ b/Code/Tools/AssetProcessor/native/ui/MainWindow.ui
@@ -337,7 +337,61 @@
         <property name="currentIndex">
          <number>0</number>
         </property>
-        <widget class="QSplitter" name="jobDialogSplitter">
+         <widget class="QWidget" name="WelcomePage">
+           <layout class="QVBoxLayout" name="WelcomeVerticalLayout">
+             <item>
+              <widget class="QLabel" name="welcomeScreenTitle">
+               <property name="font">
+                <font>
+                 <pointsize>10</pointsize>
+                 <weight>75</weight>
+                 <bold>true</bold>
+                </font>
+               </property>
+               <property name="text">
+                <string># Welcome to the O3DE Asset Processor</string>
+               </property>
+               <property name="textFormat">
+                <enum>Qt::MarkdownText</enum>
+               </property>
+              </widget>
+             </item>
+             <item>
+               <widget class="QLabel" name="WelcomeDescriptionLabel">
+                 <property name="text">
+                   <string>This tool converts an O3DE project's source assets to game ready content.</string>
+                 </property>
+               </widget>
+             </item>
+             <item>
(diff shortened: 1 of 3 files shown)


Code/Tools/AssetProcessor/native/ui/MainWindow.ui
```diff
@@ -337,7 +337,61 @@
...
+             </item>
+             <item>
+               <widget class="QLabel" name="WelcomeDescriptionLabel">
+                 <property name="text">
+                   <string>This tool converts an O3DE project's source assets to game ready content.</string>
+                 </property>
+               </widget>
+             </item>
+             <item>
+               <widget class="QLabel" name="WelcomeDescriptionLabel">
```
AMZN-stankowi, line 367: I've duplicated this widget name, adding this comment so I don't forget to fix it

gadams3 (commented):
Simple but brilliant. We should do that for the other tools as well.

AMZN-stankowi:
UX feedback: Change the word "game" to "runtime ready"

AMZN-stankowi:
I had some conversations with members of the sig-ui-ux about this pull request. Everyone's in agreement that this is a good first step, but we can do a lot more to improve the onboarding experience of this tool. I've captured some of those notes here in this issue: https://github.com/o3de/o3de/issues/15601

AMZN-alexpete (commented):
I like this, but does it mean that every time the asset processer starts every user will see this screen and have to click on `Jobs` to see the Jobs screen?  One recommendation I have is to add a "show this screen on startup [ x ]" on this page so users can switch back to `Jobs` as the default

lmbr-pip (approved):
Did we think about having stored setting so you only see this once?

amzn-mike (approved):

AMZN-stankowi:
> I like this, but does it mean that every time the asset processer starts every user will see this screen and have to click on `Jobs` to see the Jobs screen? One recommendation I have is to add a "show this screen on startup [ x ]" on this page so users can switch back to `Jobs` as the default

How big of a deal do you think this is? Besides a timing issue (the setting + UX for the setting grows the scope of this change), the main reason I didn't do that is there isn't a consistent first tab I want to start on when I launch asset processor. Based on my intended workflow (see https://github.com/o3de/o3de/issues/15601 for a list of workflows), I'm loading this tool to open different tabs and the Jobs tab isn't really the primary tab for most workflows.

So I'm clicking off to another tab most of the time already, having a flag to default to jobs doesn't save most of the workflows much time.

cgalvan (approved):

PR #15575, opened 2023-04-05
Fixes dropping a folder on itself in Asset Browser

jjjoness:
## What does this PR do?

Fixes a bug that happened because the editor allowed a folder to be dropped on itself in the Asset Browser Treeview.

## How was this PR tested?

Tried dropping a folder on itself and it no longer does anything.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Views/AssetBrowserTreeView.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Views/AssetBrowserTreeView.cpp
@@ -181,7 +181,7 @@ namespace AzToolsFramework
             CaptureTreeViewSnapshot();
         }
 
-        void AssetBrowserTreeView::dropEvent([[maybe_unused]] QDropEvent* event)
+        void AssetBrowserTreeView::dropEvent(QDropEvent* event)
         {
             QModelIndex sourceIndex = currentIndex();
             QModelIndex targetIndex = indexAt(event->pos());
@@ -189,6 +189,7 @@ namespace AzToolsFramework
             {
                 event->setDropAction(Qt::IgnoreAction);
                 event->accept();
+                return;
             }
             QTreeView::dropEvent(event);
         }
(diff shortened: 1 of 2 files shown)


forkercat (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Views/AssetBrowserTreeView.cpp
```diff
@@ -181,6 +181,18 @@ namespace AzToolsFramework
             CaptureTreeViewSnapshot();
         }
 
+        void AssetBrowserTreeView::dropEvent([[maybe_unused]] QDropEvent* event)
```
cgalvan, line 184: Nit: `event` is used below so can get rid of the `[[maybe_unused]]`

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Views/AssetBrowserTreeView.cpp
```diff
@@ -181,6 +181,18 @@ namespace AzToolsFramework
...
+        {
+            QModelIndex sourceIndex = currentIndex();
+            QModelIndex targetIndex = indexAt(event->pos());
+            if ((sourceIndex.internalPointer() == targetIndex.internalPointer()) && (sourceIndex.row() == targetIndex.row()))
+            {
+                event->setDropAction(Qt::IgnoreAction);
+                event->accept();
```
cgalvan, line 191: Should we return in this case instead of letting the base class continue to handle the drop event?

cgalvan (approved):

dtamkin1 (commented):
Does the problem also exist in when dropping folders on themselves in the thumbnail view and table view? If so, we should fix it there as well

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Views/AssetBrowserTreeView.cpp
jjjoness: Oops!

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Views/AssetBrowserTreeView.cpp
jjjoness: I'll check, if I coded it the way Qt's docs say, it didn't work at all.

jjjoness: Yes, returning still works so added it.

Issue #15561, opened 2023-04-05
Console is flooded with MeshDrawPacket errors when opening a level or resizing the Viewport


LB-KacperKapusta:
**Describe the bug**
MeshDrawPacket errors (8x) can be observed in the Console upon opening a level or creating a new one. Resizing the Viewport makes the following error spam the Console even more: `[Error] (MeshDrawPacket) - All draw items in a draw packet need to share the same root constants layout. This means that each pass (e.g. Depth, Shadows, Forward, MotionVectors) for a given materialtype should use the same layout.`

The same issue occurs on both Linux and Windows.

Please note that the issue does not occur in the latest Stabilization/2305 build.

Refer to the attached video and log for more information.

**Steps to reproduce**
1. Launch the Editor.
2. Create a new level or open an existing one.
3. Resize the Viewport.

**Expected behavior**
No errors appear in the Console when opening a level or resizing the Viewport.

**Actual behavior**
Console is flooded with MeshDrawPacket errors when opening a level or resizing the Viewport.

**Screenshots/Video**

https://user-images.githubusercontent.com/86953659/230054766-8305795a-e6a9-471c-afdb-ed342a5516fe.mp4

**Found in Branch**
Development

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
[475973d](https://github.com/o3de/o3de/commit/475973d5c2586fa2b3026b7ccddbda4296a35de4)

**Desktop/Device:**
 - Device: PC
 - OS: Windows
 - Version: 10
 - CPU: Intel Core i5-9600K
 - GPU: NVIDIA GeForce RTX 2070
 - Memory: 16GB

**Additional context**
[Editor.log](https://github.com/o3de/o3de/files/11157657/Editor.log)

LB-TomaszSerwanski:
Issue verified as fixed on the latest Development (e495f15).

PR #15583, opened 2023-04-05
Fix for GHI-15561 : MeshDrawPacket errors in console

amzn-tommy:
## What does this PR do?

Fixes https://github.com/o3de/o3de/issues/15561
Fix error message, which wasn't handling the case where the pipeline's rootConstantLayout was non-null, but its size was 0. The error message should not fire in that case.

## How was this PR tested?

Ran the editor on latest development, and the error no longer occurs.

### DIFF
--- a/Gems/Atom/RPI/Code/Source/RPI.Public/MeshDrawPacket.cpp
+++ b/Gems/Atom/RPI/Code/Source/RPI.Public/MeshDrawPacket.cpp
@@ -208,6 +208,11 @@ namespace AZ
             return false;
         }
 
+        static bool HasRootConstants(const RHI::ConstantsLayout* rootConstantsLayout)
+        {
+            return rootConstantsLayout && rootConstantsLayout->GetDataSize() > 0;
+        }
+
         bool MeshDrawPacket::DoUpdate(const Scene& parentScene)
         {
             const ModelLod::Mesh& mesh = m_modelLod->GetMeshes()[m_modelLodMeshIndex];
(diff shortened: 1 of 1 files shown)


galibzon (approved):

Gems/Atom/RPI/Code/Source/RPI.Public/MeshDrawPacket.cpp
```diff
@@ -391,8 +397,8 @@ namespace AZ
                 {
                     AZ_Error(
                         "MeshDrawPacket",
-                        (!m_rootConstantsLayout && !pipelineStateDescriptor.m_pipelineLayoutDescriptor->GetRootConstantsLayout()) ||
-                        (m_rootConstantsLayout && m_rootConstantsLayout->GetHash() == pipelineStateDescriptor.m_pipelineLayoutDescriptor->GetRootConstantsLayout()->GetHash()),
+                        (!m_rootConstantsLayout && !HasRootConstants(rootConstantsLayout)) ||
```
VickyAtAZ, line 400: I think this condition should be an assert since it may only fail if there is an implementation issue

VickyAtAZ (approved):
A minor comment

Gems/Atom/RPI/Code/Source/RPI.Public/MeshDrawPacket.cpp
```diff
@@ -391,8 +397,8 @@ namespace AZ
...
+                        (m_rootConstantsLayout && rootConstantsLayout && m_rootConstantsLayout->GetHash() == rootConstantsLayout->GetHash()),
```
rgba16f, line 401: Nit: add a function called `RootConstantsLayoutIsSame` to replace this whole line

rgba16f (approved):

Gems/Atom/RPI/Code/Source/RPI.Public/MeshDrawPacket.cpp
amzn-tommy: I think it's kind of a grey area depending on if you consider a shader more 'code' or 'data/asset'.

VickyAtAZ: I mean this part is implementation issue `(!m_rootConstantsLayout && !HasRootConstants(rootConstantsLayout)) `. I dont think it depends on shader. 
The second part is an error '(m_rootConstantsLayout && m_rootConstantsLayout->GetHash() == pipelineStateDescriptor.m_pipelineLayoutDescriptor->GetRootConstantsLayout()->GetHash()'

amzn-tommy: That case will be hit if the first DrawItem has no root constant in the shader (m_rootConstantsLayout will be null), but a subsequent DrawItem does, so that can still come from the shader.

PR #15586, opened 2023-04-05
Removed legacy unused 2d viewports

cgalvan:
## What does this PR do?

Removed legacy unused 2d viewport classes along with their usage that had been ifdef'd out with the `FEATURE_ORTHOGRAPHIC_VIEW` flag. These classes were written for the old cry renderer, and relied on the legacy object manager to paint the 2d views, along with some cry specific code that had been removed previously so that they could still compile when the atom renderer was turned on.

The hope was to update them to work with the new atom renderer, but in their current state would need to be rewritten to actually work with the atom renderer. Enabling them with the compile flag builds, but crashes as soon as you try to open a view. They are also tightly coupled with the legacy object manager, which we are in the process of removing, so I have opted to delete these classes instead of refactoring them to compile without the object manager.

## How was this PR tested?

Built/ran the Editor and verified the viewport still works as expected.


gadams3 (approved):

dtamkin1 (approved):

AMZN-daimini (approved):

rgba16f:
Why did this get closed?

PR #15587, opened 2023-04-05
DPE: BoundContainer fixes and improvements

amzn-tmryan:
**Description**
Prior to this PR, there were some issues with containers in the DPE such as broken nested container action buttons and the inability to remove nested associative containers elements from their parent container.

This PR fixes those issues and makes some improvements to the `ReflectionAdapter`'s `BoundContainer` including:
- Fix for broken nested container action buttons which weren't working due to BoundContainer combining the state for root containers, nested containers, and non-container elements
- Fix for using the IDataContainer API with a nested container by storing its element instance override (from the LegacyReflectionBridge) separately from its container instance 
- Refactored BoundContainer's element responsibilities into struct ContainerElement
  - A ContainerNode wrapper is now being used in the ReflectionAdapterReflectionImpl m_containers prefix tree so access to a BoundContainer and/or a ContainerEntry can be had
  - This solves issues brought about by having both container and element behavior contained in the same struct
- Created BoundContainer/ContainerElement factory methods to simplify their creation and maintenance in the ReflectionAdapter
- Updated usage of IDataContainer API to use the element instances it expects; previously we were passing in the key instance ptr only, which happened to work in some cases but caused undefined behavior in others
- Remove unused Add/Remove/ClearNotify notifications from the BoundContainer methods
  - These were unused, and an attempt at implementing them introduced bugs
  - They also now live in the Container node inside of PropertyNodes.h for future use
- Now passes the class size to VisitValue for instance hashes instead of using size of void*

Terms for reference:
Nested Container = `AZStd::unordered_map<int, AZStd::vector<int>>`
Non-Nested Container = `AZStd::unordered_map<int, int>`
Non-Container elements = `typename AZStd::unordered_map<int, int>::value_type` i.e the type is `AZStd::pair<const int, int>` or for `AZStd::vector<int>` the element type is `int`

**Testing**
- Verified that all action buttons function properly in the debug DPE standalone app
- Verified that all container functions such as add, remove, and clear function properly in the debug DPE standalone app
- Verified DPE containers and the DPE in general function as before in the editor

### DIFF
--- a/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/PropertyEditorNodes.h
+++ b/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/PropertyEditorNodes.h
@@ -49,7 +49,7 @@ namespace AZ::DocumentPropertyEditor::Nodes
         static constexpr AZStd::string_view Name = "Adapter";
         static constexpr auto QueryKey = CallbackAttributeDefinition<void(DocumentAdapterPtr*, AZ::Dom::Path)>("QueryKey");
         static constexpr auto AddContainerKey = CallbackAttributeDefinition<void(DocumentAdapterPtr*, AZ::Dom::Path)>("AddContainerKey");
-        static constexpr auto RejectContainerKey = CallbackAttributeDefinition<void(DocumentAdapterPtr*, AZ::Dom::Path)>("RejectContainerKey");
+        static constexpr auto RejectContainerKey = CallbackAttributeDefinition<void(AZ::Dom::Path)>("RejectContainerKey");
 
         //! Use this callback attribute if there is need to enable/disable an adapter's nodes at runtime.
         static constexpr auto SetNodeDisabled =
@@ -102,6 +102,13 @@ namespace AZ::DocumentPropertyEditor::Nodes
         FinishedEdit,
     };
 
+    struct Container : NodeWithVisiblityControl
+    {
+        static constexpr auto AddNotify = CallbackAttributeDefinition<void()>("AddNotify");
+        static constexpr auto RemoveNotify = CallbackAttributeDefinition<void()>("RemoveNotify");
+        static constexpr auto ClearNotify = CallbackAttributeDefinition<void()>("ClearNotify");
+    };
+
     //! PropertyEditor: A property editor, of a type dictated by its "type" field,
     //! that can edit an associated value.
     struct PropertyEditor : NodeWithVisiblityControl
@@ -147,11 +154,6 @@ namespace AZ::DocumentPropertyEditor::Nodes
 
         static constexpr auto ChangeNotify = CallbackAttributeDefinition<PropertyRefreshLevel()>("ChangeNotify");
         static constexpr auto RequestTreeUpdate = CallbackAttributeDefinition<void(PropertyRefreshLevel)>("RequestTreeUpdate");
-
-        // Container attributes
-        static constexpr auto AddNotify = CallbackAttributeDefinition<void()>("AddNotify");
-        static constexpr auto RemoveNotify = CallbackAttributeDefinition<void(size_t)>("RemoveNotify");
-        static constexpr auto ClearNotify = CallbackAttributeDefinition<void()>("ClearNotify");
     };
 
     struct UIElement : PropertyEditor
--- a/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/PropertyEditorNodes.h
+++ b/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/PropertyEditorNodes.h
@@ -102,10 +102,13 @@ namespace AZ::DocumentPropertyEditor::Nodes
         FinishedEdit,
     };
 
+    //! Container: A node representing a container instance.
     struct Container : NodeWithVisiblityControl
     {
+        // These notify callback attributes may be invoked when a container is modified and handled
+        // alongside other messages in the adapter's message handler.
         static constexpr auto AddNotify = CallbackAttributeDefinition<void()>("AddNotify");
-        static constexpr auto RemoveNotify = CallbackAttributeDefinition<void()>("RemoveNotify");
+        static constexpr auto RemoveNotify = CallbackAttributeDefinition<void(size_t /*index*/)>("RemoveNotify");
         static constexpr auto ClearNotify = CallbackAttributeDefinition<void()>("ClearNotify");
     };
 
--- a/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/PropertyEditorNodes.h
+++ b/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/PropertyEditorNodes.h
@@ -108,7 +108,7 @@ namespace AZ::DocumentPropertyEditor::Nodes
         // These notify callback attributes may be invoked when a container is modified and handled
         // alongside other messages in the adapter's message handler.
         static constexpr auto AddNotify = CallbackAttributeDefinition<void()>("AddNotify");
-        static constexpr auto RemoveNotify = CallbackAttributeDefinition<void(size_t /*index*/)>("RemoveNotify");
+        static constexpr auto RemoveNotify = CallbackAttributeDefinition<void(size_t index)>("RemoveNotify");
         static constexpr auto ClearNotify = CallbackAttributeDefinition<void()>("ClearNotify");
     };
 
--- a/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
+++ b/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
@@ -53,11 +53,13 @@ namespace AZ::DocumentPropertyEditor
                 AZ::SerializeContext::IDataContainer* parentContainer,
                 void* parentContainerInstance,
                 AZ::SerializeContext::IDataContainer* container,
-                void* containerInstance)
+                void* containerInstance,
+                void* overrideInstance)
                 : m_parentContainer(parentContainer)
                 , m_parentContainerInstance(parentContainerInstance)
                 , m_container(container)
                 , m_containerInstance(containerInstance)
+                , m_elementOverrideInstance(overrideInstance)
             {
             }
 
@@ -76,23 +78,28 @@ namespace AZ::DocumentPropertyEditor
                     auto parentContainerInstanceValue = attributes.Find(AZ::Reflection::DescriptorAttributes::ParentContainerInstance);
                     auto parentContainerInstance = AZ::Dom::Utils::ValueToTypeUnsafe<void*>(*parentContainerInstanceValue);
 
+                    // Check if this element is actually standing in for a direct child of a container. This is used in scenarios like
+                    // maps, where the direct children are actually pairs of key/value, but we need to only show the value as an
+                    // editable item who pretends that they can be removed directly from the container
+                    void* overrideInstance = nullptr;
+                    auto containerElementOverrideValue = attributes.Find(AZ::Reflection::DescriptorAttributes::ContainerElementOverride);
+                    if (containerElementOverrideValue)
+                    {
+                        overrideInstance = AZ::Dom::Utils::ValueToTypeUnsafe<void*>(*containerElementOverrideValue);
+                    }
+
                     if (containerValue)
                     {
                         auto container = AZ::Dom::Utils::ValueToTypeUnsafe<AZ::SerializeContext::IDataContainer*>(*containerValue);
 
                         // This is a nested container, so store data to both the parent container and nested container
-                        return BoundContainer(parentContainer, parentContainerInstance, container, instance);
+                        return BoundContainer(parentContainer, parentContainerInstance, container, instance, overrideInstance);
                     }
                     else
                     {
-                        // Check if this element is actually standing in for a direct child of a container. This is used in scenarios like
-                        // maps, where the direct children are actually pairs of key/value, but we need to only show the value as an
-                        // editable item who pretends that they can be removed directly from the container
-                        auto containerElementOverrideValue =
-                            attributes.Find(AZ::Reflection::DescriptorAttributes::ContainerElementOverride);
-                        if (containerElementOverrideValue)
+                        if (overrideInstance)
                         {
-                            instance = AZ::Dom::Utils::ValueToTypeUnsafe<void*>(*containerElementOverrideValue);
+                            instance = overrideInstance;
                         }
 
                         // This is a non-container child element, so we only store data for the parent container and the element
@@ -216,20 +223,33 @@ namespace AZ::DocumentPropertyEditor
 
             void OnRemoveElement(ReflectionAdapterReflectionImpl* impl, const AZ::Dom::Path& path)
             {
-                auto parentBoundContainer = impl->m_containers.ValueAtPath(path, Dom::PrefixTreeMatch::ExactPath);
-                if (parentBoundContainer)
-                {
-
-                }
-
+                void* instanceToRemove = nullptr;
                 if (m_elementInstance == nullptr)
                 {
-                    m_parentContainer->RemoveElement(m_parentContainerInstance, m_containerInstance, impl->m_serializeContext);
+                    instanceToRemove = m_containerInstance;
+                    auto associativeContainer = m_parentContainer->GetAssociativeContainerInterface();
+                    if (associativeContainer)
+                    {
+                        // Use the override instance if specified since it represents the
+                        // element pair instance that the IDataContainer API is expecting
+                        void* keyInstance = m_elementOverrideInstance ? m_elementOverrideInstance : m_containerInstance;
+                        const AZ::SerializeContext::ClassElement* containerClassElement =
+                            m_parentContainer->GetElement(m_parentContainer->GetDefaultElementNameCrc());
+                        instanceToRemove =
+                            associativeContainer->GetElementByKey(m_parentContainerInstance, containerClassElement, keyInstance);
+                    }
                 }
                 else
                 {
-                    m_parentContainer->RemoveElement(m_parentContainerInstance, m_elementInstance, impl->m_serializeContext);
+                    instanceToRemove = m_elementInstance;
+                    auto associativeContainer = m_parentContainer->GetAssociativeContainerInterface();
+                    if (associativeContainer)
+                    {
+                        instanceToRemove = associativeContainer->GetElementByKey(m_parentContainerInstance, nullptr, m_elementInstance);
+                    }
                 }
+
+                m_parentContainer->RemoveElement(m_parentContainerInstance, instanceToRemove, impl->m_serializeContext);
 
                 auto containerNode = GetContainerNode(impl, path);
                 Nodes::PropertyEditor::ChangeNotify.InvokeOnDomNode(containerNode);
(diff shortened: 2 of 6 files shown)


Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
```diff
@@ -405,20 +576,9 @@ namespace AZ::DocumentPropertyEditor
...
-                // maps, where the direct children are actually pairs of key/value, but we need to only show the value as an editable item
-                // who pretends that they can be removed directly from the container
-                auto containerElementOverrideAttribute = attributes.Find(AZ::Reflection::DescriptorAttributes::ContainerElementOverride);
-                if (containerElementOverrideAttribute && !containerElementOverrideAttribute->IsNull())
-                {
-                    instance = AZ::Dom::Utils::ValueToTypeUnsafe<void*>(*containerElementOverrideAttribute);
-                }
-
-                m_containers.SetValue(m_builder.GetCurrentPath(), BoundContainer{ parentContainer, parentContainerInstance, instance });
+                m_containers.SetValue(m_builder.GetCurrentPath(), AZStd::move(BoundContainer::CreateBoundContainer(instance, attributes)));
```
lumberyard-employee-dm, line 579: The AZStd::move is unnecessary as CreateBoundContainer returns temporary value, so it would bind to an rvalue reference regardless.
```suggestion
                m_containers.SetValue(m_builder.GetCurrentPath(), BoundContainer::CreateBoundContainer(instance, attributes));
```

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
```diff
@@ -455,6 +615,14 @@ namespace AZ::DocumentPropertyEditor
 
             if (visibility == Nodes::PropertyVisibility::Hide || visibility == Nodes::PropertyVisibility::ShowChildrenOnly)
             {
+                //auto containerAttribute = attributes.Find(AZ::Reflection::DescriptorAttributes::Container);
+                //if (containerAttribute && !containerAttribute->IsNull())
+                //{
+                //    m_builder.AddMessageHandler(m_adapter, Nodes::Container::AddNotify.GetName());
+                //    m_builder.AddMessageHandler(m_adapter, Nodes::Container::ClearNotify.GetName());
+                //    m_builder.AddMessageHandler(m_adapter, Nodes::Container::RemoveNotify.GetName());
+                //}
```
lumberyard-employee-dm, line 624: This commented out code needs to be removed.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
```diff
@@ -109,34 +204,99 @@ namespace AZ::DocumentPropertyEditor
...
+                {
+                    instanceToRemove = m_elementInstance;
+                    auto associativeContainer = m_parentContainer->GetAssociativeContainerInterface();
+                    if (associativeContainer)
+                    {
+                        instanceToRemove = associativeContainer->GetElementByKey(m_parentContainerInstance, nullptr, m_elementInstance);
+                    }
+                }
+
+                m_parentContainer->RemoveElement(m_parentContainerInstance, instanceToRemove, impl->m_serializeContext);
```
lumberyard-employee-dm, line 252: `instanceToRemove` needs to be check that it is not nullptr before RemoveElement is called.
GetElementByKey can return nullptr

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
amzn-tmryan: Removed the AZStd::move.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
```diff
@@ -109,34 +204,99 @@ namespace AZ::DocumentPropertyEditor
...
+                m_reservedElementInstance = nullptr;
+
+                auto containerNode = GetContainerNode(impl, containerPath);
+                Nodes::PropertyEditor::ChangeNotify.InvokeOnDomNode(containerNode);
+                impl->m_adapter->NotifyResetDocument();
+            };
+
+            void RejectAssociativeContainerKey(ReflectionAdapterReflectionImpl* impl)
+            {
+                AZ_Assert(m_reservedElementInstance != nullptr, "This BoundContainer has no reserved element to free");
```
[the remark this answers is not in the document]
amzn-tmryan: Done.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
```diff
@@ -109,34 +204,99 @@ namespace AZ::DocumentPropertyEditor
...
+                    }
+                }
+                else
+                {
+                    instanceToRemove = m_elementInstance;
+                    auto associativeContainer = m_parentContainer->GetAssociativeContainerInterface();
+                    if (associativeContainer)
+                    {
+                        instanceToRemove = associativeContainer->GetElementByKey(m_parentContainerInstance, nullptr, m_elementInstance);
+                    }
```
lumberyard-employee-dm, line 249: Is `m_elementInstance` always a pointer to the key in this case when using an associative container?

If that is the case we should document the member to indicate when it represents the value_type pair vs when it is just a key_type instance

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
amzn-tmryan: Done.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
```diff
@@ -30,27 +30,128 @@ namespace AZ::DocumentPropertyEditor
...
+                AZ::SerializeContext::IDataContainer* container,
+                void* containerInstance,
+                void* overrideInstance)
+                : m_parentContainer(parentContainer)
+                , m_parentContainerInstance(parentContainerInstance)
+                , m_container(container)
+                , m_containerInstance(containerInstance)
+                , m_elementOverrideInstance(overrideInstance)
+            {
+            }
```
lumberyard-employee-dm, line 64: It seems like these functions should be private, as it seems that you intend that Bound Containers be only created publically through `CreateBoundContainer` correct?

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
amzn-tmryan: Yes, because we've been storing the "element instance override" in `m_elementInstance`, if the `ContainerElementOverride` attribute is set in the `LegacyReflectionBridge`. I'll add a comment explaining that.

We can't do the same thing for nested containers because they need their serialized container instance ptr in many cases, but the element override for removal from their parent. I left a comment for that part of it, but I'll revisit and make sure it's clear as well.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/PropertyEditorNodes.h
```diff
@@ -102,6 +102,13 @@ namespace AZ::DocumentPropertyEditor::Nodes
         FinishedEdit,
     };
 
+    struct Container : NodeWithVisiblityControl
+    {
+        static constexpr auto AddNotify = CallbackAttributeDefinition<void()>("AddNotify");
+        static constexpr auto RemoveNotify = CallbackAttributeDefinition<void()>("RemoveNotify");
+        static constexpr auto ClearNotify = CallbackAttributeDefinition<void()>("ClearNotify");
+    };
+
```
forkercat, line 111: would be good to add some comment here

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
```diff
@@ -30,27 +30,128 @@ namespace AZ::DocumentPropertyEditor
...
+                AZ::SerializeContext::IDataContainer* container,
+                void* containerInstance,
+                void* elementInstance)
+                : m_parentContainer(container)
+                , m_parentContainerInstance(containerInstance)
+                , m_elementInstance(elementInstance)
+            {
+            }
+
+            // For constructing nested containers
```
lumberyard-employee-dm, line 51: Can you defined the difference between what a nested container is vs a non-nested container vs non-container elements?

Such as 
```c++
//! Nested Container = `AZStd::unordered_map<int, AZStd::vector<int>>`
//! Non-Nested Container = `AZStd::unordered_map<int, int>`
//! Non-Container elemments = `typename AZStd::unordered_map<int, int>::value_type` i.e the type is `AZStd::pair<const int, int>`

**Comment by forkercat** [comment]
Does this PR also fix https://github.com/o3de/o3de/issues/15157?

File: Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp (hunk 6 above)
Comment (amzn-tmryan) [reply Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp:64]: Excellent point. I was thinking there might be some case where a BoundContainer could be created otherwise, but better to make these private. This whole thing is starting to read a lot more like a class instead of a struct, semantically.

File: Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp [hunk 9]
```diff
@@ -491,8 +659,10 @@ namespace AZ::DocumentPropertyEditor
                 auto containerAttribute = attributes.Find(Reflection::DescriptorAttributes::Container);
                 if (containerAttribute && !containerAttribute->IsNull())
                 {
+                    m_containers.SetValue(
+                        m_builder.GetCurrentPath(), AZStd::move(BoundContainer::CreateBoundContainer(access.Get(), attributes)));
```
forkercat, line 663: AZStd::move() can be removed here as well

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
```diff
@@ -98,7 +193,7 @@ namespace AZ::DocumentPropertyEditor
 
                 // dataAddress is allocated instance of key type. Get the key type and send it with the dataAddress in the
                 // message, then skip the store element below until we get an AddContainerKey message back from the DPE UI
-                void* dataAddress = m_container->ReserveElement(m_instance, containerClassElement);
+                m_reservedElementInstance = m_container->ReserveElement(m_containerInstance, containerClassElement);
```
forkercat, line 196: looks like the comment is outdated

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/PropertyEditorNodes.h
amzn-tmryan: Will do!

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
amzn-tmryan: Will add something like this with the changes we talked about yesterday, even though the cases won't be separated like they are here. Would be good to document the distinction.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
amzn-tmryan: Got this with the one above.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
amzn-tmryan: Ah, thank you. Good callout.

amzn-tmryan:
Refactored BoundContainer to separate the responsibilities of interfacing with containers and managing container elements, as discussed with @lumberyard-employee-dm.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
```diff
@@ -73,74 +93,203 @@ namespace AZ::DocumentPropertyEditor
...
+
+            AZ::SerializeContext::IDataContainer* m_container = nullptr;
+            void* m_containerInstance = nullptr;
+            void* m_elementInstance = nullptr;
+        };
+
+        struct ContainerEntry
+        {
+            AZStd::unique_ptr<BoundContainer> container;
+            AZStd::unique_ptr<ContainerElement> element;
```
lumberyard-employee-dm, line 288: Could you use m_ for the members here?
```suggestion
            AZStd::unique_ptr<BoundContainer> m_container;
            AZStd::unique_ptr<ContainerElement> m_element;
```

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
```diff
@@ -927,36 +1097,28 @@ namespace AZ::DocumentPropertyEditor
...
-                    containerEntry->OnAddElement(m_impl.get(), message.m_messageOrigin);
+                    containerEntry->container->OnAddElement(m_impl.get(), message.m_messageOrigin);
                     break;
                 case ContainerAction::RemoveElement:
-                    containerEntry->OnRemoveElement(m_impl.get(), message.m_messageOrigin);
+                    containerEntry->element->OnRemoveElement(m_impl.get(), message.m_messageOrigin);
                     break;
                 case ContainerAction::Clear:
-                    containerEntry->OnClear(m_impl.get(), message.m_messageOrigin);
+                    containerEntry->container->OnClear(m_impl.get(), message.m_messageOrigin);
```
lumberyard-employee-dm, line 1106: This should check that either the BoundContainer or ContainerElement pointers aren't nullptr.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
```diff
@@ -73,74 +93,203 @@ namespace AZ::DocumentPropertyEditor
...
                     if (keyTypeAttribute)
                     {
+                        // Get the key type and send it with the dataAddress in the message, then skip the store
+                        // element below until we get an AddContainerKey message back from the DPE UI
                         auto* keyTypeData = azdynamic_cast<const AZ::Edit::AttributeData<AZ::Uuid>*>(keyTypeAttribute);
                         if (keyTypeData)
                         {
-                            auto keyType = keyTypeData->Get(nullptr);
-                            DocumentAdapterPtr reflectionAdapter = AZStd::make_shared<ReflectionAdapter>(dataAddress, keyType);
+                            auto& keyType = keyTypeData->Get(nullptr);
```
lumberyard-employee-dm, line 128: This seems like it should be a `const auto&` or just a `const AZ::TypeId&`
```suggestion
                            const auto& keyType = keyTypeData->Get(nullptr);
```

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
amzn-tmryan: Done. Same for below in the associative container key handlers.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
amzn-tmryan: Done as well. Went with `const AZ::TypeId&`

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/PropertyEditorNodes.h
```diff
@@ -102,6 +102,16 @@ namespace AZ::DocumentPropertyEditor::Nodes
         FinishedEdit,
     };
 
+    //! Container: A node representing a container instance.
+    struct Container : NodeWithVisiblityControl
+    {
+        // These notify callback attributes may be invoked when a container is modified and handled
+        // alongside other messages in the adapter's message handler.
+        static constexpr auto AddNotify = CallbackAttributeDefinition<void()>("AddNotify");
+        static constexpr auto RemoveNotify = CallbackAttributeDefinition<void(size_t /*index*/)>("RemoveNotify");
```
lumberyard-employee-dm, line 111: I don't believe you have to comment out the variable in the function declaration.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
```diff
@@ -73,74 +93,203 @@ namespace AZ::DocumentPropertyEditor
...
-                AZ_Assert(m_elementInstance != nullptr, "Attempted to remove an element without a defined element instance");
-                m_container->RemoveElement(m_instance, m_elementInstance, impl->m_serializeContext);
+                void* instanceToRemove = m_elementInstance;
+
+                auto associativeContainer = m_container->GetAssociativeContainerInterface();
+                if (associativeContainer)
+                {
+                    const AZ::SerializeContext::ClassElement* containerClassElement =
+                        m_container->GetElement(m_container->GetDefaultElementNameCrc());
+                    instanceToRemove = associativeContainer->GetElementByKey(m_containerInstance, containerClassElement, m_elementInstance);
```
lumberyard-employee-dm, line 269: The `m_elementInstance` in this case is actually what needs to be removed and not the key.
This should just be passed directly to `IDataContainer::RemoveElement`

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
```diff
@@ -566,9 +566,9 @@ namespace AZ::DocumentPropertyEditor
                 parentContainerInstanceAttribute && !parentContainerInstanceAttribute->IsNull())
             {
                 auto containerEntry = m_containers.ValueAtPath(m_builder.GetCurrentPath(), AZ::Dom::PrefixTreeMatch::ExactPath);
-                if (containerEntry)
+                if (containerEntry && containerEntry->m_element)
```
amzn-tmryan, line 569: This actually causes a bug. Removing it.

lumberyard-employee-dm: Yeah, the `m_element` should only be set if is nullptr.

If it is not nullptr then it has already been reserved and is in the container.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp
amzn-tmryan: Removed the GetElementByKey block. I didn't have it originally, but thought there might be a case which needed it. You're right; we don't store the key in the LegacyReflectionBridge, only the element instance.

lumberyard-employee-dm (approved):

alexmontAmazon (approved):
looks good :)

amzn-tmryan:
Rebasing to see if I can fix a seemingly unrelated AR test failure.

PR #15589, opened 2023-04-05
Gather by component node

kberg-amzn:
Adds new script-canvas nodes that can gather entities based on the presence of a provided component guid. This is used to fix the shield generator trap in the multiplayer sample.

## What does this PR do?

This allows the shield generator trap to precisely query for players within its explosion proximity. The previous implementation used physx queries combined with complex collision layer setup that often returned duplicated or incorrect results. This new implementation uses the IVisibility system and by filtering using componentIds requires no additional setup of collision layers.

image

### DIFF
--- a/Gems/Multiplayer/Code/Include/Multiplayer/AutoGen/AutoComponent_Source.jinja
+++ b/Gems/Multiplayer/Code/Include/Multiplayer/AutoGen/AutoComponent_Source.jinja
@@ -1596,6 +1596,10 @@ namespace {{ Component.attrib['Namespace'] }}
         AZ::BehaviorContext* behaviorContext = azrtti_cast<AZ::BehaviorContext*>(context);
         if (behaviorContext)
         {
+            behaviorContext->ConstantProperty("{{ ComponentName }}TypeId", BehaviorConstant(s_{{ LowerFirst(ComponentName) }}ConcreteUuid))
+                ->Attribute(AZ::Script::Attributes::Module, "{{ LowerFirst(Component.attrib['Namespace']) }}")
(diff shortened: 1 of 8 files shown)


AMZN-Olex (approved):

lsemp3d (approved):

mbalfour-amzn (approved):
Nice functionality! 

This makes me have a couple of wishlist items, but these are too big to get added to this PR I think:
* Would be nice to expose the creation of the sphere / aabb / etc as separate nodes. Then they could also become inputs to debug draw nodes or something to help visualize when tuning this.
* Would be nice if GatherEntitiesByComponentX could just be one generic node that takes in any shape (aabb, sphere, capsule, etc) as input. I realize that might be kinda painful though due to Overlaps() taking in specific shapes and not having a generic interface.

kberg-amzn:
> Nice functionality!
> 
> This makes me have a couple of wishlist items, but these are too big to get added to this PR I think:
> 
> * Would be nice to expose the creation of the sphere / aabb / etc as separate nodes. Then they could also become inputs to debug draw nodes or something to help visualize when tuning this.
> * Would be nice if GatherEntitiesByComponentX could just be one generic node that takes in any shape (aabb, sphere, capsule, etc) as input. I realize that might be kinda painful though due to Overlaps() taking in specific shapes and not having a generic interface.

Agreed with both of those. We should move this somewhere more central, it's not particularly tied to multiplayer but I'm not sure where the best home is, and we should incorporate your suggestions as well when we move all this to dev.

https://github.com/o3de/o3de/issues/15594

PR #15592, opened 2023-04-06
Remove m_buildTimeStamp from ShaderAsset and ShaderVariantAsset

galibzon:
## What does this PR do?

Remove m_buildTimeStamp from ShaderAsset and ShaderVariantAsset.
Having `m_buildTimeStamp` was necessary a few years ago, but with the recent improvements in the Asset System notifications it is not necessary anymore and removing it also adds stability to Asset Bundling. The main problem of having a build time stamp embedded in the asset is that the same shader built on different dates would appear as having different content, but now, regardless of the date of compilation the Shader assets will have the same binary content, assuming the code has not changed.

## How was this PR tested?

Tested empirically with AutomatedTesting (Editor) by manually changing *.shader assets and making sure the OnAssetReady and OnAssetReloaded events are processed accordingly. Also validated with AtomSampleViewer.


rgba16f (approved):

amzn-mike (approved):

AMZN-stankowi (approved):
This is awesome! Stable product assets go a long way in keeping delta bundles for patches to game projects down in size.

PR #15595, opened 2023-04-06
Change level send to be async and more gracefully handle errors.

mbalfour-amzn:
## What does this PR do?

Changed the Editor->server communication to send the level data packets over a series of frames instead of in a tight loop. For large levels, this will make it clear that the Editor isn't just frozen while sending it. Also added better status messages for successful and failed sends.

## How was this PR tested?

Ran the MPS sample, manually forced failure conditions to verify the fail message worked, slowed down the time cap to verify that the intermediate status updates worked, and ran with the code as-is to verify that success worked.

### DIFF
--- a/Gems/Multiplayer/Code/Source/Debug/MultiplayerConnectionViewportMessageSystemComponent.cpp
+++ b/Gems/Multiplayer/Code/Source/Debug/MultiplayerConnectionViewportMessageSystemComponent.cpp
@@ -382,6 +382,18 @@ namespace Multiplayer
             AZStd::fixed_string<MaxMessageLength>::format(OnEditorSendingLevelDataMessage, bytesSent, bytesTotal);
     }   
 
+    void MultiplayerConnectionViewportMessageSystemComponent::OnEditorSendingLevelDataFailed()
+    {
+        m_centerViewportDebugTextColor = AZ::Colors::Red;
+        m_centerViewportDebugText = OnEditorSendingLevelDataFailedMessage;
+    }
+
+    void MultiplayerConnectionViewportMessageSystemComponent::OnEditorSendingLevelDataSuccess()
+    {
+        m_centerViewportDebugTextColor = AZ::Colors::Yellow;
+        m_centerViewportDebugText = OnEditorSendingLevelDataSuccessMessage;
+    }
+
     void MultiplayerConnectionViewportMessageSystemComponent::OnEditorConnectionAttempt(uint16_t connectionAttempts, uint16_t maxAttempts)
     {
         m_centerViewportDebugTextColor = AZ::Colors::Yellow;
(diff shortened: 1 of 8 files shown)


kberg-amzn (approved):
👏 👏 👏

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
```diff
@@ -383,81 +392,140 @@ namespace Multiplayer
...
+    {
+        // This controls the maximum time slice to use for sending packets. Lower numbers will make the total send time take longer,
+        // but will give the Editor more time to do other work. Larger numbers will make the total send time faster, but will starve
+        // the Editor. The current value attempts to balance between the two.
+        static constexpr AZ::TimeMs MaxSendTimeMs = AZ::TimeMs{ 5 };
+
+        // These control how many retries and how to space them out for packet send failures.
+        static constexpr int MaxRetries = 20;
+        static constexpr int InitialMsDelayPerRetry = 10;
+        static constexpr int MaxMsDelayPerRetry = 1000;
```
lumberyard-employee-dm, line 440: Shouldn't these variables be of type AZ::TimeMs?

Gems/Multiplayer/Code/Source/Debug/MultiplayerConnectionViewportMessageSystemComponent.cpp
```diff
@@ -382,6 +382,18 @@ namespace Multiplayer
             AZStd::fixed_string<MaxMessageLength>::format(OnEditorSendingLevelDataMessage, bytesSent, bytesTotal);
     }   
 
+    void MultiplayerConnectionViewportMessageSystemComponent::OnEditorSendingLevelDataFailed()
```
AMZN-Gene, line 385: Thanks for including visual feedback!

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
```diff
@@ -383,81 +392,140 @@ namespace Multiplayer
...
             {
                 editorServerLevelDataPacket.SetLastUpdate(true);
+                updateFinished = true;
             }
 
             // Try to send the packet to the Editor server. Retry if necessary.
             bool packetSent = false;
-            static constexpr int MaxRetries = 20;
-            int millisecondDelayPerRetry = 10;
+            int millisecondDelayPerRetry = InitialMsDelayPerRetry;
```
AMZN-Gene, line 479: AZ::TimeMs here too?

AMZN-Gene (approved):

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorSystemComponent.cpp
mbalfour-amzn: since they're getting passed into chrono functions that take in ints, it seemed like int was the right type?

lumberyard-employee-dm: Well they get passed to an `AZStd::chrono::milliseconds` constructor which has a representation type for the integer type that is at least [45 bits](https://en.cppreference.com/w/cpp/chrono/duration#Helper_types).

It shouldn't be a problem though since the millisecond constants here are able to easily fit into an 32-bit int

PR #15628, opened 2023-04-07
Fix querying DX12 for msaa custom sample support as well as fix ios shader build related crash

moudgils:
## What does this PR do?

- Minor fix related to passing in incorrect enum to query DX12 drivers for msaa custom sample support.
- Another minor fix related ios shader not building due to passing in multiple name spaces on the same line

## How was this PR tested?

Tested on ASV


cgalvan (approved):
Thank you!

gadams3 (approved):

VickyAtAZ (approved):

PR #15645, opened 2023-04-08
Separating TAA history buffer copy into its own scope.

### LINKED ISSUE #15083: [RPI] Taa pass uses resources as UAV and as copy src/dest in the same pass
**Describe the bug**
The TAA pass has a CopyItem operation along with the execution of a compute shader. This is not allowed in the RHI since we need to add the proper barriers in order to do the copy operation. Since barriers are only added between scopes (passes) the copy operation must be in it's own scope so the RHI can add the proper resource transitions for the copy operation to be successful.

**Assets required**
Found running the Multiplayer sample (https://github.com/aws-lumberyard-dev/o3de-multiplayersample/)

**Steps to reproduce**
1. Run the Multiplayer sample on Vulkan with validation enable
2. See error

`Validation Error: [ VUID-vkCmdCopyImage-srcImageLayout-00128 ] Object 0: handle = 0x21fe48a8f10, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xef9e2bd6 | vkCmdCopyImage: Cannot use VkImage 0xe7e760000000886[Root.MainPipeline_-10.PostProcessPass.TaaPass.Accumulation2] (layer=0 mip=0) with specific layout VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL that doesn't match the previous known layout VK_IMAGE_LAYOUT_GENERAL. The Vulkan spec states: srcImageLayout must specify the layout of the image subresources of srcImage specified in pRegions at the time this command is executed on a VkDevice (https://vulkan.lunarg.com/doc/view/1.3.224.1/windows/1.3-extensions/vkspec.html#VUID-vkCmdCopyImage-srcImageLayout-00128)`

**Expected behavior**
No validation error is emitted.

**Actual behavior**
A validation error is shown.

**Found in Branch**
o3de/Development

 - Device: PC 
 - OS: Windows

invertednormal:
This PR fixes an issue where the TAA history buffer copy is done in the same scope as TAA itself, which isn't technically allowed. While these "works" it causes validation errors.

Fixes #15083 

Tested with TAA on/off in dx12/vulkan.

### DIFF
--- a/Gems/Atom/Feature/Common/Assets/Passes/PassTemplates.azasset
+++ b/Gems/Atom/Feature/Common/Assets/Passes/PassTemplates.azasset
@@ -559,6 +559,10 @@
             {
                 "Name": "SplashScreenPassTemplate",
                 "Path": "Passes/SplashScreen.pass"
+            },
+            {
+                "Name": "TaaParentTemplate",
+                "Path": "Passes/TaaParent.pass"
             }
         ]
     }
--- a/Gems/Atom/Feature/Common/Assets/Passes/PostProcessParent.pass
+++ b/Gems/Atom/Feature/Common/Assets/Passes/PostProcessParent.pass
@@ -86,7 +86,7 @@
                 },
                 {
                     "Name": "TaaPass",
-                    "TemplateName": "TaaTemplate",
+                    "TemplateName": "TaaParentTemplate",
                     "Enabled": false,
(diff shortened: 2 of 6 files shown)


rgba16f (approved):

moudgils (approved):
Nice.

PR #15656, opened 2023-04-10
Detect CPU architecture in LyTestTools

shirangj:
Make LyTestTools be able to isolate tests that are specific to or disabled on Linux ARM64.

Example usage:
Skip a test on Linux arm64 machine:
@pytest.mark.skipif(ly_test_tools.LINUX, reason="Python based file locking does not function on Linux")
@pytest.mark.skipif(ly_test_tools.ARM64, reason="Python based file locking does not function on ARM64")

Skip a test on Linux non-arm64 machine:
@pytest.mark.skipif(ly_test_tools.LINUX, reason="Python based file locking does not function on Linux")
@pytest.mark.skipif(not ly_test_tools.ARM64, reason="Python based file locking does not function on ARM64")

### DIFF
--- a/Tools/LyTestTools/ly_test_tools/__init__.py
+++ b/Tools/LyTestTools/ly_test_tools/__init__.py
@@ -8,6 +8,7 @@ OS and devices are detected and set as constants when ly_test_tools.__init__() c
 """
 import logging
 import sys
+import platform
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


evanchia-ly-sdets (approved):

Tools/LyTestTools/ly_test_tools/__init__.py
```diff
@@ -21,6 +22,9 @@
 MAC = sys.platform.startswith('darwin')
 WINDOWS = sys.platform.startswith('win')
 
+# CPU architecture
+ARM64 = platform.machine.lower() == 'arm64'
```
spham-amzn, line 26:
```suggestion

ARM64 = platform.architecture() == 'aarch64'
```

spham-amzn (approved):

Tools/LyTestTools/ly_test_tools/__init__.py
shirangj: Thanks! Fixed

Issue #8215, opened 2022-03-09
Asset Processor Asset Tab is slow to display hundreds of dependencies


amzn-mike:
**Describe the bug**
Viewing the dependencies (source or product) for an asset with hundreds of dependencies is slow and causes AP to freeze for several seconds

**Steps to reproduce**
1. Go to Assets Tab in Asset Processor with the AutomatedTesting project active
2. Find `standardpbr.azmaterialtype` (AutomatedTesting/Cache/pc/materials/types/standardpbr.azmaterialtype) in the product tab or `cinder_wall_complex.fbx` (AutomatedTesting/Assets/Destruction/cinder_wall_complex.fbx) in the sources tab
4. Select the above entry

**Expected behavior**
Details panel is shown instantly, Asset Processor UI does not freeze at all, dependencies are either populated over time, the entire list is populated instantly, or the visible part of the list is populated instantly

**Actual behavior**
Asset Processor freezes for several seconds

**Assets required**
AutomatedTesting/Cache/pc/materials/types/standardpbr.azmaterialtype
AutomatedTesting/Assets/Destruction/cinder_wall_complex.fbx

**Found in Branch**
dev

PR #15657, opened 2023-04-10
Asset Processor - Fixed performance issue displaying assets with hundreds of dependencies

amzn-mike:
## What does this PR do?

Replaced setCellWidget calls with an item delegate.  setCellWidget is too slow to be used in tables with many rows.

Fixes https://github.com/o3de/o3de/issues/8215

## How was this PR tested?

Followed repro steps in bug using the cinder_wall asset.  Dependencies now display with no noticeable delay.

Before: 2605 samples
After: 11 samples
236x speedup


gadams3 (approved):

AMZN-stankowi (approved):

Code/Tools/AssetProcessor/native/ui/GoToButtonDelegate.cpp
```diff
@@ -0,0 +1,55 @@
...
+        {
+            if (option.state & QStyle::State_MouseOver)
+            {
+
+                m_hoverIcon.paint(painter, option.rect);
+            }
+            else
+            {
+                m_icon.paint(painter, option.rect);
+            }
```
AMZN-alexpete, line 37:
```suggestion
            if (option.state & QStyle::State_MouseOver)
            {
                m_hoverIcon.paint(painter, option.rect);
            }
            else
            {
                m_icon.paint(painter, option.rect);
            }
```
Could you potentially scale the icons down a bit to give them padding so they aren't so large here?  You may also need to size the icon in the constructor with:
```
m_icon= QIcon(":/AssetProcessor_goto.svg").pixmap(width, height);
m_hoverIcon= QIcon(":/AssetProcessor_goto_hover.svg").pixmap(width, height);
```

Issue #15097, opened 2023-03-14
DPE/Prefab Overrides: Several automated component additions/removals with the DPE enabled cause an Editor deadlock/crash


jckand-amzn:
**Describe the bug**
Several automated component additions/removals with the DPE enabled cause an Editor crash at teardown.

**Assets required**
[DPE_AllComponentsAddedRemoved.zip](https://github.com/o3de/o3de/files/10970196/DPE_AllComponentsAddedRemoved.zip)
1. Unzip the file above, and extract the .py file to somewhere in the AutomatedTesting project.

**Steps to reproduce**
1. Launch the Editor with the DPE enabled:
```
Editor.exe --ed_enableDPE=true
```
2. Once the Editor has launched, in the Console type:
```
pyRunFile [Absolute Path to extracted DPE_AllComponentsAddedRemoved.py file]

**Expected behavior**
Test runs and adds/removes the Vegetation Surface Mask Filter component from the test entity.

**Actual behavior**
Test runs and adds/removes the Vegetation Surface Mask Filter component from the test entity, however the Editor then locks up. If executed via a PyTest file, the Editor will crash on teardown with the following stack:

```
E   Call Stack Trace:
E   25) 00007FFAE6A40207 (KERNELBASE) : UnhandledExceptionFilter
E   24) 00007FFAE9115530 (ntdll) : memset
E   23) 00007FFAE90FC876 (ntdll) : _C_specific_handler
E   22) 00007FFAE911241F (ntdll) : _chkstk
E   21) 00007FFAE90C14A4 (ntdll) : RtlRaiseException
E   20) 00007FFAE9110F4E (ntdll) : KiUserExceptionDispatcher
E   19) 00007FFA6F861B87 (Qt5Widgets) : QWidget::~QWidget
E   18) 00007FFA1B713951 (ComponentEntityEditorPlugin) : AzToolsFramework::Prefab::PrefabOverrideLabelHandler::`scalar deleting destructor'
E   17) 00007FFA6EDB9A24 (Qt5Core) : QMetaCallEvent::placeMetaCall
E   16) 00007FFA6EDB83C1 (Qt5Core) : QObject::event
E   15) 00007FFA6F8449A1 (Qt5Widgets) : QApplicationPrivate::notify_helper
E   14) 00007FFA6F843A23 (Qt5Widgets) : QApplication::notify
E   13) 00007FFA6ED917CB (Qt5Core) : QCoreApplication::notifyInternal2
[...]

PR #15254, opened 2023-03-21
Alexmont/dpe crashes

alexmontAmazon:
## What does this PR do?
Fixes crash enabling/disabling the DPE while the Inspector is already open
Fixes crash stemming qobject_cast failures due to casts between DLLs.

## How was this PR tested?
locally using the Editor, enabling and disabling use_dpe flag in the CVar editor

PR #15662, opened 2023-04-10
Fix handler pool issues for multiple-DPE case

forkercat:
## What does this PR do?

Fixes https://github.com/o3de/o3de/issues/15097

This PR fixes issues that would occur if there are multiple DPE views are being used. In the Entity Inspector, there is one DPE view for each component. The pool manager is responsible for creating the pool and holds a weak pointer to it. Each DPE view co-owns the instance pool if it uses its handler.

- **Why it crashes?** - The single-shot delete would cause double deletion after DPE view widget deletes the prefab label handler.
- **Why it does deletion rather than recycling?** - Pool manager's GetPoot() returns nullptr during the deletion of second DPE view, which would lead to delete the handler rather then recycling it. It happens because the prefab label handler pool has already been destroyed in the first DPE view's deletion (upon closing Entity Inspector). There is a bug in handler pool setup code in DPE that the second DPE view does not correctly co-own the shared_ptr of the pool if the pool has already been created by the first DPE view in pool manager.

**Change List:**
1. Replace Qt's single-shot call with direct delete call to prevent deletion from parent widget.
2. Make sure DPE view is registering (co-owning) the handler pool that it needs during setup.
3. Use a hash map to store the owning handler pools. Previously it was a vector.
4. Add module id to the returned handler name (similar to https://github.com/o3de/o3de/pull/15254).

## How was this PR tested?

Tested in editor. No more crash. Tested in standalone app. Didn't see new issues.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
@@ -508,7 +508,6 @@ namespace AzToolsFramework
                 auto handlerInfo = DocumentPropertyEditor::GetInfoFromWidget(childWidget);
                 if (!handlerInfo.IsNull())
                 {
-                    // propertyHandlers own their widgets, so don't destroy them here. Set them free!
                     DocumentPropertyEditor::ReleaseHandler(handlerInfo);
                 }
                 else if (auto rowWidget = qobject_cast<DPERowWidget*>(childWidget))
@@ -1684,7 +1683,6 @@ namespace AzToolsFramework
             // the registration is needed in case the handler pool is released by other DPE views
             RegisterHandlerPool(handlerName, handlerPool);
 
-            // store, then reference the unique_ptr that will manage the handler's lifetime
             auto handler = handlerPool->GetInstance();
             handler->SetValueFromDom(domValue);
             createdWidget = handler->GetWidget();
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
@@ -1707,6 +1707,7 @@ namespace AzToolsFramework
         {
             // if there is no handler pool, then delete the handler immediately; parent widgets won't delete it twice
             delete handler.handlerInterface;
+            handler.handlerInterface = nullptr;
         }
     }
 } // namespace AzToolsFramework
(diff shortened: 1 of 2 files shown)


Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
```diff
@@ -1682,18 +1698,15 @@ namespace AzToolsFramework
...
         else
         {
-            QTimer::singleShot(
-                0,
-                [interfacePointer = handler.handlerInterface]()
-                {
-                    delete interfacePointer;
-                });
+            // if there is no handler pool, then delete the handler immediately; parent widgets won't delete it twice
+            delete handler.handlerInterface;
```
lumberyard-employee-dm, line 1709: Shouldn't the handle.handlerInterface also be set to nullptr after deletion here?

Since the handler can be reused after the call, it would be better to make sure it compares to nullptr if it has been deleted.

```suggestion
            delete handler.handlerInterface;
            handler.handlerInterface = nullptr
```

forkercat: Good callout. Will do!

lumberyard-employee-dm (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
```diff
@@ -1617,9 +1616,15 @@ namespace AzToolsFramework
         message.Match(AZ::DocumentPropertyEditor::Nodes::Adapter::QueryKey, showKeyQueryDialog);
     }
 
-    void DocumentPropertyEditor::RegisterHandlerPool(AZStd::shared_ptr<AZ::InstancePoolBase> handlerPool)
+    void DocumentPropertyEditor::RegisterHandlerPool(AZ::Name handlerName, AZStd::shared_ptr<AZ::InstancePoolBase> handlerPool)
     {
-        m_handlerPools.push_back(handlerPool);
+        if (m_handlerPools.find(handlerName) != m_handlerPools.end())
+        {
+            AZ_Assert(m_handlerPools[handlerName] == handlerPool, "Registering a handler name for a different handler pool.");
```
amzn-tmryan, line 1623: This `"Registering a handler name for a different handler pool."` might be more readable as something like `"Attempted to register a handler pool to a handler name that is already in use."`.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
```diff
@@ -1617,9 +1616,15 @@ namespace AzToolsFramework
...
+        }
+
+        // will not overwrite if there is a registered handler pool for the handler name
```
amzn-tmryan, line 1626: Nit: this could mention that it won't overwrite because of the unordered_map, just to make it more clear.

michabr (approved):

alexmontAmazon (approved):
looks good, thanks!

PR #15665, opened 2023-04-11
Shadow Enhancement for FullscreenBlur

Zeang:
## What does this PR do?
(https://github.com/o3de/sig-graphics-audio/issues/130)

## How was this PR tested?
Test it in windows and have a current effect


Gems/Atom/Feature/Common/Code/Source/CoreLights/DirectionalLightFeatureProcessor.cpp
```diff
@@ -708,8 +710,25 @@ namespace AZ
 
         void DirectionalLightFeatureProcessor::CacheFullscreenPass()
         {
+#if USING_SCREENSHADOW_COMPUTEPASS
```
invertednormal, line 713: Would it be possible to make this a runtime switch? would be nice to be able to compare quality and performance.

Gems/Atom/Feature/Common/Code/Source/Shadows/FullscreenShadowComputePass.h
```diff
@@ -0,0 +1,77 @@
...
+
+            AZ::RHI::Size GetDepthBufferDimensions();
+
+            int GetDepthBufferMSAACount();
+
+            void SetConstantData();
+
+            // SRG binding indices...
+            AZ::RHI::ShaderInputNameIndex m_constantDataIndex = "m_constantData";
+            FullScreenShadowConstantData constantData;
```
invertednormal, line 65: m_constantData

Gems/Atom/Feature/Common/Code/Source/Shadows/ProjectedShadowFeatureProcessor.h
invertednormal, line 1: Is this file supposed to be in this PR?

invertednormal (changes requested):
Looks pretty good - one overall issue is it looks like this removes the code to deal with msaa sample positions. It'll work without msaa, but with msaa it'll miscalculate the world space coordinate because it assumes pixel center, however first msaa depth sample will be different depending on the msaa level. This leads to shadows that are right _most_ of the time, but then have false positives or false negatives, particularly at grazing angles.

invertednormal:
Would also love more information in the description - any performance differences? Eventually this should be able to be improved to use shared memory for additional performance.

Gems/Atom/Feature/Common/Code/Source/CoreLights/DirectionalLightFeatureProcessor.cpp
```diff
@@ -1701,20 +1721,40 @@ namespace AZ
         void DirectionalLightFeatureProcessor::SetFullscreenPassSettings()
         {
             ShadowProperty& shadowProperty = m_shadowProperties.GetData(m_shadowingLightHandle.GetIndex());
+            bool fullscreenBlurEnabled = shadowProperty.m_fullscreenBlurEnabled;
 
+#if USING_SCREENSHADOW_COMPUTEPASS
+            if (m_fullscreenShadowComputePass)
+            {
+                m_fullscreenShadowComputePass->SetEnabled(fullscreenBlurEnabled);
```
moudgils, line 1729: Do we not want full screen shadow pass without blur? Trying to understand what would happen if we disable the blur. Will shadows stop working?

Gems/Atom/Feature/Common/Assets/ShaderLib/Atom/Features/Shadow/DirectionalLightShadow.azsli
```diff
@@ -43,13 +44,13 @@ class DirectionalLightShadow
...
+bool DirectionalLightShadow::UseFullscreenShadows(uint lightIndex)
 {
 #if FORCE_OPAQUE
-    return true;
+    return o_full_screen_blur_enable;
 #else
     bool useFullscreenShadows = o_opacity_mode == OpacityMode::Opaque || o_opacity_mode == OpacityMode::Cutout;
-    return useFullscreenShadows;
+    return useFullscreenShadows && o_full_screen_blur_enable;
 #endif
```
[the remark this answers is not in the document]
Zeang: Yes,  I don't need a full-screen pass if I want to use the shadow method directly without full-screen blur. I want to save the cost of that pass.

Gems/Atom/Feature/Common/Assets/ShaderLib/Atom/Features/Shadow/FullscreenShadowCommon.azsli
```diff
@@ -0,0 +1,103 @@
...
+    
+    Sampler LinearSampler
+    {
+        MinFilter = Linear;
+        MagFilter = Linear;
+        MipFilter = Linear;
+        AddressU = Clamp;
+        AddressV = Clamp;
+        AddressW = Clamp;
+    };
```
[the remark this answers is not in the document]
Zeang: You can find it in DirectionalLightShadowCalculator.azsli

Gems/Atom/Feature/Common/Code/Source/Shadows/FullscreenShadowComputePass.h
Zeang: Thank you, It will be changed.

Gems/Atom/Feature/Common/Code/Source/Shadows/ProjectedShadowFeatureProcessor.h
Zeang: Oh, I tried to save the esm pass's cost If we don't need the esm shadow filter method. I will add it in RFC's description.

Gems/Atom/Feature/Common/Code/Source/CoreLights/DirectionalLightFeatureProcessor.cpp
Zeang: Because if we don't need the full-screen shadow to blur, the full-screen shadow pass has some performance cost. You can find it in DirectionalLightShadow.azsli. If we don't use full-screen blur, then directly calculating shadow will be used.

Gems/Atom/Feature/Common/Assets/ShaderLib/Atom/Features/Shadow/DirectionalLightShadow.azsli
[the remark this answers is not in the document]
invertednormal: I was under the impression that full screen shadows were originally done to reduce register pressure on the forward pass, the blur was just an extra feature that was enabled by the technique. Do you have any performance comparisons?

Zeang:
感谢来信，我会尽快处理~——陈泽昂

PR #15667, opened 2023-04-11
Enable Linux ARM64 Support

spham-amzn:
## What does this PR do?

This PR adds support for Linux ARM66 (AARCH64). Specifically, support will be based on ARMv8 with support for NEON instructions.

* Update cmake detection of Linux to also include detection of the system processor (through $CMAKE_SYSTEM_PROCESSOR
* Update python related files to select the prebuilt arm64 version of python for download on arm64 systems.arm64 systems
* Update Linux traits to identify and support NEON instrinsics when ARM64 is detected
* Adjust unit test tolerances that involve float comparisons when running on ARM64 accordingly (Neon has a lower prevision value than SSE)
* Fix network serialization of char to consistently use unsigned char across all platforms/architecture. This is due to the fact that on ARM64, the default for char is an unsigned char, but on x86, the default is a signed char The network serialization for char takes into the numeric limits for the type when serializing, so writing and reading a serialized char type across ARM64 and x86 will not be valid.
* Fix legacy code that assumed NEON intrinsics only on Android and iOS platforms. Also protect neon detection to use both __ARM_NEON and legacy \_\_ARM_NEON\_\_ defines.
* Update 3rd Party packaging to use system processor appropriate packages for Linux
* Remove reliance on zstd 3rd Party packaging in favor of using the system provided one.
* Fix a problem with the xxhash package was identified as a header only package, but in fact it relied on implementation as well, which was being provided by the zstd 3p package. Since we switch to using the system zstd package instead, the internal implementation for xxhash is no longer there. The fix is to force the usage of xxhash to inline all the functions instead. To protect against any impacting other areas, the usage is namespace protected inside AssetBuilderSDK namespace.
* Fix an issue when a new prefab instance is created, even though it is being called from the same prefab, generates a new unique alias, therefore  generating a new hash, causing it to always generate a new unique ID even though they are the same prefabs. The cause is that when creating a new   `PrefabDocument`, the instance is created with a unique alias, even though no alias is needed. This causes the instances of the same prefab to end up with unique IDs. The fix is to supply a 'dummyalias' instead to invoke the constructor for a prefab instance that does not generate a new instance alias for the prefab.  
* Fix internal implementation of `htonll` on platforms that don't support it (Linux, Android) to match the output of `htonll` on platforms that do support it (Windows) 
* Fix a network serialization issues related to signed integer overflow and order of operations. 
* Added unit test for type serialization for fixed expected serialization to ensure that serialization is consistent across all supported platforms/architectures

[RFC](https://github.com/o3de/sig-platform/blob/main/rfcs/linux-arm64-support.md)

## How was this PR tested?
* Existing Unit Tests for Linux x86_64 ran on Ubunutu Linux aarch64
* [Test Plan](https://github.com/o3de/sig-platform/blob/main/testing-guidance/Test-Plan-ARM64-Linux.md)


Code/Tools/LuaIDE/Source/LUA/WatchesPanel.cpp
```diff
@@ -831,7 +831,7 @@ bool WatchesDataModel::setData (const QModelIndex& index, const QVariant& value,
 
 const char* WatchesDataModel::SafetyType(char c) const
 {
-    if (c <= LUA_TNONE || c > LUA_NUMTAGS)
+    if (static_cast<int>(c) <= LUA_TNONE || c > LUA_NUMTAGS)
```
lumberyard-employee-dm, line 834: Doesn't the greater than comparison also need to cast the character to an int.
```suggestion
    if (static_cast<int>(c) <= LUA_TNONE || static_cast<int>(c) > LUA_NUMTAGS)
```

Gems/NvCloth/Code/Platform/Linux/PAL_linux.cmake
```diff
@@ -6,6 +6,12 @@
 #
 #
 
-ly_associate_package(PACKAGE_NAME NvCloth-v1.1.6-4-gd243404-pr58-rev1-linux TARGETS NvCloth PACKAGE_HASH c2e469c909fb358105ffe854476e87d9c69626aa16bd0ca0bcd941522adadc6d)
+if(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "x86_64")
+    ly_associate_package(PACKAGE_NAME NvCloth-v1.1.6-4-gd243404-pr58-rev1-linux         TARGETS NvCloth PACKAGE_HASH c2e469c909fb358105ffe854476e87d9c69626aa16bd0ca0bcd941522adadc6d)
+elseif(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "aarch64")
+    ly_associate_package(PACKAGE_NAME NvCloth-v1.1.6-4-gd243404-pr58-rev1-linux-aarch64 TARGETS NvCloth PACKAGE_HASH 5cf6afce8a65f9584b25b99fc04843cd9e5e4720d2a1736321c7768d57409161)
+else()
+    message(FATAL_ERROR "Unsupport linux architecture ${CMAKE_SYSTEM_PROCESSOR}")
```
lumberyard-employee-dm, line 14: typo: missing "ed" in "unsupport"
```suggestion
    message(FATAL_ERROR "Unsupported linux architecture ${CMAKE_SYSTEM_PROCESSOR}")
```

Gems/PhysX/Code/Source/Platform/Linux/PAL_linux.cmake
```diff
@@ -9,6 +9,13 @@
...
-     ly_associate_package(PACKAGE_NAME v-hacd-2.3-1a49edf-rev1-linux TARGETS v-hacd PACKAGE_HASH 777bed3c7805a63446ddfea51c505b35398744f9245fa7ddc58e6a25d034e682)
+    if(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "x86_64")
+        ly_associate_package(PACKAGE_NAME poly2tri-7f0487a-rev1-linux           TARGETS poly2tri PACKAGE_HASH b16eef8f0bc469de0e3056d28d7484cf42659667e39b68b239f0d3a4cbb533d0) 
+        ly_associate_package(PACKAGE_NAME v-hacd-2.3-1a49edf-rev1-linux         TARGETS v-hacd   PACKAGE_HASH 777bed3c7805a63446ddfea51c505b35398744f9245fa7ddc58e6a25d034e682)
+    elseif(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "aarch64")
+        ly_associate_package(PACKAGE_NAME poly2tri-7f0487a-rev1-linux-aarch64   TARGETS poly2tri PACKAGE_HASH a7dfa76df0b77a65c00e52bda28c3c174fe84ca115f5a015b6e937ee5b55e969)
+        ly_associate_package(PACKAGE_NAME v-hacd-2.3-1a49edf-rev1-linux-aarch64 TARGETS v-hacd   PACKAGE_HASH e5d45327099588f00f4bd0af3907f29279e23831a84c26661bb85248fb1d69c1)        
+    else()
+        message(FATAL_ERROR "Unsupport linux architecture ${CMAKE_SYSTEM_PROCESSOR}")
```
lumberyard-employee-dm, line 19: repeat: "ed" is missing from "Unsupport"
```suggestion
        message(FATAL_ERROR "Unsupported linux architecture ${CMAKE_SYSTEM_PROCESSOR}")
```

Gems/ScriptCanvas/Code/Tests/Framework/ScriptCanvasUnitTestFixture.h
```diff
@@ -15,4 +15,14 @@ namespace ScriptCanvasUnitTest
         : public UnitTest::LeakDetectionFixture
     {
     };
+
+    MATCHER_P(IsClose, expected, "")
+    {
+        #if AZ_TRAIT_USE_PLATFORM_SIMD_NEON
+        constexpr float tolerance = 0.001f;
+        #else
+        constexpr float tolerance = 0.0f;
```
lumberyard-employee-dm, line 24: So when not using cmd, the result is expected to be "exactly" equal?

That is very strange for a floating point IsClose function.

Gems/WhiteBox/Code/Source/Platform/Linux/platform_linux_tools.cmake
```diff
@@ -8,7 +8,13 @@
 
 if(PAL_TRAIT_BUILD_HOST_TOOLS)
 
-    ly_associate_package(PACKAGE_NAME OpenMesh-8.1-rev3-linux TARGETS OpenMesh PACKAGE_HASH 805bd0b24911bb00c7f575b8c3f10d7ea16548a5014c40811894a9445f17a126)
+    if(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "x86_64")
+        ly_associate_package(PACKAGE_NAME OpenMesh-8.1-rev3-linux         TARGETS OpenMesh PACKAGE_HASH 805bd0b24911bb00c7f575b8c3f10d7ea16548a5014c40811894a9445f17a126)
+    elseif(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "aarch64")
+        ly_associate_package(PACKAGE_NAME OpenMesh-8.1-rev3-linux-aarch64 TARGETS OpenMesh PACKAGE_HASH 0d53d215c4b2185879e3b27d1a4bdf61a53bcdb059eae30377ea4573bcd9ebc1)
+    else()
+        message(FATAL_ERROR "Unsupport linux architecture ${CMAKE_SYSTEM_PROCESSOR}")
```
lumberyard-employee-dm, line 16: typo: missing "ed" in the word "Unsupport"
```suggestion
        message(FATAL_ERROR "Unsupported linux architecture ${CMAKE_SYSTEM_PROCESSOR}")
```

cmake/3rdParty/Platform/Linux/BuiltInPackages_linux.cmake
```diff
@@ -12,54 +12,109 @@
...
+    if ("${OPENSSL_VERSION}" STREQUAL "")
+        message(FATAL_ERROR "OpenSSL not detected. The OpenSSL dev package is required for O3DE")
+    elseif ("${OPENSSL_VERSION}" VERSION_LESS "3.0.0")
+        ly_associate_package(PACKAGE_NAME AWSNativeSDK-1.9.50-rev2-linux-openssl-1-aarch64       TARGETS AWSNativeSDK                PACKAGE_HASH db83fb53946602495e586dfb1ef2dd52018d38d36e0f12c8baea3d1d87518e77)
+    else()
+        ly_associate_package(PACKAGE_NAME AWSNativeSDK-1.9.50-rev2-linux-openssl-3-aarch64       TARGETS AWSNativeSDK                PACKAGE_HASH 963d97090bf1a9347a4c8d512c393ff34961d3faba4eaa3a2d00e26b54ff36d4)
+    endif()
 else()
-	ly_associate_package(PACKAGE_NAME AWSNativeSDK-1.9.50-rev2-linux-openssl-3          TARGETS AWSNativeSDK                PACKAGE_HASH 9b9b5124791fb2f59b7362a95ed997944aff6cc928b9dede916e8968a09f23ff)
+    message(FATAL_ERROR "Unsupport linux architecture ${CMAKE_SYSTEM_PROCESSOR}")
```
lumberyard-employee-dm, line 117: typo: missing "ed " in "Unsupport"
```suggestion
    message(FATAL_ERROR "Unsupported linux architecture ${CMAKE_SYSTEM_PROCESSOR}")
```

Gems/ScriptCanvas/Code/Tests/Framework/ScriptCanvasUnitTestFixture.h
spham-amzn: This doesnt work, so I had to change to either a straight _EQ (for non-neon) or an isClose for NEON

cmake/3rdParty/Platform/Linux/BuiltInPackages_linux.cmake
```diff
@@ -12,54 +12,109 @@
...
+    endif()
+
+    # Certain packages are built against OpenSSL, so we must associate the proper ones based on their OpenSSL (Major) versions
+    if ("${OPENSSL_VERSION}" STREQUAL "")
+        message(FATAL_ERROR "OpenSSL not detected. The OpenSSL dev package is required for O3DE")
+    elseif ("${OPENSSL_VERSION}" VERSION_LESS "3.0.0")
+        ly_associate_package(PACKAGE_NAME AWSNativeSDK-1.9.50-rev2-linux-openssl-1-aarch64       TARGETS AWSNativeSDK                PACKAGE_HASH db83fb53946602495e586dfb1ef2dd52018d38d36e0f12c8baea3d1d87518e77)
+    else()
+        ly_associate_package(PACKAGE_NAME AWSNativeSDK-1.9.50-rev2-linux-openssl-3-aarch64       TARGETS AWSNativeSDK                PACKAGE_HASH 963d97090bf1a9347a4c8d512c393ff34961d3faba4eaa3a2d00e26b54ff36d4)
+    endif()
```
lumberyard-employee-dm, line 115: If Ubuntu 20.04 LTS is not supported on Arch, do we need to support with the OpenSSL 1.1.1?

Code/Tools/LuaIDE/Source/LUA/WatchesPanel.cpp
spham-amzn: `LUA_TNONE` was defined as `-1`, so the compiler complained because apparently `char` comparisons to a negative is a warning -> error

cmake/Platform/Linux/Configurations_linux.cmake
```diff
@@ -6,61 +6,118 @@
...
+                -lpthread
+        )
+        ly_set(CMAKE_CXX_EXTENSIONS OFF)
+
+    else()
+        message(FATAL_ERROR "Compiler ${CMAKE_CXX_COMPILER_ID} not supported in ${PAL_PLATFORM_NAME}")
+    endif()
+else()
+    message(FATAL_ERROR "Unsupport linux architecture ${CMAKE_SYSTEM_PROCESSOR}")
 endif()
```
lumberyard-employee-dm, line 118: Could we split the configuration settings up into a `Configurations_linux_x86_64.cmake` and a `Configurations_linux_aarch64.cmake` and then include each of their architecture specific files based on the SYSTEM_PROCESSOR?

cmake/3rdParty/Platform/Linux/BuiltInPackages_linux.cmake
spham-amzn: Ubuntu 20.04 LTS does support arm64

lumberyard-employee-dm: But will O3DE Arm64 need to support that version?

Code/Tools/LuaIDE/Source/LUA/WatchesPanel.cpp
lumberyard-employee-dm: Ok, so LUA_NUMTAGS is positive `9` and therefore that comparison doesn't need a cast then.

Thanks for the clarification.

cmake/3rdParty/Platform/Linux/BuiltInPackages_linux.cmake
spham-amzn: We didn't specify any specific minimum version of ubuntu in the ARM64 RFC. Officially, we still support ubuntu 20.04 LTS.

amzn-changml (approved):
ARM node config added to the AR fleet

smurly (approved):

lumberyard-employee-dm (approved):

lumberyard-employee-dm (approved):

Code/Framework/AzNetworking/AzNetworking/Serialization/NetworkInputSerializer.cpp
```diff
@@ -128,20 +123,20 @@ namespace AzNetworking
     {
         m_serializerValid &= (inputValue >= minValue);
         m_serializerValid &= (inputValue <= maxValue);
-        const uint64_t valueRange = static_cast<uint64_t>(maxValue - minValue);
+        const uint64_t valueRange = static_cast<uint64_t>(maxValue) - static_cast<uint64_t>(minValue);
         if (valueRange <= AZStd::numeric_limits<uint8_t>::max())
         {
-            return SerializeBoundedValueHelper<uint8_t>(static_cast<uint8_t>(inputValue - minValue));
+            return SerializeBoundedValueHelper<uint8_t>(static_cast<uint8_t>(inputValue) - static_cast<uint8_t>(minValue));
```
kberg-amzn, line 129: I don't think you wanted to cast this part like this..? In order to minimize risk of underflow/overflow, it's better to do all the calculations in the largest register space available and only cast the final result after subtraction.

kberg-amzn: As a case in point, consider a uint16_t type where minValue was 500 and maxValue was 510 or something.. there's only 10 integers in that range, so it can easily be transmitted in a byte or less even. 510 - 500 == 10, easily castable to a uint8_t without any risk of loss of value, however casting 500 and 510 individually to uint8_t's absolutely will truncate and lose data.

Code/Framework/AzToolsFramework/Tests/EditorTransformComponentSelectionTests.cpp
```diff
@@ -3401,7 +3409,13 @@ namespace UnitTest
         const auto distanceAway = (finalEntityTransform.GetTranslation() - viewportRay.m_origin).GetLength();
 
         // ensure final world positions match
-        EXPECT_THAT(finalEntityTransform, IsCloseTolerance(finalTransformWorld, 0.01f));
+        #if AZ_TRAIT_USE_PLATFORM_SIMD_NEON
+            constexpr float finalTransformWorldTolerance = 0.028f;
```
kberg-amzn, line 3413: ouch.. that's quite a jump in loss of precision..

kberg-amzn (commented):
Looks really good! My only concern is the casting done inside that one SerializeBoundedValue method, but otherwise this is a fantastic set of fixes! 👍

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Spawnable/PrefabDocument.cpp
```diff
@@ -14,7 +14,7 @@ namespace AzToolsFramework::Prefab::PrefabConversionUtils
 {
     PrefabDocument::PrefabDocument(AZStd::string name)
         : m_name(AZStd::move(name))
-        , m_instance(AZStd::make_unique<AzToolsFramework::Prefab::Instance>(AzToolsFramework::Prefab::EntityIdInstanceRelationship::OneToMany))
+        , m_instance(AZStd::make_unique<AzToolsFramework::Prefab::Instance>("DummyAlias", AzToolsFramework::Prefab::EntityIdInstanceRelationship::OneToMany))
```
kberg-amzn, line 17: Is this worth a comment considering it ended up costing you so much time to debug? 🤣

AMZN-alexpete (approved):
Great addition. Could you also bump the engine minor version in `engine.json` so devs have an engine version to target for minimum Linux Arm64 support?

scripts/build/Platform/Linux/build_config.json
```diff
@@ -43,7 +43,7 @@
     "steps": [
       "profile_nounity",
       "asset_profile_nounity",
-      "test_profile_nounity"
+      "test_profile_nounity_arm64"
```
lumberyard-employee-dm, line 46: Can we just use unity in this case?

The normal Ubuntu x86_64 build already covers the non-unity build

spham-amzn: In general, we use the 'nounity' builds for daily/weekly builds. I think switching to 'unity' will introduce additional metrics that may not be intended for this
```
  "profile": {
    "TAGS": [
      "periodic-incremental-daily",
      "periodic-clean-weekly-internal",
      "daily-pipeline-metrics",
      "weekly-build-metrics"
    ],
```
vs
```
  "profile_nounity": {
    "TAGS": [
      "daily-pipeline-metrics",
      "weekly-build-metrics"
    ],

```

lumberyard-employee-dm: Why can't we ad another tag.

spham-amzn: we could if its important

lumberyard-employee-dm: My suggestion was purely for saving some build time and build resources.

PR #15669, opened 2023-04-11
Save DPE inspector settings before loading

michabr:
## What does this PR do?

The DPE settings are loaded and saved during construction/destruction. This PR makes sure that settings are saved to disk before they're loaded from disk again.

## How was this PR tested?

Testing in Editor.


forkercat (approved):

gadams3 (approved):
Good catch.

PR #15676, opened 2023-04-11
Action Manager | Fix assigning widgets to action contexts at startup

AMZN-daimini:
## What does this PR do?

Fixes a regression due to changes in the order of operations during editor initialization - widgets were initialized before the action manager registration hooks were called, making it so that the assignment would fail when the editor layout is restored at startup.

Introduced a new helper function that correctly delays registration if the system has not been initialized when the widget is created.

## How was this PR tested?

Manual testing.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/ActionManager/HotKey/HotKeyWidgetRegistrationHelper.h
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/ActionManager/HotKey/HotKeyWidgetRegistrationHelper.h
@@ -38,11 +38,7 @@ namespace AzToolsFramework
 
         bool m_isRegistrationCompleted = false;
 
-        struct HotKeyActionContextPair
-        {
-            AZStd::string actionContextIdentifier;
-            QWidget* widget;
-        };
+        struct HotKeyActionContextPair;
         AZStd::vector<HotKeyActionContextPair> m_widgetContextQueue;
 
         HotKeyManagerInterface* m_hotKeyManagerInterface = nullptr;
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/ActionManager/HotKey/HotKeyWidgetRegistrationHelper.h
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/ActionManager/HotKey/HotKeyWidgetRegistrationHelper.h
@@ -1,0 +1,51 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#pragma once
+
+#include <AzCore/std/containers/vector.h>
+#include <AzCore/std/string/string.h>
+
+#include <AzToolsFramework/ActionManager/ActionManagerRegistrationNotificationBus.h>
+#include <AzToolsFramework/ActionManager/HotKey/HotKeyWidgetRegistrationInterface.h>
+
+class QWidget;
+
+namespace AzToolsFramework
+{
+    class HotKeyManagerInterface;
+    class HotKeyWidgetRegistrationInterface;
+
+    class HotKeyWidgetRegistrationHelper final
+        : private ActionManagerRegistrationNotificationBus::Handler
+        , public HotKeyWidgetRegistrationInterface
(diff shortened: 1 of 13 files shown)


gadams3 (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/ActionManager/HotKey/HotKeyWidgetRegistrationInterface.h
```diff
@@ -0,0 +1,34 @@
...
+    {
+    public:
+        AZ_RTTI(HotKeyWidgetRegistrationInterface, "{B9197853-9655-4B72-885C-F02B9B63164F}");
+
+        //! Tries assigning a widget to an Action Context.
+        //! If the system has not yet been initialized, the system will store the arguments and assign the widgets
+        //! after the registration hooks have been called.
+        //! @param contextIdentifier The identifier of the action context to assign the widget to.
+        //! @param widget The widget to assign to the action context.
+        virtual void AssignWidgetToActionContext(const AZStd::string& actionContextIdentifier, QWidget* widget) = 0;
```
lumberyard-employee-dm, line 31: The lack of efficient string parameter usage here is saddening.

These parameters are better of being an `AZStd::string` parameter, where it gets moved into the `HotKeyActionContextPair` at the end or at the very least an `AZStd::string_view` which avoids an unnecessary allocation when passing in a character literal.
The string_view parameter at least supports passing in an the value from an `AZStd::fixed_string` without any need for conversion.

Here is a link to the O3DE C++ Best Practices section on using accepting string parameters in functions: https://github.com/o3de/sig-core/blob/main/governance/C%2B%2B-Best-Practices-Guide.md#string-view-vs-const-string

Code/Framework/AzToolsFramework/AzToolsFramework/ActionManager/HotKey/HotKeyWidgetRegistrationHelper.h
```diff
@@ -0,0 +1,51 @@
...
+        // ActionManagerRegistrationNotificationBus overrides ...
+        void OnPostActionManagerRegistrationHook() override;
+
+        bool m_isRegistrationCompleted = false;
+
+        struct HotKeyActionContextPair
+        {
+            AZStd::string actionContextIdentifier;
+            QWidget* widget;
+        };
```
lumberyard-employee-dm, line 45: This struct only needs to be forward declared in the `HotKeyWidgetRegistrationHelper` class.

Its class definition can be placed in the cpp file, since an AZstd::vector class doesn't need a complete type when instantiating the template as long as the destructor isn't instantiated.

Moving the struct the cpp file would help with reducing compile time if for some reason the struct needed to be extended, as only the cpp file would need to be recompiled in that case.

The cpp file can define the struct as
```
struct HotKeyWidgetRegistrationHelper::HotKeyActionContextPair
{
    AZStd::string actionContextIdentifier;
    QWidget* widget;
};
```

AMZN-daimini: Thank you, done!

Code/Framework/AzToolsFramework/AzToolsFramework/ActionManager/HotKey/HotKeyWidgetRegistrationInterface.h
lumberyard-employee-dm: Using an `AZStd::string` as a key within an `unordered_map` is perfectly fine and interoperates well with an `AZStd::string_view`.
Both the `AZStd::string` [hash function](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/std/string/string.h#L2138-L2146) and the general [AZStd::equal_to](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/std/functional_basic.h#L137-L146) function that the hash_table class uses are "transparent".

A [transparent class](https://en.cppreference.com/w/cpp/utility/functional/equal_to_void#Notes) allows those operators to be used without converting the type to an `AZStd::string` avoiding any copies.

That being said, you don't need to address this here, however keep in mind going forward that a `const string&` is the worst case function parameter type for accepting a string.
That type can be replaced with either `string`, `string_view` or `const char*` in way that provides benefits in each case without the drawbacks.

Issue #15678, opened 2023-04-11
MeshFeatureProcessor crashes with mesh component + invalid material


amzn-tommy:
**Describe the bug**
The mesh feature processor no longer gracefully handles an invalid/missing material. Instead it crashes. This is a regression introduced by https://github.com/o3de/o3de/pull/14963

**Assets required**

**Steps to reproduce**
1. Delete Cache\pc\objects\shaderball\shaderball_default_1m_.azmaterial
2. Create a level with the default Atom environment

**Expected behavior**
The level loads with a warning: "No material provided for mesh. Skipping." The shaderball should not render, but it should not crash

**Actual behavior**
Crash ~2100 in MeshFeatureProcessor.cpp

size_t meshCount = lodAssets[lodIndex + m_lodBias]->GetMeshes().size();
for (size_t meshIndex = 0; meshIndex < meshCount; ++meshIndex)
{
   if (!r_meshInstancingEnabled)
   {
        const RHI::DrawPacket* rhiDrawPacket =
          m_drawPacketListsByLod[lodIndex + m_lodBias][meshIndex].GetRHIDrawPacket();

**Screenshots/Video**

**Found in Branch**
development

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
6184e81

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Windows
 - Version: 10
 - CPU: Intel I9-10900X
 - GPU: NVidia GeForce RTX 2080 SUPER
 - Memory: 64GB

**Additional context**

PR #15680, opened 2023-04-11
Fix for GHI-15678 MeshFeatureProcessor crashes with mesh component + invalid material

amzn-tommy:
## What does this PR do?

Fixes https://github.com/o3de/o3de/issues/15678

Fix invalid access via meshIndex when not all meshes have valid draw packets.

## How was this PR tested?
Tested in the Editor using the repro steps from the GHI.


lawsonamzn (approved):
I take it this doesn't need to go into stabilization, is the instancing code in dev only?

amzn-tommy:
> I take it this doesn't need to go into stabilization, is the instancing code in dev only?

Correct, the instancing changes are in dev only.

gadams3 (approved):

rgba16f (approved):

cgalvan (approved):

PR #15685, opened 2023-04-12
Allow usage of RT descriptors in const context

martinwinter-huawei:
## What does this PR do?

This PR only changes the signature of three methods to accept a `const` reference parameter instead of a reference, all three only make copies of that parameter internally and this change allows the usage of these descriptors in contexts where a `const` resource is present.

## How was this PR tested?

Built against the newest `o3de-atom-sampleviewer` state and does not change anything else, so should not affect run-time behaviour.


jhmueller-huawei (approved):

rgba16f (approved):

moudgils (approved):

Jackie9527 (approved):
LGTM

Issue #15626, opened 2023-04-07
Multiplayer SessionNotificationBus vs IMultiplayer Event is Confusing


AMZN-Gene:
As a dev I'm not sure when to use _SessionNotificationBus_ vs _IMultiplayer_ events.
_SessionNotificationBus_ has _OnCreateSessionBegin_ and _OnDestroySessionBegin_, while _IMultiplayer_ has _SessionInit_ and _SessionShutdown_

In reality, _IMultiplayer::SessionInit_ event isn't always called with _MultiplayerInitialized_ is called, but instead is called within _MultiplayerInitialized_ and only for server types (clients receive no callback after initializing client-mode). _SessionNotificationBus_  is currently only called from the GameLift Gem.

**Proposed Fixes**
1. Partial confusion could be cleared by renaming _SessionNotificationBus_  events which returns a result to _SessionRequestBus_. These aren't simply notification to be listened for, but rather GameLift is relying (requesting) that the multiplayer system perform an action successfully before GameLift proceed. Failure to handle the request, results in a game bad state. For example, SessionNotifications::OnCreateSessionBegin isn't called when the session begins, it's actually requesting that the MultiplayerSystem being hosting and listening on a port; if the mutliplayer system returns false, then that means the game session is unitialized.

2. Confusion might also be cleared up by removing unused events. Both _IMultiplayer::SessionInitEvent_ and _SessionShutdownEvent_ are currently only used for testing.

PR #15688, opened 2023-04-12
Deprecate Multiplayer "Session" Events

AMZN-Gene:
- Deprecate IMultiplayer SessionInit and SessionShutdown events. It's not being used, and confusing as there's already SessionNotificationBus which developers should be using instead
- Rename sv_useGameLiftServer to sv_gameLiftEnabled to follow cvar naming convention 

Deprecation GHI: https://github.com/o3de/o3de/issues/15690
Original "Bug": #15626


kberg-amzn (approved):
Looks good. I would remove the deprecated methods from dev immediately..

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
```diff
@@ -186,13 +186,13 @@ namespace Multiplayer
                     mpComponent->GetNetworkEntityManager()->ClearAllEntities();
                 })
                 ->Method("GetCurrentBlendFactor", []()
+                {
+                    if (GetMultiplayer())
```
AMZN-Olex, line 190: You might as well save a pointer here, each GetMultiplayer goes through a mutex due to using AZ::Interface.

AMZN-Olex (approved):

Gems/AWSGameLift/Code/AWSGameLiftServer/Source/AWSGameLiftServerSystemComponent.cpp
```diff
@@ -19,7 +19,7 @@
 
 namespace AWSGameLift
 {
-    AZ_CVAR(bool, sv_useGameLiftServer, false, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Activate GameLift server manager and SDK");
+    AZ_CVAR(bool, sv_gameLiftEnabled, false, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Activate GameLift server manager and SDK");
```
allisaurus, line 22: is there a corresponding o3de.org docs update to reflect this change?

Gems/Multiplayer/Code/Source/MultiplayerSystemComponent.cpp
AMZN-Gene: Thank you! Fixed.

Gems/AWSGameLift/Code/AWSGameLiftServer/Source/AWSGameLiftServerSystemComponent.cpp
AMZN-Gene: Now there is... thanks!
https://github.com/o3de/o3de.org/pull/2357

allisaurus (approved):

PR #15691, opened 2023-04-12
Fix SimplePlayerSpawner to always return valid transform

mbalfour-amzn:
## What does this PR do?

If the Simple Player Spawner component had an empty entry in its list, the Respawn code would return an uninitialized transform. This fixes the logic so that invalid entity IDs will assert and return an Identity(), and missing entities will also return an Identity() instead of uninitialized data.

## How was this PR tested?

Tried spawning in MPS with an empty entry and verified that it asserted and spawned at the Identity() location.

### DIFF
--- a/Gems/Multiplayer/Code/Source/Components/SimplePlayerSpawnerComponent.cpp
+++ b/Gems/Multiplayer/Code/Source/Components/SimplePlayerSpawnerComponent.cpp
@@ -88,7 +88,17 @@ namespace Multiplayer
         }
 
         const AZ::EntityId spawnPointEntityId = m_spawnPoints[m_spawnIndex];
-        AZ::Transform spawnPointTransform;
+
+        if (!spawnPointEntityId.IsValid())
+        {
+            AZ_Assert(
+                false,
+                "Empty spawner entry at m_spawnIndex %i. Please ensure spawn index is always valid.",
+                m_spawnIndex);
+            return AZ::Transform::Identity();
+        }
+
+        AZ::Transform spawnPointTransform = AZ::Transform::Identity();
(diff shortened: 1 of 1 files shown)


AMZN-Olex (approved):

Gems/Multiplayer/Code/Source/Components/SimplePlayerSpawnerComponent.cpp
```diff
@@ -88,7 +88,17 @@ namespace Multiplayer
         }
 
         const AZ::EntityId spawnPointEntityId = m_spawnPoints[m_spawnIndex];
-        AZ::Transform spawnPointTransform;
+
+        if (!spawnPointEntityId.IsValid())
+        {
+            AZ_Assert(
```
AMZN-Gene, line 94: Looks good! I usually leave asserts for places that require a code change, so I’d make this an error. But hard breaks are pretty nice for fixing something that’ll bite us in the long run.

AMZN-Gene (approved):

Gems/Multiplayer/Code/Source/Components/SimplePlayerSpawnerComponent.cpp
mbalfour-amzn: Ugh, fair point. This _ought_ to be an error, not an assert. I've already passed AR though, so i'll submit with the assert, and we can revisit if/when we refactor the player spawner component.

kberg-amzn: Well, yeah.. long run we need to refactor this component. Registration is backwards..

kberg-amzn: and by long-run I mean we should probably do that nowish, in dev 😄

PR #15696, opened 2023-04-13
i18n - Add basic architecture of i18n

Shirley-yhm:
## What does this PR do?

- Add basic architecture of i18n.
- Tag and translation of strings will be added in later PR.

Related RFC: https://github.com/o3de/sig-content/issues/114

## How was this PR tested?

- Set `LY_I18N_BUILD=ON` and  `LY_I18N_LANGUAGE="zh_CN;it_IT"` (An array of languages)
- Add `SOFTWARE/O3DE/O3DE Editor/Settings/Language` value manually in the registry. Once language setting is added into Installer package, this step will be performed automatically.

Please refer to https://github.com/Shirley-yhm/o3de/wiki/I18n-Guidance-for-Developers for more information.


Code/Editor/Core/QtEditorApplication.cpp
```diff
@@ -426,8 +429,16 @@ namespace Editor
...
+        QTextCodec::setCodecForLocale(QTextCodec::codecForName("utf-8"));
+
+        QStringList translatorNames = {
+            "Editor"
+        };
+
+        for (QString trans : translatorNames)
+        {
+            m_translators.append(AzToolsFramework::CreateAndLoadTranslator(trans));
+        }
```
AMZN-daimini, line 441: nit: use const reference. I'd also recommend using descriptive names.

```
        for (const QString& translatorName : translatorNames)
        {
            m_translators.append(AzToolsFramework::CreateAndLoadTranslator(translatorName ));
        }
```

Code/Framework/AzToolsFramework/AzToolsFramework/Translation/Translation.h
```diff
@@ -0,0 +1,76 @@
...
+        {
+            if ( !qApp->installTranslator(translator) )
+            {
+                AZ_Warning("Translation", false, "Error installing translation %s!", unresolvedPath.data());
+            }
+        }
+        else
+        {
+            AZ_Warning("Translation", false, "Error loading translation file %s", unresolvedPath.data());
+        }
```
AMZN-daimini, line 43: nit: this could be reordered to make it more readable:

```
        QTranslator *translator = new QTranslator();
        if (!translator->load(translationFilePath))
        {
            AZ_Warning("Translation", false, "Error loading translation file %s", unresolvedPath.data());
            return translator;
        }

        if ( !qApp->installTranslator(translator) )
        {
            AZ_Warning("Translation", false, "Error installing translation %s!", unresolvedPath.data());
        }

        return translator;
```

Code/Framework/AzToolsFramework/AzToolsFramework/Translation/Translation.h
```diff
@@ -0,0 +1,76 @@
...
+    {
+        QString filename;
+        switch (AzToolsFramework::GetLanguage())
+        {
+        case AzToolsFramework::Language::Chinese:
+            filename = "zh_CN/" + modulename + "_zh_CN.qm";
+            break;
+        default:
+            break;
+        }
```
AMZN-daimini, line 72: Any chance of storing these values somewhere (maybe a map indexed by `Language`) instead of creating a big switch/case chain?

cmake/Translation.cmake
```diff
@@ -0,0 +1,43 @@
+#
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+#
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+#
+set(LY_I18N_BUILD OFF CACHE BOOL "I18N generators")
+set(LY_I18N_LANGUAGE "en_US" CACHE STRING "I18N target language")
```
lumberyard-employee-dm, line 9: The prefix of "O3DE" should be used instead of "LY", as "LY" is left over from Lumberyard and newer variables are prefixed with "O3DE" to represent the name of the engine.

Also can you add a comment that mentions that "i18n" stands for "Internationalization" as not all developers might not know what "i18n" stands for?

Code/Editor/Core/QtEditorApplication.cpp
lumberyard-employee-dm, line 441: In theory we can initialize the translators via the Module class hook within an Internationalization Gem.

The instantiated CppToolGem template [*EditorModule.cpp](https://github.com/o3de/o3de/blob/development/Templates/CppToolGem/Template/Code/Source/Tools/%24%7BName%7DEditorModule.cpp#L30-L41) would be a good place to hook in the Qt Translators

cmake/Translation.cmake
```diff
@@ -0,0 +1,43 @@
...
+        find_program(LUPDATE_EXECUTABLE lupdate PATHS "${Qt5_DIR}/../../../bin" REQUIRED)
+        set(LANGUAGES ${LY_I18N_LANGUAGE})
+        foreach(_language ${LANGUAGES})
+            set(TS_DIR "${LY_ROOT_FOLDER}/Code/Editor/Translations/${_language}")
+            file(MAKE_DIRECTORY ${TS_DIR})
+            set(TS_FILE "${TS_DIR}/${module_name}_${_language}.ts")
+            execute_process(COMMAND ${LUPDATE_EXECUTABLE} ${target_source_dir} -ts ${TS_FILE} -silent)
+            if(LY_HOST_PLATFORM_DETECTION_Windows STREQUAL "Windows")
+                file(READ ${TS_FILE} ts_file_content)
+                file(CONFIGURE OUTPUT ${TS_FILE} CONTENT "${ts_file_content}" NEWLINE_STYLE CRLF)
```
lumberyard-employee-dm, line 39: Is this just updating the file with "carriage-return line-feed"?

Is this actually needed?

Code/Framework/AzToolsFramework/AzToolsFramework/Translation/Translation.h
```diff
@@ -0,0 +1,76 @@
...
+        delete translator;
+        translator = nullptr;
+    }
+ 
+    inline Language GetLanguage()
+    {
+        QSettings settings("O3DE", "O3DE Editor");
+        int language = settings.value("Settings/Language").toInt();
+        return static_cast<Language>(language);
+    }
```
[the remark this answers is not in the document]
lumberyard-employee-dm: +1 on that as the current setup is Windows specific using the Windows Registry, while the [wiki page](https://github.com/Shirley-yhm/o3de/wiki/I18n-Guidance-for-Developers) doesn't mention how to enable the setting for Linux.

Code/Framework/AzToolsFramework/AzToolsFramework/Translation/Translation.h
```diff
@@ -0,0 +1,76 @@
...
+    {
+        English,
+        Chinese
+    };
+ 
+    inline QTranslator* LoadTranslator(const QString& filename)
+    {
+        AZStd::array<char, AZ::IO::MaxPathLength> unresolvedPath;
+        QString dirpath = "@engroot@/Assets/Editor/Translations/" + filename;
+        AZ::IO::FileIOBase::GetInstance()->ResolvePath(dirpath.toStdString().c_str(), unresolvedPath.data(), unresolvedPath.size());
```
lumberyard-employee-dm, line 29: I would recommend using the AzCore/Utils/Utils.h [GetEngnePath](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Utils/Utils.h#L74-L77) function here.

That avoids needing the FileIOBase instance being initialized at this point.
Plus the code is a simpler to write.
```suggestion
        auto translationFilePath = (AZ::IO::FixedMaxPath(AZ::Utils::GetEnginePath()) / "Assets/Editor/Translation").LexicallyNormal();
```

lumberyard-employee-dm: Afterwards a QString can be created using `QString::fromUtf8` from the AZ::IO::FixedMaxPath

```c++
auto translationFilePath = QString::fromUtf8(translationResolvedPath.c_str(), static_cast<int>(translationResolvedPath.Native().size()));
```

cmake/Translation.cmake
lawsonamzn: I'm also concerned about using a CMAKE variable for the I18N language at all.  It being a CMake variable would indicate that in some way it is baked into the executable.  However, we don't really want to recompile executables and binaries in order to switch languages, right?  Should it not be a runtime option, with all language QMs built?

lawsonamzn: Actually discount my above comment, looksl ike this is the _array_ of languages _to build_
Perhaps the comment should be "List of languages to include in the build" rather than "target language" ?

Code/Editor/Core/QtEditorApplication.cpp
```diff
@@ -426,8 +429,16 @@ namespace Editor
 
     void EditorQtApplication::InstallEditorTranslators()
     {
-        m_editorTranslator =        CreateAndInitializeTranslator("editor_en-us.qm", ":/Translations");
-        m_assetBrowserTranslator =  CreateAndInitializeTranslator("assetbrowser_en-us.qm", ":/Translations");
+        QTextCodec::setCodecForLocale(QTextCodec::codecForName("utf-8"));
+
+        QStringList translatorNames = {
+            "Editor"
```
spham-amzn, line 435: What happened to the asset browser translator?

Code/Editor/Core/QtEditorApplication.cpp
```diff
@@ -426,8 +429,16 @@ namespace Editor
...
+        };
+
+        for (QString trans : translatorNames)
```
spham-amzn, line 438: `translaterNames` doesn't really have to be a QStringList:

```
    constexpr const char *translatorNames[] = {
        "English"
    };
    for (const char* translator: translatorNames)
    {
        m_translators.append(AzToolsFramework::CreateAndLoadTranslator(QString(translator)));
    }
```
Should also work

Code/Editor/Core/QtEditorApplication.cpp
spham-amzn: If this is intended to be expanded in the future, i'd recommend exposing the list of translator names outside of the cpp file, maybe into the header file or inside the definition of this class as a constant list.

cmake/Translation.cmake
```diff
@@ -0,0 +1,43 @@
...
+endif()
+
+function(add_translation_module target_source_dir module_name)
+    if(NOT LY_I18N_BUILD)
+        return()
+    endif()
+
+    if (PAL_TRAIT_BUILD_HOST_TOOLS)
+        find_package(Qt5 COMPONENTS LinguistTools REQUIRED)
+        find_program(LUPDATE_EXECUTABLE lupdate PATHS "${Qt5_DIR}/../../../bin" REQUIRED)
```
lumberyard-employee-dm, line 30: The `QT_LRELEASE_EXECUTABLE` cmake variable can be used here directly.

Code/Editor/Core/QtEditorApplication.cpp
Shirley-yhm: We have added i18n to most modules and they will be added in later PR. Editor is just the first module we plan to tag and translate. Translators of the whole engine will include "Editor", "AssetProcessor",  "AzCore", "AzToolsFramework", "AzFramework", "AzQtComponents", "Legacy", ... , and every Gem.
I will move the list to a better position.

cmake/Translation.cmake
Shirley-yhm: `lupdate` is needed here to generate TS files, not `lrelease`.

lumberyard-employee-dm: Ok, in that case then the QT_PATH path Cache variable that is set in the [FindQt5.cmake](https://github.com/o3de/3p-package-source/blob/main/package-system/Qt/FindQt.cmake#L15-L19) can be used to get the `lupdate` executable then

Code/Editor/Core/QtEditorApplication.cpp
spham-amzn: Not all of the modules are linked against Qt. In general, all of the 'Editor' versions of the gems, AssetProcessor, and some tools will be linked against Qt, and will have access to the translators.

cmake/Translation.cmake
Shirley-yhm: Yes, we add it just in case files are marked as modified after checkout on Windows. I'll delete it.

guillaume-haerinck:
Closed as superseeded by https://github.com/o3de/o3de/pull/19554

Issue #15157, opened 2023-03-16
DPE Debug App: Crash when changing elements in container


forkercat:
**Describe the bug**

The DPE debug app (DPEDebugViewStandalone) would crash if we try to add or clear elements for containers (vector, set, map).

![image](https://user-images.githubusercontent.com/11157226/225775229-1b89a7c0-7b32-4462-9cf3-2bda7b1e7cbe.png)

**Steps to reproduce**
1. In VS's Solution Explorer, set `DPEDebugViewStandalone` as start up project (under Framework/AzQtComponents)
2. Run the application
3. Expand "Container"
4. Click "+" button to add an element

**Expected behavior**

Element is added without any issue.

**Actual behavior**

Crash

<img height=200 src="https://user-images.githubusercontent.com/11157226/225775675-256a08f2-e8ef-478f-b1bc-5366256e36a1.png">

**Screenshots/Video**

Can be easily reproduced

**Found in Branch**

Development https://github.com/o3de/o3de/commit/b2ca8b13d5b334369f5f291e6773978a11b0ee73

**Desktop/Device**

* Device: PC
* OS: Windows 10
* CPU: AMD Ryzen 5950x
* GPU: NVIDIA RTX 3070
* Memory: 64 GB

**Additional context**

<details>
  <summary>CLICK ME TO SEE THE CALL STACK</summary>

```
> [Inline Frame] DPEDebugViewStandalone.exe!AzToolsFramework::DPEDebugModel::HandleDomChange(const AZ::Dom::Patch &) Line 531	C++
 DPEDebugViewStandalone.exe!AzToolsFramework::DPEDebugModel::SetAdapter::__l2::<lambda_2>::operator()(const AZ::Dom::Patch & patch) Line 346	C++
 [Inline Frame] DPEDebugViewStandalone.exe!AZStd::function_intermediate<void,AZ::Dom::Patch const &>::operator()(const AZ::Dom::Patch &) Line 604	C++
 [Inline Frame] DPEDebugViewStandalone.exe!AZStd::function<void __cdecl(AZ::Dom::Patch const &)>::operator()(const AZ::Dom::Patch &) Line 684	C++
 DPEDebugViewStandalone.exe!AZ::Event<AZ::Dom::Patch const &>::Signal(const AZ::Dom::Patch & <params_0>) Line 288	C++
[...]

PR #15705, opened 2023-04-13
DPE: Prevent nullptr node crash in DPEDebugModel

amzn-tmryan:
Resolves https://github.com/o3de/o3de/issues/15157

**Description**
This PR fixes a crash in the DPEDebugModel caused by paths that have path entries beyond their property. This meant that a particular `for` loop was continuing after finding the last keyed PathEntry and eventually returning nullptr.

**Testing**
- Verified no crash occurs when doing things that previously crashed in the DPEDebugViewStandalone app
- Verified that the for loop no longer continues after encountering the first keyed PathEntry


forkercat (approved):
Thanks!

alexmontAmazon (approved):

alexmontAmazon (approved):
looks good! I took the liberty of kicking off AR

PR #15707, opened 2023-04-13
Replaced the editor's base class with AZ::Render::EditorRenderCompone…

l-sempe:
…ntAdapter

## What does this PR do?

`AZ::Render::EditorRenderComponentAdapter` is more consistent with other rendering components. It also offers notification of visibility changes that can be used to toggle feature processors.

### DIFF
--- a/Templates/GraphicsGem/Template/Code/Source/Tools/Components/Editor${Name}Component.cpp
+++ b/Templates/GraphicsGem/Template/Code/Source/Tools/Components/Editor${Name}Component.cpp
@@ -85,4 +85,16 @@ namespace ${Name}
             return;
         }
     }
+
+    void Editor${Name}Component::OnEntityVisibilityChanged(bool visibility)
+    {
+        if (visibility)
+        {
+            m_controller.EnableFeatureProcessor(GetEntityId());
+        }
+        else
+        {
+            m_controller.DisableFeatureProcessor();
+        }
+    }
 }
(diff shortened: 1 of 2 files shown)


cgalvan (approved):

lumberyard-employee-dm (approved):

Templates/GraphicsGem/Template/Code/Source/Tools/Components/Editor${Name}Component.cpp
```diff
@@ -85,4 +85,16 @@ namespace ${Name}
...
+    {
+        if (visibility)
+        {
+            m_controller.EnableFeatureProcessor(GetEntityId());
+        }
+        else
+        {
+            m_controller.DisableFeatureProcessor();
+        }
+    }
```
gadams3, line 99: Looks good to me.  The editor render component adapter will activate/deactivate the component controller in response to editor visibility changes by default.  While not always optimal, it would automatically provide this functionality.  The other thing worth mentioning is that this assumes there will be a 1 to 1 relationship between a feature processor and a component.  There could be no feature processor or more than one feature processor.  It might be worth using more generic function names like enabled rendering and disable rendering.  Is it necessary to pass in the entity ID, which the controller should already have?

gadams3 (approved):

invertednormal (approved):

PR #15711, opened 2023-04-14
Remote gem version support

AMZN-alexpete:
## What does this PR do?

Extends gem version support to remote gems and to some extent, projects and templates.

### Important changes
- do not change existing `gems`, `projects`, `templates` fields in `repo.json` this would break backwards compatibility, instead use new fields to store object json data in `repo.json`  (new fields are `gems_data`, `projects_data`, `templates_data`)
- Gem Catalog now defaults to showing compatible gems.  Compatibility filtering is based on `compatible_engines`, `gem_dependencies` and `engine_api_dependecies` and not based on full dependency resolution which would be too slow.  However, full dependency resolution is performed when a user attempts to save project changes.
- Remote gems are downloaded into versioned subfolders e.g.  `c:/path/to/<gem name>/<gem version>/`  If a gem has no version then `0.0.0` is used.
- https://canonical.o3de.org remote repo is added in `engine.json`
- Remote repos can be activated/deactivated so their content is no longer accessible in the Gem Catalog, this will be necessary for recommended repos

### Other changes/fixes
- added ability to refresh a repo from the o3de CLI using `o3de.bat repo --refresh-repo <repo url>` or all repos `o3de.bat repo --refresh-all-repos`
- added ability to clone a remote object at a specific reference.  If the object data contains a `source_control_uri` and `source_control_ref`.  This is only supported by the CLI currently using `o3de.bat download <object name and optional version specifier> --use-source-control`  e.g. `o3de.bat download XR==1.2.3  --use-source-control`
- add 'Missing' filter and show missing gems versions a project uses
- add Official badge/banner for o3de canonical repo and sort it to top.
image

Screenshot of main gems tab that shows compatible filter and a remote gem available for download (PopcornFX)
image

Screenshot showing `Missing` filter and missing local and remote gem versions
image

Relevant RFC: https://github.com/o3de/sig-core/issues/44

## How was this PR tested?

Unit testing:
- updated and created unit test for new functionality, ran all o3de pytests

Manual testing:
- set up a repository and verified I could see compatible, and incompatible content
- downloaded and uninstalled remote gem versions
- verified new versions of PopcornFX could be downloaded and uninstalled
- verified refreshing remote repositories parses and loads new data and last updated field is correctly shown

# Exception Checklist
- [x] 1. Owning SIG requests the exception, by a Sig Representative (Chair/co-chair) adding the PR to the exception requests queue board.
- [x] 2. (SIG-Content) Owning SIG comments on the PR explaining impact if this code does not go into stabilization.
- [x] 3. Unrelated SIG representative (chair/cochair) approves the exception, adding a comment indicating approval and the SIG they represent.
- [x] 4. UX SIG representative (chair/cochair) approves the exception, adding a comment indicating approval and the SIG they represent.
- [x] 5. Testing SIG representative (chair/cochair) approves the exception, adding a comment indicating approval and the SIG they represent.
 (Must occur at some point before SIG-Release approval) Code has been approved for merging by maintainers
- [x] 6. SIG-Release Release Manager or co-Release Manager approves the exception, adding a comment indicating approval and the SIG they represent.
- [x] 7. If all approvals are accepted, then the exception PR may be merged
- [ ] 8. Merged PR is verified successful via testing. If Merged PR verification testing failed, consult reps from step 3-5 and discussion occurs on fixing the failure vs backing out change
- [ ] 9. Exception is complete


AMZN-alexpete:
This is mostly ready, but I intend to also include a final change that adds the canonical repo default.

lsemp3d:
As chair of SIG-Content I approve this change to go into the release, it's in important and useful mechanism to support remote gems and avoid version conflicts and complexities

Code/Tools/ProjectManager/Source/GemCatalog/GemModel.cpp
```diff
@@ -496,6 +525,41 @@ namespace O3DE::ProjectManager
...
+
+        // versions are sorted from highest to lowest so return the first compatible version
+        for (const auto& versionVariant : versionList)
+        {
+            const GemInfo& variantGemInfo = versionVariant.value<GemInfo>();
+            if(variantGemInfo.IsCompatible())
+            {
+                return variantGemInfo.m_version;
+            }
+        }
```
lumberyard-employee-dm, line 552: The for loop does the same thing as the if statement.
It checks in order that the gemInfo version is compatible from highest version to lowest version.

Therefore the if statement shouldn't be needed.

Code/Tools/ProjectManager/Source/GemCatalog/GemModel.cpp
```diff
@@ -142,6 +154,26 @@ namespace O3DE::ProjectManager
...
+    bool GemModel::ShouldUpdateItemDataFromGemInfo(const QModelIndex& modelIndex, const GemInfo& gemInfo)
+    {
+        // get the most compatible version or empty string if none are compatible
+        const QString mostCompatibleVersion = GetMostCompatibleVersion(modelIndex);
+        const bool newVersionIsCompatible = gemInfo.IsCompatible();
+        int versionResult = ProjectUtils::VersionCompare(gemInfo.m_version, modelIndex.data(RoleVersion).toString());
+
+        if (mostCompatibleVersion.isEmpty() && !newVersionIsCompatible)
+        {
+            // no compatibile versions available (yet) so refresh if version is the same or higher
```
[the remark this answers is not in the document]
AMZN-alexpete: compatibile 😆

rainbj:
Reviewed UX during a one on one meeting and provided feedback. Feedback added to the workflow. Approved Sig-UI-UX

Code/Tools/ProjectManager/Source/GemCatalog/GemModel.cpp
```diff
@@ -764,13 +847,84 @@ namespace O3DE::ProjectManager
...
+        }
+        else
+        {
+            if (currentVersion != versions.at(0).value<GemInfo>().m_version)
+            {
+                return true;
+            }
+
+            // if this is a remote gem that hasn't been downloaded, show that the update is
+            // available the user has downloaded an older version
```
lumberyard-employee-dm, line 903: There seems to be a missing conjunction `to` here?
```suggestion
            // if this is a remote gem that hasn't been downloaded, show that the update is
            // available to the user has downloaded an older version
```

Code/Tools/ProjectManager/Source/GemCatalog/GemSortFilterProxyModel.h
```diff
@@ -69,7 +69,7 @@ namespace O3DE::ProjectManager
         const QSet<QString>& GetFeatures() const { return m_featureFilter; }
         void SetFeatures(const QSet<QString>& features) { m_featureFilter = features; InvalidateFilter(); }
 
-        bool GetUpdateAvailable() const { return m_updateAvailableFilter; }
+        const bool GetCompatibleFilterFlag() const { return m_compatibleOnlyFilter; }
```
lumberyard-employee-dm, line 72: A const value type shouldn't be returned.
The compiler ignores it anyway and returns just a `bool`
```suggestion
        bool GetCompatibleFilterFlag() const { return m_compatibleOnlyFilter; }
```

Code/Tools/ProjectManager/Source/GemRepo/GemRepoItemDelegate.cpp
```diff
@@ -102,14 +112,56 @@ namespace O3DE::ProjectManager
...
+        currentHorizontalOffset += sectionSize;
+        sectionSize = m_headerWidget->m_header->sectionSize(static_cast<int>(HeaderOrder::Updated));
+        auto lastUpdated = GemRepoModel::GetLastUpdated(modelIndex);
+
+        // get the month day and year in the preferred locale's format (QLocale defaults to the OS locale)
+        QString monthDayYear = lastUpdated.toString(QLocale().dateFormat(QLocale::ShortFormat));
 
-        QString repoUpdatedDate = GemRepoModel::GetLastUpdated(modelIndex).toString(RepoTimeFormat);
+        // always show 12 hour + minutes + am/pm
+        QString hourMinuteAMPM = lastUpdated.toString("h:mmap");
```
lumberyard-employee-dm, line 159: Time zones are important here.
What if the user experiences a change in time on their machine due to daylight saving time and the value is off by an hour.
Or the user travels to another time zone.

Code/Tools/ProjectManager/Source/GemRepo/GemRepoModel.cpp
```diff
@@ -11,21 +11,28 @@
...
     QItemSelectionModel* GemRepoModel::GetSelectionModel() const
     {
         return m_selectionModel;
     }
 
+    void SetItemDataSorted(QStandardItem* item, const QSet<QString>& stringSet, int role)
+    {
+        auto stringList = QStringList(stringSet.values());
+        stringList.sort(Qt::CaseInsensitive);
+        item->setData(stringList, role);
```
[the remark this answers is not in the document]
lumberyard-employee-dm: So apparently [QSet](https://doc.qt.io/qt-6/qset.html#details) is a hash table based container, while [QMap](https://doc.qt.io/qt-6/qmap.html#details) is a red-black sorted tree container

Code/Tools/ProjectManager/Source/ProjectManagerDefs.h
```diff
@@ -30,6 +30,7 @@ namespace O3DE::ProjectManager
     static const QString ProjectCMakeCommand = "cmake";
     static const QString ProjectCMakeBuildTargetEditor = "Editor";
 
-    static const QString RepoTimeFormat = "dd/MM/yyyy hh:mmap";
+    static const QString RepoTimeFormat = "MM/dd/yyyy hh:mmap";
+    static const QString CanonicalRepoUri = "https://canonical.o3de.org";
```
lumberyard-employee-dm, line 34: A better format that doesn't have to worry about localization is the [ISO8601](https://en.wikipedia.org/wiki/ISO_8601) format. It avoids the issues with day-major(rest of the world) vs month-major(US) dates, by specify the date as `yyyy-mm-dd` so it goes from most general date value(year) to the most specific(day)

Code/Tools/ProjectManager/Source/PythonBindings.cpp
```diff
@@ -1686,23 +1768,35 @@ namespace O3DE::ProjectManager
                 gemRepoInfo.m_path = gemRepoInfo.m_directoryLink = Py_To_String(repoPath);
 
                 QString lastUpdated = Py_To_String_Optional(data, "last_updated", "");
-                gemRepoInfo.m_lastUpdated = QDateTime::fromString(lastUpdated, RepoTimeFormat);
+                // this is NOT the same as RepoTimeFormat which uses 12 hour + am/pm
+                gemRepoInfo.m_lastUpdated = QDateTime::fromString(lastUpdated, "dd/MM/yyyy HH:mm");
```
lumberyard-employee-dm, line 1772: The date format of `dd/MM/yyyy` is confusing to US users.

Values such as 04/07/2023 would mean July 4th 2023.

scripts/o3de/o3de/github_utils.py
```diff
@@ -13,18 +13,19 @@
...
+from o3de import gitproviderinterface, utils
 
 LOG_FORMAT = '[%(levelname)s] %(name)s: %(message)s'
 
 logger = logging.getLogger('o3de.github_utils')
 logging.basicConfig(format=LOG_FORMAT)
 
 class GitHubProvider(gitproviderinterface.GitProviderInterface):
-    def get_specific_file_uri(parsed_uri):
+    def get_specific_file_uri(parsed_uri:ParseResult):
```
lumberyard-employee-dm, line 28: nitpick: missing whitespace
```suggestion
    def get_specific_file_uri(parsed_uri: ParseResult):
```

scripts/o3de/o3de/github_utils.py
```diff
@@ -48,11 +49,25 @@ def get_specific_file_uri(parsed_uri):
 
         return parsed_uri
 
-    def clone_from_git(uri, download_path: pathlib.Path) -> int:
+    def clone_from_git(uri:ParseResult, download_path: pathlib.Path, force_overwrite: bool = False, ref: str = None) -> int:
```
lumberyard-employee-dm, line 52: nitpick: missing whitespace
```suggestion
    def clone_from_git(uri: ParseResult, download_path: pathlib.Path, force_overwrite: bool = False, ref: str = None) -> int:
```

scripts/o3de/o3de/download.py
```diff
@@ -363,6 +405,9 @@ def add_parser_args(parser):
     parser.add_argument('-f', '--force', action='store_true', required=False,
                             default=False,
                             help = 'Force overwrite the current object')
+    parser.add_argument('-s', '--use-source-control', action='store_true', required=False,
+                            default=False,
+                            help = 'Acquire from source control instead of downloading a .zip archive.  Requires that the object has ')
```
lumberyard-employee-dm, line 410: Seems strange that the option is `-s`, but it doesn't start with the letter of s.

How about `-usc` instead?
Also it would better for showing the usage if the long option appeared first, so that the usage text outputs `--use-source-control`
```suggestion
    parser.add_argument('--use-source-control', '-s', action='store_true', required=False,
                            default=False,
                            help = 'Acquire from source control instead of downloading a .zip archive.  Requires that the object has ')
```

scripts/o3de/o3de/git_utils.py
```diff
@@ -0,0 +1,74 @@
...
+
+        return proc.returncode
+
+def get_generic_git_provider(parsed_uri:ParseResult) -> GenericGitProvider or None:
+    # the only requirement we have is one of the path components ends in .git
+    # this could be relaxed further 
+    if parsed_uri.netloc and parsed_uri.scheme and parsed_uri.path:
+        for element in parsed_uri.path.split('/'):
+            if element.strip().endswith(".git"):
+                return GenericGitProvider
```
lumberyard-employee-dm, line 73: This is returning the GenericGitProvider class itself and not an instance of the class.

If this is the intent, then the `get_specific_file_uri` and `clone_from_git` should be decorated with `@staticmethod` to make sure they work correctly when used with an instance.
i.e This code currently will not work
```python
git_provider = GenericGitProvider()
git_provider.clone_from_git('https://github.com/o3de/o3de.get', '/home/engines/o3de')
```

Because the `git_provider` object will be passed in as the first parameter.

scripts/o3de/o3de/github_utils.py
```diff
@@ -61,9 +76,18 @@ def clone_from_git(uri, download_path: pathlib.Path) -> int:
...
+                with subprocess.Popen(params, stdout=subprocess.PIPE) as proc:
+                    print(proc.stdout.read())
+            except Exception as e:
+                logger.error(str(e))
+                return 1
+
         return proc.returncode
 
-def get_github_provider(parsed_uri) -> GitHubProvider or None:
+def get_github_provider(parsed_uri:ParseResult) -> GitHubProvider or None:
```
lumberyard-employee-dm, line 90: nitpick: missing whitespace after color
```suggestion
def get_github_provider(parsed_uri: ParseResult) -> GitHubProvider or None:
```

scripts/o3de/o3de/gitproviderinterface.py
```diff
@@ -11,10 +11,11 @@
 
 import abc
 import pathlib
+from urllib.parse import ParseResult
 
 class GitProviderInterface(abc.ABC):
     @abc.abstractmethod
-    def get_specific_file_uri(parsed_uri):
+    def get_specific_file_uri(parsed_uri:ParseResult):
```
lumberyard-employee-dm, line 18:
```suggestion
    def get_specific_file_uri(parsed_uri: ParseResult):
```

scripts/o3de/o3de/gitproviderinterface.py
```diff
@@ -23,11 +24,13 @@ def get_specific_file_uri(parsed_uri):
         pass
 
     @abc.abstractmethod
-    def clone_from_git(uri, download_path: pathlib.Path) -> int:
+    def clone_from_git(uri:ParseResult, download_path: pathlib.Path, force_overwrite: bool = False, ref: str = None) -> int:
```
lumberyard-employee-dm, line 27:
```suggestion
    def clone_from_git(uri: ParseResult, download_path: pathlib.Path, force_overwrite: bool = False, ref: str = None) -> int:
```

scripts/o3de/o3de/manifest.py
```diff
@@ -261,8 +261,17 @@ def get_manifest_restricted() -> list:
...
+
+    engine_json_data = get_engine_json_data(engine_path=get_this_engine_path())
+    if engine_json_data:
+        repos.update(set(engine_json_data.get('repos', [])))
+
+    manifest_json_data = load_o3de_manifest()
+    if manifest_json_data:
+        repos.update(set(manifest_json_data.get('repos', [])))
+
+    return list(repos)
```
lumberyard-employee-dm, line 274: Seems like projects should also be able to support repos as well.
Why not add an optional project_path parameter here, looks for any repos within the project.json as well.
`def get_manifest_repos(project_path: pathlib.PurePath = none)`

scripts/o3de/o3de/repo.py
```diff
@@ -199,6 +203,7 @@ def process_add_o3de_repo(file_name: str or pathlib.Path,
     with file_name.open('w') as f:
         try:
             time_now = datetime.now()
+            # %H is 24 hour
             time_str = time_now.strftime('%d/%m/%Y %H:%M')
```
lumberyard-employee-dm, line 207: I think it would be better to use ISO8601 timestamp format here as it is doesn't have ambiguity with regards to the user culture.

scripts/o3de/o3de/repo.py
```diff
@@ -422,3 +485,81 @@ def search_o3de_object(manifest_json, o3de_object_uris, search_func):
...
+        if not repo_json_data:
+            logger.error(f'Repository JSON {repo_json_cache_file} could not be loaded or is missing required values')
+            repo_json_cache_file.unlink()
+            return 1
+
+    with repo_json_cache_file.open('w') as f:
+        try:
+            time_now = datetime.now()
+            # %H is 24 hour
+            time_str = time_now.strftime('%d/%m/%Y %H:%M')
```
lumberyard-employee-dm, line 512: repeat: I recommend using the ISO8601 format for the [date](https://wikiless.org/wiki/ISO_8601?lang=en#Dates) part of the datetime at least.

scripts/o3de/o3de/download.py
```diff
@@ -122,23 +121,49 @@ def download_o3de_object(object_name: str, default_folder_name: str, dest_path:
...
+    if use_source_control:
+        origin_uri = downloadable_object_data.get('source_control_uri')
+        if not origin_uri:
+            logger.error(f"Tried to use source control to acquire {object_name} but the 'source_control_uri' is empty or missing.")
+            return 1
+    else:
+        origin_uri = downloadable_object_data.get('download_source_uri')
+        if not origin_uri:
+            # fallback to schemaVersion 1.0.0
+            origin_uri = downloadable_object_data.get('origin_uri')
```
lumberyard-employee-dm, line 133: What is this comment for?

Code/Tools/ProjectManager/Source/GemRepo/GemRepoItemDelegate.cpp
AMZN-alexpete: That's a good call, though because this is an existing field that didn't use UTC time it might require some extra logic to still support old fields in the old format.

Code/Tools/ProjectManager/Source/GemRepo/GemRepoModel.cpp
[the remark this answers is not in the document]
AMZN-alexpete: unfortunately neither supports case insensitive sorting, and `QSet<QString>` is not setup by default to be placed in a `QVariant`  like `QStringList` is.  I may come back to this in a future PR to make this simpler

Code/Tools/ProjectManager/Source/PythonBindings.cpp
AMZN-alexpete: I agree! which is why I changed the display value in Project Manager to show the format used in the current locale so US users see `mm/dd/yyyy` and euro users see `dd/MM/yyyy`.
If I convert the underlying format to ISO then the only reason this could will remain is for backwards compatibility.

Code/Tools/ProjectManager/Source/GemRepo/GemRepoModel.cpp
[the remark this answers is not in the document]
lumberyard-employee-dm: That is fine.
A simpler approach might be to just use an AZStd::set and then the iterator constructor for `QStringList` to construct from the AZStd::set begin and end iterators

scripts/o3de/tests/test_repo.py
```diff
@@ -65,27 +65,21 @@
...
     "repo_uri": "https://downloads.testgem3.com/o3de-repo",
-
     "summary": "Studios Repository for the testgem3 Gem.",
-
     "additional_info": "",
-
     "last_updated": "2023-01-19",
-
-    "gems": [{
+    "gems":[],
```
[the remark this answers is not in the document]
AMZN-alexpete: I'm not sure I understand your question, but maybe this will help, the fields in "gems", "projects" and "templates" are URLs, the "gems_data" fields contains JSON data.

Code/Tools/ProjectManager/Source/GemCatalog/GemFilterWidget.cpp
```diff
@@ -428,6 +471,12 @@ namespace O3DE::ProjectManager
         {
             m_filterProxyModel->SetGemActive(GemSortFilterProxyModel::GemActive::NoFilter);
         }
+
+        // Missing
+        if (button == buttons[4])
```
lumberyard-employee-dm, line 476: The above uses of `buttons<SomeInt>` all stored it in a variable which helps document what it was for
```
QAbstractButton* activeButton = buttons[2];
QAbstractButton* inactiveButton = buttons[3];
```

I also recommend doing that for `buttons[4]`
```suggestion
        const QAbstractButton* missingGemButton = buttons[4];
        if (button == missingGemButton)
```

AMZN-Dk:
Reviewed this change and it looks good to bring in, approved from the Testing SIG

Code/Tools/ProjectManager/Source/GemCatalog/GemCatalogScreen.cpp
```diff
@@ -389,11 +407,25 @@ namespace O3DE::ProjectManager
...
                 }
-                if (added && (GemModel::GetDownloadStatus(modelIndex) == GemInfo::DownloadStatus::NotDownloaded) ||
-                    (GemModel::GetDownloadStatus(modelIndex) == GemInfo::DownloadStatus::DownloadFailed))
+
+                if (added)
                 {
-                    m_downloadController->AddObjectDownload(GemModel::GetName(modelIndex), "", DownloadController::DownloadObjectType::Gem);
-                    GemModel::SetDownloadStatus(*m_gemModel, modelIndex, GemInfo::DownloadStatus::Downloading);
+                    if (newVersion.isEmpty() && (GemModel::GetDownloadStatus(modelIndex) == GemInfo::DownloadStatus::NotDownloaded) ||
+                    (GemModel::GetDownloadStatus(modelIndex) == GemInfo::DownloadStatus::DownloadFailed))
```
lumberyard-employee-dm, line 414: The second line of the `if` is normally indented.
```suggestion
                    if (newVersion.isEmpty() && (GemModel::GetDownloadStatus(modelIndex) == GemInfo::DownloadStatus::NotDownloaded) ||
                        (GemModel::GetDownloadStatus(modelIndex) == GemInfo::DownloadStatus::DownloadFailed))
```

Code/Tools/ProjectManager/Source/GemCatalog/GemInspector.cpp
```diff
@@ -236,6 +238,26 @@ namespace O3DE::ProjectManager
...
+            if(!gemInfo.m_compatibleEngines.isEmpty())
+            {
+                if (m_readOnly)
+                {
+                    m_compatibilityTextLabel->setText(tr("This version is not known to be compatible with the current engine"));
+                }
+                else
+                {
+                    m_compatibilityTextLabel->setText(tr("This version is not known to be compatible with the engine this project uses"));
+                }
```
lumberyard-employee-dm, line 253: How does the read only option change which engine the project isn't associated with?

Code/Tools/ProjectManager/Source/GemRepo/GemRepoScreen.cpp
```diff
@@ -52,34 +56,70 @@ namespace O3DE::ProjectManager
 
         vLayout->addWidget(m_contentStack);
 
-        Reinit();
+        m_notificationsView = AZStd::make_unique<AzToolsFramework::ToastNotificationsView>(this, AZ_CRC("ReposNotificationsView"));
```
lumberyard-employee-dm, line 59: This should use `AZ_CRC_CE` for compile time evauluation
```suggestion
        m_notificationsView = AZStd::make_unique<AzToolsFramework::ToastNotificationsView>(this, AZ_CRC_CE("ReposNotificationsView"));
```

cmake/Platform/Common/Install_common.cmake
```diff
@@ -492,6 +492,15 @@ function(ly_setup_cmake_install)
         list(JOIN relative_templates ",\n${indent}" O3DE_INSTALL_TEMPLATES)
     endif()
 
+    # Read the "repos" key from the source engine.json
+    o3de_read_json_array(engine_repos ${LY_ROOT_FOLDER}/engine.json "repos")
+    if(engine_repos)
+        foreach(repo_uri ${engine_repos})
+            list(APPEND repos "\"${repo_uri}\"")
+        endforeach()
+        list(JOIN repos ",\n${indent}" O3DE_INSTALL_REPOS)
```
lumberyard-employee-dm, line 501: Can you also sort the repo URIs as well?
```suggestion
        list(SORT repos CASE INSENSITIVE)
        list(JOIN repos ",\n${indent}" O3DE_INSTALL_REPOS)
```

scripts/o3de/o3de/compatibility.py
```diff
@@ -75,7 +75,15 @@ def get_most_compatible_project_engine_path(project_path:pathlib.Path,
             engine_version = '0.0.0'
 
         if has_compatible_version([project_engine], engine_name, engine_version):
-            if not most_compatible_engine_path:
+
+            # prefer the engine this project or template resides in if it is compatible
+            if pathlib.Path(engine_path) in pathlib.Path(project_path).parents:
```
lumberyard-employee-dm, line 80: pathlib [is_relative_to](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.is_relative_to) would be the more accurate check here.
```suggestion
            # prefer the engine this project or template resides in if it is compatible
            if pathlib.PurePath(project_path).is_relative_to(pathlib.PurePath(engine_path)):
```

lumberyard-employee-dm (approved):
I am signing on this PR and exception as a representative of SIG-Core which as an unrelated SIG.

Code/Tools/ProjectManager/Source/GemCatalog/GemInspector.cpp
AMZN-alexpete: I should add a comment in the code - in read only mode there is no project being configured, you are looking at all gems available via the engine and o3de manifest so I tried to make the error message not mention "project"

tkothadev (approved):

AMZN-Phil (approved):

rainbj:
@lumberyard-employee-dm - I somehow missed this ticket. Did the date format issue here get resolved? Thanks

PR #15716, opened 2023-04-14
Asset Processor - Optimize initial startup time

amzn-mike:
## What does this PR do?

* Added a string path cache to the FileStateCache to avoid re-normalizing and lowercasing the same string many times.  Profiling showed this as a hot spot and it was being called over 5M times during startup.
* Added a dependency cache to QueryAbsolutePathDependenciesRecursive.  Profiling also showed this as a hotspot for several different items.  The cache only applies during startup and is cleared/disabled after reaching idle state.
* Switched from a queue to an unordered_set as the order doesn't actually matter and profiling showed unordered_set to be faster.
* Added an optional return of the scanfolder for FindFirstMatchingFile to optimize one of the SourceAssetReference calls during the above.

## How was this PR tested?

Manually ran AP, wrote some verification code to compare the old and new results to make sure the cache was still returning the same values.

Before: 9157ms
After: 4530ms
Speedup: 2.02x

Times above are for AutomatedTesting project on an already-populated cache

### DIFF
--- a/Code/Tools/AssetProcessor/native/AssetManager/FileStateCache.cpp
+++ b/Code/Tools/AssetProcessor/native/AssetManager/FileStateCache.cpp
@@ -42,7 +42,7 @@ namespace AssetProcessor
         LockGuardType scopeLock(m_mapMutex);
         QString key = PathToKey(existingInfo.m_filePath);
         m_fileInfoMap[key] = FileStateInfo(existingInfo);
-        
+
         // it is possible to update the cache so that the info is known, but the hash is not.
         if (hash == InvalidFileHash)
         {
@@ -151,6 +151,8 @@ namespace AssetProcessor
 
     void FileStateCache::InvalidateHash(const QString& absolutePath)
     {
+        m_keyCache = {}; // Clear the key cache, its only really intended to help speedup the startup phase
+
         auto fileHashItr = m_fileHashMap.find(PathToKey(absolutePath));
 
         if (fileHashItr != m_fileHashMap.end())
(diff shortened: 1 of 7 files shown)


AMZN-stankowi (approved):

Code/Tools/AssetProcessor/native/AssetManager/FileStateCache.cpp
```diff
@@ -163,16 +165,22 @@ namespace AssetProcessor
 
     QString FileStateCache::PathToKey(const QString& absolutePath) const
     {
-        QString normalized = AssetUtilities::NormalizeFilePath(absolutePath);
+        auto cached = m_keyCache.find(absolutePath);
```
lawsonamzn, line 168: I'm surprised this is faster.  Isn't find just hashing the given path in order to look it up anyway?  Assuming the keycache is an unordered set?  If its faster, its purely becuase its avoiding the UTF8 conversion.   Its frustrating becuase in the vast majority of cases we don't actually have to lowcase or anything.  Perhaps we could use a custom hash function of some sort that doesn't involve the entire string or something

lawsonamzn: considering that m_keyCache lookup and store both depend no qhash(Qstring) which is implemented as
```
static uint hash(const QChar *p, int n)
{
    uint h = 0;
    uint g;

    while (n--) {
        h = (h << 4) + (*p++).unicode();
        if ((g = (h & 0xf0000000)) != 0)
            h ^= g >> 23;
        h &= ~g;
    }
    return h;
}
```

lawsonamzn: perhaps its purely avoiding the tolower.

lawsonamzn (approved):

Code/Tools/AssetProcessor/native/AssetManager/FileStateCache.cpp
amzn-mike: I believe it is the tolower mostly, but the qt clean function was also showing up in the profiler (and it is actually doing something, I saw paths that included .. in them).  I didn't dig into qt code to see what its doing under the hood though.  I also tried using AZ::IO::Path but that was surprisingly more expensive to use as a key compared to QString

lawsonamzn: Yeah I'll take a 2x speedup anyday.  For future there might be additional speedup opportunities here, by using a different hash function or something like that.

PR #15721, opened 2023-04-16
feature: move random color to AnimGraph

pollend:
## What does this PR do?

The colors for the animation graph are only relevant to the graph elements. I moved it under AnimGraph as a static method. I would like to remove RGBColor from EmotionFX since there is a lot of overlap with this and the existing implementation under AZ::Color.


Gems/EMotionFX/Code/EMotionFX/Source/AnimGraph.cpp
```diff
@@ -33,6 +33,146 @@ namespace EMotionFX
 {
     AZ_CLASS_ALLOCATOR_IMPL(AnimGraph, AnimGraphAllocator)
 
+
+    namespace internal
+    {
+        static constexpr AZStd::array GraphNodeColors = { AZ::u32{ 0xFF000080 },
```
gadams3, line 39: I'm assuming that this array of colors was preselected and predefined elsewhere instead of actually being randomly generated generated?

gadams3 (approved):

Gems/EMotionFX/Code/EMotionFX/Source/AnimGraph.cpp
cgalvan: Yeah I had the same question at first, then I realized it was copied from the `MCore::s_colorTable` which is why I suggested it be removed as well so it's more clear where that came from.

PR #15722, opened 2023-04-17
Enable instanced draw calls in the MeshFeatureProcessor

amzn-tommy:
## What does this PR do?

This PR creates actual instanced draw calls whenever r_meshInstancingEnabled = true. It does a bucket sort that first puts all of the instance data for all the visible objects into buckets, where each bucket is a page of instance groups. It then sorts the buckets in parallel, and then creates the instanced draw calls for each instance group in each bucket while inserting the sorted objectId's into the larger per-view instance buffer that represents all the visible objects in the view for the current frame.

There is also a debug cvar which is an easy way to force each mesh into a unique instance group, taking the 10k object levels and treating them as if there were 10k unique objects. This is helpful for testing the performance when there are mostly/only unique objects in the scene.

## How was this PR tested?

Editor using the 10kEntity and 10kVegInstances levels in Automated testing, plus ASV with both vulkan and dx12 and device validation enabled.


Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
```diff
@@ -1363,6 +1716,73 @@ namespace AZ
...
+
+        static MeshInstancingSupport CanSupportInstancing(
+            Data::Instance<RPI::Material> material, bool useForwardPassIbleSpecular, const RHI::DrawListTag& transparentDrawListTag)
+        {
+            MeshInstancingSupport result;
+            if (useForwardPassIbleSpecular)
+            {
+                // Forward pass ibl specular uses the ObjectSrg to set the closest reflection probe data
+                // Since all instances from a single instanced draw call share a single ObjectSrg, this
+                // will not work with instancing unless they happen to all share the same closes probe.
```
moudgils, line 1733: I think this will disable instancing for low end devices that want IBL.  Is the implication here we can pass reflection probe data as part of instancing buffer?

Gems/Atom/Feature/Common/Code/Source/Mesh/MeshInstanceGroupList.h
```diff
@@ -54,7 +63,7 @@ namespace AZ::Render
     public:
         using WeakHandle = StableDynamicArrayWeakHandle<MeshInstanceGroupData>;
         using OwningHandle = StableDynamicArrayHandle<MeshInstanceGroupData>;
-        using ParallelRanges = StableDynamicArray<MeshInstanceGroupData>::ParallelRanges;
+        using ParallelRanges = StableDynamicArray<MeshInstanceGroupData, 4096>::ParallelRanges;
```
moudgils, line 66: Lets move 4096 to the top in a header as a const var.

Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
```diff
@@ -425,31 +435,368 @@ namespace AZ
...
+                m_perViewInstanceDataBufferHandlers.reserve(viewCount);
+                while (m_perViewInstanceDataBufferHandlers.size() < viewCount)
                 {
-                    const ModelDataInstance::InstanceGroupHandleList* instanceGroupHandles =
-                        reinterpret_cast<const ModelDataInstance::InstanceGroupHandleList*>(visibleObject.m_userData);
-                    // For the initial commit, we naively add each draw packet for each visible object to the view.
-                    // This will be replaced with generating instance buffers for each view to use with instanced draw calls.
-                    for (auto& instanceGroupHandle : *instanceGroupHandles)
+                    m_perViewInstanceDataBufferHandlers.push_back(GpuBufferHandler(desc));
+                }
```
invertednormal, line 517: Can't you just resize() here with the GpuBufferHandler(desc) as the value?

Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
```diff
@@ -1363,6 +1716,73 @@ namespace AZ
...
+                                shaderItem.GetShaderAsset()->GetDrawListName());
+                        }
+
+                        // Check to see if the shaderItem is for a transparent pass. If any of the active shader items
+                        // are for a transparent pass, we still support instancing, but we mark it as transparent so that
+                        // we can sort by reverse-depth
+                        if (drawListTag == transparentDrawListTag)
+                        {
+                            isTransparent = true;
+                            return false;
```
invertednormal, line 1774: This causes it to not look at other shaders - wouldn't it still be possible for there to be some other shader that doesn't have the shader option to support instance which would cause `shadersSupportInstance` to be false?

invertednormal (changes requested):
Looks good, although this system is getting quite complicated. It may be good to add a high level overview of how the system works in code comments at some point. 

I think we should just turn off instancing for any transparent objects for now - that will almost certainly fail in most cases since multiple groups of instanced transparent objects can't sort with each other.

Just a few minor issues, really nice work overall!

Gems/Atom/Feature/Common/Code/Include/Atom/Feature/Mesh/MeshFeatureProcessorInterface.h
```diff
@@ -40,6 +40,33 @@ namespace AZ
...
+
+        AZ_CVAR(
+            bool,
+            r_meshInstancingDebugForceUniqueObjectsForProfiling,
+            false,
+            nullptr,
+            AZ::ConsoleFunctorFlags::Null,
+            "Enable instanced draw calls in the MeshFeatureProcessor, but force one object per draw call. "
+            "This is helpful for simulating the worst case scenario for instancing for profiling performance.");
+
```
moudgils, line 69: If you define these cars using AZ_CVAR_EXTERNED in the header file, you can move this code to the .cpp file and can just remove all the code related to console->PerformCommand from the cpp file as well. I did something similar with r_gpuMarkersMergeGroups that can be used as an example

Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
```diff
@@ -1363,6 +1748,77 @@ namespace AZ
...
+                                shaderItem.GetShaderAsset()->GetDrawListName());
+                        }
+
+                        // Check to see if the shaderItem is for a transparent pass. If any of the active shader items
+                        // are for a transparent pass, we still support instancing, but we mark it as transparent so that
+                        // we can sort by reverse-depth
+                        if (drawListTag == transparentDrawListTag)
+                        {
+                            isTransparent = true;
+                            if (!r_meshInstancingEnabledForTransparentObjects)
```
moudgils, line 1806: Should this be using m_enableMeshInstancingForTransparentObjects?

moudgils (approved):
Some minor comments.

invertednormal (approved):

PR #15723, opened 2023-04-17
Build RHI tests on Linux by default

martinwinter-huawei:
## What does this PR do?

RHI tests currently build by default only on Windows, to facilitate testing on Linux as well, this flag should be on there by default as well.
The question is, if this should be enabled on all platforms (`iOS`, `Android` and `Mac`) as well or if there are any considerations as to why not?

## How was this PR tested?

This PR only enables building the RHI test suite on Linux by default.


rgba16f (approved):

invertednormal (approved):

invertednormal:
@moudgils for question about if it makes sense to enable this on more platforms

moudgils (approved):
Yes I agree. This was probably just an oversight for Linux and Mac. Mobile however may require some work as I am not sure we are able to run unit tests fo those platforms.

Issue #15260, opened 2023-03-22
Prefabs and entities can be reparented to other prefabs using Transform component


LB-ArturZieba:
**Describe the bug**
Currently reparenting entities or prefabs into another prefab should not be possible due to https://github.com/o3de/o3de/pull/14411. While doing so via drag and dropping is blocked, it is still possible to do this via *Transform* > *Parent entity* and manually picking the parent. Parenting entities/prefabs that way produces multiple warning and errors (please refer to the Editor.log in the **Video and Editor.log** section for detailed information).

It should be noted that parenting prefab to itself this way produces an assert (but is still blocked with the cyclical dependency popup):
File:
```
/home/gpc-610/builds/o3de/Code/Framework/AzCore/./AzCore/EBus/Internal/CallstackEntry.h
```
Condition:
```
(140659235214912) 'AZ::Internal::CallstackEntry<AzToolsFramework::ToolsApplicationRequests, AzToolsFramework::ToolsApplicationRequests>::CallstackEntry(AZ::Internal::CallstackEntry::BusContextPtr, const typename Traits::BusIdType *) [Interface = AzToolsFramework::ToolsApplicationRequests, Traits = AzToolsFramework::ToolsApplicationRequests]'
Bus static const char *AZ::EBus<AzToolsFramework::ToolsApplicationRequests>::GetName() [Interface = AzToolsFramework::ToolsApplicationRequests, BusTraits = AzToolsFramework::ToolsApplicationRequests] has multiple threads in its callstack records. Configure MutexType on the bus, or don't send to it from multiple threads
```
Line: 86

It should be noted that this issue also occurs on the Development (73e54c6) build.

**Steps to reproduce**
1. Create two entities and create two separate prefabs from them.
2. Create another two entities.
3. Duplicate one of the prefabs created in step 1.
4. Select an entity and parent it to a prefab using the *Transform* > *Parent entity* > *Pick an object in the viewport* button.
5. Repeat step 4 but parent duplicated prefab to it's original.
6. Repeat step 4 but parent a prefab to another prefab.

**Expected behavior**
Parenting to prefabs is not possible (due to https://github.com/o3de/o3de/pull/14411).

**Actual behavior**
Parenting to prefabs using the *Transform* component is possible and doing so produces multiple warnings, errors and potentially an assert.

**Video and Editor.log**

https://user-images.githubusercontent.com/86952082/226882430-a044acf9-2d57-4134-bd85-08a8621216ca.mp4

[...]

PR #15726, opened 2023-04-17
Disallow Viewport selection and Drag and drop from creating cross prefab parent child relationships

raktalen:
## What does this PR do?

Fixes [this](https://github.com/o3de/o3de/issues/15260) issue

## How was this PR tested?

Tested in editor with many relevant use cases


lumberyard-employee-dm (approved):

forkercat (approved):
Feel like we can target the fix for stabilization/2305 with exceptions.

In the release of override mode, it would be spamming a lot of error/warning messages in the console if users perform this operation in inspector.

dtamkin1 (approved):

PR #15729, opened 2023-04-17
DPE: Fix various readonly issues in the EntityInspector

amzn-tmryan:
**Description**
Prior to this PR, the wrong instance pointer was being used to invoke/access ReadOnly attributes from inside the LegacyReflectionBridge. This meant that the DPE was attempting to invoke functions, or otherwise access members, on the member of the class instead of the class itself.

This PR changes the instance ptr being used to the parent data's instance ptr, which is the class that owns the reflected member being accessed.

**Testing**
- Verified that component readonly issues recently found in the EntityInspector are fixed
- Verified that readonly functionality in the DPE debug standalone app is still intact and correct


forkercat (approved):
Thanks for looking into this!

I verified that the crash that happens in adding PhysicsHeightfieldCollider component is also fixed.

cgalvan (approved):

alexmontAmazon (approved):

lumberyard-employee-dm (approved):

PR #15738, opened 2023-04-18
Sphrose/14992 open new asset browser

### LINKED ISSUE #14992: Asset Browser - Open New Window - Bugs found during testing
**Describe the bug**
Issue 1 - When right-clicking on a folder in the Asset Browser and clicking on "Open Secondary Asset Browser" window. It doesn't appear to be taking us to the location that we clicked on via right click. It currently looks like it just dropping user into the root directory of a project.  It's expected that the folder location that the users right-clicked on would be the content we are viewing in the new folder.

Issue 2 - Right-clicking on a specific folder in the Asset Browser was not where I expected to find the ability to open a new asset browser window. I expected to find this logic also present in the main hamburger menu of Asset Browser. This is an important feature, and having it so hidden makes it difficult to discover and use. 

Issue 3 - It appears that we have limited the number of Asset Browser window that can be open is 2 max. There doesn't appear to be any reason to limit this ability to have several windows open at the same time. We should re-consider this logic and open up unlimited number of windows.

**Steps to reproduce**
(all three issue can be located in the same place)

1.  Open the editor
2.  Make sure the Asset Browser feature flag is on
3. In list view, right-click on a folder and choose "open 2nd asset browser window" - text may be different. 
4. A new window will pop open. 
5. Notice the location of the new window doesn't match the right-clicked location of the last selected folder. 

**Expected behavior**
Issue 1 - I would expect that the folder selected would be the one that pops open for the user. Not putting the user back into the default project location

Issue 2 - From the main hamburger menu inside the Asset Browser. I would expect an option that says "Open new Window" and it would open from the default project location.

Issue 3 - I would not prevent user from only have 2 Asset Browser windows open at the same time. They should be unlimited. From the main dropdown. Yes, we have a checkbox that says is "asset browser" open. If any number of asset browser windows are open (1 or greater) we should keep this checkmark checked.

198414105-590674d3-675c-4300-af79-efa1e0c5927f

sphrose:
## What does this PR do?

This PR resolves #14992 by allowing multiple instances of the asset browser to be open. Each time "Open in other asset browser" is selected, a new asset browser window will be created. The tools menu option has been modified to show the tick if any instances are open, and selecting the option will close the first open instance. The hamburger menu in the AssetBrowser window has also been modified to add an option to create a new window.

## How was this PR tested?

Manual testing.


Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
```diff
@@ -461,6 +462,11 @@ void AzAssetBrowserWindow::CreateToolsMenu()
 
     if (ed_useWIPAssetBrowserDesign)
     {
+        auto* openNewAction = new QAction(tr("Open New Instance"), this);
```
dtamkin1, line 465: Could you rename this to `Open another Asset Browser` to keep the same naming as the option from the right-click menu?

sphrose: Done

dtamkin1 (approved):

Code/Editor/QtViewPaneManager.cpp
```diff
@@ -935,21 +935,22 @@ void QtViewPaneManager::CloseAllNonStandardPanes()
...
-    if (pane->IsVisible())
-    {
-        ClosePane(name);
-    }
-    else
+    pane = GetPane(name);
+    if (!pane)
     {
-        OpenPane(name);
+        Q_ASSERT(false);
```
alexmontAmazon, line 949: favor using AZ_Assert with a message, please :)

alexmontAmazon (approved):
approving, assuming ActionManager isn't neeeded...

Code/Editor/AzAssetBrowser/AzAssetBrowserWindow.cpp
```diff
@@ -461,6 +462,11 @@ void AzAssetBrowserWindow::CreateToolsMenu()
 
     if (ed_useWIPAssetBrowserDesign)
     {
+        auto* openNewAction = new QAction(tr("Open Another Asset Browser"), this);
```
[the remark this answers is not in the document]
AMZN-daimini: Ideally, yes, but it's a bigger change that would bloat up the scope of this PR. I'm up for advising if we have the time to convert the AssetBrowser's menus and toolbars to use the Action Manager anytime soon tho :D

Code/Editor/QtViewPaneManager.cpp
sphrose: Done

PR #15745, opened 2023-04-18
Cache parent instance DOM upon prefab instantiation

michabr:
## What does this PR do?

The PR prevents recreation of an instantiated prefab by caching the instance DOM of the parent instance.
Also container entity was being set up twice.

## How was this PR tested?

Tested in Editor.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabDomUtils.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabDomUtils.cpp
@@ -544,6 +544,47 @@ namespace AzToolsFramework
 
                 return prefabBuffer.GetString();
             }
+
+            bool AddOrUpdateNestedInstance(
+                PrefabDom& prefabDom,
+                const InstanceAlias& nestedInstanceAlias,
+                PrefabDomValueConstReference nestedInstanceDom)
+            {
+                auto instancesMember = prefabDom.FindMember(PrefabDomUtils::InstancesName);
+                if (instancesMember == prefabDom.MemberEnd())
+                {
+                    // Add an "Instances" member in the prefab DOM
+                    prefabDom.AddMember(
+                        rapidjson::StringRef(PrefabDomUtils::InstancesName), PrefabDomValue(), prefabDom.GetAllocator());
+                    instancesMember = prefabDom.FindMember(PrefabDomUtils::InstancesName);
+                }
+
+                PrefabDomValueReference instancesValue = instancesMember->value;
+                if (!instancesValue->get().IsObject())
+                {
+                    instancesValue->get().SetObject();
+                }
+
+                auto instanceMember = instancesValue->get().FindMember(rapidjson::StringRef(nestedInstanceAlias.c_str()));
+                if (instanceMember == instancesValue->get().MemberEnd())
+                {
+                    // Add a member for the nested instance in the prefab DOM
+                    instancesValue->get().AddMember(
+                        rapidjson::Value(nestedInstanceAlias.c_str(), prefabDom.GetAllocator()),
+                        nestedInstanceDom.has_value() ? rapidjson::Value(nestedInstanceDom->get(), prefabDom.GetAllocator())
+                                                      : PrefabDomValue(),
+                        prefabDom.GetAllocator());
+                    return true;
+                }
+                else if (nestedInstanceDom.has_value())
+                {
+                    // Nested instance already exists in the prefab DOM, so just update its contents
+                    instanceMember->value.CopyFrom(nestedInstanceDom->get(), prefabDom.GetAllocator());
+                    return true;
+                }
+
+                return false;
+            }
         } // namespace PrefabDomUtils
     } // namespace Prefab
 } // namespace AzToolsFramework
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabDomUtils.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabDomUtils.cpp
@@ -545,7 +545,7 @@ namespace AzToolsFramework
                 return prefabBuffer.GetString();
             }
 
-            bool AddOrUpdateNestedInstance(
+            void AddNestedInstance(
                 PrefabDom& prefabDom,
                 const InstanceAlias& nestedInstanceAlias,
                 PrefabDomValueConstReference nestedInstanceDom)
@@ -574,16 +574,13 @@ namespace AzToolsFramework
                         nestedInstanceDom.has_value() ? rapidjson::Value(nestedInstanceDom->get(), prefabDom.GetAllocator())
                                                       : PrefabDomValue(),
                         prefabDom.GetAllocator());
-                    return true;
                 }
-                else if (nestedInstanceDom.has_value())
+                else
                 {
-                    // Nested instance already exists in the prefab DOM, so just update its contents
-                    instanceMember->value.CopyFrom(nestedInstanceDom->get(), prefabDom.GetAllocator());
-                    return true;
+                    AZ_Assert(
+                        false,
+                        "PrefabDomUtils::AddNestedInstance - Nested instance already exists in prefab DOM.");
                 }
-
-                return false;
             }
         } // namespace PrefabDomUtils
     } // namespace Prefab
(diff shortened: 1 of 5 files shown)


galibzon (approved):

amzn-tmryan (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabSystemComponent.cpp
```diff
@@ -810,34 +810,8 @@ namespace AzToolsFramework
             PrefabDom& targetTemplateDom = targetTemplate.GetPrefabDom();
             auto memberFound = targetTemplateDom.FindMember(PrefabDomUtils::InstancesName);
 
-            PrefabDomValueReference instancesValue;
-            if (memberFound == targetTemplateDom.MemberEnd())
+            if (PrefabDomUtils::AddOrUpdateNestedInstance(targetTemplateDom, instanceAlias))
```
forkercat, line 813: Curious why a nullopt nestedInstance is used here

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabDomUtils.cpp
```diff
@@ -544,6 +544,47 @@ namespace AzToolsFramework
...
                 return prefabBuffer.GetString();
             }
+
+            bool AddOrUpdateNestedInstance(
+                PrefabDom& prefabDom,
+                const InstanceAlias& nestedInstanceAlias,
+                PrefabDomValueConstReference nestedInstanceDom)
+            {
+                auto instancesMember = prefabDom.FindMember(PrefabDomUtils::InstancesName);
+                if (instancesMember == prefabDom.MemberEnd())
```
raktalen, line 554: Is the fact that this code is called without there being a member that lists the instances an indicator of a bug somewhere else ?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabDomUtils.h
```diff
@@ -214,6 +214,16 @@ namespace AzToolsFramework
 
             AZStd::string PrefabDomValueToString(const PrefabDomValue& prefabDomValue);
 
+            //! Adds a nested instance to the the prefab DOM if it doesn't already exist and optionally update its contents.
+            //! @param prefabDom The prefab DOM to udpate.
+            //! @param nestedInstanceAlias The alias of the nested instance to be added or updated.
+            //! @param nestedInstanceDom An optional value to assign to the nested instance in the prefab DOM.
+            //! @return bool on whether the prefab DOM was updated.
+            bool AddOrUpdateNestedInstance(
```
raktalen, line 222: Suggestion: Could this method be split into two "Add" and "Update" methods. 
Reasoning : Right now debugging this method would involve watching out for both Add and Update cases. Making those operations distinct and separated out would lead to less cognitive load especially if you are debugging a complex prefab with multiple nested instances , some of which are being updated and others being added.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -566,6 +566,20 @@ namespace AzToolsFramework
 
                 CreateLink(instanceToCreate->get(), instanceToParentUnder->get().GetTemplateId(), undoBatch.GetUndoBatch(), AZStd::move(patch));
 
+                // Update the parent instance cache so that the new instance doesn't get recreated during propagation
+                if (PrefabDomReference cachedOwningInstanceDom = instanceToParentUnder->get().GetCachedInstanceDom();
+                    cachedOwningInstanceDom.has_value())
+                {
+                    const PrefabDom& targetTemplatePrefabDom = m_prefabSystemComponentInterface->FindTemplateDom(instanceToParentUnder->get().GetTemplateId());
```
raktalen, line 573: Can this template id ever be invalid ?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -566,6 +566,20 @@ namespace AzToolsFramework
 
                 CreateLink(instanceToCreate->get(), instanceToParentUnder->get().GetTemplateId(), undoBatch.GetUndoBatch(), AZStd::move(patch));
 
+                // Update the parent instance cache so that the new instance doesn't get recreated during propagation
+                if (PrefabDomReference cachedOwningInstanceDom = instanceToParentUnder->get().GetCachedInstanceDom();
```
raktalen, line 570: Are the else situations errors that bear being reported to the user ? (Or atleast logged for someone who is debugging)

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabDomUtils.cpp
michabr: Not a bug. The "Instances" array is only added when the first nested instance wants to be added.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
michabr: I don't think so. Other places in code that update the cached instance DOM don't require its existence. There's a flag that gets set for propagation to do selective deserialization so it's an option. @srikappa-amzn can you confirm?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
michabr: It shouldn't be. If it were, it would trigger an Assert in FindTemplateDom.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabSystemComponent.cpp
michabr: That call is in CreateLink() which updates the template further down in newLink.UpdateTarget().

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabDomUtils.h
michabr: That would require the caller knowing whether the nested instance already exists in the DOM or not... I'll give it some thought.

raktalen (approved):

alexmontAmazon (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
srikappa-amzn: The cached dom will be null until the first time it goes through deserialization. Here, we are checking upon the cached dom of the parent user wants to instantiate under. Since a valid prefab parent needs to exist and be active in the editor before they can instantiate, it must have already gone through deserialization and the cached dom must be already be non-null. If we want to add a check for that, sure, we can put in an assert here if the cached dom is null.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabPublicHandler.cpp
```diff
@@ -566,6 +566,20 @@ namespace AzToolsFramework
...
+                    cachedOwningInstanceDom.has_value())
+                {
+                    const PrefabDom& targetTemplatePrefabDom = m_prefabSystemComponentInterface->FindTemplateDom(instanceToParentUnder->get().GetTemplateId());
+                    PrefabDomPath instancePath = PrefabDomUtils::GetPrefabDomInstancePath(instanceToCreate->get().GetInstanceAlias().c_str());
+                    const PrefabDomValue* instanceValue = instancePath.Get(targetTemplatePrefabDom);
+                    if (instanceValue)
+                    {
+                        PrefabDomUtils::AddOrUpdateNestedInstance(
+                            cachedOwningInstanceDom->get(), instanceToCreate->get().GetInstanceAlias(), *instanceValue);
```
srikappa-amzn, line 579: If you are instantiating a new instance, then there should be no need to 'update the nested instance'. Instead of doing AddOrUpdateNestedInstance, I recommend changing this to simply add the nested instance dom to the cached dom.

michabr: Changed the function name to "AddNestedInstance", and assert if instance member already exists. However, with that I had to revert the changes in PrefabSystemComponent to use the new helper because the original logic there only wants to add the instance to the template DOM if it doesn't already exist. There are cases in CreatePrefab where the instance already exists in the template DOM (not the cached instance DOM).

srikappa-amzn (approved):

PR #15753, opened 2023-04-19
Directional shadow exclusion frustums

invertednormal:
## What does this PR do?

This PR allows directional cascades to avoid rendering objects that are completely covered in a smaller cascade. This is done by altering the culling system to reject items in the smaller cascade. This required changes in a few areas-
- The visibility system now allows frustum queries which include a "exclude" frustum. Any objects that intersect the "include" frustum, but are completely _contained_ in the "exclude" frustum will not be selected.
- The view now allows an optional world-to-clip exclude matrix. If it has one, it will be used by the culling system.
- The directional light feature processor will add the exclude matrix if the cvar `r_excludeItemsInSmallerShadowCascades` is turned on (this change turns it on by default)
- Shaders dealing with directional lights were updated to query multiple cascades in cases where a light passes through multiple cascades before hitting a sample. This tends to be a small portion of the frame (usually less than 10%).

While this does make the shader code slightly more complicated, it reduces draw call counts by 20-30%. I was also able to make some optimization to the directional shadow code so that it runs faster with this change even though it's dealing with multiple cascades.

This tends to benefit the 3rd and 4th cascades the most. In some testing in MultiplayerSample, I saw the cost of cascades 3 and 4 go from 0.4 and 0.5 to 0.3 and 0.1 respectively. Here's some images showing the difference in the 3rd cascade with this feature on:

Off:
image

On:
image

## How was this PR tested?

Tested in AtomSampleViewer (a few screenshots will be updated in another review in that repo). Also tested with some test levels as well as Multiplayer Sample


Code/Framework/AzFramework/AzFramework/Visibility/OctreeSystemComponent.cpp
```diff
@@ -185,6 +185,28 @@ namespace AzFramework
         }
     }
 
+    void OctreeNode::Enumerate(const AZ::Frustum& includeFrustum, const AZ::Frustum& excludeFrustum, const IVisibilityScene::EnumerateCallback& callback) const
+    {
+        if (AZ::ShapeIntersection::Overlaps(includeFrustum, m_bounds) && !AZ::ShapeIntersection::Contains(excludeFrustum, m_bounds))
```
kberg-amzn, line 190: Yup, looks good..

Gems/Atom/Feature/Common/Code/Source/CoreLights/DirectionalLightFeatureProcessor.h
```diff
@@ -254,6 +254,11 @@ namespace AZ
             ShadowProperty& GetShadowProperty(LightHandle handle) { return m_shadowProperties.GetData(handle.GetIndex()); }
 
         private:
+
+            // This is currently fixed, but could be exposed to allow for user configuration
+            // See DirectionalLightShadowCalculator.azsli : DirectionalShadowCalculator::CalculateCascadeBlendAmount()
+            static constexpr const float CascadeBlendArea = 0.015;
```
kberg-amzn, line 260: Worth being a cvar inside a cpp? That would allow runtime experimentation of different values..

kberg-amzn (approved):
What I could review looked pretty good to me. Nice job 👍

Code/Framework/AzFramework/AzFramework/Visibility/OctreeSystemComponent.cpp
```diff
@@ -185,6 +185,28 @@ namespace AzFramework
         }
     }
 
+    void OctreeNode::Enumerate(const AZ::Frustum& includeFrustum, const AZ::Frustum& excludeFrustum, const IVisibilityScene::EnumerateCallback& callback) const
```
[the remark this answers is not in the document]
mbalfour-amzn: Suggestion: And one _more_ thought to throw in the mix, maybe the API should take in a lambda for custom intersections, instead of specifically going for includeFrustum / excludeFrustum. Not sure how that would affect performance or if the benefits of the added flexibility would actually help any other use cases though.

The thought path that led me down this though was "what if somebody needs Overlaps(include) && !Overlaps(exclude)", since it would be a bummer to try and start explaining small API variations like that.

mbalfour-amzn (approved):

Gems/Atom/Feature/Common/Code/Source/CoreLights/DirectionalLightFeatureProcessor.h
invertednormal: Worth being exposed all the way up to ui, but that comment was there before I just moved it. That's a separate piece of work

Code/Framework/AzFramework/AzFramework/Visibility/OctreeSystemComponent.cpp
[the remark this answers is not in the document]
invertednormal: As to the first point, while that is a common pattern, I can't think of a case for wanting that with shape intersections, at least that hasn't come up yet. I'd expect adding a lambda + conditional return would slow down code that needs to be exceptionally fast.

For the second point, I can see maybe wanting this eventually, but you again have the performance issue to deal with - the function can't optimize as well when it doesn't know what the passed in lambda is going to do (maybe a templated lambda would solve the performance issue?) Since this is the first case that's a _little_ bit different than the other cases I don't think it's worth trying to completely generalize quite yet. I could maybe see a case for Karl's idea of a "Frustum Shell" type to decrease the little bit of code duplication but I'm not sure that's worth it either, it wouldn't really reduce the amount of total code required.

Code/Framework/AzFramework/AzFramework/Visibility/OctreeSystemComponent.cpp
invertednormal: Yeah, that's an option but in the end I don't think you'd end up with less code, just different code. It bugs me that this mostly duplicates the generic enumerate, but it's also pretty easy to understand what it's doing (you don't need to go look up what a FrustumShell is)

Gems/Atom/Feature/Common/Code/Source/CoreLights/DirectionalLightFeatureProcessor.cpp
```diff
@@ -1364,6 +1373,34 @@ namespace AZ
                         segment.m_aabb = viewAabb;
                         segment.m_view->SetCameraTransform(lightTransform);
                         segment.m_view->SetViewToClipMatrix(viewToClipMatrix);
+
+                        if (cascadeIndex > 0 && r_excludeItemsInSmallerShadowCascades)
+                        {
+                            Vector3 excludeAabbMin = previousAabbMin;
+                            Vector3 excludeAabbMax = previousAabbMax;
+
+                            if (property.m_blendBetwenCascades)
```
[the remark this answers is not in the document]
invertednormal: @rgba16f suggested that I use betwixt 😆

Gems/Atom/Feature/Common/Assets/ShaderLib/Atom/Features/Shadow/DirectionalLightShadowCalculator.azsli
```diff
@@ -316,54 +329,56 @@ float DirectionalShadowCalculator::GetVisibilityFromLightPcf()
...
+            if (occluder < ESM_CASCADE_MARGIN)
             {
-                ratio = GetVisibilityFromLightPcf();
+                const float exponent = -EsmExponentialShift * (distanceWithinCameraView - occluder);
+                float ratio = exp(exponent);
+
+                static const float pcfFallbackThreshold = 1.04;
+                if (ratio > pcfFallbackThreshold)
+                {
+                    ratio = GetVisibilityFromLightPcf();
```
moudgils, line 378: We are iterating through all the cascades in that function (i.e GetVisibilityFromLightPcf) too. Should we be passing in the index of the cascade for this use case?

moudgils (commented):
Looks good. Some minor comments.

Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
```diff
@@ -255,6 +258,12 @@ namespace AZ
                             continue;
                         }
                     }
+                    else if (worklistData->m_hasExcludeFrustum && ShapeIntersection::Classify(worklistData->m_excludeFrustum, c->m_cullData.m_boundingSphere) == IntersectResult::Interior)
+                    {
+                        // Only check the exclusion frustum against entries in nodes that are contained. In the case of shadow cascades, it's
```
amzn-tommy, line 263: Why is it unlikely? Imagine a node that is larger than main frustum, where the frustum itself is contained in the node. Then the node is not contained by the include frustum, but also any of its entries could be inside the exclude frustum. Might be good to verify that case with a unit test.

I think we need to always check the exclude frustum, if it it exists.

Gems/Atom/Feature/Common/Assets/ShaderLib/Atom/Features/Shadow/DirectionalLightShadowCalculator.azsli
```diff
@@ -84,20 +83,38 @@ class DirectionalShadowCalculator
...
+    float xyMax = shadowmapSize - PIXEL_MARGIN;
+    float xyMin = PIXEL_MARGIN;
+    minShadowCascade = -1;
+    maxShadowCascade = -1;
+
+    for (int index = 0; index < m_cascadeCount; ++index)
+    {
+        float3 shadowCoord = shadowCoords[index];
+        shadowCoord.xy *= shadowmapSize;
+        if (all(shadowCoord.xy >= xyMin) && all(shadowCoord.xy <= xyMax))
```
amzn-tommy, line 106: I can see from the calling code that in the case where this never becomes true and you're left with minShadowCascade and maxShadowCascade both = -1, it is treated as lit and you never actually read from the cascades. That's not obvious from within this function. I think Sid's suggestion of clarifying what is being checked in this if statement, plus a comment where the min/max are initialized to -1 (or maybe a comment above the function declaration itself, explaining the output of the function) will help that.

Gems/Atom/Feature/Common/Code/Source/CoreLights/DirectionalLightFeatureProcessor.cpp
```diff
@@ -1341,6 +1345,11 @@ namespace AZ
             {
                 const float invShadowmapSize = 1.0f / GetShadowmapSizeFromCameraView(handle, segmentIt.first);
 
+                Vector3 previousAabbMin;
+                Vector3 previousAabbMax;
```
amzn-tommy, line 1349: super nit: maybe initialize these with CreateZero? I can see that they are not actually used until cascadeIndex == 1, so they're getting set before they're used. But it took me a minute to realize that, and I think for a person reading this code for the first time, they'd be more trusting if this was initialized they way previousNear and previousFar are initialized.

Code/Framework/AzFramework/AzFramework/Visibility/OctreeSystemComponent.cpp
```diff
@@ -418,6 +440,12 @@ namespace AzFramework
         m_root.Enumerate(frustum, callback);
     }
 
+    void OctreeScene::Enumerate(const AZ::Frustum& includeFrustum, const AZ::Frustum& excludeFrustum, const EnumerateCallback& callback) const
```
amzn-tommy, line 443: This would be a good candidate for some unit tests.

amzn-tommy (changes requested):
Mostly minor comments. Good case for some unit tests, but I'll not block it based on that.

But I think skipping the exclude check when the parent node is not fully contained is a bug.

Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
amzn-tommy: You made a good point offline, this behavior is conservative, and would at worst issue draw calls for some things that don't really need to be drawn. I'm changing my blocking comment to a ship-it based on that.

amzn-tommy (approved):

Gems/Atom/RPI/Code/Source/RPI.Public/View.cpp
```diff
@@ -265,6 +273,20 @@ namespace AZ
             m_onWorldToClipMatrixChange.Signal(m_worldToClipMatrix);
         }
 
+        void View::SetViewToClipExcludeMatrix(const AZ::Matrix4x4* viewToClip)
+        {
+            if (viewToClip)
+            {
+                m_viewToClipExcludeMatrix.emplace(*viewToClip);
```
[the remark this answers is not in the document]
invertednormal: +1 on the argument name. Just looking at the class I didn't notice the operator =, figured emplace was the normal way, but you're right I'll just change it to =

Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
invertednormal: We chatted about this outside the PR, but just as an update to anyone else, The way I was doing it could create false positives (rendering things that didn't necessarily need to be) but wouldn't cause any visual artifacts. However, I decided to test out my theory and it turns out the testing _all_ the nodes is probably worth it since it reduced draw calls by another 10-15%.

Gems/Atom/Feature/Common/Assets/ShaderLib/Atom/Features/Shadow/DirectionalLightShadowCalculator.azsli
invertednormal: Adding a comment, I'll also move the if (m_maxCascade < 0) check up higher in `GetVisibility()` so it can early out sooner.

Gems/Atom/Feature/Common/Code/Source/CoreLights/DirectionalLightFeatureProcessor.cpp
invertednormal: 👍 No real issues doing the init since this is like once per frame max.

Code/Framework/AzFramework/AzFramework/Visibility/OctreeSystemComponent.cpp
invertednormal: True. Under the hood it's really just doing frustum checks which are already tested, but would be good to add something for this function just to be sure. I'll see how complicated that will be and either add it to this PR or do a separate one.

moudgils (approved):

Code/Framework/AzFramework/AzFramework/Visibility/OctreeSystemComponent.cpp
invertednormal: Ran into some complicated issues with unit tests, particularly around how the frustums created for the octree tests don't match how frustums are actually created in the culling system. I'll work on this separately so this pr can get in.

invertednormal: Nevermind, got a unit test working for the octree that's as good as the rest of the octree shape tests.

amzn-tommy (approved):

PR #15760, opened 2023-04-19
Prevent overrides created on procedural prefabs from becoming invalid on proc prefab regeneration

srikappa-amzn:
## What does this PR do?

This PR ensures that proc prefabs are created with entity aliases derived from the nodes names of SceneGraph of the source fbx. This makes sure that even if the proc prefab is recreated (due to hot reloading or clearing of cache), the same entity aliases are used for the template. This ensures that overrides created on top of proc prefabs in previous editor sessions to work. 

Another cool feature about this is that users can make edits in DCC tools that immediately get reflected in the o3de editor while also retaining overrides created in o3de. See below video for this

https://user-images.githubusercontent.com/82230713/233148756-6d6d3c5b-4dd3-4e88-ac47-c442e0627f85.mp4

## Pending issues
There seems to be a race condition where the mesh is added to the proc prefab instance and activated but the model is not ready at that point. So, new meshes added in DCC tools won't immediately reflect in the o3de editor the first time. They will show up if the proc prefab is hot reloaded again. The above video doesn't show this bug since it's intermittent and timing related

## How was this PR tested?

Manually.


Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabSystemScriptingHandler.cpp
```diff
@@ -20,6 +20,44 @@
...
+        containerEntity->CreateComponent<Components::EditorLockComponent>();
+        containerEntity->CreateComponent<Components::EditorVisibilityComponent>();
+        containerEntity->CreateComponent<Prefab::EditorPrefabComponent>();
+
+        {
+            auto transformComponent = containerEntity->CreateComponent<Components::TransformComponent>();
+            // Because procedural prefabs need to be deterministic we need to set the component ID to something unique and non-random
+            // The prefab that references the proc prefab will store a patch that references the transform component by it's component ID
+            // If this ID is not stable, the proc prefab will lose its position, parenting, etc data next time it is regenerated
+            auto hash = TypeHash64(reinterpret_cast<const uint8_t*>(filePath.data()), filePath.length(), AZ::HashValue64{ 0 });
```
lawsonamzn, line 43: is filepath a relative path?  It also needs to be stable between different peoples machines, too, ideally.  So people can collaborate and pass the files around.

srikappa-amzn: Yes, it's a relative path.

However, I don't think this is needed anymore. I believe this was needed previously component aliases were being set as 'Component_[%id]'. And this hashing scheme was being used make sure that the alias always remains the same. However, since we switched to using typename based aliases, the alias of the transform component now should always be "{27F1E1A1-8D9D-4C3B-BD3A-AFB9762449C0} TransformComponent"

I'll remove this part and see if this still works.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabSystemComponentInterface.h
```diff
@@ -84,6 +85,21 @@ namespace AzToolsFramework
...
+            //! Creates a prefab instance using the entities and nested instances provided.
+            //! @param entities The map of entities to add to the instance. Map ensures insertion order.
+            //! @param nestedInstances The instances to add to the newly created instance.
+            //! @param filePath The path at which the prefab should be created.
+            //! @param containerEntity The container entity for the new instance.
+            //! @param parent The parent instance of the newly created instance.
+            //! @return The new instance created.
+            virtual AZStd::unique_ptr<Instance> CreatePrefab(
+                const AZStd::map<EntityAlias, AZ::Entity*>& entities,
+                AZStd::vector<AZStd::unique_ptr<Instance>>&& nestedInstances,
```
forkercat, line 98: Is 'nestedInstances' the same as 'instancesToConsume' in the other 'CreatePrefab' function?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabSystemScriptingHandler.cpp
```diff
@@ -20,6 +20,44 @@
...
+
+        for (AZ::Entity* entity : topLevelEntities)
+        {
+            AzToolsFramework::Components::TransformComponent* transformComponent =
+                entity->FindComponent<AzToolsFramework::Components::TransformComponent>();
+
+            if (transformComponent)
+            {
+                transformComponent->SetParent(containerEntity->GetId());
+            }
```
forkercat, line 56: nit: I think this can be simplified. The 'AzToolsFramework' namespace can be omitted.

```cpp
if (auto transformComponent = entity->FindComponent<Components::TransformComponent>())
{
    transformComponent->SetParent(containerEntity->GetId());
}
```

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabSystemScriptingHandler.cpp
```diff
@@ -20,6 +20,44 @@
...
 namespace AzToolsFramework::Prefab
 {
+    static AZStd::unique_ptr<AZ::Entity> CreateContainerEntityAndParentEntities(
+        const AZStd::vector<AZ::Entity*> entities, const AZStd::string& filePath)
+    {
+        bool result = false;
+        [[maybe_unused]] AZ::EntityId commonRoot;
+        EntityList topLevelEntities;
+        AzToolsFramework::ToolsApplicationRequestBus::BroadcastResult(
+            result, &AzToolsFramework::ToolsApplicationRequestBus::Events::FindCommonRootInactive, entities, commonRoot, &topLevelEntities);
```
forkercat, line 30: Is [[maybe_unused]] still needed?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabSystemScriptingHandler.cpp
```diff
@@ -68,40 +106,44 @@ namespace AzToolsFramework::Prefab
...
+        for (const auto& [entityId, entityAlias] : entityIds)
         {
-            AzToolsFramework::Components::TransformComponent* transformComponent =
-                entity->FindComponent<AzToolsFramework::Components::TransformComponent>();
+            AZ::Entity* entity = nullptr;
+            AZ::ComponentApplicationBus::BroadcastResult(entity, &AZ::ComponentApplicationBus::Events::FindEntity, entityId);
 
-            if (transformComponent)
+            AZ_Warning(
+                "PrefabSystemComponent",
```
forkercat, line 134: Shouldn't it be "PrefabSystemScriptingHandler"? Same comment for other AZ_Warning and AZ_Error in this file.

forkercat (approved):
Looks cool! left some minor comments.

Can you also show some examples of how a node-name entity alias would look like in an override patch?

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabSystemComponentInterface.h
srikappa-amzn: Yes. I think it's a better name than 'consume'. I'll change in other places too.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabSystemScriptingHandler.cpp
srikappa-amzn: Thanks for the suggestions. But I've verified that this hashing is no longer necessary. So I've removed it. Persistent overrides are now guaranteed because we use typename based component aliases.

srikappa-amzn:
> Can you also show some examples of how a node-name entity alias would look like in an override patch?

It would look like any other string. In the example shown in the video, the entity aliases of the cubes are "Cube_001" and "Cube_002". These names are generated by the SceneGraph system when it parses the fbx data. And these names are guaranteed to be unique by the SceneGraph system. A lot of DCC tools like blender and maya doesn't allow duplicate names in the same level. But even for tools that does allow that, the SceneGraph system is capable of detecting such duplicates and avoiding name collisions.

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabSystemScriptingHandler.cpp
```diff
@@ -20,6 +19,31 @@
 
 namespace AzToolsFramework::Prefab
 {
+    static AZStd::unique_ptr<AZ::Entity> CreateContainerEntityAndParentEntities(const AZStd::vector<AZ::Entity*> entities)
+    {
+        bool result = false;
+        AZ::EntityId commonRoot;
+        EntityList topLevelEntities;
+        AzToolsFramework::ToolsApplicationRequestBus::BroadcastResult(
+            result, &AzToolsFramework::ToolsApplicationRequestBus::Events::FindCommonRootInactive, entities, commonRoot, &topLevelEntities);
```
alexmontAmazon, line 28: should you be checking or otherwise using this result?

alexmontAmazon (approved):
looks good :)

lawsonamzn (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/PrefabSystemScriptingHandler.cpp
srikappa-amzn: I believe the only value needed from this call is 'topLevelEntities' which is used below. 'result' and 'commonRoot' are not used and only provided because the ebus call expects them. This is preexisting code that worked fine till now. So imo, we can leave it as is.

srikappa-amzn:
> > Can you also show some examples of how a node-name entity alias would look like in an override patch?
> 
> It would look like any other string. In the example shown in the video, the entity aliases of the cubes are "Cube_001" and "Cube_002". These names are generated by the SceneGraph system when it parses the fbx data. And these names are guaranteed to be unique by the SceneGraph system. A lot of DCC tools like blender and maya doesn't allow duplicate names in the same level. But even for tools that does allow that, the SceneGraph system is capable of detecting such duplicates and avoiding name collisions.

This is where the name conflict resolution happens btw: https://github.com/o3de/o3de/blob/development/Code/Tools/SceneAPI/SceneBuilder/Importers/Utilities/RenamedNodesMap.cpp#L52-L66

PR #15762, opened 2023-04-20
RayTracing Callable Shader support

dmcdiarmid-ly:
Added support for callable shaders to the RayTracing PipelineState and ShaderTable.

Tested Editor DX12/Vulkan

### DIFF
--- a/Gems/Atom/RHI/Code/Include/Atom/RHI/RayTracingPipelineState.h
+++ b/Gems/Atom/RHI/Code/Include/Atom/RHI/RayTracingPipelineState.h
@@ -24,6 +24,7 @@ namespace AZ
 
             AZ::Name m_rayGenerationShaderName;
             AZ::Name m_missShaderName;
+            AZ::Name m_callableShaderName;
             AZ::Name m_closestHitShaderName;
             AZ::Name m_anyHitShaderName;
         };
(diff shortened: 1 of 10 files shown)


Gems/Atom/RHI/Code/Source/RHI/RayTracingPipelineState.cpp
```diff
@@ -69,6 +69,13 @@ namespace AZ
             return this;
         }
 
+        RayTracingPipelineStateDescriptor* RayTracingPipelineStateDescriptor::CallableShaderName(const AZ::Name& callableShaderName)
+        {
+            AZ_Assert(m_shaderLibraryBuildContext && m_hitGroupBuildContext == nullptr, "CallableShaderName can only be added to a ShaderLibrary");
```
moudgils, line 74: Do we need to check for m_hitGroupBuildContext == nullptr here?

Gems/Atom/RHI/Code/Include/Atom/RHI/RayTracingPipelineState.h
```diff
@@ -24,6 +24,7 @@ namespace AZ
 
             AZ::Name m_rayGenerationShaderName;
             AZ::Name m_missShaderName;
+            AZ::Name m_callableShaderName;
```
moudgils, line 27: Arent hit/miss/etc callable shaders in principal. What is the advantage of callable shaders? Are you expecting to use them in specific features?

moudgils (approved):
Looks good. Just some minor questions.

Gems/Atom/RHI/Code/Source/RHI/RayTracingPipelineState.cpp
dmcdiarmid-ly: Yes, since it's invalid to put CallableShader in a HitGroup.  RayGen and Miss perform the same check.

Gems/Atom/RHI/Code/Include/Atom/RHI/RayTracingPipelineState.h
dmcdiarmid-ly: It's a RT feature that allows the RayGen/ClosestHit/Miss shaders to invoke another shader that is scheduled separately by the GPU. This allows the main RT shaders to be simplified and invoke separate callable shaders to do different kinds of work, instead of putting it all in the main RT shaders and branching.

VickyAtAZ (approved):

PR #15769, opened 2023-04-20
Fix Hide Engine Assets Filter

dtamkin1:
## What does this PR do?

Fixes Asset Browser's "Hide Engine Assets" toggleable filter to hide folders inside the o3de engine path, excluding the project folder, rather than hiding everything outside of the project.

## How was this PR tested?

Locally with AutomatedTesting. Testing with Asset Gems and a project-centric build to ensure everything works.


HogJonny-AMZN (approved):

HogJonny-AMZN:
I tested this change and the alterations to filtering, allow Asset Gems to show up for browsing when the 'Hide Engine Folders' is active - my opinion is that this change is good.

AMZN-daimini (approved):

cgalvan (approved):

Issue #15759, opened 2023-04-19
Linux compilation error with Clang 16: `-Wenum-constexpr-conversion`


BytesOfPiDev:
**Describe the bug**
When compiling with Clang 16 or later, a compilation error is given due to the following line of code https://github.com/o3de/o3de/blob/c3512956fa0d14bbbb46c92c2bd55b37ee030e2d/Gems/Atom/RHI/Vulkan/Code/Source/RHI/RenderPass.cpp#L378 

`
error: integer value -1 is outside the valid range of values [0, 1073741823] for this enumeration type [-Wenum-constexpr-conversion]
`

**Assets required**
N/A

**Steps to reproduce**
1. Use a Linux system with Clang 16 or later.
2. Configure and generate a source build of the engine (or project-centric build).
3. Attempt to compile the engine.
4. See error.

**Expected behavior**
The compilation of the engine is successful.

**Actual behavior**
Compilation fails with an error. Example:

```sh

[290/5837] Building CXX object External/Vulkan-f0f9285a/Code/CMakeFiles/Atom_RHI_Vulkan.Private.Static.dir/profile/Source/RHI/RenderPass.cpp.o
FAILED: External/Vulkan-f0f9285a/Code/CMakeFiles/Atom_RHI_Vulkan.Private.Static.dir/profile/Source/RHI/RenderPass.cpp.o 
[...]

PR #15772, opened 2023-04-21
bugfix: address compiling issues for clang-16

pollend:
ref: https://github.com/o3de/o3de/issues/15759

when compiled on clang 16

image

## How was this PR tested?

compile on linux against clang-16

### DIFF
--- a/Gems/Atom/RHI/Vulkan/Code/Source/RHI/RenderPass.cpp
+++ b/Gems/Atom/RHI/Vulkan/Code/Source/RHI/RenderPass.cpp
@@ -375,7 +375,7 @@ namespace AZ
                 const RenderPass::Descriptor* m_descriptor = nullptr;
             };
 
-            template<typename T, VkStructureType type = static_cast<VkStructureType>(0xFFFFFF)>
+            template<typename T, VkStructureType type = AZStd::numeric_limits<VkStructureType>::max()>
             struct StructureTypeTraits
             {
                 static const VkStructureType struct_type = type;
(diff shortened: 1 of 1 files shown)


rgba16f (approved):

lumberyard-employee-dm (approved):
This is a 3 byte structure?

pollend:
> This is a 3 byte structure?

would it be better to use std::numeric_limit

moudgils (approved):

lumberyard-employee-dm:
> > This is a 3 byte structure?
> 
> would it be better to use std::numeric_limit

Well VkStructure is an enum which would be 4 byte `int` as the underlying type, but it doesn't have a good value to set as a default looking at the manual page for it: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkStructureType.html.

So the `AZStd::numeric_limits<VkStructure>::max()` would still try to set `0xffff'ffff` and result the compile warning.

Jackie9527 (approved):
LGTM

pollend:
> > > This is a 3 byte structure?
> > 
> > would it be better to use std::numeric_limit
> 
> Well VkStructure is an enum which would be 4 byte `int` as the underlying type, but it doesn't have a good value to set as a default looking at the manual page for it: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkStructureType.html.
> 
> So the `AZStd::numeric_limits<VkStructure>::max()` would still try to set `0xffff'ffff` and result the compile warning.

I think the value needs to be invalid.

PR #15793, opened 2023-04-25
DPE: Fix unwanted disabling of widgets

amzn-tmryan:
**Description**
Before this PR, it was possible for cached DOM attributes such as Disabled to be incorrect when a widget was pulled from the pool during a patch.

This PR refactors attribute caching in `DocumentPropertyEditor.cpp` so that we only remember layout relevant attributes and, unless default, always cache them the next time a widget is being configured here.

**Testing**
- Tested with an issue noticed in the DPE debug standalone app that caused certain controls to become enabled even though their DOM nodes contained the "Disabled" attribute

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
@@ -248,7 +248,7 @@ namespace AzToolsFramework
             {
                 size_t currentWidget = static_cast<size_t>(layoutIndex);
                 auto* myRow = GetRow();
-                AzToolsFramework::DPERowWidget::AttributeInfo* attributes = myRow->GetAttributes(currentWidget);
+                AzToolsFramework::DPERowWidget::AttributeInfo* attributes = myRow->GetCachedAttributes(currentWidget);
 
                 //! If the current widget is the first widget of a shared column, create the shared column layout and add widgets to it
                 if (sharedVectorIndex < m_sharePriorColumn.size() &&
@@ -266,7 +266,7 @@ namespace AzToolsFramework
                     while (sharedWidgetIndex < numItems)
                     {
                         currentWidget = m_sharePriorColumn[sharedVectorIndex][sharedWidgetIndex];
-                        attributes = myRow->GetAttributes(currentWidget);
+                        attributes = myRow->GetCachedAttributes(currentWidget);
                         // Save the alignment of the last widget in the shared column with an alignment attribute
                         if (attributes)
                         {
@@ -525,7 +525,7 @@ namespace AzToolsFramework
                 }
             }
         }
-        ClearAttributes();
+        ClearCachedAttributes();
         m_domOrderedChildren.clear();
         m_columnLayout->Clear();
     }
@@ -637,7 +637,7 @@ namespace AzToolsFramework
             }
             else if (auto handlerInfo = DocumentPropertyEditor::GetInfoFromWidget(childWidget); !handlerInfo.IsNull())
             {
-                RemoveAttributes(childIndex);
+                RemoveCachedAttributes(childIndex);
                 if (!newOwner)
                 {
                     DocumentPropertyEditor::ReleaseHandler(handlerInfo);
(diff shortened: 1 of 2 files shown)


amzn-tmryan:
Rebased.

alexmontAmazon (approved):
looks good, any noticeable performance change?

amzn-tmryan:
> looks good, any noticeable performance change?

Actually, it didn't look any different performance wise. Seems good on that front. I profiled with switching between the cvar adapter and the reflection adapter in the standalone app, and opening the cvar editor in the Editor proper (didn't expect to see much difference here since things hadn't been cached yet).

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
```diff
@@ -725,74 +725,63 @@ namespace AzToolsFramework
...
+        if (descriptionString != childWidget->toolTip())
         {
-            m_childIndexToAttributeInfo[domIndex] = updatedInfo;
+            setToolTip(descriptionString);
+            childWidget->setToolTip(descriptionString);
         }
+
+        bool isDisabled = AZ::Dpe::Nodes::PropertyEditor::Disabled.ExtractFromDomNode(domArray).value_or(false) ||
+            AZ::Dpe::Nodes::PropertyEditor::AncestorDisabled.ExtractFromDomNode(domArray).value_or(false);
+        childWidget->setEnabled(!isDisabled);
```
forkercat, line 778: Does this mean that for these two attributes they are always set by what we see from the node?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
```diff
@@ -804,17 +793,17 @@ namespace AzToolsFramework
 
     void DPERowWidget::RemoveAttributes(size_t domIndex)
     {
-        auto foundEntry = m_childIndexToAttributeInfo.find(domIndex);
-        if (foundEntry != m_childIndexToAttributeInfo.end())
+        auto foundEntry = m_childIndexToCachedAttributeInfo.find(domIndex);
+        if (foundEntry != m_childIndexToCachedAttributeInfo.end())
         {
-            m_childIndexToAttributeInfo.erase(foundEntry);
+            m_childIndexToCachedAttributeInfo.erase(foundEntry);
```
forkercat, line 799: nit: Should we rename the two functions to `RemoveCachedAttributes` and `ClearAttributes` for consistency?

forkercat (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
amzn-tmryan: For the Disabled attribute, that's correct. We can just set it directly each time because Qt is already checking for changes before performing the enabling/disabling stuff it does:
```
if (enable != q->testAttribute(Qt::WA_Disabled))
    return; // nothing to do
```

For the tooltip string, we don't need to store it twice (caching it) since we can just verify that the current tooltip string is different. There is some tradeoff since the DOM node stores the string as a string_view, so we have to convert it in order to compare with the tooltip.

The bug being fixed here happened because both the attribute cache and the DOM node said the widget was disabled, but the QWidget was actually enabled. So we weren't disabling the widget because our cache said that it already was (I think this used to be true). Same thing could happen with the tooltip; maybe with anything widget related.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
amzn-tmryan: Yes! Good catch. Will update those.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
forkercat: Thanks for the explanation. It is more clear to me now.

PR #15801, opened 2023-04-25
Fixed project templates

AMZN-Olex:
## What does this PR do?

Fixes project templates. The issue was that the project gem is already included and no longer needs a reference in project.json.

## How was this PR tested?

Testing cmake configure locally.
Currently testing with a full build.

### DIFF
--- a/Templates/DefaultProject/Template/project.json
+++ b/Templates/DefaultProject/Template/project.json
@@ -19,7 +19,6 @@
     ],
     "restricted": "${Name}",
     "gem_names": [
-        "${Name}",
         "Atom",
         "AudioSystem",
         "AWSCore",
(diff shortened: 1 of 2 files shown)


VickyAtAZ (approved):

AMZN-alexpete:
I think this is actually a different issue where CMake is pulling in external directories from multiple engines.  Locally when I repro this with engine-centric (doesn't repro with project centric)  I see a lot of these messages with `--log-level VERBOSE` param when configuring:

```-- Multiple paths were found for the same gem 'AWSCore'.
  Current:'D:/git/o3de-dev/Gems/AWSCore'
  New:'D:/workspace/o3de_install/Gems/AWSCore'
  ```
That second path is from an installer version of o3de.  Looks like engine-centric needs a little CMake ❤️ 👀

AMZN-Olex:
This isn't the right approach after talking with @AMZN-alexpete . Closing this PR.

Issue #15792, opened 2023-04-24
CMake mixes engine gems in engine-centric workflow when engine overridden or not registered correctly


AMZN-alexpete:
**Describe the bug**
When the user configures a project in an engine-centric workflow, if the engine is not registered correctly or the project has user engine overrides settings, CMake does not provide useful error messages or detect it is mixing up different engines.

NOTE: the same issue can occur if the engine is not registered in `o3de_manifest.json` or the user has an engine installed with the same name and that other engine is used instead.

**Steps to reproduce**
1. Install two different O3DE engines, e.g. development and release
2. Create a project with the first engine. The `engine` field in `project.json` should be set to the name of the first engine.
3. Create a `user/project.json` json file inside the project, and set the `engine_path` to the first engine's path, e.g.
```
{
    "engine_path":"c:/engine1"
}
```
4. Attempt to configure the project using the second engine, by running your cmake configure command from inside the second engine:
```
cd c:/engine2
cmake -B build -DLY_3RDPARTY_PATH="d:/git/3rdparty" -DLY_PROJECTS="c:/project"
```

**Expected behavior**
CMake detects the user is trying to configure the project with a different engine than the resolved engine path and issues an error with steps to set the correct engine to use, OR prefers the engine the user is using in an engine-centric way.

**Actual behavior**
CMake attempts to mix engine gems and issues multiple warnings about targets already existing such as:
```
CMake Error at cmake/Gems.cmake:80 (message):
  Target already exists, cannot create an alias for it:
  Gem::UiBasics.Builders
```

**Found in Branch**
development and stabilization/2305

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
51a65493f2cbc2d2f2708deb56b7cdd10645ea17

PR #15806, opened 2023-04-25
fix cmake dependency resolution using wrong engine

AMZN-alexpete:
## What does this PR do?

Fix a case where cmake configure fails because it uses the wrong engine in an 'engine-centric' workflow.
Fixes https://github.com/o3de/o3de/issues/15792

The solution is to always provide the engine path to the cmake.py dependency script so it always uses the engine CMake is using.

## How was this PR tested?

- configured a project in engine-centric and project-centric mode
- ran o3de pytests

### DIFF
--- a/cmake/Projects.cmake
+++ b/cmake/Projects.cmake
@@ -229,6 +229,9 @@ foreach(project ${LY_PROJECTS})
     file(REAL_PATH ${project} full_directory_path BASE_DIRECTORY ${CMAKE_SOURCE_DIR})
     string(SHA256 full_directory_hash ${full_directory_path})
 
+    # Set the project normalized path into the global O3DE_PROJECTS_PATHS property
+    set_property(GLOBAL APPEND PROPERTY O3DE_PROJECTS_PATHS ${full_directory_path})
+
     # Truncate the full_directory_hash down to 8 characters to avoid hitting the Windows 260 character path limit
     # when the external subdirectory contains relative paths of significant length
     string(SUBSTRING ${full_directory_hash} 0 8 full_directory_hash)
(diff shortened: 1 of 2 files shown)


lumberyard-employee-dm (approved):

lumberyard-employee-dm (commented):
`cmake_path(COMPARE` should be performing a case-insensitive compare on Windows

AMZN-alexpete:
> `cmake_path(COMPARE` should be performing a case-insensitive compare on Windows.
> 
> You can try normalizing the path before comparison: https://cmake.org/cmake/help/latest/command/cmake_path.html?highlight=cmake_path#normalization

I was hoping it would, but cmake_path `COMPARE` does a case sensitive string comparison of the root name https://cmake.org/cmake/help/latest/command/cmake_path.html#compare and in my tests the `NORMALIZE` option did not change the drive letter to be upper case like it does on Python, so the only way to adjust this was to either force the drive letter upper case in CMake, or update the drive case in Python to match what CMake uses, or use a case-insensitive check in CMake to avoid adding a duplicate path.   I did consider throwing `NORMALIZE` in a couple earlier bits of CMake code, but it looked like the paths were already normalized, and Python's resolve().as_posix() was providing normalized paths with upper case drive letters.  Fun!

AMZN-alexpete:
> Well the current logic breaks comparison on non-Windows system(LInux and Mac).
> 
> So an `if (PAL_HOST_PLATFORM_NAME STREQUAL "Windows")` check would be need to prevent those platforms from incorrectly comparing paths with different cases

That's a good point, I'll add that check.

lumberyard-employee-dm (approved):

PR #15814, opened 2023-04-26
Fixes first time Editor startup assert in Diffuse Probe gem

AMZN-Olex:
## What does this PR do?

The first time launch of a project with Diffuse Probe gem hits an assert as the required asset isn't ready yet. However it is already loaded in a deferred manner. So the assert here isn't necessary. Lowering the messaging down to a warning instead.

## How was this PR tested?

Testing locally with a brand new project with clean cache.

### DIFF
--- a/Gems/DiffuseProbeGrid/Code/Source/Render/DiffuseProbeGridFeatureProcessor.cpp
+++ b/Gems/DiffuseProbeGrid/Code/Source/Render/DiffuseProbeGridFeatureProcessor.cpp
@@ -125,7 +125,7 @@ namespace AZ
                 // load probe visualization model, the BLAS will be created in OnAssetReady()
                 m_visualizationModelAsset = AZ::RPI::AssetUtils::GetAssetByProductPath<AZ::RPI::ModelAsset>(
                     "Models/DiffuseProbeSphere.azmodel",
-                    AZ::RPI::AssetUtils::TraceLevel::Assert);
+                    AZ::RPI::AssetUtils::TraceLevel::Warning);
 
                 if (!m_visualizationModelAsset.IsReady())
                 {


AMZN-alexpete (approved):

Gems/DiffuseProbeGrid/Code/Source/Render/DiffuseProbeGridFeatureProcessor.cpp
```diff
@@ -125,7 +125,7 @@ namespace AZ
                 // load probe visualization model, the BLAS will be created in OnAssetReady()
                 m_visualizationModelAsset = AZ::RPI::AssetUtils::GetAssetByProductPath<AZ::RPI::ModelAsset>(
                     "Models/DiffuseProbeSphere.azmodel",
-                    AZ::RPI::AssetUtils::TraceLevel::Assert);
+                    AZ::RPI::AssetUtils::TraceLevel::Warning);
```
AMZN-alexpete, line 128: I'm actually wondering if it should even be a warning here? Isn't this model only used for debug visualization? If so we could move the warning to the point where the user turns on that debug visualization.  Also I'm hoping this model isn't required for a release build.

cgalvan (approved):

Gems/DiffuseProbeGrid/Code/Source/Render/DiffuseProbeGridFeatureProcessor.cpp
moudgils: Proper fix should be to not render visualization model if the asset is not loaded and just wait for the asset to be loaded.

AMZN-Olex: I'll file a GHI for the after this release, as the proper fix wouldn't meet the current bar.

moudgils (approved):

dmcdiarmid-ly (approved):

AMZN-Olex:
A follow work item filed as https://github.com/o3de/o3de/issues/15819 post 23.05 release.

Gems/DiffuseProbeGrid/Code/Source/Render/DiffuseProbeGridFeatureProcessor.cpp
AMZN-Olex: Filed https://github.com/o3de/o3de/issues/15819

Issue #13035, opened 2022-11-07
Automatically build project option in Add a remote project does not work


LB-JaroslawGaweda:
**Describe the bug**
After adding a remote project with the "Automatically build project" option enabled and pressing the Download & Build button, the Project manager does not build the project. The user still has to press the build button on the project tile.

**Steps to reproduce**
1. In Project Manager, select the Add Remote Project.
2. Input the Remote URL (e.g. https://github.com/AMZN-Phil/RemoteO3DEProject.git).
3. Check the Automatically build project option.
4. Press the Download & Build button.

**Expected behavior**
Project is downloaded and the building process is started.

**Actual behavior**
Project is downloaded but the building process is not started.

**Screenshots/Video**

https://user-images.githubusercontent.com/99716227/200333092-e3deb4a6-02cd-40d0-b050-35b51e4f7acd.mp4

**Found in Branch**
Development (181022c)

Desktop :
• Device: PC
• OS: Windows
• Version 10
• CPU Intel Core i7-9700K
• GPU Nvidia GeForce RTX 2070
• Memory 16GB

AMZN-alexpete:
Confirmed this issue still exists in latest [a65fc4e](https://github.com/o3de/o3de/commit/a65fc4e4a3dca10d84c65bbf624fbe1b89b3d1eb)

LB-ArturZieba:
This issue does not occur on GitHub builds of O3DE anymore (verified on Development (352640e)), however we have noticed that this still occurs on latest development installer (verified with https://o3debinaries.org/development/Latest/Windows/o3de_installer.exe downloaded around 8PM GMT+2 5/8/2023).

LB-MateuszSzweda:
The issue did not occur today on latest development installer (verified with https://o3debinaries.org/development/Latest/Windows/o3de_installer.exe downloaded around 8PM GMT+2 5/29/2023).

PR #15820, opened 2023-04-27
Prompt the user to build newly downloaded project

tkothadev:
## What does this PR do?

This is a single line fix to address the bug discovered in https://github.com/o3de/o3de/issues/13035.

In essence, auto-build did not occur because once the download happened, the current cache of projects did not account for the newly added repo. This requires updating the project information so that it is present before searching for projects to mark as needing to build.

## How was this PR tested?

Testing was done manually. Behavior was confirmed as shown in the following gif:
ProjectManagerAutoBuildRemote

### DIFF
--- a/Code/Tools/ProjectManager/Source/ProjectButtonWidget.h
+++ b/Code/Tools/ProjectManager/Source/ProjectButtonWidget.h
@@ -140,7 +140,7 @@ namespace O3DE::ProjectManager
         void CopyProject(const ProjectInfo& projectInfo);
         void RemoveProject(const QString& projectName);
         void DeleteProject(const QString& projectName);
-        void BuildProject(const ProjectInfo& projectInfo);
+        void BuildProject(const ProjectInfo& projectInfo, bool skipDialogBox = false);
         void OpenCMakeGUI(const ProjectInfo& projectInfo);
 
     private:
--- a/Code/Tools/ProjectManager/Source/ProjectsScreen.cpp
+++ b/Code/Tools/ProjectManager/Source/ProjectsScreen.cpp
@@ -645,7 +645,7 @@ namespace O3DE::ProjectManager
         SuggestBuildProjectMsg(projectInfo, true);
     }
 
-    void ProjectsScreen::QueueBuildProject(const ProjectInfo& projectInfo)
+    void ProjectsScreen::QueueBuildProject(const ProjectInfo& projectInfo, bool skipDialogBox)
     {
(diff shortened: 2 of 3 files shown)


AMZN-Phil:
I feel like having automatically build project selected in the add remote project dialog should cause it build without the additional popup. Is there a reason for the popup still appearing?

tkothadev:
The change I introduced only refreshes the project cache. I suspect it has to do with how the auto-build system was setup originally. Any queued builds are probably auto-triggering the popup window to be sure. 

I'm debating whether it's necessarily a bad thing or not, as I've found myself sometimes wanting to cancel builds (like if I change my mind).

cgalvan (approved):

dtamkin1 (approved):

AMZN-Phil (approved):
We should probably have UX look at this flow at some point, but this is a big improvement over the entire option being ignored 🙂

tkothadev:
Just spoke with UX. They're currently reviewing the workflow for building projects, but for the time being they would prefer to get rid of the dialog box.

tkothadev:
New changes have been added, and behavior is now updated to skip dialog boxes if remote projects are set to auto-build:
ProjectManagerAutoBuildRemoteTake2

AMZN-Phil (approved):

Code/Tools/ProjectManager/Source/ProjectButtonWidget.h
```diff
@@ -140,7 +140,7 @@ namespace O3DE::ProjectManager
         void CopyProject(const ProjectInfo& projectInfo);
         void RemoveProject(const QString& projectName);
         void DeleteProject(const QString& projectName);
-        void BuildProject(const ProjectInfo& projectInfo);
+        void BuildProject(const ProjectInfo& projectInfo, bool skipDialogBox = false);
```
AMZN-alexpete, line 143: This signal was changed but doesn't appear to be used.
```suggestion
        void BuildProject(const ProjectInfo& projectInfo);
```

AMZN-alexpete (changes requested):
I could be missing something but it looks like the BuildProject signal was changed, but the second parameter was not used.  Is it connected to QueueBuildProject somewhere or something?

Code/Tools/ProjectManager/Source/ProjectButtonWidget.h
AMZN-alexpete: Resolving because it looks like this is connected to `QueueBuildProject` and that is the only use

AMZN-alexpete (approved):

PR #15821, opened 2023-04-27
Syntax issue in shader: can't use a non-class enum as qualifier prefix.

siliconvoodoo:
## What does this PR do?

Straighten a nit point of code syntax in shaders

## How was this PR tested?

Ran the sample viewer and material editor to have the asset processor re-treat the change.


siliconvoodoo:
Demonstration of the warning issued by azslc

image

moudgils (commented):
What happens if we do keep the namespace? It seems like azslc allows it as it has been working up until now. Just wondering what can happen if someone adds similar code in future.

siliconvoodoo:
> What happens if we do keep the namespace? It seems like azslc allows it as it has been working up until now. Just wondering what can happen within AP if someone adds similar code in future. Will it let the shader compile or crash?

I think the worst that may happen is a reference miss, AZSLc could fail to recognize the seenat and if the enum is defined inside SRG scope the rename could fail, which in extreme coincidental cases could pass the DXC build if a similar name enum exists in the global scope.
Shouldn't be a big issue.
Normally the warning should appear in the asset processor, since the default warning level is 1 and this is a level 1 warning.
If it's not there, maybe there is an explicit --W0 somewhere we should think of removing.

moudgils (approved):

PR #15822, opened 2023-04-27
Add the ComputeViewNormalAndPosition helper function.

jonawals:
## What does this PR do?

This PR adds the helper function `ComputeViewNormalAndPosition`. This function is derived from the existing function from [here](https://github.com/o3de/o3de/blob/development/Gems/Atom/Feature/Common/Assets/Shaders/Shadow/FullscreenShadow.azsl#L77). The original PR https://github.com/o3de/o3de/pull/13358 suggested to reconcile these two functions into one and place in `ScreenSpaceUtil.azsli` but there is no way to reconcile both use cases into the same function.

## How was this PR tested?

Manual testing in the editor.

### DIFF
--- a/Gems/AtomLyIntegration/EditorModeFeedback/Assets/Shaders/EditorModeCommon.azsli
+++ b/Gems/AtomLyIntegration/EditorModeFeedback/Assets/Shaders/EditorModeCommon.azsli
@@ -17,6 +17,12 @@ float CalculateLinearDepth(const float zDepth)
     return abs(((ViewSrg::GetFarZTimesNearZ()) / (ViewSrg::GetFarZMinusNearZ() * zDepth - ViewSrg::GetFarZ()))); 
 }
 
+//! Returns the linear depth values of the specified non-linear depth values.
+float4 CalculateLinearDepth(const float4 zDepth)
+{
+    return abs(((ViewSrg::GetFarZTimesNearZ()) / (ViewSrg::GetFarZMinusNearZ() * zDepth - ViewSrg::GetFarZ()))); 
+}
+
 //! Returns the calculated color after blending original framebuffer color with the calculated feedback effect.
 float4 CalculateOutputColor(const float3 inColor, const float3 finalEffect, const float t)
 {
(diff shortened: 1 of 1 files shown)


Gems/AtomLyIntegration/EditorModeFeedback/Assets/Shaders/EditorModeCommon.azsli
```diff
@@ -77,3 +83,60 @@ partial ShaderResourceGroup PassSrg : SRG_PerPass_WithFallback
...
+    {
+        float3 positionUp   = ViewSrg::GetViewSpacePosition( float2(screenUV.x, screenUV.y - pixelSize.y), depthUpLeft.z);
+        float3 positionDown = ViewSrg::GetViewSpacePosition( float2(screenUV.x, screenUV.y + pixelSize.y), depthBottomRight.x);
+        float3 diffUp = positionVS - positionUp;
+        float3 diffDown = positionDown - positionVS;
+        diffY = (abs(diffUp.z) < abs(diffDown.z)) ? diffUp : diffDown;
+    }
+
+    float3 normalVS = normalize( cross(diffX, diffY) );
+
```
moudgils, line 124: Maybe I missed it but why cant this section of the code be refactored to a common function that can be called by Editor feedback shader as well as the shadow shader?

Gems/AtomLyIntegration/EditorModeFeedback/Assets/Shaders/EditorModeCommon.azsli
```diff
@@ -77,3 +83,60 @@ partial ShaderResourceGroup PassSrg : SRG_PerPass_WithFallback
         return CalculateOutputColor(inColor, finalEffect, PassSrg::CalculateFinalBlendAmount(t));
     }
 }
+
+//! Calculates the view position and normal from the fragment position.
+void ComputeViewNormalAndPosition(VSOutput IN, out float3 outPositionWS, out float3 outNormalWS)
```
moudgils, line 88: Who is using this function in Editormodefeedback shaders?

Gems/AtomLyIntegration/EditorModeFeedback/Assets/Shaders/EditorModeCommon.azsli
jonawals: The params sent by both are subtly different enough that one of the two will require clunky boilerplate prep for the params for each call. I would suggest we get these changes in and then if someone wishes to have another crack at reconciling the two implementations into one generic implementation then the door is open to do so 👍

Gems/AtomLyIntegration/EditorModeFeedback/Assets/Shaders/EditorModeCommon.azsli
jonawals: None of the existing effects, but future effects are free to do so. Particularly, this blog post references this function:

https://github.com/o3de/o3de.org/pull/2380

Gems/AtomLyIntegration/EditorModeFeedback/Assets/Shaders/EditorModeCommon.azsli
moudgils: To be honest I would be ok with getting this in now if it was blocking a feature or a bug and create a ghi to refactor later but the thing is no one is using these functions. So we are essentially just adding un-factored code in the hopes that it may be used in future. This just adds more confusion and tech debt. My recommendation here it to just reference the shadow code directly in the blog and just mention how it can be refactored and used in these shaders if an effect needs to built in future.

jonawals: In that case, I think your second suggestion is the better approach: As the function is short enough, I will put the source code directly in the blog post and reference it there 👍

jonawals: Updated blog post with this suggestion here: https://github.com/o3de/o3de.org/pull/2389

PR #15823, opened 2023-04-27
Allow to unregister a View from a pipeline

kh-huawei:
## What does this PR do?

Add `UnregisterView()`, `RemoveTransientView()` and `ResetPersistentView()` functions to the RenderPipeline - Class. 

Currently it is not possible to remove a registered (persistent) View from a pipeline, it is only possible to overwrite a registered View with a new View. This is fine with the default MainPipeline, but is needed if a pipeline has multiple persistent ViewTags (and Views), and you want to move a View with from one ViewTag to the other ViewTag. 

A call to `UnregisterView()` calls either `RemoveTransientView` or `ResetPersistentView` that view + viewTag.

`ResetPersistentView()` also calls `SceneNotification::OnRenderPipelinePersistentViewChanged()`, with the new view as a nullpointer, which might be unexpected.
As far as i can tell there is only one listener for that even in O3DE, and that doesn't use the viewPointers at all, but this needs to be considered by custom gems.

## How was this PR tested?
Create a pipeline with two cameras, and swap the cameras with each other between frames.


jhmueller-huawei (approved):

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/RenderPipeline.h
```diff
@@ -108,7 +108,11 @@ namespace AZ
             using PipelineViewMap = AZStd::unordered_map<PipelineViewTag, PipelineViews>;
             using ViewToViewTagMap = AZStd::map<const View*, PipelineViewTag>;
 
-            //! Assign a view for a PipelineViewTag used in this pipeline. 
+            void UnregisterView(ViewPtr view);
+            void RemoveTransientView(const PipelineViewTag viewId, ViewPtr view);
+            void ResetPersistentView(const PipelineViewTag viewId, ViewPtr view);
```
moudgils, line 113: Seems like these two function can be private

moudgils (approved):
Looks good other than the minor recommended change.

VickyAtAZ (commented):
The change looks fine. 
Btw, The transients views are added every frame and cleared when the frame end. Remove a transient view seems not necessary. But the code logic seems fine.

PR #5652, opened 2021-11-16
Viewport/editor mode UI back button

jonawals:
The back button appears in the top-right corner of the viewport highlight border if `CreateViewportBorder` is provided with a callback function. Once the back button is pressed, `RemoveViewportBorder` is called before calling the provided callback function.

The following use cases have been implemented for when the back button is pressed:
* **Component Mode:** Exits Component Mode.
* **Focus Mode:** Exits Focus Mode and selects the parent prefab Entity in the Entity Outliner.
* **Component Mode while in Focus Mode:** Exits Component Mode and returns to Focus Mode.

https://user-images.githubusercontent.com/77458631/142029180-e9bc3067-879e-49f0-9847-0e964829fff8.mp4

image

Manual testing performed, integration tests to come.

Issue #15652, opened 2023-04-10
Error dialog after reverting overrides on a parent


michabr:
**Describe the bug**
When deleting an entity that was added as an override, and then reverting overrides on the parent, an error dialog will appears the next time the level is loaded.

**Assets required**
None

**Steps to reproduce**
1. Open Editor
2. Instantiate a prefab that has nested entities in it
3. Delete a child entity
4. Right click on the deleted child's parent and select to revert overrides
5. Save the level and reload the level
6. Note the error dialog that appears each time the level is loaded

**Expected behavior**
No errors show up when loading the level.

**Actual behavior**
Error dialog shows up when loading the level.

**Screenshots/Video**
![ErrorDialog](https://user-images.githubusercontent.com/82236305/230957018-0a0a8af0-78c1-40bb-a7ef-b02542cf0023.png)

**Found in Branch**
2305 Stabilization.

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
Please provide the SHA/hash that identifies the latest commit from the o3de/o3de repository if the issue is reproducible in the development or other official branches. 
You can get the commit ID by running the `git rev-parse HEAD` command on your current branch.

**Desktop/Device (please complete the following information):**
 - Device: [e.g. PC, Mac, iPhone, Samsung] 
 - OS: [e.g. Windows, macOS, iOS, Android]
 - Version [e.g. 10, Monterey, Oreo]
 - CPU [e.g. Intel I9-9900k , Ryzen 5900x, ]
 - GPU [AMD 6800 XT, NVidia RTX 3090]
 - Memory [e.g. 16GB]

**Additional context**

Issue #15099, opened 2023-03-14
Prefab Overrides: Adding entity as override might lead to invalid override patch (sort order issue)


forkercat:
**Describe the bug**

feature flag: `EnableOutlinerOverrideManagement` is true

Performing add-entity as template edit and override edit might create invalid override patch.

(still building Editor... would add more details)

**Steps to reproduce**

TBA

**Expected behavior**

n/a. See description

**Actual behavior**

n/a. See description

**Screenshots/Video**

TBA

**Found in Branch**

Development (commit ID TBA)

**Desktop/Device (please complete the following information):**

* Device: PC
* OS: Windows 10
* CPU: AMD Ryzen 5950x
* GPU: NVIDIA RTX 3070
* Memory: 64 GB

**Additional context**

none

forkercat:
Focus on Level

```
Level
  | Car   (prefab)
    | Wheel
    | Entity1        <-- added as an override. There would be a sort order patch pointing to position /1 (add)
```

Sort order array: [Wheel, Entity1]

Focus on Car and add an entity

```
Level
  | Car   (prefab)
    | Wheel
    | Entity2       <-- added as a template edit
```

Sort order array: [Wheel, Entity2]

Warnings show up. If we save the file, the prefab file's sort order is corrupt.
- Car has two entities. But its sort array just has Wheel, which should be [Wheel, Entity2].
- Note that the `add/1` patch is stored in the Level prefab.

Issue #13437, opened 2022-11-29
Prefab Overrides: Revert overrides does not function on parent entities after add/delete overrides are applied


jckand-amzn:
**Describe the bug**
Revert overrides does not function on parent entities after add/delete overrides are applied. The override visualization is applied to parent entities both when a child entity is added under it, or a child entity is deleted from under it. Right-clicking the parent entity and selecting "Revert overrides" does not restore the deleted entity or revert the added entity, though the visualization icon is removed from the parent entity. This can also later lead to Console warnings if an additional override is applied:

```
[Warning] (Prefabs) - Prefab - EntityIdMapper: Entity with Id [11066013346202296961] has no registered owning instance
```

**Steps to reproduce**
1. Launch the Editor with prefab overrides enabled:
```
Editor.exe --regset=O3DE/Preferences/Prefabs/EnableOverridesUx=true
```
2. Create a new level.
3. Create an entity hierarchy in the level with a parent, child, and grandchild.
4. Create a prefab from the entity hierarchy.
5. Click the prefab drop-down arrow to expand the hierarchy.
6. Right-click the child entity and "Create entity" to perform an add entity override.
7. Right-click the child entity and "Revert overrides".
8. Right-click the grandchild entity and Delete.
9. Right-click the child entity and "Revert overrides".
10. Revert overrides on the created entity in step 6.
11. Repeat step 6.

**Expected behavior**
Overrides are properly reverted when performed on the parent entity. Operations are performed free of Console warnings.

**Actual behavior**
Reverting overrides does not work when performed on the parent entity, and can result in a Console warning once attempted, and another override is applied.

**Screenshots/Video**

https://user-images.githubusercontent.com/82226555/204602978-57aceb89-4694-4c24-8385-6bc5956fa6a2.mp4

**Found in Branch**
development

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
b6d77a9bb2f029008865035ec18669205a5c1272

jckand-amzn:
> I think we can downgrade this to major? @monroegm @jckand-amzn

I'm fine with Major for this. No especially ill effects, it's just confusing.

PR #15827, opened 2023-04-27
Skip warning of missing entity referenced from sort order component

michabr:
## What does this PR do?

Add an option to skip serialization warning when an entity reference does not match any registered entities. Skip these warnings when serializing the sort order component because undoing prefab overrides may leave stale entity Ids in the sort order array.

Fixes #5652

## How was this PR tested?

Tested in Editor using repro steps in ticket.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Entity/EditorEntitySortComponentSerializer.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Entity/EditorEntitySortComponentSerializer.cpp
@@ -6,6 +6,7 @@
  *
  */
 
+#include <AzCore/Component/EntityIdSerializer.h>
 #include <AzCore/Serialization/Json/JsonSerializationResult.h>
 #include <AzCore/std/sort.h>
 #include <AzToolsFramework/Entity/EditorEntitySortComponent.h>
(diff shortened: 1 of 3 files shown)


alexmontAmazon (approved):

forkercat (approved):
looks good:)

Code/Framework/AzToolsFramework/AzToolsFramework/Entity/EditorEntitySortComponentSerializer.cpp
```diff
@@ -122,10 +123,31 @@ namespace AzToolsFramework::Components
...
 
+            // Sort order array may contain stale entityIds that can be ignored. This is because
+            // adding/removing a child entity also affects its parent entity's sort order component.
+            // So when an entity is added/removed as a prefab override, two separate overrides are
+            // generated, one referencing the child and one referencing the parent.
+            // Reverting prefab overrides only affects one entity at a time, so reverting overrides on
+            // a parent whose child was deleted as an override would cause the deleted child's entityId
+            // to be added back to the parent's sort order array
+            AZ::JsonEntityIdSerializer::JsonEntityIdMapper** idMapper =
+                context.GetMetadata().Find<AZ::JsonEntityIdSerializer::JsonEntityIdMapper*>();
```
forkercat, line 134: Looks like we can link to some GHIs related to this for better reference:
> https://github.com/o3de/o3de/issues/15652
> https://github.com/o3de/o3de/issues/15099
> https://github.com/o3de/o3de/issues/13437

dtamkin1 (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Entity/EditorEntitySortComponentSerializer.cpp
```diff
@@ -122,10 +123,31 @@ namespace AzToolsFramework::Components
             const EntityOrderArray* defaultChildEntityOrderArray =
                 defaultsortComponentInstance ? &defaultsortComponentInstance->m_childEntityOrderArray : nullptr;
 
+            // Sort order array may contain stale entityIds that can be ignored. This is because
+            // adding/removing a child entity also affects its parent entity's sort order component.
+            // So when an entity is added/removed as a prefab override, two separate overrides are
+            // generated, one referencing the child and one referencing the parent.
+            // Reverting prefab overrides only affects one entity at a time, so reverting overrides on
+            // a parent whose child was deleted as an override would cause the deleted child's entityId
+            // to be added back to the parent's sort order array
```
srikappa-amzn, line 132: nit: Instead of giving a very specific repro case here, it's better to keep it general since the warning was previously being sent anytime it can't resolve an entity id reference. 
Something like "Since sort order component on an entity stores a vector of child entity ids in it, it's possible for the component to be left with stale entity ids when updates on child entities don't update the parent entity's sort order component. To reduce the noise in error log, we are   allowing the component to have stale entity ids."

srikappa-amzn (approved):

PR #15831, opened 2023-04-27
Multi-select in the Asset Browser

### LINKED ISSUE #15674: Asset Browser: Multi-select only functions in List view
**Describe the bug**
Multi-select only functions in List view.

**Steps to reproduce**
1. Launch the Editor.
2. In the Asset Browser, attempt to CTRL-select multiple items in Thumbnail or Table views.

**Expected behavior**
Multiple assets/folders can be selected in all views.

**Actual behavior**
Multi-select only functions in List view.

**Found in Branch**
stabilization

a23b7977e4d916486b7885fe4e2c550a62302a3a

dtamkin1:
## What does this PR do?

- Enables multi-select in the thumbnail view and table view of the Asset Browser.
- Fixes a bug where folders could be deleted if they were part of a multi-select and were moved via drag and drop.
- Fixes drag and drop to move empty folders.
- Disables move, delete, and duplicate operations if the multi-selection has a folder in it.

Fixes #15674 

## How was this PR tested?

Locally

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/AssetBrowserModel.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/AssetBrowserModel.cpp
@@ -260,34 +260,78 @@ namespace AzToolsFramework
 
                 if (Utils::FromMimeData(data, entries))
                 {
-                    for (auto entry : entries)
+                    if (entries.empty())
                     {
-                        using namespace AZ::IO;
-                        Path fromPath;
-                        Path toPath;
-                        bool isFolder{ true };
- 
-                        if (entry && (entry->RTTI_IsTypeOf(SourceAssetBrowserEntry::RTTI_Type())))
-                        {
-                            fromPath = entry->GetFullPath();
-                            PathView filename = fromPath.Filename();
-                            toPath = item->GetFullPath();
-                            toPath /= filename;
-                            isFolder = false;
-                        }
-                        else
-                        {
-                            fromPath = entry->GetFullPath() + "/*";
-                            Path filename = static_cast<Path>(entry->GetFullPath()).Filename();
-                            toPath = item->GetFullPath() + "/" + filename.c_str() + "/*";
-                        }
-                        AssetBrowserViewUtils::MoveEntry(fromPath.c_str(), toPath.c_str(), isFolder);
+                        return false;
                     }
-                    return true;
+
+                    if (entries.size() > 1)
+                    {
+                        for (auto assetEntry : entries)
+                        {
+                            if (assetEntry->GetEntryType() == AssetBrowserEntry::AssetEntryType::Folder)
+                            {
+                                return false;
+                            }
+                        }
+                    }
+                    bool isFolder = entries[0]->GetEntryType() == AssetBrowserEntry::AssetEntryType::Folder;
+                    if (isFolder && AssetBrowserViewUtils::IsEngineOrProjectFolder(entries[0]->GetFullPath()))
+                    {
+                        return false;
+                    }
+                    using namespace AzFramework::AssetSystem;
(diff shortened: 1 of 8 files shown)


Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/AssetBrowserModel.cpp
```diff
@@ -260,34 +260,78 @@ namespace AzToolsFramework
...
                         }
-                        else
+                    }
+                    bool isFolder = entries[0]->GetEntryType() == AssetBrowserEntry::AssetEntryType::Folder;
+                    if (isFolder && AssetBrowserViewUtils::IsEngineOrProjectFolder(entries[0]->GetFullPath()))
+                    {
+                        return false;
+                    }
+                    using namespace AzFramework::AssetSystem;
+                    EntryTypeFilter* foldersFilter = new EntryTypeFilter();
```
lsemp3d, line 284: I'm curious, who gains ownership of this? does Qt manage it?

dtamkin1: Sorry I overlooked this bit of code, the folder filter is used when performing a move operation so that the user can pick which folder to move an asset to, but this is a move via drag and drop so the folder filter isn't needed.

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/AssetBrowserModel.cpp
```diff
@@ -260,34 +260,78 @@ namespace AzToolsFramework
...
+                    foldersFilter->SetEntryType(AssetBrowserEntry::AssetEntryType::Folder);
+
+                    AZStd::string folderPath = item->GetFullPath();
+                    bool connectedToAssetProcessor = false;
+                    AzFramework::AssetSystemRequestBus::BroadcastResult(
+                        connectedToAssetProcessor, &AzFramework::AssetSystemRequestBus::Events::AssetProcessorIsReady);
```
cgalvan, line 286: Is it possible to get to this point without the AP being ready? Doesn't the asset browser model have to query the AP database to get its entries?

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Views/EntryDelegate.cpp
```diff
@@ -148,7 +148,6 @@ namespace AzToolsFramework
                 {
                     QIcon icon;
                     icon.addFile(path, size, QIcon::Normal, QIcon::Off);
-                    AZ_Assert(!icon.isNull(), "Asset Browser Icon not found for file '%s'", path.toUtf8().constData());
```
cgalvan, line 151: Should we make this an `AZ_Warning` at least so we can still capture if this case is happening?

cgalvan (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/AssetBrowserModel.cpp
dtamkin1: It actually is possible to get here without the AP being ready. If the AP is turned off after the AB model is already loaded, the entries will still appear and we shouldn't be able to perform any file operations on them.

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Views/EntryDelegate.cpp
dtamkin1: Sure, I can change it to an AZ_Warning. It didn't seem like the correct behavior to break if the icon isn't found, since the functionality still works but there's just a blank icon for that item.

lsemp3d (approved):

PR #15832, opened 2023-04-27
VMA (Vulkan Memory Allocation) library to RHI_Vulkan

akioCL:
## What does this PR do?

Add VMA library for managing all GPU memory allocations on RHI_Vulkan. This means that that all memory allocations now goes through VMA. This includes buffers, images, transient and streaming images. There's no more direct memory allocations using Vulkan API.
Consolidate all memory allocations through a per device VMA allocator. This means that device memory is shared among all resource pools (per device). There's no more paging strategy per resource pool (less fragmentation).
Remove memory allocator, buffer memory allocator, memory page allocator, buffer memory page allocator classes since they are not needed anymore.
Remove Vulkan BufferPoolDescriptor, ImagePoolDescriptor since they are not needed anymore (no more paging).
Refactor NullDescriptorManager to use resource pools instead of direct vulkan objects to help with code maintainability (less duplicated code).
Add swizziling support for Image Views on Vulkan (needed for the NullDescriptorManager).

## How was this PR tested?

Run ASV and MPS on Vulkan.
Run ASV on Android.

### DIFF
--- a/Gems/Atom/RHI/Vulkan/Code/Source/RHI/Device.cpp
+++ b/Gems/Atom/RHI/Vulkan/Code/Source/RHI/Device.cpp
@@ -342,10 +342,7 @@ namespace AZ
             physicalDevice.LoadSupportedFeatures(m_context);
 
             RHI::ResultCode resultCode = InitVmaAllocator(physicalDeviceBase);
-            if (resultCode != RHI::ResultCode::Success)
-            {
-                return resultCode;
-            }
+            return resultCode;
         }
 
         RHI::ResultCode Device::InitInternalBindlessSrg(const AZ::RHI::BindlessSrgDescriptor& bindlessSrgDesc)


Gems/Atom/RHI/Vulkan/Code/Source/RHI/Image.cpp
```diff
@@ -761,118 +778,89 @@ namespace AZ
...
-                device.GetContext().CreateImage(device.GetNativeDevice(), &createInfo, VkSystemAllocator::Get(), &m_vkImage);
-            AssertSuccess(result);
 
+            ImageCreateInfo createInfo = device.BuildImageCreateInfo(descriptor);
+            m_usageFlags = createInfo.m_vkCreateInfo.usage;
+
+            VkResult result = VK_SUCCESS;
+            if (memoryView)
+            {
+                result = vmaCreateAliasingImage2(
```
moudgils, line 819: Add a comment how this api is different from vmaCreateImage

Gems/Atom/RHI/Vulkan/Code/Source/RHI/MemoryTypeView.h
```diff
@@ -60,60 +54,66 @@ namespace AZ
...
-        size_t MemoryTypeView<T>::GetSize() const
+        inline size_t MemoryTypeView<T>::GetOffset() const
         {
-            return m_memoryAllocation.m_size;
+            return m_offset;
         }
 
         template<typename T>
-        size_t MemoryTypeView<T>::GetAlignment() const
+        size_t MemoryTypeView<T>::GetSize() const
```
moudgils, line 102: inline other methods here?

Gems/Atom/RHI/Vulkan/Code/Source/RHI/NullDescriptorManager.cpp
```diff
@@ -201,103 +194,48 @@ namespace AZ
...
+                desc.m_image = Image::Create();
+                RHI::ImageInitRequest request(*desc.m_image, desc.m_descriptor);
+                result = m_imagePool->InitImage(request);
+                RETURN_RESULT_IF_UNSUCCESSFUL(result);
+                desc.m_image->SetLayout(desc.m_layout);
+
+                desc.m_imageView = ImageView::Create();
+                ImageViewDescriptor viewDesc;
+                viewDesc.m_aspectFlags = RHI::ImageAspectFlags::Color;
+                viewDesc.m_componentMapping = ImageComponentMapping(
```
moudgils, line 210: Can you add a comment why we need swizzling here. Thanks

Gems/Atom/RHI/Vulkan/Code/Source/RHI/StreamingImagePool.cpp
```diff
@@ -59,27 +87,68 @@ namespace AZ
                 }
             }
 
-            outHeapTiles = m_tileAllocator.Allocate(blockCount);
-            if (outHeapTiles.size() == 0)
+            AZStd::vector<VmaAllocation> allocations(blockCount);
+            VmaAllocationCreateInfo allocCreateInfo = GetVmaAllocationCreateInfo(RHI::HeapMemoryLevel::Device);
+
+            auto& device = static_cast<Device&>(GetDevice());
+            VkResult vkResult = vmaAllocateMemoryPages(
```
moudgils, line 94: is this function made for allocating sparse blocks and calling it pages? This needs comments as to what it is doing under the hood.

Gems/Atom/RHI/Vulkan/Code/Source/RHI/StreamingImagePool.cpp
```diff
@@ -59,27 +87,68 @@ namespace AZ
...
+
+            AssertSuccess(vkResult);
+            if (vkResult == VK_SUCCESS)
             {
-                return RHI::ResultCode::OutOfMemory;
-            }
-            else
-            {
-                return RHI::ResultCode::Success;
+                AZStd::transform(
```
moudgils, line 105: This is fine but I prefer good old fashioned loops :) Maybe add in a  quick comment what you are doing here.

Gems/Atom/RHI/Vulkan/Code/Source/RHI/Vulkan.cpp
```diff
@@ -6,11 +6,11 @@
  *
  */
 
-#include <stdarg.h>
-#include <algorithm>
-#include <Atom/RHI.Reflect/Bits.h>
-#include <Atom/RHI.Reflect/Limits.h>
-#include <Atom/RHI.Reflect/AttachmentEnums.h>
+#define VMA_IMPLEMENTATION
```
moudgils, line 9: We can remove this correct?

VickyAtAZ:
So the VMA’s block allocation function has to use one allocation structure for one block for one allocation. This means we will have to use an array size of 256~1024 (depends on format) for one 4k image’s mip 0.
Compare to our old method (tile span) which it can be present with few data.
And inside the implementation of the vmaAllocateMemoryPages function it will use a for loop go through each of the allocation of the block to get their info. Inside the allocate function, it also allocates block one by one.
So I am positive the performance won’t as good as our previous implementation.
I will download the branch and test streaming image on windows vulkan and linux vulkan at the beginning of next week.

VickyAtAZ:
> > So the VMA’s block allocation function has to use one allocation structure for one block for one allocation. This means we will have to use an array size of 256~1024 (depends on format) for one 4k image’s mip 0. Compare to our old method (tile span) which it can be present with few data.
> 
> I didn't get this? Could you elaborate?

For example, if an image's mipmap requires 256 tiles (64k bytes per tile, 16M in total size), the tile allocator may only need 4 allocations within 4 heap pages (each page has size 2M). Each allocation is represented by one TileSpan (start with tile index 0 and the number of continues tiles is 64). 
With VMA, it uses one allocation data for one block. And for each block it would run one allocation internally (it can be quick, but it can't avoid a loop of 256). 

I am not against of using VMA to replace existing code. But I want to point out VMA's allocation for blocks is not very optimal

Another option is to use VMA for heap page allocation.  But use tile allocator for block sub allocator. We can still get rid to most of memory allocation code. But it may require some change with the TileAllocator.

Gems/Atom/RHI/Vulkan/Code/Source/RHI/MemoryAllocation.h
```diff
@@ -0,0 +1,69 @@
...
+
+namespace AZ
+{
+    namespace Vulkan
+    {
+        class Device;
+
+        //! Represents a VMA memory allocation.
+        //! The allocation may be part of a larger memory block (sub allocated) or be a dedicated memory block.
+        class MemoryAllocation :
```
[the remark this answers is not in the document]
moudgils: VmaAllocation sounds better

Gems/Atom/RHI/Vulkan/Code/Source/RHI/Device.cpp
```diff
@@ -1201,37 +1164,274 @@ namespace AZ
...
+            };
+
+            auto& instance = Instance::GetInstance();
+            StringList deviceExtensions = physicalDevice.GetDeviceExtensionNames();
+
+            VmaAllocatorCreateInfo allocatorInfo = {};
+            allocatorInfo.physicalDevice = physicalDevice.GetNativePhysicalDevice();
+            allocatorInfo.device = m_nativeDevice;
+            allocatorInfo.instance = instance.GetNativeInstance();
+            // 1.2 is our current version for glad function pointers. Update this value when updaring GLAD
```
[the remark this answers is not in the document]
moudgils: I meant to say there was a typo with the word "updaring" in the comment.

Gems/Atom/RHI/Vulkan/Code/Source/RHI/Device.h
```diff
@@ -204,6 +206,8 @@ namespace AZ
             CommandListAllocator m_commandListAllocator;
             SemaphoreAllocator m_semaphoreAllocator;
 
+            // New VkImageUsageFlags are query in a lazy way (only when they are not already in the map)
```
moudgils, line 209: grammar

Gems/Atom/RHI/Vulkan/Code/Source/RHI/CommandList.cpp
```diff
@@ -137,8 +138,8 @@ namespace AZ
 
                 VkBufferImageCopy copy{};
                 copy.bufferOffset = sourceBufferMemoryView->GetOffset() + descriptor.m_sourceOffset;
-                copy.bufferRowLength = descriptor.m_sourceBytesPerRow / GetFormatSize(format) * GetFormatDimensionAlignment(format);
-                copy.bufferImageHeight = descriptor.m_sourceSize.m_height;
+                copy.bufferRowLength = descriptor.m_sourceBytesPerRow / GetFormatSize(format) * formatDimensionAlignment;
+                copy.bufferImageHeight = RHI::AlignUp(descriptor.m_sourceSize.m_height, formatDimensionAlignment);
```
moudgils, line 142: Does CopyItemType::ImageToBuffer need to be fixed too?

Gems/Atom/RHI/Vulkan/Code/Source/RHI/MemoryAllocation.h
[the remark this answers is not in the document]
moudgils: Ahh. Lets go with VulkanAllocation.

moudgils (approved):

VickyAtAZ (approved):

PR #15833, opened 2023-04-28
Fix for MaterialEditor NoMSAA

dmcdiarmid-ly:
Fixed MaterialEditor support for NoMSAA
Added NoMSAA supervariants to the SSR RayTracing shaders

Tested MaterialEditor DX12/Vulkan
Tested Editor DX12/Vulkan

### DIFF
--- a/Gems/Atom/Tools/AtomToolsFramework/Code/Source/EntityPreviewViewport/EntityPreviewViewportScene.cpp
+++ b/Gems/Atom/Tools/AtomToolsFramework/Code/Source/EntityPreviewViewport/EntityPreviewViewportScene.cpp
@@ -74,6 +74,10 @@ namespace AtomToolsFramework
         m_scene->Activate();
 
         AZ::RPI::RPISystemInterface::Get()->RegisterScene(m_scene);
+
+        // set the application-wide MSAA state
+        // Note: this is done after the scene is registered in order to properly update the pipeline
+        AZ::RPI::RPISystemInterface::Get()->SetApplicationMultisampleState(m_activeRenderPipeline->GetRenderSettings().m_multisampleState);
     }
 
     EntityPreviewViewportScene::~EntityPreviewViewportScene()
@@ -143,10 +147,6 @@ namespace AtomToolsFramework
 
             m_activeRenderPipelineId = iter->first;
             m_activeRenderPipeline = iter->second;
-
-            // TODO SetApplicationMultisampleState should only be called once per application and will need to consider multiple viewports
-            // and pipelines. The default pipeline determines the initial MSAA state for the application.
-            AZ::RPI::RPISystemInterface::Get()->SetApplicationMultisampleState(m_activeRenderPipeline->GetRenderSettings().m_multisampleState);
         }
 
         return true;
--- a/Gems/Atom/Tools/AtomToolsFramework/Code/Source/EntityPreviewViewport/EntityPreviewViewportScene.cpp
+++ b/Gems/Atom/Tools/AtomToolsFramework/Code/Source/EntityPreviewViewport/EntityPreviewViewportScene.cpp
@@ -59,7 +59,7 @@ namespace AtomToolsFramework
         m_frameworkScene->SetSubsystem(m_scene);
         m_frameworkScene->SetSubsystem(m_entityContext.get());
 
-        ActivateRenderPipeline(defaultRenderPipelineAssetPath);
+
 
         // Create the BRDF texture generation pipeline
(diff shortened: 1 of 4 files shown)


Gems/Atom/Tools/AtomToolsFramework/Code/Source/EntityPreviewViewport/EntityPreviewViewportScene.cpp
```diff
@@ -74,6 +74,10 @@ namespace AtomToolsFramework
         m_scene->Activate();
 
         AZ::RPI::RPISystemInterface::Get()->RegisterScene(m_scene);
+
+        // set the application-wide MSAA state
+        // Note: this is done after the scene is registered in order to properly update the pipeline
+        AZ::RPI::RPISystemInterface::Get()->SetApplicationMultisampleState(m_activeRenderPipeline->GetRenderSettings().m_multisampleState);
```
gadams3, line 80: Shouldn't this still be called from the other function whenever the render pipeline changes at runtime?

Should the call to ActivateRenderPipeline(defaultRenderPipelineAssetPath) be moved down here instead?

moudgils (approved):

gadams3 (changes requested):
Just want to confirm the behavior change before submitting. Looking at these changes, the settings will only be applied for the default pipeline and not when changing pipelines.

Gems/Atom/Tools/AtomToolsFramework/Code/Source/EntityPreviewViewport/EntityPreviewViewportScene.cpp
dmcdiarmid-ly: Looking at this further, I think it's better to go with my original change.  SetApplicationMultisampleState should really only be called once at application startup.

gadams3: It's been a while. I don't recall specific issues related to initialization order, or if the main pipeline needed to be added before brdfTexturePipeline. 

You can change the render pipeline at runtime from a drop down on the viewport in material editor and material canvas. That invokes the activate render pipeline function and updates the application multi sample state for the new pipeline.

If the multi sample state can only be set once on initialization, there should at least be a comment explaining why it was moved out of the activate pipeline function until we can change it dynamically. 

In that case, should it be set to the value from the initial pipeline or to another value that works with all pipelines, even if that's disabled?
Or, should we remove the ability to change the pipeline dynamically in those tools? Ideally, this needs to be fully supported everywhere For visualization and look development.

gadams3: > Looking at this further, I think it's better to go with my original change. SetApplicationMultisampleState should really only be called once at application startup.

OK. But there's an expectation that we can change pipelines at runtime. How will this affect that? Will everything work with the multi sample setting from the default pipeline? Please add comments to the code explaining the change/ impact in case it's revisited. There was a todo in the original code for the same function call but related to supporting multiple viewports.

dmcdiarmid-ly: Changing the multisample state at runtime is currently not supported.  You can change pipelines, but changing the number of samples requires a restart.

moudgils: Actually when Chris added a drop down menu to switch between pipeline (in Material editor) it didn't crash when switching between pipelines with msaa and non-msaa. It would be nice to keep/maintain this functionality. Ideally we would want this functionality in the Editor as Carbonated onboards they will want to visualize the look of mobile (i.e mobile non msaa pipeline) in the editor itself. Switching pipelines sounds like a extremely useful feature.

moudgils: what happens if you leave the set msaa state as it currently was, i.e after activating a new pipeline?

dmcdiarmid-ly: I can move the ActivateRenderPipeline call for now to resolve this issue, but there's still the possibility of a TDR when changing the MSAA state between pipelines.

gadams3 (approved):

PR #15834, opened 2023-04-28
Test fixes for the `BasicEditorWorkflows_LevelEntityComponentCRUD` test

burelc-amzn:
## What does this PR do?

This PR fixes a few violations found with the address sanitizer on Linux of the `BasicEditorWorkflows_LevelEntityComponentCRUD` test.

## How was this PR tested?
By running the test in a build with the address sanitizer enabled, and fixing the violations until the test passed.

PR #15839, opened 2023-04-28
Test fixes for the `BasicEditorWorkflows_LevelEntityComponentCRUD'

VickyAtAZ:
Cherry pix memory fixes from development to stabilization
https://github.com/o3de/o3de/pull/15834


VickyAtAZ:
Stabilization is in lock down. And we don't feel this fix is blocking the release.

PR #15849, opened 2023-04-30
TIAF: Test target and changelist bug fixes

jonawals:
## What does this PR do?

This PR fixes two bugs with TIAF:

1. When a test target has coverage in the historical data (either in an s3 bucket or using local persistent storage) and is subsequently removed or no longer a member of the suites being run (for which the coverage data exists for), it was mistakenly being treated as a production target which causes the test selector to try and retrieve the test target for said coverage data, only to draw a `nullptr` due to the mistaken categorization of being a production target. With this fix, test targets that are not in the `testTargetMetaMap` (due to either being removed or no longer being part of the suite's coverage data) are simply not included in the build target list used by the dynamic dependency map.
2. When creating the changelist, the TIAF python script responsible was missing the `parents=True` flag to create any non-existent parent directories, meaning that under certain circumstances this call would fail, causing TIAF to fallback to a regular test run instead of a test impact analysis run. 

## How was this PR tested?

This PR was tested locally, as well as tested in AR to fix the PR that was being blocked by the first bug.

### DIFF
--- a/Code/Tools/TestImpactFramework/Runtime/Native/Code/Source/Target/Native/TestImpactNativeTargetListCompiler.cpp
+++ b/Code/Tools/TestImpactFramework/Runtime/Native/Code/Source/Target/Native/TestImpactNativeTargetListCompiler.cpp
@@ -25,18 +25,32 @@ namespace TestImpact
         AZStd::vector<NativeProductionTarget> productionTargets;
         AZStd::vector<NativeTestTarget> testTargets;
 
+        size_t numDiscardedTestTargets = 0;
         for (auto&& descriptor : buildTargetDescriptors)
         {
-            // If this build target has an associated test artifact then it is a test target, otherwise it is a production target
             if (auto&& testTargetMeta = testTargetMetaMap.find(descriptor.m_name); testTargetMeta != testTargetMetaMap.end())
             {
                 AZ_TestImpact_Eval(descriptor.m_type == TargetType::TestTarget, ArtifactException, "Target has associated target meta but is a production target");
                 testTargets.emplace_back(AZStd::move(descriptor), AZStd::move(testTargetMeta->second));
             }
-            else
+            else if (descriptor.m_type == TargetType::ProductionTarget)
             {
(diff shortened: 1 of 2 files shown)


spham-amzn (approved):

Code/Tools/TestImpactFramework/Runtime/Native/Code/Source/Target/Native/TestImpactNativeTargetListCompiler.cpp
```diff
@@ -25,18 +25,32 @@ namespace TestImpact
...
+                numDiscardedTestTargets++;
+            }
+        }
+
+        if (numDiscardedTestTargets)
+        {
+            AZ_Info(
+                "CompileNativeTargetLists",
+                "%zu test targets were either opted-out of TIAF or not in the suite(s) selected for this run.\n",
+                numDiscardedTestTargets);
```
lumberyard-employee-dm, line 53: Would it be better instead to append the test targets names without an associated test target meta artifacts in a string that can be printed out using AZ_Info statement.

It looks like the TestImpactTargetDescriptor.h has a name value which can be logged to notify the user of the CMake targets in question.

lumberyard-employee-dm (approved):

Code/Tools/TestImpactFramework/Runtime/Native/Code/Source/Target/Native/TestImpactNativeTargetListCompiler.cpp
spham-amzn: Probably just do an AZ_Warn on after `numDiscardedTestTargets++` instead of a concatenated string in conjuction with this message.

jonawals: I'll have an experiment and follow this up with the PR that will contain the amendments for the stack trace on crash 👍

PR #15860, opened 2023-05-02
Improve the Windows launcher window experience

mbalfour-amzn:
## What does this PR do?

This PR improves the behavior of the Game Launcher window on the Windows platform, which also makes it easier to implement a settings screen for a project that allows the user to change window resolutions and fullscreen/windowed mode.

Specifically, this PR contains the following:
* Introduces a GetMaximumClientAreaSize() method (currently only implemented for Windows) that lets a setting screen query for the largest resolution supported, so that unsupported resolutions can be hidden from the user.
* Add an OnFullScreenModeChanged() notification (currently only implemented for Windows) that notifies whenever the window mode changes from Fullscreen to Windowed and back. This lets a setting screen know how to refresh itself when the window state changes.
* Enabled DPI Scaling support for Windows. Without this, the window would get auto-scaled based on the user's Display Settings, so even if O3DE thought the window was only 1920x1080 in size, it could be 200% that size in each direction with OS-level scaling. With DPI Scaling enabled, the window is created at the exact resolution that O3DE believes it to be, regardless of the Display Settings scaling value.
* Changed the Launcher window on Windows to be "TOPMOST" instead of "TOP" when the window is active, and sets it to "NOTOPMOST" when the window isn't active (i.e. when the user alt-tabs away from the window). This lets the window draw on top of the taskbar instead of underneath it, and also consequently ensures that mouse actions and clicks stay within the window instead of redirecting to the overlapped taskbar.
* Removed a workaround from the Windows mouse code that tried to account for the taskbar overlap. It didn't quite work correctly, and it's no longer needed now that the window is topmost.
* Fixed bugs in the mouse and viewport code where a 0-size window (which occurs when pressing Windows-D to go to desktop) would cause several divide-by-0 cases that led to asserts and potentially crashes from invalid values and matrices entering multiple systems. These both now clamp results to produce valid cursor locations and window sizes.
* Removed cvar replication for streaming pool cvars. These should be settable per-client and shouldn't be driven by the server.

## How was this PR tested?

Tested with MultiplayerSample on Windows, using a settings screen that lets the user select Fullscreen/Windowed and change the resolution. Ran through multiple scenarios of changing both settings, pressing alt-tab to switch back and forth from the program, and windows-D to hide to desktop and restore.


mbalfour-amzn:
> Is there a GHI, or PR for an equivalent feature on Linux?

I've added this: https://github.com/o3de/o3de/issues/15861 . I'm not currently set up to test any of this on other platforms, so I don't know what the state of our windowing support currently is relative to what I've fixed here for Windows.

rgba16f:
> > Is there a GHI, or PR for an equivalent feature on Linux?
> 
> I've added this: #15861 . I'm not currently set up to test any of this on other platforms, so I don't know what the state of our windowing support currently is relative to what I've fixed here for Windows.

Linux NativeWindow implementation has equivalent features to Windows, and is at the same level of support. MPS supports Linux as well, so the same issues will need fixing.

mbalfour-amzn:
> > > Is there a GHI, or PR for an equivalent feature on Linux?
> > 
> > 
> > I've added this: #15861 . I'm not currently set up to test any of this on other platforms, so I don't know what the state of our windowing support currently is relative to what I've fixed here for Windows.
> 
> Linux NativeWindow implementation has equivalent features to Windows, and is at the same level of support. MPS supports Linux as well, so the same issues will need fixing.

Sorry, I might not have been specific enough. I don't know:
* if our Linux implementation is affected by a concept of "DPI scaling" that exists outside of the application
* if it has the same window layering issues (TOP vs TOPMOST) in windowed mode that Windows does, leading to the same problems where aspects of the Linux desktop draw on top of the Launcher window in an undesirable way
* if the user has the ability on Linux to press Alt-Enter to switch between fullscreen and windowed mode, requiring notifications back to the application as the only way to know that the mode has changed
* if there are any Linux-specific implementation bugs related to window sizes, layering, and notifications (I believe someone on Discord recently called out some Linux-specific issues in this area, but I don't remember the details)
* if there's an equivalent Linux API to the Windows SetWindowPos() that's also getting the wrong values passed in to it in an equivalent-but-different bug
* etc.

I have no doubt that _some_ subset of this should get implemented on Linux and other platforms though, which should be covered by the GHI.

Anyways, it's a fair callout that these fixes will be necessary for the MPS settings screen to reach feature parity between Windows and Linux.

lumberyard-employee-dm (approved):

Code/Framework/AzFramework/Platform/Windows/AzFramework/Windowing/NativeWindow_Windows.cpp
```diff
@@ -192,6 +197,58 @@ namespace AzFramework
...
+            AZ_Warning("NativeWindowImpl_Win32::GetMonitorRect", false, "Could not find any monitor.");
+            return { 0, 0, 0, 0 };
+        }
+
+        // Get the dimensions of the display device on which the window is currently displayed.
+        MONITORINFO monitorInfo;
+        memset(&monitorInfo, 0, sizeof(MONITORINFO)); // C4701 potentially uninitialized local variable 'monitorInfo' used
+        monitorInfo.cbSize = sizeof(MONITORINFO);
+        const BOOL success = monitor ? GetMonitorInfo(monitor, &monitorInfo) : FALSE;
+        if (!success)
```
AMZN-alexpete, line 215: I think you already early out if monitor is nullptr on 204
```suggestion
        if (!GetMonitorInfo(monitor, &monitorInfo))
```

Code/LauncherUnified/launcher_generator.cmake
```diff
@@ -6,6 +6,9 @@
 #
 #
 
+# This will be used to turn on "PerMonitor" DPI scaling support. (Currently there is no way in CMake to specify "PerMonitorV2")
+set(O3DE_DPI_AWARENESS "PerMonitor")
```
AMZN-alexpete, line 10: should this be a cache variable to make it easier for users to set?
```suggestion
# This will be used to turn on "PerMonitor" DPI scaling support. (Currently there is no way in CMake to specify "PerMonitorV2")
set(O3DE_DPI_AWARENESS "PerMonitor" CACHE STRING "Set the Launcher's VS_DPI_AWARE state.  Valid values are PerMonitor, ON, or OFF.")
```

AMZN-alexpete (approved):
Just some minor suggestions

Code/LauncherUnified/launcher_generator.cmake
mbalfour-amzn: I'm inclined not to do this for now, because it's a bit misleading. Our NativeWindow_Windows code is actually _expecting_ this flag to be turned on, and the flag being off was causing some bugs. I would suggest waiting until someone finds the need to fix up the NativeWindow_Windows code to detect the awareness setting and handle both cases correctly before exposing this as a user setting.

Code/Framework/AzFramework/Platform/Windows/AzFramework/Windowing/NativeWindow_Windows.cpp
mbalfour-amzn: Great suggestion, though if I don't have any other changes I'll probably leave this one as-is since I've already passed AR.

moudgils (approved):

Issue #15558, opened 2023-04-04
3/22 - DPE Inspector: Editor crash when an entity has Script Canvas component (not Lua)


jckand-amzn:
**Describe the bug**
DPE Inspector: Editor crash when an entity has Script Canvas component (not Lua).

**Assets required**
[!Important!] Please provide sample assets needed to reproduce the issue, either as an [attachment](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/attaching-files) or as a link to a public asset in Github.

**Steps to reproduce**
1. Launch the Editor with the DPE and Inspector Overrides enabled:
```
Editor.exe --ed_enableDPE=true --regset=O3DE/Preferences/Prefabs/EnableInspectorOverrideManagement=true
```
2. Create a new entity.
3. Add a Script Canvas component to the entity.

**Expected behavior**
Component is added without issue.

**Actual behavior**
Editor crashes

```
Call Stack Trace:
25) 00007FFDD5C20207 (KERNELBASE) : UnhandledExceptionFilter
24) 00007FFDD82B5530 (ntdll) : memset
23) 00007FFDD829C876 (ntdll) : _C_specific_handler
22) 00007FFDD82B241F (ntdll) : _chkstk
21) 00007FFDD82614A4 (ntdll) : RtlRaiseException
20) 00007FFDD82B0F4E (ntdll) : KiUserExceptionDispatcher
19) D:\GitHub\o3de\Code\Framework\AzToolsFramework\AzToolsFramework\UI\PropertyEditor\InstanceDataHierarchy.cpp (217) : AzToolsFramework::InstanceDataNode::GetInstance
18) D:\GitHub\o3de\Gems\ScriptCanvas\Code\Editor\View\Widgets\SourceHandlePropertyAssetCtrl.cpp (165) : ScriptCanvasEditor::SourceHandlePropertyHandler::ReadValuesIntoGUI
17) D:\GitHub\o3de\Code\Framework\AzToolsFramework\AzToolsFramework\UI\PropertyEditor\PropertyEditorAPI_Internals.h (589) : AzToolsFramework::TypedPropertyHandler_Internal<ScriptCanvas::SourceHandle,ScriptCanvasEditor::SourceHandlePropertyAssetCtrl>::ReadVal
[...]

PR #15863, opened 2023-05-02
DPE Inspector: Set up parent instance as node attribute for wrapped handler

forkercat:
## What does this PR do?

Fixes https://github.com/o3de/o3de/issues/15558

This PR fixes a crash that happens when we add a ScriptCanvas component in the DPE Entity Inspector.

The crash happens at the time when RPE property handler wrapper is trying to configure data inside the widget. The 'SourceHandle' property wants to set the edit notify target to its parent instance (i.e. a 'Configuration' instance owned by 'EditorScriptCanvasComponent'), but the parent instance that can be retrieved from the 'InstanceDataNode' turns out to be missing. See [my reply](https://github.com/o3de/o3de/issues/15558#issuecomment-1516582040) in the issue for more details.

In this PR, I added a new `ParentValue` attribute for 'PropertyEditor' node and set it up in legacy reflection bridge, so code in the RPE property handler wrapper is able to retrieve the parent value pointer and set it correctly in the node.

Note: This PR just fixes the crash to unblock add/delete component testing. There are some issues regarding the SC's functionality, but they can be addressed later.

## How was this PR tested?

Manually tested in editor

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyEditorAPI_Internals.h
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyEditorAPI_Internals.h
@@ -244,7 +244,9 @@ namespace AzToolsFramework
                     auto parentValue = PropertyEditor::ParentValue.ExtractFromDomNode(node);
                     if (parentValue.has_value())
                     {
-                        auto parentValuePtr = AZ::Dom::Utils::ValueToType<void*>(parentValue.value()).value_or(nullptr);
+                        auto parentValuePtr = AZ::Dom::Utils::ValueToTypeUnsafe<void*>(parentValue.value());
+                        AZ_Assert(parentValuePtr, "Parent instance was nullptr when attempting to add to instance list.");
+
                         m_proxyParentNode.m_instances.push_back(parentValuePtr);
 
                         // Set up the reference to parent node only if a parent value is available.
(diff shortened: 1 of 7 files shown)


cgalvan (approved):

alexmontAmazon (approved):
I think this is good, but I'm adding NVS just to cast his eyes over it...

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyEditorAPI_Internals.h
```diff
@@ -233,11 +233,26 @@ namespace AzToolsFramework
...
                 {
                     continue;
                 }
+                else if (name == PropertyEditor::ParentValue.GetName())
+                {
+                    auto parentValue = PropertyEditor::ParentValue.ExtractFromDomNode(node);
+                    if (parentValue.has_value())
+                    {
+                        auto parentValuePtr = AZ::Dom::Utils::ValueToType<void*>(parentValue.value()).value_or(nullptr);
+                        m_proxyParentNode.m_instances.push_back(parentValuePtr);
```
AMZN-alexpete, line 248: We don't need to check for `nullptr` here before `push_back` because `parentValue.has_value()` guarantees the `parentValuePtr` is not nullptr?

forkercat: I think the code that sets the ptr in `LegacyReflectionBridge` guarantees that it should not be 'nullptr'. But I will add an `AZ_Assert` here to check nullity.

dtamkin1 (approved):

PR #15880, opened 2023-05-04
Misc fixes to enable a clean Launcher shutdown without crashes

mbalfour-amzn:
## What does this PR do?

When exiting a launcher, these bugfixes enable the launcher to exit without deadlocking or crashing:
* EventSchedulerSystemComponent could sometimes have deallocation issues if it held onto event handlers defined in other DLLs that already unloaded on shutdown. By clearing out the handler structures on Deactivate(), the deallocations happen cleanly.
* The SpawnableEntitiesManager had a subtle bug in which a call to RetrieveTicket(), which occurs during game entity destruction, could potentially attempt to use a ticket that already had a DestroyTicket() call queued. At the point the ticket got used, the pointer was no longer valid and anything from crashes to an infinite loop processing the spawnable queue could occur.
* Exposed a new ProcessSpawnableQueuesUntilEmpty() API to enable the shutdown logic to fully empty the queue after level unload. Otherwise, unloading the level could add new entries to the queue that kept asset references living too long in the shutdown lifecycle.
* Changed the order of shutdown operations in Launcher.cpp from (shut down CrySystem, unload CrySystem.dll, shut down GameApplication) to (shut down CrySystem, shut down GameApplication, unload CrySystem.dll). Otherwise, if any EBus contexts were created in the CrySystem.dll memory space, attempts to disconnect from that bus would crash during DLL unloads.
* Added "disconnect" and "unload level" to the start of the shutdown operations inside of CrySystem. This is unfortunate placement that can be improved over time, but at least for the moment this provides minimal change for maximum benefit. Disconnect needs to happen first so that the network queues clear any references to entities and assets. Then the level unload needs to happen so that all of those entities and assets can be removed. Then all the subsystems can be shut down. If the asset references survive all the way to the Spawnable system shutdown, they will cause a number of errors and asserts because the systems that handle specific asset types and know how to destroy them will no longer exist. The level unload happens inside of CrySystem instead of later, because CrySystem owns the LevelLoader system instance.
* Fixed bug with AutoGenRegistryManager that caused intermittent crashes on shutdown due to leaked descriptor pointers. The issue was that static variables (used by the registry system) could destroy themselves in any order within a DLL, and if the manager destroyed itself first, any remaining registry descriptors would never call ReleaseDescriptor(). By moving the environment variable into GetInstance() (thanks to @lumberyard-employee-dm for identifying the solution!), everything destructs deterministically in the correct order.

## How was this PR tested?

Ran the MultiplayerSample GameLauncher and ServerLauncher and verified that both exited successfully with no asserts, deadlocks, or crashes.

(Note: Debug builds assert on PopcornFX shutdown, that's logged as a GHI for PopcornFX)


Code/Legacy/CrySystem/System.cpp
```diff
@@ -289,6 +290,30 @@ void CSystem::ShutDown()
...
+    // By unloading the level before shutting down any subsystems, we can avoid the cascade of errors.
+    ILevelSystem* levelSystem = GetILevelSystem();
+    if (levelSystem)
+    {
+        levelSystem->UnloadLevel();
+        if (auto spawnableInterface = AzFramework::RootSpawnableInterface::Get(); spawnableInterface)
+        {
+            spawnableInterface->ProcessSpawnableQueueUntilEmpty();
+        }
+    }
```
lumberyard-employee-dm, line 315: Can't we have the [SpawnableLevelSystem destructor](https://github.com/o3de/o3de/blob/development/Code/Legacy/CrySystem/LevelSystem/SpawnableLevelSystem.cpp#L109-L112) do this instead?

lumberyard-employee-dm (approved):

AMZN-alexpete (approved):
wow, great detective work!

AMZN-Gene (approved):

Code/Legacy/CrySystem/System.cpp
mbalfour-amzn: That would be really nice but it looks like it would take a larger amount of system untangling / bugfixing. In trying it, I get a crash on the level unload from ILyShine being null, which appears to be because of the work it does in OnCrySystemShutdown: https://github.com/o3de/o3de/blob/73ea5cae2d6f8ca2243e786ac8c0f5414af579c1/Gems/LyShine/Code/Source/LyShineSystemComponent.cpp#L405

I'm hesitant to push the level destructor above that notification, since destroying the system _is_ a part of shutting down. It would also then cause me to fix that LyShine code where it calls GetILevelSystem()->RemoveListener() to null guard it, and might have some other knock-off effects. I suspect it will likely start causing other small pieces to unravel and require fixing too.

For now, I'd propose we leave this here as a start. 

If/when anyone feels inclined to take another pass at improving our shutdown code, hopefully we could make the disconnect logic work just through standard Multiplayer shutdown without requiring it to happen via console command here, and hopefully we could break apart these CrySystem bits a little further so that the level system, CryFont, and LyShine can all just shut down normally in their own systems without implicit ordering dependencies.

lumberyard-employee-dm: That is good start and good with me.

It still sucks that despite being a gem LyShine is still entangled with [CrySystem](https://github.com/o3de/o3de/blob/development/Gems/LyShine/Code/Source/LyShineSystemComponent.cpp#L383-L414)

cgalvan (approved):

PR #15881, opened 2023-05-04
Add API to enumerate RHI factories.

mbalfour-amzn:
## What does this PR do?

This is useful for settings screens that enable the user to select from among the available graphics APIs.

## How was this PR tested?

Locally used it with the MultiplayerSample settings screen to filter out APIs that aren't supported by the platform.


moudgils (approved):

rgba16f (approved):

Gems/Atom/RHI/Code/Include/Atom/RHI/FactoryManagerBus.h
```diff
@@ -40,6 +40,10 @@ namespace AZ
             /// e.g. whether Vulkan or D3D should activate their debug layers and to what extent
             virtual AZ::RHI::ValidationMode DetermineValidationMode() const = 0;
 
+            /// Enumerate the list of factories that exist in the Factory Manager.
+            /// @param callback - The function that will get called for each factory that exists.
+            /// The callback should return 'true' to continue enumerating or 'false' to stop enumerating.
+            virtual void EnumerateFactories(AZStd::function<bool(Factory* factory)> callback) = 0;
```
lumberyard-employee-dm, line 46: Can you make a type alias for callback to make it easier for users specify if they wantto store the type in a variable?
```suggestion
            /// Enumerate the list of factories that exist in the Factory Manager.
            /// @param callback - The function that will get called for each factory that exists.
            /// The callback should return 'true' to continue enumerating or 'false' to stop enumerating.
            using FactoryVisitCallback = AZStd::function<bool(Factory* factory);
            virtual void EnumerateFactories(FactoryVisitCallback callback) = 0;
```

lumberyard-employee-dm (approved):

PR #15886, opened 2023-05-05
Jjjoness/restructure asset browser table to list

jjjoness:
Renames the tableView as listView and ExtendedTableView to tableview in the AssetBrowser to make the code more legible. Extending the code was more complicated than necessary because the old naming was ambiguous.

## How was this PR tested?

Built code and tested the functionality of the AssetBrowser wasn't affected.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/AssetPicker/AssetPickerDialog.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/AssetPicker/AssetPickerDialog.cpp
@@ -289,8 +289,7 @@ namespace AzToolsFramework
 
         void AssetPickerDialog::UpdatePreview() const
         {
-            auto selectedAssets = m_ui->m_assetBrowserTreeViewWidget->isVisible()
-                ? m_ui->m_assetBrowserTreeViewWidget->GetSelectedAssets()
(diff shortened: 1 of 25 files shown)


Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/AssetPicker/AssetPickerDialog.cpp
```diff
@@ -291,7 +291,7 @@ namespace AzToolsFramework
         {
             auto selectedAssets = m_ui->m_assetBrowserTreeViewWidget->isVisible()
                 ? m_ui->m_assetBrowserTreeViewWidget->GetSelectedAssets()
-                : m_ui->m_assetBrowserTableViewWidget->GetSelectedAssets();
+                                                                                  : m_ui->m_assetBrowserListViewWidget->GetSelectedAssets();
```
cgalvan, line 294: Nit: Extra whitespace here

cgalvan (changes requested):
It looks like `AssetBrowserExpandedTableView` got deleted instead of moved `git mv` to `AssetBrowserListView`. This should be moved instead so we don't lose the commit history.

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/AssetPicker/AssetPickerDialog.cpp
jjjoness: REmoved.

sphrose (approved):

PR #15889, opened 2023-05-05
No cross-prefab reparenting in editor transform component

forkercat:
## What does this PR do?

Fixes https://github.com/o3de/o3de/issues/15260 issue ([previous fix](https://github.com/o3de/o3de/pull/15726) did not fix all cases)

This PR blocks users from parenting an entity or a nested prefab to an entity owned by a different prefab in TransformComponent's parent field.

Previously, there was a PR https://github.com/o3de/o3de/pull/14411 blocking this, but it was only for drag-and-drop in Entity Outliner.

**Change List:**
- Add new check in editor's transform component
- Improve the existing check in EntityOutlinerListModel (remove `InstanceEntityMapperInterface` dependency)

## How was this PR tested?

- [x] TransformComponent: Tested in editor (as shown in video)
- [x] TransformComponent: Tested reparenting for multi-selected entities. It works for selected entities that might belong to different prefabs. Blocking still take effect.
- [x] EntityOutlinerListModel: Passed reparenting unit tests

https://user-images.githubusercontent.com/11157226/236561106-70ad77bb-ad3c-459e-b555-1524c473d8a7.mp4

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/TransformComponent.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/TransformComponent.cpp
@@ -934,36 +934,45 @@ namespace AzToolsFramework
                 return AZ::Failure(AZStd::string("Trying to set an entity ID to something that isn't an entity ID."));
             }
 
-            AZ::EntityId actualValue = static_cast<AZ::EntityId>(*((AZ::EntityId*)newValue));
+            AZ::EntityId newParentId = static_cast<AZ::EntityId>(*((AZ::EntityId*)newValue));
 
-            if (!actualValue.IsValid())
+            if (!newParentId.IsValid())
             {
                 // Handled by the calling code.
                 return AZ::Success();
             }
 
-            // Only allow reparenting the selected entities if they are all under the same instance.
-            // We check the parent entity separately because it may be a container entity and
-            // container entities consider their owning instance to be the parent instance
-            EntityIdList entitiesToCheck;
-            entitiesToCheck.reserve(2);
-            entitiesToCheck.push_back(actualValue);
-            entitiesToCheck.push_back(GetEntityId());
-            const Prefab::PrefabPublicInterface* prefabPublicInterface = AZ::Interface<Prefab::PrefabPublicInterface>::Get();
-            if (prefabPublicInterface && !prefabPublicInterface->EntitiesBelongToSameInstance(entitiesToCheck))
-            {
-                return AZ::Failure(AZStd::string("An entity in one prefab cannot be a child to an entity in another prefab"));
-            }
+            AZ::EntityId selectedEntityId = GetEntityId();
 
             // Prevent setting the parent to the entity itself.
-            if (actualValue == GetEntityId())
+            if (newParentId == selectedEntityId)
             {
                 return AZ::Failure(AZStd::string("You cannot set an entity's parent to itself."));
             }
 
+            // Prevent parenting to a different owning prefab instance.
+            // Note: The logic should be the same as defined in EntityOutlinerListModel::CanReparentEntities()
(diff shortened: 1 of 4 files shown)


Code/Framework/AzToolsFramework/AzToolsFramework/UI/Outliner/EntityOutlinerListModel.cpp
```diff
@@ -853,34 +852,32 @@ namespace AzToolsFramework
...
-        AZ_Assert(prefabPublicInterface, "EntityOutlinerListModel requires a PrefabPublicInterface instance on Initialize.");
-        if (!prefabPublicInterface->EntitiesBelongToSameInstance(selectedEntityIds))
-        {
-            return false;
-        }
 
-        // Disable parenting to a different owning instance
-        auto instanceEntityMapperInterface = AZ::Interface<AzToolsFramework::Prefab::InstanceEntityMapperInterface>::Get();
-        if (instanceEntityMapperInterface)
+        if (auto prefabPublicInterface = AZ::Interface<Prefab::PrefabPublicInterface>::Get())
```
srikappa-amzn, line 856: shouldn't this be
```
if (auto prefabPublicInterface = AZ::Interface<Prefab::PrefabPublicInterface>::Get(); prefabPublicInterface != nullptr)
```
?
If it results in a nullptr, the current code will still enter the if loop.

Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/TransformComponent.cpp
```diff
@@ -934,57 +934,66 @@ namespace AzToolsFramework
...
 
             // Prevent setting the parent to the entity itself.
-            if (actualValue == GetEntityId())
+            if (newParentId == selectedEntityId)
             {
                 return AZ::Failure(AZStd::string("You cannot set an entity's parent to itself."));
             }
 
+            // Prevent parenting to a different owning prefab instance.
+            // Note: The logic should be the same as defined in EntityOutlinerListModel::CanReparentEntities()
```
srikappa-amzn, line 954: If it's the same code, then maybe it can be moved to a utils file like EditorEntityHelpers.cpp ?

srikappa-amzn (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/UI/Outliner/EntityOutlinerListModel.cpp
AMZN-daimini: While your suggestion will work, original code is also correct. When initializing a variable in an if condition, the if condition's result will be evaluated by converting the value of the variable to bool.

So in this case
```
if (auto prefabPublicInterface = AZ::Interface<Prefab::PrefabPublicInterface>::Get())
{
    // A
}
else
{
    // B
}
```
Since `prefabPublicInterface` is a raw pointer, he `B` branch will be executed if `prefabPublicInterface` is initialized to `nullptr`, while `A` will be executed if it has any other value.

Keep in mind that this check is only executed if no semicolon is added to the end. So something like
```
if (auto prefabPublicInterface = AZ::Interface<Prefab::PrefabPublicInterface>::Get(); prefabPublicInterface->IsLevelInstanceContainerEntity(entityId))
```
will crash if `prefabPublicInterface` is `nullptr`. So in this case you need to be more verbose and add the extra check
```
if (auto prefabPublicInterface = AZ::Interface<Prefab::PrefabPublicInterface>::Get(); prefabPublicInterface && prefabPublicInterface->IsLevelInstanceContainerEntity(entityId))
```

AMZN-daimini (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/TransformComponent.cpp
forkercat: Yeah I will look into that. It makes sense to gather the same code there.

PR #15890, opened 2023-05-05
HttpRequestorGem: Add Timing Data

lmbr-pip:
Add timing information to HTTP calls
Note: The API change will break existing customer code (annoying but very easy to fix)

## How was this PR tested?
* Made calls with HTTP Requestor and measured against ping. 
Example: `ping dynamodb.us-west-2.amazonaws.com`


lmbr-pip:
Will also impact the TwitchGem, AWSClientAuthTest etc if we change the calling signature here.

Breakage is very easy to fix but annoying.

Alternative would be to have an extensible return object so we could add new fields or allow users to fetch the full response etc but much more work.

mbalfour-amzn (approved):
If you leave the API as a breaking change, this should get sent out as an "Impactful Change" notification. As a potential alternative, you could add an API to the HttpRequestManager to get the last round trip time, so that way the callback signature doesn't need to change.

Gems/AWSClientAuth/Code/Source/Authentication/GoogleAuthenticationProvider.cpp
```diff
@@ -83,7 +83,8 @@ namespace AWSClientAuth
 
         HttpRequestor::HttpRequestorRequestBus::Broadcast(&HttpRequestor::HttpRequestorRequests::AddRequestWithHeadersAndBody, m_settings->m_oAuthCodeURL
             , Aws::Http::HttpMethod::HTTP_POST, headers, body
-            , [this](const Aws::Utils::Json::JsonView& jsonView, Aws::Http::HttpResponseCode responseCode)
+            , [this](const Aws::Utils::Json::JsonView& jsonView, Aws::Http::HttpResponseCode responseCode,
+                [[maybe_unused]] AZStd::chrono::milliseconds roundTripTime)
```
spham-amzn, line 87: isn't `roundTripTime` an out parameter? ie shouldn''t this be declared as pass by reference?

AMZN-Gene: It a parameter in a callback. Could make it `const milliseconds&`, but it's really just a fancy int64.

spham-amzn (approved):

AMZN-Gene:
> If you leave the API as a breaking change, this should get sent out as an "Impactful Change" notification. As a potential alternative, you could add an API to the HttpRequestManager to get the last round trip time, so that way the callback signature doesn't need to change.

Good call, 
https://github.com/o3de/o3de/issues/16260

> If you leave the API as a breaking change, this should get sent out as an "Impactful Change" notification. As a potential alternative, you could add an API to the HttpRequestManager to get the last round trip time, so that way the callback signature doesn't need to change.

:fire: Thank you! Updated to provide a new HttpRequestorRequests::LastRoundTripTime()

Gems/HttpRequestor/Code/Source/HttpRequestManager.cpp
```diff
@@ -150,7 +155,9 @@ namespace HttpRequestor
             httpRequest->SetContentLength(AZStd::to_string(httpRequestParameters.GetBodyStream()->str().length()).c_str());
         }
 
-        auto httpResponse = httpClient->MakeRequest(httpRequest);
+        AZStd::chrono::steady_clock::time_point start = AZStd::chrono::steady_clock::now();
+        const auto httpResponse = httpClient->MakeRequest(httpRequest);
+        m_lastRoundTripTime = AZStd::chrono::duration_cast<AZStd::chrono::milliseconds>(AZStd::chrono::steady_clock::now() - start);
```
mbalfour-amzn, line 160: Just double-checking, since i see a bit of threading code in this file, only one request / last round trip time happens at a time, correct? Just making sure there aren't any race conditions or anything on m_lastRoundTripTime.

Gems/HttpRequestor/Code/Source/HttpRequestManager.h
```diff
@@ -54,6 +57,7 @@ namespace HttpRequestor
         AZStd::thread                           m_thread;                           // This is the thread that will be used for all async operations
         static const char*                      s_loggingName;                      // Name to use for log messages etc...
         bool                                    m_ownsAwsNativeInitialization = false; // Whether or not this module initialized the native layer
+        AZStd::chrono::milliseconds             m_lastRoundTripTime; // The last round trip time taken to make the http request and get a response.
```
mbalfour-amzn, line 60: Should this be stored in an atomic var? It seems like it might be possible to query it from a different thread where it's getting set.

Gems/HttpRequestor/Code/Source/HttpRequestManager.cpp
mbalfour-amzn: Also, assuming this is correct, might want to consider adding a comment above the places that set m_lastRoundTripTime explicitly calling out this assumption, so that if it ever changes, the ramifications will be more obvious.

AMZN-Gene: Correct, there's a separate single thread which makes the actual HTTP requests one at a time. During each request, m_lastRoundTripTime is set and the callback is called from that thread, which will also be when lastRoundTripTime is read.
 
There's a lock in that separate thread when it's altering the request queue since that thread will be be popping off the queue potentially while component in main thread are making new request (adding onto the queue)

Gems/HttpRequestor/Code/Source/HttpRequestManager.h
AMZN-Gene: The time you'd normally query lastRoundTripTime would be from within the callback, which is in the same thread that just set lastRoundTripTime. Hmm, you're right though, I could _technically_ call GetLastRoundTripTime anytime I wanted from the main thread...

AMZN-Gene: Update! 
m_lastRoundTripTime now atomic in case someone wants to read it outside of the callback (ie: main thread)

mbalfour-amzn (approved):

mbalfour-amzn (approved):

PR #15891, opened 2023-05-05
ImGui console cleanup

kberg-amzn:
Unifies the DebugConsole and other ImGui functionality to use a singular input mapping. ImGui is now universally toggled on and off with the ~ key. Added menu options to show and hide the debug console. Debug console is not visible during PIE mode, as the editor has its own console in the editor UI.

Removes all the input and keybinding from the CryConsole, which was still active, intercepting key requests, and modifying program input state at runtime in unpredictable ways.

image


AMZN-alexpete:
What happens if the user doesn't have the IMGUI gem active in their project? - is there a way to see/use the console?  If the console depends on IMGUI then we should probably add it to the list of "required" gems in `engine.json` in the "gem_names" field: https://github.com/o3de/o3de/blob/development/engine.json#L105

lumberyard-employee-dm:
> What happens if the user doesn't have the IMGUI gem active in their project? - is there a way to see/use the console? If the console depends on IMGUI then we should probably add it to the list of "required" gems in `engine.json` in the "gem_names" field: https://github.com/o3de/o3de/blob/development/engine.json#L105

Actually that might not be so bad.
It makes an easier way for a development to prevent access to the console, by just not having the gem enabled.

So I don't think it should be made a required gem.

Gems/ImGui/Code/Include/ImGuiBus.h
```diff
@@ -10,9 +10,11 @@
 
 #include <AzCore/EBus/EBus.h>
 #include <AzCore/EBus/Event.h>
+#include <AzCore/Interface/Interface.h>
```
lumberyard-employee-dm, line 13: Is the AZ::Interface used in the this bus header?
It doesn't seem like the include is needed here.

lumberyard-employee-dm: Or did you want to make a type alias available to users for the AZ::Interface.

Such as
```c++
using ImGuiManagerInterface = AZ::Interface<IImGuiManager>;
```

kberg-amzn: The type alias is probably unnecessary for the number of expected touchpoints. I'll remove the header instead to keep things lean and simple. Thanks!

Gems/AtomLyIntegration/ImguiAtom/Code/Source/DebugConsole.cpp
```diff
@@ -27,15 +27,12 @@
...
-////////////////////////////////////////////////////////////////////////////////////////////////////
 namespace AZ
 {
-    ////////////////////////////////////////////////////////////////////////////////////////////////
-    constexpr const char* DebugConsoleInputContext = "DebugConsoleInputContext";
-    const InputChannelId ToggleDebugConsoleInputChannelId("ToggleDebugConsole");
-    const InputChannelId ThumbstickL3AndR3InputChannelId("ThumbstickL3AndR3");
+    AZ_CVAR(bool, bg_showDebugConsole, true, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Enables or disables the debug console within imGui");
+    AZ_CVAR(float, bg_defaultDebugConsoleWidth, 960.f, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "The default width for the imGui debug console");
+    AZ_CVAR(float, bg_defaultDebugConsoleHeight, 480.f, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "The default width for the imGui debug console");
```
AMZN-alexpete, line 34:
```suggestion
    AZ_CVAR(float, bg_defaultDebugConsoleHeight, 480.f, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "The default height for the imGui debug console");
```

Gems/AtomLyIntegration/ImguiAtom/Code/Source/DebugConsole.cpp
```diff
@@ -63,11 +60,38 @@ namespace AZ
...
+
+    int InputTextCallback(ImGuiInputTextCallbackData* data)
+    {
+        DebugConsole* debugConsole = static_cast<DebugConsole*>(data->UserData);
+        if (!debugConsole)
+        {
+            return 0;
+        }
+
+        switch (data->EventFlag)
```
AMZN-alexpete, line 77: Does Clang not complain about a missing default case anymore?

Gems/ImGui/Code/Include/ImGuiBus.h
```diff
@@ -94,7 +92,7 @@ namespace ImGui
         virtual void ToggleToImGuiVisibleState(DisplayState state) = 0;
         virtual void SetDpiScalingFactor(float dpiScalingFactor) = 0;
         virtual float GetDpiScalingFactor() const = 0;
-        virtual void Render() = 0;
+        virtual ImDrawData* GetImguiDrawData() = 0;
```
AMZN-alexpete, line 95: We should probably bump the ImGui `gem.json` Major version because this looks like an API breaking change.

kberg-amzn: I can easily bump the major version, currently 0.0.0 to 1.0.0. What knock-ons might I expect from that change however? I'm not too familiar with what we have in terms of GEM dependency management, nor whether a version bump like this may require dependency changes elsewhere.

Gems/AtomLyIntegration/ImguiAtom/Code/Source/DebugConsole.cpp
kberg-amzn: Nope, apparently it does not. This is code that I moved around, I didn't edit the contents of the function in any way, so it's been functional like this for some time.

lumberyard-employee-dm (approved):

moudgils (approved):
Nice Refactor.

Gems/ImGui/Code/Include/ImGuiBus.h
AMZN-alexpete: It's safe to do, and mainly important for external devs. details here https://www.o3de.org/docs/user-guide/gems/gem-versioning/

Issue #11627, opened 2022-08-29
Remove std::is_pod from AzCore


lemonade-dm:
**Is your feature request related to a problem? Please describe.**
**Describe the bug**
The std::is_pod trait is being deprecated in C++20 and in order to ease the transition to C++20, we should replace the two usage of it from [AZStdContainers.inl](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Serialization/AZStdContainers.inl#L97-L107).

**Describe the solution you'd like**
We should replace it with the [std::is_trivial](https://en.cppreference.com/w/cpp/types/is_trivial) trait and remove the [is_pod.h](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/std/typetraits/is_pod.h) from AzCore.

**Describe alternatives you've considered**
None

**Additional context**
None

PR #15892, opened 2023-05-06
Remove std::is_pod from AzCore

ajv4288:
## What does this PR do?

this fix replaces the uses of the deprecated std::is_pod trait in AZStdContainers.inl and removes the associated header file for the deprecated trait from AzCore.

This addresses issue #11627, which can be found below:
https://github.com/o3de/o3de/issues/11627


lumberyard-employee-dm (approved):

kberg-amzn (approved):

nick-l-o3de:
Going to need to also remove the file from

[AzCore/std/azstd_files.cmake

Issue #10237, opened 2022-06-21
Inconsistency with linking hyperlinks to the source code


LB-BartoszCwiek:
**Describe the bug**
Some hyperlinks have provided URL with www domain, while others do not. 
There is also a related documentation issue already reported: https://github.com/o3de/o3de.org/issues/1705.
 
**Steps to reproduce**
1. Open Visual Studio (or another IDE) and open o3de folder.
2. In the Visual Studio go to Edit > Find and Replace > Find in Files.
3. In the search bar type [o3de.org](http://o3de.org/)

**Expected behavior**
Hyperlinks for the o3de documentation are consistently linked in the source code.

**Actual behavior**
Some hyperlinks have provided URL with www domain for the o3de documentation, while others do not.

**Screenshot**

![HyperlinkWWW](https://user-images.githubusercontent.com/100559077/174781936-3f5ad6dc-63ed-4533-afd1-b0fec8491973.png)

![Hyperlink](https://user-images.githubusercontent.com/100559077/174781938-0f27e0e8-baf0-48c6-847e-2537627c9276.png)
**Found in Branch**
Development (8d04988)

**Desktop:**
- Device: PC
- OS: Windows
- Version: 10
- CPU: Intel Core i7-11700F
- GPU: NVIDIA GeForce RTX 3070

AMZN-daimini:
Moving to @o3de/sig-docs as this sounds more like something that would need specific guidelines.

The proper url to the o3de website should be the one with the `www.` spelled out, but the other format will also work (and redirect to the correct one).

AMZN-alexpete:
fwiw I wish we would leave off the `www` subdomain syntax.  and redirect our `www` to the naked version, but so long as we are consistent with `www` we should be OK.

nonye94:
Hello, do you still need someone to complete this task.

guillaume-haerinck:
I don't think anyone is assigned so feel free to jump on it :)

nonye94:
Greate could you assign it to me please

nonye94:
cheers.

nonye94:
Sorry, can I ask if this is a javascript file? I've forked the project but can't seem to run it on the Visual Studio code terminal, "npm start."

nonye94:
And would I have to install Docker?

guillaume-haerinck:
[...]

PR #15894, opened 2023-05-06
Inconsistency with linking hyperlinks to the source code fix

ajv4288:
## What does this PR do?

Standardizes links to the o3de.org site with www spelled out. Addresses issue #10237

Relevant issue can be found here:
https://github.com/o3de/o3de/issues/10237


gadams3 (approved):
Is there any reason that WWW was left off in the first place? Or why one would be preferred over the other?

Issue #15660, opened 2023-04-10
DPE Inspector: Landscape Canvas componetn button has no label, and creating a new graph via the component displays internal data on the component


jckand-amzn:
**Describe the bug**
Landscape Canvas componetn button has no label, and creating a new graph via the component displays internal data on the component.

**Steps to reproduce**
1. Launch the Editor with the DPE enabled:
```
Editor.exe --ed_enableDPE=true
```
2. Open a new or existing level.
3. Create a new entity.
4. Add a Landscape Canvas component.
5. Click the label-less button to open the Landscape Canvas tool.
6. Close the tool, and click the button once more.

**Expected behavior**
Landscape Canvas component has a label which reads "Edit", and internal information doesn't appear on component.

**Actual behavior**
Landscape Canvas component is missing a label on the component, and after closing/reopening a graph, what appears to be the graph UUID is displayed on the component.

**Screenshots/Video**
![image](https://user-images.githubusercontent.com/82226555/231007004-eaf7c34f-e9ad-4898-8d8d-e1748248aecd.png)

**Found in Branch**
development

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
311c24f1a5264b6e4e3b6709831759d014d24168

cgalvan:
https://github.com/o3de/o3de/pull/15909 fixes the missing label part of this bug, but not the unintentional display of graph data.

jckand-amzn:
Verified the button now has label text on development@ d7ea7f7ee597079d7e246fd50f76a550a1f774d7. Internal graph data is still displaying on the component. @cgalvan is aware and investigating a separate fix.

PR #15909, opened 2023-05-09
Fix for a generic handling of invokable attributes

### LINKED ISSUE #15717: DPE Inspector: Camera component has duplicate "Match Viewport" buttons
**Describe the bug**
Camera component has duplicate "Match Viewport" buttons.

**Steps to reproduce**
```
Editor.exe --ed_enableDPE=true
```
2. Expand the Default Level Environment entity hierarchy.
3. Select the "Camera" entity.
 
**Expected behavior**
Component properties are listed only once.

**Actual behavior**
The "Match Viewport" button is duplicated on the component.

**Screenshots/Video**
image

**Found in Branch**
development

0fa6eea55514ef710fa4a0d40ea31ad43b181de9

cgalvan:
## What does this PR do?

Fixes #15717, #15660

Added handling in `LegacyAttributeToDomValue` to check if the attribute is an invokable function as opposed to a value passed directly. This will fix several cases where attributes such as `Visibility` and `ButtonText` are supplied as a function.

Also added some additional properties to the DPE standalone view to exercise these paths.

DPEInvokableAttributes

## How was this PR tested?

Tested the properties in the DPE standalone, as well as the buttons in the Camera component and Landscape Canvas component.

### DIFF
--- a/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/DocumentSchema.h
+++ b/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/DocumentSchema.h
@@ -200,18 +200,21 @@ namespace AZ::DocumentPropertyEditor
             {
                 return AZ::Reflection::ReadGenericAttributeToDomValue(instance, attribute).value_or(AZ::Dom::Value());
             }
-            // Handle the attribute providing an invokable function instead of the value directly
-            else if (attribute->CanDomInvoke(Dom::Value(Dom::Type::Array)))
-            {
-                return attribute->DomInvoke(instance, Dom::Value(Dom::Type::Array));
-            }
             else
             {
                 AZ::AttributeReader reader(instance, attribute);
                 AttributeType value;
                 if (!reader.Read<AttributeType>(value))
                 {
-                    return AZ::Dom::Value();
+                    // Handle the attribute providing an invokable function instead of the value directly
+                    if (attribute->CanDomInvoke(Dom::Value(Dom::Type::Array)))
+                    {
+                        return attribute->DomInvoke(instance, Dom::Value(Dom::Type::Array));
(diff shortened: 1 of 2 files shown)


forkercat (approved):

dtamkin1 (approved):

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/DocumentSchema.h
```diff
@@ -200,6 +200,11 @@ namespace AZ::DocumentPropertyEditor
             {
                 return AZ::Reflection::ReadGenericAttributeToDomValue(instance, attribute).value_or(AZ::Dom::Value());
             }
+            // Handle the attribute providing an invokable function instead of the value directly
```
alexmontAmazon, line 203: I'm slightly concerned how this would affect attributes that are intended to by invokable always, like changeNotify. In that case, calling LegacyAttributeToDomValue wouldn't give the DomValue for the callable attribute, but instead attempt to call it itself to retrieve a value that was not intended. Have you tried making this the final else case, like putting it below at line 214?
Something like:
if (!reader.Read<AttributeType>(value))
{
   if (attribute->CanDomInvoke(Dom::Value(Dom::Type::Array)))
    {
        return attribute->DomInvoke(instance, Dom::Value(Dom::Type::Array));
    }
   else
   {
    return AZ::Dom::Value();
    }

that way it would attempt to read the regular value first, and if it failed, it would try to invoke the attribute to get the result it was looking for.

cgalvan: Good call, updated!

alexmontAmazon (approved):
awesome, thanks!

Issue #15625, opened 2023-04-07
Adding an entity as an Override and then deleting the entity from the prefab causes the entity to be added to the level


raktalen:
**Describe the bug**
Adding an entity as an Override and then deleting the entity from the prefab causes the entity to be added to the level

**Steps to reproduce**
1. Create a prefab with a Child and a Grandchild entity (Prefab > Child > Grandchild)
2. Create an override that adds a new entity as a child of the Grandchild entity (Prefab > Child > Grandchild > Great-Grandchild)
3. Open prefab in focus mode
4. Delete the Grandchild Entity
5. Notice that the Great-Grandchild entity is added to the level

**Expected behavior**
Not 100% sure here , the expected behavior is something up for discussion.

**Actual behavior**
The entity added as an override gets added to the level.

**Found in Branch**
Release 2305

monroegm:
From sig/content triage: we should probably ignore the great-grandchild override since it is no longer valid. We should also check that the level still saves and loads properly when you're in this state.

PR #15914, opened 2023-05-09
Remove entities without a parent from instance object

### LINKED ISSUE #15627: Deleting an entity as an Override and then adding a child to that entity from the prefab causes the entity to be added to the level
**Describe the bug**
Deleting an entity as an Override and then adding a child to that entity from the prefab causes the entity to be added to the level.

**Assets required**
[!Important!] Please provide sample assets needed to reproduce the issue, either as an [attachment](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/attaching-files) or as a link to a public asset in Github.

**Steps to reproduce**
1. Create a prefab with two Child entities (Prefab > Child1  | Child2)
2. Create an override that deletes Child1  (Prefab > Child2)
3. Open prefab in focus mode
4. Add a child to 'Child1' ((Prefab > Child1>Grandchild | Child2)
5. Save prefab and close focus mode
6. Notice that the Grandchild entity is added to the level

**Expected behavior**
Not 100% sure here , the expected behavior is something up for discussion.

**Actual behavior**
The entity added as an override gets added to the level.

**Found in Branch**
Release 2305

michabr:
## What does this PR do?

When using the "add/remove entity as override" prefab feature, a prefab instance can end up with child entities that do not have a valid parent. This PR removes the invalid entities from their instance object if they do not have a parent. Otherwise, the Entity Outliner places the entities on the root, and the level is not arranged correctly in the outliner tree.

Note: Warnings on invalid patch paths will still show up in the console when propagating changes to an instance, or in the error dialog when loading the level. We may want to address this in a separate PR.

Fixes #15627, #15625

## How was this PR tested?

Tested in Editor using repro steps in ticket.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceUpdateExecutor.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceUpdateExecutor.cpp
@@ -24,6 +24,7 @@
 #include <AzToolsFramework/Prefab/PrefabSystemComponentInterface.h>
 #include <AzToolsFramework/Prefab/Template/Template.h>
 #include <AzToolsFramework/Prefab/PrefabPublicRequestBus.h>
+#include <AzToolsFramework/ToolsComponents/TransformComponent.h>
 
 namespace AzToolsFramework
 {
@@ -266,6 +267,36 @@ namespace AzToolsFramework
 
                             AzToolsFramework::EditorEntityContextRequestBus::Broadcast(
                                 &AzToolsFramework::EditorEntityContextRequests::HandleEntitiesAdded, newEntities);
+
+                            auto removeOrphanedEntities = [](Instance& instance)
+                            {
+                                instance.RemoveEntities(
+                                    [&instance](const AZStd::unique_ptr<AZ::Entity>& entity)
+                                    {
+                                        AZ::EntityId parentId;
+                                        if (AzToolsFramework::Components::TransformComponent* transformComponent =
+                                            entity->FindComponent<AzToolsFramework::Components::TransformComponent>())
+                                        {
+                                            parentId = transformComponent->GetParentId();
+                                        }
+
+                                        return (!parentId.IsValid() || !instance.GetEntityAlias(parentId).has_value());
+                                    });
+                            };
+
+                            // Remove any orphaned entities from the instance object (patch data stays intact).
+                            // Entities can contain unregistered parent aliases in the target template after
+                            // applying add/remove entity patches to the source template. These show up as
+                            // invalid parent EntityIds.
+                            // Note: This needs to be validated after the HandleEntitiesAdded call to ensure
+                            // parent entityIds have been set up (ex. when creating a new entity, its parent
+                            // is set up on a callback triggered via HandleEntitiesAdded)
+                            removeOrphanedEntities(*instanceToUpdate);
+                            instanceToUpdate->GetNestedInstances(
+                                [&](AZStd::unique_ptr<Instance>& nestedInstance)
+                                {
+                                    removeOrphanedEntities(*nestedInstance.get());
+                                });
 
                             if (!m_isRootPrefabInstanceLoaded &&
                                 instanceToUpdate->GetTemplateSourcePath() == m_rootPrefabInstanceSourcePath)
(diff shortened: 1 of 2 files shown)


Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceUpdateExecutor.cpp
```diff
@@ -267,6 +268,36 @@ namespace AzToolsFramework
...
                                 &AzToolsFramework::EditorEntityContextRequests::HandleEntitiesAdded, newEntities);
 
+                            auto removeOrphanedEntities = [](Instance& instance)
+                            {
+                                instance.RemoveEntities(
+                                    [&instance](const AZStd::unique_ptr<AZ::Entity>& entity)
+                                    {
+                                        AZ::EntityId parentId;
+                                        if (AzToolsFramework::Components::TransformComponent* transformComponent =
+                                            entity->FindComponent<AzToolsFramework::Components::TransformComponent>())
```
forkercat, line 278: nit: Can use auto here

```cpp
if (auto transformComponent = entity->FindComponent<Components::TransformComponent>())
{
}
```

forkercat (approved):
Looks good :)

If we're doing that in `InstanceUpdateExecutor::UpdateTemplateInstancesInQueue`, would there be a performance issue if the level is huge?

michabr:
> Looks good :)
> 
> If we're doing that in `InstanceUpdateExecutor::UpdateTemplateInstancesInQueue`, would there be a performance issue if the level is huge?

It could possibly affect performance for large levels. It's doing a "find" on an unordered map, so depends on how performant our unordered map implementation is. I'll see about running a performance check. We could probably optimize to only check entities if there are add/remove overrides present etc.

spham-amzn (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/Instance/InstanceUpdateExecutor.cpp
```diff
@@ -267,6 +268,36 @@ namespace AzToolsFramework
...
+                            };
+
+                            // Remove any orphaned entities from the instance object (patch data stays intact).
+                            // Entities can contain unregistered parent aliases in the target template after
+                            // applying add/remove entity patches to the source template. These show up as
+                            // invalid parent EntityIds.
+                            // Note: This needs to be validated after the HandleEntitiesAdded call to ensure
+                            // parent entityIds have been set up (ex. when creating a new entity, its parent
+                            // is set up on a callback triggered via HandleEntitiesAdded)
+                            removeOrphanedEntities(*instanceToUpdate);
```
srikappa-amzn, line 294: Rather than do it after loading the prefab, we can do this during the loading of the transform component itself : https://github.com/o3de/o3de/blob/development/Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/TransformComponentSerializer.cpp#L39-L43

In that code, if m_parentEntityId is invalid, we can mark the entity and remove it from the instance in InstanceSerializer.cpp or EntitySerializer.cpp

This will ensure that we don't need to check the parent of every entity in a prefab instance recursively(which can be a lot) and won't impact performance.

michabr: When the transform component is loaded, we are not guaranteed that its parent entity is already loaded yet.

srikappa-amzn: oh, that's right. The entity id is actually valid. However, the instance won't have an entity with that id. But we can still do the same checks you are doing here in InstanceSerializer.cpp this way:

```
Load Instance()
{
	// Create a new struct to store a vector of transform component parents and add it to JsonSerializationMetadata
	while loading each entity()
	{
		LoadTransformComponent()
		{
			add parent id to vector.
		}
	}
	
	for each (parentId : vector)
	{
		if (parentId is not present in Instance)
		{
			// Remove from instance
		}
	}
}
```

If we do it this way, we can avoid 

1. The additional overhead to recurse over every nested instance twice (once during loading and once after loading)
2. The time to find the transform component in each entity (O(n) for each entity, where n is number of components in entity)

michabr: Could you explain #1 in further detail? This logic should happen whenever an instance needs to be updated due to propagation.

michabr: The add/remove entity patches still exist despite the entity not being part of the instance. When template propagation occurs, it reapplies the patches on the target and so we need to recheck this logic.

PR #15926, opened 2023-05-10
Atom Tools: Prevent asset selection widgets from displaying non existent assets

gadams3:
## What does this PR do?

This may happen if settings are saved referencing lighting and model presets or other files that were deleted since the settings were saved.

## How was this PR tested?

Confirmed that asset selection controls in material editor and material canvas still operate as expected and reject nonexistent files.


rgba16f (approved):

amzn-changml:
Succeeded through https://jenkins.build.o3de.org/job/O3DE/view/change-requests/job/PR-15926/, but the status did not update in this PR. Overriding this check.

PR #15929, opened 2023-05-10
Atom Tools: Change file enumeration to add logging and use extension checks instead of wild cards

gadams3:
## What does this PR do?

This PR changes file enumeration functions to use a faster extension filter instead of a wild card filter. 
Logging was also added to record how long it takes to enumerate files in different systems.
Parallelization was removed from directory traversal.
This is all in an ever dude determine what might cause extended startup times on some machines.
The extended delay is usually on the first run after downloading the engine or setting up a new project where Python symbols and other data have not been saved yet.

## How was this PR tested?

Launched affected tools and ran included Python scripts

### DIFF
--- a/Gems/Atom/Tools/AtomToolsFramework/Code/Include/AtomToolsFramework/Util/Util.h
+++ b/Gems/Atom/Tools/AtomToolsFramework/Code/Include/AtomToolsFramework/Util/Util.h
@@ -290,8 +290,8 @@ namespace AtomToolsFramework
     // Visits all scan folders asynchronously, gathering all of the paths matching the filter. 
     AZStd::vector<AZStd::string> GetPathsInSourceFoldersMatchingFilter(const AZStd::function<bool(const AZStd::string&)> filterFn);
 
-    //! Collect a set of file paths from all project safe folders matching a wild card
-    AZStd::vector<AZStd::string> GetPathsInSourceFoldersMatchingWildcard(const AZStd::string& wildcard);
+    //! Collect a set of file paths from all project safe folders matching an extension
+    AZStd::vector<AZStd::string> GetPathsInSourceFoldersMatchingExtension(const AZStd::string& extension);
 
     //! Return a list of all asset safe folders except for those underneath the cache folder, usually intermediate asset folders.
     AZStd::vector<AZStd::string> GetNonCacheSourceFolders();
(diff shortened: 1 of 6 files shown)


Gems/Atom/Tools/AtomToolsFramework/Code/Include/AtomToolsFramework/Util/Util.h
```diff
@@ -290,8 +290,8 @@ namespace AtomToolsFramework
     // Visits all scan folders asynchronously, gathering all of the paths matching the filter. 
     AZStd::vector<AZStd::string> GetPathsInSourceFoldersMatchingFilter(const AZStd::function<bool(const AZStd::string&)> filterFn);
 
-    //! Collect a set of file paths from all project safe folders matching a wild card
-    AZStd::vector<AZStd::string> GetPathsInSourceFoldersMatchingWildcard(const AZStd::string& wildcard);
+    //! Collect a set of file paths from all project safe folders matching an extension
+    AZStd::vector<AZStd::string> GetPathsInSourceFoldersMatchingExtension(const AZStd::string& extension);
```
AMZN-alexpete, line 294: Could you bump the `AtomToolsFramework/gem.json` `version` to `1.0.0` if this is a breaking API change?    More info here: https://github.com/o3de/community/blob/main/guides/o3de-version-guidelines.md

Gems/Atom/Tools/AtomToolsFramework/Code/Source/EntityPreviewViewport/EntityPreviewViewportSettingsSystem.cpp
```diff
@@ -589,6 +589,7 @@ namespace AtomToolsFramework
         AZ::Job* job = AZ::CreateJobFunction(
             [toolId = m_toolId]()
             {
+                AZ_TracePrintf("EntityPreviewViewportSettingsSystem", "Enumerating presets started.");
```
AMZN-alexpete, line 592: Are these printfs debugging code?

AMZN-alexpete (approved):
Just a suggestion to update the version if it's a breaking API change

Gems/Atom/Tools/AtomToolsFramework/Code/Source/EntityPreviewViewport/EntityPreviewViewportSettingsSystem.cpp
gadams3: Yes and no. I'm mostly interested in the time stamps for when these events begin and end so users experiencing extended startup times can provide the log files with the information from that specific run. Honestly, I should sprinkle more of these to capture additional data during initialization. These are the areas that I suspected might cause a delay based on the size of the project, other than connecting to the AP, which has become negligible since optimizations, and initializing other project specific gems and features. However, the number is negligible on the hardware that I have tested, going between automated testing and MPS. MPS only takes slightly longer because it's initializing the server, splash screen, and other systems that could probably check for editor, launcher, game mode.

cgalvan (approved):

Issue #15925, opened 2023-05-10
Project Manager Gem inspector doesn't update after downloading gem


AMZN-alexpete:
**Describe the bug**
When upgrading a project to use the latest version of PopcornFX,I pressed the "Download Gem" button and the gem was downloaded (green check mark next to gem entry) but the Gem inspector still shows the "Download Gem" button.

UPDATE: it could be the actual gem is not being registered.  It appears to be downloaded fine and is visible in O3DE/Gems/PopcornFX/2.16.2  but is missing from the `o3de_manifest.json`

UPDATE 2: the gem is getting registered in another gem instead of the manifest because it was downloaded into a subfolder
e.g.
c:\users\foo\O3DE\gems\PopcornFX  <-- original gem located here
c:\users\foo\O3DE\gems\PopcornFX \2.16.2 <-- new gem downloaded here
The register logic needs to specify the gem should be registered in the manifest because the default was changed to register in the closest parent manifest. (`--force-register-with-o3de-manifest`)

**Assets required**
[!Important!] Please provide sample assets needed to reproduce the issue, either as an [attachment](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/attaching-files) or as a link to a public asset in Github.

**Steps to reproduce**
1. Open Project Manager
2. Select 'Configure Gems' for an existing project
3. Select PopcornFX 
4. In the Gem Inspector click "Download Gem"

**Expected behavior**
The gem downloads and the gem inspector updates to show the edit/remove buttons

**Actual behavior**
The download gem button persists

**Screenshots/Video**
![image](https://github.com/o3de/o3de/assets/26804013/bba25425-89d6-4d09-b198-c9b1f7b93e42)

**Found in Branch**
main (2305 release)

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
Please provide the SHA/hash that identifies the latest commit from the o3de/o3de repository if the issue is reproducible in the development or other official branches. 
You can get the commit ID by running the `git rev-parse HEAD` command on your current branch.

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Windows
 - Version 10

Issue #15893, opened 2023-05-06
PopcornFX 2.16.2 gem does not activate after adding


petrocket:
**Describe the bug**
After adding PopcornFX 2.16.2 Gem in O3DE SDK 23.05.0 Project Manager and rebuilding, the PopcornFX components are not visible in the Editor

It's possible this is related to how the gem appears in `project.json` where it includes the version `PopcornFX==2.16.2` but maybe there is a bug that prevents the dll from being added to the setreg list of dlls to load when the "==2.1.6.2" version specifier is present.

**Assets required**
PopcornFX Gem

**Steps to reproduce**
1. Install O3DE SDK 23.05.0
2. Add the PopcornFX 2.16.2 gem to a project 
3. Build the project
4. Open the Editor

**Expected behavior**
PopcornFX components are visible in the editor

**Actual behavior**
No PopcornFX components are visible

**Found in Branch**
main

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
[8c1421e](https://github.com/o3de/o3de/commit/8c1421e2a0d168d34ae9ca6c763270fe8395a809)

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Windows
 - Version 10

OBWANDO:
when adding, it puts the version number in the json list "PopcornFX=2.16.2" which doesn't work.
IF you rename to "PopcornFX" and then add or remove a gem in the project manager, it will compile.
If you just rename, save and compile, it doesn't pick it up.

lmbr-pip:
Unsure if this is build or editor issue, so tagging both SIG/build and SIG/content

PR #15933, opened 2023-05-10
Fix PopcornFX loading and registration

AMZN-alexpete:
## What does this PR do?

Fixes https://github.com/o3de/o3de/issues/15893

When a specific version of a gem is used in ProjectManager the `project.json` `gem_names` entry looks like `PopcornFX==2.16.2` but when CMake is looking for all the variants to load the `==2.16.2` part was not stripped and so none of the load targets were found.  This change removes the version specifiers so the gem name alone is used in the search for variants.

Fixes https://github.com/o3de/o3de/issues/15925

The default gem registration will not use the `o3de_manifest.json` in the case when the user downloads a version of a gem into a folder that has a `gem.json` in a parent folder - this can happen for downloaded gems that were previously downloaded without gem versions, and then a subsequent versioned gem is downloaded and ends up in a versioned subfolder.  The fix is to always register downloaded content in the `o3de_manifest.json`.  The user can re-register the content if they want to register it with a project or engine, but that would be uncommon behavior when most downloaded content is meant to be available for multiple engines and projects.

## How was this PR tested?

Reproduced the issue in python debugger, then applied the fix to verify it no longer repros and the gem appears in the `o3de_manifest.json`
Used cmake configure with Python 2.16.2. and verified the Python dlls appear in the .setreg like they should and get loaded

### DIFF
--- a/cmake/Gems.cmake
+++ b/cmake/Gems.cmake
@@ -236,6 +236,7 @@ function(ly_enable_gems)
 
     # Remove any version specifiers before looking for variants
     # e.g. "Atom==1.2.3" becomes "Atom"
+    unset(GEM_NAMES)
     foreach(gem_name_with_version_specifier IN LISTS ly_enable_gems_GEMS)
         o3de_get_name_and_version_specifier(${gem_name_with_version_specifier} gem_name spec_op spec_version)
         list(APPEND GEM_NAMES "${gem_name}")
--- a/cmake/Gems.cmake
+++ b/cmake/Gems.cmake
@@ -234,6 +234,14 @@ function(ly_enable_gems)
         set(ENABLED_GEMS ${store_temp}) # restore value of ENABLED_GEMS just in case...
     endif()
 
+    # Remove any version specifiers before looking for variants
+    # e.g. "Atom==1.2.3" becomes "Atom"
+    foreach(gem_name_with_version_specifier IN LISTS ly_enable_gems_GEMS)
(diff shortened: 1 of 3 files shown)


cmake/Gems.cmake
```diff
@@ -234,6 +234,14 @@ function(ly_enable_gems)
         set(ENABLED_GEMS ${store_temp}) # restore value of ENABLED_GEMS just in case...
     endif()
 
+    # Remove any version specifiers before looking for variants
+    # e.g. "Atom==1.2.3" becomes "Atom"
+    foreach(gem_name_with_version_specifier IN LISTS ly_enable_gems_GEMS)
+        o3de_get_name_and_version_specifier(${gem_name_with_version_specifier} gem_name spec_op spec_version)
+        list(APPEND GEM_NAMES "${gem_name}")
+    endforeach()
+    set(ly_enable_gems_GEMS ${GEM_NAMES})
```
lumberyard-employee-dm, line 243: This should unset `GEM_NAMES` before the `foreach` loop.
When functions in cmake are invoked, it's values are visible to any nested calls to other functions within it.
https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#variables

What this means is that if the caller has a GEM_NAMES variable in scope, then a copy of it would be used to initialize this function call scope version of `GEM_NAMES` and that value would be appended to.

lumberyard-employee-dm (approved):

spham-amzn (approved):

tkothadev (approved):

Issue #14691, opened 2023-02-21
Crash when deleting a component in DPE Inspector


forkercat:
**Describe the bug**

Prerequisite: behind a feature flag `ed_enableDPE = true`

The editor would crash when deleting a component in DPE Entity Inspector.

<img width=400px src="https://user-images.githubusercontent.com/11157226/220273815-cc4afda7-68fb-463f-af44-adca4037697b.png">

**Steps to reproduce**
1. Enable DPE flag
2. Create an empty level
3. Create an entity
4. Add a **Camera component** on the new entity
5. Delete the Camera component

Note: I cannot reproduce it with Comment component. Looks like the crash only occurs for some components.

**Expected behavior**

No crash

**Actual behavior**

Crash

**Screenshots/Video**

![image](https://user-images.githubusercontent.com/11157226/220273197-da847fea-986d-4506-86c0-341f471256b1.png)

**Found in Branch**

Development https://github.com/o3de/o3de/commit/0d155714601e1f90dbefce834b6374a2b3a357d6

**Desktop/Device:**
* Device: PC
* OS: Windows 10
* CPU: AMD Ryzen 5950x
* GPU: NVIDIA RTX 3070
* Memory: 64 GB

**Additional context**

<details>
  <summary>CLICK ME TO SEE THE CALL STACK</summary>

```
>	ComponentEntityEditorPlugin.dll!AZ::Component::GetEntityId() Line 53	C++
[Inline Frame] ComponentEntityEditorPlugin.dll!AZ::DocumentPropertyEditor::ComponentAdapter::HandleMessage::__l2::<lambda_1>::operator()(const AZ::Dom::Value &) Line 105	C++
[Inline Frame] ComponentEntityEditorPlugin.dll!AZStd::Internal::INVOKE(AZ::DocumentPropertyEditor::ComponentAdapter::HandleMessage::__l2::<lambda_1> &) Line 208	C++
[Inline Frame] ComponentEntityEditorPlugin.dll!AZStd::invoke(AZ::DocumentPropertyEditor::ComponentAdapter::HandleMessage::__l2::<lambda_1> &) Line 54	C++
[Inline Frame] ComponentEntityEditorPlugin.dll!AZStd::Internal::function_util::invoke_void_return_wrapper<void>::call(AZ::DocumentPropertyEditor::ComponentAdapter::HandleMessage::__l2::<lambda_1> &) Line 41	C++
[...]

Issue #15918, opened 2023-05-10
Undoing multi-edit operation causes a crash in DPE based entity inspector


srikappa-amzn:
**Describe the bug**
Undoing multi-edit operation causes a crash in DPE based entity inspector

**Steps to reproduce**
1. Create a new level or open an existing level
2. Select any two entities in the level
3. Modify the property in entity inspector
4. Select Edit> Undo

**Expected behavior**
Operation is undone and both entities go back to the previous state before the multi-edit

**Actual behavior**
Causes a crash

**Found in Branch**
development

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
eb95bee8b8e531273e69c8f7de96db1d74b3c69a

srikappa-amzn:
Notes from debugging:

The assert seems to be originating from https://github.com/o3de/o3de/blob/development/Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DPEComponentAdapter.cpp#L98 when DoRefresh() is called on what looks like an invalid adapter will garbage data. 

Once the crash is prevented, the next thing we need to fix is creating multiple undo nodes for multi-edit. Right now, https://github.com/o3de/o3de/blob/development/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/DocumentPropertyEditor/PrefabComponentAdapter.cpp#L151 is called twice. So when undo is pressed, it is undoing only the property change for the last component modified. This needs to be changed such that 
1. Either its called for all undo nodes (which is not great for performance if there are many entities)
2. A single undo node wraps all these undo nodes and treats it as one operation.

srikappa-amzn:
https://github.com/o3de/o3de/issues/15918 and https://github.com/o3de/o3de/pull/15951 combined fixed this issue. 

Another Redo issue was identified for some component properties. But it is unrelated to multi-edit and already existed in development before these PRs. Created a new issue for that : https://github.com/o3de/o3de/issues/15975

PR #15943, opened 2023-05-12
Fixes crash when undoing a multi-edit change in the DPE based Entity Inspector.

srikappa-amzn:
## What does this PR do?

This PR fixes the crash reported in https://github.com/o3de/o3de/issues/15918

Changes include:

- Listening to OnEntityDestruction() instead of OnEntityDestroyed() because the former gets sent when the entity destruction begins and the later is sent when it ends. Between the begin and the end, InvalidatePropertyDisplay() is getting called, which is queuing up a property refresh. So I changed it to listen to OnEntityDestruction
- Disconnected from buses when the entity is being destroyed. This prevents any further refreshes from being queued.
- Changed assert to if check for empty entity to handle cases where entity id is valid(eg. 0, 1, 2, etc.) but doesn't correspond to a live entity. Could happen when the component adapter holds garbage data.

## How was this PR tested?

Manually. Also verified that a previously fixed crash remains fixed : https://github.com/o3de/o3de/issues/14691

## Next steps

While the crash is fixed, the undo nodes are still not grouped. i.e, if you select 4 entities and make a multi-edit change and hit undo, you'll need to hit undo 4 times in order for that to work. Will look into fixing that next.


Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DPEComponentAdapter.h
```diff
@@ -57,7 +57,9 @@ namespace AZ::DocumentPropertyEditor
         //! @param serializedPath The serialized path to use to check whether an override is present corresponding to it.
         void CreateLabel(AdapterBuilder* adapterBuilder, AZStd::string_view labelText, AZStd::string_view serializedPath) override;
 
-        void OnEntityDestroyed(const AZ::EntityId&) override;
+
```
forkercat, line 60: nit: extra empty line

forkercat (approved):
Nice catch. Thanks for fixing this!

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DPEComponentAdapter.cpp
```diff
@@ -168,10 +173,14 @@ namespace AZ::DocumentPropertyEditor
...
 
-    void ComponentAdapter::OnEntityDestroyed(const AZ::EntityId& entityId)
+    void ComponentAdapter::OnEntityDestruction(const AZ::EntityId& entityId)
     {
         if (entityId == m_entityId)
         {
+            AzToolsFramework::PropertyEditorGUIMessages::Bus::Handler::BusDisconnect();
+            AzToolsFramework::ToolsApplicationEvents::Bus::Handler::BusDisconnect();
+            AzToolsFramework::PropertyEditorEntityChangeNotificationBus::MultiHandler::BusDisconnect(m_entityId);
+
```
michabr, line 183: I'm wondering if some of these buses are expected to get called on entity destruction (ex. to update the outliner).

michabr (approved):
Thanks!

dtamkin1 (approved):

alexmontAmazon (approved):

PR #15946, opened 2023-05-12
Fix NetBindComponent leaks / crashes

mbalfour-amzn:
## What does this PR do?

Multiplayer Sample experienced infrequent, intermittent crashes in NetBindComponent::NotifyPreRender ( https://github.com/o3de/o3de-multiplayersample/issues/349 ). These crashes appear to be caused through "leaked" pointers to the NetBindComponent and its entity, where the component registered itself with the NetworkEntityManager and NetworkEntityTracker but then never unregistered itself prior to deletion.

Upon investigation, it appears that this case was caused by deleting networked entities that had been initialized but not Activated. The registration occured in PreInit(), but the deregistration occured in Deactivate(), so if the entity was initialized but not activated, the Deactivate() would never get called and the deregistration would get skipped.

This PR fixes the problem by ensuring that the NetBindComponent always calls Deregister() on destruction.

## How was this PR tested?

1. Added a new unit test demonstrating that initializing but not activating the NetBindComponent was leaving remnants in the NetworkEntityManager and NetworkEntityTracker. Verified that with the new changes, the unit test passes, and no remnants remain.
2. Verified in MPS that there were occasional cases of the NetworkEntityManager marking enitites for removal while they were in the Init state, and verified in the NetBindComponent destructor that in those cases, it was getting destroyed while still registered.

There are still other entity lifetime issues occuring in MPS in development right now, but I haven't seen any more occurrences of this specific crash with this fix.

### DIFF
--- a/Gems/Multiplayer/Code/Include/Multiplayer/Components/NetBindComponent.h
+++ b/Gems/Multiplayer/Code/Include/Multiplayer/Components/NetBindComponent.h
@@ -50,7 +50,7 @@ namespace Multiplayer
         static void GetIncompatibleServices(AZ::ComponentDescriptor::DependencyArrayType& incompatible);
 
         NetBindComponent();
-        ~NetBindComponent() override = default;
+        ~NetBindComponent() override;
 
         //! AZ::Component overrides.
         //! @{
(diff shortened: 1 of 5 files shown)


AMZN-Olex (approved):
Nice find!

kberg-amzn (approved):
This is a lot safer, excellent change 👍 .. I wonder how we were getting entities in this state however, initialized but not activated.. very strange.

mbalfour-amzn:
> This is a lot safer, excellent change 👍 .. I wonder how we were getting entities in this state however, initialized but not activated.. very strange.

My current working theory is that I believe it could happen when the client spawns in right as an energy ball is killing itself. In one frame, the EntityReplicationManager will get change messages that cause it to spawn the entity and a delete message that will cause it to destroy the entity. It would never have time to activate the entity.

PR #15949, opened 2023-05-12
Fix resize crash on Vulkan

### LINKED ISSUE #15947: Crash when resizing windows on Vulkan
**Describe the bug**
Crash when resizing the window when running Vulkan

**Steps to reproduce**
1. Build ASV
2. Run GameLauncher
3. Resize the windows multiple times
4. See crash

**Found in Branch**
o3de/development

 - Device: PC
 - OS: Windows, Linux

akioCL:
## What does this PR do?

Fix Vulkan crash when resizing windows

Fixes #15947 

## How was this PR tested?

Run ASV Launcher and Editor while resizing the window

### DIFF
--- a/Gems/Atom/RHI/Vulkan/Code/Source/RHI/SwapChain.cpp
+++ b/Gems/Atom/RHI/Vulkan/Code/Source/RHI/SwapChain.cpp
@@ -74,9 +74,11 @@ namespace AZ
         {
             if (m_pendingRecreation)
             {
+                VkSwapchainKHR oldSwapchain = m_nativeSwapChain;
                 ShutdownImages();
-                InvalidateNativeSwapChain();
                 CreateSwapchain();
+                // Destroy the old swapchain AFTER creating the new one (because it's used for building the new one)
+                InvalidateNativeSwapChain(oldSwapchain);
                 InitImages();
 
                 m_pendingRecreation = false;
@@ -150,7 +152,8 @@ namespace AZ
                 return;
             }
 
-            InvalidateNativeSwapChainImmediately();
+            InvalidateNativeSwapChain(m_nativeSwapChain);
+            m_nativeSwapChain = VK_NULL_HANDLE;
             InvalidateSurface();
             m_presentationQueue = nullptr;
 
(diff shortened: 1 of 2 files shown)


Gems/Atom/RHI/Vulkan/Code/Source/RHI/SwapChain.cpp
```diff
@@ -74,9 +74,11 @@ namespace AZ
         {
             if (m_pendingRecreation)
             {
+                VkSwapchainKHR oldSwapchain = m_nativeSwapChain;
                 ShutdownImages();
-                InvalidateNativeSwapChain();
                 CreateSwapchain();
+                // Destroy the old swapchain AFTER creating the new one (because it's used for building the new one)
+                InvalidateNativeSwapChain(oldSwapchain);
```
moudgils, line 81: You are destroying the oldSwapchain right away with these changes. Wouldn't this cause a crash? I had to fix issues around this problem by adding support to put the old swap chain on the release queue so it is destroyed after 3 frames. In order for this PR to go in I recommend we test Editor, Material Editor, game launcher while resizing them + minimizing them on PC and Vulkan. Jus to make sure we are not introducing a previously fixed crash.

rgba16f (approved):

VickyAtAZ (commented):
Would it be an issue that we will have higher gpu peak memory usage since we have two set of swapchians at same time since we didn’t release old one when create new one
I wonder why do we need old swapchain to create new one. (I didn’t check the code)
is it possible we can avoid this restriction?

moudgils (approved):

Gems/Atom/RHI/Vulkan/Code/Source/RHI/SwapChain.cpp
martinwinter-huawei: This seems to still be an issue, it does not happen a lot, but when resizing the window often enough, we still reliably get it to hang on a condition variable in `InvalidateNativeSwapChain`.
image

Is that a know issue?

PR #15951, opened 2023-05-12
Give undo batch handling responsibility to PropertyManagerComponent

srikappa-amzn:
## What does this PR do?

This PR fixes one of the issues listed in https://github.com/o3de/o3de/issues/15918 where undoing a multi-edit action requires hitting undo 'n' times, where n is the number of entities being modified

This PR gives the responsibility of creating and ending undo batch to PropertyManagerComponent, which is the source for all DPE property edits.

This PR needs to go along with https://github.com/o3de/o3de/pull/15943 to fix the issue fully.

Another cool benefit we get with this approach is that we can do both direct template edits and overrides together using multi-edit. The prefab adapter takes care of doing the right calls. But the undo system doesn't need to know anything about it rightfully so.

## How was this PR tested?

Manually along with https://github.com/o3de/o3de/pull/15943

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/DocumentPropertyEditor/PrefabComponentAdapter.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/Prefab/DocumentPropertyEditor/PrefabComponentAdapter.cpp
@@ -119,28 +119,17 @@ namespace AzToolsFramework::Prefab
                 instanceUpdateExecutorInterface->SetShouldPauseInstancePropagation(true);
             }
 
-            if (m_currentUndoBatch)
+            AzToolsFramework::ToolsApplicationRequests::Bus::BroadcastResult(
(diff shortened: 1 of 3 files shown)


forkercat (approved):
Looks good. Suggest testing undo/redo/edit for the three scenarios:
1. RPE
2. DPE (inspector override = off)
3. DPE (inspector override = on)

cgalvan (approved):
I'm not very familiar with the `PrefabComponentAdapter` class but this seems like a safe change to move the scope of the undo batch.

alexmontAmazon (approved):
looks good to me, but I'm adding @cgalvan who might have more context here...

srikappa-amzn:
> Looks good. Suggest testing undo/redo/edit for the three scenarios:
> 
> 1. RPE
> 2. DPE (inspector override = off)
> 3. DPE (inspector override = on)

Yes, these cases are working fine. However, I saw a different issue with some slider properties not working correctly. This is happening in current development branch as well without these changes. So I created a different issue for it : https://github.com/o3de/o3de/issues/15975

Issue #15743, opened 2023-04-18
DPE Inspector: Asset assignment property in components cannot be modified


michleza:
It appears that most components with an asset property cannot assign or modify it

Repro:
1 - Launch into Editor
2 - Enable the console variable `ed_enableDPE`
3 - Open a new level
4 - Add a new entity to the scene
5 - Add a component to the entity that has an asset/file property such as Mesh or Input
6 - Attempt to change the asset property of the component

Expected behavior:
You can assign a source file to the property

Actual behavior:
The component is missing controls to modify the property

![Image](https://user-images.githubusercontent.com/98843754/232844609-f1120b87-19de-4cb9-b183-231662bd4cd3.png)

forkercat:
The asset reference does not show up for the reflected `AZ::Data::Asset<AssetData>` instance.
- For Lua component, it is [AZ::Data::Asset\<AZ::ScriptAsset> m_scriptAsset](https://sourcegraph.com/github.com/o3de/o3de/-/blob/Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/ScriptEditorComponent.h?L108)
- For mesh component, it is [AZ::Data::Asset\<RPI::ModelAsset> m_modelAsset](https://sourcegraph.com/github.com/o3de/o3de/-/blob/Gems/AtomLyIntegration/CommonFeatures/Code/Source/Mesh/MeshComponentController.h?L46)

The property handlers that handle the asset ref type are defined in [PropertyAssetCtrl.hxx](https://sourcegraph.com/github.com/o3de/o3de/-/blob/Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyAssetCtrl.hxx).

**Why it does not show up in DPE inspector?**

In [DPE code](https://sourcegraph.com/github.com/o3de/o3de/-/blob/Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp?L599) when the view tries to retrieve the handler from `PropertyEditorToolsSystem`, it doesn't match, thus returning an invalid handler id.

[...]

Issue #15953, opened 2023-05-12
DPE Inspector: Setting asset properties does not work


forkercat:
**Describe the bug**

Follow-up of https://github.com/o3de/o3de/issues/15743

Now asset property can show up in DPE inspector and you can assign assets in the field. However, the property is actually not getting updated.

**Repro steps:**

1 - Launch editor and enable the console variable `ed_enableDPE`
2 - Open a new level and add a new entity to the scene
3 - Add a component from the below list that has an asset/file property
4 - Attempt to change the asset property of the component

**Expected behavior:**

You can assign a source file to the property and the state gets updated correctly.

**Actual behavior:**

The property is not updated correctly. There are serialization warnings showing up in console.

![image](https://github.com/o3de/o3de/assets/11157226/103b9391-0fbd-4c0a-8299-f81804ce86b2)

forkercat:
The callstack leads me to [this code](https://sourcegraph.com/github.com/o3de/o3de/-/blob/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ReflectionAdapter.cpp?L752) invoked by OnChanged message in LegacyReflectionBridge where it tries to serialize the new value to json and deserialize back to object.
Callstack:

```
ComponentEntityEditorPlugin.dll!AZ::JsonSerializer::Store(rapidjson_ly::GenericValue<rapidjson_ly::UTF8<char>,rapidjson_ly::MemoryPoolAllocator<rapidjson_ly::CrtAllocator>> & output, const void * object, const void * defaultObject, const AZ::Uuid & typeId, AZ::JsonSerializer::UseTypeSerializer custom, AZ::JsonSerializerContext & context) Line 31	C++
   ComponentEntityEditorPlugin.dll!AZ::JsonSerialization::Store(rapidjson_ly::GenericValue<rapidjson_ly::UTF8<char>,rapidjson_ly::MemoryPoolAllocator<rapidjson_ly::CrtAllocator>> & output, rapidjson_ly::MemoryPoolAllocator<rapidjson_ly::CrtAllocator> & allocator, const void * object, const void * defaultObject, const AZ::Uuid & objectType, AZ::JsonSerializerSettings & settings) Line 322	C++
> [Inline Frame] ComponentEntityEditorPlugin.dll!AZ::DocumentPropertyEditor::ReflectionAdapterReflectionImpl::VisitObjectBegin::__l53::<lambda_2>::operator()(const AZ::Dom::Value &) Line 752	C++
[...]

PR #15954, opened 2023-05-12
Display the asset property in DPE inspector

forkercat:
## What does this PR do?

Fixes https://github.com/o3de/o3de/issues/15743

This PR fixes the issue with the asset property not being displayed in the DPE entity inspector.

It retrieves the generic type id for the specialized type id and compares it with the handler's type id.

**Remaining issues:**
1. Property update doesn't work: https://github.com/o3de/o3de/issues/15953
1. When switching selection on entities, non-empty asset property field would override the empty asset property field. For example, `_box_1x1` in mesh component would override the Script field in Lua Script (resolved, see updates below)

## How was this PR tested?

Tested in editor


Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/ScriptEditorComponent.cpp
```diff
@@ -1034,7 +1034,7 @@ namespace AzToolsFramework
                         ->Attribute(AZ::Edit::Attributes::PrimaryAssetType, AZ::AzTypeInfo<AZ::ScriptAsset>::Uuid())
                         ->Attribute(AZ::Edit::Attributes::ViewportIcon, "Icons/Components/Viewport/Script.png")
                         ->Attribute(AZ::Edit::Attributes::HelpPageURL, "https://o3de.org/docs/user-guide/components/reference/scripting/lua-script/")
-                        ->DataElement("AssetRef", &ScriptEditorComponent::m_scriptAsset, "Script", "Which script to use")
```
forkercat, line 1037: Setting it to default so [PropertyEditorToolsSystem::GetPropertyHandlerForNode()](https://sourcegraph.com/github.com/o3de/o3de/-/blob/Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyManagerComponent.cpp?L296) can check the default list.

michabr (commented):
I wonder if the second issue you mentioned is related to https://github.com/o3de/o3de/pull/15961.
"When switching selection on entities, non-empty asset property field would override the empty asset property field. For example, _box_1x1 in mesh component would override the Script field in Lua Script (still debugging... Issue TBA)"

michabr (approved):

forkercat:
> I wonder if the second issue you mentioned is related to https://github.com/o3de/o3de/pull/15961.

Yeah. I was trying to see if some similar fixes could resolve the issue, but I didn't make much progress yet.

PR #15955, opened 2023-05-13
Initial pass at monolithic export script for engine-centric workflow

tkothadev:
This is a first draft script for standalone monolithic project export, using an engine-centric workflow.

As of this commit the script does not yet produce the correct output folder, so a future commit is needed to fix that. 

Script is based on these instructions: https://github.com/o3de/o3de-multiplayersample/blob/development/Documentation/PackedAssetBuilds.md

### DIFF
--- a/scripts/ExportScripts/export_profile_monolithic.py
+++ b/scripts/ExportScripts/export_profile_monolithic.py
@@ -3,7 +3,7 @@ from o3de import manifest
 import argparse
 import pathlib
 import os
-
+import shutil
 #Note: project_path and engine_path are computed before this script runs, and can be accessed via o3de_context.project_path
 #for now this script assumes an engine-centric approach, and we assume to use the source built engine for the engine_path
 
@@ -22,6 +22,10 @@ projects_path_cmake_option = f"-DLY_PROJECTS={str(o3de_context.project_path)}"
 
 project_json_data = manifest.get_project_json_data(project_path = o3de_context.project_path)
 project_name= project_json_data.get('project_name')
+
+#clean the output folder if necessary
+if os.path.isdir(args.output_path):
+    shutil.rmtree(args.output_path)
 
 
 #build non-monolithic artifacts
@@ -43,13 +47,16 @@ process_command(['cmake', '-S', '.', '-B', args.mono_build_path, '-DLY_MONOLITHI
 process_command(['cmake', '--build',  args.mono_build_path, '--target', f'{project_name}.GameLauncher', '--target', f'{project_name}.ServerLauncher', '--target', f'{project_name}.UnifiedLauncher', '--config', 'profile'])
 
 #bundle content (WIP)
-# process_command([asset_bundler_exe, f'--project-path="{str(o3de_context.project_path)}"'])
+# process_command([asset_bundler_exe, 'assetList','--assetListFile {}',f'--project-path="{str(o3de_context.project_path)}"'])
 
 #copy artifacts to output directory
 
 pc_pak_path = args.output_path / 'Cache' / 'pc'
 
 os.makedirs(pc_pak_path, exist_ok=True)
+os.makedirs(args.output_path, exist_ok=True)
 # os.makedirs(args.output_path / 'Gems' / 'AWSCore')
-process_command(['copy', str(o3de_context.project_path / 'AssetBundling' / 'Bundles' / '*.pak'), str(pc_pak_path)])
-process_command(['copy', str(args.mono_build_path / 'bin' / 'profile' / '*.*'), str(args.output_path)])
+# process_command(['copy', str(o3de_context.project_path / 'AssetBundling' / 'Bundles' / '*.pak'), str(pc_pak_path)])
+source_path = str(args.mono_build_path / 'bin' / 'profile')
+output_path = str(args.output_path)
+process_command(f'xcopy {source_path} {output_path}', cwd=o3de_context.project_path)
--- a/scripts/ExportScripts/export_profile_monolithic.py
+++ b/scripts/ExportScripts/export_profile_monolithic.py
@@ -1,0 +1,55 @@
+from o3de.export_project import *
+from o3de import manifest
+import argparse
+import pathlib
+import os
+
+#Note: project_path and engine_path are computed before this script runs, and can be accessed via o3de_context.project_path
+#for now this script assumes an engine-centric approach, and we assume to use the source built engine for the engine_path
+
+#to use this script, go to the engine root directory, and issue the following command:
+#.\scripts\o3de.bat export-project -pp C:\path\to\project -es C:\path\to\export_profile_monolithic.py -nmp C:\path\to\build\non_mono -mp C:\path\to\build\mono -out C:\path\to\output -ll INFO
+
+parser = argparse.ArgumentParser(prog="Profile Monolithic exporter for windows",
+                                 description="Exports a profile monolithic packaged build for the project")
+parser.add_argument('-nmp', '--non-mono-build-path', type=pathlib.Path, required=True)
+parser.add_argument('-mp', '--mono-build-path',type=pathlib.Path, required=True)
+parser.add_argument('-out', '--output-path', type=pathlib.Path, required=True)
+
+args = parser.parse_args(o3de_context.args)
+
+projects_path_cmake_option = f"-DLY_PROJECTS={str(o3de_context.project_path)}"
+
+project_json_data = manifest.get_project_json_data(project_path = o3de_context.project_path)
+project_name= project_json_data.get('project_name')
+
+
+#build non-monolithic artifacts
+process_command(['cmake', '-S', '.', '-B', args.non_mono_build_path, '-DLY_MONOLITHIC_GAME=0', projects_path_cmake_option], cwd= o3de_context.engine_path)
+
+process_command(['cmake', '--build', args.non_mono_build_path, '--target', 'Editor', '--target', 'AssetBundler', '--target', 'AssetProcessor', '--target', 'AssetProcessorBatch', '--config', 'profile'], cwd=o3de_context.engine_path)
+
+executables_dir = args.non_mono_build_path / 'bin' / 'profile'
+
+asset_processor_exe = executables_dir / 'AssetProcessorBatch.exe'
+
+asset_bundler_exe = executables_dir / 'AssetBundler.exe'
+
+process_command([asset_processor_exe, '--project-path', o3de_context.project_path], cwd=o3de_context.engine_path)
+
+
+#build monolithic artifacts
+process_command(['cmake', '-S', '.', '-B', args.mono_build_path, '-DLY_MONOLITHIC_GAME=1', '-DALLOW_SETTINGS_REGISTRY_DEVELOPMENT_OVERRIDES=0', projects_path_cmake_option], cwd = o3de_context.engine_path)
+process_command(['cmake', '--build',  args.mono_build_path, '--target', f'{project_name}.GameLauncher', '--target', f'{project_name}.ServerLauncher', '--target', f'{project_name}.UnifiedLauncher', '--config', 'profile'])
+
+#bundle content (WIP)
+# process_command([asset_bundler_exe, f'--project-path="{str(o3de_context.project_path)}"'])
+
+#copy artifacts to output directory
+
+pc_pak_path = args.output_path / 'Cache' / 'pc'
+
+os.makedirs(pc_pak_path, exist_ok=True)
+# os.makedirs(args.output_path / 'Gems' / 'AWSCore')
+process_command(['copy', str(o3de_context.project_path / 'AssetBundling' / 'Bundles' / '*.pak'), str(pc_pak_path)])
+process_command(['copy', str(args.mono_build_path / 'bin' / 'profile' / '*.*'), str(args.output_path)])


scripts/ExportScripts/export_profile_monolithic.py
```diff
@@ -0,0 +1,55 @@
...
+from o3de import manifest
+import argparse
+import pathlib
+import os
+
+#Note: project_path and engine_path are computed before this script runs, and can be accessed via o3de_context.project_path
+#for now this script assumes an engine-centric approach, and we assume to use the source built engine for the engine_path
+
+#to use this script, go to the engine root directory, and issue the following command:
+#.\scripts\o3de.bat export-project -pp C:\path\to\project -es C:\path\to\export_profile_monolithic.py -nmp C:\path\to\build\non_mono -mp C:\path\to\build\mono -out C:\path\to\output -ll INFO
```
AMZN-Olex, line 11: Is there a way to place this script inside the scrtips/ folder and run it directly?

I am thinking that it would be good to break it up into smaller pieces and then combine different pieces together here.

scripts/ExportScripts/export_profile_monolithic.py
```diff
@@ -0,0 +1,55 @@
...
+args = parser.parse_args(o3de_context.args)
+
+projects_path_cmake_option = f"-DLY_PROJECTS={str(o3de_context.project_path)}"
+
+project_json_data = manifest.get_project_json_data(project_path = o3de_context.project_path)
+project_name= project_json_data.get('project_name')
+
+
+#build non-monolithic artifacts
+process_command(['cmake', '-S', '.', '-B', args.non_mono_build_path, '-DLY_MONOLITHIC_GAME=0', projects_path_cmake_option], cwd= o3de_context.engine_path)
```
lumberyard-employee-dm, line 28: nitpick: space after `cwd=`
```suggestion
process_command(['cmake', '-S', '.', '-B', args.non_mono_build_path, '-DLY_MONOLITHIC_GAME=0', projects_path_cmake_option], cwd=o3de_context.engine_path)
```

lumberyard-employee-dm: The tools can be built in a engine-centric workflow, but the GameLauncher/ServerLauncher/UnifiedLauncher artifacts must be in project-centric in order for the final executable to run.

scripts/ExportScripts/export_profile_monolithic.py
```diff
@@ -0,0 +1,55 @@
...
+parser.add_argument('-nmp', '--non-mono-build-path', type=pathlib.Path, required=True)
+parser.add_argument('-mp', '--mono-build-path',type=pathlib.Path, required=True)
+parser.add_argument('-out', '--output-path', type=pathlib.Path, required=True)
+
+args = parser.parse_args(o3de_context.args)
+
+projects_path_cmake_option = f"-DLY_PROJECTS={str(o3de_context.project_path)}"
+
+project_json_data = manifest.get_project_json_data(project_path = o3de_context.project_path)
+project_name= project_json_data.get('project_name')
```
lumberyard-employee-dm, line 24: nitpick missing space before `=`
```suggestion
project_name = project_json_data.get('project_name')
```

scripts/ExportScripts/export_profile_monolithic.py
lumberyard-employee-dm: I think @AMZN-Olex means to run it directly using python.
Such as
```powershell
D:\o3de>python\python.bat scripts\ExportScripts\export_profile_monolithic.py ...
```

lumberyard-employee-dm: I know @AMZN-alexpete recommend in an earlier PR that the `__main__` logic be removed from the export_project.py script, but all of our other o3de CLI scripts were setup to be run in a standalone fashion
For example the user can run the register.py script without going through the o3de.py driver (https://github.com/o3de/o3de/blob/development/scripts/o3de/o3de/register.py#L1053-L1075).

```
D:\o3de>python\python.bat scripts\o3de\register.py --this-engine
```

The only caveat with running the scripts in a standalone fashion, is that it requires to use the python that is installed inside of `<engine-root>\python` folder as we register the scripts\o3de package as a site-package as well as other requirements.

scripts/ExportScripts/export_profile_monolithic.py
lumberyard-employee-dm: The command is actually easy to adjust in that case, by specifying the build paths as relative to the project directory.

Also the `LY_MONOLITHIC_GAME=0` option is not needed since it is the default.
```suggestion
process_command(['cmake', '-S', f'{o3de_context.project_path}', '-B', f'{o3de_context.project_path}/{args.non_mono_build_path}'], cwd=o3de_context.project_path)
```

scripts/ExportScripts/export_profile_monolithic.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+#build non-monolithic artifacts
+process_command(['cmake', '-S', '.', '-B', args.non_mono_build_path, '-DLY_MONOLITHIC_GAME=0', projects_path_cmake_option], cwd= o3de_context.engine_path)
+
+process_command(['cmake', '--build', args.non_mono_build_path, '--target', 'Editor', '--target', 'AssetBundler', '--target', 'AssetProcessor', '--target', 'AssetProcessorBatch', '--config', 'profile'], cwd=o3de_context.engine_path)
```
lumberyard-employee-dm, line 30: The CMake targets can be space or comma separated, so the --target option doesn't need to be continually specified.
Furthermore building the Editor builds the AssetProcessor as it is a [runtime dependency](https://github.com/o3de/o3de/blob/development/Code/Editor/CMakeLists.txt#L187) of the Editor.

Therefore on the targets of Editor and AssetBundler needs to be specified.

```suggestion
process_command(['cmake', '--build', args.non_mono_build_path, '--target', 'Editor', 'AssetBundler' '--config', 'profile'], cwd=o3de_context.engine_path)
```

scripts/ExportScripts/export_profile_monolithic.py
```diff
@@ -0,0 +1,55 @@
...
+
+executables_dir = args.non_mono_build_path / 'bin' / 'profile'
```
lumberyard-employee-dm, line 32: I recommend replacing each instance of the the `profile` configuration, with an argparse option that allows users to specify. In fact we should replace it with two options.
One for non-monolithic builds and the other for monolithic.

We can have the default be `profile` in both cases
```python
parser.add_argument('-nmc', '--non-mono-config', type=str, default='profile', help='specifies the build configuration to use for the non-monolithic build of the tool artifacts')
parser.add_argument('-mc', '--mono-config', type=str, default='profile', help='specifies the build configuration to use for the monolithic build of the launcher artifacts')
```

scripts/ExportScripts/export_profile_monolithic.py
```diff
@@ -0,0 +1,55 @@
...
+
+asset_processor_exe = executables_dir / 'AssetProcessorBatch.exe'
+
+asset_bundler_exe = executables_dir / 'AssetBundler.exe'
+
+process_command([asset_processor_exe, '--project-path', o3de_context.project_path], cwd=o3de_context.engine_path)
+
+
+#build monolithic artifacts
+process_command(['cmake', '-S', '.', '-B', args.mono_build_path, '-DLY_MONOLITHIC_GAME=1', '-DALLOW_SETTINGS_REGISTRY_DEVELOPMENT_OVERRIDES=0', projects_path_cmake_option], cwd = o3de_context.engine_path)
```
lumberyard-employee-dm, line 42: You may also want to configure using the [LY_ARCHIVE_FILE_SEARCH_MODE](https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/AzFramework/feature_options.cmake#L9-L13) option set to `2`, to have the profile launchers replicate how release configuration load assets only from pak files.

scripts/ExportScripts/export_profile_monolithic.py
```diff
@@ -0,0 +1,55 @@
+from o3de.export_project import *
```
lumberyard-employee-dm, line 1: This script is missing the O3DE copyright

scripts/ExportScripts/export_profile_monolithic.py
```diff
@@ -0,0 +1,55 @@
...
+from o3de import manifest
+import argparse
+import pathlib
+import os
+
+#Note: project_path and engine_path are computed before this script runs, and can be accessed via o3de_context.project_path
+#for now this script assumes an engine-centric approach, and we assume to use the source built engine for the engine_path
+
```
lumberyard-employee-dm, line 9: I recommend moving this script to under a Platform/Windows folder since it is hardcoded to only work for windows.
There is however a way to generalize it.

The `asset-type` would need to be user configurable, so that they could select between `pc` (Windows), `linux` or `mac` assets

The asset platform option could be an argparse option as well that uses the appropriate default value for the platform.

For example
```python
default_asset_type = 'pc'
if sys.platform == 'win32':
    default_asset_type = 'pc'
elif sys.platform.startswith('linux'):
    default_asset_type = 'linux'
elif sys.platform.startswith('darwin'):
    default_asset_type = 'mac'
else:
    o3de_context.logger.error(f'Export script for platform {sys.platform} does not have a default asset type.' \
        " Using 'pc' as a fallback.")

parser.add_argument('-at, '--asset-type', type=str, default=default_asset_type)
```

Also with the `export_project.py` script, I recommend passing the logger through the context parameters, so that users can log info, warning and error messages

That is just a simple change of adding a logger property to the `O3DEScriptExportContext`
```python
def _run_export_script(args: argparse, passthru_args: list) -> int:
    logging.basicConfig(format=utils.LOG_FORMAT)
    logger = logging.getLogger()
    logger.setLevel(args.log_level)

    return _export_script(args.export_script, args.project_path, passthru_args, logger)
```

scripts/ExportScripts/export_profile_monolithic.py
```diff
@@ -0,0 +1,55 @@
...
+process_command(['cmake', '--build', args.non_mono_build_path, '--target', 'Editor', '--target', 'AssetBundler', '--target', 'AssetProcessor', '--target', 'AssetProcessorBatch', '--config', 'profile'], cwd=o3de_context.engine_path)
+
+executables_dir = args.non_mono_build_path / 'bin' / 'profile'
+
+asset_processor_exe = executables_dir / 'AssetProcessorBatch.exe'
+
+asset_bundler_exe = executables_dir / 'AssetBundler.exe'
+
+process_command([asset_processor_exe, '--project-path', o3de_context.project_path], cwd=o3de_context.engine_path)
```
lumberyard-employee-dm, line 38: The AssetProcessor can forward any `--regset` option values to the aggregate `bootstrap.game.*.setreg` files it builds.

The Command Line Argument documentation goes over the `--regset` option [here](https://www.o3de.org/docs/user-guide/settings/command-line-arguments/#command-line-support)

What this means, is that, you can use that option to inject into `bootstrap.game.<config>.setreg` via the [SetregBuilder](https://github.com/aws-lumberyard-dev/o3de/blob/ProjectExport/DraftReviewForMonolithicExportScript/Code/Tools/AssetProcessor/native/InternalBuilders/SettingsRegistryBuilder.cpp#L267-L276) any [Console Commands](https://www.o3de.org/docs/user-guide/settings/issue-az-console-commands/) that a user might need to run the application in a standalone.
For example the `bg_ConnectToAssetProcessor=false` option can be injected as follows to prevent the AssetProcessor from starting up on launch of the `${projectName}.GameLauncher`

```suggestion
process_command([asset_processor_exe, '--project-path', o3de_context.project_path, '--regset="/O3DE/Autoexec/ConsoleCommands/bg_ConnectToAssetProcessor=false"'], cwd=o3de_context.engine_path)
```

The generated `bootstrap.game.<config>.setreg` file in the `Cache/<asset-type>` folder for the project would then contain that setting.

scripts/ExportScripts/export_profile_monolithic.py
```diff
@@ -0,0 +1,55 @@
...
+#build non-monolithic artifacts
+process_command(['cmake', '-S', '.', '-B', args.non_mono_build_path, '-DLY_MONOLITHIC_GAME=0', projects_path_cmake_option], cwd= o3de_context.engine_path)
+
+process_command(['cmake', '--build', args.non_mono_build_path, '--target', 'Editor', '--target', 'AssetBundler', '--target', 'AssetProcessor', '--target', 'AssetProcessorBatch', '--config', 'profile'], cwd=o3de_context.engine_path)
+
+executables_dir = args.non_mono_build_path / 'bin' / 'profile'
+
+asset_processor_exe = executables_dir / 'AssetProcessorBatch.exe'
+
+asset_bundler_exe = executables_dir / 'AssetBundler.exe'
```
lumberyard-employee-dm, line 36: Removing the `.exe` would help to allow this script to be cross-platform by allowing it to run on Linux and Mac.

scripts/ExportScripts/export_profile_monolithic.py
```diff
@@ -0,0 +1,55 @@
...
+
+#bundle content (WIP)
+# process_command([asset_bundler_exe, f'--project-path="{str(o3de_context.project_path)}"'])
+
+#copy artifacts to output directory
+
+pc_pak_path = args.output_path / 'Cache' / 'pc'
+
+os.makedirs(pc_pak_path, exist_ok=True)
+# os.makedirs(args.output_path / 'Gems' / 'AWSCore')
```
lumberyard-employee-dm, line 53: Hardcoded commented code should be removed.

scripts/ExportScripts/export_profile_monolithic.py
```diff
@@ -0,0 +1,55 @@
...
+process_command(['copy', str(o3de_context.project_path / 'AssetBundling' / 'Bundles' / '*.pak'), str(pc_pak_path)])
+process_command(['copy', str(args.mono_build_path / 'bin' / 'profile' / '*.*'), str(args.output_path)])
```
lumberyard-employee-dm, line 55: The CMake install command should be used for this, as that will also fixup any `rpaths` on Linux and Mac needed to load any shared library(.so)/dynamic library(.dylib) in a standalone fashion.
Furthermore it fixes the file permissions as well, if some files need to be marked with the executable bit or read-only, etc...

It is also more portable as it doesn't assume that the CMAKE_RUNTIME_OUTPUT_DIRECTORY is always set to `<build-dir>/bin/<config>`
```suggestion
process_command(['cmake', '--install', f'{args.mono_build_path}', '--prefix', f'{args.output_path}', '--config', args.mono_config])
```

scripts/ExportScripts/export_profile_monolithic.py
tkothadev: would AssetProcessorBatch also be a runtime dependency of the editor?

scripts/ExportScripts/export_profile_monolithic.py
tkothadev: agreed, that's a better way to handle this

scripts/ExportScripts/export_profile_monolithic.py
```diff
@@ -0,0 +1,55 @@
...
+asset_processor_exe = executables_dir / 'AssetProcessorBatch.exe'
+
+asset_bundler_exe = executables_dir / 'AssetBundler.exe'
+
+process_command([asset_processor_exe, '--project-path', o3de_context.project_path], cwd=o3de_context.engine_path)
+
+
+#build monolithic artifacts
+process_command(['cmake', '-S', '.', '-B', args.mono_build_path, '-DLY_MONOLITHIC_GAME=1', '-DALLOW_SETTINGS_REGISTRY_DEVELOPMENT_OVERRIDES=0', projects_path_cmake_option], cwd = o3de_context.engine_path)
+process_command(['cmake', '--build',  args.mono_build_path, '--target', f'{project_name}.GameLauncher', '--target', f'{project_name}.ServerLauncher', '--target', f'{project_name}.UnifiedLauncher', '--config', 'profile'])
```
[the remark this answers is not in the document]
tkothadev: Modified the code to do the validation for `project_name` earlier in the script

scripts/ExportScripts/export_profile_monolithic.py
tkothadev: in case there are hidden gotchas with porting to linux, I'll opt for moving to a platform specific folder for windows

tkothadev: also we do provide the logger, but that is passed in to the user's script as `o3de_logger`:
https://github.com/o3de/o3de/blob/7b4fe7cd8ce0341376413283f15a4c40e08361ec/scripts/o3de/o3de/export_project.py#L82

scripts/ExportScripts/export_profile_monolithic.py
tkothadev: is it recommended to use this approach for specifying the start level to load in the regset as well? I'm tempted to do it that way, and require a parameter from the user, but it's difficult to tell if that's necessary, or just problematic, especially if the user has custom loading logic.

scripts/ExportScripts/export_profile_monolithic.py
lumberyard-employee-dm: No, only the [AssetProcessor](https://github.com/o3de/o3de/blob/development/Code/Editor/CMakeLists.txt#L185) is a runtime dependency of the Editor

scripts/ExportScripts/export_profile_monolithic.py
lumberyard-employee-dm: I didn't know the user script had access to the logger. That is good.

The spacing is however inconsistent with that call to load_and_execute_script. There is whitespace around the equal sign of the `o3de_context` argument but not around the `o3de_logger` argument.
```python
    return utils.load_and_execute_script(target_script_path, o3de_context=o3de_context, o3de_logger=logging.getLogger())
```

scripts/ExportScripts/export_profile_monolithic.py
```diff
@@ -43,13 +47,16 @@
...
 
 os.makedirs(pc_pak_path, exist_ok=True)
+os.makedirs(args.output_path, exist_ok=True)
 # os.makedirs(args.output_path / 'Gems' / 'AWSCore')
-process_command(['copy', str(o3de_context.project_path / 'AssetBundling' / 'Bundles' / '*.pak'), str(pc_pak_path)])
-process_command(['copy', str(args.mono_build_path / 'bin' / 'profile' / '*.*'), str(args.output_path)])
+# process_command(['copy', str(o3de_context.project_path / 'AssetBundling' / 'Bundles' / '*.pak'), str(pc_pak_path)])
+source_path = str(args.mono_build_path / 'bin' / 'profile')
+output_path = str(args.output_path)
+process_command(f'xcopy {source_path} {output_path}', cwd=o3de_context.project_path)
```
lumberyard-employee-dm, line 62: There is no reason to make this code less portable.

CMake CLI has a [copy](https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-E-arg-copy) command that works across all platforms.
It can be invoked via the `cmake -E copy_if_different` or `cmake -E copy_directory_if_different` command

PR #15957, opened 2023-05-14
Updated the Mac Bundling logic to fix the rpath of dylibs copied inside bundles

lumberyard-employee-dm:
The dylibs inside of the bundle did not have their rpaths fixed and therefore could attempt to load statically dylibs such as `libO3DEKernel.dylib` from a different location.

This was causing an issue in the AssetBuilder that is inside of the AssetProcessor.app bundle on Mac where the AssetBuilder would load the `libO3DEKernel.dylib` from the `@executable_path`, but any gems that loaded would use the `libO3DEKernel.dylib` from the `@executable_path/../Frameworks` directory which would result in multiple copies being in memory.

As the libO3DEKernel.dylib is responsible for being containing the symbols for the single copy of the `AZ::Environment` inside the executable symbol namespace, having two copies causes two `AZ::Environment` symbols to be available.

When the AssetBuilder executable namespace accessed the SystemAllocator for the first time it, used the `AZ::Environment`  in the libO3DEKernel.dylib loaded from the `@executable_path`, but when the `libCamera.Editor.dylib` exectuable namespace accessed the SystemAllocator it used the `AZ::Environment` in the libO3DEKernel.dylib loaded from the `@executable_path/../Frameworks`. Since there were two environments available, separate System Allocators instancewere stored in each. Each System Allocator contains an HphaScehma that is used to allocate/deallocate memory.

What would happen on start-up of the AssetBuilder launched from the AssetProcessor.app bundle would be that it would use the SystemAllocator from one `libO3DEKernel.dylib` to allocate the memory for the [BehaviorContext::m_checksByOperations](https://github.com/o3de/o3de/blob/018f14991c4d849a74c8a6cccf1a7cefecb4960a/Code/Framework/AzCore/AzCore/RTTI/BehaviorContext.h#L1566) unordered_map.

Eventually the Gem module of `libCamera.Editor.dylib` would add elements to that unordered_map, which would cause a rehash to occur. When a rehash occurs it attenots to deallocate memory from the unordered_map bucket vector. However the deallocation fails due to attempting to look for it, in the SystemAllocator of the other loaded `libO3DEKernel.dylib`

## How was this PR tested?

Built the AtomSampleViewer in a project-centric workflow.
Afterwards launched the AssetProcessor.app bundle and verified that the builders loaded successfully


moudgils (approved):

lumberyard-employee-dm:
I found out there is some issues with this fix, due to use having multiple copies of dylibs such as `libSceneData.dylib` and `libSceneCore.dylib` in both the `AssetProcessor.app/Contents/Frameworks` folder and the `AssetPRocessor.app/Contents/MacOS` folder.

I believe we need to fix where we place dylib content in the bundle by conforming to this layout specified in the Apple Developer documentation https://developer.apple.com/documentation/bundleresources/placing_content_in_a_bundle#3875930

I believe we need to place all of our dylibs inside the `Frameworks` folder only to avoid the duplicate copies

lawsonamzn (approved):

cgalvan (approved):

PR #15961, opened 2023-05-15
Add method for resetting property control to default

### LINKED ISSUE #15718: DPE Inspector: Selecting an entity with the Camera component causes subsequent additions of the Altitude Gradient component to fail Undo operations
**Describe the bug**
Selecting an entity with the Camera component causes subsequent additions of the Altitude Gradient component to fail Undo operations.

**Steps to reproduce**
```
Editor.exe --ed_enableDPE=true
```
2. Create a new level.
3. Expand the Default Level Environment entity hierarchy.
4. Select the "Camera" entity.
5. Create a new entity.
6. Add an Altitude Gradient component to the new entity.
7. Attempt to Undo the component addition.

**Expected behavior**
Component is removed via Undo.

**Actual behavior**
Undo fails on the Component. In the Console, the following error is shown:

```
Cannot Undo while Recording
```

**Found in Branch**
development

0fa6eea55514ef710fa4a0d40ea31ad43b181de9

michabr:
## What does this PR do?

Fixes an issue when recycling DPE controls. The DPE reuses property controls, so old settings from prior usage carry over. This was causing an issue when setting the value of a double spin box when its min/max setting was incorrect. This PR adds a method for controls to be able to reset to their default values before being reused. Currently only min/max/step attributes of the spin box control have been reset in this PR, but a follow up PR should be created to cover all attributes/controls.

More detailed info:
The double spinbox's setValue() function checks to see if the value will be clamped per the spinbox's min/max properties. If so, it queues up a "valueChanged" event. When DPE does a "readFromDom", setValue() will be called (it doesn't expect the value to be internally modified at this point). However, if min/max values were previously set, and the spinbox is being reused, the min/max values won't be accurate and can result in the "ValueChanged" message being sent. That triggers an undo node to be added to the undo stack. In this particular case, doing an "undo" undid the change but also added a new undo node to the stack therefore blocking ability to undo deeper into the stack.

RPE does not have this issue because the property controls don't seem to get recycled and are recreated each time.

Fixes #15718.

## How was this PR tested?

Testing in Editor using repro steps.


Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyDoubleSpinCtrl.cpp
```diff
@@ -85,9 +85,9 @@ namespace AzToolsFramework
         m_pSpinBox->setValue(value);
         m_pSpinBox->blockSignals(false);
 
-        //the value didn't change, so don't send a notify
-        if (!AZ::ClampIfCloseMag(value, oldValue) && notifyLater)
```
forkercat, line 89: Do you know why a value is evaluated as a condition here?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyDoubleSpinCtrl.cpp
```diff
@@ -85,9 +85,9 @@ namespace AzToolsFramework
...
+        if (notifyLater)
```
alexmontAmazon, line 88: Why isn't the AZ::ClampIfCloseMag necessary here anymore?

alexmontAmazon (approved):
looks good, but the amount of controls for which we'd need to implement this makes me nervous. Maybe we should just disable handler recycling for now?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyEditorAPI_Internals.h
```diff
@@ -326,6 +327,12 @@ namespace AzToolsFramework
                 }
             }
 
+            if (m_widget)
+            {
+                // Reset existing widget attributes and properties before reuse
+                m_rpeHandler.ResetGUIToDefaults_Internal(m_widget);
+            }
```
forkercat, line 334: Does this mean the new reset logic only applicable to DPE workflow?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyEditorAPI_Internals.h
```diff
@@ -326,6 +327,12 @@ namespace AzToolsFramework
                 }
             }
 
+            if (m_widget)
```
forkercat, line 330: If I understood right, the reason we don't call `GetWidget()` here is bc we don't need to do reset if the widget is not created?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyDoubleSpinCtrl.cpp
michabr: I couldn't figure out the purpose of `!AZ::ClampIfCloseMag(value, oldValue)`.

ClampIfCloseMag returns either x or y so it didn't make sense as a conditional.
`//! ClampIfCloseMag(x, y, epsilon) returns y when x and y are within epsilon of each other`

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyDoubleSpinCtrl.cpp
michabr: ClampIfCloseMag returns either value or oldValue so I didn't see why it would be part of the conditional...
`//! ClampIfCloseMag(x, y, epsilon) returns y when x and y are within epsilon of each other (taking magnitude into account).  Otherwise returns x.`

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyEditorAPI_Internals.h
michabr: Yes, it only applies to DPE.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyEditorAPI_Internals.h
michabr: Yes, correct.

michabr:
> looks good, but the amount of controls for which we'd need to implement this makes me nervous. Maybe we should just disable handler recycling for now?

Do we have any info on the savings we get by recycling?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyDoubleSpinCtrl.cpp
forkercat: My hunch is it should be, but maybe we can keep it unchanged in this PR?

```cpp
if (!AZ::IsCloseMag(value, oldValue) && notifyLater)
{
    ...
}
```

forkercat (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyDoubleSpinCtrl.cpp
```diff
@@ -341,6 +341,20 @@ namespace AzToolsFramework
...
+    {
+        GUI->setMinimum(-DBL_MAX);
+        GUI->setMaximum(DBL_MAX);
+        GUI->setStep(1);
+    }
+
+    void floatPropertySpinboxHandler::ResetGUIToDefaults(PropertyDoubleSpinCtrl* GUI)
+    {
+        GUI->setMinimum(std::numeric_limits<float>::lowest());
+        GUI->setMaximum(std::numeric_limits<float>::max());
```
spham-amzn, line 354:
```suggestion
        GUI->setMinimum(AZStd::numeric_limits<float>::lowest());
        GUI->setMaximum(AZStd::numeric_limits<float>::max());
```

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyDoubleSpinCtrl.cpp
```diff
@@ -341,6 +341,20 @@ namespace AzToolsFramework
         return newCtrl;
     }
 
+    void doublePropertySpinboxHandler::ResetGUIToDefaults(PropertyDoubleSpinCtrl* GUI)
+    {
+        GUI->setMinimum(-DBL_MAX);
+        GUI->setMaximum(DBL_MAX);
```
spham-amzn, line 347: Why not use `AZStd::numeric_limits<double>::lowest()` and `AZStd::numeric_limits<double>::max()` here as well?

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyEditorAPI_Internals.h
```diff
@@ -405,6 +412,12 @@ namespace AzToolsFramework
     public:
         typedef WidgetType widget_t;
 
+        // resets widget attributes and properties for reuse
+        virtual void ResetGUIToDefaults(WidgetType* widget)
+        {
+            (void)widget;
+        }
```
spham-amzn, line 419: Could also do 
```suggestion
        virtual void ResetGUIToDefaults([[maybe_used]] WidgetType* widget)
        {
            (void)widget;
        }
```

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyDoubleSpinCtrl.cpp
michabr: Cuz that's what the original code had and I copied/pasted it :). I can change it to AZStd::numeric_limits.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyDoubleSpinCtrl.cpp
michabr: I ended up just leaving it how it originally was per @forkercat's suggestion. It looks wrong, but safer to leave it as is.

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyEditorAPI_Internals.h
michabr: In this case, I tried to keep the file consistent, but I can change it to our newer convention.

spham-amzn (approved):

lawsonamzn (approved):

PR #15964, opened 2023-05-15
Fix Atom buffer crashes due to invalid pointers

mbalfour-amzn:
## What does this PR do?

This fixes intermittent crashes that could occur in `AZ::RPI::Buffer::UpdateData(const void * sourceData, unsigned __int64 sourceDataSize, unsigned __int64 bufferByteOffset)` from mysteriously invalid pointers.

The root problem was that BufferPool was expanding the size of the m_uploadPackets vector and grabbing a reference into it inside a lock, but then using it outside the lock to call Map(). If expanding the size caused m_uploadPackets to get reallocated, then any other threads that were outside the lock would have an invalid uploadRequest reference, leading to chaos and crashes.

The fix is to declare uploadRequest as a local variable, call Map() outside the lock, and then add uploadRequest into the vector inside the lock. By not holding references to the vector anywhere, the problem no longer occurs.

## How was this PR tested?

Ran MultiplayerSample with both DX12 and Vulkan and verified that everything appeared normal and the crash didn't occur.

### DIFF
--- a/Gems/Atom/RHI/DX12/Code/Source/RHI/BufferPool.cpp
+++ b/Gems/Atom/RHI/DX12/Code/Source/RHI/BufferPool.cpp
@@ -56,12 +56,12 @@ namespace AZ
                 AZ_PROFILE_FUNCTION(RHI);
 
                 MemoryView stagingMemory = m_device->AcquireStagingMemory(request.m_byteCount, Alignment::Buffer);
+                if (!stagingMemory.IsValid())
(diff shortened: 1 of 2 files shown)


moudgils (approved):
Nice fix. I will address this in a different PR for Metal when I am back on Mac.

VickyAtAZ (approved):

cgalvan (approved):

PR #15984, opened 2023-05-17
Updated o3de CLI register command to generate a CMakePresets file for project registration

lumberyard-employee-dm:
When a project is registered a
`<project-root>/user/cmake/engine/CMakePresets.json` will be added/updated which contains an "include" entry to the registered engine's ``CMakePresets.json` file

The Default and Minimal Project templates have had their `<project-root>/CMakePresets.json` file to include the generated `"user/cmake/engine/CMakePresets.json"` file.

This means that new projects will now opt-in to this functionality by default.

Added a CMakePresets.json file for the AutomatedTesting project that includes the file in the CMakePresets.json it's engine root directory.

Furthermore running cmake configfure will also update the `<project-root>/user/cmake/engine/CMakePresets.json` file to include the CMakePresets.json file using the path of the current engine being configured.

This does open up the possibility of the user being surprised by an unintended modification to use a different engine presets however.

If the user has a project called "projectA" registered with engine called `o3deA` and then decides to use engine `o3deB` in an engine-centric workflow to configure "projectA" via the `LY_PROJECTS` Cache Variable, then what would occur is that the "<projectA>/user/cmake/engine/CMakePresets.json" will be updated to point at `o3deB` CMakePresets.json even though "projectA" is registers `o3deA`.

The mitigation for this case to have "projectA" to include `o3deA` presets again are to just configure the project in a project-centric workflow, which would use the engine `o3deA` again.

## How was this PR tested?

Added unit test for validating creation of the `CMakePresets-engine.json` command.

Verified that registering a project generates the `<project-root>/user/cmake/engine/CMakePresets.json` file.

Verified that registration of an existing project, does generate a `<project-root>/user/cmake/engine/CMakePresets.json` file and **does not** modify the existing `<project-root>/CMakePresets.json` file

This means that existing projects will not have to worry about their existing CMake presets being invalidated with this change.

image

### DIFF
--- a/AutomatedTesting/CMakePresets.json
+++ b/AutomatedTesting/CMakePresets.json
@@ -6,6 +6,6 @@
         "patch": 0
     },
     "include": [
-        "user/cmake/engine/CMakePresets.json"
+        "../CMakePresets.json"
     ]
 }
--- a/scripts/o3de/tests/test_cmake.py
+++ b/scripts/o3de/tests/test_cmake.py
@@ -9,6 +9,7 @@
 import io
 import pytest
 import pathlib
+import string
 from unittest.mock import patch
 
 from o3de import cmake, manifest
@@ -143,7 +144,7 @@ class TestResolveGemDependencyPaths:
                     {"gem_name":"gemA", "version":"1.2.3", "path":pathlib.Path('gemA1Path')},
                     {"gem_name":"gemA", "version":"2.3.4", "path":pathlib.Path('gemA2Path')},
                 ]
-            }, 'gemA;gemA1Path', 0) 
+            }, 'gemA;gemA1Path', 0)
         ]
     )
     def test_resolve_gem_dependency_paths(self, engine_gem_names, project_gem_names, all_gems_json_data, expected_gem_paths, expected_result):
@@ -161,7 +162,7 @@ class TestResolveGemDependencyPaths:
             return {
                 "engine_name":"o3de",
                 "gem_names": engine_gem_names
-            } 
+            }
 
         def get_project_json_data(project_name: str = None,
                                 project_path: str or pathlib.Path = None,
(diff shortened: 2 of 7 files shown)


scripts/o3de/tests/test_cmake.py
```diff
@@ -205,10 +206,71 @@ def get_gems_json_data_by_name(engine_path:pathlib.Path = None,
...
+                return super().__enter__()
+            def __exit__(self, exc_type, exc_val, exc_tb):
+                nonlocal preset_json_content
+                preset_json_content = super().getvalue()
+                super().__exit__(exc_tb, exc_val, exc_tb)
+
+        def preset_json_open(mode):
+            if preset_path:
+                return StringBufferIOWrapper()
+            raise OSError('Cannot Open file "{preset_path}"')
```
AMZN-alexpete, line 258: I wonder if it's worth including `preset_path` as a value here because if `preset_path` is non-empty won't it not reach this line?

AMZN-alexpete (approved):
nice! will likely want to update our docs at some point after this goes in to mention this register step

AutomatedTesting/CMakePresets.json
```diff
@@ -0,0 +1,11 @@
+{
```
tkothadev, line 1: How does this affect projects made with the engine before the changes in this PR?

Will compilation break for projects that are missing a CMakePresets.json file?

AMZN-alexpete (approved):

AutomatedTesting/CMakePresets.json
lumberyard-employee-dm: This information is in the __How was this PR Tested?__ section in the description.

scripts/o3de/tests/test_cmake.py
lumberyard-employee-dm: Well the message itself isn't actually import.
The only thing need was to raise an OSError so that the UnitTest replicates being unable to open the presets file.
It could be an OSError with no message.

AMZN-alexpete (approved):

tkothadev (approved):

Issue #15546, opened 2023-04-04
Empty type-name component alias for pending and disabled components


forkercat:
**Describe the bug**

The issue was discovered in a DPE inspector issue https://github.com/o3de/o3de/issues/15509 where PrefabComponentAdapter asserts on empty component alias.

The component alias is empty after these operations:
1. Add a pending component that has dependency on other components that are missing
2. First loading a prefab with entity that has disabled components (e.g. loading a level, instantiate-prefab, etc)
3. First loading a prefab with entity that has pending components (same)

**Steps to reproduce**
1. Create an empty level
3. Create an entity
4. Disabled Component - Add any component and disable the component
5. Pending Component - Add 'Vegetable Surface Mask Filter' or 'Audio Animation' component which has dependency on components that are not available in the entity.

**Found in Branch**

Development https://github.com/o3de/o3de/commit/ec299f7b6c27f03e48a7b0a9788637c23d90b598

**Cause**

After the PR https://github.com/o3de/o3de/pull/15256 that introduces type-name component alias, a new function `OnPrepareForAdditionToEntity` is used to set up the component alias before the entity is added to the entity. It is added right after `Component::SetEntity`. However, it is not called in the following places:
1. [EditorEntityActionComponent::AddExistingComponentsToEntityById()](https://sourcegraph.com/github.com/o3de/o3de/-/blob/Code/Framework/AzToolsFramework/AzToolsFramework/Entity/EditorEntityActionComponent.cpp?L695)
2. [EditorPendingCompositionComponent::Init()](https://sourcegraph.com/github.com/o3de/o3de/-/blob/Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorPendingCompositionComponent.cpp?L96:14)
3. [EditorDisabledCompositionComponent::Init()](https://sourcegraph.com/github.com/o3de/o3de/-/blob/Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorDisabledCompositionComponent.cpp?L83:14)

**Question 1 - Should we keep track of component alias in these two editor composition components?**

Yes, we should. Before the type-name change, the component can be tracked in the composition components. Basically, we set component alias in `SetEntity` based on the component id for each component stored in the composition component.

We need to continue keeping track of the component alias to prevent issue of unexpected component alias change. Override patches are dependent on the invariant component alias, including the suffix number.

**Question 2 - Why can't we just add additional missing OnPrepareForAdditionToEntity calls?**

[...]

PR #15985, opened 2023-05-18
Handle component aliases in editor composition components

forkercat:
## What does this PR do?

Fixes empty component alias issue: https://github.com/o3de/o3de/issues/15546

This PR aims to fix the empty component alias issue that happens to components stored in pending and disabled composition components. The issue started to occur after we changed to use component type names as component aliases in PR https://github.com/o3de/o3de/pull/15256.

A non-empty component alias for a component stored in the composition component is required for ensuring correctness of override patching because patches have paths that contain component aliases. They can't be empty. Note that before using component type name as aliases, there was no issue since we used component ids as the aliases. To learn more, see the attached GHI.

**Change List:**
1. Changed to use map (not vector) to store component data, which enables us storing aliases in the serialized data.
1. Updated functions in the two composition components to work with the map and component aliases.
1. Added to set up the component alias for a new pending component in `EditorEntityActionComponent`.
1. Changed `OnPrepareForAdditionToEntity` to `OnAfterEntitySet` to indicate more clearly that it is called after 'entity is set', not 'added to entity'.
1. Improved component type name alias generation in `OnAfterEntitySet` and Changed to reuse unused suffix count.
1. Added custom serializers for composition components to support reading array type and object type data.

```cpp
// Before - EditorPendingCompositionComponent
"PendingComponents": [                                  // vector (array)
    {
        "$type": "EditorSurfaceMaskFilterComponent",
        "Id": 7435484230423003242
    }
]

// After - EditorPendingCompositionComponent
"PendingComponents": {                                  // map (object)
    "EditorSurfaceMaskFilterComopnent": {               // component alias
        "$type": "EditorSurfaceMaskFilterComponent",
        "Id": 7435484230423003242
    }
}
```

## How was this PR tested?

**Tested in editor**

Pending composition component:
1. Added a new component that is pending (e.g. `EditorSurfaceMaskFilterComponent`); Verified it has valid type name alias and could be saved to file; Reloaded the level and verified the same alias could be loaded from file.
1. Assuming there are multiple components of the same type (e.g. `EditorSurfaceMaskFilterComponent`, `_2`, `_3`), deleted the second one and added a new one; Verified `_2` would be reused, while previously it was `_4`.

Disabled composition component:
1. Same as above.
1. Assuming adding two components of the same type (e.g. `EditorCommentComponent`, `_2`), disabled the first then the second, then enabled the second then the first, and verified that they won't swap component aliases.

Other test cases:
1. Verified for RPE entity inspector, DPE inspector, and DPE inspector w/ prefab inspector override enabled. 
1. Verified that changing a pending component to disabled and back to enabled could function without issues.
1. Verified it worked for generic type component wrapper (e.g. FlyCameraInputComponent).

**Backward compatibility testing**

Added the custom serializers. I tested that it can read old array data without any issue. For example, it can read the below data and can save back to the file with the new data.

```cpp
"DisabledComponents": [                             // old array data
    {
        "$type": "EditorCommentComponent",
        "Id": 8548758530932447835,
        "Configuration": "1"
    }
]

"DisabledComponents": {                             // new map data
    "Component_[8548758530932447835]": {
        "$type": "EditorCommentComponent",
        "Id": 8548758530932447835,
        "Configuration": "1"
    }
}
```

### DIFF
--- a/Code/Framework/AzCore/AzCore/Component/Component.h
+++ b/Code/Framework/AzCore/AzCore/Component/Component.h
@@ -240,9 +240,8 @@ namespace AZ
          */
         virtual void SetEntity(Entity* entity);
 
-        //! Function to call before adding this component to the provided entity.
-        //! @param entity The entity to add the component to.
-        virtual void OnPrepareForAdditionToEntity(Entity* entity);
+        //! Function to call after setting the entity in this component.
+        virtual void OnAfterEntitySet();
 
         /**
          * Reflects the Component class.
--- a/Code/Framework/AzCore/AzCore/Component/Entity.cpp
+++ b/Code/Framework/AzCore/AzCore/Component/Entity.cpp
@@ -341,7 +341,8 @@ namespace AZ
         }
         component->SetEntity(this);
 
-        component->OnPrepareForAdditionToEntity(this);
+        component->OnAfterEntitySet();
+
         m_components.push_back(component);
 
         if (m_state == State::Init)
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorComponentBase.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorComponentBase.cpp
@@ -108,7 +108,6 @@ namespace AzToolsFramework
             captureExistingIdentifiersFunc(disabledComponents);
 
             // Generates a component typename string and updates its suffix count to avoid duplication.
-            // For example, if 'EditorCommentComponent' exists, it generates 'EditorCommentComponent_2', 'EditorCommentComponent_3', and so on.
             const AZStd::string componentTypeName = GetNameFromComponentClassData(this);
             AZStd::string serializedIdentifier = componentTypeName;
             AZ::u64 suffixOfNewComponent = 1;
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorComponentBase.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorComponentBase.cpp
@@ -53,7 +53,11 @@ namespace AzToolsFramework
 
         void EditorComponentBase::OnAfterEntitySet()
         {
-            AZ_Assert(m_entity, "OnAfterEntitySet - Entity should not be nullptr.");
+            if (!m_entity)
+            {
+                AZ_Assert(false, "OnAfterEntitySet - Entity should not be nullptr.");
+                return;
+            }
 
             // Sets up only if the component alias is empty to avoid setting the alias twice.
             // For a component added as a pending component, there are two places trying to call this function to set up the alias:
@@ -68,7 +72,11 @@ namespace AzToolsFramework
 
         AZStd::string EditorComponentBase::GenerateComponentSerializedIdentifier()
         {
-            AZ_Assert(m_entity, "GenerateComponentSerializedIdentifier - Entity should not be nullptr.");
+            if (!m_entity)
+            {
+                AZ_Assert(false, "GenerateComponentSerializedIdentifier - Entity should not be nullptr.");
+                return {};
+            }
 
             AZStd::unordered_set<AZStd::string> existingSerializedIdentifiers;
 
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorComponentBase.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorComponentBase.cpp
@@ -12,6 +12,8 @@
 #include <AzCore/Serialization/SerializeContext.h>
 #include <AzToolsFramework/API/ToolsApplicationAPI.h>
 #include <AzToolsFramework/Entity/EditorEntityHelpers.h>
+#include <AzToolsFramework/ToolsComponents/EditorDisabledCompositionBus.h>
+#include <AzToolsFramework/ToolsComponents/EditorPendingCompositionBus.h>
 
 DECLARE_EBUS_INSTANTIATION_WITH_TRAITS(AzToolsFramework::Components::EditorComponentDescriptor, AZ::ComponentDescriptorBusTraits);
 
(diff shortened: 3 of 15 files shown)


Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorComponentBase.cpp
```diff
@@ -49,45 +51,74 @@ namespace AzToolsFramework
             m_transform = nullptr;
         }
 
-        void EditorComponentBase::OnPrepareForAdditionToEntity(AZ::Entity* entity)
+        void EditorComponentBase::OnAfterEntitySet()
         {
-            if (entity && m_alias.empty())
+            AZ_Assert(m_entity, "OnAfterEntitySet - Entity should not be nullptr.");
```
lawsonamzn, line 56: is m_entity even used in this function?  Its probably not useful to assert here if thats the case.

Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorComponentBase.cpp
```diff
@@ -49,45 +51,74 @@ namespace AzToolsFramework
...
-                        return;
-                    }
-                    if (RTTI_GetType() == componentInEntity->RTTI_GetType())
-                    {
-                        serializedIdentifiersMatchingType.emplace(componentInEntity->GetSerializedIdentifier());
-                    }
-                }
+        AZStd::string EditorComponentBase::GenerateComponentSerializedIdentifier()
+        {
+            AZ_Assert(m_entity, "GenerateComponentSerializedIdentifier - Entity should not be nullptr.");
```
lawsonamzn, line 71: if m_entity is null here, this will cause a null pointer reference, and crash later.   Asserts keep going - they do not stop execution or return.

Perhaps add a bail out clause (return early if no entity?), or perhaps move this assert later down in the function to right before it uses m_entity, and maybe add a way for it to safely proceed, ie, by not calling any entity-related functions, skipping that part, so that in the worst case, the entity-related stuff doesn't get involved in making the identifer.

Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorDisabledCompositionComponent.cpp
```diff
@@ -46,39 +47,70 @@ namespace AzToolsFramework
...
+                components.insert(components.end(), pair.second);
+            }
         }
 
         void EditorDisabledCompositionComponent::AddDisabledComponent(AZ::Component* componentToAdd)
         {
-            AZ_Assert(componentToAdd, "Unable to add a disabled component that is nullptr");
-            if (componentToAdd && AZStd::find(m_disabledComponents.begin(), m_disabledComponents.end(), componentToAdd) == m_disabledComponents.end())
+            AZ_Assert(componentToAdd, "Unable to add a disabled component that is nullptr.");
+
```
lawsonamzn, line 59: Again, should we just return (after asserting) so that we don't force a crash?

lawsonamzn (approved):
In general, it looks good to me, just concerned about the choice to assert-and-continue in general.  I'm not sure what the outcome is but it seems like if we can avoid an actual crash we should?

Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorComponentBase.cpp
forkercat: I think an assert is needed here to ensure `OnAfterEntitySet` is called after `SetEntity`. But I would also add a bail out clause to return early if no entity.

forkercat:
> In general, it looks good to me, just concerned about the choice to assert-and-continue in general. I'm not sure what the outcome is but it seems like if we can avoid an actual crash we should?

Thanks for the suggestion. Yeah, I think I need to re-evaluate some of the assertion usage here and change it to return early if possible.

Code/Framework/AzCore/AzCore/Component/Entity.cpp
```diff
@@ -341,7 +341,8 @@ namespace AZ
         }
         component->SetEntity(this);
 
-        component->OnPrepareForAdditionToEntity(this);
+        component->OnAfterEntitySet();
+
```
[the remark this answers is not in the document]
forkercat: `OnPrepareForAdditionToEntity` was just added from the previous component type name PR https://github.com/o3de/o3de/pull/15256. So I think it is okay to change it now if the new name can better describe its purpose.

From what I understood, 'Add' and 'Set' have the same meaning here. But I changed to use 'Set' to better align with the `SetEntity` function. @srikappa-amzn

Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorDisabledCompositionComponent.cpp
```diff
@@ -95,16 +127,18 @@ namespace AzToolsFramework
...
             {
                 // It's possible to get null components in the list if errors occur during serialization.
                 // Guard against that case so that the code won't crash.
                 if (disabledComponent)
                 {
                     auto editorComponentBaseComponent = azrtti_cast<Components::EditorComponentBase*>(disabledComponent);
                     AZ_Assert(editorComponentBaseComponent, "Editor component does not derive from EditorComponentBase");
+
                     editorComponentBaseComponent->SetEntity(GetEntity());
+                    editorComponentBaseComponent->SetSerializedIdentifier(componentAlias);
```
forkercat, line 141: While looking at the custom serializer for composition components, I think it might be good to set up the serialized identifier over there instead and remove this line here.

Code/Framework/AzCore/AzCore/Component/Component.h
```diff
@@ -240,9 +240,8 @@ namespace AZ
          */
         virtual void SetEntity(Entity* entity);
 
-        //! Function to call before adding this component to the provided entity.
-        //! @param entity The entity to add the component to.
-        virtual void OnPrepareForAdditionToEntity(Entity* entity);
+        //! Function to call after setting the entity in this component.
+        virtual void OnAfterEntitySet();
```
AMZN-alexpete, line 244: looks like the API change wasn't included in the last release so this change should be OK - otherwise you'd want to  follow the API deprecation guidelines https://github.com/o3de/community/blob/main/guides/o3de-deprecation-guidelines.md

forkercat: Oh, I didn't notice that. thanks for checking this!

Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorPendingCompositionComponent.cpp
```diff
@@ -7,18 +7,92 @@
...
+                    static AZ::TypeId genericComponentWrapperTypeId = azrtti_typeid<GenericComponentWrapper>();
+
+                    for (auto& [componentKey, component] : componentMap)
+                    {
+                        // If the component didn't serialize (i.e. is null) or the underlying type is GenericComponentWrapper,
+                        // the template is null and the component should not be added.
+                        if (component && component->GetUnderlyingComponentType() != genericComponentWrapperTypeId)
+                        {
+                            componentInstance->m_pendingComponents.emplace(componentKey, component);
+                        }
```
AMZN-alexpete, line 71: Do we have any concept of "missing" component types that are made visible in the case where someone opens a prefab that uses a component that is defined by a gem that is missing from their project - and they still see in the entity inspector that there is a component that exists on the entity but it hasn't been registered so it is disabled.  
e.g. I send you a prefab file that contains an entity with a PopcornFX component on it, you try to open it in your AutomatedTesting project which doesn't have the PopcornFX gem, does the editor show the unknown component in the entity inspector in a disabled state - or is it invisible and if the user saves it, that component is erased?

AMZN-alexpete (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/ToolsComponents/EditorPendingCompositionComponent.cpp
forkercat: I think it would show up a dialog saying the component cannot be loaded and it is missing in the inspector. I don't think we currently treat it as a disabled component.

lawsonamzn (approved):

PR #16007, opened 2023-05-23
Fixed signature of Red-Black Tree node_handle_insert

lemonade-dm:
The `node_handle_insert_unique` and `node_handle_insert_equal` functions are expected to work with a `const_iterator` which is iterator to a constant view of the pair data and not just a `const iterator` which is a constant iterator to a non-constant view of the pair data.

## How was this PR tested?

Added unitest to validate that node extraction and re-insertion with an iterator compiles successfully.


alexmontAmazon (approved):
looks good, thank you!

lawsonamzn (approved):

cgalvan (approved):

PR #16013, opened 2023-05-24
Multiplayer: Send updated properties in delete messages

mbalfour-amzn:
## What does this PR do?

This changes the delete messages for entity replication to include any properties that have been changed since the last update. This way, we can guarantee that at the point that the entity OnDeactivate() is called, the entity has access to the correct state at the time of deletion. By providing this extra state, it becomes easier to build networked entities that perform some sort of functionality at destruction time. Without this state, there's no easy way to guarantee ordering between any desired property updates or RPCs and destruction, so destruction could potentially happen without triggering the desired destruction event sequence.

These changes are used by the Energy Balls in MultiplayerSample in a separate PR to fix a number of errors with movement and explosions.

The functional changes boil down to the following:
* NetworkEntityUpdateMessage now always contains serialized changes. Before, with delete messages it just contained a delete flag and no data.
* EntityReplicationManager specifically checks for redundant deletes and deletes that would implicitly cause a creation just to let the delete process and drops the messages. The net result is that the entity is still gone, but with less processing. Also, when processing delete messages, the "update" part of the message is processed first so that the properties are up-to-date when the delete occurs.
* EntityReplicator / PropertyPublisher will create and cache off a delete message at the point that an entity is marked for removal. This way, the changed property values will remain available for sending even after the entity has finished deleting. However, if the entity has never replicated a creation, the delete won't get cached, and it will just never get sent.

There are some refactoring changes below as well, to make the internal APIs easier to read and use.
* PropertyPublisher / PropertySubscriber APIs are now private inside of EntityReplicator instead of exposed outward through GetPropertyPublisher() / GetPropertySubscriber(). This makes it easier to control their usage.
* The API names inside of PropertyPublisher and EntityReplicator have been reworked a bit to better reflect their intent. ex: "HasChangesToPublish()" instead of "RequiresSerialization()". The goal of the classes are to publish and replicate and not just to serialize, so the API names were changed to better reflect that. These are all internal APIs, so backwards compatibility isn't a concern.
* More responsibility for generating the packets to publish has been pushed down into PropertyPublisher from EntityReplicator.  The previous division was awkward and was forcing the delete message caching to happen up at the EntityReplicator level, which didn't seem architecturally correct.

## How was this PR tested?

Tested with MultiplayerSample and the "Net_RepDeletes" logging enabled to ensure that all the entities were behaving as expected and that all of the different edge cases were being hit over the course of the run.

Also added the following unit tests:
* If an entity is deleted before the creation has been sent, then both the creation and deletion packets won't get sent.
* If an entity delete is received before the creation has been received, then the entity will never get created and the delete won't get processed.
* The delete message contains a payload with serialized data in it.
* Delete messages will continue to get resent every frame until they've been acknowledged.


Gems/Multiplayer/Code/Include/Multiplayer/NetworkEntity/EntityReplication/EntityReplicator.h
```diff
@@ -51,28 +53,47 @@ namespace Multiplayer
...
         void SetPendingRemoval(AZ::TimeMs pendingRemovalTimeMs);
-        bool IsPendingRemoval() const;
         void ClearPendingRemoval();
-        bool IsDeletionAcknowledged() const;
         bool WasMigrated() const;
         void SetWasMigrated(bool wasMigrated);
         // If an entity is part of a network hierarchy then it is only ready to activate when its direct parent entity is active.
         bool IsReadyToActivate() const;
 
+        // Interface for ReplicationManager to manage publishing entity changes
```
kberg-amzn, line 71: I absolutely love all the comments you've added to this interface, this is awesome! 👍

Gems/Multiplayer/Code/Source/NetworkEntity/EntityReplication/EntityReplicationManager.cpp
```diff
@@ -210,10 +209,10 @@ namespace Multiplayer
                         m_remoteEntitiesPendingCreation.erase(*iter);
                     }
 
-                    if (canSend && propPublisher->RequiresSerialization())
+                    if (canSend && replicator->HasChangesToPublish())
```
kberg-amzn, line 212: Your refactors here really paid off. This is extremely readable.

Gems/Multiplayer/Code/Source/NetworkEntity/EntityReplication/EntityReplicationManager.cpp
```diff
@@ -812,30 +821,65 @@ namespace Multiplayer
...
+                // By default, trust everything from the server
                 result = UpdateValidationResult::HandleMessage;
+
+                // For delete messages, we can potentially skip processing them if the entity hasn't been created or has already
+                // been deleted.
+                if (updateMessage.GetIsDelete())
+                {
+                    if (!entityReplicator)
+                    {
+                        // If the entity hasn't been created (or has already been deleted), drop the delete message.
```
kberg-amzn, line 833: Again, huge appreciation for all the comments explaining everything.

Gems/Multiplayer/Code/Source/NetworkEntity/EntityReplication/EntityReplicator.cpp
```diff
@@ -321,6 +326,11 @@ namespace Multiplayer
 
     void EntityReplicator::MarkForRemoval()
     {
+        // NOTE: It's possible that heading into this function, m_entityHandle is already invalid if this was triggered
+        // via OnEntityRemoved.
+
+        AZLOG(NET_RepDeletes, "Marking entity %llu for removal.", (AZ::u64)m_entityHandle.GetNetEntityId());
```
AMZN-Gene, line 332: Prefer C++ casts?  The AZLOGs are using c-style casts, but the other logs are using static_cast<>

AMZN-Gene (approved):

Gems/Multiplayer/Code/Source/NetworkEntity/EntityReplication/EntityReplicator.cpp
```diff
@@ -465,38 +534,76 @@ namespace Multiplayer
...
+        return m_propertyPublisher->RequiresSerialization();
+    }
 
-        InputSerializer inputSerializer(updateMessage.ModifyData().GetBuffer(), static_cast<uint32_t>(updateMessage.ModifyData().GetCapacity()));
-        m_propertyPublisher->UpdateSerialization(inputSerializer);
-        updateMessage.ModifyData().Resize(inputSerializer.GetSize());
+    void EntityReplicator::SetRebasing()
+    {
+        AZ_Assert(m_propertyPublisher, "Expected to have a property publisher");
+        m_propertyPublisher->SetRebasing();
```
kberg-amzn, line 573: Is it worth being a bit more defensive and null-guarding the actual implementation calls in case a replicator gets in a bad state? It might allow execution to continue for a time, albeit incorrectly and with unpredictable side-effects.

mbalfour-amzn: yeah, wouldn't hurt. i'll make sure they all assert as well, so that way the problem is pretty obvious if it happens. it's definitely an API misuse if something tries to use them when the publisher or subscriber is null, but since it's state dependent and lifetime dependent, it's easy to see how someone could accidentally misuse the API. Guarded!

Gems/Multiplayer/Code/Source/NetworkEntity/EntityReplication/PropertyPublisher.cpp
```diff
@@ -75,72 +76,83 @@ namespace Multiplayer
...
+        // Only update the pending record if we don't have a cached delete message already.
+        if (!m_cachedDeleteMessage.GetIsDelete())
+        {
+            AZ_Assert(netBindComponent, "NetBindComponent is nullptr");
+            netBindComponent->FillReplicationRecord(m_pendingRecord);
+        }
     }
 
-    bool PropertyPublisher::HasUpdateEntityRecord()
+    bool PropertyPublisher::HasEntityChangesToSend()
```
kberg-amzn, line 89: So as I understand the architecture currently, we have an entity, and that entity has a number of attached replicators. Say for a server with 5 connected clients that can all see the entity, that would be 5 replicators. For each replicator for each frame we're keeping a bitset of properties that represent which properties were modified on that frame. That means for 5 replicators we'd basically have a history of 5 identical bitsets, one copy for each replicator. The only difference being how many entries we keep for each replicator, which depends on the last acked update for each.

mbalfour-amzn: More specifically, we're keeping a bitset of properties with _unacknowledged_ modifications. If the 5 connected clients all acknowledge at slightly different times, I believe the bitsets can vary from each other on any given frame.

Gems/Multiplayer/Code/Source/NetworkEntity/EntityReplication/PropertyPublisher.cpp
```diff
@@ -75,72 +76,83 @@ namespace Multiplayer
...
 
-    bool PropertyPublisher::PrepareUpdateEntityRecord()
+    void PropertyPublisher::PrepareUpdateEntityRecord(NetBindComponent* netBindComponent)
     {
-        bool didPrepare = true;
         if (m_sentRecords.size() >= net_EntityReplicatorRecordsMax)
         {
-            // If we reach the maximum outstanding records, reset the replication state
-            didPrepare = PrepareAddEntityRecord();
+            // If we reach the maximum outstanding records, reset the replication state by creating an "Add" record.
```
kberg-amzn, line 148: This could use slightly more detail about what an Add record is. I'm guessing it's just a total reset record, all values that differ from the original prefab state?

Gems/Multiplayer/Code/Source/NetworkEntity/EntityReplication/PropertyPublisher.h
```diff
@@ -27,25 +35,51 @@ namespace Multiplayer
             False,
         };
 
-        PropertyPublisher(NetEntityRole remoteNetworkRole, OwnsLifetime ownsLifetime, NetBindComponent* netBindComponent, AzNetworking::IConnection& connection);
+        PropertyPublisher(NetEntityRole remoteNetworkRole, OwnsLifetime ownsLifetime, AzNetworking::IConnection& connection);
 
+        //! Set the publishing state to "rebasing". The next record sent will be a rebase record.
```
kberg-amzn, line 40: Again, delete is pretty obvious, but a reader may not understand what a "rebase" is. A really quick description could help quite a bit.

mbalfour-amzn: This part is slim on comments because I don't feel 100% confident that I understand what a rebase is or what conditions would cause it to be set. I mostly left the rebasing piece of this alone.

Gems/Multiplayer/Code/Source/NetworkEntity/EntityReplication/PropertyPublisher.h
```diff
@@ -66,35 +100,45 @@ namespace Multiplayer
...
         //! Check if we have data to send
-        bool HasUpdateEntityRecord();
+        bool HasEntityChangesToSend();
 
         //! Phase 1, setup of the record
-        bool PrepareAddEntityRecord();
-        bool PrepareRebaseEntityRecord();
-        bool PrepareUpdateEntityRecord();
-        bool PrepareDeleteEntityRecord();
+        void PrepareAddEntityRecord(NetBindComponent* netBindComponent);
```
kberg-amzn, line 106: Oh.. perhaps I got this wrong and an add record is actually a create record, which includes the prefabentityId?

Gems/Multiplayer/Code/Tests/NetworkEntityTests.cpp
```diff
@@ -364,6 +365,122 @@ namespace Multiplayer
         EXPECT_TRUE(m_entityReplicationManager->HandleEntityUpdateMessage(m_mockConnection.get(), header, constMessage));
     }
 
+    TEST_F(MultiplayerNetworkEntityTests, EntityReplicatorNoDeleteSentIfCreateWasNotSent)
```
kberg-amzn, line 368: Big kudos for adding new tests! 👍  Awesome..

Gems/Multiplayer/Code/Tests/NetworkEntityTests.cpp
```diff
@@ -364,6 +365,122 @@ namespace Multiplayer
...
+        // The delete should now be seen as sent too.
+        EXPECT_FALSE(m_root->m_replicator->HasChangesToPublish());
+    }
+
+    TEST_F(MultiplayerNetworkEntityTests, EntityReplicatorDeleteMessageResentUntilAcknowledged)
+    {
+        // When sending a delete message, the message should keep getting resent until it has been acknowledged.
+
+        // Start by mocking that no packets were acknowledged.
+        ON_CALL(*m_mockConnection, WasPacketAcked).WillByDefault(::testing::Return(false));
```
kberg-amzn, line 450: This is a cool trick with gtest mocks.. I'll have to remember it.

Gems/Multiplayer/Code/Source/NetworkEntity/EntityReplication/PropertyPublisher.h
mbalfour-amzn: Hmmm. Let me poke at this a bit and make sure I get it right before I comment it. I *think* it's marking _every_ networked property as changed indiscriminately, regardless of whether or not it differs from the prefab. I assume it's called an "add" record because you'd typically use that at replicated creation time, though as you pointed out, it also gets used when too many updates get lost, so it's not _literally_ an add or create.

kberg-amzn (approved):
Wow, looks great to me!

Gems/Multiplayer/Code/Tests/NetworkEntityTests.cpp
mbalfour-amzn: Heh, to me this is the whole point to having mocks, we can easily inject behaviors on a per-test basis!

Gems/Multiplayer/Code/Source/NetworkEntity/EntityReplication/PropertyPublisher.h
kberg-amzn: It might just be everything, since that's easier than diffing against the prefab version and probably safer too (in the case where something changed away from default and then got restored back to a default value later).

Gems/Multiplayer/Code/Source/NetworkEntity/EntityReplication/PropertyPublisher.h
mbalfour-amzn: OK, after searching through the engine, there are no active code paths or tests that set rebasing. 

There's an entry point on EntityReplicationManager that is "SetEntityRebasing()", and from the code it's expected to be an Autonomous entity and it creates a full replication record minus any predictable values, and claims to be useful for client migrations. I'll repeat this information in a comment, but since it isn't used anywhere, I can't guarantee correctness.

mbalfour-amzn: Looking at this closer, I _suspect_ we could kill the separate rebase state and either just use PrepareAddEntityRecord() to send everything on migration, or change that method to always subtract out predictable properties for Autonomous entities. I don't understand why migrations would want to behave differently than creations in that sense, since even "creation" doesn't line up with local creation necessarily, it's remote creation for an entity that might have been processing for a little while already.

But again, since nothing uses rebasing right now, and there's no way to test migrations currently, I'll just leave this as-is other than the couple extra comments I added.

kberg-amzn: I double-checked against the multiserver code just to be sure, and couldn't see anything there either that relies on rebasing at the moment.

PR #16027, opened 2023-05-25
[Atom Renderer] Explicit render resolution support

VickyAtAZ:
## What does this PR do?
This PR added implicit render resolution support for render pipeline which renders to a swapchain. 
The render resolution can be different than the swapchain's size.
It also added support for swapchain scaling if the backend device supports it. 
A cvar is added for setting up implicit render resolution:
r_resolutionMode: "0: render solution same as window client area size, 1: render solution use a specified resolution"

Fullscreen on 4x monitor with swapchain scaling, dx12
dx12_swapchain
Fullscreen on 4x monitor by utilize CopyToSwapChain pass to do scaling , Vulkan
image

## How was this PR tested?
MPS setting menu in game launcher 
O3DE Editor. 
Material Editor.
ASV full test suites. 
All applications were tested with both DX12 and Vulkan backend on windows with 150% DPI

### DIFF
--- a/Code/Editor/Lib/Tests/test_ModularViewportCameraController.cpp
+++ b/Code/Editor/Lib/Tests/test_ModularViewportCameraController.cpp
@@ -153,6 +153,8 @@ namespace UnitTest
             // note: WindowRequests is used internally by ModularViewportCameraController, this ensures it returns the viewport size we want
             ON_CALL(m_mockWindowRequests, GetClientAreaSize())
                 .WillByDefault(Return(AzFramework::WindowSize(WidgetSize.width(), WidgetSize.height())));
+            ON_CALL(m_mockWindowRequests, GetRenderResolution())
+                .WillByDefault(Return(AzFramework::WindowSize(WidgetSize.width(), WidgetSize.height())));
 
             m_mockViewportInteractionRequests.Connect(TestViewportId);
 
--- a/Code/Framework/AzFramework/AzFramework/Windowing/NativeWindow.cpp
+++ b/Code/Framework/AzFramework/AzFramework/Windowing/NativeWindow.cpp
@@ -118,6 +118,26 @@ namespace AzFramework
         return m_pimpl->SupportsClientAreaResize();
     }
 
+    void NativeWindow::SetEnableCustomizedResolution(bool enable)
+    {
+        m_pimpl->SetEnableCustomizedResolution(enable);
+    }
+
+    bool NativeWindow::IsCustomizedResolutionEnabled()
+    {
+        return m_pimpl->IsCustomizedResolutionEnabled();
+    }
+
+    WindowSize NativeWindow::GetRenderResolution() const
+    {
+        return m_pimpl->GetRenderResolution();
+    }
+
+    void NativeWindow::SetRenderResolution(WindowSize resolution)
+    {
+        m_pimpl->SetRenderResolution(resolution);
+    }
+
     bool NativeWindow::GetFullScreenState() const
     {
         return m_pimpl->GetFullScreenState();
@@ -243,6 +263,9 @@ namespace AzFramework
         {
             m_activated = true;
             WindowNotificationBus::Event(GetWindowHandle(), &WindowNotificationBus::Events::OnWindowResized, m_width, m_height);
+
+            WindowSize resolution = GetRenderResolution();
+            WindowNotificationBus::Event(GetWindowHandle(), &WindowNotificationBus::Events::OnResolutionChanged, resolution.m_width, resolution.m_height);
         }
     }
 
@@ -282,6 +305,61 @@ namespace AzFramework
     {
         // Default to client area resize is unsupported, supported platforms will override this function
         return false;
+    }
+
+    void NativeWindow::Implementation::SetEnableCustomizedResolution(bool enable)
+    {
+        if (m_enableCustomizedResolution != enable)
+        {
+            // If disable customized resolution, we need to update render resolution based on client area size
+            WindowSize resolution;
+            if (!enable)
+            {
+                SetRenderResolution(GetClientAreaSize());
+            }
+            m_enableCustomizedResolution = enable;
+            if (enable)
+            {
+                // if enabling customized resolution, send notification if the resolution is different than client area size
+                if (m_customizedRenderResolution != GetClientAreaSize())
+                {
+                    WindowNotificationBus::Event(GetWindowHandle(), &WindowNotificationBus::Events::OnResolutionChanged, m_customizedRenderResolution.m_width, m_customizedRenderResolution.m_height);
+                }
+            }
+        }
+    }
+
+    bool NativeWindow::Implementation::IsCustomizedResolutionEnabled()
+    {
+        return m_enableCustomizedResolution;
+    }
+
+    WindowSize NativeWindow::Implementation::GetRenderResolution() const
+    {
+        if (m_enableCustomizedResolution)
+        {
+            return m_customizedRenderResolution;
+        }
+        else
+        {
+            return GetClientAreaSize();
+        }
+    }
+
+    void NativeWindow::Implementation::SetRenderResolution(WindowSize resolution)
+    {
+        if (m_enableCustomizedResolution)
+        {
+            if (m_customizedRenderResolution != resolution)
+            {
+                m_customizedRenderResolution = resolution;
+                WindowNotificationBus::Event(GetWindowHandle(), &WindowNotificationBus::Events::OnResolutionChanged, resolution.m_width, resolution.m_height);
+            } 
+        }
+        else
+        {
+            AZ_Assert(false, "Use SetEnableCustomizedResolution(true) to enable customized render resolution first");
+        }
     }
 
     bool NativeWindow::Implementation::GetFullScreenState() const
--- a/Code/Framework/AzFramework/AzFramework/Windowing/NativeWindow.cpp
+++ b/Code/Framework/AzFramework/AzFramework/Windowing/NativeWindow.cpp
@@ -123,7 +123,7 @@ namespace AzFramework
         m_pimpl->SetEnableCustomizedResolution(enable);
     }
 
-    bool NativeWindow::IsCustomizedResolutionEnabled()
+    bool NativeWindow::IsCustomizedResolutionEnabled() const
     {
         return m_pimpl->IsCustomizedResolutionEnabled();
     }
@@ -331,7 +331,7 @@ namespace AzFramework
         }
     }
 
-    bool NativeWindow::Implementation::IsCustomizedResolutionEnabled()
+    bool NativeWindow::Implementation::IsCustomizedResolutionEnabled() const
     {
         return m_enableCustomizedResolution;
     }
--- a/Code/Framework/AzFramework/AzFramework/Windowing/NativeWindow.cpp
+++ b/Code/Framework/AzFramework/AzFramework/Windowing/NativeWindow.cpp
@@ -311,16 +311,18 @@ namespace AzFramework
     {
         if (m_enableCustomizedResolution != enable)
         {
-            // If disable customized resolution, we need to update render resolution based on client area size
-            WindowSize resolution;
+            // If disabling customized resolution, we need to update render resolution based on client area size
+            // Note: this need to be done when m_enableCustomizedResolution is true
             if (!enable)
             {
                 SetRenderResolution(GetClientAreaSize());
             }
+
             m_enableCustomizedResolution = enable;
+
+            // after enabling customized resolution, send notification if the customized resolution is different than client area size
             if (enable)
             {
-                // if enabling customized resolution, send notification if the resolution is different than client area size
                 if (m_customizedRenderResolution != GetClientAreaSize())
                 {
                     WindowNotificationBus::Event(GetWindowHandle(), &WindowNotificationBus::Events::OnResolutionChanged, m_customizedRenderResolution.m_width, m_customizedRenderResolution.m_height);
--- a/Code/Framework/AzFramework/AzFramework/Windowing/WindowBus.h
+++ b/Code/Framework/AzFramework/AzFramework/Windowing/WindowBus.h
@@ -30,6 +30,16 @@ namespace AzFramework
 
         uint32_t m_width = 0;
         uint32_t m_height = 0;
+        
+        bool operator==(const WindowSize& rhs) const
+        {
+            return m_width == rhs.m_width && m_height == rhs.m_height;
+        }
+
+        bool operator!=(const WindowSize& rhs) const
+        {
+            return m_width != rhs.m_width || m_height != rhs.m_height;
+        }
     };
 
     //! Options for resizing and moving the window.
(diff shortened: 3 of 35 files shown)


Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
```diff
@@ -178,6 +179,24 @@ namespace AZ
...
+                        if (AZ::StringFunc::LooksLikeInt(valueStr.c_str()))
+                        {
+                            auto resolutionMode = AZ::StringFunc::ToInt(valueStr.c_str());
+                            if (resolutionMode >= 0)
+                            {
+                                r_resolutionMode = resolutionMode;
+                            }
+                        }
+                    }
+                }
```
lumberyard-employee-dm, line 198: The AZ CVAR system automatically handles command line parameters.

Since this is an AZ CVAR handling the command line parameters is not needed.

I would also recommend using the callback parameter to [AZ_CVAR](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Console/IConsole.h#L211-L222) to handle updating the render resolution support if the CVar is changed while the application is running.

rgba16f: > The AZ CVAR system automatically handles command line parameters.
> 
> Since this is an AZ CVAR handling the command line parameters is not needed.
> 
> I would also recommend using the callback parameter to [AZ_CVAR](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Console/IConsole.h#L211-L222) to handle updating the render resolution support if the CVar is changed while the application is running.

I believe these values are used before CVAR system is active and updated from the command line, hence the need to actively query the command line for any non-default values.

Code/Framework/AzFramework/AzFramework/Windowing/NativeWindow.cpp
```diff
@@ -284,6 +307,61 @@ namespace AzFramework
...
+        if (m_enableCustomizedResolution != enable)
+        {
+            // If disable customized resolution, we need to update render resolution based on client area size
+            WindowSize resolution;
+            if (!enable)
+            {
+                SetRenderResolution(GetClientAreaSize());
+            }
+            m_enableCustomizedResolution = enable;
+            if (enable)
```
mbalfour-amzn, line 321: Tiny nit: If you move the `m_enableCustomizedResolution = enable;` down to the bottom, you could make the `if (enable)` just an `else`.

If it needs to appear exactly where it is due to side effects of it being set before SetRenderResolution() or OnResolutionChanged, then please add a comment describing that.

Code/Framework/AzFramework/Platform/Windows/AzFramework/Windowing/NativeWindow_Windows.cpp
```diff
@@ -322,40 +322,40 @@ namespace AzFramework
...
+
+            //        // If we're going to fullscreen, then we presumably want to be TOPMOST whether or not a debugger is attached.
+            //        windowPriority = HWND_TOPMOST;
+            //    }
+
+            //    // When becoming active again, transition from NOTOPMOST to TOPMOST. (Or stay NOTOPMOST if a debugger is attached)
+            //    SetWindowPos(
+            //        nativeWindowImpl->m_win32Handle,
+            //        windowPriority, 0, 0, 0, 0, SWP_NOMOVE | SWP_NOSIZE | SWP_NOACTIVATE | SWP_NOSENDCHANGING);
+            //}
```
mbalfour-amzn, line 358: Should this change get reverted? It looks like something removed locally while debugging.

Gems/Atom/RHI/Code/Include/Atom/RHI.Reflect/SwapChainDescriptor.h
```diff
@@ -64,6 +64,11 @@ namespace AZ
             // Return the index of a XR swapchain
             // as you can have multiple XR swapchains (one per view).
             uint32_t m_xrSwapChainIndex = 0;
+
+            // The scaling mode when presents the swapchain's back buffer to the target
```
mbalfour-amzn, line 68:
```suggestion
            // The scaling mode to use when presenting the swapchain's back buffer to the target
```

Code/Framework/AzFramework/Platform/Windows/AzFramework/Windowing/NativeWindow_Windows.cpp
VickyAtAZ: Yes, it need to be reverted.

Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
mbalfour-amzn: https://github.com/o3de/o3de/commit/c1d56b73c63f01da92ad0bf7a0da85783bf5292c

It's definitely a good idea to make the cvar handle updates at runtime, but i'd suggest possibly saving that for a different PR, because the other related cvars (r_width, r_height, r_fullscreen) would all need to get updated for this.

Code/Framework/AzFramework/AzFramework/Windowing/NativeWindow.cpp
VickyAtAZ: I will add comment

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Pass/Specific/SwapChainPass.h
```diff
@@ -52,22 +52,38 @@ namespace AZ
...
 
             // The SwapChain used when rendering this pass
             Ptr<PassAttachment> m_swapChainAttachment;
 
-            RHI::SwapChainDimensions m_swapChainDimensions;
+            // The intermediate attachment used for render pipeline's output
+            Ptr<PassAttachment> m_pipelinOutputAttachment;
+
+            // The pass which copies pipeline output to swapchain 
+            Ptr<Pass> m_copyOutputPass;
```
moudgils, line 82: Can you explain a bit about this copy. Aren't we already writing to an intermediate RT and then doing a copy to swapchain so why do we need a second copy?

VickyAtAZ: This copy pass implementation is expanding the SwapchainPass the functionality to giving the render pipeline an output attachment other than the swapchain attachment. 
It works with any render pipeline pass template. The intermediate RT of MainPipeline is a choice of the MainPipeline. MainPipeline may choose to remove that RT by changing its pass template. And it won't effect the SwapchainPass's ability of having a different resolution for render pipeline output.

Gems/Atom/RHI/Code/Include/Atom/RHI.Reflect/Base.h
```diff
@@ -130,7 +130,23 @@ namespace AZ
             DepthThenKey,
             ReverseDepthThenKey
         };
+        
+        enum class Scaling : uint32_t
+        {
+            None = 0,               // No scaling
+            Stretch,                // Scale the source to fit the target
+            AspectRatioStretch,     // Scale the source to fit the target while preserving the aspect ratio of the source
```
AMZN-alexpete, line 138: does this mode leave grey bars in the areas not drawn, or does it stretch to fill and clip so you don't see the entire rendered content?

AMZN-alexpete: Ah looks like there are borders containing the window background color https://learn.microsoft.com/en-us/windows/win32/api/dxgi1_2/ne-dxgi1_2-dxgi_scaling

AMZN-alexpete:
Could this feature be used to scale the rendered content but keep the UI at full resolution?  I'm wondering if it will also be useful for supporting FSR

VickyAtAZ:
> Could this feature be used to scale the rendered content but keep the UI at full resolution? I'm wondering if it will also be useful for supporting FSR

No. That would be a different feature which require RenderPipeline's pass template support. The render pipeline's 3d passes need to use a resolution while the 2d passes use another resolution. This actually a render pipeline support dynamic resolution rendering would do

VickyAtAZ:
Btw, using builtin swapchain scaling saves gpu memory compare to using swapchain which has same size as the window. With the test done on a 4k monitor, I observed more than 0.5G vram usage different for a 2k resolution application.

moudgils (commented):
I am trying to better understand the fallback solution. Please correct me if I am wrong here but I think this is the current setup on Vk and metal.

**CopyToSwapChain**
Transient FrameBuffer (custom resolution) -> MainPipeline_Output (custom resolution)
**CopyOutputToSwapChain**
MainPipeline_Output (custom resolution) -> SwapChainTexture (Window resolution) -> Present


In the above case we are doing an extra unnecessary copy (which is probably un-acceptable on mobile given how expensive full screen copies are). Is it not possible to just do this?

**CopyToSwapChain**
Transient FrameBuffer (custom resolution) -> SwapChainTexture (window resolution) -> Present

mbalfour-amzn:
> I am trying to better understand the fallback solution. Please correct me if I am wrong here but I think this is the current setup on Vk and metal.
> 
> **CopyToSwapChain** Transient FrameBuffer (custom resolution) -> MainPipeline_Output (custom resolution) **CopyOutputToSwapChain** MainPipeline_Output (custom resolution) -> SwapChainTexture (Window resolution) -> Present
> 
> In the above case we are doing an extra unnecessary copy (which is probably un-acceptable on mobile given how expensive full screen copies are). Is it not possible to just do this?
> 
> **CopyToSwapChain** Transient FrameBuffer (custom resolution) -> SwapChainTexture (window resolution) -> Present

I thought PipelineOutput was used for things like DepthPrePass etc? Wouldn't skipping it break any passes that rely on the output of the main pipeline?

moudgils:
> > I am trying to better understand the fallback solution. Please correct me if I am wrong here but I think this is the current setup on Vk and metal.
> > **CopyToSwapChain** Transient FrameBuffer (custom resolution) -> MainPipeline_Output (custom resolution) **CopyOutputToSwapChain** MainPipeline_Output (custom resolution) -> SwapChainTexture (Window resolution) -> Present
> > In the above case we are doing an extra unnecessary copy (which is probably un-acceptable on mobile given how expensive full screen copies are). Is it not possible to just do this?
> > **CopyToSwapChain** Transient FrameBuffer (custom resolution) -> SwapChainTexture (window resolution) -> Present
> 
> I thought PipelineOutput was used for things like DepthPrePass etc? Wouldn't skipping it break any passes that rely on the output of the main pipeline?

I think PipelineOutput can be used for capturing screenshots (but not DepthPrepass). I am not saying to skip writing to pipeline output. Just to not introduce a second swapchain copy for no reason.

Code/Editor/Lib/Tests/test_ModularViewportCameraController.cpp
mbalfour-amzn, line 1: Any idea why this entire file is showing up as a diff? Did the line endings change? I only see a couple of added lines, no other obvious changes...

Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
```diff
@@ -53,6 +53,7 @@ AZ_CVAR(AZ::CVarFixedString, r_default_openxr_right_pipeline_name, "passes/XRRig
 AZ_CVAR(uint32_t, r_width, 1920, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Starting window width in pixels.");
 AZ_CVAR(uint32_t, r_height, 1080, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Starting window height in pixels.");
 AZ_CVAR(uint32_t, r_fullscreen, false, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Starting fullscreen state.");
+AZ_CVAR(uint32_t, r_resolutionMode, 0, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "0: render resolution same as window client area size, 1: render resolution use the values specified by r_width and r_height");
```
mbalfour-amzn, line 56: should this be a bool instead of a uint32_t? (and maybe "r_useCustomResolution" or another more descriptive name?)

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/RenderPipeline.h
mbalfour-amzn, line 1: And again here... whole file is marked as changed.

Code/Framework/AzFramework/AzFramework/Windowing/WindowBus.h
```diff
@@ -127,6 +152,9 @@ namespace AzFramework
 
         //! This is called once when the window is Activated and also called if the user resizes the window.
         virtual void OnWindowResized(uint32_t width, uint32_t height) { AZ_UNUSED(width); AZ_UNUSED(height); };
+        
+        //! This is called when the window's desired render resolution is changed.
+        virtual void OnResolutionChanged(uint32_t width, uint32_t height) { AZ_UNUSED(width); AZ_UNUSED(height); };
```
mbalfour-amzn, line 157: I'm totally good with this API addition, but I just want to point out for consideration that the usage below seemed like OnWindowResized / OnResolutionChanged are typically called in pairs. If any listeners end up needing to know about both of these, they might need to create some workarounds or implicit assumptions about the calling order so that they only process once they have complete information from both notifications.

I think in this case though, the listeners that I saw below in the PR were only listening for one or the other, not both, which is why I don't think it's _actually_ a problem in this case. It's just a tricky thing to watch out for with API design, so I wanted to call it out for consideration in case you ran into any of this yourself while making these changes.

mbalfour-amzn (approved):
Approved, except for the whole-file changes should get examined and fixed first.

Code/Editor/Lib/Tests/test_ModularViewportCameraController.cpp
VickyAtAZ: The the line ending change. The previous line ending was /n. And it was changed to /r/n
I asked in the internal discussion and the /r/n should be the preferred/required the ending format

Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
VickyAtAZ: I am thinking about this might be used for future function expansion, especially when we introduce the resolution only for 3d rendering.

lumberyard-employee-dm: We support enums for AZ_CVAR as well

Code/Framework/AzFramework/AzFramework/Windowing/WindowBus.h
VickyAtAZ: Currently, the implementation is that when both values are changed, the notifications would be called after both values are changed. So if people is trying to get render resolution in OnWindowResized() function, they would get the latest the render solution (the one paired with window size). 

There are use cases that the two values are not changed together. For example, when MPS's full screen states change, the window size is changed, but not the render resolution, in this case, we don't need to handle swapchain resize (for dx12).

moudgils (approved):

Code/Editor/Lib/Tests/test_ModularViewportCameraController.cpp
mbalfour-amzn: ah. i thought the answer was that git was supposed to auto-convert it, but i guess you're saying these were originally not converted and this now fixes it, so that works for me. thx!

Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
VickyAtAZ: Good to know.
It seems there are only two places which use enums for AZ_CVAR

PR #16032, opened 2023-05-26
Shadow cascade pass optimization by rejecting more meshes based on frustum intersection between light and camera.

moudgils:
## What does this PR do?

Optimization
- This optimization is disabled by default unless triggered by the project. Companion PR for MPS - https://github.com/o3de/o3de-multiplayersample/pull/429
- For shadow cascades that are greater than index 0 we can do another check to see if we can reject any Octree node that do not intersect with the camera frustum. We do this by checking for an overlap between the camera frustum and the Obb created from the node's AABB but rotated and extended towards light direction.
- We pick the camera view via inquiring the render pipeline associated with the shadow pass used by the shadow light view
- Perf data
    - CPU - On the CPU we add extra time doing intersection tests for the shadow cascade views but at the same time we save time encoding meshes if they get rejected. Given that all this time is done in parallel the extra time spent as well as saved is hidden across various job threads. In total there is no significant impact on cpu.
    - GPU - Since we reject draw calls we can see savings from 0ms in the worst case up to 0.5ms in the best case for the tests done on MPS. The worst case is when the camera frustum looks across the whole level and the best case is when the camera is looking at a small section of the level.

Before Opt - Total draw calls = **4148**. Gpu time = **0.754ms**
image

After Opt - Total draw calls = **2123**. Gpu time = **0.435ms**
image

Reference camera view (Ignore the high sun intensity done for debugging purposes)
image

## How was this PR tested?

Tested MPS and ASV. Tested with multiple render pipelines


Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
```diff
@@ -54,6 +52,10 @@ namespace AZ
         // Node work lists using node count
         AZ_CVAR(uint32_t, r_numNodesPerCullingJob, 25, nullptr, AZ::ConsoleFunctorFlags::Null, "Controls amount of nodes to collect for jobs when not using the entry count");
 
+        //This value dictates the amount to extrude the octree node OBB when doing a frustum intersection test against the camera frustum to help cut draw calls. Default is set to 20 units as a reasonable value.
+        AZ_CVAR(int, r_shadowCascadeExtrutionAmount, 20, nullptr, AZ::ConsoleFunctorFlags::Null, "The amount to extrude the Obb towards light direction when doing frustum overlap test against camera frustum");
```
mbalfour-amzn, line 56:
```suggestion
        AZ_CVAR(int, r_shadowCascadeExtrusionAmount, 20, nullptr, AZ::ConsoleFunctorFlags::Null, "The amount to extrude the Obb towards light direction when doing frustum overlap test against camera frustum");
```

Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
```diff
@@ -585,10 +589,50 @@ namespace AZ
...
+                if (worklistData->m_applyCameraFrustumIntersectionTest && worklistData->m_hasExcludeFrustum)
+                {
+                    // Build an Obb from the Octree node's aabb
+                    AZ::Obb extrudedBounds = AZ::Obb::CreateFromAabb(nodeData.m_bounds);
+
+                    // Rotate the Obb in the direction of the light
+                    AZ::Quaternion directionalLightRot = worklistData->m_view->GetCameraTransform().GetRotation();
+                    extrudedBounds.SetRotation(directionalLightRot);
+                    
+                    AZ::Vector3 halfLegth = 0.5f * nodeData.m_bounds.GetExtents();
```
mbalfour-amzn, line 619:
```suggestion
                    AZ::Vector3 halfLength = 0.5f * nodeData.m_bounds.GetExtents();
```

Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
```diff
@@ -585,10 +589,50 @@ namespace AZ
...
+                    // After converting AABB to OBB we apply a rotation and this can incorrectly fail intersection test. If you have an OBB cube built from an octree node,
+                    // rotating it can cause it not encapsulate meshes it encapsulated beforehand. The type of shape we want here is essentially a capsule that starts from the
```
mbalfour-amzn, line 621:
```suggestion
                    // rotating it can cause it to not encapsulate meshes it encapsulated beforehand. The type of shape we want here is essentially a capsule that starts from the
```

Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
```diff
@@ -585,10 +589,50 @@ namespace AZ
...
+                    // light and wraps the aabb of the octree node cube and extends towards light direction. This capsule's diameter needs to the size of the body diagonal
+                    // of the cube. Since using capsule shape will make intersection test expensive we simply expand the Obb to have each side be atleast the size of the body diagonal
```
mbalfour-amzn, line 623:
```suggestion
                    // of the cube. Since using capsule shape will make intersection test expensive we simply expand the Obb to have each side be at least the size of the body diagonal
```

Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
```diff
@@ -585,10 +589,50 @@ namespace AZ
...
+                    // which is sqrt(3) * side size. Hence we expand the Obb by 73%. 
+                    halfLegth += halfLegth * Vector3(0.36f, 0.36f, 0.36f);
```
mbalfour-amzn, line 625:
```suggestion
                    // which is sqrt(3) * side size. Hence we expand the Obb by 73%. Since this is half length, we expand it by 73% / 2, or 36.5%.
                    halfLength *= Vector3(1.365f);
```

Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
```diff
@@ -585,10 +589,50 @@ namespace AZ
...
             
             auto nodeVisitorLambda = [worklistData, taskGraph, parentJob, &worklist](const AzFramework::IVisibilityScene::NodeData& nodeData) -> void
             {
+                // For shadow cascades that are greater than index 0 we can do another check to see if we can reject any Octree node that do not
+                // intersect with the camera frustum. We do this by checking for an overlap between the camera frustum and the Obb created
+                // from the node's AABB but rotated and extended towards light direction. 
+                if (worklistData->m_applyCameraFrustumIntersectionTest && worklistData->m_hasExcludeFrustum)
+                {
+                    // Build an Obb from the Octree node's aabb
+                    AZ::Obb extrudedBounds = AZ::Obb::CreateFromAabb(nodeData.m_bounds);
```
AMZN-alexpete, line 613: are the node bounds in world space or clip space like the camera frustum?

Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
```diff
@@ -585,10 +589,50 @@ namespace AZ
...
+                    // After converting AABB to OBB we apply a rotation and this can incorrectly fail intersection test. If you have an OBB cube built from an octree node,
+                    // rotating it can cause it not encapsulate meshes it encapsulated beforehand. The type of shape we want here is essentially a capsule that starts from the
+                    // light and wraps the aabb of the octree node cube and extends towards light direction. This capsule's diameter needs to the size of the body diagonal
+                    // of the cube. Since using capsule shape will make intersection test expensive we simply expand the Obb to have each side be atleast the size of the body diagonal
+                    // which is sqrt(3) * side size. Hence we expand the Obb by 73%. 
+                    halfLegth += halfLegth * Vector3(0.36f, 0.36f, 0.36f);
+                    
+                    // Next we extrude the Obb in the direction of the light in order to ensure we capture meshes that are behind the camera but cast a shadow within it's frustum
+                    halfLegth.SetY(halfLegth.GetY() + r_shadowCascadeExtrutionAmount);
```
AMZN-alexpete, line 628: if you have a directional light component on an entity, is "behind the camera" relative to the entity world space position?  and is the r_shadowCascadeExtrusionAmount in meters in the -Y direction relative to the (shadow) camera?  it's unclear to me when this parameter would need to be changed, but it seems if you had a directional light on an entity that was in the middle of your scene and a building more than 20 meters "behind" the directional light, it would no longer be drawn into the shadow cascade.

moudgils: - Correct, "Behind the Camera" is relative to the entity's world space position. 
- Correct, r_shadowCascadeExtrusionAmount is a value in meters in the forward direction which is +Y (O3de is Z-up in world space).
- Correct, consider this cvar as extra padding but it is not a perfect solution. We already expand the Obb in the y-axis by 36% so it will capture a lot of the objects in the near vicinity. I added support for this cvar in case people are still missing shadows for an object that was much further away and outside the frustum and want to extrude the Obb even further they can modify the cvar. We dont want to make this cvar too large as then you are loosing the perf gain by this optimizaiton.

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/View.h
```diff
@@ -194,6 +198,9 @@ namespace AZ
             //! Attempt to create a shader resource group.
             void TryCreateShaderResourceGroup();
 
+            //! Update ViewToWorld matrix as well us the view transform
```
VickyAtAZ, line 201: *as

VickyAtAZ (approved):

kberg-amzn (approved):

Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
```diff
@@ -54,6 +52,10 @@ namespace AZ
         // Node work lists using node count
         AZ_CVAR(uint32_t, r_numNodesPerCullingJob, 25, nullptr, AZ::ConsoleFunctorFlags::Null, "Controls amount of nodes to collect for jobs when not using the entry count");
 
+        //This value dictates the amount to extrude the octree node OBB when doing a frustum intersection test against the camera frustum to help cut draw calls. Default is set to 20 units as a reasonable value.
+        AZ_CVAR(int, r_shadowCascadeExtrusionAmount, 20, nullptr, AZ::ConsoleFunctorFlags::Null, "The amount to extrude the Obb towards light direction when doing frustum overlap test against camera frustum");
```
mbalfour-amzn, line 56: Nit: Would be nice to say what the units are - are these meters?

mbalfour-amzn (approved):

Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
kberg-amzn: As discussed offline, perhaps consider replacing this distance with the projected distance between the cell and the camera.
image
therefore
image

moudgils: This only works in some use cases. An example where it wont work is a following. By projecting onto light vector we get cy which does not extend to the frustum.
IMG_1092

invertednormal:
If I understand it right, I'd think this should be an optional optimization since it can ignore shadow casters that are outside the view's frustum that still result in a shadow that the view frustum can see. Only draws outside the view frustum opposite the direction of the light are actually safe to cull.

moudgils:
> If I understand it right, I'd think this should be an optional optimization since it can ignore shadow casters that are outside the view's frustum that still result in a shadow that the view frustum can see. Only draws outside the view frustum opposite the direction of the light are actually safe to cull.

I am extruding the Ochre's OBB by 36% (bigger OBBs will extend further) + another 20 meters in order to catch nodes that are outside the frustum but still cast shadow into the frustum. It is not ideal but it should be good for most cases and keeps the cpu cost down. In theory, I could calculate light direction vector to camera frustum intersection and extrude to that distance but then the cpu cost will become significant enough that this optimization is not worth. 

A possible direction I was considering was to have the extrusion direction set to -1 as default so the optimization is disabled by default but if a content developer sets this car to a positive number, it triggers this optimization and it forces the developer to set the correct amount of extrusion for their content. So if you have a very large scene with massive buildings you may want to extrude to 50m. I am curious to know your thoughts @invertednormal.

moudgils:
> > > If I understand it right, I'd think this should be an optional optimization since it can ignore shadow casters that are outside the view's frustum that still result in a shadow that the view frustum can see. Only draws outside the view frustum opposite the direction of the light are actually safe to cull.
> > 
> > 
> > I am rotating the Octree OBB towards light direction and then extruding the Octee's OBB by 36% (bigger OBBs will extend further) + another 20 meters in order to catch nodes that are outside the frustum but still cast shadow into the frustum. It is not ideal but it should be good for most cases and keeps the cpu cost down. In theory, I could calculate light direction vector to camera frustum intersection and extrude to that distance but then the cpu cost will become significant enough that this optimization is not worth.
> > A possible direction I was considering was to have the extrusion direction set to -1 as default so the optimization is disabled by default but if a content developer sets this car to a positive number, it triggers this optimization and it forces the developer to set the correct amount of extrusion for their content. So if you have a very large scene with massive buildings the content developers can set the extrusion cvar to 50m in their project. I am curious to know your thoughts @invertednormal.
> 
> Should this optimization be disabled for any level that uses terrain?

No. Its not about using terrain or not. Its more about how much to extrude the OBB to catch some weird edge cases for your content. You can get these edge cases with or without terrain.

rgba16f:
> > If I understand it right, I'd think this should be an optional optimization since it can ignore shadow casters that are outside the view's frustum that still result in a shadow that the view frustum can see. Only draws outside the view frustum opposite the direction of the light are actually safe to cull.
> 
> I am rotating the Octree OBB towards light direction and then extruding the Octee's OBB by 36% (bigger OBBs will extend further) + another 20 meters in order to catch nodes that are outside the frustum but still cast shadow into the frustum. It is not ideal but it should be good for most cases and keeps the cpu cost down. In theory, I could calculate light direction vector to camera frustum intersection and extrude to that distance but then the cpu cost will become significant enough that this optimization is not worth.
> 
> A possible direction I was considering was to have the extrusion direction set to -1 as default so the optimization is disabled by default but if a content developer sets this cvar to a positive number, it triggers this optimization and it forces the developer to set the correct amount of extrusion for their content. So if you have a very large scene with massive buildings the content developers can set the extrusion cvar to 50m in their project. I am curious to know your thoughts @invertednormal.

I think @invertednormal is thinking of cases where a directional light (or any other high intensity light) is low to the horizon and will cast shadows across most of the level. The optimization will make it faster at the cost of having incorrect shadows in some cases.

mbalfour-amzn:
> > > > If I understand it right, I'd think this should be an optional optimization since it can ignore shadow casters that are outside the view's frustum that still result in a shadow that the view frustum can see. Only draws outside the view frustum opposite the direction of the light are actually safe to cull.
> > > 
> > > 
> > > I am rotating the Octree OBB towards light direction and then extruding the Octee's OBB by 36% (bigger OBBs will extend further) + another 20 meters in order to catch nodes that are outside the frustum but still cast shadow into the frustum. It is not ideal but it should be good for most cases and keeps the cpu cost down. In theory, I could calculate light direction vector to camera frustum intersection and extrude to that distance but then the cpu cost will become significant enough that this optimization is not worth.
> > > A possible direction I was considering was to have the extrusion direction set to -1 as default so the optimization is disabled by default but if a content developer sets this car to a positive number, it triggers this optimization and it forces the developer to set the correct amount of extrusion for their content. So if you have a very large scene with massive buildings the content developers can set the extrusion cvar to 50m in their project. I am curious to know your thoughts @invertednormal.
> > 
> > 
> > Should this optimization be disabled for any level that uses terrain?
> 
> No. Its not about using terrain or not. Its more about how much to extrude the OBB to catch some weird edge cases for your content. You can get these edge cases with or without terrain.

To rephrase the question, terrain is a use case that very often has tall features that can appear very far behind the camera. If this optimization is turned on by default, won't this be extremely likely to cause artifacts in most standard terrain use cases? 

Assuming I'm understanding the ramifications correctly, I'm basically just +1ing the idea of having this optimization off by default, or at a _minimum_ to default it off when terrain is present, so that our defaults favor correctness over speed.

moudgils:
> > > If I understand it right, I'd think this should be an optional optimization since it can ignore shadow casters that are outside the view's frustum that still result in a shadow that the view frustum can see. Only draws outside the view frustum opposite the direction of the light are actually safe to cull.
> > 
> > 
> > I am rotating the Octree OBB towards light direction and then extruding the Octee's OBB by 36% (bigger OBBs will extend further) + another 20 meters in order to catch nodes that are outside the frustum but still cast shadow into the frustum. It is not ideal but it should be good for most cases and keeps the cpu cost down. In theory, I could calculate light direction vector to camera frustum intersection and extrude to that distance but then the cpu cost will become significant enough that this optimization is not worth.
> > A possible direction I was considering was to have the extrusion direction set to -1 as default so the optimization is disabled by default but if a content developer sets this cvar to a positive number, it triggers this optimization and it forces the developer to set the correct amount of extrusion for their content. So if you have a very large scene with massive buildings the content developers can set the extrusion cvar to 50m in their project. I am curious to know your thoughts @invertednormal.
> 
> I think @invertednormal is thinking of cases where a directional light (or any other high intensity light) is low to the horizon and will cast shadows across most of the level. The optimization will make it faster at the cost of having incorrect shadows in some cases.



> > > If I understand it right, I'd think this should be an optional optimization since it can ignore shadow casters that are outside the view's frustum that still result in a shadow that the view frustum can see. Only draws outside the view frustum opposite the direction of the light are actually safe to cull.
> > 
> > 
> > I am rotating the Octree OBB towards light direction and then extruding the Octee's OBB by 36% (bigger OBBs will extend further) + another 20 meters in order to catch nodes that are outside the frustum but still cast shadow into the frustum. It is not ideal but it should be good for most cases and keeps the cpu cost down. In theory, I could calculate light direction vector to camera frustum intersection and extrude to that distance but then the cpu cost will become significant enough that this optimization is not worth.
> > A possible direction I was considering was to have the extrusion direction set to -1 as default so the optimization is disabled by default but if a content developer sets this cvar to a positive number, it triggers this optimization and it forces the developer to set the correct amount of extrusion for their content. So if you have a very large scene with massive buildings the content developers can set the extrusion cvar to 50m in their project. I am curious to know your thoughts @invertednormal.
> 
> I think @invertednormal is thinking of cases where a directional light (or any other high intensity light) is low to the horizon and will cast shadows across most of the level. The optimization will make it faster at the cost of having incorrect shadows in some cases.

yeah in that circumstance the developer will have to use higher extrusion value if they see missing shadows. I am ok with having this optimization be off by default and only enabled by the developer who will be forced to set the value based on their content.

invertednormal:
I'm cool with the optimization in general - I'm mostly advocating for it being configurable so that users can choose how much to bias for correctness vs performance. I don't like making it where the optimization is forced and correctness is no longer possible in  many cases.

kberg-amzn:
> If I understand it right, I'd think this should be an optional optimization since it can ignore shadow casters that are outside the view's frustum that still result in a shadow that the view frustum can see. Only draws outside the view frustum opposite the direction of the light are actually safe to cull.

As per the diagrams above, the optimization can be perfectly safe as long as a correct extrude distance is correctly chosen. I would prefer an algorithmic solution for finding this distance so it can be guaranteed to be correct regardless of content, every time. As this check is done hierarchically by traversing a spatial tree, I suspect the aggregate cost of calculating the points of intersection would still be small enough for this to be a nice performance win.

Gems/Atom/RPI/Code/Source/RPI.Public/Culling.cpp
kberg-amzn: It's guaranteed to intersect the camera if we applied the same algorithm to each point on the original bounding volume.

image

Note all the points of intersection should lie on the B-D line segment. It seems to be close to a workable solution however, if we used the maximum segment length, say the distance between A-B and applied that to all points on the bounding volume to get the volume defined by A, B, C, E, we'd definitely get something that intersected the view frustum.

rgba16f: > It's guaranteed to intersect the camera if we applied the same algorithm to each point on the original bounding volume.
> 
> image
> 
> Note all the points of intersection should lie on the B-D line segment. It seems to be close to a workable solution however, if we used the maximum segment length, say the distance between A-B and applied that to all points on the bounding volume to get the volume defined by A, B, C, E, we'd definitely get something that intersected the view frustum.

Directional lights are an orthographic projection and don't have an origin to work from. That means the calculation is then relative to the obb of the shadow cascade, which is not as cheap to calculate for every object potentially in the shadow casting frustum.

moudgils: Actually, I dont really need to get to point x which was going to be expensive. The dot product between ac and d will get me to  the camera and I will extend it a little further to get past the near plane and that should work. Its just a few vector ALUs so hopefully not significant on cpu. Ok let me try that approach.

moudgils: @kberg-amzn yes you are correct for that use case. I could in theory project the vector (centre of aabb to camera) onto light direction and it will tell me the distance from aabb to the camera and then I can extrude a little more to go past the near plane. But it break for other use cases. 

- Use case 1 is when the aabb is in front of the camera but still does not intersect the frustum as shown below.

IMG_1154

- Use case 2 is when the aabb is behind the camera but further away such that even extending it to the calculated dot product may still miss the frustum unless you pad the frustum a little and then we are in the same boat as we are in right now where we dont know how much to pad. 

IMG_1155

You could calculate the distance between center of aabb to the frustum by doing a vector to plane intersection for all the planes of the camera frustum and then extending the aabb to that distance. This will be expensive on cpu given that we calculate all the meshes of a cascade on one thread. It is not worth as this still doesnt help the worst case on gpu so in theory we are making the worst case even worse on cpu by doing all these extra tests but not rejecting any meshes. This solution also has a minor problem where some aabbs can eventually intersect with camera frustum under the ground and we will be incorrectly accepting them.

moudgils:
> > If I understand it right, I'd think this should be an optional optimization since it can ignore shadow casters that are outside the view's frustum that still result in a shadow that the view frustum can see. Only draws outside the view frustum opposite the direction of the light are actually safe to cull.
> 
> As per the diagrams above, the optimization can be perfectly safe as long as a correct extrude distance is correctly chosen. I would prefer an algorithmic solution for finding this distance so it can be guaranteed to be correct regardless of content, every time. As this check is done hierarchically by traversing a spatial tree, I suspect the aggregate cost of calculating the points of intersection would still be small enough for this to be a nice performance win.

I have added use cases above where the doing a dot product will fail. You would need to do vector plane intersection for each plane and that will be expensive on cpu. I have added support to disable the optimization by default in the core engine and added a cvar in MPS to enable it for that project.

PR #16050, opened 2023-05-31
Add useCompactStyle attribute to AutoGen function nodes

mddvisu:
## What does this PR do?

This adds an attribute to the ScriptCanvasFunction.xsd schema file to allow for users to specify a node created through AzAutoGen to have a compact style. Changes the ScriptCanvasFunctionRegistry_Source.jinja template file to add this data as an attribute to the behavior context. No visible changes happen yet when this attribute is enabled for a particular function node. I am currently working on implementing these changes, where adding the attribute will shrink down the size of the node and make it into a more compact style.

## How was this PR tested?

I preformed manual testing by adding the attribute to different function nodes created with AzAutoGen.

### DIFF
--- a/Code/Framework/AzCore/AzCore/ScriptCanvas/ScriptCanvasAttributes.h
+++ b/Code/Framework/AzCore/AzCore/ScriptCanvas/ScriptCanvasAttributes.h
@@ -68,6 +68,8 @@ namespace AZ
 
         static const AZ::Crc32 BranchOnResult = AZ_CRC("BranchOnResult", 0x7a741f05);
 
+        static const AZ::Crc32 UseCompactStyle = AZ_CRC_CE("UseCompactStyle");
+
         ///< \see enum class Script::Attributes::OperatorType
         static const AZ::Crc32 OperatorOverride = AZ_CRC("OperatorOverride", 0x6b0e3ffb);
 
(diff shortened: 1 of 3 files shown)


Code/Framework/AzCore/AzCore/ScriptCanvas/ScriptCanvasAttributes.h
```diff
@@ -68,6 +68,8 @@ namespace AZ
 
         static const AZ::Crc32 BranchOnResult = AZ_CRC("BranchOnResult", 0x7a741f05);
 
+        static const AZ::Crc32 UseCompactStyle = AZ_CRC_CE("UseCompactStyle");
+
```
lsemp3d, line 72: Please add a comment that explains what this attribute is for

Gems/ScriptCanvas/Code/Include/ScriptCanvas/AutoGen/ScriptCanvasFunctionRegistry_Source.jinja
```diff
@@ -123,6 +123,10 @@ namespace ScriptCanvas
 {% endif %}
 {% if hasbranch|booleanTrue == true %}
                     ->Attribute(AZ::ScriptCanvasAttributes::BranchOnResult, branchResultInfo)
+{% endif %}
+{% set useCompactStyle = macros.CleanName(function.attrib['UseCompactStyle']) %}
+{% if useCompactStyle is defined and useCompactStyle.lower() == "true" %}
+                    ->Attribute(AZ::ScriptCanvasAttributes::UseCompactStyle, true)
 {% endif %}
```
lsemp3d, line 130: without seeing how you use the attribute, I wonder if it's even necessary when you could just have the template produce the necessary code. The risk of using attributes is that we lose the ability to change the templates and produce different code, instead we go back to the more traditional style in which we disconnect the code generation from what we produce.

PR #16051, opened 2023-05-31
Support scripted network spawning

mbalfour-amzn:
## What does this PR do?

This adds the ability to do scripted network spawning of .network.spawnable files by adding the NetworkSpawnableScriptAssetRef subclass. This subclass can be used with the existing Prefab/Spawn APIs in ScriptCanvas to specifically spawn and despawn *.network.spawnable assets.

For this change to work, SpawnableScriptAssetRef is now a base class that can be inherited from. The overall design of .spawnable assets is to allow for other systems to create more .spawnable subtypes over time for different purposes, which is what networking did with .network.spawnable. By making this a base class, other systems can now create specifically-filtered subtypes that still work with the base spawning APIs, but restrict asset selection to just the subtype of spawnable desired.

This PR also includes additional error messaging to SimplePlayerSpawnerComponent, so that anyone who tries to use it to spawn arbitrary prefabs will get errors letting them know they're using it wrong.

Also, the NetworkSpawnable class has been updated to allow empty assets to be selected. Otherwise, once you fill in an asset reference for SimplePlayerSpawnerComponent or NetworkSpawnableScriptAssetRef, you could never delete it.

After this PR is submitted, there will _likely_ be a few follow-up PRs:
1. Instead of using ChangeValidate to catch bad Spawnable selections, it would be nice to have a new AssetPicker attribute that allows for custom asset filtering so that only valid choices are shown in the first place.
2. The documentation for SimplePlayerSpawnerComponent needs to be updated to explain its use a little better.
3. We'll probably want documentation for NetworkSpawnableScriptAssetRef to explictly show how to use it to spawn networked spawnables.

## How was this PR tested?

Created a test level that spawns a .network.spawnable on the server and replicates it to the client.

image
image

### DIFF
--- a/Code/Framework/AzFramework/AzFramework/Spawnable/Script/SpawnableScriptAssetRef.cpp
+++ b/Code/Framework/AzFramework/AzFramework/Spawnable/Script/SpawnableScriptAssetRef.cpp
@@ -38,10 +38,12 @@ namespace AzFramework::Scripts
                     ->Attribute(AZ::Edit::Attributes::AutoExpand, true)
                     // m_asset
                     ->DataElement(AZ::Edit::UIHandlers::Default, &SpawnableScriptAssetRef::m_asset, "asset", "")
-                    ->Attribute(AZ::Edit::Attributes::ShowProductAssetFileName, false)
-                    ->Attribute(AZ::Edit::Attributes::HideProductFilesInAssetPicker, true)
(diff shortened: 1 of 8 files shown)


cgalvan (approved):

mbalfour-amzn:
As it turns out, I don't think we _want_ to expose being able to directly spawn .network.spawnable files. I'll open a new PR with just the couple of bugfixes from here, but without the NetworkSpawnableScriptAssetRef.

PR #16056, opened 2023-06-01
Fixed Vegetation Layer Debugger component not showing cube size field

cgalvan:
## What does this PR do?

The cube size field in the `Vegetation Layer Debugger` component is a float field, but was trying to register to use a color handler to display it, which is why no widget showed up. This was seemingly working by accident with the RPE (because it ended up showing a float widget somehow), but with the DPE we just didn't show a widget because the handler type mismatched.

Fixing the registration allows the field to be shown as expected.

BEFORE:
LayerDebugger_BEFORE

AFTER:
LayerDebugger_AFTER

## How was this PR tested?

Tested adding a `Vegetation Layer Component` to an entity and verified the cube size field now appears and is usable as expected.

### DIFF
--- a/Gems/Vegetation/Code/Source/Debugger/AreaDebugComponent.cpp
+++ b/Gems/Vegetation/Code/Source/Debugger/AreaDebugComponent.cpp
@@ -48,7 +48,7 @@ namespace Vegetation
                     "Vegetation Layer Debugger Config", "")
                     ->ClassElement(AZ::Edit::ClassElements::EditorData, "")
                     ->DataElement(AZ::Edit::UIHandlers::Color, &AreaDebugConfig::m_debugColor, "Debug Visualization Color", "")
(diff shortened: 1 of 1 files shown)


mbalfour-amzn (approved):
Lol thanks for fixing this!

AMZN-daimini (approved):
Could/should the DPE just show a label stating "Error - property handler type mismatch" or something along those lines?

dtamkin1 (approved):

gadams3 (approved):

cgalvan:
> Lol thanks for fixing this!

Yeah I'm not quite sure how it was working before, but works now :)

cgalvan:
> Could/should the DPE just show a label stating "Error - property handler type mismatch" or something along those lines?

There is a log message, but I think some components actually use not having a handler to show just a label with items grouped underneath it :/ I'll keep a lookout for them.

alexmontAmazon (approved):

Issue #15958, opened 2023-05-15
[Linux] User is unable to leave the Game Mode upon entering it


LB-KacperKapusta:
**Describe the bug**
After entering the Game Mode there is no way of leaving it which makes closing the Editor using standard methods also impossible. "Esc" button does not work. We were only able to terminate the Editor by killing the Terminal it has been run from.

The issue does not reproduce on Windows.
Please refer to the attached video for more information.

**Steps to reproduce**
1. Launch the Editor.
2. Create a level or open an already existing one.
3. Enter the Game Mode (Ctrl+G).
4. Try to leave the Game Mode.

**Expected behavior**
User is able to leave the Game Mode by pressing the "Esc" button.

**Actual behavior**
User is unable to leave the Game Mode by any means and restart of the Editor is required.

**Screenshots/Video**

https://github.com/o3de/o3de/assets/86953659/5f2ed01b-a195-41f1-85cc-de1baed70dd2

**Found in Branch**
Development
Installer build (https://o3debinaries.org/development/Latest/Linux/O3DE_latest.deb) downloaded on 5/15/2023 at 8:30 AM GMT+2

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
[018f149](https://github.com/o3de/o3de/commit/018f14991c4d849a74c8a6cccf1a7cefecb4960a) 

**Desktop/Device**
- Device: PC
- OS: Linux
- Version: Ubuntu 20.04.3 LTS
- CPU: AMD Ryzen 5 3600
- GPU: NVIDIA GeForce RTX 2060 SUPER
- Memory: 16GB

**Additional context**
[Editor.log](https://github.com/o3de/o3de/files/11477860/Editor.log)

nicholas-rh:
Reproduced on my machine using Fedora 38

nicholas-rh:
@LB-KacperKapusta Is this a recent issue that popped up or something that has been a problem for a while? I've been looking through the Qt code to try and figure out what's causing it to happen, but haven't found anything yet. My current guess is that some widget is stealing the focus when entering game mode but I haven't tracked it down/confirmed this yet, and also it wouldn't explain why it happens on Linux but not Windows.

LB-KacperKapusta:
@nicholas-rh This rather seems like a recent issue that popped up. Previously it did not happen and entering/exiting Game Mode is often times verified.

nicholas-rh:
Found a working commit & will run a git bisect to track down the cause.

[...]

PR #16074, opened 2023-06-01
Linux game mode esc exit hotfix

nicholas-rh:
This is a temporary hotfix (may not be the ideal solution) for the current issue of not being able to exit the game mode in Linux after entering. Opening a PR for it to gather feedback.

## What does this PR do?

Allow for exiting from game mode using ESC as intended.

#15958

## How was this PR tested?

Tested on Fedora 38, by checking to see if exiting game mode was possible by pressing ESC.

### DIFF
--- a/Gems/ImGui/Code/Source/ImGuiManager.cpp
+++ b/Gems/ImGui/Code/Source/ImGuiManager.cpp
@@ -27,8 +27,10 @@
 #include <AzFramework/Input/Devices/Touch/InputDeviceTouch.h>
 #include <AzFramework/Input/Devices/VirtualKeyboard/InputDeviceVirtualKeyboard.h>
 #include <AzFramework/Viewport/ViewportBus.h>
+#include <CrySystem/System.h>
 #include <IConsole.h>
 #include <imgui/imgui_internal.h>
+#include <ISystem.h>
 #include <sstream>
 #include <string>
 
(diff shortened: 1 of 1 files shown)


michalpelka (approved):
I've cherry-picked changes and tested them. It works as expected.

AMZN-alexpete:
Is the issue that IMGUI is swallowing the input, or that the ESC key input is not correctly captured?  IMGUI at one time, was not a required gem, in which case, would projects without IMGUI gem enabled still have this issue?

mbalfour-amzn:
@kberg-amzn is this possibly related to https://github.com/o3de/o3de/pull/15891 ?

Gems/ImGui/Code/Source/ImGuiManager.cpp
```diff
@@ -452,6 +454,21 @@ bool ImGuiManager::OnInputChannelEventFiltered(const InputChannel& inputChannel)
             {
                 ToggleThroughImGuiVisibleState();
             }
+
+            if (inputChannel.GetInputChannelId() == AzFramework::InputDeviceKeyboard::Key::Escape)
+            {
+                CSystem *cSys = (CSystem*)GetISystem();
```
lumberyard-employee-dm, line 460: standards: static_cast should be preferred here instead of a C-style cast according to the coding standards: https://github.com/o3de/sig-core/blob/main/governance/Coding-Standards-and-Style-Guide.md#casting

nicholas-rh: Fixed in 368d2201a0598928e0e9ee5b1fed65de4fd55658

lumberyard-employee-dm (approved):

nicholas-rh:
> It looks okay to me. It also sounds like this fix is considered a temp fix? It may be helpful to either open an issue about what might be the "real" full fix and link it here, so that it doesn't get lost.
> 
> My remaining concern would be how this behaves on other operating systems besides linux. If linux was the odd one out, adding this code in a non-linux section would affect the others, which were _presumably_ working.

Sorry for the delay I haven't forgotten about this, I set up a Windows guest VM but was running into some problems so I'll double check everything for compatibility once I get an actual installation up and running.

kberg-amzn:
> @kberg-amzn is this possibly related to #15891 ?

It's possible. I removed extra imgui and legacy cry console key binding in that change. I didn't add any new key binding that might swallow input unexpectedly, but that invisible cry console was definitely ticking and doing strange things like capturing all inputs and making it appear the player had lost all input binding. It could possibly still have been critical path for escape key handling and I missed it during testing.

kberg-amzn (approved):

Issue #16020, opened 2023-05-25
PhysX Joints not created in Multiplayer Mode


lgleim:
**Describe the bug**
PhysX Joint of any kind (e.g. Fixed or Hinge) between PhysX Rigid Bodies are not created for Network Spawnables. 

**Steps to reproduce**
1. In a multiplayer project, create two entities with PhysX Rigid Body Component
2. Connect the two entities using a PhysX Joint Component
3. Add a Network Transform (and Network Binding) components to both entities
4. Create a Prefab containing the previously created two entities joined by a joint
5. Set this prefab up to be spawned as a network spawnable
6. Launch the simulation with client and server
7. Observe the joints not being created

**Expected behavior**
Joints are being created and subsequently simulated upon spawning the prefab

**Actual behavior**
The Prefab is spawned but the Joint is not created

**Screenshots/Video**

https://github.com/o3de/o3de/assets/3898904/709627ef-3908-4057-8494-0f8cecaa5589

In the video, the prefab `test`, consisting of two boxes `base` and `top` connected by a fixed joint, is once statically included inside the level and once instantiated as a network spawnable upon multiplayer simulation using the simple multiplayer spawner component at `SpawnPoint1` (shown at the beginning of the video).

In the server console, the following corresponding warnings are logged:
![Screenshot from 2023-05-25 16-57-15](https://github.com/o3de/o3de/assets/3898904/2c58f3dc-93af-450e-b23c-12d3910e1008)
The client console does not show the warning related to `TransformComponent`s but logs the warning regarding Joints not being created twice.

**Found in Branch**
`main` and `development` 

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
`0b04b763f011dc87e07bba9e977183c364c2c25a` 

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Ubuntu 22.04

**Additional context**

- It appears that at the time that the PhysX gem's `JointComponent.Activate()` function is called, there is no rigid body attached to the Joint's base entity (i.e. the entity that the joint should attach the current entity to), causing joint construction to fail and the rigid bodies to just "fall apart"  based on the physics simulation. This apparently is only an issue for  `.network.spawnable` and not for `.spawnable` prefabs.
- Not having support for Joints on networked entities is a **blocker** for our current effort towards creating a multiplayer template for multirobot simulation.
[...]

PR #16089, opened 2023-06-01
Simple Player Spawner error checking fixes

mbalfour-amzn:
## What does this PR do?

1. Adds support for clearing the .network.spawnable asset again after it's been assigned on the Simple Player Spawner.
2. Adds error messages when trying to use the Simple Player Spawner with a prefab that has anything other than 1 networked entity in it.

## How was this PR tested?

Manually tested both cases above.

### DIFF
--- a/Gems/Multiplayer/Code/Include/Multiplayer/MultiplayerTypes.h
+++ b/Gems/Multiplayer/Code/Include/Multiplayer/MultiplayerTypes.h
@@ -234,6 +234,7 @@ namespace Multiplayer
                     ->Attribute(AZ::Edit::Attributes::AppearsInAddComponentMenu, AZ_CRC_CE("Game"))
 
                     ->DataElement(AZ::Edit::UIHandlers::Default, &NetworkSpawnable::m_spawnableAsset, "Network Spawnable Asset", "")
+                        ->Attribute(AZ::Edit::Attributes::AssetPickerTitle, "Network Spawnable Asset")
                         ->Attribute(AZ::Edit::Attributes::ChangeValidate, &NetworkSpawnable::ValidatePotentialSpawnableAsset)
                     ;
             }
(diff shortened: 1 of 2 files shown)


Gems/Multiplayer/Code/Include/Multiplayer/MultiplayerTypes.h
```diff
@@ -249,14 +250,17 @@ namespace Multiplayer
         }
 
         const auto potentialNetworkSpawnable = reinterpret_cast<AZ::Data::Asset<AzFramework::Spawnable>*>(newValue);
-        if (!potentialNetworkSpawnable->GetHint().ends_with(NetworkSpawnableFileExtension))
+        const auto& hint = potentialNetworkSpawnable->GetHint();
```
AMZN-alexpete, line 253: are hints removed in the final release game build .pak bundle output? Is that why you allowed `hint.empty()`?

Gems/Multiplayer/Code/Source/Components/SimplePlayerSpawnerComponent.cpp
```diff
@@ -152,13 +152,30 @@ namespace Multiplayer
...
+            // Success: The player prefab has exactly one networked entity in it.
             controlledEntity = entityList[0];
         }
         else
         {
-            AZLOG_WARN("Attempt to spawn prefab %s failed. Check that prefab is network enabled.", prefabEntityId.m_prefabName.GetCStr());
+            // Failure: The player prefab has too many networked entities in it.
+            AZLOG_ERROR(
+                "Attempt to spawn prefab '%s' failed, it contains too many networked entities. "
+                "A player prefab must only have a single networked entity inside of it.",
```
AMZN-alexpete, line 173: what happens if you have multiple networked entities?  Does it only work if the player prefab is the first one in the order or does something really bad happen?  reason I ask is I recall an occasion where a game was spawning multiple entities for players that were networked - one that was the visible character with all the various skin attachments, and another that was the character model used for silhouettes for when the player needed to be drawn behind walls (low-res proxy).

AMZN-alexpete (approved):

Gems/Multiplayer/Code/Include/Multiplayer/MultiplayerTypes.h
mbalfour-amzn: No, the hint.empty() is because you should be allowed to clear the reference on the Simple Network Player Spawner. Without this change, the following occurs in the Editor:
1. Add a Simple Network Player Spawner component to the level, which by default has an empty hint / asset.
2. Set the spawner asset to something valid.
3. Try to clear the asset by pressing the X. It refuses to let you clear the asset, it instead prints an error message.

With this change, the X works, and you can clear the asset.

It's just a small workflow improvement, it's a little bit annoying to have an asset set that you can't clear without deleting the component.

Gems/Multiplayer/Code/Source/Components/SimplePlayerSpawnerComponent.cpp
mbalfour-amzn: Right now, the other entities are getting untracked and possibly despawned. I'm not 100% sure, I haven't exactly gone tracing it to see _precisely_ what's happening. This came up though because a user was trying to spawn a multi-entity prefab using the player spawner, and it was breaking pretty badly: https://github.com/o3de/o3de/issues/16020 . 

This code gets back a list of spawned entities that it promptly ignores except for the first one, so in a best-case scenario those entities get immediately despawned. In a worst-case scenario, they leak and _never_ despawn. In either case, the other entities aren't associated with the player entity, they aren't passed around with OnPlayerJoin / OnPlayerLeave, only one will get set to Autonomous and the others won't, only one will get respawned on a Rejoin, etc.

I'm not personally enamored of this design, but until / unless the design changes, we should at least explicitly mark it as an error if someone tries to use this component in a way that isn't supported.

cgalvan (approved):

Gems/Multiplayer/Code/Include/Multiplayer/MultiplayerTypes.h
AMZN-Gene: Nice find/fix!

PR #16102, opened 2023-06-05
Fixed some node display label handling in DPE

### LINKED ISSUE #16066: DPE Inspector: MultiLineEdit fields showing labels that should be hidden
**Describe the bug**
Some MultiLineEdit fields that should take up the full width of the row are instead showing a label as well.

Comment

**Steps to reproduce**

3. Create a new entity.
4. Add a `Comment` component to the entity.

**Expected behavior**
MultlineEdit widget should take up the full width of the row.

**Actual behavior**
Label is shown in first column of the row.

cgalvan:
## What does this PR do?

Fixes #16066 

There were some cases where the DPE wasn't following the same logic for selecting a node's display label as the RPE. Part of this fix was consolidating the separate checks into a single location for easier readability of how a node's display label is chosen. I also re-worked it to follow a similar logic flow as the method the RPE uses (`AzToolsFramework::GetNodeDisplayName`).

This method will be expanded in further PRs as we handle some of the more complicated label name override cases (which the RPE handles in a different method by modifying the `m_name` in the metadata prior to reaching `GetNodeDisplayName`).

Two other changes made during this PR:
1. Changed the `EnumerateInstanceCallContext` to use the `ENUM_ACCESS_FOR_READ` instead of write. The RPE also uses the read flag. I don't believe we need to be modifying the node data as we enumerate.
2. Added small helper method to check if a node has valid `GetElementEditMetadata`

The simplest use-case to demonstrate this particular fix can be seen here:

BEFORE:
NodeLabel_BEFORE

AFTER:
NodeLabel_AFTER

## How was this PR tested?

Tested the repro steps as noted, and also tested various components to verify labels still show as expected (that were previously working). Also verified basic container case of elements labels still work as intended (e.g. `[0]`, `[1]`, etc...)

### DIFF
--- a/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/Reflection/LegacyReflectionBridge.cpp
+++ b/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/Reflection/LegacyReflectionBridge.cpp
@@ -166,6 +166,16 @@ namespace AZ::Reflection
                 AZ::SerializeContext::IDataContainer* m_parentContainerInfo = nullptr;
                 void* m_parentContainerOverride = nullptr;
                 void* m_containerElementOverride = nullptr;
+
+                const AZ::Edit::ElementData* GetElementEditMetadata() const
+                {
+                    if (m_classElement)
+                    {
+                        return m_classElement->m_editData;
+                    }
+
+                    return nullptr;
+                }
             };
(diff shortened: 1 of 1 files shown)


dtamkin1 (approved):
Very cool changes!

alexmontAmazon (approved):
Nice! Hooray for RPE compatibility!

PR #16103, opened 2023-06-05
Add support to the ByteContainerStream for an `Open` function

lumberyard-employee-dm:
### ByteContainerStream changes
The ByteContainerStream now supports an Open member, that can be used to associate a different container buffer with the ByteContainerStream instance.

The ByteContainerStream member functions that mutate the underlying container buffer (`Seek`, `Read`, `Write`, etc...), now check if the buffer is in an "open" state before performing any of those operations.
`AZ_Error` messages gets logged in that case.

### Task System changes
Updated the `TaskExecutor::Submit` function to not increment the `m_graphsRemaining` member until it has been validated there is at least one compiled task to run.

### Archive Writer PR feedback adressing.

The feedback from the ArchiveWriter PR post commit has been addressed: https://github.com/o3de/o3de/pull/16052

## How was this PR tested?

Added UnitTest to validate the ByteContainerStream functionality.

### DIFF
--- a/Code/Framework/AzCore/Tests/GenericStreamTests.cpp
+++ b/Code/Framework/AzCore/Tests/GenericStreamTests.cpp
@@ -133,3 +133,125 @@ TEST_F(GenericStreamTest, MemoryStreamWriteFromStream_MoreBytesThanTempBuffer_Co
     // just to help catch any possible errors.
     TestMemoryStreamWriteFromStream(aznumeric_cast<size_t>(AZ::IO::GenericStream::StreamToStreamCopyBufferSize * 2.6f));
 }
+
+namespace AZ::IO::Test
+{
+    class ByteContainerStreamTest
+        : public UnitTest::LeakDetectionFixture
+    {
+    };
+
+    TEST_F(ByteContainerStreamTest, IsDefaultConstructible)
+    {
+        using ContainerType = AZStd::vector<AZStd::byte>;
+        AZ::IO::ByteContainerStream<ContainerType> byteContainerStream;
+        EXPECT_FALSE(byteContainerStream.IsOpen());
+    }
+
+    TEST_F(ByteContainerStreamTest, CanBeOpened_PostConstruction)
+    {
+        using ContainerType = AZStd::vector<AZStd::byte>;
+        AZ::IO::ByteContainerStream<ContainerType> byteContainerStream;
+        EXPECT_FALSE(byteContainerStream.IsOpen());
+
+        ContainerType byteBuffer;
+        EXPECT_TRUE(byteContainerStream.Open(byteBuffer));
+        ASSERT_TRUE(byteContainerStream.IsOpen());
+        EXPECT_EQ(0, byteContainerStream.GetCurPos());
+    }
+
+    TEST_F(ByteContainerStreamTest, Open_OnAlreadyOpenStream_Fails)
+    {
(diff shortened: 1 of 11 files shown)


Code/Framework/AzCore/Tests/GenericStreamTests.cpp
```diff
@@ -133,3 +133,125 @@ TEST_F(GenericStreamTest, MemoryStreamWriteFromStream_MoreBytesThanTempBuffer_Co
...
+        ASSERT_TRUE(byteContainerStream.IsOpen());
+
+        AZ::IO::SizeType seekPos = byteContainerStream.GetLength();
+        EXPECT_EQ(seekPos, byteContainerStream.GetLength());
+        // Seek to the end of the stream
+        byteContainerStream.Seek(seekPos, AZ::IO::GenericStream::SeekMode::ST_SEEK_BEGIN);
+        EXPECT_EQ(seekPos, byteContainerStream.GetCurPos());
+
+        // Yes, "byte" was on purpose
+        constexpr AZStd::string_view secondString("Goodbyte World");
```
mbalfour-amzn, line 238: ouch

Gems/Archive/Code/Include/Archive/Tools/ArchiveWriterAPI.h
```diff
@@ -48,7 +48,11 @@ namespace Archive
         ArchiveWriterErrorString m_errorMessage;
     };
 
-    //! Stores settings to configure how Archive Writer Settings
+    //! Stores settings to configure how Archive Writer perform specific operations
+    //! This can be used to change if the Archive TOC should be compressed on Commit
+    //! It also supports configuring an optional error callback to invoke if an
+    //! error occurs in a function that can't return a outcome value such as a constructor/destructor
+    //! The number of compression task that can run in parallel is also configurable
```
mbalfour-amzn, line 55:
```suggestion
    //! Stores settings to configure how Archive Writer performs specific operations
    //! This can be used to change if the Archive TOC should be compressed on Commit
    //! It also supports configuring an optional error callback to invoke if an
    //! error occurs in a function that can't return a outcome value such as a constructor/destructor
    //! The number of compression tasks that can run in parallel is also configurable
```

Gems/Archive/Code/Source/Clients/ArchiveTOC.h
```diff
@@ -43,6 +43,8 @@ namespace Archive
 
         //! Wrapper path structure to ensure the Table of Contents only contain paths that
         //! uses the Posix Path Separator '/'
+        //! This is used to normalize how the paths within the Table of Cotnents
```
mbalfour-amzn, line 46:
```suggestion
        //! This is used to normalize how the paths within the Table of Contents
```

Gems/Archive/Code/Source/Clients/ArchiveTOC.h
```diff
@@ -80,7 +82,7 @@ namespace Archive
         };
 
         //! vector storing a copy of each file path in memory
-//! It's length matches the value of m_fileCount
+        //! It's length matches the value of m_fileCount
```
mbalfour-amzn, line 85:
```suggestion
        //! Its length matches the value of m_fileCount
```

mbalfour-amzn (approved):

cgalvan (approved):

PR #16104, opened 2023-06-05
Fixed read-only handling in DPE to get applied to UIElements as well

### LINKED ISSUE #16067: DPE Inspector: Read-only attribute not being applied to UIElements
**Describe the bug**
The read-only attribute isn't being parsed for UIElements (e.g. buttons).

Transform

**Steps to reproduce**

3. Create a new entity.
4. Add a `Transform` component to the entity.
5. Click the "Add non-uniform scale" button on the `Transform` component.

**Expected behavior**
The `Add non-uniform scale` button should be marked as read-only.

**Actual behavior**
The `Add non-uniform scale` button is still enabled/clickable.

cgalvan:
## What does this PR do?

Fixes #16067 

Fixed an issue with the DPE handling of read-only attributes that was preventing them being handled for UIElements (e.g. buttons, groups). I moved the check to the generic `checkAttribute` location in the `LegacyReflectionBridge` so it gets all the same benefits of that method, but kept the special-case to ignore it if we'd already detected the ancestor node is disabled.

Here is a simple use-case in the `Transform` component where the `Add non-uniform scale` button should get disabled based on whether or not the  `Non-uniform Scale component` is present for this Entity. Before this change, the button is just always enabled.

BEFORE:
ReadOnly_BEFORE

AFTER:
ReadOnly_AFTER

## How was this PR tested?

Tested according to repro steps listed in bug and above. Tested various other read-only usages on normal data elements (e.g. Mesh Stats in Mesh component) and verified they still work as expected.

### DIFF
--- a/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/PropertyEditorNodes.cpp
+++ b/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/PropertyEditorNodes.cpp
@@ -40,6 +40,7 @@ namespace AZ::DocumentPropertyEditor::Nodes
     {
         system->RegisterNode<NodeWithVisiblityControl>();
         system->RegisterNodeAttribute<NodeWithVisiblityControl>(NodeWithVisiblityControl::Visibility);
+        system->RegisterNodeAttribute<NodeWithVisiblityControl>(NodeWithVisiblityControl::ReadOnly);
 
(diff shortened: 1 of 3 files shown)


dtamkin1 (approved):

alexmontAmazon (approved):
sweet :)

gadams3 (approved):

PR #16106, opened 2023-06-06
MeshFeatureProcessor: Consider emissive intensity unit for raytracing materials

msat-huawei:
## What does this PR do?

The light intensity in the StandardPBR material is defined in Ev100 in the Editor. However the shader expects Nits.
For the non-raytracing materials the intensity is converted using a Functor that is registered in the material (in `EmissivePropertyGroup.json`). This Functor is never called for the raytracing material.

My solution is to search for the specific Functor when creating the raytracing material, and convert the unit using this functor.

The solution is kind of dirty, but I can't think of another way, to convert the unit, with the current material system. The raytracing support, for materials, is quite limited currently.

## How was this PR tested?

Vulkan and Windows

### DIFF
--- a/Gems/Atom/Feature/Common/Code/Source/Material/ConvertEmissiveUnitFunctor.h
+++ b/Gems/Atom/Feature/Common/Code/Source/Material/ConvertEmissiveUnitFunctor.h
@@ -31,7 +31,7 @@ namespace AZ
             using RPI::MaterialFunctor::Process;
             void Process(RPI::MaterialFunctorAPI::RuntimeContext& context) override;
             void Process(RPI::MaterialFunctorAPI::EditorContext& context) override;
-            float GetProcessedValue(float originalEmissiveIntensity, uint32_t lightUnitIndex);
+            float GetProcessedValue(float originalEmissiveIntensity, uint32_t lightUnitIndex) const;
 
         private:
 
--- a/Gems/Atom/Feature/Common/Code/Source/Material/ConvertEmissiveUnitFunctor.h
+++ b/Gems/Atom/Feature/Common/Code/Source/Material/ConvertEmissiveUnitFunctor.h
@@ -31,6 +31,8 @@ namespace AZ
             using RPI::MaterialFunctor::Process;
             void Process(RPI::MaterialFunctorAPI::RuntimeContext& context) override;
             void Process(RPI::MaterialFunctorAPI::EditorContext& context) override;
+            float GetProcessedValue(float originalEmissiveIntensity, uint32_t lightUnitIndex);
+
         private:
 
             AZ::RPI::MaterialPropertyIndex m_intensityPropertyIndex;
--- a/Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
@@ -6,20 +6,23 @@
  *
  */
 
-#include <Atom/RHI/RHISystemInterface.h>
-#include <Atom/RHI/RHIUtils.h>
-#include <Atom/RHI.Reflect/InputStreamLayoutBuilder.h>
-#include <Atom/Feature/RenderCommon.h>
+#include <Atom/Feature/CoreLights/PhotometricValue.h>
 #include <Atom/Feature/Mesh/MeshCommon.h>
 #include <Atom/Feature/Mesh/MeshFeatureProcessor.h>
 #include <Atom/Feature/Mesh/ModelReloaderSystemInterface.h>
+#include <Atom/Feature/RenderCommon.h>
 #include <Atom/Feature/Utils/GpuBufferHandler.h>
+#include <Atom/RHI.Reflect/InputStreamLayoutBuilder.h>
+#include <Atom/RHI/RHISystemInterface.h>
+#include <Atom/RHI/RHIUtils.h>
+#include <Atom/RPI.Public/AssetQuality.h>
+#include <Atom/RPI.Public/Culling.h>
 #include <Atom/RPI.Public/Model/ModelLodUtils.h>
 #include <Atom/RPI.Public/Model/ModelTagSystemComponent.h>
-#include <Atom/RPI.Public/Scene.h>
-#include <Atom/RPI.Public/Culling.h>
 #include <Atom/RPI.Public/RPIUtils.h>
-#include <Atom/RPI.Public/AssetQuality.h>
+#include <Atom/RPI.Public/Scene.h>
+
+#include <Material/ConvertEmissiveUnitFunctor.h>
 
 #include <Atom/Utils/StableDynamicArray.h>
 #include <ReflectionProbe/ReflectionProbeFeatureProcessor.h>
@@ -30,6 +33,7 @@
 
 #include <AtomCore/Instance/InstanceDatabase.h>
 
+#include <AzCore/Asset/AssetCommon.h>
 #include <AzCore/Console/IConsole.h>
 #include <AzCore/Jobs/Algorithms.h>
 #include <AzCore/Jobs/JobCompletion.h>
@@ -39,8 +43,7 @@
 #include <AzCore/RTTI/RTTI.h>
 #include <AzCore/RTTI/TypeInfo.h>
 #include <AzCore/Serialization/SerializeContext.h>
-#include <AzCore/Asset/AssetCommon.h>
-#include <AzCore/Name/NameDictionary.h>
+
 
 #include <algorithm>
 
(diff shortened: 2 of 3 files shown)


jhmueller-huawei:
I agree that the solution is hacky, but I cannot quickly come up with a better solution. Maybe someone else knows a better way to do this without changing the whole material system? Otherwise, I'd approve this PR. It seems to be another case, where ray-tracing appears to be somewhat of an afterthought in Atom, especially in conjunction with the material system. I wonder how this could be improved in general so that somewhat hacky solutions like this are not necessary anymore, but this is not the point of this PR.

Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
```diff
@@ -2230,7 +2232,29 @@ namespace AZ
...
+                                {
+                                    auto emissiveFunctor = azdynamic_cast<ConvertEmissiveUnitFunctor*>(functor);
+                                    if (emissiveFunctor != nullptr)
+                                    {
+                                        intensity = emissiveFunctor->GetProcessedValue(intensity, unit);
+                                        foundEmissiveUnitFunctor = true;
+                                        break;
+                                    }
+                                }
+                                material->GetAsset()->GetId().ToFixedString().c_str();
```
moudgils, line 2252: Is this a typo?

Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
```diff
@@ -2230,7 +2232,29 @@ namespace AZ
                             propertyIndex = material->FindPropertyIndex(s_emissive_intensity_Name);
                             if (propertyIndex.IsValid())
                             {
-                                subMesh.m_emissiveColor *= material->GetPropertyValue<float>(propertyIndex);
+                                auto unitPropertyIndex = material->FindPropertyIndex(s_emissive_unit_Name);
```
moudgils, line 2235: Add a comment indicating what you are trying to do here, i.e adjusting for light unit.

Gems/Atom/Feature/Common/Code/Source/Material/ConvertEmissiveUnitFunctor.h
```diff
@@ -31,6 +31,8 @@ namespace AZ
             using RPI::MaterialFunctor::Process;
             void Process(RPI::MaterialFunctorAPI::RuntimeContext& context) override;
             void Process(RPI::MaterialFunctorAPI::EditorContext& context) override;
+            float GetProcessedValue(float originalEmissiveIntensity, uint32_t lightUnitIndex);
```
AMZN-alexpete, line 34:
```suggestion
            float GetProcessedValue(float originalEmissiveIntensity, uint32_t lightUnitIndex) const;
```

Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
```diff
@@ -128,9 +129,10 @@ namespace AZ
 
             m_rayTracingFeatureProcessor = GetParentScene()->GetFeatureProcessor<RayTracingFeatureProcessor>();
             m_reflectionProbeFeatureProcessor = GetParentScene()->GetFeatureProcessor<ReflectionProbeFeatureProcessor>();
-            m_handleGlobalShaderOptionUpdate = RPI::ShaderSystemInterface::GlobalShaderOptionUpdatedEvent::Handler
-            {
-                [this](const AZ::Name&, RPI::ShaderOptionValue) { m_forceRebuildDrawPackets = true; }
+            m_handleGlobalShaderOptionUpdate =
+                RPI::ShaderSystemInterface::GlobalShaderOptionUpdatedEvent::Handler{ [this](const AZ::Name&, RPI::ShaderOptionValue)
+                                                                                     {
+                                                                                         m_forceRebuildDrawPackets = true; }
```
AMZN-alexpete, line 135: this formatting change can probably be reverted

Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
```diff
@@ -2230,7 +2232,29 @@ namespace AZ
                             propertyIndex = material->FindPropertyIndex(s_emissive_intensity_Name);
                             if (propertyIndex.IsValid())
                             {
-                                subMesh.m_emissiveColor *= material->GetPropertyValue<float>(propertyIndex);
+                                auto unitPropertyIndex = material->FindPropertyIndex(s_emissive_unit_Name);
+                                AZ_Assert(
+                                    propertyIndex.IsValid(),
+                                    "MeshFeatureProcessor: Materials with an emissive intensity need a unit for the intensity");
```
AMZN-alexpete, line 2238:
```suggestion
                                    "MeshFeatureProcessor: Emissive intensity property missing in material %s. Materials with an emissive intensity need a unit for the intensity.",
                                    material->GetAsset()->GetId().ToFixedString().c_str());
```

Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
```diff
@@ -2230,7 +2232,29 @@ namespace AZ
                             propertyIndex = material->FindPropertyIndex(s_emissive_intensity_Name);
                             if (propertyIndex.IsValid())
                             {
-                                subMesh.m_emissiveColor *= material->GetPropertyValue<float>(propertyIndex);
+                                auto unitPropertyIndex = material->FindPropertyIndex(s_emissive_unit_Name);
+                                AZ_Assert(
```
AMZN-alexpete, line 2236: Maybe this should be a warning so it's visible when not in debug?
```suggestion
                                AZ_WarningOnce(
```

gadams3: In addition to changing this to a warning, the rest of the logic should be wrapped in an if block. No reason to execute any of the other logic if the index is invalid. That's consistent with checks involving other material property indexes.

Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
gadams3, line 2252: It looks like this line can be deleted

Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
```diff
@@ -2230,7 +2232,29 @@ namespace AZ
...
+                                        foundEmissiveUnitFunctor = true;
+                                        break;
+                                    }
+                                }
+                                material->GetAsset()->GetId().ToFixedString().c_str();
+                                AZ_Assert(
+                                    foundEmissiveUnitFunctor,
+                                    "MeshFeatureProcessor: Could not find ConvertEmissiveUnitFunctor for material %s",
+                                    material->GetAsset()->GetId().ToFixedString().c_str());
+                                subMesh.m_emissiveColor *= intensity;
```
gadams3, line 2257: Same as with the material property index, I think there should be an if block checking for the functor pointer and the assert should be a warning because it's data-driven and possible that other materials will not have the functor.

gadams3 (commented):
Where is this being tested, editor or atom sample viewer? Does the functor get called correctly for the non ray tracing path? If so, what is the ray tracing path doing differently with respect to the material? The functor should get executed whenever the material is initialized, reinitialized, or properties change. That may require calling compile on the material instance, which will happen automatically with anything that uses the material component if the changes go through the material component interface and not directly through the instance. There are outstanding tasks to update the material system to automatically recompile modified material instances.

msat-huawei:
> Where is this being tested, editor or atom sample viewer? Does the functor get called correctly for the non ray tracing path? If so, what is the ray tracing path doing differently with respect to the material? The functor should get executed whenever the material is initialized, reinitialized, or properties change. That may require calling compile on the material instance, which will happen automatically with anything that uses the material component if the changes go through the material component interface and not directly through the instance. There are outstanding tasks to update the material system to automatically recompile modified material instances.

It's tested in the editor, with a custom raytracing pass.
The functor is correctly called in the non-raytracing path. All functors are called in the `Material::Compile` function, and they are responsible for setting members in the Material Srg. The compilation of the Material does not change the intensity value in the material instance, only the one in the Srg.

moudgils (approved):

gadams3:
> > Where is this being tested, editor or atom sample viewer? Does the functor get called correctly for the non ray tracing path? If so, what is the ray tracing path doing differently with respect to the material? The functor should get executed whenever the material is initialized, reinitialized, or properties change. That may require calling compile on the material instance, which will happen automatically with anything that uses the material component if the changes go through the material component interface and not directly through the instance. There are outstanding tasks to update the material system to automatically recompile modified material instances.
> 
> It's tested in the editor, with a custom raytracing pass. The functor is correctly called in the non-raytracing path. All functors are called in the `Material::Compile` function, and they are responsible for setting members in the Material Srg. The compilation of the Material does not change the intensity value in the material instance, only the one in the Srg.

I understand now. IIRC, there were multiple attempts to handle the usability issues surrounding supporting multiple lighting units. One attempt allowed the functor to convert the value based on the selected unit but there were other issues and support for doing so was removed in general. I think the current implementation stores a unique material property value for each unit. Changing the active unit setting would just display a different value in the material editor. I don't think they are synchronized with each other. So the functor or shader code has to select or convert the value at runtime. I'm trying to remember the complications with changing this to set a material property value instead of setting the shader value directly. It's probably just synchronization issues updating the material editor or other systems to reflect the changes.

gadams3 (approved):

msat-huawei:
I forgot to sign off the last commit, so I had to push it again

jhmueller-huawei (approved):

PR #16123, opened 2023-06-07
Added support for DPE to handle visibility attribute values of boolean or 0/1

### LINKED ISSUE #16079: DPE Inspector: Visibility attribute missing support for bool or 0/1
**Describe the bug**
The RPE has special logic for just the Visibility attribute that lets the user return a bool or 0/1 (for Hide/Show) instead of the CRC. The DPE needs to add support for these special cases because it is used by a lot of components.

Light, PhysX Force Region, m_lodSampleRates on Simple LOD Distance component, PhysX Primitive Collider,
    PhysX Ragdoll - "Maximum Mass Ratio", UI Canvas Asset Ref,
    Vegetation Asset List - "External Assets",
    Specular Reflections (Level component)

NOTE: On these additional components, the behavior is showing an extra button that produces a crash, but the Visibility attribute should be making them mutually exclusive:
PhysX Heightfield Collider, PhysX Mesh Collider, PhysX Primitive Collider, PhysX Shape Collider

**Steps to reproduce**

3. Create a new entity.
4. Add a `Light` component to the entity.

**Expected behavior**
All the fields besides Light Type should be hidden until you pick a Light Type.

**Actual behavior**
All the fields are visible.

cgalvan:
## What does this PR do?

Fixes #16079 

Most attributes only expect a certain type of value, but the `Visibility` attribute is a rare exception that has additional special-case handling in the RPE. In addition to the default type of a hash, it also supports values of 0/1 and boolean false/true, for Hide/Show respectively. I updated the visibility parsing to account for these cases in the DPE.

The boolean case is used by many components to Show/Hide fields. The best example of this is the Light component, where before this fix it was showing every field all at once, instead of hiding them until the user selects a light type.

BEFORE:
VisibilityBoolean_BEFORE

AFTER:
VisibilityBoolean_AFTER

NOTE: There are still some resizing bugs and grouping issues that are being addressed separately.

## How was this PR tested?

Tested all components mentioned in #16079 and verified their boolean visibility checks now work as expected.

### DIFF
--- a/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/Reflection/LegacyReflectionBridge.cpp
+++ b/Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/Reflection/LegacyReflectionBridge.cpp
@@ -183,6 +183,8 @@ namespace AZ::Reflection
             using HandlerCallback = AZStd::function<bool()>;
             AZStd::unordered_map<AZ::TypeId, HandlerCallback> m_handlers;
 
+            static constexpr auto VisibilityBoolean = AZ::DocumentPropertyEditor::AttributeDefinition<bool>("VisibilityBoolean");
+
             // Specify whether the visit starts from the root of the instance.
             bool m_visitFromRoot = true;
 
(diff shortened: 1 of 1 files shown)


Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/Reflection/LegacyReflectionBridge.cpp
```diff
@@ -688,11 +690,48 @@ namespace AZ::Reflection
...
+                                AZ::u32 visibilityNumericValue = static_cast<AZ::u32>(visibility);
+                                if (visibilityNumericValue == 0)
+                                {
+                                    visibility = PropertyVisibility::Hide;
+                                }
+                                else if (visibilityNumericValue == 1)
+                                {
+                                    visibility = PropertyVisibility::Show;
+                                }
+                                return;
```
AMZN-daimini, line 722: Could a switch work better here?

alexmontAmazon (approved):
looks good! Soon we won't be able to tell if we're using the RPE or DPE...

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/Reflection/LegacyReflectionBridge.cpp
```diff
@@ -688,11 +690,48 @@ namespace AZ::Reflection
...
+                            {
+                                auto visibilityBoolValue = VisibilityBoolean
+                                    .DomToValue(VisibilityBoolean.LegacyAttributeToDomValue(instance, it->second));
+                                if (visibilityBoolValue.has_value())
+                                {
+                                    bool isVisible = visibilityBoolValue.value();
+                                    visibility = isVisible ? PropertyVisibility::Show : PropertyVisibility::Hide;
+                                    return;
+                                }
+                            }
```
AMZN-daimini, line 734: nit: Since the previous if always returns, we could remove the else and avoid the indentation.

Also, initialize the auto inside the if:
```
if (auto visibilityBoolValue = VisibilityBoolean.DomToValue(VisibilityBoolean.LegacyAttributeToDomValue(instance, it->second)); visibilityBoolValue.has_value())
{
    bool isVisible = visibilityBoolValue.value();
    visibility = isVisible ? PropertyVisibility::Show : PropertyVisibility::Hide;
    return;
}
```

or even better, if `visibilityBoolValue` can implicitly convert to bool:

```
if (auto visibilityBoolValue = VisibilityBoolean.DomToValue(VisibilityBoolean.LegacyAttributeToDomValue(instance, it->second)))
{
// ...
```

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/Reflection/LegacyReflectionBridge.cpp
cgalvan: Sure thing!

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/Reflection/LegacyReflectionBridge.cpp
cgalvan: Good call I'll clean up!

dtamkin1 (approved):

PR #16129, opened 2023-06-07
Adding Archive Reader implementation

lumberyard-employee-dm:
### Archive Gem changes
Adding complete implementation of the new Archive ArchiveReader API

This pairs with the implementation of the ArchiveWriter API and
therefore can be used to write and O3AR archive files.

The Archive Reader supports the following operations

1. `ListFileInArchive` - This function can query a file within an
   archive by file path or file token which was queried previously from
   a ListFileInArchive call
1. `ExtractFileFromArchive` - This function can extract file from the
   archive using either the relative file path of the contained file or
   the file token queried from a call to `ListFileInArchive`
   The file is extracted to a user provided allocated buffer via an
   `AZStd::span<AZStd::byte>` paramter
1. `EnumerateFilesInArchive` - This functions enumerates all non-deleted files within the archive and invokes a user provided callback with the ArchiveListFileResult structure for each file. The `ArchiveListFileResult` structure is the same type returned by `ListFileInArchive`

#### Archive Factory addition

Added an Archive Reader and Writer Factory implementation

The Archive Reader Factory allows IArchiveReader instances to be created
by external Client gem modules by depending on the Archive.Clients.API target

The Archive Writer Factory allows IArchiveWriter instances to be created
by external Tool gem modules by depending on the Archive.Tools.API
target.

This other Gems such as the AssetValidation or some other library such as the AssetBundler to create Archive class instances needed to interact with the API

### Compression Gem changes
Adding a new interface function of `GetCompressionAlgorithmName` to the ICompressionInterface/IDecompressionInterface interfaces.
Overriding this function can used to provide a human readable name associated with the Compression algorithm

The plan for this change is to make it easier for a future CLI to allow uses to specify the compression algorithm to use when adding files to an archive.

## How was this PR tested?

Added UnitTest for the ArchiveReader functionality to the Archive.Editor.Tests module
Validated the existing ArchiveWriter test still passes.

### DIFF
--- a/Code/Framework/AzCore/AzCore/Math/Crc.h
+++ b/Code/Framework/AzCore/AzCore/Math/Crc.h
@@ -80,13 +80,16 @@ namespace AZ
          * Calculates the value from a block of raw data.
          */
         Crc32(const void* data, size_t size, bool forceLowerCase = false);
-        constexpr Crc32(const uint8_t* data, size_t size, bool forceLowerCase = false);
-        constexpr Crc32(const char* data, size_t size, bool forceLowerCase = false);
+        template<class ByteType, class = AZStd::enable_if_t<sizeof(ByteType) == 1>>
+        constexpr Crc32(const ByteType* data, size_t size, bool forceLowerCase = false);
+        constexpr Crc32(AZStd::span<const AZStd::byte> inputSpan);
 
         constexpr void Add(AZStd::string_view view);
         void Add(const void* data, size_t size, bool forceLowerCase = false);
-        constexpr void Add(const uint8_t* data, size_t size, bool forceLowerCase = false);
-        constexpr void Add(const char* data, size_t size, bool forceLowerCase = false);
+        template<class ByteType>
+        constexpr auto Add(const ByteType* data, size_t size, bool forceLowerCase = false)
+            -> AZStd::enable_if_t<sizeof(ByteType) == 1>;
+        constexpr void Add(AZStd::span<const AZStd::byte> inputSpan);
 
         constexpr operator u32() const               { return m_value; }
 
@@ -101,8 +104,9 @@ namespace AZ
         // A constant expression cannot contain a conversion from const-volatile void to any pointer to object type
         // nor can it contain a reinterpret_cast, therefore overloads for const char* and uint8_t are added
         void Set(const void* data, size_t size, bool forceLowerCase = false);
-        constexpr void Set(const uint8_t* data, size_t size, bool forceLowerCase = false);
-        constexpr void Set(const char* data, size_t size, bool forceLowerCase = false);
+        template<class ByteType>
+        constexpr auto Set(const ByteType* data, size_t size, bool forceLowerCase = false)
+            -> AZStd::enable_if_t<sizeof(ByteType) == 1>;
         constexpr void Combine(u32 crc, size_t len);
 
         u32 m_value;
--- a/Code/Framework/AzCore/AzCore/PlatformDef.h
+++ b/Code/Framework/AzCore/AzCore/PlatformDef.h
@@ -281,35 +281,45 @@
 
 // define builtin functions used by char_traits class for efficient compile time and runtime
 // operations
+// The list of builtins can be found on the clang and GCC documentation pages at
+// https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html
+// https://clang.llvm.org/docs/LanguageExtensions.html#builtin-functions
+
+// NOTE: __builtin_memcpy and __builtin_memmove for GCC is not usable in a compile time context currently
+// The following is a permalink to godbolt that checks which builtin intrinsics
+// are available for GCC 10.1+ and clang 12.0.0+
+// The non-short link is being used here to not rely on a URL shortener service
+// https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGe1wAyeAyYAHI%2BAEaYxBIAHKQADqgKhE4MHt6%2BekkpjgJBIeEsUTFc8XaYDmlCBEzEBBk%2Bfly2mPZ5DDV1BAVhkdFxtrX1jVktCsM9wX3FA2UAlLaoXsTI7BzmAMzByN5YANQmm24TxMHAAPoAbniYAO5H2CYaAILPLwD0H/sAatF4NEwCn2ES8dEcDH2wQIZwYKWQwIICCYBH2dUw%2By8CiYEXoUMhTH2aBYCToGMcbCJskwqgI7zQcIINISxHxqKZEwuoPBwQUEHmhwA7FZXvsxVTGczWRN0CAQKdztdbnd9gplqsMUcACKHMxmAASbVoqH2AHUSLR0OYzEcRS9xRKJlLVQRZSA7gqjEr7vs7ngsEJ1WtDpsdQFrYbaMazRarXrbe93g7kMjWVgJiYAKwWTY2zNakzCwsFzZ2h13FN1C6ov1YLVAunZ3NZgtFwUlu1J8VffZsFjIBIATxBYNoEIUibM2yo%2BwuF2RCi5o4hEDn3LHwQufYHg/mXbFa%2BXm%2B3Q4g6YIpFVQcwADp0CimPzL2qVmsbykAF6Yfl70uTzauAC/4ArO85MIu64roePIMBcdwnru%2B6gZBm7wZg/anrWmD1hMl5YYGr63vetRPr6/qYARGrvngX4/gmrxbIBVCJgxordt8fYsKgVwYihcLATOc4LkuMGriJG6wZx3GYHubEHuJEJbuhXE8WeDbPted4PqRL5UZ%2B37zL%2BnZTkxAmgcJfFiXxcFSTxsn2uK0ESTZynSRAWE4ReZEBppxGPos3kUZp%2Bl0X%2BDEmQw%2BDMaxDlij2pz0JCfETuF07meBCnBFZR6wQljD2Q6TmKXlDAQLpb5%2BaFxkAZFQGpSBQkZZZRWoQiiUFY5mWwRWCiJe55GURV2mGfRbwRVFLGxfsPbbggrLJWZjUQTl2UwUp/ZzR18nWbNxBlb5w2XhoGmEdRtEjWFY01RN9WCWBy2iS13W7VtyE5S5G17fhB0kQFx2BYNt4hRd1WmTFDozehyAkiOMEpVdDX3V1q3OQhr1Pet0MJPtp2VQFEZGia5rEJa1onXpNEGUZ/5gwjd0WStGNof2JLo11H1Y/1Pm44d%2BzhnqkbRsTpN6nhA3BZTVU07V0VvHJ%2BzEJgBArJCGijcWk3Qr2TBZfZhZluKDJOqoLJsgrQJeGOIY6hyBBdXy1Pg%2BKivK8QkKKwolt0pdxYcIstCcJmvB%2BBwWikKgnBuNY1hXoRuqbDwpAEJofuLAA1iAmybDeWe53n%2BcAGz6JwkjByn4ecLwCggMdyeh37pBwLAMCICgqAkmSZAUBAxKkvQMTAFwCekFgNxrD8yoAPIJIwnCJzQY7RNXEAROXETBHUg5z7w6/MMQg6TxE2iVHXifEmwgiTwwtBb/XI%2BYKCwBuGItDV9wvBYCwhjAOId/4IrVQeJvzDjSSoXgmTb3IIINo5daB4AiMQTeHgsDlxhHgFg29FhUAMMABQE97jT1nu/GQggRBiHYFIEh8glBqHLroFoBgjAoGjpYfQ8Dq6QEWKgBIHQ34AFoZTalMJYawXBBS8GksQM46Z4CLAqFUZwEBXCjGaKQQI0wiglGyMkVIAgVHaNyGkXomi5itHaNUSY%2BjxhtBPhY7oxj%2BilCGN0Kxzj6gONmKUOR14JD%2B0DmXO%2BEcOD7FULEAufCC6SH2MAZAyB9hDxvFwfYEBcCEBIPHLg8xeB1y0IZUgGdMzHQDhwUupAMGFNICHMOQSq41yTinRYTdW7LAIAkcB5BKC907qEVg6xQnhMidE2J8Ts6J0wPgIg0i9D8FIaIcQlCZnUJUOoO%2B9DSB3EQQkTBxcOBB0qeXIJk9wFtNRKgGc/SIlRJiXEhJSSIAeA7v3DJWT6n1zyQgTATAsAxCfMU0p5TjpVIkZXWwdScmp3yZnbO%2BcYW5yLsUzYATqkgvBY0luTSIBIC6f3DpPd259wGLsb%2BXAzAaGOqPPA48p4zxDvPcES9KCrzvrvTekCWX70PsfBwkDz6MAIFfG%2B5csCP2flGN%2BidP7f1/mHf%2BtigHl1AcgcB6xE7QhgXfOBCCkEYHWGHNBGD35YJwXg6lRD56yDIfM6QizFDLLoSAMw%2Bhv7MJEawzVHCnzcN4ZwARrohEsIsGIiRPEpHkQ9XImxCi/BKMiq49RhRHEGN0ekTwTQk0dA8Vo6x5iBBdBGKmsYZjbG5smJm0xEwXEFtURW9xGjE2ZKWIRXxOy9lAorsEi5gyiVGHiWYG8Gh%2B3JNSZM552SGnpyhTnWFMKdn/JABUttNTQW1waY3dFSAWknNxdi6IPS2CcE7VE7twBe39v7bwcZaSpktEWZaih1rZC2toash16zNnbOKa2g5nAjmtPAfsM5ISwmXKJIwk9JKz0aGSQ8glrIthmBeaixYHyvkDF%2BSXXgAL9mBJBdXFdbyJ1ZyndOvOOzEXYeRRwMdBGdlmCRcCqjrzcmLBDSkZwkggA%3D%3D%3D
+
 #if defined(__has_builtin)
-    #if __has_builtin(__builtin_memcpy) && (!defined(AZ_COMPILER_GCC) || AZ_COMPILER_GCC >= 130000)
+    #if __has_builtin(__builtin_memcpy) && (!defined(AZ_COMPILER_GCC))
         #define az_has_builtin_memcpy true
     #endif
-    #if __has_builtin(__builtin_wmemcpy) && (!defined(AZ_COMPILER_GCC) || AZ_COMPILER_GCC >= 130000)
+    #if __has_builtin(__builtin_wmemcpy)
         #define az_has_builtin_wmemcpy true
     #endif
-    #if __has_builtin(__builtin_memmove) && (!defined(AZ_COMPILER_GCC) || AZ_COMPILER_GCC >= 130000)
+    #if __has_builtin(__builtin_memmove) && (!defined(AZ_COMPILER_GCC))
         #define az_has_builtin_memmove true
     #endif
-    #if __has_builtin(__builtin_wmemmove) && (!defined(AZ_COMPILER_GCC) || AZ_COMPILER_GCC >= 130000)
+    #if __has_builtin(__builtin_wmemmove)
         #define az_has_builtin_wmemmove true
     #endif
-    #if (__has_builtin(__builtin_strlen) && (!defined(AZ_COMPILER_GCC) || AZ_COMPILER_GCC >= 130000)) || defined(AZ_COMPILER_MSVC)
+    #if (__has_builtin(__builtin_strlen)) || defined(AZ_COMPILER_MSVC)
         #define az_has_builtin_strlen true
     #endif
-    #if (__has_builtin(__builtin_wcslen) && (!defined(AZ_COMPILER_GCC) || AZ_COMPILER_GCC >= 130000)) || defined(AZ_COMPILER_MSVC)
+    #if (__has_builtin(__builtin_wcslen)) || defined(AZ_COMPILER_MSVC)
         #define az_has_builtin_wcslen true
     #endif
-    #if (__has_builtin(__builtin_char_memchr) && (!defined(AZ_COMPILER_GCC) || AZ_COMPILER_GCC >= 130000)) || defined(AZ_COMPILER_MSVC)
+    #if (__has_builtin(__builtin_char_memchr)) || defined(AZ_COMPILER_MSVC)
         #define az_has_builtin_char_memchr true
     #endif
-    #if (__has_builtin(__builtin_wmemchr) && (!defined(AZ_COMPILER_GCC) || AZ_COMPILER_GCC >= 130000)) || defined(AZ_COMPILER_MSVC)
+    #if (__has_builtin(__builtin_wmemchr)) || defined(AZ_COMPILER_MSVC)
         #define az_has_builtin_wmemchr true
     #endif
-    #if (__has_builtin(__builtin_memcmp) && (!defined(AZ_COMPILER_GCC) || AZ_COMPILER_GCC >= 130000)) || defined(AZ_COMPILER_MSVC)
+    #if (__has_builtin(__builtin_memcmp) && (!defined(AZ_COMPILER_GCC))) || defined(AZ_COMPILER_MSVC)
         #define az_has_builtin_memcmp true
     #endif
-    #if (__has_builtin(__builtin_wmemcmp) && (!defined(AZ_COMPILER_GCC) || AZ_COMPILER_GCC >= 130000)) || defined(AZ_COMPILER_MSVC)
+    #if (__has_builtin(__builtin_wmemcmp)) || defined(AZ_COMPILER_MSVC)
         #define az_has_builtin_wmemcmp true
     #endif
 #elif defined(AZ_COMPILER_MSVC)
--- a/Gems/Archive/Code/Include/Archive/Clients/ArchiveInterfaceStructs.h
+++ b/Gems/Archive/Code/Include/Archive/Clients/ArchiveInterfaceStructs.h
@@ -10,6 +10,7 @@
 
 #include <AzCore/base.h>
 
+#include <AzCore/Math/Crc.h>
 #include <AzCore/IO/Path/Path_fwd.h>
 #include <AzCore/std/containers/array.h>
 #include <AzCore/std/containers/span.h>
@@ -238,7 +239,7 @@ namespace Archive
         " the ArchiveDefaultBlockAlignment");
 
     //! Represents an entry of a single file within the Archive
-    struct alignas(16) ArchiveTocFileMetadata
+    struct alignas(32) ArchiveTocFileMetadata
     {
         ArchiveTocFileMetadata();
 
@@ -280,9 +281,19 @@ namespace Archive
         //! The actual cap for Archive V1 layout is around 64TiB, since the m_blockTable can only
         //! represent 2^25 "2 MiB" blocks, which is (2^25 * 2^21) = 2^46 = 64TiB
         AZ::u64 m_offset : 39; // 64-bits
+
+        //! offset = 16
+        //! Stores a checksum value of the file uncompressed data
+        //! This can be used to validate that uncompressed file contents
+        AZ::Crc32 m_crc32{};
+
+        //! offset = 20
+        //! Add padding bytes to fill the File Metadata structure
+        //! with 0 bytes on construction
+        AZStd::byte m_unused[12]{};
     };
 
-    static_assert(sizeof(ArchiveTocFileMetadata) == 16, "File Metadata size should be 16 bytes");
+    static_assert(sizeof(ArchiveTocFileMetadata) == 32, "File Metadata size should be 16 bytes");
 
     //! Stores the size of a file path and an offset into the file path blob table for a single file
     //! in the archive
--- a/Gems/Archive/Code/Include/Archive/Clients/ArchiveInterfaceStructs.h
+++ b/Gems/Archive/Code/Include/Archive/Clients/ArchiveInterfaceStructs.h
@@ -62,6 +62,12 @@ namespace Archive
     {
         ArchiveHeader();
 
+        //! An actual implementation of the ArchiveHeader copy constructor
+        //! is required as the const AZ::u32 m_magicBytes member
+        //! causes the copy constructor to be implicitly deleted
+        ArchiveHeader(const ArchiveHeader& other);
+        ArchiveHeader& operator=(const ArchiveHeader& other);
+
         //! Retrieves the Uncompressed Table of Contents(TOC) size
         //! by adding up the size of the TOC File Metadata table
         //! + TOC File Path Index table
@@ -167,11 +173,12 @@ namespace Archive
 
         //! Stores 32-bit IDS of up to 7 compression algorithms that this archive can use
         //! Each entry is initialized to the Invalid CompressionAlgorithmId
+        //! The capacity of the array is the value of the uncompressed algorithm index
         //! offset = 48
-        AZStd::array<Compression::CompressionAlgorithmId, 7> m_compressionAlgorithmsIds =
+        AZStd::array<Compression::CompressionAlgorithmId, UncompressedAlgorithmIndex> m_compressionAlgorithmsIds =
             []() constexpr
         {
-            AZStd::array<Compression::CompressionAlgorithmId, 7> compressionIdInitArray{};
+            AZStd::array<Compression::CompressionAlgorithmId, UncompressedAlgorithmIndex> compressionIdInitArray{};
             compressionIdInitArray.fill(Compression::Invalid);
             return compressionIdInitArray;
         }();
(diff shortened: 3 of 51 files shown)


Gems/Archive/Code/Include/Archive/Clients/ArchiveInterfaceStructs.h
```diff
@@ -299,16 +306,27 @@ namespace Archive
     //! There are 3 blocks per block line as 3 "2 MiB" chunks can be encoded in a 64-bit integer
     //! This is done by storing the compressed block size using 21-bits
     constexpr AZ::u64 BlocksPerBlockLine = 3;
+    //! For block line with a jump entry one of instead of having 3 21-bit compressed block sizes
+    //! one of the block entries is borrowed for the jump table entry
```
AMZN-alexpete, line 310:
```suggestion
    //! For a block line with a jump entry of one, instead of having 3 21-bit compressed block sizes
    //! one of the block entries is borrowed for the jump table entry
```

Gems/Archive/Code/Include/Archive/Clients/ArchiveInterfaceStructs.inl
```diff
@@ -283,5 +321,197 @@ namespace Archive
...
+        // h(x) = f(x) + g(x)
+
+        if (const AZ::u64 firstIndexForFinalBlockLineSetOf3 = GetFirstIndexForFinalBlockLineSetOf3(blockCount);
+            blockIndex < firstIndexForFinalBlockLineSetOf3)
+        {
+            // Convert the blocks to skip with a jump entry -> block lines to skip with a jump entry
+            const AZ::u64 skippedBlockLines = blockIndex / BlocksToSkipWithJumpEntry * BlockLinesToSkipWithJumpEntry;
+            result.m_blockLineIndex = skippedBlockLines +
+                ((blockIndex % BlocksToSkipWithJumpEntry) + 1) / BlockLinesToSkipWithJumpEntry;
+            // For block lines sets of 3 that include a jump entries, the blocks per line area layout is
```
AMZN-alexpete, line 423:
```suggestion
            // For block lines sets of 3 that include a jump entry, the blocks per line area layout is
```

Gems/Archive/Code/Include/Archive/Clients/ArchiveInterfaceStructs.inl
```diff
@@ -283,5 +321,197 @@ namespace Archive
...
+            // 2 -> 3 -> 3
+            // block line 0 -> block line 1 -> block line 2
+            // A phase shift is used here to pivot the relativeblock index of to 0 and then modulo arithmetic with
```
AMZN-alexpete, line 426:
```suggestion
            // A phase shift is used here to pivot the relative block index to 0 and then modulo arithmetic with
```

Gems/Archive/Code/Include/Archive/Clients/ArchiveInterfaceStructs.inl
```diff
@@ -283,5 +321,197 @@ namespace Archive
...
+            // the blocks per block line (which is 3) is used to determine the offset within a block
+            // If the block index < 2 then it's value is used directly
```
AMZN-alexpete, line 428:
```suggestion
            // If the block index < 2 then its value is used directly
```

Gems/Archive/Code/Include/Archive/Clients/ArchiveReaderAPI.h
```diff
@@ -0,0 +1,283 @@
...
+    {
+        //! returns true if the ResultOutcome has a non-error value
+        explicit constexpr operator bool() const;
+        //! Stores error info about enumerating all files within the archive
+        ResultOutcome m_resultOutcome;
+    };
+
+
+    //! Interface for the ArchiveReader of O3DE Archive format
+    //! An ArchiveReaderSettings object can be used to customize the how
```
AMZN-alexpete, line 181:
```suggestion
    //! An ArchiveReaderSettings object can be used to customize how
```

Gems/Archive/Code/Include/Archive/Clients/ArchiveReaderAPI.h
```diff
@@ -0,0 +1,283 @@
...
+    //! an archive is read.
+    //! Such as the ability to specify the number of read and decompression task that
+    //! can run in parallel
+
+    //! The user can supply there own stream of archive data via the AZ::IO::GenericStream interface
```
AMZN-alexpete, line 186:
```suggestion
    //! The user can supply their own stream of archive data via the AZ::IO::GenericStream interface
```

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
```diff
@@ -0,0 +1,885 @@
...
+namespace Archive
+{
+    // Implement TypeInfo, Rtti and Allocator support
+    AZ_TYPE_INFO_WITH_NAME_IMPL(ArchiveReader, "ArchiveReader", ArchiveReaderTypeId);
+    AZ_RTTI_NO_TYPE_INFO_IMPL(ArchiveReader, IArchiveReader);
+    AZ_CLASS_ALLOCATOR_IMPL(ArchiveReader, AZ::SystemAllocator);
+
+    // ArchiveReader forwarding functions
+    // This is used to create an ArchiveReader instance that is pointed
+    // to from an IArchiveReader* to allow use of the ArchiveReader in outside of the Archive Gem modules
```
AMZN-alexpete, line 30:
```suggestion
    // to from an IArchiveReader* to allow use of the ArchiveReader in modules outside of the Archive Gem 
```

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
```diff
@@ -0,0 +1,885 @@
...
+            if (AZ::IO::SizeType bytesRead = m_archiveStream->ReadAtOffset(blockCompressedSize,
+                compressedBlockToReadInto.data(), absoluteSeekOffset);
+                bytesRead != blockCompressedSize)
+            {
+                return AZStd::unexpected(ResultString::format("Cannot read all of compressed block for"
+                    " block %llu. The compressed block size is %llu, but only %llu was able to be read",
+                    blockIndex, blockCompressedSize, bytesRead));
+            }
+
+            // As the read was successful add the aligned compressed size to the the fileRelativeSeekOffset
```
AMZN-alexpete, line 576:
```suggestion
            // As the read was successful add the aligned compressed size to the fileRelativeSeekOffset
```

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
```diff
@@ -0,0 +1,885 @@
...
+                        &decompressedBlockResult = decompressedBlockResults[decompressTaskSlot]]()
+                {
+                    // Decompressed the compressed block
+                    decompressedBlockResult = decompressionInterface->DecompressBlock(
+                        decompressionBlockSpan, compressedDataForBlock, decompressionOptions);
+                };
+
+                    taskGraph.AddTask(decompressTaskDescriptor, AZStd::move(decompressTask));
+                    // Increment to the block index
+                    ++blockRange.first;
```
AMZN-alexpete, line 634:
```suggestion
                taskGraph.AddTask(decompressTaskDescriptor, AZStd::move(decompressTask));
                // Increment to the block index
                ++blockRange.first;
```

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
```diff
@@ -0,0 +1,885 @@
...
+            }
+
+            taskGraph.SubmitOnExecutor(m_taskExecutor, taskDecompressGraphEvent.get());
+            // Sync on the task completion
+            taskDecompressGraphEvent->Wait();
```
AMZN-alexpete, line 639: Is the sync here where we wait on all tasks in the task graph to complete preferred to using a work queue to avoid the locks on the queue and be faster overall?  I'm not recommending changing the approach in this PR, just want to understand the tradeoffs.

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
```diff
@@ -0,0 +1,885 @@
...
+                    return AZStd::unexpected(AZStd::move(decompressedBlockResult.m_decompressionOutcome.m_resultString));
+                }
+            }
+        }
+
+        // Return a subspan that accounts for the start offset within the compressed file to start
+        // reading from, up to the bytes read amount
+        // Due to the logic in the function only reading the set of 2 MiB blocks that are needed
+        // the start offset for reading is calculated  a modulo operation
+        // with with the ArchiveBlockSizeForCompression (2 MiB).
```
AMZN-alexpete, line 657:
```suggestion
        // the start offset for reading is calculated by a modulo operation
        // with the ArchiveBlockSizeForCompression (2 MiB).
```

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
```diff
@@ -0,0 +1,885 @@
...
+                    taskGraph.AddTask(decompressTaskDescriptor, AZStd::move(decompressTask));
+                    // Increment to the block index
+                    ++blockRange.first;
+            }
+
+            taskGraph.SubmitOnExecutor(m_taskExecutor, taskDecompressGraphEvent.get());
+            // Sync on the task completion
+            taskDecompressGraphEvent->Wait();
+
+            // Validate the decompression for all blocks
```
AMZN-alexpete, line 641: Is it the responsibility of the decompressor to store a CRC of the decompressed contents alongside the data and verify that the decompressed contents match after decompression?

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
```diff
@@ -0,0 +1,885 @@
...
+                    " but the file path stored in the TOC is empty."
+                    "This indicates that the token is referring to a deleted file",
+                    static_cast<AZ::u64>(archiveFileToken)));
+            return errorResult;
+        }
+
+        // The file has been found and has a non-empty path
+        // Populate the ArchiveListFileResult structure
+        ArchiveListFileResult listResult;
+        // Now the extract the path has stored in the file path blob into the extract result
```
AMZN-alexpete, line 695:
```suggestion
        // Extract the path stored in the file path blob into the extract result
```

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
```diff
@@ -0,0 +1,885 @@
...
+                {
+                    ArchiveListFileResult listResult;
+                    listResult.m_relativeFilePath = contentPathView;
+                    listResult.m_filePathToken = static_cast<ArchiveFileToken>(filePathIndex);
+
+                    // Gather the file metadata
+                    const ArchiveTocFileMetadata& fileMetadata = m_archiveToc.m_tocView.m_fileMetadataTable[filePathIndex];
+
+                    // Use the compression algorithm index to lookup the compression algorithm ID
+                    // if the file the value is less than the size of the compression AlgorithmIds array
```
AMZN-alexpete, line 768:
```suggestion
                    // if the file value is less than the size of the compression AlgorithmIds array
```

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
```diff
@@ -0,0 +1,885 @@
...
+            };
+        EnumerateFilePathIndexOffsets(EnumerateAllFiles, m_archiveToc.m_tocView);
+
+        // There are currently no error messages that enumerate file path sets
+        // So a default constructed instance which converts to boolean true is returned
+        return {};
+    }
+
+    bool ArchiveReader::WriteArchiveMetadata(AZ::IO::GenericStream& metadataStream,
+        const ArchiveMetadataSettings& metadataSettings) const
```
AMZN-alexpete, line 791: It seems odd that the reader is writing data 😄

Gems/Archive/Code/Source/Clients/ArchiveReader.h
```diff
@@ -0,0 +1,217 @@
...
+        //! binding a lambda that populates a vector
+        //!
+        //! @param listFileCallback Callback which is invoked for each file in the archive
+        //! @return result structure that is convertible to a boolean value indicating if enumeration was successful
+        EnumerateArchiveResult EnumerateFilesInArchive(ListFileCallback listFileCallback) const override;
+
+        //! Writes metadata for the archive to the supplied generic stream
+        //! @param metadataStream with human readable data about files within the archive will be written to the stream
+        //! @return true if metadata was successfully written
+        bool WriteArchiveMetadata(AZ::IO::GenericStream& metadataStream,
```
AMZN-alexpete, line 107: I know technically it's writing into the stream, but consider `ReadArchiveMetaData` because it is also reading it out of the archive and might be easier to find by developers searching for this functionality if this isn't internal

Gems/Archive/Code/Source/Tools/ArchiveWriter.cpp
```diff
@@ -424,7 +431,12 @@ namespace Archive
             return false;
         }
 
-        return ReadArchiveHeaderAndToc();
+        if (!ReadArchiveHeaderAndToc())
+        {
+            UnmountArchive();
```
AMZN-alexpete, line 436: might be worth a comment saying why we're calling Unmount, which it looks like is to make sure to clean up/reset any stream settings that may have been changed when reading the header and Toc

Gems/Archive/Code/Tests/Tools/ArchiveReaderTest.cpp
```diff
@@ -0,0 +1,681 @@
...
+#include <Archive/Tools/ArchiveWriterAPI.h>
+
+#include <Compression/CompressionLZ4API.h>
+
+namespace Archive::Test
+{
+    // Note the ArchiveReader unit test are placed in the Archive.Editor.Tests
+    // Tools module to have to the ArchiveWriter which is used to create
+    // test archives to be read
+    // In theory if the all archives were written to a file on disk
```
AMZN-alexpete, line 24:
```suggestion
    // In theory if all archives were written to a file on disk
```

Gems/Archive/Code/Tests/Tools/ArchiveReaderTest.cpp
```diff
@@ -0,0 +1,681 @@
...
+
+            // Write a compressed file this time
+            fileData = "My Prefab Data in an Archive";
+            levelPrefabFileData = fileData;
+            fileSettings.m_compressionAlgorithm = CompressionLZ4::GetLZ4CompressionAlgorithmId();
+            fileSettings.m_relativeFilePath = "subdirectory/Level.prefab";
+            EXPECT_TRUE(archiveWriter->AddFileToArchive(AZStd::as_bytes(AZStd::span(fileData)), fileSettings));
+
+            // Write a second text file
+            fileData = "Box Box, Box Box";
```
AMZN-alexpete, line 345: image

Gems/Archive/Code/Tests/Tools/ArchiveReaderTest.cpp
```diff
@@ -0,0 +1,681 @@
...
+            mountErrorOccurred = true;
+        };
+        AZStd::unique_ptr<IArchiveReader> archiveReader = CreateArchiveReader(AZStd::move(archiveStreamPtr),
+            readerSettings);
+
+        EXPECT_TRUE(mountErrorOccurred);
+        EXPECT_FALSE(archiveReader->IsMounted());
+
+            // reset the mountErrorOccurred boolean
+            mountErrorOccurred = false;
```
AMZN-alexpete, line 107:
```suggestion
        // reset the mountErrorOccurred boolean
        mountErrorOccurred = false;
```

Gems/Archive/Code/Include/Archive/Clients/ArchiveReaderAPI.h
AMZN-alexpete: Or you are sneakily putting them in to see if anyone is reading them! 😆 But seriously, I really appreciate these comments, very useful!

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
lumberyard-employee-dm: Well in this case the number of decompression task is limited by the lower between the number of blocks remaining to decompress into a 2-MIB chunk of the file and the core count of the system.

Task slots are preallocated up front and only run on a block of a compressed file that are segmented to not overlap.

The sync is preferable here so that that the next chunk of blocks being decompressed don't have to search for an available task slot.
Otherwise a synchronization primitive would have to be used to keep track of the active set of decompression jobs that are being run.

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
lumberyard-employee-dm: Well this is for writing out Archive file informational data for diagnostics.
So if the user wanted to just write out all the file names within the archive in a machine parseable format

Gems/Archive/Code/Source/Clients/ArchiveReader.h
lumberyard-employee-dm: How about DumpArchiveMetadata instead?

Gems/Archive/Code/Source/Tools/ArchiveWriter.cpp
lumberyard-employee-dm: Will do.

Gems/Archive/Code/Tests/Tools/ArchiveReaderTest.cpp
lumberyard-employee-dm: s🅱️innala

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
lumberyard-employee-dm: I added support for writing a CRC32 for each file added to the Archive in this commit https://github.com/o3de/o3de/pull/16129/commits/efaf58563abac11f3eddd9548b6620a8b9593412

Gems/Archive/Code/Source/Tools/ArchiveWriter.cpp
```diff
@@ -1517,8 +1561,13 @@ namespace Archive
                         // Only output compressed size if an compression that compresses data is being used
                         if (contentFileMetadata.m_compressionAlgoIndex < UncompressedAlgorithmIndex)
                         {
-                            fileMetadataString += MetadataString::format(R"(, compressed_size=%llu)",
-                                contentFileMetadata.m_compressedSizeInSectors * ArchiveDefaultBlockAlignment);
+                            if (auto rawFileSizeResult = GetRawFileSize(contentFileMetadata, m_archiveToc.m_blockOffsetTable);
+                                rawFileSizeResult)
+                            {
+                                AZ::u64 compressedSize = rawFileSizeResult.value();
+                                fileMetadataString += MetadataString::format(R"(, compressed_size=%llu)",
```
AMZN-alexpete, line 1568:
```suggestion
                                fileMetadataString += MetadataString::format(R"(compressed_size=%llu)",
```

AMZN-alexpete (approved):
Nice!

Code/Framework/AzCore/AzCore/Math/Crc.h
```diff
@@ -80,13 +80,16 @@ namespace AZ
          * Calculates the value from a block of raw data.
          */
         Crc32(const void* data, size_t size, bool forceLowerCase = false);
-        constexpr Crc32(const uint8_t* data, size_t size, bool forceLowerCase = false);
-        constexpr Crc32(const char* data, size_t size, bool forceLowerCase = false);
+        template<class ByteType, class = AZStd::enable_if_t<sizeof(ByteType) == 1>>
+        constexpr Crc32(const ByteType* data, size_t size, bool forceLowerCase = false);
+        constexpr Crc32(AZStd::span<const AZStd::byte> inputSpan);
```
spham-amzn, line 85: Is it possible to templatize this for span as well? 
```suggestion
        template<class ByteType, class = AZStd::enable_if_t<sizeof(ByteType) == 1>>
        constexpr Crc32(AZStd::span<const ByteType> inputSpan);
```

lumberyard-employee-dm: Yes the span could be templatized as well, however the code is using a convention that passing text is `AZStd::string_view` and passing in binary is `AZStd::span<const AZStd::byte>`

Since the Crc32 class has both, that covers the text and binary cases

Code/Framework/AzCore/AzCore/PlatformDef.h
```diff
@@ -281,35 +281,45 @@
...
+
 #if defined(__has_builtin)
-    #if __has_builtin(__builtin_memcpy) && (!defined(AZ_COMPILER_GCC) || AZ_COMPILER_GCC >= 130000)
+    #if __has_builtin(__builtin_memcpy) && (!defined(AZ_COMPILER_GCC))
```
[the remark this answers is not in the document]
lumberyard-employee-dm: This is the opposite.
The GCC build was failing because it was trying to use `__builtin_memcpy` in a constant evaluation context and it was failing.

So the change is to have GCC13+ use the behavior of GCC12 and below.

Gems/Archive/Code/Include/Archive/Clients/ArchiveInterfaceStructs.inl
```diff
@@ -283,5 +321,197 @@ namespace Archive
...
+        return m_errorMessage.empty();
+    }
+
+    constexpr GetBlockLineIndexResult GetBlockLineIndexFromBlockIndex(AZ::u64 blockCount, AZ::u64 blockIndex)
+    {
+        GetBlockLineIndexResult result;
+        if (blockIndex >= blockCount)
+        {
+            result.m_errorMessage = "Block index is out of range of the number of block count."
+                " A block line index cannot be returned";
```
spham-amzn, line 381: nit: missing period
```suggestion
                " A block line index cannot be returned.";
```

Gems/Archive/Code/Include/Archive/Clients/ArchiveInterfaceStructs.inl
```diff
@@ -283,5 +321,197 @@ namespace Archive
...
+            return result;
+        }
+
+        auto GetFirstIndexForFinalBlockLineSetOf3 = [](AZ::u64 blockCount) constexpr
+        {
+            auto GetFinalIndexForFinalBlockLineSetOf3 = [](AZ::u64 blockCount) constexpr
+            {
+                return blockCount <= MaxBlocksNoJumpEntry
+                    ? MaxBlocksNoJumpEntry
```
spham-amzn, line 390: The `blockCount <= MaxBlocksNoJumpEntry` is checked before using this function in the lines below, so is this check necessary here?

lumberyard-employee-dm: No it is not.
I'll remove in this PR if there is a need for a non-trivial update.

spham-amzn: Also make sure which logic is right. The one below returns zero, this one returns `MaxBlocksNoJumpEntry` if `blockCount <= MaxBlocksNoJumpEntry`

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
```diff
@@ -0,0 +1,928 @@
...
+
+        return true;
+    }
+
+    // Define the Archive TableOfContentReader constructors
+    ArchiveReader::ArchiveTableOfContentsReader::ArchiveTableOfContentsReader() = default;
+
+    // Takes ownership of the buffer containing the Table of Contents raw data
+    // and an ArchiveTableOfContentsView instance
+    ArchiveReader::ArchiveTableOfContentsReader::ArchiveTableOfContentsReader(AZStd::vector<AZStd::byte> tocBuffer,
```
spham-amzn, line 73: Shouldn't this use move operators?

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
```diff
@@ -0,0 +1,928 @@
...
+        ArchiveTableOfContentsView tocView)
```
spham-amzn, line 74: Doesn't this need `&&` operators?
```suggestion
    ArchiveReader::ArchiveTableOfContentsReader::ArchiveTableOfContentsReader(AZStd::vector<AZStd::byte> && tocBuffer,
        ArchiveTableOfContentsView && tocView)
```

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
```diff
@@ -0,0 +1,928 @@
...
+
+    auto ArchiveReader::ReadRawFileIntoBuffer(AZStd::span<AZStd::byte> fileBuffer, AZ::u64 offset,
+        AZ::u64 fileSize,
+        const ArchiveReaderFileSettings& fileSettings)
+        -> ReadRawFileOutcome
+    {
+        // Calculate the start offset where to read the file content from
+        // It must be within the the range of [offset, offset + size)
+        AZ::IO::OffsetType readOffset = AZStd::clamp(offset, offset + fileSettings.m_startOffset, offset + fileSize);
+        // Next clamp the bytesToRead to be not read pass the end of the file
```
spham-amzn, line 397:
```suggestion
        // Next clamp the bytesToRead to not read pass the end of the file
```

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
```diff
@@ -0,0 +1,928 @@
...
+                " archive file offset %lld."
+                " Buffer size is %zu, while %llu is required.", readOffset, fileBuffer.size(), bytesToRead));
+        }
+
+        AZStd::scoped_lock archiveReadLock(m_archiveStreamMutex);
+        if (AZ::IO::SizeType bytesRead = m_archiveStream->ReadAtOffset(bytesToRead, fileBuffer.data(), readOffset);
+            bytesRead < bytesToRead)
+        {
+            return AZStd::unexpected(ResultString::format("Attempted to read %llu bytes from the archive at offset %lld."
+                " But only %llu bytes were able to be read", bytesToRead, readOffset, bytesRead));
```
spham-amzn, line 414: nit: missing period
```suggestion
                " But only %llu bytes were able to be read.", bytesToRead, readOffset, bytesRead));
```

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
```diff
@@ -0,0 +1,928 @@
...
+        // The cap is uncompressed size of the file
+        if (fileSettings.m_startOffset > extractFileResult.m_uncompressedSize)
+        {
+            return AZStd::unexpected(ResultString::format("Start offset %llu to read file data from is larger."
+                " than the size of the file %llu for file %s", fileSettings.m_startOffset,
+                extractFileResult.m_compressedSize,
+                extractFileResult.m_relativeFilePath.c_str()));
+        }
+
+        // Clamp the the bytes that can read for the file to be at
```
spham-amzn, line 469:
```suggestion
        // Clamp the bytes that can read for the file to be at
```

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
```diff
@@ -0,0 +1,928 @@
...
+            static_cast<AZ::u64>(archiveFileToken)];
+
+        if (filePathOffsetSize.m_size == 0)
+        {
+            ArchiveListFileResult errorResult;
+            errorResult.m_filePathToken = archiveFileToken;
+            errorResult.m_resultOutcome = AZStd::unexpected(
+                ResultString::format(R"(A file token "%llu" is being used to extract the file,)"
+                    " but the file path stored in the TOC is empty."
+                    "This indicates that the token is referring to a deleted file",
```
spham-amzn, line 702:
```suggestion
                    "This indicates that the token is referring to a deleted file.",
```

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
```diff
@@ -0,0 +1,928 @@
...
+    }
+
+    ArchiveListFileResult ArchiveReader::ListFileInArchive(AZ::IO::PathView relativePath) const
+    {
+        if (relativePath.empty())
+        {
+            ArchiveListFileResult errorResult;
+            errorResult.m_resultOutcome = AZStd::unexpected(
+                ResultString(R"(An empty file path has been supplied."
+                " An empty path cannot be found in the archive.)"));
```
spham-amzn, line 754: will the `R"` string literal span across two string literals that are broken across lines? Probably not needed at all here

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
lumberyard-employee-dm: No, the move happens internally using the value type.
If the caller wants to move the tocBuffer into the ArchiveTableOfContentsReader, the choice is left up to them with this API.
In their calling code `ArchiveTableOfContentsReader(AZStd::move(tocBuffer))`, then they would move the buffer contents
If they instead invoke `ArchiveTbleOfContentsReader(tocBuffer)`, they would make the copy of contents

If the signature was `AZStd::vector<AZStd::byte>&&` then the user would be forced to transfer ownership over which isn't explicitly required.

The comment above the function that says "Takes ownership of the buffer" is wrong though.

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
```diff
@@ -0,0 +1,928 @@
...
+        }
+
+        if (metadataSettings.m_writeFilePaths)
+        {
+            // Validate the file path and file metadata tables are in sync
+            if (m_archiveToc.m_tocView.m_filePathIndexTable.size() != m_archiveToc.m_tocView.m_fileMetadataTable.size())
+            {
+                auto errorString = MetadataString::format("Error: The Archive TOC of contents has a mismatched size between"
+                    " the file path index vector (size=%zu) and the file metadata vector (size=%zu).\n"
+                    "This indicates a code error in the ArchiveReader in keeping both vectors in sync and should never occur",
```
spham-amzn, line 845:
```suggestion
                    "This indicates a code error in the ArchiveReader in keeping both vectors in sync and should never occur.",
```

spham-amzn: Also not sure if the "should never occur" should be here, its not something that is actionable outside of the ArchiveReader

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
spham-amzn: yeah, I was basing this on the description

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
```diff
@@ -0,0 +1,928 @@
...
+        // If the compression algorithm id is invalid return the invalid algorithm index
+        if (compressionAlgorithmId == Compression::Invalid)
+        {
+            return InvalidAlgorithmIndex;
+        }
+
+        // If the compression algorithm id is uncompressed
+        // return a value of 7 which is the highest 3-bit value(0b111)
+        // The compression algorithm Id array contains 7 elements for registering compression algorithms.
+        // The index of 7 is used to represent the file is uncompressed the purposes of storing
```
spham-amzn, line 913:
```suggestion
        // The index of 7 is used to represent the file is uncompressed for the purpose of storing
```

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
lumberyard-employee-dm: It will expand across multiple lines, but the raw string literal isn't needed here, so I'll remove it.

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
lumberyard-employee-dm: Yeah, I fix up the Error message to remove the should never occur comment.

Gems/Archive/Code/Include/Archive/Clients/ArchiveInterfaceStructs.inl
lumberyard-employee-dm: Well the inner lambda condition is always returning the "else" part of the Ternary since the condition is checked before calling it.

The entire purpose of the two function is to return the minimum block index for the final set of block lines that don't have a jump entry (confusing and hard to explain, I know).

But pretty much every compressed file in the Archive that is >16 MiB starts with its first block line (index 0) with a jump entry.
And then every three block lines afterwards(index 3, 6, 9, ...) would have a Jump entry if there more than 18 MiB (9 blocks) remaining in the file from the start of that block line.

It is explained better in the Streaming Archive document at [here](https://github.com/o3de/sig-core/blob/decc359e5fbc3a718e193bb92e6d80e55e19fb4f/rfcs/rfc-core-2023-04-20-streaming-archive.md#block-offset-table-section), but the implementation is complicated.

Gems/Archive/Code/Source/Clients/ArchiveReader.h
```diff
@@ -0,0 +1,217 @@
...
+            ArchiveTableOfContentsView m_tocView;
+        private:
+            AZStd::vector<AZStd::byte> m_tocBuffer;
+        };
+        ArchiveTableOfContentsReader m_archiveToc;
+
+        //! Stores mapping of FilePath to index within the file path table in the Archive TOC
+        //! The index is used to as the ArchiveFileToken
+        //! IMPORTANT: The PathView is a view into the m_archiveToc TOC buffer
+        //! and therefore this map should be clear before re-reading another archive TOC
```
lumberyard-employee-dm, line 202: clear -> cleared
```suggestion
        //! and therefore this map should be cleared before re-reading another archive TOC
```

Gems/Archive/Code/Source/Clients/ArchiveReader.cpp
lumberyard-employee-dm: I updated to the class constructor comment to remove the bit about taking ownership of the buffer.

Gems/Archive/Code/Source/Clients/ArchiveReader.h
```diff
@@ -0,0 +1,217 @@
...
+        //! View of the Archive TOC within the supplied archive stream
+        //! Since the ArchiveReader doesn't mutate the archive, a Table of Contents View is used
+        //! and paired with a raw buffer of the Table of Contents
+        struct ArchiveTableOfContentsReader
+        {
+            // Default a table of contents reader that has an empty vector
+            // and default constructed view
+            ArchiveTableOfContentsReader();
+            // Takes ownership of the buffer containing the Table of Contents raw data
+            // and an ArchiveTableOfContentsView instance
```
spham-amzn, line 190: 'Takes ownership' is mentioned here as well

Gems/Archive/Code/Source/Clients/ArchiveTOCView.h
```diff
@@ -105,5 +105,48 @@ namespace Archive
...
 
+// Separate namespace block to create visual spacing around utility functions
+// for querying the the Table of Contents View
+namespace Archive
+{
+   //! Retrieves a subspan from the Table of Contents Archive BlockLine Offset Table
+   //! that contains the block lines associated with a specific file
+   //! @param fileMetadataTableIndex index into the file metadata table
+   //!        This is used to lookup the uncompressed size and the first block line offset entry
+   //!        for a file
```
spham-amzn, line 116: @params here don't match the declaration

Gems/Archive/Code/Source/Clients/ArchiveReader.h
lumberyard-employee-dm: I will update this comment as well to remove the mention of taking over ownership of the buffer

Gems/Archive/Code/Source/Clients/ArchiveTOCView.h
lumberyard-employee-dm: The missing tocView parameter has been added.

Gems/Archive/Code/Source/Clients/ArchiveReader.h
```diff
@@ -0,0 +1,217 @@
...
+        //! Queries the archive header for the index matching the specified compression
+        //! algorithm id
+        //! @param compressionAlgorithmId Id to query for the index in the archive header
+        //! for the compression algorithm Id array
+        //! @return the index of the registered compression algorithm Id
+        //! Special Case: If the input ID is `Compression::Uncompressed`, then
+        //! the value of `Compression::UncompressedAlgorithmIndex` is return instead
+        //! If the algorithm Id is not registered then the value of `InvalidAlgorithmIndex`
+        //! is returned
+        size_t FindCompressionAlgorithmId(Compression::CompressionAlgorithmId);
```
spham-amzn, line 120: The @param compressionAlgorithmId was not declared (just the type) in the header. I know its not needed for the header, but that argument is referenced in the doc for this function

lumberyard-employee-dm: Actually that function shouldn't be declared anymore.
It was moved to a free function in ArchiverBaseAPI.h.
I'll remove the declaration.

spham-amzn: yeah, i wish there was a way to tell the linker to detect undefined function references to catch these, otherwise people who try to use undefined functions will find out the hard way.

spham-amzn (approved):
Mostly notes on comments/docs. nice work

PR #16130, opened 2023-06-08
Add edit-repo-properties CLI function and unit test

VeryKaylin-player1:
## What does this PR do?

Added a edit-repo-CLI function that edits the repo.json fields -> 
gems_data: add gems / remove gems / replace gems
template_data: add template / remove template / replace template
project_data: add project / remove project / replace project

the command to add a gem is: './o3de.bat edit-repo-properties -rp "C:\o3de_Projects\test1\repo.json" -ag "C:\o3de_Projects\o3de\Gems\Archive"'

## How was this PR tested?
I wrote unit test test_repo_properties.py file and tested through the command line argument to add, delete, and replace different objects.

### DIFF
--- a/scripts/o3de/o3de/repo_properties.py
+++ b/scripts/o3de/o3de/repo_properties.py
@@ -94,7 +94,7 @@ def create_remote_object_archive(src_data_path: pathlib.Path,
     zip_path = releases_path / archive_filename
     # check if a object.zip folder already exist in the path - ask user if they want to overwrite the current zip
     if not force and zip_path.exists():
-        logger.error(f'{zip_path} already exists.  Use the --force to overwrite the existing zip or '
+        logger.error(f'{zip_path} already exists.  Use --force command to overwrite the existing zip or '
                      'provide a new location to save your archive.')
         return 1
     else:
--- a/scripts/o3de/o3de/repo_properties.py
+++ b/scripts/o3de/o3de/repo_properties.py
@@ -48,9 +48,13 @@ def merge_json_data(json_path: pathlib.Path, json_data: dict) -> dict:
             existing_json_data = json.load(f)
 
         merged_json_data = existing_json_data | json_data
+        
+        if merged_json_data == existing_json_data and 'version' in existing_json_data:
+            # no changes were made, avoid making unnecessary change to the file modification time
+            return existing_json_data
 
         if not 'version' in merged_json_data:
-            merged_json_data['version'] = '1.0.0'
+            merged_json_data['version'] = '0.0.0'
 
         with open(json_path, 'w') as f:
             f.write(json.dumps(merged_json_data, indent=4) + '\n')
@@ -62,7 +66,8 @@ def create_remote_object_archive(src_data_path: pathlib.Path,
                    archive_filename: pathlib.PurePath, 
                    releases_path: pathlib.Path, 
                    repo_uri:str,
-                   download_prefix:str) -> dict:
+                   force: bool,
+                   download_prefix:str) -> dict or int:
     """
     Creates a release of a specific version of a 
     remote object for the given src_data_path.
@@ -86,23 +91,27 @@ def create_remote_object_archive(src_data_path: pathlib.Path,
         ValueError: If the json_data_path is not a dict.
     """
 
-    json_data = merge_json_data(json_data_path, {
-        'repo_uri':repo_uri,
-        'download_source_uri':f'{download_prefix}/{archive_filename}',
-    })
-
-    logging.info(f"Creating '{releases_path / archive_filename}'")
-
-    # create the release zip file
-    shutil.make_archive(releases_path / archive_filename.stem, 'zip', src_data_path)
-
     zip_path = releases_path / archive_filename
+    # check if a object.zip folder already exist in the path - ask user if they want to overwrite the current zip
+    if not force and zip_path.exists():
+        logger.error(f'{zip_path} already exists.  Use the --force to overwrite the existing zip or '
+                     'provide a new location to save your archive.')
+        return 1
+    else:
+        # create the release zip file
+        json_data = merge_json_data(json_data_path, {
+            'repo_uri':repo_uri,
+            'download_source_uri':f'{download_prefix}/{archive_filename}',
+        })
 
-    with zip_path.open('rb') as f:
-        json_data['sha256'] = hashlib.sha256(f.read()).hexdigest()
+        logging.info(f"Creating '{releases_path / archive_filename}'")
+
+        shutil.make_archive(releases_path / archive_filename.stem, 'zip', src_data_path)
+        with zip_path.open('rb') as f:
+            json_data['sha256'] = hashlib.sha256(f.read()).hexdigest()
 
     return json_data
-
+    
 # retrieves json data of the repository from a file path
 def get_repo_props(path: pathlib.Path) -> dict or None:
     repo_json = manifest.get_json_data_file(path, "Repo", validation.valid_o3de_repo_json)
@@ -134,9 +143,10 @@ def _edit_objects(object_typename:str,
                   delete_objects: str or list = None,
                   replace_objects: pathlib.Path or list = None,
                   release_archive_path: pathlib.Path = None,
+                  force: bool = None,
                   download_prefix: str = None):
     """
-    Edits and modifies the 'gem_data/projects_data/template_data' in the repo_json parameter
+    Modifies the 'gems_data/projects_data/templates_data' in repo_json
     :param object_typename: The type object field you want to change
     :param validator: validates the object you want to edit
     :param add_objects: Any object paths you want to add to your repo_json object field
@@ -158,6 +168,8 @@ def _edit_objects(object_typename:str,
             # object JSON data of the object you want to add
             json_data = manifest.get_json_data(object_typename, object_path, validator)
             if json_data:
+                # for a project called TestProject that is version 1.0.0, 
+                # --release_archive_path would create a filename of `testproject-1.0.0-project.zip`
                 if release_archive_path:
                     version = json_data.get('version','0.0.0')
                     archive_filename = pathlib.PurePath(f"{json_data[f'{object_typename}_name']}-{version}-{object_typename}.zip".lower())
@@ -166,7 +178,9 @@ def _edit_objects(object_typename:str,
                                 archive_filename, 
                                 release_archive_path, 
                                 repo_json['repo_uri'], 
+                                force,
                                 download_prefix)
+                    if type(json_data) == int: return 1
                 index = _find_index(repo_objects_data, f'{object_typename}_name', json_data[f'{object_typename}_name'])
                 # index will be non-negative if an instance of this object already exists in the repo.json
                 if index != -1:
@@ -180,8 +194,10 @@ def _edit_objects(object_typename:str,
 
                     # versions data must always include the version
                     version_data['version'] = json_data.get('version','0.0.0')
-
+                
                     # versions data must always include one of the download/source uris
+                    if not 'download_source_uri' in json_data and not 'source_control_uri' in json_data :
+                        logger.warning('No download_source_uri found and sorce_control_uri found!')
                     if 'download_source_uri' in json_data:
                         version_data['download_source_uri'] = json_data['download_source_uri']
                     if 'source_control_uri' in json_data:
@@ -253,6 +269,7 @@ def edit_repo_props(repo_path: pathlib.Path = None,
                        delete_templates: str or list = None,
                        replace_templates: pathlib.Path or list = None,
                        release_archive_path: pathlib.Path = None,
+                       force: bool = None,
                        download_prefix: str = None
                        ) -> int:
     """
@@ -284,13 +301,13 @@ def edit_repo_props(repo_path: pathlib.Path = None,
         repo_json['repo_name'] = repo_name
 
     if add_gems or delete_gems or replace_gems:
-        _edit_objects('gem', validation.valid_o3de_gem_json, repo_json, add_gems, delete_gems, replace_gems, release_archive_path, download_prefix)
+        _edit_objects('gem', validation.valid_o3de_gem_json, repo_json, add_gems, delete_gems, replace_gems, release_archive_path, force, download_prefix)
 
     if add_projects or delete_projects or replace_projects:
-        _edit_objects('project', validation.valid_o3de_project_json, repo_json, add_projects, delete_projects, replace_projects, release_archive_path, download_prefix)
+        _edit_objects('project', validation.valid_o3de_project_json, repo_json, add_projects, delete_projects, replace_projects, release_archive_path, force, download_prefix)
 
     if add_templates or delete_templates or replace_templates:
-        _edit_objects('template', validation.valid_o3de_template_json, repo_json, add_templates, delete_templates, replace_templates, release_archive_path, download_prefix)
+        _edit_objects('template', validation.valid_o3de_template_json, repo_json, add_templates, delete_templates, replace_templates, release_archive_path, force, download_prefix)
 
     return 0 if manifest.save_o3de_manifest(repo_json, repo_path) else 1
 
@@ -312,6 +329,7 @@ def _edit_repo_props(args: argparse) -> int:
                               replace_templates=args.replace_templates,
 
                               release_archive_path=args.release_archive_path,
+                              force=args.force,
                               download_prefix=args.download_prefix
                               )
 
@@ -354,6 +372,8 @@ def add_parser_args(parser):
                                                   description='path arguments to use with the --add-gems or --replace-gems option')
     modify_gems_group.add_argument('--release-archive-path','-rap',  type=pathlib.Path, required=False,
                             help='Create a release archive at the specified local path and update the download_source_uri and sha256 fields.')
+    modify_gems_group.add_argument('--force', '-f', action='store_true', default=False,
+                                                help='Copies over instantiated template directory even if it exist.')
     modify_gems_group.add_argument('--download-prefix','-dp',  type=str, required=False,
                             help='a URL prefix for a file attached to a GitHub release might look like this:'
                             '-dp https://github.com/o3de/o3de-extras/releases/download/2305.0/')
--- a/scripts/o3de/o3de/repo_properties.py
+++ b/scripts/o3de/o3de/repo_properties.py
@@ -15,7 +15,8 @@ import json
 import logging
 import hashlib
 import shutil
-
+from packaging.version import Version, InvalidVersion
+from packaging.specifiers import SpecifierSet
 from o3de import manifest, utils, validation
 
 logger = logging.getLogger('o3de.repo_properties')
@@ -56,17 +57,18 @@ def merge_json_data(json_path: pathlib.Path, json_data: dict) -> dict:
 
     return merged_json_data
 
-def create_release(src_data_path: pathlib.Path, 
+def create_remote_object_archive(src_data_path: pathlib.Path, 
                    json_data_path: pathlib.Path, 
                    archive_filename: pathlib.PurePath, 
                    releases_path: pathlib.Path, 
                    repo_uri:str,
                    download_prefix:str) -> dict:
     """
-    Creates a release for the given src_data_path.
+    Creates a release of a specific version of a 
+    remote object for the given src_data_path.
     The release is saved to the releases_path.
-    Returns the json data.
-    
+    Returns the json data. 
+ 
     Args:
         src_data_path (pathlib.Path): The path to the src root folder.
         json_data_path (pathlib.Path): The path to the json data.
@@ -83,22 +85,18 @@ def create_release(src_data_path: pathlib.Path,
         FileNotFoundError: If the json_data_path does not exist.
         ValueError: If the json_data_path is not a dict.
     """
-    # basename = f'{object_name}_release_1.0.0'
-    # zip_filename = f'{archive_filename}.zip'
-    basename = archive_filename.stem
-    zip_filename = archive_filename
 
     json_data = merge_json_data(json_data_path, {
         'repo_uri':repo_uri,
-        'download_source_uri':f'{download_prefix}/{zip_filename}',
+        'download_source_uri':f'{download_prefix}/{archive_filename}',
     })
 
-    logging.info(f"Creating '{releases_path / zip_filename}'")
+    logging.info(f"Creating '{releases_path / archive_filename}'")
 
     # create the release zip file
-    shutil.make_archive(releases_path / basename, 'zip', src_data_path)
+    shutil.make_archive(releases_path / archive_filename.stem, 'zip', src_data_path)
 
-    zip_path = releases_path / zip_filename
+    zip_path = releases_path / archive_filename
 
     with zip_path.open('rb') as f:
         json_data['sha256'] = hashlib.sha256(f.read()).hexdigest()
@@ -106,13 +104,12 @@ def create_release(src_data_path: pathlib.Path,
     return json_data
 
 # retrieves json data of the repository from a file path
-def get_repo_props(path: pathlib.Path) -> dict:
+def get_repo_props(path: pathlib.Path) -> dict or None:
     repo_json = manifest.get_json_data_file(path, "Repo", validation.valid_o3de_repo_json)
     if not isinstance(repo_json, dict):
-        logger.error(f'Could not retrieve repo.json file for {path}')
+        logger.error(f'Could not retrieve repo.json file from {path} or the repo.json file is invalid.')
         return None
     return repo_json
-
 
 def _find_index(data:list, key:str, value:str) -> int:
     for index, item in enumerate(data):
@@ -145,39 +142,44 @@ def _edit_objects(object_typename:str,
     :param add_objects: Any object paths you want to add to your repo_json object field
     :param delete_objects: Any object_names to be removed from your repo_json object field
     :param replace_objects: A list of object paths that will completely replace the current object_list
-    :param release_archive_path: Path where you want your release to be located
+    :param release_archive_path: Optional local path to a folder where a release archive should be written
     :param download_prefix: The prefix of the download uri
     """
     # The beginning remote repo json template
     repo_objects_data = repo_json.get(f'{object_typename}s_data',[])
-
-    if add_objects:
-        paths = add_objects.split() if isinstance(add_objects, str) else add_objects
+   
+    if add_objects or replace_objects:
+        if replace_objects:
+            repo_objects_data = []
+            paths = replace_objects.split() if isinstance(replace_objects, str) else replace_objects
+        else:
+            paths = add_objects.split() if isinstance(add_objects, str) else add_objects
         for object_path in paths:
-            # gem data of the gem you want to add
+            # object JSON data of the object you want to add
             json_data = manifest.get_json_data(object_typename, object_path, validator)
             if json_data:
                 if release_archive_path:
                     version = json_data.get('version','0.0.0')
                     archive_filename = pathlib.PurePath(f"{json_data[f'{object_typename}_name']}-{version}-{object_typename}.zip".lower())
-                    json_data = create_release(object_path, 
-                                   object_path / f'{object_typename}.json', 
-                                   archive_filename, 
-                                   release_archive_path, 
-                                   repo_json['repo_uri'], 
-                                   download_prefix)
+                    json_data = create_remote_object_archive(object_path, 
+                                object_path / f'{object_typename}.json', 
+                                archive_filename, 
+                                release_archive_path, 
+                                repo_json['repo_uri'], 
+                                download_prefix)
                 index = _find_index(repo_objects_data, f'{object_typename}_name', json_data[f'{object_typename}_name'])
-                # if the added gem is same with partial changes
+                # index will be non-negative if an instance of this object already exists in the repo.json
                 if index != -1:
                     # we want changes only
                     version_data = _changed(repo_objects_data[index], json_data)
                     if not version_data:
-                        # there is no difference between the gem being added and 
-                        # what already exists in gems_data
+                        # there is no difference between the object being added and 
+                        # what already exists in repo.json
+                        logger.warning(f"No changes were found between the object JSON data in repo.json and the object data being added from {object_path}")
                         continue
 
                     # versions data must always include the version
-                    version_data['version'] = json_data['version']
+                    version_data['version'] = json_data.get('version','0.0.0')
 
                     # versions data must always include one of the download/source uris
                     if 'download_source_uri' in json_data:
@@ -192,7 +194,10 @@ def _edit_objects(object_typename:str,
                         # update the existing version
                         if version_data:
                             repo_objects_data[index]['versions_data'][version_index] = version_data
-                    else:
+                    # if version is same, update field                        
+                    elif repo_objects_data[index].get('version','0.0.0') == version_data['version']:
+                        repo_objects_data[index].update(version_data)
+                    else:    
                         # add the new version
                         if 'versions_data' not in repo_objects_data[index]:
                             repo_objects_data[index]['versions_data'] = [version_data] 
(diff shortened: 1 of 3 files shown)


scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,401 @@
...
+    Returns:
+        dict: The json data.
+    
+    Raises:
+        FileNotFoundError: If the src_data_path does not exist.
+        FileNotFoundError: If the json_data_path does not exist.
+        ValueError: If the json_data_path is not a dict.
+    """
+    #basename = f'{object_name}_release_1.0.0'
+    #zip_filename = f'{archive_filename}.zip'
```
AMZN-Gene, line 87: add single-space after #

scripts/o3de/tests/test_repo_properties.py
```diff
@@ -0,0 +1,180 @@
...
+    "origin_url": "${OriginURL}",
+    "summary": "${Summary}",
+    "additional_info": "${AdditionalInfo}",
+    "last_updated": "",
+    "gems_data": [],
+    "projects_data": [],
+    "templates_data": []
+}
+'''
+# cls returns a dictionary of the template repo json
```
AMZN-Gene, line 31: what does "cls" stand for?

scripts/o3de/tests/test_repo_properties.py
```diff
@@ -0,0 +1,180 @@
...
+
+        with patch('o3de.repo_properties.get_repo_props', side_effect=get_repo_props) as get_repo_props_patch, \
+                patch('o3de.manifest.save_o3de_manifest', side_effect=save_o3de_manifest) as save_o3de_manifest_patch, \
+                patch('o3de.manifest.get_json_data', side_effect=get_json_data) as get_json_data_patch:
+            result = repo_properties.edit_repo_props(repo_path, repo_name, add_gems, delete_gems, replace_gems,
+                                                     add_projects, delete_projects, replace_projects, 
+                                                     add_templates, delete_templates, replace_templates)
+            assert result == expected_result
+            if result == 0:
+                assert self.repo_json.data.get('repo_name') == 'Repotest1'
```
AMZN-Gene, line 169: This assert should check repo_name not the hardcoded "Repotest1'
```
assert self.repo_json.data.get('repo_name') == repo_name
```

scripts/o3de/tests/test_repo_properties.py
```diff
@@ -0,0 +1,180 @@
...
+                         None, None, None, None,
+                         'template1 template2 template3', None, "template48", [{'template_name':'template48'}],
+                         0),
+            # test add all objects
+            pytest.param(pathlib.Path('F:/repotest'),'Repotest1', 
+                         'gem1', None, None, [{'gem_name':'gem1'}],
+                         'project1', None, None, [{'project_name':'project1'}],
+                         'template1 template2', None, None, [{'template_name':'template1'}, {'template_name':'template2'}],
+                         0),
+            # test remove all objects
```
AMZN-Gene, line 126: This "test remove all object types" not "test remove all objects" right?

AMZN-Gene (approved):

scripts/o3de/tests/test_repo_properties.py
VeryKaylin-player1: yes

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,399 @@
...
+def main():
+    the_parser = argparse.ArgumentParser()
+    add_parser_args(the_parser)
+    the_args = the_parser.parse_args()
+    ret = the_args.func(the_args) if hasattr(the_args, 'func') else 1
+    sys.exit(ret)
+
+
+if __name__ == "__main__":
+    main()
```
lumberyard-employee-dm, line 399: nitpick: The file is missing a newline at the end.

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,399 @@
...
+def create_release(src_data_path: pathlib.Path, 
+                   json_data_path: pathlib.Path, 
+                   archive_filename: pathlib.PurePath, 
+                   releases_path: pathlib.Path, 
+                   repo_uri:str,
+                   download_prefix:str) -> dict:
+    """
+    Creates a release for the given src_data_path.
+    The release is saved to the releases_path.
+    Returns the json data.
```
lumberyard-employee-dm, line 68: Informational question, but what is a "release"?

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,399 @@
...
+
+    logging.info(f"Creating '{releases_path / zip_filename}'")
+
+    # create the release zip file
+    shutil.make_archive(releases_path / basename, 'zip', src_data_path)
+
+    zip_path = releases_path / zip_filename
+
+    with zip_path.open('rb') as f:
+        json_data['sha256'] = hashlib.sha256(f.read()).hexdigest()
```
lumberyard-employee-dm, line 104: What if the user doesn't want to use a `zip`, but another commonly used format for different platform such as Linux or MacOS?
On Linux `tar.gz` and `tar.xz` is popular and MacOS `.xz` is popular

How about data driving the archive format via an argparse argument?

That way the user can choose the options that [shutil.make_archive](https://docs.python.org/3/library/shutil.html#shutil.make_archive) makes available.

AMZN-alexpete: Currently [our download archive code](https://github.com/o3de/o3de/blob/d9285c8e0c7804b64f51762ac6ef3efbd4907e83/scripts/o3de/o3de/download.py#L168) is looking for zip archive.  But we should definitely make this more flexible in the future!

scripts/o3de/o3de/repo_properties.py
AMZN-alexpete: The naming could probably be improved, this name refers to a release of a specific version of a remote object.  So every time you update a `repo.json` with a new version of a gem/project/template and tell us to make an archive, it calls this function and creates a "release", maybe we should instead calls this `create_remote_object_archive`

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,399 @@
...
+    Returns:
+        dict: The json data.
+    
+    Raises:
+        FileNotFoundError: If the src_data_path does not exist.
+        FileNotFoundError: If the json_data_path does not exist.
+        ValueError: If the json_data_path is not a dict.
+    """
+    # basename = f'{object_name}_release_1.0.0'
+    # zip_filename = f'{archive_filename}.zip'
```
AMZN-alexpete, line 87: lets remove this commented out code

```suggestion
```

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,399 @@
...
+    basename = archive_filename.stem
+    zip_filename = archive_filename
```
AMZN-alexpete, line 89: lets not use a temporary variable here - you can replace `zip_filename` with `archive_filename`
```suggestion
```

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,399 @@
...
+
+    json_data = merge_json_data(json_data_path, {
+        'repo_uri':repo_uri,
+        'download_source_uri':f'{download_prefix}/{zip_filename}',
+    })
+
+    logging.info(f"Creating '{releases_path / zip_filename}'")
+
+    # create the release zip file
+    shutil.make_archive(releases_path / basename, 'zip', src_data_path)
```
AMZN-alexpete, line 99: we don't really need the `basename` temporary variable either, can just use `archive_filename.stem`
```suggestion
    shutil.make_archive(releases_path / archive_filename.stem, 'zip', src_data_path)
```

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,399 @@
...
+
+    zip_path = releases_path / zip_filename
+
+    with zip_path.open('rb') as f:
+        json_data['sha256'] = hashlib.sha256(f.read()).hexdigest()
+
+    return json_data
+
+# retrieves json data of the repository from a file path
+def get_repo_props(path: pathlib.Path) -> dict:
```
AMZN-alexpete, line 109:
```suggestion
def get_repo_props(path: pathlib.Path) -> dict or None:
```

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,399 @@
...
+    :param release_archive_path: Path where you want your release to be located
+    :param download_prefix: The prefix of the download uri
+    """
+    # The beginning remote repo json template
+    repo_objects_data = repo_json.get(f'{object_typename}s_data',[])
+
+    if add_objects:
+        paths = add_objects.split() if isinstance(add_objects, str) else add_objects
+        for object_path in paths:
+            # gem data of the gem you want to add
```
AMZN-alexpete, line 157:
```suggestion
            # object data of the object you want to add
```

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,399 @@
...
+                  release_archive_path: pathlib.Path = None,
+                  download_prefix: str = None):
+    """
+    Edits and modifies the 'gem_data/projects_data/template_data' in the repo_json parameter
+    :param object_typename: The type object field you want to change
+    :param validator: validates the object you want to edit
+    :param add_objects: Any object paths you want to add to your repo_json object field
+    :param delete_objects: Any object_names to be removed from your repo_json object field
+    :param replace_objects: A list of object paths that will completely replace the current object_list
+    :param release_archive_path: Path where you want your release to be located
```
AMZN-alexpete, line 148:
```suggestion
    :param release_archive_path: Optional local path to a folder where a release archive should be written
```

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,399 @@
...
+                    version = json_data.get('version','0.0.0')
+                    archive_filename = pathlib.PurePath(f"{json_data[f'{object_typename}_name']}-{version}-{object_typename}.zip".lower())
+                    json_data = create_release(object_path, 
+                                   object_path / f'{object_typename}.json', 
+                                   archive_filename, 
+                                   release_archive_path, 
+                                   repo_json['repo_uri'], 
+                                   download_prefix)
+                index = _find_index(repo_objects_data, f'{object_typename}_name', json_data[f'{object_typename}_name'])
+                # if the added gem is same with partial changes
```
AMZN-alexpete, line 170:
```suggestion
                # index will be non-negative if an instance of this object already exists in the repo.json
```

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,399 @@
...
+                if index != -1:
+                    # we want changes only
+                    version_data = _changed(repo_objects_data[index], json_data)
+                    if not version_data:
+                        # there is no difference between the gem being added and 
+                        # what already exists in gems_data
```
AMZN-alexpete, line 176:
```suggestion
                        # what already exists in gems_data
                        logger.warning(f"No changes were found between the object JSON data in repo.json and the object data being added from {object_path}")
```

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,399 @@
...
+                                   release_archive_path, 
+                                   repo_json['repo_uri'], 
+                                   download_prefix)
+                index = _find_index(repo_objects_data, f'{object_typename}_name', json_data[f'{object_typename}_name'])
+                # if the added gem is same with partial changes
+                if index != -1:
+                    # we want changes only
+                    version_data = _changed(repo_objects_data[index], json_data)
+                    if not version_data:
+                        # there is no difference between the gem being added and
```
AMZN-alexpete, line 175:
```suggestion
                        # there is no difference between the object being added and 
```

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,399 @@
...
+                    repo_objects_data.append(json_data)
+
+    # Remove duplicates from list 
+    if delete_objects:      
+        removal_list = delete_objects.split() if isinstance(delete_objects, str) else delete_objects
+        # find name field of object you want to delete
+        object_key = str(f'{object_typename}_name')
+        for removal_object in removal_list:
+            # Remove the JSON object(s) with the specified object_name value
+            repo_objects_data = [object_typename for object_typename in repo_objects_data if object_typename.get(object_key) != removal_object]
```
AMZN-alexpete, line 211: we need to handle cases where the user wants to remove a specific version of an object and not every version of that object.  For example, if a repo.json has a single gem entry with 2 versions, the user should be able to just remove a single version of that gem from repo.json.  The way that command should look is something like this:
`o3de.bat repo delete-gems foo==2.0.0` 
The above command would only remove the gem with `gem_name` of `foo` and `version` `2.0.0`

If the user wanted to remove every version of a gem named `foo` they would use the command:
`o3de.bat repo delete-gems foo`
```suggestion
        removal_list = delete_objects.split() if isinstance(delete_objects, str) else delete_objects
        # find name field of object you want to delete
        object_key = str(f'{object_typename}_name')
        for removal_object in removal_list:
            object_name, version_specifier = utils.get_object_name_and_optional_version_specifier(removal_object)
            if not version_specifier :
                # Remove the JSON object(s) with the specified object_name value
                repo_objects_data = [object_typename for object_typename in repo_objects_data if object_typename.get(object_key) != object_name]
            else:
               index = _find_index(repo_objects_data, f'{object_typename}_name', object_name)
               if index > -1:
                   versions_data = repo_objects_data[index].get('versions_data',[])
                   if versions_data:
                       # remove matching versions
                       versions_data = [data for data in versions_data if Version(data.get('version', '0.0.0') not in  SpecifierSet(version_specifier)]
                       repo_objects_data[index]['versions_data'] = versions_data
                              
                   # remove/replace base version if it matches
                   if Version(repo_objects_data[index].get('version','0.0.0')) in SpecifierSet(version_specifier):
                       if versions_data:
                           # replace the base version content with all values from the first entry in versions_data
                           ...
                       else:
                           # versions data is empty so remove the whole object entry
                           del repo_objects_data[index]
               
```

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,399 @@
...
+
+    
+    # Replace 
+    if replace_objects:
+        paths = replace_objects.split() if isinstance(replace_objects, str) else replace_objects
+        for object_path in paths:
+            # clear suggect object field
+            repo_objects_data = []
```
AMZN-alexpete, line 219: instead of copying the code from above I recommend you either make a helper function and call it, or modify the logic for `add_objects` as suggested above

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,399 @@
...
+    :param replace_objects: A list of object paths that will completely replace the current object_list
+    :param release_archive_path: Path where you want your release to be located
+    :param download_prefix: The prefix of the download uri
+    """
+    # The beginning remote repo json template
+    repo_objects_data = repo_json.get(f'{object_typename}s_data',[])
+
+    if add_objects:
+        paths = add_objects.split() if isinstance(add_objects, str) else add_objects
+        for object_path in paths:
```
AMZN-alexpete, line 156: If looks like replace_objects is just like add_objects but cleears the repo_objects_data first so you could just do this:
```suggestion
    if add_objects or replace_objects:
        if replace_objects:
            repo_objects_data = []
            paths = replace_objects.split() if isinstance(replace_objects, str) else replace_objects
        else:
            paths = add_objects.split() if isinstance(add_objects, str) else add_objects
            
        for object_path in paths:
```

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,399 @@
...
+    group.add_argument('--repo-path', '-rp', type=pathlib.Path, required=False,
+                       help='The local path to the remote repository.')
+    group = parser.add_argument_group('properties', 'arguments for modifying individual remote repo properties.')
+    group.add_argument('--repo-name','-rn',  type=str, required=False,
+                       help='The name of the remote repository.')
+    group = parser.add_mutually_exclusive_group(required=False)
+    group.add_argument('--add-gems', '-ag', type=pathlib.Path, nargs='*', required=False,
+                       help="Adds gem(s) to the 'gems_data' property. Space delimited list (ex. -ag c:/gem1 c:/gem2)")
+    group.add_argument('--delete-gems', '-dg', type=str, nargs='*', required=False,
+                       help='Removes gems(s) from the gems_data property. Space delimited list (ex. -db A B C')
```
AMZN-alexpete, line 356:
```suggestion
                       help='Removes gems(s) from the gems_data property by gem name with optional version specifier. Space delimited list (ex. -dg gemA==1.0.0 gemB gemC>2.0.0')
```

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,399 @@
...
+    group.add_argument('--delete-templates', '-dt', type=str, nargs='*', required=False,
+                       help='Removes templates(s) from the templates_data property. Space delimited list (ex. -dt A B C')
+    group.add_argument('--replace-templates', '-rt', type=pathlib.Path, nargs='*', required=False,
+                       help='Replace entirety of templates_data property with the provided templates')
+
+    group = parser.add_mutually_exclusive_group(required=False)
+    group.add_argument('--add-projects', '-apr', type=pathlib.Path, nargs='*', required=False,
+                       help="Adds projects(s) to the 'projects_data' property. Space delimited list (ex. -at c:/project1 c:/project2)") 
+    group.add_argument('--delete-projects', '-dpr', type=str, nargs='*', required=False,
+                       help='Removes projects(s) from the projects_data property. Space delimited list (ex. -dp A B C')
```
AMZN-alexpete, line 372:
```suggestion
                       help='Removes projects(s) from the projects_data property by project name with optional version specifier. Space delimited list (ex. -dpr projectA==1.0.0 projectB projectC>2.0.0')
```

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,399 @@
...
+    group.add_argument('--delete-gems', '-dg', type=str, nargs='*', required=False,
+                       help='Removes gems(s) from the gems_data property. Space delimited list (ex. -db A B C')
+    group.add_argument('--replace-gems', '-rg', type=pathlib.Path, nargs='*', required=False,
+                       help='Replace entirety of gems_data property with the provided gems')
+
+    group = parser.add_mutually_exclusive_group(required=False)
+    group.add_argument('--add-templates', '-at', type=pathlib.Path, nargs='*', required=False,
+                       help="Adds template(s) to the 'templates_data' property. Space delimited list (ex. -at c:/template1 c:/template2)")
+    group.add_argument('--delete-templates', '-dt', type=str, nargs='*', required=False,
+                       help='Removes templates(s) from the templates_data property. Space delimited list (ex. -dt A B C')
```
AMZN-alexpete, line 364:
```suggestion
                       help='Removes templates(s) from the templates_data property by template name with optional version specifier. Space delimited list (ex. -dt templateA templateB==1.0.0 templateC>1.0.0')
```

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,399 @@
...
+                       help='Removes projects(s) from the projects_data property. Space delimited list (ex. -dp A B C')
+    group.add_argument('--replace-projects', '-rpr', type=pathlib.Path, nargs='*', required=False,
+                       help='Replace entirety of projects_data property with the provided projects')
+
+    modify_gems_group = parser.add_argument_group(title='modify gems',
+                                                  description='path arguments to use with the --add-gems or --replace-gems option')
+    modify_gems_group.add_argument('--release-archive-path','-rap',  type=pathlib.Path, required=False,
+                            help='Create a release archive at the specified local path and update the download_source_uri and sha256 fields.')
+    modify_gems_group.add_argument('--download-prefix','-dp',  type=str, required=False,
+                            help='Download prefix for the release archive.')
```
AMZN-alexpete, line 381:
```suggestion
                            help='Download prefix for the release archive. (ex. -dp https://github.com/o3de/o3de-extras/releases/download/2305.0/')
```

scripts/o3de/tests/test_repo_properties.py
```diff
@@ -0,0 +1,180 @@
...
+import pytest
+import pathlib
+from unittest.mock import patch
+
+from o3de import repo_properties
+
+TEST_TEMPLATE_REPO_JSON = '''
+{
+    "repo_name": "repotest",
+    "repo_uri": "https://test",
```
AMZN-alexpete, line 19: Let's use a valid URL
```suggestion
    "repo_uri": "https://test.com",
```

scripts/o3de/tests/test_repo_properties.py
```diff
@@ -0,0 +1,180 @@
...
+
+# use the TEST_TEMPLATE_REPO_JSON created above as the repo json template
+@pytest.mark.usefixtures('init_repo_json_data')
+class TestEditRepoProperties:
+
+    @staticmethod
+    def resolve(self):
+        return self
+
+    @pytest.mark.parametrize("repo_path, repo_name, \
```
AMZN-alexpete, line 48: The repo_path and repo_name are the same in all the tests, so I recommend removing them from the params or modify some of the tests to use different paths and names

scripts/o3de/tests/test_repo_properties.py
```diff
@@ -0,0 +1,180 @@
...
+                         'project1', None, None, [{'project_name':'project1'}],
+                         'template1 template2', None, None, [{'template_name':'template1'}, {'template_name':'template2'}],
+                         0),
+            # test remove all object types
+            pytest.param(pathlib.Path('F:/repotest'),'Repotest1', 
+                         'gem1 gem2', 'gem1', None, [{'gem_name':'gem2'}],
+                         'project1', 'project1', None, [],
+                         'template1 template2 template3', 'template1 template3', None, [{'template_name':'template2'}],
+                         0)
+            ]
```
AMZN-alexpete, line 132: We need to test adding multiple versions of gems and removing gems by version.  I recommend making a new test in this file specifically for version testing because it'll likely be more complex.  It should test:
1. Adding a gem with version 1.0.0 and then the same gem with version 2.0.0
2. Adding a gem with version 1.0.0 and then the same gem with version 2.0.0 and then removing version 2.0.0
3. Adding a gem with version 1.0.0 and then the same gem with version 2.0.0 and then removing version 1.0.0
4. Adding a gem with version 1.0.0 and then the same gem with version 2.0.0 and then removing version 1.0.0 and 2.0.0
5. Adding a gem with version 1.0.0 and then the same gem with the same version but other updated gem.json fields
6. Adding a gem with version 1.0.0 and then the same gem with version 2.0.0 and then adding version 2.0.0 with updated gem.json fields
7. Adding a gem with multiple versions and then replacing all gems with another set of gems

We will likely also need to add tests for users who have gems that don't contain a version field in their gem.json because they're older gems.  Something like:
1. Adding a gem without a version, then adding a gem with the same name but with a version
2. Adding a gem without a version, then adding a gem with the same name but with a version and then removing the unversioned gem by using the delete_gems = ['gemA==0.0.0']  or delete_gems = ['gemA<1.0.0']

scripts/o3de/tests/test_repo_properties.py
```diff
@@ -0,0 +1,180 @@
...
+        with patch('o3de.repo_properties.get_repo_props', side_effect=get_repo_props) as get_repo_props_patch, \
+                patch('o3de.manifest.save_o3de_manifest', side_effect=save_o3de_manifest) as save_o3de_manifest_patch, \
+                patch('o3de.manifest.get_json_data', side_effect=get_json_data) as get_json_data_patch:
+            result = repo_properties.edit_repo_props(repo_path, repo_name, add_gems, delete_gems, replace_gems,
+                                                     add_projects, delete_projects, replace_projects, 
+                                                     add_templates, delete_templates, replace_templates)
+            assert result == expected_result
+            if result == 0:
+                assert self.repo_json.data.get('repo_name') == repo_name
+                assert self.repo_json.data.get('repo_uri') == 'https://test'
```
AMZN-alexpete, line 170:
```suggestion
                assert self.repo_json.data.get('repo_uri') == 'https://test.com'
```

scripts/o3de/o3de/repo_properties.py
AMZN-Gene: I like seeing the example comment of what's in those vars. I couldnt intuit what "basename" stored

scripts/o3de/o3de/repo_properties.py
AMZN-Gene: Is implementing object removal based on version a large enough task (including tests) to move to a 2nd PR? Or is it required now, because you're saying removal would be currently broken without this logic?

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,377 @@
...
+    merged_json_data = {}
+    if json_path.exists():
+        existing_json_data = {}
+        with open(json_path, 'r') as f:
+            existing_json_data = json.load(f)
+
+        merged_json_data = existing_json_data | json_data
+
+        if not 'version' in merged_json_data:
+            merged_json_data['version'] = '1.0.0'
```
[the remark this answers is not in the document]
AMZN-alexpete: Yeah, that's the use case we want to consider here.  Currently, a missing "version" field means `0.0.0` and so we should use `0.0.0` if the value is missing.  I believe my code initially used 1.0.0 specifically for my use case which was updating the o3de-extras content.
```suggestion
        if not 'version' in merged_json_data:
            merged_json_data['version'] = '0.0.0'
```

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,377 @@
...
+        FileNotFoundError: If the json path does not exist.
+        ValueError: If the json data is not a dict.
+    """
+    merged_json_data = {}
+    if json_path.exists():
+        existing_json_data = {}
+        with open(json_path, 'r') as f:
+            existing_json_data = json.load(f)
+
+        merged_json_data = existing_json_data | json_data
```
AMZN-alexpete, line 50: We should check here if the data is the same and not modify the file if we don't have to which causes the file modification time to change and might make the file look modified to source control.
```suggestion
        with open(json_path, 'r') as f:
            existing_json_data = json.load(f)

        merged_json_data = existing_json_data | json_data
        
        if merged_json_data == existing_json_data and 'version' in existing_json_data then:
            # no changes were made, avoid making unnecessary change to the file modification time
            return existing_json_data
```

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,377 @@
...
+def _edit_objects(object_typename:str,
+                  validator:callable,
+                  repo_json: dict,
+                  add_objects: pathlib.Path or list = None,
+                  delete_objects: str or list = None,
+                  replace_objects: pathlib.Path or list = None,
+                  release_archive_path: pathlib.Path = None,
+                  download_prefix: str = None):
+    """
+    Edits and modifies the 'gem_data/projects_data/template_data' in the repo_json parameter
```
AMZN-alexpete, line 139:
```suggestion
    Modifies the 'gems_data/projects_data/templates_data' in repo_json 
```

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,377 @@
...
+                       if versions_data:
+                            # updates the base version with newer version
+                            repo_objects_data[index].update(versions_data.pop(0))
+                            repo_objects_data[index]['versions_data'] = versions_data
+                       else: 
+                            # versions data is empty so remove the whole object entry
+                            del repo_objects_data[index]              
+
+    repo_json[f'{object_typename}s_data'] = repo_objects_data
+     #if older version is removed, merged the items thats changed in the newer version if newer version exist and update version
```
AMZN-alexpete, line 242: I don't think this comment is in the correct place any more?
```suggestion
```

VeryKaylin-player1: yes, this should just be removed, sorry about that

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,377 @@
...
+            paths = replace_objects.split() if isinstance(replace_objects, str) else replace_objects
+        else:
+            paths = add_objects.split() if isinstance(add_objects, str) else add_objects
+        for object_path in paths:
+            # object JSON data of the object you want to add
+            json_data = manifest.get_json_data(object_typename, object_path, validator)
+            if json_data:
+                if release_archive_path:
+                    version = json_data.get('version','0.0.0')
+                    archive_filename = pathlib.PurePath(f"{json_data[f'{object_typename}_name']}-{version}-{object_typename}.zip".lower())
```
lumberyard-employee-dm, line 163: Can you add a comment with an example of the kind of path that this logic produces?

Such as 
```python
# for a project called TestProject that is version 1.0.0, this would create a filename of `testproject-1.0.0-project.zip`
```

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,377 @@
...
+                    # versions data must always include the version
+                    version_data['version'] = json_data.get('version','0.0.0')
+
+                    # versions data must always include one of the download/source uris
+                    if 'download_source_uri' in json_data:
+                        version_data['download_source_uri'] = json_data['download_source_uri']
+                    if 'source_control_uri' in json_data:
+                        version_data['source_control_uri'] = json_data['source_control_uri']
+                    if 'source_control_ref' in json_data:
+                        version_data['source_control_ref'] = json_data['source_control_ref']
```
lumberyard-employee-dm, line 190: Should there be a check for if none of the uris are provided?

AMZN-alexpete: We should add a warning here in case there isn't at least a `download_source_uri` or `source_control_uri`.  In theory there could be a `download_source_uri` in the gem data and not in the `versions_data` list that it could fall back to, but I think that would almost certainly not be what the user intends.

scripts/o3de/o3de/repo_properties.py
```diff
@@ -0,0 +1,377 @@
...
+
+    if add_gems or delete_gems or replace_gems:
+        _edit_objects('gem', validation.valid_o3de_gem_json, repo_json, add_gems, delete_gems, replace_gems, release_archive_path, download_prefix)
+
+    if add_projects or delete_projects or replace_projects:
+        _edit_objects('project', validation.valid_o3de_project_json, repo_json, add_projects, delete_projects, replace_projects, release_archive_path, download_prefix)
+
+    if add_templates or delete_templates or replace_templates:
+        _edit_objects('template', validation.valid_o3de_template_json, repo_json, add_templates, delete_templates, replace_templates, release_archive_path, download_prefix)
+
```
AMZN-alexpete, line 294: In the future, this is where we want to validate the modified `repo.json` file and backup the existing `repo.json` file before we write the updates to disk.

scripts/o3de/o3de/repo_properties.py
AMZN-Gene: source_control_uri isn't being used in o3de-extras. is there a working of this? assuming it's just the github.com path to the object's (gem/template/project) folder? or maybe it's a https://raw.githubusercontent.com/ link?

AMZN-alexpete: it's not required, you just have to have `download_source_uri` or `source_control_uri` or both.  For o3de-extras we prefer to use a downloadable archive because if we used the source control method we would be cloning multiple copies of the entire repo.

scripts/o3de/o3de/repo_properties.py
```diff
@@ -86,23 +91,27 @@ def create_remote_object_archive(src_data_path: pathlib.Path,
...
-
-    logging.info(f"Creating '{releases_path / archive_filename}'")
-
-    # create the release zip file
-    shutil.make_archive(releases_path / archive_filename.stem, 'zip', src_data_path)
-
     zip_path = releases_path / archive_filename
+    # check if a object.zip folder already exist in the path - ask user if they want to overwrite the current zip
+    if not force and zip_path.exists():
+        logger.error(f'{zip_path} already exists.  Use the --force to overwrite the existing zip or '
```
AMZN-Gene, line 97: Use --force (sans "the")

scripts/o3de/o3de/repo_properties.py
```diff
@@ -158,6 +168,8 @@ def _edit_objects(object_typename:str,
             # object JSON data of the object you want to add
             json_data = manifest.get_json_data(object_typename, object_path, validator)
             if json_data:
+                # for a project called TestProject that is version 1.0.0,
```
AMZN-Gene, line 171: +1 good comment

scripts/o3de/o3de/repo_properties.py
```diff
@@ -166,7 +178,9 @@ def _edit_objects(object_typename:str,
                                 archive_filename, 
                                 release_archive_path, 
                                 repo_json['repo_uri'], 
+                                force,
                                 download_prefix)
+                    if type(json_data) == int: return 1
```
AMZN-Gene, line 183: add a comment about what would cause json_data to be an int.
can also change this to be `if type(json_data) != Dict:`

scripts/o3de/o3de/repo_properties.py
```diff
@@ -180,8 +194,10 @@ def _edit_objects(object_typename:str,
 
                     # versions data must always include the version
                     version_data['version'] = json_data.get('version','0.0.0')
-
+
```
AMZN-Gene, line 197: added spaces

scripts/o3de/o3de/repo_properties.py
```diff
@@ -354,6 +372,8 @@ def add_parser_args(parser):
                                                   description='path arguments to use with the --add-gems or --replace-gems option')
     modify_gems_group.add_argument('--release-archive-path','-rap',  type=pathlib.Path, required=False,
                             help='Create a release archive at the specified local path and update the download_source_uri and sha256 fields.')
+    modify_gems_group.add_argument('--force', '-f', action='store_true', default=False,
+                                                help='Copies over instantiated template directory even if it exist.')
```
AMZN-Gene, line 376: "Copies over instantiated template directory even if it exist." is that accurate? This is used so that the release_archive (zip) will still be created even if a zip already exists (overwrite the existing zip)

AMZN-Gene (approved):

lumberyard-employee-dm (approved):

scripts/o3de/o3de/repo_properties.py
AMZN-alexpete: boo! I'm telling Obi-wan!

AMZN-alexpete (approved):

PR #16133, opened 2023-06-08
DPE: Reworked finding generic property handlers

### LINKED ISSUE #16057: DPE Inspector: Crash on loading certain combo boxes
**Describe the bug**
There are several different ways to reflect enum data to be shown in combo boxes, and some of them currently crash due to a recent change to attempt fixing a separate issue. We need to back out that part of the change and continue addressing the separate issue (#15953)

**Steps to reproduce**

3. Create a new entity.
4. Add a `Display Mapper` component to the entity.

**Expected behavior**
Component is added and Editor doesn't crash.

**Actual behavior**
Editor crashes.

The `DPEDebugViewStandalone` executable also crashes due to this bug.

cgalvan:
## What does this PR do?

Fixes #16057 

The initial fix to get asset property handlers to show (#15954) introduced a crash on certain enum/combo boxes. The first commit was to back out just the logic from that PR in `PropertyEditorAPI_Internals.h`. The new approach is similar in that we need to account for generic class info when searching for property handlers by type, but with two major differences:

1. No longer modify the `m_proxyClassData.m_typeId`, which led to crashes in other widgets (enum/combo boxes)
2. Instead of checking generic info in each handler every check, now we only fallback to looking at the generic info if the initial searches in the default and registered handlers fail to find a handler, which matches the logic flow for the RPE `PropertyManagerComponent::ResolvePropertyHandler`. 

## How was this PR tested?

Verified the `DPEDebugViewStandalone` no longer crashes and the `Editor` no longer crashes when adding a `Display Mapper` component. Tested `Mesh` component and other components with asset properties and verified the asset property widgets still show up.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/PropertyEditorToolsSystem.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/PropertyEditorToolsSystem.cpp
@@ -6,6 +6,7 @@
  *
  */
 
+#include <AzCore/Component/ComponentApplicationBus.h>
 #include <AzCore/Interface/Interface.h>
 #include <AzToolsFramework/UI/DocumentPropertyEditor/PropertyEditorToolsSystem.h>
 #include <AzToolsFramework/UI/DocumentPropertyEditor/ContainerActionButtonHandler.h>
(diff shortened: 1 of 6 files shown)


forkercat (approved):
Nice!

gadams3 (approved):

mbalfour-amzn (approved):

Issue #16135, opened 2023-06-09
DPE Inspector: Re-enable property handler recycling


cgalvan:
This is a future follow-up task to re-enable handler recycling in the DPE after it was disabled in #16127

The DPE has many optimizations over the RPE, one of them being the recycling of property handlers. The RPE creates a new property handler for every field that needs one, whereas the DPE recycling the instances once its done using them, so then subsequent properties that need a handler of the same type can just retrieve from the pool. The problem with this though is that many of the widgets, and even some of the handlers themselves, were written under the assumption of the RPE behavior where they'd be re-created everytime, so they store transient data specific to that field which never gets reset, causing unrelated fields to receive attributes and other data they shouldn't.

A mechanism was recently added to fix a specific use-case for this in a widget through a `ResetGUIToDefaults`, but this is currently only implemented for one widget, so it would need to be implemented for all of our supported widgets. In addition, this just applies to the widget, but we would also need to reset the handlers themselves as well.

We will need to come up with a plan for effectively resetting both the handlers and widgets and compile a list of all our supported handlers/widgets to update with reset methods in order to turn this back on.

PR #16136, opened 2023-06-09
Disable property handler recycling in the DPE

### LINKED ISSUE #16127: DPE Inspector: Turn off handler recycling
This needs more data, but essentially the handlers in the RPE are re-created for each field as needed, whereas we are recycling them in the DPE. But I've found many handlers (e.g. MeshNodeHandler, PropertyAssetCtrl, etc...) store transient data about the widget, such as info gathered during consuming of attributes, which is then carried over to other fields when they shouldn't be.

We need to take a deeper look at this, but possibly might need to disable the handler recycling in the short-term.

cgalvan:
## What does this PR do?

Fixes #16127 

We need to disable the property handler recycling feature in the DPE in the short-term. The RPE creates a new property handler for each field that needs one, and because of this behavior many of our property handlers and the widgets they create store transient data about the field they are attached to. With the recycling feature in the DPE, this transient data never gets reset, and so bleeds over into unrelated fields causing random inconsistencies.

I've created a backlog item to revisit the property handler recycling in the future: #16135
There was a mechanism that was created to fix a specific issue with re-use of one handler's widget, but this same approach would need to be done for all of our supported handlers and widgets.

## How was this PR tested?

Tested with breakpoints in several property handlers to verify that they were getting re-created each time instead of getting recycled.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
@@ -1736,15 +1736,8 @@ namespace AzToolsFramework
 
     void DocumentPropertyEditor::ReleaseHandler(HandlerInfo& handler)
     {
-        auto poolManager = static_cast<AZ::InstancePoolManager*>(AZ::Interface<AZ::InstancePoolManagerInterface>::Get());
-        auto handlerName = GetNameForHandlerId(handler.handlerId);
-        auto handlerPool = poolManager->GetPool<PropertyHandlerWidgetInterface>(handlerName);
-
-        if (handlerPool)
-        {
-            handlerPool->RecycleInstance(handler.handlerInterface);
-        }
-        else
+        // GHI-16135: Revisit recycling handler instances once we have a mechanism to reset the handlers/widgets for re-use
+        // and have implemented
         {
             // if there is no handler pool, then delete the handler immediately; parent widgets won't delete it twice
             delete handler.handlerInterface;


gadams3 (approved):

dtamkin1 (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
```diff
@@ -1736,15 +1736,8 @@ namespace AzToolsFramework
...
-        auto handlerName = GetNameForHandlerId(handler.handlerId);
-        auto handlerPool = poolManager->GetPool<PropertyHandlerWidgetInterface>(handlerName);
-
-        if (handlerPool)
-        {
-            handlerPool->RecycleInstance(handler.handlerInterface);
-        }
-        else
+        // GHI-16135: Revisit recycling handler instances once we have a mechanism to reset the handlers/widgets for re-use
+        // and have implemented
```
alexmontAmazon, line 1740: could you leave the code in but commented? Otherwise someone will have to go searching for this PR, or worse,  actually remember how to do this!

cgalvan:
> Should the code be disabled rather than deleted?

I prefer not to have commented out code. I left the GHI number/comment so we can always go back and find this PR to restore the previous code :)

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
cgalvan: Spoke offline, I've added the removed code into the follow-up task: https://github.com/o3de/o3de/issues/16135

PR #16147, opened 2023-06-12
Add CMake var for debug symbols in release builds, make Gem Assets inside Project folders higher priority

AMZN-alexpete:
## What does this PR do?

- Add `O3DE_BUILD_WITH_DEBUG_SYMBOLS_RELEASE` to output .pdb symbol files for release builds
- Set the default asset processor priority for Gems inside Project folders to be higher than the Project, so that if you place a Gem inside your Project, the Gem's asset paths in the cache are not relative to the Project root, but relative to the Gem `Assets` folder like they are when the Gem is outside the Project folder.

## How was this PR tested?

- Placed PopcornFX gem inside a project's folder and verified the paths were relative to the Popcorn `Assets` folder by looking at the source and product output in the Asset Processor.
- Verified when the `O3DE_BUILD_WITH_DEBUG_SYMBOLS_RELEASE` is enabled .pdb symbol files are generated in release mode

### DIFF
--- a/Templates/DefaultProject/Template/Registry/assetprocessor_settings.setreg
+++ b/Templates/DefaultProject/Template/Registry/assetprocessor_settings.setreg
@@ -16,7 +16,8 @@
                     "watch": "@PROJECTROOT@/Registry",
                     "recursive": 1,
                     "order": 3
-                }
+                },
+                "ProjectRelativeGemsScanFolderPriority":"higher"
             }
         }
     }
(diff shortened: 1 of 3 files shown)


lumberyard-employee-dm (approved):
Looks good to me.
We may want to add this to the reference page of the User Guide documentation of o3de.org: https://github.com/o3de/o3de.org/blob/development/content/docs/user-guide/build/reference.md

mbalfour-amzn (approved):

PR #16149, opened 2023-06-12
Fixed assigning asset properties in the DPE

### LINKED ISSUE #15953: DPE Inspector: Setting asset properties does not work

cgalvan:
## What does this PR do?

Fixes #15953 

This fix now makes it possible to write property value changes to the DOM for property types whose handler is for a sub-class (generic class info). The best use-case of this is being able to modify asset properties, which previous would cause a DOM error and throw many errors to the console.

The underlying issue is that for most properties, there is a handler for that exact type (e.g. primitives, AZ::Vector3, etc...) so we can infer the type from the value itself, for which we can just use `ValueFromType` when writing out the new value to the DOM. However, this fails for things like asset properties where there is a default handler for the underlying asset type `AZ::Data::Asset<AZ::Data::AssetData>`, but the actual type for the property is a derived class (e.g. `AZ::Data::Asset<RPI::ModelAsset>` for meshes). For this workflow, the handler is using the base `AssetData` for the `TypeId`, which is fine when reading the value in, but then when it tries to write it back to the DOM, the DOM is expecting the actual derived type. So we need to marshal the pointer to the expected type explicitly, which is how it is initially read in from the DOM at the very beginning.

BEFORE:
AssigningAssets_BEFORE

AFTER:
AssigningAssets_AFTER

## How was this PR tested?

Tested by assigning asset properties for various asset types (e.g. mesh, input) and verified they can now be assigned, whereas before they threw multiple errors and never assigned the new asset. Also saved the level after making asset property changes, and then re-loaded the level and verified the asset properties change stuck. Also tested modifying other properties (e.g. strings, boolean flags, transform position/rotation/scale, etc...) and verified they still go down the previous code path when writing out the value and still work as expected.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyEditorAPI_Internals.h
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyEditorAPI_Internals.h
@@ -398,7 +398,30 @@ namespace AzToolsFramework
             using AZ::DocumentPropertyEditor::Nodes::PropertyEditor;
 
             m_rpeHandler.WriteGUIValuesIntoProperty_Internal(GetWidget(), &m_proxyNode);
-            const AZ::Dom::Value newValue = AZ::Dom::Utils::ValueFromType(m_proxyValue);
+
+            auto typeIdAttribute = m_domNode.FindMember(PropertyEditor::ValueType.GetName());
+            AZ::TypeId typeId = AZ::TypeId::CreateNull();
+            if (typeIdAttribute != m_domNode.MemberEnd())
+            {
+                typeId = AZ::Dom::Utils::DomValueToTypeId(typeIdAttribute->second);
+            }
+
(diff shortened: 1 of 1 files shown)


alexmontAmazon (approved):
Nice! Super helpful fix :)

dtamkin1 (approved):

forkercat (approved):

AMZN-daimini (approved):

PR #16151, opened 2023-06-12
Fixes for Compile errors related to macOS16 (Ventura)

moudgils:
## What does this PR do?

Compile fixes for Ventura upgrade.

## How was this PR tested?

Tested ASV

### DIFF
--- a/Gems/Atom/RHI/Metal/Code/Source/RHI/PipelineState.cpp
+++ b/Gems/Atom/RHI/Metal/Code/Source/RHI/PipelineState.cpp
@@ -168,7 +168,12 @@ namespace AZ
                 [depthStencilDesc release];
                 depthStencilDesc = nil;
             }
+#if defined(__IPHONE_16_0) || defined(__MAC_13_0)
+            m_renderPipelineDesc.rasterSampleCount = descriptor.m_renderStates.m_multisampleState.m_samples;
+#else
             m_renderPipelineDesc.sampleCount = descriptor.m_renderStates.m_multisampleState.m_samples;
+#endif
+            
             m_renderPipelineDesc.alphaToCoverageEnabled = descriptor.m_renderStates.m_blendState.m_alphaToCoverageEnable;
             
             PipelineLibrary* pipelineLibrary = static_cast<PipelineLibrary*>(pipelineLibraryBase);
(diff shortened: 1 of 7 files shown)


Gems/Atom/RHI/Metal/Code/Source/RHI/PipelineState.cpp
```diff
@@ -168,7 +168,12 @@ namespace AZ
                 [depthStencilDesc release];
                 depthStencilDesc = nil;
             }
+#if defined(__IPHONE_16_0) || defined(__MAC_13_0)
```
AMZN-Olex, line 171: Is this okay by PAL standards?

moudgils: I think so. Even though the define macro uses platform names it is not a platform related update but an OS related update. We dont do PAL-ification for OS updates. So in this case we are protecting the api for specific OS version.

VickyAtAZ: How about the compatibility with the versions in the future? 
The new rasterSampleCount could be used for iphone 16 or any versions moving forward. 
The sampleCount should be versions which are lower then iphone 16 or  mac 13

VickyAtAZ (commented):
One small issue

Gems/Atom/RHI/Metal/Code/Source/RHI/PipelineState.cpp
lumberyard-employee-dm: Is this define for new versions of MacOS and iOS?

If MacOS 14 comes out is this still defined?

Shouldn't we check the  `__APPLE__` and `__IOS__` defines?

moudgils: So the two macros are guaranteed to always be defined going forward for every OS version but will not be defined for any versions before ios_16_0 and mac_13_0. That way rasterSampleCount will be picked up going forward but not backwards.

VickyAtAZ (approved):

lumberyard-employee-dm (approved):

Gems/Atom/RHI/Metal/Code/Source/RHI/PipelineState.cpp
moudgils: As an example here are details related to this from the header within macOS13.1
....
#define __MAC_11_6          110600
#define __MAC_12_0          120000
#define __MAC_12_1          120100
#define __MAC_12_2          120200
#define __MAC_12_3          120300
#define __MAC_13_0          130000
#define __MAC_13_1          130100
/* __MAC_NA is not defined to a value but is used as a token by macros to indicate that the API is unavailable */

PR #16154, opened 2023-06-12
Fix vulkan issues for Intel GPUs

akioCL:
## What does this PR do?

Fixes related to running Vulkan on Intel GPUs

## How was this PR tested?

Run ASV on Windows Vulkan using Integrated Intel GPU

### DIFF
--- a/Gems/Atom/RHI/Vulkan/Code/Source/RHI/PhysicalDevice.cpp
+++ b/Gems/Atom/RHI/Vulkan/Code/Source/RHI/PhysicalDevice.cpp
@@ -17,9 +17,9 @@ namespace AZ
 {
     namespace Vulkan
     {
-        static const uint32_t VendorID_Nvidia = 0x10DE;
-        static const uint32_t VendorID_AMD = 0x1002;
-        static const uint32_t VendorID_Intel = 0x8086;
+        [[maybe_unused]] static constexpr uint32_t VendorID_Nvidia = 0x10DE;
+        [[maybe_unused]] static constexpr uint32_t VendorID_AMD = 0x1002;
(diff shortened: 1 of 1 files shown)


AMZN-alexpete (approved):

Gems/Atom/RHI/Vulkan/Code/Source/RHI/PhysicalDevice.cpp
```diff
@@ -17,6 +17,10 @@ namespace AZ
 {
     namespace Vulkan
     {
+        static const uint32_t VendorID_Nvidia = 0x10DE;
```
moudgils, line 20: Shouldn't these be in the RHI common layer?

Gems/Atom/RHI/Vulkan/Code/Source/RHI/PhysicalDevice.cpp
```diff
@@ -284,6 +288,10 @@ namespace AZ
                 (m_vulkan12Features.separateDepthStencilLayouts));
             m_features.set(static_cast<size_t>(DeviceFeature::DescriptorIndexing), VK_DEVICE_EXTENSION_SUPPORTED(context, EXT_descriptor_indexing));
             m_features.set(static_cast<size_t>(DeviceFeature::BufferDeviceAddress), VK_DEVICE_EXTENSION_SUPPORTED(context, EXT_buffer_device_address));
+            // Disable memory budget extension for now since it's crashing the driver when querying heap budgets.
```
moudgils, line 291: It would be nice to add the crash related details in the comment here so if this fixed in future we will know when to remove this or just in general have better context around this issue.

moudgils (approved):

PR #16155, opened 2023-06-13
Fixed calculation of `bootstrap.<launcher-type>.<config>.setreg` SubIDs

lumberyard-employee-dm:
The SubID calculation was taking into account the project name and project launcher executable name which would cause a different Asset SubID to be generated based on the active project.

That is not needed since each project has it's own Asset Cache and therefore the SubIDs values should be the same across all projects.

## How was this PR tested?

Currently building the AssetProcessor right now to validate that the SubIDs of the `bootstrap.<launcher-type>.<config>.setreg` files are the same when created across two different projects.

Update: Verified that both the AtomSampleViewer and AutomatedTesting project produces the same SubIds for the `bootstrap.<launcher-type>.<config>.setreg` files

### DIFF
--- a/Code/Tools/AssetProcessor/native/InternalBuilders/SettingsRegistryBuilder.cpp
+++ b/Code/Tools/AssetProcessor/native/InternalBuilders/SettingsRegistryBuilder.cpp
@@ -35,7 +35,7 @@ namespace AssetProcessor
         builderDesc.m_busId = m_builderId;
         builderDesc.m_createJobFunction = AZStd::bind(&SettingsRegistryBuilder::CreateJobs, this, AZStd::placeholders::_1, AZStd::placeholders::_2);
         builderDesc.m_processJobFunction = AZStd::bind(&SettingsRegistryBuilder::ProcessJob, this, AZStd::placeholders::_1, AZStd::placeholders::_2);
-        builderDesc.m_version = 2;
+        builderDesc.m_version = 3;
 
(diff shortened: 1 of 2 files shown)


AMZN-Olex (approved):

AMZN-Gene (approved):

Code/Tools/AssetProcessor/native/InternalBuilders/SettingsRegistryBuilder.cpp
```diff
@@ -307,12 +307,12 @@ namespace AssetProcessor
...
                     file.Close();
 
-                    // Hash all specializations tags
+                    // Hash only the launcher type and build config specializations tags
                     size_t hashedSpecialization{};
-                    for (size_t specIndex{}; specIndex < specialization.GetCount(); ++specIndex)
-                    {
-                        AZStd::hash_combine(hashedSpecialization, specialization.GetSpecialization(specIndex));
-                    }
+                    // Get the launcher type specialization tag
```
AMZN-Gene, line 312: I'm assuming 0 and 1 are the indices to look up the launcher type and build config? Is there a variable tracking that special index? Maybe just to replace the "magic" numbers.

AMZN-Gene (approved):

Code/Tools/AssetProcessor/native/InternalBuilders/SettingsRegistryBuilder.cpp
lumberyard-employee-dm: Yes, 0 is the launcher type and 1 is the build config.
I'll add constants for both.

AMZN-daimini (approved):

PR #16158, opened 2023-06-13
AWS Util to Extracting Data from an ARN

AMZN-Gene:
Adding a new AWS Util for extracting data from an AWS ARN. 
Multiplayer Sample uses this to automatically find the AWS region in order to configure the client.

Helps fix: https://github.com/o3de/o3de-multiplayersample/pull/436

### DIFF
--- a/Gems/AWSCore/Code/Include/Framework/Util.h
+++ b/Gems/AWSCore/Code/Include/Framework/Util.h
@@ -39,8 +39,12 @@ namespace AWSCore
         //! Be aware that the ARNs for some resources omit the Region, the account ID, or both the Region and the account ID.
         inline AZStd::string ExtractArnData(AZStd::string_view awsArn, ArnFormatDataIndex arnDataIndex)
         {
+            // ARNs that omit region or account id will leave an empty entry.
+            // For example, IAM ARNs will omit the region (notice the 2 colons '::' specifying no region)
+            //     arn:aws:iam::123456789012:<user_data>
+            constexpr bool keepEmptyStrings = true;
             AZStd::vector<AZStd::string> tokenizedGameLiftArn;
-            AZ::StringFunc::Tokenize(awsArn, tokenizedGameLiftArn, ":");
+            AZ::StringFunc::Tokenize(awsArn, tokenizedGameLiftArn, ":", keepEmptyStrings);
             if (tokenizedGameLiftArn.size() >= arnDataIndex)
             {
                 return tokenizedGameLiftArn[arnDataIndex];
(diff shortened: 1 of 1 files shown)


AMZN-daimini (approved):

Gems/AWSCore/Code/Include/Framework/Util.h
```diff
@@ -9,12 +9,53 @@
...
+        //! Extracts information from an AWS ARN (Amazon Resource Name)
+        //! @param awsArn An AWS ARN. ARNs are formatted as arn:partition:service:region:account-id:<resource-type/id>
+        //! @param arnDataIndex The type of ARN data to extract. Possible values are Partition, Service, Region, or AccountId
+        //! @return A string representing extracted ARN data. Empty string is returned if data type isn't found.
+        //! Be aware that the ARNs for some resources omit the Region, the account ID, or both the Region and the account ID.
+        inline AZStd::string ExtractArnData(AZStd::string_view awsArn, ArnFormatDataIndex arnDataIndex)
+        {
+            AZStd::vector<AZStd::string> tokenizedGameLiftArn;
+            AZ::StringFunc::Tokenize(awsArn, tokenizedGameLiftArn, ":");
+            if (tokenizedGameLiftArn.size() >= arnDataIndex)
```
tkothadev, line 44: Are we sure this would work? If for example, partition, service, and accountId are specified, but region is not specified, then you would have a list of:
`[<partition>, <service>, <accountId>]`

So the resulting arn list size is >= region index (2), but it would fetch the wrong data. This also depends on how many ':' characters are present in the ARN string. There may be strings that contain redundant ':' characters (similar to how IPv6 addresses allow redundant characters in addressing).

lumberyard-employee-dm: Well empty entries are represented in the ARN as well with consecutive <colon> characters (`service::account-id`).
So It could be something such "resource::partition:AWSSSH::123456789:resource-id"

However the `StringFunc::Tokenize` function only adds empty entries if the [keepEmptyStrings](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/StringFunc/StringFunc.h#L274) option is supplied, which would be needed in this case

AMZN-Gene: Good call. I could make this specific to Multiplayer Sample? The format is based on AWS doc. I've placed a warning about some resources not including data, but without knowing exactly which resources, I think a warning with have to suffice.

https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html

AMZN-Gene: Thank you @lumberyard-employee-dm! This works great, to preserve empty strings
ARNs that omit region or account id will leave an empty entry.
For example, IAM ARNs will omit the region (notice the 2 colons '::' specifying no region)
    arn:aws:iam::123456789012:<user_data>

lumberyard-employee-dm (approved):

tkothadev (approved):

PR #16171, opened 2023-06-15
User can switch render pipeline at runtime for Editor's main viewport and GameLauncher's viewport

VickyAtAZ:
## What does this PR do?

User can set the pipeline's azasset file file to r_default_pipeline_name cvar at runtime to switch the render pipeline for Editor's 3d viewport or game launcher

The viewport debug info now has information of current render pipeline's name. Resolve the pass injections errors form hair FP and editorModeFeedback FP when switched to low end pipeline. Output warnings.

## How was this PR tested?

Tested with graphics/Stanza level and Graphics/PbrMaterialSheet level of AutomatedTesting project in Editor (dx12, Vulkan) and GameLauncher. 
Multiplayer Sample Project: works fine with dx12; got a device lost crash with vulkan when switch render pipeline.

### DIFF
--- a/Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
+++ b/Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
@@ -46,7 +46,35 @@
 #include <AzCore/Console/IConsole.h>
 #include <BootstrapSystemComponent_Traits_Platform.h>
 
-AZ_CVAR(AZ::CVarFixedString, r_default_pipeline_name, AZ_TRAIT_BOOTSTRAPSYSTEMCOMPONENT_PIPELINE_NAME, nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Default render pipeline name");
+void cvar_default_pipeline_name_Changed(const AZ::CVarFixedString& newPipelinePath)
+{
+    auto viewportContextManager = AZ::Interface<AZ::RPI::ViewportContextRequestsInterface>::Get();
+    if (!viewportContextManager)
+    {
+        return;
+    }
+    auto viewportContext = viewportContextManager->GetDefaultViewportContext();
+    if (!viewportContext)
+    {
+        return;
+    }
+
+    AZ::Data::Asset<AZ::RPI::AnyAsset> pipelineAsset =
+        AZ::RPI::AssetUtils::LoadAssetByProductPath<AZ::RPI::AnyAsset>(newPipelinePath.data(), AZ::RPI::AssetUtils::TraceLevel::Error);
+    if (pipelineAsset)
+    {
+        AZ::RPI::RenderPipelineDescriptor renderPipelineDescriptor =
+            *AZ::RPI::GetDataFromAnyAsset<AZ::RPI::RenderPipelineDescriptor>(pipelineAsset); // Copy descriptor from asset
+
+        AZ::Render::Bootstrap::RequestBus::Broadcast(&AZ::Render::Bootstrap::RequestBus::Events::SwitchRenderPipeline, renderPipelineDescriptor, viewportContext);
+    }
+    else
+    {
+        AZ_Warning("SetDefaultPipeline", false, "Failed to switch default render pipeline to %s: can't load the asset", newPipelinePath.data());
+    }
+}
+
+AZ_CVAR(AZ::CVarFixedString, r_default_pipeline_name, AZ_TRAIT_BOOTSTRAPSYSTEMCOMPONENT_PIPELINE_NAME, cvar_default_pipeline_name_Changed, AZ::ConsoleFunctorFlags::DontReplicate, "The asset (.azasset) path for default render pipeline");
 AZ_CVAR(AZ::CVarFixedString, r_default_openxr_pipeline_name, "passes/MultiViewRenderPipeline.azasset", nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Default openXr render pipeline name");
 AZ_CVAR(AZ::CVarFixedString, r_default_openxr_left_pipeline_name, "passes/XRLeftRenderPipeline.azasset", nullptr, AZ::ConsoleFunctorFlags::DontReplicate, "Default openXr Left eye render pipeline name");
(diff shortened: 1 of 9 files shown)


Gems/AtomLyIntegration/AtomViewportDisplayInfo/Code/Source/AtomViewportDisplayInfoSystemComponent.cpp
```diff
@@ -257,10 +259,8 @@ namespace AZ::Render
         RPI::PassSystemFrameStatistics passSystemFrameStatistics = AZ::RPI::PassSystemInterface::Get()->GetFrameStatistics();
 
         DrawLine(AZStd::string::format(
-            "RenderPasses: %d Vertex Count: %lld Primitive Count: %lld",
-            passSystemFrameStatistics.m_numRenderPassesExecuted,
-            aznumeric_cast<long long>(stats.m_vertexCount),
-            aznumeric_cast<long long>(stats.m_primitiveCount)
+            "RenderPasses: %d",
```
moudgils, line 262: Why remove Vert count / prim count?

Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
```diff
@@ -514,6 +541,38 @@ namespace AZ
...
+                        pipelineDescriptor.m_name.c_str());
+                    pipelineDescriptor.m_renderSettings.m_multisampleState.m_customPositions = {};
+                    pipelineDescriptor.m_renderSettings.m_multisampleState.m_customPositionsCount = 0;
+                }
+
+                // Create new render pipeline
+                auto oldRenderPipeline = viewportContext->GetRenderScene()->GetDefaultRenderPipeline();
+                RPI::RenderPipelinePtr newRenderPipeline = RPI::RenderPipeline::CreateRenderPipelineForWindow(
+                        pipelineDescriptor, *viewportContext->GetWindowContext().get(), AZ::RPI::ViewType::Default);
+
```
moudgils, line 567: Can you not just call LoadPipeline and then just add the code to remove the old pipeline?

Gems/AtomLyIntegration/AtomViewportDisplayInfo/Code/Source/AtomViewportDisplayInfoSystemComponent.cpp
VickyAtAZ: They are always showing as 0. there is no code to count those counts

Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
VickyAtAZ: I think it's better to have the API function to accept RenderPipelineDescriptor than a product asset string. This way, user may use code to switch to a render pipeline without an actual asset file. Also, I think the switching render pipeline shouldn't need as strict as start up which requires the asset must exist. It can fail as long as we have feedback when it failed.

Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
```diff
@@ -46,7 +46,35 @@
...
+        AZ::Render::Bootstrap::RequestBus::Broadcast(&AZ::Render::Bootstrap::RequestBus::Events::SwitchRenderPipeline, renderPipelineDescriptor, viewportContext);
+    }
+    else
+    {
+        AZ_Warning("SetDefaultPipeline", false, "Failed to switch default render pipeline to %s: can't load the asset", newPipelinePath.data());
+    }
+}
+
+AZ_CVAR(AZ::CVarFixedString, r_default_pipeline_name, AZ_TRAIT_BOOTSTRAPSYSTEMCOMPONENT_PIPELINE_NAME, cvar_default_pipeline_name_Changed, AZ::ConsoleFunctorFlags::DontReplicate, "The asset (.azasset) path for default render pipeline");
```
AMZN-alexpete, line 77: it is a little confusing that this cvar is a relative path instead of a name.

VickyAtAZ: The cvar's name is confusing. We need a path of an asset (azasset with RenderPipelineDescriptor) anyway. 
The trait's name is also using "_PIPELINE_NAME" but its data is actually a path.

I updated the comment which should help user know the cvar. 

I can rename the cvar since I think this was newly added and it was only used for Atom automated tests.

Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
```diff
@@ -514,6 +541,38 @@ namespace AZ
...
+                }
+
+                // Create new render pipeline
+                auto oldRenderPipeline = viewportContext->GetRenderScene()->GetDefaultRenderPipeline();
+                RPI::RenderPipelinePtr newRenderPipeline = RPI::RenderPipeline::CreateRenderPipelineForWindow(
+                        pipelineDescriptor, *viewportContext->GetWindowContext().get(), AZ::RPI::ViewType::Default);
+
+                // Switch render pipeline
+                viewportContext->GetRenderScene()->RemoveRenderPipeline(oldRenderPipeline->GetId());
+                viewportContext->GetRenderScene()->AddRenderPipeline(newRenderPipeline);
```
AMZN-alexpete, line 570: Is having both render pipelines loaded while switching resource intensive?

VickyAtAZ: No. The render pipelines don't cost much of memory.

AMZN-alexpete (approved):

Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
moudgils: I was mainly wondering about the duplicated code in this method and BootstrapSystemComponent::LoadPipeline and was wondering if we should consider refactoring it. But it's a minor point.

moudgils (approved):

galibzon (approved):

PR #16175, opened 2023-06-16
Add support for the JSON Importer to the Settings Registry

lumberyard-employee-dm:
The `$import` directive that is supported with JSON Serialization is now supported in the Settings Registy to trigger merges of referenced files.

## How was this PR tested?

Verified JsonImporter and SettingsRegistryTest all pass.
Added UnitTest to validate `$import` directive in JSON files.


Code/Framework/AzCore/AzCore/Serialization/Json/JsonImporter.cpp
```diff
@@ -144,21 +159,170 @@ namespace AZ
...
+
+    JsonSerializationResult::ResultCode JsonImportResolver::StoreImports(const rapidjson::Value& jsonDoc,
+        ImportPathStack& importPathStack,
+        JsonImportSettings& settings, StackedString& element)
+    {
+        using namespace JsonSerializationResult;
+
+        if (settings.m_importer == nullptr)
+        {
+            AZStd::string reportMessage = "Json Importer is nullptr for ResolveImports.";
```
AMZN-alexpete, line 203:
```suggestion
            AZStd::string reportMessage = "Json Importer is nullptr for StoreImports.";
```

Code/Framework/AzCore/AzCore/Serialization/Json/JsonImporter.cpp
```diff
@@ -144,21 +159,170 @@ namespace AZ
...
+            return settings.m_reporting(reportMessage,
+                ResultCode(Tasks::Import, Outcomes::Skipped), element);
+        }
+
+        if (jsonDoc.IsObject())
+        {
+            for (auto& field : jsonDoc.GetObject())
+            {
+                if (strncmp(field.name.GetString(), JsonSerialization::ImportDirectiveIdentifier, field.name.GetStringLength()) == 0)
+                {
```
AMZN-alexpete, line 217: `ResolveImports` and `StoreImports` are similar enough it may be useful to move some of this inner code to a common function if it will help prevent fixing something in one function and forgetting to make the fix/change in the other.

Code/Framework/AzCore/AzCore/Settings/SettingsRegistryImpl.cpp
```diff
@@ -1356,104 +1191,127 @@ namespace AZ
...
-            pointer.Create(m_settings, m_settings.GetAllocator()).SetObject()
-                .AddMember(StringRef("Error"), StringRef("Unable to read registry file."), m_settings.GetAllocator())
-                .AddMember(StringRef("Path"), Value(path, m_settings.GetAllocator()), m_settings.GetAllocator());
+            // If the file cannot be complete read into memoryy
+            // return 0 as the size for the string
+            return fileReader.Read(size, buffer) == size ? size : 0;
+        };
+        jsonData.resize_and_overwrite(fileSize, ReadJsonIntoString);
 
+        // If the strin is empty then the file could not be read
```
AMZN-alexpete, line 1248:
```suggestion
        // If the string is empty then the file could not be read
```

Code/Framework/AzCore/AzCore/Settings/SettingsRegistryImpl.cpp
```diff
@@ -1463,58 +1321,64 @@ namespace AZ
             break;
         case Format::JsonMergePatch:
             mergeApproach = JsonMergeApproach::JsonMergePatch;
-            if (!jsonPatch.IsObject())
+             if (anchorKey.empty() && !jsonPatch.IsObject())
```
AMZN-alexpete, line 1324: looks like a space snuck in at the beginning of the line
```suggestion
            if (anchorKey.empty() && !jsonPatch.IsObject())
```

Code/Framework/AzCore/AzCore/Serialization/Json/JsonImporter.cpp
lumberyard-employee-dm: Oops.
Will fix.

Code/Framework/AzCore/AzCore/Serialization/Json/JsonImporter.cpp
lumberyard-employee-dm: Well the way the logic is interleaved makes it hard to combine them, especially with the calls to ResolvedNestedImports vs StoreNestedImports

Code/Framework/AzCore/AzCore/Settings/SettingsRegistryImpl.cpp
AMZN-alexpete: is it OK if the json patch is not an object if an anchor key is provided?

Code/Framework/AzCore/AzCore/Settings/SettingsRegistryImpl.cpp
```diff
@@ -1533,57 +1397,52 @@ namespace AZ
             };
         }
 
-        ScopedMergeEvent scopedMergeEvent(*this, { path, rootKey });
-        JsonSerializationResult::ResultCode mergeResult(JsonSerializationResult::Tasks::Merge);
-        SettingsType anchorType;
-        if (rootKey.empty())
+        // Create a scoped merge event to triger the PreMerge notification on construction
```
AMZN-alexpete, line 1400:
```suggestion
        // Create a scoped merge event to trigger the PreMerge notification on construction
```

Code/Framework/AzCore/Tests/Settings/SettingsRegistryTests.cpp
```diff
@@ -1473,16 +1439,44 @@ namespace SettingsRegistryTests
...
-        bool result = m_registry->Get(history, AZ_SETTINGS_REGISTRY_HISTORY_KEY "/0");
-        EXPECT_TRUE(result);
-        EXPECT_STREQ(path.c_str(), history.c_str());
-
         AZ::s64 intValue{};
         EXPECT_TRUE(m_registry->Get(intValue, "/TestObject/IntValue"));
         EXPECT_EQ(7, intValue);
     }
 
+    TEST_F(SettingsRegistryTest, MergeSettingsFile_ObjectWithImport_MergesImportedFile)
```
AMZN-alexpete, line 1447: If we don't have one already, it might be good to include a test that imports multiple files in the same setreg - e.g.
```
{
    "$import":"x.setreg",
    "$import":"y.setreg"
}
```

Code/Framework/AzCore/AzCore/Settings/SettingsRegistryImpl.cpp
```diff
@@ -703,126 +703,11 @@ namespace AZ
...
-
-        MergeSettingsResult mergeResult;
-        mergeResult.m_returnCode = MergeSettingsReturnCode::Success;
-        return mergeResult;
+        return MergeSettingsString(data, format, anchorKey, AZ::IO::PathView{});
     }
 
     auto SettingsRegistryImpl::MergeSettingsFile(AZStd::string_view path, Format format, AZStd::string_view rootKey,
-        AZStd::vector<char>* scratchBuffer)
+        [[maybe_unused]] AZStd::vector<char>* scratchBuffer)
```
AMZN-alexpete, line 710: I'm not familiar enough with this functionality to know the implications of removing this scratch buffer functionality - or is it not needed here because the JSON importer allows you to provide the scratch buffer with a different method?

lumberyard-employee-dm: It wasn't useful in the first place.
It was just used to load the file content into.
It was premature optimization for a system that doesn't need it.

lumberyard-employee-dm: If anything, any such buffers, should be supplied as part of some settings structure that gets passed into the function itself, instead of being a dedicated parameter.

Code/Framework/AzCore/AzCore/Settings/SettingsRegistryImpl.cpp
lumberyard-employee-dm: Yes, it is, in the case, because only the the value at the anchor field will be overridden.

The anchor key check here is to prevent the **entire** Settings Registry from being overridden if the user supplies a JSON non-object type.

The following would override the entire settings registry with an array of [1, 2, 3], which we disallow, as it would override other systems settings inadvertently
```c++
settingsRegistry->MergeSettings(R"(
    [
        1,
        2,
        3
    ]
)", SettingsRegistryInterface::Format::JsonMergePatch, "");
```

However we do allow using a JSON non-object if the JSON is not anchored at the root
```
```c++
settingsRegistry->MergeSettings(R"(
    [
        1,
        2,
        3
    ]
)", SettingsRegistryInterface::Format::JsonMergePatch, "/O3DE/MyAtom/Custom/Settings");
```

Code/Framework/AzCore/AzCore/Settings/SettingsRegistryImpl.cpp
```diff
@@ -1397,41 +1372,55 @@ namespace AZ
...
 
-        // Create a scoped merge event to triger the PreMerge notification on construction
-        // and PostMerge event on destruction
-        ScopedMergeEvent scopedMergeEvent(*this, { importSettings.m_loadedJsonPath.Native(), anchorKey});
-
-        // First resolve any $import directives in the merge JSON blob
+        // First resolve any $import directives in the source JSON blob in place, while maintaining
+        // the order of fields
+        // Any fields that are after $import directives override the imported JSON data, while fields before
+        // are overridden by the $import directive
```
AMZN-alexpete, line 1378: nice!

Code/Framework/AzCore/AzCore/Serialization/Json/JsonImporter.cpp
```diff
@@ -159,6 +160,311 @@ namespace AZ
...
+            applyPatchSettings.m_reporting = issueReporter;
+
+            rapidjson::Value& importedDoc = importedObject.GetValue();
+            if (importDirective.IsObject())
+            {
+                auto patchField = importDirective.FindMember("patch");
+                if (patchField != importDirective.MemberEnd())
+                {
+                    JsonSerialization::ApplyPatch(importedDoc, allocator, patchField->value,
+                        AZ::JsonMergeApproach::JsonMergePatch, applyPatchSettings);
```
AMZN-alexpete, line 185: this part is a little confusing to me, is the scenario where you use this when you have a file with an import like this?

 `test.setreg` 
```json
{
   "$import":{
       "patch":"somefile.setreg"
   }
   ...
}
```

So first it merges the patch from `somefile.setreg` into `test.setreg` and then merge the result into `target`?

lumberyard-employee-dm: Well the "patch" isn't a file, but a JSON string.
So it merges the patch data into the current JSON file that is being imported and then merges imported JSON to the target.

Code/Framework/AzCore/AzCore/Serialization/Json/JsonImporter.cpp
```diff
@@ -159,6 +160,311 @@ namespace AZ
...
+
+                    mergePatchResult.Combine(ResolveImportsInOrder(newFieldMemberIter->value, allocator, *fieldValue,
+                        importPathStack, settings, element));
+                }
+                else if (fieldValue->IsNull())
+                {
+                    if (targetField != targetDataAtCurrentDepth.MemberEnd())
+                    {
+                        targetDataAtCurrentDepth.RemoveMember(targetField);
+                    }
```
AMZN-alexpete, line 441: could importing/merging `"field":null` could be used to remove a field?

lumberyard-employee-dm: Actually yes.
The algorithm here is actually a modified [JSON Merge Patch](https://datatracker.ietf.org/doc/html/rfc7386#section-1) algorithm, that has a special case were if it finds an `"$import"` field, it instead merges the content of the import file into the parent field of the `$import` key.

AMZN-alexpete (approved):

Code/Framework/AzCore/AzCore/Serialization/Json/JsonImporter.cpp
AMZN-alexpete: Cool thanks!

Code/Framework/AzCore/AzCore/Serialization/Json/JsonImporter.cpp
lumberyard-employee-dm: Well it turns out removing null entries with the ResolveImportsInorder causes a problem when reading fields that are legitimately null. 

For example reading in a JSON patch that is trying to set a field to the value of `null` is legitimate use of regular [JSON Patch](https://datatracker.ietf.org/doc/html/rfc6902) (Note: [JSON Merge Patch](https://datatracker.ietf.org/doc/html/rfc7386#section-1) is a different algorithm from [JSON Patch](https://datatracker.ietf.org/doc/html/rfc6902)).

Due to the currently logic fields with null values are being removed before they can be actually merged.
That is causing UnitTest to fail such as this [test](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/Tests/Settings/SettingsRegistryTests.cpp#L682-L691)

Code/Framework/AzCore/AzCore/Serialization/Json/JsonImporter.cpp
```diff
@@ -144,21 +160,606 @@ namespace AZ
...
+        JsonImportSettings& settings, const AZ::IO::FixedMaxPath& importPath, StackedString& element)
+    {
+        using namespace JsonSerializationResult;
+
+        for (auto& path : importPathStack)
+        {
+            if (importPath == path)
+            {
+                return settings.m_reporting(
+                    AZStd::string::format("'%s' was already imported in this chain. This indicates a cyclic dependency.", importPath.c_str()),
```
AMZN-alexpete, line 609: nit it looks like there are a few functions with ImportPathStack (489, 412) and some of the reporting uses c_str and others use `AZ_PATH_ARG` - not sure if it really matters, very minor thing.

AMZN-alexpete (approved):

Code/Framework/AzCore/AzCore/Serialization/Json/JsonImporter.cpp
```diff
@@ -144,21 +160,606 @@ namespace AZ
...
+        pathsResult.m_importRelPath = importName;
+
+        return pathsResult;
+    }
+
+    // The algorithm below is a modified JSON Merge Patch algorithm that is for patching a specific field
+    // in the target JSON value
+    // The modifications are that any `null` entries do not revome existing fields from the target JSON
+    // The reason this is needed is because the Json Import operation should preserve existing fields
+    // in case the JSON data is used for an JSON patch or JSON merge patch operation explicitly
```
mbalfour-amzn, line 300:
```suggestion
    // in case the JSON data is used for a JSON patch or JSON merge patch operation explicitly
```

Code/Framework/AzCore/AzCore/Serialization/Json/JsonImporter.cpp
```diff
@@ -144,21 +160,606 @@ namespace AZ
...
+                    AZ_PATH_ARG(importAbsPath), importedObject.GetError().c_str()),
+                JsonSerializationResult::ResultCode(JsonSerializationResult::Tasks::Import,
+                    JsonSerializationResult::Outcomes::Catastrophic), element.Get());
+        }
+    }
+
+    JsonSerializationResult::ResultCode JsonImportResolver::ResolveNestedImportsInOrder(rapidjson::Value& target,
+        rapidjson::Document::AllocatorType& allocator, const rapidjson::Value& source,
+        ImportPathStack& importPathStack,
+        JsonImportSettings& settings, StackedString& element, AZ::IO::PathView importPath)
```
mbalfour-amzn, line 224: There are a lot of parameters on these new methods. Just double-checking - do the order of the parameters on these make logical sense from a caller point of view? I'm just asking because the placement of target vs source and importPath vs importPathStack seem a bit weird to me just looking at it, but maybe it makes more sense in the larger context. 

Either way, I don't have any blocking concerns since the parameters are well-named and easy to understand.

Code/Framework/AzCore/AzCore/Serialization/Json/JsonImporter.cpp
```diff
@@ -144,21 +160,606 @@ namespace AZ
...
+    // in case the JSON data is used for an JSON patch or JSON merge patch operation explicitly
+    JsonSerializationResult::ResultCode JsonImportResolver::PatchField(rapidjson::Value& target,
+        rapidjson::Document::AllocatorType& allocator, const rapidjson::Value& source,
+        rapidjson::Value::StringRefType fieldNameStringRef, const rapidjson::Value& fieldValue,
+        ImportPathStack& importPathStack,
+        JsonImportSettings& settings, StackedString& element)
+    {
+        using namespace JsonSerializationResult;
+
+        using ReporterString = AZStd::fixed_string<1024>;
```
mbalfour-amzn, line 309: Should this definition get pushed out somewhere a little more global, so that we don't have a bunch of hard-coded fixed_string<1024> entries everywhere?

Code/Framework/AzCore/AzCore/Serialization/Json/JsonImporter.cpp
```diff
@@ -144,21 +160,606 @@ namespace AZ
...
+                    // Add an empty object or array field depending on the field value type
+                    target.AddMember(AZStd::move(name),
+                        rapidjson::Value{ rapidjson::kObjectType },
+                        allocator);
+                    auto newFieldMemberIter = target.FindMember(fieldNameStringRef);
+                    if (newFieldMemberIter == target.MemberEnd())
+                    {
+                        AZStd::string_view fieldName(static_cast<const char*>(fieldNameStringRef),
+                            fieldNameStringRef.length);
+                        auto reportMessage = ReporterString::format("Failed to add member %.*s to JSON while importing.",
```
mbalfour-amzn, line 352: If you wanted, you could use AZ_STRING_FORMAT instead of %.*s. I tend to use the former, but I'm totally supportive of both. Just mentioning it here because it's a companion define for AZ_STRING_ARG.

Code/Framework/AzCore/AzCore/Serialization/Json/JsonImporter.cpp
```diff
@@ -144,21 +160,606 @@ namespace AZ
...
+                    }
+                    return settings.m_reporting(reportMessage, ResultCode(Tasks::Import, Outcomes::Catastrophic), element);
+                }
+
+                newFieldValue = &target[oldArrayCount];
+            }
+
+            // Array fields are recursively checked for children $import directives
+            mergePatchResult.Combine(ResolveImportsInOrder(*newFieldValue, allocator, fieldValue,
+                importPathStack, settings, element));
```
mbalfour-amzn, line 443: This all looks cut-n-pasted from fieldValue.IsObject() case directly above. Can it be turned into a helper function (or even internal lambda) to make it shared logic? It would be extremely hard to spot if the two ever got slightly out of sync accidentally.

Code/Framework/AzCore/AzCore/Serialization/Json/JsonImporter.cpp
```diff
@@ -144,21 +160,606 @@ namespace AZ
...
+        // into a local JSON value that will be moved to the target JSON value at the end of this function
+        rapidjson::Value targetDataAtCurrentDepth{ source.IsObject()
+            ? rapidjson::kObjectType : rapidjson::kArrayType };
+
+        // If this point is reached, the source is either an object or an array
+        size_t fieldCount = source.IsObject() ? source.MemberCount() : source.Size();
+        for (size_t fieldIndex = 0; fieldIndex < fieldCount; ++fieldIndex)
+        {
+            // Used to store the converted field name for an array field
+            using FieldNameString = AZStd::fixed_string<256>;
```
mbalfour-amzn, line 515: If this is an expected limitation of field name sizes, the 256 should get promoted somewhere more visible.

Code/Framework/AzCore/AzCore/Serialization/Json/JsonImporter.h
```diff
@@ -85,25 +97,85 @@ namespace AZ
             rapidjson::Document::AllocatorType& allocator, ImportPathStack& importPathStack,
             JsonImportSettings& settings, StackedString& element);
 
+        //! Resolves the imports from the source JSON object in the field order.
+        //! It does this by iterating over the fields of the source JSON object and copies them
+        //! to the target object as it processes them.
+        //! Any $import directives are merged  to the target object at that point using the JSON Merge Patch
```
mbalfour-amzn, line 103:
```suggestion
        //! Any $import directives are merged to the target object at that point using the JSON Merge Patch
```

Code/Framework/AzCore/AzCore/Serialization/Json/JsonMerger.cpp
```diff
@@ -285,7 +285,7 @@ namespace AZ
         }
 
         rapidjson::Value newValue;
-        newValue.CopyFrom(value->value, allocator);
+        newValue.CopyFrom(value->value, allocator, true);
```
mbalfour-amzn, line 288: Suggestion: Consider using a constexpr bool X = true for these CopyFrom calls so that the parameter has a descriptive name. Just reading it in the context of the PR, i can't explain what this parameter is or why its presence is good.

Code/Framework/AzCore/AzCore/Settings/SettingsRegistry.h
```diff
@@ -394,6 +394,10 @@ namespace AZ
         //! into the Settings Registry
         struct MergeSettingsResult
         {
+            //! Any values that is >= MergeSettingsReturnCode::Unset
+            //! is treated as a successful operation
+            //! This means that a default initialized MergeSettingsResult
+            //! is convertible to true
```
mbalfour-amzn, line 400: What are the implications of changing the default initialization from "convertible false" to "convertible true"?

Code/Framework/AzCore/AzCore/Settings/SettingsRegistryImpl.h
```diff
@@ -112,9 +110,16 @@ namespace AZ
...
-            const rapidjson::Pointer& historyPointer, AZStd::string_view folderPath);
+            AZStd::string_view folderPath);
         bool ExtractFileDescription(RegistryFile& output, AZStd::string_view filename, const Specializations& specializations);
-        MergeSettingsResult MergeSettingsFileInternal(const char* path, Format format, AZStd::string_view rootKey, AZStd::vector<char>& scratchBuffer);
+        MergeSettingsResult MergeSettingsFileInternal(const char* path, Format format, AZStd::string_view rootKey);
+        MergeSettingsResult MergeSettingsJsonDocument(const rapidjson::Document& jsonPatch, Format format, AZStd::string_view rootKey,
+            AZ::IO::PathView filePath);
+
+        //! The filePath here is for the loaded json content in the string parameter
+        MergeSettingsResult MergeSettingsString(AZStd::string insituData, Format format, AZStd::string_view anchorKey,
```
mbalfour-amzn, line 120: "in situ" is two words, so...

```suggestion
        MergeSettingsResult MergeSettingsString(AZStd::string inSituData, Format format, AZStd::string_view anchorKey,
```

Is this supposed to copy the inSituData as the parameter, instead of take a reference? It feels like it's literally NOT using the data in place if it's making a copy...

Code/Framework/AzCore/AzCore/Settings/SettingsRegistryImpl.cpp
```diff
@@ -1533,57 +1372,66 @@ namespace AZ
...
-        JsonSerializationResult::ResultCode mergeResult(JsonSerializationResult::Tasks::Merge);
+        // First resolve any $import directives in the source JSON blob in place, while maintaining
+        // the order of fields
+        // Any fields that are after $import directives override the imported JSON data, while fields before
+        // are overridden by the $import directive
+        AZ::StackedString importedFieldKey(AZ::StackedString::Format::JsonPointer);
+        JsonImportResolver::ImportPathStack importPathStack;
+        importPathStack.push_back(filePath);
+
+        // JSON Importer lifetime needs to be larger than the improt settings
```
mbalfour-amzn, line 1383:
```suggestion
        // JSON Importer lifetime needs to be larger than the import settings
```

Code/Framework/AzCore/AzCore/Settings/SettingsRegistryImpl.cpp
```diff
@@ -1533,57 +1372,66 @@ namespace AZ
...
+        AZ::BaseJsonImporter jsonImporter;
+        AZ::JsonImportSettings importSettings;
+        importSettings.m_reporting = [&mergeResult](AZStd::string_view message,
+            AZ::JsonSerializationResult::ResultCode result, AZStd::string_view)
+        {
+            // Store any JSON Importer messages as par of the merge result
```
mbalfour-amzn, line 1389:
```suggestion
            // Store any JSON Importer messages as part of the merge result
```

Code/Framework/AzCore/Tests/Settings/SettingsRegistryTests.cpp
```diff
@@ -1473,16 +1439,125 @@ namespace SettingsRegistryTests
...
+        EXPECT_EQ("override", stringValue);
+
+        intValue = {};
+        EXPECT_TRUE(m_registry->Get(intValue, FixedValueString(anchorKey) + "/Memory"));
+        EXPECT_EQ(1, intValue);
+
+        // Check that the "pre_field" is object "first" field is overridden
+        // but the "post_field" overrides the import files value
+        // The "Memory.mobile.setreg" would merge over the "Memory.iphone13.setreg"
+        // because it's import values appear afterwards
```
mbalfour-amzn, line 1484:
```suggestion
        // because its import values appear afterwards
```

Code/Framework/AzCore/Tests/Settings/SettingsRegistryTests.cpp
```diff
@@ -1473,16 +1439,125 @@ namespace SettingsRegistryTests
...
         EXPECT_TRUE(m_registry->Get(intValue, "/TestObject/IntValue"));
         EXPECT_EQ(7, intValue);
     }
 
+    TEST_F(SettingsRegistryTest, MergeSettingsFile_JsonWithImport_MergesImportedFiles_InOrder)
+    {
+        auto filePath = CreateTestFile("Memory.iphone13.setreg", R"({ "pre_field": {"first": 7 },
+            "$import": "Memory.ios.setreg", "non_override": 7, "post_field": {"2nd": "twenty-one"} })");
+        CreateTestFile("Memory.ios.setreg", R"({ "$import": { "filename": "Memory.mobile.setreg" },
+            "non_override2": "Hello World", "override_imported_settings": "override" })");
```
mbalfour-amzn, line 1452: Is it ok to use the names "ios" and "iphone" on all platforms like this? I'd suggest using more generic names in the test (ex: "os", "phone") just in case.

Code/Framework/AzCore/Tests/Settings/SettingsRegistryTests.cpp
```diff
@@ -1473,16 +1439,125 @@ namespace SettingsRegistryTests
...
+        intValue = {};
+        EXPECT_TRUE(m_registry->Get(intValue, FixedValueString(anchorKey) + "/pre_field/first"));
+        EXPECT_EQ(14, intValue);
+
+        stringValue.clear();
+        EXPECT_TRUE(m_registry->Get(stringValue, FixedValueString(anchorKey) + "/pre_field/second"));
+        EXPECT_EQ("fourteen", stringValue);
+
+        stringValue.clear();
+        // The "Memory.iphone13.setreg" "post_field" should retain it's value
```
mbalfour-amzn, line 1494:
```suggestion
        // The "Memory.iphone13.setreg" "post_field" should retain its value
```

Code/Framework/AzCore/AzCore/Serialization/Json/JsonImporter.cpp
lumberyard-employee-dm: Well this function in particular is a private function.

The arguments are in a simliar order to the `ResolveNestedImports` function

Code/Framework/AzCore/AzCore/Serialization/Json/JsonImporter.cpp
lumberyard-employee-dm: I avoid AZ_STRING_FORMAT because it is longer than "%.*s", but I used AZ_STRING_ARG because it is shorter than a `static_cast<int>(variableName.size()), variableName.data())`

Code/Framework/AzCore/Tests/Settings/SettingsRegistryTests.cpp
lumberyard-employee-dm: Well these filesnames won't affect any functionality at all and they aren't being loaded as part of a `SettingsRegistryInterface::MergeSettingsFolder` call.
I'll generalized the names anyway.

Code/Framework/AzCore/AzCore/Serialization/Json/JsonImporter.cpp
lumberyard-employee-dm: Anything that is using "%s" is using the `.c_str()` method to retrieve a null-terminated string.
Anything format that is using "%.*s" is using either an AZ::IO::PathView or AZStd::string_view which may not be null-terminated and therefore uses the AZ_STRING_ARG/AZ_PATH_ARG method.

Code/Framework/AzCore/AzCore/Serialization/Json/JsonImporter.cpp
lumberyard-employee-dm: The FieldNameString is just used to store a converted array index to into a string value.

The Field length is only limited by the size of rapidjson.
That being said the FieldNameString can be a lot smaller, since only 19 characters are needed to represent index values up to 2^64.

Code/Framework/AzCore/Tests/Settings/SettingsRegistryTests.cpp
mbalfour-amzn: I wasn't concerned about functionality, just that it's generally better to avoid trademarked names.

Code/Framework/AzCore/AzCore/Settings/SettingsRegistry.h
lumberyard-employee-dm: The implications is that if any of the Settings Registry `MergeSettings*` functions don't explicitly set a failure return code as part of it's logic, the operation will be treated as successful.

Code/Framework/AzCore/AzCore/Serialization/Json/JsonImporter.cpp
mbalfour-amzn: Ah, thanks! I'd suggest that you add that information into the comment above it, I didn't realize it was holding an array index.
```
// Used to store the converted array index for an array field. Since it holds a converted 64-bit number, it will never be larger than 19 characters.
```

Code/Framework/AzCore/AzCore/Settings/SettingsRegistryImpl.h
lumberyard-employee-dm: The in situ data here is for Rapidjson to modify the data in place.
That can be seen a few lines [below](https://github.com/o3de/o3de/pull/16175/files/f85ede2297e4f7553df949c831635cec8248396e#diff-5a2efe42854b40635f68162ad700e69ce5ed23c3d1ec3c894098f6d1eafb0b95R1263-R1269)

This function is called in two places.
1. `MergeSettingsFileInternal`, where the JSON data read from a file on disk is moved into the parameter.
2. `MergeSettings`, where the JSON data is a string_view parameter and there needs to copied into a buffer in order to allow rapidjson to modify it in place

Code/Framework/AzCore/AzCore/Serialization/Json/JsonImporter.cpp
lumberyard-employee-dm: Yes, I'll add this to the [JsonSerializationResult.h](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Serialization/Json/JsonSerializationResult.h#L117) header since the primary purpose of the typedef is to construct a temporary string without dynamic allocations for use in the JsonIssueCallback function to pass in the error message.

Code/Framework/AzCore/AzCore/Serialization/Json/JsonImporter.cpp
lumberyard-employee-dm: I am updating it to calculate the exact number of digits needed to store the maximum value of the rapidjson size type rounded up to the nearest power of 2.

Code/Framework/AzCore/AzCore/Serialization/Json/JsonMerger.cpp
lumberyard-employee-dm: I'll do that.
The `rapidjson:Value::CopyFrom` doesn't actually copy const strings by default which causes dangling string references when the source `rapidjson::Value` being copied from goes out of scope.

Code/Framework/AzCore/AzCore/Serialization/Json/JsonImporter.cpp
lumberyard-employee-dm: Yeah, the logic is similar with some changes in the middle to make a rapidjson::Value that is an Array instead of an Object.

Code/Framework/AzCore/AzCore/Settings/SettingsRegistryImpl.h
mbalfour-amzn: OK, I guess I'd suggest the following:
1. Add a comment explaining that this intentionally takes in a string copy, not a reference, because the data inside it will get modified inside the method. To be extra-thorough, you could add a comment above the call to the Rapidjson insitu call to explain that it's modifying the settings string in place.
2. Maybe call the parameter settingsString instead of inSituData? It's just an implementation detail that it's getting changed, it doesn't need to be in the parameter name. Because the string is getting copied anyways, the caller doesn't need to know what's happening to it.

lumberyard-employee-dm: I change the name of the parameter.

Code/Framework/AzCore/AzCore/Settings/SettingsRegistryImpl.cpp
```diff
@@ -1463,58 +1296,64 @@ namespace AZ
...
-                    .AddMember(StringRef("Path"), Value(path, m_settings.GetAllocator()), m_settings.GetAllocator());
-
-                MergeSettingsResult result;
-                result.m_returnCode = MergeSettingsReturnCode::Failure;
-                result.m_operationMessages = AZStd::string::format(
-                    R"(Attempting to merge the settings registry file "%s" where the root element is a)"
-                    R"( non-JSON Object using the JSON MergePatch approach. The JSON MergePatch algorithm would therefore)"
+                mergeResult.Combine(MergeSettingsReturnCode::Failure);
+                mergeResult.m_operationMessages = AZStd::string::format(
+                    R"(Cannot merge settings registry JSON data to to root element field with a non-JSON Object)"
```
nick-l-o3de, line 1303: spelling, I think this should be "due to" instead of "to to" ?

nick-l-o3de (approved):
Nice cleanup in the settings registry init, too.

Code/Framework/AzCore/AzCore/Settings/SettingsRegistryImpl.cpp
lumberyard-employee-dm: Will fix.
```suggestion
                    R"(Cannot merge settings registry JSON data due to root element field with a non-JSON Object)"
```

PR #16179, opened 2023-06-16
Added functionality for implicit execution connections

mddvisu:
## What does this PR do?

This PR allows execution slots to be given the attribute m_createsImplicitConnections. This makes it so the slot automatically connects itself to other nodes' execution slots behind the scenes when their data slots are connected. I do this by having implicit connections only create their entity in the Script Canvas gem without creating a corresponding visual connection entity in the Graph Canvas gem.
I also added a few more attributes to nodeables to allow for nodeable input execution slots to be given the m_createsImplicitConnections attribute, and for nodeble execution slots to be hidden.
Finally, I made a few changes in the AbstractCodeModel to ensure that execution out slots with multiple connections will always parse nodes with implicit connections first. I did this so people can make nodes without execution slots that preform a certain calculation first so that nodes can always pull the data from it later on in execution (examples below).

## How was this PR tested?

Script canvas tests were run locally to ensure I wasn't breaking anything else. I also did manual testing by connecting, disconnecting, moving, undoing, redoing, and deleting the data connections of nodes with an implicit connection caused by that data connection to ensure that the implicit connection remained/was removed as expected. Saving and loading graphs was also tested extensively.
Functionality of the implicit connections was tested by making a simple increment nodeable node and giving it an execution input that creates implicit connections:

Screenshot 2023-06-15 165713
Screenshot 2023-06-15 165742

Increment nodeables can also be chained together:

Screenshot 2023-06-15 165856
Screenshot 2023-06-15 165919

### DIFF
--- a/Gems/ScriptCanvas/Code/Editor/Components/EditorGraph.cpp
+++ b/Gems/ScriptCanvas/Code/Editor/Components/EditorGraph.cpp
@@ -1145,6 +1145,11 @@ namespace ScriptCanvasEditor
         const ScriptCanvas::Node* sourceNode = FindNode(sourceEndpoint.GetNodeId());
         const ScriptCanvas::Node* targetNode = FindNode(targetEndpoint.GetNodeId());
 
+        if (!sourceNode || !targetNode)
+        {
+            return;
+        }
+
         // Retrieve the source node's execution out and target node's execution in
         const ScriptCanvas::Slot* sourceNodeExecutionSlot;
         const ScriptCanvas::Slot* targetNodeExecutionSlot;
--- a/Gems/ScriptCanvas/Code/Editor/Components/EditorGraph.cpp
+++ b/Gems/ScriptCanvas/Code/Editor/Components/EditorGraph.cpp
@@ -1074,7 +1074,7 @@ namespace ScriptCanvasEditor
             ? *AZStd::any_cast<AZ::EntityId>(connectionUserData)
             : AZ::EntityId();
 
-        if (ScriptCanvas::Connection* connectionComponent = AZ::EntityUtils::FindFirstDerivedComponent<ScriptCanvas::Connection>(scConnectionId))
+        if (auto connectionComponent = AZ::EntityUtils::FindFirstDerivedComponent<ScriptCanvas::Connection>(scConnectionId))
         {
             ScriptCanvas::Endpoint sourceEndpoint = connectionComponent->GetSourceEndpoint();
             ScriptCanvas::Endpoint targetEndpoint = connectionComponent->GetTargetEndpoint();
--- a/Gems/ScriptCanvas/Code/Editor/Components/EditorGraph.cpp
+++ b/Gems/ScriptCanvas/Code/Editor/Components/EditorGraph.cpp
@@ -1074,13 +1074,18 @@ namespace ScriptCanvasEditor
             ? *AZStd::any_cast<AZ::EntityId>(connectionUserData)
             : AZ::EntityId();
 
-        if (AZ::EntityUtils::FindFirstDerivedComponent<ScriptCanvas::Connection>(scConnectionId))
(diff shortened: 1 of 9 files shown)


Gems/ScriptCanvas/Code/Editor/Components/EditorGraph.cpp
```diff
@@ -1130,10 +1135,70 @@ namespace ScriptCanvasEditor
...
-
+        UpdateCorrospondingImplicitConnection(scSourceEndpoint, scTargetEndpoint);
+        
         return scConnected;
     }
 
+    void EditorGraph::UpdateCorrospondingImplicitConnection(const ScriptCanvas::Endpoint& sourceEndpoint, const ScriptCanvas::Endpoint& targetEndpoint)
+    {
+        const ScriptCanvas::Node* sourceNode = FindNode(sourceEndpoint.GetNodeId());
+        const ScriptCanvas::Node* targetNode = FindNode(targetEndpoint.GetNodeId());
```
spham-amzn, line 1146: How are we handling situations if either (or both) the source / target endpoint node id cannot be found? There should at least be an assert here if we expect that they always exist just in case something underlying changes

mddvisu: I added a check for this, thanks!

mddvisu: I am not sure if I should add an assert here, I am thinking that could possibly cause the editor make an assert when a user tries to make an invalid connection or doesn't drag a connection all the way to a node if someone were to reuse this function elsewhere. There aren't any other places in this file where I found an assert being made from a node not being found from its endpoints

spham-amzn: Asserts are usually placed where things are expected and could be indicative of a programming/logic error. For example, you would not assert on situations where the inputs are expected to be one value or another. If its expected that there are situations where either of these nodes could not be found and the caller expects this possibility, then asserts is not the right thing to do. Its your call since you are the author of this

Gems/ScriptCanvas/Code/Editor/Components/EditorGraph.cpp
```diff
@@ -1130,10 +1135,75 @@ namespace ScriptCanvasEditor
...
+            else
+            {
+                AZStd::vector<const ScriptCanvas::Slot*> sourceNodeDataSlots = sourceNode->GetAllSlotsByDescriptor(ScriptCanvas::SlotDescriptors::DataOut(), false);
+
+                int numDataConnectionsBetween = 0;
+
+                // Count the number of data connections between the nodes of the two endpoints
+                for (const ScriptCanvas::Slot* dataSlot : sourceNodeDataSlots)
+                {
+                    for (ScriptCanvas::EndpointResolved connectedNode : sourceNode->GetConnectedNodes(*dataSlot))
```
lsemp3d, line 1190: I recommend using a const reference here `const ScriptCanvas::EndpointResolved&` to avoid making a copy, it's a small performance improvement, but it's a good general practice

lsemp3d (approved):

Gems/ScriptCanvas/Code/Editor/Components/EditorGraph.cpp
```diff
@@ -1074,13 +1074,18 @@ namespace ScriptCanvasEditor
             ? *AZStd::any_cast<AZ::EntityId>(connectionUserData)
             : AZ::EntityId();
 
-        if (AZ::EntityUtils::FindFirstDerivedComponent<ScriptCanvas::Connection>(scConnectionId))
+        if (ScriptCanvas::Connection* connectionComponent = AZ::EntityUtils::FindFirstDerivedComponent<ScriptCanvas::Connection>(scConnectionId))
```
cgalvan, line 1077: Nit: Can specify type as "auto" for readability here since the type is already being passed to the templated method

cgalvan (approved):

Issue #16172, opened 2023-06-15
O3DE CLI Create-Repo Fails to Convert Folder to Repo


AMZN-Gene:
Converting an existing folder into a repo results in a python error without any suggestion for resolving.

**Steps to reproduce**
1. Create a folder with a gem (or some O3D object) inside of it
2. Attempt to turn this folder in an O3D repo for sharing with others by using the O3D CLI
   ```  .\scripts\o3de.bat create-repo -rp <folder_path> -ru https://github.com/AMZN-Gene/Open3dRepoTest -s "An awesome repo"```

**Expected behavior**
Create-Repo CLI catches the fact that the folder already exists and suggests user use the --force option
**better yet...**
Create-Repo CLI automatically allows existing folders to be converted into an O3D repo (essentially apply the --force option automatically). The normal use case would be that people turn their existing gems into a repo, and not starting a repo.

**Actual behavior**
Script errors

 _.\scripts\o3de.bat create-repo -rp D:\prj\Open3dRepoTest -ru https://github.com/AMZN-Gene/Open3dRepoTest -s "An awesome repo"_

```
Traceback (most recent call last):
  File "D:\prj\o3de\scripts\o3de.py", line 111, in <module>
    ret = known_args.func(known_args) if hasattr(known_args, 'func') else 1
  File "D:\prj\o3de\scripts\o3de\o3de\engine_template.py", line 2459, in _run_create_repo
    return create_repo(args.repo_path,
  File "D:\prj\o3de\scripts\o3de\o3de\engine_template.py", line 2343, in create_repo
    os.makedirs(repo_path, exist_ok=force)
  File "D:\prj\o3de\python\runtime\python-3.10.5-rev1-windows\python\lib\os.py", line 225, in makedirs
    mkdir(name, mode)
FileExistsError: [WinError 183] Cannot create a file when that file already exists: 'D:\\prj\\Open3dRepoTest'
```
**Found in Branch**
development

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
Mon Jun 12 2023
d5a8cc28f5f420cc70e6aa1c4314717eac1c7959

PR #16180, opened 2023-06-16
Repo Path Folder Exist Error Fix

VeryKaylin-player1:
## What does this PR do?
1. Checks if a folder already exists at the user provided path when user calls the create-repo cli command. Provides an error message telling user to use --force if they would want to override the current directory. 
2. If there is already a repo.json file in the directory, provided error message saying that a repo.json already exist and ask user if they want to edit the repo.json properties instead, if so, they should be using edit-repo-properties function.
3. fixes https://github.com/o3de/o3de/issues/16172 

## How was this PR tested?
I called create-repo function and pass in a path that already has a folder, and I now see an error message instead of a crash.  I choose --force to override that folder when prompted.

Here is the command I used when there was a already a folder existed at the path I provided: ./o3de.bat create-repo -rp "C:\o3de_Projects\bugtest"  -ru https://github.com/AMZN-Gene/Open3dRepoTest -s "An awesome repo", there would be a error message "o3de.engine_template: C:\o3de_Projects\bugtest already exists. Use the --force command to overwrite the existing directory." 
Then, calling the same function again  ./o3de.bat create-repo -rp "C:\o3de_Projects\bugtest"  -ru https://github.com/AMZN-Gene/Open3dRepoTest -s "An awesome repo" -f will override the current directory and create the repo.json file.

### DIFF
--- a/scripts/o3de/o3de/engine_template.py
+++ b/scripts/o3de/o3de/engine_template.py
@@ -2331,20 +2331,16 @@ def create_repo(repo_path: pathlib.Path,
         logger.error('Remote repository URI cannot be empty.')
         return 1
     
-    # check if path is empty and
-    if not force and repo_path.is_dir():
-        logger.error(f'{repo_path} already exists. Use the --force command to overwrite the existing directory.')
-        return 1
     repo_json_path = repo_path / 'repo.json'
     # if a repo.json file already exist in directory - can only have one repo.json file per project
-    if repo_json_path.is_file():
+    if not force and repo_json_path.is_file():
         logger.error(f'{repo_json_path} already exists. If you want to edit the repo.json properties, use edit-repo-properties command.'
-                     'If you want to create a new remote repo template, create a new directory.')
+                     'Use --force if you want to override the current repo.json file')
         return 1
     elif repo_path.is_file():
         logger.error(f'{repo_path} already exists and is a file instead of a directory.')
         return 1
-    else:
+    elif not repo_path.is_dir():
         os.makedirs(repo_path, exist_ok=force)
 
     if not repo_name:
--- a/scripts/o3de/o3de/engine_template.py
+++ b/scripts/o3de/o3de/engine_template.py
@@ -2341,7 +2341,7 @@ def create_repo(repo_path: pathlib.Path,
         logger.error(f'{repo_path} already exists and is a file instead of a directory.')
         return 1
     elif not repo_path.is_dir():
-        os.makedirs(repo_path, exist_ok=force)
+        os.makedirs(repo_path, exist_ok=True)
 
     if not repo_name:
         # repo name default is the last component of repo_path
--- a/scripts/o3de/o3de/engine_template.py
+++ b/scripts/o3de/o3de/engine_template.py
@@ -2306,7 +2306,7 @@ def create_repo(repo_path: pathlib.Path,
                 origin_url: str = None,
                 summary: str = None,
                 additional_info: str = None,
-                force: bool = False,
+                force: bool = None,
                 replace: list = None,
                 no_register: bool = False) -> int:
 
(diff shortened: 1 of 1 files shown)


scripts/o3de/o3de/engine_template.py
```diff
@@ -2331,10 +2331,15 @@ def create_repo(repo_path: pathlib.Path,
         logger.error('Remote repository URI cannot be empty.')
         return 1
     
-    # check if a repo.json file already exist in directory - can only have one repo.json file per project
+    # check if path is empty and
+    if not force and repo_path.is_dir():
+        logger.error(f'{repo_path} already exists. Use the --force command to overwrite the existing directory.')
+        return 1
```
AMZN-alexpete, line 2337: It's OK to write to an existing folder, that's how you would be able to write the `repo.json` file to an existing Gem folder for example.  You just don't want to overwrite any existing files.

```suggestion
```

scripts/o3de/o3de/engine_template.py
```diff
@@ -2331,10 +2331,15 @@ def create_repo(repo_path: pathlib.Path,
...
     repo_json_path = repo_path / 'repo.json'
-    if not force and repo_json_path.is_file():
-        logger.error(f'{repo_json_path} already exists.  Use the --force to overwrite the existing file.')
+    # if a repo.json file already exist in directory - can only have one repo.json file per project
+    if repo_json_path.is_file():
+        logger.error(f'{repo_json_path} already exists. If you want to edit the repo.json properties, use edit-repo-properties command.'
+                     'If you want to create a new remote repo template, create a new directory.')
```
AMZN-alexpete, line 2342: It's possible the user may want to re-create the file, and they shouldn't have to create a new directory to use this command.
```suggestion
        logger.error(f'{repo_json_path} already exists. If you want to re-create the existing file, use the --force option.  If you want to edit repo.json properties, use the edit-repo-properties command.')
```

AMZN-alexpete (commented):
What file is engine_template.py complaining about here https://github.com/o3de/o3de/issues/16172?

VeryKaylin-player1:
> What file is engine_template.py complaining about here #16172?

When using a path that already has a folder there when calling create-repo, the system will through an error if we don't check if the directory exists

AMZN-alexpete (approved):

lumberyard-employee-dm (approved):

lumberyard-employee-dm:
> > What file is engine_template.py complaining about here #16172?
> 
> When using a path that already has a folder there when calling create-repo, the system will through an error if we don't check if the directory exists

That makes sense from the engine-template code perspective as it doesn't want to overwrite a user project or gem with template files without the user being explicit about it.

scripts/o3de/o3de/engine_template.py
```diff
@@ -2331,15 +2331,16 @@ def create_repo(repo_path: pathlib.Path,
...
     if not force and repo_json_path.is_file():
-        logger.error(f'{repo_json_path} already exists.  Use the --force to overwrite the existing file.')
+        logger.error(f'{repo_json_path} already exists. If you want to edit the repo.json properties, use edit-repo-properties command.'
+                     'Use --force if you want to override the current repo.json file')
         return 1
     elif repo_path.is_file():
         logger.error(f'{repo_path} already exists and is a file instead of a directory.')
         return 1
-    else:
+    elif not repo_path.is_dir():
```
spham-amzn, line 2343: Why is this check needed? Also, if this check is done here, then the following call
`os.makedirs(repo_path, exist_ok=force)` won't need the `exist_ok` argument, since it will never exist.

VeryKaylin-player1: Hi Steve, the check is needed when user wants to create a folder at a path that already has a existed folder. If no check is performed then there will be a windows error with no appropriate error message: FileExistsError: [WinError 183] Cannot create a file when that file already exists: 'C:\\o3de_Projects\\o3deRemoteRepo' and exist_ok = force to allow user to override that folder.

spham-amzn: To me, the original logic looks fine without this check, just as long as you do `exist_ok=True` as an argument to `os.makedirs` instead. `force` doesn't look like the appropriate value to set here because it is not actually forcing anything. If true, it will skip creating a folder if it already exists. If false, it will say 'you are creating a folder that already exists`. (see https://docs.python.org/3/library/os.html#os.makedirs )  The `force` argument should only apply to overriding any existing `repo.json` file.

VeryKaylin-player1: Thank you for the suggestion, this way would be more clear.

spham-amzn (approved):

Issue #16181, opened 2023-06-16
Edit-Repo-Properties Archive Removes Download Source Uri when rezipped 


VeryKaylin-player1:
**Describe the bug**
When the download source uri is changed and user re zipped the archive, the download source uri field will automatically be revert to the fallback

**Steps to reproduce**
1. update repo.json download_source_uri field to the new uri 
2. use ./o3de.bat edit-repo-properties -rp "C:\o3de_Projects\o3deRemoteRepo\repo.json" -ag "C:\o3de_Projects\o3deRemoteRepo\Gems\KaylinGem\" --release-archive-path "C:\o3de_Projects" -f   to rezip the repo.json file 

Then the download_source_uri field would change to fallback

**Expected behavior**
The newly zipped repo.json should get zipped with the updated download_source_uri field and only change the version but not override the hosting uri

**Actual behavior**
When we rezipped the repo.json file, it will update the download_source_uri to the fallback 

**Found in Branch**
development

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
2a7cbf1c9939e269cd95f24ffc7a71b11a238f85   6/16/2023

AMZN-alexpete:
Without the `--download-prefix` argument, the current logic is unable to create a valid download url which is wny you are seeing `None` in the download_source_uri path.

PR #16182, opened 2023-06-16
Add Check To Require Download-prefix To Be Provided With Usage Suggestion

VeryKaylin-player1:
## What does this PR do?

If user use the --release-archive-path command without --download-prefix and error message would require user to pass in a --download-prefix along with an example of a github url

fix https://github.com/o3de/o3de/issues/16181

## How was this PR tested?
I used this command: ./o3de.bat edit-repo-properties -rp "C:\o3de_Projects\o3deRemoteRepo\repo.json" -ag "C:\o3de_Projects\o3deRemoteRepo\Gems\TestGem\" --release-archive-path "C:\o3de_Projects"
an error message would be: [ERROR] o3de.repo_properties: The -download-prefix must be provided. A url prefix for a file attached to a Github release might look like this: -dp https://github.com/o3de/o3de-extras/releases/download/2305.0

### DIFF
--- a/scripts/o3de/o3de/repo_properties.py
+++ b/scripts/o3de/o3de/repo_properties.py
@@ -91,7 +91,7 @@ def create_remote_object_archive(src_data_path: pathlib.Path,
         ValueError: If the json_data_path is not a dict.
     """
     if download_prefix is None:
-        logger.error('The -download-prefix must be provided. A url prefix for a file attached to a Github release might look like this:'
+        logger.error('The --download-prefix argument must be provided. A url prefix for a file attached to a Github release might look like this:'
                      '-dp https://github.com/o3de/o3de-extras/releases/download/2305.0')
         return {}
     zip_path = releases_path / archive_filename
--- a/scripts/o3de/o3de/repo_properties.py
+++ b/scripts/o3de/o3de/repo_properties.py
@@ -93,6 +93,7 @@ def create_remote_object_archive(src_data_path: pathlib.Path,
     if download_prefix is None:
         logger.error('The -download-prefix must be provided. A url prefix for a file attached to a Github release might look like this:'
                      '-dp https://github.com/o3de/o3de-extras/releases/download/2305.0')
+        return {}
     zip_path = releases_path / archive_filename
     # check if a object.zip folder already exist in the path - ask user if they want to overwrite the current zip
     if not force and zip_path.exists():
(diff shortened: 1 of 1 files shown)


scripts/o3de/o3de/repo_properties.py
```diff
@@ -90,7 +90,9 @@ def create_remote_object_archive(src_data_path: pathlib.Path,
         FileNotFoundError: If the json_data_path does not exist.
         ValueError: If the json_data_path is not a dict.
     """
-
+    if download_prefix is None:
+        logger.error('The -download-prefix must be provided. A url prefix for a file attached to a Github release might look like this:'
+                     '-dp https://github.com/o3de/o3de-extras/releases/download/2305.0')
```
spham-amzn, line 95: This check only logs the error but the flow will continue. If you want to follow the existing patterns in this function, then add a 'return {}'
Also, this doesn't seem to be the appropriate place to raise an error related to arg parsing. This should be done by the arg parser object itself, and the `download-prefix` argument in the argparser should be marked required.

VeryKaylin-player1: Hi steve! yes let me add the return {},
for the arg parse, If I use the required = '--create-archive-path' in sys.argv, the logic would be good but no appropriate error message can be displayed. The argv library would just give message to tell user that  --download-prefix is required and that's it. Unless user use --help to show detailed message.

spham-amzn: Another option would be to not make it required but default it to none, and then in the same function that the argparse is evaluated, do this check there and display the appropriate error message. The reason is that this error message describes a specific argument that is defined in the argparse (`-download-prefix'), but doesn't really exist inside the context of this function itself. If another function called this function but didn't use argparse, then it would out of context.

scripts/o3de/o3de/repo_properties.py
```diff
@@ -90,7 +90,10 @@ def create_remote_object_archive(src_data_path: pathlib.Path,
         FileNotFoundError: If the json_data_path does not exist.
         ValueError: If the json_data_path is not a dict.
     """
-
+    if download_prefix is None:
+        logger.error('The -download-prefix must be provided. A url prefix for a file attached to a Github release might look like this:'
```
AMZN-alexpete, line 94: The argument is missing a '-'
```suggestion
        logger.error('The --download-prefix argument must be provided. A url prefix for a file attached to a Github release might look like this:'
```

AMZN-alexpete (approved):

AMZN-Gene (approved):

PR #16186, opened 2023-06-19
[LINUX] Fix "white" background of textfield

yaakuro:
## What does this PR do?

In the current version, there is an issue where the background color and font background color appear to be identical, making the font unreadable in the GemCatalog. This problem is particularly noticeable on my system. Additionally, the StyleSheet is already set globally for the Editor, so there is no need to set it manually in this specific instance.

## How was this PR tested?

To test this pull request, the following steps were taken:

1. Started the Editor with the project on Linux.
2. Opened the Project Settings and accessed the Gems configuration.
3. Observed that the font was not visible in the search field and with this PR the font gets the correct color.


mbalfour-amzn (approved):

spham-amzn (approved):

spham-amzn:
This change isn't Linux-specific (or any platform specific). Is this happening on windows?

AMZN-alexpete (approved):
Thanks @yaakuro !

lumberyard-employee-dm (approved):

PR #16187, opened 2023-06-19
Highlight search terms in asset browser views.

### LINKED ISSUE #15966: Asset Browser: Search results not clear - highlight text
As a user, when I use the search in the Asset Browser, looking for a specific asset, it's hard to tell why an item was displayed. This is because the name could be somewhere in the title OR because the folder name included those details. But currently we don't highlight the results, so it's unclear why you are seeing some results.

Please see the attached video to describe the problem as well as see an example of the solution in ScriptCanvas.  

https://github.com/o3de/o3de/assets/82551958/9cb92a82-e690-4cbe-8acf-93e1f69c175b

**Although, highlighting the text will fix the main issue here.** A secondary issue will need to be considered.  Previously, we supported a search panel that would highlight the search term if the name came up as part of a directory name. To continue to support this, we could do one of the following. 

A) Force everyone to use the same layout. Let's say table view for search. Where you can highlight all the specific details. However, if someone was looking for thumbnails in a search view, then this becomes more difficult.

B) Make search for directory work in all layouts, including thumbnail view. Example below. (Small error I just noticed with this image. It will most likely show the results will be folder in the "search by folder" section and not actual files.) Actual files would appear in search by name. 

C) We don't support directory name search.

sphrose:
## What does this PR do?

This PR resolves #15966 by highlighting the selected search term in the AssetBrowser Tree, Thumbnail and Table views.

## How was this PR tested?

Manual testing, turned of the disabling of search in the table view to check that for future use.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Views/AssetBrowserTableView.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Views/AssetBrowserTableView.cpp
@@ -19,6 +19,7 @@
 #include <AzToolsFramework/API/EditorAssetSystemAPI.h>
 #include <AzToolsFramework/Editor/ActionManagerIdentifiers/EditorContextIdentifiers.h>
 #include <AzToolsFramework/Editor/ActionManagerUtils.h>
+#include <AzToolsFramework/Editor/RichTextHighlighter.h>
 #include <AzCore/Utils/Utils.h>
 
 #include <AzQtComponents/Components/Widgets/AssetFolderTableView.h>
@@ -509,6 +510,11 @@ namespace AzToolsFramework
             m_assetFilterModel->SetFilter(FilterConstType(filterCopy));
         }
 
+        void AssetBrowserTableView::SetSearchString(const QString& searchString)
+        {
+            m_expandedTableViewProxyModel->SetSearchString(searchString);
+        }
+
         ExpandedTableViewDelegate::ExpandedTableViewDelegate(QWidget* parent)
             : QStyledItemDelegate(parent)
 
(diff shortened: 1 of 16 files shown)


dtamkin1 (approved):

AutomatedTesting/Gem/PythonTests/editor/EditorScripts/AssetBrowser_SearchFiltering.py
```diff
@@ -122,7 +122,8 @@ def verify_files_appeared(model, allowed_asset_extensions, parent_index=QtCore.Q
         # 4) Click the "X" in the search bar.
         clear_search = asset_browser.findChild(QtWidgets.QToolButton, "ClearToolButton")
         clear_search.click()
-
+        
+        print("test1")
```
cgalvan, line 126: Remove this debug statement and below as well

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Views/AssetBrowserTableView.cpp
```diff
@@ -542,10 +547,15 @@ namespace AzToolsFramework
...
                 textRect.setX(textRect.x() + 4);
-                style->drawItemText(
-                    painter, options.rect, Qt::AlignLeft | Qt::AlignVCenter, options.palette, options.state & QStyle::State_Enabled, text.toString());
+
+                QStyleOptionViewItem optionV4{ option };
+                initStyleOption(&optionV4, index);
+                optionV4.state &= ~(QStyle::State_HasFocus | QStyle::State_Selected);
+
+                RichTextHighlighter::PaintHighlightedRichText(text.toString(), painter, optionV4, textRect);
+
```
cgalvan, line 558: Nit: Can remove extra newline here

cgalvan (approved):

AutomatedTesting/Gem/PythonTests/editor/EditorScripts/AssetBrowser_SearchFiltering.py
sphrose: Removed

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Views/AssetBrowserTableView.cpp
sphrose: Removed.

PR #16193, opened 2023-06-20
fix to use labels for row identifiers when saving expander state in the DPE

### LINKED ISSUE #16087: DPE Inspector: Complete expansion state fix
**Describe the bug**
There's an issue with restoring expansion state of fields where it's not always restored with chained adapters. Work on a fix had been started previously and is in a draft PR here: https://github.com/o3de/o3de/pull/15038

**Steps to reproduce**

3. Create a new entity.
4. Add several components to the entity.
5. Toggle several expansion buttons.
6. Unselect the Entity and then re-select it.

**Expected behavior**
Expansion state of fields should be restored exactly as before.

**Actual behavior**
Expansion state of some fields might be back to the default.

alexmontAmazon:
## What does this PR do?
Refactor DocumentPropertyEditorSettings as ExpanderSettings that can be individually tooled for each adapter. Different adapters may want to save their expanded settings differently -- notably, the ReflectionAdapter cannot rely on DOM paths to reliably address the same logical elements in different Entities, since each Entity's components can have fields that are shown or hidden based on the component's state. This PR allows an adapter to specify the path nomenclature/storage, or even to disable expander saving altogether.

some components had to be moved from the AzToolsFramework to the lower level AzFramework to allow code at the Adapter level to access them.

Before all DPEs would attempt to save the expanded state of each component's elements based on their DOM paths. This PR makes that flexible based on virtual functions provided by the individual adapters. It also implements the new LabeledRowDPEExpanderSettings class, that allows the Entity Inspector to specify its paths based on the first label (the display name) of each component element, instead of the direct DOM path.

This PR fixes #16087 

## How was this PR tested?
Locally in the Editor project


Code/Framework/AzToolsFramework/AzToolsFramework/aztoolsframework_files.cmake
```diff
@@ -407,8 +407,6 @@ set(FILES
     UI/DocumentPropertyEditor/DPEComponentAdapter.cpp
     UI/DocumentPropertyEditor/DocumentPropertyEditor.cpp
     UI/DocumentPropertyEditor/DocumentPropertyEditor.h
-    UI/DocumentPropertyEditor/DocumentPropertyEditorSettings.cpp
-    UI/DocumentPropertyEditor/DocumentPropertyEditorSettings.h
```
lumberyard-employee-dm, line 411: Shouldn't the `DocumentPropertyEditorSettings.h` also be removed as well?

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ExpanderSettings.cpp
```diff
@@ -0,0 +1,201 @@
...
+            {
+                auto& currChild = *arrayIter;
+                if (arrayIter->GetNodeName() == AZ::Dpe::GetNodeName<AZ::Dpe::Nodes::Label>())
+                {
+                    newPath.Append(AZ::Dpe::Nodes::Label::Value.ExtractFromDomNode(currChild).value_or(""));
+                }
+            }
+        }
+        return newPath;
+    }
```
lumberyard-employee-dm, line 199: A DomPath can return a string via it's ToString function.
The use of AZ::IO::Path seems unnecessary here

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ExpanderSettings.cpp
```diff
@@ -0,0 +1,201 @@
...
+        const auto& fullContents = m_adapter->GetContents();
+        AZ::Dom::Path subPath;
+        for (auto pathComponent : rowPath)
+        {
+            subPath.Push(pathComponent);
+            auto rowValue = fullContents[subPath];
+            for (auto arrayIter = rowValue.ArrayBegin(), endIter = rowValue.ArrayEnd(); arrayIter != endIter; ++arrayIter)
+            {
+                auto& currChild = *arrayIter;
+                if (arrayIter->GetNodeName() == AZ::Dpe::GetNodeName<AZ::Dpe::Nodes::Label>())
```
lumberyard-employee-dm, line 192: currChild child should be used in the if statement over arrayIter if it is going to be assigned here.
```suggestion
                auto& currChild = *arrayIter;
                if (currChild .GetNodeName() == AZ::Dpe::GetNodeName<AZ::Dpe::Nodes::Label>())
```

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ExpanderSettings.cpp
```diff
@@ -0,0 +1,201 @@
...
+        m_settingsRegistrar.StoreObjectSettings(m_fullSettingsRegistryPath, this);
+
+        // Configuration that the dumper util will use when collecting our data from the SettingsRegistry
+        AZ::SettingsRegistryMergeUtils::DumperSettings dumperSettings;
+        dumperSettings.m_prettifyOutput = true;
+        dumperSettings.m_includeFilter = [pathFilter = m_settingsRegistryBasePath](AZStd::string_view path)
+        {
+            return AZ::SettingsRegistryMergeUtils::IsPathAncestorDescendantOrEqual(pathFilter, path);
+        };
+        dumperSettings.m_jsonPointerPrefix = m_settingsRegistryBasePath;
```
lumberyard-employee-dm, line 85: The include filter isn't needed as the the key parameter passed to the SaveSettingsToFile function below will only save the settings that are at or below the `m_settingsRegistryBasePath` anchor field.

That is detailed in the Output the Settings Registry documentation: https://www.docs.o3de.org/docs/user-guide/settings/output-settings-registry/#storing-editor-preferences-using-a-key-and-an-anchor

Code/Framework/AzToolsFramework/AzToolsFramework/aztoolsframework_files.cmake
alexmontAmazon: I'm confused, it is, right? It's in the line right where you added this comment

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ExpanderSettings.cpp
alexmontAmazon: LabeledRowDPEExpanderSettings uses the value of the first row label, not the path, so the ToString wouldn't work here. That is, however, exactly what the base class ExpanderSettings::GetStringPathForDomPath does.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ExpanderSettings.h
```diff
@@ -10,30 +10,36 @@
...
     //! This serializable class stores and loads the DocumentPropertyEditor settings such as tree node expansion state.
-    class DocumentPropertyEditorSettings
+    class ExpanderSettings
     {
     public:
-        AZ_RTTI(DocumentPropertyEditorSettings, "{7DECB0A1-A1AB-41B2-B31F-E52D3C3014A6}");
+        AZ_RTTI(ExpanderSettings, "{7DECB0A1-A1AB-41B2-B31F-E52D3C3014A6}");
 
-        using ExpanderStateMap = AZStd::unordered_map<AZStd::string, bool>;
+        using PathType = AZ::IO::BasicPath<AZStd::string>;
```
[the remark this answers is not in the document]
alexmontAmazon: fixed, thanks!

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ExpanderSettings.cpp
alexmontAmazon: fixed, thanks!

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ExpanderSettings.cpp
alexmontAmazon: fixed, thanks!

AMZN-daimini (approved):

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ExpanderSettings.cpp
```diff
@@ -0,0 +1,198 @@
...
+        // Configuration that the dumper util will use when collecting our data from the SettingsRegistry
+        AZ::SettingsRegistryMergeUtils::DumperSettings dumperSettings;
+        dumperSettings.m_prettifyOutput = true;
+        dumperSettings.m_jsonPointerPrefix = m_settingsRegistryBasePath;
+
+        auto outcome = m_settingsRegistrar.SaveSettingsToFile(m_settingsFilepath, dumperSettings, m_settingsRegistryBasePath);
+        if (!outcome.IsSuccess())
+        {
+            AZ_Warning("DocumentPropertyEditorSettings", false, outcome.GetError().c_str());
+        }
```
cgalvan, line 87: Could also do:

```suggestion
        AZ_Warning("DocumentPropertyEditorSettings", outcome.IsSuccess(), outcome.GetError().c_str());
```

cgalvan (approved):

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/ExpanderSettings.cpp
```diff
@@ -0,0 +1,198 @@
...
+        SaveAndCleanExpanderStates();
+    }
+
+    void ExpanderSettings::Reflect(AZ::ReflectContext* context)
+    {
+        AZ::SerializeContext* serializeContext = azrtti_cast<AZ::SerializeContext*>(context);
+        if (serializeContext)
+        {
+            serializeContext->Class<ExpanderSettings>()->Version(0)->Field(
+                "ExpandedElements", &ExpanderSettings::m_expandedElementStates);
```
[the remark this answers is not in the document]
alexmontAmazon: yeah, that was our auto-format. I never know what to do in those situations...

PR #16197, opened 2023-06-20
Fix the first-time Editor crash bugs

mbalfour-amzn:
## What does this PR do?

There are a lot of miscellaneous bug fixes lumped together in this PR that together fix the set of problems that were causing the asserts and crashes when the Editor was run for the first time with no assets processed.

The crash bug that was reported was that the DiffuseProbeGridFeatureProcessor would assert and crash when loading the DiffuseProbeSphere model on the first run of the Editor if assets hadn't been processed yet. The root cause of the crash was that the model would sometimes process and load before the DefaultVertexBufferPool.resourcepool had been processed, which caused problems when trying to use the model's data.

In investigating the problem, I discovered a number of other related problems with Atom assets where they would try and fail to load on startup if they didn't exist yet, and there were no recovery mechanisms to ensure that they would load at a later point. This led to a condition where the Editor was "mostly working", but various rendering features wouldn't quite be working correctly, and they didn't self-correct until the next relaunch of the Editor.

The set of fixes below improve this state but doesn't fix it completely. In particular, I learned the following:
* Any call to GetAsset<> before an asset has been processed will fail to find the asset in the catalog, and so the asset will fail to load and won't self-correct, UNLESS the asset type has support for a default "missing" asset and supports auto-reloads. (The changes below add "missing" support to azmodel)
* "Missing" asset support _only_ works if the swapped-out asset doesn't have dependencies. The swapped-out asset is also expected to exist on disk, it takes a special effort to create it in-memory.
* Any place where the engine synchronously loads an asset and doesn't listen for reloads or catalog notifications is a likely failure point on first-time startups. 
* Atom model reloads occur in the MeshFeatureProcessor, not the Asset Manager system, so systems like the DiffuseProbeGrid that use the Asset Manager to load models don't automatically get reload support.
* CompileSyncAsset() calls can stall the engine for several minutes if the compiled asset relies on intermediate assets (ex: models relying on materials) because those assets get bumped to the back of the compile line while waiting for the intermediate assets to show up.

With the fixes in this PR, the actual crash is fixed, Atom Models add support for "missing" assets that can be auto-reloaded to correct assets once they're processed, and a number of Atom engine asset loads are switched to the "LoadCriticalAsset" pathway that tries to synchronously compile the asset before loading it as a way to sidestep the issue of GetAsset<> calls failing if the asset isn't compiled yet. 

Hard-coded model asset loads still need to be examined and reworked to ensure that they support auto-reloading, other hard-coded asset loads still need to be examined to make sure they recover gracefully if the asset doesn't exist, and material assets would benefit from "missing" asset support to help all the places that have hard-coded material asset loads.

Larger fixes in the PR:
* THE ACTUAL PRIMARY BUGFIX: Added a job dependency to model building on DefaultVertexBufferPool. Without this, Atom models would try loading as soon as the AP was done and then crash when trying to use their buffers.
* Fixed the compile lookup by UUID in AssetProcessor to correctly handle "not found". Previously, it was returning those compile results as successful, now they're returned as failures. When they were returned as success, the Editor would go on to try (and fail) to load the assets, because they still didn't exist yet.
* Consolidated a lot of redundant Atom loading code that would compile and synchronously load an asset to all go through LoadCriticalAsset<> and RPI::LoadStreamingTexture (which now also goes through LoadCriticalAsset<>). This makes the calls easier to find and makes it easier to improve the compile & sync load function in one place instead of many places.
* Changed many of the Atom hard-coded engine assets to go through LoadCriticalAsset<> because the code loading them wasn't set up to handle asynchronously loading or reloading those assets and would fail in unrecoverable ways if the asset didn't exist yet.
* Fixed the MeshReloader code to correctly handle missing assets on deletion and loading.
* Added a default missing asset (an invisible cube) to ModelAsset so that models can now load successfully when missing and get reloaded as the assets become available.
* Changed the DiffuseProbeFeatureProcessor to gracefully handle the visualization model load without requiring a blocking synchronous asset compile first.

Smaller fixes:
* Fixed an invalid DB query in the AssetProcessor. This didn't have much of an effect because the LegacyGuids don't seem to be used much or at all.
* From @lumberyard-employee-dm - changed the location of the CriticalAssetsCompiled signal to after the Asset Catalog load, so that anything reacting to the signal that tries to load an asset will have an ability to succeed in the asset load. Otherwise, the asset loads silently fail and never try again during that Editor run.
* Made the warning in AssetCatalog a little less spammy. To make it even less spammy, the errors themselves should get corrected (i.e. more asset types need to be created and registered).
* Standardized the compile request message logging in AssetProcessor to all use AZ_Info. Previously, some were Trace and some were TracePrintf (Info), causing the logging to be incomplete.

## How was this PR tested?

Ran on both PC and Linux, verified that both the AutomatedTesting and MPS projects could be started and a level opened without any crashes or asserts.

Performed lots of tests of deleting, copying, and modifying models to ensure that they swapped in and out correctly as they went from "missing" to "existing" and back to "missing".


Gems/Atom/Feature/Common/Code/Source/OcclusionCullingPlane/OcclusionCullingPlane.cpp
mbalfour-amzn, line 1: Switched to LoadCriticalAsset because these are expected to exist at this point and the loads won't recover if the assets don't exist. Note - even with this change, the model compile can still fail sometimes (but less often) due to missing material dependencies, causing this model to fail to load. It also won't use the "missing asset" fallback because it's loading by path instead of id, causing it to fail to find an id to try loading. This would need to change to use an async load with an id to be more robust.

Gems/Atom/Feature/Common/Code/Source/ScreenSpace/DeferredFogSettings.cpp
mbalfour-amzn, line 1: Consolidated into the compile-and-sync-load of LoadStreamingTexture(). This adds the compile asset step because the asset lookup would otherwise fail and not recover.

Gems/Atom/RPI/Code/Include/Atom/RPI.Reflect/Asset/AssetUtils.h
mbalfour-amzn, line 1: Moved compilation into a helper function for easier debugging and less code repetition in the templated functions.

Gems/Atom/RPI/Code/Include/Atom/RPI.Reflect/Asset/AssetUtils.inl
mbalfour-amzn, line 1: Added compiling to the async loader because the one place that uses the async loader requires the asset to exist, and asset id lookups by path will fail. Eventually this should get changed to use id loading instead of path loading so that it can use the missing asset fallbacks without requiring sync compiling.

Gems/Atom/RPI/Code/Source/RPI.Builders/Model/ModelAssetBuilderComponent.cpp
mbalfour-amzn, line 1: Added the job dependency on DefaultVertexBufferPool to ensure that models aren't marked as complete until that exists as well. This solves the crash bug.

Gems/Atom/RPI/Code/Source/RPI.Public/RPIUtils.cpp
mbalfour-amzn, line 1: Logic is an exact duplicate of LoadCriticalAsset<>, so consolidated.

Gems/Atom/RPI/Code/Source/RPI.Reflect/Image/StreamingImageAssetHandler.cpp
mbalfour-amzn, line 1: Stream length can be 0 if someone calls FindOrCreateAsset() with a streaming image with the flag set to auto-add a catalog entry, and then tries to load that asset. This prevents asserts / crashes in that scenario.

Also fixes a variable that would be uninitialized if the AssetSystemRequestBus isn't hooked up.

Gems/Atom/RPI/Code/Source/RPI.Reflect/Model/ModelAsset.cpp
mbalfour-amzn, line 1: Adds support for creating an invisible box model as a default missing asset. It's invisible because the engine doesn't seem to have any guaranteed-to-exist default materials that i can assign to it.

Gems/Atom/RPI/Code/Source/RPI.Reflect/Model/ModelAssetHelpers.cpp
mbalfour-amzn, line 1: Split out the method to create a unit cube model into a separate helper. It looks like there are a couple places in the code that might be able to get refactored to use this as desired (unit tests, somewhere else that was generating a cube for visualizations, etc).

Gems/DiffuseProbeGrid/Code/Source/Render/DiffuseProbeGridFeatureProcessor.cpp
mbalfour-amzn, line 1: This is changed to successfully load and use the "missing asset" cube if the sphere isn't compiled yet. It won't replace it with the sphere though, it needs a larger amount of refactoring to correctly handle the model reload.

Code/Tools/AssetProcessor/native/resourcecompiler/rccontroller.cpp
```diff
@@ -383,12 +383,22 @@ namespace AssetProcessor
...
-            AZ_TracePrintf(AssetProcessor::DebugChannel, "OnRequestCompileGroup:  %s - %s requested, but no matching source assets found.\n", searchTerm.toUtf8().constData(), assetId.ToString<AZStd::string>().c_str());
+            Q_EMIT CompileGroupCreated(groupID, AzFramework::AssetSystem::AssetStatus_Unknown);
         }
         else
         {
+            AZ_Info(
+                AssetProcessor::DebugChannel,
+                "GetAssetStatus: OnRequestCompileGroup:  %s - %s requested and queued, found %d results.\n",
+                searchTerm.toUtf8().constData(),
+                assetId.ToString<AZStd::string>().c_str(), results.size());
```
lumberyard-employee-dm, line 400: FYI, the [AssetId::ToFixedString](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Asset/AssetCommon.h#L87-L89) function can be used to avoid a heap allocation
```suggestion
                assetId.ToFixedString().c_str(), results.size());
```

Gems/Atom/Feature/Common/Code/Source/OcclusionCullingPlane/OcclusionCullingPlane.cpp
lumberyard-employee-dm: Since the file path is known, we can generated a Uuid from the file path by using a lowercase path following the Algorithm in [CreateSafeSourceUUIDFromName](https://github.com/o3de/o3de/blob/development/Code/Tools/AssetProcessor/native/utilities/assetUtils.cpp#L866-L875).

Gems/Atom/RPI/Code/Source/RPI.Builders/Model/ModelAssetBuilderComponent.cpp
```diff
@@ -2420,5 +2420,59 @@ namespace AZ
...
+            // DefaultVertexBufferPool has been processed.
+
+            AssetBuilderSDK::SourceFileDependency defaultVertexBufferPoolSource;
+            defaultVertexBufferPoolSource.m_sourceFileDependencyPath = ModelAssetBuilderComponent::s_defaultVertexBufferPoolSourcePath;
+
+            AssetBuilderSDK::JobDependency jobDependency;
+            jobDependency.m_jobKey = "Model Asset Builder (Default Vertex Buffer Pool)";
+            jobDependency.m_sourceFile = defaultVertexBufferPoolSource;
+            jobDependency.m_platformIdentifier = platformIdentifier;
+            jobDependency.m_productSubIds.push_back(0);
```
lumberyard-employee-dm, line 2471: Can you make this SubID a constant as well?
Perhaps
`constexpr AZ::u32 ModelAssetDefaultSubID = 0;`

Gems/Atom/RPI/Code/Source/RPI.Reflect/Asset/AssetUtils.cpp
```diff
@@ -33,19 +33,46 @@ namespace AZ
...
+                        reporting,
+                        AZStd::string::format(
+                            "Could not compile asset '%s', status = %u.", assetFilePath.c_str(), static_cast<uint32_t>(status))
+                            .c_str());
+
+                    return false;
+                }
+
+                return true;
+            }
```
lumberyard-employee-dm, line 54: Should this try to lookup the AssetID first using the assetFilePath and if found use `CompileAssetSyncById` in that case?

Gems/Atom/RPI/Code/Source/RPI.Reflect/Model/ModelAssetHelpers.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+            // 3 values per position, 4 positions per face, 6 faces
+            const AZStd::array<float, 3 * 4 * 6> bitangents = {
+                +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, // front (+X)
+                +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, // right (+Y)
+                -1.0f, +0.0f, +0.0f, -1.0f, +0.0f, +0.0f, -1.0f, +0.0f, +0.0f, -1.0f, +0.0f, +0.0f, // back (-X)
+                +0.0f, -1.0f, +0.0f, +0.0f, -1.0f, +0.0f, +0.0f, -1.0f, +0.0f, +0.0f, -1.0f, +0.0f, // left (-Y)
+                +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, // top (+X)
+                +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, // bottom (+X)
+            };
```
lumberyard-employee-dm, line 127: All of these arrays can be marked constexpr
```suggestion
            constexpr AZStd::array<uint32_t, 6 * 6> indices = {
                 0,  1,  2,  0,  2,  3,   // front face
                 4,  5,  6,  4,  6,  7,   // right face
                 8,  9, 10,  8, 10, 11,   // back face
                12, 13, 14, 12, 14, 15,   // left face
                16, 17, 18, 16, 18, 19,   // top face
                20, 21, 22, 20, 22, 23    // bottom face
            };

            // 3 values per position, 4 positions per face, 6 faces
            constexpr AZStd::array<float, 3 * 4 * 6> positions = {
                -0.5f, -0.5f, -0.5f, +0.5f, -0.5f, -0.5f, +0.5f, -0.5f, +0.5f, -0.5f, -0.5f, +0.5f,     // front
                +0.5f, -0.5f, -0.5f, +0.5f, +0.5f, -0.5f, +0.5f, +0.5f, +0.5f, +0.5f, -0.5f, +0.5f,     // right
                +0.5f, +0.5f, -0.5f, -0.5f, +0.5f, -0.5f, -0.5f, +0.5f, +0.5f, +0.5f, +0.5f, +0.5f,     // back
                -0.5f, +0.5f, -0.5f, -0.5f, -0.5f, -0.5f, -0.5f, -0.5f, +0.5f, -0.5f, +0.5f, +0.5f,     // left
                -0.5f, -0.5f, +0.5f, +0.5f, -0.5f, +0.5f, +0.5f, +0.5f, +0.5f, -0.5f, +0.5f, +0.5f,     // top
                -0.5f, +0.5f, -0.5f, +0.5f, +0.5f, -0.5f, +0.5f, -0.5f, -0.5f, -0.5f, -0.5f, -0.5f,     // bottom
            };

            // 2 values per position, 4 positions per face, 6 faces
            // This aribtrarily maps the UVs to use the full texture on each face.
            // This choice can be changed if a different mapping would be more usable.
            constexpr AZStd::array<float, 2 * 4 * 6> uvs = {
                0.0f, 0.0f, 1.0f, 0.0f, 1.0f, 1.0f, 0.0f, 1.0f,     // front
                0.0f, 0.0f, 1.0f, 0.0f, 1.0f, 1.0f, 0.0f, 1.0f,     // right
                0.0f, 0.0f, 1.0f, 0.0f, 1.0f, 1.0f, 0.0f, 1.0f,     // back
                0.0f, 0.0f, 1.0f, 0.0f, 1.0f, 1.0f, 0.0f, 1.0f,     // left
                0.0f, 0.0f, 1.0f, 0.0f, 1.0f, 1.0f, 0.0f, 1.0f,     // top
                0.0f, 0.0f, 1.0f, 0.0f, 1.0f, 1.0f, 0.0f, 1.0f,     // bottom
            };

            // 3 values per position, 4 positions per face, 6 faces
            constexpr AZStd::array<float, 3 * 4 * 6> normals = {
                +0.0f, -1.0f, +0.0f, +0.0f, -1.0f, +0.0f, +0.0f, -1.0f, +0.0f, +0.0f, -1.0f, +0.0f,     // front (-Y)
                +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f,     // right (+X)
                +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f,     // back (+Y)
                -1.0f, +0.0f, +0.0f, -1.0f, +0.0f, +0.0f, -1.0f, +0.0f, +0.0f, -1.0f, +0.0f, +0.0f,     // left (-X)
                +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, +1.0f,     // top (+Z)
                +0.0f, +0.0f, -1.0f, +0.0f, +0.0f, -1.0f, +0.0f, +0.0f, -1.0f, +0.0f, +0.0f, -1.0f,     // bottom (-Z)
            };

            // 4 values per position, 4 positions per face, 6 faces
            constexpr AZStd::array<float, 4 * 4 * 6> tangents = {
                0.0f, 0.0f, 1.0f, 1.0f, 0.0f, 0.0f, 1.0f, 1.0f, 0.0f, 0.0f, 1.0f, 1.0f, 0.0f, 0.0f, 1.0f, 1.0f, // front (+Z)
                0.0f, 0.0f, 1.0f, 1.0f, 0.0f, 0.0f, 1.0f, 1.0f, 0.0f, 0.0f, 1.0f, 1.0f, 0.0f, 0.0f, 1.0f, 1.0f, // right (+Z)
                0.0f, 0.0f, 1.0f, 1.0f, 0.0f, 0.0f, 1.0f, 1.0f, 0.0f, 0.0f, 1.0f, 1.0f, 0.0f, 0.0f, 1.0f, 1.0f, // back (+Z)
                0.0f, 0.0f, 1.0f, 1.0f, 0.0f, 0.0f, 1.0f, 1.0f, 0.0f, 0.0f, 1.0f, 1.0f, 0.0f, 0.0f, 1.0f, 1.0f, // left (+Z)
                0.0f, 1.0f, 0.0f, 1.0f, 0.0f, 1.0f, 0.0f, 1.0f, 0.0f, 1.0f, 0.0f, 1.0f, 0.0f, 1.0f, 0.0f, 1.0f, // top (+Y)
                0.0f, -1.0f, 0.0f, 1.0f, 0.0f, -1.0f, 0.0f, 1.0f, 0.0f, -1.0f, 0.0f, 1.0f, 0.0f, -1.0f, 0.0f, 1.0f, // bottom (-Y)
            };

            // 3 values per position, 4 positions per face, 6 faces
            constexpr AZStd::array<float, 3 * 4 * 6> bitangents = {
                +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, // front (+X)
                +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, // right (+Y)
                -1.0f, +0.0f, +0.0f, -1.0f, +0.0f, +0.0f, -1.0f, +0.0f, +0.0f, -1.0f, +0.0f, +0.0f, // back (-X)
                +0.0f, -1.0f, +0.0f, +0.0f, -1.0f, +0.0f, +0.0f, -1.0f, +0.0f, +0.0f, -1.0f, +0.0f, // left (-Y)
                +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, // top (+X)
                +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, +1.0f, +0.0f, +0.0f, // bottom (+X)
            };
```

Gems/AtomLyIntegration/CommonFeatures/Code/Include/AtomLyIntegration/CommonFeatures/Mesh/MeshComponentBus.h
```diff
@@ -106,6 +106,10 @@ namespace AZ
             : public ComponentBus
         {
         public:
+            // Notifications can be triggered from job threads, so this uses a mutex to guard against
+            // listeners joining or leaving the ebus on other threads mid-notification.
+            typedef AZStd::recursive_mutex MutexType;
```
lumberyard-employee-dm, line 111: using is preferred over typedef
```suggestion
            // Notifications can be triggered from job threads, so this uses a mutex to guard against
            // listeners joining or leaving the ebus on other threads mid-notification.
            using MutexType = AZStd::recursive_mutex;
```

Gems/DiffuseProbeGrid/Code/Source/Render/DiffuseProbeGridPreparePass.cpp
```diff
@@ -44,7 +44,7 @@ namespace AZ
         {
             // load shader
             // Note: the shader may not be available on all platforms
-            AZStd::string shaderFilePath = "Shaders/DiffuseGlobalIllumination/DiffuseProbeGridPrepare.azshader";
+            AZStd::string shaderFilePath = "shaders/diffuseglobalillumination/diffuseprobegridprepare.azshader";
```
lumberyard-employee-dm, line 47: Can we make a constant for this at the top of the namespace within this file?
i.e
```c++
constexpr AZStd::string_view DiffuseProbeGridPrepareShader = "shaders/diffuseglobalillumination/diffuseprobegridprepare.azshader";
```

lumberyard-employee-dm (approved):

Gems/Atom/RPI/Code/Source/RPI.Reflect/Asset/AssetUtils.cpp
mbalfour-amzn: It _could_, but my preferred route would be to phase out the direct use of hard-coded paths entirely and have IDs passed in. (I'll be entering GHIs for these) It also gets a bit tricky because of the way we potentially need _product_ paths for the ID lookup, but _source_ paths for the compile request.

The biggest problem I've seen though with passing in paths is that it prevents us from using the "missing asset" flow and requires the compile to happen to even get an assetId from the lookup. This is especially a problem with the places where azmodel assets are passed in, because they can fail the CompileAssetSync due to their reliance on intermediate assets.

I'll change this now if you feel strongly, but I'd rather see folks work on changing the path usage over to ID usage.

lumberyard-employee-dm: What is currently there is fine?

Gems/Atom/Feature/Common/Code/Source/OcclusionCullingPlane/OcclusionCullingPlane.cpp
mbalfour-amzn: Yep, though it needs a bit more than just that to handle reloads correctly, so I'm writing it up as a separate GHI.

Gems/Atom/RPI/Code/Source/RPI.Reflect/Asset/AssetUtils.cpp
VickyAtAZ: What's the benefit of that? I mean using get asset id by file path then use CompileAssetSyncById compare to CompileAssetSync(assetPath)

Code/Framework/AzFramework/AzFramework/Asset/AssetCatalog.cpp
```diff
@@ -841,7 +841,7 @@ namespace AzFramework
                     }
 
 #if defined(AZ_ENABLE_TRACING)
-                    if (message.m_assetType == AZ::Data::s_invalidAssetType)
+                    if (isNewAsset && (message.m_assetType == AZ::Data::s_invalidAssetType))
                     {
                         AZ_TracePrintf(
```
AMZN-alexpete, line 846: nit, could possibly convert the AZ_TracePrintf on 846 to AZ_Trace?

VickyAtAZ (approved):

Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
```diff
@@ -577,17 +577,9 @@ namespace AZ
...
-                    status, &AzFramework::AssetSystemRequestBus::Events::CompileAssetSync, pipelineName.data());
-                AZ_Assert(
-                    status == AzFramework::AssetSystem::AssetStatus_Compiled || status == AzFramework::AssetSystem::AssetStatus_Unknown,
-                    "Could not compile the default render pipeline at '%s'",
-                    pipelineName.data());
-
+                // When running with an Asset Processor, this will attempt to compile the asset before loading it.
                 Data::Asset<RPI::AnyAsset> pipelineAsset =
-                    RPI::AssetUtils::LoadAssetByProductPath<RPI::AnyAsset>(pipelineName.data(), RPI::AssetUtils::TraceLevel::Error);
+                    RPI::AssetUtils::LoadCriticalAsset<RPI::AnyAsset>(pipelineName.data(), RPI::AssetUtils::TraceLevel::Error);
```
AMZN-alexpete, line 582: I wonder if the previous code was the way it was to provide better error messaging if the pipeline couldn't compile? Though, the error messaging for LoadCriticalAsset looks OK to me.  Maybe Sid or Vicky have an opinion?

Gems/Atom/RPI/Code/Source/RPI.Reflect/Image/StreamingImageAssetHandler.cpp
```diff
@@ -39,10 +39,13 @@ namespace AZ
             Data::AssetHandler::LoadResult loadResult = Data::AssetHandler::LoadResult::Error;
             if (assetData)
             {
-                loadResult = Utils::LoadObjectFromStreamInPlace<StreamingImageAsset>(*stream, *assetData, m_serializeContext) ?
-                    Data::AssetHandler::LoadResult::LoadComplete :
-                    Data::AssetHandler::LoadResult::Error;
-                
+                if (stream->GetLength() > 0)
```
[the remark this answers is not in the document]
mbalfour-amzn: Right now, it doesn't happen because of other changes I made, but there's nothing _preventing_ it from happening again. 

The way it happened before was that the AssetUtils::AsyncAssetLoader was passing "true" to auto-generate an ID in the AssetUtils::GetAssetIdFromProductPath method:
https://github.com/o3de/o3de/blob/development/Gems/Atom/RPI/Code/Include/Atom/RPI.Reflect/Asset/AssetUtils.inl#L22
https://github.com/o3de/o3de/blob/development/Gems/Atom/RPI/Code/Source/RPI.Reflect/Asset/AssetUtils.cpp#L39-L44

The AsyncAssetLoader is only used in one place right now, which is here:
https://github.com/o3de/o3de/blob/development/Gems/Atom/Feature/Common/Code/Source/CoreLights/LtcCommon.cpp#L52-L55

If these images weren't compiled by the time this was called, the "true" would cause an entry to get added into the AssetCatalog for an empty 0-length stream for these assets. At that point, the asset load would try to load the asset, see the 0 length, and immediately go into the StreamingImageAssetHandler::LoadAssetData. The 0-length buffer would get passed into the ObjectStream code, and would assert (and maybe crash, I forget, but probably just asserts).

In this PR, I've removed the code that allows you to auto-register an ID with the asset catalog, so instead the assets will show up as missing, which will cause them to successfully load the StreamingImage placeholder assets instead and _in theory_ should cause it to get auto-replaced with the correct assets assuming StreamingImage handles the reloads correctly without LtcCommon needing to do anything (I still need to verify).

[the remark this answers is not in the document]
mbalfour-amzn: Oh, and so what I mean by "nothing is _preventing_ it from happening again" is that if anyone else adds code to auto-add entries into the AssetCatalog for a StreamingImageAsset then tries to load it, the 0-length pathway could get triggered again. So it didn't seem like a bad idea to leave in the little safeguard in the loading code.

Gems/Atom/RPI/Code/Source/RPI.Reflect/Asset/AssetUtils.cpp
mbalfour-amzn: The main benefit is that CompileAssetSync goes through a whole set of heuristics to try to figure out what to compile based on string matching and multiple O(N) searches, where CompileAssetSyncById just does a single O(N) search:
https://github.com/o3de/o3de/blob/development/Code/Tools/AssetProcessor/native/resourcecompiler/rcjoblistmodel.cpp#L393-L563
vs
https://github.com/o3de/o3de/blob/development/Code/Tools/AssetProcessor/native/resourcecompiler/rcjoblistmodel.cpp#L565-L590

Gems/Atom/RPI/Code/Source/RPI.Reflect/Model/ModelAssetHelpers.cpp
```diff
@@ -0,0 +1,206 @@
...
+#include <Atom/RPI.Reflect/Buffer/BufferAssetCreator.h>
+#include <Atom/RPI.Reflect/Model/ModelAssetHelpers.h>
+#include <Atom/RPI.Reflect/Model/ModelLodAssetCreator.h>
+#include <Atom/RPI.Reflect/ResourcePoolAssetCreator.h>
+
+namespace AZ
+{
+    namespace RPI
+    {
+        // Build a buffer asset that contains the given data.
```
moudgils, line 18: Ideally you want the method comments in the header file

moudgils (approved):
Looks good. just some minor comments

Gems/Atom/RPI/Code/Include/Atom/RPI.Reflect/Model/ModelAsset.h
```diff
@@ -81,9 +82,23 @@ namespace AZ
...
+                AZStd::span<AZ::Name> tags);
+
             // AssetData overrides...
             bool HandleAutoReload() override
             {
+                // Automatic asset reloads via the AssetManager are disabled for Atom models and their dependent assets because reloads
+                // need to happen in a specific order to refresh correctly. They require more complex code than what the default
+                // AssetManager reloading provides. See ModelReloader() for the actual handling of asset reloads.
+                // Models need to be loaded via the MeshFeatureProcessor to reload correctly, and reloads can be listened
+                // to by using MeshFeatureProcessor::ConnectModelChangeEventHandler().
```
AMZN-alexpete, line 101: nice! very helpful

Code/Framework/AzFramework/AzFramework/Asset/AssetCatalog.cpp
mbalfour-amzn: I'll change it to AZ_Info, but not AZ_Trace. Even though the messages are spammy, they are legitimately notifying of problems. Specifically, we have files in the Cache/ folder that don't have a registered asset type. The worst offenders are the *.abdata.json files that appear to be recent Asset Browser metadata additions, but there are a few others too.

Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
mbalfour-amzn: FWIW if the pipeline fails to compile then error messages should appear in the AP that are far more specific than this one, describing exactly how/why it didn't compile.

Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
```diff
@@ -2120,6 +2121,14 @@ namespace AZ
                     material->GetAsset()->GetMaterialTypeAsset()->GetUvNameMap());
                 AZ_Assert(result, "Failed to retrieve mesh stream buffer views");
 
+                // The code below expects streams for positions, normals, tangents, bitangents, and uvs.
+                constexpr size_t NumExpectedStreams = 5;
+                if (streamBufferViews.size() < NumExpectedStreams)
+                {
+                    AZ_Warning("MeshFeatureProcessor", false, "Model is missing one or more expected streams, skipping.");
```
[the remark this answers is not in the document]
mbalfour-amzn: There's _equally_ unhelpful messaging right above this where it asserts and says "Failed to retrieve mesh stream buffer views". ;) I'll change the warning to this: "Model is missing one or more expected streams (positions, normals, tangents, bitangents, uvs), skipping the raytracing data generation." FWIW, I'm not sure how this error would occur in the common use cases though, since the model exporter should produce all of these. I only ran into this when generating my in-memory unit cube model because I only generated positions and normals at first.

Also, I can't speak to why those streams are set to optional above, since I can say from first-hand observation the code used to crash when dereferencing the stream buffer views for those streams if they don't exist. :) I suspect that the code for getting the byte counts and descriptors for the optional streams should probably have guards around them, but at least this code won't crash in the meantime until someone does that (if ever).

Gems/Atom/RPI/Code/Include/Atom/RPI.Reflect/Asset/AssetUtils.h
```diff
@@ -133,21 +135,10 @@ namespace AZ
...
-                    AzFramework::AssetSystem::AssetStatus status = AzFramework::AssetSystem::AssetStatus_Unknown;
-                    AzFramework::AssetSystemRequestBus::BroadcastResult(
-                        status, &AzFramework::AssetSystemRequestBus::Events::CompileAssetSync, assetFilePath);
-                    if (status != AzFramework::AssetSystem::AssetStatus_Compiled && status != AzFramework::AssetSystem::AssetStatus_Unknown)
-                    {
-                        AssetUtilsInternal::ReportIssue(reporting, AZStd::string::format("Could not compile asset '%s'", assetFilePath.c_str()).c_str());
-                        return {};
-                    }
-                }
+                TryToCompileAsset(assetFilePath, reporting);
```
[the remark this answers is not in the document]
mbalfour-amzn: It's a redundant check because CompileAssetSync also checks for an AP connection and returns back "Unknown" if the AP isn't connected, so I removed it.

Gems/Atom/RPI/Code/Source/RPI.Reflect/Model/ModelAssetHelpers.cpp
```diff
@@ -0,0 +1,206 @@
...
+            return asset;
+        }
+
+        void ModelAssetHelpers::CreateUnitCube(ModelAsset* modelAsset)
+        {
+            // Build a mesh containing a unit cube.
+            // The vertices are duplicated for each face so that we can have correct per-face normals and UVs.
+
+            // 6 vertices per face, 6 faces.
+            constexpr AZStd::array<uint32_t, 6 * 6> indices = {
```
AMZN-alexpete, line 68: I'm wondering if there might be a simpler to create a single face struct and transform it 6 times to generate the values for all the buffers.  Kinda like the icosphere version https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Math/Geometry3DUtils.cpp#L51 e.g.
```c++
// a single forward facing quad
struct quad {
    float positions[6] = {};
    float normal[3] = {0,1,0};
    uint32_t indicies[6] = {0,1,2,0,2,3};
    float uvs[2 * 4] = {0.0f, 0.0f, 1.0f, 0.0f, 1.0f, 1.0f, 0.0f, 1.0f};
    float tangent[4] = {0.0f, 0.0f, 1.0f, 1.0f};
    float bitangent[4] = {1.0f, +0.0f, +0.0f, +1.0f}
};

// reserve 
indices.reserve(blah);
positions.reserve(blah);

for(faceRotation : faceRotations)
{
     // rotate face and fill in buffers
}
```

mbalfour-amzn: Probably. I was thinking about that too, especially now that I've added a second version that generates an X instead of a cube based on UX feedback (they were concerned the cube would look too correct if viewed in-scene with a material on it). 

On the other hand, the version I have is still pretty straightforward, I'm not fully convinced a data generation version of the function will actually be smaller or simpler. If you want to refactor it in a follow-up PR I'd be fully supportive of that though. ;)

AMZN-alexpete (approved):

PR #16201, opened 2023-06-20
Add Unit Test For Create Archive Path CLI Function

VeryKaylin-player1:
## What does this PR do?

Add Unit test to make sure the sha256 and URL is being updated by using the existing function validate_downloaded_zip_sha256 in download.py file to validate.

## How was this PR tested?

Passed Unit test by calling the edit_repo_props function from repo_properties.py file and made sure release_archive_path function got called and correct release archive zip was formed by using the validate_downloaded_zip_sha256 in download.py file.

### DIFF
--- a/scripts/o3de/tests/test_repo_properties.py
+++ b/scripts/o3de/tests/test_repo_properties.py
@@ -7,14 +7,15 @@
 #
 
 import json
+import os
+import shutil
 import pytest
 import pathlib
 from unittest.mock import patch
 
-from o3de import repo_properties
+from o3de import download, repo_properties
 
-TEST_TEMPLATE_REPO_JSON = '''
-{
+TEST_TEMPLATE_REPO_JSON = '''{
     "repo_name": "repotest",
     "repo_uri": "https://test.com",
     "$schemaVersion": "1.0.0",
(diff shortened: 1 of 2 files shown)


AMZN-alexpete (approved):
Now that I see these changes and that were only using the temp folder for one test, I think it would have been cleaner to create a separate test function using the built in tmpdir parameter.  Maybe we can circle back to it later if we have time.

scripts/o3de/tests/test_repo_properties.py
```diff
@@ -310,7 +395,11 @@ def get_json_data(object_typename: str,
                 assert self.repo_json.data.get('origin') == 'o3de'
 
                 for gem in self.repo_json.data.get('gems_data', []):
-                    assert gem in expected_gems
+                    if release_archive_path:
+                        zip_path = release_archive_path / pathlib.Path(gem['download_source_uri']).name
+                        download.validate_downloaded_zip_sha256(gem, zip_path, "gem.json")
```
AMZN-Gene, line 400: do we need to assert download.validate_downloaded_zip_sha256 returns true or something?

VeryKaylin-player1: The function will raise an error if it fails, since its not mocked, the original function won't pass

AMZN-Gene: Spoke offline with @VeryKaylin-player1.
Unit test will error inside that method and so will fail (as expected)

AMZN-Gene (approved):

Issue #16076, opened 2023-06-01
DPE Inspector: Some enum fields show value name instead of display name


cgalvan:
**Describe the bug**
Some enum fields show the actual name of the enum variable in the combo-box instead of the appropriate display name.

This issue has been found in the following components:
Image Gradient, Directional Light, PhysX Dynamic Rigid Body, Terrain Macro Material

**Steps to reproduce**

1. Launch the Editor with the DPE enabled:
`Editor.exe --ed_enableDPE=true`
2. Open a new or existing level.
3. Create a new entity.
4. Add a `Image Gradient` component to the entity.

**Expected behavior**
The "Auto-Save Mode" field should have the following options in the combo-box:
Save As..., Auto Save, Auto Save With Incrementing Names

**Actual behavior**
The "Auto-Save Mode" field has the following options in the combo-box:
SaveAs, AutoSave, AutoSaveWithIncrementalNames

alexmontAmazon:
this is an issue with enums that don't use u64 as the base type.

PR #16203, opened 2023-06-21
DPE | Fix PhysX Dynamic Rigid Body type property labeling

AMZN-daimini:
- Defines `SetTrueLabel` and `SetFalseLabel` attributes so that the `Type` property of the `PhysX Dynamic Rigid Body
` component is populated correctly.
- Also fixes minor issues in comments.

BEFORE:
image

AFTER:
image

### DIFF
--- a/Code/Framework/AzCore/AzCore/RTTI/ReflectContext.h
+++ b/Code/Framework/AzCore/AzCore/RTTI/ReflectContext.h
@@ -269,7 +269,7 @@ namespace AZ
             return {};
         }
 
-        /// Gets a marshaleld Dom::Value representation of this attribute bound to a given instance.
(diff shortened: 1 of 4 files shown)


cgalvan (approved):

alexmontAmazon (approved):
Looks good! I feel like might be a bug in our backlog related to this (or at least other True False stuff), so maybe check if this fixes one of our GHIs

AMZN-daimini:
> Looks good! I feel like might be a bug in our backlog related to this (or at least other True False stuff), so maybe check if this fixes one of our GHIs

It partially fixes https://github.com/o3de/o3de/issues/16076 but the issue on other components has a different root cause (the "enum underlying type different than AZ::u64" issue I was looking into and could not resolve last week :(

dtamkin1 (approved):

PR #16206, opened 2023-06-21
Hide Slice asset type from search categories.

### LINKED ISSUE #10813: Bug Report -Asset Browser Filter menu shows Slices instead of Prefabs
**Describe the bug**
When using the Asset Browser dropdown menu, you see a filter option for slices. Because we no longer officially support slices, I recommend removing this option from the dropdown menu.

ResourceIMG-7-21-2022-04-32-32-PM

**Steps to reproduce**
1. Go to Asset Browser
2. Click on the hamburger menu
4. Scroll down to slices
5. See error

**Expected behavior**
I would expect to not see Slices

sphrose:
## What does this PR do?

This PR resolves #10813 by changing the GetGroup return value from Slice to Hidden, which causes it to be ignored in the filter category list.

## How was this PR tested?

Checked in search category list in the AssetBrowser, slice no longer appears, but prefab is still there.


AMZN-daimini (commented):
Ideally, it is still possible to disable prefabs and use slices. So maybe this should be:

```
bool isPrefabSystemEnabled = false;
AzFramework::ApplicationRequests::Bus::BroadcastResult(isPrefabSystemEnabled, &AzFramework::ApplicationRequests::IsPrefabSystemEnabled);

if (isPrefabSystemEnabled)
{
    return "Hidden";
}
else
{
    return "Slice";
}
```

AMZN-daimini (approved):

dtamkin1 (approved):

cgalvan (approved):

PR #9698, opened 2022-05-21
Fix for crashes on Linux when actions related to deleting nodes are invoked in Script Canvas

spham-amzn:
There are a few bugs logged ( https://github.com/o3de/o3de/issues/9335, https://github.com/o3de/o3de/issues/7579, https://github.com/o3de/o3de/issues/9335, https://github.com/o3de/o3de/issues/8016, https://github.com/o3de/o3de/issues/7849 ) that causes crashes in Script Canvas were all related to nodes being directly or indirectly deleted. 

This also undoes a previous fix attempt ( https://github.com/o3de/o3de/pull/9629 ) that instead tried to protect from accessing the nodes member variables in the middle of a deletion. (This partially fixed it in some situations, but the root cause was the timing issue that this change fixes), the description of the problem and a working solution was proposed in PR https://github.com/o3de/o3de/pull/9679 .  

But it seems all of the crashes can be prevented by applying the same fix at the base call to ```SceneComponent::Delete``` .

This change will:

1. Insert a force of MainWindow::OnSystemTick whenever entities are deleted
2. Rollback previous fix for crash that just tried to protect the object during deletion

Signed-off-by: Steve Pham <82231385+spham-amzn@users.noreply.github.com>

PR #16222, opened 2023-06-22
Parsing for implicit connections to nodes with multiple data slots

mddvisu:
## What does this PR do?

This PR allows for nodes with an execution in slot marked with the m_createsImplicitConnections attribute to be able to parse when it has more than one data in slot. This is accomplished during parsing, where implicit connections connected to nodes with implicit connections that have not been parsed yet will be ignored.
This PR also fixes a crash that happens when nodes that have implicit connections are deleted by refreshing selections before the deletion of nodes. This crash previously only happened on Linux and was fixed here: https://github.com/o3de/o3de/pull/9698.

## How was this PR tested?

Tests for ScriptCanvas and ScriptCanvas Editor were run locally. I also did manual testing by creating, deleting, undoing, and redoing connections. Three simple nodeables were created to test functionality, one that increments a number, one that adds two numbers to each other, and one that adds three numbers to each other:

multoutgraph1
<br>
multoutout1
 <br><br>
multoutgraph2
<br>
multoutout2
<br><br>
multoutgraph3
<br>
multoutout3

### DIFF
--- a/Gems/ScriptCanvas/Code/Include/ScriptCanvas/Grammar/AbstractCodeModel.cpp
+++ b/Gems/ScriptCanvas/Code/Include/ScriptCanvas/Grammar/AbstractCodeModel.cpp
@@ -3317,7 +3317,27 @@ namespace ScriptCanvas
 
             // \todo Infinite loop handling both in code and in the graph will have to occur here.
             auto executionOutNodes = execution->GetId().m_node->GetConnectedNodes(outSlot);
-            auto numConnections = executionOutNodes.size();
+
+            // Sort the out node connections to ensure that implicit execution connections are always parsed first
+            EndpointsResolved sortedExecutionOutNodes;
+            for (const EndpointResolved& endpoint : executionOutNodes)
+            {
+                // If this connection is implicit, only parse it if the recieving node has parsed all of its implicit
+                // connections already
+                if (endpoint.second->CreatesImplicitConnections() && !HasUnparsedImplicitConnections(&outSlot, endpoint.second))
+                {
+                    sortedExecutionOutNodes.emplace_back(endpoint);
+                }
+            }
+            for (const EndpointResolved& endpoint : executionOutNodes)
+            {
+                if (!endpoint.second->CreatesImplicitConnections())
+                {
+                    sortedExecutionOutNodes.emplace_back(endpoint);
+                }
+            }
+
+            auto numConnections = sortedExecutionOutNodes.size();
 
             if (numConnections == 0)
             {
@@ -3332,13 +3352,13 @@ namespace ScriptCanvas
             }
             else if (numConnections == 1)
             {
-                ParseExecutionFunctionRecurse(execution, execution->ModChild(0), outSlot, executionOutNodes[0]);
+                ParseExecutionFunctionRecurse(execution, execution->ModChild(0), outSlot, sortedExecutionOutNodes[0]);
             }
             else
             {
                 AZStd::vector<const Slot*> outSlots;
-                outSlots.resize(executionOutNodes.size(), &outSlot);
-                ParseExecutionMultipleOutSyntaxSugar(execution, executionOutNodes, outSlots);
+                outSlots.resize(sortedExecutionOutNodes.size(), &outSlot);
+                ParseExecutionMultipleOutSyntaxSugar(execution, sortedExecutionOutNodes, outSlots);
             }
 
             ParseMultiExecutionPost(execution);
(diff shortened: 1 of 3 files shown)


Gems/ScriptCanvas/Code/Include/ScriptCanvas/Grammar/AbstractCodeModel.cpp
```diff
@@ -3317,7 +3317,27 @@ namespace ScriptCanvas
...
+                    sortedExecutionOutNodes.emplace_back(endpoint);
+                }
+            }
+            for (const EndpointResolved& endpoint : executionOutNodes)
+            {
+                if (!endpoint.second->CreatesImplicitConnections())
+                {
+                    sortedExecutionOutNodes.emplace_back(endpoint);
+                }
+            }
```
spham-amzn, line 3338: If you are just grouping to make sure the parsed implicit nodes are done first, you can probably optimize this to only loop through once by emplace_front and emplace_back
```suggestion
            for (const EndpointResolved& endpoint : executionOutNodes)
            {
                // If this connection is implicit, only parse it if the recieving node has parsed all of its implicit
                // connections already
                if (endpoint.second->CreatesImplicitConnections() && !HasUnparsedImplicitConnections(&outSlot, endpoint.second))
                {
                    sortedExecutionOutNodes.emplace_front(endpoint);
                }
                else if (!endpoint.second->CreatesImplicitConnections())
                {
                    sortedExecutionOutNodes.emplace_back(endpoint);
                }
            }
```

lsemp3d (approved):

Gems/ScriptCanvas/Code/Include/ScriptCanvas/Grammar/AbstractCodeModel.cpp
mddvisu: Behind the scenes, EndpointsResolved seems to be a vector which means that it doesn't have an emplace_front method. I could insert the items into the front of the list, but each time I do that the vector would have to shift all other elements which would hurt efficiency https://www.geeksforgeeks.org/vector-insert-function-in-cpp-stl/. Not a huge deal though, since I don't expect this vector to consist of any more than 2-3 elements in most scenarios. But I do need the executionOutNodes with the implicit connections added to that vector in that order (first in first out) for parsing to work with implicitly connected nodes that are chained together.

spham-amzn: Yeah, its really just a minor observation and not worth it to over analyze (I didn't catch that it was a vector). But the shift isn't a big deal in this case anyways since its a vector of references. Leaving your original solution will at least preserve the order of like items anyways, if that was important at all

spham-amzn (approved):

Gems/ScriptCanvas/Code/Include/ScriptCanvas/Grammar/AbstractCodeModel.cpp
```diff
@@ -3800,6 +3806,51 @@ namespace ScriptCanvas
...
+                {
+                    connectedNodes = inNode->GetConnectedNodes(*inSlot);
+                }
+                for (const EndpointResolved& connectedNode : connectedNodes)
+                {
+                    // If the endpoint's slot is not the same as the output slot currently being processed, check if it has already been parsed
+                    if (connectedNode.second != outSlot)
+                    {
+                        // If there are no connections to ignore yet, that must mean there are still more slots on this node to parse
+                        if (m_parsedImplicitConnections.size() == 0)
```
cgalvan, line 3825: Nit: Could also check for `empty()` to make the line a little shorter

Gems/ScriptCanvas/Code/Include/ScriptCanvas/Grammar/AbstractCodeModel.cpp
```diff
@@ -3800,6 +3806,51 @@ namespace ScriptCanvas
...
+                        {
+                            m_parsedImplicitConnections.push_back(AZStd::make_pair(outSlot, inSlot));
+                            return true;
+                        }
+                        else
+                        {
+                            bool foundConnectionInParsed = false;
+                            for (auto connection : m_parsedImplicitConnections)
```
cgalvan, line 3833: Nit: Could use `const auto&` here so it doesn't make a local coppy
```suggestion
                            for (const auto& connection : m_parsedImplicitConnections)
```

cgalvan (approved):

PR #16224, opened 2023-06-23
Add README.md Files And Generate Repo.json.bak File When Change Is Made

VeryKaylin-player1:
## What does this PR do?

1. Auto Creates Gems/Projects/Templates folder when create-repo command is called, each with README.md inside of them that clarifies what should be placed in those folders. 
2. Allow repo.json to generate a backup file called repo.json.bak every time user made a change to their repo.json file

## How was this PR tested?
1. Auto generation of folders is tested by calling this command to create a remote repo:
> ./o3de.bat create-repo -rp "C:\o3de_Projects\RemoteRepoCheck\repo.json" -ru "https://github.com/VeryKaylin-player1/RemoteRepoCheck"
 Result: Three folders would show up with appropriate README.md files inside them with appropriate message

- Rerun all unit test in test_repo_properties.py file to make sure adding this function won't affect the previous functions

2. To test repo.json.bak file call add command on edit-repo-properties: 
> ./o3de.bat edit-repo-properties -rp "C:\o3de_Projects\RemoteRepoCheck\repo.json" --add-gem "C:\o3de_Projects\RemoteRepoCheck\Gems\MyGem"
Result: This will create a backup of the current repo.json file before adding the new gem called repo.json.bak0

### DIFF
--- a/scripts/o3de/o3de/repo_properties.py
+++ b/scripts/o3de/o3de/repo_properties.py
@@ -306,12 +306,15 @@ def edit_repo_props(repo_path: pathlib.Path = None,
         repo_json['repo_name'] = repo_name
 
     if add_gems or delete_gems or replace_gems:
+        utils.backup_file(repo_path)
         _edit_objects('gem', validation.valid_o3de_gem_json, repo_json, add_gems, delete_gems, replace_gems, release_archive_path, force, download_prefix)
 
     if add_projects or delete_projects or replace_projects:
+        utils.backup_file(repo_path)
         _edit_objects('project', validation.valid_o3de_project_json, repo_json, add_projects, delete_projects, replace_projects, release_archive_path, force, download_prefix)
 
     if add_templates or delete_templates or replace_templates:
+        utils.backup_file(repo_path)
         _edit_objects('template', validation.valid_o3de_template_json, repo_json, add_templates, delete_templates, replace_templates, release_archive_path, force, download_prefix)
 
     return 0 if manifest.save_o3de_manifest(repo_json, repo_path) else 1
--- a/scripts/o3de/o3de/repo_properties.py
+++ b/scripts/o3de/o3de/repo_properties.py
@@ -9,6 +9,7 @@
 This file contains all the code that has to do with editing the remote repo template
 """
 import argparse
+import copy
 import pathlib
 import sys
 import json
@@ -298,6 +299,8 @@ def edit_repo_props(repo_path: pathlib.Path = None,
     if not repo_json:
         return 1
 
+    repo_json_original = copy.deepcopy(repo_json)
+
     if isinstance(repo_name, str):
         if not utils.validate_identifier(repo_name):
             logger.error(f'Repo name must be fewer than 64 characters, contain only alphanumeric, "_" or "-"'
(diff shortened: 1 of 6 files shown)


scripts/o3de/o3de/repo_properties.py
```diff
@@ -306,12 +306,15 @@ def edit_repo_props(repo_path: pathlib.Path = None,
         repo_json['repo_name'] = repo_name
 
     if add_gems or delete_gems or replace_gems:
+        utils.backup_file(repo_path)
```
tkothadev, line 309: In the testing framework for the O3DE CLI, we have unit tests outlined for utils.py, as seen in this path: scripts/o3de/tests/test_utils.py

I think the `backup_file` method could warrant a unit test to verify its intended behavior. Currently that is missing from test_utils.py. That may be out of scope for the current PR, but something to consider.

tkothadev:
I recommend running the unit tests for repo_properties.py (found in test_repo_properties.py) for this PR. This is to ensure that any introduced changes do not introduce new error states. We use the pytest framework for running unit tests.

Once you can confirm that unit tests work correctly, I recommend posting the results of the tests to the "How this PR was tested" section.

scripts/o3de/o3de/repo_properties.py
spham-amzn: yeah, out of scope for this PR, it would be more relevant for this PR to create unit tests for `repo_properties` and mock the `utils.backup_file` instead. 
In other news, if we decide to eventually unit test 'backup_file', make sure to use pytest's temp directory as it is automatically cleaned up for you

https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html

spham-amzn (approved):

scripts/o3de/o3de/repo_properties.py
VeryKaylin-player1: Thank you T.J and Steve!! Will definitely look into this

tkothadev (approved):

scripts/o3de/o3de/repo_properties.py
```diff
@@ -306,12 +306,15 @@ def edit_repo_props(repo_path: pathlib.Path = None,
...
         _edit_objects('gem', validation.valid_o3de_gem_json, repo_json, add_gems, delete_gems, replace_gems, release_archive_path, force, download_prefix)
 
     if add_projects or delete_projects or replace_projects:
+        utils.backup_file(repo_path)
```
AMZN-alexpete, line 313: Rather than placing three backup calls inside each if block , I recommend you save a copy of `repo_json` at the start of the function and then at the end of this function on line 319 you call `utils.backup_file` only if the contents changed like this:
```
if repo_json_original != repo_json:
    utils.backup_file(repo_path)
```

There a case when the backup file could overwrite an existing backup when you don't want it to - for example, say you already have a backup file and you run this function with a syntax error that causes _edit_objects() to fail, your backup file would get replaced even though no changes were made to the file.

Also, the backup file would be created multiple times if you combined add and remove commands.

AMZN-alexpete (changes requested):
Lets only backup the contents if they've changed, and assert that the backup function is called in the test.

AMZN-alexpete (approved):

Issue #15960, opened 2023-05-15
Installation debian package does not work for ubuntu 22.04


spham-amzn:
**Describe the bug**
The nightly development Linux debian installer will not install due to the fact that it was built on Ubuntu 20.04, which uses OpenSSL 1.1.1.x for its system OpenSSL library, while Ubuntu 22.04 (and forward) uses OpenSSL 3.x. This issue is not only an issue with Ubuntu, but with other Linux distros depending on the major version of OpenSSL is installed at that system level (generally, they should not have a side-by-side installation of both). 

**Steps to reproduce**
1. Go to https://o3debinaries.org/development/Latest/Linux/O3DE_latest.deb
2. Download to your local Ubuntu 22.04 host
3. Install the package with `dpkg -i` '
4. Launch o3de

**Expected behavior**
The project manager launches successfully

**Actual behavior**
An error is reported indicating missing openssl 
```
/o3de: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory
```

**Found in Branch**
Development

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
https://github.com/o3de/o3de/commit/1db85409ed279b16fc0bc068bce428a2e7e960a0

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Linux
 - Version Ubuntu 22.04
 - CPU : N/A
 - GPU : N/A
 - Memory : 16 GB

julekbr:
In the release notes from version 2111.2 they're saying they build only on Ubuntu 20 LTS and that other Linux versions might not be supported. Dunno if this info is still up to date, because in the notes of later releases I didn't find any info. I think probably you will either have to use Ubuntu 20 or build it on your own. I built it for Ubuntu 22 but there are problems when trying to open Projects over the Project Manager.

lmbr-pip:
Believe SIG/build owns installer. Please coordinate with SIG/release if a patch is needed.

PR #16226, opened 2023-06-23
Fix debian package to work for both Ubuntu 20.04 and 22.04

spham-amzn:
## What does this PR do?
This request updates the references to the new 3rd Party packages for linux which now link against the static version of the O3DE vendored OpenSSL rather than using the system one for the current ubuntu host. 
The issue before was that if a debian package was built on a system that uses OpenSSL 1.x (Ubuntu 20.04), then the debian package will rely on OpenSSL 1.x packages. The debian package will will fail to install on systems that use OpenSSL 3.x (Ubuntu 22.04) because they are not API compatible.

## How was this PR tested?

Debian Package built on 20.04 performed the following steps on Ubuntu 20.04 and Ubuntu 22.04 separately

1. Install the debian package (through dpkg -i) on clean images of ubuntu 20.04
2. Launch O3DE
3. Create a new project based on the default template from the Project Manager
4. Launch the new project from the Project Manager
5. Repeat steps 1-4 for Ubuntu 22.04

fixes https://github.com/o3de/o3de/issues/15960


cmake/Platform/Linux/Packaging_linux.cmake
```diff
@@ -25,26 +25,27 @@ elseif("$ENV{O3DE_PACKAGE_TYPE}" STREQUAL "DEB")
...
     set(CPACK_PACKAGING_INSTALL_PREFIX "/opt/${CPACK_PACKAGE_NAME}/${CPACK_PACKAGE_VERSION}")
 
-    # get all the package dependencies, extracted from scripts\build\build_node\Platform\Linux\package-list.ubuntu-focal.txt
+    # Define all the debian package dependencies needed to build and run
     set(package_dependencies
-        libffi7
-        clang-12
+        # Required Tools
+        "cmake (>=3.22)"                        # Cmake required (minimum version 3.22.0)
+        "clang (>=12.0)"                        # Clang required (minimum version 12.0)
```
amzn-changml, line 32: Nice. Something we should have thought to do earlier

amzn-changml (approved):

cmake/Platform/Linux/PAL_linux.cmake
```diff
@@ -53,16 +53,8 @@ ly_set(LY_PYTHON_CMD ${CMAKE_CURRENT_SOURCE_DIR}/python/python.sh)
 set(PAL_TRAIT_LINUX_WINDOW_MANAGER "xcb" CACHE STRING "Sets the Window Manager type to use when configuring Linux")  
 set_property(CACHE PAL_TRAIT_LINUX_WINDOW_MANAGER PROPERTY STRINGS xcb wayland)
 
-# Use system default OpenSSL library instead of maintaining an O3DE version for Linux
-include(${CMAKE_CURRENT_LIST_DIR}/OpenSSL_linux.cmake)
-
 # Use system default libunwind  instead of maintaining an O3DE version for Linux
```
lumberyard-employee-dm, line 56: While you are in here, can you remove the extra space between the words of "libunwind" and "instead"
```suggestion
# Use system default libunwind instead of maintaining an O3DE version for Linux
```

cmake/Platform/Linux/Packaging_linux.cmake
lumberyard-employee-dm: I didn't know apt-get supported dependency strings.
Apparently this was added in 2019 as part of the [apt-get satisfy](https://askubuntu.com/questions/1413251/how-does-the-apt-satisfy-work) command

lumberyard-employee-dm (approved):

cmake/Platform/Linux/Packaging_linux.cmake
spham-amzn: https://cmake.org/cmake/help/v3.23/cpack_gen/deb.html?highlight=cpack_debian_package_depends#variable:CPACK_DEBIAN_PACKAGE_DEPENDS

amzn-changml: Yeah, though dep strings have actually been a feature for CMake since 3.3 (as @spham-amzn just linked)

lumberyard-employee-dm: So CPack supports it as well.

That gives me an idea that we should update the Linux environment setup scripts to use `apt-get satisfy` then for

PR #16227, opened 2023-06-23
Fix for RayTracing MaterialInfo Vulkan validation errors

dmcdiarmid-ly:
Adjusted padding in the MaterialInfo and ReflectionProbeData structures


Gems/Atom/Feature/Common/Assets/ShaderLib/Atom/Features/RayTracing/RayTracingMaterialSrg.azsli
```diff
@@ -33,15 +33,15 @@ ShaderResourceGroup RayTracingMaterialSrg : SRG_RayTracingMaterial
...
+        float3 m_emissiveColor;
         float  m_metallicFactor;
         float  m_roughnessFactor;
-        float3 m_emissiveColor;
         uint   m_textureFlags;
         uint   m_textureStartIndex;
+        uint   m_padding0;
 
         ReflectionProbeData m_reflectionProbeData;
         uint   m_reflectionProbeCubeMapIndex;
```
moudgils, line 44: Would be safer if we move m_reflectionProbeCubeMapIndex above the struct (i.e m_reflectionProbeData) as if someone messes up ReflectionProbeData and it doesn't end with 16 byte alignment this variable is in danger. 

Also, in the shader you want to use **[[pad_to(16)]]** at the end of ReflectionProbeData and in c++ side you can declare this struct as **struct alignas(16) ReflectionProbeData**

dmcdiarmid-ly: The total padding still needs to be aligned to 16 bytes, so moving that member won't save anything.

moudgils (approved):

Issue #7400, opened 2022-02-03
Linux: Enabling Bloom on Bloom component results in populating Editor console with asserts


LB-JakubSkorupka:
**Describe the bug**
The console is populated with multiple similar asserts after enabling **Bloom** in the **Bloom** component in of an entity.
Please refer to the attached Editor.log file due to a large console output.

**File in asserts**
`/home/gpc-xxx/temp/o3de/Gems/Atom/RHI/Vulkan/Code/Source/RHI/ImageView.cpp(128)`

**Steps to reproduce**
1. Create a new entity.
2. Add **Bloom** component to the entity.
3. Enable **Bloom** in the **Bloom** component.

**Expected behavior**
**Bloom** is enabled in the **Bloom** component, no asserts are being printed to the console.

**Actual behavior**
The console is populated with asserts after enabling **Bloom** in the **Bloom** component.

**Video**

https://user-images.githubusercontent.com/86951967/152358047-a297bd07-5075-41dd-9268-edf96c6b5ae3.mp4

**Found in Branch**
Development (611598c)

**Desktop:**
 - Device: PC
 - OS: Linux
 - Version: Ubuntu 20.04.3 LTS
 - CPU: Intel Core i7-8700K
 - GPU: NVIDIA GeForce RTX 2080 Ti
 - Memory: 16GB

**Additional context**
[Editor.log](https://github.com/o3de/o3de/files/7995051/Editor.log)

LB-KacperKapusta:
It has been observed using Stabilization_2205 (3e47df2) build that in addition to the asserts being printed, modifying Bloom settings inside the component has no effect on the Viewport on Linux.

Issue #4812, opened 2021-10-20
[GNU/LINUX] The editor is frozens from Atom Bloom on AMD Cards (PostFX)


Darkinggq:
**Describe the bug**
Post effects add bloom enable immediately freeze the entire editor will not be able to exit only restart the PC

**To Reproduce**
Run editor ->

https://user-images.githubusercontent.com/17591539/138050553-5ea5506f-91d9-43de-be39-8eb19bfba644.mp4

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Xubuntu
 - Version: 20.04.03 LTS
 -  CPU: AMD 3 3100
 - GPU: AMD 5500 XT [Mesa 21.0.3]
 - Memory: 16GB

Mobs1024:
stabilization/2110 ba329378677590cdad0019c476cccccb5dec102f

https://user-images.githubusercontent.com/17591539/143003764-45d2b1dd-836a-4588-857d-e5a9efa73dea.mp4

[Editor.log](https://github.com/o3de/o3de/files/7587086/Editor.log)
[crash.log](https://github.com/o3de/o3de/files/7587087/crash.log)

Magical10240:
PostFX bloom unknown problem in ubuntu 20.04 lts the button immediately crashes, and other distro opensuse tumbleweed works does not crash this is strange

https://user-images.githubusercontent.com/17591539/154491702-92c33c01-574d-403d-82ae-05e2a674e1d4.mp4

could it be a mesa driver problem?

ubuntu 20.04 lts **Mesa 21.0.3**
VS
opensuse Tumbleweed **Mesa 21.3.6**

Darkinggq:
enable FX Bloom works fine in other non-Ubuntu distributions, Ubuntu still crashes (development 14076b4f932a768e9d12b041ca16a13903ac6be5)

Darkinggq:
Verified fix on stabilization 2310.0 (ubuntu 22.04 lts)

PR #16241, opened 2023-06-26
[LINUX] Fix crash when enabling bloom component.

yaakuro:
# What does this PR do?

This PR addresses a crash issue on Linux that occurs when the Bloom component is enabled. The fix resolves the problem, allowing the Bloom component to function properly.

The root cause of the crash was identified as the usage of strings for the WidthMultiplier and HeightMultiplier parameters in some pass templates. This resulted in zero-sized images being generated in the RHI::Size PassAttachmentSizeMultipliers::ApplyModifiers(const RHI::Size& size) const function.

It is worth noting that this issue may also be related to the following tickets:

#4812
#7400

Additionally, there are other template passes that utilize the same approach and may require similar fixes.
# How was this PR tested?

The changes in this PR were tested on Linux within the O3DE Editor environment to ensure its functionality and verify the resolution of the crash issue.

### DIFF
--- a/Gems/Atom/Feature/Common/Assets/Passes/BloomDownsample.pass
+++ b/Gems/Atom/Feature/Common/Assets/Passes/BloomDownsample.pass
@@ -44,8 +44,8 @@
                             "Attachment": "Input"
                         },
                         "Multipliers":{
-                            "WidthMultiplier" : "0.5",
-                            "HeightMultiplier": "0.5"
+                            "WidthMultiplier" : 0.5,
+                            "HeightMultiplier": 0.5
                         }
                     },
                      "FormatSource": {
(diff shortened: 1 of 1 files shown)


spham-amzn (approved):
Nice catch

cgalvan (approved):

Jackie9527 (approved):

moudgils:
Tagging @amzn-changml

amzn-changml:
This could be due to the DCO infra failing to fire (it's external to this project). The commit is signed, so we can bypass it and merge if we're ready

yaakuro:
5 maintainers checked ok. can we merge this fix now.

amzn-changml:
Merging this now

PR #16252, opened 2023-06-27
DPE | Refactor LegacyReflectionBridge, reimplement grouping

AMZN-daimini:
## What does this PR do?

Refactors the `BeginNode` and `EndNode` functions to better separate the purpose of each block and make them more readable.
Also re-writes grouping support to make it match the output of the RPE. The DPE so far assumed that the items would be iterated on in the same order as they appear in the EditContext registration code, but that does not seem to be the case - attributes nested under different groups can alternate in weird ways.
The RPE would generate one widget per group and append the properties to them even if they were presented out of order.
The DPE so far would open a group after the property that came before them in the elements list, but given that the ordering is not guaranteed that would result in groups not matching their definitions.

This re-writing does a first pass to organize the nodes, organize them into different "buckets", then to a pass in `EndNode` to correctly visit them. I'm sure there are better and more efficient ways to achieve the same result, so I'm open to suggestions.

## How was this PR tested?

Manual testing on components whose grouping was messed up. Example below.
Will test further before closing the related issues.
Note that some components still have ordering issues due to the `DisplayOrder` attribute not being consumed correctly, but the grouping should be good now.

Component as shown by the RPE:
Grouping_RPE

DPE representation before this fix:
Grouping_BEFORE

DPE representation after this fix:
Grouping_AFTER


Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/Reflection/LegacyReflectionBridge.cpp
```diff
@@ -484,87 +567,169 @@ namespace AZ::Reflection
...
+                    // Iterate over non serialized elements to see if any of them should be added
+                    const char* elementName = "";
+                    if (parentData.m_classElement && parentData.m_classElement->m_editData && parentData.m_classElement->m_editData->m_name)
+                    {
+                        elementName = parentData.m_classElement->m_editData->m_name;
+                    }
+                    auto iter = m_nonSerializedElements.begin();
+                    while (iter != m_nonSerializedElements.end())
+                    {
+                        // TODO - Fix this
```
alexmontAmazon, line 706: github really mangled this diff, but as far as I can tell, this is a new todo. Could you file a GHI for this? We've been asked to file GHIs for any known todos to help the community track this stuff...

alexmontAmazon (approved):

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/Reflection/LegacyReflectionBridge.cpp
AMZN-daimini: Hey, no this TODO already existed. I'll check if it was already tracked and add a GHI if that's not the case.

alexmontAmazon: oh, no worries then. I just didn't see it in the diff

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/Reflection/LegacyReflectionBridge.cpp
```diff
@@ -270,137 +278,201 @@ namespace AZ::Reflection
...
                     }
                 }
 
-                //! This is the case to create UIElements, Groups and Editor Data if any one of these is the first element in a class
-                auto iter = m_nonSerializedElements.begin();
-                while (iter != m_nonSerializedElements.end())
+                nodeData.m_path = AZStd::move(path);
+            }
+
+            bool IsInParentGroup(const StackEntry& nodeData, StackEntry& parentData)
```
cgalvan, line 305: Nit: By this name I thought it was just checking if the node is in a parent group, but it's actually adding it to the group vector to be handled later, so maybe a different name like `NodeShouldBeHandledByParentGroup`. Not married to the verbiage, my suggestion is actually a little wordy, but something like that would be helpful.

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/Reflection/LegacyReflectionBridge.cpp
```diff
@@ -270,137 +278,201 @@ namespace AZ::Reflection
...
+                                nodeData.m_groupEntries.insert(groupName);
+                                nodeData.m_groups.emplace_back(AZStd::make_pair(groupName, AZStd::move(entry)));
+                            }
+                            else
+                            {
+                                // If the group name is empty, this is the end of the previous group.
+                                // Create an group for entries that should be displayed after the previous group.
+                                if (!groupName.empty())
+                                {
+                                    groupName = AZStd::string::format("WHATEVER[%d]", ++groupCounter);
```
cgalvan, line 360: I'm guessing "WHATEVER" was a placeholder for something more descriptive :P

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/Reflection/LegacyReflectionBridge.cpp
```diff
@@ -270,137 +278,201 @@ namespace AZ::Reflection
...
+                                                     nodeData.m_classData,
+                                                     UIElement };
+
+                                nodeData.m_groupEntries.insert(groupName);
+                                nodeData.m_groups.emplace_back(AZStd::make_pair(groupName, AZStd::move(entry)));
+                            }
+                            else
+                            {
+                                // If the group name is empty, this is the end of the previous group.
+                                // Create an group for entries that should be displayed after the previous group.
```
cgalvan, line 357:
```suggestion
                                // Create a group for entries that should be displayed after the previous group.
```

cgalvan (approved):

dtamkin1 (approved):

Issue #16253, opened 2023-06-27
Editor menu actions can be run when in game mode


AMZN-alexpete:
**Describe the bug**
When in game mode, a user can run any editor menu action registered with ActionManager including CTRL+S (save), CTRL+N (new level) etc.  Only the "exit play mode" action should be allowed to run when in game.

**Steps to reproduce**
1. Open any level in the editor and enter game mode (CTRL+G)
2. Once in game mode, press CTRL+N

**Expected behavior**
Nothing happens because the editor actions should be disabled (except ESC action)

**Actual behavior**
The new level GUI appears the the user is allowed to create a new level and open it, which will put the editor in a broken state where it seems to think it is in game mode still.

**Screenshots/Video**
n/a

**Found in Branch**
development

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) Repository**
0e7d3bc92f40aa1d97666a3bf99c7557caf6c58b

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Windows
 - Version 10

**Additional context**

PR #16254, opened 2023-06-27
Allow Qt keyboard events only in Editor game mode when console open, and detect ESC key in Editor instead of IMGUI

### LINKED ISSUE #16198: [LINUX] Can' t stop Play In Editor with Escape key
**Describe the bug**
I am having issues to escape from Play in Editor (PIE) mode. When in game mode pressing the Escape key does not react.

**Assets required**
Not required. Any project that opens the Editor will do.

**Steps to reproduce**
1. Open Project
2. Press Play Game
3. Try to stop pressing Escape key
4. Experience the issue.

**Expected behavior**
PIE to stop after pressing the Escape key.

**Actual behavior**
Does not stop the PIE and the Editor gets useless and needs to be terminated.

**Screenshots/Video**
None

**Found in Branch**
development

17dacb028d41fb1c24aeb556dbaa431eab917fe8

 - Device: PC
 - OS: Linux, KUbuntu 22.04 (KDE Plasma Desktop)
 - Version 22.04
 - CPU:  Ryzen 5950x
 - GPU: AMD 6950 XT
 - Memory: 128Gb

**Additional context**
Looking into it a bit shows me somehow that the Qt events are being like 100% consumed by nativeEventFilter (runing the XcbEventHanderBus that is). So whatever got changed in the meanwhile, special keys (like stopping PIE with Escape) doesn't pass those keys to Qt widgets anymore (MainWindow for example).  For my personal usage, I added the **AzFramework::InputChannelNotificationBus::Handler** to **QtEditorApplication_linux.cpp**:

```cpp
void EditorQtApplicationXcb::OnInputChannelEvent(const AzFramework::InputChannel& inputChannel, bool& hasBeenConsumed)
{
  if(GetIEditor()->IsInGameMode() && 
    (inputChannel.GetInputChannelId() == AzFramework::InputDeviceKeyboard::Key::Escape)
  )
  {
	hasBeenConsumed = true;
	GetIEditor()->SetInGameMode(false);
  }

  InputChannelNotifications::OnInputChannelEvent(inputChannel, hasBeenConsumed);
}
```
What I am doing is, checking if the escape key is pressed during Game Play and stop it if the event occurs. I know this might be not a real solution and also not sure if this is only related to my System.

AMZN-alexpete:
## What does this PR do?

Fixes #16198 (Unable to leave game mode in Linux)
Fixes the cursor being hidden when the user closed the console after exiting game mode with the console open

In Linux all input events are consumed, but in Windows some key events were are allowed through.   This allowed QAction shortcuts like CTRL+N to be used.

Other issues uncovered:
- Any QAction can be run while in game mode including CTRL+S, CTRL+N etc and these should not be allowed, only the editor action to leave game mode.  Related GHI filed: https://github.com/o3de/o3de/issues/16253
- The in-editor console window will not work if we don't allow Qt events through,  I tested consuming input events on windows and confirmed that users can no longer type in the editor console when in game mode (after pressing the tilda key to reveal the cursor and let the user type in the console window).  Potentially we will need this behavior eventually if we want to let other windows have limited interaction so users can modify component properties while the game is running in the future.

## How was this PR tested?

- entered and exited game mode on Linux and Windows, verifying the console could be opened and interacted with when in game mode and out of game mode
- verified the cursor is visible when the console is open and isn't hidden when the console is closed in the editor when not in game mode.
- OSX was not tested, will see if I can get someone to check this for me


Code/Editor/GameEngine.cpp
```diff
@@ -794,6 +795,22 @@ void CGameEngine::Update()
...
+        // When the user opens the console, Qt events will be allowed 
+        // so the user can interact with limited Editor content like the console.
+        const AzFramework::InputChannel* inputChannel = nullptr;
+        const AzFramework::InputChannelId channelId(AzFramework::InputDeviceKeyboard::Key::Escape);
+        AzFramework::InputChannelRequestBus::EventResult(inputChannel, channelId, &AzFramework::InputChannelRequests::GetInputChannel);
+        if(inputChannel && inputChannel->GetState() == AzFramework::InputChannel::State::Began)
+        {
+            // leave game mode
+            RequestSetGameMode(false);
+        }
```
lumberyard-employee-dm, line 813: Is the input only processed once per frame?

Because if not, could the input channel state go from Began ->Updated before it is processed?

Gems/ImGui/Code/Source/ImGuiManager.cpp
```diff
@@ -729,10 +715,19 @@ void ImGuiManager::ToggleThroughImGuiVisibleState()
...
+            // This prevents incorrectly hiding the cursor when a user is in
+            // game mode, opens the console, exits game mode, and then hides the
+            // console resulting in the cursor being restored to what it was
+            // in game mode, usually making the cursor hidden in the editor.
+            if (m_previousSystemCursorStateInGame == inGame)
+            {
+                AzFramework::InputSystemCursorRequestBus::Event(AzFramework::InputDeviceMouse::Id,
+                    &AzFramework::InputSystemCursorRequests::SetSystemCursorState,
+                    m_previousSystemCursorState);
+            }
```
lumberyard-employee-dm, line 730: Since this is all for Editor game mode which is used for development, we should bias the behavior for showing a cursor when it shouldn't be shown as that is the safer approach.

lumberyard-employee-dm (approved):

Code/Editor/GameEngine.cpp
AMZN-alexpete: The input is only processed once per tick/frame at the beginning so we should always see the 'Began' state unless someone is doing something unusual with the input system.  I considered putting this block of code before the game ticks, but thought it made more sense to let the game application get to process the Escape key too before the Editor uses it to leave game mode.  Sometimes it's good to see your game code is able to detect the Escape key for testing even though in Editor it results in exiting game mode.  Ideally, we'd led the user choose the key bind for this instead of hardcoding it to Escape.

Gems/ImGui/Code/Source/ImGuiManager.cpp
AMZN-alexpete: I almost set this code to always show the cursor in Editor when not in game, but wasn't sure if maybe there's some case where the cursor is hidden and that's why the code was written this way?

Gems/ImGui/Code/Source/ImGuiManager.cpp
```diff
@@ -729,10 +715,19 @@ void ImGuiManager::ToggleThroughImGuiVisibleState()
         case DisplayState::Visible:
             m_clientMenuBarState = DisplayState::Hidden;
 
-            // Restore old cursor state
-            AzFramework::InputSystemCursorRequestBus::Event(AzFramework::InputDeviceMouse::Id,
-                &AzFramework::InputSystemCursorRequests::SetSystemCursorState,
-                m_previousSystemCursorState);
+            const bool inGame = !gEnv->IsEditor() || gEnv->IsEditorGameMode();
```
mbalfour-amzn, line 718: Suggestion: Move the definition of `inGame` outside of the switch statement, then in the DisplayState::Hidden case, you can set `m_previousSystemCursorStateInGame = inGame;`. This would make the logic a little bit easier to read because it provides explicit continuity between how the variable is set and how it's being checked.

Gems/ImGui/Code/Source/ImGuiManager.cpp
mbalfour-amzn: I can't think of a case where the cursor should remain hidden outside of game mode if in the Editor.

Suggestion: Instead of `m_previousSystemCursorStateInGame`, how about `m_inGameWhenImGuiActivated`? While this logic is only affecting the cursor, the actual logic check here is "did the game mode state change between when we made ImGui visible and when we made it hidden", so a name tied to the logic check instead of the action we want to take might make this a little bit more readable as well.

mbalfour-amzn (approved):

lumberyard-employee-dm:
You should update the summary of the PR to include MacOS as well

lumberyard-employee-dm (approved):

Gems/ImGui/Code/Source/ImGuiManager.cpp
AMZN-alexpete: I went ahead and updated this so we just never hide the cursor if we're not in game.

PR #16259, opened 2023-06-28
Add Auto Update CLI and Unit Test

VeryKaylin-player1:
## What does this PR do?

Add the --auto-update CLI option to repo-properties.py and include the associated unit test. This CLI should be capable of automatically detecting any Gems/Projects/Templates present in the remote repository folder but not registered with the repo.json file, then update that object in the user's repo.json file.

- User can specify which type of object they want to auto-update. The command for using this CLI is:  `o3de.bat edit-repo-properties -rp "C:\o3de_Projects\test2\repo.json" -au gem`
- you can use the last argument parameter with other objects
For example : `-au gem`, `-au project`, `-au template`, or `-au gem project template`

## How was this PR tested?

1. I added unit tests that made sure the _auto_update function got called and mocked that function to make sue the correct object gets updated in the repo.json file. Run all unit tests to make sure all functions still passes the tests.
2. Run above command and made sure that repo.json file gets updated with the correct object and data.

### DIFF
--- a/scripts/o3de/o3de/repo_properties.py
+++ b/scripts/o3de/o3de/repo_properties.py
@@ -395,7 +395,7 @@ def add_parser_args(parser):
     group.add_argument('--auto-update', '-au', type=str, nargs='*', required=False,
                        help='Checks for any new Gems/Projects/Templates in the associated directories that have not been added'
                        ' to your repo.json fields.'
-                       ' Pass in lower case object type as args that should look like this: gem/ project/ template'
+                       ' Optionally, provide the object types to update as args like this: --auto-update gem project template'
                        ' Note: This does not update the deleted gems/projects/templates, please use'
                         '--delete-gems if you want to delete data from repo.json file')
     
--- a/scripts/o3de/o3de/repo_properties.py
+++ b/scripts/o3de/o3de/repo_properties.py
@@ -265,8 +265,8 @@ def _edit_objects(object_typename:str,
     repo_json[f'{object_typename}s_data'] = repo_objects_data
 
 def _auto_update_json(object_type: str or list,
-                      repo_path: pathlib.Path = None,
-                      repo_json: dict = None,
+                      repo_path: pathlib.Path,
+                      repo_json: dict,
                       ):
     
     repo_directory = os.path.dirname(repo_path)
@@ -281,7 +281,8 @@ def _auto_update_json(object_type: str or list,
                 expected_files[filename].append(file_path)
     
     objects = object_type.split() if isinstance(object_type, str) else object_type
-    for object_name in objects:
+    for object in objects:
+        object_name = object.lower()
         if object_name == 'gem':
             _edit_objects(object_name, validation.valid_o3de_gem_json, repo_json, expected_files.get("gem.json"))
         if object_name == 'project':
@@ -319,7 +320,7 @@ def edit_repo_props(repo_path: pathlib.Path = None,
     :add_templates: Any template paths to be added to the list
     :delete_templates: Any template names to be removed from the list
     :replace_templates: A list of template paths that will completely replace the current list of path
-    :auto_update: Automatically updates your selected gem/project/template to your remote repository
+    :auto_update: List of object types (gem/project/template) to automatically to your remote repository. `None` if no auto update is desired.
     :release_archive_path: Path where you want your release to be located
     :force: Replaces current directory with new user input directory or file
     :download_prefix: The string prefix of the download uri
@@ -391,7 +392,7 @@ def add_parser_args(parser):
     group.add_argument('--repo-name','-rn',  type=str, required=False,
                        help='The name of the remote repository.')
     group.add_argument('--auto-update', '-au', type=str, nargs='*', required=False,
-                       help='Checks for any new Gems/Projects/Templates in the asscoiated directories that has not been added'
+                       help='Checks for any new Gems/Projects/Templates in the associated directories that have not been added'
                        ' to your repo.json fields.'
                        ' Pass in lower case object type as args that should look like this: gem/ project/ template'
                        ' Note: This does not update the deleted gems/projects/templates, please use'
--- a/scripts/o3de/o3de/repo_properties.py
+++ b/scripts/o3de/o3de/repo_properties.py
@@ -266,20 +266,22 @@ def _edit_objects(object_typename:str,
 
 def _auto_update_json(object_type: str or list,
                       repo_path: pathlib.Path,
-                      repo_json: dict,
-                      ):
+                      repo_json: dict):
     
     repo_directory = os.path.dirname(repo_path)
     expected_files = {"project.json": [],
                       "gem.json": [],
                       "template.json": []}
     for directory, sub_directory, files, in os.walk(repo_directory):
-        current_depth = directory[len(repo_directory):].count(os.sep)
+        found_object_json = False
         for filename in files:
-            if filename in expected_files.keys() and current_depth < 3:
+            if filename in expected_files.keys():
+                found_object_json = True
                 file_path = pathlib.Path(directory)
                 expected_files[filename].append(file_path)
-    
+        if found_object_json:
+            sub_directory.clear()
+
     objects = object_type.split() if isinstance(object_type, str) else object_type
     for object in objects:
         object_name = object.lower()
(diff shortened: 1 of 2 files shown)


scripts/o3de/o3de/repo_properties.py
```diff
@@ -352,7 +390,13 @@ def add_parser_args(parser):
     group = parser.add_argument_group('properties', 'arguments for modifying individual remote repo properties.')
     group.add_argument('--repo-name','-rn',  type=str, required=False,
                        help='The name of the remote repository.')
-    group = parser.add_mutually_exclusive_group(required=False)
+    group.add_argument('--auto-update', '-au', type=str, nargs='*', required=False,
+                       help='Checks for any new Gems/Projects/Templates in the asscoiated directories that has not been added'
```
spham-amzn, line 394: spelling
```suggestion
                       help='Checks for any new Gems/Projects/Templates in the associated directories that have not been added'
```

scripts/o3de/o3de/repo_properties.py
```diff
@@ -263,6 +264,32 @@ def _edit_objects(object_typename:str,
 
     repo_json[f'{object_typename}s_data'] = repo_objects_data
 
+def _auto_update_json(object_type: str or list,
+                      repo_path: pathlib.Path = None,
+                      repo_json: dict = None,
```
spham-amzn, line 269: These arguments should not have default values if they are required. I am assuming at least `repo_path` is required since the call `    repo_directory = os.path.dirname(repo_path)` below will raise a type error

scripts/o3de/o3de/repo_properties.py
```diff
@@ -263,6 +264,32 @@ def _edit_objects(object_typename:str,
...
+    for directory, sub_directory, files, in os.walk(repo_directory):
+        current_depth = directory[len(repo_directory):].count(os.sep)
+        for filename in files:
+            if filename in expected_files.keys() and current_depth < 3:
+                file_path = pathlib.Path(directory)
+                expected_files[filename].append(file_path)
+    
+    objects = object_type.split() if isinstance(object_type, str) else object_type
+    for object_name in objects:
+        if object_name == 'gem':
```
spham-amzn, line 285: if you do `.lower()` to `object_name`, then the auto_update list doesn't have to be case sensitive and you can update documentation for the argument.
```suggestion
        if object_name.lower() == 'gem':
```

scripts/o3de/o3de/repo_properties.py
```diff
@@ -291,7 +319,9 @@ def edit_repo_props(repo_path: pathlib.Path = None,
     :add_templates: Any template paths to be added to the list
     :delete_templates: Any template names to be removed from the list
     :replace_templates: A list of template paths that will completely replace the current list of path
+    :auto_update: Automatically updates your selected gem/project/template to your remote repository
```
spham-amzn, line 322: This description isn't clear. Looking at it I would assume that this is a boolean argument, which it clearly isn't. 
```suggestion
    :auto_update: List of object types (gem/project/template) to automatically to your remote repository. `None` if no auto update is desired.
```

scripts/o3de/o3de/repo_properties.py
VeryKaylin-player1: This makes the function a lot more robust 👍

scripts/o3de/o3de/repo_properties.py
```diff
@@ -263,6 +264,33 @@ def _edit_objects(object_typename:str,
 
     repo_json[f'{object_typename}s_data'] = repo_objects_data
 
+def _auto_update_json(object_type: str or list,
+                      repo_path: pathlib.Path,
+                      repo_json: dict,
```
AMZN-alexpete, line 269:
```suggestion
                      repo_json: dict
```

scripts/o3de/o3de/repo_properties.py
```diff
@@ -263,6 +264,33 @@ def _edit_objects(object_typename:str,
...
+                      ):
+    
+    repo_directory = os.path.dirname(repo_path)
+    expected_files = {"project.json": [],
+                      "gem.json": [],
+                      "template.json": []}
+    for directory, sub_directory, files, in os.walk(repo_directory):
+        current_depth = directory[len(repo_directory):].count(os.sep)
+        for filename in files:
+            if filename in expected_files.keys() and current_depth < 3:
```
AMZN-alexpete, line 279: I think we should change this to not default to looking for gems inside gems or gems inside projects because then you have to add special logic to ignore the project gem, and also ignore gems or projects inside templates...

I think it's best to change this logic to walk the directories until we find a gem.json, project.json, template.json or there are no more subdirectories to look in.  If we find a valid .json file don't look in any more subdirectories in that path (don't look for gems, projects, templates inside the current directory)

scripts/o3de/o3de/repo_properties.py
```diff
@@ -352,7 +391,13 @@ def add_parser_args(parser):
     group = parser.add_argument_group('properties', 'arguments for modifying individual remote repo properties.')
     group.add_argument('--repo-name','-rn',  type=str, required=False,
                        help='The name of the remote repository.')
-    group = parser.add_mutually_exclusive_group(required=False)
+    group.add_argument('--auto-update', '-au', type=str, nargs='*', required=False,
+                       help='Checks for any new Gems/Projects/Templates in the associated directories that have not been added'
+                       ' to your repo.json fields.'
+                       ' Pass in lower case object type as args that should look like this: gem/ project/ template'
+                       ' Note: This does not update the deleted gems/projects/templates, please use'
+                        '--delete-gems if you want to delete data from repo.json file')
```
AMZN-alexpete, line 399: Can we default to updating gems projects and templates if they don't specify the object types?  That seems like it'd be the most convenient/common use.
```suggestion
    group.add_argument('--auto-update', '-au', type=str, nargs='*', default='gem project template', required=False,
                       help='Checks for any new Gems/Projects/Templates in the associated directories that have not been added'
                       ' to your repo.json fields.'
                       ' Updates all object types if no object types are provided.  (e.g. --auto-update). '
                       ' If object types are provided, auto update will  limit the update to those specific object types. e.g. --auto-update gem project  (only update Gems and Projects, not Templates)'
                       ' Note: This does not update the deleted gems/projects/templates, please use'
                        '--delete-gems if you want to delete data from repo.json file')
```

scripts/o3de/tests/test_repo_properties.py
```diff
@@ -379,18 +441,34 @@ def get_json_data(object_typename: str,
                 return {'project_name': object_path}
             if object_typename == "template":
                 return {'template_name': object_path}
+            
+        def _auto_update_json(object_type: str or list = None,
+                      repo_path: pathlib.Path = None,
+                      repo_json: dict = None,
```
AMZN-alexpete, line 447:
```suggestion
                      repo_json: dict = None
```

scripts/o3de/o3de/repo_properties.py
```diff
@@ -263,6 +264,35 @@ def _edit_objects(object_typename:str,
...
+                      "template.json": []}
+    for directory, sub_directory, files, in os.walk(repo_directory):
+        found_object_json = False
+        for filename in files:
+            if filename in expected_files.keys():
+                found_object_json = True
+                file_path = pathlib.Path(directory)
+                expected_files[filename].append(file_path)
+        if found_object_json:
+            sub_directory.clear()
```
AMZN-alexpete, line 283: does this prevent the logic from looking into other sub directories once we've found one of the .json files?

scripts/o3de/o3de/repo_properties.py
```diff
@@ -352,7 +392,13 @@ def add_parser_args(parser):
     group = parser.add_argument_group('properties', 'arguments for modifying individual remote repo properties.')
     group.add_argument('--repo-name','-rn',  type=str, required=False,
                        help='The name of the remote repository.')
-    group = parser.add_mutually_exclusive_group(required=False)
+    group.add_argument('--auto-update', '-au', type=str, nargs='*', required=False,
+                       help='Checks for any new Gems/Projects/Templates in the associated directories that have not been added'
+                       ' to your repo.json fields.'
+                       ' Pass in lower case object type as args that should look like this: gem/ project/ template'
```
AMZN-alexpete, line 398: we don't really need the forward slashes to pass multiple args do we?
```suggestion
                       ' Optionally, provide the object types to update as args like this: --auto update gem project template'
```

scripts/o3de/o3de/repo_properties.py
VeryKaylin-player1: The Flag statement: 'found_object_json = True' would trigger this sub_directory.clear() so it stops looking at the rest of the directories when the object.json file is found.

AMZN-alexpete (approved):

spham-amzn (approved):

PR #16269, opened 2023-06-28
Feature: Use the PipelineViewTag of a parent pass in the pipeline hierarchy

kh-huawei:
## What does this PR do?

Currently many passes have a hardcoded `"PipelineViewTag" : "MainCamera"` set in the pass template. This makes it somewhat impossible to use existing pass-templates in a context where more than one active Camera exists. Also, many post-processing passes (e.g. `TaaPass.cpp`) use `GetDefaultView()` of the pipeline to get access to a view, since they do not want to bind a View-SRG, but need access to a view. 

This PR now splits the functionality of the `"PipelineViewTag"` into two separate concepts: 
  - The first one is determining which view-tag is relevant for the current pass
    - either set explicitly for each pass or taken from the pass hierarchy, ending with the Main view-tag of the pipeline.
  - The second one is binding the View-SRG. 

The concrete changes for this are:
- Move the `PipelineViewTag` from the `RenderPassData` to the `PassData` struct, so parent-passes can also specify a view-tag relevant for their child-passes.
    - This should not change the json-structure at all, but it does mean assets need to be reprocessed 
- Moved the `m_viewTag` from the `RenderPass` class to the `Pass` class.
- Changed the function `GetPipelineViewTag()` so it always returns a view-tag:
    - from the current pass, if set explicitly
    - recursively from the first parent pass that has an explicitly set view tag 
    - the `MainViewTag` of the pipeline, if the hierarchy is at the root pass
- Added the flag `BindViewSrg` to the `RenderPassData`, to explicitly mark that the shader needs a view-Srg.
- Renamed the flag `m_flags.m_hasPipelineViewTag` to `m_flags.m_bindViewSrg` 
    - this is set to true if the pass has an explicitly declared view tag, so the behaviour should stay unchanged for exising passes
- Renamed all `"PipelineViewTag" : "MainCamera"` to `"BindViewSrg": true` via search/replace of all .pass files
- Added `GetFirstView(viewTag)` to the `RenderPipeline` class
- Changed the usage of `GetDefaultView()` in all pass-classes to `GetFirstView(GetPipelineViewTag())`

This is the first solution we had for our issue. It works as is, but i am welcoming feedback on how to improve this.

## How was this PR tested?

- Confirmed that the Main Pipeline still works unchanged
- Use a single pipeline with two cameras, where each camera has its own post-processing passes

### DIFF
--- a/Gems/Atom/Feature/Common/Code/Source/PostProcessing/BlendColorGradingLutsPass.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/PostProcessing/BlendColorGradingLutsPass.cpp
@@ -417,7 +417,7 @@ namespace AZ
             if (scene)
             {
                 PostProcessFeatureProcessor* fp = scene->GetFeatureProcessor<PostProcessFeatureProcessor>();
-                AZ::RPI::ViewPtr view = scene->GetDefaultRenderPipeline()->GetDefaultView();
+                AZ::RPI::ViewPtr view = m_pipeline->GetFirstView(GetPipelineViewTag());
                 if (fp)
                 {
                     PostProcessSettings* postProcessSettings = fp->GetLevelSettingsFromView(view);
--- a/Gems/Atom/Feature/Common/Code/Source/PostProcessing/BloomBlurPass.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/PostProcessing/BloomBlurPass.cpp
@@ -109,7 +109,7 @@ namespace AZ
 
             RPI::Scene* scene = GetScene();
             PostProcessFeatureProcessor* fp = scene->GetFeatureProcessor<PostProcessFeatureProcessor>();
-            RPI::ViewPtr view = scene->GetDefaultRenderPipeline()->GetDefaultView();
+            RPI::ViewPtr view = m_pipeline->GetFirstView(GetPipelineViewTag());
             if (fp)
             {
                 PostProcessSettings* postProcessSettings = fp->GetLevelSettingsFromView(view);
--- a/Gems/Atom/Feature/Common/Code/Source/PostProcessing/BloomCompositePass.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/PostProcessing/BloomCompositePass.cpp
@@ -94,7 +94,7 @@ namespace AZ
  
             RPI::Scene* scene = GetScene();
             PostProcessFeatureProcessor* fp = scene->GetFeatureProcessor<PostProcessFeatureProcessor>();
-            RPI::ViewPtr view = scene->GetDefaultRenderPipeline()->GetDefaultView();
+            RPI::ViewPtr view = m_pipeline->GetFirstView(GetPipelineViewTag());
             if (fp)
             {
                 PostProcessSettings* postProcessSettings = fp->GetLevelSettingsFromView(view);
--- a/Gems/Atom/Feature/Common/Code/Source/PostProcessing/BloomDownsamplePass.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/PostProcessing/BloomDownsamplePass.cpp
@@ -92,7 +92,7 @@ namespace AZ
         {
             RPI::Scene* scene = GetScene();
             PostProcessFeatureProcessor* fp = scene->GetFeatureProcessor<PostProcessFeatureProcessor>();
-            RPI::ViewPtr view = scene->GetDefaultRenderPipeline()->GetDefaultView();
+            RPI::ViewPtr view = m_pipeline->GetFirstView(GetPipelineViewTag());
             if (fp)
             {
                 PostProcessSettings* postProcessSettings = fp->GetLevelSettingsFromView(view);
--- a/Gems/Atom/Feature/Common/Code/Source/PostProcessing/ChromaticAberrationPass.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/PostProcessing/ChromaticAberrationPass.cpp
@@ -38,11 +38,11 @@ namespace AZ
                 return false;
             }
             PostProcessFeatureProcessor* fp = scene->GetFeatureProcessor<PostProcessFeatureProcessor>();
-            const RPI::ViewPtr view = GetRenderPipeline()->GetDefaultView();
             if (!fp)
             {
                 return false;
             }
+            const RPI::ViewPtr view = GetRenderPipeline()->GetFirstView(GetPipelineViewTag());
             PostProcessSettings* postProcessSettings = fp->GetLevelSettingsFromView(view);
             if (!postProcessSettings)
             {
@@ -71,7 +71,7 @@ namespace AZ
             PostProcessFeatureProcessor* fp = scene->GetFeatureProcessor<PostProcessFeatureProcessor>();
             if (fp)
             {
-                RPI::ViewPtr view = scene->GetDefaultRenderPipeline()->GetDefaultView();
+                RPI::ViewPtr view = m_pipeline->GetFirstView(GetPipelineViewTag());
                 PostProcessSettings* postProcessSettings = fp->GetLevelSettingsFromView(view);
                 if (postProcessSettings)
                 {
--- a/Gems/Atom/Feature/Common/Code/Source/PostProcessing/HDRColorGradingPass.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/PostProcessing/HDRColorGradingPass.cpp
@@ -128,7 +128,7 @@
             if (scene)
             {
                 PostProcessFeatureProcessor* fp = scene->GetFeatureProcessor<PostProcessFeatureProcessor>();
-                AZ::RPI::ViewPtr view = scene->GetDefaultRenderPipeline()->GetDefaultView();
+                AZ::RPI::ViewPtr view = m_pipeline->GetFirstView(GetPipelineViewTag());
                 if (fp)
                 {
                     PostProcessSettings* postProcessSettings = fp->GetLevelSettingsFromView(view);
--- a/Gems/Atom/Feature/Common/Code/Source/PostProcessing/LookModificationCompositePass.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/PostProcessing/LookModificationCompositePass.cpp
@@ -132,7 +132,7 @@ namespace AZ
             if (scene)
             {
                 PostProcessFeatureProcessor* fp = scene->GetFeatureProcessor<PostProcessFeatureProcessor>();
-                AZ::RPI::ViewPtr view = scene->GetDefaultRenderPipeline()->GetDefaultView();
+                AZ::RPI::ViewPtr view = m_pipeline->GetFirstView(GetPipelineViewTag());
                 if (fp)
                 {
                     PostProcessSettings* postProcessSettings = fp->GetLevelSettingsFromView(view);
@@ -157,7 +157,7 @@ namespace AZ
             if (scene)
             {
                 PostProcessFeatureProcessor* fp = scene->GetFeatureProcessor<PostProcessFeatureProcessor>();
-                AZ::RPI::ViewPtr view = scene->GetDefaultRenderPipeline()->GetDefaultView();
+                AZ::RPI::ViewPtr view = m_pipeline->GetFirstView(GetPipelineViewTag());
                 if (fp)
                 {
                     PostProcessSettings* postProcessSettings = fp->GetLevelSettingsFromView(view);
--- a/Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Pass/ParentPass.h
+++ b/Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Pass/ParentPass.h
@@ -7,6 +7,7 @@
  */
 #pragma once
 
+#include <Atom/RPI.Public/Base.h>
 #include <Atom/RPI.Public/Pass/Pass.h>
 
 namespace AZ
--- a/Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Pass/Pass.h
+++ b/Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Pass/Pass.h
@@ -153,7 +153,10 @@ namespace AZ
             virtual bool IsEnabled() const { return m_flags.m_enabled; }
 
             bool HasDrawListTag() const { return m_flags.m_hasDrawListTag; }
-            bool HasPipelineViewTag() const { return m_flags.m_hasPipelineViewTag; }
+            bool BindViewSrg() const
+            {
+                return m_flags.m_bindViewSrg;
+            }
 
             // Searches this pass's attachment bindings for one with the provided Name (nullptr if none found)
             PassAttachmentBinding* FindAttachmentBinding(const Name& slotName);
@@ -422,6 +425,9 @@ namespace AZ
             // Pointer to the parent pass if this pass is a child pass
             ParentPass* m_parent = nullptr;
 
+            // View tag used to associate a pipeline view for this pass or its child-passes.
+            PipelineViewTag m_viewTag;
+
             struct
             {
                 union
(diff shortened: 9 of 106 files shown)


Gems/Atom/RPI/Code/Source/RPI.Public/Pass/RenderPass.cpp
```diff
@@ -47,6 +52,12 @@ namespace AZ
...
         const PipelineViewTag& RenderPass::GetPipelineViewTag() const
         {
+            if (m_viewTag.IsEmpty())
+            {
+                // This returns either the view-tag of the parent, the main ViewTag of the Pipeline, or a reference to a static but empty
+                // viewTag.
+                return Pass::GetPipelineViewTag();
+            }
             return m_viewTag;
         }
```
VickyAtAZ, line 62: I think you can just remove this function so it will just use Pass::GetPipelineViewTag();

Gems/Atom/Feature/Common/Code/Source/PostProcessing/BlendColorGradingLutsPass.cpp
```diff
@@ -417,7 +417,7 @@ namespace AZ
             if (scene)
             {
                 PostProcessFeatureProcessor* fp = scene->GetFeatureProcessor<PostProcessFeatureProcessor>();
-                AZ::RPI::ViewPtr view = scene->GetDefaultRenderPipeline()->GetDefaultView();
+                AZ::RPI::ViewPtr view = scene->GetDefaultRenderPipeline()->GetFirstView(GetPipelineViewTag());
```
VickyAtAZ, line 420: Maybe irrelevant to your PR, it probably shouldn't use GetDefaultRenderPipeline() but use m_pipeline

Gems/Atom/Feature/Common/Code/Source/PostProcessing/BloomBlurPass.cpp
```diff
@@ -109,7 +109,7 @@ namespace AZ
 
             RPI::Scene* scene = GetScene();
             PostProcessFeatureProcessor* fp = scene->GetFeatureProcessor<PostProcessFeatureProcessor>();
-            RPI::ViewPtr view = scene->GetDefaultRenderPipeline()->GetDefaultView();
+            RPI::ViewPtr view = scene->GetDefaultRenderPipeline()->GetFirstView(GetPipelineViewTag());
```
VickyAtAZ, line 112: Maybe irrelevant to your PR, it probably shouldn't use GetDefaultRenderPipeline() but use m_pipeline

Gems/Atom/Feature/Common/Code/Source/PostProcessing/BloomCompositePass.cpp
```diff
@@ -94,7 +94,7 @@ namespace AZ
  
             RPI::Scene* scene = GetScene();
             PostProcessFeatureProcessor* fp = scene->GetFeatureProcessor<PostProcessFeatureProcessor>();
-            RPI::ViewPtr view = scene->GetDefaultRenderPipeline()->GetDefaultView();
+            RPI::ViewPtr view = scene->GetDefaultRenderPipeline()->GetFirstView(GetPipelineViewTag());
```
VickyAtAZ, line 97: Maybe irrelevant to your PR, it probably shouldn't use GetDefaultRenderPipeline() but use m_pipeline

Gems/Atom/Feature/Common/Code/Source/PostProcessing/BloomDownsamplePass.cpp
```diff
@@ -92,7 +92,7 @@ namespace AZ
         {
             RPI::Scene* scene = GetScene();
             PostProcessFeatureProcessor* fp = scene->GetFeatureProcessor<PostProcessFeatureProcessor>();
-            RPI::ViewPtr view = scene->GetDefaultRenderPipeline()->GetDefaultView();
+            RPI::ViewPtr view = scene->GetDefaultRenderPipeline()->GetFirstView(GetPipelineViewTag());
```
VickyAtAZ, line 95: Maybe irrelevant to your PR, it probably shouldn't use GetDefaultRenderPipeline() but use m_pipeline

Gems/Atom/Feature/Common/Code/Source/PostProcessing/ChromaticAberrationPass.cpp
```diff
@@ -71,7 +71,7 @@ namespace AZ
             PostProcessFeatureProcessor* fp = scene->GetFeatureProcessor<PostProcessFeatureProcessor>();
             if (fp)
             {
-                RPI::ViewPtr view = scene->GetDefaultRenderPipeline()->GetDefaultView();
+                RPI::ViewPtr view = scene->GetDefaultRenderPipeline()->GetFirstView(GetPipelineViewTag());
```
VickyAtAZ, line 74: Maybe irrelevant to your PR, it probably shouldn't use GetDefaultRenderPipeline() but use m_pipeline

Gems/Atom/Feature/Common/Code/Source/PostProcessing/HDRColorGradingPass.cpp
```diff
@@ -128,7 +128,7 @@
             if (scene)
             {
                 PostProcessFeatureProcessor* fp = scene->GetFeatureProcessor<PostProcessFeatureProcessor>();
-                AZ::RPI::ViewPtr view = scene->GetDefaultRenderPipeline()->GetDefaultView();
+                AZ::RPI::ViewPtr view = scene->GetDefaultRenderPipeline()->GetFirstView(GetPipelineViewTag());
```
VickyAtAZ, line 131: Maybe irrelevant to your PR, it probably shouldn't use scene->GetDefaultRenderPipeline() but use m_pipeline (or GetRenderPipeline()

Gems/Atom/Feature/Common/Code/Source/PostProcessing/LookModificationCompositePass.cpp
```diff
@@ -132,7 +132,7 @@ namespace AZ
             if (scene)
             {
                 PostProcessFeatureProcessor* fp = scene->GetFeatureProcessor<PostProcessFeatureProcessor>();
-                AZ::RPI::ViewPtr view = scene->GetDefaultRenderPipeline()->GetDefaultView();
+                AZ::RPI::ViewPtr view = scene->GetDefaultRenderPipeline()->GetFirstView(GetPipelineViewTag());
```
VickyAtAZ, line 135: Maybe irrelevant to your PR, it probably shouldn't use scene->GetDefaultRenderPipeline() but use m_pipeline (or GetRenderPipeline()

Gems/Atom/Feature/Common/Code/Source/PostProcessing/LookModificationCompositePass.cpp
```diff
@@ -157,7 +157,7 @@ namespace AZ
...
```
VickyAtAZ, line 160: Maybe irrelevant to your PR, it probably shouldn't use scene->GetDefaultRenderPipeline() but use m_pipeline (or GetRenderPipeline()

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Pass/ParentPass.h
```diff
@@ -7,6 +7,7 @@
  */
 #pragma once
 
+#include "Atom/RPI.Public/Base.h"
```
VickyAtAZ, line 10: #include <Atom/RPI.Public/Base.h>

Gems/Atom/RPI/Code/Source/RPI.Public/Pass/RenderPass.cpp
```diff
@@ -367,9 +378,14 @@ namespace AZ
...
-                const AZStd::vector<ViewPtr>& views = m_pipeline->GetViews(m_viewTag);
+                const auto& viewTag = GetPipelineViewTag();
+                if (viewTag.IsEmpty())
+                {
+                    return nullptr;
+                }
+                const AZStd::vector<ViewPtr>& views = m_pipeline->GetViews(viewTag);
                 if (views.size() > 0)
                 {
                     return views[0];
```
VickyAtAZ, line 391: This block can be replaced by m_pipeline->GetFirstView(viewTag)

Gems/Atom/RPI/Code/Source/RPI.Public/RenderPipeline.cpp
```diff
@@ -6,7 +6,7 @@
  *
  */
 
-
+#include "Atom/RPI.Public/Base.h"
```
VickyAtAZ, line 9: #include <Atom/RPI.Public/Base.h>

and move it to above this line
#include <Atom/RPI.Public/Pass/PassFilter.h>

Gems/Atom/RPI/Code/Source/RPI.Public/RenderPipeline.cpp
```diff
@@ -423,13 +423,22 @@ namespace AZ
...
         ViewPtr RenderPipeline::GetDefaultView()
         {
-            ViewPtr defaultView;
             const AZStd::vector<ViewPtr>& views = GetViews(m_mainViewTag);
             if (!views.empty())
             {
-                defaultView = views[0];
+                return views[0];
+            }
+            return {};
```
VickyAtAZ, line 431: return GetFirstView(m_mainViewTag);

VickyAtAZ (changes requested):
I like the idea of splitting the functionality of PipelineViewTag. 
There are only some small issues need to be addressed. 

Also, you can change the version number of the PassBuilder so the asset rebuilding will be triggered automatically.
The version number is at line 56
of o3de\Gems\Atom\RPI\Code\Source\RPI.Builders\Pass\PassBuilder.cpp file

builder.m_version = 15; // Changed dependency type to OrderOnce

You may change it to 

builder.m_version = 16; // Move PipelineViewTag from RenderPassData to PassData; add BindViewTag to RenderPassData

Note: we update the comment after the version number to avoid two commits have same version number change at same time so it can be detected by merge conflict.

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/RenderPipeline.h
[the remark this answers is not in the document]
kh-huawei: This is surprisingly difficult:
I see ~360 changed lines for this file, and as I interpret this, the file has CRLF in the repo, which my git automagically converts to LF upon commit, regardless of what I'm using locally (seemingly with core.autocrlf=true). 

Do I understand that right that the file has CRLF in the repo? 
If yes, why? 
Wouldn't a basic 'LF' in the repo work better? 

Also, how?
While I can change the file ending type for a single file in VS Code, that change doesn't register as a modification in git status, which means I apparently can't commit only line ending changes?

Gems/Atom/Feature/Common/Code/Source/PostProcessing/BlendColorGradingLutsPass.cpp
kh-huawei: changed, here and for the other instances as well

Gems/Atom/RPI/Code/Source/RPI.Public/Pass/RenderPass.cpp
kh-huawei: removed

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Pass/ParentPass.h
kh-huawei: done

Gems/Atom/RPI/Code/Source/RPI.Public/Pass/RenderPass.cpp
kh-huawei: done

Gems/Atom/RPI/Code/Source/RPI.Public/RenderPipeline.cpp
kh-huawei: done

Gems/Atom/RPI/Code/Source/RPI.Public/RenderPipeline.cpp
kh-huawei: done

VickyAtAZ (commented):
This change can be a breaking change since it requires people to update their pass files. The pass building process won't fail since "PipelinViewTag" is part of PassData now and all the other pass data types are derived from PassData. But the m_bindViewSrg won't be set true for them. And this could lead to runtime crash due to ViewSrg wasn't bound. 
Unfortunately we don't have pass asset versioning support in place. 

I would suggest two things be done here:
1. Update PassBuilder code to handle the cases user may have not updated their pass files. For example, if the pass data is a RenderPassData or its derived classes, and it has PipelineViewTag set to MainCamera and m_bindViewSrg is false, then we may assume it's an old data format and update the m_bindViewSrg to true. Maybe it's safe to set PipelineViewTag to empty too. 
2. Send an [Impactful Change] message to slack channel.

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/RenderPipeline.h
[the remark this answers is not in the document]
kh-huawei: created #16344 to change the CRLF to LF for this file, which should be merged before this.

kh-huawei:
I've updated the PassBuilder to write a deprecation warning when the json contains `"PipelineViewTag" : "MainCamera"`, and i'm setting `BindViewSrg` to true if any view-tag is provided.

Gems/Atom/RPI/Code/Source/RPI.Builders/Pass/PassBuilder.cpp
```diff
@@ -329,6 +331,36 @@ namespace AZ
...
+                    else if (!passData->m_bindViewSrg)
+                    {
+                        // Explicitly set "PipelineViewTag": implicitly set BindViewSrg to true as well, if it isn't yet.
+                        AZ_Info(
+                            PassBuilderName,
+                            "Asset %s: Pass has explicit PipelineViewTag '%s' -> setting \"BindViewSrg\" to true as well.",
+                            request.m_fullPath.c_str(),
+                            passData->m_pipelineViewTag.GetCStr());
+                        passData->m_bindViewSrg = true;
+                    }
```
VickyAtAZ, line 360: The logic here doesn't set m_bindViewSrg to true if  m_pipelineViewTag was set to "MainCamera" for RenderPass.
Is this correct?

VickyAtAZ (commented):
One small issue

Gems/Atom/RPI/Code/Source/RPI.Builders/Pass/PassBuilder.cpp
kh-huawei: good catch. Fixed it.

VickyAtAZ (approved):

moudgils (approved):
Looks good. Dont forget to send an Impactful Change message.

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Pass/Pass.h
```diff
@@ -153,7 +153,10 @@ namespace AZ
             virtual bool IsEnabled() const { return m_flags.m_enabled; }
 
             bool HasDrawListTag() const { return m_flags.m_hasDrawListTag; }
-            bool HasPipelineViewTag() const { return m_flags.m_hasPipelineViewTag; }
+            bool BindViewSrg() const
+            {
+                return m_flags.m_bindViewSrg;
+            }
```
AMZN-alexpete, line 159: This looks like a breaking public API change - I think we should keep the old function for deprecation purposes for one release - and increment the minor gem version in `Gems/Atom/RPI/gem.json` from `0.0.0` to `1.0.0`

```suggestion
            // Deprecated - replaced with BindViewSrg()
            bool HasPipelineViewTag() const { return BindViewSrg(); }
            bool BindViewSrg() const
            {
                return m_flags.m_bindViewSrg;
            }
```

PR #16284, opened 2023-06-29
Fix AB Inspector Image Previewer Scaling

### LINKED ISSUE #16250: Resize Asset Browser Inspector Previewer Image
The space given to the image previewer for the asset browser inspector can be shrunk or enlarged, but the image itself does not resize along with the space. I would like the image to scale along with the space available to it, while keeping its aspect ratio.

dtamkin1:
## What does this PR do?

Allow previewer images to be resized to fit the space available, while keeping their aspect ratio.

Fixes #16250 

Editor_H7v08Dx80h

## How was this PR tested?

Locally


alexmontAmazon (approved):

cgalvan:
This seems like it would be a generally useful widget. We have something similar in `AzQtComponents::ExtendedLabel`. That one currently ignores the aspect ratio, but might be worth investigating to change it to keep aspect ratio (which I believe would be desired) and then see if you can re-use it for your purposes.

cgalvan (approved):

PR #16285, opened 2023-06-29
creating a standardized export script based on MultiplayerSample.

tkothadev:
## What does this PR do?
This PR introduces the first ExportScript for O3DE, using the project export CLI and API functionality. This is modeled after the export script available for Multiplayer Sample (https://github.com/o3de/o3de-multiplayersample/pull/433). This script includes the following new arguments:

```
parser.add_argument('-out', '--output-path', type=pathlib.Path, required=True, help='Path that describes the final resulting Release Directory path location.')
parser.add_argument('-cfg', '--config', type=str, default='profile', choices=['release', 'profile'],
                        help='The CMake build configuration to use when building project binaries. If tool binaries are built with this script, they will use profile mode.')
parser.add_argument('-aof', '--archive-output-format',
                        type=str,
                        help="Format of archive to create from the output directory",
                        choices=["zip", "gzip", "bz2", "xz"], default="zip")
parser.add_argument('-sl', '--seedlist', type=pathlib.Path, dest='seedlist_paths', action='append',
                        help='Path to a seed list file for asset bundling. Specify multiple times for each seed list.')
parser.add_argument('-gf', '--gem-folder-name-to-copy', type=str, dest='gem_folder_names_to_copy', action='append',
                        help="The name of a Gem folder to include in release. Only specify folder name. This should match what is in the project directory.")
parser.add_argument('-pfp', '--project-file_pattern-to-copy', type=str, dest='project_file_patterns_to_copy', action='append',
                        help="Any additional file patterns located in the project directory. File patterns will be relative to the project path.")
parser.add_argument('-bnmt', '--build-non-mono-tools', action='store_true')
parser.add_argument('-nmbp', '--non-mono-build-path', type=pathlib.Path, default=None)
parser.add_argument('-mbp', '--mono-build-path', type=pathlib.Path, default=None)
parser.add_argument('-nogame', '--no-game-launcher', action='store_true', help='this flag skips building the Game Launcher on a platform if not needed.')
parser.add_argument('-noserver', '--no-server-launcher', action='store_true', help='this flag skips building the Server Launcher on a platform if not needed.')
parser.add_argument('-nounified', '--no-unified-launcher', action='store_true', help='this flag skips building the Unified Launcher on a platform if not needed.')
parser.add_argument('-pl', '--platform', type=str, default=host_asset_platform, choices=['pc', 'linux', 'mac'])
parser.add_argument('-a', '--archive-output', action='store_true', help='This option places the final output of the build into a compressed archive')
parser.add_argument('-q', '--quiet', action='store_true', help='Suppresses logging information unless an error occurs.')
```

## How was this PR tested?
The script was tested using Windows 10 to export O3DE Multiplayer sample in release mode. The following command was used:
```
C:\workspace\mps-test-build\o3de\scripts\o3de.bat export-project -es C:\workspace\mps-test-build\o3de\scripts\o3de\ExportScripts\export_standalone_monolithic_engine_centric.py -pp C:\workspace\mps-test-build\o3de-multiplayersample -ll INFO -out C:\workspace\mps-test-build\output -bnmt -cfg release -a -aof zip -sl C:\workspace\mps-test-build\o3de-multiplayersample\AssetBundling\SeedLists\BasePopcornFxSeedList.seed -sl C:\workspace\mps-test-build\o3de-multiplayersample\AssetBundling\SeedLists\GameSeedList.seed -sl C:\workspace\mps-test-build\o3de-multiplayersample\AssetBundling\SeedLists\VFXSeedList.seed -gf AWSCore -pfp launch_*.*
```

Afterwards, MPS was tested using the following commands:
```
.\MultiplayerSample.ServerLauncher.exe --console-command-file=launch_server.cfg --rhi=null -NullRenderer -bg_ConnectToAssetProcessor=0 -sys_PakPriority=2 -sv_terminateOnPlayerExit=false
```
```
.\MultiplayerSample.ServerLauncher.exe --console-command-file=launch_server.cfg --rhi=null -NullRenderer -bg_ConnectToAssetProcessor=0 -sys_PakPriority=2 -sv_terminateOnPlayerExit=false
```

The game was successfully launched.


scripts/o3de/o3de/export_project.py
```diff
@@ -49,6 +50,17 @@ def args(self) -> list:
         return self._args
 
 # Helper API
+def determine_host_asset_platform():
```
AMZN-Gene, line 53: why have this mapping? Iinstruct users to use "windows", "linux", or "darwin" when providing the exporting platform, and keep the O3D standard :)

AMZN-Gene (approved):

scripts/o3de/o3de/export_project.py
tkothadev: Currently that is used to determine a default asset platform on the user's machine if no argument is supplied to the export script.
```
parser.add_argument('-pl', '--platform', type=str, default=host_asset_platform, choices=['pc', 'linux', 'mac'])
```
So if that argument is not supplied, the assumption is that the user wants to build it on their own machine. Doing it this way would make it cross platform for each host that o3de supports (the technique was suggested by @lumberyard-employee-dm ). As for O3DE standard, if I understand correctly, `[pc, linux, mac]` are currently the o3de standard internally.

AMZN-alexpete: Maybe it'd be clearer if you rename the function to be something like `get_default_asset_pllatform`

AMZN-alexpete (commented):
how many export scripts are we expecting to have?  As a user, I think I'd prefer if we could just have one export script - a default - that can handle engine centric or project centric, and if I need to  do something fancy, I can make my own export script.  If this is the case, then I would name this file, `default_project_export.py` - it can also be the default export script the Project Manager uses.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
```diff
@@ -0,0 +1,161 @@
...
+import sys
+import time
+
+from o3de.export_project import process_command
+from o3de import manifest
+
+project_json_data = manifest.get_project_json_data(project_path=o3de_context.project_path)
+project_name = project_json_data.get('project_name')
+host_asset_platform = exp.determine_host_asset_platform()
+parser = argparse.ArgumentParser(prog='Exporter for standalone builds',
```
AMZN-alexpete, line 25: What other kinds of builds are we thinking of exporting besides standalone?

scripts/o3de/ExportScripts/export_standalone_monolithic_engine_centric.py
```diff
@@ -0,0 +1,160 @@
...
+os.makedirs(output_cache_path, exist_ok=True)
+
+# Build project assets and the engine (non-monolithic)
+if args.build_non_mono_tools:
+    process_command(['cmake', '-S', o3de_context.engine_path, '-B', str(non_mono_build_path), '-DLY_MONOLITHIC_GAME=0', f'-DLY_PROJECTS={o3de_context.project_path}'],
+                    cwd=o3de_context.engine_path)
+    process_command(['cmake', '--build', str(non_mono_build_path), '--target', 'AssetBundler', 'AssetBundlerBatch', 'AssetProcessor', 'AssetProcessorBatch', '--config','profile'],
+                    cwd=o3de_context.engine_path)
+    process_command(['cmake', '--build', str(non_mono_build_path), '--target', f'{project_name}.Assets', '--config', 'profile'],
+                    cwd=o3de_context.engine_path)
```
spham-amzn, line 79: You shouldn't need do do this in 2 different cmake commands, you should be able to specify all the targets including`${project_name}.Assets`. (And also, you probably don't need `AssetProcessor`. `AssetProcessorBatch` should be enough to compile the assets.
```suggestion
    process_command(['cmake', '--build', str(non_mono_build_path), '--target', 'AssetBundler', 'AssetBundlerBatch', 'AssetProcessor', 'AssetProcessorBatch', f'{project_name}.Assets', '--config','profile'],
                    cwd=o3de_context.engine_path)
```

tkothadev:
> how many export scripts are we expecting to have? As a user, I think I'd prefer if we could just have one export script - a default - that can handle engine centric or project centric, and if I need to do something fancy, I can make my own export script. If this is the case, then I would name this file, `default_project_export.py` - it can also be the default export script the Project Manager uses.

I agree with that instinct, but the plan is to iterate to that point. Right now we have the engine centric export script to start with. The plan is to modify the script to include a project-centric approach as well. If, and only if, that turns out to be too bloated an approach will I need to split it up into a separate export script. 

Atm, I'm opting to name it engine-centric for the sake of transparency. That way folks with project-centric workflows wont be surprised if the current script is incompatible.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
```diff
@@ -0,0 +1,161 @@
...
+parser.add_argument('-pfp', '--project-file_pattern-to-copy', type=str, dest='project_file_patterns_to_copy', action='append',
+                        help="Any additional file patterns located in the project directory. File patterns will be relative to the project path.")
+parser.add_argument('-bnmt', '--build-non-mono-tools', action='store_true')
+parser.add_argument('-nmbp', '--non-mono-build-path', type=pathlib.Path, default=None)
+parser.add_argument('-mbp', '--mono-build-path', type=pathlib.Path, default=None)
+parser.add_argument('-nogame', '--no-game-launcher', action='store_true', help='this flag skips building the Game Launcher on a platform if not needed.')
+parser.add_argument('-noserver', '--no-server-launcher', action='store_true', help='this flag skips building the Server Launcher on a platform if not needed.')
+parser.add_argument('-nounified', '--no-unified-launcher', action='store_true', help='this flag skips building the Unified Launcher on a platform if not needed.')
+parser.add_argument('-pl', '--platform', type=str, default=host_asset_platform, choices=['pc', 'linux', 'mac'])
```
AMZN-alexpete, line 47: not currently supported but should eventually include 'ios' and 'android'

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
tkothadev: Some projects may want an incremental approach. For example, if we plan to have mobile use cases, it may be good to add the ability to only build patches, or DLC content that can be pipelined into a live service.

scripts/o3de/ExportScripts/export_standalone_monolithic_engine_centric.py
```diff
@@ -0,0 +1,160 @@
...
+                        help='The CMake build configuration to use when building project binaries. If tool binaries are built with this script, they will use profile mode.')
+parser.add_argument('-aof', '--archive-output-format',
+                        type=str,
+                        help="Format of archive to create from the output directory",
+                        choices=["zip", "gzip", "bz2", "xz"], default="zip")
+parser.add_argument('-sl', '--seedlist', type=pathlib.Path, dest='seedlist_paths', action='append',
+                        help='Path to a seed list file for asset bundling. Specify multiple times for each seed list.')
+parser.add_argument('-gf', '--gem-folder-name-to-copy', type=str, dest='gem_folder_names_to_copy', action='append',
+                        help="The name of a Gem folder to include in release. Only specify folder name. This should match what is in the project directory.")
+parser.add_argument('-pfp', '--project-file_pattern-to-copy', type=str, dest='project_file_patterns_to_copy', action='append',
```
AMZN-alexpete, line 38:
```suggestion
parser.add_argument('-pfp', '--project-file-pattern-to-copy', type=str, dest='project_file_patterns_to_copy', action='append',
```

scripts/o3de/ExportScripts/export_standalone_monolithic_engine_centric.py
```diff
@@ -0,0 +1,160 @@
...
+if not args.no_server_launcher:
+    process_command(['cmake', '--build', str(mono_build_path), '--target', f'{project_name}.ServerLauncher', '--config', args.config],
+                    cwd=o3de_context.engine_path)
+
+if not args.no_unified_launcher:
+    process_command(['cmake', '--build', str(mono_build_path), '--target', f'{project_name}.UnifiedLauncher', '--config', args.config],
+                    cwd=o3de_context.engine_path)
+
+# Before bundling content, make sure that the necessary executables exist
+asset_bundler_batch_path = non_mono_build_path / 'bin' / 'profile' / ('AssetBundlerBatch' + ('.exe' if host_asset_platform=='pc' else ''))
```
spham-amzn, line 98: `host_asset_platform` is not the right value to check the executable extension. 
```suggestion
asset_bundler_batch_path = non_mono_build_path / 'bin' / 'profile' / ('AssetBundlerBatch' + ('.exe' if sys.platform=='windows' else ''))
```

scripts/o3de/ExportScripts/export_standalone_monolithic_engine_centric.py
tkothadev: This came up during development for MPS export script as well. The main reason I chose to split it up this way was to make troubleshooting easier. Doing it this way makes it easier to trace logs whether building assets failed, or building the executables failed, as each command is isolated.

Note taken on `AssetProcessor`. I can remove that.

scripts/o3de/ExportScripts/export_standalone_monolithic_engine_centric.py
```diff
@@ -0,0 +1,160 @@
...
+from o3de import manifest
+
+project_json_data = manifest.get_project_json_data(project_path=o3de_context.project_path)
+project_name = project_json_data.get('project_name')
+host_asset_platform = exp.determine_host_asset_platform()
+parser = argparse.ArgumentParser(prog='Exporter for standalone builds',
+                                 description = "Exports a project as standalone to the desired output directory with release layout. "
+                                                "In order to use this script, the engine and project must be setup and registered beforehand. ")
+parser.add_argument('-out', '--output-path', type=pathlib.Path, required=True, help='Path that describes the final resulting Release Directory path location.')
+parser.add_argument('-cfg', '--config', type=str, default='profile', choices=['release', 'profile'],
```
AMZN-alexpete, line 28: Eventually I we'll want this to default to `release` once the SDK supports that

scripts/o3de/ExportScripts/export_standalone_monolithic_engine_centric.py
```diff
@@ -0,0 +1,160 @@
...
+                    cwd=o3de_context.engine_path)
+
+if not args.no_unified_launcher:
+    process_command(['cmake', '--build', str(mono_build_path), '--target', f'{project_name}.UnifiedLauncher', '--config', args.config],
+                    cwd=o3de_context.engine_path)
+
+# Before bundling content, make sure that the necessary executables exist
+asset_bundler_batch_path = non_mono_build_path / 'bin' / 'profile' / ('AssetBundlerBatch' + ('.exe' if host_asset_platform=='pc' else ''))
+if not asset_bundler_batch_path.is_file():
+    o3de_logger.error(f"AssetBundlerBatch not found at path '{asset_bundler_batch_path}'. In order to bundle the data for project, this executable must be present!")
```
spham-amzn, line 100: Won't AssetBundlerBatch be available from the `cmake` build command above? If this is an expected file from the build, then an assert is more appropriate here. (In other words, we expect this executable to always exist, if it doesn't then there is something wrong in this workflow)

scripts/o3de/ExportScripts/export_standalone_monolithic_engine_centric.py
```diff
@@ -0,0 +1,160 @@
...
+                        help="Any additional file patterns located in the project directory. File patterns will be relative to the project path.")
+parser.add_argument('-bnmt', '--build-non-mono-tools', action='store_true')
+parser.add_argument('-nmbp', '--non-mono-build-path', type=pathlib.Path, default=None)
+parser.add_argument('-mbp', '--mono-build-path', type=pathlib.Path, default=None)
+parser.add_argument('-nogame', '--no-game-launcher', action='store_true', help='this flag skips building the Game Launcher on a platform if not needed.')
+parser.add_argument('-noserver', '--no-server-launcher', action='store_true', help='this flag skips building the Server Launcher on a platform if not needed.')
+parser.add_argument('-nounified', '--no-unified-launcher', action='store_true', help='this flag skips building the Unified Launcher on a platform if not needed.')
+parser.add_argument('-pl', '--platform', type=str, default=host_asset_platform, choices=['pc', 'linux', 'mac'])
+parser.add_argument('-a', '--archive-output', action='store_true', help='This option places the final output of the build into a compressed archive')
```
AMZN-alexpete, line 47: It might be simpler to just use the `--archive-output-format` instead of having two parameters,  in fact, we could rename `--archive-output-format` to be `--archive-output` and the format can be an optional value.

scripts/o3de/ExportScripts/export_standalone_monolithic_engine_centric.py
```diff
@@ -0,0 +1,160 @@
...
+selected_platform = args.platform
+
+if not selected_platform:
+    o3de_logger.error("Unable to identify default host platform! Please supply the platform using '--platform'. Options are [pc, linux, mac].")
+    sys.exit(1)
+
+o3de_logger.info(f"Selected Project path: {o3de_context.project_path}")
+o3de_logger.info(f"Selected Engine path : {o3de_context.engine_path}")
+o3de_logger.info(f"Build path for non-monolithic executables: {args.non_mono_build_path}")
+o3de_logger.info(f"Build path for monolithic executables: {args.mono_build_path}")
```
spham-amzn, line 67: nit: Lets line up the paths for better reading
```suggestion
o3de_logger.info(f"Selected Project path                     : {o3de_context.project_path}")
o3de_logger.info(f"Selected Engine path                      : {o3de_context.engine_path}")
o3de_logger.info(f"Build path for non-monolithic executables : {args.non_mono_build_path}")
o3de_logger.info(f"Build path for monolithic executables     : {args.mono_build_path}")
```

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
tkothadev: That will likely be added in a future PR when we're ready to tackle exporting to mobile.

scripts/o3de/ExportScripts/export_standalone_monolithic_engine_centric.py
```diff
@@ -0,0 +1,160 @@
...
+
+output_cache_path = args.output_path / 'Cache' / selected_platform
+os.makedirs(output_cache_path, exist_ok=True)
+
+# Build project assets and the engine (non-monolithic)
+if args.build_non_mono_tools:
+    process_command(['cmake', '-S', o3de_context.engine_path, '-B', str(non_mono_build_path), '-DLY_MONOLITHIC_GAME=0', f'-DLY_PROJECTS={o3de_context.project_path}'],
```
AMZN-alexpete, line 74: LY_MONOLITHIC_GAME 0 is the default, is this here in case the user tries to re-use an existing folder that had a monolithic build in it?
```suggestion
    process_command(['cmake', '-S', o3de_context.engine_path, '-B', str(non_mono_build_path),  f'-DLY_PROJECTS={o3de_context.project_path}'],
```

scripts/o3de/o3de/export_project.py
```diff
@@ -49,6 +51,17 @@ def args(self) -> list:
...
+    system_platform = platform.system().lower()
+    if not system_platform:
+        return ""
+    if system_platform == "windows":
+        return "pc"
+    elif system_platform == "linux":
+        return "linux"
+    elif system_platform == "darwin":
+        return "mac"
+
```
spham-amzn, line 64:
```suggestion
    host_platform_to_asset_platform_map = { 'windows': 'pc',
                                            'linux':   'linux',
                                            'darwin':  'mac' }
    return host_platform_to_asset_platform_map.get(platform.system().lower(),"")
```

scripts/o3de/ExportScripts/export_standalone_monolithic_engine_centric.py
```diff
@@ -0,0 +1,160 @@
...
+    output_gem_path = args.output_path / 'Gems'/ gem_folder_name
+    os.makedirs(output_gem_path, exist_ok=True)
+    for file in glob.glob(str(pathlib.PurePath(mono_build_path / 'bin' / args.config / 'Gems' / gem_folder_name / '*.*'))):
+        shutil.copy(file, output_gem_path)
+
+for project_file_pattern in args.project_file_patterns_to_copy:
+    for file in glob.glob(str(pathlib.PurePath(o3de_context.project_path / project_file_pattern))):
+        shutil.copy(file, args.output_path)
+
+# Optionally zip the layout directory if the user requests
```
AMZN-alexpete, line 154:
```suggestion
# Optionally compress the layout directory into an archive if the user requests
```

scripts/o3de/ExportScripts/export_standalone_monolithic_engine_centric.py
tkothadev: Yes, that was the intent.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
spham-amzn: Keep in mind we already have a [deployment script for android](https://www.docs.o3de.org/docs/user-guide/platforms/android/ )

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
```diff
@@ -0,0 +1,161 @@
...
+parser.add_argument('-cfg', '--config', type=str, default='profile', choices=['release', 'profile'],
+                        help='The CMake build configuration to use when building project binaries. If tool binaries are built with this script, they will use profile mode.')
+parser.add_argument('-aof', '--archive-output-format',
+                        type=str,
+                        help="Format of archive to create from the output directory",
+                        choices=["zip", "gzip", "bz2", "xz"], default="zip")
+parser.add_argument('-sl', '--seedlist', type=pathlib.Path, dest='seedlist_paths', action='append',
+                        help='Path to a seed list file for asset bundling. Specify multiple times for each seed list.')
+parser.add_argument('-gf', '--gem-folder-name-to-copy', type=str, dest='gem_folder_names_to_copy', action='append',
+                        help="The name of a Gem folder to include in release. Only specify folder name. This should match what is in the project directory.")
```
spham-amzn, line 38: Can't we extrapolate the gem folder name for the project from the project settings?

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
```diff
@@ -0,0 +1,161 @@
...
+
+project_json_data = manifest.get_project_json_data(project_path=o3de_context.project_path)
+project_name = project_json_data.get('project_name')
+host_asset_platform = exp.determine_host_asset_platform()
+parser = argparse.ArgumentParser(prog='Exporter for standalone builds',
+                                 description = "Exports a project as standalone to the desired output directory with release layout. "
+                                                "In order to use this script, the engine and project must be setup and registered beforehand. ")
+parser.add_argument('-out', '--output-path', type=pathlib.Path, required=True, help='Path that describes the final resulting Release Directory path location.')
+parser.add_argument('-cfg', '--config', type=str, default='profile', choices=['release', 'profile'],
+                        help='The CMake build configuration to use when building project binaries. If tool binaries are built with this script, they will use profile mode.')
```
spham-amzn, line 30: Does this mean that I cannot set `release` if I built the tools in profile?

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
tkothadev: We could, but the question is "which one?". From what I saw with MPS, that explicitly includes only the AWSCore folder. This is meant to allow users to manually specify what folders they know they want to include (incase existing cmake build processes do not cover it)

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
tkothadev: You can set it to release. The final executables for the project will match what the chosen config is. Only tool executables (like AssetProcessor/Bundler), will be built with profile mode.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
```diff
@@ -0,0 +1,161 @@
...
+parser.add_argument('-aof', '--archive-output-format',
+                        type=str,
+                        help="Format of archive to create from the output directory",
+                        choices=["zip", "gzip", "bz2", "xz"], default="zip")
+parser.add_argument('-sl', '--seedlist', type=pathlib.Path, dest='seedlist_paths', action='append',
+                        help='Path to a seed list file for asset bundling. Specify multiple times for each seed list.')
+parser.add_argument('-gf', '--gem-folder-name-to-copy', type=str, dest='gem_folder_names_to_copy', action='append',
+                        help="The name of a Gem folder to include in release. Only specify folder name. This should match what is in the project directory.")
+parser.add_argument('-pfp', '--project-file_pattern-to-copy', type=str, dest='project_file_patterns_to_copy', action='append',
+                        help="Any additional file patterns located in the project directory. File patterns will be relative to the project path.")
```
spham-amzn, line 40: IMHO, specifying a package manifest file is a better experience than to have to specify it on a command line every time i need to build an image. I think a manifest can also contain other arguments that are contained in this arg list. Thats not to say that we remove the arguments, but the manifest can contain the default values and the arguments can be a way to override them if needed.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
```diff
@@ -0,0 +1,161 @@
...
+parser.add_argument('-bnmt', '--build-non-mono-tools', action='store_true')
+parser.add_argument('-nmbp', '--non-mono-build-path', type=pathlib.Path, default=None)
+parser.add_argument('-mbp', '--mono-build-path', type=pathlib.Path, default=None)
```
spham-amzn, line 43: Can we add descriptions for these? And what are the defaults? If not required, what happens?

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
tkothadev: does o3de already have a mechanism for a package manifest file, or would we have to create one specifically for this process? If the latter, that may be better done in a dedicated PR.

scripts/o3de/ExportScripts/export_standalone_monolithic_engine_centric.py
tkothadev: How would we handle when it is unspecified? If `none`, then maybe that can indicate turning off the compressing functionality, and if any other value, goes ahead and does it.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
spham-amzn: Maybe manifest is not the right word, but more like a packaging rule. (Think of CMake presets, but per project). It would be directly related to this process, but if you want to do another PR for it, then this script will need to be updated accordingly. But probably better to think more generally about this problem and create an RFC + Design on the packaging process.

scripts/o3de/ExportScripts/export_standalone_monolithic_engine_centric.py
tkothadev: There is the possibility that the user won't want to run build tools compilation, so that cmake command may not run. It is expected, so we can assert, but I chose this method since I want to provide more information to the user so they can pick the best recourse. That includes debug statements instructing users to run the build tool flag if it does turn out they are missing it. 

An assert's error trace clutters up the view in the logs imo, making it harder to read in this case.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
tkothadev: That may need to be a separate process altogether then. Atm, I believe this export script is best suited to PC games

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
tkothadev: I'll add the descriptions

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
```diff
@@ -0,0 +1,161 @@
...
+    o3de_logger.setLevel(logging.ERROR)
+
+non_mono_build_path = (o3de_context.engine_path) / 'build' / 'non_mono' if args.non_mono_build_path is None else args.non_mono_build_path
+mono_build_path     = (o3de_context.engine_path) / 'build' / 'mono' if args.mono_build_path is None else args.mono_build_path
+
+selected_platform = args.platform
+
+if not selected_platform:
+    o3de_logger.error("Unable to identify default host platform! Please supply the platform using '--platform'. Options are [pc, linux, mac].")
+    sys.exit(1)
```
spham-amzn, line 63: This situation should never happen. The argparse argument for `platform` is a choice of those 3 options, and if none of them are set, the arg parser will have raised an exception already

scripts/o3de/ExportScripts/export_standalone_monolithic_engine_centric.py
spham-amzn: Probably don't need `AssetBundler` either

scripts/o3de/ExportScripts/export_standalone_monolithic_engine_centric.py
spham-amzn: We should also strip out debug symbols as well `-DLY_STRIP_DEBUG_SYMBOLS=TRUE`

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
spham-amzn: Maybe allow this parameter to be a file pattern as well (like `MPS*`)

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
spham-amzn: Eventually this script should be part of a `export and deploy` workflow by platform. The android script currently does the packaging and deployment to the device. I think this will be useful for things like 'deploy server launcher to linux' which will uses this script to create the image but then will upload (through ssh or whatever) to a server.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
```diff
@@ -0,0 +1,161 @@
...
+                        help="Any additional file patterns located in the project directory. File patterns will be relative to the project path.")
+parser.add_argument('-bnmt', '--build-non-mono-tools', action='store_true')
+parser.add_argument('-nmbp', '--non-mono-build-path', type=pathlib.Path, default=None)
+parser.add_argument('-mbp', '--mono-build-path', type=pathlib.Path, default=None)
+parser.add_argument('-nogame', '--no-game-launcher', action='store_true', help='this flag skips building the Game Launcher on a platform if not needed.')
+parser.add_argument('-noserver', '--no-server-launcher', action='store_true', help='this flag skips building the Server Launcher on a platform if not needed.')
+parser.add_argument('-nounified', '--no-unified-launcher', action='store_true', help='this flag skips building the Unified Launcher on a platform if not needed.')
+parser.add_argument('-pl', '--platform', type=str, default=host_asset_platform, choices=['pc', 'linux', 'mac'])
+parser.add_argument('-a', '--archive-output', action='store_true', help='This option places the final output of the build into a compressed archive')
```
spham-amzn, line 48: This description is a little confusing to me. Maybe
```suggestion
parser.add_argument('-a', '--archive-output', action='store_true', help='Option to create a compressed archive the output ')
```

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
```diff
@@ -0,0 +1,161 @@
...
+import os
+import o3de.export_project as exp
+import pathlib
+import sys
+import time
+
+from o3de.export_project import process_command
+from o3de import manifest
+
+project_json_data = manifest.get_project_json_data(project_path=o3de_context.project_path)
```
spham-amzn, line 22: Can we move most of the code into a function and have a `if __name__ == "__main__"` condition in this script to call the function? (Or even better yet, isolate these steps into more separate concrete actions and call them in sequence).
This pattern is done to prevent this logic from executing if someone just imports this script as a module.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
tkothadev: Agreed, an RFC would be warranted for that. I think it's out of scope for the PR at present, but I'll file an issue related to this.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
tkothadev: A pattern may be a good compromise, but there's currently an issue with this:

Right now the gem folder name is used to specify the glob pattern of files to copy (which your suggestion would match), but also the folder destination in the output release directory for gems.

That may require splitting this up into "gem-patterns-to-copy" and "gem-folder-output" variables, but  that would make this clunkier for sure.

This is what the current code for copying files from gem folders looks like for context:
```

for gem_folder_name in args.gem_folder_names_to_copy:
    output_gem_path = args.output_path / 'Gems'/ gem_folder_name
    os.makedirs(output_gem_path, exist_ok=True)
    for file in glob.glob(str(pathlib.PurePath(mono_build_path / 'bin' / args.config / 'Gems' / gem_folder_name / '*.*'))):
        shutil.copy(file, output_gem_path)
```

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
tkothadev: So it turns out `if __name__ == "__main__"` will not work because the export script is run by the export_project.py script. As such it is not `__main__`.

We can get an equivalent gating mechanism using `if "o3de_context" in globals()`, which provides the extra benefit that we know only O3DE export related code will run in that section.

spham-amzn: I see that now
```
utils.load_and_execute_script(target_script_path, o3de_context = o3de_context, o3de_logger=logging.getLogger())
```

spham-amzn (approved):

scripts/o3de/ExportScripts/export_standalone_monolithic_engine_centric.py
```diff
@@ -0,0 +1,246 @@
+#
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+#
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+#
+
```
AMZN-alexpete, line 8: We need some comment docs in here about how to use this script with o3de export

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
```diff
@@ -0,0 +1,246 @@
+#
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+#
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+#
+
```
AMZN-alexpete, line 8: We need comment docs in here for users to indicate how to use o3de export with this script

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
```diff
@@ -0,0 +1,246 @@
+#
```
AMZN-alexpete, line 1: This script should also be in the MinimalProject template

Templates/DefaultProject/Template/Registry/assetprocessor_settings.setreg
```diff
@@ -17,6 +17,9 @@
                     "recursive": 1,
                     "order": 3
                 },
+                "Exclude export scripts": {
+                    "pattern": ".*\\\\/ExportScripts\\\\/.*"
+                },
```
AMZN-alexpete, line 22: This should also go in the MinimalProject template Registry folder

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
```diff
@@ -0,0 +1,263 @@
...
+# --seedlist path\to\seedlist1.seed --seedlist path\to\seedlist2.seed <etc..> 
+
+# The following parameters are optional, but may be helpful:
+#   seedlist: an O3DE seedlist for describing what assets of the project should be bundled
+#   gem-folder-name-to-copy: A Gem Folder from <ProjectRoot>/Gems that should be copied to the output location of the exported project
+#                               Files found in these folders will go into <output-path>/Gems/GemFolderName
+#   project-file-pattern-to-copy: Any regex pattern starting from <ProjectRoot> that indicates what files should be copied to output location.
+#                                   Files found in these patterns will go into <output-path>
+
+# For more information on the available parameters for this script, check the parse_arguments function defined at the bottom of the file
```
AMZN-alexpete, line 39: nit: I think it'd be better for this comment to be the first thing after the copyright header

AMZN-alexpete (approved):

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic_engine_centric.py
lumberyard-employee-dm: Instead of using the gem folder name to copy an more exact approach is to use the registered gem name.
The o3de CLI [manifest.get_registered](https://github.com/o3de/o3de/blob/development/scripts/o3de/o3de/manifest.py#L1132-L1139) function can use the gem name to determine the path to gem root folder.
Once that is found the [pathlib.PurePath.name](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.name) could be used to extract the current the folder name

lumberyard-employee-dm (commented):
The export scripts in the DefaultProject and Minimal Project template should be called project centric and named `export_standalone_monolithic.py`.

lumberyard-employee-dm:
Also the files that were added to the DefaultProject and MinimalProject template will NOT be used until the [template.json](https://github.com/o3de/o3de/blob/development/Templates/DefaultProject/template.json) in each template is updated to copy those files over.
Right now those files aren't copied to the project.

PR #16289, opened 2023-06-30
Fixed numeric attributes not being handled properly

### LINKED ISSUE #16125: DPE Inspector: NumericEditor attributes not registered for some sliders/spinboxes
**Describe the bug**
It appears the Min/Max/SoftMin/SoftMax/etc... node attributes for the `NumericEditor` aren't being registered for the Slider and Spinbox variants. So those fields aren't consuming the attributes.

**Steps to reproduce**

3. Create a new entity.
4. Add a Light component to the entity.

**Expected behavior**
The intensity field should show a default value of `100.0`

**Actual behavior**
The intensity field should shows a default value of `1.0` because the `Max` attribute was never set to let it go past `1.0`.

cgalvan:
## What does this PR do?

Fixes #16125 

The numeric editor attributes (e.g. min/max/softmin/softmax) weren't being registered for the different numeric editor types, so they wouldn't get handled properly.

The best use-case was the above bug where the max attribute wasn't being parsed for the light intensity value, so it would actually lose the default value because it was getting moved down to the default max of `1`.

BEFORE:
Editor_m8oEJeHKBq

AFTER:
Editor_CCoa0rc0F2

## How was this PR tested?

Tested repro steps as above and verified it now works as expected.


dtamkin1 (approved):

alexmontAmazon (approved):
nice!

PR #16291, opened 2023-06-30
vulkan switch pipeline fix

VickyAtAZ:
## What does this PR do?
The Editor which has a MPS level open may crash when switch render pipeline from MainRenderPipeline to LowEndRenderPipeline. This is because the DrawPacket which was created with MSAA sample count 2 is used render in a pass which output MSAA sample count 1 render targets. 
This PR fixed issue by re-create DrawPacket after the render pipeline switch.
The previous old was intended to do so but it didn't connect to the SceneNotificationBus which the function won't be called.

MISC changes:
- added a helper function in InstanceDatabase to visit all the instances in the database. 
- update TerrainMeshManager to build a DrawPacket with all the shaders but not just common shaders. This is required to support multiple material pipelines. 

## How was this PR tested?

Switching render pipelines in Editor with MPS project

### DIFF
--- a/Gems/Terrain/Code/Source/TerrainRenderer/TerrainMeshManager.h
+++ b/Gems/Terrain/Code/Source/TerrainRenderer/TerrainMeshManager.h
@@ -219,6 +219,7 @@ namespace Terrain
             AZ::RHI::DrawListTag m_drawListTag;
             AZ::RHI::Ptr<AZ::RHI::ShaderResourceGroupLayout> m_drawSrgLayout;
             AZ::RPI::ShaderVariant m_shaderVariant;
+            AZ::Name m_materialPipelineName;
         };
 
         struct HeightNormalVertex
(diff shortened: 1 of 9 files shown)


mbalfour-amzn (approved):

Gems/Terrain/Code/Source/TerrainRenderer/TerrainMeshManager.h
```diff
@@ -218,6 +219,7 @@ namespace Terrain
             AZ::RHI::DrawListTag m_drawListTag;
             AZ::RHI::Ptr<AZ::RHI::ShaderResourceGroupLayout> m_drawSrgLayout;
             AZ::RPI::ShaderVariant m_shaderVariant;
+            AZ::Name m_materialPipelineName;
```
moudgils, line 222: Where does this get set?

VickyAtAZ: m_cachedDrawData.push_back({ shader, shaderOptions, pipelineState, drawListTag, drawSrgLayout, variant, materialPipelineName });

AMZN-alexpete (approved):

moudgils (approved):

PR #16300, opened 2023-07-03
Cognito User Pools Warning Cleanup

AMZN-Gene:
- Adding HasResourceType to AWSResourceMappingManager, thus allowing users to check if a property exists instead of requesting the property and receiving a warning that it doesn't exist.
- Cleaning up AWS warning about undefined Cognito User Pools. User Pools aren't required, so the warning was giving developers a sense that something is wrong when it's fine.

### DIFF
--- a/Gems/AWSCore/Code/Source/ResourceMapping/AWSResourceMappingManager.cpp
+++ b/Gems/AWSCore/Code/Source/ResourceMapping/AWSResourceMappingManager.cpp
@@ -118,6 +118,11 @@ namespace AWSCore
         }, resourceKeyName);
     }
 
+    bool AWSResourceMappingManager::HasResourceType(const AZStd::string& resourceKeyName) const
+    {
+        return m_resourceMappings.contains(resourceKeyName);
+    }
+
     AZStd::string AWSResourceMappingManager::GetServiceUrlByServiceName(const AZStd::string& serviceName) const
     {
         return GetServiceUrlByRESTApiIdAndStage(
--- a/Gems/AWSCore/Code/Source/ResourceMapping/AWSResourceMappingManager.cpp
+++ b/Gems/AWSCore/Code/Source/ResourceMapping/AWSResourceMappingManager.cpp
@@ -82,6 +82,11 @@ namespace AWSCore
         return m_defaultRegion;
     }
 
+    bool AWSResourceMappingManager::HasResource(const AZStd::string& resourceKeyName) const
+    {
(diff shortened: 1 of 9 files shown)


Gems/AWSCore/Code/Tests/Framework/ServiceClientJobConfigTest.cpp
```diff
@@ -69,6 +69,12 @@ class ServiceClientJobConfigTest
         return "";
     }
 
+    bool HasResourceType(const AZStd::string& resourceKeyName) const override
+    {
+        AZ_UNUSED(resourceKeyName);
+        return false;
+    }
```
spham-amzn, line 76: should use `[[maybe_unused]]` instead
```suggestion
    bool HasResourceType([[maybe_unused]] const AZStd::string& resourceKeyName) const override
    {
        return false;
    }
```

AMZN-Gene: Fixed!
The Gem is using both AZ_UNUSED and [[maybe_unused]]... changing this file to all use [[maybe_unused]]

spham-amzn (approved):

Gems/AWSCore/Code/Tests/Framework/ServiceClientJobConfigTest.cpp
spham-amzn: I think `AZ_UNUSED` should be deprecated since all of the compilers we support supports it. We should remove it from the coding standards if its there

Gems/AWSCore/Code/Source/ResourceMapping/AWSResourceMappingManager.cpp
```diff
@@ -118,6 +118,11 @@ namespace AWSCore
         }, resourceKeyName);
     }
 
+    bool AWSResourceMappingManager::HasResourceType(const AZStd::string& resourceKeyName) const
```
AMZN-alexpete, line 121: `HasResourceType` sounds like you are providing the type which would match the `attribute.resourceType`, but you are passing in a `resourceKeyName` which does not appear to be a `resourceType`, what if this function was simply called `HasResource`?

AMZN-Gene: Yes, I like it! Fixed :D

AMZN-alexpete (approved):

PR #16308, opened 2023-07-05
Adding native support for VectorN and MatrixMxN to script canvas

kberg-amzn:
Adding native support for VectorN and MatrixMxN as script canvas math primitives. Adjusts the operators on MatrixMxN to be less dangerous (removal of operator /, conversion of operator * from a hadamard product to standard matrix multiplication.

image

### DIFF
--- a/Code/Framework/AzCore/AzCore/Math/MatrixMxN.h
+++ b/Code/Framework/AzCore/AzCore/Math/MatrixMxN.h
@@ -107,7 +107,6 @@ namespace AZ
 
         MatrixMxN& operator+=(const MatrixMxN& rhs);
         MatrixMxN& operator-=(const MatrixMxN& rhs);
-        MatrixMxN& operator*=(const MatrixMxN& rhs);
 
         MatrixMxN& operator+=(float sum);
         MatrixMxN& operator-=(float difference);
--- a/Code/Framework/AzCore/AzCore/Math/MatrixMxN.h
+++ b/Code/Framework/AzCore/AzCore/Math/MatrixMxN.h
@@ -41,6 +41,10 @@ namespace AZ
 
         ~MatrixMxN() = default;
 
+        MatrixMxN& operator=(MatrixMxN&&) = default;
+
+        MatrixMxN& operator=(const MatrixMxN&) = default;
+
         //! Creates an M by N matrix with all elements set to zero.
         static MatrixMxN CreateZero(AZStd::size_t rowCount, AZStd::size_t colCount);
 
@@ -86,18 +90,24 @@ namespace AZ
         MatrixMxN GetClamp(const MatrixMxN& min, const MatrixMxN& max) const;
         //! @}
 
-        //! Returns a new VectorN containing the absolute value of all elements in the source MatrixMxN.
+        //! Returns a new MatrixMxN containing the absolute value of all elements in the source MatrixMxN.
         MatrixMxN GetAbs() const;
 
-        //! Returns a new VectorN containing the square of all elements in the source MatrixMxN.
+        //! Returns a new MatrixMxN containing the square of all elements in the source MatrixMxN.
         MatrixMxN GetSquare() const;
 
-        //! These operators perform basic arithmetic on the individual elements within the respective matrices.
+        //! Aside from multiplication, these operators perform basic arithmetic on the individual elements within the respective matrices.
         //! @{
+        MatrixMxN operator-() const;
+        MatrixMxN operator+(const MatrixMxN& rhs) const;
+        MatrixMxN operator-(const MatrixMxN& rhs) const;
+        MatrixMxN operator*(const MatrixMxN& rhs) const;
+        MatrixMxN operator*(float multiplier) const;
+        MatrixMxN operator/(float divisor) const;
+
         MatrixMxN& operator+=(const MatrixMxN& rhs);
         MatrixMxN& operator-=(const MatrixMxN& rhs);
-        MatrixMxN& operator*=(const MatrixMxN& rhs); //! Hadamard product, not matrix multiplication.
-        MatrixMxN& operator/=(const MatrixMxN& rhs);
+        MatrixMxN& operator*=(const MatrixMxN& rhs);
 
         MatrixMxN& operator+=(float sum);
         MatrixMxN& operator-=(float difference);
--- a/Code/Framework/AzCore/AzCore/Math/MatrixMxN.inl
+++ b/Code/Framework/AzCore/AzCore/Math/MatrixMxN.inl
@@ -293,7 +293,7 @@ namespace AZ
 
     AZ_MATH_INLINE MatrixMxN MatrixMxN::operator*(const MatrixMxN& rhs) const
     {
-        MatrixMxN result;
+        MatrixMxN result(GetRowCount(), rhs.GetColumnCount());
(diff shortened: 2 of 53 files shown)


Code/Framework/AzCore/AzCore/Math/MatrixMxN.h
```diff
@@ -86,18 +90,23 @@ namespace AZ
...
+        MatrixMxN operator-() const;
+        MatrixMxN operator+(const MatrixMxN& rhs) const;
+        MatrixMxN operator-(const MatrixMxN& rhs) const;
+        MatrixMxN operator*(const MatrixMxN& rhs) const;
+        MatrixMxN operator*(float multiplier) const;
+        MatrixMxN operator/(float divisor) const;
+
         MatrixMxN& operator+=(const MatrixMxN& rhs);
         MatrixMxN& operator-=(const MatrixMxN& rhs);
-        MatrixMxN& operator*=(const MatrixMxN& rhs); //! Hadamard product, not matrix multiplication.
```
lumberyard-employee-dm, line 99: Good deal.
Frankly for these kind of operations, I prefer to have an explicit function name for the operation instead of operating overloading as different users have different assumptions what the operations mean (Hadamard product vs matrix Product

Code/Framework/AzNetworking/Azcg/Temp/AzNetworking_input_files.txt
```diff
@@ -0,0 +1 @@
+AzNetworking/AutoGen/AutoPacketDispatcher_Header.jinja;AzNetworking/AutoGen/AutoPacketDispatcher_Inline.jinja;AzNetworking/AutoGen/AutoPackets_Header.jinja;AzNetworking/AutoGen/AutoPackets_Inline.jinja;AzNetworking/AutoGen/AutoPackets_Source.jinja;AzNetworking/AutoGen/CorePackets.AutoPackets.xml
```
lumberyard-employee-dm, line 1: This file should be removed.

Gems/ScriptCanvas/Code/Include/ScriptCanvas/Libraries/Math/VectorN.cpp
```diff
@@ -0,0 +1,116 @@
...
+        {
+            return a.IsClose(b, aznumeric_cast<float>(tolerance));
+        }
+
+        Data::BooleanType IsZero(const Data::VectorNType& source, Data::NumberType tolerance)
+        {
+            return source.IsZero(aznumeric_cast<float>(tolerance));
+        }
+
+        Data::VectorNType& ReLU(Data::VectorNType& source)
```
lumberyard-employee-dm, line 50: So this performs a [Rectified Linear Unit](https://www.mathworks.com/help/deeplearning/ref/nnet.cnn.layer.relulayer.html) operation.
This seems machine learning specific and not part of a normal math vector API

Gems/ScriptCanvas/Code/Include/ScriptCanvas/Libraries/Operators/Math/OperatorDiv.cpp
```diff
@@ -103,6 +103,30 @@ namespace ScriptCanvas
...
+            struct OperatorDivImpl<Data::VectorNType>
+            {
+                OperatorDivImpl(Node* node)
+                    : m_node(node)
+                {}
+
+                Data::VectorNType operator()(const Data::VectorNType& a, const Datum& b)
+                {
+                    const Data::VectorNType* divisor = b.GetAs<Data::VectorNType>();
+                    if (divisor->IsZero())
```
lumberyard-employee-dm, line 116: This should also check that the divisor isn't nullptr either and probably return 0 in that case as well.

Code/Framework/AzNetworking/Azcg/Temp/AzNetworking_input_files.txt
kberg-amzn: Good catch, thanks!

Gems/ScriptCanvas/Code/Include/ScriptCanvas/Libraries/Math/VectorN.cpp
kberg-amzn: Yes, it's ML specific. I could remove it, but I anticipate it will be heavily used.

Gems/ScriptCanvas/Code/Include/ScriptCanvas/Libraries/Operators/Math/OperatorDiv.cpp
kberg-amzn: Yup, sounds good 👍

lumberyard-employee-dm (approved):

Code/Framework/AzCore/AzCore/Math/MatrixMxN.inl
```diff
@@ -261,54 +261,82 @@ namespace AZ
...
+        {
+            returnValue.m_values[i] = m_values[i] * multiplier;
+        }
+        return returnValue;
     }
 
-    AZ_MATH_INLINE MatrixMxN& MatrixMxN::operator*=(const MatrixMxN& rhs)
+    AZ_MATH_INLINE MatrixMxN MatrixMxN::operator/(float divisor) const
+    {
+        MatrixMxN returnValue(m_rowCount, m_colCount);
```
spham-amzn, line 313: Maybe an assert `divisor != 0.0` here?

Code/Framework/AzCore/AzCore/Math/VectorN.h
```diff
@@ -39,6 +39,10 @@ namespace AZ
 
         ~VectorN() = default;
 
+        VectorN& operator=(VectorN&&) = default;
```
spham-amzn, line 42: Don't we have to implement the move assignment manually because we use AZStd::move?

lumberyard-employee-dm: No, the move assignment operator would only need to be manually implemented if the default wasn't sufficient.
The default move assignment/move constructor is to invoke move on each of the members.

spham-amzn: ok, just wanted clarification, because i've seen some PRs in the past that implemented the move constructor but left the normal copy constructor as default.

mbalfour-amzn (approved):

Gems/ScriptCanvas/Code/Include/ScriptCanvas/Data/DataType.h
```diff
@@ -56,6 +58,8 @@ namespace ScriptCanvas
             // Function,
             // List,
             AssetId,
+            VectorN,
+            MatrixMxN
```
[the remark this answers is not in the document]
kberg-amzn: Yeah, that's a good suggestion. I was considering something similar last night after I fixed that test.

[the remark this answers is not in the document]
lumberyard-employee-dm: AZ_ENUM_CLASS_WITH_UNDERLYING_TYPE(eType, AZ::u32, ....) can be used to automatically add a count to the enum.

Which can then be accessed as `AZ::AzEnumTraits<eType>::Count`

PR #16320, opened 2023-07-07
Add Dry-Run CLI and Unit Test

VeryKaylin-player1:
## What does this PR do?

Added a new feature, --dry-run (CLI) function that can be used with the --auto-update function. Its purpose is to allow users to verify the expected changes before applying them to actual repo.json file. When the --dry-run option is used, the program will print the number of expected objects that will be added or modified, along with their names.

The --dry-run function can also be used with other functions such as Add-gem, Add-project, and Add-template to preview the anticipated results before executing those actions.

image

## How was this PR tested?

I tested the dry function with this command `./o3de.bat edit-repo-properties -rp "C:\o3de_Projects\test2\repo.json" --auto-update gem project template --dry-run` and made sure results are what should be expected. 

Additionally, I included unit tests for the --dry-run feature and ensured that all existing tests for the repo_properties functions passed successfully.


spham-amzn:
Naive question: I wonder if its easier to just always apply the changes to a temporary file, and the only difference between a dry run and an actual run is that the actual run will overwrite the target file with the temporary file, while a dry run can either print the temporary file, or do a diff between the target file and the temporary file.

VeryKaylin-player1:
>

lumberyard-employee-dm:
> Naive question: I wonder if its easier to just always apply the changes to a temporary file, and the only difference between a dry run and an actual run is that the actual run will overwrite the target file with the temporary file, while a dry run can either print the temporary file, or do a diff between the target file and the temporary file.

Well temporary files causes unnecessary writes to disk and most dry-run operations avoid changing the disk state.
For example, rsync --dry-run option just outputs the files that it would transfer, but doesn't actually cause any transfers itself.
Reference: https://linuxhint.com/use-rsync-dry-run/

VeryKaylin-player1:
> Naive question: I wonder if its easier to just always apply the changes to a temporary file, and the only difference between a dry run and an actual run is that the actual run will overwrite the target file with the temporary file, while a dry run can either print the temporary file, or do a diff between the target file and the temporary file.

This is also a great approach, I didn't do a temp file because if anything goes wrong the file might stay in memory. And if dry run prints the repo.json file data to console, the result can be really long and difficult to read for users. I figure printing somethings simple would be more direct forward for users then to call diff on the two file, which you have to know where your temp file is stored right?

spham-amzn:
> > Naive question: I wonder if its easier to just always apply the changes to a temporary file, and the only difference between a dry run and an actual run is that the actual run will overwrite the target file with the temporary file, while a dry run can either print the temporary file, or do a diff between the target file and the temporary file.
> 
> Well temporary files causes unnecessary writes to disk and most dry-run operations avoid changing the disk state. For example, rsync --dry-run option just outputs the files that it would transfer, but doesn't actually cause any transfers itself. Reference: https://linuxhint.com/use-rsync-dry-run/

It doesn't have to be a temporary file in that case, it could just use a memory buffer to store the content which can either be used to overwrite the repo.json itself or just print to stdout for dry-runs. The point was really to avoid having dry runs use a different part of the code than the actual update itself. Ideally, the mechanism to do the update should flow through the same code, and the divergent code is where we decide to commit the change or display it (or anything else).

spham-amzn:
> > Naive question: I wonder if its easier to just always apply the changes to a temporary file, and the only difference between a dry run and an actual run is that the actual run will overwrite the target file with the temporary file, while a dry run can either print the temporary file, or do a diff between the target file and the temporary file.
> 
> This is also a great approach, I didn't do a temp file because if anything goes wrong the file might stay in memory. And if dry run prints the repo.json file data to console, the result can be really long and difficult to read for users. I figure printing somethings simple would be more direct forward for users then to call diff on the two file, which you have to know where your temp file is stored.
The suggestion to use a temp file was really based on not needing to re-implement the logic that does the actual update for just performing the dry-run. With temp files, you can take advantage of existing tools like diff to compare the temp file and the existing file if you don't want to print out the whole thing. You can even use python's [difflib](https://docs.python.org/3/library/difflib.html) to compare the contents and instead output that for dry runs.

VeryKaylin-player1:
> > > Naive question: I wonder if its easier to just always apply the changes to a temporary file, and the only difference between a dry run and an actual run is that the actual run will overwrite the target file with the temporary file, while a dry run can either print the temporary file, or do a diff between the target file and the temporary file.
> > 
> > 
> > This is also a great approach, I didn't do a temp file because if anything goes wrong the file might stay in memory. And if dry run prints the repo.json file data to console, the result can be really long and difficult to read for users. I figure printing somethings simple would be more direct forward for users then to call diff on the two file, which you have to know where your temp file is stored.
> 
> The suggestion to use a temp file was really based on not needing to re-implement the logic that does the actual update for just performing the dry-run. With temp files, you can take advantage of existing tools like diff to compare the temp file and the existing file if you don't want to print out the whole thing. You can even use python's [difflib](https://docs.python.org/3/library/difflib.html) to compare the contents and instead output that for dry runs.

AMZN-Gene:
> Naive question: I wonder if its easier to just always apply the changes to a temporary file, and the only difference between a dry run and an actual run is that the actual run will overwrite the target file with the temporary file, while a dry run can either print the temporary file, or do a diff between the target file and the temporary file.

Hmm, nice point. When doing a normal run (not dry run) we _already_ generate a backup _.bak_ file, so users can already diff the json and json.bak if they wanted. I figured the --dry-run was nice as users know for sure nothing will actually be altered.

scripts/o3de/o3de/repo_properties.py
```diff
@@ -293,6 +293,44 @@ def _auto_update_json(object_type: str or list,
             _edit_objects(object_name, validation.valid_o3de_template_json, repo_json, expected_files.get("template.json"))
     return 0
 
+def _dry_run(repo_json: dict,
```
AMZN-Gene, line 296: Suggest changing this method name to "print_repo_diff". The name "dry_run" seems like this is different logic to dry run, but really this is just pretty printing the differences between 2 repo files.

AMZN-Gene (changes requested):

scripts/o3de/o3de/repo_properties.py
AMZN-Gene: Or use python's [difflib](https://docs.python.org/3/library/difflib.html) as @spham-amzn suggested. Not sure how hard it would be to "pretty-print" the differences like you are here. There's a lot of noise in repo.json files since the all the object (gem/template/project) properties are added into the repo.json, not something nice and streamline looking such as just the object name (which you're printing now)

scripts/o3de/o3de/repo_properties.py
```diff
@@ -293,6 +293,44 @@ def _auto_update_json(object_type: str or list,
             _edit_objects(object_name, validation.valid_o3de_template_json, repo_json, expected_files.get("template.json"))
     return 0
 
+def print_repo_diff(repo_json: dict,
+             repo_json_original: dict):
+    
+    if repo_json_original == repo_json:
+            logger.warning('No difference is found between your current repo.json file and your dry run')
+            return 0
```
spham-amzn, line 301: I would not consider not having any differences to be an error or warning. I think its valid to just not print anything.

scripts/o3de/o3de/repo_properties.py
```diff
@@ -293,6 +293,44 @@ def _auto_update_json(object_type: str or list,
...
+                if dry_run_repo[f'{object_key}_name'] == original_repo[f'{object_key}_name']
+                and any(field not in original_repo for field in dry_run_repo)
+            ]    
+            if modified_objects:
+                pretty_print_string.append(f'{object_key.capitalize()}s Modified: {len(modified_objects)}')
+                for i, object_name in enumerate(modified_objects, start=1):
+                    pretty_print_string.append(f"  {i}. {object_name}")
+            else:
+                pretty_print_string.append(f'{object_key.capitalize()}s Modified: {len(modified_objects)}')
+    logger.warning('\n'.join(pretty_print_string))
```
spham-amzn, line 331: This isn't an appropriate use for a `logger.warning`. The function's purpose is to print, so we should use `print` here instead of `logger.warning`

scripts/o3de/o3de/repo_properties.py
```diff
@@ -305,6 +341,7 @@ def edit_repo_props(repo_path: pathlib.Path = None,
                        delete_templates: str or list = None,
                        replace_templates: pathlib.Path or list = None,
                        auto_update: str or list = None,
+                       dry_run: bool = None,
```
spham-amzn, line 344: `dry_run` should default to `False`, not `None`. `None` is an indeterminant state, `False` communicates the parameter better

spham-amzn (approved):

AMZN-Gene (approved):

PR #16326, opened 2023-07-08
Remove duplicate XR .pass files

antonmic:
Removing XRLeft and XRRight pipeline pass files since they are the same as multiviewpipeline, so that can be used instead


moudgils (approved):
Nice. Thanks.

VickyAtAZ (approved):

PR #16334, opened 2023-07-10
Remove Obsolete CVAR: sv_isDedicated

### LINKED ISSUE #15646: Obsolete Cvar: sv_isDedicated
I don't think the sv_isDedicated cvar is needed now that we can check if a dedicated server is running using  if(AZ_TRAIT_SERVER && !AZ_TRAIT_CLIENT)

Replace sv_isDedicated with if(AZ_TRAIT_SERVER && !AZ_TRAIT_CLIENT), or provide a wrapper inside IMultiplayer

BytesOfPiDev:
## What does this PR do?

Removes instances of the `sv_isDedicated` CVAR and replaces them with AZ_TRAIT_CLIENT/AZ_TRAIT_SERVER. 

Adds two macros located in `MultiplayerSystemComponent.h`:
- AZ_DEDICATED_SERVER
- AZ_CLIENT_SERVER

To avoid linking to the Multiplayer gem, added a new CVAR, `cl_enableNvClothSimulation`, which defaults to true. It can be set to false on servers, or any other situation someone may want to disable the simulation, but keep the component enabled. I renamed and updated the tests associated with enabling/disabling the simulation.

Fixes #15646

## How was this PR tested?

Compiled on Linux (Ubuntu 22.04.03 LTS)

I ran the following test:

```sh
./python/python.sh scripts/build/ci_build.py -p Linux -t test_profile_nounity
```

```sh
100% tests passed, 0 tests failed out of 140

Label Time Summary:
COMPONENT_Atom                  = 357.18 sec*proc (6 tests)
COMPONENT_MetadataRelocation    =  46.45 sec*proc (1 test)
COMPONENT_Multiplayer           =  57.52 sec*proc (1 test)
COMPONENT_Physics               =  47.47 sec*proc (1 test)
COMPONENT_ScriptCanvas          =  17.64 sec*proc (1 test)
COMPONENT_Smoke                 =   0.52 sec*proc (1 test)
COMPONENT_Terrain               =  51.46 sec*proc (1 test)
COMPONENT_TestTools             = 249.24 sec*proc (7 tests)
COMPONENT_WhiteBox              =  46.45 sec*proc (1 test)
FRAMEWORK_googletest            = 1589.22 sec*proc (96 tests)
FRAMEWORK_pytest                = 1031.11 sec*proc (43 tests)
SUITE_main                      = 2516.10 sec*proc (117 tests)
SUITE_smoke                     = 105.88 sec*proc (23 tests)
TIAF_shard_fixture              =  81.81 sec*proc (1 test)

Total Test time (real) = 1386.41 sec
~/repos/o3de
--------------------------------------------------------------------------------
[ci_build] OK
```
I opened the AutomatedTesting project and the `BasicConnectivity_Connects` level, set `editorsv_clientserver true` `editorsv_enabled true`, and pressed play. The editor began hosting and the player sphere was visible.


Gems/NvCloth/Code/Source/Components/ClothComponent.cpp
```diff
@@ -14,6 +14,15 @@
 
 namespace NvCloth
 {
+    AZ_CVAR(
+        bool,
+        bg_enableNvClothSimulation,
```
AMZN-Gene, line 19: Could we force _bg_enableNvClothSimulation_ to false inside of the MultiplayerSystemCompoent if it's a dedicated server? Not a fan of cvars being set automatically, but I think most (all?) people running a dedicated server will want this disabled.

AMZN-Gene (approved):

Gems/NvCloth/Code/Source/Components/ClothComponent.cpp
```diff
@@ -14,6 +14,15 @@
...
+        true,
+        nullptr,
+        AZ::ConsoleFunctorFlags::DontReplicate,
+        "Enable NvCloth simulation. It is useful to set this to false on servers for increase performance.\n"
+        "Affects components activated after this cvar is set.");
```
AMZN-Gene, line 24: Suggest: "Only affects components activated after this cvar is set"

Gems/NvCloth/Code/Source/Components/ClothComponent.cpp
AMZN-Gene: Or is there a way for ClothComponent to exclude itself from server entities? We do something similar with Popcorn Fx particle components, where they simply won't be created on server.

AMZN-Gene:
Be sure to conduct an extra test by hand in editor using editorsv_clientserver true. This causes editor to become its own server after pressing CTRL+G instead of spawning a new ServerLauncher in the background. Jenkins Automated Tests only tests Editor when using dedicated servers, so if this change accidently breaks _editorsv_clientserver_, it won't catch it.

Great to see 1 less cvar for devs to keep track of!

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorConnection.cpp
```diff
@@ -42,28 +41,29 @@ namespace Multiplayer
         // - LegacySystemInterfaceCreated is signaled, so that the logging system is ready. Automated testing listens for these logs.
         // - LegacyCommandLineProcessed is signaled, so that everything has initialized and finished their blocking loads, so that it
         // should be relatively safe to start receiving packets without as much fear of too much time passing between system ticks.
-        if (editorsv_isDedicated)
+#if AZ_TRAIT_SERVER && !AZ_TRAIT_CLIENT
```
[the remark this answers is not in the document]
AMZN-Gene: The Mutliplayer Sample Game Lift Gem already does something similar:
```
#define AZ_DEDICATED_SERVER_ONLY (AZ_TRAIT_SERVER && !AZ_TRAIT_CLIENT)
```

https://github.com/o3de/o3de-multiplayersample/blob/dd1c6e1b1bc5f3aca2d5b70f7b4c4687271463d8/MPSGameLift/Code/Source/MPSGameLiftModuleInterface.h#L18-L19

Gems/NvCloth/Code/Source/Components/ClothComponent.cpp
BytesOfPiDev: I removed the server target from NvCloth and renamed it to `cl_enableNvClothSimulation`

Gems/Multiplayer/Code/Source/Debug/MultiplayerConnectionViewportMessageSystemComponent.cpp
```diff
@@ -115,22 +115,10 @@ namespace Multiplayer
         switch (agentType)
```
captain-hoity-toity, line 115: The code is generally well-written and organized. Here are some minor suggestions for improvement:

1. In the `MultiplayerConnectionViewportMessageSystemComponent::DrawConnectionStatus` method, the `switch` statement for `connectionState` does not have a `default` case. It's good practice to always include a `default` case in a `switch` statement even if you think all possible cases are covered.

2. In the `MultiplayerConnectionViewportMessageSystemComponent::OnEditorSendingLevelData` method, consider checking if `bytesTotal` is zero before performing division to avoid potential division by zero.

3. There are several hardcoded strings used for messages. Consider moving these to a constants file or using a localization system for easier management and potential localization in the future.

4. `MultiplayerConnectionViewportMessageSystemComponent::OnEditorConnectionAttemptsFailed` and `MultiplayerConnectionViewportMessageSystemComponent::OnConnectToSimulationFail` methods have similar code. Consider creating a helper method to avoid code duplication.

5. In the `MultiplayerConnectionViewportMessageSystemComponent::OnEditorConnectionAttempt` method, consider renaming the parameters to something more descriptive than `connectionAttempts` and `maxAttempts`.

6. In the `MultiplayerConnectionViewportMessageSystemComponent::OnRenderTick` method, consider breaking up this method into smaller, more manageable methods. This would improve readability and maintainability. 

7. There are magic numbers used in the code (e.g., `0.5f`, `0.01f`, `1.0f`). Consider defining these as named constants to improve code readability and maintainability. 

8. Consider adding more comments to explain what each block of code is doing. This can help other developers understand the code more quickly and easily.

lsemp3d (approved):

Gems/Multiplayer/Code/Source/Editor/MultiplayerEditorConnection.cpp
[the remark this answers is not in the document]
BytesOfPiDev: I reverted this change back to the original because `editorsv_isDedicated` is currently necessary to toggle the editor being a dedicated server at runtime. I wanted to try to get rid of the CVAR, but without it, I'd have to set whether it is or is not one at compile time.

cgalvan (approved):

spham-amzn:
I'm not sure if this is the right approach. While this will work for monolithic release builds, the fact that a launcher is running in client or server mode depends on the specific launcher that is being run (GameLauncher, ServerLauncher, UnifiedLauncher, HeadlessServerLauncher). Replacing with a compile time flag locks the MultiplayerGem into one mode or the other, even though either a Game or Server Launcher can be launched using the same shared Multiplayer Gem.

spham-amzn:
What I would suggest is the following:

1. Add an `isDedicatedServer()` to [ApplicationTypeQuery](https://github.com/o3de/o3de/blob/fc7f51a5748612eb069ec4623908311e2af99915/Code/Framework/AzCore/AzCore/Component/ComponentApplicationBus.h#L40) and add the appropriate enum `DedicatedServer` mask.  That will allow you to [query](https://github.com/o3de/o3de/blob/fc7f51a5748612eb069ec4623908311e2af99915/Code/Framework/AzCore/AzCore/Component/ComponentApplicationBus.h#L196) to application type instead of using the #ifdef guards.
2. Add both a `SetDedicatedServer(bool dedicated)` method, and a `bool m_dedicatedServer{ false };` attribute to [GameApplication](https://github.com/o3de/o3de/blob/fc7f51a5748612eb069ec4623908311e2af99915/Code/Framework/AzGameFramework/AzGameFramework/Application/GameApplication.h#L16).
3. In GameApplication::QueryApplicationType, add a check for `m_dedicatedServer` after the `m_headless` check to see if you need to add the (bitwise or) `AZ::ApplicationTypeQuery::Masks::DedicatedServer` to `appType.m_maskValue`
4. Right after the block that sets the [Headless ](https://github.com/o3de/o3de/blob/fc7f51a5748612eb069ec4623908311e2af99915/Code/LauncherUnified/Launcher.cpp#L421-L426) flag, and code to set the new DedicatedServer flag:
   ```
   gameApplication.SetDedicatedServer(IsDedicatedServer());
   ```
5. Go back to the other modifications and replace the `if AZ_DEDICATED_SERVER` blocks with 
   ```
        AZ::ApplicationTypeQuery appType;
        AZ::ComponentApplicationBus::Broadcast(&AZ::ComponentApplicationBus::Events::QueryApplicationType, appType);
        if (appType.IsDedicatedServer())
        {
          ...
        }
   ```

This is just paraphrasing a suggestion, so I there may be some things in this suggestion that needs updating.

PR #12009, opened 2022-09-15
[Atom] Add indirect raytracing command

mprettner-huawei:
## What does this PR do?

This PR adds support for indirect ray tracing (similar do indirect draw commands) to the Vulkan and DX12 RHI.
The accompanying change to AtomSampleViewer can be found in https://github.com/o3de/o3de-atom-sampleviewer/pull/510

## How was this PR tested?

I tested the PR with the AtomSampleViewer using Samples->Features->SSAO and changeing to RTAO in the sidebar on the right, with both Vulkan and DX12. The results looked the same before and after the change.

PR #16341, opened 2023-07-10
Fixed Indirect Raytracing for DX12

msat-huawei:
## What does this PR do?

Indirect raytracing, as implemented in https://github.com/o3de/o3de/pull/12009, does not work in DX12.
Why? Because the buffer, that ExecuteIndirect expects has a different layout than the one we currently have.
The buffer contains information about the Raytracing shader table, as well as the number of threads (See [here](https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#d3d12_dispatch_rays_desc)).
The buffer in Vulkan only contains information about the number of threads, not the shader table. (Or not, see sidenote)

This PR solves this problem, by copying the indirect call information (number of threads) from the indirect buffer, e.g. passed as a slot, to a buffer also containing the shader table.
This new buffer must be stored somewhere. For this, the `DispatchRaysIndirectBuffer` is introduced.
An instance of it must be passed to the `DispatchRaysIndirect` constructor. It's also necessary to pass the shader table to the buffer, every time the shader table changes.

I also added a new BufferPool for indirect buffers. I'm not sure if this is the right solution, or adding the `Indirect` flag to another BufferPool would be better.

Sidenote: There is also a Vulkan [call](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkTraceRaysIndirectCommand2KHR) with the shader table, but I don't think it's a good idea to use that.
This would mean, that we couldn't pass indirect arguments, by slot, to the Raytracing pass. Instead the pass that fills the indirect buffer would need to get the buffer from the Raytracing pass. Meaning that a pass would need a buffer from another pass, that is defined after itself. Which I don't think would currently work. It would also not be possible to use the same indirect buffer for multiple passes.

## How was this PR tested?

Windows and DX12, with a custom indirect raytracing pass.


Gems/Atom/RHI/DX12/Code/Source/RHI/DispatchRaysIndirectBuffer.cpp
```diff
@@ -0,0 +1,156 @@
...
+                barrier.Type = D3D12_RESOURCE_BARRIER_TYPE_TRANSITION;
+                barrier.Flags = D3D12_RESOURCE_BARRIER_FLAG_NONE;
+                barrier.Transition.pResource = dx12OriginalBuffer->GetMemoryView().GetMemory();
+                barrier.Transition.Subresource = D3D12_RESOURCE_BARRIER_ALL_SUBRESOURCES;
+                barrier.Transition.StateBefore = D3D12_RESOURCE_STATE_INDIRECT_ARGUMENT;
+                barrier.Transition.StateAfter = D3D12_RESOURCE_STATE_COPY_SOURCE;
+                commandList->ResourceBarrier(1, &barrier);
+            }
+
+            commandList->CopyBufferRegion(
```
moudgils, line 126: Add a comment what is getting copied over here

moudgils (commented):
I wonder if a better design is to have the new class named RHI::IndirectBuffer which simply holds RHI::Buffer for the most part and maybe info if the buffer will be used for dispatching rays. The Dx12 version of it will create multiple buffers to update the original Buffer as done by DX12::DispatchRaysIndirectBuffer here (as long as the buffer is used for dispatching rays). The extra buffers can maybe use the same pool as the original buffer so the bufferpool doesnt need to be passed in at higher level. Then RHI::IndirectBufferView will hold RHI::IndirectBuffer instead of RHI::Buffer. I would also move **copy step** into the DX12::Commandlist as that seems a more natural place for it. Thoughts? Tagging @akioCL for visibility.

msat-huawei:
I'm not sure having a RHI::IndirectBuffer is a better solution.
As far as I see, the pros are, that we can easily reuse the BufferPool, and it will be impossible to forget adding the buffers to the call.
The cons are, that we now need a generic IndirectBuffer that contains raytracing specific data (we will need a SetShaderTable function in there). We will then also need to change all Compute/Raster passes that use indirect calls.

About the passed BufferPools: I could get rid of the stagingBufferPool, if a staging buffer pool is introduced in the DX12 device. Vulkan and Metal already have one in there. I would pretty much copy the code from there.

I agree on moving the actualy copy step to the DX12::Commandlist.

msat-huawei:
I realized that the code in this PR does not build when PIX is enabled.
The error was hower not in my code. The issue was created in the QueryPoolResolver.cpp file which includes PIX before including the DirectX headers. I think the issue surfaced because the unity build files are now created slightly differently, due to the new files. My solution is to include the DX12 headers in Device_Windows.h

moudgils:
> I'm not sure having a RHI::IndirectBuffer is a better solution. As far as I see, the pros are, that we can easily reuse the BufferPool, and it will be impossible to forget adding the buffers to the call. The cons are, that we now need a generic IndirectBuffer that contains raytracing specific data (we will need a SetShaderTable function in there). We will then also need to change all Compute/Raster passes that use indirect calls.
> 
> About the passed BufferPools: I could get rid of the stagingBufferPool, if a staging buffer pool is introduced in the DX12 device. Vulkan and Metal already have one in there. I would pretty much copy the code from there.
> 
> I agree on moving the actualy copy step to the DX12::Commandlist.


I guess another option could be to add a second RHI::Buffer to RHI::DispatchRaysIndirect (instead of DispatchRaysIndirectBuffer). Lets call this buffer RHI::Buffer* m_shaderTableAndDim. Add new API within RHI::BufferPool with something like UpdateShaderTable(RHI* ShaderTable) and the dx12 version of it would update it as needed via BufferPoolResolver (so no need for staging buffer). And then the higher level code can pass this buffer in to RHI during Submit time via DispatchRaysIndirect to be used as needed. It eliminates the need to have a new RHI object.  The thing that I dont like about this approach is the new API within RHI::BufferPool. So our options is wether we have a new RHI class just for this purpose or add new api to RHI::BufferPool. Either are not ideal as I hate to make big API changes at RHI level just to solve very backend specific detail. I guess I would be ok with this approach although not ideal but lets clean up  and remove the code around staging pool and keep that as part of dx12 back end at the very least.

msat-huawei:
I decided to keep using the implemented approach, and just cleaned it up a bit:
I removed the staging BufferPool from the class. The DX12 Device already had a staging memory allocator. Just not a BufferPool like the other RHIs. I didn't see that before, but use it now.
I also moved the copy commands to the CommandList.cpp

PR #16342, opened 2023-07-10
DX12: Raytracing: Support multiple shaders in one source file

msat-huawei:
## What does this PR do?

Having multiple raytracing shaders (RayGen, Hit, Miss) in the same `.shader` file leads to errors in the DX12 debug layer. The errors look something like this:
`D3D12 ERROR: ID3D12Device::CreateStateObject: Export "RayGen" already defined. [ STATE_CREATION ERROR #1194: CREATE_STATE_OBJECT_ERROR]`

This is because the shader libraries, passed to the `RayTracingPipelineStateDescriptor` are duplicates of each other.
In this PR, we check for duplicates in the shader libraries, so each one is only attached once.

## How was this PR tested?

Windows and DX12/Vulkan. For DX12 it now works, Vulkan still works.

### DIFF
--- a/Gems/Atom/Feature/Common/Code/Source/RayTracing/RayTracingPass.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/RayTracing/RayTracingPass.cpp
@@ -67,44 +67,56 @@ namespace AZ
                 return;
             }
 
-            // ray generation shader
-            m_rayGenerationShader = LoadShader(m_passData->m_rayGenerationShaderAssetReference);
-            if (m_rayGenerationShader == nullptr)
+            struct RTShaderLib
             {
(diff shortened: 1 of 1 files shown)


jhmueller-huawei (approved):

moudgils (commented):
This only handles RayTracingPass. Other passes may need a similar solution. Also, 2 shaders in one source file may break on Vulkan backend which I think was why we didn't add this support initially. Can we confirm if this works on Vk? And if it doesn't we should add an assert for the Vk backend providing a solution in the assert message (i.e break the shader into separate files).

msat-huawei:
I don't think the other passes need something like that. Raytracing is a bit special here, as we usually need at least three shaders here (RayGen,Hit,Miss). For Compute there is usually only one shader, and the Vertex/Fragment shaders in Raster passes can be in the same file already..

As for Vulkan: Yes this works there. We developed the passes using Vulkan, and when we tried to run them with DX12, we found out that they don't work there.

moudgils (approved):

PR #16343, opened 2023-07-10
Pass attachment: Fetch *sources for attachments recursively

msat-huawei:
## What does this PR do?

When creating a pass request, e.g. the size of an attachment might depend on the size of another image/buffer. For this a size/format/multisampling/arraysize can be specified.
This does currently not work, if:
- The referenced image/buffer attachment, also has a size source
- The referenced attachment has a `Transient' lifetime
- The current attachment has an `Imported` lifetime

This happens e.g. when an images size is dependent on the input image size, which again is dependent on the PipelineOutput size.

This PR solves this issue, by recursively searching for the "original" size source, and use it.

The same is also true for the format/multipsampling/arraysize members.

## How was this PR tested?

Windows and DX12/Vulkan.

### DIFF
--- a/Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Pass/PassAttachment.h
+++ b/Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Pass/PassAttachment.h
@@ -112,9 +112,19 @@ namespace AZ
                     u8 m_getSizeFromPipeline : 1;
                     u8 m_getFormatFromPipeline : 1;
                     u8 m_getMultisampleStateFromPipeline : 1;
+                    u8 m_updatingImageFormat : 1;
+                    u8 m_updatingMultisampleState : 1;
+                    u8 m_updatingSize : 1;
+                    u8 m_updatingArraySize : 1;
                 };
                 u8 m_allFlags = 0;
             };
+
+        protected:
+            void UpdateImageFormat();
+            void UpdateImageMultisampleState();
+            void UpdateImageSize();
+            void UpdateImageArraySize();
         };
 
         //! An attachment binding points to a PassAttachment and specifies how the pass uses that attachment.
--- a/Gems/Atom/RPI/Code/Source/RPI.Public/Pass/PassAttachment.cpp
+++ b/Gems/Atom/RPI/Code/Source/RPI.Public/Pass/PassAttachment.cpp
@@ -127,6 +127,16 @@ namespace AZ
                 UpdateImageMultisampleState();
                 UpdateImageSize();
                 UpdateImageArraySize();
+
+                if (m_generateFullMipChain)
+                {
+                    uint32_t width = m_descriptor.m_image.m_size.m_width;
+                    uint32_t height = m_descriptor.m_image.m_size.m_height;
+
+                    double maxDimension = static_cast<double>(AZStd::max(width, height));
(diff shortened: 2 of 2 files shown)


antonmic:
Can you give an example of a configuration of passes where the previous size source did not work?

VickyAtAZ (commented):
I am not a fan of this recursion tries either. This solution also have flaws if the reference attachment using get format or other states from render pipeline. 

I am suggesting we could add these functions to PassAttachment
```
UpdateImageFormat
UpdateImageMultisampleState
UpdateImageSize
UpdateImageArraySize
```
For example, the code may look like: 
```
void PassAttachment::UpdateImageFormat()
{
	if (m_allFlags.m_updatingImageFormat)
	{
		AZ_Assert(false, "Error: round trip reference");
		return;
	}
	m_allFlags.m_updatingImageFormat = true;
	if (m_getFormatFromPipeline && m_renderPipelineSource)
	{
		m_descriptor.m_image.m_format = m_renderPipelineSource->GetRenderSettings().m_format;
	}
	else if (m_formatSource)
	{
		auto& refAttachment = m_formatSource->GetAttachment();
		if (refAttachment && refAttachment->m_descriptor.m_type == RHI::AttachmentType::Image)
		{
			refAttachment->UpdateImageFormat();
			m_descriptor.m_image.m_format = refAttachment->m_descriptor.m_image.m_format;
		}
	}
	m_allFlags.m_updatingImageFormat = false;
}

void PassAttachment::Update(bool updateImportedAttachments)
{
	if (m_descriptor.m_type == RHI::AttachmentType::Image && (m_lifetime == RHI::AttachmentLifetimeType::Transient || updateImportedAttachments == true))
	{
		UpdateImageFormat();
		UpdateImageMultisampleState();
		UpdateImageSize();
		UpdateImageArraySize();
	}
}
```

Note: I wanted to add  a flag for PassAttachemnt such as m_imageFormatSet, the issue was we don't have a good place to set the flag to false for all the PassAttachemnts incase we want to update the size in FrameBeginInternal due to binding change. 

I think we need something better to resolve many pass update order issues. For now, we may have to compromise for this individual problem.

msat-huawei:
@VickyAtAZ I like your solution much better than the one I implemented. I've added it to the PR.

Just one question: I've now made the new Update* functions protected, as they should usually not be called from outside the class. Is this OK in a struct, according to the coding standard?

VickyAtAZ:
> @VickyAtAZ I like your solution much better than the one I implemented. I've added it to the PR.
> 
> Just one question: I've now made the new Update* functions protected, as they should usually not be called from outside the class. Is this OK in a struct, according to the coding standard?

I don't think there are any coding guideline regarding to the private members of struct. I think it's good idea to have these functions to be private. FYI, this is the link to O3DE coding standards
https://github.com/o3de/sig-core/blob/main/governance/Coding-Standards-and-Style-Guide.md#private-members-need-love-too

VickyAtAZ (approved):

jhmueller-huawei (approved):

msat-huawei:
I just realized that I accidentally removed the GenerateFullMipChain if in the Update function. I added it again. This didn't show up in my tests because it is only used in the Luminance downsampling passes, which we don't use.

moudgils (approved):

PR #16344, opened 2023-07-10
Change line endings to LF

kh-huawei:
## What does this PR do?

Change the line endings of `RenderPipeline.h` to 'LF'


jhmueller-huawei (approved):
Thanks for the fix, source code files shouldn't be CRLF.

moudgils (approved):

spham-amzn:
How did that happen? The git client should been handling line-endings based on the platform youre on

kh-huawei:
`core.autocrlf=true` commits files with the 'LF' ending, regardless of what is used locally. As far as i can tell, the file hat 'CRLF' in the repo, and i wanted to change something (in the other MR), which caused my git to convert each line to 'LF', which makes the diff somewhat useless.

lumberyard-employee-dm (approved):

PR #16350, opened 2023-07-10
Multiple fixes with game launcher for loft scene 

VickyAtAZ:
## What does this PR do?
- Crash when start loft scene game launcher: it was trying to build cullable before mesh was loaded.
- Lighting was wrong when load a new level while there is a level loaded: the new level entities were activated before entities from old level are deactivated. This leads to IBL fp reset its diffuse & specular maps to default.

## How was this PR tested?
Run game launcher for loft-arch-vis-sample project. 
Load Interior_03.prefab level.
Load Interior_03.prefab level again. 
Make sure the 3d rendering is same before and after load the Interior_03.prefab level the second time. 
Try switch to different levels then load Interior_03.prefab again. The rendering result should be the same.

### DIFF
--- a/Code/Legacy/CrySystem/LevelSystem/SpawnableLevelSystem.cpp
+++ b/Code/Legacy/CrySystem/LevelSystem/SpawnableLevelSystem.cpp
@@ -591,7 +591,7 @@ namespace LegacyLevelSystem
         AzFramework::RootSpawnableInterface::Get()->ReleaseRootSpawnable();
         
         // Process the queued deactivate calls for the unloaded entities
-        AzFramework::RootSpawnableInterface::Get()->ProcessSpawnableQueue();
+        AzFramework::RootSpawnableInterface::Get()->ProcessSpawnableQueueUntilEmpty();
 
         m_lastLevelName.clear();
 
--- a/Code/Legacy/CrySystem/LevelSystem/SpawnableLevelSystem.cpp
+++ b/Code/Legacy/CrySystem/LevelSystem/SpawnableLevelSystem.cpp
@@ -589,6 +589,9 @@ namespace LegacyLevelSystem
 
         // Delete level entities and remove them from the game entity context
         AzFramework::RootSpawnableInterface::Get()->ReleaseRootSpawnable();
+        
+        // Process the queued deactivate calls for the unloaded entities
+        AzFramework::RootSpawnableInterface::Get()->ProcessSpawnableQueue();
 
         m_lastLevelName.clear();
 
--- a/Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
@@ -845,6 +845,12 @@ namespace AZ
                 // Per mesh shader option flags was on, but now turned off, so reset all the shader options.
                 for (auto& modelHandle : m_modelData)
                 {
+                    // skip if the model need to be initialized
+                    if (modelHandle.m_flags.m_needsInit)
+                    {
+                        continue;
+                    }
+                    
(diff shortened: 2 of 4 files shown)


spham-amzn (approved):

lumberyard-employee-dm (approved):
This makes sense.
If the system is going to load a new spawnable, the old spawnable should have all of its pending request flushed.

Code/Legacy/CrySystem/LevelSystem/SpawnableLevelSystem.cpp
```diff
@@ -589,6 +589,9 @@ namespace LegacyLevelSystem
 
         // Delete level entities and remove them from the game entity context
         AzFramework::RootSpawnableInterface::Get()->ReleaseRootSpawnable();
+        
+        // Process the queued deactivate calls for the unloaded entities
+        AzFramework::RootSpawnableInterface::Get()->ProcessSpawnableQueue();
```
AMZN-alexpete, line 594: Should this instead use `ProcessSpawnableQueueUntilEmpty` to make sure everything is processed?
```suggestion
        AzFramework::RootSpawnableInterface::Get()->ProcessSpawnableQueueUntilEmpty();
```

AMZN-alexpete (changes requested):
I think we might need to use `ProcessSpawnableQueueUntilEmpty` to ensure all off the queue is processed

Code/Legacy/CrySystem/LevelSystem/SpawnableLevelSystem.cpp
VickyAtAZ: Based the description, the ProcessSpawnableQueue() should be good for the loading a new level use case
```

        //! Force processing all SpawnableEntitiesManager requests immediately
        //! This is useful when loading a different level while SpawnableEntitiesManager still has
        //! pending requests
        virtual void ProcessSpawnableQueue() = 0;
        //! Force processing all SpawnableEntitiesManager requests immediately and keep reprocessing until the queue is empty
        //! This is useful when unloading and shutting down to ensure that all pending requests are cleared.
        virtual void ProcessSpawnableQueueUntilEmpty() = 0;
```
Currently the ProcessSpawnableQueueUntilEmpty is only used at the system shutting down at this moment.

By looking at the code, it seems it's possible that some requests get re-queued. It might be safer to use ProcessSpawnableQueueUntilEmpty.

Gems/Atom/Feature/Common/Code/Source/Mesh/MeshFeatureProcessor.cpp
```diff
@@ -878,6 +878,11 @@ namespace AZ
                 {
                     if (modelHandle.m_cullable.m_prevShaderOptionFlags != modelHandle.m_cullable.m_shaderOptionFlags)
                     {
+                        // skip if the model need to be initialized
+                        if (modelHandle.m_flags.m_needsInit)
+                        {
+                            continue;
+                        }
```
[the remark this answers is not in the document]
VickyAtAZ: Nice catch. That needs this check too

Gems/Atom/Feature/Common/Code/Source/Shadows/FullscreenShadowPass.h
```diff
@@ -52,9 +52,10 @@ namespace AZ
                 m_receiverShadowPlaneBiasEnable = enable;
             }
 
-            void SetLightIndex(int lightIndex)
+            // Set directional light's raw index which is used for access directional light in shader
```
mbalfour-amzn, line 55:
```suggestion
            // Set directional light's raw index which is used for accessing the directional light in the shader
```

mbalfour-amzn (approved):

AMZN-alexpete (approved):

PR #16353, opened 2023-07-11
DPE | Fixes to UI Element and Group management

AMZN-daimini:
Fixes multiple issues with UI Element and Group management in the LegacyReflectionBridge class.

- Fixes node skipping mechanism so that when a node execution is queued, that node's descendants are not explored. This caused nodes to be correctly reordered, but their children to be displayed in the wrong spot;
- Changes the way nodes are restored out of order when added to groups; instead of just visiting the node, we now call EnumerateInstance on it so that all descendants are handled subsequently;
- Removes `m_iteratingOnGroups` variable, which was a test mechanism to prevent re-entry. I verified that's not a problem and as such the variable would just be misleading;
- Relocates the code that handles incrementing `m_childElementIndex`, which had incorrectly been placed into a helper function and would set the value of all indices to `[1]`;
- Addresses issue with generating the path to a UI Element in `HandleNodeUiElementsRetrieval`. The code incorrectly assumed Groups could be nested, while instead the EditContext behavior is to close a Group when a new one is defined;
- Fixes initialization of `m_instanceToInvoke` when creating an entry for a Group;
- Minor unrelated fix to a comment referring to an old Qt version that is no longer used.

These fixes contribute to addressing the issues with the Standalone DPE test dialog.

BEFORE:
image

AFTER:
image


cgalvan (approved):

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/Reflection/LegacyReflectionBridge.cpp
```diff
@@ -652,7 +644,7 @@ namespace AZ::Reflection
                 if (HandleNodeInParentGroup(nodeData, parentData))
                 {
                     m_nodeWasSkipped = true;
-                    return true;
+                    return false;
```
alexmontAmazon, line 647: I know these magic bools were not done by you, but would you mind putting a comment here as to why we're returning false instead of true?

alexmontAmazon (approved):
hooray! I think we can get some more DPE testers now!

Code/Framework/AzFramework/AzFramework/DocumentPropertyEditor/Reflection/LegacyReflectionBridge.cpp
AMZN-daimini: Done!

PR #16355, opened 2023-07-11
Use general regex to find repo url

amzn-changml:
## What does this PR do?

Removes the Github url regex pattern to allow for usage with other Git services

## How was this PR tested?

Tested in sandbox Jenkins

### DIFF
--- a/scripts/build/Jenkins/Jenkinsfile
+++ b/scripts/build/Jenkins/Jenkinsfile
@@ -1192,7 +1192,7 @@ try {
                     timestamps {
                         repositoryUrl = scm.getUserRemoteConfigs()[0].getUrl()
                         // repositoryName is the full repository name
-                        repositoryName = (repositoryUrl =~ /https:\/\/(.*)\/(.*)\.git/)[0][1]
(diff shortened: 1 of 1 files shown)


brianherrera (approved):

scripts/build/Jenkins/Jenkinsfile
```diff
@@ -1192,7 +1192,7 @@ try {
                     timestamps {
                         repositoryUrl = scm.getUserRemoteConfigs()[0].getUrl()
                         // repositoryName is the full repository name
-                        repositoryName = (repositoryUrl =~ /https:\/\/github.com\/(.*)\.git/)[0][1]
+                        repositoryName = (repositoryUrl =~ /https:\/\/(.*)\/(.*)\.git/)[0][1]
```
lumberyard-employee-dm, line 1195: Should `.+` be used here instead to make sure there is at least one character between https:// and `/`.

```suggestion
                        repositoryName = (repositoryUrl =~ /https:\/\/(.+)\/(.+)\.git/)[0][1]
```

amzn-changml: Yup, that's fine. It'll still fail out if the URL is invalid, but better to be sure nothing weird is injected

PR #16356, opened 2023-07-11
Add integration tests for implicitly connecting slots

mddvisu:
## What does this PR do?

This PR adds two integration tests to test nodes with slots that make implicit connections:
- SlotConnecting_ImplicitConnections, which tests if implicit connections are being made correctly between two nodes that do not have a Slot Execution Map
- SlotConnecting_ImplicitConnectionsSlotExecutionMap, which tests if implicit connections are being made correctly with a more complex node that does have a Slot Execution Map

Both tests make ConfigurableUnitTestNodes and configure one to have an execution input that makes implicit connections, and another to have execution output(s). Each node is given data slots that are connected and disconnected with checks to make sure implicit connections are being properly made/removed.

## How was this PR tested?

Manual testing was done to ensure that the tests passed and failed when expected. The tests were run while the rest of the codebase was left unchanged which passed. I also tested by temporarily changing the UpdateCorrespondingImplicitConnections function to not work properly and the tests failed as expected.

### DIFF
--- a/Gems/ScriptCanvasTesting/Code/Source/Framework/ScriptCanvasTestFixture.h
+++ b/Gems/ScriptCanvasTesting/Code/Source/Framework/ScriptCanvasTestFixture.h
@@ -185,23 +185,17 @@ namespace ScriptCanvasTests
 
         ScriptCanvas::Graph* CreateGraph()
         {
-            if (m_graph == nullptr)
-            {
-                m_graph = aznew ScriptCanvas::Graph();
-                m_graph->Init();
-            }
-
+            AZ_Assert(!m_graph, "Only one graph should be created per test.");
+            m_graph = aznew ScriptCanvas::Graph();
+            m_graph->Init();
             return m_graph;
         }
 
         ScriptCanvasEditor::EditorGraph* CreateEditorGraph()
         {
-            if (m_graph == nullptr)
-            {
-                m_graph = aznew ScriptCanvasEditor::EditorGraph();
-                m_graph->Init();
-            }
-
+            AZ_Assert(!m_graph, "Only one graph should be created per test.");
+            m_graph = aznew ScriptCanvasEditor::EditorGraph();
+            m_graph->Init();
             return static_cast<ScriptCanvasEditor::EditorGraph*>(m_graph);
         }
 
@@ -210,9 +204,11 @@ namespace ScriptCanvasTests
             AZ::Entity* configurableNodeEntity = new AZ::Entity(entityName.c_str());
             auto configurableNode = configurableNodeEntity->CreateComponent<TestNodes::ConfigurableUnitTestNode>();
 
-            if (m_graph == nullptr)
+            AZ_Assert(m_graph, "A graph must be created before any nodes are created.");
+
+            if (!m_graph)
             {
-                CreateGraph();
+                return nullptr;
             }
 
             ScriptCanvas::ScriptCanvasId scriptCanvasId = m_graph->GetScriptCanvasId();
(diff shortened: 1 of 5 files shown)


Gems/ScriptCanvasTesting/Code/Source/Framework/ScriptCanvasTestFixture.h
```diff
@@ -193,6 +194,17 @@ namespace ScriptCanvasTests
...
+        {
+            if (m_graph == nullptr)
+            {
+                m_graph = aznew ScriptCanvasEditor::EditorGraph();
+                m_graph->Init();
+            }
+
+            return static_cast<ScriptCanvasEditor::EditorGraph*>(m_graph);
+        }
+
```
spham-amzn, line 207: This doesn't look safe. The `CreateGraph` method above will create an instance of `ScriptCanvas::Graph` if `m_graph` does not exist. But this function will similarly create an instance of `ScriptCanvasEditor::EditorGraph` if it doesn't exist, so what you get will depend on who gets called first. Should the editor graph be stored in a different variable?

spham-amzn (changes requested):

Gems/ScriptCanvasTesting/Code/Source/Framework/ScriptCanvasTestFixture.h
mddvisu: Each test that uses CreateGraph() or CreateEditorGraph() should really only use one or the other once to determine the type of graph the test is working with. And I don't see much point in converting between them in the middle of a test, since an EditorGraph is a Graph. I probably shouldn't create a separate variable, because the other functions in ScriptCanvasTestFixture use m_graph specifically

Gems/ScriptCanvasTesting/Code/Source/Framework/ScriptCanvasTestFixture.h
```diff
@@ -245,6 +257,71 @@ namespace ScriptCanvasTests
             EXPECT_EQ(m_graph->CanCreateConnectionBetween(targetEndpoint, sourceEndpoint).IsSuccess(), isValid);
         }
 
+        // Test if there is an existing connection between the provided endpoints
+        void TestIsConnectionBetween(const ScriptCanvas::Endpoint& sourceEndpoint, const ScriptCanvas::Endpoint& targetEndpoint, bool isValid = true)
+        {
+            AZ::Entity* ent;
+
+            EXPECT_EQ(m_graph->FindConnection(ent, sourceEndpoint, targetEndpoint), isValid);
+            EXPECT_EQ(m_graph->FindConnection(ent, targetEndpoint, sourceEndpoint), isValid);
```
cgalvan, line 266: We can git rid of this second line because `FindConnection` is source/target agnostic.

Gems/ScriptCanvasTesting/Code/Tests/ScriptCanvas_Slots.cpp
```diff
@@ -286,6 +290,194 @@ TEST_F(ScriptCanvasTestFixture, SlotConnecting_ExecutionBasic)
...
+    using namespace ScriptCanvas;
+
+    auto editorGraph = CreateEditorGraph();
+
+    // These vectors store each "set" of data output endpoints that correspond with one execution out
+    AZStd::vector<Endpoint> dataOutEndpointSet1;
+    AZStd::vector<Endpoint> dataOutEndpointSet2;
+    AZStd::vector<Endpoint> dataOutEndpointSet3;
+    AZStd::vector<Endpoint> dataOutEndpointSet4;
+    AZStd::vector<Endpoint> dataOutEndpointSet5;
```
cgalvan, line 328: I see the comments below explaining the scenario these sets are used for testing. Might be worth adding a little blurb here for them as well so folks can see right away what these are used for, since set1/set2/etc... isn't prescriptive on it's own.

cgalvan (approved):

spham-amzn (approved):

PR #16357, opened 2023-07-11
Fix various issues related to iOS build and run on Iphone 8

moudgils:
## What does this PR do?

- Crash related to incorrect shader language. Explicitly setting it to metal shader language 2.2 as oppose to it picking one based not he operating system of the Mac where AP is running
- Crash related incorrect deployment target. Explicitly setting the OS version to iOS 14.0.
- Crash related to Imgui - Fixed Imgui's UV index stream to hold 4 byte integers as oppose to 1 byte as the drivers do not like that. Other option was to uncompress explicitly in the shader but it wasn't worth it to save 48 bytes. 
- Crash related to incorrect scissors when writing to non zero mip - Add support to clamp scissors based on the mip level of the render target
- Address Slow AP speeds by increasing thread priority - Increased thread priority for Apple to help AP processing.
- Address iOS compile/link errors

## How was this PR tested?

Tested ASV on Mac and iOS

### DIFF
--- a/Code/Framework/AzCore/Platform/Common/Apple/AzCore/std/parallel/internal/thread_Apple.cpp
+++ b/Code/Framework/AzCore/Platform/Common/Apple/AzCore/std/parallel/internal/thread_Apple.cpp
@@ -70,6 +70,9 @@ namespace AZStd
         // QOS_CLASS_BACKGROUN         |           0          |           4          |  Will be prevented from using whole core.
         uint8_t GetDefaultThreadPriority()
         {
+            // We currently set the thread priority somewhere in the middle of QOS_CLASS_DEFAULT by default, since our threads generally
+            // don't need to finish in a single frame. Set this value higher if the default thread priorities are causing the threads to
+            // get starved too often.
             return 25;
         }
     }
--- a/Code/Framework/AzCore/Platform/Common/Apple/AzCore/std/parallel/internal/thread_Apple.cpp
+++ b/Code/Framework/AzCore/Platform/Common/Apple/AzCore/std/parallel/internal/thread_Apple.cpp
@@ -70,7 +70,7 @@ namespace AZStd
         // QOS_CLASS_BACKGROUN         |           0          |           4          |  Will be prevented from using whole core.
         uint8_t GetDefaultThreadPriority()
         {
-            return 10;
+            return 25;
         }
     }
 }
--- a/Gems/Atom/Feature/Common/Code/Source/ImGui/ImGuiPass.cpp
+++ b/Gems/Atom/Feature/Common/Code/Source/ImGui/ImGuiPass.cpp
@@ -535,7 +535,11 @@ namespace AZ
             desc.m_byteCount = 64;
             desc.m_bufferData = instanceData;
             m_instanceBuffer = RPI::BufferSystemInterface::Get()->CreateBufferFromCommonPool(desc);
-            m_instanceBufferView = RHI::StreamBufferView(*m_instanceBuffer->GetRHIBuffer(), 0, 64, 4);
+            m_instanceBufferView = RHI::StreamBufferView(
+                *m_instanceBuffer->GetRHIBuffer(),
+                0,
+                aznumeric_cast<uint32_t>(desc.m_byteCount),
+                aznumeric_cast<uint32_t>(desc.m_elementSize));
 
             ImGui::NewFrame();
             AzFramework::InputChannelEventListener::Connect();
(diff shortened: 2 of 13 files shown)


Code/Framework/AzCore/Platform/Common/Apple/AzCore/std/parallel/internal/thread_Apple.cpp
```diff
@@ -70,7 +70,7 @@ namespace AZStd
         // QOS_CLASS_BACKGROUN         |           0          |           4          |  Will be prevented from using whole core.
         uint8_t GetDefaultThreadPriority()
         {
-            return 10;
+            return 25;
```
lumberyard-employee-dm, line 73: So this value is somewhere between DEFAULT and USER_INITIATED?

lumberyard-employee-dm (approved):

Code/Framework/AzCore/Platform/Common/Apple/AzCore/std/parallel/internal/thread_Apple.cpp
moudgils: yeah the thread is now QOS_CLASS_DEFAULT instead of QOS_CLASS_UTILITY. Apple docs want you to use their api (i.e dispatch_queues) where you can specify the level and OS decides the priority. In our case we are using pthread so I am picking 25  to be in the middle.

mbalfour-amzn: A comment explaining the rationale of picking 25 as the value would be nice. Otherwise, future code maintainers won't understand why to leave or change the value. Maybe something like:
```
// We currently set the thread priority somewhere in the middle of QOS_CLASS_DEFAULT by default, since our threads generally don't need to finish in a single frame. 
// Set this value higher if the default thread priorities are causing the threads to get starved too often.
```
As a side question, if this is for all our job manager threads, wouldn't USER_INITIATED be more appropriate? The description of "Asynchronous / cross-frame work" sounds more like what we use the job manager for by default than "streaming / multiple frames deadline".

Gems/Atom/Feature/Common/Code/Source/ImGui/ImGuiPass.cpp
```diff
@@ -527,15 +527,15 @@ namespace AZ
...
             desc.m_poolType = RPI::CommonBufferPoolType::StaticInputAssembly;
             desc.m_bufferName = "InstanceBuffer";
-            desc.m_elementSize = 1;
-            desc.m_byteCount = 16;
+            desc.m_elementSize = 4;
+            desc.m_byteCount = 64;
             desc.m_bufferData = instanceData;
             m_instanceBuffer = RPI::BufferSystemInterface::Get()->CreateBufferFromCommonPool(desc);
-            m_instanceBufferView = RHI::StreamBufferView(*m_instanceBuffer->GetRHIBuffer(), 0, 16, 1);
+            m_instanceBufferView = RHI::StreamBufferView(*m_instanceBuffer->GetRHIBuffer(), 0, 64, 4);
```
mbalfour-amzn, line 538: Suggestion: Since these numbers appear to be tied to byte count and element size, why not make that explicit?

```suggestion
            m_instanceBufferView = RHI::StreamBufferView(*m_instanceBuffer->GetRHIBuffer(), 0, desc.m_byteCount, dest.m_elementSize);
```

Gems/Atom/RPI/Code/Source/RPI.Public/Pass/FullscreenTrianglePass.cpp
```diff
@@ -243,6 +243,32 @@ namespace AZ
         void FullscreenTrianglePass::SetupFrameGraphDependencies(RHI::FrameGraphInterface frameGraph)
         {
             RenderPass::SetupFrameGraphDependencies(frameGraph);
+            
+            uint16_t viewMinMip = RHI::ImageSubresourceRange::HighestSliceIndex;
+            for (const PassAttachmentBinding& attachmentBinding : m_attachmentBindings)
+            {
+                if (attachmentBinding.GetAttachment() != nullptr &&
+                    frameGraph.GetAttachmentDatabase().IsAttachmentValid(attachmentBinding.GetAttachment()->GetAttachmentId()) &&
+                    attachmentBinding.m_unifiedScopeDesc.GetType() ==RHI::AttachmentType::Image &&
```
mbalfour-amzn, line 252:
```suggestion
                    attachmentBinding.m_unifiedScopeDesc.GetType() == RHI::AttachmentType::Image &&
```

mbalfour-amzn (approved):

Code/Framework/AzCore/Platform/Common/Apple/AzCore/std/parallel/internal/thread_Apple.cpp
moudgils: Yeah I did think that this could be an issue but didnt want to go higher as AP could in theory interfere with Editor usage. For now I am just jumping one level and if cpu threading becomes an issue we can jump to another level. Would be nice to pass in priority when creating threads (at higher level) as then we could set the correct one based on it's usage.

PR #16358, opened 2023-07-12
Shader Management Console evolutions

siliconvoodoo:
## Disclaimer

<s>This diff is NOT intended to be a merge candidate. It is a discussion base draft, for how to implement properly the suggested changes.</s>
The current status is now a merge candidate.

## PR is

Adding features to Shader Management Console

- seeing proper asset types in the asset browser when selecting "shader" filter
- being able to start a variantlist from scratch from a .shader by using "create new variant list" in the context menu of asset browswer
- visual clue for row following using alternating colors
- easier widget clearing from cells after use (focus out)
- add new variant (row) from context action (or (+) button in the future)
- delete variant (row)
- order options by actual analyzed cost
- better UX to communicate what is an "empty" cell to the user (right now using GPU icon & tooltip)

## tests

Execute SMC through Multiplayer Sample repository as a project.

## environment

If you want to reproduce the situation, I suggest you edit:

`Gems/Atom/Asset/Shader/Registry/atom_shaders.setreg`
add line
`,"AzslCompilerOverridePath": "<yourclonepath>o3de-azslc/bin/win_x64/Release/azslc.exe"`

and

`Registry/AssetProcessorPlatformConfig.setreg`
remove `vulkan,null` from `tags` to accelerate asset processing


Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/AssetBrowserBus.h
```diff
@@ -260,6 +260,8 @@ namespace AzToolsFramework
                 return SourceFileDetails();
             }
 
+            virtual void CreateNewVariantListRequested(const char* /*fromShaderSourcePath*/) {}
+
```
siliconvoodoo, line 264: Obviously this shouldn't be here. It's a specific feature that creeps in a generic tool, so I need a suggestion for "dependency inversion".

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Search/Filter.cpp
```diff
@@ -362,6 +362,13 @@ namespace AzToolsFramework
 
         bool AssetGroupFilter::MatchInternal(const AssetBrowserEntry* entry) const
         {
+            AZStd::string ext;
+            AzFramework::StringFunc::Path::GetExtension(entry->GetVisiblePath().c_str(), ext, false);
+            if (m_group == "Shader" && ext.starts_with("shadervariant"))
+            {
+                return true;
+            }
```
siliconvoodoo, line 370: Here is a hardcoded change that results in the desired effect. To properly implement this behavior, we have a problem because the filter system is now checking the AssetGroup static string member which only exists on product assets, but the `shaderviarant` is a source.

Gems/Atom/Tools/AtomToolsFramework/Code/Source/AssetBrowser/AtomToolsAssetBrowserInteractions.cpp
```diff
@@ -83,6 +84,18 @@ namespace AtomToolsFramework
     void AtomToolsAssetBrowserInteractions::AddContextMenuActionsForSourceEntries(
         [[maybe_unused]] QWidget* caller, QMenu* menu, const AzToolsFramework::AssetBrowser::AssetBrowserEntry* entry)
     {
+        QFileInfo fileInfo(entry->GetFullPath().c_str());
+        QString extension = fileInfo.completeSuffix();
+        if (extension == "shader")
+        {
+            menu->addAction("Create new variantlist", [entry]()
```
siliconvoodoo, line 91: Same here, I'd need a "dependency inversion" method where the top tool can "inject" an action in this menu.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleTableView.cpp
```diff
@@ -13,9 +13,35 @@
...
+        {
+            BaseWidget::hidePopup();
+            if (m_onExit)
+            {
+                m_onExit();
+            }
+        }
+
+        std::function<void()> m_onExit;
+    };
```
siliconvoodoo, line 43: this thingie didn't work, so it might go away

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleTableView.cpp
```diff
@@ -25,10 +51,42 @@ namespace ShaderManagementConsole
         setEditTriggers(QAbstractItemView::NoEditTriggers);
         setSelectionBehavior(QAbstractItemView::SelectItems);
         setSelectionMode(QAbstractItemView::SingleSelection);
-        horizontalHeader()->setSectionResizeMode(QHeaderView::ResizeToContents);
+        verticalHeader()->setSectionResizeMode(QHeaderView::ResizeToContents);
+        horizontalHeader()->setSectionResizeMode(QHeaderView::Interactive);
```
siliconvoodoo, line 55: interactive is nice, but it gets unfortunately reset at each table rebuild, which is often if we're still tinkering with variant count.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleTableView.cpp
```diff
@@ -78,17 +139,17 @@ namespace ShaderManagementConsole
             m_shaderOptionCount, m_documentId, &ShaderManagementConsoleDocumentRequestBus::Events::GetShaderOptionDescriptorCount);
 
         // Only clear the table if the number of columns or rows have changed
-        if (rowCount() != m_shaderVariantCount || columnCount() != m_shaderOptionCount)
+        if (rowCount() != m_shaderVariantCount || (columnCount() + 1) != m_shaderOptionCount)
```
siliconvoodoo, line 142: Adding the deleter widgets column created the need for this index hack.
I personally feel it as risky and ugly, so if anybody sees a method to virtualize column access to hide that index hack under a safer accessor that'd be great.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleTableView.cpp
```diff
@@ -99,36 +160,89 @@ namespace ShaderManagementConsole
...
+                AZStd::sort(
+                    m_shaderOptionDescriptors.begin(),
+                    m_shaderOptionDescriptors.end(),
+                    [](const auto& a, const auto& b)
+                    {   //      big first (high score is more important)
+                        return a.GetCostEstimate() > b.GetCostEstimate();
+                    });
+            }
+            break;
+        }
```
siliconvoodoo, line 201: If someone sees a way to factorize this better, maybe through a lambda inside the lambda?

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleTableView.cpp
```diff
@@ -99,36 +160,89 @@ namespace ShaderManagementConsole
...
-            const auto& shaderOptionDescriptor = m_shaderOptionDescriptors[column];
-            setHorizontalHeaderItem(column, new QTableWidgetItem(shaderOptionDescriptor.GetName().GetCStr()));
+            const auto& shaderOptionDescriptor = m_shaderOptionDescriptors[column - 1];
+            auto* tableItem = new QTableWidgetItem(shaderOptionDescriptor.GetName().GetCStr());
+            tableItem->setToolTip(tr("cost %1").arg(shaderOptionDescriptor.GetCostEstimate()));
+            setHorizontalHeaderItem(column, tableItem);
         }
+        setHorizontalHeaderItem(0, new QTableWidgetItem(""));
+
+        QIcon emptyOptionIcon("D:/emptyoption.png");
```
siliconvoodoo, line 213: I have no idea of the resource submission proper process. Do we need to prepare multi DPI images, or can I just push that png somewhere in the repository, is there a impact on installation (I mean deployment of the app to user machine) etc..

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleTableView.h
```diff
@@ -36,11 +47,33 @@ namespace ShaderManagementConsole
...
+    };
+
+    class ShaderManagementConsoleContainer : public QVBoxLayout
+    {
+    public:
+        AZ_CLASS_ALLOCATOR(ShaderManagementConsoleContainer, AZ::SystemAllocator);
+
+        ShaderManagementConsoleContainer(QWidget* container, const AZ::Crc32& toolId, const AZ::Uuid& documentId, QWidget* parent);
+
+        ShaderManagementConsoleTableView m_tableView;
```
siliconvoodoo, line 74: about this, all other code places have dynamic per object allocations, and raw pointer passing to Qt. I find this disturbing but if that's Qt's culture, that class will need to be changed from storing its own widgets to simply newing them in the cpp

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleWindow.cpp
```diff
@@ -41,6 +41,25 @@ namespace ShaderManagementConsole
...
+            action,
+            &QAction::triggered,
+            this,
+            []()
+            {
+                
+            });
+        contextMenu.addAction(action);
+
+        contextMenu.exec(mapToGlobal(pos));
```
siliconvoodoo, line 62: I figured it'd be convenient to have a small set of "File" menu actions repeated directly from the blank space context menu. Not implemented yet.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleWindow.cpp
```diff
@@ -104,8 +123,8 @@ namespace ShaderManagementConsole
     {
         Base::CreateMenus(menuBar);
 
-        // Add statistic button
-        QAction* action = new QAction(tr("Generate Shader Variant Statistic..."), m_menuFile);
+        // Add statistics button
+        QAction* action = new QAction(tr("Generate shader variant statistics..."), m_menuFile);
```
siliconvoodoo, line 127: I must revert this, I didn't realize Title Case was a thing.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Document/ShaderManagementConsoleDocument.cpp
```diff
@@ -60,6 +60,33 @@ namespace ShaderManagementConsole
...
+                : m_shaderVariantListSourceData.m_shaderVariants.back().m_stableId + 1;
+            m_shaderVariantListSourceData.m_shaderVariants.push_back(variantInfo);
+            EndEdit();
+            AtomToolsFramework::AtomToolsDocumentNotificationBus::Event(
+                m_toolId, &AtomToolsFramework::AtomToolsDocumentNotificationBus::Events::OnDocumentModified, m_id);
+            return variantInfo.m_stableId;
+        }
+        return {};
+    }
+
```
gadams3, line 89: Refer to SetShaderVariantListSourceData to see all of the notifications that must be sent in order to update all of the observers and the inspector. This will prevent the crash that was demonstrated earlier.

This function could be simplified by checking to see if the document is open, copying the shader variant list source data, making any modifications, then calling SetShaderVariantListSourceData to update the data with all of the correct notifications. 

I would recommend moving the begin edit and end edit to the call site instead of calling them from within the function. That would allow the function to be called more than once as part of a batch between begin edit and edit calls.

This function should also be added to the shader management console document request bus and the behavior context for this class.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/ShaderManagementConsoleApplication.cpp
```diff
@@ -93,8 +95,11 @@ namespace ShaderManagementConsole
...
         // Overriding default document type info to provide a custom view
         auto documentTypeInfo = ShaderManagementConsoleDocument::BuildDocumentTypeInfo();
-        documentTypeInfo.m_documentViewFactoryCallback = [this](const AZ::Crc32& toolId, const AZ::Uuid& documentId) {
-            return m_window->AddDocumentTab(documentId, new ShaderManagementConsoleTableView(toolId, documentId, m_window.get()));
+        documentTypeInfo.m_documentViewFactoryCallback = [this](const AZ::Crc32& toolId, const AZ::Uuid& documentId)
+        {
+            auto* container = new QWidget;
+            new ShaderManagementConsoleContainer(container, toolId, documentId, m_window.get());
+            return m_window->AddDocumentTab(documentId, container);
         };
```
gadams3, line 103: What does the container widget do in this case?

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleWindow.cpp
gadams3: The main file menu has opened recent functionality built into it. I believe you can require a pointer to that menu and display it as a pop up menu here.

gadams3 (commented):
Initial round of comments. Will continue reviewing.

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Search/Filter.cpp
gadams3: See AtomToolsAssetBrowser::CreateFilter() const for how the atom tools set up the asset browser filter. I created a custom filter type that just uses a standard function/lambda. The default function excludes files in the cache folder. You can update the class to pass in a custom filter function to the constructor that includes the cache folder and your specific extensions. Try to preserve the default behavior and verify in the material editor and material canvas.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/ShaderManagementConsoleApplication.cpp
siliconvoodoo: Oh yes, it serves to adapt the excepted pointer type. AddDocumenTab takes a QWidget*. I got this solution from a stackoverflow answer. The SMCContainer class derives from Layout* so it wasn't compatible without using this dummy intermediary.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Document/ShaderManagementConsoleDocument.cpp
```diff
@@ -60,6 +60,33 @@ namespace ShaderManagementConsole
...
+        if (shaderAssetResult)
+        {
+            BeginEdit();
+            AZ::Data::Asset<AZ::RPI::ShaderAsset> shaderAsset = shaderAssetResult.GetValue();
+            AZStd::vector<AZ::RPI::ShaderOptionDescriptor> options = shaderAsset->GetShaderOptionGroupLayout()->GetShaderOptions();
+            AZ::RPI::ShaderVariantListSourceData::VariantInfo variantInfo;
+            for (auto& optionDescriptor : options)
+            {
+                variantInfo.m_options[optionDescriptor.GetName()] = ""; // all unset
+            }
```
gadams3, line 77: I'm not sure what other changes have been made with respect to shader options. Previously, I think the intention was to only add options to the map if the value was actually overridden. I don't remember exactly but I think that might have reduced the amount of serialized data. If the option was not in the map then it fell back to the default value. In that case the function could be reduced to resizing the shader variant vector and updating the stable IDs.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleTableView.cpp
```diff
@@ -25,10 +51,42 @@ namespace ShaderManagementConsole
...
+        this->setContextMenuPolicy(Qt::CustomContextMenu);
+        this->connect(this, &QTableWidget::customContextMenuRequested, this, &ShaderManagementConsoleTableView::ShowContextMenu);
+    }
+
+    void ShaderManagementConsoleTableView::mousePressEvent(QMouseEvent* e)
+    {
+        QTableWidget::mousePressEvent(e);
+        if (e->button() == Qt::RightButton)
+        {
+            ShowContextMenu(e->pos());
```
gadams3, line 69: Do we need this and the custom context menu requested connection?

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleTableView.cpp
```diff
@@ -25,10 +51,42 @@ namespace ShaderManagementConsole
...
+            this,
+            [this]()
+            {
+                ShaderManagementConsoleDocumentRequestBus::Event(
+                    m_documentId,
+                    &ShaderManagementConsoleDocumentRequestBus::Events::AddOneVariantRow);
+            });
+        contextMenu.addAction(action);
+
+        contextMenu.exec(mapToGlobal(pos));
```
gadams3, line 89: If you connect to the bus and override the functions listed below you can add this action to the main menu edit menu instead of managing your own context menu. This will also give you the opportunity to add other actions with shortcut keys. Rather than having the extra row and offset logic for the delete button or the context menu for the add button, they could both be in the edit menu with standard shortcut keys.

        // AtomToolsMainMenuRequestBus::Handler overrides...
        void CreateMenus(QMenuBar* menuBar) override;
        void UpdateMenus(QMenuBar* menuBar) override;

There is a bit of trickery when you have multiple documents open. You just have to connect or disconnect to the bus when the selected document changes. Material canvas GraphDocumentView does this, listening to the document notification bus for document changes, disconnecting on most events and only reconnecting if the document you care about is being opened.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleTableView.cpp
gadams3: There is an icon folder under Source/Window. Put your files there and update the QRC in the folder above. Regarding DPI, UI has moved most images away from PNG to SVG, scalable vector graphics, files.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleTableView.cpp
```diff
@@ -209,20 +342,30 @@ namespace ShaderManagementConsole
...
                 {
                     shaderVariant.m_options.erase(AZ::Name{optionItem->text().toUtf8().constData()});
+                    variantItem->setIcon(QIcon("D:/emptyoption.png"));
+                    variantItem->setToolTip(tr("runtime variable"));
                 }
                 else
                 {
                     shaderVariant.m_options[AZ::Name{optionItem->text().toUtf8().constData()}] = variantItem->text().toUtf8().constData();
+                    variantItem->setIcon({});
+                    variantItem->setToolTip("");
```
gadams3, line 359: Rather than using images to depict the empty value, one thing I had considered was just displaying the default value and changing the background color of the cell to denote whether or not it's overridden. This will allow the user to always see a value associated with the cell instead of having to guess what the default is.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleTableView.cpp
```diff
@@ -232,15 +375,51 @@ namespace ShaderManagementConsole
...
+        m_subLayout.addWidget(&m_sortLabel);
+        m_subLayout.addWidget(&m_sortComboBox);
+        m_subLayout.addStretch();
+        this->addLayout(&m_subLayout);
+        this->addWidget(&m_tableView);
+        connect(&m_sortComboBox, static_cast<void(QComboBox::*)(int)>(&QComboBox::currentIndexChanged), this, [this](int index) {
+            m_tableView.SetColumnSortMode(ColumnSortMode(index));
+        });
+
+    }
```
gadams3, line 421: This could also optionally be moved to the edit or view menu.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Document/ShaderManagementConsoleDocument.cpp
siliconvoodoo: I don't know if we should strive to keep the stableID compacts also. Because with the introduction of delete row, it creates gaps. Also, it's perfectly possible to make many rows that are similar in all points, that may mess up RHI internals if they work with equality/hashes. Maybe we need to run some verification to make the number of variants lean by ensuring uniquity. Also since verifying all against all risks to be quadratic, we can imagine to verify at save only. especially because during edition, it's ok to have temporary broken invariants.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleTableView.cpp
siliconvoodoo: I think the connection is useless it doesn't work, thus this manual popup.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleTableView.cpp
siliconvoodoo: I'm not too sure; because the runtime is free to set any value, probably even encouraged to not let it be the default. If it's just in runtime mode to remain at default, it can be hardcoded in the shader and not be an option. Or maybe populate a variant row in the list with all explicit defaults for the same effect. The unset options are expected to have unpredictable values because their purpose is to have a C++ runtime free to set them on the fly according to game state.
I wonder if even, an option that doesn't explicitly receives a value from the runtime, and lets it be the default, shouldn't be a sort of "uninitialized error", at least a "validation layer" warning.

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Search/Filter.cpp
siliconvoodoo: I attempted a change here to let this `GroupAssetFilter` pass sources asset. As it was, it only tolerated `Products`. Which is odd because of how you setup the root filter (with an `AND` policy) in `..::CreateFilter()` where it does the opposite, products asset can't pass the `!=Source !=Folder` test. According to first order logic, when an asset group filter is set, nothing should be displayed ever. But it's not what we observe. Anyways, since .shadervariantlist is a source asset, and that's what we're interested-in, in the SMC anway, I added a possible implementation to let source asset pass that filter. Not sure about it though, because it uses a fuzzy "get primary asset type".

Gems/Atom/Tools/AtomToolsFramework/Code/Source/AssetBrowser/AtomToolsAssetBrowserInteractions.cpp
siliconvoodoo: What I'd find most meaningful in terms of function and UX is that the .shader is NOT supported. This is not a valid source document type. The document type is meant to be .shadervariantlist and that's it.
BUT, `.shader` is indeed the side-file type to extract the option list and create an empty .shadervariantlist starter document.
I'd ideally mean to make this distinction clear so that .shader opening just end up in the json opening in a text editor.
But right click "generate new doc from..." (currently phrased "create new variantlist") would work.
Maybe I need to create a phony .shadervariantlist placeholder with just a "please kickstart me from <path.shader>" flag data inside only, and then the SMC::Open can "parse" that.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleWindow.cpp
siliconvoodoo: I did that, it works very poorly because of some 1 or 2 second delay in the construction of that menu, requiring to pass the mouse cursor a second time over the submenu to see it pop.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Document/ShaderManagementConsoleDocument.cpp
```diff
@@ -360,7 +389,13 @@ namespace ShaderManagementConsole
...
     {
-        return OpenFailed();
+        // return OpenFailed();  // TODO: it should be that. what we need is something like "PopulateNewVariantListFromShaderAsset"
+
+        AZ::RPI::ShaderVariantListSourceData shaderVariantListSourceData;
+        shaderVariantListSourceData.m_shaderFilePath = m_absolutePath;
+        SetShaderVariantListSourceData(shaderVariantListSourceData);
+        m_modified = {};
+        return OpenSucceeded();
     }
```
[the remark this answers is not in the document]
siliconvoodoo: In my sense the population from materials should be runnable after the document created empty. By calling the script. So We don't really need to rereport it to C++

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Document/ShaderManagementConsoleDocument.cpp
```diff
@@ -207,6 +235,7 @@ namespace ShaderManagementConsole
         documentType.m_documentFactoryCallback = [](const AZ::Crc32& toolId, const AtomToolsFramework::DocumentTypeInfo& documentTypeInfo) {
             return aznew ShaderManagementConsoleDocument(toolId, documentTypeInfo); };
         documentType.m_supportedExtensionsToOpen.push_back({ "Shader Variant List", AZ::RPI::ShaderVariantListSourceData::Extension });
+        documentType.m_supportedExtensionsToOpen.push_back({ "Shader Asset", AZ::RPI::ShaderSourceData::Extension });
         documentType.m_supportedExtensionsToSave.push_back({ "Shader Variant List", AZ::RPI::ShaderVariantListSourceData::Extension });
         return documentType;
     }
```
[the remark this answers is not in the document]
siliconvoodoo: Ok I used the "ExtensionsToCreate" and it behaves much better (like described in the previous comment about the phony file with a create-me flag).
Now just the thing is I got an automatic "create shader variant list..." context that is redundant with mine, except this one prompts for a save path.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Document/ShaderManagementConsoleDocument.cpp
siliconvoodoo: As discussed 2 weeks ago we can implement a "recompact" button.
But I'm also concerned about the builder, and the system that makes the economy of rebuilding all the variants by checking what stableID changed, if the stableID check doesn't incorporate the options's values, it will have lots of false negatives. We might have to extend the stableID with the whole key, even perhaps, replace it WITH the key to ensure natively unicity. But that could cause edition curiosities (like a difference between the view model and model)

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleTableView.cpp
siliconvoodoo: I see. We can do something like that in a second PR

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/ShaderManagementConsoleApplication.cpp
```diff
@@ -93,14 +100,50 @@ namespace ShaderManagementConsole
 
         // Overriding default document type info to provide a custom view
         auto documentTypeInfo = ShaderManagementConsoleDocument::BuildDocumentTypeInfo();
-        documentTypeInfo.m_documentViewFactoryCallback = [this](const AZ::Crc32& toolId, const AZ::Uuid& documentId) {
-            return m_window->AddDocumentTab(documentId, new ShaderManagementConsoleTableView(toolId, documentId, m_window.get()));
+        documentTypeInfo.m_documentViewFactoryCallback = [this](const AZ::Crc32& toolId, const AZ::Uuid& documentId)
+        {
+            auto* container = new QWidget;
+            new ShaderManagementConsoleContainer(container, toolId, documentId, m_window.get());
+            return m_window->AddDocumentTab(documentId, container);
```
moudgils, line 107: are you able to add a comment with the explanation about using a dummy intermediary

gadams3 (commented):
Here are some initial comments and requests. I'll continue reviewing tomorrow.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Document/ShaderManagementConsoleDocument.cpp
```diff
@@ -60,6 +63,97 @@ namespace ShaderManagementConsole
...
+        // WIP
+        //BeginEdit();
+
+        // Prepare a map of equivalences
+
+        // Prepare a whole new source data
+        //AZ::RPI::ShaderVariantListSourceData newSourceData{ m_shaderVariantListSourceData };
+
+        //EndEdit();
+        //SetAndNotifyModified();
```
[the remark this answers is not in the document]
siliconvoodoo: removed

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Document/ShaderManagementConsoleDocument.cpp
```diff
@@ -71,6 +165,7 @@ namespace ShaderManagementConsole
         {
             m_shaderAsset = shaderAssetResult.GetValue();
 
+            // TODO: check this comment. "material"?
```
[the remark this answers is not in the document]
siliconvoodoo: removed and following comment adapted

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Search/Filter.cpp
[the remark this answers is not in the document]
siliconvoodoo: ok, removed

Gems/Atom/RPI/Code/Include/Atom/RPI.Edit/Shader/ShaderVariantListSourceData.h
```diff
@@ -77,6 +78,9 @@ namespace AZ
             AZStd::string m_shaderFilePath; // .shader file.
             AZStd::vector<VariantInfo> m_shaderVariants;
 
+            //! non-mandatory list of option-names that have been gatherd from material properties.
+            //! This allows for UX hints in differenciating system options from material options.
+            AZStd::unordered_set<Name> m_materialOptionsHint;
```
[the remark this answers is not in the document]
siliconvoodoo: That method is subject to evolve in coming works, because for the "enumeration" script as you said, we'll need to mask out the material options, so this will need to be more robust, it will probably come from C++ in the following works.
I'd suggest to keep this like this for now.

Gems/Atom/Tools/AtomToolsFramework/Code/Source/Document/AtomToolsDocument.cpp
```diff
@@ -298,15 +298,15 @@ namespace AtomToolsFramework
...
+        AZ_TracePrintf("AtomToolsDocument", "Document opened: '%s' (uuid %s)\n", m_absolutePath.c_str(), m_id.ToString<AZStd::string>().c_str());
         AzToolsFramework::AssetSystemBus::Handler::BusConnect();
         AtomToolsDocumentNotificationBus::Event(m_toolId, &AtomToolsDocumentNotificationBus::Events::OnDocumentOpened, m_id);
         return true;
     }
 
     bool AtomToolsDocument::OpenFailed()
     {
-        AZ_TracePrintf("AtomToolsDocument", "Document could not opened: '%s'.\n", m_absolutePath.c_str());
+        AZ_TracePrintf("AtomToolsDocument", "Document could not open: '%s'.\n", m_absolutePath.c_str());
```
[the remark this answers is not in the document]
siliconvoodoo: Goccha.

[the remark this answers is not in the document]
siliconvoodoo: I tried to run `TEST_SUITE_main` is that the thing?
It doesn't run very well. Lots of python probles, dozens of fail in image processing test and it seems to stop at 24/100

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Document/ShaderManagementConsoleDocument.cpp
```diff
@@ -335,6 +431,15 @@ namespace ShaderManagementConsole
...
 
+    void ShaderManagementConsoleDocument::SetAndNotifyModified()
+    {
+        m_modified = true;
+        AtomToolsFramework::AtomToolsDocumentNotificationBus::Event(
+            m_toolId, &AtomToolsFramework::AtomToolsDocumentNotificationBus::Events::OnDocumentObjectInfoInvalidated, m_id);
+        AtomToolsFramework::AtomToolsDocumentNotificationBus::Event(
+            m_toolId, &AtomToolsFramework::AtomToolsDocumentNotificationBus::Events::OnDocumentModified, m_id);
+    }
+
```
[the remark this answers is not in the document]
siliconvoodoo: Good point. 👍  Done.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Document/ShaderManagementConsoleDocument.cpp
```diff
@@ -60,6 +63,97 @@ namespace ShaderManagementConsole
...
 
+    AZ::u32 ShaderManagementConsoleDocument::AddOneVariantRow()
+    {
+        auto shaderAssetResult =
+            AZ::RPI::AssetUtils::LoadAsset<AZ::RPI::ShaderAsset>(m_absolutePath,
+                                                                 m_shaderVariantListSourceData.m_shaderFilePath);
+        if (shaderAssetResult)
+
+        {
+            BeginEdit();
```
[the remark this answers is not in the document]
siliconvoodoo: ✔️

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Document/ShaderManagementConsoleDocument.cpp
```diff
@@ -60,6 +63,97 @@ namespace ShaderManagementConsole
...
+            for (auto& optionDescriptor : options)
+            {
+                variantInfo.m_options[optionDescriptor.GetName()] = ""; // all unset
+            }
+            variantInfo.m_stableId = m_shaderVariantListSourceData.m_shaderVariants.empty()
+                ? 0
+                : m_shaderVariantListSourceData.m_shaderVariants.back().m_stableId + 1;
+            m_shaderVariantListSourceData.m_shaderVariants.push_back(variantInfo);
+            EndEdit();
+            SetAndNotifyModified();
```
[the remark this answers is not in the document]
siliconvoodoo: ✔️

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/ShaderManagementConsoleRequestBus.h
```diff
@@ -42,6 +43,8 @@ namespace ShaderManagementConsole
         //! A wrapper to call GenerateRelativeSourcePath from python
         virtual AZStd::string GenerateRelativeSourcePath(const AZStd::string& fullShaderPath) = 0;
 
+        virtual AZ::RPI::ShaderOptionValue MakeShaderOptionValueFromInt(int value) = 0;
```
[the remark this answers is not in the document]
siliconvoodoo: In the absolute I don't think this is strictly needed. It's more that I didn't find how to generate the python binders for this type. I noticed that there are a bunch of *.pyi things that have partial python definitions for the "exported" types but I have no idea how they are generated.
For instance `user\python_symbols\azlmbr\rpi.pyi` maybe could contain a "constructor" function for ShaderOptionValue.
In the meantime I used the bus system that I know how will be exposed to python thanks to BehaviorContext to make the instances I needed in the enumeration script.

Will make a comment.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleTableView.cpp
```diff
@@ -52,12 +116,31 @@ namespace ShaderManagementConsole
...
+        return uiColumnIndex - 1;  // because column #0 is the "X" (deleters)
+    }
+
+    int ShaderManagementConsoleTableView::GetColumnsCount(CountQuery query) const
+    {
+        switch (query)
+        {
+            case CountQuery::ForUi:   return columnCount();
+            case CountQuery::Options: return UiColumnToOption(columnCount());
+        }
```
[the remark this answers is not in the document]
siliconvoodoo: No way. Changing the implementation to ternary.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Window/ShaderManagementConsoleTableView.cpp
```diff
@@ -99,36 +182,96 @@ namespace ShaderManagementConsole
...
+            auto* deleterButton = new QPushButton;
+            deleterButton->setText(u8"\u274C");  // cross sign
+            deleterButton->setToolTip(tr("delete row"));
+            connect(deleterButton, &QPushButton::clicked, this, [this, row](){
+                auto& vec = m_shaderVariantListSourceData.m_shaderVariants;
+                vec.erase(vec.begin() + row);
+                TransferViewModelToModel(CallOnModified);
+                AtomToolsFramework::AtomToolsDocumentNotificationBus::Event(
+                    m_toolId, &AtomToolsFramework::AtomToolsDocumentNotificationBus::Events::OnDocumentObjectInfoInvalidated, m_documentId);
+            });
```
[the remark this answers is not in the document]
siliconvoodoo: Indeed you're right. seems like the notifications moved in EndEdit fixed the need for that.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/ShaderManagementConsoleApplication.cpp
```diff
@@ -93,14 +100,52 @@ namespace ShaderManagementConsole
...
+                    AzFramework::StringFunc::Path::ReplaceExtension(savePath, AZ::RPI::ShaderVariantListSourceData::Extension);
+                    menu->addAction(
+                        "Create New Variant List (side by side to source)",
+                        [entry, savePath, this]()
+                        {
+                            AtomToolsFramework::AtomToolsDocumentSystemRequestBus::Event(
+                                m_toolId,
+                                &AtomToolsFramework::AtomToolsDocumentSystemRequestBus::Handler::CreateDocumentFromFilePath,
+                                entry->GetFullPath().c_str(),
+                                savePath);
```
gadams3, line 145: This looks like the open and save paths are the same. Should the save be empty or its extension changed to shadervariantlist? I could register a generic implementation of this menu command with the other asset browser document related commands that adds a “create X document” for files matching registered create from extensions.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/ShaderManagementConsoleApplication.cpp
```diff
@@ -320,4 +365,9 @@ namespace ShaderManagementConsole
             return "";
         }
     }
+
+    AZ::RPI::ShaderOptionValue ShaderManagementConsoleApplication::MakeShaderOptionValueFromInt(int value)
+    {
+        return AZ::RPI::ShaderOptionValue(value);
```
gadams3, line 371: Does shader option value have Behavior context mapping? Does it have a constructor that takes an integer registered with the behavior context that might make this function unnecessary?

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/ShaderManagementConsoleApplication.h
```diff
@@ -13,13 +13,15 @@
...
 #include <Window/ShaderManagementConsoleWindow.h>
+#include <AzToolsFramework/AssetBrowser/AssetBrowserBus.h>
 
 namespace ShaderManagementConsole
 {
     class ShaderManagementConsoleApplication
         : public AtomToolsFramework::AtomToolsDocumentApplication
         , private ShaderManagementConsoleRequestBus::Handler
         , private AzToolsFramework::EditorWindowRequestBus::Handler
+        , public AzToolsFramework::AssetBrowser::AssetBrowserInteractionNotificationBus::Handler
```
gadams3, line 24: I don't see where the asset browser interaction notification bus is being used, except for where it's connected. Can this be Removed or reverted if it's not necessary anymore?

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/ShaderManagementConsoleApplication.cpp
gadams3: That behavior is already implemented. Since you registered “shader” as a create from extension, don't you already get the “create shader variant list document” by right clicking on a shader in the asset browser? 

\ShaderManagementConsoleDocument should be renamed to ShaderVariantListDocument. Whether or not it gets reimplemented, the idea was that a document class could potentially support opening multiple file types, like a text or image editor document type would support opening several text and image formats.

You can also register a generic document type for shader source data so that you can open it directly in the editor like material canvas. That would not interfere with you being able to right click and create a shader variant list from a shader source data document. They would work together.

    void MaterialCanvasApplication::InitShaderSourceDataDocumentType()
    {
        // Register document type for editing shader source data files. This document type does not have a central view
        // and will display a label widget that directs users to edit using the inspector.
        auto documentTypeInfo = AtomToolsFramework::AtomToolsAnyDocument::BuildDocumentTypeInfo(
            "Shader Source Data",
            { "shader" },
            {},
            AZStd::any(AZ::RPI::ShaderSourceData()),
            AZ::RPI::ShaderSourceData::TYPEINFO_Uuid()); // Supplying ID because it is not included in the JSON file

        documentTypeInfo.m_documentViewFactoryCallback = [this]([[maybe_unused]] const AZ::Crc32& toolId, const AZ::Uuid& documentId)
        {
            auto viewWidget = new QLabel("Shader Source Data properties can be edited in the inspector.", m_window.get());
            viewWidget->setAlignment(Qt::AlignCenter);
            return m_window->AddDocumentTab(documentId, viewWidget);
        };

        AtomToolsFramework::AtomToolsDocumentSystemRequestBus::Event(
            m_toolId, &AtomToolsFramework::AtomToolsDocumentSystemRequestBus::Handler::RegisterDocumentType, documentTypeInfo);
    }

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Search/Filter.cpp
```diff
@@ -362,28 +363,27 @@ namespace AzToolsFramework
...
 
             if (m_group.compare("Other", Qt::CaseInsensitive) == 0 && group.isEmpty())
             {
                 return true;
             }
 
-            return (m_group.compare(group, Qt::CaseInsensitive) == 0);
+            bool match = m_group.compare(group, Qt::CaseInsensitive) == 0;
+            return match;
         }
```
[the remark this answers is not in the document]
siliconvoodoo: roger

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Document/ShaderManagementConsoleDocument.cpp
```diff
@@ -60,6 +62,77 @@ namespace ShaderManagementConsole
...
+            for (int column = 0; column < count; ++column)
+            {
+                auto& descriptor = GetShaderOptionDescriptor(column);
+                auto& optionName = descriptor.GetName();
+                auto indexIt = nameToHeaderIndex.find(optionName);
+                if (indexIt == nameToHeaderIndex.end())
+                {
+                    mapOfOptionNameToValues[optionName] = ""; // unset
+                }
+                else
```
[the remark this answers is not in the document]
siliconvoodoo: I was wondering about why we need to do this as well, but it appeared to me we don't have the descriptors actually. So that's why the RebuildTable system can't know about options that have no keys in the map.
We don't really conserve them after document opening. This is the reason why `AddOneVariantRow` needs to re-access the original shader asset using LoadAsset which I suspect isn't efficient.
Personally, I would prefer the behavior to be like you described. We would probably just need to save a `AZStd::vector<AZ::RPI::ShaderOptionDescriptor> m_descriptors;` somewhere in the document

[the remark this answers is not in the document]
gadams3: The descriptors are accessible from the shader asset.
The shader management console document loads the shader asset as part of opening a shader or shader variant list file.
This is enforced when calling ShaderManagementConsoleDocument::SetShaderVariantListSourceData, which does some additional validation and uses the descriptor to retrieve default values and prune empty options.
In an earlier comment, I think I recommended rewriting some of these functions in terms of ShaderManagementConsoleDocument::SetShaderVariantListSourceData.
For example, the add one row function could simply add the row directly to the shader variant list source data with the unique option ID, then call SetShaderVariantListSourceData. This might seem like overkill but should cover most of the edge cases and other validation. 
All of the table view logic was previously making changes to a local copy of the shader variant list source data then calling this function to update the document.
Either way, it can be addressed after the fact but I remember issues and comments from past PRs when I created the table view and originally moved everything from Python to C++.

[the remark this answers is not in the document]
siliconvoodoo: Your remark is correct as far as I know. What I'm concerned with is the opening of the shader asset, it's probably not that heavy but the console itself is already super slow, for example the python script that generates enumerations can prepare the enumerated table in memory for tens of thousands of variants (using the embedded pycharm interpreter) in a split second. But adding the lines in Qt takes hours. I left it run for 4 hours and it got to ~9000 table rows only (in profile mode).
We have serious performance problems so I'd rather not make it worse by opening the shader asset at each little delta on the data model. Maybe if we need it, we can keep a cached version of the descriptor vector.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/ShaderManagementConsoleApplication.cpp
siliconvoodoo: Yes we have a double behavior, the generic tool system already adds a "create new variant list" in the context menu of the asset browser, which is good. But it doesn't give a method to the client tool to specify the save path, thus it opens a file picker which risks to puzzle the user because it starts in %home%/document path so probably not where it needs to be.
If we can fix that, it would remove the need for the custom extension

siliconvoodoo: > This looks like the open and save paths are the same. Should the save be empty or its extension changed to shadervariantlist? I could register a generic implementation of this menu command with the other asset browser document related commands that adds a “create X document” for files matching registered create from extensions.

The `savePath` extension is changed line 136 to .shadervariantlist

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/ShaderManagementConsoleApplication.cpp
siliconvoodoo: ShaderOptionValue is an alias of Handle<uint32> which has some behavior context for IsValid and GetIndex but not the constructor.

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/ShaderManagementConsoleApplication.h
siliconvoodoo: Indeed it wasn't necessary (perhaps anymore). Removed it

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Search/Filter.cpp
```diff
@@ -13,6 +13,7 @@
 #include <AzToolsFramework/AssetBrowser/Search/Filter.h>
 #include <AzToolsFramework/AssetBrowser/Entries/AssetBrowserEntry.h>
 #include <AzToolsFramework/AssetBrowser/Entries/ProductAssetBrowserEntry.h>
+#include <AzToolsFramework/AssetBrowser/Entries/SourceAssetBrowserEntry.h>
```
siliconvoodoo, line 16: maybe i can remove that as well, seems useless now no?

gadams3 (approved):

Gems/Atom/RPI/Code/Source/RPI.Edit/Shader/ShaderVariantListSourceData.cpp
```diff
@@ -29,9 +29,10 @@ namespace AZ
                     ;
 
                 serializeContext->Class<ShaderVariantListSourceData>()
-                    ->Version(1)
+                    ->Version(2)
```
moudgils, line 32: Let's add a quick comment to induce the changes going with this version. We do this in a bunch of other places too.

Gems/Atom/RPI/Code/Source/RPI.Reflect/Shader/ShaderOptionGroupLayout.cpp
```diff
@@ -85,14 +85,16 @@ namespace AZ
             if (auto* serializeContext = azrtti_cast<SerializeContext*>(context))
             {
                 serializeContext->Class<ShaderOptionDescriptor>()
-                    ->Version(4)
+                    ->Version(5)
```
moudgils, line 88: Let's add a quick comment to induce the changes going with this version. We do this in a bunch of other places too.

Gems/Atom/Tools/AtomToolsFramework/Code/Source/AssetBrowser/AtomToolsAssetBrowserInteractions.cpp
```diff
@@ -8,6 +8,7 @@
 
 #include <Atom/RPI.Edit/Common/AssetUtils.h>
 #include <AtomToolsFramework/AssetBrowser/AtomToolsAssetBrowserInteractions.h>
+#include <AtomToolsFramework/Document/AtomToolsDocumentSystemRequestBus.h>
```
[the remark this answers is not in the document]
moudgils: Yes. Lets revert if not needed.

moudgils (approved):

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Document/ShaderManagementConsoleDocument.cpp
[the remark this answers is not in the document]
siliconvoodoo: In fact this fits nicely with the near term goals listed in here https://github.com/o3de/o3de/wiki/%5BAtom%5D-Shader-Management-Console-(SMC) (this PR answers many points coincidentally, I guess it was that natural to go this way)

Gems/Atom/Tools/ShaderManagementConsole/Code/Source/Document/ShaderManagementConsoleDocument.cpp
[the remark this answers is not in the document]
siliconvoodoo: Yes but it opens it only when the list is empty, I'm thinking it was intended as a detector for "the first time".
We should revisit it because it's not a very robust detector, we can cache what's necessary, or keep the whole asset itself as a memory object copy, I don't think it's heavy.

PR #16360, opened 2023-07-12
Fixes runtime crash that was introduced by

galibzon:
## What does this PR do?

Fixes runtime crash that was introduced by
PR https://github.com/o3de/o3de/pull/16326

It caused the AP to fail building
Gems/Atom/Feature/Common/Assets/Passes/PassTemplates.azasset

Which in turn caused the Editor to crash at startup in Debug mode

## How was this PR tested?

The AP doesn't complain anymore and The Editor doesn't crash anymore when loading.


AMZN-alexpete (approved):
Thanks @galibzon !

lumberyard-employee-dm (approved):

Issue #12037, opened 2022-09-16
[Android] DefaultProject application crashes with assert after coming back from device's screen timeout or screen being locked


LB-KacperKapusta:
**Describe the bug**
Android application of DefaultProject crashes after coming back from device's screen timeout or screen lock.

The following assert is printed in the Game.log (found on PC - /DefaultProject/user/log):
```
Trace::Assert
 C:/o3de/Gems/Atom/RHI/Vulkan/Code/Source\RHI/Vulkan.h(96): (509911637184) 'void AZ::Vulkan::AssertSuccess(VkResult)' 
(System) - ASSERT: Vulkan API method failed: Native window in use.
```
Please refer to the attached logs. Video is not attached as locking the screen stops the recording.

**Prerequisites**
- Follow Project Requirements and any of the Workflows from https://github.com/moraaar/sig-platform/blob/Andoid_LKG_Workflow/testing-guidance/workflow-tests/android/android-lkg-workflows.md.

**Steps to reproduce**
1. Launch the deployed DefaultProject app.
2. Wait until level is loaded.
3. Wait until screen timeout disables the screen and unlock the screen.
or alternatively
3. Lock the screen, unlock it and bring the app back.

**Expected behavior**
Application keeps running after screen is locked and reenabled.

**Actual behavior**
Application crashes with an assert when being brought back from the disabled state.

**Found in Branch**
Development 511a88c

**Device**
 - Device: Samsung S20+
 - OS: Android
 - Version 11 (API 30)

**Additional context**
[Game.log](https://github.com/o3de/o3de/files/9582686/Game.log)
[logcat_DefaultProject_crash.txt](https://github.com/o3de/o3de/files/9582862/logcat_DefaultProject_crash.txt)

LB-MateuszSzweda:
Issue verified as fixed on Development (b305c20) . Both locking the screen manually and waiting for screen timeout and unlocking the phone no longer crash the deployed app.

PR #16363, opened 2023-07-12
Android window destruction when suspending

akioCL:
## What does this PR do?

Add support for handling app suspend/resume for Android. The window gets destroyed and recreated when suspending and resuming the app on Android.

Fixes https://github.com/o3de/o3de/issues/12037

## How was this PR tested?

Run app on Android while suspending and resuming it.


AMZN-alexpete (approved):

Gems/Atom/Bootstrap/Code/Source/BootstrapSystemComponent.cpp
```diff
@@ -681,7 +704,9 @@ namespace AZ
             {
                 m_windowHandle = nullptr;
                 m_viewportContext.reset();
+#if AZ_TRAIT_BOOTSTRAPSYSTEMCOMPONENT_EXIT_ON_WINDOW_CLOSE
```
moudgils, line 707: Can you add a comment about this macro. Why is it disabled for android only?

moudgils (approved):

PR #16368, opened 2023-07-13
Fix invalid replace arg count error handling

### LINKED ISSUE #13644: As a developer I want the o3de.bat create-from-template Destination Path ERROR to Suggest -F
When using o3de.bat create-from-template to create a component for my project, I received an error that the project folder already exists. What's required is for the -f (force) command to be used.
`
[ERROR] o3de.engine_template: Destination path D:\prj\Day7\Gem\Code already exists.
`

It would be nice if the error was updated to suggest -f in the case of the path already existing.
`
[ERROR] o3de.engine_template: Destination path D:\prj\Day7\Gem\Code already exists; use -f to force creation.
`

Maybe update the documentation for creating a component to include -f. Any component someone makes will be in an already existing directory.

AMZN-alexpete:
## What does this PR do?
When the replacement arg count is odd, one of the pairs will be missing a value  resulting in a script error.  This  PR checks for this case and provides better error messaging and a unit test.

Related documentation here: https://docs.o3de.org/docs/user-guide/programming/components/create-component/

Related command:
```cmd
scripts\o3de.bat create-from-template -dp MyCode -dn MyTest -tn DefaultComponent -kr -r ${GemName} MyGem
```

Also fixes #13644
Also fixes an issue where a file download would silently fail on Windows if a file URI was provided in an invalid format.

## How was this PR tested?

New unit tests were added and commands were testing in Windows terminal (bash and cmd).
Used the new remote repo functionality to generate a repo, add a gem with an archive to it and test downloading it using a `file:///` uri in Project Manager and via the o3de CLI

### DIFF
--- a/scripts/o3de/o3de/engine_template.py
+++ b/scripts/o3de/o3de/engine_template.py
@@ -1243,10 +1243,18 @@ def create_from_template(destination_path: pathlib.Path,
     # if replacements are provided we need an even number of
     # replace arguments because they are A->B pairs
     if replace and len(replace) % 2 == 1:
+        replacement_pairs = []
+        for replace_index in range(0, len(replace), 2):
+            if replace_index + 1 < len(replace):
+                replacement_pairs.append(f' {replace[replace_index]} -> {replace[replace_index + 1]}')
+            else:
+                replacement_pairs.append(f' {replace[replace_index]} is missing replacement')
         logger.error("Invalid replacement argument pairs.  "
                      "Verify you have provided replacement match and value pairs "
                      "and your replacement arguments are in single quotes "
-                     "e.g. -r '${GemName}' 'NameValue'")
+                     "e.g. -r '${GemName}' 'NameValue'\n\n"
+                     "The current set of replacement pairs are:\n" + 
+                     ("\n".join(replacement_pairs)))
         return 1
 
     if template_name:
--- a/scripts/o3de/o3de/engine_template.py
+++ b/scripts/o3de/o3de/engine_template.py
@@ -1240,6 +1240,15 @@ def create_from_template(destination_path: pathlib.Path,
         logger.error(f'Template Name or Template Path must be specified.')
         return 1
 
+    # if replacements are provided we need an even number of
+    # replace arguments because they are A->B pairs
+    if replace and len(replace) % 2 == 1:
+        logger.error("Invalid replacement argument pairs.  "
+                     "Verify you have provided replacement match and value pairs "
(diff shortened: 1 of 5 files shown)


scripts/o3de/o3de/engine_template.py
```diff
@@ -1240,6 +1240,15 @@ def create_from_template(destination_path: pathlib.Path,
         logger.error(f'Template Name or Template Path must be specified.')
         return 1
 
+    # if replacements are provided we need an even number of
+    # replace arguments because they are A->B pairs
+    if replace and len(replace) % 2 == 1:
+        logger.error("Invalid replacement argument pairs.  "
+                     "Verify you have provided replacement match and value pairs "
+                     "and your replacement arguments are in single quotes "
+                     "e.g. -r '${GemName}' 'NameValue'")
```
lumberyard-employee-dm, line 1249: Can you also log the replacement pairs as well?
```suggestion
    if replace and len(replace) % 2 == 1:
        replacement_pairs = []
        for replace_index in range(0, len(replace), 2):
            if replace_index + 1 < len(replace):
                replacement_pairs.append(f'{replace[replace_index]} -> {replace[replace_index + 1]}')
            else:
                replacement_pairs.append(f'{replace[replace_index]} is missing replacement')
        logger.error("Invalid replacement argument pairs.  "
                     "Verify you have provided replacement match and value pairs "
                     "and your replacement arguments are in single quotes "
                     "e.g. -r '${GemName}' 'NameValue'\n"
                     "The current set of replacement pairs are as follows:\n"
                     "\n".join(replacement_pairs))
```

AMZN-alexpete: Good idea

tkothadev (approved):

scripts/o3de/o3de/utils.py
```diff
@@ -432,8 +432,17 @@ def download_progress(downloaded_bytes):
             logger.error(f'URL Error {e.reason} opening {parsed_uri.geturl()}')
             return 1
     else:
-        origin_file = pathlib.Path(parsed_uri.geturl()).resolve()
+        parsed_uri_path = urllib.parse.unquote(parsed_uri.path)
+        if isinstance(download_path, pathlib.PureWindowsPath):
+            # On Windows we want to remove the initial slash in front of the drive letter
+            if parsed_uri_path[0] == '/':
```
lumberyard-employee-dm, line 438: It is safer to use `startswith` here in case the path is empty.

PR #16373, opened 2023-07-13
Fixed issues with handling enums in DPE

### LINKED ISSUE #16063: DPE Inspector: Some enum type fields are empty
**Describe the bug**
Some enum type fields are empty with no values in the combo box.

Audio Preload "Load Type",
Audio Trigger "Obstruction Type", 
Multi-position Audio "Behavior Type",
PhysX Ball Joint - "Display Setup in Viewport"

**Steps to reproduce**

3. Create a new entity.
4. Add a `Audio Preload` component to the entity.

**Expected behavior**
"Load Type" field combo box has values.

**Actual behavior**
"Load Type" field combo box is empty, no values to choose from.

cgalvan:
## What does this PR do?

Fixes #16063 and #16076 

This resolves two separate enum handling issues in the DPE. Cases that reflect the enum to the EditContext need to also have them reflected to the SerializeContext if they want to provide the values through EnumAttribute attributes. The other case is for specialized enum types that don't use u64 as the base type, they need special handling to make sure their value list gets passed on properly and that the changed value uses the proper proxy value.

FixedEnums

## How was this PR tested?

Tested the repro cases in the listed issues and verified they now work as expected. Also tested various enum cases in the Editor.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyEditorAPI_Internals.h
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyEditorAPI_Internals.h
@@ -256,6 +256,19 @@ namespace AzToolsFramework
                     }
                     continue;
                 }
+                // Use the SerializedPath as the prxy class element name, which prior to this is always an empty string
+                // since it isn't used by the property handler. This helps when debugging the ConsumeAttribute method
+                // in the property handlers because the class element name gets passed as the debugName, so its easier
+                // to tell which property is currently processing the attributes.
+                else if (name == AZ::Reflection::DescriptorAttributes::SerializedPath)
+                {
+                    auto elementName = AZ::Dom::Utils::ValueToType<AZStd::string_view>(attributeIt->second);
+                    if (elementName.has_value())
+                    {
+                        m_proxyClassElement.m_name = elementName.value().data();
+                    }
+                    continue;
+                }
 
                 AZ::Crc32 attributeId = AZ::Crc32(name.GetStringView());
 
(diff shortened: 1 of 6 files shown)


Gems/LmbrCentral/Code/Source/Audio/EditorAudioPreloadComponent.cpp
```diff
@@ -25,6 +25,11 @@ namespace LmbrCentral
                 ->Field("Load Type", &EditorAudioPreloadComponent::m_loadType)
                 ;
 
+            serializeContext->Enum<AudioPreloadComponent::LoadType>()
```
alexmontAmazon, line 28: hmmm, we should probably document this somewhere if this is a new requirement...

alexmontAmazon (approved):
looks great, as long as people know to reflect their enums to the SC from now on

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyEditorAPI_Internals.h
```diff
@@ -256,6 +256,19 @@ namespace AzToolsFramework
                     }
                     continue;
                 }
+                // Use the SerializedPath as the prxy class element name, which prior to this is always an empty string
```
spham-amzn, line 259:
```suggestion
                // Use the SerializedPath as the proxy class element name, which prior to this is always an empty string
```

spham-amzn (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyEditorAPI_Internals.h
cgalvan: Oopsie, thank you!

Gems/LmbrCentral/Code/Source/Audio/EditorAudioPreloadComponent.cpp
cgalvan: Yeah we should note it wherever we are planning to have DPE documentation. This is the least used variant of all the different ways we support reflecting enum type display values. Most of the variants involve specifying the values in either a list, or generic values, or with EnumAttributes. This is the only place where you specify only the enum as a data element and rely on the values to be reflected separately, which for whatever reason only these 4 fields only did it in the EditContext, whereas the other several ones did it in both places, so it's one of those "I'm not actually sure how it was working before in the RPE."

PR #16380, opened 2023-07-14
Edit Download Specific Object Version on Project Manager and test_repo.py 

VeryKaylin-player1:
## What does this PR do?

This PR update the Project Manager Interface to support downloading different versions of gems based on the new remote repository template behavior.
- Previous Behavior: The previous behavior involved reading version data from the `version_data` field, where all version data was stored in the remote repository.
- New Behavior: The updated behavior reads the new remote repository template. Under this template, version one data is located under the `version` field, and any subsequent changes to the same gem are stored under the `version_data` field. 
- Changes to test_repo.py: The test_repo.py file has been updated to reflect the changes made to the repo.py functionality.

## How was this PR tested?

- The PR was tested by creating two instances of the same gem with different versions. The aim was to ensure that the Project Manager is capable of downloading a user-specified version of the gem.
- All unit tests passed successfully after adapting to the new changes.


scripts/o3de/o3de/repo.py
```diff
@@ -249,8 +249,13 @@ def get_object_versions_json_data(remote_object_list:list, required_json_key:str
 
         versions_data = remote_object_json_data.pop('versions_data', None)
         if versions_data:
+            version_found = False
             for version_json_data in versions_data:
+                if remote_object_json_data.get('version') == version_json_data.get('version'):
+                    version_found = True
                 object_json_data_list.append(remote_object_json_data | version_json_data)
```
spham-amzn, line 256: Unless its possible to have a `None` in the `versions_data` list, this line can remove `remote_object_json_data` from the `|` logic:
```
object_json_data_list.append( version_json_data)
```
since youve added a condition to add it anyways (if `version_found`),

spham-amzn (approved):

scripts/o3de/o3de/repo.py
VeryKaylin-player1: Version_data with the same version only contains what has changed from the current version, if version didn't change but other fields changed, we would still need to get the other fields from remote_object_json_data.

AMZN-alexpete (approved):

amzn-changml (approved):

PR #16382, opened 2023-07-14
Compact node basic operators

mddvisu:
## What does this PR do?

This is the first of a series of 3 PRs that will add new compact style nodes into the node palette that are designed to manipulate data without needing to connect them to an execution path:

image

This PR adds compact nodes that complete some basic operations. Here is a list of all that were added:
- +=
- -=
- *=
- /=
- *-1
- ++
- \-\-

All of these nodes are located in the node palette under `Compact/Basic Operators`:

image

An edit was also made to `ScriptCanvasNodeable_Source.jinja` and `ScriptCanvas_Nodeable_Macros.jinja` to allow for simple names of nodes to contain slashes (for the /= node).

In addition, two unit tests were added to test that each of these nodes work as expected.

## How was this PR tested?

As mentioned above, unit tests was added to make sure each node works as expected. The first unit test checks that each individual node works as expected and preforms the desired operation:
image

The second unit test chains compact nodes together with one another to ensure that the multiple implicit connections between them are being handled by the parser correctly in more complex graphs:
image


Gems/ScriptCanvas/Code/Include/ScriptCanvas/AutoGen/ScriptCanvas_Nodeable_Macros.jinja
```diff
@@ -11,10 +11,10 @@ SPDX-License-Identifier: Apache-2.0 OR MIT
 {# Macros for Nodeables #}
 
 {# CleanName #}
-{# Returns a clean name without quotations, commas, slashes or white space #}
+{# Returns a clean name without quotations, commas, or white space #}
 
 {%- macro CleanName(name) -%}
-{{name|replace(' ','')|replace('"','')|replace('/','')|replace(':','')}}
+{{name|replace(' ','')|replace('"','')|replace(':','')}}
```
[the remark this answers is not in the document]
mddvisu: There don't seem to be any other nodeables with slashes in their name. It makes sense to remove double quotes, but I am not sure why this macro needs to remove slashes, colons, commas, or even spaces for node names in case people want to include those characters in the name. I will likely remove it altogether when I get to cleaning up the jinja templates. At worst, this would only affect the name of the nodeable that appears in the editor, not the name that is used internally. So it wouldn't break existing nodeables

Gems/ScriptCanvas/Code/Include/ScriptCanvas/Libraries/Compact/BasicOperators/CompactNegativeNodeable.cpp
```diff
@@ -0,0 +1,17 @@
...
+ *
+ */
+
+#include "CompactNegativeNodeable.h"
+
+namespace ScriptCanvas::Nodeables
+{
+    float CompactNegativeNodeable::In(float a)
+    {
+        return a * -1;
```
spham-amzn, line 15: I dont think you need to use '*' here
```suggestion
        return -a;
```

Gems/ScriptCanvas/Code/Include/ScriptCanvas/Libraries/Compact/BasicOperators/CompactDecrementNodeable.ScriptCanvasNodeable.xml
```diff
@@ -0,0 +1,19 @@
...
+
+<ScriptCanvas Include="Include/ScriptCanvas/Libraries/Compact/BasicOperators/CompactDecrementNodeable.h" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+			  xsi:noNamespaceSchemaLocation="../../../AutoGen/ScriptCanvasNodeable.xsd">
+    <Class Name="CompactDecrementNodeable"
+        QualifiedName="Nodeables::CompactDecrementNodeable"
+        PreferredClassName="--"
+        Category="Compact/Basic Operators"
+        Namespace="ScriptCanvas"
+        Description="Decrements the input number by 1"
+		Preset="compact">
```
spham-amzn, line 11: nit: Can we align xml attributes? (here and all of the other nodeables)
```suggestion
    <Class Name="CompactDecrementNodeable"
           QualifiedName="Nodeables::CompactDecrementNodeable"
           PreferredClassName="--"
           Category="Compact/Basic Operators"
           Namespace="ScriptCanvas"
           Description="Decrements the input number by 1"
           Preset="compact">
```
Or at least fix the tabs in front of `Preset`

Gems/ScriptCanvas/Code/Include/ScriptCanvas/Libraries/Compact/BasicOperators/CompactDivideNodeable.ScriptCanvasNodeable.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<ScriptCanvas Include="Include/ScriptCanvas/Libraries/Compact/BasicOperators/CompactDivideNodeable.h" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+			  xsi:noNamespaceSchemaLocation="../../../AutoGen/ScriptCanvasNodeable.xsd">
+    <Class Name="CompactDivideNodeable"
+        QualifiedName="Nodeables::CompactDivideNodeable"
+        PreferredClassName="/="
+        Category="Compact/Basic Operators"
+        Namespace="ScriptCanvas"
+        Description="Divides first input number by the second input number"
```
spham-amzn, line 10: Should we instead of using `first` and `second`, use `parameter 'a'` and `parameter 'b'`?
```suggestion
        Description="Divides parameter 'a' by parameter 'b'"
```

spham-amzn: This comment applies to all of the related node descriptions

Gems/ScriptCanvas/Code/Include/ScriptCanvas/Libraries/Compact/BasicOperators/CompactNegativeNodeable.ScriptCanvasNodeable.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<ScriptCanvas Include="Include/ScriptCanvas/Libraries/Compact/BasicOperators/CompactNegativeNodeable.h" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+			  xsi:noNamespaceSchemaLocation="../../../AutoGen/ScriptCanvasNodeable.xsd">
+    <Class Name="CompactNegativeNodeable"
```
spham-amzn, line 5: Is this the right name of this node? It may imply that f(-1) = -1, since the number is still negative. Maybe the name should be `CompactNegateNumberNodeable` ?

Gems/ScriptCanvas/Code/Include/ScriptCanvas/Libraries/Compact/BasicOperators/CompactDivideNodeable.ScriptCanvasNodeable.xml
mddvisu: I am worried that using parameter `a` and `b` instead of `first` and `second` might be a little confusing since the labels for the inputs are hidden by default.

Gems/ScriptCanvas/Code/Include/ScriptCanvas/Libraries/Compact/BasicOperators/CompactNegativeNodeable.ScriptCanvasNodeable.xml
mddvisu: Hmm yeah I see what you are saying, but I feel like `CompactNegateNumberNodeable` could imply the same thing

mddvisu: Though, I think I can see how it might make it a bit clearer to the user, I'll change that

Gems/ScriptCanvas/Code/Include/ScriptCanvas/Libraries/Compact/BasicOperators/CompactDivideNodeable.ScriptCanvasNodeable.xml
spham-amzn: Well, if the labels are hidden, then 'a' and 'b' won't make sense then.

spham-amzn (approved):

AMZN-alexpete (approved):
Some of these nodeable files are so small I kinda wish they were all combined into one file.

PR #16383, opened 2023-07-14
Improve Android project generation error messages

spham-amzn:
## What does this PR do?

The Android project generation script `generate_android_project.py` will report misleading errors because it is not evaluating the responses from the exec calls to `sdkmanager` in the android sdk command line tools. A common android SDK setup issue is the host environment not having the corrent version of the JDK installed, which will lead to the following error:
```
Exception in thread "main" java.lang.UnsupportedClassVersionError: com/android/prefs/AndroidLocationsProvider has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
...
```
The error stream and result code was being ignored, so instead it would report a different error because the stdout will empty:
```
Invalid Android SDK Package Android SDK Platform 29: Bad package path platforms;android-29
```
This change will both raise an error indicating a failure with the call to `sdkmanager` and also provide a more concise error if possible in this scenario

## How was this PR tested?

Ran the generate project script without `JAVA_HOME` being set and validated the better messaging


cmake/Tools/Platform/Android/android_support.py
```diff
@@ -1650,7 +1662,7 @@ def _factory_available_update(package_map, item_components):
             package_map[item_components[0]] = AndroidSDKResolver.AvailableUpdate(item_components)
 
         # Use the SDK manager to collect the available and installed packages
-        result_code, result_stdout, result_stderr = self.sdk_manager.exec(['--list'], capture_stdout=True, suppress_stderr=True)
+        result_stdout = self.call_sdk_manager(['--list'], "retreiving package list")
```
AMZN-alexpete, line 1665:
```suggestion
        result_stdout = self.call_sdk_manager(['--list'], "retrieving package list")
```

AMZN-alexpete (approved):

cmake/Tools/Platform/Android/android_support.py
```diff
@@ -1631,6 +1631,18 @@ def __init__(self, android_sdk_path, command_line_tools_version):
         self.available_updates = {}
         self.refresh_sdk_installation()
 
+    def call_sdk_manager(self, arguments, action):
+        result_code, result_stdout, result_stderr = self.sdk_manager.exec(arguments, capture_stdout=True, suppress_stderr=False)
+        if result_code != 0:
+            # Provide a more concise error if a known execption is thrown indicating the wrong version of JAVA is set in the environment
+            if "java.lang.UnsupportedClassVersionError" in result_stderr:
+                err_msg = "The current installed java runtime is not compatible with the sdkmanager. Make sure that " \
+                          "java runtime version 11 is installed and JAVA_HOME is set to that path."
```
lumberyard-employee-dm, line 1640: Java runtime version 11 is quite old.
Why would the error message specific that version?

spham-amzn: The compiled JAR files that have the sdkmanager and other class files were compiled with that version, so JDK8 was incompatible. The Android studio installation includes a local subfolder with its own java runtime ([jbr](https://developer.android.com/build/jdks#jdk-android-studio)), and is set in the studio.jdk config file, but when running from command line, it relies on %PATH% or %JAVA_HOME%.

lumberyard-employee-dm: Is there a way to parameterized that version then?

lumberyard-employee-dm (approved):

cmake/Tools/Platform/Android/android_support.py
spham-amzn: I don't think parameterizing the version is the right approach, nor is calling out a specific version. I'll look into the right way to determine what the expected version of java is required to run `sdkmanager.bat` and provide that message instead

byrcolin (approved):

PR #16384, opened 2023-07-14
Fix AB's Hide Engine Toggle

### LINKED ISSUE #16370: Bug Report - Assets outside the project folder are hidden by default, Hide engine folders hides more than engine folders and is on by default
**Describe the bug**
Asset browser hides all folders except the project folder by default AND `Hide engine folders` filter is on by default which is hiding all folders outside the project, which it shouldn't and was supposedly fixed in https://github.com/o3de/o3de/pull/15769/files#top
Related Issue https://github.com/o3de/o3de/issues/15765

The current method for saying what assets to hide by default is too aggressive because it assumes the O3DE Gems and Assets folder do not have commonly used assets for users.  Examples:
1.  All the Atom default materials, skies textures, every sample gem that comes with the engine, all the LyShine examples, scripts, script libraries - all of this is excluded if we ignore the Gems folder inside O3DE.
1. Assets/Editor contains default font, default level prefab
1. Assets/Engine contains default config files, default cube map (unused?), additional fonts, utility Lua scripts, misc default textures and moon texture (old/unused?)
1. Registry contains numerous settings files - we don't currently have an editor for these, but technically these files are combined and shipped with your game, should you be able to find them in the Asset Browser? maybe?

I don't think it ever makes sense to hide the engine Gem's asset folders and I think we should do an asset clean up and move stuff to the correct folders before hiding the remaining "engine" folders that are processed by the AP and contain assets that might be shipped with your game. 

**Steps to reproduce**
1. Open the Editor with the AutomatedTesting project
2. Look at the Asset Browser
3. Toggle the "Hide engine folders" filter on and off

**Expected behavior**
All the available assets INCLUDING the engine assets are visible by default.  
Toggling the "Hide engine folders" only shows/hides the assets in `o3de/Assets/Engine`

**Actual behavior**
Assets from all Gems and the engine are hidden by default because "Hide engine folders" is on by default
Pressing the "Hide engine folders" toggle shows/hides every folder outside the project, not just the engine assets.
Why would you want to not show Gem assets????  Gems are one of the main ways we share assets between projects, we should make it very difficult to accidently hide them!

**Found in Branch**
Dev

1b4595b7c20ac1c00ce21ba51c3f1c4e0c9f515f

### LINKED ISSUE #16371: Bug Report - Asset Browser filter does not show assets in gems
**Describe the bug**
When typing the name of an asset in the Search box, it is not found if the asset is in a Gem

**Steps to reproduce**
1. Open the Editor and type 'geogap' in the Asset Browser search bar
2. See nothing is found.
3. Make sure the "Hide engine folders" is off in the Asset Browser menu
4. See nothing is found
5. Remove your search text and browse to `Gems/Atom/Feature/Common/Assets/LightingPreset/HighConstrast`
6. See multiple files with `geogap` in the name.

**Expected behavior**
Files in the Asset Browser are found with the search terms.

**Actual behavior**
Files are not found.

**Found in Branch**
development

1b4595b7c20ac1c00ce21ba51c3f1c4e0c9f515f

dtamkin1:
## What does this PR do?
Removes the Engine Filter, which was filtering out all folders inside of the o3de directory that were not the project. Instead, Only o3de/Assets/Engine and o3de/Registry are hidden by the "Hide Engine Assets" toggle

Fixes #16370 

## How was this PR tested?

Locally


forkercat (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Search/SearchWidget.cpp
```diff
@@ -175,10 +175,21 @@ namespace AzToolsFramework
...
 
-            AZ::IO::Path projectPath{ AZ::Utils::GetProjectPath() };
-            AZ::IO::Path enginePath{ AZ::Utils::GetEnginePath() };
-            auto engineFilter = new EngineFilter();
-            engineFilter->SetEngineAndProject(enginePath, projectPath);
+            const AZ::IO::Path enginePath{ AZ::Utils::GetEnginePath() };
+            AZ::IO::Path engineAssets = enginePath;
+            engineAssets = engineAssets.Append("Assets/Engine").LexicallyNormal();
+            AZ::IO::Path setregAssets = enginePath;
+            setregAssets = setregAssets.Append("Registry").LexicallyNormal();
```
cgalvan, line 182: Nit: This can be condensed a bit with the `/` operators that you can use with `AZ::IO::Path`
```suggestion
            AZ::IO::Path engineAssets = (enginePath / "Assets" / "Engine").LexicallyNormal();
            AZ::IO::Path setregAssets = (enginePath / "Registry").LexicallyNormal();
```

cgalvan (approved):

lsemp3d (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Search/SearchWidget.cpp
```diff
@@ -175,10 +175,21 @@ namespace AzToolsFramework
...
+
+            auto filterFn = [engineAssets, setregAssets](const AssetBrowserEntry* entry)
+            {
+                AZ::IO::Path absolutePath = entry->GetEntryType() == AssetBrowserEntry::AssetEntryType::Product
+                    ? entry->GetParent()->GetFullPath()
+                    : entry->GetFullPath();
+
+                return !(absolutePath.IsRelativeTo(engineAssets) || absolutePath.IsRelativeTo(setregAssets));
+            };
```
AMZN-alexpete, line 191: Correct me if I'm wrong but are we not doing a lot of string create/copies here?
Would it be more efficient if AssetBrowserEntry could provide GetFullPathView that returned an AZ::IO::PathView that we could use for the IsRelativeTo() call?  @lumberyard-employee-dm may have some additional thoughts but if we are running this filter on every file and folder it could be expensive no?

lumberyard-employee-dm: Yeah an `AZ::IO::PathView` has the IsRelativeTo function which requires no string copies.

An alternative as well is to use an `const AZ::IO::Path&` variable if the `GetFullPath` function return as a const reference.

lumberyard-employee-dm: Update: It does not, it returns an AZStd::string [by value](https://github.com/o3de/o3de/blob/b4ca369207868ba025f73edd8de71769610b9b37/Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Entries/AssetBrowserEntry.cpp#L193-L204).

Also because the function constructs a new path on the fly, returning an `AZ::IO::PathView` would be result in a dangling reference, so adding `GetFullPathView()` function should NOT be done.

The best way to avoid heap allocations is to return an `AZ::IO::FixedMaxPath`.
However a string copy would still need to occur.

If the desire is to avoid string copies, then the AssetBrowserEntry class would need to be update to cache off the full path as a member instead of performing an on the fly call to ResolvePath.

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Search/SearchWidget.cpp
lumberyard-employee-dm: Plus the paths don't need to be normalized.

AZ::IO::Path members functions can deal with paths that aren't normalized, normalization is for when the path is used outside of the AZ::IO::Path API.

So using the `IsRelativeTo` API below doesn't require normalization

dtamkin1: Sounds good, done!

Code/Framework/AzToolsFramework/AzToolsFramework/AssetBrowser/Search/SearchWidget.cpp
AMZN-alexpete: Thanks! For now I'm happy to get this fix checked in, but am wary of performance concerns after running into the AssetBrowser freezing the Editor for minutes lately when processing many changes.  Since it's not a simple as using a reference or path view then we should probably not delay this fix and we could even do some proper profiling to determine how much the filter is affecting Editor perf.

AMZN-alexpete (approved):

PR #16388, opened 2023-07-16
Improved Script Canvas node registration

l-sempe:
# Impactful Change

## Improved Script Canvas node registration.

This change was made to make it straightforward to add Script Canvas nodes, whether they are added at the ScriptCanvas gem or in standalone gems the process is the same.

This changes replaces the ScriptCanvasDataRegistry system that created different collections of node depending on their function or behavior. This is unnecessary as Script Canvas nodes are all expected to be properly serialized and reflected types of O3DE.

This change relies on static time automatic registration of types, the only important thing to do when creating a Script Canvas node is to make sure that its _.generated.h_ is in fact included and compiled somewhere, otherwise, the node will be subject to the compiler's stripping unused and unreferenced code. The project will compile, but the node will not be available.

**Notable changes:**

- The ScriptCanvasDataRegistry AutoGen template and functionality is no longer a separate system, registering data types can be done within a .ScriptCanvasFunction.xml file. This consolidates the concept of reflecting functions and/or data to the Behavior Context through AzAutoGen.

- The ScriptCanvasDataRegistry was rewritten in order to reduce the explicit and verbose system, instead, there is one registration methodology and nearly all Script Canvas types are registered to the same, singular, registry within an AZ::EnvironmentVariable (Note: some exceptions remain, those will be addressed separately).

- The ScriptCanvasDataRegistry AzAutoGen templates for Functions, Grammar and Nodeables have been removed as they are no longer needed.

- The Script Canvas node palette has been updated to search for nodes within the new registry, however, the Library mechanism was left as-is, there are a few nodes that are manually added into that Library that could be ported in the future.

- The ScriptCanvas Nodeable Schema (.xsd) was updated to support all possible attributes

- All former registry includes and uses of the REGISTER_SCRIPTCANVAS_AUTOGEN_FUNCTION macro have been removed

- Few minor fixes, hiding the Nodeling type from the node palette, removed unused tags in .xml files, removed the word "Node" when creating function child entries for objects in the node palette, added newlines to the AzAutoGen copyright banner

This change was tested by running all unit tests and testing the MultiplayerGameSample

## Migration Guide

Transitioning to the new system is straightforward and requires removing more code than it does adding.

Open the CMakeLists.txt for your Gem

From the AUTOGEN_RULES section remove any of these lines if present:

```
*.ScriptCanvasData.xml,ScriptCanvasDataRegistry_Header.jinja,AutoGenDataRegistry.generated.h
*.ScriptCanvasData.xml,ScriptCanvasDataRegistry_Source.jinja,AutoGenDataRegistry.generated.cpp
*.ScriptCanvasFunction.xml,ScriptCanvasFunctionRegistry_Header.jinja,AutoGenFunctionRegistry.generated.h
*.ScriptCanvasFunction.xml,ScriptCanvasFunctionRegistry_Source.jinja,AutoGenFunctionRegistry.generated.cpp
*.ScriptCanvasGrammar.xml,ScriptCanvasGrammarRegistry_Header.jinja,AutoGenGrammarRegistry.generated.h
*.ScriptCanvasGrammar.xml,ScriptCanvasGrammarRegistry_Source.jinja,AutoGenGrammarRegistry.generated.cpp
*.ScriptCanvasNodeable.xml,ScriptCanvasNodeableRegistry_Header.jinja,AutoGenNodeableRegistry.generated.h
*.ScriptCanvasNodeable.xml,ScriptCanvasNodeableRegistry_Source.jinja,AutoGenNodeableRegistry.generated.cpp 
```

From any files in your gem remove any instances of these lines:

```cpp
#include <AutoGenDataRegistry.generated.h>
#include <AutoGenFunctionRegistry.generated.h>
#include <AutoGenNodeableRegistry.generated.h>
#include <AutoGenGrammarRegistry.generated.h>

REGISTER_SCRIPTCANVAS_AUTOGEN_DATA(ScriptCanvasStatic);
REGISTER_SCRIPTCANVAS_AUTOGEN_FUNCTION(ScriptCanvasStatic);
REGISTER_SCRIPTCANVAS_AUTOGEN_NODEABLE(ScriptCanvasStatic);
REGISTER_SCRIPTCANVAS_AUTOGEN_GRAMMAR(ScriptCanvasStatic);
```

From the file list .cmake files for your Gem remove any of these lines if present:

```
Include/ScriptCanvas/AutoGen/ScriptCanvasDataRegistry_Header.jinja
Include/ScriptCanvas/AutoGen/ScriptCanvasDataRegistry_Source.jinja
Include/ScriptCanvas/AutoGen/ScriptCanvasFunctionRegistry_Header.jinja
Include/ScriptCanvas/AutoGen/ScriptCanvasFunctionRegistry_Source.jinja
Include/ScriptCanvas/AutoGen/ScriptCanvasGrammarRegistry_Header.jinja
Include/ScriptCanvas/AutoGen/ScriptCanvasGrammarRegistry_Source.jinja
Include/ScriptCanvas/AutoGen/ScriptCanvasNodeableRegistry_Header.jinja
Include/ScriptCanvas/AutoGen/ScriptCanvasNodeableRegistry_Source.jinja
```

Rename any *.ScriptCanvasData.xml files to *.ScriptCanvasFunction.xml

Very important final step!

For any use of *.ScriptCanvasFunction.xml, open the .cpp file that implements the body of the functions defined in the xml file and #include the generated header file

For example:

AABB.cpp now has

```cpp
#include "Include/ScriptCanvas/Libraries/Math/AABB.generated.h"
```

This is important, without it, the automatic registration mechanism may get stripped by the compiler and your functions will not show up in the Script Canvas node palette.


Gems/ScriptCanvas/Code/Include/ScriptCanvas/AutoGen/ScriptCanvasAutoGenRegistry.h
```diff
@@ -6,89 +6,168 @@
...
+{
+public:
 
-        //! Init specified AutoGen registry by given name
-        static void Init(const char* registryName);
+    static inline AZStd::string Encode(AZStd::string input)
+    {
+        AZStd::hash<AZStd::string> hash_string;
+        auto h = hash_string(input);
+        return AZStd::string::format("%lX", h);
```
lumberyard-employee-dm, line 117: This is not the correct formatter.
`%lx` is only applicable on Linux/Android where size_t has a type of `unsigned long`.
It is not applicable on MacOS, iOS and Windows wheer size_t has a type of `unsigned long long`.

The correct formatter here is `%zX`
```suggestion
        return AZStd::string::format("%zX", h);
```

lumberyard-employee-dm (changes requested):

Gems/ScriptCanvas/Code/Include/ScriptCanvas/AutoGen/ScriptCanvasAutoGenRegistry.h
l-sempe: thank you!

Gems/ScriptCanvasTesting/Code/Source/Framework/ScriptCanvasTestUtilities.cpp
```diff
@@ -57,7 +57,7 @@ namespace ScriptCanvasTests
     {
         using namespace ScriptCanvas;
 
-        AZ_TEST_START_TRACE_SUPPRESSION;
+  //      AZ_TEST_START_TRACE_SUPPRESSION;
```
lumberyard-employee-dm, line 60: Is this code no longer throwing expected errors.
Because if not the commented trace suppression block should be removed.

l-sempe: oh, this was a test and I didn't clean it up, will fix

Gems/ScriptCanvasTesting/Code/Source/Framework/ScriptCanvasTestUtilities.cpp
```diff
@@ -83,22 +83,22 @@ namespace ScriptCanvasTests
 
         if (!reporter.IsGraphLoaded())
         {
-            ADD_FAILURE() << "Graph was not successfully loaded.";
+            ADD_FAILURE() << "Graph was not successfully loaded.\n" << reporter.GetFilePath().c_str();
```
lumberyard-employee-dm, line 86: We have a `PrintTo` function overload for GTest for O3DE string classes that allows pretty printing using AZStd types (https://github.com/o3de/o3de/blob/development/Code/Framework/AzTest/AzTest/Printers.h#L14-L31)

Gems/ScriptCanvas/Code/Editor/Framework/ScriptCanvasReporter.h
```diff
@@ -190,6 +194,7 @@ namespace ScriptCanvasEditor
         bool m_isParseAttemptMade = false;
         bool m_isReportFinished = false;
         bool m_processOnly = false;
+        AZStd::string m_filePath;
```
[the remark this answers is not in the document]
l-sempe: I'm having a few bigger problems I could use help with... any thoughts on this?

```
<19:01:30> [Error] (Serialize) - Cannot enumerate data from data element (element:{E48A0B26-B6B7-4AF3-9341-9E5C5C1F0DE8}) from parent data element (element:{75F9EC0D-D8D6-5410-BD79-960E22076B03}) with class name "AZStd::unique_ptr<Contract>" because the ClassData does not exist. This can indicate that the class is not reflected at the point of this call. If this is class is reflected as part of a gem check if that gem is loaded
```
the gem is loaded and I saw `Contract` reflected, I even tried adding some `AZStd::unique_ptr<Contract>` to trigger on demand reflection but still not working...

lsemp3d:
Abandoned in favor of #16678

PR #16406, opened 2023-07-19
Adding a new project centric export script, which also works as a project template

tkothadev:
## What does this PR do?
As part of the Project Export effort in this [RFC](https://github.com/o3de/sig-content/issues/123), this PR presents a standard project-centric method of exporting a project to a release layout directory. This is a continuation of the work found in this PR: https://github.com/o3de/o3de/pull/16285.

This one differs from the engine-centric approach in that all build artifacts by default are centered around the project folder (unless otherwise specified by the user). Most of the arguments are the same as https://github.com/o3de/o3de/pull/16285, and any differences can be found in the `parse_args` function defined at the bottom of the new export script.

This export script splits up the release package into separate folders, one for each of ServerLauncher, GameLauncher, and/or UnifiedLauncher. If the user decides any of these are irrelevant for their project, that launcher can be disabled, and won't show in release.

As well, this script replaces the old script in the Project Template system.
### Test 1: The Export Script
The first test checked the accuracy of the export script itself. This was done with a recent version of O3DE MultiplayerSample. I used the following batch script from a relevant current directory to perform the test:
```batch
@REM From current working directory C:\workspace\dev
git clone -b development https://github.com/o3de/o3de.git 
git clone -b development https://github.com/o3de/o3de-multiplayersample.git
git clone -b development https://github.com/o3de/o3de-multiplayersample-assets.git
cd o3de-multiplayersample-assets
git submodule update --init --recursive 
cd ..

call %~dp0\o3de\python\get_python

call %~dp0\o3de\scripts\o3de register --this-engine
call %~dp0\o3de\scripts\o3de register -p .\o3de-multiplayersample
call %~dp0\o3de\scripts\o3de register --all-gems-path .\o3de-multiplayersample-assets\Gems
call %~dp0\o3de\scripts\o3de export-project -es %~dp0\o3de\scripts\o3de\ExportScripts\export_standalone_monolithic_project_centric.py -pp %~dp0\o3de-multiplayersample -out %~dp0\output-game2 -cfg release -a zip -code -assets -nounified -ll INFO -sl %~dp0\o3de-multiplayersample\AssetBundling\SeedLists\BasePopcornFxSeedList.seed -sl %~dp0\o3de-multiplayersample\AssetBundling\SeedLists\GameSeedList.seed -sl %~dp0\o3de-multiplayersample\AssetBundling\SeedLists\VFXSeedList.seed -gf AWSCore -gpfp launch_client.cfg -spfp launch_server.cfg
```

This created a directory `C:\workspace\dev\output-game2`, which can be seen here:
image

The ServerPackage:
image

The GamePackage:
image

note: the `usr` folders only showed up after I did some playtesting. I confirmed that the server and the game launchers both work, and was able to play around with the test scene for 30 minutes.

I tested them from each package folder respectively with the following commands:
```batch
 cd MultiplayerSampleServerPackage\
 .\MultiplayerSample.ServerLauncher.exe --console-command-file=launch_server.cfg --net_udpDefaultTimeoutMs=20000 -sv_terminateOnPlayerExit=false -bg_ConnectToAssetProcessor=0 -bg_traceLogLevel=5

 cd MultiplayerSampleGamePackage\
 .\MultiplayerSample.GameLauncher.exe --console-command-file=launch_client.cfg --connect=127.0.0.1 --net_udpDefaultTimeoutMs=20000 -bg_ConnectToAssetProcessor=0
```
Sample Screenshot:
image

### Test 2: Template Copying
I confirmed that the DefaultProject Template and MinimalProject Template both generated the export script when a new project is created from ProjectManager.


Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
```diff
@@ -0,0 +1,321 @@
...
+
+
+# This script is meant to be run via o3de cli export-project command, not from Python directly!
+# This export script operates in a project-centric manner, meaning all build artifacts are centered around the project folder. 
+# In order to use this script, the command should look like the following (on windows use o3de.bat, on linux use o3de.sh):
+
+# cd path\to\o3de\engine
+# .\scripts\o3de.bat  export-project --export-scripts path\to\o3de\ExportScripts\export_standalone_monolithic_project_centric.py \
+# --project-path path\to\project --log-level INFO --output-path path\to\output --build-non-mono-tools --config release --archive-output zip \
+# --seedlist path\to\seedlist1.seed --seedlist path\to\seedlist2.seed <etc..>
```
lumberyard-employee-dm, line 33: Alternatively the current working directory can be in the project folder and the path to the engine can be specified.

```python
# Example invocation on Linux
'''
$ cd <project-path>
$ <engine-path>/scripts/o3de.sh export-project --export-scripts ExportScripts/export_standalone_monolithic.py \
 --project-path . --log-level INFO --output-path <install-path> --build-non-mono-tools --config release --archive-output xz \
 --seedlist ./Assets/seedlist1.seed --seedlist ./Assets/seedlist2.seed <etc..> 
'''
```

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
```diff
@@ -0,0 +1,321 @@
...
+# In order to use this script, the command should look like the following (on windows use o3de.bat, on linux use o3de.sh):
+
+# cd path\to\o3de\engine
+# .\scripts\o3de.bat  export-project --export-scripts path\to\o3de\ExportScripts\export_standalone_monolithic_project_centric.py \
+# --project-path path\to\project --log-level INFO --output-path path\to\output --build-non-mono-tools --config release --archive-output zip \
+# --seedlist path\to\seedlist1.seed --seedlist path\to\seedlist2.seed <etc..> 
+
+# The following parameters are optional, but may be helpful:
+#   seedlist: an O3DE seedlist for describing what assets of the project should be bundled
+#   gem-folder-name-to-copy: A Gem Folder from <ProjectRoot>/Gems that should be copied to the output location of the exported project
```
lumberyard-employee-dm, line 37: Not all projects have a Gems folder.
There are projects that have a `<ProjectRoot>/Gem` folder, `<ProjectRoot>/Code` folder or even no folder at all.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
```diff
@@ -0,0 +1,321 @@
...
+    process_command(["cmake", "--build", non_mono_build_path, "--target", "AssetProcessorBatch", "AssetBundlerBatch", "--config", "profile", "--", "/m"])
+
+def build_monolithic_code_modules( project_path: pathlib.Path,
+                        project_name: str,
+                        build_config: str,
+                        mono_build_path: pathlib.Path,
+                        should_build_game_launcher: bool,
+                        should_build_server_launcher: bool,
+                        should_build_unified_launcher: bool):
+    process_command(["cmake", "-B", mono_build_path, "-S", project_path, "-DLY_MONOLITHIC_GAME=1", "-DALLOW_SETTINGS_REGISTRY_DEVELOPMENT_OVERRIDES=0"])
```
lumberyard-employee-dm, line 62: The [ALLOW_SETTINGS_REGISTRY_DEVELOPMENT_OVERRIDES](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/CMakeLists.txt#L23-L24) should be configurable by the user.

They weren't given a choice as to whether they want to settings registry files to load from external sources when they ran the script.
I recommended adding an argparse option which defaults to `False` for the Cache variable

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
```diff
@@ -0,0 +1,321 @@
...
+
+    if should_build_game_launcher:
+        process_command(["cmake", "--build", mono_build_path, "--target", f"{project_name}.GameLauncher", "--config", build_config, "--", "/m"])
+    
+    if should_build_unified_launcher:
+        process_command(["cmake", "--build", mono_build_path, "--target", f"{project_name}.UnifiedLauncher", "--config", build_config, "--", "/m"])
+
+
+def kill_existing_processes(project_name: str):
+    o3de_process_names = ['o3de', 'editor', 'assetprocessor', f'{project_name.lower()}.serverlauncher', f'{project_name.lower()}.gamelauncher' ]
```
lumberyard-employee-dm, line 77: Killing the `o3de`, `editor` and `assetprocessor` applications aren't relevant, since none of those applications are being built.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
```diff
@@ -0,0 +1,321 @@
...
+
+    process_command(["cmake", "--build", non_mono_build_path, "--target", "AssetProcessorBatch", "AssetBundlerBatch", "--config", "profile", "--", "/m"])
+    process_command(["cmake", "--build", non_mono_build_path, "--target", f"{project_name}.Assets", "--config", "profile", "--", "/m"])
+
+def bundle_assets(project_path: pathlib.Path,
+                  selected_platform: str,
+                  seedlist_paths: List[str],
+                  non_mono_build_path: pathlib.Path):
+    asset_bundler_batch_path = non_mono_build_path / 'bin' / 'profile' / ('AssetBundlerBatch' + ('.exe' if platform.system().lower()=='windows' else ''))
+    asset_list_path = project_path / 'AssetBundling' / 'AssetLists'
```
lumberyard-employee-dm, line 108: I think this AssetBundling AssetList path should be configurable via an argparse option with a default value of `<project-path>/AssetBundling/AssetLists`.

Also python pathlib automatically normalize paths that contain multiple directories as part of one string so
even the following works
```suggestion
    asset_list_path = project_path / 'AssetBundling/AssetLists'
```
image

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
```diff
@@ -0,0 +1,321 @@
...
+    process_command(gen_engine_asset_list_command)
+
+    #generate the bundles. We will place it in the project directory for now, since the files need to be copied multiple times (one for each separate launcher distribution)
+    process_command([asset_bundler_batch_path, "bundles", "--maxSize", "2048", "--platform", selected_platform, "--allowOverwrites",
+                         "--outputBundlePath", bundles_path / f"game_{selected_platform}.pak",
+                         "--assetListFile", game_asset_list_path])
+    
+    process_command([asset_bundler_batch_path, "bundles", "--maxSize", "2048", "--platform", selected_platform, "--allowOverwrites",
+                         "--outputBundlePath", bundles_path / f"engine_{selected_platform}.pak",
+                         "--assetListFile", engine_asset_list_path])
```
lumberyard-employee-dm, line 132: The maxSize option here should be user configurable. 
It can default to 2GiB, but this is specifically a setting the package exporter will want to have full control over.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
```diff
@@ -0,0 +1,321 @@
...
+def create_launcher_layout_directory(project_path: pathlib.Path,
+                                     output_path: pathlib.Path,
+                                     platform: str,
+                                     mono_build_path: pathlib.Path,
+                                     bundles_to_copy: List[pathlib.Path],
+                                     gem_folder_names_to_copy: List[str],
+                                     project_file_patterns_to_copy: List[str],
+                                     build_config: str,
+                                     archive_output_format: str = "none",
+                                     logger: logging.Logger|None =None):
```
lumberyard-employee-dm, line 145: nitpick: missing space between = and None
```suggestion
                                     logger: logging.Logger|None = None):
```

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
```diff
@@ -0,0 +1,321 @@
...
+    for bundle in bundles_to_copy:
+        shutil.copy(bundle, output_cache_path)
+
+    for file in glob.glob(str(pathlib.PurePath(mono_build_path / 'bin' / build_config / '*.*'))):
+        shutil.copy(file, output_path)
+
+    for gem_folder_name in gem_folder_names_to_copy:
+        output_gem_path = output_path / 'Gems'/ gem_folder_name
+        os.makedirs(output_gem_path, exist_ok=True)
+        for file in glob.glob(str(pathlib.PurePath(mono_build_path / 'bin' / build_config / 'Gems' / gem_folder_name / '*.*'))):
```
lumberyard-employee-dm, line 163: Gem Folders names don't have to be unique and are not in the project gem case(notice the folder is called "Gem" in most cases).

What is unique for a Gem is it's name.
So the gem name should be used as the output folder to prevent two gems with the same folder name from colliding.

The Gem name can be query via getting the gem.json file by using by invoking [manifest.get_gem_json_data](https://github.com/o3de/o3de/blob/development/scripts/o3de/o3de/manifest.py#L948-L949) method.
```python
from o3de import manifest
# ...
gem_json_data = manifest.get_gem_json_data(gem_path=path_to_gem)
gem_name = gem_json_data.get('gem_name', gem_folder_name) if gem_json_data else gem_folder_name
```

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
```diff
@@ -0,0 +1,321 @@
...
+                                                [] if not args.project_file_patterns_to_copy else args.project_file_patterns_to_copy,
+                                                args.should_build_code,
+                                                args.should_build_assets,
+                                                not args.no_game_launcher,
+                                                not args.no_server_launcher,
+                                                not args.no_unified_launcher,
+                                                args.non_mono_build_path,
+                                                args.mono_build_path, 
+                                                args.archive_output,
+                                                o3de_logger)
```
lumberyard-employee-dm, line 321: nitpick: missing newline at the end of this script.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
```diff
@@ -0,0 +1,321 @@
...
+                                    "This should match what is in the project directory.")
+        parser.add_argument('-ggf', '--game-gem-folder-name-to-copy', type=str, dest='game_gem_folder_names_to_copy', action='append',
+                                help="The name of a Gem folder to include in release for a Game Package. Only specify folder name."
+                                    "This should match what is in the project directory.")
+        parser.add_argument('-sgf', '--server-gem-folder-name-to-copy', type=str, dest='server_gem_folder_names_to_copy', action='append',
+                                help="The name of a Gem folder to include in release for a Server Package. Only specify folder name."
+                                    "This should match what is in the project directory.")
+        parser.add_argument('-gpfp', '--game-project-file-pattern-to-copy', type=str, dest='game_project_file_patterns_to_copy', action='append',
+                                help="Any additional file patterns located in the project directory. File patterns will be relative to the project path. Will be included in GameLauncher.")
```
lumberyard-employee-dm, line 278: Terminology wise I think we should use the term `client` over `game`, since it pairs with the term `server` and is also applicable for content that is not a game (such as a running a simulation).

We can't change the `${project}.GameLauncher` name currently, but we can use the correct terminology going forward.

Any Thoughts @tkothadev @AMZN-alexpete @mbalfour-amzn @AMZN-Olex @gene-walt @spham-amzn ?

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
```diff
@@ -0,0 +1,321 @@
...
+# For more information on the available parameters for this script, check the parse_arguments function defined at the bottom of the file
+
+def  build_non_monolithic_code_modules( project_path: pathlib.Path,
+                                        non_mono_build_path: pathlib.Path):
+    #build the pre_requisite tools if needed
+    os.makedirs(non_mono_build_path, exist_ok=True)
+
+    process_command(["cmake", "-B", non_mono_build_path, "-S", project_path])
+
+    process_command(["cmake", "--build", non_mono_build_path, "--target", "AssetProcessorBatch", "AssetBundlerBatch", "--config", "profile", "--", "/m"])
```
spham-amzn, line 53: "/m" is visual studio specific. In fact, everything after `--` are arguments that are passed through to the underlying compiler. Instead of auto-building it, i think  and error should be raised to inform the user to build `AssetProcessorBatch` manually.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
tkothadev: I could see the use case with a networked game like MPS, but with an offline game, that terminology won't make sense. Game is probably the closest compromise implying client nature, which is why I suspect that stuck.

One counter point though to my argument there is that you could just use the general `gem-folder-name` or `project-pattern-file`, and prevent building server via `-noserver`, so all general file parameters only go to the game launcher. 

That would get the same effect, but it doesn't erase the fact that `ClientPackage` for a singleplayer game implies networking where there is none, imo.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
AMZN-Olex: Why do we need to copy Gems folder?

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
AMZN-Olex: I think Game Launcher is fine, since that works for single player games just fine. It's not necessarily a client. In fact, if you link Game Launcher with Multiplayer.Unified.Static it can serve as a host not just a client.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
tkothadev: I dont think requiring manual build of AssetProcessorBatch is appropriate for this script, since the purpose is to automate the entire build process, including dependencies. That being said, there is a `build-non-mono-tools` flag, so I could rework it so that if it's enabled and it does not detect the executable, it can rebuild it. Otherwise if it isnt there, it could error out.

I'll remove the visual studio specific parts.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
tkothadev: Only some projects do. MPS for instance has data in the AWSCore Gems folder, which was what necessitated the argument to be made. It's an optional parameter, so it can be omitted if that project doesnt use it.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
spham-amzn: I agree, `client` should be use in the terminology. I think there should be a deprecation path for `.GameLauncher` in the future, since O3DE can be used for simulations as well, in which the term `Game` would be confusing or misleading.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
lumberyard-employee-dm: Should the data be part of the Asset Bundler Seed List then?

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
```diff
@@ -0,0 +1,321 @@
...
+def kill_existing_processes(project_name: str):
+    o3de_process_names = ['o3de', 'editor', 'assetprocessor', f'{project_name.lower()}.serverlauncher', f'{project_name.lower()}.gamelauncher' ]
+    for process in psutil.process_iter():
+        processes_to_kill = []
+        # strip off .exe and check process name
+        if os.path.splitext(process.name())[0].lower() in o3de_process_names:
+            processes_to_kill.append(process)
+        
+        if processes_to_kill:
+            logging.error("The following processes are still running: "+[process.name() for process in processes_to_kill])
```
spham-amzn, line 85: nit:
```suggestion
            logging.error(f"The following processes are still running: {', '.join([process.name() for process in processes_to_kill])})
```

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
```diff
@@ -0,0 +1,321 @@
...
+            elif user_input.lower() == 'q':
+                quit()
+
+
+def process_assets( project_name: str,
+                    project_path: pathlib.Path,
+                    non_mono_build_path: pathlib.Path):
+    process_command(["cmake", "-B", non_mono_build_path, "-S", project_path])
+
+    process_command(["cmake", "--build", non_mono_build_path, "--target", "AssetProcessorBatch", "AssetBundlerBatch", "--config", "profile", "--", "/m"])
```
spham-amzn, line 100: Can't this be done by calling `build_non_monolithic_code_modules` instead?

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
```diff
@@ -0,0 +1,321 @@
...
+    game_asset_list_path = asset_list_path /  f'game_{selected_platform}.assetlist'
+    engine_asset_list_path = asset_list_path / f'engine_{selected_platform}.assetlist'
+    bundles_path = project_path / 'AssetBundling' / 'Bundles'
+
+    #generate the asset lists for the project
+    gen_game_asset_list_command = \
+        f"{asset_bundler_batch_path} assetLists --assetListFile {game_asset_list_path} --platform {selected_platform} --allowOverwrites"
+    for seed in seedlist_paths:
+        gen_game_asset_list_command += " --seedListFile " + str(seed)
+    process_command(gen_game_asset_list_command.split())
```
spham-amzn, line 119: Why are we concatenating to a string only to split it into an array anyways? `gen_game_asset_list_command` can just be initialized as a list and passed into `process_command` directly. If `asset_bundler_batch_path` has a space in it (even if its surrounded by double quotes) will be split as well and will cause problems with the command exec.

```suggestion
    #generate the asset lists for the project
    gen_game_asset_list_command = [
        asset_bundler_batch_path,
        'assetLists',
        '--assetListFile',
        game_asset_list_path,
        '--platform',
        selected_platform,
        '--allowOverwrites']
 
    for seed in seedlist_paths:
        gen_game_asset_list_command.append("--seedListFile")
        gen_game_asset_list_command.append(str(seed))

    process_command(gen_game_asset_list_command)
```

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
```diff
@@ -0,0 +1,321 @@
...
+
+    gen_engine_asset_list_command = \
+        f"{asset_bundler_batch_path} assetLists --assetListFile {engine_asset_list_path} --platform {selected_platform} --allowOverwrites --addDefaultSeedListFiles"
+    process_command(gen_engine_asset_list_command)
```
spham-amzn, line 123: I would apply the same list method mentioned above here as well

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
```diff
@@ -0,0 +1,321 @@
...
+#This code is only run by the 'export-project' O3DE CLI command
+if "o3de_context" in globals():
+    def parse_args(unprocessed_args: List[str]):
+        parser = argparse.ArgumentParser(
+                    prog='Exporter for standalone builds',
+                    description="Exports a project as standalone to the desired output directory with release layout. "
+                                "In order to use this script, the engine and project must be setup and registered beforehand. ")
+        parser.add_argument('-out', '--output-path', type=pathlib.Path, required=True, help='Path that describes the final resulting Release Directory path location.')
+        parser.add_argument('-cfg', '--config', type=str, default='profile', choices=['release', 'profile'],
+                                help='The CMake build configuration to use when building project binaries. If tool binaries are built with this script, they will use profile mode.')
```
spham-amzn, line 258:
```suggestion
                                help='The CMake build configuration to use when building project binaries. Tool binaries built with this script will always be built with the profile configuration')
```

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
spham-amzn: If we are going to perform long steps like building in this script, there should be at least a status messaging to stdout informing that the process is underway. Also, does the stdout/stderr streams captured or outputted to the console? or does it get eaten up by this script?

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
spham-amzn, line 132: Looks like these two processes can be parallelized. Have you considered having a thread to signal when a process_command returns instead of waiting and blocking for the result?

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
tkothadev: `process_command` does indicate a status that the process is underway, if the user sets the log level to `INFO` in the arguments. And yes, stdout/stderr is outputted to the console.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
tkothadev: I've thought about doing that, but in practice I've noticed the engine bundle does not take long at all (only a second or so). The bottle neck is almost exclusively with the game bundle in all of the build tests I've done. As such I dont find it necessary yet.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
tkothadev: The second issue is what I ran into during test builds. Locked files in the project folder b/c I accidentally left o3de/project manager running during builds halted the build process. By checking for this first, it doesn't leave the build process in an unfinished state (either it's able to start, or not at all).

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
tkothadev: So to clarify, the function on it's own should not auto-fail and require manual building. The purpose of the function is to build the tools if desired.

In the main project centric export function, we have the following code snippet:
```python

    if should_build_all_code:
        if should_build_non_mono_tools:
            build_non_monolithic_code_modules(project_path, non_mono_build_path)
        
        build_monolithic_code_modules(project_path, project_name, build_config, mono_build_path,
                                      should_build_game_launcher, should_build_server_launcher, should_build_unified_launcher,
                                      allow_registry_overrides)

```
So if the user desired to build non_mono_tools, then building asset processor and batch are expected. But in both cases, I can add a check for the existence of the tool binaries before proceeding on. If those two are not present at the expected path, then the process failed, and manual intervention is required.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
tkothadev: good catch! Adding that in now.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
spham-amzn: Instead of killing those processes, maybe the script should error out saying that you must close those processes before continuing

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
tkothadev: For now I will defer this discussion for a future PR. I will make an issue of this for future reference.

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
tkothadev: I think in those cases you can just pass in the seed list files manually using the `seedlist` parameter

Templates/DefaultProject/Template/ExportScripts/export_standalone_monolithic.py
lumberyard-employee-dm: > For now I will defer this discussion for a future PR. I will make an issue of this for future reference.

I am good with that decision

lumberyard-employee-dm (approved):

PR #16421, opened 2023-07-21
Volume Texture Support for AttachmentReadback

galibzon:
## What does this PR do?

New capabilities have been added to `AttachmentReadback` class:  

- Multiple Mip Levels can be read at once from the same attachment.
- For image attachments, Volume Texture (aka Texture3D) support was added too.

Added minor improvement to DdsFile class to suppport Volume Texture. With this improvement it is possible to generate a single DDS file that contains all mip levels of a Volume Texture.

In the future, I may add a new sample to AtomSampleViewer that reads multiple attachments at once.

TODO: With this change we can generate Volumetric DDS source assets but the Image builder still generates a single 2D StreamingImage. Not a big deal at the moment as nothing in the engine expects to load a 3D StreamingImage asset. But I'll add the support in another PR.

## How was this PR tested?

All changes were validated against the test suites in AtomSampleViewer project (and the AttachmentReadback sample work as expected). The new functionalities were validated in my Volumetric Clouds Gem, where a Noise Volume Texture with up to 9 mip levels is generated in a Compute shader and all mips are read back to CPU in a single pass.


Gems/Atom/RPI/Code/Source/RPI.Public/Pass/AttachmentsReadbackGroup.cpp
```diff
@@ -0,0 +1,367 @@
...
+#include <Atom/RHI/RHIUtils.h>
+#include <Atom/RHI/ScopeProducerFunction.h>
+
+#include <AzCore/Serialization/Json/JsonUtils.h>
+
+#include <AzCore/std/smart_ptr/make_shared.h>
+
+namespace AZ
+{
+    namespace RPI
```
AMZN-alexpete, line 33: very minor but we can save an indent level by using
```suggestion
namespace AZ::RPI
{
```

galibzon: Thanks for the feedback. I'm also waiting for @VickyAtAZ feedback if she has a preference to integrate the new functionality into the existing AttachmentReadback class or keep this original design that introduces a new subclass called AttachmentsReadbackGroup. 

The more I look at the code, the more "clunky" AttachmentsReadbackGroup feels to me, but if no one else feels the clunkiness, I'll leave it as is.

galibzon: @AMZN-alexpete , after talking to Vicky, she also agreed that it was better to add the new capabilities to `AttachmentReadback`. So I've made the corresponding changes and removed `AttachmentsReadbackGroup`. Decided not to do the `namespace AZ::RPI` suggestion because it would impact ALL the file and make code review harder.

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Pass/AttachmentReadback.h
```diff
@@ -43,14 +46,37 @@ namespace AZ
...
 
-            //! Readback a pass attachment.
+            //! Reads back a single pass attachment.
             //! @param readbackName is the name for the readback buffer. And it will be saved in ReadbackResult::m_name. If the name is empty, a name will be generated automatically.
             //! Return true if the pass attachment readback request was submitted.  
             //! The callback function set by SetCallback(CallbackFunction callback) will be called once the readback is finished
-            bool ReadPassAttachment(const PassAttachment* attachment, const AZ::Name& readbackName);
+            //! @param imageViewDescriptor If null, by default it is assumed that @attachment refers to Mip Level 0.
+            //!        When different than null, the image view descriptor can be used to specify which mip level is the attachment bound to.
+            bool ReadPassAttachment(const PassAttachment* attachment, const AZ::Name& readbackName, const RHI::ImageViewDescriptor* imageViewDescriptor = nullptr);
```
VickyAtAZ, line 57: RHI::ImageViewDescriptor is used to represent a group of continuous sub-resources. If you want to just read one sub-resource, it's better to use RHI::ImageSubresource

I think RHI::ImageViewDescriptor is useful if user want to get a group of continuous sub-resources with only need setup one struct for the purpose. And it can be useful for most of use cases which try to read all the mipmaps. 

But we may need to update the implementation so it can create multiple copy items rather than one. 

To support read multiple mipmaps which may not be continues, we can have another function use RHI::ImageSubresource as input. 

`bool ReadPassAttachment(const PassAttachment* attachment, const AZ::Name& readbackName, const RHI::ImageSubresource * imageSubresource= nullptr);`

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Pass/AttachmentReadback.h
```diff
@@ -43,14 +46,37 @@ namespace AZ
...
+                // Same semantic as ReadPassAttachment Arg 0.
+                const PassAttachment* m_attachment = nullptr;
+                // Same semantic as ReadPassAttachment Arg 1.
+                Name m_readbackName;
+                // With a properly set imageViewDescriptor we can readback
+                // a particular mip level.
+                RHI::ImageViewDescriptor m_imageViewDescriptor;
+            };
+            //! Reads back multiple pass attachments in one frame.
+            bool ReadPassAttachments(const AZStd::vector<ReadbackRequestInfo>& readbackAttachmentRequests);
```
VickyAtAZ, line 71: Do you intent to support a list of ReadbackRequestInfo which may have different PassAttachment?

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Pass/AttachmentReadback.h
VickyAtAZ: Maybe we should only have this one function 
`bool ReadPassAttachment(const PassAttachment* attachment, const AZ::Name& readbackName, const RHI::ImageViewDescriptor* imageViewDescriptor = nullptr)`

and remove the new `bool ReadPassAttachments(const AZStd::vector<ReadbackRequestInfo>& readbackAttachmentRequests);`function. 

This is because PassAttachment in a scope (ImageScopeAttachment) already contains ImageViewDescriptor information which makes sense if we want to use that when read the PassAttachment. 

If we only want to read one mip, we can construct the ImageViewDescriptor which only contains one mip.

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Pass/AttachmentReadback.h
galibzon: Yes.

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Pass/AttachmentReadback.h
galibzon: Great feedback, but I need a little bit of clarification... In my Compute Shader I have:
```
RWTexture3D m_mips[6]
```
I have several attachments "Mip0"..."Mip5", and I bind them independently where all of them are referring to the same Image, but different subresource, in other words one RHI::ImageViewDescriptor for each attachment:
```c++
AZ::RHI::ImageViewDescriptor viewDesc = AZ::RHI::ImageViewDescriptor::Create3D(m_texture3DAttachment->GetDescriptor().m_format,
                mipIndex, mipIndex, 0, pixelSizeForMip - 1);
binding->m_unifiedScopeDesc.SetAsImage(viewDesc);
``` 
So, the code snippet above is called for each Mip level, because each Mip is an attachment.  
So If I understand your point, it seems I could bind all Mips with a single attachment and assuming total mips = 6,
the code would look like this:
```
AZ::RHI::ImageViewDescriptor viewDesc = AZ::RHI::ImageViewDescriptor::Create3D(m_texture3DAttachment->GetDescriptor().m_format,
                0 /*top mip index*/, 5/*last mip index*/, 0, pixelSizeForMip - 1);
binding->m_unifiedScopeDesc.SetAsImage(viewDesc);
```

Gems/Atom/RHI/Vulkan/Code/Source/RHI/CommandList.cpp
```diff
@@ -224,6 +224,14 @@ namespace AZ
                 copy.imageExtent.height = descriptor.m_sourceSize.m_height;
                 copy.imageExtent.depth = descriptor.m_sourceSize.m_depth;
 
+                // [GFX TODO] It was found that after submitting the copy command, there could occur
```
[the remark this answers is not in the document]
galibzon: Done: https://github.com/o3de/o3de/issues/16444
And added it to the code.

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Pass/AttachmentReadback.h
```diff
@@ -25,8 +25,11 @@ namespace AZ
 
     namespace RPI
     {
-        //! A scope producer which reads back buffer or image attachments 
-        class AttachmentReadback
+        //! A scope producer which reads back one or more attachments to CPU memory.
```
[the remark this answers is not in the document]
galibzon: Done. Changed to:  
```
        //! Adds one or two scope producers ((the second is to decompose an MS Texture)
        //! which read back one or more attachments to CPU memory.
        //! Both buffer and image attachments are supported.
        //! In case of images it can also capture specific Mip Levels, as defined in the image view descriptors.
        //! Also, for images, Volume Texture image attachments (aka Texture3D) are supported too.
```

Gems/Atom/RPI/Code/Include/Atom/RPI.Public/Pass/AttachmentReadback.h
galibzon: @VickyAtAZ , per our conversation...
When **writing** to several Mip levels of the same UAV, we need to specify a unique ImageViewDescriptor for each mip level, and each mip level becomes its own attachment.
BUT when reading back, we can read back all Mips at once by referencing the attachment to Mip 0, and having  a single ImageViewDescriptor that describes the range of mip levels.  
  
Based on this clarification, I'm going to change the API, per your suggestion, to a single public function:  
```
bool ReadPassAttachment(const PassAttachment* attachment, const AZ::Name& readbackName, const RHI::ImageViewDescriptor* imageViewDescriptor = nullptr)
```
  
In the future if there's interest in being able to read back from different image attachments, the We'd have to modify `AZ::RPI::Pass` class to hold a list of `AttachmentReadback`s. But it is not necessary at the moment.

Gems/Atom/RPI/Code/Source/RPI.Public/Pass/AttachmentReadback.cpp
```diff
@@ -226,6 +225,12 @@ namespace AZ
                 {
                     m_bufferAttachmentByteSize = attachment->m_descriptor.m_buffer.m_byteCount;
                 }
+                m_readbackItems.push_back({});
+                auto& item = m_readbackItems.back();
+                for(uint32_t i = 0; i < RHI::Limits::Device::FrameCountMax; i++)
+                {
+                    item.m_readbackBufferArray.push_back(nullptr);
+                }
```
AMZN-alexpete, line 233: I haven't used `Data::instance<>`, can you use the `resize` method with a nullptr? or would `resize()` init with nullptr types?
```suggestion
                item.m_readbackBufferArray.resize(RHI::Limits::Device::FrameCountMax, nullptr);
```

AMZN-alexpete (approved):

Gems/Atom/RPI/Code/Source/RPI.Public/Pass/AttachmentReadback.cpp
galibzon: I like your idea. Will give it a try.

galibzon: Done. Looks better.

Gems/Atom/RPI/Code/Source/RPI.Public/Pass/AttachmentReadback.cpp
```diff
@@ -339,7 +355,10 @@ namespace AZ
                     }
                     if (m_callback)
                     {
-                        m_callback(GetReadbackResult());
+                        for (const auto& readbackItem : m_readbackItems)
+                        {
+                            m_callback(GetReadbackResult(readbackItem));
+                        }
```
VickyAtAZ, line 361: It would be better if the m_callback is only called once for all the mips. 
Although this requires change with ReadbackResult structure to include all the mips' data.

galibzon: I'm going to modify `struct ReadbackItem` with new member variables (the new list of data buffers), so the existing code base does not get impacted. And keep it as a one-time dispatched callback.

galibzon: Done. Now the callback is executed once.

VickyAtAZ (approved):
Great work!

PR #16446, opened 2023-07-25
Refactor to support Headless Server Launchers

spham-amzn:
## What does this PR do?

This is a refactor of the O3DE to support headless server launchers in accordance to the [Headless Server RFC](https://github.com/o3de/sig-network/blob/main/rfcs/rfc-net-20230821-1-headless-server-launcher.md)

**Overview of changes**
- `AzFramework` has been split to `AzFramework` and `AzFramework.NativeUI`. `AzFramework.NativeUI` will have the platform specific implementation to manage native windows and device input / output.
- All targets that end up using a native window and/or native input devices (including keyboard) will end up having a dependency `AzFramework.NativeUI` The main UI targets that will include:
  - Editor
  - AssetProcessor
  - Game/Server launchers
- Only the `<Game>.HeadlessServerLauncher` will omit this dependency
- Replaced original static `Implementation* Create()` in the `Implementation` pimpl classes with an equivalent `ImplementationFactory` instead. This was done to improve management of optional implementations through the use of AZ::Interface<> instead of always calling `Implementation::Create`
- Created new `AzFramework.UI` target, with a corresponding `AzFrameworkNativeUIModule` to manage the UI private implementations in platform-specific ways through PAL
- Created a new `AzFrameworkNativeUIModule` component that manages the lifecycle of the Platform-specific UI Pimpl classes for `AzFramework.UI`
- Updated the export scripts to support `<Game>.HeadlessServerLauncher`

## How was this PR tested?

- **Windows** Built [o3de multiplayer sample](https://github.com/o3de/o3de-multiplayersample) on window, tested with:
  - Editor : View starbase level in edit mode, launched game mode within the editor
  - MultiplayerSample.HeadlessServerLauncher.exe + MultiplayerSample.GameLauncher.exe : Launched `MultiplayerSample.HeadlessServerLauncher.exe --console-command-file=launch_server.cfg` and then launched `MultiplayerSample.GameLauncher.exe --console-command-file=launch_client.cfg`. Confirmed headless server console running, while client launched and can connect to play the game
  - MultiplayerSample.ServerLauncher.exe + MultiplayerSample.GameLauncher.exe: Confirmed server launcher running (with client window), and game launcher launched separate game window and can connect and play the game
- **Linux** Built [o3de multiplayer sample](https://github.com/o3de/o3de-multiplayersample) on window, tested with:  
  - Editor : View starbase level in edit mode, launched game mode within the editor
  - MultiplayerSample.HeadlessServerLauncher + MultiplayerSample.GameLauncher : Launched `MultiplayerSample.HeadlessServerLauncher --console-command-file=launch_server.cfg` and then launched `MultiplayerSample.GameLauncher --console-command-file=launch_client.cfg`. Confirmed headless server console running, 
- **Android** Built AutomatedTesting and launched on Android using the `DefaultLevel`


Code/Framework/AzCore/AzCore/Component/ComponentApplicationBus.h
```diff
@@ -42,14 +42,16 @@ namespace AZ
         bool IsEditor() const;
         bool IsTool() const;
         bool IsGame() const;
+        bool IsHeadless() const;
         bool IsValid() const;
-
+
```
AMZN-alexpete, line 47:
```suggestion

```

Code/Editor/CMakeLists.txt
```diff
@@ -106,6 +106,8 @@ ly_add_target(
             Legacy::CryCommon
             Legacy::EditorCommon
             AZ::AzCore
+            AZ::AzFramework
+            AZ::AzFramework.NativeUI
```
lumberyard-employee-dm, line 110: AzToolsFramewok brings in the AZ::Framework dependency already

Code/Framework/AzFramework/AzFramework/Application/Application.cpp
```diff
@@ -213,7 +213,9 @@ namespace AzFramework
 
     void Application::StartCommon(AZ::Entity* systemEntity)
     {
-        m_pimpl.reset(Implementation::Create());
+        
+        auto implementationFactory = AZ::Interface<Application::ImplementationFactory>::Get();
+        m_pimpl = (implementationFactory != nullptr) ? implementationFactory->Create() : nullptr;
```
lumberyard-employee-dm, line 218: super nitpick: extra line here.
```suggestion
        auto implementationFactory = AZ::Interface<Application::ImplementationFactory>::Get();
        m_pimpl = (implementationFactory != nullptr) ? implementationFactory->Create() : nullptr;
```

Code/Framework/AzFramework/AzFramework/Components/NativeUISystemComponent.h
```diff
@@ -0,0 +1,67 @@
...
+        void InitializeNativeWindowImplementationFactory();
+
+        AZStd::unique_ptr<Application::ImplementationFactory> m_applicationImplFactory {};
+        AZStd::unique_ptr<InputDeviceGamepad::ImplementationFactory> m_deviceGamepadImplFactory {};
+        AZStd::unique_ptr<InputDeviceKeyboard::ImplementationFactory> m_deviceKeyboardImplFactory {};
+        AZStd::unique_ptr<InputDeviceMotion::ImplementationFactory> m_deviceMotionImplFactory {};
+        AZStd::unique_ptr<InputDeviceMouse::ImplementationFactory> m_deviceMouseImplFactory {};
+        AZStd::unique_ptr<InputDeviceTouch::ImplementationFactory> m_deviceTouchImplFactory {};
+        AZStd::unique_ptr<InputDeviceVirtualKeyboard::ImplementationFactory> m_deviceVirtualKeyboardImplFactory {};
+        AZStd::unique_ptr<NativeWindow::ImplementationFactory> m_nativeWindowImplFactory {};
```
lumberyard-employee-dm, line 65: The curly braces aren't needed as a unique_ptr initializes to nullptr by default.

Code/Framework/AzFramework/Platform/Mac/AzFramework/Input/Devices/Gamepad/InputDeviceGamepad_Mac.mm
```diff
@@ -0,0 +1,334 @@
...
+            { DIGITAL_BUTTON_MASK_R3,       &InputDeviceGamepad::Button::R3 },
+
+            { DIGITAL_BUTTON_MASK_A,        &InputDeviceGamepad::Button::A },
+            { DIGITAL_BUTTON_MASK_B,        &InputDeviceGamepad::Button::B },
+            { DIGITAL_BUTTON_MASK_X,        &InputDeviceGamepad::Button::X },
+            { DIGITAL_BUTTON_MASK_Y,        &InputDeviceGamepad::Button::Y },
+
+            { DIGITAL_BUTTON_MASK_PAUSE,    &InputDeviceGamepad::Button::Start },
+            { DIGITAL_BUTTON_MASK_SELECT,   &InputDeviceGamepad::Button::Select }
+        };
```
lumberyard-employee-dm, line 64: The using namespace AzFramework isn't needed here
```suggestion
    ////////////////////////////////////////////////////////////////////////////////////////////////
    //! Digital button bitmasks
    const AZ::u32 DIGITAL_BUTTON_MASK_DU    = 0x0001;
    const AZ::u32 DIGITAL_BUTTON_MASK_DD    = 0x0002;
    const AZ::u32 DIGITAL_BUTTON_MASK_DL    = 0x0004;
    const AZ::u32 DIGITAL_BUTTON_MASK_DR    = 0x0008;

    const AZ::u32 DIGITAL_BUTTON_MASK_L1    = 0x0010;
    const AZ::u32 DIGITAL_BUTTON_MASK_R1    = 0x0020;
    const AZ::u32 DIGITAL_BUTTON_MASK_L3    = 0x0040;
    const AZ::u32 DIGITAL_BUTTON_MASK_R3    = 0x0080;

    const AZ::u32 DIGITAL_BUTTON_MASK_A     = 0x0100;
    const AZ::u32 DIGITAL_BUTTON_MASK_B     = 0x0200;
    const AZ::u32 DIGITAL_BUTTON_MASK_X     = 0x0400;
    const AZ::u32 DIGITAL_BUTTON_MASK_Y     = 0x0800;

    const AZ::u32 DIGITAL_BUTTON_MASK_PAUSE = 0x1000;
    const AZ::u32 DIGITAL_BUTTON_MASK_SELECT = 0x2000;

    ////////////////////////////////////////////////////////////////////////////////////////////////
    //! Map of digital button ids keyed by their button bitmask
    const AZStd::unordered_map<AZ::u32, const AzFramework::InputChannelId*> GetDigitalButtonIdByBitMaskMap()
    {
        const AZStd::unordered_map<AZ::u32, const AzFramework::InputChannelId*> map =
        {
            { DIGITAL_BUTTON_MASK_DU,       &AzFramework::InputDeviceGamepad::Button::DU },
            { DIGITAL_BUTTON_MASK_DD,       &AzFramework::InputDeviceGamepad::Button::DD },
            { DIGITAL_BUTTON_MASK_DL,       &AzFramework::InputDeviceGamepad::Button::DL },
            { DIGITAL_BUTTON_MASK_DR,       &AzFramework::InputDeviceGamepad::Button::DR },

            { DIGITAL_BUTTON_MASK_L1,       &AzFramework::InputDeviceGamepad::Button::L1 },
            { DIGITAL_BUTTON_MASK_R1,       &AzFramework::InputDeviceGamepad::Button::R1 },
            { DIGITAL_BUTTON_MASK_L3,       &AzFramework::InputDeviceGamepad::Button::L3 },
            { DIGITAL_BUTTON_MASK_R3,       &AzFramework::InputDeviceGamepad::Button::R3 },

            { DIGITAL_BUTTON_MASK_A,        &AzFramework::InputDeviceGamepad::Button::A },
            { DIGITAL_BUTTON_MASK_B,        &AzFramework::InputDeviceGamepad::Button::B },
            { DIGITAL_BUTTON_MASK_X,        &AzFramework::InputDeviceGamepad::Button::X },
            { DIGITAL_BUTTON_MASK_Y,        &AzFramework::InputDeviceGamepad::Button::Y },

            { DIGITAL_BUTTON_MASK_PAUSE,    &AzFramework::InputDeviceGamepad::Button::Start },
            { DIGITAL_BUTTON_MASK_SELECT,   &AzFramework::InputDeviceGamepad::Button::Select }
        };
```

Code/Framework/AzFramework/Platform/Mac/AzFramework/Input/Devices/Gamepad/InputDeviceGamepad_Mac.mm
```diff
@@ -0,0 +1,334 @@
...
+    ////////////////////////////////////////////////////////////////////////////////////////////////
+    void InputDeviceGamepadMac::SetVibration(float leftMotorSpeedNormalized,
+                                               float rightMotorSpeedNormalized)
+    {
+        // The apple game controller framework does not (yet?) support force-feedback
+    }
+
+    ////////////////////////////////////////////////////////////////////////////////////////////////
+    bool InputDeviceGamepadMac::GetPhysicalKeyOrButtonText(const InputChannelId& inputChannelId,
+                                                             AZStd::string& o_keyOrButtonText) const
```
lumberyard-employee-dm, line 165: The spacing of the parameters are off in the function declarations
```suggestion
    void InputDeviceGamepadMac::SetVibration(float leftMotorSpeedNormalized,
                                             float rightMotorSpeedNormalized)
    {
        // The apple game controller framework does not (yet?) support force-feedback
    }

    ////////////////////////////////////////////////////////////////////////////////////////////////
    bool InputDeviceGamepadMac::GetPhysicalKeyOrButtonText(const InputChannelId& inputChannelId,
                                                           AZStd::string& o_keyOrButtonText) const
```

Code/Framework/AzFramework/Platform/Windows/AzFramework/Application/Application_Windows.h
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#pragma once
+
+#include <AzFramework\Application\Application.h>
+#include <AzFramework\API\ApplicationAPI_Windows.h>
+#include <AzCore\Memory\SystemAllocator.h>
```
lumberyard-employee-dm, line 13: This must be changed to forward slashes
```suggestion
#include <AzFramework/Application/Application.h>
#include <AzFramework/API/ApplicationAPI_Windows.h>
#include <AzCore/Memory/SystemAllocator.h>
```

lumberyard-employee-dm (changes requested):

Code/Framework/AzGameFramework/AzGameFramework/Application/GameApplication.cpp
```diff
@@ -57,6 +62,26 @@ namespace AzGameFramework
...
+    {
+        AZ::ComponentTypeList components = AzFramework::Application::GetRequiredSystemComponents();
+
+        #if !O3DE_HEADLESS_SERVER
+        components.insert(
+            components.end(),
+            {
+                azrtti_typeid<AzFramework::NativeUISystemComponent>(),
+            });
+        #endif // O3DE_HEADLESS_SERVER
```
lumberyard-employee-dm, line 75: Why is this `#if !O3DE_HEADLESS_SERVER` indented, but the one below is not?

Code/Framework/AzFramework/Platform/iOS/AzFramework/Components/NativeUISystemComponentFactories_iOS.h
```diff
@@ -0,0 +1,63 @@
...
+    public:
+        AZStd::unique_ptr<InputDeviceVirtualKeyboard::Implementation> Create(InputDeviceVirtualKeyboard& inputDevice) override;
+    };
+
+    class IosNativeWindowFactory 
+        : public NativeWindow::ImplementationFactory
+    {
+    public:
+        AZStd::unique_ptr<NativeWindow::Implementation> Create() override;
+    };
```
lumberyard-employee-dm, line 61: These factory classes can be defined in the cpp file.

Code/Framework/AzGameFramework/AzGameFramework/CMakeLists.txt
```diff
@@ -14,6 +14,24 @@ ly_add_target(
...
+            AZ::AzCore
+            AZ::AzFramework
+            AZ::AzFramework.NativeUI
+)
+
+ly_add_target(
+    NAME AzGameFramework.Headless STATIC
+    NAMESPACE AZ
+    FILES_CMAKE
+        azgameframework_files.cmake
```
lumberyard-employee-dm, line 28: Can you add a comment that the AzGameFramework.Headless static library is using the same source files as the AzGameFramework target, so it is compiled twice, with the difference of there being a `O3DE_HEADLESS_SERVER` define?

Also have looked into ways at reducing this duplicate compilation, like with do in the launcher, by only having the files that need the define into it's own target (https://github.com/o3de/o3de/blob/development/Code/LauncherUnified/CMakeLists.txt#L51-L54)

Code/Framework/AzFramework/Platform/Android/AzFramework/Components/NativeUISystemComponentFactories_Android.h
```diff
@@ -0,0 +1,70 @@
...
+    public:
+        AZStd::unique_ptr<InputDeviceVirtualKeyboard::Implementation> Create(InputDeviceVirtualKeyboard& inputDevice) override;
+    };
+
+    class AndroidNativeWindowFactory 
+        : public NativeWindow::ImplementationFactory
+    {
+    public:
+        AZStd::unique_ptr<NativeWindow::Implementation> Create() override;
+    };
```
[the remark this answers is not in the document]
spham-amzn: They were placed in a header file on purpose because the implementation of the ::Create functions are in the pimpl class functions. 
For example in [InputDeviceKeyboard_Android](https://github.com/aws-lumberyard-dev/o3de/blob/f4212b1bf9296f08b7ae59c996d11dd3c2171cf9/Code/Framework/AzFramework/Platform/Android/AzFramework/Input/Devices/Keyboard/InputDeviceKeyboard_Android.cpp#L438-L441)

```
 AZStd::unique_ptr<InputDeviceKeyboard::Implementation> AndroidDeviceKeyboardImplFactory::Create(InputDeviceKeyboard& inputDevice)
    {
        return AZStd::make_unique<InputDeviceKeyboardAndroid>(inputDevice);
    }
```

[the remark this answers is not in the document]
lumberyard-employee-dm: But for Linux and Windows, you placed the implementation in their cpp files, so it is inconsistent with those platforms.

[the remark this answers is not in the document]
spham-amzn: I was able to do it for Linux because there was another level of abstraction using Xcb, so most of the device specific implementation was wrapped in other headers anyways. For windows, it was inverted where the pimpl classes had header files and was included in the component class. 
I could remove the header files if i move the PAL platform implementation of in the pimpl classes, but that would mean i need to keep the base ones for non-supported devices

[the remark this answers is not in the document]
lumberyard-employee-dm: Alright this is fine then, I am not going to block the PR over moving the Android, MacOS and iOS code out of the header.

lumberyard-employee-dm (approved):

Code/Framework/AzFramework/Platform/Android/AzFramework/Components/NativeUISystemComponentFactories_Android.h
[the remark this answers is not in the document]
spham-amzn: Its not a bad idea, i'm going to make that pattern consistent

lumberyard-employee-dm (approved):
Good with me.

Code/Framework/AzFramework/Platform/iOS/AzFramework/Input/Devices/Gamepad/InputDeviceGamepad_iOS.mm
```diff
@@ -112,50 +113,37 @@ bool GetPhysicalKeyOrButtonText(const InputChannelId& inputChannelId,
...
         , m_rawGamepadState(GetDigitalButtonIdByBitMaskMap())
         , m_controller(nullptr)
         , m_wasPausedHandlerCalled(false)
     {
-        AZ_Assert(inputDevice.GetInputDeviceId().GetIndex() < InputDeviceGamepad::GetMaxSupportedGamepads(),
-                  "Creating InputDeviceGamepadApple with index %d that is greater than the max supported by the game controller framework: %d",
-                  inputDevice.GetInputDeviceId().GetIndex(), InputDeviceGamepad::GetMaxSupportedGamepads());
+        AZ_Assert(inputDevice.GetInputDeviceId().GetIndex() <= GCControllerPlayerIndex4,
+                  "Creating InputDeviceGamepadiOS with index %d that is greater than the max supported by the game controller framework: %d",
+                  inputDevice.GetInputDeviceId().GetIndex(), (GCControllerPlayerIndex4+1));
```
mbalfour-amzn, line 124: Nit: If you can't still use GetMaxSupportedGamepads() here, it would be nice if you put "GCControllerPlayerIndex4+1" into a constant somewhere that this code and GetMaxSupportedGamepads() could both use.

Gems/Atom/RHI/DX12/CMakeLists.txt
```diff
@@ -13,3 +13,5 @@ o3de_restricted_path(${gem_json} gem_restricted_path gem_parent_relative_path)
 ly_add_external_target_path(${CMAKE_CURRENT_LIST_DIR}/3rdParty)
 
 add_subdirectory(Code)
+
+ly_create_alias(NAME Atom_RHI_DX12.HeadlessServers NAMESPACE Gem)
```
mbalfour-amzn, line 17: Just curious - why do we need this? Wouldn't a headless server always use a null RHI?

Gems/Atom/RPI/Code/Source/RPI.Private/RPISystemComponent.cpp
```diff
@@ -100,24 +100,36 @@ namespace AZ
...
-            if (commandLine)
+            AZ::ApplicationTypeQuery appType;
+            ComponentApplicationBus::Broadcast(&AZ::ComponentApplicationBus::Events::QueryApplicationType, appType);
+            
+            bool isNullRenderer = false;
+            if (appType.IsHeadless())
+            {
+                // if the application is `headless`, merge `NullRenderer` attribute to the setting registry
+                isNullRenderer = true;
+            }
```
mbalfour-amzn, line 113: Shouldn't this also set -rhi=null somewhere?

mbalfour-amzn (approved):
One final really dumb question - why does this need a compile-time flag (O3DE_HEADLESS_SERVER) vs just switching out the factory implementation at runtime based on the -rhi=null / -NullRenderer flags? Is it important to not include the windowing libraries at _all_ into the executable or something like that?

Gems/Atom/RHI/DX12/CMakeLists.txt
spham-amzn: It ended up during my testing that this was needed because monolithic builds do not dynamically load the Atom RHIs, they are all linked together. So this wasn't necessary for profile builds, but monolithic builds were pulling in Xcb* dependencies (see response to the other question below)

Code/Framework/AzFramework/AzFramework/AzFrameworkNativeUIModule.cpp
```diff
@@ -0,0 +1,40 @@
...
+    {
+        AZ::ComponentTypeList components{};
+
+        #if !O3DE_HEADLESS_SERVER
+        components.insert(
+            components.end(),
+            {
+                azrtti_typeid<AzFramework::NativeUISystemComponent>(),
+            });
+        #endif // O3DE_HEADLESS_SERVER
```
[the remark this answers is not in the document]
spham-amzn: Probably not (good catch)

Gems/Atom/RPI/Code/Source/RPI.Private/RPISystemComponent.cpp
spham-amzn: Not really needed, this supercedes it

mbalfour-amzn: How so? From everything I've seen, they're independent switches. -NullRenderer controls the RPI layer only, and -rhi=null controls the RHI layer only. To turn off rendering completely, you need to pass both. Otherwise, I believe the RHI level will still try to initialize with DX12, Vulkan, etc even if no RPI features then use it.

spham-amzn: sorry, my mistake, you are right, this supercedes the block immediately below it
```
                // Otherwise if the command line contains -NullRenderer merge it to setting registry
                const char* nullRendererOption = "NullRenderer"; // command line option name
                const AzFramework::CommandLine* commandLine = nullptr;
                AzFramework::ApplicationRequests::Bus::BroadcastResult(commandLine, &AzFramework::ApplicationRequests::GetApplicationCommandLine);
                if (commandLine->GetNumSwitchValues(nullRendererOption) > 0)
                {
                    isNullRenderer = true;
                }
                // Otherwise if the command line contains -NullRenderer merge it to setting registry
                const char* nullRendererOption = "NullRenderer"; // command line option name
                const AzFramework::CommandLine* commandLine = nullptr;
                AzFramework::ApplicationRequests::Bus::BroadcastResult(commandLine, &AzFramework::ApplicationRequests::GetApplicationCommandLine);
                if (commandLine->GetNumSwitchValues(nullRendererOption) > 0)
                {
                    isNullRenderer = true;
                }
```
The the `-rhi=null` isn't really necessary because IIRC, making the RHI aliases for Vulkan, DX12, Metal for the new `HeadlessServer` cmake variants pointing to nothing, it will actually exclude the dependencies on those RHI modules and they won't launch (when building headless server launcher).

Code/Framework/AzFramework/AzFramework/Input/Devices/Gamepad/InputDeviceGamepad.cpp
```diff
@@ -144,7 +144,7 @@ namespace AzFramework
         }
 
         // Create the platform specific or custom implementation
-        m_pimpl.reset(implementationFactory ? implementationFactory(*this) : nullptr);
+        m_pimpl = (implementationFactory != nullptr) ? implementationFactory->Create(*this) : nullptr;
```
AMZN-alexpete, line 147: likely not an issue here, because this should only be called once, but I believe `reset` would actually delete any existing implementation as part of a swap.

Code/Framework/AzFramework/Platform/Mac/platform_nativeui_mac.cmake
```diff
@@ -0,0 +1,22 @@
...
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+#
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+#
+
+find_library(APPKIT_LIBRARY AppKit)
+find_library(GAME_CONTROLLER_LIBRARY GameController)
+find_library(CARBON_LIBRARY Carbon)
```
AMZN-alexpete, line 11: Do we use this? if not we should probably remove it if it's the same as `Carbon Core` which is deprecated https://developer.apple.com/documentation/coreservices/carbon_core?language=objc
Can be a future PR.

Code/Framework/AzFramework/Platform/iOS/AzFramework/Input/Devices/Gamepad/InputDeviceGamepad_iOS.mm
```diff
@@ -68,23 +67,28 @@
...
 {
+    static constexpr int GetMaxIosSupportedGamepads()
+    {
+        return static_cast<int>(GCControllerPlayerIndex4) + 1;
+    }
+
     ////////////////////////////////////////////////////////////////////////////////////////////////
     //! Platform specific implementation for apple game-pad input devices
-    class InputDeviceGamepadApple : public InputDeviceGamepad::Implementation
+    class InputDeviceGamepadiOS : public InputDeviceGamepad::Implementation
```
AMZN-alexpete, line 77: Not sure I understand the renaming of GamepadApple -> GamepadiOS? Are gamepads really supported on iOS? I would think this would be for gamepads connected to Macs?

Gems/AtomLyIntegration/AtomBridge/Code/CMakeLists.txt
```diff
@@ -63,10 +63,46 @@ ly_add_target(
...
+        PRIVATE
+            Gem::Atom_AtomBridge.Static
+    RUNTIME_DEPENDENCIES
+        Gem::Atom_RHI.Private
+        Gem::Atom_RPI.Private
+        Gem::Atom_RHI_Null.Private
+        Gem::Atom_Feature_Common.Headless
+        Gem::Atom_Bootstrap
+        Gem::Atom_Component_DebugCamera
+        Gem::AtomImGuiTools
```
AMZN-alexpete, line 90: Can this be removed from headless since IMGUI is GUI related?

Gems/AtomLyIntegration/AtomBridge/Code/CMakeLists.txt
```diff
@@ -63,10 +63,46 @@ ly_add_target(
...
+        Gem::AtomLyIntegration_CommonFeatures
+        Gem::EMotionFX_Atom
+        Gem::ImguiAtom
+        Gem::AtomFont
```
AMZN-alexpete, line 94: Can this be removed from headless since IMGUI and AtomFont are GUI related?

Gems/AtomLyIntegration/AtomBridge/Code/CMakeLists.txt
```diff
@@ -63,10 +63,46 @@ ly_add_target(
...
+        Gem::AtomViewportDisplayInfo
+)
+
+if (PAL_TRAIT_BUILD_SERVER_SUPPORTED)
+
+endif()
```
AMZN-alexpete, line 100:
```suggestion
```

cmake/Gems.cmake
```diff
@@ -363,19 +365,28 @@ function(ly_add_gem_dependencies_to_project_variants)
 
     # apply the list of gem targets.  Adding a gem really just means adding the appropriate dependency.
     foreach(gem_name ${ly_add_gem_dependencies_GEM_DEPENDENCIES})
+
+        # Construct the gem target name to determine if the target exists
         set(gem_target ${gem_name}.${ly_add_gem_dependencies_VARIANT})
 
-        # if the target exists, add it.
+        # Also construct a fall if the target doesnt exist for the given variant, then check if the variant has a fallback
```
AMZN-alexpete, line 372:
```suggestion
        # Also construct a fallback target if the target doesn't exist for the given variant, then check if the variant has a fallback
```

AMZN-alexpete (approved):
Just some minor comments/questions, but over-all good with this.

Code/Framework/AzFramework/AzFramework/Input/Devices/Gamepad/InputDeviceGamepad.cpp
spham-amzn: That would actually be good since the pimp is meant to be singleton. But there isn't any path (currently) that would make this code execute more than once.

Code/Framework/AzFramework/Platform/iOS/AzFramework/Input/Devices/Gamepad/InputDeviceGamepad_iOS.mm
spham-amzn: I'm not sure, this use to be a shared file between both iOS and Mac, and I don't have the context on why gamepad is on iOS. I removed the sharing of files.

Gems/AtomLyIntegration/AtomBridge/Code/CMakeLists.txt
spham-amzn: This was to minimize the changes needed to have a headless version of atom bridge. Although i havent tested, it may need a lot of #if defs to the code to keep this module compilable, which i wanted to avoid. The need to make this headless was purely to not pull in XCB/X11 dependencies that the original AzFramework library did.

Gems/AtomLyIntegration/AtomBridge/Code/CMakeLists.txt
spham-amzn: same reason as above.

Code/Framework/AzFramework/AzFramework/Input/Devices/Gamepad/InputDeviceGamepad.cpp
spham-amzn: Actually, in this case, i think `reset` and `=` will have the same effect.
https://en.cppreference.com/w/cpp/memory/unique_ptr/operator%3D 
```
3) Effectively the same as calling reset().
```

Code/Framework/AzFramework/Platform/Mac/platform_nativeui_mac.cmake
spham-amzn: Yeah, this was a move to a new file, and removing it should be a different PR
Created a GHI for this: https://github.com/o3de/o3de/issues/16644

PR #16457, opened 2023-07-26
Added feature to prompt user on container clear for DPE

### LINKED ISSUE #16437: DPE Inspector: PostFX Layer component does not throw a warning when deleting all Camera Tags
**Describe the bug**
Attempting to remove all elements from either Select Camera Tags Only or Excluded Camera Tags containers should result in a warning pop-up prompting the user to confirm the removal. The pop-up is thrown properly while DPE is disabled, but when DPE is enabled the pop-up does not appear and all elements are removed.
Issue occurs on both Windows and Linux.

**Steps to reproduce**
1. Create an Entity and add the PostFX Layer component to it.
2. Make sure DPE is disabled (enter `ed_enableDPE 0` in the Console).
3. Add empty elements to either Select Camera Tags Only or Excluded Camera Tags containers.
4. Attempt to remove all elements by pressing the bin icon next to Select Camera Tags Only or Excluded Camera Tags.
5. Enable DPE (enter `ed_enableDPE 1` in the Console).
6. Add empty elements to either Select Camera Tags Only or Excluded Camera Tags containers.
7. Attempt to remove all elements by pressing the bin icon next to Select Camera Tags Only or Excluded Camera Tags.

**Expected behavior**
After step 4 and 7 a warning appears prompting the user to confirm removal of all elements.

**Actual behavior**
After step 7 no warning appears and all elements are removed without confirmation from the user.

**Screenshots/Video**

https://github.com/o3de/o3de/assets/87059746/d0c8968c-743c-428a-ba5f-31362a2b389a

**Found in Branch**
Development

[3210473](https://github.com/o3de/o3de/commit/321047321e3b6b22d842db8fec9c4f2d4ed646b8)

**Desktop/Device:**
- Device: PC
- OS: Linux
- Version: Ubuntu 22.04.2 LTS
- CPU: Ryzen 5600H
- GPU: NVidia RTX 3070
- Memory: 16GB

cgalvan:
## What does this PR do?

Fixes #16437 

This adds the feature for prompting the user when clearing a container in the DPE. I've added an improvement over the RPE, where previously this prompt was always displayed, but now in the DPE you can op-out of it for your container by setting the `PromptOnContainerClear` on your container element to `false`.

ClearContainerPrompt

## How was this PR tested?

Tested all container buttons to make sure they all work as expected. Verified that the prompt only shows when attempting to clear at the container level, not just deleting a single element.

### DIFF
--- a/Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/ContainerActionButtonHandler.cpp
+++ b/Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/ContainerActionButtonHandler.cpp
@@ -8,6 +8,8 @@
 
 #include <AzToolsFramework/UI/DocumentPropertyEditor/ContainerActionButtonHandler.h>
 
+#include <QMessageBox>
+
 namespace AzToolsFramework
 {
     ContainerActionButtonHandler::ContainerActionButtonHandler()
@@ -25,8 +27,8 @@ namespace AzToolsFramework
         using AZ::DocumentPropertyEditor::Nodes::ContainerAction;
         using AZ::DocumentPropertyEditor::Nodes::ContainerActionButton;
 
-        ContainerAction action = ContainerActionButton::Action.ExtractFromDomNode(node).value_or(ContainerAction::AddElement);
-        switch (action)
+        m_action = ContainerActionButton::Action.ExtractFromDomNode(node).value_or(ContainerAction::AddElement);
+        switch (m_action)
         {
         case ContainerAction::AddElement:
             setIcon(s_iconAdd);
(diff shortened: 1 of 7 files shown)


Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/ContainerActionButtonHandler.cpp
```diff
@@ -42,4 +44,30 @@ namespace AzToolsFramework
...
+        using AZ::DocumentPropertyEditor::Nodes::ContainerAction;
+
+        if (m_action == ContainerAction::Clear)
+        {
+            // Default to true if the PromptOnContainerClear attribute isn't present
+            bool shouldPromptOnClear = Container::PromptOnContainerClear.ExtractFromDomNode(m_node).value_or(true);
+
+            if (shouldPromptOnClear)
+            {
+                auto result = QMessageBox::question(
```
alexmontAmazon, line 60: don't we have a styled version of this that we're supposed to use now? I vaguely recall LY having a QMessageBox replacement... ignore me, if you don't know what I'm talking about...

alexmontAmazon (approved):

Code/Framework/AzToolsFramework/AzToolsFramework/UI/DocumentPropertyEditor/ContainerActionButtonHandler.cpp
cgalvan: We don't have AZ versions of the `QMessageBox`, because we already override their style in the StyleManager, so we get the styled version for free without having to change all the API calls. You might be thinking of the `AzMessageBox` variants we have, but they aren't a separate class they just let you specify a setreg first so that you can get the behavior of "Do not show this message again" checkbox functionality.

PR #16458, opened 2023-07-26
Fix Vulkan validation errors on Android

akioCL:
## What does this PR do?

Remove PVRTC format for Vulkan. PVRTC support is provided by the VK_IMG_format_pvrtc extension, which has been deprecated. Querying information about this format triggers validation errors on Android. We are not currently using PVRTC images on Android, only ASTC for compressed images.
Also enable Int16 support for shaders on Vulkan.

## How was this PR tested?

Run Vulkan on Android

### DIFF
--- a/Gems/Atom/RHI/Vulkan/Code/Source/RHI.Reflect/Formats.inl
+++ b/Gems/Atom/RHI/Vulkan/Code/Source/RHI.Reflect/Formats.inl
@@ -99,10 +99,6 @@
         _Func(ETC2A1_UNORM_SRGB,       VK_FORMAT_ETC2_R8G8B8A1_SRGB_BLOCK,      1,    0,    0) \
         _Func(ETC2A_UNORM,             VK_FORMAT_ETC2_R8G8B8A8_UNORM_BLOCK,     1,    0,    0) \
         _Func(ETC2A_UNORM_SRGB,        VK_FORMAT_ETC2_R8G8B8A8_SRGB_BLOCK,      1,    0,    0) \
-        _Func(PVRTC2_UNORM,            VK_FORMAT_PVRTC2_2BPP_UNORM_BLOCK_IMG,   1,    0,    0) \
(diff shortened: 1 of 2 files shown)


Gems/Atom/RHI/Vulkan/Code/Source/RHI.Reflect/Formats.inl
```diff
@@ -99,10 +99,6 @@
         _Func(ETC2A1_UNORM_SRGB,       VK_FORMAT_ETC2_R8G8B8A1_SRGB_BLOCK,      1,    0,    0) \
         _Func(ETC2A_UNORM,             VK_FORMAT_ETC2_R8G8B8A8_UNORM_BLOCK,     1,    0,    0) \
         _Func(ETC2A_UNORM_SRGB,        VK_FORMAT_ETC2_R8G8B8A8_SRGB_BLOCK,      1,    0,    0) \
-        _Func(PVRTC2_UNORM,            VK_FORMAT_PVRTC2_2BPP_UNORM_BLOCK_IMG,   1,    0,    0) \
-        _Func(PVRTC2_UNORM_SRGB,       VK_FORMAT_PVRTC2_2BPP_SRGB_BLOCK_IMG,    1,    0,    0) \
-        _Func(PVRTC4_UNORM,            VK_FORMAT_PVRTC2_4BPP_UNORM_BLOCK_IMG,   1,    0,    0) \
-        _Func(PVRTC4_UNORM_SRGB,       VK_FORMAT_PVRTC2_4BPP_SRGB_BLOCK_IMG,    1,    0,    0) \
```
AMZN-alexpete, line 105: It might be a good idea to leave a comment in here indicating PVRTC formats are deprecated and have been removed because they cause validation errors in case some developer tries to add them back

AMZN-alexpete (approved):

Gems/Atom/RHI/Vulkan/Code/Source/RHI.Reflect/Formats.inl
VickyAtAZ: We shouldn't need to. There is no PVRTC format support in image builder.

VickyAtAZ (approved):

moudgils (approved):

PR #16465, opened 2023-07-27
Added Volume(3D) Texture support to StreamingImage.

galibzon:
## What does this PR do?

In addition to Width(x) and Height(y), the IImageObject interface now has awareness of Depth(z). This means the interface now supports Volume Textures, including mip levels. Consequently, relevant updates were made to CImageObject.

In terms of Asset Processing, only the `DdsLoader` was added the capability of loading Volume Textures.

Unlike Image2D, Volume Textures do not support
automatic generation of Mip Levels during asset processing. The Volume Texture is assumed to contain all the mips and the output streaming mip chain assets will contain the pixels for all Mip Levels.

The runtime can now load a StreaminImage from a dds.streamingimage asset that contains 3D Texture data.

## How was this PR tested?

First, it was validated that the existing 2D Image functionality is not broken. All ASV samples work as expected.
Regarding loading StreamingImage resoruces from a Volume Texture, it was validated with my Volumetric Cloudscape Gem.

### DIFF
--- a/Gems/Atom/Asset/ImageProcessingAtom/Code/Source/Processing/ImageAssetProducer.cpp
+++ b/Gems/Atom/Asset/ImageProcessingAtom/Code/Source/Processing/ImageAssetProducer.cpp
@@ -93,6 +93,14 @@ namespace ImageProcessingAtom
         RPI::StreamingImageAssetCreator builder;
         builder.Begin(imageAssetId);
 
+        uint32_t imageDepth = m_imageObject->GetDepth(0);
+        const bool isVolumeTexture = m_imageObject->HasImageFlags(EIF_Volumetexture) || (imageDepth > 1);
+        if (m_imageObject->HasImageFlags(EIF_Cubemap) && isVolumeTexture)
+        {
+            AZ_Assert(false, "An image can not be a cubemap and a volume texture at the same time!");
+            return false;
+        }
+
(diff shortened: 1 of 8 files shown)


VickyAtAZ (approved):

moudgils (approved):

Gems/Atom/Asset/ImageProcessingAtom/Code/Source/Processing/ImageAssetProducer.cpp
```diff
@@ -93,14 +93,24 @@ namespace ImageProcessingAtom
         RPI::StreamingImageAssetCreator builder;
         builder.Begin(imageAssetId);
 
+        uint32_t imageDepth = m_imageObject->GetDepth(0);
+        const bool isVolumeTexture = m_imageObject->HasImageFlags(EIF_Volumetexture) || (imageDepth > 1);
+        if (m_imageObject->HasImageFlags(EIF_Cubemap) && isVolumeTexture)
+        {
+            AZ_Assert(false, "An image can not be a cubemap and a volume texture at the same time!");
```
AMZN-alexpete, line 100: Should this be an AZ_Error so it is visible outside of debug builds?

AMZN-alexpete: Actually, looking at AZ_Assert and AZ_Error it looks like both should exist in a profile build, for some reason I thought asserts were not visible outside debug builds.

